PR #517, opened 2018-03-23
Fix Appveyor builds which fail during dependency installation

speth:
Attempting to fix failing Appveyor builds


bryanwweber:
Seems like this fixes that particular error. Not sure why not using the most recent version of pip is suddenly a build breaking error, but at least the fix is easy.

Issue #313, opened 2015-12-06
Add examples from "Defining Phases" to the documentation


bryanwweber:
This file: https://github.com/Cantera/cantera/blob/master/doc/sphinx/cti/example-combustion.rst is presently blank. Is it left over from an old version of the documentation?

As reported on the Google Group: https://groups.google.com/d/msg/cantera-users/h0s-vJ6iELY/RgqsyqGnCQAJ

speth:
It's left over from when I was converting the "Defining Phases" guide ([SourceForge download](http://sourceforge.net/projects/cantera/files/cantera%20documentation/Input%20Files/)) to RST. I never got around to adding the examples from that guide to the CTI guide (and it's not entirely clear to me that they belong there anyway).

bryanwweber:
OK I changed the title to reflect what could be done to resolve the issue.

kyleniemeyer:
This is a somewhat stale issue, but adding an example/some documentation here seems like something I can do...

sameshl:
I am new and would like to start contributing.Can I take up this issue?

bryanwweber:
Sure, see [the comment on the other PR](https://github.com/Cantera/cantera/pull/518#issuecomment-376164976) for where a good place to get started it.

speth:
Given the plans to deprecate the CTI format in the near future (#423, #584), I would hold off on this.

bryanwweber:
@speth Is the example on page 52 (55 of the PDF, copied below) already in the examples? It seems functionally the same as https://cantera.org/examples/matlab/reactor2.m.html

```matlab
gas = importPhase(’h2mech.cti’,’hydrogen_oxygen’);
nsp = nSpecies(gas);
% set the initial conditions
set(gas,’T’,1001.0,’P’,0.1*oneatm,’X’,’H2:2, O2:1, AR:4’); % create a reactor, and insert the gas
r = Reactor(gas);
t = 0;
dt = 3.0e-4;
tnow = 0;
m = 0;
for n = 1:15
   t = t + dt;
   while tnow < t
      tnow = step(r, t); % take one internal timestep m = m + 1;
      tim(m) = tnow;
      temp(m) = temperature(r);
       xm(:,m) = moleFractions(gas);
   end
end
% make plots
figure(1);
plot(tim, temp);
xlabel(’time’);
ylabel(’Temperature’);
figure(2);
plot(tim, xm);
xlabel(’time’);
ylabel(’Mole Fraction’);
[...]

PR #518, opened 2018-03-24
doc: hydrogen combustion example blank page removed

vdevgan:
- "hydrogen combustion reaction " page in doc was empty. 
that has been removed.


bryanwweber:
This is being tracked in #313, where the solution is to add the example rather than removing the page

vdevgan:
could I be shown another example in the doc or elsewhere to know in what specific format is the example to be filled?

bryanwweber:
The example in question is on pg. 55 of the PDF linked in #313. Other examples can be found in https://github.com/Cantera/cantera/tree/master/interfaces/cython/cantera/examples for Python and https://github.com/Cantera/cantera/tree/master/samples/matlab for Matlab. Note that the script in the PDF is a Matlab script, but we would prefer to have either Python or both Python and Matlab. You can also directly use the `h2o2.xml` file as the chemistry file, rather than a subset of GRI-3.0 as in the PDF.

speth:
I'm inclined to accept the solution of just removing the blank page, since it's not clear that those examples really belong as part of the CTI documentation anyway. The only change needed to this PR is that the entry in `index.rst` also needs to be removed. Issue #313 should remain open as a reminder that those examples have not been migrated.

vdevgan:
combustion example reference from cti/ index.rst has been removed.

PR #521, opened 2018-03-26
Add get_equivalence_ratio function

jsantner:
- Add function to get the equivalence ratio of a mixture. Assumes all species are "fuel" except for O2. Assumes C is oxidized to CO2, H to H2O, and S to SO2. All other elements are considered inert.


interfaces/cython/cantera/thermo.pyx
```diff
@@ -632,7 +632,45 @@ cdef class ThermoPhase(_SolutionBase):
...
-
+        
+    def get_equivalence_ratio(self):
+        """Considers the oxidation of C to CO2, H to H2O and S to SO2.
+        Other elements are assumed not to participate in oxidation
+        (i.e. N ends up as N2).
+        O2 is considered as the only oxidizer (i.e. NO is considered a "fuel")
+        """
+        alpha = 0
+        for species in range(self.n_species):
```
bryanwweber: In the Python interface, you can loop directly through the species list:

```python
for s in self.species():
    if s.composition == {'O': 2}:
        ...
    else:
        n_C = s.composition.get('C', 0)
        n_H = s.composition.get('H', 0)
        ....
        alpha += ...
```

bryanwweber (commented):
Aside from the style comment about the loop, I'd suggest adding a test that returns infinity, to check that is being done properly. That way, if this is refactored in the future, we don't have to worry about that case getting dropped by accident.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -632,7 +632,30 @@ cdef class ThermoPhase(_SolutionBase):
...
+                mol_O2 = self[s.name].X[0]
+            else:
+                nC = s.composition.get('C', 0)
+                nH = s.composition.get('H', 0)
+                nO = s.composition.get('O', 0)
+                nS = s.composition.get('S', 0)
+
+                alpha += (nC + nH/4 + nS - nO/2) * self[s.name].X[0]
 
+        if mol_O2 == 0:
```
bryanwweber: Hmm, sorry just noticed two more things here. One is that there may be cases where `mol_O2` is very close to zero, but not identically (`1E-15` or something), and this comparison would result in an equivalence ratio that is huge. I'm not totally sure what the best way to handle that situation is (`if mol_O2 < some_small_value`, or something else, maybe @speth has an idea here), but I don't think we should be doing floating point equality comparisons.

Second, in the case where there is no species with the composition `{'O': 2}` in the mechanism, this will throw a `NameError`, which should be handled somehow. Again, not sure of the best way to handle this...

interfaces/cython/cantera/thermo.pyx
```diff
@@ -632,7 +632,30 @@ cdef class ThermoPhase(_SolutionBase):
         stoichAirFuelRatio = - (Of - 2*Cf - 2*Sf - Hf/2.0) / (Oo - 2*Co - 2*So - Ho/2.0)
         Xr = phi * Xf + stoichAirFuelRatio * Xo
         self.TPX = None, None, Xr
+        
+    def get_equivalence_ratio(self):
+        """Considers the oxidation of C to CO2, H to H2O and S to SO2.
+        Other elements are assumed not to participate in oxidation
+        (i.e. N ends up as N2).
+        O2 is considered as the only oxidizer (i.e. NO is considered a "fuel")
```
bryanwweber: Sorry, and two more here. Can we add an argument to the function definition that allows the user to specify the oxidizer species, with the default being `'O2'` (or maybe better, whichever species has the composition `{'O': 2}`)? Would that complicate the logic awfully? Is it worth it to do that?

It would also be nice if this docstring was a little more complete (see the `set_equivalence_ratio` immediately above function for an example).

bryanwweber:
@jsantner Thanks for this! I forgot to mention before I think this will be a nice addition. I made a few more line comments on the code suggesting some changes. As I said, I'm not sure what the best way to implement the ideas are, so if you want to push code here we can iterate on it, or you can make comments to discuss if you'd rather.

interfaces/cython/cantera/thermo.pyx
jsantner: I think that if `mol_O2` is very close to zero, it's okay that this would result in a huge equivalence ratio. With `mol_O2 == 1e-15` , then it would be accurate to return a huge equivalence ratio. I only put the case with `mol_O2 == 0` to avoid a division by zero error.

If the mechanism doesn't have a species with the composition `{'O': 2}`, then that means there is no molecular oxygen, so the equivalence ratio should be `inf`, right? I can add that exception handling if you agree.

interfaces/cython/cantera/thermo.pyx
jsantner: Interesting idea. I could add a dict argument to specify the oxidizer, with the default being `{'O': 2}`. That would allow "non-traditional" oxidizers, like ozone, atomic oxygen, or NO.

jsantner: Now that I think about it more, maybe it's most accurate to have the oxidizer defined as any species that contains oxygen atoms that doesn't contain sulfur, carbon, or hydrogen. Then it would automatically include ozone, atomic oxygen, and NOx

speth:
So, the function introduced here can calculate the equivalence ratio for an unburned mixture. The question I have is whether we can think of the equivalence ratio as a quantity that is preserved as a mixture burns. If you restrict yourself to pure hydrocarbons, it is not hard to define in this way. For example, if we write

CnHm + 1/&phi; * (n + m/4) O2 -> products

Then the elemental mole fractions *Z* of each element are *Z_C = n*, *Z_H = m*, and *Z_O = 2/&phi; * (n + m/4)* and we can compute *&phi; = 2/Z_O * (Z_C + Z_H/4)*. This definition holds even after equilibrating a mixture, unlike the one defined in this PR:
```python
>>> import cantera as ct
>>> gas = ct.Solution('gri30.xml')
>>> gas.set_equivalence_ratio(1.8, 'CH4', 'O2:1.0, N2:3.76')
>>> print(2/gas.elemental_mole_fraction('O') * (gas.elemental_mole_fraction('C') + gas.elemental_mole_fraction('H')/4))
1.7999999999999998
>>> gas.equilibrate('HP')
>>> print(2/gas.elemental_mole_fraction('O') * (gas.elemental_mole_fraction('C') + gas.elemental_mole_fraction('H')/4))
1.799999999999997
>>> print(gas.get_equivalence_ratio())
350171124.406
```
One challenge with this is extending it to oxygenated fuels. In this case, you would need to provide a way of defining the C/O or H/O ratio in the fuel to compensate, perhaps by specifying a fuel composition. I think there is a similar issue if you consider CO2 or H2O to be present in the oxidizer.

jsantner:
@bryanwweber I added functionality so that the user can choose their oxidizer. The default behavior assigns every species that doesn't contain C, H, or S to be an oxidizer
@speth That's an interesting idea, but I'm not sure how useful it is. Usually, the equivalence ratio is defined by the fuel/oxidizer ratio. But, in the burned gas, I wouldn't consider a fuel to be present. As you mentioned, it also requires knowledge of the fuel composition when using oxygenated fuels. I saw this function as a tool to provide more information on a mixture without requiring knowledge of the mixture composition. Also, your method is simple enough that it may not warrant its own function.
I'm still new to GitHub. Am I doing this right? Bryan mentioned that I could "push code here we can iterate on it...". Am I doing this in a way that lets you edit the code, or am I only able to make changes?

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -262,6 +262,29 @@ def test_set_equivalence_ratio_sulfur(self):
...
+    def test_get_equivalence_ratio(self):
+        gas = ct.Solution('gri30.xml')
+        for phi in np.linspace(0.5, 2.0, 5):
+            gas.set_equivalence_ratio(phi, 'CH4:0.8, CH3OH:0.2', 'O2:1.0, N2:3.76')
+            self.assertNear(phi, gas.get_equivalence_ratio())  
+        # Check sulfur species
+        sulfur_species = [k for k in ct.Species.listFromFile('nasa_gas.xml') if k.name in ("SO", "SO2")]
+        gas = ct.Solution(thermo='IdealGas', kinetics='GasKinetics',
+                          species=ct.Species.listFromFile('gri30.xml') + sulfur_species,
+                          reactions=ct.Reaction.listFromFile('gri30.xml'))
```
speth: No need to add the reactions from GRI 3.0 for this test

interfaces/cython/cantera/thermo.pyx
```diff
@@ -632,7 +632,52 @@ cdef class ThermoPhase(_SolutionBase):
         stoichAirFuelRatio = - (Of - 2*Cf - 2*Sf - Hf/2.0) / (Oo - 2*Co - 2*So - Ho/2.0)
         Xr = phi * Xf + stoichAirFuelRatio * Xo
         self.TPX = None, None, Xr
+        
+    def get_equivalence_ratio(self, oxidizers=[], ignore=[]):
+        """Get the composition of a fuel/oxidizer mixture.
+        Considers the oxidation of C to CO2, H to H2O and S to SO2.
```
speth: Maybe the extra bit of clarity that would be useful to add here would be to say that this is the equivalence ratio of an _unburned_ fuel/oxidizer mixture, to distinguish this definition from one which is conserved after combustion.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -632,7 +632,52 @@ cdef class ThermoPhase(_SolutionBase):
...
+        >>> gas.set_equivalence_ratio(0.5, 'CH3:0.5, CH3OH:.5, N2:0.125', 'O2:0.21, N2:0.79, NO:0.01')
+        >>> gas.get_equivalence_ratio()
+        0.50000000000000011
+        >>> gas.get_equivalence_ratio(['O2'])  # Only consider O2 as the oxidizer instead of O2 and NO
+        0.48809523809523814
+        >>> gas.X = 'CH4:1, O2:2, NO:0.1'
+        >>> gas.get_equivalence_ratio(ignore=['NO'])
+        1.0
+        """
+        if oxidizers==[]:  # Default behavior, find all possible oxidizers
```
speth: better: `if not oxidizers:`

interfaces/cython/cantera/thermo.pyx
```diff
@@ -632,7 +632,52 @@ cdef class ThermoPhase(_SolutionBase):
...
+            oxidizers = [s.name for s in self.species() if
+                         all(y not in s.composition for y in ['C', 'H', 'S'])]
+        alpha = 0
+        mol_O = 0
+        for s in self.species():
+            if s.name in ignore:
+                pass
```
speth: I would prefer `continue` here, although the effect is the same.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -632,7 +632,52 @@ cdef class ThermoPhase(_SolutionBase):
...
+        >>> gas.X = 'CH4:1, O2:2, NO:0.1'
+        >>> gas.get_equivalence_ratio(ignore=['NO'])
+        1.0
+        """
+        if oxidizers==[]:  # Default behavior, find all possible oxidizers
+            oxidizers = [s.name for s in self.species() if
+                         all(y not in s.composition for y in ['C', 'H', 'S'])]
+        alpha = 0
+        mol_O = 0
+        for s in self.species():
```
speth: If you do `for k, s in enumerate(self.species()):` here, you can then just get the mole fractions as `self.X[k]` later, which is more efficient.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -632,7 +632,52 @@ cdef class ThermoPhase(_SolutionBase):
...
+        
+    def get_equivalence_ratio(self, oxidizers=[], ignore=[]):
+        """Get the composition of a fuel/oxidizer mixture.
+        Considers the oxidation of C to CO2, H to H2O and S to SO2.
+        Other elements are assumed not to participate in oxidation
+        (i.e. N ends up as N2)::
+        
+        :param oxidizers:
+            List of oxidizer species names as strings
+            Default: with oxid=[], every species that contains O but does not
```
speth: Should be "with oxidizers=[]"

speth (changes requested):
OK, I've thought about this a little bit more, and I think it's fine as long as there's a clarification that this definition is really only meaningful for an unburned mixture. I still like the idea of a definition that corresponds to a conserved quantity after combustion, but this implementation doesn't preclude doing that at some point in the future, e.g. by adding an additional keyword argument.

Your method of adding the commits is OK. We'll squash all of these commits into one when merging the PR.

speth (approved):

PR #527, opened 2018-04-16
Add cython func defn to expose species viscosity & add test

skyreflectedinmirrors:
Fixes (N/A)

Changes proposed in this pull request:
- Expose (read-only) species viscosity from transport to user in python interface 

The proposed use case here is that OpenFOAM currently doesn't implement a species transport model based on standard kinetic theory (i.e., involving Lennard-Jones, dipole moment, well-depth etc.).  Long term, it would be nice to build a Cantera interface for OpenFOAM (IIRC, this is one of the GSoC challenges?), but in the meantime this can be worked around as follows:

1. Use a [`sutherland` transport model](https://cfd.direct/openfoam/user-guide/thermophysical/#x35-2610007.1.2) in OpenFOAM.
2. Write a simple script using scipy `curve_fit` to determine best fit Sutherland coefficients based on Cantera

PR #529, opened 2018-04-21
Fix pip appveyor

bryanwweber:
- Fix errors on Appveyor because of outdated/removed options to pip

Should we use Python 3.6 for the builds and tests here? It's probably a good idea to keep testing with Python 3.5 as long as we support that, but I'm not sure we test with Python 3.6 anywhere else (except the conda packages, I guess)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,14 +1,14 @@
 version: 1.0.{build}
 install:
 - ps: |
-    C:\Python27-x64\python.exe -m pip install --upgrade pip
     C:\Python27-x64\Scripts\pip.exe install --egg scons
-    C:\Python27-x64\Scripts\pip.exe install numpy
+    C:\Python27-x64\python.exe -m pip install --no-cache-dir --upgrade pip
+    C:\Python27-x64\Scripts\pip.exe install --no-cache-dir numpy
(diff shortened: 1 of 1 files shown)


speth:
Oh, so SCons can finally be installed without the `--egg` option? Good. The last time I tried, that was not the case. A minor point, but I would suggest that the commit message note that this option is no longer *required*, rather than kind of implying that we were using a deprecated and now removed option for no apparent reason.

I don't think there's much difference in testing with Python 3.5 vs 3.6. Testing with an older version at least makes sure that we don't accidentally use features that were only introduced in Python 3.6. I know I do all my local development and testing using Python 3.6, so this might be the main source of testing for Python 3.5. I'd rather not expand the build matrix, just because that would slow down all the CI builds.

bryanwweber:
I think the change is that more recent versions of SCons have support for more recent versions of setuptools (or maybe require more recent versions of setuptools). Also, installing from the wheel (because the wheel package is installed) rather than the sdist may have helped. In any case, we would have to downgrade pip on Appveyor (or maybe just not upgrade pip) to keep using the `--egg` option, since on #527 it seems like the error is because the option was removed.

Anyways, this is squashed and should be good to merge now.

PR #535, opened 2018-05-12
[SCons] implement 'googletest' option

### LINKED ISSUE #515: Optional "googletest" package dependency to build Cantera
### Cantera version 
2.3.0 and 2.4.0b1(development)

### Operating System 
Linux

### Python/MATLAB version 
none

### Expected Behavior 
Successful build the Cantera package in case of without planing to run `scons test`

### Actual Behavior
config error: Googletest not found and submodule checkout failed.

### Steps to reproduce

1. Unpack cantera tarball.
2. Ensure that googletest package isn't installed in system (and maybe disable internet connection).
3. Run in terminal in cantera root working directory command `scons build`

Result: `config error: Googletest not found and submodule checkout failed.`

Issue Message:
As I understand "googletest" (gtest/gmock) package is needed only for performing "regression" code test and is unnecessary itself to build and use Cantera libraries, isn't it? 

Attached simple patches (one is for cantera-2.3.0 and another is for current "master"-branch) add "googletest" configuration option to disable/enable checking of googletest package presence in system and if it's enable (is set to "y") or isn't set ('default') then check "system_googletest" configuration option.

E.i. if call `scons build googletest='n'` then the absence of gtest/gmock will not give error message and cantera is successfully compile.

[cantera_is_googletest_patches.zip](https://github.com/Cantera/cantera/files/1811253/cantera_is_googletest_patches.zip)

Sorry for absence of pull request right now. I'm not good with git(hub) and that take some time to create PR and attache it to issue.

band-a-prend:
Fixes #515

## Changes proposed in this pull request:

-  New scons build option introduced to replace 'system_googletest'
and allow to not use 'googletest' module while running test
(all tests that require this module could be omited)

- New option description:
```
'googletest',
"""Select whether to use gtest/gmock from system
    installation ('system'), from a Git submodule ('submodule'), to decide
    automatically ('default') or don't look for gtest/gmock ('none')
    and don't run tests that depend on gtest/gmock.
    Selection of 'none' will suppress the depricated 'system_googletest' option."""
```

- Old option is still presented and mentioned as 'depricated'.
Option 'googletest = none' supress the old 'system_googletest'.

- In other cases if they is setted explicitly and contradict to eacher other - an ERROR message appears. 

P.S.
See [attached file 'scons_test_results.txt'](https://github.com/Cantera/cantera/files/1997371/scons_test_results.txt) to compare results of running 'scons test' command before and after implimentation of new option.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -540,14 +540,14 @@ config_options = [
            installation ('y'), from a Git submodule ('n'), or to decide
            automatically ('default'). Deprecated option, please use 'googletest' instead.
            This option is supressed by 'googltest' option.""",
-        'default', ('default', 'y', 'n')),
+        'unspecified', ('unspecified','default', 'y', 'n')),
     EnumVariable(
         'googletest',
         """Select whether to use gtest/gmock from system
            installation ('system'), from a Git submodule ('submodule'), to decide
            automatically ('default') or don't look for gtest/gmock ('none')
-           and don't run tests that depend on gtest/gmock.
-           If this option is set then it suppress the deprecated 'system_googletest' option.""",
+           and don't run tests that depend on gtest/gmock. If this option is
+           set then it suppresses the deprecated 'system_googletest' option.""",
         'default', ('default', 'system', 'submodule', 'none')),
     (
         'env_vars',
@@ -916,19 +916,29 @@ except ValueError:
     env['FMT_VERSION'] = '0.0.0'
     print('INFO: Could not find version of fmt')
 
+# Convert setting of the deprecated system_googletest option
+if env['system_googletest'] != 'unspecified':
+    print("WARNING: The 'system_googletest' option is deprecated. "
+        "Use the 'googletest' option instead.")
+if env['system_googletest'] == 'default':
+    env['googletest'] = 'default'
+elif env['system_googletest'] == 'y':
+    env['googletest'] = 'system'
+elif env['system_googletest'] == 'n':
+    env['googletest'] = 'submodule'
+
 # Check for googletest and checkout submodule if needed
 if env['googletest'] in ('system', 'default'):
     has_gtest = conf.CheckCXXHeader('gtest/gtest.h', '""')
     has_gmock = conf.CheckCXXHeader('gmock/gmock.h', '""')
     if has_gtest and has_gmock:
-        env['system_googletest'] = True
         env['googletest'] = 'system'
-    elif env['system_googletest'] == 'y' or env['googletest'] == 'system':
+        print("""INFO: Using system installation of Googletest""")
+    elif env['googletest'] == 'system':
         config_error('Expected system installation of Googletest-1.8.0, but it '
                      'could not be found.')
 
 if env['googletest'] in ('submodule', 'default'):
-    env['system_googletest'] = False
     env['googletest'] = 'submodule'
     has_gtest = os.path.exists('ext/googletest/googletest/include/gtest/gtest.h')
     has_gmock = os.path.exists('ext/googletest/googlemock/include/gmock/gmock.h')
@@ -945,9 +955,10 @@ if env['googletest'] in ('submodule', 'default'):
             config_error('Googletest not found and submodule checkout failed.\n'
                          'Try manually checking out the submodule with:\n\n'
                          '    git submodule update --init --recursive ext/googletest\n')
+    print("""INFO: Using Googletest from Git submodule""")
 
-#do nothing "if env['googletest'] in ('none'):"
-# i.e. don't run tests that use 'googletest' module
+if env['googletest'] == 'none':
+    print("""INFO: Not using Googletest -- unable to run complete test suite""")
 
 # Check for Eigen and checkout submodule if needed
 if env['system_eigen'] in ('y', 'default'):
--- a/SConstruct
+++ b/SConstruct
@@ -538,8 +538,17 @@ config_options = [
         'system_googletest',
         """Select whether to use gtest/gmock from system
            installation ('y'), from a Git submodule ('n'), or to decide
-           automatically ('default').""",
+           automatically ('default'). Deprecated option, please use 'googletest' instead.
+           This option is supressed by 'googltest' option.""",
         'default', ('default', 'y', 'n')),
+    EnumVariable(
+        'googletest',
+        """Select whether to use gtest/gmock from system
+           installation ('system'), from a Git submodule ('submodule'), to decide
+           automatically ('default') or don't look for gtest/gmock ('none')
+           and don't run tests that depend on gtest/gmock.
+           If this option is set then it suppress the deprecated 'system_googletest' option.""",
+        'default', ('default', 'system', 'submodule', 'none')),
     (
         'env_vars',
         """Environment variables to propagate through to SCons. Either the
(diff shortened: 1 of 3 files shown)


SConstruct
```diff
@@ -538,8 +538,16 @@ config_options = [
         'system_googletest',
         """Select whether to use gtest/gmock from system
            installation ('y'), from a Git submodule ('n'), or to decide
-           automatically ('default').""",
+           automatically ('default'). Depricated option, please use 'googletest' instead.""",
```
speth: The correct spelling is 'Deprecated'.

SConstruct
```diff
@@ -881,32 +889,45 @@ if env['system_fmt'] in ('n', 'default'):
...
+    if env['system_googletest'] in ('y', 'default'):
+        has_gtest = conf.CheckCXXHeader('gtest/gtest.h', '""')
+        has_gmock = conf.CheckCXXHeader('gmock/gmock.h', '""')
+        if has_gtest and has_gmock:
+            env['system_googletest'] = True
+        elif env['system_googletest'] == 'y' or env['googletest'] == 'system':
+            config_error('Expected system installation of Googletest-1.8.0, but it '
+                         'could not be found.')
+    elif env['system_googletest'] == 'n' and env['googletest'] == 'system':
+        config_error('Options "googletest=system" and "system_googletest=n" contradict each other.\n'
```
speth: I would be fine with having the `googletest` option always override the `system_googletest` option, with no need to check for consistency.

SConstruct
```diff
@@ -538,8 +538,16 @@ config_options = [
         'system_googletest',
         """Select whether to use gtest/gmock from system
            installation ('y'), from a Git submodule ('n'), or to decide
-           automatically ('default').""",
+           automatically ('default'). Depricated option, please use 'googletest' instead.""",
         'default', ('default', 'y', 'n')),
```
speth: It might be useful to introduce a new default here, e.g. `unspecified`, treated as equivalent to `default`, and print a warning if any of the current options are explicitly given, `default` included.

SConstruct
```diff
@@ -881,32 +889,45 @@ if env['system_fmt'] in ('n', 'default'):
...
-    has_gtest = conf.CheckCXXHeader('gtest/gtest.h', '""')
-    has_gmock = conf.CheckCXXHeader('gmock/gmock.h', '""')
-    if has_gtest and has_gmock:
-        env['system_googletest'] = True
+if env['googletest'] in ('system', 'default'):
+    if env['system_googletest'] in ('y', 'default'):
+        has_gtest = conf.CheckCXXHeader('gtest/gtest.h', '""')
+        has_gmock = conf.CheckCXXHeader('gmock/gmock.h', '""')
+        if has_gtest and has_gmock:
+            env['system_googletest'] = True
```
speth: Should be setting `env['googletest'] = 'system'` here, right?

test/SConscript
```diff
@@ -15,11 +15,15 @@ else:
 
 localenv.Prepend(CPPPATH=['#include'],
                  LIBPATH='#build/lib')
-if not env['system_googletest']:
-    localenv.Prepend(CPPPATH=['#ext/googletest/googletest/include',
+if env['googletest'] in ('system', 'submodule', 'default'):
```
speth: Once we know whether there is either a system or submodule googletest, we should no longer be looking for `default`

test/SConscript
```diff
@@ -87,10 +91,11 @@ def addTestProgram(subdir, progName, env_vars={}):
     passedFile = File(pjoin(str(program[0].dir), '%s.passed' % program[0].name))
     PASSED_FILES[progName] = str(passedFile)
     testResults.tests[passedFile.name] = program
-    run_program = testenv.Command(passedFile, program, gtestRunner)
-    env.Depends(run_program, env['build_targets'])
-    env.Depends(env['test_results'], run_program)
-    Alias('test-%s' % progName, run_program)
+    if env['googletest'] in ('system', 'submodule', 'default'):
```
speth: More cleanly written as `if env['googletest'] != 'none':`

speth (changes requested):
I think it would help to structure this such that after the initial checks, we only look at the `googletest` variable, which should have as its value one of `system`, `submodule`, or `none`, and all later references to the `system_googletest` variable should be removed.

When it comes to running the test suite, I think there need to be some very loud and clear warnings that these tests are being skipped, printed in the test summary at the very end.

SConstruct
band-a-prend: Corrected.

SConstruct
band-a-prend: Done.

SConstruct
band-a-prend: I still don't implement this. Could you clarify what is the warning message should be?
That "it's will be skipped if 'googletest' is set explicitly" and if not then notify that "this option is deprecated and in future user should use another"?


And about your offering about additional message
> When it comes to running the test suite, I think there need to be some very loud and clear warnings that these tests are being skipped, printed in the test summary at the very end.

It seems that I need to edit 'buildutils.py' and I currently have no idea how to implement this behaviour.
So far I added additional warning before running tests after the ending of Cantera build process.

Sorry for delaying with my answer. That's usually a problem for me to edit and rebase commits on weekdays.

SConstruct
band-a-prend: Yes, you are right. Also I add string `env['googletest'] = 'submodule'` to another block below.

test/SConscript
band-a-prend: Removed 'default' here.

test/SConscript
band-a-prend: Done. Yes it's more clarifying condition.

band-a-prend:
I made some updates in SCounstruct and test/SConscript.

About item 3 in your "changes request":

> It might be useful to introduce a new default here, e.g. unspecified, treated as equivalent to default, and print a warning if any of the current options are explicitly given, default included.

I still don't implement this. Could you clarify what is the warning message should be?
That "it's will be skipped if 'googletest' is set explicitly" and if not then notify that "this option is deprecated and in future user should use another"?

And about your offering about additional message:

> When it comes to running the test suite, I think there need to be some very loud and clear warnings that these tests are being skipped, printed in the test summary at the very end.

It seems that I need to edit 'buildutils.py' and I currently have no idea how to implement this behaviour.
So far I added additional warning before running tests after the ending of Cantera build process.

Sorry for delaying with my answer. That's usually a problem for me to edit and rebase commits on weekdays.

band-a-prend:
@speth 

Could you check why OSX "The Travis CI build" is failed?
The Linux build was completed successfully.

band-a-prend:
I'm in a shame spiral. The build problem was due to I inserted few tabulations in two strings while editing.

test/SConscript
```diff
@@ -15,11 +15,15 @@ else:
...
+    if not env['system_googletest']:
+        localenv.Prepend(CPPPATH=['#ext/googletest/googletest/include',
                               '#ext/googletest/googlemock/include'])
-localenv.Append(LIBS=['gtest', 'gmock'] + cantera_libs,
-                CCFLAGS=env['warning_flags'])
+    localenv.Append(LIBS=['gtest', 'gmock'] + cantera_libs,
+                    CCFLAGS=env['warning_flags'])
+else:
+    localenv.Append(LIBS=cantera_libs,
+                    CCFLAGS=env['warning_flags'])
```
speth: Since we append `env['warning_flags']` to `CCFLAGS` regardless, this should be done outside the if/else block.

speth (commented):
Thanks, this is looking quite a bit cleaner now. I think there's still a bit that can be done to help with the transition from the `system_googletest` to `googletest` option. If we introduce `unspecified` as the default choice for the `system_googletest` option, taking this to mean that the user has not given any of the three existing choices, then I think logic should be something like the following:
```python
if env['system_googletest'] != 'unspecified':
    warn about deprecated option
if env['system_googletest'] == 'default':
    env['googletest'] = 'default'
elif env['system_googletset'] == 'y':
    env['googletest'] = 'system'
elif env['system_googletest'] == 'n':
    env['googletest'] = 'submodule'
```
After which point the 'system_googletest' variable should not be used at all.

I'll see about putting together a patch that gives the behavior I'd like for running the test suite when use of googletest is disabled.

speth (approved):

PR #536, opened 2018-05-14
Allow convertMech function to be called multiple times

### LINKED ISSUE #528: Multiple runs of ck2cti.Parser result in incorrect duplicate species error
### Cantera version
2.4.0b1 (3d59914e6)

### Operating System
Any

### Python/MATLAB version
Any

### Expected Behavior
The ck2cti parser should raise an error about problems in the thermo.txt file regarding an incorrect specification of the thermo entry:
```
ERROR: Unable to parse 'mech_debug/thermo.txt' near line 526:
```

### Actual Behavior
Instead, the parser raises an error about an additional declaration of species `H`, although there are not duplicated species in the mechanism:
```
InputParseError: Found additional declaration of species H
```

It appears that some information is cached between runs of the parser, resulting in this duplicate species declaration. If this is actually the expected behavior, it is not clear that this would happen, especially when input files are changed.

### Steps to reproduce
```python
import cantera.ck2cti as ck2cti
parser = ck2cti.Parser()
parser.convertMech('mech.txt', 'thermo.txt', 'tran.txt', outName='mech.cti')
parser.convertMech('mech_fixed.txt', 'thermo.txt', 'tran.txt', outName='mech.cti')
```

[mech_fixed.txt](https://github.com/Cantera/cantera/files/1934520/mech_fixed.txt)
[mech.txt](https://github.com/Cantera/cantera/files/1934521/mech.txt)
[thermo.txt](https://github.com/Cantera/cantera/files/1934522/thermo.txt)
[tran.txt](https://github.com/Cantera/cantera/files/1934523/tran.txt)

speth:
Each call to convertMech now creates a new Parser object to do the conversion, rather than requiring the user to do so themselves. The implementation as a `@classmethod` should fix the current method of using a `Parser` object to work, as well as the simpler free function `ck2cti.convertMech`.

Fixes #528.

### DIFF
--- a/interfaces/cython/cantera/ck2cti.py
+++ b/interfaces/cython/cantera/ck2cti.py
@@ -2132,7 +2132,8 @@ class Parser(object):
 
         return surface_names
 
-    def showHelp(self):
+    @staticmethod
+    def showHelp():
         print("""
 ck2cti.py: Convert Chemkin-format mechanisms to Cantera input files (.cti)
 
@@ -2165,9 +2166,12 @@ duplicate transport data) to be ignored.
 
 """)
 
-    def convertMech(self, inputFile, thermoFile=None,
-                    transportFile=None, surfaceFile=None, phaseName='gas',
-                    outName=None, quiet=False, permissive=None):
+    @staticmethod
+    def convertMech(inputFile, thermoFile=None, transportFile=None,
+                    surfaceFile=None, phaseName='gas', outName=None,
+                    quiet=False, permissive=None):
+
+        parser = Parser()
         if inputFile:
             inputFile = os.path.expanduser(inputFile)
         if thermoFile:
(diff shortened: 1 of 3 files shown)


interfaces/cython/cantera/ck2cti.py
```diff
@@ -2132,7 +2132,8 @@ def writeCTI(self, header=None, name='gas', transportModel='Mix',
 
         return surface_names
 
-    def showHelp(self):
+    @classmethod
```
bryanwweber: I think this should be a `@staticmethod` since it doesn't use the `cls` parameter (or the old `self` parameter for that matter)

interfaces/cython/cantera/ck2cti.py
```diff
@@ -2165,9 +2166,12 @@ def showHelp(self):
 
 """)
 
-    def convertMech(self, inputFile, thermoFile=None,
+    @classmethod
```
bryanwweber: I think this could also be a `@staticmethod`

bryanwweber (approved):
I think the two `@classmethods` could actually be `@staticmethod`, but otherwise, this looks good to me.

speth:
Oh, good point! These are even better as `@staticmethod`s.

PR #540, opened 2018-05-24
Fixes for fmtlib 5.0.0

bryanwweber:
In fmtlib 5.0.0, `MemoryWriter` was removed. The [recommendation](https://github.com/fmtlib/fmt/issues/710) is to replace it with `memory_buffer`

- Switch `MemoryWriter` uses to `memory_buffer` uses

I have no idea if this is the right fix, I'm putting it out there to get some feedback and start to learn the C++ better. I also didn't test this with fmtlib 4.x, so I'm not sure if it's backwards compatible. Once this commit is applied, there is another error I don't know how to fix:

```
g++ -o build/src/kinetics/Group.os -c -std=c++0x -O3 -Wno-inline -g -Wall -include src/pch/system.h -fPIC -DNDEBUG -Iinclude -Iinclude/cantera/ext -Ibuild/src -I/Users/bryan/miniconda3/envs/cantera-dev/include src/kinetics/Group.cpp
In file included from src/kinetics/Group.cpp:9:
include/cantera/kinetics/Group.h:128:40: error: too many arguments provided to function-like macro invocation
    std::ostream& fmt(std::ostream& s, const std::vector<std::string>& esymbols) const;
                                       ^
/usr/local/include/fmt/format.h:3655:10: note: macro 'fmt' defined here
# define fmt(s) FMT_STRING(s)
         ^
In file included from src/kinetics/Group.cpp:9:
include/cantera/kinetics/Group.h:128:22: error: expected ';' at end of declaration list
    std::ostream& fmt(std::ostream& s, const std::vector<std::string>& esymbols) const;
                     ^
                     ;
include/cantera/kinetics/Group.h:23:5: error: constructor for 'Cantera::Group' must explicitly initialize the reference member 'fmt'
    Group() : m_sign(-999) { }
    ^
include/cantera/kinetics/Group.h:128:19: note: declared here
    std::ostream& fmt(std::ostream& s, const std::vector<std::string>& esymbols) const;
                  ^
include/cantera/kinetics/Group.h:24:5: error: constructor for 'Cantera::Group' must explicitly initialize the reference member 'fmt'
    Group(size_t n) : m_sign(0) {
    ^
include/cantera/kinetics/Group.h:128:19: note: declared here
    std::ostream& fmt(std::ostream& s, const std::vector<std::string>& esymbols) const;
                  ^
include/cantera/kinetics/Group.h:27:5: error: constructor for 'Cantera::Group' must explicitly initialize the reference member 'fmt'
    Group(const vector_int& elnumbers) :
    ^
include/cantera/kinetics/Group.h:128:19: note: declared here
    std::ostream& fmt(std::ostream& s, const std::vector<std::string>& esymbols) const;
                  ^
include/cantera/kinetics/Group.h:31:5: error: constructor for 'Cantera::Group' must explicitly initialize the reference member 'fmt'
    Group(const std::vector<size_t>& elnumbers) :
    ^
include/cantera/kinetics/Group.h:128:19: note: declared here
    std::ostream& fmt(std::ostream& s, const std::vector<std::string>& esymbols) const;
                  ^
include/cantera/kinetics/Group.h:38:5: error: constructor for 'Cantera::Group' must explicitly initialize the reference member 'fmt'
    Group(const Group& g) :
    ^
include/cantera/kinetics/Group.h:128:19: note: declared here
    std::ostream& fmt(std::ostream& s, const std::vector<std::string>& esymbols) const;
                  ^
src/kinetics/Group.cpp:40:26: error: too many arguments provided to function-like macro invocation
                         const std::vector<std::string>& esymbols) const
                         ^
/usr/local/include/fmt/format.h:3655:10: note: macro 'fmt' defined here
# define fmt(s) FMT_STRING(s)
         ^
src/kinetics/Group.cpp:39:22: error: non-static data member defined out-of-line
std::ostream& Group::fmt(std::ostream& s,
              ~~~~~~~^
src/kinetics/Group.cpp:39:22: error: declaration of reference variable 'fmt' requires an initializer
std::ostream& Group::fmt(std::ostream& s,
                     ^~~
src/kinetics/Group.cpp:39:25: error: expected ';' after top level declarator
std::ostream& Group::fmt(std::ostream& s,
                        ^
                        ;
src/kinetics/Group.cpp:41:1: error: expected unqualified-id
{
^
12 errors generated.
```

It looks like the `Group::fmt` function shadows the macro of the same name defined by fmtlib, but I'm not totally sure...

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -881,7 +881,7 @@ if env['system_fmt'] in ('y', 'default'):
 if env['system_fmt'] in ('n', 'default'):
     env['system_fmt'] = False
     print("""INFO: Using private installation of fmt library.""")
-    if not os.path.exists('ext/fmt/fmt/ostream.h'):
+    if not os.path.exists('ext/fmt/include/fmt/ostream.h'):
         if not os.path.exists('.git'):
             config_error('fmt is missing. Install source in ext/fmt.')
 
@@ -894,6 +894,18 @@ if env['system_fmt'] in ('n', 'default'):
             config_error('fmt submodule checkout failed.\n'
                          'Try manually checking out the submodule with:\n\n'
                          '    git submodule update --init --recursive ext/fmt\n')
+
+fmt_include = '<fmt/format.h>' if env['system_fmt'] else '"../ext/fmt/include/fmt/format.h"'
+fmt_version_source = get_expression_value([fmt_include], 'FMT_VERSION')
+retcode, fmt_lib_version = conf.TryRun(fmt_version_source, '.cpp')
+try:
+    fmt_lib_version = divmod(float(fmt_lib_version.strip()), 10000)
+    (fmt_maj, (fmt_min, fmt_pat)) = fmt_lib_version[0], divmod(fmt_lib_version[1], 100)
+    env['FMT_VERSION'] = '{major:.0f}.{minor:.0f}.{patch:.0f}'.format(major=fmt_maj, minor=fmt_min, patch=fmt_pat)
+    print('INFO: Found fmt version {}'.format(env['FMT_VERSION']))
+except ValueError:
+    env['FMT_VERSION'] = '0.0.0'
+    print('INFO: Could not find version of fmt')
 
 # Check for googletest and checkout submodule if needed
 if env['system_googletest'] in ('y', 'default'):
--- a/SConstruct
+++ b/SConstruct
@@ -852,6 +852,21 @@ def config_error(message):
 if not conf.CheckCXXHeader('cmath', '<>'):
     config_error('The C++ compiler is not correctly configured.')
 
+
+def get_expression_value(includes, expression):
+    s = ['#include ' + i for i in includes]
+    s.extend(('#define Q(x) #x',
+              '#define QUOTE(x) Q(x)',
+              '#include <iostream>',
+              '#ifndef SUNDIALS_PACKAGE_VERSION', # name change in Sundials >= 3.0
+              '#define SUNDIALS_PACKAGE_VERSION SUNDIALS_VERSION',
+              '#endif',
+              'int main(int argc, char** argv) {',
+              '    std::cout << %s << std::endl;' % expression,
+              '    return 0;',
+              '}\n'))
+    return '\n'.join(s)
+
 # Check for fmt library and checkout submodule if needed
 # Test for 'ostream.h' to ensure that version >= 3.0.0 is available
 if env['system_fmt'] in ('y', 'default'):
@@ -912,12 +927,14 @@ if env['system_googletest'] in ('n', 'default'):
 if env['system_eigen'] in ('y', 'default'):
     if conf.CheckCXXHeader('Eigen/Dense', '<>'):
         env['system_eigen'] = True
+        print("""INFO: Using system installation of Eigen.""")
     elif env['system_eigen'] == 'y':
         config_error('Expected system installation of Eigen, but it '
                      'could not be found.')
 
 if env['system_eigen'] in ('n', 'default'):
     env['system_eigen'] = False
+    print("""INFO: Using private installation of Eigen.""")
     if not os.path.exists('ext/eigen/Eigen/Dense'):
         if not os.path.exists('.git'):
             config_error('Eigen is missing. Install Eigen in ext/eigen.')
@@ -932,20 +949,12 @@ if env['system_eigen'] in ('n', 'default'):
                          'Try manually checking out the submodule with:\n\n'
                          '    git submodule update --init --recursive ext/eigen\n')
 
-
-def get_expression_value(includes, expression):
-    s = ['#include ' + i for i in includes]
-    s.extend(('#define Q(x) #x',
-              '#define QUOTE(x) Q(x)',
-              '#include <iostream>',
-              '#ifndef SUNDIALS_PACKAGE_VERSION', # name change in Sundials >= 3.0
-              '#define SUNDIALS_PACKAGE_VERSION SUNDIALS_VERSION',
-              '#endif',
-              'int main(int argc, char** argv) {',
-              '    std::cout << %s << std::endl;' % expression,
-              '    return 0;',
-              '}\n'))
-    return '\n'.join(s)
+eigen_include = '<Eigen/Core>' if env['system_eigen'] else '"../ext/eigen/Eigen/Core"'
+eigen_versions = 'QUOTE(EIGEN_WORLD_VERSION) "." QUOTE(EIGEN_MAJOR_VERSION) "." QUOTE(EIGEN_MINOR_VERSION)'
+eigen_version_source = get_expression_value([eigen_include], eigen_versions)
+retcode, eigen_lib_version = conf.TryRun(eigen_version_source, '.cpp')
+env['EIGEN_LIB_VERSION'] = eigen_lib_version.strip()
+print('INFO: Found Eigen version {}'.format(env['EIGEN_LIB_VERSION']))
 
 # Determine which standard library to link to when using Fortran to
 # compile code that links to Cantera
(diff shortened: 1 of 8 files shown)


bryanwweber:
And as it turns out, this is not backwards compatible. Does that matter? Should I just bump the commit in the submodule?

speth:
We absolutely need to be backwards compatible. Otherwise we won't be able to do things like build the Ubuntu packages, which will not have the version of fmt.

speth:
I think the problem with that macro (which is hideous) can be resolved if we add `#define FMT_NO_FMT_STRING_ALIAS` near the top of `fmt.h`.

bryanwweber:
So how should we handle the backwards compatibility? `#if FMT_VERSION =...` directives to change which function is used?

speth:
Yes, it will need to be `#ifdef`ed based on `FMT_VERSION`. To avoid cluttering the code with too many `#ifdefs`, what I think I would do is, for the case of `FMT_VERSION < 50000`, define/alias the functions needed to match the new API. This might get a bit tricky given that `format_to` is a variadic template.

bryanwweber:
Would be nice if they offered a clean upgrade path with deprecation of old functions before summarily removing them... 

I'm not really sure, but it seems like aliasing variadic template functions isn't really possible: https://stackoverflow.com/q/24330300 how would you recommend proceeding?

bryanwweber:
By the way, if its easier, feel free to supersede this PR or [push commits to my branch](https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/)

speth:
Sorry, using the term 'alias' may have been a red herring. This is what I was thinking of. The variadic templates actually end up being pretty clean.

bryanwweber:
Ah, I should have focused more on the "define" and less on the "alias" 😄 Thank you for adding the code!

I added some additional output to SConstruct with the versions of Eigen and fmt (googletest doesn't have a way to get the version easily) and also whether or not the system or private installation of these libraries is being used. I also noticed that the system Eigen check didn't seem to be looking in the right directory on my Mac for the files, so I fixed that. I'm not sure if the headers are installed in a different place on other OS.

I also bumped fmt to 5.0.0 since we were back at 3.0.1. Not sure if this was appropriate, so I put it into its own commit to be easily reverted.

bryanwweber:
And this branch also fixes the build for Homebrew

SConstruct
```diff
@@ -910,14 +938,16 @@ if env['system_googletest'] in ('n', 'default'):
 
 # Check for Eigen and checkout submodule if needed
 if env['system_eigen'] in ('y', 'default'):
-    if conf.CheckCXXHeader('Eigen/Dense', '<>'):
+    if conf.CheckCXXHeader('eigen3/Eigen/Dense', '<>'):
```
speth: I'm not sure what the right way of handling Eigen is, but this change is insufficient. The Eigen docs suggest that the includes be written as `#include <Eigen/Dense>`, which is what we do, e.g. in `numerics/eigen_dense.h`. If the headers were installed in `/usr/include/eigen3/Eigen` (Ubuntu does this, e.g.), then you could always write `#include <eigen3/Eigen/Dense>`, and pick this up from the fact that `/usr/include` is on the default search path. However, it would not work for other locations for the Eigen headers. The current approach is that to use the system Eigen on such a system, you need to add the right directory to `extra_inc_dirs`, as described in the docstring for the `system_eigen` option.

SConstruct
```diff
@@ -932,20 +962,16 @@ if env['system_eigen'] in ('n', 'default'):
...
-              '#include <iostream>',
-              '#ifndef SUNDIALS_PACKAGE_VERSION', # name change in Sundials >= 3.0
-              '#define SUNDIALS_PACKAGE_VERSION SUNDIALS_VERSION',
-              '#endif',
-              'int main(int argc, char** argv) {',
-              '    std::cout << %s << std::endl;' % expression,
-              '    return 0;',
-              '}\n'))
-    return '\n'.join(s)
+env['EIGEN_VERSION'] = '.'.join(eigen_version)
```
speth: You should be able to get all of this info in a single call to `get_expression_value`. The expression should be something like
```cpp
QUOTE(EIGEN_WORLD_VERSION) "." QUOTE(EIGEN_MAJOR_VERSION) "." QUOTE(EIGEN_MINOR_VERSION)
```
where the `QUOTE` macro is already defined in the `get_expression_value` program.

include/cantera/base/fmt.h
```diff
@@ -1,16 +1,47 @@
...
+#ifndef CT_FMT_H
+#define CT_FMT_H
 #include "ct_defs.h"
 
+//! Do not use the fmt macro from fmtlib because it shadows a function of
+//! the same name in kinetics/Group.h
+#define FMT_NO_FMT_STRING_ALIAS
+
+//! Only use the header portions of fmtlib
+#define FMT_HEADER_ONLY
```
speth: Is there a reason we're changing to using this as header only? We were specifically compiling the compilable portion before.

SConstruct
```diff
@@ -880,18 +895,31 @@ if env['system_fmt'] in ('n', 'default'):
...
     if has_gtest and has_gmock:
         env['system_googletest'] = True
+        print("""INFO: Using system installation of googletest library.""")
     elif env['system_googletest'] == 'y':
         config_error('Expected system installation of Googletest-1.8.0, but it '
                      'could not be found.')
 
 if env['system_googletest'] in ('n', 'default'):
     env['system_googletest'] = False
+    print("""INFO: Using private installation of googletest library.""")
```
speth: Can we not do this right now? I'm anticipating merge conflicts with #535 if there are any changes to the googletest options.

SConstruct
bryanwweber: OK that makes sense. The directory is the same on my Mac as well (via Homebrew), so I'm not sure when they would end up in a different directory, but if that's what the docs say, that's what we'll do.

include/cantera/base/fmt.h
bryanwweber: Well, the short answer is because I didn't know that and this fixed the compile problems I was having. I think I fixed this now, but fmtlib >= 5.0.0 is apparently distributed with a pre-compiled static library (at least from Homebrew) so I had to work around that, or I got a bunch of duplicated symbols errors when linking the final Cantera library together.

SConstruct
bryanwweber: I dropped this commit.

bryanwweber:
@speth With the exception of needing to squash the last commit here, I think this is good :+1: Any other comments?

SConstruct
```diff
@@ -932,20 +961,16 @@ if env['system_eigen'] in ('n', 'default'):
                          'Try manually checking out the submodule with:\n\n'
                          '    git submodule update --init --recursive ext/eigen\n')
 
+if env['system_eigen']:
+    eigen_include = ['<Eigen/Core>', '<Eigen/src/Core/util/Macros.h>']
+else:
+    eigen_include = ['"../ext/eigen/Eigen/Core"', '"../ext/eigen/Eigen/src/Core/util/Macros.h"']
```
speth: `Eigen/Core` already has `#include "src/Core/util/Macros.h"`. Is there a reason to explicitly include this?

ext/SConscript
```diff
@@ -37,16 +37,24 @@ def prep_gmock(env):
...
     return localenv
 
 # each element of libs is: (subdir, (file extensions), prepfunction)
-libs = [('libexecstream', ['cpp'], prep_default),
-        ('fmt/fmt', ['cc'], prep_fmt)]
+libs = [('libexecstream', ['cpp'], prep_default)]
+
+# fmtlib versions less than 5.0.0 were not distributed with a compiled library,
+# so we should build it ourselves. Otherwise, only need to build if not using the system fmt
+if LooseVersion(env['FMT_VERSION']) < LooseVersion('5.0.0'):
```
speth: Distributed by whom? I think the point is, if we are using a non-system `fmt`, we want to compile the source files (which are, of course, in a different directory in 5.0.0). And I guess, in the case of the system `fmt` library, we are expecting to find a compiled library to link to.

ext/SConscript
```diff
@@ -37,16 +37,24 @@ def prep_gmock(env):
 def prep_fmt(env):
     localenv = prep_default(env)
     if not env['system_fmt']:
+        if LooseVersion(env['FMT_VERSION']) >= LooseVersion('5.0.0'):
```
speth: If we're bumping the submodule to fmt 5.0.0, this check is unnecessary, right?

speth:
Sorry for the earlier suggestions to the contrary, but I guess there is a good reason to use the header-only form of `fmt`. Namely, that it only builds static libraries by default, which on Linux causes problems related to relocatable symbols (similar to the problem that crops of with Sundials if you compile without `-fPIC`).

bryanwweber:
OK I squashed a few of the commits and removed the now-unnecessary `posix.h` include.

PR #544, opened 2018-06-12
Move docs around

bryanwweber:
- This PR moves the documentation files around and restyles them to match the new website
- This PR should be merged and new docs pushed to their respective repositories before deploying the new website

The main changes here are to remove examples, "science" documentation (governing equations, etc.), tutorials, and information about Cantera - basically anything whose new home is somewhere else in the new website. Other changes:

* Switch to KaTeX for math rendering instead of MathJax. [KaTeX](https://khan.github.io/KaTeX/) is much faster than MathJax, and is used on the main website. The only caveat is that it adds one more dependency to building the Sphinx docs (`sphinxcontrib.katex`), and that dependency doesn't work right in the released version (or, at least, it didn't for me). So I made a [PR over there](https://github.com/hagenw/sphinxcontrib-katex/pull/2) to fix the problem. We'll see what happens, but I'm not opposed to dropping this, since the math in the Sphinx docs is minimal. Also, I just realized Doxygen is still using MathJax, so I'll need to figure out if I can change that.
* Switch to the alabaster Sphinx theme and redo `layout.html` to match the main website theme a little better.

To-Do still:

- [ ] ~Check if KaTeX can be used with Doxygen~ It seems it cannot, but the Doxygen docs already use a completely different theme anyways, so I'm not going to muck around with it
- [ ] ~Fix that the documentation files for `kinetics.h` and `Kinetics.h` shadow each other on case-insensitive file systems (Mac, Windows)~ Don't have time to do this now, and everything will be built on Linux (Travis) anyways
- [x] Remove files whose content has migrated to elsewhere in the new website (`about.rst`, `faq.rst`, etc.)
- [x] Set up auto-push to the `dev-docs` repository on successful build of the `master` branch
- [x] Fix as many Sphinx and Doxygen warnings as I can
- [x] Regenerate Spinx conf.py and make sure options are up-to-date


bryanwweber:
@speth Provided this passes the tests, I would appreciate some 👀 on this PR. I tested the pushing to GH of the dev-docs over on the [`dev_docs_push`](https://github.com/Cantera/cantera/tree/dev_docs_push) branch, since PRs from forks don't have access to encrypted variables. Pending any requested changes, I think this has all the updates needed for the new site.

doc/sphinx/conf.py
```diff
@@ -41,12 +41,11 @@
               'sphinx.ext.autodoc',
               'sphinx.ext.todo',
               'sphinx.ext.autosummary',
-              'sphinx.ext.mathjax',
               'sphinxcontrib.doxylink',
+              'sphinxcontrib.katex',  # Use KaTeX because it's faster and the main site uses it
```
speth, line 45: Is this requirement documented somewhere?

speth (commented):
Is there something we can do about all of the absolute URLs which are used in this which require things to be found at the root of the web server? If I build the docs locally and want to view them using a `file:///` URL, then none of the CSS or JS stuff is loaded. Is there some other way to easily view the content?

doc/sphinx/conf.py
bryanwweber: Yes, here: http://engr.uconn.edu/~weber/cantera/compiling/dependencies.html#sec-dependencies

bryanwweber:
Three ways I can think of would be to copy the JS and CSS from the other theme over here, or recreate whatever minimal portion of that we use in these docs, or not follow the style of the "main" site quite so closely.

FWIW, other than the header and footer, all the links internal to the Sphinx docs here should work, so you should be able to navigate within these docs reasonably well using the sidebar navigation.

doc/sphinx/cttheme/theme.conf
```diff
@@ -1,3 +1,3 @@
 [theme]
-inherit = sphinxdoc
+inherit = alabaster
```
speth, line 2: This theme certainly matches the main page better, but I think we should to switch to a the same sans-serif fonts that are used there as well.

include/cantera/thermo/WaterProps.h
```diff
@@ -22,18 +22,19 @@ class PDSS_Water;
 /**
  * @defgroup relatedProps Electric Properties of Phases
  *
- * ### Treatment of the %Phase Potential and the electrochemical potential of
- * a species
+ * Computation of the electric properties of phases
  *
- * The electrochemical potential of species *k* in a phase *p*, \f$ \zeta_k \f$,
+ * ### Treatment of the Phase Potential and the electrochemical potential of a species
```
speth: I guess this should really just be lowercase "phase potential"?

interfaces/cython/cantera/ck2cti.py
```diff
@@ -822,7 +822,7 @@ class Sri(object):
     """
     A kinetic model of a phenomenological rate coefficient k(T, P) using the
     "SRI" formulation of the blending function :math:`F` using either 3 or
-    5 parameters. See :ref:`sec-sri-falloff`.
+    5 parameters. See https://cantera.org/science/reactions.html#sec-sri-falloff.
```
speth: Can we links with titles that are different from the URLs, as we had here before (also applies to the other instances where the use of the `:ref:` role has just been replaced by a plain URL).

interfaces/cython/cantera/reactor.pyx
```diff
@@ -983,19 +983,18 @@ cdef class ReactorNet:
...
         `Reactor` or `IdealGasReactor`:
-
-          - 0  - mass
-          - 1  - volume
-          - 2  - internal energy or temperature
-          - 3+ - mass fractions of the species
+        - 0  - mass
+        - 1  - volume
+        - 2  - internal energy or temperature
+        - 3+ - mass fractions of the species
```
speth, line 993: I thought this was formatted correctly before. Now, it's just a smushed onto a single line.

.gitignore
```diff
@@ -1,3 +1,4 @@
+github_deploy_key
```
speth: I'd like to keep the `github_deploy_key.enc` out of the root of the repository -- I think it can go in the `doc` subdirectory, and this file could just be written to `/tmp` so that nothing would need to be added to the `.gitignore`. (assuming we stick to deploying to a Github repo).

doc/sphinx/_templates/layout.html
```diff
@@ -1,10 +1,165 @@
...
-{% block relbar1 %}
-<div style="background-color: white; text-align: left; padding: 10px 10px 15px 15px">
-<a href="{{ pathto('index') }}">
-<img src="{{pathto("_static/cantera-logo.png", 1) }}" border="0" alt="Cantera"/></a>
-</div>
-{{ super() }}
-{% endblock %}
+
+{%- macro css() %}
+  <link href="/assets/css/bootstrap.min.css" rel="stylesheet" type="text/css" />
```
speth: I would like to see these pages look passable if you build and view them locally. I think loading Bootstrap from a CDN would get us 90% of the way there.

speth (commented):
I noticed that links from the Python docs to the C++ docs (e.g. within http://engr.uconn.edu/~weber/cantera/documentation/docs-2.4/sphinx/html/cython/thermo.html#species) do not seem to be going to the correct location.

The only other question that came to mind was whether we should deploy directly to `cantera.org`, or if not, what the mechanism would be for making the dev docs available there.

interfaces/cython/cantera/reactor.pyx
bryanwweber: I think it ended up formatted as a blockquote, which is why I changed it. Docutils is such a PITA with the input format

interfaces/cython/cantera/ck2cti.py
bryanwweber: I'm not sure if this will be possible, since the content being referred to has moved to the main site, and I'm not sure how to get that information into Sphinx at build-time of the API docs. I agree it would be nice to have the arbitrary title references back though...

bryanwweber: 👍

interfaces/cython/cantera/reactor.pyx
bryanwweber: :+1:

bryanwweber:
In d9c67eb I enabled generation of XML output in addition to the HTML output.  I think the XML output will probably be useful for the stable documentation to be able to link to different places from the website to the C++ docs. See https://github.com/Cantera/cantera-website/issues/16 However, since these files would only be useful for the stable docs, I set rsync here not to upload them to the dev-docs folder on the website.

bryanwweber:
This is good to go on my end now I think

PR #547, opened 2018-07-21
Add temperature dependence of rotational relaxation

speth:
The temperature dependence of the rotational relaxation is given by Equation 12.112 of _Chemically Reacting Flow_ (1st ed.). This factor results in an increase in mixture-averaged thermal conductivity of ~1% or less, and a similar increase in laminar flame speeds, at least for some test cases.

Originally reported here: https://groups.google.com/forum/#!topic/cantera-users/2_u6PjBqPcY


bryanwweber (approved):
The equation is 11.115 in the 2nd edition of Chemically Reacting Flow. Can you add a comment to that effect (maybe including both equation numbers) in the code?

speth:
If we're going to start reliably referencing equations, I think we need a convenient shorthand, and a consistent place to put the full citations. As a starting point, I'll suggest just writing `[Kee2017] Eq. 11.115` in the code comments, and providing the full citation, e.g. 
```
[Kee2017] R. J. Kee, M. E. Coltrin, P. Glarborg, and H. Zhu. Chemically Reacting Flow: Theory and Practice. 2nd Ed. John Wiley and Sons, 2017.
````
in either the class docstring (i.e. the `MixTransport` class, in this case), a consolidated bibliography somewhere in the docs, or both.

bryanwweber:
For now, I think this format is fine, with the bibliography going in the class docstring. Doxygen does have a [`\cite`](https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdcite) command that we might be able to make use of. It requires `bibtex` to be in the `PATH`, so maybe not worth it and we can just develop our own standard.

PR #549, opened 2018-08-09
[Doc] Fix CSS specification bug

bryanwweber:
The font themes must be specified as a string.

### DIFF
--- a/doc/sphinx/conf.py
+++ b/doc/sphinx/conf.py
@@ -129,40 +129,42 @@ pygments_style = 'sphinx'
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 html_theme = 'cttheme'
-html_sidebars = {'**': ['localtoc.html', 'relations.html', 'sourcelink.html', 'searchbox.html', 'numfocus.html']}
+html_sidebars = {
+    '**': ['localtoc.html', 'relations.html', 'sourcelink.html', 'searchbox.html', 'numfocus.html'],
+}
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
 
-# Copy the Bootstrap 4 font families
+# Copy the Bootstrap 4 font families.
 font_families = [
     # Default on Apple
     '-apple-system',
     # Default for older versions of Chrome on Mac
     'BlinkMacSystemFont',
     # Windows
-    'Segoe UI',
+    '"Segoe UI"',
     # Android
     'Roboto',
     # Standard fallbacks
-    'Helvetica Neue', 'Arial', 'sans-serif',
+    '"Helvetica Neue"', 'Arial', 'sans-serif',
     # Emoji fonts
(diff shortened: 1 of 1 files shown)


speth:
Presumably, `doc/sphinx/cttheme/static/cantera.css` also needs to be updated accordingly.

I'm not quite sure what `!default` is supposed to do, but it's not supposed to end up as a member of the list of fonts directly, is it? That's what's currently happening.

Also, it looks like this results in doubled semicolons in the generated `alabaster.css`.

bryanwweber:
Actually, the line in `doc/sphinx/cttheme/static/cantera.css` is correct, I corrected `conf.py`. The `!default` is a line from the SASS (?) preprocessor directive that I misinterpreted from the Bootstrap docs. Fixed, thanks!

speth:
With this change, `alabaster.css` now has:
```css
font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Helvetica Neue", "Arial", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
```
where I guess the syntax is that specific fonts are quoted, and special names like `sans-serif` are not (I have no idea how `BlinkMacSystemFont` is meant to be interpreted. However, `cantera.css` still has:
```css
font-family: -apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,"Helvetica Neue",Arial,sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol" !important;
```
Where quotes are only used for font names containing spaces.

bryanwweber:
Apparently, the quotes aren't required at all (see https://stackoverflow.com/a/13752149/2449192), even for fonts with spaces (although it is recommended), so I've amended the comment in the code. The `BlinkMacSystemFont` is the name of the font that older versions of Chrome use on Mac OS X. I've normalized the fonts without spaces to not have double quotes in the `alabaster.css` file, to save the characters.

doc/sphinx/conf.py
```diff
@@ -129,40 +129,42 @@
...
     # Default on Apple
     '-apple-system',
     # Default for older versions of Chrome on Mac
     'BlinkMacSystemFont',
     # Windows
-    'Segoe UI',
+    '"Segoe UI"',
     # Android
-    'Roboto',
+    '"Roboto"',
```
speth: Per your last comment, this should not have quotes, right?

doc/sphinx/conf.py
```diff
@@ -129,40 +129,42 @@
...
-    'SFMono-Regular',
-    'Menlo',
-    'Monaco',
-    'Consolas',
-    'Liberation Mono',
-    'Courier New', 'monospace'
+    '"SFMono-Regular"',
+    '"Menlo"',
+    '"Monaco"',
+    '"Consolas"',
```
speth: Likewise here, right?

bryanwweber:
Aww man, I had made those changes but forgot to stage them before I commited. Anyways, should be fixed now.

PR #550, opened 2018-08-16
New combustor example

speth:
The main purpose of this is to update the `combustor.py` example so that when people copy it and use it as the basis for their WSR codes, they get a reasonable method for solving such a problem. The hydrogen radical igniter used here is inefficient and complicated, and the method for specifying the inlet mixture is also overly complicated, both for setting the equivalence ratio as well as the mass flow rate.

Adds a new example (`fuel_injection.py`) where the mass flow rate is explicitly time-dependent so that we still have an example demonstrating this capability, but hopefully this configuration will not lead users astray as often.

### DIFF
--- a/interfaces/cython/cantera/examples/reactors/combustor.py
+++ b/interfaces/cython/cantera/examples/reactors/combustor.py
@@ -1,93 +1,83 @@
 """
-A combustor. Two separate stream - one pure methane and the other air, both at
-300 K and 1 atm flow into an adiabatic combustor where they mix and burn.
+A combustor, modeled as a single well-stirred reactor.
 
-We are interested in the steady-state burning solution. Since at 300 K no
-reaction will occur between methane and air, we need to use an 'igniter' to
-initiate the chemistry. A simple igniter is a pulsed flow of atomic hydrogen.
-After the igniter is turned off, the system approaches the steady burning
-solution.
+We are interested in the steady-state burning solution. This example explores
+the effect of changing the residence time on completeness of reaction (through
+the burned gas temperature) and on the total heat release rate.
+
+Demonstrates the use of a MassFlowController where the mass flow rate function
+depends on variables other than time by capturing these variables from the
+enclosing scope. Also shows the use of a PressureController to create a constant
+pressure reactor with a fixed volume.
 """
 
-import math
-import csv
-
+import numpy as np
+import matplotlib.pyplot as plt
 import cantera as ct
 
-# use reaction mechanism GRI-Mech 3.0
-
+# Use reaction mechanism GRI-Mech 3.0
 gas = ct.Solution('gri30.xml')
 
-# create a reservoir for the fuel inlet, and set to pure methane.
-gas.TPX = 300.0, ct.one_atm, 'CH4:1.0'
-fuel_in = ct.Reservoir(gas)
-fuel_mw = gas.mean_molecular_weight
+# Create a Reservoir for the inlet, set to a methane/air mixture at a specified
+# equivalence ratio
+equiv_ratio = 0.5  # lean combustion
+gas.TP = 300.0, ct.one_atm
+gas.set_equivalence_ratio(equiv_ratio, 'CH4:1.0', 'O2:1.0, N2:3.76')
+inlet = ct.Reservoir(gas)
 
-# use predefined function Air() for the air inlet
-air = ct.Solution('air.xml')
-air_in = ct.Reservoir(air)
-air_mw = air.mean_molecular_weight
-
-# to ignite the fuel/air mixture, we'll introduce a pulse of radicals. The
-# steady-state behavior is independent of how we do this, so we'll just use a
-# stream of pure atomic hydrogen.
-gas.TPX = 300.0, ct.one_atm, 'H:1.0'
-igniter = ct.Reservoir(gas)
-
-# create the combustor, and fill it in initially with N2
-gas.TPX = 300.0, ct.one_atm, 'N2:1.0'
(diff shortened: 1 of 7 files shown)


interfaces/cython/cantera/examples/reactors/fuel_injection.py
```diff
@@ -0,0 +1,93 @@
...
+# Create Reactor and set initial contents to be products of lean combustion
+gas.TP = 1000, 20*ct.one_atm
+gas.set_equivalence_ratio(0.30, 'c12h26', 'n2:3.76, o2:1.0')
+gas.equilibrate('TP')
+r = ct.IdealGasReactor(gas)
+r.volume = 0.001 # 1 liter
+
+# Create an inlet for the fuel, supplied as a Gaussian pulse
+def fuel_mdot(t):
+    total = 3.0e-3 # mass of fuel [kg]
```
bryanwweber: Inline comments should have two spaces before the `#`

interfaces/cython/cantera/examples/reactors/fuel_injection.py
```diff
@@ -0,0 +1,93 @@
...
+    if tnow-tprev > 1e-2 or i == 10:
+        i = 0
+        tprev = tnow
+        states.append(r.thermo.state, t=tnow)
+
+if __name__ == '__main__':
+    # Plot the concentrations of some species of interest, including PAH species
+    # which can be considered as precursors to soot formation.
+
+    import matplotlib.pyplot as plt
```
bryanwweber: Any reason not to import this at the top of the file?

interfaces/cython/cantera/examples/reactors/fuel_injection.py
```diff
@@ -0,0 +1,93 @@
...
+sim = ct.ReactorNet([r])
+
+# Integrate for 10 seconds, storing the results for later plotting
+tfinal = 10.0
+tnow = 0.0
+i = 0
+tprev = tnow
+states = ct.SolutionArray(gas, extra=['t'])
+
+while tnow < tfinal:
```
bryanwweber, line 48: Should the integration loop be in the `if __name__...` block, or inside of it's own function? Typically, that pattern is used to protect long-running code from being run if the module is imported. I guess this module is probably not intended to be imported, but then why have the `if __name__...` block at all?

interfaces/cython/cantera/examples/reactors/combustor.py
```diff
@@ -1,93 +1,83 @@
...
+# saving the state after each iteration.
+states = ct.SolutionArray(gas, extra=['tres'])
+
+residence_time = 0.1 # starting residence time
+while combustor.T > 500:
+    sim.set_initial_time(0.0) # reset the integrator
+    sim.advance_to_steady_state()
+    print('tres = {:.2e}; T = {:.1f}'.format(residence_time, combustor.T))
+    states.append(combustor.thermo.state, tres=residence_time)
+    residence_time *= 0.9 # decrease the residence time for the next iteration
```
bryanwweber: This reduces the residence time at a very slow rate. I've had success using the `np.logspace` function with two ranges, typically something like `np.logspace(0, -3, 50)` and `np.logspace(-3, -5, 100)`. This approach is able to capture the extinction point pretty closely.

interfaces/cython/cantera/examples/reactors/combustor.py
```diff
@@ -1,93 +1,83 @@
...
+
+if __name__ == '__main__':
```
bryanwweber: Same comment here as for the other example - why use this pattern if the expensive code is outside it?

bryanwweber (changes requested):
I like this change, I think it makes a lot of sense. It will be good to have an example of a PSR that people can follow. Just a few style comments.

interfaces/cython/cantera/examples/reactors/combustor.py
bryanwweber: With the specific exponents in `logspace` tuned for that particular reactor, of course.

speth: This method is precisely the same as log spacing, just without requiring you to specify a second endpoint.

interfaces/cython/cantera/examples/reactors/fuel_injection.py
speth: Sigh, caught by my least-favorite PEP8 rule...

interfaces/cython/cantera/examples/reactors/combustor.py
speth: I think I confused myself while writing this. The idea I had was "what if you don't have Matplotlib installed?", but after looking at our other examples, there are quite a few where we just assume it's available. And of course, this doesn't actually help in that case. We do have a couple where we put all of the plotting in a `try / except ImportError` block, so that might be a better option here, or we can just go straight for the plotting.

interfaces/cython/cantera/examples/reactors/combustor.py
bryanwweber: Right, it threw me off because of the `0.9` instead of repeated `0.1` multiplication. My mistake!

interfaces/cython/cantera/examples/reactors/combustor.py
bryanwweber: I say just go for it. People should have matplotlib installed anyways, otherwise, how do they visualize their results? Export to Excel? 🤢

bryanwweber (approved):

PR #560, opened 2018-09-10
Complete migration to Python 3

speth:
Complete migration to Python 3. Removes support for Python 2.7.

This is a work in progress. I'm creating the PR now for the sake of testing the modifications to the Travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,10 +40,7 @@ before_script: |
       pip3 install --user cython
 
       # Install packages for the documentation
-      pip3 install --user sphinx sphinxcontrib-matlabdomain
-      # These two packages have specific bug fixes from these branches
-      pip3 install --user https://github.com/Cantera/doxylink/archive/fix_relative_path_source_14.tar.gz
-      pip3 install --user https://github.com/bryanwweber/sphinxcontrib-katex/archive/fix-options-bug.tar.gz
+      pip3 install --user sphinx sphinxcontrib-matlabdomain sphinxcontrib-doxylink sphinxcontrib-katex
   fi
   rm -f cantera.conf
 script: |
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,7 @@ os:
 addons:
   apt:
     packages:
-      - python-dev
-      - python-numpy
-      - python-pip
+      - python3-pip
       - python3-dev
       - python3-numpy
       - python3-setuptools
@@ -33,28 +31,27 @@ before_script: |
   if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       brew update > /dev/null
       brew upgrade python  # Installs Python 3
-      brew install python@2  # Installs Python 2
       brew install scons  # Install SCons which *should* use Python 3
       pip3 install numpy cython  # Install numpy and Cython for Python 3
-      pip2 install numpy 3to2  # Install numpy and 3to2 for Python 2
       brew install boost
   else
-      pip2 install --user --install-option="--no-cython-compile" cython
-      pip2 install --user 3to2
+      pip3 install --user --upgrade pip
+      pip3 install --user --upgrade setuptools wheel
+      pip3 install --user cython
 
       # Install packages for the documentation
-      pip2 install --user sphinx sphinxcontrib-matlabdomain
+      pip3 install --user sphinx sphinxcontrib-matlabdomain
       # These two packages have specific bug fixes from these branches
-      pip2 install --user https://github.com/Cantera/doxylink/archive/fix_relative_path_source_14.tar.gz
-      pip2 install --user https://github.com/bryanwweber/sphinxcontrib-katex/archive/fix-options-bug.tar.gz
+      pip3 install --user https://github.com/Cantera/doxylink/archive/fix_relative_path_source_14.tar.gz
+      pip3 install --user https://github.com/bryanwweber/sphinxcontrib-katex/archive/fix-options-bug.tar.gz
   fi
   rm -f cantera.conf
 script: |
   if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
       set -e
-      scons build -j2 VERBOSE=y python2_package=full python3_package=full python3_cmd=/usr/bin/python3 blas_lapack_libs=lapack,blas optimize=n coverage=y
+      scons build -j2 python_cmd=/usr/bin/python3 VERBOSE=y python_package=full blas_lapack_libs=lapack,blas optimize=n coverage=y
       scons test
-      scons build sphinx_docs=y doxygen_docs=y
+      scons build sphinx_docs=y doxygen_docs=y sphinx_cmd="/usr/bin/python3 `which sphinx-build`"
       if [[ "${TRAVIS_PULL_REQUEST}" == "false" ]] && [[ "${TRAVIS_BRANCH}" == "master" ]] && [[ "${TRAVIS_REPO_SLUG}" == "Cantera/cantera" ]]; then
         openssl aes-256-cbc -k "${ctdeploy_pass}" -in ./doc/ctdeploy_key.enc -out ./doc/ctdeploy_key -d
         chmod 0600 ./doc/ctdeploy_key
@@ -77,7 +74,7 @@ script: |
         echo "Skipping documentation upload from source other than Cantera/cantera:master"
       fi
   else
-      python3 `which scons` build -j2 VERBOSE=y python2_package=full python3_package=full python2_cmd=/usr/local/opt/python@2/bin/python2 blas_lapack_libs=lapack,blas optimize=n coverage=y
+      python3 `which scons` build -j2 VERBOSE=y python_package=full blas_lapack_libs=lapack,blas optimize=n coverage=y
       python3 `which scons` test
   fi
 after_success: |
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,8 +71,8 @@ script: |
         echo "Skipping documentation upload from source other than Cantera/cantera:master"
       fi
   else
-      python3 `which scons` build -j2 VERBOSE=y python_package=full blas_lapack_libs=lapack,blas optimize=n coverage=y
-      python3 `which scons` test
+      scons build -j2 python_cmd=python3 VERBOSE=y python_package=full blas_lapack_libs=lapack,blas optimize=n coverage=y
+      scons test
   fi
 after_success: |
   if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
--- a/SConstruct
+++ b/SConstruct
@@ -1110,6 +1110,7 @@ if env['VERBOSE']:
 env['python_cmd_esc'] = quoted(env['python_cmd'])
 
 # Python Package Settings
+python_min_version = LooseVersion('3.3')
 cython_min_version = LooseVersion('0.23')
 numpy_min_test_version = LooseVersion('1.8.1')
 
@@ -1176,12 +1177,18 @@ if env['python_package'] in ('full', 'default'):
             print(err, err.output)
         warn_no_python = True
     else:
-        env['python_version'] = info[0]
+        python_version = LooseVersion(info[0])
         numpy_version = LooseVersion(info[1])
         cython_version = LooseVersion(info[2])
         if len(info) > 3:
             print("WARNING: Unexpected output while checking Python / Numpy / Cython versions:")
             print('| ' + '\n| '.join(info[3:]))
+
+        if python_version < python_min_version:
+            print("WARNING: Python is an incompatible version: "
+                "Found {0} but {1} or newer is required".format(
+                python_version, python_min_version))
+            warn_no_python = True
 
         if numpy_version == LooseVersion('0.0.0'):
             print("NumPy not found.")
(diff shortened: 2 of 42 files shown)


.travis.yml
```diff
@@ -33,28 +31,28 @@ before_script: |
...
+      pip3 install --user scons
+      pip3 install --user --install-option="--no-cython-compile" cython
 
       # Install packages for the documentation
-      pip2 install --user sphinx sphinxcontrib-matlabdomain
+      pip3 install --user sphinx sphinxcontrib-matlabdomain
       # These two packages have specific bug fixes from these branches
-      pip2 install --user https://github.com/Cantera/doxylink/archive/fix_relative_path_source_14.tar.gz
-      pip2 install --user https://github.com/bryanwweber/sphinxcontrib-katex/archive/fix-options-bug.tar.gz
+      pip3 install --user https://github.com/Cantera/doxylink/archive/fix_relative_path_source_14.tar.gz
```
bryanwweber: With Python 3, we can use the most recent version of doxylink. Also, the bug in sphinxcontrib-katex has been fixed in the most recent release as well, so we don't need to install from GitHub.

speth:
The CI failure is in building the docs, and is due to the release of Sphinx 1.8.0. This will be fixed pending a release of `sphinxcontrib-katex` which incorporates https://github.com/hagenw/sphinxcontrib-katex/pull/8 or different fix for this problem. Besides that, this is pending your review, @bryanwweber.

.travis.yml
```diff
@@ -77,8 +71,8 @@ script: |
         echo "Skipping documentation upload from source other than Cantera/cantera:master"
       fi
   else
-      python3 `which scons` build -j2 VERBOSE=y python2_package=full python3_package=full python2_cmd=/usr/local/opt/python@2/bin/python2 blas_lapack_libs=lapack,blas optimize=n coverage=y
-      python3 `which scons` test
+      scons build -j2 python_cmd=python3 VERBOSE=y python_package=full blas_lapack_libs=lapack,blas optimize=n coverage=y
```
bryanwweber, line 74: Will this build with Python 2 or Python 3 running SCons? If Python 2 (as I suspect, I think this will pick up the system Python), is that intentional? Should it be intentional?

SConstruct
```diff
@@ -345,81 +343,43 @@ config_options = [
...
-        """Use this option if you want to install the Cantera Python 2 package to
+        """Use this option if you want to install the Cantera Python package to
            an alternate location. On Unix-like systems, the default is the same
            as the 'prefix' option. If the 'python_prefix' option is set to
            the empty string or the 'prefix' option is not set, then the package
            will be installed to the system default 'site-packages' directory.
            To install to the current user's 'site-packages' directory, use
            'python_prefix=USER'.""",
         defaults.python_prefix, PathVariable.PathAccept),
-    EnumVariable(
```
bryanwweber, line 364: This PR is going to need a corresponding PR over at cantera-website to remove the config options and/or update the existing options, depending on how we want to handle documenting building the dev branch.

SConstruct
```diff
@@ -1178,257 +1146,142 @@ if env['python_package'] == 'new':
...
         import sys
         print('{v.major}.{v.minor}'.format(v=sys.version_info))
+        err = ''
+        try:
+            import numpy
+            print(numpy.__version__)
+        except ImportError as np_err:
+            print('0.0.0')
+            err += str(np_err) + '\\n'
+        try:
```
bryanwweber, line 1166: The previous behavior was that Cython was installed for the Python that was running SCons, while Numpy was installed for the Python that would eventually run Cantera. Any reason to change that here?

appveyor.yml
```diff
@@ -1,23 +1,20 @@
...
-    C:\Python27-x64\python.exe -m pip install --upgrade setuptools
-    C:\Python27-x64\python.exe -m pip install --upgrade --no-warn-script-location wheel
-    C:\Python27-x64\Scripts\pip.exe install scons
-    C:\Python27-x64\Scripts\pip.exe install --no-cache-dir --no-warn-script-location numpy
-    C:\Python27-x64\Scripts\pip.exe install --no-warn-script-location cython
-    C:\Python27-x64\Scripts\pip.exe install 3to2
-    C:\Python27-x64\Scripts\pip.exe install pypiwin32
-    C:\Python35-x64\python.exe -m pip install --no-cache-dir --upgrade pip
-    C:\Python35-x64\Scripts\pip.exe install --no-cache-dir --no-warn-script-location numpy
+    C:\Python37-x64\python.exe -m pip install --no-cache-dir --upgrade pip
```
bryanwweber, line 4: 👏 building with 3.7!

interfaces/cython/SConscript
```diff
@@ -135,64 +27,113 @@ for f in ['inputs/h2o2.inp', 'inputs/gri30.inp', 'transport/gri30_tran.dat']:
...
+            localenv.Append(CPPDEFINES='MS_WIN64')
+        # fix for http://bugs.python.org/issue11566
+        localenv.Append(CPPDEFINES={'_hypot':'hypot'})
+elif localenv['OS'] == 'Cygwin':
+    # extract 'pythonX.Y' from 'libpythonX.Y.dll.a'
+    localenv.Append(LIBS=pylib[3:-6])
+
+# Build the Python module
+obj = localenv.SharedObject('#build/temp-py/_cantera', 'cantera/_cantera.cpp')
+ext = localenv.LoadableModule('#build/python/cantera/_cantera%s' % module_ext,
```
bryanwweber: Since we're changing all of this anyways, can we use the `format` method here (and everywhere else), instead of `%` formatting?

interfaces/cython/cantera/onedim.py
```diff
@@ -27,7 +22,7 @@ def __init__(self, domains, gas, grid=None):
         if grid is None:
             grid = np.linspace(0.0, 0.1, 6)
         self.flame.grid = grid
-        super(FlameBase, self).__init__(domains)
+        super().__init__(domains)
```
bryanwweber, line 25: 👏 Here's one of the Python 3 features I didn't think of, that is super useful, if only in making it easier for new people to jump in without having to know which arguments are which, etc.

bryanwweber (changes requested):
This is awesome! A few mostly cosmetic comments here. I want to dig into SConstruct a little more, but I want to open it in a real text editor instead of the diff view here; the diff is too confusing. I'll try to do that this weekend.

This isn't directly related to dropping Python 2, but since we're working on the build script anyways, can you add the home directory expansion for the variables such as `boost_inc_dir`, `python_cmd`, etc.?

SConstruct
bryanwweber: Maybe a separate page with the options for the master branch? Something like,  if you're compiling a tag, go here. If you're compiling master, go here.

.travis.yml
speth: Yes, I believe it will build with the system Python 2.7, and that is intentional. Many users are going to try running `scons build ...` and end up in exactly this situation, so I think it's a good one to test.

SConstruct
speth: Yeah, I think we need a separate page on the website. I would say that the primary page corresponds to the current release, and that on that page there is a link for "If you're compiling the development version..."

.travis.yml
bryanwweber: OK, agreed. Just wanted to make sure! 😄

SConstruct
speth: In the cases where the Python running SCons ends up being different from the Python which will run Cantera, I think it's better to use the latter. If you are using an old, system version of Python to run SCons, it would be confusing to say "install Cython for *this* copy of Python and numpy for *this* copy".

interfaces/cython/SConscript
speth: In places where there are already other changes, sure. Elsewhere I'm inclined to just leave things as-is, especially since we'll just have the urge to replace things with the `f''`-style formatting once we require Python 3.6 or newer.

SConstruct
bryanwweber: I think the reason that Cython had to be installed for the Python running SCons was the `import Cython` stuff in the `interfaces/cython/SConscript` file. If you changed that, 👍 I can't recall, I'll have to go look

interfaces/cython/SConscript
bryanwweber: Agreed, if the line is being changed here, let's update it. If not, it can wait.

SConstruct
speth: Yes, both the check for Cython and the actual cythonization are now done using `$python_cmd`

interfaces/cython/cantera/onedim.py
speth: This is one of my favorite little clean-ups in Python 3, removing the need to specify completely redundant information.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -156,9 +155,7 @@ def test_setCompositionDict(self):
         self.assertNear(X[0], 0.25)
         self.assertNear(X[3], 0.75)
 
-        # Change back to u'xx' syntax when support for Python 3.2 is dropped
-        #self.phase.Y = {u'H2':1.0, u'O2':3.0}
-        self.phase.Y = {'H2':1.0, 'O2':3.0}
+        self.phase.Y = {u'H2':1.0, u'O2':3.0}
```
bryanwweber: I don't think this is necessary; in Python 3, all strings are unicode strings. I think it would only be necessary for Python 2 support, and this appears to be testing the cython interface, and not `ctml_writer`. But this whole unicode/str/bytes business has me very confused...

speth: Oh, good point, the only case where this can/needs to be tested is if we were supporting Python 2.7 Python 3.3+.

SConstruct
speth: New version of the docs with separate pages for both https://testing.cantera.org/compiling/configure-build-dev.html and https://testing.cantera.org/compiling/config-options-dev.html are up on the testing site.

bryanwweber: That looks good to me

SConstruct
```diff
@@ -345,81 +343,43 @@ config_options = [
...
-        defaults.python_prefix, PathVariable.PathAccept),
     EnumVariable(
         'python3_package',
-        """Controls whether or not the Python 3 module will be built. By
-           default, the module will be built if the Python 3 interpreter
-           and the required dependencies (NumPy for Python 3 and Cython
-           for the version of Python for which SCons is installed) can be
-           found.""",
+        """Deprecated synonym for the 'python_package' option. Will be overridden
+        if 'python_package' is set.""",
```
bryanwweber, line 360: We should be consistent about using \`\`option_name`` or 'option_name' in these strings. I would prefer the former, to make auto-generation of the corresponding docs page easier. That said, it seems like most of them use the latter form, and it makes the terminal-formatted `scons help` look nicer.

SConstruct
```diff
@@ -1178,257 +1156,136 @@ if env['python_package'] == 'new':
...
+            print(Cython.__version__)
+        except ImportError as cython_err:
+            print('0.0.0')
+            err += str(cython_err) + '\\n'
+        if err:
+            print(err)
     """)
 
+    if env['python_array_home']:
+        script = "sys.path.append({})\n".format(env['python_array_home']) + script
```
bryanwweber: I think if `python_array_home` is set this will result in a `NameError` because `sys` won't be imported when the `sys.path.append` is called. I think this is probably a mistake I made when I redid this a few months ago.

SConstruct
```diff
@@ -1178,257 +1156,136 @@ if env['python_package'] == 'new':
...
 
-def configure_minimal_python(py_ver):
-    # Test to see if we can run Python if the minimal interface is to be built
+if env['python_package'] in ('full', 'default'):
+    # Test to see if we can import numpy and Cython
     warn_no_python = False
     script = textwrap.dedent("""\
-        from __future__ import print_function
         import sys
         print('{v.major}.{v.minor}'.format(v=sys.version_info))
```
bryanwweber: It looks like this check for Python is duplicated when we check for the `minimal-default` case. Is there a way we can not have to do the check (that's just going to fail again) twice? Can we have a different bool for the case where Python itself can't be found, vs. NumPy or Cython not being found?

SConstruct
bryanwweber: Maybe this is just something that should just be handled when https://github.com/Cantera/cantera-website/issues/26 is resolved

bryanwweber (changes requested):
A couple of small things, once these are fixed/resolved this is 👍 from me

SConstruct
speth: I'm 👍 on deferring this to that other issue, given that ``'option_name'`` is currently used in all but two options.

SConstruct
speth: Digging into it, it's from an earlier change: https://github.com/Cantera/cantera/commit/7d3ebdd3581c06f6294e4fb64e355007c2ea766e, which happens to be before 2.3.0 was released. And it's broken not only because `sys` hasn't been imported, but also because the path isn't quoted.

Which actually provides us with some useful information: nobody uses this option! I was even thinking about whether or not we could drop this, and this is definitive evidence that we can.

bryanwweber: Great! I was wondering if we could drop it too. I suppose its from a time when NumPy wasn't so easy to install, but that time is long gone.

PR #562, opened 2018-09-25
Sphinx docs

band-a-prend:
This pull request offers next improvements:

1. Fix installation of Cantera Sphinx Documentation.

~~2. Rework Cantera 2.4.0 Sphinx Documentation menubar with new one
that now contains only links to local documentation sections. The previous links are intended for cantera website and point to relative paths so they don't work for standalone documentation. The sphinx 'pathto()' variable is used during build of documentation to generate relative paths for links within menubar for each page instead of absolute. The Cantera Sphinx Documentation page with modified menubar now looks like:~~
cantera_sphinx_menu_crop

~~3~~ 2. Add asynchronous loading of remote javascript and css files within Cantera Sphinx Documentation to avoid too long render of pages in case of offline usage of documentation.

Please review this pull request.

### DIFF
--- a/doc/sphinx/_templates/layout.html
+++ b/doc/sphinx/_templates/layout.html
@@ -10,20 +10,24 @@ lang="en">
 {%- macro script() %}
     <script type="text/javascript" id="documentation_options" data-url_root="{{ pathto('', 1) }}" src="{{ pathto('_static/documentation_options.js', 1) }}"></script>
     {%- for scriptfile in script_files %}
-    <script type="text/javascript" src="{{ pathto(scriptfile, 1) }}"></script>
+    {%- if scriptfile.startswith("https://cdn.jsdelivr.net") %}
+        <script defer type="text/javascript" src="{{ pathto(scriptfile, 1) }}"></script>
+    {%- else %}
+        <script type="text/javascript" src="{{ pathto(scriptfile, 1) }}"></script>
+    {%- endif %}
     {%- endfor %}
 {%- endmacro %}
 
 
 {%- macro css() %}
-  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.1.2/css/bootstrap.min.css" integrity="sha384-Smlep5jCw/wG7hdkwQ/Z5nLIefveQRIY9nfy6xoR1uRYBtpZgI6339F5dgvm/e9B" crossorigin="anonymous" />
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.1.2/css/bootstrap.min.css" media="none" onload="this.media='all'" integrity="sha384-Smlep5jCw/wG7hdkwQ/Z5nLIefveQRIY9nfy6xoR1uRYBtpZgI6339F5dgvm/e9B" crossorigin="anonymous" />
   <link rel="stylesheet" href="{{ pathto('_static/' + style, 1) }}" type="text/css" />
   <link rel="stylesheet" href="{{ pathto('_static/pygments.css', 1) }}" type="text/css" />
   {%- for css in css_files %}
     {%- if css|attr("rel") %}
-  <link rel="{{ css.rel }}" href="{{ pathto(css.filename, 1) }}" type="text/css"{% if css.title is not none %} title="{{ css.title }}"{% endif %} />
+  <link rel="{{ css.rel }}" href="{{ pathto(css.filename, 1) }}" media="none" onload="this.media='all'" type="text/css"{% if css.title is not none %} title="{{ css.title }}"{% endif %} />
     {%- else %}
-  <link rel="stylesheet" href="{{ pathto(css, 1) }}" type="text/css" />
+  <link rel="stylesheet" href="{{ pathto(css, 1) }}" media="none" onload="this.media='all'" type="text/css" />
     {%- endif %}
   {%- endfor %}
 {%- endmacro %}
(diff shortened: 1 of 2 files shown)


band-a-prend:
It seems I need to rework part with async loading of javascripts because sometimes "auto-render.min.js" executes early than "katex.min.js" script.

bryanwweber:
@band-a-prend Thanks for this suggestion. I think the change to fix the installation of the docs is valuable.

However, I cannot support changing the menubar for the documentation page. That menubar is needed to integrate with the rest of the Cantera website, and was designed that way intentionally. I might support some sort of conditional in the template, that renders one set of links when a flag is set, if we can find a way that is not complicated to maintain/support. Why do you want to remove the links to the rest of the website? Are the "sidebar" links not sufficient?

doc/sphinx/_templates/layout.html
```diff
@@ -10,20 +10,20 @@
...
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.1.2/css/bootstrap.min.css" media="none" onload="this.media='all'" integrity="sha384-Smlep5jCw/wG7hdkwQ/Z5nLIefveQRIY9nfy6xoR1uRYBtpZgI6339F5dgvm/e9B" crossorigin="anonymous" />
   <link rel="stylesheet" href="{{ pathto('_static/' + style, 1) }}" type="text/css" />
   <link rel="stylesheet" href="{{ pathto('_static/pygments.css', 1) }}" type="text/css" />
   {%- for css in css_files %}
     {%- if css|attr("rel") %}
-  <link rel="{{ css.rel }}" href="{{ pathto(css.filename, 1) }}" type="text/css"{% if css.title is not none %} title="{{ css.title }}"{% endif %} />
+  <link rel="{{ css.rel }}" href="{{ pathto(css.filename, 1) }}" media="none" onload="this.media='all'" type="text/css"{% if css.title is not none %} title="{{ css.title }}"{% endif %} />
     {%- else %}
-  <link rel="stylesheet" href="{{ pathto(css, 1) }}" type="text/css" />
+  <link rel="stylesheet" href="{{ pathto(css, 1) }}" media="none" onload="this.media='all'" type="text/css" />
```
bryanwweber, line 30: Can you please explain what these changes do?

band-a-prend: @bryanwweber 
It seems I fixed async loading in third commit by adding conditional addition of "defer" attribute
that allow now async loading of some remote javascripts in proper order (with direct loading of local javascripts)

As for `media="none" onload="this.media='all'" `:  except of external javascript there is also external css files presented that also result in page load delay if PC is offline. So this temporary disable loading of such css and allow to load page content first.
I refered this article (https://www.filamentgroup.com/lab/async-css.html) when used this method.

As for Menubar, I assumed that sphinx documentation now is fully independent of official site, at least it's "_templates/layout.html". If not then it's really mustn't be changed and second commit is unnecessary. If needed I could change it while packaging of documentation while my local build of sphinx documentation.

Also I still has no idea how to fix one error within Cantera 2.4.0 Sphinx Documentation source code: 
the `sphinx/html/_static/katex_autorenderer.js` results in error
> Uncaught /usr/share/cantera/doc/sphinx/html/_static/katex_autorenderer.js:2 
> SyntaxError: Unexpected token ,

i.e.
```
katex_options = {
,
delimiters: [
        { left: "\\(", right: "\\)", display: false },
        { left: "\\[", right: "\\]", display: true }
        ],
}

document.addEventListener("DOMContentLoaded", function() {
  renderMathInElement(document.body, katex_options);
});
```
shouldn't have comma at line 2  and this comma actually absence in [Cantera- website katex_autorenderer.js](https://github.com/Cantera/cantera-website/blob/master/api-docs/docs-2.4/sphinx/html/_static/katex_autorenderer.js) as also comma at line 6 but it's presence doesn't result in error. So Cantera website renders formulas without error.
I can't find where and how this file is generated in cantera source code.

bryanwweber: This is generated from the [`sphinxcontrib-katex`](https://github.com/hagenw/sphinxcontrib-katex) plugin. That specific error has been fixed in a more recent release of the package, I think 0.3.0. `pip install -U sphinxcontrib-katex` should get you the updated version, but even the newest version doesn't (yet) work with Sphinx 1.8.0, so just make sure you're using 1.7.x

bryanwweber: After reading that site, I'm not convinced that this change should be part of the main Cantera documentation. I recognize this adds some delay to loading the pages when offline, but some of these options are not available in all browsers (from the article "along with additional workarounds for old/still-active browsers that don't support onload events on link elements."). As these docs are primarily meant to be read online in HTML format, I don't think we should include this change either. Is there another way to speed up the loading of the page while offline?

band-a-prend: "Travis CI" check is failed again :(

bryanwweber: Yes, that's because of the error I mentioned about Sphinx 1.8.0

band-a-prend: I made some additional checks. The every page of Shinx documentation render rather fast if PC has unpluged cable or disabled network adapter. If PC has connection to local network with it's own dns-server but without access to Internet then time of rendering of every page is about 1 minute.

By the data of https://caniuse.com/#search=onload the "onload" event isn't supported by IE 8 and earlier version of it but supported since firefox 2, chrome 4, opera 10.1, safari 3.1.

I tested with IE 8 and it just ignore this event and render page within local network without internet as long as without this "onload" event.

Unfortunately I can't find another way to speedup the offline rendering under such conditions. But if this docs, as you mentioned, are primarily to be read online then this commit is unnecessary indeed.

band-a-prend: Thank you, the update of sphinxcontrib-katex from 0.2.0 to 0.3.0 resolves issue with katex_autorenderer.js

speth: Based on that site, I think this feature is standard enough that we could use it.

bryanwweber: OK, I'm convinced :smile: If you drop the commit that changes the menubar, I think this is good to merge. I will just quickly check the install step locally. Thank you!

band-a-prend: I rebased pull request with removing second commit that changed the menubar.

band-a-prend: Forgot to mention that I used namely `defer` attribute instead of `async="async"` for some remote javascripts because they should be loaded consistently. And local scripts have no any such attributes.

PR #563, opened 2018-09-27
BinarySolutionTabulatedThermo Class

benjaminkee:
- Thermo class ConstDensityTabulatedThermo
- Thermo class IdealSolidSolnTabulatedThermo
- Test files for both thermo classes


speth:
Thanks for this PR! I'm glad to see this getting into shape. As a first step toward getting this PR merged, could you revert the merge commit and rebase this onto the current `master` instead?

speth:
Sorry, I should perhaps not have said 'revert'. What I really mean is to reset your branch to the state before the merge commit, e.g. `git reset --hard 2a0838e`. Then you should be able to run `git rebase upstream/master` (replacing `upstream` with whatever you have named the `Cantera/cantera` remote) and then do a force push.

benjaminkee:
Ah ok, thanks for the git tips! I just pushed now with the instructions provided, with the exception of the reset commit, which was changed to one later commit, to include the test suite files.

benjaminkee:
The Linux part of the travis check is failing due to 
"Exception occurred:
  File "/home/travis/.local/lib/python3.4/site-packages/docutils/nodes.py", line 567, in __getitem__
    return self.attributes[key]
KeyError: 'latex'" 
Any suggestion on how to troubleshoot this?

bryanwweber:
@lionkey This is due to an error in a dependency that has been fixed if you rebase your branch onto the current tip of the master branch. The command @speth gave you in his last comment should do it.

benjaminkee:
Thanks for the help! Looks like all tests pass now.

include/cantera/thermo/ConstDensityTabulatedThermo.h
```diff
@@ -0,0 +1,113 @@
+/*
+ * ConstDensityTabulatedThermo.h
+ *
+ *  Created on: 04.06.2016
```
bryanwweber: Please do not add author information to each file; instead, add your name(s) to the `AUTHORS` file in the root of the repository.

include/cantera/thermo/ConstDensityTabulatedThermo.h
```diff
@@ -0,0 +1,113 @@
...
+/**
+ *  @file ConstDensityTabulatedThermo.h
+ * Definition file for a derived class of ThermoPhase that assumes either
+ * an ideal gas or ideal solution approximation and handles
+ * variable pressure standard state methods for calculating
+ * thermodynamic properties (see \ref thermoprops and
+ * class \link Cantera::ConstDensityTabulatedThermo ConstDensityTabulatedThermo\endlink).
+ */
+/*
+ * Copyright (2005) Sandia Corporation. Under the terms of
```
bryanwweber: Please see one of the [existing header files](https://github.com/Cantera/cantera/blob/d7292b33459e64921074d75c92e0cd8faf37a31b/include/cantera/Edge.h#L3) for how to specify the copyright/license information.

include/cantera/thermo/IdealSolidSolnPhaseTabulatedThermo.h
```diff
@@ -0,0 +1,135 @@
...
+ * IdealSolidSolnPhaseTabulatedThermo.h
+ *
+ *  Created on: 04.06.2016
+ *      Author: mmayur
+ */
+
+/**
+ *  @file IdealSolidSolnPhaseTabulatedThermo.h
+ * Definition file for a derived class of ThermoPhase that assumes either
+ * an ideal gas or ideal solution approximation and handles
```
bryanwweber: If this is for an ideal solid, the documentation should be updated to reflect that it is not a gas

include/cantera/thermo/IdealSolidSolnPhaseTabulatedThermo.h
```diff
@@ -0,0 +1,135 @@
...
+     *  derived from ThermoPhase even if the application only has
+     *  a pointer to ThermoPhase to work with.
+     */
+    virtual ThermoPhase* duplMyselfAsThermoPhase() const;
+
+    virtual std::string type() const {
+        return "IdealSolidSolutionTabultedThermo";
+    }
+
+    //! Was added because of the DENIS LFP Runaway simulation.
```
bryanwweber: Can you please explain a little more what this comment means? Remember that the comments should make sense to a developer who hasn't looked at the code before, but who needs to maintain it. What is the DENIS LFP Runaway simulation? Why is it important? How does this function help?

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -798,6 +809,27 @@ def build(self, t):
         u["size"] = "9"
         u["name"] = "coeffs"
 
+class standardState(object):
+    """Base class for species standard-state properties."""
+    def _build(self, p):
+        return p.addChild("standardState")
+    def export(self, f, fmt = 'CSV'):
```
bryanwweber: Keyword arguments in function definitions in Python should not have spaces around the `=`. There are a few places where this happens in the code you're adding (no need to fix existing code), and if you're adding to an existing function that has spaces around the `=`, then you should match that style.

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -798,6 +809,27 @@ def build(self, t):
...
+    def __init__(self,
+                 molarVolume = 0.0):
+        """
+        :param molarVolume:
+            Reference-state molar volume. Default: 0.0.
+        """
+        self._mv = molarVolume
+    def build(self, ss):
+        #t = self._build(p)
+        ss['model']='constant_incompressible'
```
bryanwweber: Please don't commit commented code. Also, there should be spaces on both sides of this `=`.

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -2163,6 +2199,93 @@ def build(self, p):
...
+                 data_file = '',
+                 options = []):
+        phase.__init__(self, name, 3, elements, species, note, 'None',
+                       initial_state, options)
+        self._dens = density
+        self._pure = 0
+        self._modSpecies = modifiable_species
+        self._data = data_file
+        if self._dens is None:
+            raise CTI_Error('In phase '+name+': density must be specified.')
```
bryanwweber: There should be spaces around the `+` here and below as well.

src/thermo/ConstDensityTabulatedThermo.cpp
```diff
@@ -0,0 +1,203 @@
+/**
+ *  @file ConstDensityTabulatedThermo.cpp
+ * Definition file for a derived class of ThermoPhase that assumes either
+ * an ideal gas or ideal solution approximation and handles
```
bryanwweber: If the phase is constant density, it cannot be an ideal gas.

src/thermo/ConstDensityTabulatedThermo.cpp
```diff
@@ -0,0 +1,203 @@
...
+void ConstDensityTabulatedThermo::_updateThermo()
+{
+	doublereal tnow = temperature();
+	doublereal xnow = moleFraction(m_kk_mod);
+	doublereal c[4];
+	doublereal dS_corr = 0.0;
+	doublereal tlow = 0.0, thigh = 0.0;
+	int type = 0;
+	if (m_tlast != tnow || m_xlast != xnow) {
+		c[0]=tnow;
```
bryanwweber: Please be consistent with spaces around `=` in the code here. I think the preference in C++ code is to include the spaces.

src/thermo/ConstDensityTabulatedThermo.cpp
```diff
@@ -0,0 +1,203 @@
...
+	doublereal a=0,b=0,c=0;
+	std::ifstream t;
+	std::string line;
+	int count = 0;
+
+	if (phaseNode.hasChild("thermo")) {
+		XML_Node& thermoNode = phaseNode.child("thermo");
+		std::string model = thermoNode["model"];
+		if (model == "ConstDensityTabulatedThermo") {
+			if (thermoNode.hasChild("data")) {
```
bryanwweber: Please use spaces and not tabs to indent the code. Four spaces = one tab

src/thermo/ConstDensityTabulatedThermo.cpp
```diff
@@ -0,0 +1,203 @@
...
+}
+
+doublereal ConstDensityTabulatedThermo::interp_h(doublereal x) const
+{
+	// Assumes that "table" is sorted by .first
+	// Check if x is out of bound
+	if (x > molefrac_h.back().first) return molefrac_h.back().second;
+	if (x < molefrac_h[0].first) return molefrac_h[0].second;
+	std::vector<std::pair<double, double> >::const_iterator it, it2;
+	// INFINITY is defined in math.h in the glibc implementation
```
bryanwweber: I don't see how this comment is related to the adjacent lines of code. Can you explain a bit more please?

test/data/IdealSolidSolnPhaseTabulatedThermo.cti
```diff
@@ -0,0 +1,31 @@
+#=====================================================================================
+#   Electrodes
+#=====================================================================================
+
+# Graphite (anode)
+#------------------------------------------------------------------
+IdealSolidSolutionTabulatedThermo(
+                                  name = "anode",
```
bryanwweber: This line, and the following lines, can be indented by only 4 spaces to limit the column width

bryanwweber (changes requested):
Hello! :wave: Thank you again for making this pull request! I've gone through and made some stylistic comments about the code and the formatting. In general, things look pretty good, but there are some things to change to make your code consistent with the rest of Cantera. Take a look at our [Contributing guide](https://github.com/Cantera/cantera/blob/master/CONTRIBUTING.md) for more information, in addition to my comments here.

It would be super if you could add an example of how to use this new functionality, aside from just the tests - what motivated you to add these classes? What kind of problems are you solving?

include/cantera/thermo/ConstDensityTabulatedThermo.h
```diff
@@ -0,0 +1,113 @@
...
+
+namespace Cantera
+{
+
+//!   Overloads the virtual methods of class ThermoPhase to implement the
+//!   incompressible equation of state.
+/**
+ * <b> Specification of Solution Thermodynamic Properties </b>
+ *
+ * The density is assumed to be constant, no matter what the concentration of the solution.
```
speth: The class documentation should be explain the model that you're implementing, rather than just duplicating that of the parent class. I would suggest that this include some background on what the model is for, what its mathematical background is, what assumptions and limitations of the model are, and what data users need to specify in order to use it.

include/cantera/thermo/ConstDensityTabulatedThermo.h
```diff
@@ -0,0 +1,113 @@
...
+	 * species have been added. For example, it might be used to
+	 * resize internal work arrays that must have an entry for
+	 * each species.  The base class implementation does nothing,
+	 * and subclasses that do not require initialization do not
+	 * need to overload this method.  When importing a CTML phase
+	 * description, this method is called from ThermoPhase::initThermoXML(),
+	 * which is called from importPhase(),
+	 * just prior to returning from function importPhase().
+	 */
+	virtual void initThermoXML(XML_Node& phaseNode, const std::string& id_);
```
speth: For functions where there is no difference in the docstring from that of the parent class, it should be omitted, e.g. as in those for the following functions.

include/cantera/thermo/ConstDensityTabulatedThermo.h
```diff
@@ -0,0 +1,113 @@
...
+	virtual void setMoleFractions(const doublereal* const x);
```
speth: Please use the built-in name `double` rather than the typedef `doublereal`.

test/data/Anode_Graphite.dat
```diff
@@ -0,0 +1,108 @@
...
+*
+*Source:
+*
+*dH: from Delacourt yet unpublished 2011
+*dS: Reynier 2004 JElchemSoc - Thermodynamics of Lithium Intercalation into Graphites and Disordered Carbons
+*original range: 0.01 - 0.8
+* fit of original data
+*
+*Author: Wolfgang Mielke 21.03.2011
+*File format: '*' starts a comment line. Values may be separated by one or multiple CR, TAB or SPACE.
```
speth: I know the CTI/XML input file duality is kind of a pain, but I think the data here should be included in those input files, rather than introducing a new, ad hoc format just for this class. I would suggest taking a look at the additional pure fluid parameters are added in the Redlich-Kwong model as an example of how to include extra data and get that in.

test/thermo/ConstDensityTabulatedThermo_Test.cpp
```diff
@@ -0,0 +1,87 @@
...
+       	-1523717.129950,
+       	-2332251.932434,
+	    -3630175.923912,
+	    -4707368.353702,
+	    -5177756.746737,
+	    -4674479.634792,
+	    -5341855.069058,
+	    -5661616.443350,
+	    -4929484.137938
+    };
```
speth: Is there a basis for these expected results, or are these purely a regression test? It's fine either way, but a comment explaining where these come from would be helpful.

test/thermo/IdealSolidSolnPhaseTabulatedThermo_Test.cpp
```diff
@@ -0,0 +1,168 @@
...
+    {
+        set_defect_X(xmin + i*dx);
+        test_phase->getPartialMolarEntropies(&partialMolarEntropies[0]);
+        EXPECT_NEAR(expected_result[i], partialMolarEntropies[0], 1.e-6);
+    }
+
+}
+
+
+}
```
speth: There should be a newline at the end of the file.

test/thermo/IdealSolidSolnPhaseTabulatedThermo_Test.cpp
```diff
@@ -0,0 +1,168 @@
...
+       - 9019730.125078,
+       - 6917660.798931
+    };
+
+    double xmin = 0.10;
+    double xmax = 0.75;
+    int numSteps= 9;
+    double dx = (xmax-xmin)/(numSteps-1);
+    vector_fp chemPotentials(2);
+    for(int i=0; i < 9; ++i)
```
speth: add a space between `for` and `(` (since `for` is a keyword, not a function).

There should also be spaces on each side of the `=`.

include/cantera/thermo/IdealSolidSolnPhaseTabulatedThermo.h
```diff
@@ -0,0 +1,135 @@
...
+ */
+#ifndef CT_IDEALSOLIDSOLNPHASETABULATEDTHERMO_H
+#define CT_IDEALSOLIDSOLNPHASETABULATEDTHERMO_H
+
+#include "IdealSolidSolnPhase.h"
+#include "cantera/base/utilities.h"
+#include <vector>
+#include <iterator>
+#include <iostream>
+#include <fstream>
```
speth: I believe most, if not all, of these standard library includes are unnecessary, as they will implicitly included from other Cantera headers which include `ct_defs.h`.

include/cantera/thermo/IdealSolidSolnPhaseTabulatedThermo.h
```diff
@@ -0,0 +1,135 @@
...
+     */
+	IdealSolidSolnPhaseTabulatedThermo& operator=(const IdealSolidSolnPhaseTabulatedThermo& right);
+
+    //! Duplication routine for objects which inherit from ThermoPhase
+    /*!
+     *  This virtual routine can be used to duplicate objects
+     *  derived from ThermoPhase even if the application only has
+     *  a pointer to ThermoPhase to work with.
+     */
+    virtual ThermoPhase* duplMyselfAsThermoPhase() const;
```
speth: This method, along with the copy constructor and assignment operator, were deprecated in Cantera 2.3.0 and removed in Cantera 2.4.0, and so should not be implemented in any new classes.

src/thermo/IdealSolidSolnPhaseTabulatedThermo.cpp
```diff
@@ -0,0 +1,332 @@
...
+	 */
+	XML_Node& speciesList = phaseNode.child("speciesArray");
+	XML_Node* speciesDB = get_XML_NameID("speciesData", speciesList["datasrc"],
+			&phaseNode.root());
+
+	for (size_t k = 0; k < m_kk; k++) {
+		XML_Node* s =  speciesDB->findByAttr("name", speciesName(k));
+		XML_Node* ss = s->findByName("standardState");
+		m_speciesMolarVolume[k] = getFloat(*ss, "molarVolume", "toSI");
+	}
```
speth: Is the species molar volume used in this model? If not, this can simply be removed.

src/thermo/IdealSolidSolnPhaseTabulatedThermo.cpp
```diff
@@ -0,0 +1,332 @@
...
+		} else if (caseInsensitiveEquals(formString, "solvent_volume")) {
+			m_formGC = 2;
+		} else {
+			throw CanteraError("IdealSolidSolnPhaseTabulatedThermo::initThermoXML",
+					"Unknown standardConc model: " + formString);
+		}
+	} else {
+		throw CanteraError("IdealSolidSolnPhaseTabulatedThermo::initThermoXML",
+				"Unspecified standardConc model");
+	}
```
speth: Is the `m_formGC` variable used in this model? If not, this should be removed.

include/cantera/thermo/IdealSolidSolnPhaseTabulatedThermo.h
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * The density is assumed to be constant, no matter what the concentration of the solution.
+ *
+ * @ingroup thermoprops
+ */
+class IdealSolidSolnPhaseTabulatedThermo : public IdealSolidSolnPhase
+{
+public:
+    //! Constructor.
+	IdealSolidSolnPhaseTabulatedThermo() {}
```
speth: If the default constructor does nothing, it does not need to be explicitly implemented.

include/cantera/thermo/IdealSolidSolnPhaseTabulatedThermo.h
```diff
@@ -0,0 +1,135 @@
...
+    virtual void setMassFractions(const doublereal* const y);
+    virtual void setMassFractions_NoNorm(const doublereal* const y);
+    virtual void setConcentrations(const doublereal* const conc);
+
+    //! Species thermodynamics interpolation functions
+	doublereal interp_h(doublereal x) const;
+	doublereal interp_s(doublereal x) const;
+
+    //! Current modifiable species index
+    size_t m_kk_mod;
```
speth: The `m_kk_mod` variable, along with the `interp_h` and `interp_s` functions, should be `protected`, not `public`.

src/thermo/ConstDensityTabulatedThermo.cpp
```diff
@@ -0,0 +1,203 @@
...
+#include "cantera/base/ctml.h"
+
+namespace Cantera
+{
+
+void ConstDensityTabulatedThermo::setMoleFractions(const doublereal* const x)
+{
+	Phase::setMoleFractions(x);
+	_updateThermo();
+}
```
speth: Rather than introducing the `_updateThermo` method and having to add all of these wrapper functions for the composition-setting functions, you can put the actions in the `_updateThermo` method in an override of the `compositionChanged` function from the `Phase` class, which is automatically called when any of these composition-setting functions is called.

include/cantera/thermo/IdealSolidSolnPhaseTabulatedThermo.h
```diff
@@ -0,0 +1,135 @@
...
+    //! Duplication routine for objects which inherit from ThermoPhase
+    /*!
+     *  This virtual routine can be used to duplicate objects
+     *  derived from ThermoPhase even if the application only has
+     *  a pointer to ThermoPhase to work with.
+     */
+    virtual ThermoPhase* duplMyselfAsThermoPhase() const;
+
+    virtual std::string type() const {
+        return "IdealSolidSolutionTabultedThermo";
```
speth: "Tabulated" is misspelled.

test/data/ConstDensityTabulatedThermo.cti
```diff
@@ -0,0 +1,25 @@
...
+
+# Graphite (anode)
+#------------------------------------------------------------------
+ConstDensityTabulatedThermo(
+	name = "anode",
+	elements = "Li C",
+	species = "Li[anode] V[anode]",  # Lithium atoms, vacancies, and electrons
+	density = (2541, 'kg/m3'),#(2420, 'kg/m3'),
+	initial_state = state( mole_fractions = "Li[anode]:0.5 V[anode]:0.5"),
+    modifiable_species = "Li[anode]",
```
speth: The terminology "modifiable species" leaves a bit to be desired. Perhaps "tabulated species" would be a bit more clear?

src/thermo/ConstDensityTabulatedThermo.cpp
```diff
@@ -0,0 +1,203 @@
...
+	std::sort(molefrac_s.begin(), molefrac_s.end());
+	t.close();
+	ThermoPhase::initThermoXML(phaseNode, id_);
+}
+
+doublereal ConstDensityTabulatedThermo::interp_h(doublereal x) const
+{
+	// Assumes that "table" is sorted by .first
+	// Check if x is out of bound
+	if (x > molefrac_h.back().first) return molefrac_h.back().second;
```
speth: We do not use the single-line `if` statement. This should be written as
```c++
if (x > molefrac_h.back().first) {
    return molefrac_h.back().second;
}
```

src/thermo/ConstDensityTabulatedThermo.cpp
```diff
@@ -0,0 +1,203 @@
...
+	if (x < molefrac_h[0].first) return molefrac_h[0].second;
+	std::vector<std::pair<double, double> >::const_iterator it, it2;
```
speth: Since we require a compiler supporting C++11, there is no need for the space between the two closing angle brackets.

src/thermo/ConstDensityTabulatedThermo.cpp
```diff
@@ -0,0 +1,203 @@
...
+	t.close();
+	ThermoPhase::initThermoXML(phaseNode, id_);
+}
+
+doublereal ConstDensityTabulatedThermo::interp_h(doublereal x) const
+{
+	// Assumes that "table" is sorted by .first
+	// Check if x is out of bound
+	if (x > molefrac_h.back().first) return molefrac_h.back().second;
+	if (x < molefrac_h[0].first) return molefrac_h[0].second;
```
speth: One nice way to avoid all of these special cases in the linear interpolation is to augment the original vector with values at +/- infinity. For example, if your original data were `[(0.1, a), (0.2, b), (0.5, c)]`, store this as `[(-inf, a), (0.1, a), (0.2 b), (0.5, c), (inf, c)]`. This is the approach taken in the interpolation for `Plog` reaction rates.

src/thermo/ConstDensityTabulatedThermo.cpp
```diff
@@ -0,0 +1,203 @@
...
+	std::vector<std::pair<double, double> >::const_iterator it, it2;
+	// INFINITY is defined in math.h in the glibc implementation
+	it = std::lower_bound(molefrac_h.begin(), molefrac_h.end(), std::make_pair(x, molefrac_h[0].second));
+	// Corner case
+	if (it == molefrac_h.begin()) return it->second;
+	it2 = it;
+	--it2;
+	return it2->second + (it->second - it2->second)*(x - it2->first)/(it->first - it2->first);
+}
```
speth: This linear interpolation algorithm is repeated 4 times, which really suggests that it needs to be refactored into a standalone function. I am also wondering whether it would be simpler / more clear to just store three vectors: mole fractions, enthalpies, and entropies. In that case, you could just calculate
```c++
size_t i = std::distance(molefrac.begin(), std::lower_bound(molefrac.begin(), molefrac.end(), x));
```
and do the interpolation of h and s between elements `i` and `i-1`.

include/cantera/thermo/IdealSolidSolnPhaseTabulatedThermo.h
```diff
@@ -0,0 +1,135 @@
...
+
+    //! Current modifiable species mole fraction
+    mutable doublereal m_xlast;
+
+    //! File name for IdealSolidSolnPhaseTabulatedThermo thermo data
+    std::string m_dataFile;
+
+    //! Vector pairs for IdealSolidSolnPhaseTabulatedThermo thermo
+    std::vector<std::pair<doublereal,doublereal> > molefrac_h;
+    std::vector<std::pair<doublereal,doublereal> > molefrac_s;
```
speth: member variable names should be prefixed with `m_`.

src/thermo/ConstDensityTabulatedThermo.cpp
```diff
@@ -0,0 +1,203 @@
...
+		} else {
+			dS_corr = GasConstant*std::log(xnow/(1.0-xnow));
+		}
+		c[2]=interp_s(xnow) * 1e3 + dS_corr; // 1e3 for conversion J/K/mol -> J/K/kmol
+		c[3]=0.0;
+		type = m_spthermo.reportType(m_kk_mod);
+		tlow = m_spthermo.minTemp(m_kk_mod);
+		thigh = m_spthermo.maxTemp(m_kk_mod);
+		shared_ptr<SpeciesThermoInterpType> stit(
+				newSpeciesThermoInterpType(type, tlow, thigh, OneAtm, c));
```
speth: Rather than recreating the `SpeciesThermoInterpType` object every time this method is called, which is rather inefficient, I would recommend leaving these objects unmodified, relying on them only to represent the temperature-dependence of the pure species thermodynamics, and add the tabulated contributions to `m_h0_RT` ant `m_s0_R` separately here.

include/cantera/thermo/IdealSolidSolnPhaseTabulatedThermo.h
```diff
@@ -0,0 +1,135 @@
...
+//!   Overloads the virtual methods of class ThermoPhase to implement the
+//!   incompressible equation of state.
+/**
+ * <b> Specification of Solution Thermodynamic Properties </b>
+ *
+ * The density is assumed to be constant, no matter what the concentration of the solution.
+ *
+ * @ingroup thermoprops
+ */
+class IdealSolidSolnPhaseTabulatedThermo : public IdealSolidSolnPhase
```
speth: In what way is this phase type ideal? I think we should consider a different name for this class that better reflects the model being implemented.

include/cantera/thermo/IdealSolidSolnPhaseTabulatedThermo.h
```diff
@@ -0,0 +1,135 @@
...
+    //! File name for IdealSolidSolnPhaseTabulatedThermo thermo data
+    std::string m_dataFile;
+
+    //! Vector pairs for IdealSolidSolnPhaseTabulatedThermo thermo
+    std::vector<std::pair<doublereal,doublereal> > molefrac_h;
+    std::vector<std::pair<doublereal,doublereal> > molefrac_s;
+
+    //! Function to update the reference state thermo functions
+    //! Changed from private to protected so that the methods of the parent class could use this overridden function
+    void _updateThermo();
```
speth: You can't change a function's private/protected status by changing where it is declared in a derived class. This would need to be changed in the parent class. In addition, docstring comments are not the right place to document changes to the code -- no one reading the code later should have to understand the history of how this was previously structured. Instead, such comments are appropriate as part of the commit message.

interfaces/cython/cantera/ctml_writer.py
speth: What is this method for? It seems to be an unused stub.

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -798,6 +809,27 @@ def build(self, t):
         u["size"] = "9"
         u["name"] = "coeffs"
 
+class standardState(object):
+    """Base class for species standard-state properties."""
+    def _build(self, p):
+        return p.addChild("standardState")
```
speth: Does this method get called anywhere? All I see is the commented-out call from `constantIncompressible.build`.

src/thermo/ConstDensityTabulatedThermo.cpp
```diff
@@ -0,0 +1,203 @@
...
+	if (m_tlast != tnow || m_xlast != xnow) {
+		c[0]=tnow;
+		c[1]=interp_h(xnow) * 1e3; // 1e3 for conversion J/mol -> J/kmol
+		if (xnow==0) {
+			dS_corr = -BigNumber;
+		} else if (xnow==1) {
+			dS_corr = BigNumber;
+		} else {
+			dS_corr = GasConstant*std::log(xnow/(1.0-xnow));
+		}
```
speth: Why does the entropy of the "modifiable species" go to infinity as X->1? Shouldn't it just go to the pure species value, which should be well-defined?

src/thermo/IdealSolidSolnPhaseTabulatedThermo.cpp
```diff
@@ -0,0 +1,332 @@
...
+}
+
+void IdealSolidSolnPhaseTabulatedThermo::getPartialMolarEnthalpies(doublereal* result) const {
+	doublereal vdp = (pressure() - m_spthermo.refPressure()) / molarDensity();
+	doublereal rt = temperature() * GasConstant;
+	const vector_fp& h_RT = enthalpy_RT_ref();
+	for (size_t k = 0; k < m_kk; k++) {
+		result[k] = rt*h_RT[k] + vdp;
+	}
+}
```
speth: Why is this different from the corresponding method in `IdealSoldSolnPhase`? Why do we add a "v dP" term here, but not for that model?

src/thermo/IdealSolidSolnPhaseTabulatedThermo.cpp
```diff
@@ -0,0 +1,332 @@
...
+void IdealSolidSolnPhaseTabulatedThermo::getPartialMolarEntropies(doublereal* result) const {
+	doublereal xx;
+	doublereal r = GasConstant;
+	const vector_fp& s_R = entropy_R_ref();
+	for (size_t k = 0; k < m_kk; k++) {
+		xx = moleFraction(k);
+		if(xx < SmallNumber) xx = SmallNumber;
+		result[k] = r*(s_R[k] - log(xx));
+	}
+}
```
speth: How is this functionally different from the corresponding method in `IdealSolidSolnPhase`?

test/data/Anode_Graphite.dat
```diff
@@ -0,0 +1,108 @@
...
+*
+*columns:		
+*Stoichiometry x [], dH [J/mol],  dS [J/mol/K]
```
speth: By default, units in Cantera should be J/kmol and J/kmol/K, with the ability to specify alternate units on a case-specific basis.

speth (changes requested):
I have a few overarching questions/suggestions for this PR, as well as some more detail-oriented issues which will ultimately need to be addressed. I would suggest we address these questions first, before worrying about some of the implementation details.

- Is there a need for introducing two classes here? What does the `ConstDensityTabulatedThermo` class do that the `IdealSolidSolnPhaseTabulatedThermo` class can't?
- Can you provide a physical/mathematical description of how this model is meant to work? There are a few things that aren't quite clear when trying to work backward from the implementation.
- I would strongly prefer the input format for the new model to make use of the existing CTI/XML input file formats, rather than introducing a new, ad hoc format. It should also be possible to provide all necessary inputs by calling functions in the C++ code, rather than needing to use any files. I am happy to help with this if need be.

manikmayur:
We've updated this pull request to reflect the changes brought up by @bryanwweber and @speth.
-  The scientific relevance of the new thermodynamic class is to model lithium-ion battery intercalation materials. The need for this class is to be able to use experimentally-obtained half-cell data as function of lithium mole fraction as tabulated input for standard-state thermodynamics. We (Mayur, DeCaluwe, Kee, Kee, Bessler) are currently preparing a journal manuscript demonstrating the capabilities of the new class.
-  The tabulated thermo class has been condensed into a single class called BinarySolutionTabulatedThermo.
-  We've cleaned up the formatting issues.
-  The data input was modified to accept the inline format (CTI level and XML level) instead of another file. Optionally at the CTI level, an CSV file can still be read; but this option is not used in the test suite.
- The tabulated species is now referred to as the tabulated species instead of the modifiable species.

Please do let us know if any further changes are required.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -0,0 +1,159 @@
...
+public:
+    /**
+     * Constructor for BinarySolutionTabulatedThermo.
+     * The generalized concentrations can have three different forms
+     * depending on the value of the member attribute #m_formGC, which
+     * is supplied in the constructor or read from the XML data file.
+     *
+     * @param formCG This parameter initializes the #m_formGC variable.
+     */
+	BinarySolutionTabulatedThermo(int formCG=0);
```
speth: I would eliminate the `m_formGC` argument to all of the constructors. The lower-maintenance method of dealing with this setting is to just call the `setStandardConcentrationModel` method which is inherited from the `IdealSolidSolnPhase` class.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -0,0 +1,159 @@
...
+     * subclasses to perform any initialization required after all
+     * species have been added. For example, it might be used to
+     * resize internal work arrays that must have an entry for
+     * each species.  The base class implementation does nothing,
+     * and subclasses that do not require initialization do not
+     * need to overload this method.  When importing a CTML phase
+     * description, this method is called from ThermoPhase::initThermoXML(),
+     * which is called from importPhase(),
+     * just prior to returning from function importPhase().
+     */
```
speth: This docstring is just copied from `ThermoPhase::initThermo` and should be deleted.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -0,0 +1,159 @@
...
+    virtual void initThermoXML(XML_Node& phaseNode, const std::string& id_);
+
+protected:
+
+    /**
+    * The standard concentrations can have one of three different forms:
+    * 0 = 'unity', 1 = 'molar_volume', 2 = 'solvent_volume'. See
+    * setStandardConcentrationModel().
+    */
+    int m_formGC;
```
speth: The base class `IdealSolidSolnPhase` already has this member variable, so this declaration is not needed (and actually shadows the variable of the parent class in a way that is very error-prone).

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+    * Value of the reference pressure for all species in this phase. The T
+    * dependent polynomials are evaluated at the reference pressure. Note,
+    * because this is a single value, all species are required to have the same
+    * reference pressure.
+    */
+    double m_Pref;
```
speth: Again, shadows the corresponding base class member variable.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -0,0 +1,159 @@
...
+    * The density variable which is inherited as part of the State class,
+    * m_dens, is always kept current whenever T, P, or X[] change.
+    */
+    double m_Pcurrent;
+
+    //! Vector of molar volumes for each species in the solution
+    /**
+    * Species molar volumes \f$ m^3 kmol^-1 \f$
+    */
+    vector_fp m_speciesMolarVolume;
```
speth: Ditto for this and `m_Pcurrent`

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -0,0 +1,230 @@
...
+double* BinarySolutionTabulatedThermo::interpolate(double x) const
+{
+	static double c[2];
+    // Assumes that "table" is sorted by .first
+    // Check if x is out of bound
+    if (x > m_molefrac_tab.back()) {
+    	c[0] = m_enthalpy_tab.back();
+    	c[1] = m_entropy_tab.back();
+    	return c;
+    }
```
speth: Indentation should be 4 spaces per level, so the body of this `if` should be indented a total of 8 spaces.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -0,0 +1,230 @@
...
+            std::string tabulated_species_name = speciesNode["name"];
+            m_kk_tab = speciesIndex(tabulated_species_name);
+            if (m_kk_tab == npos) {
+                throw CanteraError("BinarySolutionTabulatedThermo::initThermoXML",
+                        "Species " + tabulated_species_name + " not found.");
+            }
+            m_xlast = moleFraction(m_kk_tab);
+        }
+        if (thermoNode.hasChild("tabulatedThermo")) {
+        	XML_Node& dataNode = thermoNode.child("tabulatedThermo");
```
speth: Rogue tab on this line, and a few others.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -0,0 +1,230 @@
...
+    /*
+     * Call the base initThermo, which handles setting the initial
+     * state.
+     */
+    ThermoPhase::initThermoXML(phaseNode, id_);
+}
+
+double* BinarySolutionTabulatedThermo::interpolate(double x) const
+{
+	static double c[2];
```
speth: non-const static variables are not thread safe. The better solution is to just create and return a `std::pair<double, double>`.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -0,0 +1,230 @@
...
+    	return c;
+    }
+    if (x < m_molefrac_tab[0]) {
+    	c[0] = m_enthalpy_tab[0];
+    	c[1] = m_entropy_tab[0];
+    	return c;
+    }
+    size_t i = std::distance(m_molefrac_tab.begin(), std::lower_bound(m_molefrac_tab.begin(), m_molefrac_tab.end(), x));
+    c[0] = m_enthalpy_tab[i-1] + (m_enthalpy_tab[i] - m_enthalpy_tab[i-1]) * (x - m_molefrac_tab[i-1])/(m_molefrac_tab[i]- m_molefrac_tab[i-1]);
+    c[1] = m_entropy_tab[i-1] + (m_entropy_tab[i] - m_entropy_tab[i-1]) * (x - m_molefrac_tab[i-1])/(m_molefrac_tab[i]- m_molefrac_tab[i-1]);
```
speth: Please wrap these lines, ideally to 80 characters maximum.

test/thermo/BinarySolutionTabulatedThermo_Test.cpp
```diff
@@ -0,0 +1,170 @@
...
+        3852.587527,
+	    5260.898245,
+	    5764.709566,
+	    7786.429343,
+	   10411.473830,
+	   15276.785622,
+	   17900.243026,
+	   22085.482446,
+	   25989.143405
+    };
```
speth, line 72: Lots of rogue tabs in here.

test/thermo/BinarySolutionTabulatedThermo_Test.cpp
```diff
@@ -0,0 +1,170 @@
...
+        set_defect_X(xmin + i*dx);
+        EXPECT_NEAR(expected_result[i], test_phase->enthalpy_mole(), 1.e-6);
+    }
+
+}
+
+TEST_F(BinarySolutionTabulatedThermo_Test,interp_s)
+{
+
+	test_phase->setState_TP(298.15, 101325.);
```
speth: Please remove unnecessary blank lines, e.g. at the start and end of function bodies.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -0,0 +1,159 @@
...
+    virtual void compositionChanged();
+
+    //! Species thermodynamics interpolation functions
+    double* interpolate(double x) const;
+
+    //! Current modifiable species index
+    size_t m_kk_tab;
+
+    //! Current modifiable species mole fraction
+    mutable double m_xlast;
```
speth: If the terminology is "tabulated species", these docstrings should be updated accordingly

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -0,0 +1,230 @@
...
+        if (xnow == 0)
+        {
+            dS_corr = -BigNumber;
+        } else if (xnow == 1)
+        {
+            dS_corr = BigNumber;
+        } else
+        {
+            dS_corr = GasConstant*std::log(xnow/(1.0-xnow)) + GasConstant/Faraday*std::log(this->standardConcentration(this->nSpecies()-m_kk_tab)/this->standardConcentration(m_kk_tab));
+        }
```
speth: For one, it is almost never necessary to write `this->methodName`, you can just call the method directly.

I'm more concerned about what concentrations this formula is supposed to be referencing. If `m_kk_tab == 0`, then this will attempt to get the concentration of species index 2, which is out of bounds. If `m_kk_tab == 1`, then it reduces to `standardConcentration(1)/standardConcentration(1)`, which doesn't make sense either.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -0,0 +1,230 @@
...
+     * Call the base initThermo, which handles setting the initial
+     * state.
+     */
+    ThermoPhase::initThermoXML(phaseNode, id_);
+}
+
+double* BinarySolutionTabulatedThermo::interpolate(double x) const
+{
+	static double c[2];
+    // Assumes that "table" is sorted by .first
```
speth: This comment seems to be out of date / inapplicable

samples/matlab/lithium_ion_battery/lithium_ion_battery.m
```diff
@@ -0,0 +1,99 @@
...
+    
+% This function returns the Cantera calculated anode current (in A)
+    function anCurr = anode_curr(phi_s,phi_l,X_Li_an)
+        % Set the active material mole fraction
+        set(anode,'X',['Li[anode]:' num2str(X_Li_an) ', V[anode]:' num2str(1-X_Li_an)]);
+        
+        % Set the electrode and electrolyte potential
+        setElectricPotential(elde,phi_s);
+        setElectricPotential(elyt,phi_l);
+
```
speth: Watch out for whitespace on otherwise blank lines

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -2163,7 +2200,148 @@ def build(self, p):
...
+        sc = ph.addChild('standardConc')
+        sc['model'] = self._stdconc
+
+class table(thermo):
+    """User provided thermo table for BinarySolutionTabulatedThermo"""
+    def __init__(self,
+                 moleFraction = ([],''),
+                 enthalpy = ([],''),
+                 entropy = ([],'')):
+        self.x = moleFraction
```
speth, line 2267: Please document the arguments to the constructor for the `table` class. For an example, see the `species` class.

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -2163,7 +2200,148 @@ def build(self, p):
...
+        self.h = enthalpy
+        self.s = entropy
+    def build(self,t):
+        x = ', '.join('{0:12.5e}'.format(val) for val in self.x[0])
+        u1 = t.addChild("moleFraction", x)
+        u1['units'] = self.x[1]
```
speth: Why not just omit units for the mole fractions, since they are by definition unitless?

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -2163,7 +2200,148 @@ def build(self, p):
...
+        
+class csvfile(thermo):
+    """User provided CSV file for BinarySolutionTabulatedThermo"""
+    def __init__(self,filename):
+        fh = open(filename)
+        x = []
+        for line in fh.readlines():
+            li=line.strip()
+            if not li.startswith("*"):
+                y = [float(value) for value in re.split(';|,\s|\t',line)]
```
speth: Based on the way this file is parsed, I think it's a bit of a stretch to refer to it as a "CSV" file. I would really like to avoid introducing an input file format that is used for only one purpose, when it's not really that hard to represent this data in the existing CTI/XML formats.

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -2163,7 +2200,148 @@ def build(self, p):
...
+        u2['size'] = str(len(self.h[0]))
+        s = ', '.join('{0:12.5e}'.format(val) for val in self.s[0])
+        u3 = t.addChild("entropy", s)
+        u3['units'] = self.s[1]
+        u3['size'] = str(len(self.s[0]))
+        
+class csvfile(thermo):
+    """User provided CSV file for BinarySolutionTabulatedThermo"""
+    def __init__(self,filename):
+        fh = open(filename)
```
speth: One source of ambivalence I have about introducing a separate input file for the tabulated data is the odd behavior this introduces regarding where this "CSV" file is versus where the CTI file is. Cantera will look for the CTI file anywhere on its data path, but this function will look for a file using a path relative to the current directory only.

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -2163,7 +2200,148 @@ def build(self, p):
...
+    def conc_dim(self):
+        return (1,-3)
+    def build(self, p):
+        ph = phase.build(self, p)
+        e = ph.child("thermo")
+        e['model'] = 'BinarySolutionTabulatedThermo'
+        if self._tabSpecies:
+            e1 = e.addChild('tabulatedSpecies')
+            e1['name'] = self._tabSpecies
+        if self._tabThermo:
```
speth: In the constructor, `_tabSpecies` and `_tabThermo` are required to be specified, so these checks are unnecessary.

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -2163,7 +2200,148 @@ def build(self, p):
...
+            t = e.addChild("tabulatedThermo")
+            if isinstance(self._tabThermo, thermo):
+                self._tabThermo.build(t)
```
speth: Does it matter if `_tabThermo` is a `thermo` instance or not? As long as it has an appropriate `build` method, that should be fine. And if it doesn't, the alternate case of silently skipping it doesn't seem better than just letting it fail with whatever exception otherwise would have occurred.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -0,0 +1,230 @@
...
+                throw CanteraError("BinarySolutionTabulatedThermo::initThermoXML",
+                        "Species " + tabulated_species_name + " not found.");
+            }
+            m_xlast = moleFraction(m_kk_tab);
+        }
+        if (thermoNode.hasChild("tabulatedThermo")) {
+        	XML_Node& dataNode = thermoNode.child("tabulatedThermo");
+            getFloatArray(dataNode, x, false, "", "moleFraction");
+            getFloatArray(dataNode, h, false, "", "enthalpy");
+            getFloatArray(dataNode, s, false, "", "entropy");
```
speth: These arrays should be converted to Cantera's internal units here (i.e. SI+kmol). As written, the units specified in the CTI file are being ignored.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -0,0 +1,230 @@
...
+    importPhase(root, this);
+}
+
+void BinarySolutionTabulatedThermo::compositionChanged()
+{
+	IdealSolidSolnPhase::compositionChanged();
+    _updateThermo();
+}
+
+void BinarySolutionTabulatedThermo::_updateThermo()
```
speth: Since the only place this function is called is the one location in `compositionChanged`, it would be simpler to just include the contents of this function in `compositionChanged` directly.

speth (changes requested):
Thank you for the updates. This looks to be significantly improved, however there are still a number of issues that should be addressed. I believe all of the comments made earlier that are not already marked as resolved are still applicable. Please ask if you have questions about any of them.

* Please make sure to commit files without using Windows-style line endings. You can do this with the git setting:
  ```
  git config --global core.autocrlf true
  ```
  Then, to fix the existing affected files (`BinarySolutionTabulatedThermo.h`, `BinarySolutionTabulatedThermo.cpp`, and `lithium_ion_battery.cti`), I think you should be able to do something like
  ```
  git add --renormalize path/to/file
  ```
  You may need to update your version of Git (2.16 or newer), as this is a fairly new feature.

* I very much appreciate the added Matlab example. One thing that would make this even more useful would be including a script which calls this function, e.g. doing a parameter sweep of some sort and plotting the results. Also, I'm assuming that it would eventually make sense to provide a similar example in Python, in which case I would suggest moving the `lithium_ion_battery.cti` file to the `data/inputs` directory.

* I have a number of concerns about the the implementation of the `_updateThermo` method. I am not sure that it is doing what you intend in all cases, and I think there may be a simpler way to do some of these things, which is why I was hoping you could share a description of the model.

* Please remember to add yourselves to the `AUTHORS` file

src/thermo/BinarySolutionTabulatedThermo.cpp
decaluwe: The possible exception here is that the parent class (`IdealSolidSolution`) also calls `_updateThermo()`.  It is not clear to me whether these calls from the parent class also need to check whether the composition has been changed.

include/cantera/thermo/ConstDensityTabulatedThermo.h
decaluwe: Fixed.

include/cantera/thermo/ConstDensityTabulatedThermo.h
decaluwe: The constDensityTabulatedThermo class has been deleted, but I have added what I think is sufficient documentation to IdealSolidSolutionTabulatedThermo.h

test/data/Anode_Graphite.dat
decaluwe: The updated PR now has the data loaded in the CTI format.  I believe that ctml_writer.py still includes functionality to convert from CSV and load to the phase XML, if the user so desires. But the CSV is not loaded as any official part of the distribution or the demonstration of these capabilities.  Is this sufficient, or do you think all traces of the CSV should be removed?

test/data/Anode_Graphite.dat
decaluwe: This file no longer exists.

src/thermo/ConstDensityTabulatedThermo.cpp
decaluwe: This file no longer exists in the PR.

interfaces/cython/cantera/ctml_writer.py
decaluwe: This appears to have been resolved (except for the many many cases where it was already present :) ).

interfaces/cython/cantera/ctml_writer.py
decaluwe: Is it called in `species.build`, lines 651-652?

src/thermo/ConstDensityTabulatedThermo.cpp
decaluwe: The `dS_corr` variable implements a correction term on the standard state entropy (and enthalpy), such that the calculated entropy of reaction matches the experimental value.  

<a href="https://www.codecogs.com/eqnedit.php?latex=\Delta&space;S^{\rm&space;expt}_{\rm&space;rxn}&space;=&space;\sum&space;\nu_i&space;s^{\rm&space;o}_i&space;&plus;&space;RT\ln\left(\prod_k&space;X_k^{\nu_k}&space;\right)" target="_blank"></a>

As X->1 the log term would go to +/- infinity, but from the tabulated data there is an actual value, so the standard entropy contains a term to correct for this log term.  Admittedly, there might be a more efficient way to do this, such that one is not adding the correction on, only to subtract it off, later.  I will also note that tabulated data (i.e. measured from experiments) typically does not approach X=1 or X=0, but typically stops well short of these limits.

speth:
Somehow, a number of commits that shouldn't have been part of this PR (related to Steven's other PR with the critical properties database) got included here. I removed those from this PR and squashed several commits to reduce the churn associated with addressing some of the comments on the PR, and rebased this on top of the current master. Please base all further work on this updated branch to make merging this PR easier.

A more serious issue arose while I was trying to fix the problem of the shadowed member variables, specifically `m_formGC`. As written here, the setting in the input is used to set the variable `BinarySolutionTabulatedThermo::m_formGC`, which, since it is declared in both header files, is different from the variable `IdealSolidSolnPhase::m_formGC`. However, it is the only latter variable that is actually used in the methods that are inherited from the `IdealSolidSolnPhase` class. When I fixed this (see https://github.com/speth/cantera/tree/tabulated-thermo), all of the tests for the class started to fail. What seems to be happening is that when using the `molar_volume` setting for the standard concentration, all the thermo properties end up being `inf`. I haven't had a chance to look any deeper into the cause.

decaluwe:
Okay, I think I have fixed up the tests.  Main thing was that IdealSolidSolnPhase.h needed to be included in the test file header.  After fixing this, there was still a small discrepancy between some of the test return values and the `expected_result` values.  I did the hand calculations and confirmed that the actual class behavior was correct, and overwrote the `expected_result` values.

My latest push to this branch incorporates your fixes from your branch, and then adds my additional fixes.

I think the only things left to do are:
1. Remove residual `csv` functionality from `ctml_writer.py`, which is no longer used.
2. Perhaps modify the example file in manner you mention, to show some form of parametric variation (i.e. for example, plot open circuit voltage as a function of composition, or perhaps run a charge-discharge curve).
3. Perhaps construct a similar example file in Python or Jupyter.

@speth and @bryanwweber , does this list sum up the remaining changes, in your view?

decaluwe:
Sorry - not quite sure how I closed this PR, but it is now re-opened...

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -0,0 +1,191 @@
+/**
+ * @file BinarySolutionTabulatedThermo.h
+ * Header file for an binary solution model with tabulated standard state
+ * thermodynamic data (see \ref thermoprops and class
+ * \link Cantera::BinarySolutionTabulatedThermo BinarySolutionTabulatedThermo\endlink).
+ */
+
+// This file is part of Cantera. See License.txt in the top-level directory or
+// at http://www.cantera.org/license.txt for license and copyright information.
```
bryanwweber: I think in new files we should link to `https://cantera.org/license.txt`, and as we edit older headers, update them to point to the new link.

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -527,6 +528,7 @@ def __init__(self,
                  note = '',
                  thermo = None,
                  transport = None,
+                 standardState = None,
```
bryanwweber: Is there a reason to add this argument here, rather than as the last argument? If it is added here, creating any instances of this class have to modify the order of their arguments if they're not using keyword arguments.

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -639,12 +645,21 @@ def build(self, p):
                 nt = len(self._transport)
                 for n in range(nt):
                     self._transport[n].build(t)
+        if self._standardState:
```
bryanwweber, line 648: Just want to check that the behavior here is expected: this will be falsey if `_standardState` is `None`, `''`, `0`, `[]`, etc. If any of those should be allowed values, the check should instead be `if self._standardState is not None:`

include/cantera/thermo/BinarySolutionTabulatedThermo.h
decaluwe: Fixed (for this class).

interfaces/cython/cantera/ctml_writer.py
decaluwe: I'm not sure if there's a reason, but I'm hard pressed to think of a case when someone is going to attempt to create a species from input (cti, xml, or yaml) without using any keywords...

That said, i don't really have a problem with moving it.  Hopefully @manikmayur can comment

interfaces/cython/cantera/ctml_writer.py
decaluwe: This is correct.  The accepted inputs are all strings: `unity`, `molar_volume`, and `solvent_volume`.  These are re-assigned to an int value, c.f. `IdealSolidSolnPhase::setStandardConcentrationModel`: 
```
void IdealSolidSolnPhase::setStandardConcentrationModel(const std::string& model)
{
    if (caseInsensitiveEquals(model, "unity")) {
        m_formGC = 0;
    } else if (caseInsensitiveEquals(model, "molar_volume")) {
        m_formGC = 1;
    } else if (caseInsensitiveEquals(model, "solvent_volume")) {
        m_formGC = 2;
    } else {
        throw CanteraError("IdealSolidSolnPhase::setStandardConcentrationModel",
                           "Unknown standard concentration model '{}'", model);
    }
}
```

So it should, in fact be "falsey" for any of those mentioned cases.  😄 

Now, whether we want to enable someone to directly set the int value from input is a valid question, but imo is separate from this PR.

bryanwweber: I think users setting strings is much better than "magic numbers", so I vote against allowing to set the integers directly :smile:

speth: I think this discussion is a bit confused -- the "standard state" that's being set here would be an object of type `constantIncompressible`, or nothing. The strings `molar_volume` etc. are for setting the `standard_concentration` property of the `phase` object.

interfaces/cython/cantera/ctml_writer.py
speth: I agree that the new argument needs to be last.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -0,0 +1,191 @@
...
+private:
+    void _updateThermo();
+
+    //@}
+};
+}
+
+//#endif
+
+#endif /* INCLUDE_CANTERA_THERMO_BINARYSOLUTIONTABULATEDTHERMO_H_ */
```
speth: The commented-out `#endif` should be removed.

If there is a desire to note what the include-guard `endif` matches, the string should match. But we don't normally mark this, so I'd just remove the comment.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -0,0 +1,191 @@
...
+     * subclasses to perform any initialization required after all
+     * species have been added. For example, it might be used to
+     * resize internal work arrays that must have an entry for
+     * each species.  The base class implementation does nothing,
+     * and subclasses that do not require initialization do not
+     * need to overload this method.  When importing a CTML phase
+     * description, this method is called from ThermoPhase::initThermoXML(),
+     * which is called from importPhase(),
+     * just prior to returning from function importPhase().
+     */
```
speth: The above comment, which is a duplicate of the documentation for `ThermoPhase::initThermo()` should be removed.

interfaces/cython/cantera/ctml_writer.py
manikmayur: I agree too, it can be moved to the end.

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -1627,11 +1662,13 @@ def __init__(self,
                  kf = None,
                  id = '',
                  order = '',
+                 rateCoeff = '',
```
speth: This new argument also should be added last, to avoid regressions (even though I would be surprised if anyone were using positional arguments for these later options).

Also, would it be better to name this argument `rate_coeff_type`, since it is specifying how the rate expression is interpreted, rather than the actual coefficients which are given by the `kf` argument?

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -2126,6 +2163,158 @@ def build(self, p):
...
+                 note = '',
+                 transport = 'None',
+                 initial_state = None,
+                 standard_concentration = None,
+                 tabulated_species = '',
+                 tabulated_thermo = None,
+                 options = []):
+        phase.__init__(self, name, 3, elements, species, note, 'None',
+                       initial_state, options)
+        self._pure = 0
```
speth: The `_pure` attribute does not seem to be used, so should be removed, both here and for `IdealSolidSolution`.

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -2126,6 +2163,158 @@ def build(self, p):
...
+        self._tabSpecies = tabulated_species
+        self._tabThermo = tabulated_thermo
+        self._stdconc = standard_concentration
+        self._tr = transport
+        if self._stdconc is None:
+            raise CTI_Error('In phase ' + name + ': standard_concentration must be specified.')
+        if tabulated_species is None:
+            raise CTI_Error('In phase ' + name + ': tabulated_species must be specified.')
```
speth: Checking `if tabulated_species is None` is not compatible with the default value specified above, `tabulated_species = ''`.

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -2126,6 +2163,158 @@ def build(self, p):
...
+        if tabulated_thermo is None:
+            raise CTI_Error('In phase ' + name + ': A tabulated thermo data must be specified.')
+
+    def conc_dim(self):
+        return (1,-3)
```
speth: Redefining `conc_dim` should be unnecessary, since the default from the `phase` class is to return `(1, -self._dim)`.

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -2126,6 +2163,158 @@ def build(self, p):
...
+        self._tabSpecies = tabulated_species
+        self._tabThermo = tabulated_thermo
+        self._stdconc = standard_concentration
+        self._tr = transport
+        if self._stdconc is None:
+            raise CTI_Error('In phase ' + name + ': standard_concentration must be specified.')
+        if tabulated_species is None:
+            raise CTI_Error('In phase ' + name + ': tabulated_species must be specified.')
+        if tabulated_thermo is None:
+            raise CTI_Error('In phase ' + name + ': A tabulated thermo data must be specified.')
```
speth: The grammar in the error message could use a bit of work.

interfaces/cython/cantera/ctml_writer.py
decaluwe: Oops - you're right.  Okay, so can you comment on @bryanwweber 's query?  It would look like it is used correctly.

If somebody enters something other than an instance of `contantIncompressible`, it looks like it repeats the build of the species thermo...?

interfaces/cython/cantera/ctml_writer.py
decaluwe: Done.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -0,0 +1,201 @@
...
+        c[1] = d[0] * 1e3; // 1e3 for conversion J/mol -> J/kmol
+        if (xnow == 0)
+        {
+            dS_corr = -BigNumber;
+        } else if (xnow == 1)
+        {
+            dS_corr = BigNumber;
+        } else
+        {
+            dS_corr = GasConstant*std::log(xnow/(1.0-xnow)) + GasConstant/Faraday*std::log(this->standardConcentration(1-m_kk_tab)/this->standardConcentration(m_kk_tab));
```
speth: It is almost never necessary to write `this->` before calling a member function. Also, can we wrap this line to keep things at least close to the guideline of an 80-character width?

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -0,0 +1,201 @@
...
+    /*
+     * Call the base initThermo, which handles setting the initial
+     * state.
+     */
+    ThermoPhase::initThermoXML(phaseNode, id_);
+}
+
+double* BinarySolutionTabulatedThermo::interpolate(double x) const
+{
+    static double c[2];
```
speth: The use of a `static` array here is dangerous, and introduces thread-safety issues. To return two doubles, you'd be better off using a `std::pair<double, double>` (where the two elements are then `c.first` and `c.second`.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -0,0 +1,201 @@
...
+            dS_corr = GasConstant*std::log(xnow/(1.0-xnow)) + GasConstant/Faraday*std::log(this->standardConcentration(1-m_kk_tab)/this->standardConcentration(m_kk_tab));
+        }
+        c[2] = d[1] * 1e3 + dS_corr; // 1e3 for conversion J/K/mol -> J/K/kmol
+
+        c[3] = 0.0;
+        type = m_spthermo.reportType(m_kk_tab);
+        tlow = m_spthermo.minTemp(m_kk_tab);
+        thigh = m_spthermo.maxTemp(m_kk_tab);
+        shared_ptr<SpeciesThermoInterpType> stit(
+                newSpeciesThermoInterpType(type, tlow, thigh, OneAtm, c));
```
speth: The way this is written, this requires that the species thermo model be `ConstCpPoly` (given the assumed structure of the coefficient array). However, there's nothing actually checking that this is the case.

As I commented on an earlier version of this PR, rather than recreating the `SpeciesThermoInterpType` object every time this method is called, which is rather inefficient, I would recommend leaving these objects unmodified, relying on them only to represent the temperature-dependence of the pure species thermodynamics, and add the tabulated contributions to `m_h0_RT` ant `m_s0_R` separately here.  Unless there's something I'm missing about what that wouldn't work?

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -0,0 +1,191 @@
...
+    virtual void compositionChanged();
+
+    //! Species thermodynamics interpolation functions
+    double* interpolate(double x) const;
+
+    //! Current tabulated species index
+    size_t m_kk_tab;
+
+    //! Current tabulated species mole fraction
+    mutable double m_xlast;
```
speth: This variable does not need to be marked `mutable`.

interfaces/cython/cantera/ctml_writer.py
speth: I agree, the `else` case for the `_standardState` makes no sense. I think it should be more like `if self._standardState:` build, otherwise do nothing.

interfaces/cython/cantera/ctml_writer.py
speth: I see a call to `build` in `species.build`, but not a call to `_build`.

speth (changes requested):
This looks like some good progress. Besides the issues noted below, there are a couple of items that I noted previously that I think should still be addressed:
- issue with units in calls to `getFloatArray`
- question about `_build` function of `standardState` CTI class

src/thermo/BinarySolutionTabulatedThermo.cpp
decaluwe: Hmm, even worse: they're assumed to be J/mol and J/mol/K (see lines 56 and 67 in BinarySolutionTabulatedThermo.cpp).  So ignored and assumed of a form different from Cantera's internal units.  Will fix in a second.

decaluwe: Do I leave the moleFraction entry as is, since it is unitless/dimensionless?  Or is there some entry which specifies this?  Other than that, I've got this one fixed.

decaluwe:
Okay, unit conversion on `getFloatArray` calls should be fixed, now.

interfaces/cython/cantera/ctml_writer.py
decaluwe: Okay, hitting the limits of my knowledge, here, but is the `standardState` class needed at all?

Nevermind - answered my own question.  Commenting out `_build` causes no tests to fail, but commenting out the entire class causes 19 test to fail, in just the suite of thermo tests.

interfaces/cython/cantera/ctml_writer.py
decaluwe: So skip the entire check to make sure the string assigned is actually an instance of the `standardState` class?  Just based on how other classes operate in this file, I would think it would be:

```
if self._standardState:
            ss = s.addChild("standardState")
            ss['model'] = id
            if isinstance(self._standardState, standardState):
                self._standardState.build(ss)
```

and just omit the `else` statement.  No?

include/cantera/thermo/BinarySolutionTabulatedThermo.h
decaluwe: Fixed.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
decaluwe: Fixed.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
decaluwe: Done, and done.

interfaces/cython/cantera/ctml_writer.py
decaluwe: I agree - I've never really liked the tag `rateCoeff`...nor the option `exchangecurrentdensity`, which one can use to specify the rate coefficient type.

1. I agree that `rate_coeff_type` is the way to go for this label.
2. For the latter, I would rather use `butler_volmer`, both because it's shorter, but also because the exchange current density can be used in multiple ways.  For the Butler-Volmer implementation, the charge-transfer current would be:
<a href="https://www.codecogs.com/eqnedit.php?latex=i&space;=&space;i_\circ\left[\exp\left(\frac{\beta&space;n_{\rm&space;elec}F&space;\eta}{RT}&space;\right)-\exp\left(\frac{-(1-\beta)&space;n_{\rm&space;elec}F&space;\eta}{RT}&space;\right)\right&space;]" target="_blank"></a>
Another implementation would be the Tafel equation, where at large enough overpotentials the 2nd term in Butler-Volmer becomes negligible, and is dropped:
<a href="https://www.codecogs.com/eqnedit.php?latex=i&space;=&space;i_\circ\exp\left(\frac{\beta&space;n_{\rm&space;elec}F&space;\eta}{RT}&space;\right)" target="_blank"></a>
Anyway, I think both would be valuable to enable, and the simplest would be to have the specified at the `rate_coeff_type` level.

interfaces/cython/cantera/ctml_writer.py
decaluwe: ...so I read this as "change the default to `tabulated_species = None`.  Correct?

speth: Yes, that's a good way to resolve this (and I see, already done).

src/thermo/BinarySolutionTabulatedThermo.cpp
decaluwe: Okay, this is fixed, now.

interfaces/cython/cantera/ctml_writer.py
decaluwe: Fixed.

interfaces/cython/cantera/ctml_writer.py
decaluwe: Fixed.

src/thermo/BinarySolutionTabulatedThermo.cpp
decaluwe: Fixed.

interfaces/cython/cantera/ctml_writer.py
decaluwe: This has been removed.

speth (approved):

PR #572, opened 2018-11-17
Use RPATH for linking (to avoid missing shared libraries at runtime) …

skyreflectedinmirrors:
…for appropriate (i.e., non-Windows) systems

Please fill in the issue number this pull request is fixing: N/A

- Automatically include `-rpath` linkage to user-supplied libdirs if not disabled by the user

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -610,6 +610,11 @@ config_options = [
            as symlinks.
            """,
         defaults.versionedSharedLibrary),
+    BoolVariable(
+        'use_rpath_linkage',
+        """If enabled, link to all shared libraries using 'rpath', i.e., a fixed
+           run-time search path for dynamic library loading.""",
+        True),
     EnumVariable(
         'layout',
         """The layout of the directory structure. 'standard' installs files to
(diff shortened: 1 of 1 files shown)


SConstruct
```diff
@@ -719,9 +724,20 @@ elif env['env_vars']:
 env['extra_inc_dirs'] = [d for d in env['extra_inc_dirs'].split(':') if d]
 env['extra_lib_dirs'] = [d for d in env['extra_lib_dirs'].split(':') if d]
 
+# setup rpath
+if env['debug']:
+    env['use_rpath'] = defaults.includeRPath and listify(
+        env['debug_linker_flags']) != defaults.debugLinkFlags
+else:
+    env['use_rpath'] = defaults.includeRPath and listify(
+        env['no_debug_linker_flags']) != defaults.noDebugLinkFlags
```
speth: Might be easier to handle this as a SCons option

SConstruct
```diff
@@ -277,18 +279,21 @@ elif env['CC'] == 'cl': # Visual Studio
     defaults.buildPch = True
     env['pch_flags'] = ['/FIpch/system.h']
     env['openmp_flag'] = '/openmp'
+    defaults.includeRPath = False
```
speth: Do we need to special case this, or will SCons just ignore this if it's not an option for Visual Studio?

skyreflectedinmirrors: It looks like you're right, [SCons ignores it for windows](https://scons.org/doc/2.3.6/HTML/scons-man.html#cv-RPATH)

PR #464, opened 2017-07-03
[thermo] `RedlichKwongMFTP` critical property lookups

decaluwe:
Changes proposed in this pull request:
-Makes it so that `Thermo::RedlichKwongMFTP` input files do not necessarily have to provide `pureFluidParameters` for every species.  Since these can be calculated from the critical temperature and pressure for the Redlich Kwong EoS, we provide the ability to look up Tc and Pc for well-known species.
-Provides a database of species with well-known critical properties.  Species are currently matched by chemical formula, so isomers are currently prohibited.  Can possibly add functionality to scan by species name, but this seems like it could lead to significant ambiguity.  My preference is to keep this database limited to species where potential ambiguity is near zero.  Database is stored in the `data` folder.  Updating the database is relatively straight-foward for any end users - user simply has to lookup Tc and Pc and enter them into the database, rather than lookup Tc and Pc, calculate a and b parameters for R-K EoS, and then update the cti file.
-Adds test suite coverage for this capability.

PR #574, opened 2018-11-17
Crit props 2018

decaluwe:
This supercedes #464

### DIFF
--- a/src/thermo/RedlichKwongMFTP.cpp
+++ b/src/thermo/RedlichKwongMFTP.cpp
@@ -79,9 +79,17 @@ void RedlichKwongMFTP::setSpeciesCoeffs(const std::string& species,
         if (k == j) {
             continue;
         }
-        double a0kj = sqrt(a_coeff_vec(0, j + m_kk * j) * a0);
-        double a1kj = sqrt(a_coeff_vec(1, j + m_kk * j) * a1);
-        if  (a_coeff_vec(0, j + m_kk * k) == 0) {
+
+        // a_coeff_vec is initialized to -1, so screen for unidentified species to prevent
+        // imaginary numbers on the off-diagonals:
+        if (a_coeff_vec(0, j + m_kk * j) < 0 || a_coeff_vec(1, j + m_kk * j) < 0){
+            // The diagonal element of the jth species has not yet been defined.
+            continue;
+        } else if (a_coeff_vec(0, j + m_kk * k) == -1) {
+            // Only use the mixing rules if the off-diagonal element has not already been defined by a
+            // user-specified crossFluidParameters entry:
+            double a0kj = sqrt(a_coeff_vec(0, j + m_kk * j) * a0);
+            double a1kj = sqrt(a_coeff_vec(1, j + m_kk * j) * a1);
             a_coeff_vec(0, j + m_kk * k) = a0kj;
             a_coeff_vec(1, j + m_kk * k) = a1kj;
             a_coeff_vec(0, k + m_kk * j) = a0kj;
@@ -546,9 +554,11 @@ bool RedlichKwongMFTP::addSpecies(shared_ptr<Species> spec)
     bool added = MixtureFugacityTP::addSpecies(spec);
     if (added) {
         a_vec_Curr_.resize(m_kk * m_kk, 0.0);
-        b_vec_Curr_.push_back(0.0);
 
-        a_coeff_vec.resize(2, m_kk * m_kk, 0.0);
+        // Initialize a_vec and b_vec to -1, to screen for species with
+        //     pureFluidParameters which are undefined in the input file:
+        b_vec_Curr_.push_back(-1);
+        a_coeff_vec.resize(2, m_kk * m_kk, -1);
 
         m_pp.push_back(0.0);
         m_tmpV.push_back(0.0);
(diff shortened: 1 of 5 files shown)


src/thermo/RedlichKwongMFTP.cpp
```diff
@@ -573,13 +583,69 @@ void RedlichKwongMFTP::initThermoXML(XML_Node& phaseNode, const std::string& id)
...
+
+
+            // If any species exist which have undefined pureFluidParameters,
+            //     search the database in 'critProperties.xml' to find critical
+            //     temperature and pressure and calculate a and b based on that.
+
+            // Get the array of species names:
+            XML_Node& sparrays = phaseNode.child("speciesArray");
+            std::vector<std::string> spnames;
+            getStringArray(sparrays, spnames);
```
speth: This doesn't seem to be used

src/thermo/RedlichKwongMFTP.cpp
```diff
@@ -573,13 +583,69 @@ void RedlichKwongMFTP::initThermoXML(XML_Node& phaseNode, const std::string& id)
...
+
+                // Check if a and b are already populated (only the diagonal elements of a).
+                size_t counter = iSpecies + m_kk * iSpecies;
+
+                // If not, then search the database:
+                if (a_coeff_vec(0, counter) == -1 ||
+                    b_vec_Curr_[iSpecies] == -1) {
+
+					          vector<double> coeffArray;
+
```
speth, line 621: It would be nice to get the weird tabs out of this so that the indentation actually looks right.

src/thermo/RedlichKwongMFTP.cpp
```diff
@@ -589,6 +655,84 @@ void RedlichKwongMFTP::initThermoXML(XML_Node& phaseNode, const std::string& id)
...
+                    throw CanteraError("RedlichKwongMFTP::GetCoeff",
+                                       "Critical Pressure must be positive ");
+                }
+                P_crit = vParams;
+            }
+
+            //Assuming no temperature dependence
+
+            spCoeff[0] = 0.42748 * pow(GasConstant, 2) * pow(T_crit, 2.5) / P_crit; //coeff a
+            spCoeff[1] = 0.08664 * GasConstant * T_crit / P_crit; // coeff b
```
speth: Should use the more precise constants (`omega_a` and `omega_b`) that are defined at the top of the file.

src/thermo/critProperties.xml
```diff
@@ -0,0 +1,6331 @@
...
+ file.  In the case that they are not provided for a species, the function
+ initThermoXML will search this database for critical properties, which can
+ also be used to calculate the a and b coefficients.  If critical properties
+ are not found and coefficients are not provided, then an error is thrown.
+ 
+ These critical properties have been compiled from various sources by
+ C. Karakaya (https://orcid.org/0000-0001-5102-5476)    
+ -->
+
+<ctml>
```
speth: This file in the `src/thermo` directory isn't actually used -- the only one that will get included with an installed copy of Cantera is the one in `data/inputs`. This one seems to have more species, but I'm not sure which one is preferred. In any case, there should only be one file added to the repository for this.

src/thermo/critProperties.xml
```diff
@@ -0,0 +1,6331 @@
...
+           <Pc value= "35E6 "> </Pc>
+           <Vc value="93.1"> </Vc>
+           <Zc value="0.294"> </Zc>
+           <omega value="0.066"> </omega>
+       </species>
+       <species name= "CO2">
+           <ChemName name ="carbon-dioxide"> </ChemName>
+           <MW value = "44.01"> </MW>
+           <Tc value= "304.2"> </Tc>
+           <Pc value= "73.9E6 "> </Pc>
```
speth: I think all of these critical pressures are off by an order of magnitude -- the critical pressure for CO2 is 7.39 MPa, not 73.9 MPa.

src/thermo/critProperties.xml
```diff
@@ -0,0 +1,6331 @@
...
+           <Tc value= "704.35"> </Tc>
+           <Pc value= "39.36E6 "> </Pc>
+           <Vc value="419.5"> </Vc>
+           <Zc value="0.282"> </Zc>
+           <omega value="0.275"> </omega>
+       </species>
+       <species name= "C2CL3F">
+           <ChemName name ="3  112trichlorotrifluoroet"> </ChemName>
+           <MW value = "hane 187.376"> </MW>
+           <Tc value= "487.3"> </Tc>
```
speth: How is this file being generated? There is clearly a parsing error here -- the `3` at the start of `ChemName` should be the end of the `species` name, and the `hane` at the start of the molecular weight should be at the end of the `ChemName`. There are over 100 species with errors of this sort in the file.

src/thermo/critProperties.xml
```diff
@@ -0,0 +1,6331 @@
...
+           <ChemName name ="2-methylthiophene"> </ChemName>
+           <MW value = "98.162"> </MW>
+           <Tc value= "610"> </Tc>
+           <Pc value= "48.5E6 "> </Pc>
+           <Vc value="275.5"> </Vc>
+           <Zc value="0.265"> </Zc>
+           <omega value="0.238"> </omega>
+       </species>
+       <species name= "C5H6S">
+           <ChemName name ="3-methylthiophene"> <
```
speth: There are 7 entries for `C10H14`. Right now, only the first entry will ever be used, which is a little troubling. At a minimum, maybe we should specify an alternate name for the other isomers, e.g. `C10H14-1`, `C10H14-2`, etc. This applies to quite a few other species as well.

test/thermo/RedlichKwongMFTP_Test.cpp
```diff
@@ -165,4 +178,29 @@ TEST_F(RedlichKwongMFTP_Test, setTP)
...
+    //
+    // // Check that the critical properties (temperature and pressure) are calculated correctly for
+    // // pure fluids, both for those with pureFluidParameters provided in the cti file (i.e., h2) and
+    // // those where the pureFluidParameters are calculated based on the tabulated critical properties
+    // // (i.e. co2):
+    //
+    // // CO2 - should match tabulated values in critProperties.xml
+    // set_r(1.0);
+    // EXPECT_NEAR(test_phase->critTemperature(),304.2,1.e-3);
+    // EXPECT_NEAR(test_phase->critPressure(),7390000,100);
```
speth: The contents of the test shouldn't be commented out :laughing:

test/thermo/RedlichKwongMFTP_Test.cpp
```diff
@@ -9,8 +9,9 @@ namespace Cantera
 class RedlichKwongMFTP_Test : public testing::Test
 {
 public:
-    RedlichKwongMFTP_Test() {
-        test_phase.reset(newPhase("../data/co2_RK_example.cti"));
+    void initializeTestPhase(const std::string & filename)
+    {
+        test_phase.reset(newPhase(filename));
```
speth: Leaving the constructor as it was would allow you to eliminate the boilerplate code in each test for constructing the `ThermoPhase` object.

speth (changes requested):
It looks like the implementation for reading the critical properties is more or less working, but I have some concerns about the current version of the property database.

src/thermo/RedlichKwongMFTP.cpp
decaluwe: Agreed.  Deleted.

src/thermo/RedlichKwongMFTP.cpp
decaluwe: Fixed.

src/thermo/RedlichKwongMFTP.cpp
decaluwe: Done.

src/thermo/critProperties.xml
decaluwe: Yes, I deleted the one previously in data/inputs, and then moved the newer, more extensive file over from src/thermo.

src/thermo/critProperties.xml
decaluwe: Fixed.  Thanks for catching that.

src/thermo/critProperties.xml
decaluwe: Yes, this was odd.  There appeared to be a max length for both `name` and `ChemName`, which led to these parsing errors.  I went through and corrected them, and am fairly certain that I caught them all.

src/thermo/critProperties.xml
decaluwe: All duplicate/isomers now have numbers appended, as suggested.

test/thermo/RedlichKwongMFTP_Test.cpp
decaluwe: Lol.  I've seen it both ways... 😉 

(this was a remnant from when I was trying to figure out why the tests were failing.  I just forgot to uncomment this last test, once I got it working.

test/thermo/RedlichKwongMFTP_Test.cpp
decaluwe: So just to be clear, if I add  the `test_phase.reset(newPhase...)` back in , I can remove the function `initializeTestPhase()`  and then also the lines 
    const std::string valid_file("../data/co2_RK_example.cti");
    initializeTestPhase(valid_file);
at the beginning of each test?

speth (approved):

PR #581, opened 2018-12-06
expose setting of max-steps to python interface & test

skyreflectedinmirrors:
- Expose the ability to set the maximum number of CVODEs steps to the python interface
- Add test to ensure setting works correctly.

I am using Cantera (as typically) to validate other ODE solvers.  As such I typically want to set the network reactor tolerances to very stringent values (e.g., `atol=1e-20`, `rtol=1e-15`) and run ignition problems over a wide range of conditions.

The issue I'm currently facing is that the default # of steps (appears to be 20k?) works for _most_ cases, but not all.  Rather than playing the game of trying to modify the tolerances to fit in the number of steps, it would be nice if we could just expose this property to the user via the python interface.

Let me know if there are any comments / concerns,

Best,

Nick

### DIFF
--- a/include/cantera/numerics/CVodesIntegrator.h
+++ b/include/cantera/numerics/CVodesIntegrator.h
@@ -53,6 +53,7 @@ public:
     virtual void setMaxStepSize(double hmax);
     virtual void setMinStepSize(double hmin);
     virtual void setMaxSteps(int nmax);
+    virtual int maxSteps();
     virtual void setMaxErrTestFails(int n);
     virtual void setBandwidth(int N_Upper, int N_Lower) {
         m_mupper = N_Upper;
--- a/include/cantera/zeroD/ReactorNet.h
+++ b/include/cantera/zeroD/ReactorNet.h
@@ -207,6 +207,21 @@ public:
         m_integrator_init = false;
     }
 
+    //! Set the maximum number of internal integration time-steps the
+    //! integrator will take before reaching the next output time
+    //! @param nmax The maximum number of steps, setting this value
+    //!             to zero disables this option.
+    virtual void setMaxSteps(int nmax) {
+        m_integ->setMaxSteps(nmax);
+    }
+
+    //! Returns the maximum number of internal integration time-steps the
+    //!  integrator will take before reaching the next output time
+    //!
+    virtual int maxSteps() {
+        return m_integ->maxSteps();
+    }
+
 protected:
     //! Initialize the reactor network. Called automatically the first time
     //! advance or step is called.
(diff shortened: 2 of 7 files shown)


bryanwweber:
Hey @arghdos! Thanks for this. Can you also add a method to get the max steps as well? Does CVODES expose that information?

skyreflectedinmirrors:
@bryanwweber Hmm, it doesn't look like they allow you to query the maximum # of steps, however we should be able to just return whatever value is stored in the integrator.

skyreflectedinmirrors:
@bryanwweber -- implemented

interfaces/cython/cantera/test/test_reactor.py
```diff
@@ -140,6 +140,23 @@ def test_timestepping(self):
...
+        max_steps = 10
+        max_step_size = 1e-07
+        self.net.set_initial_time(0)
+        self.net.set_max_time_step(max_step_size)
+        self.net.max_steps = max_steps
+        try:
+            self.net.advance(1e-04)
+            self.assertTrue(False)
+        except ct.CanteraError:
+            self.assertTrue(self.net.time <= max_steps * max_step_size)
```
speth: Rather than the try/except, I would recommend using `assertRaisesRegex`. You can see some of the examples in `test_kinetics`. Also, instead of the `assertTrue`, prefer the more specific functions like `assertLess` and `assertEqual`.

include/cantera/numerics/CVodesIntegrator.h
```diff
@@ -53,6 +53,7 @@ class CVodesIntegrator : public Integrator
     virtual void setMaxStepSize(double hmax);
     virtual void setMinStepSize(double hmin);
     virtual void setMaxSteps(int nmax);
+    virtual int getMaxSteps();
```
speth: Rather than `getMaxSteps`, I think just `maxSteps` would be more consistent with the other methods.

include/cantera/zeroD/ReactorNet.h
```diff
@@ -207,6 +207,22 @@ class ReactorNet : public FuncEval
...
+     *  @param nmax The maximum number of steps, setting this value
+     *              to zero disables this option.
+     */
+    virtual void setMaxSteps(int nmax) {
+        m_integ->setMaxSteps(nmax);
+    }
+
+    /** Returns the maximum number of internal integration time-steps the
+     *  integrator will take before reaching the next output time
+     */
```
speth: If there's only a 'brief' docstring, I tend to prefer just using the `//!` on each line rather than the C-style block comment.

speth (commented):
I had just a few minor suggestions, but this generally looks fine to me.

speth:
Sorry, I didn't notice that you updated this. LGTM!

Issue #569, opened 2018-10-31
The examples should use except ct.CanteraError where appropriate


bryanwweber:
In some of the flame examples [here](https://github.com/Cantera/cantera/blob/5c783c708f0c08def6f6a3d8328b336d65842357/interfaces/cython/cantera/examples/reactors/ic_engine.py#L149), [here](https://github.com/Cantera/cantera/blob/5c783c708f0c08def6f6a3d8328b336d65842357/interfaces/cython/cantera/examples/onedim/diffusion_flame_extinction.py#L116), [here](https://github.com/Cantera/cantera/blob/5c783c708f0c08def6f6a3d8328b336d65842357/interfaces/cython/cantera/examples/onedim/diffusion_flame_batch.py#L119), and [here](https://github.com/Cantera/cantera/blob/5c783c708f0c08def6f6a3d8328b336d65842357/interfaces/cython/cantera/examples/onedim/diffusion_flame_batch.py#L170) we use a bare `except Exception`, probably as a leftover from before `CanteraError`s could be caught in Python. The latter two examples rely on catching a bare `Exception` to determine extinction, but it would probably be better to give a class name to that exception to distinguish solver errors vs. extinction.

CyberDrudge:
I'm unsure about the exception to distinguish solver errors vs. extinction, should i keep that as it is (that `if-else` clause) or implement a new error class for maybe each of them or something else.

bryanwweber:
@CyberDrudge The "Flame extinguished" string comes from this line: https://github.com/Cantera/cantera/blob/5c783c708f0c08def6f6a3d8328b336d65842357/interfaces/cython/cantera/examples/onedim/diffusion_flame_batch.py#L56 I would suggest adding a simple class that derives from `Exception` to indicate that the flame has extinguished, like
```python
class FlameExtinguished(Exception): pass
```
That should be defined outside the interrupt function and then `raise`d inside the interrupt. Then, you should have two `except` statements I think
```python
try:
    f.solve(...)
    ...
except FlameExtinguished:
    print('Flame extinguished')
    break
except ct.CanteraError as e:
    print('Error occurred while solving', e)
    break
```
That way, something like an `OSError` isn't caught and stops the simulation, but we handle the errors we know how to handle. Please also try to fix as many typos as you can find in that file. Thanks!

CyberDrudge:
I was thinking about implementing something similar, just wasn't sure.

PR #587, opened 2019-01-10
Use CanteraError

CyberDrudge:
Fixes part of #569 

Implemented use of _ct.CanteraError_ in **ic_engine.py**

### DIFF
--- a/interfaces/cython/cantera/examples/reactors/ic_engine.py
+++ b/interfaces/cython/cantera/examples/reactors/ic_engine.py
@@ -150,7 +150,7 @@ for n1, t_i in enumerate(t):
         except ct.CanteraError:
             sim.set_max_time_step(1e-6 * 10. ** -n2)
             n_last_refinement = n1
-        if n2 == 3 and solved == False:
+        if not solved:
             raise ct.CanteraError('Refinement limit reached')
     # coarsen time step if too long ago
     if n1 - n_last_refinement == n_wait_coarsening:
--- a/interfaces/cython/cantera/examples/reactors/ic_engine.py
+++ b/interfaces/cython/cantera/examples/reactors/ic_engine.py
@@ -141,14 +141,19 @@ for n1, t_i in enumerate(t):
         injector_mfc.set_mass_flow_rate(0)
 
     # perform time integration, refine time step if necessary
+    solved = False
     for n2 in range(4):
         try:
             sim.advance(t_i)
+            solved = True
+            break
         except ct.CanteraError:
             sim.set_max_time_step(1e-6 * 10. ** -n2)
             n_last_refinement = n1
+        if n2 == 3 and solved == False:
+            raise ct.CanteraError('Refinement limit reached')
     # coarsen time step if too long ago
-    if n1 - n_last_refinement is n_wait_coarsening:
+    if n1 - n_last_refinement == n_wait_coarsening:
         sim.set_max_time_step(1e-5)
 
     # write output data
--- a/interfaces/cython/cantera/examples/reactors/ic_engine.py
+++ b/interfaces/cython/cantera/examples/reactors/ic_engine.py
@@ -142,8 +142,6 @@ for n1, t_i in enumerate(t):
 
     # perform time integration, refine time step if necessary
     for n2 in range(4):
-        if n2 is 4:
-            raise 'Error: Refinement limit reached'
         try:
             sim.advance(t_i)
         except ct.CanteraError:
--- a/interfaces/cython/cantera/examples/reactors/ic_engine.py
+++ b/interfaces/cython/cantera/examples/reactors/ic_engine.py
@@ -150,8 +150,8 @@ for n1, t_i in enumerate(t):
         except ct.CanteraError:
             sim.set_max_time_step(1e-6 * 10. ** -n2)
             n_last_refinement = n1
-        if not solved:
-            raise ct.CanteraError('Refinement limit reached')
+    if not solved:
+        raise ct.CanteraError('Refinement limit reached')
     # coarsen time step if too long ago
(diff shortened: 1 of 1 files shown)


interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -146,7 +146,7 @@ def piston_speed(t):
             raise 'Error: Refinement limit reached'
```
bryanwweber: There is a bug here too, can you fix it by raising a `ct.CanteraError`?

bryanwweber (commented):
This looks like a good start, keep going with the other files!

interfaces/cython/cantera/examples/reactors/ic_engine.py
bryanwweber: And on the line prior, can you change it to use `==` rather than `is`?

CyberDrudge: If that for loop iterates from 0 to 3, will the if condition ever be fulfilled? What is its significance?

speth: Indeed, that check makes no sense as written.

CyberDrudge: I'll remove it then.

bryanwweber: It would be nice if when that loop exited because the limit was reached without success if the script did not continue, and an error was reported to the user. There's also another `is` comparison that should be `==` a few lines below.

CyberDrudge: I'll change `is`  to `==` . And for the error to be reported to the user if limit was reached without success, shouldn't the condition `if n2 == 4` be after the loop and not inside it.

bryanwweber: The more I think about it, the more I think the whole logic in that loop is borked. What it does right now is try to advance the reactor to `t_i`. If it succeeds, it tries again up to 3 more times (effectively not doing anything other than wasting time). If it fails, the max time step is reduced and the loop goes around to `advance` the simulation again. If it fails 4 times, it just silently continues with the script, even though the simulation isn't complete. Oy.

The most straightforward thing to do is add a Boolean variable before the start of the loop called `solved` initially `False`, set `solved` to `True` and `break` the loop if the `advance` doesn't `raise`, and check the value outside the loop to `raise` an error. The fun way is to use the Python [loop-else](http://python-notes.curiousefficiency.org/en/latest/python_concepts/break_else.html) clause to raise the error if the loop doesn't get broken 😄 (I recommend the straightforward way).

speth: The logic is even more borked than that. The real problem is from changing model parameters in the loop without calling `sim.reinitialize()`. If that problem is fixed, there won't be any need for any of this try/except nonsense. But that seems well outside the scope of this PR.

CyberDrudge: @speth Currently i am unfamiliar with the working of  `sim.reinitialize()`. I implemented the changes suggested by @bryanwweber .

interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -141,16 +141,19 @@ def piston_speed(t):
...
-            raise 'Error: Refinement limit reached'
         try:
             sim.advance(t_i)
-        except Exception:
+            solved = True
+            break
+        except ct.CanteraError:
             sim.set_max_time_step(1e-6 * 10. ** -n2)
             n_last_refinement = n1
+        if n2 == 3 and solved == False:
```
bryanwweber: The `n2 == 3` check is redundant, I'd suggest removing it.
```suggestion
        if solved == False:
```

bryanwweber: Actually, this should just be
```suggestion
        if not solved:
```

bryanwweber:
@speth This change causes the `Refinement limit reached` error to be raised and the example effectively doesn't work. Prior to this change, as we've noted, the logic in the loop was not correct but at least it produced a figure at the end. What did you mean if `sim.reinitialize()` is used, the `try-except` isn't necessary? Won't the `try` be needed to catch when the time step isn't fine enough? Should the max time step even be adjusted here at all?

I think we need to figure out how to fix this example before this PR can be merged, otherwise we'd distribute broken examples 😄

speth:
The reason it doesn't work is because the logic in the loop isn't right -- the `if not solved` check should be outside the loop. Right now, it gets hit on the first failure, and you never make a second pass through the loop.

I think my comment about `sim.reinitialize()` is well beyond the scope of this PR. Because of the way the CVODES integrator works, where it can need function evaluations at times other than the specified output times, changing parameters of the ODE outside of the ODE function (where the time that the integrator chooses to evaluate is actually known) can lead to failures. One way around this is to call `reinitialize` after any such discontinuous change. The other is to set the parameters from a function that is called as part of the ODE evaluation, e.g. the way we allow `MassFlowController` to have a mass flow rate as a function of time.

bryanwweber:
@CyberDrudge Please see the first part of @speth's comment. I missed the indentation there!

bryanwweber:
@CyberDrudge I rebased the commits to combine a few of them and updated it to the current master branch. This PR was merged by d9b95b2efbe6c512f00f72fe2e0bf4fda76ba7e6 Thank you for doing this work!

bryanwweber:
@CyberDrudge Since you have this experience now, I'd encourage you to go through the rest of the places noted in #569 to fix those as well 😄

CyberDrudge:
Sure, I'll work on them as soon as i'll get some time.

PR #592, opened 2019-02-02
Add solvePseudoSteadyStateProblem() in cython interface

lavdwall:
This fixes my post in the Cantera Users' Group:
https://groups.google.com/d/msg/cantera-users/cZZXmLGPXaM/qpwBDh47GAAJ

### DIFF
--- a/interfaces/cython/cantera/kinetics.pyx
+++ b/interfaces/cython/cantera/kinetics.pyx
@@ -366,6 +366,12 @@ cdef class InterfaceKinetics(Kinetics):
         coverages for a specified amount of time.
         """
         (<CxxInterfaceKinetics*>self.kinetics).advanceCoverages(dt)
+      
+    def advance_coverages_to_steady_state(self):
+        """
+        This method advances the surface coverages to (pseudo) steady state.
+        """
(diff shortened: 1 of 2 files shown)


interfaces/cython/cantera/kinetics.pyx
```diff
@@ -366,6 +366,12 @@ cdef class InterfaceKinetics(Kinetics):
         coverages for a specified amount of time.
         """
         (<CxxInterfaceKinetics*>self.kinetics).advanceCoverages(dt)
+      
+    def advance_coverages_to_steady_state(self):
+        """
+        This method advances the surface coverages to (pseudo) steady state.
```
speth: I know this comes from the name of the underlying method, but can you either explain what "pseudo" means here, or remove it from the description, depending on what you think makes more sense?

speth (changes requested):
The implementation looks good to me. I just have a small question about the docstring.

interfaces/cython/cantera/kinetics.pyx
lavdwall: For me it is OK just to remove it from the description.

speth: Can you push a revised version of the PR, then?

PR #596, opened 2019-02-12
Use Cantera Error in leftover examples

### LINKED ISSUE #569: The examples should use except ct.CanteraError where appropriate

CyberDrudge:
Fixes #569 
Implemented use of CanteraError in examples [diffusion_flame_extinction.py](https://github.com/Cantera/cantera/blob/master/interfaces/cython/cantera/examples/onedim/diffusion_flame_extinction.py) and [diffusion_flame_batch.py](https://github.com/Cantera/cantera/blob/master/interfaces/cython/cantera/examples/onedim/diffusion_flame_batch.py)

### DIFF
--- a/interfaces/cython/cantera/examples/onedim/diffusion_flame_batch.py
+++ b/interfaces/cython/cantera/examples/onedim/diffusion_flame_batch.py
@@ -12,13 +12,18 @@ The tutorial makes use of the scaling rules derived by Fiala and Sattelmayer
 (doi:10.1155/2014/484372). Please refer to this publication for a detailed
 explanation. Also, please don't forget to cite it if you make use of it.
 
-This example can e.g. be used to iterate to a counterflow diffusion flame to an
+This example can, for example, be used to iterate to a counterflow diffusion flame to an
 awkward  pressure and strain rate, or to create the basis for a flamelet table.
 """
 
 import cantera as ct
 import numpy as np
 import os
+
+
+class FlameExtinguished(Exception): 
+    pass
+
 
 # Create directory for output data files
 data_directory = 'diffusion_flame_batch_data/'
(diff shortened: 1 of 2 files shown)


interfaces/cython/cantera/examples/onedim/diffusion_flame_batch.py
```diff
@@ -76,8 +85,8 @@ def interrupt_extinction(t):
 
 # Exponents for the initial solution variation with changes in pressure Taken
 # from Fiala and Sattelmayer (2014). The exponents are adjusted such that the
-# strain rates increases proportional to p^(3/2), which results in flames
-# similar with respect to the extinction strain rate.
+# strain rates increases proportional to p^(3/2), which results in similar
+# flames with respect to the extinction strain rate.
```
bryanwweber: I think these lines should actually not be changed. In this case, the comment is using "similar" in a technical sense to mean that there is a similarity relationship between the conditions.

interfaces/cython/cantera/examples/onedim/diffusion_flame_extinction.py
```diff
@@ -113,11 +113,11 @@
     f.set_profile('lambda', normalized_grid, f.L * strain_factor ** exp_lam_a)
     try:
         f.solve(loglevel=0)
-    except Exception as e:
-        # Throw Exception if solution fails
+    except ct.CanteraError as e:
+        # Throw Error if solution fails
```
bryanwweber: I think you can probably just delete the comment, it doesn't add anything anyways.

bryanwweber (changes requested):
Thanks @CyberDrudge I restarted the failing build because I think it was a problem with the CodeCov service, not with anything you did here. A few changes below!

CyberDrudge:
I applied the suggested changes and squashed the commits.

bryanwweber (approved):

PR #597, opened 2019-02-13
Support for Langmuir-Hinshelwood kinetics

lavdwall:
These changes should allow to use Langmuir-Hinshelwood type of kinetics, in the way it is done in Chemkin. 
This answers my question here: https://groups.google.com/forum/#!topic/cantera-users/B-l_HFkJy0c

### DIFF
--- a/include/cantera/kinetics/InterfaceKinetics.h
+++ b/include/cantera/kinetics/InterfaceKinetics.h
@@ -368,6 +368,11 @@ public:
     virtual void determineFwdOrdersBV(ElectrochemicalReaction& r, vector_fp& fwdFullorders);
 
 protected:
+
+    //! Add Langmuir-Hinshelwood type of reaction
+    void addLangmuirReaction(LangmuirReaction& r);
+    void modifyLangmuirReaction(size_t i, LangmuirReaction& rNew);
+
     //! Build a SurfaceArrhenius object from a Reaction, taking into account
     //! the possible sticking coefficient form and coverage dependencies
     //! @param i  Reaction number
@@ -392,6 +397,10 @@ protected:
      *  The class SurfaceArrhenius is described in RxnRates.h
      */
     Rate1<SurfaceArrhenius> m_rates;
+    Rate1<LangmuirRate> m_LH_rates;
+
+     //! Array of partial pressures for each species
+    vector_fp m_pp;
 
     bool m_redo_rates;
 
--- a/include/cantera/kinetics/RxnRates.h
+++ b/include/cantera/kinetics/RxnRates.h
@@ -451,6 +451,69 @@ protected:
     vector_fp dotProd_; //!< dot product of chebCoeffs with the reduced pressure polynomial
 };
 
+//! Langmuir-Hinshelwood reaction rate type
+class LangmuirRate
+{
+public:
+    //! return the rate coefficient type.
+    static int type() {
+        return LANGMUIR_REACTION_RATECOEFF_TYPE;
+    }
+     //! Default constructor.
+    LangmuirRate();
+     /// Constructor.
+    /// @param A pre-exponential. The unit system is
+    ///     (kmol, m, s). The actual units depend on the reaction
+    ///     order and the dimensionality (surface or bulk).
+    /// @param b Temperature exponent. Non-dimensional.
+    /// @param E Activation energy in temperature units. Kelvin.
+    LangmuirRate(doublereal A, doublereal b, doublereal E);
+     //! Add an adsorption dependency for species *k*, with equilibrium
+    //! constant \f$ E_k = A_k T^{b_k} exp(-H_k/RT) \f$, where A_k is in terms of
+    //! pressure. The parameter *d* is the species exponent in the denominator.
+    //! The parameter *n* indicates whether or not to include the
+    //! equilibrium constant in the numerator.
+    void addAdsorptionDependence(size_t k, doublereal a,
+                               doublereal b, doublereal h,
+                               doublereal d, bool n);
+
+    void correctOrderEffect(doublereal order);
+
+    //! Set the overal exponent of the denominator
+    void setDenominatorExponent(doublereal m);
+     //! Update concentration-dependent parts of the rate coefficient.
+    //! @param conc the concentrations in kmol/m3
+    void update_C(const doublereal* conc) {
+        size_t k;
+        for (size_t n = 0; n < m_cc.size(); n++) {
+            k = m_sp[n];
+            m_cc[n] = conc[k];
+        }
+    }
+     /**
+     * Update the value the rate constant.
+     *
+     * This function returns the actual value of the rate constant. It can be
+     * safely called for negative values of the pre-exponential factor.
+     */
+    doublereal updateRC(doublereal logT, doublereal recipT) {
+        m_denom = 1.0;
+        m_num = m_A * std::exp(m_b*logT - m_E*recipT);
+        for (size_t n = 0; n < m_cc.size(); n++) {
+            m_Kc[n] = m_ac[n] * std::exp(m_bc[n]*logT - m_hc[n]*recipT);
+            m_denom += m_Kc[n] * std::pow(m_cc[n]*GasConstant/recipT, m_dc[n]);
+            if (m_nc[n]) m_num *= m_Kc[n];
+        }
+        return m_num / std::pow(m_denom, m_md);
+    }
+ protected:
+    doublereal m_A, m_E, m_b;
+    doublereal m_md, m_denom, m_num;
+    vector_fp m_Kc, m_ac, m_hc, m_bc, m_cc, m_dc;
+    std::vector<size_t> m_sp;
+    std::vector<bool> m_nc;
+};
+
 }
 
 #endif
--- a/interfaces/cython/cantera/ck2cti.py
+++ b/interfaces/cython/cantera/ck2cti.py
@@ -550,6 +550,30 @@ class PDepArrhenius(KineticsModel):
         return '\n'.join(lines)
 
 
+class Langmuir(Arrhenius):
+    """
+    A Langmuir-Hinshelwood type of reaction
+    """
+    def __init__(self, *args, **kwargs):
+        Arrhenius.__init__(self, *args, **kwargs)
+        self.adsorption = []
+        self.lhnu = []
+    def rateStr(self):
+        s = Arrhenius.rateStr(self)
+        s = '\n{0}Langmuir({1},'.format(' '*17, s[1:-1])
+        if len(self.adsorption) > 0:
+            s +='\n{0}adsorption=['.format(' '*17)
+            for species,A,b,H,m,n in self.adsorption:
+                if self.Ea[1] != self.parser.output_energy_units:
+                    H = (H, self.Ea[1])
+                s += '[{0!r}, {1}, {2}, {3}, {4}, {5!r}],'.format(str(species), A, b, H, m, str(n))
+            s = s.rstrip(',') + '],'
+        return s.rstrip(',') + ')'
+    def to_cti(self, reactantstr, arrow, productstr, indent=0):
+        rxnstring = reactantstr + arrow + productstr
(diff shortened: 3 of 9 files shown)


interfaces/cython/cantera/ctml_writer.py
```diff
@@ -1053,6 +1053,47 @@ def build(self, p, name='', a=None):
...
+                    raise CTI_Error("Incorrect number of langmuir adsorption parameters")
+        else:
+            self._ads = None
+    def build(self, p, name='', a=None):
+        if a is None:
+            a = p.addChild('Langmuir')
+        if name:
+            a['name'] = name
+         # if a pure number is entered for A, multiply by 10 (mol/cm2 to kmol/m2).
+        addFloat(a,'A',self._c[0]*1.0e1, fmt = '%14.6E')
```
speth, line 1079: 1. This doesn't seem to address the case where units for `A` are supplied explicitly
2. The implied units of `A` are not necessarily `mol/cm^2` - the CTI file could specify any set of quantity and length units, so this should follow the implementation of `Arrhenius.build`. Ideally, the `Langmuir` class here would be derived from the `Arrhenius` class to avoid repetition of this code.

include/cantera/kinetics/RxnRates.h
```diff
@@ -451,6 +451,69 @@ class ChebyshevRate
...
+    }
+     //! Default constructor.
+    LangmuirRate();
+     /// Constructor.
+    /// @param A pre-exponential. The unit system is
+    ///     (kmol, m, s). The actual units depend on the reaction
+    ///     order and the dimensionality (surface or bulk).
+    /// @param b Temperature exponent. Non-dimensional.
+    /// @param E Activation energy in temperature units. Kelvin.
+    LangmuirRate(doublereal A, doublereal b, doublereal E);
```
speth, line 470: Per the [contributor guidelines](https://github.com/Cantera/cantera/blob/master/CONTRIBUTING.md), please use just `double` as opposed to `doublereal` in all new code.

include/cantera/kinetics/InterfaceKinetics.h
```diff
@@ -392,6 +397,10 @@ class InterfaceKinetics : public Kinetics
      *  The class SurfaceArrhenius is described in RxnRates.h
      */
     Rate1<SurfaceArrhenius> m_rates;
+    Rate1<LangmuirRate> m_LH_rates;
```
speth, line 400: If we're using the name `Langmuir` elsewhere, I would recommend naming this `m_Langmuir_rates`for consistency, and to avoid adding another obscure initialism.

include/cantera/kinetics/InterfaceKinetics.h
```diff
@@ -392,6 +397,10 @@ class InterfaceKinetics : public Kinetics
...
+
+     //! Array of partial pressures for each species
+    vector_fp m_pp;
```
speth, line 403: This doesn't seem to be used.

include/cantera/kinetics/RxnRates.h
```diff
@@ -451,6 +451,69 @@ class ChebyshevRate
     vector_fp dotProd_; //!< dot product of chebCoeffs with the reduced pressure polynomial
 };
 
+//! Langmuir-Hinshelwood reaction rate type
+class LangmuirRate
```
speth, line 455: The equations defining the Lanmuir-Hinshelwood form need to be documented somewhere. While this has been done for other reaction rate classes in this file, I would actually be more interested in seeing it added to the Cantera "science" documentation on the website by making modifications to https://github.com/Cantera/cantera-website/blob/master/pages/science/reactions.rst, and then simply providing a link to https://cantera.org/science/reactions.html#langmuir-hinshelwood-reactions here.

include/cantera/kinetics/RxnRates.h
```diff
@@ -451,6 +451,69 @@ class ChebyshevRate
...
+     //! Add an adsorption dependency for species *k*, with equilibrium
+    //! constant \f$ E_k = A_k T^{b_k} exp(-H_k/RT) \f$, where A_k is in terms of
+    //! pressure. The parameter *d* is the species exponent in the denominator.
+    //! The parameter *n* indicates whether or not to include the
+    //! equilibrium constant in the numerator.
+    void addAdsorptionDependence(size_t k, doublereal a,
+                               doublereal b, doublereal h,
+                               doublereal d, bool n);
+
+    void correctOrderEffect(doublereal order);
```
speth, line 480: This function needs a docstring

include/cantera/kinetics/RxnRates.h
```diff
@@ -451,6 +451,69 @@ class ChebyshevRate
...
+
+    //! Set the overal exponent of the denominator
```
speth, line 482: Fix spelling of "overall"

interfaces/cython/cantera/ck2cti.py
```diff
@@ -1313,6 +1342,7 @@ def parseExpression(expression, dest):
         efficiencies = {}
         chebyshevCoeffs = []
         revReaction = None
+        langmuir = None
```
speth, line 1345: I don't see a case where this variable takes on a value other than `None`

interfaces/cython/cantera/ck2cti.py
```diff
@@ -1297,6 +1322,10 @@ def parseExpression(expression, dest):
 
         # The rest of the first line contains Arrhenius parameters
         reaction_type = SurfaceArrhenius if surface else Arrhenius
+
+        if self.pressure_units:
+            reaction_type = Langmuir
```
speth, line 1327: This doesn't seem right -- the `reaction_type` for all reactions should not depend on whether or not non-default pressure units have been given.

interfaces/cython/cantera/ck2cti.py
```diff
@@ -550,6 +550,30 @@ def to_cti(self, reactantstr, arrow, productstr, indent=0):
...
 
 
+class Langmuir(Arrhenius):
+    """
+    A Langmuir-Hinshelwood type of reaction
+    """
+    def __init__(self, *args, **kwargs):
+        Arrhenius.__init__(self, *args, **kwargs)
+        self.adsorption = []
+        self.lhnu = []
```
speth, line 560: This variable does not seem to be used (and so the `LHNU` does not seem to be parsed/utilized)

interfaces/cython/cantera/ck2cti.py
```diff
@@ -1815,6 +1847,8 @@ def readline():
                             self.motz_wise = True
                         elif token == 'MWOFF':
                             self.motz_wise = False
+                        elif token == 'PASCALS':
+                            self.pressure_units = True
```
speth, line 1851: Why is `PASCALS` the only allowed pressure unit? The valid pressure units on the `REACTIONS` line are, according to the Chemkin Input manual, `ATM`, `BAR`, `DYN`, `TOR` and `PASCAL`. And within the `LHPR` declaration, they are `ATM`, `BAR`, `TORR`, `PASC` and `DYNE` (slightly different for no apparent reason).

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -1053,6 +1053,47 @@ def build(self, p, name='', a=None):
...
 
+class Langmuir(rate_expression):
+    def __init__(self,
+                 A = 0.0,
+                 b = 0.0,
+                 E = 0.0,
+                 adsorption = []):
+        self._c = [A, b, E]
+        if adsorption:
+            if isinstance(adsorption[0], basestring):
```
speth, line 1064: Use `string_types` (defined in this file) instead of `basestring`, in order to provide compatibility with both Python 2.7 and Python 3.

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -524,6 +527,9 @@ bool InterfaceKinetics::addReaction(shared_ptr<Reaction> r_base)
         return false;
     }
 
+    if (r_base->reaction_type == LANGMUIR_RXN) {
+        addLangmuirReaction(dynamic_cast<LangmuirReaction&>(*r_base));
+    } else {
```
speth, line 532: Instead of enclosing the rest of this function in braces but not indenting it, the first branch of the `if` should end with `return true;`.

Also, I think most of the other initialization that is done in the second branch of the `if` still needs to be done for instances of `LangmuirReaction`, at least to keep the indexing for later reactions correct -- basically everything starting with the settings of `m_revindex` and `m_irrev`.

include/cantera/kinetics/RxnRates.h
```diff
@@ -451,6 +451,69 @@ class ChebyshevRate
...
+        return LANGMUIR_REACTION_RATECOEFF_TYPE;
+    }
+     //! Default constructor.
+    LangmuirRate();
+     /// Constructor.
+    /// @param A pre-exponential. The unit system is
+    ///     (kmol, m, s). The actual units depend on the reaction
+    ///     order and the dimensionality (surface or bulk).
+    /// @param b Temperature exponent. Non-dimensional.
+    /// @param E Activation energy in temperature units. Kelvin.
```
speth, line 469: While both options are valid, I would suggest using `//!` to mark Doxygen comment blocks rather than `///`, given that the former is far more widely used within Cantera.

include/cantera/kinetics/RxnRates.h
```diff
@@ -451,6 +451,69 @@ class ChebyshevRate
...
 };
 
+//! Langmuir-Hinshelwood reaction rate type
+class LangmuirRate
+{
+public:
+    //! return the rate coefficient type.
+    static int type() {
+        return LANGMUIR_REACTION_RATECOEFF_TYPE;
+    }
```
speth, line 461: I know this was written for consistency with the other classes in `RxnRates.h`, but I don't think it's actually used anywhere.

src/kinetics/RxnRates.cpp
```diff
@@ -152,4 +152,42 @@ ChebyshevRate::ChebyshevRate(double Tmin, double Tmax, double Pmin, double Pmax,
     }
 }
 
+LangmuirRate::LangmuirRate()
+    : m_A(0.0)
+    , m_b(0.0)
+    , m_E(0.0)
+    , m_md(2.0)
+{
```
speth, line 160: These fields should be initialized in the order they appear in the header file.

include/cantera/kinetics/RxnRates.h
```diff
@@ -451,6 +451,69 @@ class ChebyshevRate
...
+     /**
+     * Update the value the rate constant.
+     *
+     * This function returns the actual value of the rate constant. It can be
+     * safely called for negative values of the pre-exponential factor.
+     */
+    doublereal updateRC(doublereal logT, doublereal recipT) {
+        m_denom = 1.0;
+        m_num = m_A * std::exp(m_b*logT - m_E*recipT);
```
speth, line 501: There is no need for `m_denom` and `m_num` to be member variables of the class instead of just local variables for this function.

speth (changes requested):
Thanks for making this pull request. I think it will be a valuable feature addition to Cantera. In addition to the comments left on specific sections of code, I would also suggest the following:
* Please add some tests of this new functionality. There are two sets that make sense:
    * To test the conversions from Chemkin->CTI->XML, you can add tests to `test_convert.py` in the `chemkinConverterTest` class.
    * To test that the new rate calculations are doing what is expected, you can add tests to either `test_kinetics.py` or `test/kinetics/rates.cpp`
* Please do not change the mode of the `ctml_writer.py` and `ck2cti.py` scripts - they should be left as non-executable in the Git repository.
* Please check the indentation -- there are a number of places where things are indented by one space more than they should be (i.e. 5 instead of 4)

ischoegl:
@lavdwall ... is this PR still being pursued?

ischoegl:
@lavdwall - why did you decide to close?

lavdwall:
This is no longer being pursued..

ischoegl:
ok. From the perspective of the community, it may have been a nice addition, and someone will ultimately benefit. I do understand that going through reviews takes time

PR #435, opened 2017-02-20
Preserve units in PLOG entries in ck2cti

rwest:
This PR preserves the units when converting PLOG entries from a chemkin file.

For example of how it should work, if you read and output both in cal/mol then you get this for an Arrhenius
`reaction('ch3co3h <=> ch3co2 + oh', [5.010000e+14, 0.0, 40150.0])`
but if you read in kcal/mol and output in cal/mol then you get this
`reaction('ch3co3h <=> ch3co2 + oh', [5.010000e+14, 0.0, (40.15, 'kcal/mol')])`
which is great. It preserves the units on Ea.
Also works for high and low pressure limits of all the fall-off types, as far as I can tell.
But if you have a pdep_arrhenius (PLOG) reaction the units are ignored.
Eg. if you have a reaction that ought to look (in cal/mol) like
```
pdep_arrhenius('ch2cho + o2 <=> o2ch2cho',
               [(0.01, 'atm'), 1.580000e+77, -21.9, 19350.0],
               [(0.1, 'atm'), 3.880000e+69, -18.84, 19240.0],
               [(1.0, 'atm'), 7.800000e+59, -15.4, 17650.0],
               [(10.0, 'atm'), 3.050000e+50, -12.2, 15630.0])
```
but you read it from a file in kcal/mol then your output looks like 
```
pdep_arrhenius('ch2cho + o2 <=> o2ch2cho',
               [(0.01, 'atm'), 1.580000e+77, -21.9, 19.35],
               [(0.1, 'atm'), 3.880000e+69, -18.84, 19.24],
               [(1.0, 'atm'), 7.800000e+59, -15.4, 17.65],
               [(10.0, 'atm'), 3.050000e+50, -12.2, 15.63])
```

After this PR it becomes
```
pdep_arrhenius('ch2cho + o2 <=> o2ch2cho',
               [(0.01, 'atm'), 1.580000e+77, -21.9, (19.35, 'kcal/mol')],
               [(0.1, 'atm'), 3.880000e+69, -18.84, (19.24, 'kcal/mol')],
               [(1.0, 'atm'), 7.800000e+59, -15.4, (17.65, 'kcal/mol')],
               [(10.0, 'atm'), 3.050000e+50, -12.2, (15.63, 'kcal/mol')])
```
because we re-use the unit-preserving code in the Arrhenius class.

I haven't actually checked that the resulting .cti syntax can be parsed by Cantera....
I guess that should be done before merging the PR :)

[...]

PR #600, opened 2019-02-26
Fix Chebyshev rate constant when using local quantity units

speth:
Since the leading Chebyshev coefficient has effective units like log(cm^3/kmol), it needs to be converted directly to the default units of the CTI file.

Analogous to the fix for PLOG reactions in #435.


bryanwweber:
Hey @speth, just curious what you changed to fix the previous CI errors so I can learn from it :smile:

speth:
The first version of this didn't update all of the affected tests. I hadn't noticed the failure because I was only running a subset of the test suite locally.

interfaces/cython/cantera/ck2cti.py
```diff
@@ -65,6 +65,8 @@
                 'KJOULES/MOL': 'kJ/mol',
                 'KJOULES/MOLE': 'kJ/mol'}
 
+Avogadro = 6.02214129e23  # in molec/mol
```
bryanwweber: Can you document where this number comes from? For instance, it is different from the CODATA value: https://physics.nist.gov/cgi-bin/cuu/Value?na

bryanwweber (approved):

interfaces/cython/cantera/ck2cti.py
speth: It is the 2010 CODATA value, consistent with the value in `ct_defs.h`.

Issue #404, opened 2016-12-05
Question on defaults for Troe Reaction parameters


skyreflectedinmirrors:
### Cantera version
v2.3.0.b1

### Operating System
RHEL7

### Python/MATLAB version
python2.7

### Expected Behavior
Undefined, see below.

### Actual Behavior
Currently, on line 33-43 of [Falloff.cpp](https://github.com/Cantera/cantera/blob/ae6e7fdff770009e4b6ba7db1543c73227fbe4eb/src/kinetics/Falloff.cpp) the (inverse of) the Troe temperature parameters are set to 1000 in the case that the mechanism incorrectly specifies T1 or T3 to be zero.  

### Steps to reproduce
[DME_cti.txt](https://github.com/Cantera/cantera/files/631688/DME_cti.txt) (note, I had to rename it to a text file to upload to github)
```

import cantera as ct
gas = ct.Solution('mechs/DME.cti')
print gas.reactions()[47].falloff.parameters
[  0.00000000e+00   5.70000000e+02   1.00000000e-03   1.00000000e+30]
```

Of note here is that T1 is set to 1e-3 (or 1 / 1000) when the mechanism has the parameter (incorrectly) set to zero.  The modeler seemingly did the incorrect (but often used in engineering) simplification that:
`0 * exp(-C / 0) == 0`

However, this is actually undefined in a strict mathematical sense (and to boot, is crappy input)

A few questions:
1.  Are these default values significant? e.g. if they were from the original paper I would understand, but in this case we should add the defaults to the documentation
2.  Is this the desired behavior? 

Generally I think we don't want a magical changing of numbers behind the scenes.  Indeed in the SRI initialization we have checks that throw errors instead, e.g.:

```
if (c[2] < 0.0) {
        throw CanteraError("SRI::init()",
                           "m_c parameter is less than zero: {}", c[2]);
    }
```

If we agree that it should result in an error to specify T1 or T3 = 0, I'll submit a pull request along these lines

skyreflectedinmirrors:
Going through the original papers (thanks @bryanwweber), I don't see any of the above default values

speth:
I don't think there's anything particularly magical going on here, although the existing implementation may be generating a bit of unnecessary confusion.

[...]

PR #601, opened 2019-02-28
Additional check for Troe coefficients being zero

g3bk47:
I encountered the problem that the "init" function of Troe reactions generates a floating point exception due to division by zero. I am using the FORCE Caltech mechanism from here:
http://www.theforce.caltech.edu/CaltechMech/
In the mechanism file
http://cvs.theforce.caltech.edu/cgi-bin/viewvc.cgi/CaltechMech/CaltechMech.chmech?view=co
there is the following reaction:

 H+O2(+M) = HO2(+M)                      5.590e+13    0.200      0.00
   O2/0.89/ H2/1.50/ H2O/11.70/ CO2/2.18/ AR/0.52/ CO/1.09/ 
   LOW  /  2.650e+19   -1.300      0.00 /
   TROE/     0.7     0.00 10000000000.00 10000000000.00 /

The second Troe parameter is zero. According to the Chemkin Format
http://akrmys.com/public/chemkin/CKm_inp.html.en
this corresponds to T*** or "m_rt3" in Cantera:
https://github.com/Cantera/cantera/blob/5c783c708f0c08def6f6a3d8328b336d65842357/src/kinetics/Falloff.cpp#L33

Using the standard installation of Cantera, this does not generate any errors. But in my case, I use Cantera together with another program (OpenFOAM), which throws hardware exceptions if this occurs. Of course, I could change this by hand in the mechanism file itself, but maybe it is nicer to handle this in the code.
This was actually handled in previous versions, but was removed in this commit:
https://github.com/Cantera/cantera/commit/84e6775ee304185ec11f736358a82f6da2a2c04d#diff-4210fe882743537c1bb1cacdbe10255b
and this issue:
https://github.com/Cantera/cantera/issues/404
I would argue to include a special treatment for zero in case Cantera is used together with programs like OpenFOAM which will trigger hardware exceptions if division by zero occurs.

Because this part of the code is not relevant to performance, I introduced an additional check. This will prevent hardware exceptions in case c[1] or c[2] are zero but will exactly keep the current behaviour if c[1] and c[2] are not zero.

### DIFF
--- a/src/kinetics/Falloff.cpp
+++ b/src/kinetics/Falloff.cpp
@@ -30,8 +30,18 @@ void Troe::init(const vector_fp& c)
             c.size());
     }
     m_a = c[0];
-    m_rt3 = 1.0/c[1];
-    m_rt1 = 1.0/c[2];
+    if (std::abs(c[1]) < SmallNumber) {
+        m_rt3 = std::numeric_limits<double>::infinity();
+    } else {
+        m_rt3 = 1.0 / c[1];
+    }
+
+    if (std::abs(c[2]) < SmallNumber) {
+        m_rt1 = std::numeric_limits<double>::infinity();
+    } else {
+        m_rt1 = 1.0 / c[2];
+    }
+
     if (c.size() == 4) {
         m_t2 = c[3];
     }


src/kinetics/Falloff.cpp
```diff
@@ -30,8 +30,17 @@ void Troe::init(const vector_fp& c)
             c.size());
     }
     m_a = c[0];
-    m_rt3 = 1.0/c[1];
-    m_rt1 = 1.0/c[2];
+    
+    if (std::abs(c[1]) < SmallNumber)
+        m_rt3 = 1.0/SmallNumber;
```
speth: * I would suggest setting the value to `std::numeric_limits<double>::infinity()`, so that the inverse operation in `Troe::getParameters` provides the correct output.
* Please do not use `if/else` constructs without braces.

speth (changes requested):

speth:
Just pushed an updated version to fix some formatting issues (trailing whitespace and placement of curly braces). Good to merge once the CI completes.

speth (approved):

PR #606, opened 2019-03-05
Fix property updates in BinarySolutionTabulatedThermo

speth:
Because `IdealSolidSolnPhase::_updateThermo` wasn't a virtual method, and the signature wasn't the same as `BinarySolutionTabulatedThermo::_updateThermo` (const vs non-const), calls to this method
from `IdealSolidSolnPhase` weren't being overridden by ` BinarySolutionTabulatedThermo::_updateThermo` as expected.

After fixing this, there was still a problem in the case the temperature changed but the mole fractions were the same, where the tabulated correction to the enthalpy / entropy was being replaced by the pure species values. Now, we store the tabulated values and apply them every time.

### DIFF
--- a/include/cantera/thermo/BinarySolutionTabulatedThermo.h
+++ b/include/cantera/thermo/BinarySolutionTabulatedThermo.h
@@ -158,6 +158,12 @@ protected:
     //! Current tabulated species mole fraction
     mutable double m_xlast;
 
+    //! Tabulated contribution to h0[m_kk_tab] at the current composition
+    mutable double m_h0_tab;
+
+    //! Tabulated contribution to s0[m_kk_tab] at the current composition
+    mutable double m_s0_tab;
+
     //! Vector for storing tabulated thermo
     vector_fp m_molefrac_tab;
     vector_fp m_enthalpy_tab;
--- a/include/cantera/thermo/BinarySolutionTabulatedThermo.h
+++ b/include/cantera/thermo/BinarySolutionTabulatedThermo.h
@@ -156,7 +156,7 @@ protected:
     size_t m_kk_tab;
 
     //! Current tabulated species mole fraction
-    double m_xlast;
+    mutable double m_xlast;
 
     //! Vector for storing tabulated thermo
     vector_fp m_molefrac_tab;
@@ -164,7 +164,7 @@ protected:
     vector_fp m_entropy_tab;
 
 private:
-    void _updateThermo();
+    virtual void _updateThermo() const;
 };
 }
 
--- a/src/thermo/BinarySolutionTabulatedThermo.cpp
+++ b/src/thermo/BinarySolutionTabulatedThermo.cpp
@@ -113,7 +113,6 @@ void BinarySolutionTabulatedThermo::initThermoXML(XML_Node& phaseNode, const std
                 throw CanteraError("BinarySolutionTabulatedThermo::initThermoXML",
                         "Species " + tabulated_species_name + " not found.");
             }
-            m_xlast = moleFraction(m_kk_tab);
         }
         if (thermoNode.hasChild("tabulatedThermo")) {
             XML_Node& dataNode = thermoNode.child("tabulatedThermo");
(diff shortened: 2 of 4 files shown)


decaluwe:
Thanks, @speth - this logic sounds correct, and explains a few oddities I was seeing (but didn’t comprehend) when I was finalizing this PR.  I’ll take a closer look in the next day or 2.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -156,15 +156,15 @@ class BinarySolutionTabulatedThermo : public IdealSolidSolnPhase
...
     //! Current tabulated species mole fraction
-    double m_xlast;
+    mutable double m_xlast;
 
     //! Vector for storing tabulated thermo
     vector_fp m_molefrac_tab;
     vector_fp m_enthalpy_tab;
     vector_fp m_entropy_tab;
 
 private:
```
decaluwe, line 172: Just for my own growth, can you explain what adding `const` after the method name does?

test/thermo/BinarySolutionTabulatedThermo_Test.cpp
```diff
@@ -52,6 +52,9 @@ TEST_F(BinarySolutionTabulatedThermo_Test,interp_h)
     {
         set_defect_X(xmin + i*dx);
         EXPECT_NEAR(expected_result[i], test_phase->enthalpy_mole(), 1.e-6);
+        // enthalpy is temperature-independent
```
decaluwe: This is correct for the cti file on which this test is built, but am I correct that it is not necessarily true of the phase model (i.e. the excess terms are added to the species "standard state" terms, which are calculated according to whatever is in the cti file?) 

If I'm correct about that, there are two rather minor concerns:
1. The test could easily be broken if the underlying cti file is updated w/ temperature-dependent species thermo.  Thinking while I write, here, but this is actually a good thing, in that the cti probably shouldn't be changed in that way.  It is good to be able to verify that temperature updates and composition updates are 'disentangled,' and the temperature-independent thermo provides a good pathway for this.
2. Also a very minor concern, but someone reading this test suite to learn more about the class might get the wrong impression.  Chances are vanishingly small that this would happen, particularly as there are other, better places to learn about the class.  But it crossed my mind...

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -50,53 +50,36 @@ void BinarySolutionTabulatedThermo::compositionChanged()
...
             m_h0_RT[k] += deltaE;
             m_g0_RT[k] = m_h0_RT[k] - m_s0_R[k];
         }
-        m_h0_RT[m_kk_tab] += d.first*rrt;
-        m_s0_R[m_kk_tab] += (d.second + dS_corr)*rr;
-        m_g0_RT[m_kk_tab] = m_h0_RT[m_kk_tab] - m_s0_R[m_kk_tab];
-
         m_tlast = tnow;
-        m_xlast = xnow;
-    } else if (m_tlast != tnow) {
```
decaluwe, line 89: Thanks for catching this.  I know I thought through the logic when I was doing this, and am having a hard time figuring out if I actually thought this previous approach was satisfactory, or just forgot to go back and fix it.

Anyway, the new approach is a lot better, regardless.

decaluwe (approved):
Hi Ray, thanks for catching this and cleaning it up.  This all looks good to me.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
speth: A method with `const` at the end of the declaration cannot modify any member variables of the class, except those that are marked as `mutable`. See [this SO answer](https://stackoverflow.com/a/3141107/310477) for a more detailed explanation.

The fact that a non-`const` function does not override a `const` function which otherwise has the same signature is one of C++'s many subtle "gotchas".

test/thermo/BinarySolutionTabulatedThermo_Test.cpp
speth: Correct, that is a property of the example input file which sets the standard state to be h(T) = 0, not of the model as a whole. In general, I would expect changes to the test data files to cause some of the tests to fail -- if you can change the data files and all the tests still succeed, I'd argue that the tests aren't checking the right things.

I will update the comment to indicate *why* the enthalpy is temperature independent in this particular case, to avoid any confusion.

PR #610, opened 2019-03-07
add integrator parameters to the InterfaceKinetics::advanceCoverages …

### LINKED ISSUE #608: Allow setting of integrator options for InterfaceKinetics.advanceCoverages
Right now, there is nothing a user can do to resolve integration failures in `InterfaceKinetics.advanceCoverages`. Providing access to some of the integrator options would allow users to work around at least some failures. The following methods (from the `CVodesIntegrator` class) come to mind as the most useful:
- `setTolerances`
- `setMaxErrTestFails`
- `setMaxStepSize`

skyreflectedinmirrors:
…& ImplicitSurfChem constructor to allow users to modfiy

Fixes #608

- Adds user-settable integrator parameters to the InterfaceKinetics::advanceCoverages method
- Adds the same to the Cython interface

@speth -- what would be a good test here?  Something like simultaneously limiting the maximum step-size to `t_out/N` while limiting the maximum number of steps to `N-1` would be a good test from the cython interface for those values, but I'm not sure how to test that the rtol / atol or error test failures are correctly set without lowering the tests into C++?  I guess I could add some information methods (similar to those in the base Integrator class) to query these directly, but that doesn't seem to fit well with the very short lived nature of the integrator inside of InterfaceKinetics...

### DIFF
--- a/include/cantera/kinetics/ImplicitSurfChem.h
+++ b/include/cantera/kinetics/ImplicitSurfChem.h
@@ -61,8 +61,16 @@ public:
      * @param k  Vector of pointers to InterfaceKinetics objects Each object
      *           consists of a surface or an edge containing internal degrees of
      *           freedom representing the concentration of surface adsorbates.
+     * @param rtol   The relative tolerance for the integrator
+     * @param atol   The absolute tolerance for the integrator
+     * @param maxStepSize   The maximum step-size the integrator is allowed to take
+     * @param maxSteps   the maximum number of time-steps the integrator can take
+     * @param maxErrTestFails   the maximum permissible number of error test failures
      */
-    ImplicitSurfChem(std::vector<InterfaceKinetics*> k);
+    ImplicitSurfChem(std::vector<InterfaceKinetics*> k,
+                     doublereal rtol=1.e-7, doublereal atol=1.e-14,
+                     doublereal maxStepSize=0, size_t maxSteps=0,
+                     size_t maxErrTestFails=0);
 
     virtual ~ImplicitSurfChem() {};
 
@@ -230,6 +238,8 @@ protected:
     std::unique_ptr<Integrator> m_integ;
     doublereal m_atol, m_rtol; // tolerances
     doublereal m_maxstep; //!< max step size
+    size_t m_nmax; //!< maximum number of steps allowed
+    size_t m_maxErrTestFails; //!< maximum number of error test failures allowed
     vector_fp m_work;
 
     /**
--- a/include/cantera/kinetics/ImplicitSurfChem.h
+++ b/include/cantera/kinetics/ImplicitSurfChem.h
@@ -63,14 +63,17 @@ public:
      *           freedom representing the concentration of surface adsorbates.
      * @param rtol   The relative tolerance for the integrator
      * @param atol   The absolute tolerance for the integrator
-     * @param maxStepSize   The maximum step-size the integrator is allowed to take
-     * @param maxSteps   the maximum number of time-steps the integrator can take
+     * @param maxStepSize   The maximum step-size the integrator is allowed to take.
+     *                      If zero, this option is disabled.
+     * @param maxSteps   The maximum number of time-steps the integrator can take.
+     *                   If not supplied, uses the default value in the CVodesIntegrator (20000).
      * @param maxErrTestFails   the maximum permissible number of error test failures
+     *                           If not supplied, uses the default value in CVODES (7).
      */
     ImplicitSurfChem(std::vector<InterfaceKinetics*> k,
-                     doublereal rtol=1.e-7, doublereal atol=1.e-14,
-                     doublereal maxStepSize=0, size_t maxSteps=0,
-                     size_t maxErrTestFails=0);
+                     double rtol=1.e-7, double atol=1.e-14,
+                     double maxStepSize=0, size_t maxSteps=20000,
+                     size_t maxErrTestFails=7);
 
     virtual ~ImplicitSurfChem() {};
 
@@ -78,6 +81,27 @@ public:
      *  Must be called before calling method 'advance'
      */
     virtual void initialize(doublereal t0 = 0.0);
+
+    /*!
+     *  Set the maximum integration step-size.  Note, setting this value to zero
+     *  disables this option
+     */
+    virtual void setMaxStepSize(double maxstep = 0.0);
+
+    /*!
+     *  Set the relative and absolute integration tolerances.
+     */
+    virtual void setTolerances(double rtol=1.e-7, double atol=1.e-14);
+
+    /*!
+     *  Set the maximum number of CVODES integration steps.
+     */
+    virtual void setMaxSteps(size_t maxsteps = 20000);
+
+    /*!
+     *  Set the maximum number of CVODES error test failures
+     */
+    virtual void setMaxErrTestFails(size_t maxErrTestFails = 7);
 
     //! Integrate from t0 to t1. The integrator is reinitialized first.
     /*!
--- a/include/cantera/kinetics/InterfaceKinetics.h
+++ b/include/cantera/kinetics/InterfaceKinetics.h
@@ -242,14 +242,16 @@ public:
      * @param tstep  Time value to advance the surface coverages
      * @param rtol   The relative tolerance for the integrator
      * @param atol   The absolute tolerance for the integrator
-     * @param maxStepSize   The maximum step-size the integrator is allowed to take
-     * @param maxSteps   the maximum number of time-steps the integrator can take
-     *                   before reaching tstep
+     * @param maxStepSize   The maximum step-size the integrator is allowed to take.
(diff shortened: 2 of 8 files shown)


speth:
Yes, I think your first suggested test is a good one. For the tolerances, I agree that creating 
"getter" methods would be overkill. You could try a test that runs with two different sets of tolerances, and make sure that the solution is similar but not identical.

skyreflectedinmirrors:
Not sure why the OSX TravisCI job timed out?

bryanwweber:
@arghdos I restarted the job, let's see what happens

interfaces/cython/cantera/test/test_reactor.py
```diff
@@ -1542,3 +1542,39 @@ def test_ConstPressureReactor(self):
...
+        dt = 1.0
+        max_steps = 10
+        max_step_size = dt / (max_steps + 1)
+        # this should throw an error, as we can't reach dt
+        try:
+            surf.advance_coverages(
+                dt=dt, max_step_size=max_step_size, max_steps=max_steps)
+            self.fail()
+        except ct.CanteraError:
+            pass
```
speth: Instead of writing the `try`/`except` block, you can use `with self.assertRaises(ct.CanteraError):`.

interfaces/cython/cantera/test/test_reactor.py
```diff
@@ -1542,3 +1542,39 @@ def test_ConstPressureReactor(self):
         self.assertEqual(states.X[-2], 1)
         for i in range(3,7):
             self.assertNear(states.T[i], states.T[2])
+
+
+class ImplicitSurfChemTest(utilities.CanteraTest):
```
speth: I'd suggest a different name for the test suite, rather than dragging in the incomprehensible name of the implementation class which is never seen in user code. Maybe `SurfaceIntegratorTest` or even `AdvanceCoveragesTest`?

interfaces/cython/cantera/test/test_reactor.py
```diff
@@ -1542,3 +1542,39 @@ def test_ConstPressureReactor(self):
...
+        try:
+            surf.advance_coverages(
+                dt=dt, max_step_size=max_step_size, max_steps=max_steps)
+            self.fail()
+        except ct.CanteraError:
+            pass
+
+        # next, run with different tolerances
+        gas1 = ct.Solution('ptcombust.xml', 'gas')
+        surf = ct.Interface('ptcombust.xml', 'Pt_surf', [gas1])
```
speth: This should probably be a separate test method, with the initialization of the `gas1` and `surf1` objects done in the `setUp` method.

interfaces/cython/cantera/test/test_reactor.py
```diff
@@ -1542,3 +1542,39 @@ def test_ConstPressureReactor(self):
...
+        surf.advance_coverages(dt=dt, rtol=1e-5, atol=1e-12)
+        cov = surf.coverages[:]
+
+        surf.coverages = 'O(S):0.1, PT(S):0.5, H(S):0.4'
+        gas1.TP = surf.TP
+        surf.advance_coverages(dt=dt, rtol=1e-7, atol=1e-14)
+
+        # check that the solutions are similar, but not identical
+        self.assertArrayNear(cov, surf.coverages)
+        assert any(cov != surf.coverages)
```
speth: All of the test assertions should be methods of the test class, rather than plain `assert` statements, in order to get good error messages out of them.

src/kinetics/ImplicitSurfChem.cpp
```diff
@@ -107,6 +112,18 @@ void ImplicitSurfChem::getState(doublereal* c)
 void ImplicitSurfChem::initialize(doublereal t0)
 {
     m_integ->setTolerances(m_rtol, m_atol);
+    if (m_maxstep > 0)
+    {
```
speth: Please place the opening brace on the preceding line.

include/cantera/kinetics/ImplicitSurfChem.h
```diff
@@ -61,8 +61,16 @@ class ImplicitSurfChem : public FuncEval
...
+     * @param atol   The absolute tolerance for the integrator
+     * @param maxStepSize   The maximum step-size the integrator is allowed to take
+     * @param maxSteps   the maximum number of time-steps the integrator can take
+     * @param maxErrTestFails   the maximum permissible number of error test failures
      */
-    ImplicitSurfChem(std::vector<InterfaceKinetics*> k);
+    ImplicitSurfChem(std::vector<InterfaceKinetics*> k,
+                     doublereal rtol=1.e-7, doublereal atol=1.e-14,
+                     doublereal maxStepSize=0, size_t maxSteps=0,
+                     size_t maxErrTestFails=0);
```
speth: Please use `double` instead of `doublereal` in any new code.

interfaces/cython/cantera/kinetics.pyx
```diff
@@ -376,13 +376,16 @@ cdef class InterfaceKinetics(Kinetics):
             self._phase_indices[phase.name] = i
             self._phase_indices[i] = i
 
-    def advance_coverages(self, double dt):
+    def advance_coverages(self, double dt, double rtol=1e-7, double atol=1e-14,
+                          double max_step_size=0, int max_steps=0,
+                          int max_err_failures=0):
```
speth: I would suggest the parameter name be `max_err_test_fails` or `max_error_test_failures`

include/cantera/kinetics/InterfaceKinetics.h
```diff
@@ -240,8 +240,16 @@ class InterfaceKinetics : public Kinetics
...
+     * @param rtol   The relative tolerance for the integrator
+     * @param atol   The absolute tolerance for the integrator
+     * @param maxStepSize   The maximum step-size the integrator is allowed to take
+     * @param maxSteps   the maximum number of time-steps the integrator can take
+     *                   before reaching tstep
+     * @param maxErrTestFails   the maximum permissible number of error test failures
+     */
+    void advanceCoverages(doublereal tstep, doublereal rtol=1.e-7,
+                          doublereal atol=1.e-14, doublereal maxStepSize=0,
+                          size_t maxSteps=0, size_t maxErrTestFails=0);
```
speth: Instead of using zero as the default value for `maxSteps` and `maxErrTestFails`, it might be more instructive to use the default values from CVODES, which are 500 and 7. For `maxStepSize`, we should document that 0 means no limit is enforced.

src/kinetics/ImplicitSurfChem.cpp
```diff
@@ -107,6 +112,18 @@ void ImplicitSurfChem::getState(doublereal* c)
...
+        m_integ->setMaxStepSize(m_maxstep);
+    }
```
speth, line 117: Given the call to `setMaxStepSize` in `ImplicitSurfChem::integrate`, this has no effect, right?

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -767,11 +767,14 @@ doublereal InterfaceKinetics::electrochem_beta(size_t irxn) const
...
-void InterfaceKinetics::advanceCoverages(doublereal tstep)
+void InterfaceKinetics::advanceCoverages(doublereal tstep, doublereal rtol,
+                                         doublereal atol, doublereal maxStepSize,
+                                         size_t maxSteps, size_t maxErrTestFails)
 {
     if (m_integrator == 0) {
         vector<InterfaceKinetics*> k{this};
-        m_integrator = new ImplicitSurfChem(k);
+        m_integrator = new ImplicitSurfChem(k, rtol, atol, maxStepSize, maxSteps,
+                                            maxErrTestFails);
```
speth: This probably isn't a big issue given the typical use cases, but as written here, changes to the integrator parameters on subsequent calls to `advanceCoverages` will have no effect. This could be fixed by moving the setting of these parameters to a different method that can be called either on a new or existing `ImplicitSurfChem` object, rather than requiring them to be constructor arguments.

speth (changes requested):
There's a little bit of cleanup to be done, but this looks pretty good as far as the actual user-facing part of it goes.

include/cantera/kinetics/InterfaceKinetics.h
skyreflectedinmirrors: @speth -- my only concern here is that 500 steps is very low and usually causes a mxsteps exception... Maybe we should use the default from `CVodesIntegrator` (20000)?

speth: Ah, good point. I had forgotten that `CVodesIntegrator` overrides the default `maxSteps` of CVODES with its own value. And certainly for the case of `advanceCoverages`, where the most common use case seems to be just advancing to steady state without caring about the intermediate timesteps, maintaining that higher value seems fair.

skyreflectedinmirrors:
@speth -- updated this, should be good now

PR #611, opened 2019-03-08
Add Cython language_level directive to _cantera.pyx

bryanwweber:
This fixes the warnings generated by recent versions of Cython that
the language_level will be changed in the future. By setting this
directive, all the code in the .pyx files should be written in
Python 3 syntax. This required several changes to the import
syntax in the files to fix relative vs. absolute imports

See https://stackoverflow.com/questions/54900723/what-does-language-level-in-setup-py-for-cython-do for more information about this setting

### DIFF
--- a/interfaces/cython/cantera/onedim.pyx
+++ b/interfaces/cython/cantera/onedim.pyx
@@ -1,7 +1,7 @@
 # This file is part of Cantera. See License.txt in the top-level directory or
 # at http://www.cantera.org/license.txt for license and copyright information.
 
-import interrupts
+from .interrupts import no_op
 import warnings
 
 # Need a pure-python class to store weakrefs to
@@ -589,7 +589,7 @@ cdef class Sim1D:
 
         self.sim = new CxxSim1D(D)
         self.domains = tuple(domains)
-        self.set_interrupt(interrupts.no_op)
+        self.set_interrupt(no_op)
         self._initialized = False
         self._initial_guess_args = ()
         self._initial_guess_kwargs = {}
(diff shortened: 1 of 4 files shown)


interfaces/cython/cantera/onedim.pyx
```diff
@@ -1,7 +1,7 @@
 # This file is part of Cantera. See License.txt in the top-level directory or
 # at http://www.cantera.org/license.txt for license and copyright information.
 
-import interrupts
+from .interrupts import no_op
```
speth, line 4: The more directly equivalent way to write this would be `from . import interrupts`, right?

bryanwweber: Yes, I think so, although since its just one function I didn't see the need to keep the namespace. Since that's the only function in that module and that's the only place its used, I thought about removing the module entirely. But since `no_op` is pure Python, putting it into the `.pyx` file would cause it to be compiled and expect the type annotations, right?

speth: I guess it's the balance between just one function and the fact that it's only used one time -- the  form you used certainly makes sense in the case where the function is used many times.

The reason for this module existing is written in the docstring for the `no_op` function: "a pure Python function is needed in order for KeyboardInterrupt events to be captured.". I'm not sure if there are any additional considerations regarding type annotations.

bryanwweber: So if the function were in the .pyx file, it would be compiled and couldn't be used as an interrupt, got it. Thanks!

PR #624, opened 2019-04-16
Enable additional states / equations in onedim

ischoegl:
Fixes # [cantera-users] Additional Equations in StFlow

- add `nSpecies() { return m_nsp; }` to `StFlow.h`
- replace  `m_nsp = m_flow->nComponents() - c_offset_Y;` by `m_nsp = m_flow->nSpecies();` in `boundaries1D.cpp`

Proposed change enables decoupling of number of components from the number of species.

### DIFF
--- a/include/cantera/oneD/StFlow.h
+++ b/include/cantera/oneD/StFlow.h
@@ -70,6 +70,11 @@ public:
         m_thermo = &th;
     }
 
+    //! Return number of species
+    virtual size_t nSpecies() {
+        return m_nsp;
+    }
+  
     //! Set the kinetics manager. The kinetics manager must
     void setKinetics(Kinetics& kin) {
         m_kin = &kin;
--- a/include/cantera/oneD/StFlow.h
+++ b/include/cantera/oneD/StFlow.h
@@ -70,11 +70,6 @@ public:
         m_thermo = &th;
     }
 
-    //! Return number of species
-    virtual size_t nSpecies() {
-        return m_nsp;
-    }
-  
     //! Set the kinetics manager. The kinetics manager must
     void setKinetics(Kinetics& kin) {
         m_kin = &kin;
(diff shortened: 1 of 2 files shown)


ischoegl:
Not sure about codecov flagging files that were not changed (GasTransport.cpp?). The push request is limited to 5 lines of code ...

PS: a functional (albeit rudimentary) example for code with added equations is posted on [ischoegl/ctapp](https://github.com/ischoegl/ctapp) (object `NewFlow` defined in C++)

include/cantera/oneD/StFlow.h
```diff
@@ -70,6 +70,11 @@ class StFlow : public Domain1D
         m_thermo = &th;
     }
 
+    //! Return number of species
+    virtual size_t nSpecies() {
+        return m_nsp;
+    }
+
```
speth: Instead of introducing a new method here, you could just access the number of species from `Inlet1D` as `m_flow->phase().nSpecies()`.

speth (commented):
I can see how the change in `Inlet1D` is necessary to be able to add state variables, but I'm kind of surprised that this is sufficient. Certainly, there's a lot more work to be done to make the 1D solver more generically extensible, but this seems fine for what it does.

The coverage failures are confusing but not a problem. The test suite passing is what's important.

include/cantera/oneD/StFlow.h
ischoegl: Yes, the simplicity of the fix surprised me also. Thanks for the suggestion: I didn't think of calling `m_flow->phase()`; I created the new method as neither `m_thermo` nor `m_nsp` were accessible.

I'll report back.

ischoegl:
@speth per your suggestion I eliminated the `StFLow::nSpecies()` method I had added, now accessing `m_flow->phase().nSpecies()` instead. The code compiles and the test suite passes.

PR #629, opened 2019-04-23
Improve output handling in zeroD simulations

### LINKED ISSUE #628: Enhancement: automatic reduction of advance step in ReactorNet simulations
### Cantera version
2.5.0a2
### Current Behavior
There is no option for an automatic reduction of time increments between outputs for `ReactorNet.advance` when states rapidly change (e.g. example simulations for `IdealGasConstPressureReactor` in `reactor1.py`)
### Enhancement
An automatic reduction of time steps ensures that processes are properly resolved without having to guess an appropriate step size.
### Steps to reproduce
1. run example script `reactor1.py`, where the maximum temperature difference exceeds 500 K, i.e.
```
    t [s]      T [K]     P [Pa]       u [J/kg]
 ...
 3.100e-04   1207.525 101325.000   5.456362e+05
 3.200e-04   1718.185 101325.000   3.697981e+05
...
```
2. At `t = 3.100e-4`, a maximum temperature difference of ~50K between outputs requires a time advance of 1.25e-06 (8-fold reduction) at the given conditions.
3. For simulations of ignition processes, delays change over several orders of magnitude. Using small time steps by default may create unnecessarily large output files; the determination of an appropriate time step may involve trial and error.

ischoegl:
Fixes #628 

- use `CVodesIntegrator` to calculate time derivatives via `CVodeGetDky`
- implement a Taylor series expansion to estimate future states of a `ReactorNet` object
- specify maximum difference ("atol") for states between advance states
- create new `ReactorNet::advanceTowards` function as an alternative to `ReactorNet::advance`, which automatically reduces the time advance when excessive state differences are predicted (time advances are successively reduced by `2^(-1)` until differences remain within specified limits. Internal cvode time steps remain unaffected).

All functions are implemented in C++ and are accessible from the Python interface via cython. The new example `reactor3.py` illustrates differences to the standard `ReactorNet` example `reactor1.py`.

### DIFF
--- a/include/cantera/numerics/Integrator.h
+++ b/include/cantera/numerics/Integrator.h
@@ -7,7 +7,7 @@
  */
 
 // This file is part of Cantera. See License.txt in the top-level directory or
-// at http://www.cantera.org/license.txt for license and copyright information.
+// at https://cantera.org/license.txt for license and copyright information.
 
 #ifndef CT_INTEGRATOR_H
 #define CT_INTEGRATOR_H
@@ -138,6 +138,18 @@ public:
     //! The current value of the solution of the system of equations.
     virtual doublereal* solution() {
         warn("solution");
+        return 0;
+    }
+
+    //! n-th derivative of the output function at time tout.
+    virtual double* derivative(double tout, int n) {
+        warn("derivative");
+        return 0;
+    }
+
+    //! Order used during the last solution step
+    virtual int lastOrder() const {
+        warn("lastOrder");
         return 0;
     }
 
--- a/include/cantera/zeroD/Reactor.h
+++ b/include/cantera/zeroD/Reactor.h
@@ -132,6 +132,20 @@ public:
     //! @see componentIndex()
     virtual std::string componentName(size_t k);
 
+    //! Set absolute step size limits during advance
+    //! @param limits array of step size limits with length neq
+    virtual void setAdvanceLimits(const double* limits);
+
+    //! Retrieve absolute step size limits during advance
+    //! @param[out] limits array of step size limits with length neq
+    //! @returns           True if at least one limit is set, False otherwise
+    virtual bool getAdvanceLimits(double* limits);
+
+    //! Set individual step size limit for compoment name *nm*
+    //! @param nm component name
+    //! @param limit value for step size limit
+    virtual void setAdvanceLimit(const std::string& nm, const double limit);
+
 protected:
     //! Set reaction rate multipliers based on the sensitivity variables in
     //! *params*.
@@ -179,6 +193,8 @@ protected:
     bool m_chem;
     bool m_energy;
     size_t m_nv;
+
+    vector_fp m_advancelimits; //!< Advance step limit
 
     // Data associated each sensitivity parameter
     std::vector<SensitivityParameter> m_sensParams;
--- a/include/cantera/zeroD/ReactorNet.h
+++ b/include/cantera/zeroD/ReactorNet.h
@@ -1,7 +1,7 @@
 //! @file ReactorNet.h
 
 // This file is part of Cantera. See License.txt in the top-level directory or
-// at http://www.cantera.org/license.txt for license and copyright information.
+// at https://cantera.org/license.txt for license and copyright information.
 
 #ifndef CT_REACTORNET_H
 #define CT_REACTORNET_H
@@ -80,6 +80,17 @@ public:
      * @param time Time to advance to (s).
      */
     void advance(doublereal time);
+
+    /**
+     * Advance the state of all reactors in time. Take as many internal
+     * timesteps as necessary towards *time*. If *applylimit* is true,
+     * the advance step will be automatically reduced if needed to
+     * stay within limits (set by setAdvanceLimit).
+     * Returns the time at the end of integration.
+     * @param time Time to advance to (s).
+     * @param applylimit Limit advance step (boolean).
+     */
+    double advance(double time, bool applylimit);
 
     //! Advance the state of all reactors in time.
     double step();
@@ -164,6 +175,9 @@ public:
 
     virtual void getState(doublereal* y);
 
+    //! Return k-th derivative at the current time
+    virtual void getDerivative(int k, double* dky);
+
     virtual size_t nparams() {
         return m_sens_params.size();
     }
@@ -195,6 +209,10 @@ public:
         return m_paramNames.at(p);
     }
 
+    //! Initialize the reactor network. Called automatically the first time
+    //! advance or step is called.
+    void initialize();
+
     //! Reinitialize the integrator. Used to solve a new problem (different
     //! initial conditions) but with the same configuration of the reactor
     //! network. Can be called manually, or automatically after calling
@@ -222,10 +240,25 @@ public:
         return m_integ->maxSteps();
     }
 
+    //! Set absolute step size limits during advance
+    virtual void setAdvanceLimits(const double* limits);
+
+    //! Retrieve absolute step size limits during advance
+    virtual bool getAdvanceLimits(double* limits);
+
 protected:
-    //! Initialize the reactor network. Called automatically the first time
-    //! advance or step is called.
-    void initialize();
+
+    //! Estimate a future state based on current derivatives.
+    //! The function is intended for internal use by ReactorNet::advance
+    //! and deliberately not exposed in external interfaces.
+    virtual void getEstimate(double time, int k, double* yest);
+
+    //! Returns the order used for last solution step of the ODE integrator
+    //! The function is intended for internal use by ReactorNet::advance
+    //! and deliberately not exposed in external interfaces.
+    virtual int lastOrder() {
+        return m_integ->lastOrder();
+    }
 
     std::vector<Reactor*> m_reactors;
     std::unique_ptr<Integrator> m_integ;
@@ -251,6 +284,8 @@ protected:
     std::vector<std::string> m_paramNames;
 
     vector_fp m_ydot;
+    vector_fp m_yest;
+    vector_fp m_advancelimits;
 };
 }
 
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -502,7 +502,7 @@ cdef extern from "cantera/zerodim.h" namespace "Cantera":
         size_t neq()
         void getState(double*)
         void addSurface(CxxReactorSurface*)
-
+        void setAdvanceLimit(string&, double) except +translate_exception
         void addSensitivityReaction(size_t) except +translate_exception
         void addSensitivitySpeciesEnthalpy(size_t) except +translate_exception
         size_t nSensParams()
@@ -580,8 +580,9 @@ cdef extern from "cantera/zerodim.h" namespace "Cantera":
     cdef cppclass CxxReactorNet "Cantera::ReactorNet":
         CxxReactorNet()
         void addReactor(CxxReactor&)
-        void advance(double) except +translate_exception
+        double advance(double, cbool) except +translate_exception
         double step() except +translate_exception
+        void initialize() except +translate_exception
         void reinitialize() except +translate_exception
         double time()
         void setInitialTime(double)
@@ -596,8 +597,11 @@ cdef extern from "cantera/zerodim.h" namespace "Cantera":
         void setVerbose(cbool)
         size_t neq()
         void getState(double*)
+        void getDerivative(int, double *) except +translate_exception
+        void setAdvanceLimits(double*)
+        cbool getAdvanceLimits(double*)
         string componentName(size_t) except +translate_exception
-
+        size_t globalComponentIndex(string&, int) except +translate_exception
         void setSensitivityTolerances(double, double)
         double rtolSensitivity()
         double atolSensitivity()
--- a/interfaces/cython/cantera/examples/reactors/reactor1.py
+++ b/interfaces/cython/cantera/examples/reactors/reactor1.py
@@ -12,16 +12,22 @@ gas.TPX = 1001.0, ct.one_atm, 'H2:2,O2:1,N2:4'
 r = ct.IdealGasConstPressureReactor(gas)
 
 sim = ct.ReactorNet([r])
-time = 0.0
+sim.verbose = True
+
+# limit advance when temperature difference is exceeded
+delta_T_max = 20.
+r.set_advance_limit('temperature', delta_T_max)
+
+dt_max = 1.e-5
+t_end = 100 * dt_max
 states = ct.SolutionArray(gas, extra=['t'])
 
-print('%10s %10s %10s %14s' % ('t [s]','T [K]','P [Pa]','u [J/kg]'))
-for n in range(100):
-    time += 1.e-5
-    sim.advance(time)
-    states.append(r.thermo.state, t=time*1e3)
-    print('%10.3e %10.3f %10.3f %14.6e' % (sim.time, r.T,
-                                           r.thermo.P, r.thermo.u))
+print('{:10s} {:10s} {:10s} {:14s}'.format('t [s]','T [K]','P [Pa]','u [J/kg]'))
+while sim.time < t_end:
+    sim.advance(sim.time + dt_max)
+    states.append(r.thermo.state, t=sim.time*1e3)
+    print('{:10.3e} {:10.3f} {:10.3f} {:14.6f}'.format(sim.time, r.T,
+                                                       r.thermo.P, r.thermo.u))
 
 # Plot the results if matplotlib is installed.
 # See http://matplotlib.org/ to get it.
--- a/interfaces/cython/cantera/reactor.pyx
+++ b/interfaces/cython/cantera/reactor.pyx
@@ -315,6 +315,17 @@ cdef class Reactor(ReactorBase):
         self.reactor.getState(&y[0])
         return y
 
+    def set_advance_limit(self, name, limit):
+        """
+        Limit absolute change of component *name* during `ReactorNet.advance`.
+        (positive *limit* values are considered; negative values disable a
+        previously set advance limit for a solution component). Note that
+        limits are disabled by default (with individual values set to -1.).
+        """
+        if limit is None:
+            limit = -1.
+        self.reactor.setAdvanceLimit(stringify(name), limit)
+
 
 cdef class Reservoir(ReactorBase):
     """
@@ -859,12 +870,17 @@ cdef class ReactorNet:
         self._reactors.append(r)
         self.net.addReactor(deref(r.reactor))
 
-    def advance(self, double t):
+    def advance(self, double t, pybool apply_limit=True):
         """
-        Advance the state of the reactor network in time from the current
-        time to time *t* [s], taking as many integrator timesteps as necessary.
+        Advance the state of the reactor network in time from the current time
+        towards time *t* [s], taking as many integrator timesteps as necessary.
+        If *apply_limit* is true and an advance limit is specified, the reactor
+        state at the end of the timestep is estimated prior to advancing. If
+        the difference exceed limits, the end time is reduced by half until
+        the projected end state remains within specified limits.
+        Returns the time reached at the end of integration.
         """
-        self.net.advance(t)
+        return self.net.advance(t, apply_limit)
 
     def step(self):
         """
@@ -872,6 +888,12 @@ cdef class ReactorNet:
         returned.
         """
         return self.net.step()
+
+    def initialize(self):
+        """
+        Force initialization of the integrator after initial setup.
+        """
+        self.net.initialize()
 
     def reinitialize(self):
         """
@@ -965,6 +987,17 @@ cdef class ReactorNet:
             return pybool(self.net.verbose())
         def __set__(self, pybool v):
             self.net.setVerbose(v)
+
+    def global_component_index(self, name, int reactor):
+        """
+        Returns the index of a component named *name* of a reactor with index
+        *reactor* within the global state vector. I.e. this determines the
+        (absolute) index of the component, where *reactor* is the index of the
+        reactor that holds the component. *name* is either a species name or the
+        name of a reactor state variable, e.g. 'int_energy', 'temperature', etc.
+        depending on the reactor's equations.
+        """
+        return self.net.globalComponentIndex(stringify(name), reactor)
 
     def component_name(self, int i):
         """
@@ -1070,6 +1103,38 @@ cdef class ReactorNet:
         cdef np.ndarray[np.double_t, ndim=1] y = np.zeros(self.n_vars)
         self.net.getState(&y[0])
         return y
+
+    def get_derivative(self, k):
+        """
+        Get the k-th time derivative of the state vector of the reactor network.
+        """
+        if not self.n_vars:
+            raise CanteraError('ReactorNet empty or not initialized.')
+        cdef np.ndarray[np.double_t, ndim = 1] dky = np.zeros(self.n_vars)
+        self.net.getDerivative(k, & dky[0])
+        return dky
+
+    property advance_limits:
+        """
+        Get or set absolute limits for state changes during `ReactorNet.advance`
+        (positive values are considered; negative values disable a previously
+        set advance limit for a solution component). Note that limits are
+        disabled by default (with individual values set to -1.).
+        """
+        def __get__(self):
+            cdef np.ndarray[np.double_t, ndim=1] limits = np.empty(self.n_vars)
+            self.net.getAdvanceLimits(&limits[0])
+            return limits
+
+        def __set__(self, limits):
+            if limits is None:
+                limits = -1. * np.ones([self.n_vars])
+            elif len(limits) != self.n_vars:
+                raise ValueError('array must be of length n_vars')
+
+            cdef np.ndarray[np.double_t, ndim=1] data = \
+                np.ascontiguousarray(limits, dtype=np.double)
+            self.net.setAdvanceLimits(&data[0])
 
     def advance_to_steady_state(self, int max_steps=10000,
                                 double residual_threshold=0., double atol=0.,
--- a/interfaces/cython/cantera/test/test_reactor.py
+++ b/interfaces/cython/cantera/test/test_reactor.py
@@ -121,6 +121,22 @@ class TestReactor(utilities.CanteraTest):
         self.assertNear(P1, self.r1.thermo.P)
         self.assertNear(P2, self.r2.thermo.P)
 
+    def test_derivative(self):
+        T1, P1 = 300, 101325
+
+        self.make_reactors(n_reactors=1, T1=T1, P1=P1)
+        self.net.advance(1.0)
+
+        # compare cvode derivative to numerical derivative
+        dydt = self.net.get_derivative(1)
+        dt = -self.net.time
+        dy = -self.net.get_state()
+        self.net.step()
+        dt += self.net.time
+        dy += self.net.get_state()
+        for i in range(self.net.n_vars):
+            self.assertNear(dydt[i], dy[i]/dt)
+
     def test_timestepping(self):
         self.make_reactors()
 
@@ -204,6 +220,63 @@ class TestReactor(utilities.CanteraTest):
 
         self.assertTrue(n_baseline > n_rtol)
         self.assertTrue(n_baseline > n_atol)
+
+    def test_advance_limits(self):
+        P0 = 10 * ct.one_atm
+        T0 = 1100
+        X0 = 'H2:1.0, O2:0.5, AR:8.0'
+        self.make_reactors(n_reactors=1, T1=T0, P1=P0, X1=X0)
+
+        limit_H2 = .01
+        ix = self.net.global_component_index('H2', 0)
+        self.r1.set_advance_limit('H2', limit_H2)
+        self.assertEqual(self.net.advance_limits[ix], limit_H2)
+
+        self.r1.set_advance_limit('H2', None)
+        self.assertEqual(self.net.advance_limits[ix], -1.)
+
+        self.r1.set_advance_limit('H2', limit_H2)
+        self.net.advance_limits = None
+        self.assertEqual(self.net.advance_limits[ix], -1.)
+
+        self.r1.set_advance_limit('H2', limit_H2)
+        self.net.advance_limits = 0 * self.net.advance_limits - 1.
+        self.assertEqual(self.net.advance_limits[ix], -1.)
+
+    def test_advance_with_limits(self):
+        def integrate(limit_H2 = None, apply=True):
+            P0 = 10 * ct.one_atm
+            T0 = 1100
+            X0 = 'H2:1.0, O2:0.5, AR:8.0'
+            self.make_reactors(n_reactors=1, T1=T0, P1=P0, X1=X0)
+            if limit_H2 is not None:
+                self.r1.set_advance_limit('H2', limit_H2)
+                ix = self.net.global_component_index('H2', 0)
+                self.assertEqual(self.net.advance_limits[ix], limit_H2)
+
+            tEnd = 1.0
+            tStep = 1.e-3
+            nSteps = 0
+
+            t = tStep
+            while t < tEnd:
+                t_curr = self.net.advance(t, apply_limit=apply)
+                nSteps += 1
+                if t_curr == t:
+                    t += tStep
+
+            return nSteps
+
+        n_baseline = integrate()
+        n_advance_coarse = integrate(.01)
+        n_advance_fine = integrate(.001)
+        n_advance_negative = integrate(-.001)
+        n_advance_override = integrate(.001, False)
+
+        self.assertTrue(n_advance_coarse > n_baseline)
+        self.assertTrue(n_advance_fine > n_advance_coarse)
+        self.assertTrue(n_advance_negative == n_baseline)
+        self.assertTrue(n_advance_override == n_baseline)
 
     def test_heat_transfer1(self):
         # Connected reactors reach thermal equilibrium after some time
--- a/src/zeroD/Reactor.cpp
+++ b/src/zeroD/Reactor.cpp
@@ -103,6 +103,7 @@ void Reactor::initialize(doublereal t0)
         }
     }
     m_work.resize(maxnt);
+    m_advancelimits.resize(m_nv, -1.0);
 }
 
 size_t Reactor::nSensParams()
@@ -427,6 +428,52 @@ void Reactor::resetSensitivity(double* params)
     if (m_kin) {
         m_kin->invalidateCache();
     }
+}
+
+void Reactor::setAdvanceLimits(const double *limits)
+{
+    if (m_thermo == 0) {
+        throw CanteraError("getState",
+                           "Error: reactor is empty.");
+    }
+    for (size_t j = 0; j < m_nv; j++) {
+        m_advancelimits[j] = limits[j];
+    }
+}
+
+bool Reactor::getAdvanceLimits(double *limits)
+{
+    bool has_limit = false;
+
+    for (size_t j = 0; j < m_nv; j++) {
+        limits[j] = m_advancelimits[j];
+        has_limit |= (limits[j] > 0.);
+    }
+
+    return has_limit;
+}
+
+void Reactor::setAdvanceLimit(const string& nm, const double limit)
+{
+    size_t k = componentIndex(nm);
+
+    if (m_thermo == 0) {
+        throw CanteraError("getState",
+                           "Error: reactor is empty.");
+    }
+    if (m_nv == 0) {
+        if (m_net == 0) {
+            throw CanteraError("Reactor::setAdvanceLimit",
+                               "Cannot set limit on a reactor that is not "
+                               "assigned to a ReactorNet object.");
+        } else {
+            m_net->initialize();
+        }
+    } else if (k > m_nv) {
+        throw CanteraError("Reactor::setAdvanceLimit",
+                           "Index out of bounds.");
+    }
+    m_advancelimits[k] = limit;
 }
 
 }
--- a/src/zeroD/ReactorNet.cpp
+++ b/src/zeroD/ReactorNet.cpp
@@ -1,7 +1,7 @@
 //! @file ReactorNet.cpp
 
 // This file is part of Cantera. See License.txt in the top-level directory or
-// at http://www.cantera.org/license.txt for license and copyright information.
+// at https://cantera.org/license.txt for license and copyright information.
 
 #include "cantera/zeroD/ReactorNet.h"
 #include "cantera/zeroD/FlowDevice.h"
@@ -98,6 +98,8 @@ void ReactorNet::initialize()
     }
 
     m_ydot.resize(m_nv,0.0);
+    m_yest.resize(m_nv,0.0);
+    m_advancelimits.resize(m_nv,-1.0);
     m_atol.resize(neq());
     fill(m_atol.begin(), m_atol.end(), m_atols);
     m_integ->setTolerances(m_rtol, neq(), m_atol.data());
@@ -136,6 +138,60 @@ void ReactorNet::advance(doublereal time)
     updateState(m_integ->solution());
 }
 
+double ReactorNet::advance(double time, bool applylimit)
+{
+    if (!m_init) {
+        initialize();
+    } else if (!m_integrator_init) {
+        reinitialize();
+    }
+
+    if (!applylimit) {
+        // take full step
+        advance(time);
+        return time;
+    }
+
+    bool limitadvance = getAdvanceLimits(m_advancelimits.data());
+    if (!limitadvance) {
+        // take full step
+        advance(time);
+        return time;
+    }
+
+    // ensure that gradient is available
+    while (lastOrder() < 1) {
+        step();
+    }
+
+    int k = lastOrder();
+    double t = time, delta;
+    double* y = m_integ->solution();
+
+    // reduce time step if limits are exceeded
+    while (true) {
+        bool exceeded = false;
+        getEstimate(t, k, &m_yest[0]);
+        for (size_t j = 0; j < m_nv; j++) {
+            delta = abs(m_yest[j] - y[j]);
+            if ( (m_advancelimits[j] > 0.) && ( delta > m_advancelimits[j]) ) {
+                exceeded = true;
+                if (m_verbose) {
+                    writelog("    Limiting global state vector component {:d} (dt = {:9.4g}):"
+                             "{:11.6g} > {:9.4g}\n",
+                             j, t - m_time, delta, m_advancelimits[j]);
+                }
+            }
+        }
+        if (!exceeded) {
+            break;
+        }
+        t = .5 * (m_time + t);
+    }
+    advance(t);
+    return t;
+}
+
 double ReactorNet::step()
 {
     if (!m_init) {
@@ -146,6 +202,26 @@ double ReactorNet::step()
     m_time = m_integ->step(m_time + 1.0);
     updateState(m_integ->solution());
     return m_time;
+}
+
+void ReactorNet::getEstimate(double time, int k, double* yest)
+{
+    // initialize
+    double* cvode_dky = m_integ->solution();
+    for (size_t j = 0; j < m_nv; j++) {
+        yest[j] = cvode_dky[j];
+    }
+
+    // Taylor expansion
+    double factor = 1.;
+    double deltat = time - m_time;
+    for (int n = 1; n <= k; n++) {
+        factor *= deltat / n;
+        cvode_dky = m_integ->derivative(m_time, n);
+        for (size_t j = 0; j < m_nv; j++) {
+            yest[j] += factor * cvode_dky[j];
+        }
+    }
 }
 
 void ReactorNet::addReactor(Reactor& r)
@@ -216,6 +292,35 @@ void ReactorNet::getState(double* y)
     for (size_t n = 0; n < m_reactors.size(); n++) {
         m_reactors[n]->getState(y + m_start[n]);
     }
+}
+
+void ReactorNet::getDerivative(int k, double* dky)
+{
+    double* cvode_dky = m_integ->derivative(m_time, k);
+    for (size_t j = 0; j < m_nv; j++) {
+        dky[j] = cvode_dky[j];
+    }
+}
+
+void ReactorNet::setAdvanceLimits(const double *limits)
+{
+    if (!m_init) {
+        initialize();
+    }
+    for (size_t n = 0; n < m_reactors.size(); n++) {
+        m_reactors[n]->setAdvanceLimits(limits + m_start[n]);
+    }
+}
+
+bool ReactorNet::getAdvanceLimits(double *limits)
+{
+    bool has_limit = false;
+
+    for (size_t n = 0; n < m_reactors.size(); n++) {
+        has_limit |= m_reactors[n]->getAdvanceLimits(limits + m_start[n]);
+    }
+
+    return has_limit;
 }
 
 size_t ReactorNet::globalComponentIndex(const string& component, size_t reactor)
--- a/include/cantera/numerics/CVodesIntegrator.h
+++ b/include/cantera/numerics/CVodesIntegrator.h
@@ -3,7 +3,7 @@
  */
 
 // This file is part of Cantera. See License.txt in the top-level directory or
-// at http://www.cantera.org/license.txt for license and copyright information.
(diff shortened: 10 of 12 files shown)


ischoegl:
I can add unit tests once I hear back.

ischoegl (commented):
The fix in commit 38ab0b1 is unrelated to the pull request, and simply fixes a warning during compilation.

ischoegl:
Based on first comments in #628, I believe the following adjustments will create a more intuitive interface:
 * change `bool ReactorNet::progress(time)` (returning success) to `double ReactorNet::advance_towards(time)` (returning the final time)
 * rename `set/getDeltaStateProgress` to `set/getAdvanceAtol` (and similar)

ischoegl:
I incorporated first feedback from #628 and renamed some functions for clarity (plus squashed commits). I can add unit tests once I hear back (edit: rudimentary unit tests are now added).

interfaces/cython/cantera/examples/reactors/reactor3.py
```diff
@@ -0,0 +1,61 @@
...
+print('%10s %10s %10s %14s' % ('t [s]','T [K]','P [Pa]','u [J/kg]'))
+for n in range(100):
+    time += 1.e-5
+    while sim.advance_towards(time) < time:
+        states.append(r.thermo.state, t=sim.time*1e3)
+        print('%10.3e %10.3f %10.3f %14.6e' % (sim.time, r.T,
+                                               r.thermo.P, r.thermo.u))
+    states.append(r.thermo.state, t=time*1e3)
+    print('%10.3e %10.3f %10.3f %14.6e' % (sim.time, r.T,
+                                           r.thermo.P, r.thermo.u))
```
speth: You can avoid the nested loop and achieve essentially the same effect by writing this as:
```python
while sim.time < 1e-3:
    sim.advance_towards(sim.time + 1e-5)
    states.append(r.thermo.state, t=sim.time*1e3)
    print('%10.3e %10.3f %10.3f %14.6e' % (sim.time, r.T,
                                           r.thermo.P, r.thermo.u))
```

interfaces/cython/cantera/examples/reactors/reactor3.py
```diff
@@ -0,0 +1,61 @@
...
+r = ct.IdealGasConstPressureReactor(gas)
+
+sim = ct.ReactorNet([r])
+time = 0.0
+states = ct.SolutionArray(gas, extra=['t'])
+
+# limit advance when temperature difference is exceeded
+ix = r.component_index('temperature')
+delta_T_max = 20.
+sim.set_atol_advance(ix, delta_T_max)
```
speth: Since `component_index` returns the index within the variables for just this reactor, not the whole `ReactorNet`, this will only work for a network with just one reactor. To be able to set this quantity for a multi-reactor network, we need to add a function for getting the component index within the whole `ReactorNet`.

include/cantera/zeroD/ReactorNet.h
```diff
@@ -81,9 +81,21 @@ class ReactorNet : public FuncEval
...
+     * Returns the time at the end of integration.
+     * @param time Time to advance to (s).
+     */
+    double advanceTowards(double time);
+
     //! Advance the state of all reactors in time.
     double step();
 
+    //! Estimate a future state based on current derivatives.
+    virtual void estimate(double time, int k, double* yest);
```
speth: Parameters need to be described.

The usual convention in Cantera for functions which fill in an array that is passed in as an argument is for the name to start with `get`, e.g. `getEstimate`.

include/cantera/zeroD/ReactorNet.h
```diff
@@ -222,6 +237,20 @@ class ReactorNet : public FuncEval
         return m_integ->maxSteps();
     }
 
+    //! Returns the order used during the last solution step
```
speth: I would suggest clarifying that 'order' is the order used by the ODE integrator.

include/cantera/zeroD/ReactorNet.h
```diff
@@ -222,6 +237,20 @@ class ReactorNet : public FuncEval
...
+    virtual int lastOrder() {
+        return m_integ->lastOrder();
+    }
+
+    //! Set absolute tolerances during advanceTowards
+    virtual void setAtolAdvance(const double* atols);
```
speth: I'm a little ambivalent about the re-use of the term "tolerances" for this new feature, in addition to the absolute integrator tolerances. Would it make sense to refer to these as "thresholds", or something else?

interfaces/cython/cantera/examples/reactors/reactor3.py
```diff
@@ -0,0 +1,61 @@
+"""
+Constant-pressure, adiabatic kinetics simulation.
+Simulation is equivalent to reactor1.py, except for an automatically 
+reduced step size when large changes of temperature are anticipated.
+"""
```
speth: I would support using the new method in the existing example(s), rather than having the first examples that the users encounter showcase suboptimal ways of using Cantera.

src/zeroD/ReactorNet.cpp
```diff
@@ -148,6 +191,28 @@ double ReactorNet::step()
...
 }
 
+void ReactorNet::estimate(double time, int k, double* yest)
+{
+    double* cvode_dky;
+    double deltat = time - m_time;
+    double factor = 1.;
+
+    // initialize
+    cvode_dky = m_integ->solution();
```
speth: you can both declare and initialize `cvode_dky` on this line.

src/zeroD/ReactorNet.cpp
```diff
@@ -148,6 +191,28 @@ double ReactorNet::step()
...
+    for (size_t j = 0; j < m_nv; j++) {
+        yest[j] = cvode_dky[j];
+    }
+
+    // taylor expansion
+    for (int n = 1; n <= k; n++) {
+        factor *= deltat / ((double)n);
```
speth: I don't think the explicit cast is necessary - the result of dividing a `double` by an `int` should automatically be a `double`.

src/zeroD/ReactorNet.cpp
```diff
@@ -218,6 +283,45 @@ void ReactorNet::getState(double* y)
     }
 }
 
+void ReactorNet::getDerivative(int k, double* dky)
+{
+    double* cvode_dky = m_integ->derivative(m_time, k);
+    for (size_t j = 0; j < m_nv; j++) dky[j] = cvode_dky[j];
```
speth: All `for` loops should be multi-line, and get curly braces.

src/zeroD/ReactorNet.cpp
```diff
@@ -218,6 +283,45 @@ void ReactorNet::getState(double* y)
...
+}
+
+void ReactorNet::setAtolAdvance(const double *atols)
+{
+    if (!m_init) {
+        initialize();
+    }
+    m_limitadvance = false;
+    for (size_t j = 0; j < m_nv; j++) {
+        if (atols[j] > 0.) m_limitadvance = true;
```
speth: All `if` statements should be multi-line and get curly braces.

speth (changes requested):
I certainly understand the desire for control over the output interval when integrating, and the use of CVODES to extrapolate the solution in order to do so is an intriguing approach. I am a bit concerned about the complexity, however, given that you can get fairly similar control from very a very simple approach such as writing the integration loop as:
```python
tlast = 0
Tlast = 0
while sim.time < 1e-3:
    sim.step()
    if sim.time > tlast + 1e-5 or abs(Tlast - r.T) > delta_T_max:
        tlast = sim.time
        Tlast = r.T
        states.append(r.thermo.state, t=sim.time*1e3)
        print('%10.3e %10.3f %10.3f %14.6e' % (sim.time, r.T,
                                               r.thermo.P, r.thermo.u))
```
This doesn't do quite the same thing as the `advance_towards` method (it results in having output just after either condition is exceeded, rather than just before), but the effect is similar, and I'm not sure that this is any more complicated for the end user than the new `set_atol_advance` function introduced here.

One particular concern I have is that adding this would make it difficult for us to change to an alternate integrator instead of CVODES at some point in the future, by locking us into requiring an integrator which has the specific feature of being able to access information about the derivatives.

Perhaps what would make this more compelling would be to provide defaults for the tolerances such that decent performance could be obtained without the user having to do anything except replace `step` with `advance_towards`.

bryanwweber:
My skepticism I expressed about this change in the related issue remains. I've been thinking about this, but been swamped with other work lately. I'll try to make some time to collect my thoughts next week. Sorry for the delay! Thanks @speth for poking this back onto my radar.

interfaces/cython/cantera/examples/reactors/reactor3.py
bryanwweber: I would strongly prefer not to use the old-style `%` formatting. Since Python 3.5 is retired from Red Hat support this month (!), we should consider using the 3.6-and-up "f"-strings. If not that, at least the `str.format` method should be used.

interfaces/cython/cantera/examples/reactors/reactor3.py
ischoegl: This is correct: I used this here as it is the simplest approach for the given example. In general, a reverse lookup from `component_name` works as long as the user knows the structure of the `ReactorNet`, which, however, may require more Python skills than what we can expect from some users. I believe that overall the issue is similar for `sensitivies`: I'd welcome suggestions.

interfaces/cython/cantera/examples/reactors/reactor3.py
ischoegl: > I would strongly prefer not to use the old-style `%` formatting. 

Noted. Will convert when copying code going forward

speth: Python 3.5 isn't reaching end-of-life from python.org until [September 2020](https://devguide.python.org/#status-of-python-branches), so I think it's a little premature to make 3.6 the minimum requirement already.

ischoegl: > You can avoid the nested loop and achieve essentially the same effect by writing this as:
> 
> ```python
> while sim.time < 1e-3:
>     sim.advance_towards(sim.time + 1e-5)
>     states.append(r.thermo.state, t=sim.time*1e3)
>     print('%10.3e %10.3f %10.3f %14.6e' % (sim.time, r.T,
>                                            r.thermo.P, r.thermo.u))
> ```

It doesn't do quite the same: my intent for `advance_towards` was to generate output with prescribed step size when changes are moderate, and add interstitial points when needed, i.e. something that is more similar to `advance` than to `step`. 

As `sim.time` is moved forward by `advance_towards`,  you will end up with different output times, as the 1e-5 will always be added to the last time. For this example, I wanted to get the same output points as the regular `advance` with added points when needed.

src/zeroD/ReactorNet.cpp
ischoegl: OK. While not required, I was not following the established style. Will be fixed.

src/zeroD/ReactorNet.cpp
ischoegl: OK. While not required, I was not following the established style. Will be fixed.

src/zeroD/ReactorNet.cpp
ischoegl: Old habits. I may misremember compilers complaining eons ago. Will check on this.

ischoegl:
@speth ... thank you for the review, but I will wait for @bryanwweber's comments before continuing.

To briefly restate my motivation for this PR: I specifically wanted to improve the behaviour of `advance`, and not necessarily do another version of `step`. The overarching objective was to provide something simple that allows for straight-forward output control. (I do realize that there is the issue of `component_name` not having a reverse function for `ReactorNet` objects.) 

At the moment, there are three options for output handling:
 1. Use `step` and live with unnecessarily large data files,
 2. Use `advance` at the peril that essential processes may not be properly resolved, or
 3. Apply actual python skills to get what you want

Based on experience, (3) is unfortunately not something that can be expected when teaching an UG combustion class at a generic college, (1) is overkill: some students _will_ use Excel to process data (!), leaving (2) with the stated caveats. Hence my desire to improve upon (2).

Regarding the comment about default arguments: it is a reasonable idea. However, would it make sense to merge `advance_towards` with `advance`, leaving no step size reduction as the default (i.e. not changing current behaviour unless specified otherwise?). IMHO, introducing a simple way to get to indices within the reactor state vector (i.e. reverse of `component_name` which remains to be discussed) would create a very intuitive solution.

interfaces/cython/cantera/examples/reactors/reactor3.py
ischoegl: Will certainly do this when the opinions on the PR converge: I left it separate for the time being.

include/cantera/zeroD/ReactorNet.h
ischoegl: I have renamed this function multiple times thus far and will settle on any good suggestion.

include/cantera/zeroD/ReactorNet.h
ischoegl: noted.

include/cantera/zeroD/ReactorNet.h
ischoegl: noted.

interfaces/cython/cantera/examples/reactors/reactor3.py
bryanwweber: @speth Well, that's surprising. I assumed RHEL would be the last to support a given version of Python. Then `str.format` should be preferred, I suppose.

interfaces/cython/cantera/examples/reactors/reactor3.py
ischoegl: merged files.

interfaces/cython/cantera/examples/reactors/reactor3.py
ischoegl: Switched to `string.format`.

interfaces/cython/cantera/examples/reactors/reactor3.py
ischoegl: I defined a new function `ReactorNet.component_index`, allowing for
```
ix = sim.component_index(r.name, 'temperature')
```

ischoegl:
@speth and @bryanwweber ... at this point I resolved all but one of the requests, but am still hoping to obtain feedback. Following up on my comment from a week ago, i.e.

>  Regarding the comment about default arguments: it is a reasonable idea. However, would it make sense to merge `advance_towards` with `advance`, leaving no step size reduction as the default (i.e. not changing current behaviour unless specified otherwise?). IMHO, introducing a simple way to get to indices within the reactor state vector (i.e. reverse of `component_name` which remains to be discussed) would create a very intuitive solution.

I defined a new method `ReactorNet.component_index` (currently defined in cython, but can be easily done in C++ also), i.e.
```python
ix = sim.component_index(r.name, 'temperature')
```
 and am still suggesting to merge `advance` with `advance_towards`.

bryanwweber:
Right, thanks for your patience @ischoegl! My biggest concern with this change as implemented is that now we have yet another way for users to advance time in a simulation. Already, I see many people using `advance()` when they should be using `step()` and vice-versa. I think the example that @speth showed of the Python necessary to control the output timing (with the caveat that he noted) is easy enough to teach even to a generic US university undergrad. So, I'm not excited about adding a new method to do almost the same thing that we already have two ways of doing.

I'm also not keen on adding additional optional arguments to the `advance()` function if they will change the current behavior of the function. The current understanding of `advance()` is that it reaches the end time and changing that behavior by adding some arguments seems like it might cause trouble, i.e., the user has to understand that sometimes this function reaches the end time, and sometimes it doesn't. In this sense, I much prefer the name `advance_towards()`.

That said, the way I would like to see it work if optional arguments are added to `advance()` is even with tolerances, the integrator still advanced all the way to the passed in time and returned an array with the collected values. Bonus points if it returns a `SolutionArray` in Python :smile:

Overall, I'm a soft :-1: on this change right now. If it can be merged with `advance()` in a useful way, then I could be more on board. Before this gets merged, a Matlab implementation will also be needed.

On a related but separate note, I think the same task could be accomplished by adding a `filter()` method to `SolutionArray` to reduce a set of data given some user-specified tolerances. Could also add a new method or an argument to `advance()` to automatically return a `SolutionArray` filled with data taken by `step`ping.

bryanwweber:
I want to add some more thoughts to my last comment. I should have started by saying I think this is a valuable idea before launching into my concerns :smile: By "soft" disagreement, I mainly meant that if we could come to consensus that the cognitive overhead for users is offset by the gain in functionality, then I would be on board. My concern is that, as currently implemented, that balance is on the side of too much overhead.

However, I had another thought. I think the general idea of the user saying "I want to integrate for x units" where "x" and "units" are described by the user has great utility. I think framing it in terms of limiting the **time** that the solver advances is somewhat confusing though, and the idea that the `advance_towards()` method might not reach the end time confused me for a while when this was first proposed.

So what if the `advance()` method were modified to take two arguments, a double and a string/int, where the double is the increment and the string/int is a reactor component name/index? The default would remain the same, passing in just the first argument is interpreted as a time, and the solver advances until it reaches that time. But the user could also do `advance(50., 'temperature')` and advance 50 K in temperature units, and this would have exactly the same conceptual meaning as `advance(0.5, 'time')` (which is the same as `advance(0.5)`). That should hopefully avoid some of the concerns I raised in my earlier point while still staying with the spirit of this PR.

This could be largely the same implementation as is currently here, just presented to the user differently. That said, I would be interested in a comparison with an interpolation method rather than (or in addition to) the current gradient estimation method, i.e. similar to the one done by `CVode` when the passed value of `tout` is exceeded.  Another comparison would be repeatedly calling `step()` and checking when the limit is exceeded; I'd be interested to know if this would be much slower than `CVode` with the `CV_NORMAL` argument if everything was done in the C++ layer (I think that returning control to Python to do the comparison would be quite a bit slower, but I'm curious if leaving everything in C++ would incur much penalty).

ischoegl:
Thanks for your detailed thoughts, @bryanwweber! I believe the suggestion of `advance(50., 'temperature')` deserves thought, although time will have to be part of an advance step: otherwise the next output may never be reached. I also agree that including a third way of integrating (beyond `step` and `advance`) is not a good idea. C++ allows for different return types as long as the argument list is distinct, which provides an avenue for a solution.

Regarding `SolutionArray`: I have significant concerns about this approach, as the object is only defined in Cython and lacks a C++ core implementation. Speed of compiled cython code may be adequate (although output handling can be an issue: see e.g. the `ic_engine.py` example). Nevertheless, there are significant repercussions when it comes to suitable Matlab implementations. I.e. while I am not opposed to the idea of returning `SolutionArray` objects, there is a lack of infrastructure.

One thing that I am interested in is the acceptance for creating an interface for time derivatives computed by CVODE (which my current implementation hinges on). If there is agreement that it is a useful feature to be broken out in the `ReactorNet` interface, I'd propose tabling the discussion for `advance`, and to restrict this PR to implementing the `ReactorNet.get_derivative` and/or `ReactorNet.get_estimate` portion.

bryanwweber:
> although time will have to be part of an advance step: otherwise the next output may never be reached

Yes, I avoided some of the difficulty in the implementation of the idea as an initial thought 😃 

> there are significant repercussions when it comes to suitable Matlab implementations. 

The idea would be that C++ would return (a pointer to) an array, and the separate implementations would decide what to do with that, e.g., Cython could turn that into a `SolutionArray`, Matlab could do whatever was most natural in Matlab, etc.

> One thing that I am interested in is the acceptance for creating an interface for time derivatives computed by CVODE

I don't immediately see the utility of a general interface to this beyond the capability that you've shown in this PR, can you provide some other examples?

Once we had converged on an idea, I was actually going to suggest removing at least `get_estimate()` from the public-facing API, and probably also `get_derivative()` as well. The names imply that they would be useful for end users to work with, which I don't think is the case. Especially the `get_derivative()` name would be confusing because there are many kinds of derivatives that would be very useful (e.g., dT/dt, dP/dt, etc.) that this function is not intended to compute (at least, not to my understanding, please correct me if that's wrong).

ischoegl:
> Especially the `get_derivative()` name would be confusing because there are many kinds of derivatives that would be very useful (e.g., dT/dt, dP/dt, etc.) that this function is not intended to compute

In the current implementation, `ReactorNet.get_derivative` retrieves the k-th time derivative from CVODE, i.e. it does indeed compute dT/dt for certain reactor types, together with other time derivatives. Specifically, it is the derivative of the composite state vector of the reactor network (which itself is interfaced via `ReactorNet.get_state`). The function `get_estimate` would be a Taylor series expansion to estimate the reactor network state at a future time.

PS: one thing I have been wondering about in a loosely related context is the choice of the _Feature Scaling_ approach in `advance_to_steady_state`. The step size is chosen by CVODE and thus neither constant nor defined? It obviously works, but convergence appears to depend on the CVODE integrator state as much as it does on `residual_threshold`.

bryanwweber:
> In the current implementation, ReactorNet.get_derivative retrieves the k-th time derivative from CVODE, i.e. it does indeed compute dT/dt for certain reactor types, together with other time derivatives.

OK, so that may have some utility, if a user is interested in dT/dt or d[X]/dt (given the current state vectors).

> The function get_estimate would be a Taylor series expansion to estimate the reactor network state at a future time.

Right, so how is that useful for a user? Why wouldn't they just have the integrator do the integration?

> `advance_to_steady_state()`

I'm not sure what you're asking, but the Users' Group is probably a better place for general questions about features (unless you think it should be an Issue on GitHub). I'd like to keep this discussion focused on the features you're proposing. Thanks 😄

ischoegl:
Alright. I’ll have to think about this for some time: I don’t see a clear path forward and may just limit this PR to get_derivative. 

Regarding advance_to_steady_state, it really was a rhetorical question, as convergence is poorly defined in the current code. It’s a minor issue not worth raising by itself, but a more rigorous definition could build on an actual time derivative, which is why I brought it up.

speth:
> [BW] That said, the way I would like to see it work if optional arguments are added to advance() is even with tolerances, the integrator still advanced all the way to the passed in time and returned an array with the collected values. Bonus points if it returns a SolutionArray in Python

I'm not sure there's a useful, general way to implement automatic collection of integrator states. In C++, there's not an obvious, efficient container, and in Python, SolutionArray would only be useful for a network consisting of a single reactor.

> [BW] Before this gets merged, a Matlab implementation will also be needed.

Given the current state of the Matlab toolbox, how difficult adding things is, and how often ideas for improving it involve effectively starting over, I would not insist on this.

> [BW] That said, I would be interested in a comparison with an interpolation method rather than (or in addition to) the current gradient estimation method, i.e. similar to the one done by CVode when the passed value of tout is exceeded.

That's an interesting thought, and CVODE already provides access to this interpolation via the `CVodeGetDky` function. Doing something like this would be a little complicated, though. For example, you would have to step past your threshold for (say) temperature change and interpolate back to an output time that satisfies your threshold. However, the next call to `advance` or `advance_towards` could end up being either within this interval, or beyond it and require taking additonal steps.

> [IS] One thing that I am interested in is the acceptance for creating an interface for time derivatives computed by CVODE (which my current implementation hinges on).

While I can see that the computed derivative function does seem to behave reasonably well in your extrapolation method, I think that this polynomial is only meant to apply to the time interval between the last output time and the current time, given the way the BDF method is defined. So, I'm a little concerned about providing a method whose main purpose is to use this data outside the time interval for which it is applicable.

bryanwweber:
@speth do you have any thoughts about my concern about adding another user interface method for integration?

> However, the next call to `advance` or `advance_towards` could end up being either within this interval, or beyond it and require taking additonal steps.

I'm not sure I understand what you're saying here. Why is this a concern?

ischoegl:
> >    [BW] That said, the way I would like to see it work if optional arguments are added to advance() is even with tolerances, the integrator still advanced all the way to the passed in time and returned an array with the collected values. Bonus points if it returns a SolutionArray in Python

> I'm not sure there's a useful, general way to implement automatic collection of integrator states. In C++, there's not an obvious, efficient container, and in Python, SolutionArray would only be useful for a network consisting of a single reactor.

I didn't have a good idea here either, especially as there is no good data structure candidate in C++ implemented at this point (and I am not even sure whether implementing one makes sense: it definitely goes way beyond the original intent of this PR).

> >    [BW] That said, I would be interested in a comparison with an interpolation method rather than (or in addition to) the current gradient estimation method, i.e. similar to the one done by CVode when the passed value of tout is exceeded.

> That's an interesting thought, and CVODE already provides access to this interpolation via the `CVodeGetDky` function. Doing something like this would be a little complicated, though. For example, you would have to step past your threshold for (say) temperature change and interpolate back to an output time that satisfies your threshold. However, the next call to advance or advance_towards could end up being either within this interval, or beyond it and require taking additonal steps.

Agreed. `step` and `advance` are calling CVODE with different flags (`CV_NORMAL` and `CV_ONE_STEP`, respectively). A solution using interpolation would have to bypass `CV_NORMAL` completely. One of my questions here is whether this is developing too far away from the philosophy of the underlying integrator. 

> >    [IS] One thing that I am interested in is the acceptance for creating an interface for time derivatives computed by CVODE (which my current implementation hinges on).

> While I can see that the computed derivative function does seem to behave reasonably well in your extrapolation method, I think that this polynomial is only meant to apply to the time interval between the last output time and the current time, given the way the BDF method is defined. So, I'm a little concerned about providing a method whose main purpose is to use this data outside the time interval for which it is applicable.

Correct. The derivatives are accurate within the time interval of the last CVODE step, although I am not sure how they are internally calculated. What I used for the current implementation was a Taylor series expansion, with all its well known caveats. Time derivatives are, however, also of interest within other contexts: one would be my comment above about the current `advance_to_stead_state` implementation (which imho should use the derivative instead of an increment over an unspecified time interval in a more rigorous definition).

ischoegl:
> @speth do you have any thoughts about my concern about adding another user interface method for integration?
> 
> > However, the next call to `advance` or `advance_towards` could end up being either within this interval, or beyond it and require taking additonal steps.
> 
> I'm not sure I understand what you're saying here. Why is this a concern?

@bryanwweber This is linked to how CVODE is being called (i.e. with flags `CV_NORMAL` or `CV_ONE_STEP`). `CV_NORMAL` uses multiple single steps and interpolates within the last step to the targeted output time. Interpolation is only available for the last step, i.e. the implementation you are suggesting would have to bypass `CV_NORMAL` entirely. Thus, `advance_towards` would have to emulate the `CV_NORMAL` behavior within the C++ layer instead of using the native CVODE code.

ischoegl:
@speth and @bryanwweber ... I figure that the intent of my original PR was to treat the step size reduction via thresholds  (with bounds estimated using a Taylor series expansion). The implementation is straight-forward, and does not bypass CVODE internals.

Summarizing my other thoughts from before, I believe the following Cantera interface makes sense:
 * `ReactorNet.step()` as is
 * `ReactorNet.advance(tout)` as is
 * `ReactorNet.advance(tout, limit_step=True)` replacing `advance_towards`, with the step thresholds being set by `ReactorNet.set_advance_limit` (renamed from current `ReactorNet.set_atol_advance`), or a `ReactorNet.advance_limit` attribute (renamed from `ReactorNet.atol_advance`). If no thresholds are set, the behavior defaults to the regular `advance` behavior.
 * `ReactorNet.component_index(reactor_name, name)` locates components within various ReactorNet structures

That way, not reaching the output time will not be a surprise to the user, which is really the main concern that I can see.

Otherwise, I do not see a simple solution emerging from this discussion.

speth:
I like this proposal for the interface, using an optional argument to `advance`. 

I think I would prefer the `set_advance_limit` function to a property, since the most common way to use it would be setting the limit for a specific solution component. This name is definitely an improvement over `atol_advance`.

While adding the `ReactorNet.component_index` function would be useful in terms of completing the interface, you could avoid some awkward function calls if the `set_advance_limit` were also implemented on the `Reactor` object.

I also agree that having a way of calculating the derivative would be helpful for fixing the flawed implementation of `advance_to_steady_state` (as part of a separate PR, of course :smile:).

ischoegl:
> While adding the `ReactorNet.component_index` function would be useful in terms of completing the interface, you could avoid some awkward function calls if the `set_advance_limit` were also implemented on the `Reactor` object.

This will require some work (and affect more class definitions), but it does make sense.

> I also agree that having a way of calculating the derivative would be helpful for fixing the flawed implementation of `advance_to_steady_state` (as part of a separate PR, of course smile).

I will open an issue (#654)

include/cantera/zeroD/ReactorNet.h
ischoegl: Will be renamed to `advance_limit`/`set_advance_limit`, per discussion.

ischoegl:
@speth ... the suggested change to the interface is implemented, together with your request to implement `set_advance_limit` for `Reactor` objects (as well as for `ReactorNet` objects). Regarding the comment about properties: I kept the attribute setter methods for the time being (I also kept `get_estimate` in cython). While some of those methods can be arguably removed from the cython interface, a few are being used internally, i.e. exposing them via cython may make sense.

interfaces/cython/cantera/examples/reactors/reactor1.py
```diff
@@ -1,5 +1,7 @@
 """
 Constant-pressure, adiabatic kinetics simulation.
+Simulation is equivalent to reactor1.py, except for an automatically
```
ischoegl: Needs updating (note to self).

interfaces/cython/cantera/reactor.pyx
```diff
@@ -315,6 +315,31 @@ cdef class Reactor(ReactorBase):
         self.reactor.getState(&y[0])
         return y
 
+    property advance_limits:
```
ischoegl: property could be removed from cython (although getter is needed in C++ layer).

interfaces/cython/cantera/reactor.pyx
```diff
@@ -873,6 +903,27 @@ cdef class ReactorNet:
         """
         return self.net.step()
 
+    def get_estimate(self, double t, int k=-1):
```
ischoegl: could be removed from cython (although needed internally)

interfaces/cython/cantera/reactor.pyx
```diff
@@ -966,6 +1017,17 @@ cdef class ReactorNet:
         def __set__(self, pybool v):
             self.net.setVerbose(v)
 
+    def global_component_index(self, name, int reactor):
```
ischoegl, line 991: this is now exposed to cython (it already existed in C++ layer)

interfaces/cython/cantera/reactor.pyx
```diff
@@ -1071,6 +1133,48 @@ cdef class ReactorNet:
...
+
+        def __set__(self, limits):
+            if len(limits) != self.n_vars:
+                raise ValueError('array must be of length n_vars')
+
+            cdef np.ndarray[np.double_t, ndim=1] data = \
+                np.ascontiguousarray(limits, dtype=np.double)
+            self.net.setAdvanceLimits(&data[0])
+
+    def set_advance_limit(self, int index, double limit):
```
ischoegl: could be removed

include/cantera/zeroD/ReactorNet.h
```diff
@@ -195,6 +212,10 @@ class ReactorNet : public FuncEval
         return m_paramNames.at(p);
     }
 
+    //! Initialize the reactor network. Called automatically the first time
```
ischoegl, line 212: this is now exposed in cython, as reactors need to be initialized before limits can be applied. I.e. automatic initialization happens too late.

ischoegl (commented):
I left a couple of comments.

include/cantera/numerics/Integrator.h
```diff
@@ -141,6 +141,18 @@ class Integrator
         return 0;
     }
 
+    //! n-th derivative of the output function at time tout.
+    virtual double* derivative(double tout, int n) {
+        warn("derivative");
+        return 0;
+    }
+
+    //! Order used during the last solution step
```
ischoegl, line 150: Needs change

interfaces/cython/cantera/reactor.pyx
```diff
@@ -859,12 +884,17 @@ cdef class ReactorNet:
         self._reactors.append(r)
         self.net.addReactor(deref(r.reactor))
 
-    def advance(self, double t):
+    def advance(self, double t, pybool apply_limit=True):
```
ischoegl, line 873: Default value true or false?

interfaces/cython/cantera/reactor.pyx
```diff
@@ -1071,6 +1133,48 @@ cdef class ReactorNet:
...
+        """
+        Get the k-th time derivative of the state vector of the reactor network.
+        """
+        if not self.n_vars:
+            raise CanteraError('ReactorNet empty or not initialized.')
+        cdef np.ndarray[np.double_t, ndim = 1] dky = np.zeros(self.n_vars)
+        self.net.getDerivative(k, & dky[0])
+        return dky
+
+    property last_order:
```
ischoegl: Could be removed from cython

interfaces/cython/cantera/reactor.pyx
bryanwweber: I think this should definitely be removed from Cython. Can you also add documentation to the C++ function that indicates the function should not be used to get an estimate of the solution a long time after the time the derivatives were calculated, and if users want the solution at a future time they should just integrate? And also that the function is deliberately not exposed in Cython? I want to make sure that our future selves know why that function is not in Cython, and that it was an intentional choice not a mistake 😄

interfaces/cython/cantera/reactor.pyx
bryanwweber: I think this is the preferred name instead of the property.

interfaces/cython/cantera/reactor.pyx
bryanwweber: False. We need to maintain backwards compatibility.

interfaces/cython/cantera/reactor.pyx
bryanwweber: I don't see a use for this outside the estimation routine. Do you have one in mind? If not, I would be in favor of removing.

interfaces/cython/cantera/reactor.pyx
```diff
@@ -315,6 +315,31 @@ cdef class Reactor(ReactorBase):
         self.reactor.getState(&y[0])
         return y
 
+    property advance_limits:
+        """
+        Get or set absolute limits for state changes during `ReactorNet.advance`
+        (positive values are considered; negative values are ignored).
```
bryanwweber: Should we raise or throw (since I guess it would be from C++) if negative values are passed? Hypothetically, negative values could be useful, e.g., fuel mole fraction decrease

interfaces/cython/cantera/reactor.pyx
```diff
@@ -859,12 +884,17 @@ cdef class ReactorNet:
...
-    def advance(self, double t):
+    def advance(self, double t, pybool apply_limit=True):
         """
-        Advance the state of the reactor network in time from the current
-        time to time *t* [s], taking as many integrator timesteps as necessary.
+        Advance the state of the reactor network in time from the current time
+        towards time *t* [s], taking as many integrator timesteps as necessary.
+        If *apply_limit* is true and an advance limit is specified, the reactor 
+        state at the end of the timestep is estimated prior to advancing. If 
+        the difference exceed limits, the step size is reduced by half until
```
bryanwweber: Technically, the step size is not reduced (the step size is still controlled by the integrator), right? It should the end time that is reduced by half.

interfaces/cython/cantera/test/test_reactor.py
```diff
@@ -205,6 +244,40 @@ def integrate(atol, rtol):
         self.assertTrue(n_baseline > n_rtol)
         self.assertTrue(n_baseline > n_atol)
 
+    def test_advance_with_limits(self):
```
bryanwweber, line 246: If negative values are ignored or throw, there should be a test case for that path as well.

interfaces/cython/cantera/reactor.pyx
ischoegl: Correct.

ischoegl: Correct.

ischoegl: Correct.

interfaces/cython/cantera/reactor.pyx
ischoegl: Eliminated.

interfaces/cython/cantera/reactor.pyx
ischoegl: Eliminated, as redundant (now handled by reactor object).

interfaces/cython/cantera/reactor.pyx
speth: A default of `True` is also backwards compatible, though, since by default there aren't any limits to apply.

interfaces/cython/cantera/reactor.pyx
ischoegl: Eliminated. Property now only defined for ReactorNet object.

interfaces/cython/cantera/_cantera.pxd
```diff
@@ -596,8 +600,13 @@ cdef extern from "cantera/zerodim.h" namespace "Cantera":
         void setVerbose(cbool)
         size_t neq()
         void getState(double*)
+        void getDerivative(int, double *) except +translate_exception
+        int lastOrder() except +translate_exception
+        void setAdvanceLimits(double*)
+        void getAdvanceLimits(double*)
```
speth: Return type declaration is incorrect.

include/cantera/zeroD/Reactor.h
```diff
@@ -132,6 +132,18 @@ class Reactor : public ReactorBase
     //! @see componentIndex()
     virtual std::string componentName(size_t k);
 
+    //! Set absolute step size limits during advance
+    virtual void setAdvanceLimits(const double* limits);
+
+    //! Retrieve absolute step size limits during advance
+    virtual bool getAdvanceLimits(double* limits);
```
speth, line 142: Return value should be documented. Specifying the length of the `limits` array (i.e. `nEquations()`) would also be useful.

include/cantera/zeroD/Reactor.h
```diff
@@ -162,7 +174,7 @@ class Reactor : public ReactorBase
 
     //! Get initial conditions for SurfPhase objects attached to this reactor
     virtual void getSurfaceInitialConditions(double* y);
-
+
```
speth: Trailing whitespace.

interfaces/cython/cantera/reactor.pyx
bryanwweber: OK, that's fair. When I wrote that comment, I was also thinking about whether it would make sense to automatically apply the limits if the user set a limit, or if they should be forced to pass `True` here. I couldn't decide between it being a useful default ("I set some limit, you should just automatically use them") and having the user be explicit. I don't have a strong opinion either way.

In either case, I think a test of what happens when limits are not specified and `True` is passed here (whether by default or not) would be good (i.e., is an error thrown, or does it just do the same thing as before?).

ischoegl: @speth is correct. After manually setting a limit, a user would expect the limit to be applied, which is why I set the default to `True`; the flag allows this to be disabled.

interfaces/cython/cantera/reactor.pyx
ischoegl: `get_estimate` always pulls the most recent derivative from CVODE. I moved the function to the `protected` section (together with `lastOrder`) and left comments that they are deliberately not exposed.

interfaces/cython/cantera/reactor.pyx
ischoegl: Limits are currently defined as absolute values (and thus apply to both increases and decreases). I thus was only checking for strictly positive values and set the default limit to -1 (no limit). I can accomplish the same behavior setting the default to 0 though, but believe -1 to be more intuitive.

interfaces/cython/cantera/test/test_reactor.py
ischoegl: done.

interfaces/cython/cantera/reactor.pyx
ischoegl: Going with "I set some limit, you should just automatically use them", I am leaving the default as `True`, but added a test showing the expected behavior when this is overridden by passing `False`.

ischoegl:
@bryanwweber and @speth ... I believe I took care of everything, but left the two items that were not trivial open (despite considering both as resolved). Let me know whether there is anything else.

interfaces/cython/cantera/reactor.pyx
bryanwweber: Is the current behavior retained when no limits are set but `True` is passed? Is there a test for that?

ischoegl: The current default behavior is preserved, and there is a test (see `n_baseline` in `test_advance_with_limits` ... the `integrate` function explicitly passes `True` to the `advance` call).

interfaces/cython/cantera/reactor.pyx
bryanwweber: OK, I think absolute values make sense. I do think the user should be informed somehow that negative values are ignored (apologies if they are and I missed it), especially if negative values will turn off the limit for a particular component. How does a user turn off the limiting for a component if they want to?

I think `-1` is a fine sentinel value for C++, but I think it might be nice to have a way to disable a particular limit more naturally in Python, perhaps by giving it the value `False`, `None`, and/or `0`?

interfaces/cython/cantera/reactor.pyx
bryanwweber: Great, thank you!

interfaces/cython/cantera/reactor.pyx
ischoegl: Specifying `None` is certainly intuitive. Implementation (and test) will follow shortly. In addition, I added specific information about the default value to the docstring.

ischoegl:
The last two remaining points are now addressed.

src/zeroD/Reactor.cpp
```diff
@@ -429,4 +430,46 @@ void Reactor::resetSensitivity(double* params)
...
+    }
+}
+
+bool Reactor::getAdvanceLimits(double *limits)
+{
+    bool has_limit = false;
+
+    for (size_t j = 0; j < m_nv; j++) {
+        limits[j] = m_advancelimits[j];
+	has_limit |= (limits[j] > 0.);
```
speth: This line is indented with a tab instead of spaces.

src/zeroD/Reactor.cpp
```diff
@@ -429,4 +430,46 @@ void Reactor::resetSensitivity(double* params)
...
+    size_t k = componentIndex(nm);
+
+    if (m_thermo == 0) {
+        throw CanteraError("getState",
+                           "Error: reactor is empty.");
+    }
+    if (m_nv == 0) {
+        throw CanteraError("Reactor::setAdvanceLimit",
+                           "Cannot set limit prior to initialization "
+			   "of ReactorNet object.");
```
speth: More tabs.

src/zeroD/ReactorNet.cpp
```diff
@@ -136,6 +138,54 @@ void ReactorNet::advance(doublereal time)
...
 
+double ReactorNet::advance(double time, bool applylimit)
+{
+    if (!m_init) {
+        initialize();
+    } else if (!m_integrator_init) {
+        reinitialize();
+    }
+
+    bool limitadvance = getAdvanceLimits(&m_advancelimits[0]);
```
speth:
```suggestion
    bool limitadvance = getAdvanceLimits(m_advancelimits.data());
```

src/zeroD/ReactorNet.cpp
```diff
@@ -136,6 +138,54 @@ void ReactorNet::advance(doublereal time)
...
+    if (!limitadvance || !applylimit) {
+        // take full step
+        advance(time);
```
speth, line 158: Perhaps this is an unnecessary optimization, but the call to `getAdvanceLimits` could be avoided if `applylimit` is `false`.

interfaces/cython/cantera/examples/reactors/reactor1.py
```diff
@@ -12,16 +12,26 @@
 r = ct.IdealGasConstPressureReactor(gas)
 
 sim = ct.ReactorNet([r])
+sim.initialize()
```
speth: I'm not keen on requiring users to explicitly call the `initialize` function. Is there a reason not to simply have `setAdvanceLimit` call `initialize` if needed, as we do in other methods such as `ReactorNet::advance` and `ReactorNet::globalComponentIndex`?

interfaces/cython/cantera/examples/reactors/reactor1.py
```diff
@@ -12,16 +12,26 @@
...
 
-print('%10s %10s %10s %14s' % ('t [s]','T [K]','P [Pa]','u [J/kg]'))
+print('{:10s} {:10s} {:10s} {:14s}'.format('t [s]','T [K]','P [Pa]','u [J/kg]'))
 for n in range(100):
     time += 1.e-5
-    sim.advance(time)
+    while sim.advance(time) < time:
+        states.append(r.thermo.state, t=sim.time*1e3)
+        print('{:10.3e} {:10.3f} {:10.3f} {:14.6f}'.format(sim.time, r.T,
+                                                           r.thermo.P, r.thermo.u))
```
speth: I realize it provides slightly different output, but I think the simpler integration loop that users will more often want to emulate would be one that limits the maximum Δt and maximum ΔT, but without necessarily needing outputs at a pre-defined set of times, e.g.
```python
while sim.time < t_end:
    sim.advance(sim.time + dt_max)
    states.append(...)
    print(...)
```

interfaces/cython/cantera/reactor.pyx
```diff
@@ -315,6 +315,16 @@ cdef class Reactor(ReactorBase):
         self.reactor.getState(&y[0])
         return y
 
+    def set_advance_limit(self, name, limit):
+        """
+        Limit absolute change of component *name* during `ReactorNet.advance`.
+        (positive *limit* values are considered; negative values are ignored). Note
+        that limits are disabled by default (with individual values set to -1.).
```
speth: I think it would be more clear to describe the handling of negative values as "Negative values disable a previously-set advance limit for a solution component" rather than saying that they are ignored.

speth (changes requested):

include/cantera/zeroD/Reactor.h
ischoegl: removed.

interfaces/cython/cantera/_cantera.pxd
ischoegl: corrected

include/cantera/zeroD/Reactor.h
ischoegl: added more detailed doc strings.

src/zeroD/Reactor.cpp
ischoegl: replaced.

src/zeroD/ReactorNet.cpp
ischoegl: that's reasonable.

src/zeroD/ReactorNet.cpp
ischoegl: done.

src/zeroD/Reactor.cpp
ischoegl: fixed.

interfaces/cython/cantera/examples/reactors/reactor1.py
ischoegl: changed.

interfaces/cython/cantera/reactor.pyx
ischoegl: done.

interfaces/cython/cantera/examples/reactors/reactor1.py
ischoegl: Good point. I opted for this as the `setAdvanceLimit` is called on a `Reactor` object, not `ReactorNet`. However, I found a way to eliminate this call by using a callback from `Reactor` via `Reactor:m_net`, which allows for the required `ReactorNet::initialize` call.

ischoegl:
@speth ... thanks for the review and comments: I believe everything is addressed. Also let me know if I should rebase the PR (afaik, it's not necessary).

speth (approved):

PR #631, opened 2019-04-27
Improve exception handling in Sim1D.solve

### LINKED ISSUE #615: Flame solver does not retain transport options after automatically expanding domain
When solving the freely propagating flame with the `auto` option, if the solver has to expand the domain to fit the flame, the settings for `transport_model` and `soret_enabled` are not retained, but are replaced with the values used in the first solving stage of `'Mix'` and `False`.

Example:
```py
import cantera as ct
print("Running Cantera version", ct.__version__)

gas = ct.Solution('gri30.cti', 'gri30_multi')
gas.set_equivalence_ratio(1.0, 'CH4', {'O2':1.0, 'N2':3.76})
gas.TP = 300, 101325
width = 0.01  # domain that is too narrow

flame = ct.FreeFlame(gas, width=width)
flame.set_refine_criteria(ratio=3, slope=0.1, curve=0.1)

flame.transport_model = 'Multi'
flame.soret_enabled = True

flame.solve(loglevel=1, auto=True)

Su0 = flame.u[0]
print("Flame Speed is: {:.2f} cm/s".format(Su0*100))
flame.show_stats()
print(flame.transport_model)
print(flame.soret_enabled)
```
Changing the initial domain width to 0.06 m eliminates the need to expand the domain, and results in the use of the multicomponent transport model with the Soret effect enabled for the final solve. This affects version 2.4.0 and newer, since the automatic grid expansion was added in that version.

ischoegl:
Fixes #615

During an `auto` solution by `Sim1D.solve`, flags for `m_do_soret` and `m_do_multidiffusion` are temporarily reset. If an exception is raised, they need to be restored. Affects `FreeFlame` and `BurnerFlame`.

- buffer and restore flags

Note: Affected exceptions (`DomainTooNarrow` / `FlameBlowoff`) are raised by functions defined in overloaded classes (`FreeFlame.solve` / `BurnerFlame.solve`), so buffering and restoring of flags within `Sim1D.solve` requires catching and re-raising of exceptions.

### DIFF
--- a/interfaces/cython/cantera/onedim.pyx
+++ b/interfaces/cython/cantera/onedim.pyx
@@ -881,6 +881,12 @@ cdef class Sim1D:
             self.sim.solve(loglevel, <cbool>refine_grid)
             return
 
+        def set_transport(multi):
+            self.gas.transport_model = multi
+            for dom in self.domains:
+                if isinstance(dom, _FlowBase):
+                    dom.set_transport(self.gas)
+
         have_user_tolerances = any(dom.have_user_tolerances for dom in self.domains)
         if have_user_tolerances:
             # Save the user-specified tolerances
@@ -894,16 +900,18 @@ cdef class Sim1D:
 
         # Do initial steps without Soret diffusion
         soret_doms = [dom for dom in self.domains if getattr(dom, 'soret_enabled', False)]
-        for dom in soret_doms:
-            dom.soret_enabled = False
+
+        def set_soret(soret):
+            for dom in soret_doms:
+                dom.soret_enabled = soret
+
+        set_soret(False)
 
         # Do initial solution steps without multicomponent transport
+        transport = self.gas.transport_model
         solve_multi = self.gas.transport_model == 'Multi'
         if solve_multi:
-            self.gas.transport_model = 'Mix'
-            for dom in self.domains:
-                if isinstance(dom, _FlowBase):
-                    dom.set_transport(self.gas)
+            set_transport('Mix')
 
         def log(msg, *args):
             if loglevel:
@@ -934,6 +942,11 @@ cdef class Sim1D:
             except CanteraError as e:
                 log(str(e))
                 solved = False
+            except Exception as e:
+                # restore settings before re-raising exception
+                set_transport(transport)
+                set_soret(True)
+                raise e
 
             # If initial solve using energy equation fails, fall back on the
             # traditional fixed temperature solve followed by solving the energy
@@ -947,6 +960,11 @@ cdef class Sim1D:
                 except CanteraError as e:
                     log(str(e))
                     solved = False
+                except Exception as e:
+                    # restore settings before re-raising exception
+                    set_transport(transport)
+                    set_soret(True)
+                    raise e
 
                 if solved:
                     log('Solving on {} point grid with energy equation re-enabled', N)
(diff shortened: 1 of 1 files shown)


interfaces/cython/cantera/onedim.pyx
```diff
@@ -1003,7 +1028,13 @@ cdef class Sim1D:
...
         # Final call with expensive options enabled
         if have_user_tolerances or solve_multi or soret_doms:
-            self.sim.solve(loglevel, <cbool>refine_grid)
+            try:
+                self.sim.solve(loglevel, <cbool>refine_grid)
+            except Exception as e:
+                # restore settings before re-raising exception
+                set_transport(self.gas, self.domains, solve_multi)
+                set_soret(soret_doms, solve_soret)
+                raise e
```
speth: I'm not sure I understand the need for this `try/except` block -- the settings should have their intended final values at this point whether or not this final call to `solve` succeeds, right?

interfaces/cython/cantera/onedim.pyx
```diff
@@ -894,16 +904,15 @@ cdef class Sim1D:
 
         # Do initial steps without Soret diffusion
         soret_doms = [dom for dom in self.domains if getattr(dom, 'soret_enabled', False)]
-        for dom in soret_doms:
-            dom.soret_enabled = False
+        solve_soret = len(soret_doms)>0
```
speth: You could define `set_soret` after this point, and then not need to pass `soret_doms` in as an argument.

interfaces/cython/cantera/onedim.pyx
```diff
@@ -881,6 +881,16 @@ cdef class Sim1D:
...
 
+        def set_soret(soret_doms, soret):
+            for dom in soret_doms:
+                dom.soret_enabled = soret
+
+        def set_transport(gas, doms, multi):
+            gas.transport_model = multi
+            for dom in doms:
+                if isinstance(dom, _FlowBase):
+                    dom.set_transport(gas)
```
speth: This function doesn't need `gas` and `doms` as arguments, since they always have the values `self.gas` and `self.domains`, respectively.

interfaces/cython/cantera/onedim.pyx
```diff
@@ -934,6 +943,11 @@ cdef class Sim1D:
             except CanteraError as e:
                 log(str(e))
                 solved = False
+            except Exception as e:
+                # restore settings before re-raising exception
+                set_transport(self.gas, self.domains, transport)
+                set_soret(soret_doms, solve_soret)
```
speth: Since `set_soret` will only enable the Soret term for domains where it was originally enabled, you can just pass `True` as the argument, rather than using the boolean `solve_soret`. Then, you can use just use the Pythonic `if soret_doms:` instead of defining `solve_soret` in the first place.

speth (changes requested):
Thanks, this fix makes sense to me. I just had a few minor suggestions for cleaning up the implementation.

interfaces/cython/cantera/onedim.pyx
ischoegl: correct.

interfaces/cython/cantera/onedim.pyx
ischoegl: ok. My objective was clarity / symmetry to `solve_multi`, but it is indeed not necessary.

interfaces/cython/cantera/onedim.pyx
ischoegl: done

interfaces/cython/cantera/onedim.pyx
ischoegl: done

ischoegl:
@speth ... thanks for the review! All requested fixes are applied.

speth (approved):

PR #632, opened 2019-04-28
Improve ReactorNet infrastructure for alternative FlowDevice and Wall objects

### LINKED ISSUE #618: Feature request: C++ factories for flow devices and walls
### Feature Request

A consistent use of `Factory` objects would simplify the inclusion of custom `Wall` or `FlowDevice` classes in C++. The `ReactorFactory` approach (mostly) avoids hard-coding, and illustrates how custom objects can be included in reactor networks.

### Cantera version
2.5.0a2

### Expected Behavior
Consistent handling of `ReactorNet` related objects in clib and cython interfaces allows for addition of new derived objects with minimal effort.

### Actual Behavior
Only `ReactorBase` objects are flexible. Instantiation of `FlowDevice` objects is hardcoded, whereas there are no alternatives to a single `Wall` class.

ischoegl:

Fixes #618 

- add a base class for `WallBase` objects, enabling alternatives to the `Wall` class.
- add factories for `FlowDevice` objects and `WallBase` objects.
- create a uniform approach for instantiation of `ReactorNet` related objects from the cython interface via factories (instead of using a factory for `ReactorBase`, hard-coded creation for `FlowDevice`, and a single `Wall` class handling everything wall related)
 - a consistent definition of factories allows for the compilation of external code against unmodified cantera with minimal effort (which is currently not convenient for `FlowDevice` objects and impossible for alternative `Wall` classes)

~__Comment:__ To illustrate, a `ReciprocatingPiston` wall object is introduced, which handles wall movement using native C++ code. Further, defining a class `Cylinder` (reactor with added geometry) and a class `Solenoid` (timed valve), the setup for an IC-engine type simulation (e.g. example `reactors/ic_engine.py`) becomes more intuitive:~

```python
# reactor with geometry information
r = ct.Cylinder(gas)

# create reciprocating piston
piston = ct.ReciprocatingPiston(r, ambient_air)
piston.omega = 2. * np.pi * f
# piston.rod_radius_ratio = 2.

# install piston and set geometry
r.install(piston)
r.bore = d_piston
r.displacement = V_H
r.compression_ratio = epsilon

# create solenoid acting as inlet valve
inlet_valve = ct.Solenoid(inlet, r, coeff=inlet_valve_coeff)
inlet_valve.timing = inlet_open, inlet_close
inlet_valve.set_master(piston)

# create solenoid acting as injector
injector_mfc = ct.Solenoid(injector, r, mass=injector_mass)
injector_mfc.timing = injector_open, injector_close
injector_mfc.set_master(piston)

# create solenoid acting as outlet valve
outlet_valve = ct.Solenoid(r, outlet, coeff=outlet_valve_coeff)
outlet_valve.timing = outlet_open, outlet_close
outlet_valve.set_master(piston)

# create a reactor network containing the cylinder
sim = ct.ReactorNet([r])
```
~Beyond more readable code (I use cantera's stock examples for class assignments), there are technical advantages: while the current `Func1` approach is very flexible, execution is considerably slower than native C++ code; Python code executed between `sim.advance(t)` is faster, but does not continuously update parameters, potentially introducing error if time steps are chosen too large.~

### DIFF
--- a/include/cantera/zeroD/FlowDevice.h
+++ b/include/cantera/zeroD/FlowDevice.h
@@ -1,7 +1,7 @@
 //! @file FlowDevice.h
 
 // This file is part of Cantera. See License.txt in the top-level directory or
-// at http://www.cantera.org/license.txt for license and copyright information.
+// at https://cantera.org/license.txt for license and copyright information.
 
 #ifndef CT_FLOWDEVICE_H
 #define CT_FLOWDEVICE_H
@@ -27,9 +27,7 @@ const int Valve_Type = 3;
 class FlowDevice
 {
 public:
-    FlowDevice() : m_mdot(0.0), m_func(0), m_type(0),
-        m_nspin(0), m_nspout(0),
-        m_in(0), m_out(0) {}
+    FlowDevice();
 
     virtual ~FlowDevice() {}
     FlowDevice(const FlowDevice&) = delete;
--- a/include/cantera/zeroD/FlowDeviceFactory.h
+++ b/include/cantera/zeroD/FlowDeviceFactory.h
@@ -1,0 +1,72 @@
+//! @file FlowDeviceFactory.h
+
+// This file is part of Cantera. See License.txt in the top-level directory or
+// at https://cantera.org/license.txt for license and copyright information.
+
+#ifndef FLOWDEVICE_FACTORY_H
+#define FLOWDEVICE_FACTORY_H
+
+#include "cantera/base/FactoryBase.h"
+#include "cantera/zeroD/FlowDevice.h"
+
+namespace Cantera
+{
+
+class FlowDeviceFactory : public Factory<FlowDevice>
+{
+public:
+    static FlowDeviceFactory* factory() {
+        std::unique_lock<std::mutex> lock(flowDevice_mutex);
+        if (!s_factory) {
+            s_factory = new FlowDeviceFactory;
+        }
+        return s_factory;
+    }
+
+    virtual void deleteFactory() {
+        std::unique_lock<std::mutex> lock(flowDevice_mutex);
+        delete s_factory;
+        s_factory = 0;
+    }
+
+    //! Create a new flow device by type identifier.
+    /*!
+     * @param n the type to be created.
+     */
+    virtual FlowDevice* newFlowDevice(int n);
+
+    //! Create a new flow device by type name.
+    /*!
+     * @param flowDeviceType the type to be created.
+     */
+    virtual FlowDevice* newFlowDevice(const std::string& flowDeviceType);
+
+    //! Register a new flow device type identifier.
+    /*!
+     * @param name the name of the flow device type.
+     * @param type the type identifier of the flow device.
+     * Integer type identifiers are used by clib and matlab interfaces.
+     */
+    void reg_type(const std::string& name, const int type) {
+        m_types[type] = name;
+    }
+
+protected:
+    //! Map containing flow device type identifier / type name pairs.
+    std::unordered_map<int, std::string> m_types;
+
+private:
+    static FlowDeviceFactory* s_factory;
+    static std::mutex flowDevice_mutex;
+    FlowDeviceFactory();
+};
+
+//! Create a FlowDevice object of the specified type
+inline FlowDevice* newFlowDevice(const std::string& model)
+{
+    return FlowDeviceFactory::factory()->newFlowDevice(model);
+}
+
+}
+
+#endif
--- a/include/cantera/zeroD/Wall.h
+++ b/include/cantera/zeroD/Wall.h
@@ -1,125 +1,79 @@
-//! @file Wall.h Header file for class Wall.
+//! @file Wall.h Header file for base class WallBase.
 
 // This file is part of Cantera. See License.txt in the top-level directory or
-// at http://www.cantera.org/license.txt for license and copyright information.
+// at https://cantera.org/license.txt for license and copyright information.
 
 #ifndef CT_WALL_H
 #define CT_WALL_H
 
 #include "cantera/base/ctexceptions.h"
-#include "cantera/numerics/Func1.h"
-#include "cantera/zeroD/ReactorBase.h"
 #include "cantera/zeroD/ReactorSurface.h"
+#include "cantera/zeroD/ReactorBase.h"
 
 namespace Cantera
 {
 
 class Kinetics;
 class SurfPhase;
+class Func1;
 
-//! Represents a wall between between two ReactorBase objects.
-/*!
- * Walls can move (changing the volume of the adjacent reactors) and allow heat
- * transfer between reactors.
+const int WallType = 1;
+
+/**
+ * Base class for 'walls' (walls, pistons, etc.) connecting reactors.
+ * @ingroup reactor0
  */
-class Wall
+class WallBase
 {
 public:
-    Wall();
+    WallBase();
 
-    virtual ~Wall() {}
-    Wall(const Wall&) = delete;
-    Wall& operator=(const Wall&) = delete;
+    virtual ~WallBase() {}
+    WallBase(const WallBase&) = delete;
+    WallBase& operator=(const WallBase&) = delete;
+
+    //! String indicating the wall model implemented. Usually
+    //! corresponds to the name of the derived class.
+    virtual std::string type() const {
+        return "WallBase";
+    }
 
     //! Rate of volume change (m^3/s) for the adjacent reactors.
     /*!
-     * The volume rate of change is given by
-     * \f[
-     *     \dot V = K A (P_{left} - P_{right}) + F(t)
-     * \f]
-     * where *K* is the specified expansion rate coefficient, *A* is the wall
-     * area, and *F(t)* is a specified function of time. Positive values for
-     * `vdot` correspond to increases in the volume of reactor on left, and
-     * decreases in the volume of the reactor on the right.
+     * This method is called by Reactor::evalWalls(). Base class method
+     * does nothing (i.e. constant volume), but may be overloaded.
      */
-    virtual doublereal vdot(doublereal t);
+    virtual double vdot(double t) {
+        return 0.0;
+    }
 
     //! Heat flow rate through the wall (W).
     /*!
-     * The heat flux is given by
-     * \f[
-     *     Q = h A (T_{left} - T_{right}) + A G(t)
-     * \f]
-     * where *h* is the heat transfer coefficient, *A* is the wall area, and
-     * *G(t)* is a specified function of time. Positive values denote a flux
-     * from left to right.
+     * This method is called by Reactor::evalWalls(). Base class method
+     * does nothing (i.e. adiabatic wall), but may be overloaded.
      */
-    virtual doublereal Q(doublereal t);
+    virtual double Q(double t) {
+        return 0.0;
+    }
 
-    //! Area in m^2.
-    doublereal area() {
+    //! Area in (m^2).
+    double area() {
         return m_area;
     }
 
     //! Set the area [m^2].
-    void setArea(doublereal a) {
-        m_area = a;
-        m_surf[0].setArea(a);
-        m_surf[1].setArea(a);
-    }
+    virtual void setArea(double a);
 
     //! Get the area [m^2]
+    /*!
+     * Redundant function (same as WallBase::area()).
+     * @deprecated To be removed after Cantera 2.5.
+     */
     double getArea() const {
+        warn_deprecated("WallBase::getArea()", 
+                        "To be removed after Cantera 2.5. "
+                        "Replace with WallBase::area().");        
         return m_area;
-    }
-
-    void setThermalResistance(doublereal Rth) {
-        m_rrth = 1.0/Rth;
-    }
-
-    //! Set the overall heat transfer coefficient [W/m^2/K].
-    void setHeatTransferCoeff(doublereal U) {
-        m_rrth = U;
-    }
-
-    //! Get the overall heat transfer coefficient [W/m^2/K].
-    double getHeatTransferCoeff() const {
-        return m_rrth;
-    }
-
-    //! Set the emissivity.
-    void setEmissivity(doublereal epsilon) {
-        if (epsilon > 1.0 || epsilon < 0.0) {
-            throw CanteraError("Wall::setEmissivity",
-                               "emissivity must be between 0.0 and 1.0");
-        }
-        m_emiss = epsilon;
-    }
-
-    double getEmissivity() const {
-        return m_emiss;
-    }
-
-    //! Set the wall velocity to a specified function of time
-    void setVelocity(Func1* f=0) {
-        if (f) {
-            m_vf = f;
-        }
-    }
-
-    //! Set the expansion rate coefficient.
-    void setExpansionRateCoeff(doublereal k) {
-        m_k = k;
-    }
-
-    //! Get the expansion rate coefficient
-    double getExpansionRateCoeff() const {
-        return m_k;
-    }
-
-    //! Specify the heat flux function \f$ q_0(t) \f$.
-    void setHeatFlux(Func1* q) {
-        m_qf = q;
     }
 
     //! Install the wall between two reactors or reservoirs
@@ -149,12 +103,118 @@ protected:
 
     std::vector<ReactorSurface> m_surf;
 
-    doublereal m_area, m_k, m_rrth;
-    doublereal m_emiss;
-    Func1* m_vf;
-    Func1* m_qf;
+    double m_area;
 };
 
+//! Represents a wall between between two ReactorBase objects.
+/*!
+ * Walls can move (changing the volume of the adjacent reactors) and allow heat
+ * transfer between reactors.
+ */
+class Wall : public WallBase
+{
+public:
+    Wall();
+
+    //! String indicating the wall model implemented. Usually
+    //! corresponds to the name of the derived class.
+    virtual std::string type() const {
+        return "Wall";
+    }
+
+    //! Set the wall velocity to a specified function of time, i.e. \f$ v(t) \f$.
+    void setVelocity(Func1* f=0) {
+        if (f) {
+            m_vf = f;
+        }
+    }
+
+    //! Rate of volume change (m^3/s) for the adjacent reactors.
+    /*!
+     * The volume rate of change is given by
+     * \f[
+     *     \dot V = K A (P_{left} - P_{right}) + F(t)
+     * \f]
+     * where *K* is the specified expansion rate coefficient, *A* is the wall
+     * area, and *F(t)* is a specified function of time. Positive values for
+     * `vdot` correspond to increases in the volume of reactor on left, and
+     * decreases in the volume of the reactor on the right.
+     */
+    virtual double vdot(double t);
+
+    //! Specify the heat flux function \f$ q_0(t) \f$.
+    void setHeatFlux(Func1* q) {
+        m_qf = q;
+    }
+
+    //! Heat flow rate through the wall (W).
+    /*!
+     * The heat flux is given by
+     * \f[
+     *     Q = h A (T_{left} - T_{right}) + A G(t)
+     * \f]
+     * where *h* is the heat transfer coefficient, *A* is the wall area, and
+     * *G(t)* is a specified function of time. Positive values denote a flux
+     * from left to right.
+     */
+    virtual double Q(double t);
+
+    void setThermalResistance(double Rth) {
+        m_rrth = 1.0/Rth;
+    }
+
+    //! Set the overall heat transfer coefficient [W/m^2/K].
+    void setHeatTransferCoeff(double U) {
+        m_rrth = U;
+    }
+
+    //! Get the overall heat transfer coefficient [W/m^2/K].
+    double getHeatTransferCoeff() const {
+        return m_rrth;
+    }
+
+    //! Set the emissivity.
+    void setEmissivity(double epsilon) {
+        if (epsilon > 1.0 || epsilon < 0.0) {
+            throw CanteraError("WallBase::setEmissivity",
+                               "emissivity must be between 0.0 and 1.0");
+        }
+        m_emiss = epsilon;
+    }
+
+    //! Get the emissivity.
+    double getEmissivity() const {
+        return m_emiss;
+    }
+
+    //! Set the expansion rate coefficient.
+    void setExpansionRateCoeff(double k) {
+        m_k = k;
+    }
+
+    //! Get the expansion rate coefficient
+    double getExpansionRateCoeff() const {
+        return m_k;
+    }
+
+protected:
+
+    //! expansion rate coefficient
+    double m_k;
+
+    //! heat transfer coefficient
+    double m_rrth;
+
+    //! emissivity
+    double m_emiss;
+
+    //! Velocity function
+    Func1* m_vf;
+
+    //! Heat flux function
+    Func1* m_qf;
+};
+  
 }
 
 #endif
--- a/include/cantera/zeroD/WallFactory.h
+++ b/include/cantera/zeroD/WallFactory.h
@@ -1,0 +1,72 @@
+//! @file WallFactory.h
+
+// This file is part of Cantera. See License.txt in the top-level directory or
+// at https://cantera.org/license.txt for license and copyright information.
+
+#ifndef WALL_FACTORY_H
+#define WALL_FACTORY_H
+
+#include "cantera/base/FactoryBase.h"
+#include "cantera/zeroD/Wall.h"
+
+namespace Cantera
+{
+
+class WallFactory : public Factory<WallBase>
+{
+public:
+    static WallFactory* factory() {
+        std::unique_lock<std::mutex> lock(wall_mutex);
+        if (!s_factory) {
+            s_factory = new WallFactory;
+        }
+        return s_factory;
+    }
+
+    virtual void deleteFactory() {
+        std::unique_lock<std::mutex> lock(wall_mutex);
+        delete s_factory;
+        s_factory = 0;
+    }
+
+    //! Create a new wall by type identifier.
+    /*!
+     * @param n the type to be created.
+     */
+    virtual WallBase* newWall(int n);
+
+    //! Create a new wall by type name.
+    /*!
+     * @param wallType the type to be created.
+     */
+    virtual WallBase* newWall(const std::string& wallType);
+
+    //! Register a new wall type identifier.
+    /*!
+     * @param name the name of the wall type.
+     * @param type the type identifier of the wall.
+     * Integer type identifiers are used by clib and matlab interfaces.
+     */
+    void reg_type(const std::string& name, const int type) {
+        m_types[type] = name;
+    }
+
+protected:
+    //! Map containing wall type identifier / wall type name pairs.
+    std::unordered_map<int, std::string> m_types;
+
+private:
+    static WallFactory* s_factory;
+    static std::mutex wall_mutex;
+    WallFactory();
+};
+
+//! Create a Wall object of the specified type
+inline WallBase* newWall(const std::string& model)
+{
+    return WallFactory::factory()->newWall(model);
+}
+
+}
+
+#endif
--- a/include/cantera/zerodim.h
+++ b/include/cantera/zerodim.h
@@ -1,13 +1,34 @@
+//! @file zerodim.h
+
+// This file is part of Cantera. See License.txt in the top-level directory or
+// at https://cantera.org/license.txt for license and copyright information.
+
 #ifndef CT_INCL_ZERODIM_H
 #define CT_INCL_ZERODIM_H
-#include "zeroD/Reactor.h"
+
+// reactor network
 #include "zeroD/ReactorNet.h"
-#include "zeroD/Reservoir.h"
-#include "zeroD/Wall.h"
-#include "zeroD/flowControllers.h"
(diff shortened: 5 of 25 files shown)


include/cantera/zeroD/DeviceConst.h
```diff
@@ -0,0 +1,37 @@
+//! @file DeviceConst.h
+
+// This file is part of Cantera. See License.txt in the top-level directory or
+// at http://www.cantera.org/license.txt for license and copyright information.
```
bryanwweber: Can you please use https://cantera.org/license.txt for all new files? Thank you :smile:

bryanwweber: And update the URL in any other files that you're changing, if it hasn't been done already.

ischoegl: I can certainly do that ... I guess I copied the block from files that hadn't been updated.

ischoegl: Done.

ischoegl:
At this point, the illustration example takes up many more lines than the actual modification that is required to enable a more flexible infrastructure (I just force-pushed a modification that implements an initial version to be used for RCM's). To facilitate review, there are two commits: one creating the infrastructure, and a second adding new classes.

ischoegl:
@bryanwweber per discussion from bryanwweber/CanSen#45, I added a Wall type that supports RCM data  (i.e. pr-omethe-us/PyKED compatible input): 
```python
cyl = ct.Cylinder(gas)
piston = ct.DrivenPiston(cyl, amb)
cyl.install(piston)
cyl.set_profile(vs, ts)
# piston.stroke_length = stroke
```
where `vs` and `ts` are numpy arrays for volume and time. Note: while I have not tested this, it would eliminate the dependency of PyKED on CanSen, see pr-omethe-us/PyKED#125

src/zeroD/Piston.cpp
```diff
@@ -0,0 +1,129 @@
...
+    } else {
+        return 0.;
+    }
+}
+
+double DrivenPiston::velocity(const double t)
+{
+    int siz = m_tprof.size();
+    if (siz>1 && t >= m_tprof[0] && t < m_tprof[siz-1]) {
+        if (t<m_tprof[m_ix]) m_ix = 0;
```
ischoegl: curly brackets to be consistent with cantera style (need to check throughout)

speth:
> while the current Func1 approach is very flexible, execution is considerably slower than native C++ code

Can you provide an example demonstrating this? While Python is certainly much slower than C++, the vast majority of the work in solving the reactor network consists of (a) evaluating the species production rates and (b) factorizing / solving linear systems in CVODES. I remember checking this when I introduced the Python callback function, and don't remember there being any significant performance impacts. It's possible something has crept in since then, though.

ischoegl:
> > while the current Func1 approach is very flexible, execution is considerably slower than native C++ code
> 
> Can you provide an example demonstrating this? While Python is certainly much slower than C++, the vast majority of the work in solving the reactor network consists of (a) evaluating the species production rates and (b) factorizing / solving linear systems in CVODES. I remember checking this when I introduced the Python callback function, and don't remember there being any significant performance impacts. It's possible something has crept in since then, though.

I can look into this next week in more detail. For right now, the ic_engine.py example is slowed down more than necessary due to the way the valves are handled (time step is overly small and the majority of the time is spent saving data). With my code in the other PR I believe a speed up of x10 is possible, but not necessarily only because of Func1. The modified example of the PR runs around 4x faster as the valve times are handled not within the advance loop but are updated at every step. The main advantage over an imposed python implementation is really more readable code.

However, I would like to point out that removing/bypassing Func1 is _not_ the primary objective. I implemented this PR as I was unable to overload wall objects for a very different research application within the available infrastructure.

Overall, I believe that Func1 is powerful, but not easy for someone new to cantera. I picked this example as a demo as I thought that having a simple ic_engine example would benefit the user base.

bryanwweber:
@ischoegl I also agree with @speth, seeing an example where Python significantly slows things down would be helpful.

@speth @decaluwe This is the PR that prompted the email that @ischoegl sent to us. I'm not sure where we want to draw the line between "core" Cantera and "applications" of Cantera, and what should/shouldn't be in the main code base. That's something we've talked about in rough terms, so I asked him to send an email to prompt some more discussion of that.

speth:
As you note, the valves in ic_engine.py are handled outside the ODE integrator, and I expect that this is the reason it's slow. What it's doing is inconsistent with the way the multistep CVODES integrator works, causing it to take unnecessarily small time steps. I don't think the speed has anything to do with the piston velocity function.

While I support the idea of providing an easy-to-use interface for some of these common applications, this ends up being a lot of code and a lot of new classes to implement something that I think is equivalent to a transformation from a domain-specific set of input parameters to the generic formulation of the
reactor/wall/valve classes.

I'm also having some trouble seeing how adopting the factory pattern actually increases flexibility here, at least at present. One of the main things that the derived classes that are being introduced here do is add new member functions for interacting with them, which can't be used generically, e.g. if you only
have a pointer to the base class. In cases like this, where you just end up re-casting to the derived type that you know has the right function, you may as well just create an object of the derived type directly.

ischoegl:
@speth and @bryanwweber ... I do not think that the discussion is going in the right direction: as mentioned, `Func1` was not my main priority. Thus, I dropped the second commit implementing the `ic_engine` altogether.

> I'm also having some trouble seeing how adopting the factory pattern actually increases flexibility here, at least at present. One of the main things that the derived classes that are being introduced here do is add new member functions for interacting with them, which can't be used generically, e.g. if you only
have a pointer to the base class. In cases like this, where you just end up re-casting to the derived type that you know has the right function, you may as well just create an object of the derived type directly.

The issue this PR is addressing explicitly is that while it is possible to create derived classes for `Wall` and `FlowDevice` objects, it is impossible to register them to be used within the context of `reactor.pyx` for the cython interface or `ctreactor.h` for C++. My research group's approach is to develop code that is compiled against stock cantera, and faces substantial restrictions due to this matter. I'd appreciate if this use case could be considered.

To illustrate: a bare bones example for how compilation against cantera looks like is shown in [ischoegl/ctapp](https://github.com/ischoegl/ctapp), where I create a clone of a `Reactor` class, and register it to be used with the existing cantera integrator. In the current infrastructure, the same cannot be done for `Wall` and `FlowDevice` objects, as registration is hard coded. This PR eliminates this restriction.

ischoegl:
PS: to be explicit ... the key line for the `ctapp` example is in [`ctapp/ClonedReactor.h`](https://github.com/ischoegl/ctapp/blob/master/ctapp/ClonedReactor.h), where a new reactor object is registered using 
```C++
inline void registerClonedReactor() {
  ReactorFactory::factory()->reg("ClonedReactor",
                                 []() { return new ClonedReactor(); });
}
```
Using the factory pattern, the cython interface [clonedreactor.pyx](https://github.com/ischoegl/ctapp/blob/master/ctapp/clonedreactor.pyx) generates fully functional code when compiled against stock cantera:
```python
cdef __register():
    registerClonedReactor()

__register()

cdef class ClonedReactor(ct.IdealGasConstPressureReactor):
    reactor_type = "ClonedReactor"
```
Again, the goal is to create this capability for all `zerod` classes.

include/cantera/zeroD/ReactorClasses.h
```diff
@@ -0,0 +1,16 @@
...
+
+#ifndef CT_REACTOR_CLASSES_H
+#define CT_REACTOR_CLASSES_H
+
+#include "cantera/zeroD/Reservoir.h"
+#include "cantera/zeroD/Reactor.h"
+#include "cantera/zeroD/FlowReactor.h"
+#include "cantera/zeroD/ConstPressureReactor.h"
+#include "cantera/zeroD/IdealGasReactor.h"
+#include "cantera/zeroD/IdealGasConstPressureReactor.h"
```
speth: I don't really see the need for this file -- I would rather leave these includes in `zerodim.h` directly.

include/cantera/zeroD/DeviceConst.h
```diff
@@ -0,0 +1,33 @@
...
+// flow device definitions
+const int MFCType = 1;
+const int PressureControllerType = 2;
+const int ValveType = 3;
+
+// wall type definitions
+const int WallType = 1;
+
+// other constants
+const double pi = 3.141592653589793;
```
speth: Already defined as `Pi` in `ct_defs.h`.

include/cantera/zeroD/DeviceConst.h
```diff
@@ -0,0 +1,33 @@
...
+const int ReactorType = 2;
+const int FlowReactorType = 3;
+const int ConstPressureReactorType = 4;
+const int IdealGasReactorType = 5;
+const int IdealGasConstPressureReactorType = 6;
+
+// flow device definitions
+const int MFCType = 1;
+const int PressureControllerType = 2;
+const int ValveType = 3;
```
speth: I am trying very hard to get rid of the use of magic numbers as type constants throughout Cantera. Rather than trying to organize them better, I would prefer to just leave them where they are, and touch them as little as possible until they can be deprecated and removed.

include/cantera/zeroD/Wall.h
```diff
@@ -1,35 +1,40 @@
...
-//! Represents a wall between between two ReactorBase objects.
-/*!
- * Walls can move (changing the volume of the adjacent reactors) and allow heat
- * transfer between reactors.
+/**
+ * Base class for 'walls' (walls, pistons, etc.) connecting reactors.
+ * @ingroup reactor0
  */
-class Wall
+class WallBase
```
speth, line 26: I think some more clarity is needed on the separation between `WallBase` and `Wall`. I recognize that introducing a base class with fewer "features" may make adding new derived classes simpler, but I don't fully follow the logic of what is included in `WallBase` versus `Wall`. Can you please elaborate?

include/cantera/zeroD/Wall.h
```diff
@@ -1,35 +1,40 @@
...
-    Wall(const Wall&) = delete;
-    Wall& operator=(const Wall&) = delete;
+    virtual ~WallBase() {}
+    WallBase(const WallBase&) = delete;
+    WallBase& operator=(const WallBase&) = delete;
+
+    //! Return an integer indicating the type of wall
+    int type() {
+        return m_type;
+    }
```
speth, line 39: I suggest a function with the signature `virtual std::string type() const`, like in class `ThermoPhase`.

include/cantera/zeroD/WallClasses.h
```diff
@@ -0,0 +1,11 @@
...
+
+// This file is part of Cantera. See License.txt in the top-level directory or
+// at https://cantera.org/license.txt for license and copyright information.
+
+#ifndef CT_WALL_CLASSES_H
+#define CT_WALL_CLASSES_H
+
+#include "cantera/zeroD/Wall.h"
+
+#endif
```
speth: Again, I would prefer to just leave wall implementations as a includes directly in `zerodim.h`.

include/cantera/zeroD/WallFactory.h
```diff
@@ -0,0 +1,61 @@
...
+        delete s_factory;
+        s_factory = 0;
+    }
+
+    /**
+     * Create a new Wall.
+     * @param n the type to be created.
+     */
+    virtual WallBase* newWall(int n);
+    virtual WallBase* newWall(const std::string& wallType);
```
speth, line 42: I would prefer to have only the string-based constructor.

interfaces/cython/cantera/reactor.pyx
```diff
@@ -676,6 +686,16 @@ cdef class FlowDevice:
...
+    # property upstream:
+    #     """Upstream reactor"""
+    #     def __get__(self):
+    #         return self._upstream
+
+    # property downstream:
+    #     """Downstream reactor"""
+    #     def __get__(self):
+    #         return self._downstream
+
```
speth: Remove commented-out code.

src/zeroD/FlowDeviceFactory.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+FlowDeviceFactory* FlowDeviceFactory::s_factory = 0;
+std::mutex FlowDeviceFactory::flowDevice_mutex;
+
+FlowDeviceFactory::FlowDeviceFactory()
+{
+    reg("MassFlowController", []() { return new MassFlowController(); });
+    reg("PressureController", []() { return new PressureController(); });
+    reg("Valve", []() { return new Valve(); });
+    // reg("FlowDevice", []() { return new FlowDevice(); }); // unused?
```
speth: I don't think the base class can be used directly, can it?

src/zeroD/Wall.cpp
```diff
@@ -1,24 +1,23 @@
...
+// at https://cantera.org/license.txt for license and copyright information.
 
-#include "cantera/zeroD/Wall.h"
-#include "cantera/zeroD/ReactorNet.h"
+#include "cantera/base/stringUtils.h"
 #include "cantera/numerics/Func1.h"
+#include "cantera/zeroD/Wall.h"
 #include "cantera/thermo/SurfPhase.h"
-#include "cantera/base/stringUtils.h"
+//#include "cantera/zeroD/ReactorNet.h"
```
speth: Remove commented-out code.

speth (changes requested):
Sorry, the implementation of the `ic_engine` example was indeed distracting, as the majority of the code added was related to that rather than the factory functions. I've provided some comments on the implementation of the factories for these classes.

In addition, your `ctapp` example answers the question I had of why implementing the extensible factory for these classes is useful -- the key is less that you can register a user-defined class from a C++ function, since you can just as easily instantiate the user-defined class directly from C++ as well, but that it makes it easy to wrap these classes in Python.

This `ctapp` example is very good demonstration of extensibility that I had hoped was possible but had never had a chance to try out. I am wondering if this method currently only works on Linux, where all of the symbols in the Cantera Python extension module (which includes the full Cantera C++ library) are exposed, versus Windows and macOS which I believe only export symbols that are explicitly declared. But in any case, it's a good starting point.

include/cantera/zeroD/ReactorClasses.h
ischoegl: the complete list is used in `zerodim.h` and also in `ctreactor.cpp`. I tried to limit having to add additional files to more than one file. I don't have a strong opinion on this.

include/cantera/zeroD/DeviceConst.h
ischoegl: This was specific to ic_engine. Will be removed.

include/cantera/zeroD/DeviceConst.h
ischoegl: Understood. Magic numbers are still essential to the implementation for `ctreactor.cpp`: the reason I solved it that way was that I used `Reactor` as a template. Would you prefer to work ahead for the other factories (which would create a non-uniform approach)?

include/cantera/zeroD/Wall.h
ischoegl: The distinction is arbitrary, but I wanted to mimic `ReactorBase` vs `Reactor` in the `WallBase`/`Wall` area. From my perspective, the `Func1` is a specialization for a multi-purpose object communicating with Python, which may not be necessary for another case. There are many other ways to differentiate. Suggestions are welcome.

include/cantera/zeroD/Wall.h
ischoegl: as mentioned above, I used `Reactor` (and also `FlowDevice`) classes as a template and didn't want to go a different direction. ~It's easy to change if so desired.~

include/cantera/zeroD/WallClasses.h
ischoegl: These `.h` files made more sense in the context of the `ic_engine` implementation. I don't have a strong opinion.

include/cantera/zeroD/WallFactory.h
ischoegl: At this point, `ctreactor.cpp` still depends on integer-based constructors; I kept this approach to mimic implementations of `Reactor`-type objects.

interfaces/cython/cantera/reactor.pyx
ischoegl: ok. I did miss that one.

src/zeroD/FlowDeviceFactory.cpp
ischoegl: It did have a constructor call in `ctreactor.cpp`, which was puzzling (hence my comment)

src/zeroD/Wall.cpp
ischoegl: Another one I overlooked.

ischoegl:
@speth ... thanks for the thorough review - I truly appreciate it.

At this point I'd ask you to give some directions on whether you want to do without integer-based constructors for `WallBase` (making the approach distinct from `ReactorBase` and `FlowDevice` - specifically considering `ctreactor.cpp`'s implementation). Once I know I'll address all your requests. I can also give deprecation a shot if you prefer (although it may take me a couple of days as I have to take care of some other matter first).

> This ctapp example is very good demonstration of extensibility that I had hoped was possible but had never had a chance to try out. I am wondering if this method currently only works on Linux, where all of the symbols in the Cantera Python extension module (which includes the full Cantera C++ library) are exposed, versus Windows and macOS which I believe only export symbols that are explicitly declared. But in any case, it's a good starting point.

I am quite happy you liked the `ctapp` example. I haven't tested any of this on a non-linux machine (to be exact: outside of a linux-docker image), but can ask my students. I'll definitely let you know.

ischoegl:
@speth ... I looked into eliminating magic numbers for object constructors (for the moment, only for walls): I am afraid that they are an integral part of the CLIB interface. I.e. object creation in `ctreactor.h/.cpp` would have to be redefined as `int wall_new(std::string type)`, which breaks backward compatibility and has repercussions all the way back to the matlab interface. This requires heavier lifting that I believe is best handled by a separate PR.

include/cantera/zeroD/Wall.h
ischoegl: At close inspection, this involves major changes in CLIB - which builds on the constructor  `int wall_new(int type)`, - potentially breaking backwards compatibility. I propose to defer elimination/deprecation of magic numbers in a separate effort.

include/cantera/zeroD/DeviceConst.h
ischoegl: At close inspection, this involves major changes in CLIB - which builds on the constructor  `int wall_new(int type)`, - potentially breaking backwards compatibility. I propose to defer elimination/deprecation of magic numbers in a separate effort.

include/cantera/zeroD/ReactorClasses.h
ischoegl: Now using `zerodim.h` more extensively (in both `ctreactor.cpp` and `_cantera.pxd`)

include/cantera/zeroD/WallClasses.h
ischoegl: removed.

include/cantera/zeroD/DeviceConst.h
ischoegl: the entire file is removed.

include/cantera/zeroD/WallFactory.h
ischoegl: As mentioned above, this involves major changes in CLIB - which builds on the constructor `int wall_new(int type)`, - potentially breaking backwards compatibility. I propose to defer elimination/deprecation of magic numbers in a separate effort.

include/cantera/zeroD/Wall.h
ischoegl: To elaborate further: anyone interested in changing the behavior of wall objects will likely overload heat transfer (`WallBase::qdot`) and/or wall velocity (`WallBase::vdot`), which is what motivated the proposed separation.

ischoegl:
@speth ... I resolved all but three requests, two of which would require extensive changes in the CLIB interface (to address the same issue for current `ReactorBase` and `FlowDevice` implementations). I opened issue #638 that allows this to be addressed by a separate PR.

include/cantera/zeroD/FlowDevice.h
```diff
@@ -15,9 +15,9 @@ namespace Cantera
 class Func1;
 class ReactorBase;
 
-const int MFC_Type = 1;
-const int PressureController_Type = 2;
-const int Valve_Type = 3;
+const int MFCType = 1;
+const int PressureControllerType = 2;
+const int ValveType = 3;
```
speth: Given the eventual goal of deprecating and removing these constants, I don't think they should be renamed. If nothing else, renaming them now would break compatibility with any code that was using these names without warning.

include/cantera/zeroD/Wall.h
speth: Except that `wall_new` currently doesn't use the argument passed in since there is only one type of wall. So, no one can be using that argument to actually do anything.

I agree that figuring out how to transition away from the existing magic numbers is best left to another PR, but we can make the task at least a little simpler if we avoid putting things in our own way. So, while I see the temporary need for being able to assign an integer constant to a custom `Wall` class via the `reg_type` function, I'd rather we use the name `type` here for the string constant, and something else for the integer constant (if that's even really necessary).

include/cantera/zeroD/Wall.h
speth: Right, making `vdot` and `qdot` virtual certainly makes sense. I don't really see the use of `Func1` as being applicable only to Python -- it's meant to be used any time you want to set some parameter as a function of time. That interface could probably be revised to make use of C++11 lambdas rather than the custom `Func1` class, but the concept would remain the same.

I guess you could move all the functions related to heat transfer and motion away from the base wall class. That doesn't leave much left in `WallBase`, but it would mean that you'd have a pretty clean and clear interface for any new derived classes to be built on.

include/cantera/zerodim.h
```diff
@@ -1,13 +1,29 @@
...
+#include "cantera/zeroD/IdealGasConstPressureReactor.h"
+
+// flow devices
+#include "cantera/zeroD/flowControllers.h"
+
+// walls
+#include "cantera/zeroD/Wall.h"
+
+// surface
+#include "cantera/zeroD/ReactorSurface.h"
```
speth, line 27: I think the factory class header files should be included here as well, which would be consistent with `thermo.h` and `kinetics.h`, as well as saving you a few other `#include`s elsewhere.

speth (changes requested):
Thanks for the revisions, I think this is pretty close now. I just had a few more suggestions which I don't think will pose any problems.

include/cantera/zeroD/FlowDevice.h
ischoegl: understood

include/cantera/zerodim.h
ischoegl: done.

include/cantera/zeroD/Wall.h
ischoegl: The way I implemented things, instantiation of objects via `WallFactory` is an exact replica of objects created via `ReactorFactory` calls (note that in `ctreactor.cpp` I changed calls `wall_new` and `flowdev_new` to work exactly like `reactor_new`, i.e. the argument is actually used). 

I'd much prefer to use the _old_ coding style for the time being and change everything in a separate PR. In other words, since there are precedents for `int FlowDevice::type()` and `int ReactorBase::type()` signatures, I am hesitant to introduce a `std::string WallBase::type()` at this point as it is not consistent.

include/cantera/zeroD/Wall.h
ischoegl: Good suggestion. It is now implemented.

ischoegl:
@speth ... thanks for the re-review of the PR. 

I implemented all requests but one: since there are precedents for `int FlowDevice::type()` and `int ReactorBase::type()` signatures, I am hesitant to introduce a `std::string WallBase::type()` at this point as it is not consistent. This is imho a trivial change that can be easily implemented when magic numbers are deprecated. Instead, I would suggest to maintain structure and interface of `WallFactory` and `WallBase` in exact analogy to `ReactorFactory` and `ReactorBase`.

speth:
Thanks for the updates. Replacing `int WallBase::type()` to `std::string WallBase::type()` in a later release would actually be rather involved. Since functions can't be overloaded based on return type, in order to allow incremental updates without breaking backwards compatibility, it would take three releases to effect this change:
* one to deprecate the `int`-returning function and add a `string`-returning function with a different name (say `WallBase::wallType()`)
* one to remove the `int`-returning function, add a `string`-returning version of `WallBase::type()`, and deprecate `WallBase::wallType()`
* one to remove `WallBase::wallType()`.

This is already what we will be stuck with to do this for the other reactor-related classes, but I thought it would be nice to avoid doing so here. And while the pattern that you've followed is consistent with these other classes, Cantera also uses the style of `string type()` elsewhere, e.g. in class `ThermoPhase`.

include/cantera/zeroD/Wall.h
ischoegl: OK, the clarification makes sense. Since this is currently not used by any call, changing signature is trivial.

ischoegl:
Thanks for the clarification, being inconsistent makes sense in this context: based on your comments, I changed the return type to `std::string`, so all requests are resolved.

PS: I noticed two calls for `WallBase::area()` and `WallBase::getArea()` that appear to be redundant ... 

~___I am not suggesting to touch this in this PR___~, but am curious about the appropriate way to deprecate the former call (which is not consistent with the typical set/get approach you mentioned in another PR). While I haven't found `warn_deprecated` being used much, would it be 
```
warn_deprecated("WallBase::area()", 
                "To be removed after Cantera 2.5. Replace with WallBase::getArea().");
```
? ~(which I haven't tried)~ (which works)

ischoegl:
@speth ... while I haven’t changed anything that should affect coverage, codecov is now no longer passing. Apart from that I believe we have converged.

PS: due to the coverage issue, I added a `type` attribute for `WallBase` objects to the cython interface, as well as a trivial unit test. I also went ahead and replaced `area` with `getArea` and added a deprecation warning for `WallBase::area()`.

speth:
Good catch on the redundancy of `area()` and `getArea()`. However, I would suggest keeping `area()`. The guideline that I mentioned before on this (which I hope I got the right way around) is that we tend to have the bare property name as the accessor when you can just return the value, e.g. `double density()`, but use the `get` prefix when the access is done by passing in a pointer to an object (array) that is then modified, e.g. `void getPartialMolarEnthalpies(double*)`.

And yes, that would be the way to deprecate the function. You should also add a `@deprecated To be removed...` message to the docstring so it shows up in the list of deprecated functions in the Doxygen docs.

And don't worry too much about the coverage report. Passing those tests is not a requirement for merging a PR.

ischoegl:
> Good catch on the redundancy of `area()` and `getArea()`. However, I would suggest keeping `area()`. The guideline that I mentioned before on this (which I hope I got the right way around) is that we tend to have the bare property name as the accessor when you can just return the value, e.g. `double density()`, but use the `get` prefix when the access is done by passing in a pointer to an object (array) that is then modified, e.g. `void getPartialMolarEnthalpies(double*)`.
> 
> And yes, that would be the way to deprecate the function. You should also add a `@deprecated To be removed...` message to the docstring so it shows up in the list of deprecated functions in the Doxygen docs.
> 
> And don't worry too much about the coverage report. Passing those tests is not a requirement for merging a PR.

Alright, no problem. I will revert to `area` ... should be done shortly.

Regarding coverage, I noticed that having inline constructors vs constructors defined in cpp files does make a difference.

ischoegl:
@speth ... done.

ischoegl:
@speth ... is there anything else that should be taken care of? (I believe I have resolved all mentioned issues). Thanks in advance!

include/cantera/zeroD/FlowDeviceFactory.h
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Create a new FlowDevice.
+     * @param n the type to be created.
+     */
+    virtual FlowDevice* newFlowDevice(int n);
+    virtual FlowDevice* newFlowDevice(const std::string& flowDeviceType);
+
+    //! Register a new object type
+    void reg_type(const std::string& name, const int type) {
```
speth, line 50: This function could use a bit more documentation, to explain that to register a new type, you must call the `reg` method and assign a string identifier, and that calling this method is then necessary only if you need to assign an integer identifier for creating this type from clib or Matlab.

include/cantera/zeroD/FlowDeviceFactory.h
```diff
@@ -0,0 +1,61 @@
...
+        m_types[type] = name;
+    }
+
+protected:
+    std::unordered_map<int, std::string> m_types;
```
speth, line 56: Please add a docstring for this member variable.

speth (commented):
Thanks, the implementation all looks fine to me. I have just one last request for improvements to the documentation.

include/cantera/zeroD/FlowDeviceFactory.h
ischoegl: done.

include/cantera/zeroD/FlowDeviceFactory.h
ischoegl: done.

ischoegl:
Done. Let me know if the two overloaded `ReactorFactory::newReactor`/`WallFactory::newWall`/`FlowDeviceFactory::newFlowDevice` calls should not have separated docstrings (I split them in my revision)

speth (approved):

PR #637, opened 2019-05-06
Improved lithium-ion battery cti file and Matlab example

wbessler:
Extended and clarified comments, corrected density data, improved functionality

- the lithium_ion_battery.cti and .m files had minor errors (wrong density of the materials, missing T AND P setting inialization of the interfaces, unrealistic stoichiometry ranges) 
- comments were extended and improved for more clarity, and entries in cti file were re-ordered.

### DIFF
--- a/samples/matlab/lithium_ion_battery.m
+++ b/samples/matlab/lithium_ion_battery.m
@@ -1,4 +1,4 @@
-% This example file calculates the cell voltage of a lithium-ion % battery
+% This example file calculates the cell voltage of a lithium-ion battery
 % at given temperature, pressure, current, and range of state of charge (SOC).
 %
 % The thermodynamics are based on a graphite anode and a LiCoO2 cathode,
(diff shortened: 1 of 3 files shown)


samples/matlab/lithium_ion_battery.m
```diff
@@ -32,18 +26,16 @@
...
-R_elyt = 0;
-T = 300;
-P = oneatm;
-
-global F
-% Parameters
+% Input parameters
+SOC = 0:0.02:1; % [-] Input state of charge (0...1)
+X_Li_an = (0.75-0.01)*SOC+0.01; % anode balancing
+X_Li_ca = (0.99-0.49)*(1-SOC)+0.49; % cathode balancing
```
decaluwe: Tiniest little quibble here, @wbessler - might it be clearer for the user if this balancing were made more explicit?

- Would it make sense to specify the upper and lower SOC limits as variables?  I'm not sure if a user would ever want to change them, so there's a risk there, but then the user would understand what 0.01, 0.75, et al. represent
- Or would it be better to just add a comment, explaining what we mean by balancing?
- Lastly, the comment you have removed above, which commented on the anode and cathode capacities being perfectly balanced, has some relevant info here, as well.  Just thinking of ways to make the example a little more transparent to somebody not well versed in battery modeling.

decaluwe:
Thanks, @wbessler - I think these changes are all pretty clearly improvements!  I have one minor question on the matlab example (see earlier comment), but other than that @speth I think this is good to go.

decaluwe:
Thanks, @wbessler - these changes all look good to me.  I caught one small typo in the matlab example, which I corrected, and also converted the hand-coded Faraday constant to instead use the function now included in the matlab toolbox.  There was something weird with a rebase I did on my local machine, so I had to do a force push to get this back to the correct state.

@speth and @bryanwweber  - any comments?  I think this should be ready to go.

bryanwweber:
It seems like there's a problem with the lithium_ion_battery.yaml input file, the elements key cannot contain an empty list it seems like. I'm not sure why only the macOS build is passing. @decaluwe I think the tests should pass before we merge this.

speth:
I think the YAML conversion issue should be fixed. I also removed some end-of-line whitespace from the CTI and Matlab files. Otherwise, this looks OK to me.

PR #640, opened 2019-06-07
Adding faradayconstant.m to Matlab toolbox

decaluwe:
- Adds a function `faradayconstant.m` to the Matlab toolbox which, akin to `gasconstant.m`, returns the value of the Faraday constant, in Coulombs per kmol of electron.

### DIFF
--- a/interfaces/matlab/toolbox/faradayconstant.m
+++ b/interfaces/matlab/toolbox/faradayconstant.m
@@ -1,0 +1,8 @@
+function r = faradayconstant
+% FARADAYCONSTANT  Get the Faraday constant in C/kmol of electron.
+% r = faradayconstant
+% :return:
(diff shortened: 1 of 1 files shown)


bryanwweber:
Can you get this number from the built-in constants? Would it be possible to create a Matlab class that has all of the constants that Cantera defines?

decaluwe:
C’mon, man, you’re ruining my “easiest PR ever!” 😂

bryanwweber:
If the answer to either of those is "its complicated", I'm fine with this as-is. But if its easy... why not, you know?

speth:
This is consistent with how other constants are currently defined in the Matlab toolbox, so I think this is fine as-is. I'd save anything more complicated for the long overdue rewrite of the Matlab toolbox.

PR #643, opened 2019-06-11
Deprecate PureFluid.h

bryanwweber:
- Deprecate `PureFluid.h`, the convenience wrapper around `PureFluidPhase`. It can be replaced with `PureFluidPhase::initThermoFile()` or other `ThermoPhase` creation methods.
- Update the `rankine.cpp` sample, which was the only code in the repository that used `PureFluid.h`

This is based on the YAML branch from @speth's fork. It should not be merged until after the YAML PR is finished, due to the update in the `rankine.cpp` to use the YAML input file.


bryanwweber:
Need to add `warn_deprecated()` for deprecation in the class constructors.

speth:
Can you rebase (or maybe `cherry-pick` would be easier) this onto the current master?

bryanwweber:
So tests are failing on macOS but its not failing the job: https://travis-ci.org/Cantera/cantera/jobs/551315672#L3474 Ugh

speth:
I think the build status is set by the exit code of the last command. Now that `scons samples` has been added to the script, `scons test` isn't the last command. For some reason, we had `set -e`, which causes a failing command to terminate the batch job, set only for the Linux build and not the macOS build. Let's see if this fixes it.

bryanwweber:
Thanks @speth that seems to have fixed it. I'll fix up the tests here and push again.

include/cantera/PureFluid.h
```diff
@@ -6,17 +6,25 @@
...
+/*!
+ * @deprecated To be removed after Cantera 2.5.0. Replaceable with PureFluidPhase::initThermoFile
+ */
 class PureFluid : public PureFluidPhase
 {
 public:
-    PureFluid() : m_ok(false), m_r(0) {}
+    PureFluid() : m_ok(false), m_r(0) {
+        warn_deprecated("class PureFluid", "To be removed after Cantera 2.5.0. "
+        "Replaceable with PureFluidPhase::initThermoFile.");
```
speth: The default constructor is replaceable with just `PureFluidPhase`. The constructor from input file name is the one that is replaceable by using the default constructor then calling `PureFluidPhase::initThermoFile`.

Also, I would indent the the continuation line by an additional level.

test_problems/pureFluidTest/testPureWater.cpp
```diff
@@ -24,30 +24,28 @@ int main()
     double pres;
     try {
 
-        XML_Node* xc = get_XML_File("liquidvapor.xml");
-        XML_Node* const xs = xc->findNameID("phase", "water");
-        ThermoPhase* water_tp = newPhase(*xs);
-        PureFluidPhase* w = dynamic_cast <PureFluidPhase*>(water_tp);
+        PureFluidPhase w;
+        w.initThermoFile(std::string("liquidvapor.yaml"), std::string("water"));
```
speth: I don't think the explicit call to the `std::string` constructor is necessary, e.g. you should be able to write just `w.initThermoFile("liquidvapor.yaml", "water");`

test_problems/rankine_democxx/rankine.cpp
```diff
@@ -82,7 +77,7 @@ int main()
     try {
         return openRankine(0, 0);
     } catch (CanteraError& err) {
-        std::cout << err.what() << std::endl;
+        printf("%s", err.what());
```
speth: Why is this preferable to using `std::cout`?

test_problems/rankine_democxx/rankine.cpp
```diff
@@ -26,21 +20,22 @@ void saveState(F& fluid, string name)
...
-    double etap = 0.6;     // pump isentropic efficiency
-    double etat = 0.8;     // turbine isentropic efficiency
-    double phigh = 8.0e5;  // high pressure
+    double etap = 0.6; // pump isentropic efficiency
+    double etat = 0.8; // turbine isentropic efficiency
+    double phigh = 8.0e5; // high pressure
 
-    Water w;
+    PureFluidPhase w;
+    w.initThermoFile(std::string("liquidvapor.yaml"), std::string("water"));
```
speth: Likewise here with `std::string`

test_problems/rankine_democxx/rankine.cpp
```diff
@@ -70,7 +65,7 @@ int openRankine(int np, void* p)
     double heat_in = h["3"] - h["2"];
     double efficiency = work/heat_in;
 
-    cout << "efficiency = " << efficiency << endl;
+    printf("efficiency = %8.6g", efficiency);
```
speth: If you don't want to use `cout`, I'd recommend Cantera's `writelog` function over `printf`.

test_problems/rankine_democxx/rankine.cpp
```diff
@@ -26,21 +20,22 @@ void saveState(F& fluid, string name)
 
 void printStates()
 {
-    size_t nStates = states.size();
-    for (size_t n = 0; n < nStates; n++) {
-        string name = states[n];
-        printf(" %5s %10.6g %10.6g  %12.6g %12.6g %5.2g \n",
+    int nStates = states.size();
+    for (int n = 0; n < nStates; n++) {
```
speth, line 24: What's wrong with `size_t`?

test_problems/rankine_democxx/rankine.cpp
```diff
@@ -26,21 +20,22 @@ void saveState(F& fluid, string name)
...
+        std::string name = states[n];
+        printf(" %5s %10.6g %10.6g %12.6g %12.6g %5.2g \n",
```
speth: How does this not leave a trailing space on each line in the new blessed output file?

speth (changes requested):
The miscellaneous changes to `test_problems/rankine_democxx/rankine.cpp` should at least be noted in the commit message, if not split into a separate commit.

test_problems/rankine_democxx/rankine.cpp
bryanwweber: In this change and changing `size_t` to `int`, I was making the test code to be the same as the sample. I'm not sure why the code is nearly duplicated but I didn't want to modify SConscruct to copy the files around, so I made them as close as possible 🤷‍♂

test_problems/rankine_democxx/rankine.cpp
bryanwweber: Probably because I opened the blessed output file in the text editor, which automatically strips end of line whitespace. I changed the function so it doesn't include the last space to match the blessed output.

test_problems/rankine_democxx/rankine.cpp
bryanwweber: See the other comment

test_problems/rankine_democxx/rankine.cpp
```diff
@@ -26,21 +20,22 @@ void saveState(F& fluid, string name)
...
+    int nStates = states.size();
+    for (int n = 0; n < nStates; n++) {
+        std::string name = states[n];
+        writelog(" {:5s} {:10.6g} {:10.6g} {:12.6g} {:12.6g} {:5.2g}\n",
+                 name, T[name], P[name], h[name], s[name], x[name]);
     }
 }
 
-int openRankine(int np, void* p)
+int openRankine()
```
bryanwweber, line 31: Here and in the other Rankine file, I wasn't sure what these arguments were for, and they don't appear to be used, so I deleted them

speth:
Sorry, I wasn't trying to send you down a rabbit hole replacing uses of `printf` with `writelog`. I have a general preference to `writelog` (which uses `fmt`) over `cout` over `printf`, but refactoring the `test_problems` tests to change this alone isn't worth it, given the myriad other issues with those tests.

In any case, I don't think `writelog` works with the `%` style format strings -- it uses the Python-style ones implemented by `fmt`.

speth:
Also, the almost-but-not-quite duplication of the `rankine.cpp` sample in the `test_problems` test suite is a wart in the test suite that I'm not really expecting to be fixed here.

speth (approved):

test_problems/pureFluidTest/testPureWater.cpp
```diff
@@ -24,32 +24,29 @@ int main()
     double pres;
     try {
 
-        XML_Node* xc = get_XML_File("liquidvapor.xml");
-        XML_Node* const xs = xc->findNameID("phase", "water");
-        ThermoPhase* water_tp = newPhase(*xs);
-        PureFluidPhase* w = dynamic_cast <PureFluidPhase*>(water_tp);
+        unique_ptr<ThermoPhase> w(newPhase("liquidvapor.yaml", "water"));
```
bryanwweber, line 27: @speth For my own education, why is this better than using `PureFluidPhase w`?

speth: It's more similar to the method of instantiating the phase that was being used before, in that it allocates the object on the heap and you just have a pointer to it. I prefer it in this case because (a) it means you don't have to change all the pointer dereferencing operations (`w->foo`) to member access (`w.foo`), and (b) there is no point in this test where the fact that this  is a `PureFluidPhase` object is actually used -- all of the member functions called are ones that are defined in the base class, so it's fine if all we have is a pointer to a `ThermoPhase`.

bryanwweber: Awesome, thanks for the explanation!

PR #645, opened 2019-06-12
Deprecating ConstDensityThermo class

decaluwe:
As previously discussed, the `incompressible_solid` cti phase actually points to the `ConstDensityThermo` class.  This is misleading to a user, and moreover, a thermo class which enforces constant mass density has both dubious physical underpinning and marginal utility for a user.  It is therefore appropriate to deprecate and remove this class, and refer interested users to either the `IdealSolidSolnPhase` or `Lattice` thermo classes, which are appropriate replacements.

- Marks `ConstDensityThermo` thermo class for deprecation, as of release 2.5, and for removal, thereafter.
- Adds appropriate messages to `ctml_writer.py` parsing routine, `ConstDensityThermo.cpp`, and `ConstDensityThermo.h` to indicate as much, and suggest use of `Lattice` or `IdealSolidSolnPhase` classes as alternates
- Removes the `incompressible_solid` phase in `sofc.cti` (which is used for the sofc example, and which aliases to the `ConstDensityThermo` class) and replaces it with an equivalent `Lattice` phase.
- Changes the cython test in `test_convert.py` which interrogates the `incompressible_solid` phase mass density in `sofc.cti` so that it instead interrogates the `Lattice` phase molar density.

### DIFF
--- a/interfaces/cython/cantera/ctml_writer.py
+++ b/interfaces/cython/cantera/ctml_writer.py
@@ -2137,7 +2137,16 @@ class semiconductor(phase):
 
 
 class incompressible_solid(phase):
-    """An incompressible solid."""
+    """An incompressible solid.
+
+    .. deprecated:: 2.5
+
+        To be deprecated with version 2.5, and removed thereafter.
+        This phase pointed to an ill-considered constant_density ThermoPhase
+        model, which assumed a constant mass density. This underlying phase is
+        simultaneously being deprecated. Please consider switching to
+        either `IdealSolidSolution` or `lattice` phase, instead.
+    """
     def __init__(self,
                  name = '',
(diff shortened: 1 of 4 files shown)


interfaces/cython/cantera/ctml_writer.py
```diff
@@ -2295,15 +2304,14 @@ def __init__(self,
                         initial_state, options)
         self._tr = transport
         self._n = site_density
-        self._species = species
+
```
speth, line 2307: As minor as it is, I'd suggest that this fix for `lattice` be done in a separate commit, since it serves a distinct purpose, and 'Enable instantiation of lattice class in CTI files` is a nice thing to see in the commit log.

PR #647, opened 2019-06-12
Fix OpenMP on macOS

bryanwweber:
- Using the macOS clang compiler distributed by Apple requires different compiler options
- Explicitly include `Func1.h` in `zerodim.h`, it had been previously implicitly included but was inadvertently removed in #632 
- Change CMakeLists.txt to use OpenMP *and* Accelerate in the samples on macOS


bryanwweber:
Superseded by #649

PR #649, opened 2019-06-12
Build the samples on the CI services

bryanwweber:
- Using the macOS clang compiler distributed by Apple requires different compiler options
- Explicitly include `Func1.h` in `zerodim.h`, it had been previously implicitly included but was inadvertently removed in #632 
- Change CMakeLists.txt to use OpenMP *and* Accelerate in the samples on macOS
- Build the samples on the CI services using scons

### DIFF
--- a/samples/cxx/openmp_ignition/openmp_ignition.cpp
+++ b/samples/cxx/openmp_ignition/openmp_ignition.cpp
@@ -34,10 +34,10 @@ void run()
     }
 
     // Points at which to compute ignition delay time
-    size_t nPoints = 50;
+    int nPoints = 50;
     vector_fp T0(nPoints);
     vector_fp ignition_time(nPoints);
-    for (size_t i = 0; i < nPoints; i++) {
+    for (int i = 0; i < nPoints; i++) {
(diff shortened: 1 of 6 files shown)


bryanwweber:
I guess this supersedes #647. Whoops!

samples/cxx/openmp_ignition/openmp_ignition.cpp
```diff
@@ -51,7 +51,7 @@ void run()
     // thread in cases where the workload is biased, e.g. calculations for low
     // T0 take longer than calculations for high T0.
     #pragma omp parallel for schedule(static, 1)
-    for (size_t i = 0; i < nPoints; i++) {
+    for (int i = 0; i < nPoints; i++) {
```
speth, line 54: Needing to change this to a signed type is unfortunate. In this case, `nPoints` should also be changed to an `int`, as well other variables which are compared to `nPoints` (i.e. the loop index in the preceding loop) to avoid compiler warnings about "comparison between signed and unsigned integer expressions".

speth (approved):
Looks good. Feel free to merge after making the one suggested change.

PR #653, opened 2019-06-15
Update atomic weights and CODATA constants

### LINKED ISSUE #650: Update to 2018 CODATA physical constants
With the release of the new definition of the kilogram, we should update the CODATA physical constants and atomic weights of the elements. This issue serves as a discussion platform for how the constants should be defined.

## Physical constants

The most recent values of the constants can be found from NIST: https://physics.nist.gov/cuu/Constants/Table/allascii.txt

Some of the constants have a defined value, e.g., [Avogadro's Number](https://physics.nist.gov/cgi-bin/cuu/Value?k|search_for=adopted_in!) and the [Boltzmann Constant](https://physics.nist.gov/cgi-bin/cuu/Value?k|search_for=adopted_in!). However, in `ct_defs.h`, the Boltzmann constant is calculated:

https://github.com/Cantera/cantera/blob/ba8ac1d519b12fd19e80d0b05688d720a734664c/include/cantera/base/ct_defs.h#L77

Should we reverse this, so the Gas Constant is calculated from Boltzmann and Avogadro? Leave it as is? Or "overspecify" the values, since the [Gas Constant](https://physics.nist.gov/cgi-bin/cuu/Value?r|search_for=physchem_in!) is also given. I personally prefer the latter approach to avoid differences between platforms/compilers, etc. due to floating point arithmetic (small though the difference may be).

Also, should we add more constants to the list? Are there any others that would be useful to include?

## Atomic masses

There appear to be two sources for atomic masses, the [Standard Atomic Weights](http://www.ciaaw.org/atomic-weights.htm), which include a range of values for some elements depending on the relative isotopic abundance, or the [Current Atomic Masses](http://www.ciaaw.org/atomic-masses.htm). I'm not sure which of these we should use. It is also not clear to me whether or not these values have been adjusted for the new definition of the kilogram. Presumably not, since they were only updated in August 2018 and/or 2017.

bryanwweber:

Fixes #650 

- Update CODATA physical constants to the 2018 values
- Update atomic weights from the IUPAC periodic table (https://iupac.org/what-we-do/periodic-table-of-elements/)
- Reformat elements.xml


bryanwweber:
Expecting the tests to fail here, @speth can you take a look at these values and make sure they make sense? Especially for the atomic weights, most of them are just reducing the precision of the values that were already there. You can look specifically at the changes in 083b5ed and c19447c to see the actual value updates, aside from the other formatting changes.

include/cantera/base/ct_defs.h
```diff
@@ -49,62 +49,83 @@ using std::isnan; // workaround for bug in libstdc++ 4.8
...
-const doublereal Pi = 3.14159265358979323846;
+const double Pi = 3.14159265358979323846;
 
 /*!
- *   @name Variations of the Gas Constant
- * %Cantera uses the MKS system of units. The unit for moles
- * is defined to be the kmol.
+ * @name Defined Constants
+ * These constants are defined by CODATA to have a particular value.
+ * https://physics.nist.gov/cgi-bin/cuu/Category?view=html&Adopted+values.x=71&Adopted+values.y=14
```
speth: This URL looks pretty fragile to me. I would be more inclined to link to the landing page at `https://physics.nist.gov/cuu/Constants/index.html`.

include/cantera/base/ct_defs.h
```diff
@@ -49,62 +49,83 @@ using std::isnan; // workaround for bug in libstdc++ 4.8
...
+
+//! Stefan-Boltzmann constant [W/m2/K4]
+const double StefanBoltz = Pi * Pi * std::pow(Boltzmann, 4.0) / (60.0 * std::pow(Planck_bar, 3.0) * lightSpeed * lightSpeed); // 5.670374419e-8
+
+//! Faraday constant [C/kmol]
+const double Faraday = ElectronCharge * Avogadro;
 
-/// Permeability of free space \f$ \mu_0 \f$ in N/A^2.
-const doublereal permeability_0 = 4.0e-7*Pi;
+//! Permeability of free space [N/A2]
```
speth: I thought noting that the usual symbol for this is mu_0 was a useful reminder.

speth (changes requested):
* For the atomic weights, I think the approach should be as follows:
  * Use the "standard atomic weight" from http://www.ciaaw.org/atomic-weights.htm whenever a single value is given.
  * Otherwise, use the "conventional atomic weight" from the IUPAC PDF.
  The other values shown in the PDF are, despite the legend, the "abridged atomic weights".

* The atomic weights in `src/thermo/Elements.cpp` also need to be updated.

* The comment in `ct_defs.h` needs to be updated to mention that these are the 2018 CODATA values, rather than 2010.

* Physical constants show up in a couple of other files, where they can't be referenced back to `ct_defs.h`. The ones I am aware of are:
  * `ctml_writer.py` (hence `cti2yaml.py`)
  * `ck2cti.py`
  * `gasconstant.m`
  * `faradayconstant.m`

* I would skip the reformatting of `elements.xml` -- that file is on its way out the door with the XML format

include/cantera/base/ct_defs.h
bryanwweber: I agree the link is not ideal. I wanted to document that these numbers are the ones defined specifically under the "Defined Constants" page, i.e., they have no uncertainty because they are not measured. Do you think that distinction is worth preserving?

include/cantera/base/ct_defs.h
bryanwweber: OK, I removed it for consistency's sake. I will restore this and add symbols for other constants.

include/cantera/base/ct_defs.h
speth: I think noting that these are defined constants with exact values is useful. I just don't think the deep link is needed. Anyone who wants to check on these values can go to the higher-level page and search for the specific constant of interest.

bryanwweber:
> Use the "standard atomic weight" from http://www.ciaaw.org/atomic-weights.htm whenever a single value is given.

What about for elements that don't have a weight listed, e.g., technitium (atomic number 43), presumably because there aren't any stable isotopes? From [Wikipedia](https://en.wikipedia.org/wiki/Technetium)

> It is the lightest element whose isotopes are all radioactive; none are stable

I could potentially find a newer copy of the CRC handbook in the library at UConn to get those values, since they were apparently put in there for the 1989 edition.

bryanwweber:
I actually found an online version of the CRC handbook and that also doesn't list a value for the same elements. It does, however, have updated values for the weights of deuterium and tritium :tada:

speth:
I'm thinking we should remove the weights for the elements that IUPAC doesn't give values for due to the lack of stable isotopes. This is going to require a few changes in `elements.cpp`, since it will break the correspondence between list position and atomic number.

bryanwweber:
@speth There are still three Python tests failing here, but I can't see why. The test report is:
```
FAIL: test_mass_basis (cantera.test.test_thermo.TestThermoPhase)
  File "/Users/bryan/GitHub/cantera/build/python/cantera/test/test_thermo.py", line 325, in test_mass_basis
    self.assertEqual(self.phase.enthalpy_mass, self.phase.h)
AssertionError: 26468.504562941045 != 26468.50456294105

FAIL: test_molar_basis (cantera.test.test_thermo.TestThermoPhase)
  File "/Users/bryan/GitHub/cantera/build/python/cantera/test/test_thermo.py", line 339, in test_molar_basis
    self.assertEqual(self.phase.volume_mole, self.phase.v)
AssertionError: 24.617209824287904 != 24.6172098242879

FAIL: test_reaction_units (cantera.test.test_convert.chemkinConverterTest)
  File "/Users/bryan/GitHub/cantera/build/python/cantera/test/test_convert.py", line 324, in test_reaction_units
    [300, 800, 1450, 2800], [5e3, 1e5, 2e6], 1e-7)
  File "/Users/bryan/GitHub/cantera/build/python/cantera/test/test_convert.py", line 63, in checkKinetics
    self.assertNear(ref_kf[i], gas_kf[i], rtol=tol, msg='kf '+message)
  File "/Users/bryan/GitHub/cantera/build/python/cantera/test/utilities.py", line 56, in assertNear
    self.fail(message)
AssertionError: kf  for reaction 2 at T = 300, P = 5000.0
AssertNear: 286246917679.53 - 286246845542.49 = 72137.040466309
Relative error of 2.520099e-07 exceeds rtol = 1.000000e-07
```

In the first two cases, its comparing the long name of a property with the shorthand, so I'm not sure how its coming up different. In the last case, I'm not sure where the comparison data is coming from. I'll keep looking over the weekend maybe or next week, but if you have any ideas quickly I'd be happy to hear them 😄 Otherwise, I think this PR has all the necessary changes.

bryanwweber:
OK, I managed to fix the test failures. For the first two, I ended up just using `assertNear()` since it seemed to be failing because of floating point roundoff (`gas.h` is computed by `enthalpy_mol*_mole_factor`, whereas `enthalpy_mass` just goes right to C++). Similar for `v`. The other one I had to update the `A`s and `Ea`s in the input file due to the change in Avogadro's number and the Boltzmann constant.

data/inputs/elements.xml
```diff
@@ -1,929 +1,846 @@
 <ctml>
   <elementData caseSensitive="no">
-	<element name="H" atomicWt = "1.00794" atomicNumber = "1"> 
+  <!--  Values are used from CIAAW. Atomic weights of the elements 2017
+ when a single value is given. Available online at
+ http://www.ciaaw.org/atomic-weights.htm
+
+ When a range of values is given in the CIAAW table, the "conventional
+ atomic weight" from the IUPAC Periodic Table is used. Available
+ online at https://iupac.org/wp-content/uploads/2018/12/IUPAC_Periodic_Table-01Dec18.jpg
```
speth: The PDF at https://iupac.org/wp-content/uploads/2018/12/IUPAC_Periodic_Table-01Dec18.pdf is more readable.

data/inputs/elements.xml
```diff
@@ -1,929 +1,846 @@
...
+  <element name="Rh" atomicWt = "102.90549" atomicNumber = "45" >
            <entropy298 value = "31.54E3">
              <source>
                      The standard entropy was taken from Robie and
                      Hemingway (1979), Thermodynamic Properties of
-		     Minerals and Related Substances at 298.15 K
-		     and 1 bar (10^5 Pascals) Pressure abd at Higher 
-		     Temperatures, USGS Bulletin 1452, p. 90.
+         Minerals and Related Substances at 298.15 K
+         and 1 bar (10^5 Pascals) Pressure abd at Higher
```
speth: If you're changing so much else in this file, you may as well fix the repeated typo of "and".

data/inputs/elements.xml
```diff
@@ -1,929 +1,846 @@
...
-		    Handbook (Chase 1998), Journal of Physical and 
-		    Chemical Reference Data, Monograph 9, p. 1040.  
+        The standard entropy (1/2 D2 gas) was taken from the NIST-JANAF
+        Handbook (Chase 1998), Journal of Physical and
+        Chemical Reference Data, Monograph 9, p. 1040.
              </source>
            </entropy298>
         </element>
-	<element name="Tr" atomicWt = "3.016327" atomicNumber = "1" >
+  <element name="Tr" atomicWt = "2.010666" atomicNumber = "1" >
```
speth: Surely this should be 3.something. Wikipedia has 3.01604928199 as the mass, but irritatingly has no citation, and I can't find the value on the CIAAW site.

src/thermo/Elements.cpp
```diff
@@ -158,24 +195,33 @@ static struct atomicWeightData atomicWeightTable[] = {
  * The size of the table is given by the initial instantiation.
  */
 static struct isotopeWeightData isotopeWeightTable[] = {
-    {"D",  "deuterium", 2.0, 1},
-    {"Tr", "tritium",   3.0, 1},
+    {"D",  "deuterium", 2.014, 1},
+    {"Tr", "tritium",   3.016, 1},
```
speth: These values should be the same as in `elements.xml`, right?

src/thermo/Elements.cpp
```diff
@@ -158,24 +195,33 @@ static struct atomicWeightData atomicWeightTable[] = {
...
-            return atomicWeightTable[i].atomicWeight;
+            elementWeight = atomicWeightTable[i].atomicWeight;
+            break;
         }
     }
+    if (elementWeight > 0.0) {
+        return elementWeight;
+    } else if (elementWeight < 0.0) {
+        throw CanteraError("getElementWeight",
+            "element has no stable isotopes: " + ename);
```
speth:
```suggestion
            "element '{}' has no stable isotopes", ename);
```

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1300,20 +1300,24 @@ def test_element_atomic_number_big(self):
         with self.assertRaisesRegex(ct.CanteraError, 'IndexError'):
             ct.Element(num_elements + 1)
 
+    def test_element_no_weight(self):
+        with self.assertRaisesRegex(ct.CanteraError, 'CanteraError'):
```
speth: You can check for something more specific in the Regex, given that just about any `CanteraError` will contain the string `CanteraError`.

include/cantera/thermo/Elements.h
```diff
@@ -93,7 +93,8 @@ namespace Cantera
  *
  * @param ename String, name or symbol of the element
  * @return The atomic weight of the element
- * @exception CanteraError If a match is not found, throws a CanteraError
+ * @exception CanteraError if a match for ename is not found or the
+ * element has no stable isotopes
```
speth:
```suggestion
 * element has no stable isotopes (and therefore no standard atomic weight)
```

src/thermo/Elements.cpp
```diff
@@ -192,7 +238,13 @@ double getElementWeight(int atomicNumber)
     if (atomicNumber > num || atomicNumber < 1) {
         throw IndexError("getElementWeight", "atomicWeightTable", atomicNumber, num);
     }
-    return atomicWeightTable[atomicNumber - 1].atomicWeight;
+    double elementWeight = atomicWeightTable[atomicNumber - 1].atomicWeight;
+    if (elementWeight < 0.0) {
+        string ename = getElementName(atomicNumber);
+        throw CanteraError("getElementWeight",
+            "element has no stable isotopes: " + ename);
```
speth:
```suggestion
            "element '{}' has no stable isotopes: ", getElementName(atomicNumber));
```
and eliminate the variable `ename`.

speth:
I had a few suggested modifications, and a couple small additions which I would appreciate your confirmation on.

speth (changes requested):

data/inputs/elements.xml
bryanwweber: That looks like exactly the same link that I put in the file?

speth: You linked to the `.jpg`

bryanwweber: What the.... How did I manage that? Well, I wanted the PDF so thanks for catching it 🤣

data/inputs/elements.xml
bryanwweber: I found the data for deuterium and tritium from Pubchem and added links to `Elements.cpp`.

bryanwweber:
OK I think I got all the updates.

Commit 0784eff fixes the IonBurnerFlame test that started failing on my machine. According to git bisect, the problem commit was 5dea810 but that passed the CI tests, so I'm not sure what's going on there. Also, the equivalent IonFreeFlame test is not failing, so that seems very strange and like something is wrong with the tests or the implementation (i.e., if one fails, both should fail).

data/inputs/elements.xml
bryanwweber: I think I originally found the weight of tritium (T<sub>2</sub>) from the CRC handbook and divided by 3 instead of 2 🤦‍♂

speth (approved):

speth:
Where did this reference value of 580.09 show up? I'm still seeing 552.33 as the result locally and on all of the CI servers.

bryanwweber:
Running the tests locally here results in that number. I'm really not sure why, but `git bisect` localized it to the commit I referenced earlier. I guess I'll just drop that commit since it doesn't result anywhere else.

bryanwweber:
@speth Can you confirm that I've gotten all the masses correct, particularly the isotopes that you caught before?

speth:
I found the more precise values for deuterium and tritium by following the source that is referenced by CIAAW on this page: http://www.ciaaw.org/atomic-masses.htm.

PR #657, opened 2019-06-20
Rename ambiguous variable diag

ischoegl:
This PR clarifies the meaning of ambiguously named variables (`diag`/`diagg`) in the C++ core of `OneD`, which indicate whether the governing equation within the state vector is to be handled as transient or a steady state. 

While the function signature for `Domain1D::eval` documents a parameter `mask`, there is no consistency in overloaded functions (`StFlow1D::eval` uses `diag`/`diagg`, etc.). Further, function parameter names in `StFlow.cpp` do not match declarations in `StFlow.h`.

- use consistent variable names (i.e. rename `diagg` to `tmaskg` and `diag` to `tmask`, and `OneDim::m_mask` to `OneDim::m_tmask`)
- consistently add `g` to global state/residual vectors and parameters
- assign boolean values instead of `1` and `0` for clarity, and take advantage of C++ Integral promotion
- this PR also differentiates `MultiJac::m_mask` from `OneDim::m_tmask`, which are now clearly separated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -25,6 +25,7 @@ Jon Kristofer
 Kyle Linevitch, Jr.
 Christopher Leuth
 Nicholas Malaya, University of Texas at Austin
+Thanasis Mattas, Aristotle University of Thessaloniki
 Evan McCorkle
 Ivan Mitrichev, Mendeleev University of Chemical Technology of Russia
 Harry Moffat, Sandia National Laboratory
@@ -33,6 +34,8 @@ Kyle Niemeyer, Oregon State University
 @pwcnorthrop
 Andreas Rücker
 Jeff Santner
+Satyam Saxena
+Ingmar Schoegl, Louisiana State University
 Santosh Shanbhogue, Massachusetts Institute of Technology
 David Sondak
 Raymond Speth, Massachusetts Institute of Technology
--- a/include/cantera/oneD/Boundary1D.h
+++ b/include/cantera/oneD/Boundary1D.h
@@ -42,12 +42,12 @@ public:
     }
 
     /// Set the temperature.
-    virtual void setTemperature(doublereal t) {
+    virtual void setTemperature(double t) {
         m_temp = t;
     }
 
     /// Temperature [K].
-    virtual doublereal temperature() {
+    virtual double temperature() {
         return m_temp;
     }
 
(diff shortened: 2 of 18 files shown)


ischoegl:
The OSX check currently appears to fail for reasons unrelated to the PR itself: scons is not found (?!)
```
Users/travis/.travis/functions: line 132: scons: command not found
```

speth:
I'm generally on-board with renaming these variables, having been confused about them myself at various times. The best interpretation I have of the name `diag` is that it means "is there a false transient on the diagonal for this equation?"

Don't worry about the Travis failure for macOS -- I have fixes for this that are currently in the YAML PR.

ischoegl:
Repeated confusion about the meaning of `diag` is indeed what prompted this PR. You are probably correct about the original rationale for `diag` (... see `MultiJac::updateTransient`)

ischoegl:
~~As an aside: the `tmask` is really used as a boolean vector, but implemented as an integer vector: is this worth changing? (this *will* affect function signatures, which I've been avoiding thus far).~~ Also, what is the guideline for `doublereal` vs. `double` and `integer` vs. `int`? 

I can also wrap another comment of yours into this PR:
> Make class and file names clear and consistent: Inlet1D.h should be named Boundary1D.h, and class Bdry1D should be named Boundary1D.

ischoegl:
PS: to follow up on my own questions: 
* ~~an alternative `vector<bool>` has some inconvenient features (does not save as contiguous array), so an alternative implementation is not obvious ... i.e. it's not worth changing.~~
* I also realized that `MultiJac::transientMask()` and the member variable `MultiJac::m_mask` are not used anywhere (the mask is exclusively owned by `OneDim::m_tmask`). I added a deprecation warning in case anyone is using it
* names for `bdry` -> `boundary` are replaced in a second commit within this PR; a transitional `Bdry1D` class raises a deprecation warning.

ischoegl:
@speth ... after rebasing, travis indeed runs through.

ischoegl:
Closing as 1D solvers need much more work. Will put together proposal for a more comprehensive PR.

bryanwweber:
@ischoegl Glad to see you updated the AUTHORS list. I was going to remind you to do that as this PR progressed 😄

ischoegl:
@bryanwweber ... I moved the AUTHORS list to my other PR. I honestly believe that 1D needs major work that will be better done once 2.5.0 is stable. I am currently developing code that builds on the current code base and it is quite apparent that `oneD` should be updated with some of the same philosophy that was put into `zeroD`. Ray's project ideas go into that direction and I don't think it makes much sense to incrementally improve the status quo.

PR #659, opened 2019-06-25
Exposing getDeltaEnthalpies to Matlab interface.

decaluwe:
The general intent here was to enable calculating reaction enthalpies in the Matlab toolbox, as part of the li-ion battery simulations in PR #563.

This required several changes:

- Create getDeltaEnthalpies.m in Matlab toolbox/@Kinetics
- Add kin_getDelta to kineticsmethods.cpp.  Note that this can be easily extended to create other functions in the matlab Kinetics toolbox (see the other fucntions listed in kin_getDelta in src/clib/ct.cpp).
- Add getPartialMolarEnthalpies to metalPhase class (it returns all zeros).

### DIFF
--- a/interfaces/matlab/toolbox/@Kinetics/getDeltaEnthalpies.m
+++ b/interfaces/matlab/toolbox/@Kinetics/getDeltaEnthalpies.m
@@ -1,0 +1,14 @@
+function dH = getDeltaEnthalpies(a)
+% GETDELTAENTHALPIES  Get the enthalpy of reaction for each reaction.
+% dH = getDeltaEnthalpies(a)
+%
+% :param a:
+%     Instance of class :mat:func:`Kinetics` (or another
+%     object deriving from Kinetics) for which the enthalpies of
+%     reaction are desired.
+% :return:
+%     Returns a vector of the enthalpy of reaction for each
+%     reaction. Units: J/kmol
+%
+
+dH = kinetics_get(a.id, 17, 0);
--- a/interfaces/matlab/toolbox/@Kinetics/getDeltaEntropies.m
+++ b/interfaces/matlab/toolbox/@Kinetics/getDeltaEntropies.m
@@ -1,0 +1,14 @@
+function dS = getDeltaEntropies(a)
+% GETDELTAENTROPIES  Get the entropy of reaction for each reaction.
+% dS = getDeltaEntropies(a)
+%
+% :param a:
+%     Instance of class :mat:func:`Kinetics` (or another
+%     object deriving from Kinetics) for which the entropies of
+%     reaction are desired.
+% :return:
+%     Returns a vector of the entropy of reaction for each
+%     reaction. Units: J/kmol-K
+%
+
+dS = kinetics_get(a.id, 17, 2);
--- a/src/matlab/kineticsmethods.cpp
+++ b/src/matlab/kineticsmethods.cpp
@@ -103,6 +103,9 @@ void kineticsmethods(int nlhs, mxArray* plhs[],
             case 16:
                 ok = kin_getRevRateConstants(kin,1,nr,h);
                 break;
+            case 17:
+                ok = kin_getDelta(kin,getInt(prhs[3]),nr,h);
+                break;
             default:
                 ;
             }
--- a/include/cantera/thermo/MetalPhase.h
+++ b/include/cantera/thermo/MetalPhase.h
@@ -85,6 +85,11 @@ public:
             c[n] = 1.0;
         }
     }
+    virtual void getPartialMolarEnthalpies(doublereal *h) const {
+        for (size_t n = 0; n < nSpecies(); n++) {
+            h[n] = 0.0;
+        }
+    }
 
     virtual Units standardConcentrationUnits() const {
         return Units(1.0);
--- a/interfaces/matlab/toolbox/@Kinetics/getDeltaGibbs.m
+++ b/interfaces/matlab/toolbox/@Kinetics/getDeltaGibbs.m
@@ -1,0 +1,14 @@
+function dG = getDeltaGibbs(a)
+% GETDELTAGIBBS  Get the Gibbs free energy of reaction for each reaction.
+% dG = getDeltaGibbs(a)
+%
+% :param a:
+%     Instance of class :mat:func:`Kinetics` (or another
+%     object deriving from Kinetics) for which the Gibbs free
+%     energies of reaction are desired.
+% :return:
+%     Returns a vector of the Gibbs free energy of reaction
+%     for each reaction. Units: J/kmol
+%
+
+dG = kinetics_get(a.id, 17, 1);


interfaces/matlab/toolbox/@Kinetics/getDeltaEnthalpies.m
```diff
@@ -0,0 +1,13 @@
+function dH = getDeltaEnthalpies(a)
+% getDeltaEnthalpies  Get the enthalpy of reaction for each reaction.
+% dH = getDeltaEnthalpies(a)
+%
+% :param a:
+%     Instance of class :mat:func:`Kinetics` (or another
+%     object deriving from Kinetics)
+%     for which the net stoichiometric coefficients are desired.
+% :return:
+%     Returns a vector of the enthalpy of reactions for each reaction.  If a
```
bryanwweber: Missing the end of this sentence 😄

interfaces/matlab/toolbox/@Kinetics/getDeltaEnthalpies.m
```diff
@@ -0,0 +1,13 @@
...
+%
+% :param a:
+%     Instance of class :mat:func:`Kinetics` (or another
+%     object deriving from Kinetics)
+%     for which the net stoichiometric coefficients are desired.
+% :return:
+%     Returns a vector of the enthalpy of reactions for each reaction.  If a 
+%
+
+dH = kinetics_get(a.id, 17, 0);
```
bryanwweber: Should be one blank line at the end of the file

interfaces/matlab/toolbox/@Kinetics/getDeltaEnthalpies.m
```diff
@@ -0,0 +1,13 @@
+function dH = getDeltaEnthalpies(a)
+% getDeltaEnthalpies  Get the enthalpy of reaction for each reaction.
+% dH = getDeltaEnthalpies(a)
+%
+% :param a:
+%     Instance of class :mat:func:`Kinetics` (or another
+%     object deriving from Kinetics)
+%     for which the net stoichiometric coefficients are desired.
```
bryanwweber: Copy-pasta 😸

interfaces/matlab/toolbox/@Kinetics/getDeltaEnthalpies.m
```diff
@@ -0,0 +1,13 @@
+function dH = getDeltaEnthalpies(a)
+% getDeltaEnthalpies  Get the enthalpy of reaction for each reaction.
```
bryanwweber: The function name on this line should be ALL CAPS

src/matlab/kineticsmethods.cpp
```diff
@@ -97,6 +97,9 @@ void kineticsmethods(int nlhs, mxArray* plhs[],
             case 16:
                 ok = kin_getRevRateConstants(kin,1,nr,h);
                 break;
+            case 17:
+                ok = kin_getDelta(kin,getInt(prhs[3]),nr,h);
```
bryanwweber, line 107: Why is it called just `kin_getDelta`?

bryanwweber (changes requested):
Seems fine, some typos to fix though 😃

src/matlab/kineticsmethods.cpp
decaluwe: As referenced in the PR message, this is a pre-existing function in clib/ct.cpp that provides an interface to a range of 'delta' functions for reaction thermo - a single interface which provides access to deltaGibbs, deltaEntropy, etc., based on the 'job' number fed in as prhs[3].  As with other Matlab toolbox issues, rather than parsing the cost-benefit of this interface and the use of "magic numbers," I decided to punt and use as is, until the Matlab toolbox revamp.

interfaces/matlab/toolbox/@Kinetics/getDeltaEnthalpies.m
decaluwe: Ha - opening in atom fixed it for me, automatically (the original was created in VS Code).

bryanwweber: Yes, its annoying that's not the default in VS Code.

src/matlab/kineticsmethods.cpp
bryanwweber: Right, reading comprehension is apparently not something I'm capable of today 🙄  I'm torn, then, between adding the `.m` files for the other `case`s in `kin_getDelta` and just leaving well enough alone. On the one hand, the files are basically trivial and potentially useful; on the other, "just do it in the rewrite that may or may not ever happen".

interfaces/matlab/toolbox/@Kinetics/getDeltaEnthalpies.m
decaluwe: 😳

src/matlab/kineticsmethods.cpp
decaluwe: As referenced in the PR message, `kin_getDelta` is a pre-existing function in ctlib/ct.cpp.  It accepts a job number (`prhs[3]`), and calculates a range of reaction thermo terms (`deltaGibbs`, `deltaEntropy`, and standard state versions of all three).

Rather than wade into the cost benefit of this "magic number" approach, as with other recent changes I'm going to go with the Matlab toolbox as is, and wait for the toolbox overhaul before making any broader edits to the Matlab interface...

decaluwe: Yeah, it's basically trivial - I'm fine adding them.

bryanwweber: Another thought I had is that you're doubling up on the magic numbers in the `.m` file here by using `prhs[3]`. Why not just eliminate that argument (`0`) from the call in `getDeltaEnthalpies.m` and just specify the `0` in `kin_getDelta`?

decaluwe: So the question is if we want to also expose those other `getDelta` functions.  If we do, your suggestion above would mean we need a separate magic number (`17`, `18`, etc.) for each of `getDeltaEnthalpies`, `getDeltaGibbs`, etc., at the `src/matlab/kineticsmethods.cpp` level.  In the original, there is just one magic number at this level, then a second magic number at the `kin_getDelta` level.

The latter is a little clearer in reflecting the structure of the code - they all call a common function, and pass a 2nd argument to that function.  But I don't know if that really matters, much.  The former has the benefit of exposing the user to fewer magic numbers, and also avoids the conflict `kineticsmethods.cpp` wherein `prhs[3]` refers to this flag in our case, but is also assigned to `irxn` on line 159.

I don't have a super-strong preference.  If forced to choose, I'd go with omitting the `0` from `getDeltaEnthalpies.m`.  But then my vote on adding the other 'delta' functions to to the matlab interface would probably be "no."

bryanwweber: As a counter-counter-point, all these kinds of kinetics functions are supposed to fit between job ids 10 and 19. Since its already at 17, there aren't enough "spaces" to add all 6 functions as each their own `case` in `kineticsmethods.cpp`. So that suggests using the one magic number (`17`) as the job id and the second magic number as the item to retrieve. Boy, this is awful.

> avoids the conflict `kineticsmethods.cpp` wherein `prhs[3]` refers to this flag in our case, but is also assigned to `irxn` on line 159.

Given the other ridiculosity in this interface, I feel like "overloading" `prhs[3]` is the least-confusion-causing thing about this 😂

decaluwe: Lol. 

Okay, I will stick with the current approach, in terms of `getDeltaEnthalpies.m` and the two magic numbers.

bryanwweber (approved):

interfaces/matlab/toolbox/@Kinetics/getDeltaEntropies.m
```diff
@@ -0,0 +1,14 @@
...
+% :param a:
+%     Instance of class :mat:func:`Kinetics` (or another
+%     object deriving from Kinetics) for which the entropies of
+%     reaction are desired.
+% :return:
+%     Returns a vector of the entropy of reaction for each
+%     reaction. Units: J/kmol-K
+%
+
+dH = kinetics_get(a.id, 17, 2);
```
bryanwweber: Sorry, I know I already approved this, but this should probably be dS and the Gibbs should probably be dG

speth: I was about to make the same request! :grin:

decaluwe: You guys are _so rigid_ in your style preferences. :)  

Fixed now - thanks for the catch!

PR #660, opened 2019-06-26
Deprecate magic numbers in zeroD CLIB

### LINKED ISSUE #638: Deprecate magic numbers in zeroD CLIB interface
### Cantera version

All / current development version (2.5.0a2)

### Operating System

All

### Python/MATLAB version

N/A

### Expected Behavior

The use of _magic numbers_ as type constants should be deprecated throughout Cantera in favor of string types. I.e. type queries should have a signature `virtual std::string type() const`, like in class `ThermoPhase`.

### Actual Behavior

All current objects in `zeroD` still use magic numbers with a signature `int type() const`. While a resolution of this issue in the cython interface is straight-forward, the current structure of the CLIB interface depends on integer-based object creation, i.e. internal integer `m_type` class variables are still essential.

### Steps to reproduce
N/A

ischoegl:

Fixes #638

- add deprecation warning for `int ReactorBase::type()` (to be changed after Cantera 2.5)
- introduce temporary `std::string ReactorBase::typeStr()` (to be renamed after Cantera 2.5) 
- deprecate all functions using the old call and introduce associated temporary functions

Temporary functions are necessary as overloading by return argument type is not possible in C++ (same for overloading of C functions). 

~~PS: Updates are threaded through MATLAB, although I do not have a way to test at the moment (I am unsure whether unit tests are implemented)~~ Edit: MATLAB testing is included in this PR.

### DIFF
--- a/include/cantera/zeroD/IdealGasReactor.h
+++ b/include/cantera/zeroD/IdealGasReactor.h
@@ -21,7 +21,18 @@ class IdealGasReactor : public Reactor
 public:
     IdealGasReactor() {}
 
+    virtual std::string typeStr() const {
+        return "IdealGasReactor";
+    }
+
+    /*!
+     * @deprecated To be changed after Cantera 2.5.
+     */
     virtual int type() const {
+        warn_deprecated("IdealGasReactor::type()",
+                        "To be changed after Cantera 2.5. "
+                        "Return string instead of magic number; use "
+                        "IdealGasReactor::typeStr() during transition");
         return IdealGasReactorType;
     }
 
--- a/include/cantera/zeroD/ReactorBase.h
+++ b/include/cantera/zeroD/ReactorBase.h
@@ -15,7 +15,10 @@ class FlowDevice;
 class WallBase;
 class ReactorNet;
 class ReactorSurface;
+class Kinetics;
 
+//! Magic numbers
+//! @deprecated To be removed after Cantera 2.5.
 const int ReservoirType = 1;
 const int ReactorType = 2;
 const int FlowReactorType = 3;
@@ -49,8 +52,19 @@ public:
     ReactorBase(const ReactorBase&) = delete;
     ReactorBase& operator=(const ReactorBase&) = delete;
 
+    //! String indicating the reactor model implemented. Usually
+    //! corresponds to the name of the derived class.
+    virtual std::string typeStr() const {
+        return "ReactorBase";
+    }
+
     //! Return a constant indicating the type of this Reactor
+    //! @deprecated To be changed after Cantera 2.5.
     virtual int type() const {
+        warn_deprecated("ReactorBase::type()",
+                        "To be changed after Cantera 2.5. "
+                        "Return string instead of magic number; use "
+                        "ReactorBase::typeStr() during transition");
         return 0;
     }
 
@@ -76,6 +90,21 @@ public:
     //! this substance is stored, and as the integration proceeds, the state of
     //! the substance is modified.
     virtual void setThermoMgr(thermo_t& thermo);
+
+    //! Specify chemical kinetics governing the reactor.
+    virtual void setKineticsMgr(Kinetics& kin) {
+        throw NotImplementedError("ReactorBase::setKineticsMgr");
+    }
+
+    //! Enable or disable changes in reactor composition due to chemical reactions.
+    virtual void setChemistry(bool cflag = true) {
+        throw NotImplementedError("ReactorBase::setChemistry");
+    }
+
+    //! Set the energy equation on or off.
+    virtual void setEnergy(int eflag = 1) {
+        throw NotImplementedError("ReactorBase::setEnergy");
+    }
 
     //! Connect an inlet FlowDevice to this reactor
     void addInlet(FlowDevice& inlet);
--- a/interfaces/matlab/toolbox/@Reactor/Reactor.m
+++ b/interfaces/matlab/toolbox/@Reactor/Reactor.m
@@ -18,14 +18,14 @@ function x = Reactor(contents, typ)
 %     Instance of class :mat:func:`Solution` representing the contents of the
 %     reactor
 % :param typ:
-%     Integer, reactor type. Options are:
+%     Character array, reactor type. Options are:
 %
-%     1. Reservoir
-%     2. Reactor
-%     3. Flow Reactor
-%     4. Constant Pressure Reactor
-%     5. Ideal Gas Reactor
-%     6. Ideal Gas Constant Pressure Reactor
+%     'Reservoir'
+%     'Reactor'
+%     'FlowReactor'
+%     'ConstPressureReactor'
+%     'IdealGasReactor'
+%     'IdealGasConstPressureReactor'
 %
 % :return:
 %     Instance of class :mat:func:`Reactor`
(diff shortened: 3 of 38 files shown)


include/cantera/zeroD/IdealGasReactor.h
```diff
@@ -21,7 +21,18 @@ class IdealGasReactor : public Reactor
 public:
     IdealGasReactor() {}
 
+    virtual std::string typeStr() const {
+        return "IdealGasReactor";
+    }
+
+    /*!
+     * @deprecated To be removed after Cantera 2.5.
```
speth:
```suggestion
     * @deprecated To be changed after Cantera 2.5.
```

src/matlab/reactormethods.cpp
```diff
@@ -23,7 +22,9 @@ void reactormethods(int nlhs, mxArray* plhs[],
 
     // constructor
     if (job == 0) {
-        n = reactor_new(i);
+        char* type = getInt(prhs[2]);
```
speth: Shouldn't this be `getString`?

src/matlab/flowdevicemethods.cpp
```diff
@@ -20,7 +18,9 @@ void flowdevicemethods(int nlhs, mxArray* plhs[],
 
     // constructor
     if (job == 0) {
-        n = flowdev_new(i);
+        char* type = getInt(prhs[2]);
```
speth: Shouldn't this be `getString`?

include/cantera/zeroD/ReactorBase.h
```diff
@@ -228,6 +241,11 @@ class ReactorBase
     //! Set the ReactorNet that this reactor belongs to.
     void setNetwork(ReactorNet* net);
 
+    //! Return whether ReactorBase object handles equations
```
speth: A slightly more detailed explanation would be useful here, although I'm not exactly sure of how to phrase the distinction. 

And actually, looking at the place where this is used in `clib`, I wonder whether simply removing the check would resolve the associated "todo" items, and eliminate the need for adding this method.

include/cantera/zeroD/ReactorBase.h
```diff
@@ -49,8 +51,19 @@ class ReactorBase
...
     ReactorBase& operator=(const ReactorBase&) = delete;
 
+    //! String indicating the reactor model implemented. Usually
+    //! corresponds to the name of the derived class.
+    virtual std::string typeStr() const {
+        return "ReactorBase";
+    }
+
     //! Return a constant indicating the type of this Reactor
+    //! * @deprecated To be changed after Cantera 2.5.
```
speth:
```suggestion
    //! @deprecated To be changed after Cantera 2.5.
```

src/matlab/reactormethods.cpp
ischoegl: correct.

include/cantera/zeroD/IdealGasReactor.h
ischoegl: changed.

src/matlab/flowdevicemethods.cpp
ischoegl: correct.

include/cantera/zeroD/ReactorBase.h
ischoegl: fixed.

include/cantera/zeroD/ReactorBase.h
ischoegl: Good point. I eliminated `hasEquations` and instead defined `setKineticsMgr` in `ReactorBase` as virtual (raising a `NotImplementedError`). In this way, `clib` throws a clearly defined error. For the remaining instances (`setChemisty` / `setEnergy`), I simply removed the check as you suggested.

ischoegl:
@speth ... I believe comments are addressed.

~~As mentioned earlier, I currently don't have a way to check matlab (I don't have a working toolchain at the moment).~~

src/matlab/wallmethods.cpp
```diff
@@ -23,7 +22,8 @@ void wallmethods(int nlhs, mxArray* plhs[],
 
     // constructor
     if (job == 0) {
-        n = wall_new(i);
+        char* type = getInt(prhs[2]);
+        n = wall_new2(i);
```
ischoegl: Typo: Needs to be `type`

ischoegl:
WIP ... at this point, I finally have a working toolchain for matlab. I will report once the remaining issues are addressed.

ischoegl:
@speth ... Matlab testing is now complete: the example script `test_examples.m` now runs without errors (tested for R2019).

interfaces/matlab/toolbox/@Reactor/Reactor.m
```diff
@@ -18,33 +18,42 @@
...
+%     "IdealGasConstPressureReactor"
 %
 % :return:
 %     Instance of class :mat:func:`Reactor`
 %
 
 if nargin == 0
     contents = 0;
-    typ = 2;
+    typ = "Reactor";
```
speth: One potential issue here is that "strings" (double-quoted) are only supported in Matlab 2016b or newer -- Matlab 2016a and earlier only have "character arrays (single quoted) which behave a bit differently. I'm not sure what our minimum Matlab version is at present (the [website](https://cantera.org/compiling/dependencies.html#optional-programs) claims it works with 2009a, but I'm not sure if that's still true, and don't think we need to support anything that old in any case), but if we want to use these, we will need to note the version requirement.

src/matlab/flowdevicemethods.cpp
```diff
@@ -1,5 +1,5 @@
 // This file is part of Cantera. See License.txt in the top-level directory or
-// at http://www.cantera.org/license.txt for license and copyright information.
+// at http://cantera.org/license.txt for license and copyright information.
```
speth: I think always using `https` is the main change that Bryan wants to have in these; omitting the `www` is secondary.

src/matlab/flowdevicemethods.cpp
```diff
@@ -32,7 +32,9 @@ void flowdevicemethods(int nlhs, mxArray* plhs[],
 
     // options that do not return a value
 
+    int i = getInt(prhs[2]);
     if (job < 20) {
+
```
speth: Blank line here seems unnecessary.

speth (commented):
I think the only decision to be made here is on the Matlab version requirement. If changing this to support Matlab < 2016b is easy, then I would say we should do that. If it's annoying to refactor to do so, I don't really have a problem with setting that as the version requirement.

interfaces/matlab/toolbox/@Reactor/Reactor.m
ischoegl: That's an interesting point. Comparisons of character arrays are a little more annoying than strings. I wasn't aware of the fact that strings were only introduced 3 years ago, but will look into this as it's not reason enough to lose compatibility.

src/matlab/flowdevicemethods.cpp
ischoegl: corrected.

ischoegl:
@speth ... thank you for the review. Replacing strings in Matlab was actually quite straight-forward (cell arrays of char arrays), and it's back to being compatible beyond 2016b.

speth (approved):

speth:
@ischoegl The reason your commits aren't being properly attributed is because the "author" field is being filled in using your GMail address, presumably based on your Git configuration on one of your machines. You can get Github to link these commits to your account by adding that address as an alternate address for your Github account.

ischoegl:
@speth ... this leaves me stumped: I have my primary email set to my gmail and backup set to the lsu domain (I added the latter recently). I just looked at my commit and it shows my gmail (!). I.e. I already have both my e-mails associated with the account.

PR #664, opened 2019-06-28
Fix YAML/Cython tests

bryanwweber:
These changes fix a few things that broke when testing the installed Cython interface (via `python -m unittest -v cantera.test`)

- Install the `test/data/test_subdir` in the Cython interface
- Avoid using paths in the test files that implicitly relied on being in the source tree
- Write test conversions into the `test_work_dir` instead of the cwd

### DIFF
--- a/interfaces/cython/cantera/test/test_convert.py
+++ b/interfaces/cython/cantera/test/test_convert.py
@@ -1,6 +1,7 @@
 import os
 from os.path import join as pjoin
 import itertools
+from pathlib import Path
 
 from . import utilities
 import cantera as ct
@@ -565,7 +566,8 @@ class cti2yamlTest(utilities.CanteraTest):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        cti2yaml.convert(pjoin(cls.cantera_data, 'gri30.cti'), 'gri30.yaml')
+        cti2yaml.convert(Path(cls.cantera_data).joinpath('gri30.cti'),
+                         Path(cls.test_work_dir).joinpath('gri30.yaml'))
 
     def checkConversion(self, basename, cls=ct.Solution, ctiphases=(),
                         yamlphases=(), **kwargs):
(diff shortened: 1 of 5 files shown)


interfaces/cython/cantera/test/test_convert.py
```diff
@@ -565,7 +566,8 @@ class cti2yamlTest(utilities.CanteraTest):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        cti2yaml.convert(pjoin(cls.cantera_data, 'gri30.cti'), 'gri30.yaml')
+        cti2yaml.convert(Path(cls.cantera_data).joinpath('gri30.cti'),
+                         Path(cls.test_work_dir).joinpath('gri30.yaml'))
```
speth, line 570: If we're going all in on the use of `pathlib`, why don't we just go ahead and convert the `cantera_data`, `test_work_dir`, and `test_data_dir` variables to instances of `pathlib.Path`?

speth:
Don't forget to `rebase -i` so that your fixup commit gets squashed correctly (I'm assuming GitHub doesn't do `autosquash`).

interfaces/cython/cantera/test/test_convert.py
bryanwweber: Sure. The test failures are because cti2yaml expects a string as the filename, so I'm working on making that pathlib compatible.

bryanwweber: So at the moment, most of the other test cases assume that `cantera_data` et al. are strings, and pass them to various `os.path` functions that only work with strings in Python 3.5. 3.6 upgrades all the `os.path` functions to also work with pathlib-like objects, but since we're still supporting 3.5 and I don't think it makes sense to update all of the ck2cti tests that will be going away, I'm not going to change this for now.

bryanwweber:
> Don't forget to `rebase -i` so that your fixup commit gets squashed correctly (I'm assuming GitHub doesn't do `autosquash`).

Once the changes are approved, I'll autosquash and then it can be merged.

bryanwweber:
This is rebased and squashed now. :+1:?

bryanwweber:
I'll merge this tomorrow if no further comments. This is blocking for Cantera/conda-recipes#19

speth (approved):

PR #672, opened 2019-07-24
Fix to build Cantera against Sundials 4.0.0 and newer

### LINKED ISSUE #620: SUNDIALS 4.1.0 does not work with Cantera 2.4.0
### Cantera version
2.4.0
### Operating System
Linux, RHEL6
### Python/MATLAB version
3.6.3
### Expected Behavior
Build successfully
### Actual Behavior
Failed because `CV_NEWTON` cannot be found in sundials 4.1.0

last part of `config.log` is 
```
scons: Configure: Checking for CVodeCreate(CV_BDF, CV_NEWTON) in C++ library sundials_cvodes... 
.sconf_temp/conftest_10.cpp <-
  |
  |
  |#include "cvodes/cvodes.h"
  |int
  |main() {
  |  CVodeCreate(CV_BDF, CV_NEWTON);
  |return 0;
  |}
  |
g++ -o .sconf_temp/conftest_10.o -c -std=c++0x -pthread -O3 -Wno-inline -g -DNDEBUG -I/PARA/pp569/share/eigen/3.3.7/include/eigen3 -I/PARA/pp569/share/fmt/5.3.0/include -I/PARA/pp569/share/googletest/1.8.1/include -I/PARA/pp569/share/boost/1.69.0/include -I/PARA/pp569/share/sundials/4.1.0/include .sconf_temp/conftest_10.cpp
.sconf_temp/conftest_10.cpp: In function 'int main()':
.sconf_temp/conftest_10.cpp:6:23: error: 'CV_NEWTON' was not declared in this scope
   CVodeCreate(CV_BDF, CV_NEWTON);
                       ^
scons: Configure: no
```
### Steps to reproduce

just install sundials 4.1.0 and compile cantera 2.4.0

I found this:
`With the introduction of SUNNonlinSol modules, the input parameter iter to CVodeCreate has been **removed** along with the function CVodeSetIterType and the **constants CV_NEWTON** and CV_FUNCTIONAL.`
in https://computation.llnl.gov/projects/sundials/release-history

And I searched the headers of sundials but no `CV_NEWTON` keyword was found.

According to 

https://github.com/Cantera/cantera/blob/master/.gitmodules

and 

https://github.com/Cantera/sundials-mirror

sundials 3.1.0 was used by default.

So I think it is caused by upgrade of sundials.

band-a-prend:

Fixes #620

- Simple fix for `Cantera` compatibility with `Sundials 3.2`;
- Fix to make `Cantera` build against `>=Sundials-4.0`.

### DIFF
--- a/include/cantera/numerics/Integrator.h
+++ b/include/cantera/numerics/Integrator.h
@@ -175,10 +175,12 @@ public:
         warn("setMethodType");
     }
 
-    //! Set the linear iterator.
-    virtual void setIterator(IterType t) {
-        warn("setInterator");
-    }
+    #if CT_SUNDIALS_VERSION < 40
+        //! Set the linear iterator.
+        virtual void setIterator(IterType t) {
+            warn("setInterator");
+        }
+    #endif
 
     //! Set the maximum step size
     virtual void setMaxStepSize(double hmax) {
--- a/include/cantera/numerics/Integrator.h
+++ b/include/cantera/numerics/Integrator.h
@@ -175,12 +175,11 @@ public:
         warn("setMethodType");
     }
 
-    #if CT_SUNDIALS_VERSION < 40
-        //! Set the linear iterator.
-        virtual void setIterator(IterType t) {
-            warn("setInterator");
-        }
-    #endif
+    //! Set the linear iterator.
+    //! @deprecated Unused. To be removed after Cantera 2.5.
+    virtual void setIterator(IterType t) {
+        warn_deprecated("Integrator::setIterator", "To be removed after Cantera 2.5.");
+    }
 
     //! Set the maximum step size
     virtual void setMaxStepSize(double hmax) {
--- a/SConstruct
+++ b/SConstruct
@@ -1038,23 +1038,29 @@ env['has_demangle'] = conf.CheckDeclaration("boost::core::demangle",
 
 import SCons.Conftest, SCons.SConf
 context = SCons.SConf.CheckContext(conf)
-ret = SCons.Conftest.CheckLib(context,
-                              ['sundials_cvodes'],
-                              header='#include "cvodes/cvodes.h"',
-                              language='C++',
-                              call='CVodeCreate(CV_BDF, CV_NEWTON);',
-                              autoadd=False,
-                              extra_libs=env['blas_lapack_libs'])
-if ret:
+
+# Check initially for Sundials<=3.2 and then for Sundials>=4.0
+for cvode_call in ['CVodeCreate(CV_BDF, CV_NEWTON);','CVodeCreate(CV_BDF);']:
+    ret = SCons.Conftest.CheckLib(context,
+                                  ['sundials_cvodes'],
+                                  header='#include "cvodes/cvodes.h"',
+                                  language='C++',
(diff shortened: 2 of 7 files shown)


include/cantera/numerics/Integrator.h
```diff
@@ -175,10 +175,12 @@ class Integrator
...
-    //! Set the linear iterator.
-    virtual void setIterator(IterType t) {
-        warn("setInterator");
-    }
+    #if CT_SUNDIALS_VERSION < 40
+        //! Set the linear iterator.
+        virtual void setIterator(IterType t) {
+            warn("setInterator");
+        }
+    #endif
```
speth: Given that we never use any method besides Newton iteration, my suggestion would be to:
* deprecate `setIterator` and remove the `m_iter` variable
* remove the existing calls to `setIterator`, and
* directly call `CVodeCreate` with the `CV_NEWTON` constant if we're using an older version of Sundials.

speth (changes requested):
Thanks for looking into the required changes to work with the newer SUNDIALS releases. I think the suggested change will simplify things a bit and localize the preprocessor checks on `CT_SUNDIALS_VERSION` to the `CVodesIntegrator` class.

band-a-prend:
I checked that Cantera build and pass tests with `SUNDIALS v.5.0.0-dev.1` but refrained mentioning it's `5.0` in the SConstruct version check until `5.0` final release.

include/cantera/numerics/Integrator.h
```diff
@@ -175,11 +175,6 @@ class Integrator
         warn("setMethodType");
     }
 
-    //! Set the linear iterator.
-    virtual void setIterator(IterType t) {
-        warn("setInterator");
-    }
-
```
speth: While it's not likely that this function is used directly by any external users, I don't think the `setIterator` function should be removed without first being deprecated for one release. That is, for now it should have an `@deprecated` message added to the docstring, and a call to `warn_deprecated` in the function body (for all implementations of the function). See for example the [`absmax`](https://github.com/Cantera/cantera/blob/master/include/cantera/base/utilities.h#L178) function, which will be removed after Cantera 2.5.

speth (changes requested):

include/cantera/numerics/Integrator.h
band-a-prend: Am I correct in assuming that 
https://github.com/Cantera/cantera/blob/26e47e59fe36465cf3fbc988077650efd8f46804/include/cantera/numerics/CVodesIntegrator.h#L54-L56
still should be removed on deprecation?

bryanwweber:
@band-a-prend can you remove the "cantera" from the beginning of the commit message? If you want, you can classify the commit with `[Doc]`, `[Thermo]`, `[Tests]`, etc.

band-a-prend:
@bryanwweber 
I removed it. It really was redundant here.

speth (approved):

include/cantera/numerics/Integrator.h
speth: Yes, that's correct -- it is safe to remove the implementation from the derived class in this case.

PR #677, opened 2019-08-06
Add set_equivalence_ratio to SolutionArray objects

### LINKED ISSUE #604: Add `set_equivalence_ratio` to `SolutionArray` objects
Currently, the Python `SolutionArray` class has no equivalent of `ThermoPhase.set_equivalence_ratio`. A version of this function could be added to the `SolutionArray` class so that one can set the equivalence ratio to a range of values. This would make calculation of an adiabatic flame temperature curve very simple:
```py
states = ct.SolutionArray(gas, 100)
phi = np.arange(0, 2, 100)
states.set_equivalence_ratio(phi, 'CH4:1.0', 'O2:1.0, N2:3.76')
states.equilibrate('HP')
plt.plot(phi, states.T)
```

This function could also enable specifying a range of fuel and/or oxidizer compositions, in the case where a `dict` is used for the composition:
```py
states.set_equivalence_ratio(phi, {'CH4': np.linspace(0, 1, 100), 'H2': np.linspace(1, 0, 100)})
```

ischoegl:

Fixes #604

- Wrap `ThermoPhase.set_equivalence_ratio`, and check dimensions
- Add unit test

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -565,6 +565,23 @@ class SolutionArray:
             self._phase.equilibrate(*args, **kwargs)
             self._states[index][:] = self._phase.state
 
+    def set_equivalence_ratio(self, phi, *args, **kwargs):
+        """
+        See `ThermoPhase.set_equivalence_ratio`
+
+        Note that *phi* either needs to be a scalar value or dimensions have
+        to be matched to the SolutionArray.
+        """
+
+        # broadcast argument shape
+        phi, _ = np.broadcast_arrays(phi, self._output_dummy)
+
+        # loop over values
+        for index in self._indices:
+            self._phase.state = self._states[index]
+            self._phase.set_equivalence_ratio(phi[index], *args, **kwargs)
+            self._states[index][:] = self._phase.state
+
     def collect_data(self, cols=('extra','T','density','Y'), threshold=0,
                      species='Y'):
         """
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -1474,6 +1474,25 @@ class TestSolutionArray(utilities.CanteraTest):
         states.TP = np.linspace(400, 500, 5), 101325
         self.assertArrayNear(states.X.squeeze(), np.ones(5))
 
+    def test_set_equivalence_ratio(self):
+        states = ct.SolutionArray(self.gas, 8)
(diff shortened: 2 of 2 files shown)


interfaces/cython/cantera/composite.py
```diff
@@ -565,6 +566,30 @@ def equilibrate(self, *args, **kwargs):
...
+
+        Note that *phi* either needs to be a scalar value or dimensions have
+        to be matched to the SolutionArray.
+        """
+
+        # test arguments
+        if isinstance(phi, _numbers.Real):
+            phi = phi * np.ones(self._shape)
+        elif isinstance(phi, np.ndarray):
+            if phi.shape!=self._shape:
```
```suggestion
            if phi.shape != self._shape:
```

interfaces/cython/cantera/composite.py
```diff
@@ -565,6 +566,30 @@ def equilibrate(self, *args, **kwargs):
...
+                raise ValueError('Shape mismatch.')
+        else:
+            if len(phi)!=len(self._indices):
```
```suggestion
            if len(phi) != len(self._indices):
```

interfaces/cython/cantera/composite.py
```diff
@@ -565,6 +566,30 @@ def equilibrate(self, *args, **kwargs):
...
+                raise ValueError('Incompatible dimensions.')
```
speth: The alternative to this block is to use the same approach as is used in the state setting functions:
```python
    phi, _ = np.broadcast_arrays(phi, self._output_dummy)
```
This also allows things like broadcasting a `phi` array to a higher-dimensional `SolutionArray`, e.g. from 1xN to MxN.

speth (changes requested):
I had an alternative suggestion for mapping the input equivalence ratio to an array of the appropriate shape.

interfaces/cython/cantera/composite.py
ischoegl: NP. I was looking for something existing but obviously missed that one ...

interfaces/cython/cantera/composite.py
ischoegl: N/A.

interfaces/cython/cantera/composite.py
ischoegl: N/A.

ischoegl:
Done and pushed. Let's see whether unit tests do their job ...

ischoegl:
unit tests work ;)

speth (approved):

PR #680, opened 2019-08-06
Add write_hdf to SolutionArray objects

### LINKED ISSUE #679: [Enhancement] Add HDF output for SolutionArray objects
### Cantera version

2.5.0a3

### Operating System

All

### Python/MATLAB version

Python 3

### Expected Behavior

The Python-exclusive `SolutionArray` object should include options for efficient storage. HDF is widely used, offers compression, and allows for storage of multiple data sets in a single file.

### Actual Behavior

Currently, only output to CSV is provided (`cantera.SolutionArray.write_csv`), which is not space-efficient.

### Steps to reproduce
1. N/A

ischoegl:

Fixes #679 

- The commit implements saving of data extracted from SolutionArrays to HDF containers using pandas infrastructure.
- The method only works if the pandas module can be imported; an exception is only raised if the method is called without a working pandas installation.

A simple example for the intended use would be:
```python
# create SolutionArray and/or run some manipulations
states = ct.SolutionArray(gas, 10)
...

states.write_hdf('some_file.h5')
```


bryanwweber:
Does pandas require a particular HDF5 package? I don't think that's a default dependency of Pandas, but I could be wrong

ischoegl:
@bryanwweber ... it uses PyTables as backend. I keep a pretty sparse build tool chain, and apt-get install python3-pandas took care of everything. Haven't checked pip yet, but can look into this.

ischoegl:
PS: if this has legs, I am still planning on implementing unit tests, but wanted to get some feedback before getting into the CI settings.

interfaces/cython/cantera/composite.py
```diff
@@ -637,22 +647,68 @@ def collect_data(self, cols=('extra','T','density','Y'), threshold=0,
 
         return np.hstack(data), labels
 
-    def write_csv(self, filename, cols=('extra','T','density','Y'),
+    def write_csv(self, file_name, cols=('extra','T','density','Y'),
```
bryanwweber: I think this should remain `filename` to avoid problems with anyone who might be using the keyword arguments here. Filename can be considered one word 😄 Same applies to `write_hdf()`.

interfaces/cython/cantera/composite.py
```diff
@@ -637,22 +647,68 @@ def collect_data(self, cols=('extra','T','density','Y'), threshold=0,
...
+        """
+
+        # local import avoids explicit dependence of cantera on pandas
+        import pandas as pd
+
+        data, labels = self.collect_data(cols, *args, **kwargs)
+        return pd.DataFrame(data=data, columns=labels)
+
+    def write_hdf(self, file_name, cols=('extra','T','density','Y'),
+                  key='df', mode='a', append=False, complevel=5, *args, **kwargs):
```
bryanwweber: How did you pick this `complevel`? Why use the `a` file mode and not `w`? Can you document what happens to the `*args` and `**kwargs` arguments? Can you add spaces after the commas in the default argument for `cols`?

bryanwweber (changes requested):
I'm definitely in favor of this change, just some minor comments. I also don't think there's much to test here, since all of the functionality is either tested elsewhere (`collect_states()`) or part of the Pandas package, which we don't really need to test.

I'm also interested in how a `pip`-installed or `conda`-installed Pandas manages the HDF5 dependency.

interfaces/cython/cantera/composite.py
ischoegl: NP. had pep8 in mind and it wasn't a keyword. But will change as it follows precedent.

bryanwweber: A user can still use the argument as a keyword argument, even if there's no default value provided, i.e., this will work:

```python
states.write_csv(filename='states.csv')
```

The argument is mandatory, that's all.

interfaces/cython/cantera/composite.py
ischoegl: OK. Spaces aren't a problem: I am refraining from autopep8'ing the file, but will add the spaces wherever I can find them
* Regarding default values: I was trying to keep things in line with the pandas defaults. It may be cleaner to set them all to None, and construct a kwarg dictionary so it always defaults to pandas, no matter what they do. As an aside, the mid-range `complevel` was a decent compromise between compression and speed.
* `*args` and `**kwargs` are documented with `collect_data`, and was following the approach taken for `write_csv` ...

interfaces/cython/cantera/composite.py
ischoegl: ah. for some reason I have never used that. Thanks for pointing that out!

interfaces/cython/cantera/composite.py
ischoegl: PS: the append `a` mode is a good choice as you can open existing files and add new data (with a different key) at a later point without overwriting what's already there.

bryanwweber: re the spaces: This is a difference that @speth and I have for function arguments, I prefer to be more strict about following PEP8 where he's a little more flexible. I would say that code that you're adding/modifying, add them where it makes sense, but I wouldn't bother in other code. Where it makes sense = some combination of avoiding weird line breaks and following PEP8

re defaults: I think it would be better to default to the Pandas defaults with `None` values.

re `*args`: I meant document as being passed through to `collect_data()`, if that's what's happening to them (they could also be passed through to `pandas.write_hdf()`).

bryanwweber:
I forgot to mention in the review, there's a few places where `pandas` should be uppercase Pandas in the docstring.

interfaces/cython/cantera/composite.py
speth: I know it's not strictly PEP8-compliant, but my rationale for this specific case (literal tuple as an argument) is that I think it makes it easier to see the argument and its default value as a single unit (which I think is the point of the PEP8 recommendation of not having spaces around the `=` here), while adding the spaces inside the tuple makes it easier to misread elements of the tuple as being individual arguments to the function.

ischoegl: @speth and @bryanwweber ... I see both points, but they're conflicting. I do believe there is enough visual clue from `cols=(`, so I'm leaning towards PEP8. Will run with this until I hear otherwise.

ischoegl:
Alright. pip dependencies require `pandas` and `tables` (i.e. Pandas and PyTables). Docstring is updated accordingly.

ischoegl:
Ugh. Last push was autopep'd. Can revert if so desired.

speth:
Please do. It makes the intentional changes hard to see in the diff.

ischoegl:
> Please do. It makes the intentional changes hard to see in the diff.

done. thankfully the file was mostly compliant with autopep8's choices ;) ... unintentionally opened emacs from within a conda environment with an autopep8-on-save hook enabled.

interfaces/cython/cantera/composite.py
```diff
@@ -653,6 +664,60 @@ def write_csv(self, filename, cols=('extra','T','density','Y'),
             for row in data:
                 writer.writerow(row)
 
+    def to_pandas(self, cols=('extra', 'T', 'density', 'Y'),
+                  *args, **kwargs):
+        """
+        Returns the data specified by *cols* in a single Pandas DataFrame.
+
+        Additional arguments are passed on to `collect_data`. This method works
+        only with 1D SolutionArray objects and requires a working Pandas
```
speth: Reading the *pandas* documentation, it appears that the name is usually not capitalized.

interfaces/cython/cantera/composite.py
```diff
@@ -653,6 +664,60 @@ def write_csv(self, filename, cols=('extra','T','density','Y'),
...
+
+        Arguments *key*, *mode*, *append*, and *complevel* are mapped to
+        parameters for `pandas.DataFrame.to_hdf`; the choice `None` for *mode*,
+        *append*, and *complevel* results in default values set by Pandas.
+
+        Additional arguments (i.e. *args* and *kwargs*) are passed on to
+        `collect_data` via `to_pandas`; see `collect_data` for further
+        information. This method works only with 1D SolutionArray objects
+        and requires working installations of Pandas and PyTables. Use
+        pip or conda to install pandas and tables to enable this method.
```
```suggestion
        pip or conda to install pandas and pytables to enable this method.
```

interfaces/cython/cantera/composite.py
```diff
@@ -653,6 +664,60 @@ def write_csv(self, filename, cols=('extra','T','density','Y'),
...
+        the `to_pandas` method.
+
+        :param filename: name of the HDF container file; typical file extensions
+            are `.hdf`, `.hdf5` or `.h5`.
+        :param cols: A list of any properties of the solution being exported.
+        :param key: Identifier for the group in the container file.
+        :param mode: Mode to open the file {None,'a','w','r+}.
+        :param append: use less efficient structure that makes HDF entries
+            appendable or append to existing appendable HDF entry {None,True,False}.
+        :param complevel: Specifies a compression level for data {None,0-9}.
```
speth: If you really want to have spaces following commas, that should probably apply to the docstrings as well.

speth (changes requested):
I'm in favor of this addition. My only suggested changes relate to the docstrings for the new methods.

interfaces/cython/cantera/composite.py
ischoegl: pip installs PyTables as `pip install tables`, so the naming is not consistent, see [PyPI](https://pypi.org/project/tables/). Will leave this instance as is.

I figure this edit should clarify this:
```
        Additional arguments (i.e. *args* and *kwargs*) are passed on to
        `collect_data` via `to_pandas`; see `collect_data` for further
        information. This method works only with 1D SolutionArray objects
        and requires working installations of pandas and PyTables. Use
        pip or conda to install `pandas` and `tables` to enable this method.
```

interfaces/cython/cantera/composite.py
bryanwweber: Oh! My mistake then!

interfaces/cython/cantera/composite.py
speth: But conda calls it `pytables`, so perhaps the last sentence of this should be "These packages can be installed using pip (`pandas` and `tables`) or conda (`pandas` and `pytables`)."

ischoegl: No problem. I had originally cut & paste this from an error message raised by pandas, but your suggestion is much cleaner.

ischoegl:
@speth ... I just noticed that two issues you wanted to have addressed didn't make it into the last commit (I had taken care of it but the commit was lost). It is now fixed.

speth (approved):

ischoegl:
@bryanwweber ... if this looks good to you, could you approve so I can add HDF support to #687?

bryanwweber (approved):

PR #681, opened 2019-08-06
Update ic_engine example

### LINKED ISSUE #665: Update ic_engine.py example
### Cantera version

2.4 / 2.5a3

### Operating System

all

### Python/MATLAB version

python

### Expected Behavior

The example should run simulations in an efficient manner (i.e. use `Func1` routines for valve train and injection timing)

### Actual Behavior

Implementing valve train and injection in the main simulation loop (between sim.advance steps) requires a time step that is set much smaller than necessary, making the example highly inefficient.

### Steps to reproduce
1. run stock example `ic_engine.py`

ischoegl:

Fixes #665 

- Eliminate if-else blocks that switch valves and replace by timed valves
- Use limited advance steps for adaptive refinement 
- Logic for refinement strategy is no longer necessary
- Streamline use of solution array
- Run autopep8

Excluding plots, simulation time decreases from 16.5s to 5.4s. Results are unaffected.


ischoegl:
Something went awry when force-pushing a different branch. Continued as #682 ...

PR #682, opened 2019-08-08
Update ic_engine example using new cantera capabilities

### LINKED ISSUE #665: Update ic_engine.py example

ischoegl:

Fixes #665

- Eliminate if-else blocks that switch valves and replace by timed valves
- Use limited advance steps for adaptive refinement
- Logic for refinement strategy is no longer necessary
- Streamline use of solution array

Excluding plots, simulation time decreases from 16.5s to 5.4s. Results are unaffected.

### DIFF
--- a/interfaces/cython/cantera/examples/reactors/ic_engine.py
+++ b/interfaces/cython/cantera/examples/reactors/ic_engine.py
@@ -2,27 +2,36 @@
 """
 Simulation of a (gaseous) Diesel-type internal combustion engine.
 
-The use of pure propane as fuel requires an unrealistically high compression
-ratio.
-
+The simulation uses n-Dodecane as fuel, which is injected close to top dead
+center. Note that this example uses numerous simplifying assumptions and
+thus serves for illustration purposes only.
 """
 
 import cantera as ct
 import numpy as np
 
-#######################################################################
+from scipy.integrate import trapz
+import matplotlib.pyplot as plt
+
+#########################################################################
 # Input Parameters
-#######################################################################
+#########################################################################
+
+# reaction mechanism, kinetics type and compositions
+reaction_mechanism = 'nDodecane_Reitz.xml'
+phase_name = 'nDodecane_IG'
+comp_air = 'o2:1, n2:3.76'
+comp_fuel = 'c12h26:1'
 
 f = 3000. / 60.  # engine speed [1/s] (3000 rpm)
 V_H = .5e-3  # displaced volume [m**3]
-epsilon = 50.  # compression ratio [-]
+epsilon = 20.  # compression ratio [-]
 d_piston = 0.083  # piston diameter [m]
 
 # turbocharger temperature, pressure, and composition
 T_inlet = 300.  # K
 p_inlet = 1.3e5  # Pa
-comp_inlet = 'O2:1, N2:3.76'
+comp_inlet = comp_air
 
 # outlet pressure
 p_outlet = 1.2e5  # Pa
@@ -30,15 +39,12 @@ p_outlet = 1.2e5  # Pa
 # fuel properties (gaseous!)
 T_injector = 300.  # K
 p_injector = 1600e5  # Pa
-comp_injector = 'C3H8:1'
+comp_injector = comp_fuel
 
 # ambient properties
 T_ambient = 300.  # K
 p_ambient = 1e5  # Pa
-comp_ambient = 'O2:1, N2:3.76'
-
-# Reaction mechanism name
-reaction_mechanism = 'gri30.xml'
+comp_ambient = comp_air
 
 # Inlet valve friction coefficient, open and close timings
 inlet_valve_coeff = 1.e-6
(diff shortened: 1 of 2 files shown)


interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -54,200 +56,203 @@
...
+
+t = 1.e-3 * states.t
+
+# #####################################################################
 # Plot Results in matplotlib
-#####################################################################
+# #####################################################################
+
+
+def ca_ticks(t): return np.round(crank_angle(t) * 180 / np.pi, decimals=1)
```
bryanwweber: Can you add a docstring for this function?

interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -54,200 +56,203 @@
...
+                  mdot_out=outlet_valve.mdot(sim.time),
+                  dWv_dt=dWv_dt,
+                  heat_release_rate=heat_release_rate)
+
+t = 1.e-3 * states.t
+
+# #####################################################################
 # Plot Results in matplotlib
-#####################################################################
+# #####################################################################
```
bryanwweber: The spaces like `# #` seem weird to me. I presume this is an autopep8 thing, but since these are intended to be separators, I don't think the space makes sense.

interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -54,200 +56,203 @@
...
-inlet_valve = ct.Valve(inlet, r)
-injector_mfc = ct.MassFlowController(injector, r)
-outlet_valve = ct.Valve(r, outlet)
-piston = ct.Wall(ambient_air, r)
+# piston is modeled as a moving wall with sinusoidal velocity profile
+piston = ct.Wall(ambient_air, cyl)
+piston.area = A_piston
+
+
+def piston_speed(t): return - stroke / 2 * 2 * \
```
bryanwweber: If this has to be on two lines anyways, why not put the whole statement on one line? Also, can you add a docstring?

interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -54,200 +56,203 @@
...
-#####################################################################
+# ####################################################################
 
-from scipy.integrate import trapz
-Q = trapz(heat_release_rate, t)
-W = trapz(d_W_v_d_t, t)
+
+# heat release
+Q = trapz(states.heat_release_rate, t)
+print('{:45s}{:>4.1f} kW'.format('Heat release rate per cylinder (estimate):',
```
bryanwweber, line 244: I think you used the `{:45s}` to space these strings all the same, but it looks weird to me to have a literal string as an argument to `format()` 🤷‍♂

interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -54,200 +56,203 @@
...
-        np.sum(gas.standard_enthalpies_RT * r.thermo.net_production_rates, 0)
-
-
-#####################################################################
+states = ct.SolutionArray(cyl.thermo,
+                          extra=('t', 'ca', 'V', 'm',
+                                 'mdot_in', 'mdot_out',
+                                 'dWv_dt', 'heat_release_rate'))
+
+# simulate with a minimum resolution of 1 deg crank angle
```
bryanwweber: I think this should be a maximum resolution, since if `delta_T_max` is exceeded, the integration will stop and a point will be recorded?

interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -54,200 +56,203 @@
...
+
+    # calculate results to be stored
+    dWv_dt = - (cyl.thermo.P - ambient_air.thermo.P) * A_piston * \
+        piston_speed(sim.time)
+    heat_release_rate = - cyl.volume * ct.gas_constant * cyl.T * \
+        np.sum(gas.standard_enthalpies_RT * cyl.thermo.net_production_rates, 0)
+
+    # append output data
+    states.append(cyl.thermo.state,
+                  t=sim.time*1e3, ca=crank_angle(sim.time),
```
bryanwweber: The time in ms as set here doesn't appear to be used anywhere, so you could consider removing this multiplication

interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -54,200 +56,203 @@
...
+fig, ax = plt.subplots(nrows=2)
+ax[0].plot(t, states.P / 1.e5)
+ax[0].set_ylabel('$p$ [bar]')
+ax[0].set_xlabel('$\phi$ [deg]')
+ax[0].set_xticklabels([])
+ax[1].plot(t, states.T)
+ax[1].set_ylabel('$T$ [K]')
+ax[1].set_xlabel('$\phi$ [deg]')
+ax[1].set_xticklabels(ca_ticks(ax[1].get_xticks()))
+fig.savefig('ic_engine_t_p_T.png')
```
bryanwweber: I know that this was the previous behavior of the script, but I'm not a fan of polluting the working directory when the example is run. For one, it prevents a clean run in, say, the CI service. I don't have a strong opinion that this line could be removed, but let's say its a mild preference 😸

interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -1,18 +1,20 @@
+#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
```
bryanwweber: Since Cantera 2.5.0 and up will only support Python 3, you can remove this line.

bryanwweber (changes requested):
Thanks @ischoegl! This looks good to me overall.

Similar to your other PR, we need some way to mark that this will work for the development version only. I think the best solution for the examples is for the website repo to check out the `2.4` branch of the Cantera repository during the CI build process, to make sure we get the examples that will be compatible with 2.4. Do you feel comfortable with that change, or do you want me to do it? I should be just one line.

bryanwweber:
One other comment, again forgot to add it in the review 🙄 I'm sure you know that gri30 is not intended to be used for propane chemistry. I think it's used in this example because it's distributed with Cantera and is relatively small (=fast), but I don't want to encourage people to use gri30 for problems with propane. Can you try this out with the Reitz dodecane mechanism or another mechanism in `data/inputs`?

interfaces/cython/cantera/examples/reactors/ic_engine.py
ischoegl: I share that sentiment, but didn't want to deviate too much from the original.

interfaces/cython/cantera/examples/reactors/ic_engine.py
ischoegl: That was indeed the rationale (replaced a bunch of `\t`'s). Better idea?

ischoegl:
> Thanks @ischoegl! This looks good to me overall.
> 
> Similar to your other PR, we need some way to mark that this will work for the development version only. I think the best solution for the examples is for the website repo to check out the `2.4` branch of the Cantera repository during the CI build process, to make sure we get the examples that will be compatible with 2.4. Do you feel comfortable with that change, or do you want me to do it? I should be just one line.

Not sure how CI differentiates (I had always assumed that 2.4 didn't pull from master). I'd be interested, so could you provide some pointers?

> One other comment, again forgot to add it in the review roll_eyes I'm sure you know that gri30 is not intended to be used for propane chemistry. I think it's used in this example because it's distributed with Cantera and is relatively small (=fast), but I don't want to encourage people to use gri30 for problems with propane. Can you try this out with the Reitz dodecane mechanism or another mechanism in `data/inputs`?

Yes, I am keenly aware of that, but didn't want to change without prior discussion. Likely, the compression ratio will have to change; at least the simulations are more efficient now. It won't take long to test, and will push a revision once I see some decent results (hopefully we'll get a more reasonable efficiency also).

bryanwweber:
In the Travis config for the website, it clones this repo. All you have to do is add a `git checkout 2.4` right below the clone. Let me know if any jargon needs defining 😊

ischoegl:
> In the Travis config for the website, it clones this repo. All you have to do is add a `git checkout 2.4` right below the clone. Let me know if any jargon needs defining blush

Not familiar with it, but it's probably obvious. Just haven't poked around there yet.

ischoegl:
As an aside, there appear a bunch of `nan`'s in the entropy calculations (as evidenced by the intermittent lines in the T-s plot). The root cause clearly goes beyond this PR, but I'm curious whether this is a known issue? (there *are* some negative concentrations)

ischoegl:
> > In the Travis config for the website, it clones this repo. All you have to do is add a `git checkout 2.4` right below the clone. Let me know if any jargon needs defining blush
> 
> Not familiar with it, but it's probably obvious. Just haven't poked around there yet.

@bryanwweber  ... in `.travis.yml`:
```
before_install:
  - git clone --depth=1 https://github.com/Cantera/cantera ../cantera  --branch 2.4
  - git clone --depth=1 https://github.com/Cantera/cantera-jupyter ../cantera-jupyter
```
I assume website updates involve PR's from forked repos?

bryanwweber:
@ischoegl No, the website will only be updated when new changes are merged to the `master` branch of the main fork. Also, it looks like the `2.4` branch is already checked out, so there shouldn't be anything that you need to do!

I'm not sure about the `nan` calculations here, could the problem be the mechanism?

ischoegl:
@bryanwweber ... ok, I did look into Reitz, but realized that the mechanism is not compatible with `IdealGasReactor`. Switching to `Reactor`, I got things to run after cranking up tolerances (`rtol=1.e-12, atol=1.e-16`), but suffered erratic core dumps when exiting out of ipython (`corrupted double-linked list` or `corrupted size vs. prev_size` that I believe stem from the core ... i.e. this goes beyond this PR).

As there aren't any good alternatives in terms of gas mechanisms, I opted for a more detailed disclaimer in the docstring:
```
"""
Simulation of a (gaseous) Diesel-type internal combustion engine.

Note that this example serves for illustration purposes only. While the
approximation of an IC engine as a simple reactor network has inherent
limitations by itself, additional considerations are:
1. The reaction mechanism GRI3.0 was never intended for use with pure propane
   (it was developed for natural gas mixtures), but is used here as it allows 
   for fast simulations, i.e. predictive capabilities are limited.
2. Simulations require an unrealistically high compression ratio, i.e.
   resulting efficiency estimates are also not realistic.
"""
```

ischoegl:
> @ischoegl No, the website will only be updated when new changes are merged to the `master` branch of the main fork. Also, it looks like the `2.4` branch is already checked out, so there shouldn't be anything that you need to do!

Thanks. FYI: it was me who added the `--branch 2.4` in my previous post. I guess it needs to go into `.travis.yml` right there?

bryanwweber:
Yes, that should do the trick. You might also consider adding the `--single-branch` switch to help speed up the clone a little more.

ischoegl:
> Yes, that should do the trick. You might also consider adding the `--single-branch` switch to help speed up the clone a little more.

PR is already issued (I'll let CI run through before `--single-branch`). See Cantera/cantera-website#86

speth:
A couple of comments from just reading the discussion:
1. You can use the Reitz mechanism as an ideal gas phase - the instantiation would just be
`gas = ct.Solution('nDodecane_Reitz.cti', 'nDodecane_IG')`, corresponding to the second phase declaration in the input file.
2. You will get `NaN`s for entropy any time you have negative mass/mole fractions. The main time that this is encountered is when you're directly using the `Solution` object in the state that is left in by CVODES (which must be allowed to contain negative numbers), rather than the "corrected" state that you will get if you access it as `reactor.thermo`.

ischoegl:
@speth ... thank you for the suggestions!
>
>   1.  You can use the Reitz mechanism as an ideal gas phase - the instantiation would just be
>        `gas = ct.Solution('nDodecane_Reitz.cti', 'nDodecane_IG')`, corresponding to the second phase declaration in the input file.
 
This works nicely! Also, no more core dumps ... 

>  
> 2. You will get `NaN`s for entropy any time you have negative mass/mole fractions. The main time that this is encountered is when you're directly using the `Solution` object in the state that is left in by CVODES (which must be allowed to contain negative numbers), rather than the "corrected" state that you will get if you access it as `reactor.thermo`.

Thanks for this hint ... however, the script *does* grab from `Reactor.thermo.state` when saving `states`. A quick check for the `SolutionArray` at the end of the simulation yields the following (which imho is not consistent).
```
In [6]: np.isnan(states.s).any()
Out[6]: True

In [7]: np.isnan(states.h).any()
Out[7]: False
```

speth:
Sorry, my mistake. The state setting done implicitly by `Reactor.thermo` does not clean up the mass fraction array, and using the `state` attribute will also continue to use the methods for setting the "unnormalized" mass fractions.

I'm not sure why it's inconsistent for the entropy to end up as `NaN` while the the enthalpy does not -- the equation for the entropy requires (effectively) taking `log(Y_k)`, which is undefined for negative numbers while the equation for the enthalpy has no similar problems.

However, I did notice that we have a workaround for this problem implemented in other methods, such as `IdealGasPhase::getPartialMolarEntropies`, and I don't think there's any reason not to do the same for `enthalpy_mole` (and implicitly, `enthalpy_mass`), which I've created a PR for now (#686).

interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -2,43 +2,49 @@
...
+import matplotlib.pyplot as plt
+
+#########################################################################
 # Input Parameters
-#######################################################################
+#########################################################################
+
+# reaction mechanism, kinetics type and compositions
+reaction_mechanism = 'nDodecane_Reitz.xml'
+kinetics_type = 'nDodecane_IG'
```
speth: I would suggest calling this the `phase_name` -- this is the name of the phase definition which includes the thermodynamic, kinetics and transport models. In this case, both of the phase definitions in the input file use the gas phase kinetics model -- it's the thermodynamic model where they differ.

interfaces/cython/cantera/examples/reactors/ic_engine.py
speth: Anything is better than tabs in my book.

interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -54,200 +60,203 @@
...
-injector_t_open = (injector_close - injector_open) / 2. / np.pi / f
 
-# Simulation time and resolution
-sim_n_revolutions = 8.
-sim_n_timesteps = 100000.
+# Simulation time and parameters
+sim_n_revolutions = 8
+delta_T_max = 20.
+rtol = 1.e-12
+atol = 1.e-16
```
speth, line 68: Are these more stringent tolerances necessary/helpful?

speth (changes requested):
Generally looks good to me. Here are a couple of minor suggestions.

interfaces/cython/cantera/examples/reactors/ic_engine.py
ischoegl: wouldn't solve otherwise. I didn't investigate what the minimum requirements were, but the default settings didn't work. I.e. they are *necessary*

interfaces/cython/cantera/examples/reactors/ic_engine.py
ischoegl: NP.

speth (approved):

bryanwweber (approved):

PR #686, opened 2019-08-09
[Thermo] Avoid NaN in entropy with small negative mass fractions

speth:
Avoid NaN results in entropy_mole calculations when there are small negative
mass fractions. Consistent with the approach used elsewhere,
e.g. IdealGasPhase::getPartialMolarEntropies.

This is a result of investigating the issue brought up by @ischoegl in #682.


ischoegl:
I’d give this one a 👍 , with the specific reason that it should always be possible to set a state based on `S`. Currently, a valid state can generate an invalid entropy value, I.e. the state cannot be reconstructed via 'SP' or 'SV'.

Issue #644, opened 2019-06-11
Setting of density in `LatticePhase`


decaluwe:
### Cantera version 
2.4.0

### Operating System: 
Any

### Python/MATLAB version: 
Any

### Expected Behavior:  
The `LatticePhase` implements a `thermoPhase` representing a solid solution with a fixed/specified molar density (kmol/m3).  In theory, the mass density is therefore not an independent variable - it should adjust, as a function of composition, to maintain this set molar density.

### Actual Behavior: 
In reality, the density may be set by a user via any number of means (either as part of a property pair (e.g. 'TD'), or as part of a property pair + composition (e.g., 'TDY').  In either instance, the user-supplied molar density (`site_density` in the cti file) is explicitly changed/overwritten.

### Steps to reproduce 
(using Python syntax)
1.  ```ysz = ct.Solution('sofc.cti','oxide_bulk')```
2.  ```print(ysz.density_mole)```
3.  ```ysz.TD = 1073, 200```
4.  ```print(ysz.density_mole)```

There is some question, however, as to whether this is really a problem.  A user who sets the density for this phase is explicitly violating the assumptions of the model they have chosen to implement, and it is not apparent that it is the software's job to prevent them from this accomplishment.  Moreover, it is conceivable that a user would intentionally perform this operation, e.g. a numerical study where they vary the lattice's molar density.

The purpose of this issue, therefore, is to document the concern, so as to i) solicit input from users, and ii) encourage follow-up at a later date.

Issue #595, opened 2019-02-07
save/restore ThermoPhase state assumes density is an independent variable


speth:
The functions `Phase::saveState` and `Phase::restoreState` assume that temperature and density are the independent thermodynamic variables, which is not true for several of Cantera's thermo models. For these models, the save/restore feature does not work correctly. For example (in Python):
```python
import cantera as ct
phase = ct.ThermoPhase('test_problems/cathermo/issp/IdealSolidSolnPhaseExample.xml')
T0, p0 = phase.TP # T0 = 500, p0 == 101325
z = phase.state # calls Phase::saveState
phase.TP = 600, 2*ct.one_atm
phase.state = z # calls Phase::restoreState
assert phase.T == T0, 'temperature'
assert phase.P == p0, 'pressure'
```
Here, the pressure is left at 2 atm, despite the call to `restoreState`. In other cases, the `restoreState` call will fail because the derived class overrides the `setDensity` function to throw an exception.

Among other things, this error contributes to making these phases difficult to use with the `Solution` class in Python and Matlab, where the default behavior of instantiating a dummy `Transport` object triggers calls to `saveState` and `restoreState`, with the latter failing.

This problem can be resolved by making the `saveState` and `restoreState` methods virtual and having them work with the appropriate set of independent variables for each phase type.

ThanasisMattas:
Hello! I would like to have a look on this issue. I think this could be handled parallel to this @todo entry commented in Phase.h:

  @todo
    - Make the concept of saving state vectors more general, so that it can
      handle other cases where there are additional internal state variables,
      such as the voltage, a potential energy, or a strain field.

 
So, possible overrides at derived classes will use their access methods for every other state variable, like voltage(). So, is just a virtual keyword enough here? And both methods can be overridden at the derived class. More specificly, I would like to ask if this requires any re-implementation of the overridden methods so that they handle the proper state variables (and, corresponding to the todo entry, to add any state variable to the state vector).

PR #308, opened 2015-11-12
[Python] Fix full-state (e.g. TPX) properties of PureFluid

speth:
The implementations which come from ThermoPhase expect "X" to mean mole
fractions, but for PureFluid "X" means vapor fraction. No setter is provided
since all three properties cannot be specified simultaneously.

Resolves #307

Issue #307, opened 2015-11-10
Setting TPX of a PureFluid results in an error (but should probably be undefined)


bryanwweber:
There is no reason for a PureFluid to be able to set the temperature, pressure, and X (where X for a PureFluid is the vapor quality, not the mole fractions). Example:

```
>> import cantera as ct
>> w = ct.Water()
>> w.TX = 380, 0.5
>> w()

   water:

        temperature             380  K
           pressure          128618  Pa
            density         1.49268  kg/m^3
   mean mol. weight          18.016  amu
     vapor fraction             0.5

>> w.TPX = 373.15, 101325, 0.5
>> w.TPX = 373.15, 101325, 0.5
   ---------------------------------------------------------------------------
   ValueError                                Traceback (most recent call last)
   <ipython-input-22-8a529abadb62> in <module>()
   ----> 1 w.TPX = 373.15, 101325, 0.5

   interfaces/cython/cantera/thermo.pyx in cantera._cantera.ThermoPhase.TPX.__set__ (interfaces/cython/cantera/_cantera.cpp:20706)()

   interfaces/cython/cantera/thermo.pyx in cantera._cantera.PureFluid.X.__set__ (interfaces/cython/cantera/_cantera.cpp:26306)()

   ValueError: Cannot set vapor quality outside thetwo-phase region
```

The error comes out because the generic `TPX` function sets the `X` of the phase (thinking it should be the mole fractions) first, then tries to set the temperature and pressure. It works the first time, because the quality is set first, and then the state is overridden by the given temperature and pressure. Then the second time around it tries to set the quality, but can't because its outside the two-phase region.

Since T and P are not independent if X is specified, it should probably be undefined to set all three at the same time. This is also related to #306.

speth:
The case where it doesn't raise an exception is definitely wrong.

I don't think it's possible to undefine the methods, since `PureFluid` inherits these methods from `ThermoPhase`. They can be overridden, either to throw an exception, or just defined to be `None`.

#306 is actually somewhat different -- overloading "X" to be both quality and mole fraction is only done in the Python interface.

bryanwweber:
I think probably throwing a ValueError (or CanteraError? Can we throw those from the Python interface?) would be the best solution, something like

```
Cannot set the temperature, pressure, and vapor fraction simultaneously
```

To resolve this overloading, should we change the Python interface then, to use something like `q` for the quality?

[...]

PR #687, opened 2019-08-09
Implement recreation of SolutionArray objects from stored data

### LINKED ISSUE #543: SolutionArray save / load
Is there a way to save/load the SolutionArray.object

```
import pickle
# mal schauen ob der zwischenstand gesichert werden kann
pickle.dump( states, open( "save.p", "wb" ) )
```

doesn't work.... I want to save my Solution within the SolutionArray.object and load the data in a other python source. Yes i can save it in CSV but then the nice functions of SolutionArray.object is missing.

Is there a way to save multipy scenarios on disk an reuse them later?

ischoegl:

Fixes #543

This commit implements two new methods for SolutionArray:
* `restore_data` can restore data previously exported by `collect_data`
* `read_csv` restores data previously saved by `write_csv`
* `read_hdf` restores data saved by `write_hdf`

Intended usage is:
```python
# create a SolutionArray, do some manipulations and save results
states = ct.SolutionArray(gas)
...
states.write_csv('some_file.csv')
```
and, at a later point
```python
# restore prior work in a new python session
states = ct.SolutionArray(gas)
states.read_csv('some_file.csv')
```
~~PS: once PR #680 is merged, an equivalent method for `read_hdf` will be added to this PR.~~ Done.

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -576,6 +576,117 @@ class SolutionArray:
             self._phase.equilibrate(*args, **kwargs)
             self._states[index][:] = self._phase.state
 
+    def restore_data(self, data, labels):
+        """
+        Restores a SolutionArray based on *data* specified in a single
+        2D Numpy array and a list of corresponding column *labels*. Thus,
+        this method allows to restore data exported by `collect_data`.
+
+        :param data: a 2D Numpy array holding data to be restored.
+        :param labels: a list of labels corresponding to SolutionArray entries.
+
+        The receiving SolutionArray either has to be empty or should have
+        matching dimensions. Essential state properties and extra entries
+        are detected automatically whereas stored information of calculated
+        properties is omitted. If the receiving SolutionArray has extra
+        entries already specified, only those will be imported; if *labels* does
+        not contain those entries, an error is raised.
+        """
+
+        # check arguments
+        if not isinstance(data, np.ndarray) or data.ndim != 2:
+            raise TypeError("restore_data only works for 2D ndarrays")
+        elif len(labels) != data.shape[1]:
+            raise ValueError("inconsistent data and label dimensions")
+        rows = data.shape[0]
+        if self._shape!=(0,) and self._shape!=(rows,):
+            raise ValueError('incompatible dimensions.')
+
+        # get full state information (may differ depending on type of ThermoPhase)
+        full_states = [fs for fs in self._phase._full_states.values()]
+        if isinstance(self._phase, PureFluid):
+            # make sure that potentially non-unique state definitions are checked last
+            last = ['TP', 'TX', 'PX']
+            full_states = [fs for fs in full_states
+                           if fs not in last] + ['TPX'] + last
+
+        # determine whether complete concentration is available (mass or mole)
+        # assumes that `X` or `Y` is always in last place
+        mode = ''
+        has_species = False
+        for prefix in ['X_', 'Y_']:
+            spc = ['{}{}'.format(prefix, s) for s in self.species_names]
+            valid_species = {s[2:]: labels.index(s) for s in spc
+                             if s in labels}
+            all_species = [l for l in labels if l[:2] == prefix]
+            if len(valid_species):
+                mode = prefix[0]
+                full_states = [v[:-1] for v in full_states if mode in v]
+                break
+        if len(valid_species) != len(all_species):
+            raise ValueError('incompatible species information.')
+        if mode == '':
+            full_states = {v[:2] for v in full_states}
+
+        # determine suitable thermo properties for reconstruction
+        basis = {'molar': 'mole', 'mass': 'mass'}[self.basis]
+        prop = {'T': ('T'), 'P': ('P'),
+                'D': ('density', 'density_{}'.format(basis)),
+                'U': ('u', 'int_energy_{}'.format(basis)),
+                'V': ('v', 'volume_{}'.format(basis)),
+                'H': ('h', 'enthalpy_{}'.format(basis)),
+                'S': ('s', 'entropy_{}'.format(basis))}
+        for fs in full_states:
+            state = [{fs[i]: labels.index(p) for p in prop[fs[i]] if p in labels}
+                     for i in range(len(fs))]
+            found = [len(state[i]) for i in range(len(fs))]
+            if all(found):
+                mode = fs + mode
+                break
+        if len(mode) == 1:
+            raise ValueError('invalid/incomplete state information.')
+
+        # raise warning if state is potentially not uniquely defined
+        if isinstance(self._phase, PureFluid) and mode in last:
+            # note: adding a setter for PureFluid.TPX would would be beneficial
+            warnings.warn('Using  mode `{}` to restore data: may not '
+                          'be sufficient to define unique state '
+                          'for a PureFluid phase'.format(mode),
+                          UserWarning)
+
+        # assemble and restore state information
+        state_data = tuple([data[:, state[i][mode[i]]] for i in range(len(state))])
+        if len(valid_species):
+            state_data += (np.zeros((rows, self.n_species)),)
+            for i, s in enumerate(self.species_names):
+                if s in valid_species:
+                    state_data[-1][:, i] = data[:, valid_species[s]]
+
+        # labels may include calculated properties that must not be restored
+        calculated = self._scalar + self._n_species + self._n_reactions
+        exclude = [l for l in labels
+                   if any([v in l for v in calculated])]
+        extra = {l: list(data[:, i]) for i, l in enumerate(labels)
+                 if l not in exclude}
+        if len(self._extra_lists):
+            extra_lists = {k: extra[k] for k in self._extra_arrays}
+        else:
+            extra_lists = extra
+
+        # ensure that SolutionArray accommodates dimensions
+        if self._shape == (0,):
+            self._states = [self._phase.state] * rows
+            self._indices = range(rows)
+            self._output_dummy = self._indices
+            self._shape = (rows,)
+
+        # restore data
+        for i in self._indices:
+            setattr(self._phase, mode, [st[i, ...] for st in state_data])
+            self._states[i] = self._phase.state
+        self._extra_lists = extra_lists
+        self._extra_arrays = {l: np.array(v) for l, v in extra_lists.items()}
+
     def set_equivalence_ratio(self, phi, *args, **kwargs):
         """
         See `ThermoPhase.set_equivalence_ratio`
@@ -680,6 +791,19 @@ class SolutionArray:
             writer.writerow(labels)
             for row in data:
                 writer.writerow(row)
+
+    def read_csv(self, filename):
+        """
+        Read a CSV file named *filename* and restore data to the SolutionArray
+        using `restore_data`. This method allows for recreation of data
+        previously exported by `write_csv`.
+        """
+        # read data block and header separately
+        data = np.genfromtxt(filename, skip_header=1, delimiter=',')
+        labels = np.genfromtxt(filename,
+                               max_rows=1, delimiter=',', dtype=str)
+
+        self.restore_data(data, list(labels))
 
     def to_pandas(self, cols=('extra', 'T', 'density', 'Y'),
                   *args, **kwargs):
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -5,6 +5,12 @@ from ._cantera import *
 import numpy as np
 import csv as _csv
 
+# avoid explicit dependence of cantera on pandas
+try:
+    import pandas as _pandas
+except ImportError as err:
+    _pandas = err
+
 
 class Solution(ThermoPhase, Kinetics, Transport):
     """
@@ -367,15 +373,28 @@ class SolutionArray:
 
         >>> states.write_csv('somefile.csv', cols=('T', 'P', 'X', 'net_rates_of_progress'))
 
-    As an alternative, data extracted from SolutionArray objects can be saved
-    to a pandas compatible HDF container file using the `write_hdf` method::
+    As long as stored columns specify a valid thermodynamic state, the contents of
+    a SolutionArray can be restored using the `read_csv` method::
+
+        >>> states = ct.SolutionArray(gas)
+        >>> states.read_csv('somefile.csv')
+
+    As an alternative to comma separated export and import, data extracted from
+    SolutionArray objects can also be saved to and restored from a pandas compatible
+    HDF container file using the `write_hdf`::
 
         >>> states.write_hdf('somefile.h5', cols=('T', 'P', 'X'), key='some_key')
 
-    In this case, the (optional) key argument allows for saving and accessing
-    multiple solutions in a single container file. Note that `write_hdf` requires
-    working installations of pandas and PyTables. These packages can be installed
-    using pip (`pandas` and `tables`) or conda (`pandas` and `pytables`).
+    and `read_hdf` methods::
+
+        >>> states = ct.SolutionArray(gas)
+        >>> states.read_hdf('somefile.h5', key='some_key')
+
+    For HDF export and import, the (optional) key argument *key* allows for saving
+    and accessing of multiple solutions in a single container file. Note that
+    `write_hdf` and `read_hdf` require working installations of pandas and
+    PyTables. These packages can be installed using pip (`pandas` and `tables`)
+    or conda (`pandas` and `pytables`).
 
     :param phase: The `Solution` object used to compute the thermodynamic,
         kinetic, and transport properties
@@ -648,7 +667,6 @@ class SolutionArray:
 
         # raise warning if state is potentially not uniquely defined
         if isinstance(self._phase, PureFluid) and mode in last:
-            # note: adding a setter for PureFluid.TPX would would be beneficial
             warnings.warn('Using  mode `{}` to restore data: may not '
                           'be sufficient to define unique state '
                           'for a PureFluid phase'.format(mode),
@@ -815,11 +833,25 @@ class SolutionArray:
         installation. Use pip or conda to install `pandas` to enable this method.
         """
 
-        # local import avoids explicit dependence of cantera on pandas
-        import pandas as pd
+        if isinstance(_pandas, ImportError):
+            raise ImportError(_pandas.msg)
 
         data, labels = self.collect_data(cols, *args, **kwargs)
-        return pd.DataFrame(data=data, columns=labels)
+        return _pandas.DataFrame(data=data, columns=labels)
+
+    def from_pandas(self, df):
+        """
+        Restores SolutionArray data from a pandas DataFrame *df*.
+
+        This method is intendend for loading of data that were previously
+        exported by `to_pandas`. The method requires a working pandas
+        installation. The package 'pandas' can be installed using pip or conda.
+        """
+
+        data = df.to_numpy(dtype=float)
+        labels = [col for col in df.columns]
+
+        self.restore_data(data, list(labels))
 
     def write_hdf(self, filename, cols=('extra', 'T', 'density', 'Y'),
                   key='df', mode=None, append=None, complevel=None,
@@ -859,6 +891,24 @@ class SolutionArray:
         pd_kwargs = {'mode': mode, 'append': append, 'complevel': complevel}
         pd_kwargs = {k: v for k, v in pd_kwargs.items() if v is not None}
         df.to_hdf(filename, key, **pd_kwargs)
+
+    def read_hdf(self, filename, key=None):
+        """
+        Read a dataset identified by *key* from a HDF file named *filename*
+        and restore data to the SolutionArray object. This method allows for
+        recreation of data previously exported by `write_hdf`.
+
+        The method imports data using `restore_data` via `from_pandas` and
+        requires working installations of pandas and PyTables. These
+        packages can be installed using pip (`pandas` and `tables`) or conda
+        (`pandas` and `pytables`).
+        """
+
+        if isinstance(_pandas, ImportError):
+            raise ImportError(_pandas.msg)
+
+        pd_kwargs = {'key': key} if key else {}
+        self.from_pandas(_pandas.read_hdf(filename, **pd_kwargs))
 
 
 def _make_functions():
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -299,7 +299,7 @@ class SolutionArray:
     states using the same `Solution` object and computing properties for that
     array of states.
 
-    SolutionArray can represent both 1D and multi-dimensional arrays of states,
+    `SolutionArray` can represent both 1D and multi-dimensional arrays of states,
     with shapes described in the same way as Numpy arrays. All of the states
     can be set in a single call::
 
@@ -311,7 +311,7 @@ class SolutionArray:
         >>> states.TPX = T, P, X
 
     Similar to Numpy arrays, input with fewer non-singleton dimensions than the
-    SolutionArray is 'broadcast' to generate input of the appropriate shape. In
+    `SolutionArray` is 'broadcast' to generate input of the appropriate shape. In
     the above example, the single value for the mole fraction input is applied
     to each input, while each row has a constant temperature and each column has
     a constant pressure.
@@ -337,7 +337,7 @@ class SolutionArray:
         >>> states.equilibrate('HP')
         >>> states.T # -> adiabatic flame temperature at various equivalence ratios
 
-    SolutionArray objects can also be 'sliced' like Numpy arrays, which can be
+    `SolutionArray` objects can also be 'sliced' like Numpy arrays, which can be
     used both for accessing and setting properties::
 
         >>> states = ct.SolutionArray(gas, (6, 10))
@@ -347,7 +347,7 @@ class SolutionArray:
     If many slices or elements of a property are going to be accessed (i.e.
     within a loop), it is generally more efficient to compute the property array
     once and access this directly, rather than repeatedly slicing the
-    SolutionArray object, e.g.::
+    `SolutionArray` object, e.g.::
 
         >>> mu = states.viscosity
         >>> for i,j in np.ndindex(mu.shape):
@@ -368,19 +368,19 @@ class SolutionArray:
         >>> s.reaction_equation(10)
         'CH4 + O <=> CH3 + OH'
 
-    Data represented by a SolutionArray can be extracted and saved to a CSV file
+    Data represented by a `SolutionArray` can be extracted and saved to a CSV file
     using the `write_csv` method::
 
         >>> states.write_csv('somefile.csv', cols=('T', 'P', 'X', 'net_rates_of_progress'))
 
     As long as stored columns specify a valid thermodynamic state, the contents of
-    a SolutionArray can be restored using the `read_csv` method::
+    a `SolutionArray` can be restored using the `read_csv` method::
 
         >>> states = ct.SolutionArray(gas)
         >>> states.read_csv('somefile.csv')
 
     As an alternative to comma separated export and import, data extracted from
-    SolutionArray objects can also be saved to and restored from a pandas compatible
+    `SolutionArray` objects can also be saved to and restored from a pandas compatible
     HDF container file using the `write_hdf`::
 
         >>> states.write_hdf('somefile.h5', cols=('T', 'P', 'X'), key='some_key')
@@ -399,7 +399,7 @@ class SolutionArray:
     :param phase: The `Solution` object used to compute the thermodynamic,
         kinetic, and transport properties
     :param shape: A tuple or integer indicating the dimensions of the
-        SolutionArray. If the shape is 1D, the array may be extended using the
+        `SolutionArray`. If the shape is 1D, the array may be extended using the
         `append` method. Otherwise, the shape is fixed.
     :param states: The initial array of states. Used internally to provide
         slicing support.
@@ -597,35 +597,40 @@ class SolutionArray:
 
     def restore_data(self, data, labels):
         """
-        Restores a SolutionArray based on *data* specified in a single
+        Restores a `SolutionArray` based on *data* specified in a single
         2D Numpy array and a list of corresponding column *labels*. Thus,
         this method allows to restore data exported by `collect_data`.
 
         :param data: a 2D Numpy array holding data to be restored.
-        :param labels: a list of labels corresponding to SolutionArray entries.
+        :param labels: a list of labels corresponding to `SolutionArray` entries.
 
-        The receiving SolutionArray either has to be empty or should have
+        The receiving `SolutionArray` either has to be empty or should have
         matching dimensions. Essential state properties and extra entries
         are detected automatically whereas stored information of calculated
-        properties is omitted. If the receiving SolutionArray has extra
-        entries already specified, only those will be imported; if *labels* does
-        not contain those entries, an error is raised.
+        properties is omitted. If the receiving `SolutionArray` has extra
+        entries already specified, only those will be imported; if *labels*
+        does not contain those entries, an error is raised.
         """
 
         # check arguments
         if not isinstance(data, np.ndarray) or data.ndim != 2:
             raise TypeError("restore_data only works for 2D ndarrays")
         elif len(labels) != data.shape[1]:
-            raise ValueError("inconsistent data and label dimensions: "
+            raise ValueError("inconsistent data and label dimensions "
                              "({} vs. {})".format(len(labels), data.shape[1]))
         rows = data.shape[0]
         if self._shape != (0,) and self._shape != (rows,):
-            raise ValueError('incompatible dimensions.')
+            raise ValueError(
+                "incompatible dimensions ({} vs. {}): the receiving "
+                "SolutionArray object either needs to be empty "
+                "or have a length that matches data rows "
+                "to be restored".format(self._shape[0], rows)
+            )
 
-        # get full state information (may differ depending on type of ThermoPhase)
+        # get full state information (may differ depending on ThermoPhase type)
         full_states = self._phase._full_states.values()
         if isinstance(self._phase, PureFluid):
-            # make sure that potentially non-unique state definitions are checked last
+            # ensure that potentially non-unique state definitions are checked last
             last = ['TP', 'TX', 'PX']
             full_states = [fs for fs in full_states
                            if fs not in last] + ['TPX'] + last
@@ -633,14 +638,13 @@ class SolutionArray:
         # determine whether complete concentration is available (mass or mole)
         # assumes that `X` or `Y` is always in last place
         mode = ''
-        has_species = False
         for prefix in ['X_', 'Y_']:
             spc = ['{}{}'.format(prefix, s) for s in self.species_names]
             # solution species names also found in labels
             valid_species = {s[2:]: labels.index(s) for s in spc
                              if s in labels}
             # labels that start with prefix (indicating concentration)
-            all_species = [l for l in labels if l[:2] == prefix]
+            all_species = [l for l in labels if l.startswith(prefix)]
             if valid_species:
                 # save species mode and remaining full_state candidates
                 mode = prefix[0]
@@ -655,7 +659,7 @@ class SolutionArray:
             full_states = {v[:2] for v in full_states}
 
         # determine suitable thermo properties for reconstruction
-        basis = {'molar': 'mole', 'mass': 'mass'}[self.basis]
+        basis = 'mass' if self.basis == 'mass' else 'mole'
         prop = {'T': ('T'), 'P': ('P'),
                 'D': ('density', 'density_{}'.format(basis)),
                 'U': ('u', 'int_energy_{}'.format(basis)),
@@ -671,7 +675,10 @@ class SolutionArray:
                 mode = fs + mode
                 break
         if len(mode) == 1:
-            raise ValueError('invalid/incomplete state information.')
+            raise ValueError(
+                "invalid/incomplete state information (detected "
+                "partial information as mode='{}')".format(mode)
+            )
 
         # assemble and restore state information
         state_data = tuple([data[:, state[i][mode[i]]] for i in range(len(state))])
@@ -734,8 +741,8 @@ class SolutionArray:
             are specified, then either the mass or mole fraction of that species
             will be taken, depending on the value of *species*. *cols* may also
             include any arrays which were specified as 'extra' variables when
-            defining the SolutionArray object. The special value 'extra' can be
-            used to include all 'extra' variables.
+            defining the `SolutionArray` object. The special value 'extra' can
+            be used to include all 'extra' variables.
         :param threshold: Relative tolerance for including a particular column.
             The tolerance is applied by comparing the maximum absolute value for
             a particular column to the maximum absolute value in all columns for
@@ -802,7 +809,7 @@ class SolutionArray:
         *cols*. The first row of the CSV file will contain column labels.
 
         Additional arguments are passed on to `collect_data`. This method works
-        only with 1D SolutionArray objects.
+        only with 1D `SolutionArray` objects.
         """
         data, labels = self.collect_data(cols, *args, **kwargs)
         with open(filename, 'w') as outfile:
@@ -813,7 +820,7 @@ class SolutionArray:
 
     def read_csv(self, filename):
         """
-        Read a CSV file named *filename* and restore data to the SolutionArray
+        Read a CSV file named *filename* and restore data to the `SolutionArray`
         using `restore_data`. This method allows for recreation of data
         previously exported by `write_csv`.
         """
@@ -829,7 +836,7 @@ class SolutionArray:
         Returns the data specified by *cols* in a single pandas DataFrame.
 
         Additional arguments are passed on to `collect_data`. This method works
-        only with 1D SolutionArray objects and requires a working pandas
+        only with 1D `SolutionArray` objects and requires a working pandas
         installation. Use pip or conda to install `pandas` to enable this method.
         """
 
@@ -841,7 +848,7 @@ class SolutionArray:
 
     def from_pandas(self, df):
         """
-        Restores SolutionArray data from a pandas DataFrame *df*.
+        Restores `SolutionArray` data from a pandas DataFrame *df*.
 
         This method is intendend for loading of data that were previously
         exported by `to_pandas`. The method requires a working pandas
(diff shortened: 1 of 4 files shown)


interfaces/cython/cantera/composite.py
```diff
@@ -565,6 +565,98 @@ def equilibrate(self, *args, **kwargs):
...
+
+        # ensure that SolutionArray accommodates dimensions
+        if self._shape == (0,):
+            self._states = [self._phase.state] * rows
+            self._indices = range(rows)
+            self._output_dummy = self._indices
+            self._shape = (rows,)
+
+        # restore data
+        for i in self._indices:
```
ischoegl: Could not get `setattr(self, mode, state_data)` to work here ...

ischoegl:
Unit tests are currently missing ... will restore them once I'm back in my office.

interfaces/cython/cantera/composite.py
```diff
@@ -565,6 +565,98 @@ def equilibrate(self, *args, **kwargs):
...
+        # check arguments
+        if not isinstance(data, np.ndarray) or data.ndim != 2:
+            raise TypeError("restore_data only works for 2D ndarrays")
+        elif len(labels) != data.shape[1]:
+            raise ValueError("inconsistent data and label dimensions")
+        rows = data.shape[0]
+        if self._shape!=(0,) and self._shape!=(rows,):
+            raise ValueError('incompatible dimensions.')
+
+        # determine whether complete concentration is available (mass or mole)
```
ischoegl: This block assumes a `ThermoPhase`. Should be bypassed if the solution is a `PureFluid`. Implementation that works for both cases should pull `self._phase._full_states` instead of `self._state2`.

interfaces/cython/cantera/composite.py
```diff
@@ -565,6 +565,98 @@ def equilibrate(self, *args, **kwargs):
...
+
+        # determine suitable thermo properties for reconstruction
+        basis = {'molar': 'mole', 'mass': 'mass'}[self.basis]
+        prop = {'T': ('T'), 'P': ('P'),
+                'D': ('density', 'density_{}'.format(basis)),
+                'U': ('u', 'int_energy_{}'.format(basis)),
+                'V': ('v', 'volume_{}'.format(basis)),
+                'H': ('h', 'enthalpy_{}'.format(basis)),
+                'S': ('s', 'entropy_{}'.format(basis))}
+        for st2 in self._state2:
```
ischoegl: Should loop over first two characters of `self._phase._full_states`. `X` (quality) should be included in `prop`.

bryanwweber:
Moving this discussion here so if that commit is deleted, the discussion isn't lost.

> where a user may save T, P, X histories for a phase transition (being able to reconstruct this history is a completely legitimate assumption).

I agree this is a legitimate case.

> The proposed setter checks for inconsistent states, and picks the most suitable setter based on whether the state is or is not saturated.

I don't think the setter is correct. If, say, the quality is saved such that it's close to saturated vapor, T and p are not unique compared to the saturated liquid, so using T, p to set the state isn't unique. At least, that's how I'm reading the `elif` condition.

>  In view of existing setters for TP, TX and PX (all of which involve implicit assumptions), I feel like allowing for TPX import is a cleaner solution.

I'm not sure what assumptions you mean here. TP is only valid for single-phase states, while TX and PX are only valid for two-phase. I believe TP will raise an exception if T and p are not independent.

In any case, shouldn't it be possible to implement the check in the `SolutionArray` method? If the phase is a `PureFluid`, and `X` is given and is not `None`, use that plus another independent variable. Or, try using T, P and if it fails, try T, X.

bryanwweber:
Here is the rest of the conversation about the commit adding a TPX setter for PureFluid phases:

@bryanwweber: 

> I'd have to search through the commit history, but I believe this was explicitly removed previously. For pure fluids, t, p, and x overspecify the state.

@ischoegl:

> This is true, but prevents an import of saved data being loaded back into a SolutionArray, where a user may save T, P, X histories for a phase transition (being able to reconstruct this history is a completely legitimate assumption). The proposed setter checks for inconsistent states, and picks the most suitable setter based on whether the state is or is not saturated.
>
> It can be said that saving T, D (or other histories) would make this unnecessary. In view of existing setters for TP, TX and PX (all of which involve implicit assumptions), I feel like allowing for TPX import is a cleaner solution.

(Sorry about the close/reopen, I fat-fingered the buttons on my phone)

ischoegl:
> 
> I don't think the setter is correct. If, say, the quality is saved such that it's close to saturated vapor, T and p are not unique compared to the saturated liquid, so using T, p to set the state isn't unique. At least, that's how I'm reading the `elif` condition.
> 

I believe the setter is correct. `X` is only a legitimate property when conditions are saturated, i.e. under a case when `T` and `P` are not uniquely defining the state during a phase change. Conversely, in a single phase `T` and `P` are not linked, and `X` either has to be 0 or 1. I.e., in a single phase, `X` becomes essentially meaningless as a property. Afaik, no other values are permissible for a simple compressible substance.

Edit: *re-reading* the scenario, we're actually on the same page. Note that this is only possible if P=P_sat(T), which is when the `if` condition catches (also: there already is a unit test for this).

> 
> In any case, shouldn't it be possible to implement the check in the `SolutionArray` method? If the phase is a `PureFluid`, and `X` is given and is not `None`, use that plus another independent variable. Or, try using T, P and if it fails, try T, X.

The current implementation uses a single setter once suitable properties are identified, i.e. implementing a `TPX` setter is much simpler. It would require convoluted logic to handle `PureFluid` `TP`/`TX` setting as a special case.

ischoegl:
Looking over open issues, there are some repercussions for #595, which relocates restoring states to the C++ core. A proper solution would entail defining `_full_states` (or equivalent) for C++ objects rather than defining them in the Cython layer. 

Edit: #644 is another example where setting of states is not handled consistently with phase definitions.

As mentioned, I believe that a list of legitimate setters should be generated within C++. Nevertheless, I do not think that `SolutionArray.restore_data` needs to defer state restoration to the underlying C++ objects. The data labeling for export/import is defined within Python, and there is no underlying data structure that exists within C++.

bryanwweber:
@ischoegl See #307 and #308 where the `TPX` was "removed" (more specifically, defined only to have a getter and no setter). I'll have some more thoughts later, but I didn't want to lose those issue numbers! 😄

ischoegl:
To summarize some thoughts on setting states within C++ and the python interface:
* C++ should have well-defined independent parameters that specify the state; the choice of parameters may differ depending on the type of phase implemented (see #595)
* While `SolutionArray` objects only exist in Python, they do utilize setters that are implemented in the Python layer and tie into C++. I.e. Python setters will have to be aware of what states need to be assigned
* At the moment, most Python setters make use of `ThermoPhase`, with `PureFluid` being the only inheriting class that redefines setters. 
* Creating a more flexible C++ state definition will require a more differentiated approach for Python representations of different phase objects.

interfaces/cython/cantera/composite.py
```diff
@@ -691,11 +833,25 @@ def to_pandas(self, cols=('extra', 'T', 'density', 'Y'),
...
+
+        This method is intendend for loading of data that were previously
+        exported by `to_pandas`. The method requires a working pandas
+        installation. The package 'pandas' can be installed using pip or conda.
+        """
+
+        data = df.to_numpy(dtype=float)
+        labels = [col for col in df.columns]
+
+        self.restore_data(data, list(labels))
```
speth: `labels` is already a list here, isn't it? Or better still, I think you could just write `list(df.columns)` and get rid of the list comprehension.

interfaces/cython/cantera/composite.py
```diff
@@ -691,11 +833,25 @@ def to_pandas(self, cols=('extra', 'T', 'density', 'Y'),
         installation. Use pip or conda to install `pandas` to enable this method.
         """
 
-        # local import avoids explicit dependence of cantera on pandas
-        import pandas as pd
+        if isinstance(_pandas, ImportError):
+            raise ImportError(_pandas.msg)
```
speth: This can just be `raise _pandas`, right? The traceback will be different, but one doesn't seem better than the other to me.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1520,3 +1521,156 @@ def test_write_csv(self):
...
+        states.TPX = np.linspace(300, 1000, 7), 2e5, 'H2:0.5, O2:0.4'
+        try:
+            # this will run through if pandas is installed
+            df = states.to_pandas()
+            self.assertTrue(df.shape[0]==7)
+        except ImportError as err:
+            # pandas is not installed and correct exception is raised
+            pass
+        except Exception as err:
+            raise(err)
```
speth: `raise` is a statement, not a function, so the parentheses here are misleading. Also, I think the `except Exception as err` case can simply be omitted -- if an exception other than `ImportError` occurs, it will not be caught, and the test will fail, which is the desired result.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1520,3 +1521,156 @@ def test_write_csv(self):
...
+
+        # test ThermoPhase
+        a = ct.SolutionArray(self.gas)
+        for i in range(10):
+            T = 300 + 1800*np.random.random()
+            P = ct.one_atm*(1 + 10*np.random.random())
+            X = np.random.random(self.gas.n_species)
+            X[-1] = 0.
+            X /= X.sum()
+            a.append(T=T, P=P, X=X)
```
speth, line 1595: This is a lot of distinct tests in a single function. I would usually suggest making these at least a bit more atomic. You could put the creation of the initial `SolutionArray` `a` into the `setupClass` method if you want to avoid the excess initializations.

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,116 @@ def equilibrate(self, *args, **kwargs):
...
+        not contain those entries, an error is raised.
+        """
+
+        # check arguments
+        if not isinstance(data, np.ndarray) or data.ndim != 2:
+            raise TypeError("restore_data only works for 2D ndarrays")
+        elif len(labels) != data.shape[1]:
+            raise ValueError("inconsistent data and label dimensions")
+        rows = data.shape[0]
+        if self._shape!=(0,) and self._shape!=(rows,):
```
```suggestion
        if self._shape != (0,) and self._shape != (rows,):
```

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,116 @@ def equilibrate(self, *args, **kwargs):
...
+            raise ValueError('incompatible dimensions.')
+
+        # get full state information (may differ depending on type of ThermoPhase)
+        full_states = [fs for fs in self._phase._full_states.values()]
```
speth: While the `dict_values` object returned by `dict.values()` is a little weird, I think it can be used directly without needing to convert it to a true `list`. And if the list really is needed, the best way of doing that would just be `list(self._phase._full_states.values())`.

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,116 @@ def equilibrate(self, *args, **kwargs):
...
+                             if s in labels}
+            all_species = [l for l in labels if l[:2] == prefix]
+            if len(valid_species):
+                mode = prefix[0]
+                full_states = [v[:-1] for v in full_states if mode in v]
+                break
+        if len(valid_species) != len(all_species):
+            raise ValueError('incompatible species information.')
+        if mode == '':
+            full_states = {v[:2] for v in full_states}
```
speth, line 659: I think the logic here lets through the case where no mass or mole fractions have been provided for a phase with more than one species, which should presumably be an error.

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,116 @@ def equilibrate(self, *args, **kwargs):
...
+            spc = ['{}{}'.format(prefix, s) for s in self.species_names]
+            valid_species = {s[2:]: labels.index(s) for s in spc
+                             if s in labels}
+            all_species = [l for l in labels if l[:2] == prefix]
+            if len(valid_species):
+                mode = prefix[0]
+                full_states = [v[:-1] for v in full_states if mode in v]
+                break
+        if len(valid_species) != len(all_species):
+            raise ValueError('incompatible species information.')
```
speth: It would be nice if this error message could be a bit more informative, e.g. listing the missing / extra species names. One way to make doing so a bit easier would be to form the lists of species names as `set`s and just take the difference in each direction to get the missing / extra species.

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,116 @@ def equilibrate(self, *args, **kwargs):
...
+        properties is omitted. If the receiving SolutionArray has extra
+        entries already specified, only those will be imported; if *labels* does
+        not contain those entries, an error is raised.
+        """
+
+        # check arguments
+        if not isinstance(data, np.ndarray) or data.ndim != 2:
+            raise TypeError("restore_data only works for 2D ndarrays")
+        elif len(labels) != data.shape[1]:
+            raise ValueError("inconsistent data and label dimensions")
```
speth: Adding the two non-matching lengths to the error message might help in debugging.

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,116 @@ def equilibrate(self, *args, **kwargs):
...
+                'V': ('v', 'volume_{}'.format(basis)),
+                'H': ('h', 'enthalpy_{}'.format(basis)),
+                'S': ('s', 'entropy_{}'.format(basis))}
+        for fs in full_states:
+            state = [{fs[i]: labels.index(p) for p in prop[fs[i]] if p in labels}
+                     for i in range(len(fs))]
+            found = [len(state[i]) for i in range(len(fs))]
+            if all(found):
+                mode = fs + mode
+                break
```
speth, line 676: This section is complex enough that a few additional comments explaining what's being done here and why would be helpful.

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,116 @@ def equilibrate(self, *args, **kwargs):
...
+                'D': ('density', 'density_{}'.format(basis)),
+                'U': ('u', 'int_energy_{}'.format(basis)),
+                'V': ('v', 'volume_{}'.format(basis)),
+                'H': ('h', 'enthalpy_{}'.format(basis)),
+                'S': ('s', 'entropy_{}'.format(basis))}
+        for fs in full_states:
+            state = [{fs[i]: labels.index(p) for p in prop[fs[i]] if p in labels}
+                     for i in range(len(fs))]
+            found = [len(state[i]) for i in range(len(fs))]
+            if all(found):
```
speth: You can use the fact that empty `dict`s evaluate to `False` and non-empty to `True` to skip creating the `found` dict and just write `if all(state):`.

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,116 @@ def equilibrate(self, *args, **kwargs):
...
+        # raise warning if state is potentially not uniquely defined
+        if isinstance(self._phase, PureFluid) and mode in last:
+            warnings.warn('Using  mode `{}` to restore data: may not '
+                          'be sufficient to define unique state '
+                          'for a PureFluid phase'.format(mode),
+                          UserWarning)
+
+        # assemble and restore state information
+        state_data = tuple([data[:, state[i][mode[i]]] for i in range(len(state))])
+        if len(valid_species):
```
speth: This is equivalent, and preferred:
```suggestion
        if valid_species:
```

interfaces/cython/cantera/composite.py
```diff
@@ -681,6 +810,19 @@ def write_csv(self, filename, cols=('extra', 'T', 'density', 'Y'),
...
+    def read_csv(self, filename):
+        """
+        Read a CSV file named *filename* and restore data to the SolutionArray
+        using `restore_data`. This method allows for recreation of data
+        previously exported by `write_csv`.
+        """
+        # read data block and header separately
+        data = np.genfromtxt(filename, skip_header=1, delimiter=',')
+        labels = np.genfromtxt(filename,
+                               max_rows=1, delimiter=',', dtype=str)
```
speth: A very minor point, but I think this fits on one line -- 78 characters versus the nominal limit of 80.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1520,3 +1521,156 @@ def test_write_csv(self):
...
+
+        data, labels = a.collect_data()
+
+        # basic restore
+        b = ct.SolutionArray(self.gas)
+        b.restore_data(data, labels)
+        self.assertTrue(check(a, b))
```
speth: Rather than using `assertTrue(check(...))`, I would recommend making `check` a member function which calls other assert functions. This will help provide more granular error messages, e.g. you will know exactly which of the arrays didn't match.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1520,3 +1521,156 @@ def test_write_csv(self):
...
+        except Exception as err:
+            raise(err)
+
+    def test_restore(self):
+
+        def check(a, b, atol=None):
+            if atol is None:
+                fcn = lambda c, d: np.allclose(c, d)
+            else:
+                fcn = lambda c, d: np.allclose(c, d, atol=atol)
```
speth: It's not identical to `np.allclose`, but in much of the test suite we use a custom function `CanteraTest.assertArrayNear` which provides extra data about failed comparisons, e.g.
```
AssertionError: AssertNear: 354.98584001686 - 999 = -644.01415998314
Relative error for element (3,) of 9.512863e-01 exceeds rtol = 1.000000e-08
```
Rather than the sometimes mysterious `AssertionError: False is not true`.

speth (changes requested):
Overall, this looks very good to me. The set of cases handled is quite comprehensive, and I appreciate the effort put into making any valid combination of properties work for setting the state.

The only one that I question is reintroducing the `TPX` setter to `PureFluid`. There are better choices easily available -- the default properties include the best one, temperature and density, so I don't think we need to create a fallback option that mostly works but has potential for weirdness for points that fall within the threshold of the `isclose` checks.

interfaces/cython/cantera/composite.py
ischoegl: True! Both `valid_species` and `all_species` would be empty ... however, I used this as a feature (see one of my unit tests):
```python
        b.restore_data(data[:, :2], labels[:2])
```
i.e. the original concentrations of `self.gas` would be applied to all entries.

interfaces/cython/cantera/test/test_thermo.py
ischoegl: Excellent suggestion.

interfaces/cython/cantera/test/test_thermo.py
ischoegl: NP - done!

interfaces/cython/cantera/test/test_thermo.py
ischoegl: I wanted to keep all of the tests that pertain to `restore_data` together (which makes it easy to keep the `check` function in the function's namespace).

ischoegl:
@speth ... thanks for the review (excellent comments like always)

> The only one that I question is reintroducing the TPX setter to PureFluid. There are better choices easily available -- the default properties include the best one, temperature and density, so I don't think we need to create a fallback option that mostly works but has potential for weirdness for points that fall within the threshold of the isclose checks.

I am extremely torn on this one. I see the following two options:
1. disable import for what I reserved as fallbacks (i.e. `last = ['TP', 'TX', 'PX']`) altogether, or
2. add the `TPX` setter as suggested

I was leaning towards (2), as setters for `TP`, `TX` etc. do exist at the moment (at the peril of the user). In my implementation, I made sure that the fallback is only used if alternatives (e.g. `TD`) are not an option. However, as I had mentioned to @bryanwweber, my rationale was that a user may expect that saving `TPX` states throughout a phase transition (subcooled all the way to superheated) should allow for a reconstruction at a later point.

PS: another option is to not use a setter for a `TPX` assignment (i.e. alternate route to `setattr`); it's not consistent though (and adds unnecessary complexity).

interfaces/cython/cantera/composite.py
ischoegl: Agreed - I added several comments throughout.

interfaces/cython/cantera/composite.py
speth: Got it - the comment makes this more clear.

interfaces/cython/cantera/composite.py
```diff
@@ -736,6 +899,24 @@ def write_hdf(self, filename, cols=('extra', 'T', 'density', 'Y'),
...
+        recreation of data previously exported by `write_hdf`.
+
+        The method imports data using `restore_data` via `from_pandas` and
+        requires working installations of pandas and PyTables. These
+        packages can be installed using pip (`pandas` and `tables`) or conda
+        (`pandas` and `pytables`).
+        """
+
+        if isinstance(_pandas, ImportError):
+            raise ImportError(_pandas.msg)
```
speth: Another instance of the prior comment about just using `raise _pandas`. I think I prefer the message generated by this over that generated by `raise ImportError(_pandas)` -- the former traceback points to the `import pandas as _pandas` statement, while the other points to this line.

speth (changes requested):
All the updates look good to me. Just one more instance to fix of re-raising the stored `ImportError`.

I'm also on the fence with the PureFluid / TPX issue is tricky. Both options have pitfalls (I'm not a fan of option 3, either), and I'm not sure which one I think is better. For option 1, there will be users who try to do this, but the fix is pretty easy -- just store T and density or another fully independent property pair. Running into issues with option 2 is probably a lot less likely, but I think they will be harder to pin down as being related to this overspecification.

ischoegl:
@speth ... thanks for the quick turn-around! Ad `TPX` ...

> Running into issues with option 2 is probably a lot less likely, but I think they will be harder to pin down as being related to this overspecification.

I believe providing an improved error message should address this sufficiently? I. e.

```python
                raise ValueError(
                    'invalid thermodynamic state: the `TPX` setter '
                    'received a combination of values that does not '
                    'represent an valid state. As an alternative, '
                    'specify the state using two fully independent '
                    'properties (e.g. `TD`)')
```
I hope this adequately addresses the legitimate concerns.

bryanwweber:
My understanding is that the user can control which properties are output when saving, say, a DataFrame. So it also seems reasonable they would in some cases not store independent properties across a phase transition (e.g., T and D as you suggest), but still want to be able to restore the state from their data file. However, I also agree that adding back the `TPX` setter is not an ideal solution. I'm afraid I don't have anything new to add to the discussion, even after almost 2 months of thinking about it...

One comment I have is on the UserWarning that's raised when X is used with a PureFluid. I'm not sure this warning is worth adding. Either the state will be uniquely defined, and the warning is unnecessary, or it won't, and the C++ will raise an error (I hope this is the case, at least), in which case you get a warning and an error.

interfaces/cython/cantera/composite.py
ischoegl: I adopted your suggestion in both instances.

ischoegl:
> My understanding is that the user can control which properties are output when saving, say, a DataFrame. So it also seems reasonable they would in some cases not store independent properties across a phase transition (e.g., T and D as you suggest), but still want to be able to restore the state from their data file. 

This is my rationale.

> However, I also agree that adding back the `TPX` setter is not an ideal solution. I'm afraid I don't have anything new to add to the discussion, even after almost 2 months of thinking about it...

:man_shrugging: (hence my suggestion to clarify the error message)

> One comment I have is on the UserWarning that's raised when X is used with a PureFluid. I'm not sure this warning is worth adding. Either the state will be uniquely defined, and the warning is unnecessary, or it won't, and the C++ will raise an error (I hope this is the case, at least), in which case you get a warning and an error.

Legitimate (while it may be overspecified, if things work, there is no problem).

PS: the `UserWarning` is gone now.

bryanwweber:
~On the `ValueError`, I think it is too strict, for instance, the following should work but does not:~
```
w = ct.Water
w.TPX
(300, 101325, 0.0)
w.TPX = None, 200000, 0.5
ValueError: ...
```

ETA: just kidding, I didn't think this through enough... hold on, something is nagging at me here.

Edit 2: OK, I think this is fine logically. What's nagging at me is there just doesn't seem to be a good reason for anyone to use `TPX` for PureFluids aside from the specific use case in this PR of restoring bunches of states.

Is there any way to detect that the user is calling the setter vs. the functions added here? In other words, I think I'm coming around to considering option 3 a little more. PureFluids are already inconsistent with the base ThermoPhase, so adding one more difference doesn't seem that bad to me. The complexity is something to consider though...

Last, one minor comment about the formatting in the error messages... no need to include backticks around code/properties/etc. It doesn't get read by Sphinx, so doesn't get formatted, and it is monospace on the terminal anyways

ischoegl:
> On the `ValueError`, I think it is too strict.

Hmm .. also got a unit test failure, but that was due to the removal of the UserError. 

Ad Edit 2: option 3 can be implemented. The backtick comment is noted.

speth:
While my initial instinct was that there was a way to get yourself into trouble with the TPX setter, the logic seems pretty solid and I haven't been able to figure out a way to get any surprising behavior out of it.

@bryanwweber I think that the `ValueError` should be expected in that case, right? Passing `None` for T means that you want it to keep its current value, and there's no way to do that and also be a two phase state at the specified pressure.

ischoegl:
> While my initial instinct was that there was a way to get yourself into trouble with the TPX setter, the logic seems pretty solid and I haven't been able to figure out a way to get any surprising behavior out of it.
> 
:thumbsup:

@bryanwweber  suggested a hidden setter, i.e. `_TPX`. Thoughts?

bryanwweber:
@speth Yeah, I was mis-remembering what a `None` value for a property meant in the setters. My original thought was that `None` should not set any value for that property, when in fact it is holding that property constant, and `T=300`, `P=200000`, and `X=0.5` is indeed inconsistent and should raise the error. 🤦‍♂️

I still think the `ValueError` would be surprising behavior in many cases from the user perspective, even if it is the correct behavior. What about using `_TPX` as the setter for PureFluids? Then a user won't accidentally use it, but the logic doesn't have to be reimplemented in the restore functions. I guess the question is, is an attribute error better than a value error?

ischoegl:
The hidden setter `_TPX` is a solution I hadn't thought of ... it won't create much overhead (except for adding a character in the `mode` string). I don't have a strong preference here though ...

speth:
Well, the `TPX` property exists in either case, so there's no way to prevent a user from *trying* to set it. Currently, they get an `AttributeError` that doesn't really explain why that value isn't writable, but with the implementation in this PR, they will get a pretty clear description of what is wrong.

ischoegl:
@bryanwweber ...

> What's nagging at me is there just doesn't seem to be a good reason for anyone to use `TPX` for PureFluids aside from the specific use case in this PR of restoring bunches of states.

True. At the same time, any user has to get pretty lucky to 'guess' a valid 3-property state without getting the `ValueError` that points to what's awry.

> Is there any way to detect that the user is calling the setter vs. the functions added here? 

Could we resolve this by applying PEP 20 (it is implemented in Python after all?) :wink:

bryanwweber:
Here are my thoughts with the "hidden" `_TPX` setter: 

1. The main benefit to me is that it sends a clear signal that `TPX` is not meant to be set for `PureFluid` phases. This might help relatively inexperienced users avoid a pitfall and clearly indicates the intent to future developers.
2. The disadvantage is that any users who do try to set `TPX` anyways get an `AttributeError` instead of a possibly more useful `ValueError`. In addition, it adds another property that would be specific to `PureFluid` phases.
3. Since it is specific to `PureFluid`, it doesn't generalize if, in the future, we decide to allow more complicated EOS (R-K or P-R comes to mind) to have their state set with the mixture quality. I'm not sure if this is something we can or will support, but on my mind.

I did some experimenting this morning trying to do weird things with the regular `TPX` setter and I also couldn't find any really weird results. One suggestion I have is to do a bit of type checking on the `X` part of the setting. For instance, I think it would help to improve the error message in the following situation, which I think would be common if people try to set mole fractions of a `PureFluid`:
```
In [6]: w.TX = 500, 0.5

In [7]: w.TPX
Out[7]: (500.0, 2634957.1389629007, 0.5)

In [8]: w.TPX = None, None, {'CH4': 1.0}
<ipython-input-8-1ecf254b1d00> in <module>
----> 1 w.TPX = None, None, {'CH4': 1.0}

~/GitHub/cantera/interfaces/cython/cantera/thermo.pyx in cantera._cantera.PureFluid.TPX.__set__()
   1578             X = values[2] if values[2] is not None else self.X
   1579             if np.isclose(P, self.thermo.satPressure(T)):
-> 1580                 self.TX = T, X
   1581             elif np.isclose(X, 0.) or np.isclose(X, 1.):
   1582                 self.TP = T, P

~/GitHub/cantera/interfaces/cython/cantera/thermo.pyx in cantera._cantera.PureFluid.TX.__set__()
   1464             T = values[0] if values[0] is not None else self.T
   1465             X = values[1] if values[1] is not None else self.X
-> 1466             self.thermo.setState_Tsat(T, X)
   1467
   1468     property PX:

TypeError: must be real number, not dict
```
That `TypeError` is pretty clear, but I think it could be more helpful if `TPX` checked for a number and raised an error, rather than deferring to `TX`.

One other suggestion is to print the T, P, and X values in the `ValueError` message. It's already pretty long, and I don't immediately see a way to shorten it, but I think that would be good too.

ischoegl:
> That TypeError is pretty clear, but I think it could be more helpful if TPX checked for a number and raised an error, rather than deferring to TX.

> One other suggestion is to print the T, P, and X values in the ValueError message. It's already pretty long, and I don't immediately see a way to shorten it, but I think that would be good too.

Both suggestions are reasonable and can be implemented with little effort. Will get to it later today (which also allows @speth to chime in).

PS: ad other EOS ... the dual-purpose `X` being used for mole fraction and mixture quality depending on context raises some 'interesting' questions for R-K, etc.. I do not have an opinion beyond this observation at this point (it certainly goes well beyond this PR - and I haven't looked into what is being done in other PR's that are currently implementing these).

bryanwweber:
To my understanding, the other EOS types are multi-species capable and `X` means mole fraction. Given this is the case, it may make sense to use `Q` for quality at some point in the future. Clearly out of scope for this PR, but just throwing it out there...

ischoegl:
> Given this is the case, it may make sense to use `Q` for quality at some point in the future. Clearly out of scope for this PR, but just throwing it out there...

:thumbsup: that would have been my go-to solution also

speth:
All of the choices here are nonideal. I don't really like the idea of adding the `_TPX` property which will only be used in this one scenario. At the same time, that same logic basically applies to adding a setter for `TPX`.

I'm also starting to think that option 3 doesn't sound so bad, but I'm still OK with option 1 as implemented.

> Given this is the case, it may make sense to use Q for quality at some point in the future. Clearly out of scope for this PR, but just throwing it out there...

This would also be consistent with other naming conventions that have been adopted in the Python module, e.g. D for density instead of R for Rho. I think there are additional, larger complexities for the multi-species two-phase phases that are currently unsolved, e.g. how to access the composition of each phase which currently can't be done.

ischoegl:
:man_shrugging: ... I can go either way. I have a mild preference for 1 (mostly PEP 20 related), but do not have a problem with 3. On the other hand, since the `ValueError` will appear for anyone not respecting thermodynamics, I feel the facetious urge to leave things in place :grin:

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,117 @@ def equilibrate(self, *args, **kwargs):
             self._phase.equilibrate(*args, **kwargs)
             self._states[index][:] = self._phase.state
 
+    def restore_data(self, data, labels):
+        """
+        Restores a SolutionArray based on *data* specified in a single
```
bryanwweber: `SolutionArray` should be in backticks here, I think, so it auto-links. Same below.

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,117 @@ def equilibrate(self, *args, **kwargs):
...
+
+        # check arguments
+        if not isinstance(data, np.ndarray) or data.ndim != 2:
+            raise TypeError("restore_data only works for 2D ndarrays")
+        elif len(labels) != data.shape[1]:
+            raise ValueError("inconsistent data and label dimensions: "
+                             "({} vs. {})".format(len(labels), data.shape[1]))
+        rows = data.shape[0]
+        if self._shape != (0,) and self._shape != (rows,):
+            raise ValueError('incompatible dimensions.')
```
bryanwweber: Can you clarify which dimensions are incompatible here?

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,117 @@ def equilibrate(self, *args, **kwargs):
...
+        if isinstance(self._phase, PureFluid):
+            # make sure that potentially non-unique state definitions are checked last
+            last = ['TP', 'TX', 'PX']
+            full_states = [fs for fs in full_states
+                           if fs not in last] + ['TPX'] + last
+
+        # determine whether complete concentration is available (mass or mole)
+        # assumes that `X` or `Y` is always in last place
+        mode = ''
+        has_species = False
```
bryanwweber: My linter says this variable is assigned here but not used anywhere else. Can it be removed?

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,117 @@ def equilibrate(self, *args, **kwargs):
...
+        for prefix in ['X_', 'Y_']:
+            spc = ['{}{}'.format(prefix, s) for s in self.species_names]
+            # solution species names also found in labels
+            valid_species = {s[2:]: labels.index(s) for s in spc
+                             if s in labels}
+            # labels that start with prefix (indicating concentration)
+            all_species = [l for l in labels if l[:2] == prefix]
```
bryanwweber: This might be simpler as:
```
all_species = list(filter(lambda x: x.startswith(prefix), labels))
```
or at least, `str.startswith` could be used instead of `l[:2] == prefix`

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,117 @@ def equilibrate(self, *args, **kwargs):
...
+        if len(valid_species) != len(all_species):
+            incompatible = list(set(valid_species) ^ set(all_species))
+            raise ValueError('incompatible species information for '
+                             '{}'.format(incompatible))
+        if mode == '':
+            # concentration specifier ('X' or 'Y') is not used
+            full_states = {v[:2] for v in full_states}
+
+        # determine suitable thermo properties for reconstruction
+        basis = {'molar': 'mole', 'mass': 'mass'}[self.basis]
```
bryanwweber: I'm not sure about this construct, would it be better to say
```
basis = 'mass' if self.basis == 'mass' else 'mole'
```
?

interfaces/cython/cantera/composite.py
```diff
@@ -576,6 +595,117 @@ def equilibrate(self, *args, **kwargs):
...
+        for fs in full_states:
+            # identify property specifiers
+            state = [{fs[i]: labels.index(p) for p in prop[fs[i]] if p in labels}
+                     for i in range(len(fs))]
+            if all(state):
+                # all property identifiers match
+                mode = fs + mode
+                break
+        if len(mode) == 1:
+            raise ValueError('invalid/incomplete state information.')
```
bryanwweber: Can you add the mode to this error message?

interfaces/cython/cantera/thermo.pyx
```diff
@@ -1564,9 +1565,33 @@ cdef class PureFluid(ThermoPhase):
...
+            elif np.isclose(X, 0.) or np.isclose(X, 1.):
+                self.TP = T, P
+            else:
+                raise ValueError(
+                    'invalid thermodynamic state: the TPX setter '
+                    'received a combination of property values that '
+                    'do not represent a valid state. As an alternative, '
+                    'specify the state using two fully independent '
+                    'properties (e.g. TD) instead of: '
+                    'T={}, P={}, X={}'.format(T, P, X))
```
bryanwweber: The last parentheses here should be on the next line and dedented.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -1564,9 +1565,33 @@ cdef class PureFluid(ThermoPhase):
...
+        An Exception is raised if the thermodynamic state is not consistent.
+        """
         def __get__(self):
             return self.T, self.P, self.X
+        def __set__(self, values):
+            T = values[0] if values[0] is not None else self.T
+            P = values[1] if values[1] is not None else self.P
+            X = values[2] if values[2] is not None else self.X
+            if not isinstance(X, (np.ndarray, _numbers.Number)):
+                raise ValueError(
```
bryanwweber, line 1581: I don't think this needs to be on a new line, at least at this length. Can you change `X` to `vapor fraction (X)`? I think that would help clarify.

interfaces/cython/cantera/test/test_purefluid.py
```diff
@@ -149,11 +149,18 @@ def test_fd_properties_twophase(self):
...
-        T,P,X = self.water.TPX
+        T, P, X = self.water.TPX
         self.assertNear(T, 400)
         self.assertNear(X, 0.8)
-        with self.assertRaises(AttributeError):
-            self.water.TPX = 500, 101325, 0.3
+
+        self.water.TPX = T, P, X
+        self.assertNear(X, 0.8)
+        with self.assertRaises(ValueError):
```
bryanwweber: Since these `ValueError`s are different (at least, the third is different from the first two), can you use `assertRaisesRegex` to check the message in addition to the type?

bryanwweber (changes requested):
I'm :+1: with the properties as they are here (i.e., option 1), and I've suggested a few small changes here.

interfaces/cython/cantera/composite.py
ischoegl: :thumbsup: ... there were quite a few instances in other portions I added in other PR's. I replaced them all.

interfaces/cython/cantera/composite.py
ischoegl: :thumbsup: ... I routinely use this for longer selections, but your suggestion is clearly better in this instance!

interfaces/cython/cantera/composite.py
ischoegl: fancy. not using `filter` much (probably I should)

interfaces/cython/cantera/composite.py
ischoegl: Sure (i.e. done). Also finally displaying column numbers in emacs (sigh). Was 'guessing' thus far ...

interfaces/cython/cantera/composite.py
bryanwweber: I wish the filter could just directly use `str.startswith` instead of needing the `lambda`, but yeah, I've not used it too much either.

bryanwweber (approved):

interfaces/cython/cantera/composite.py
speth: I'm :+1: on using `startswith`, but I think the list comprehension is preferable to the `lambda`+`filter` combo. Also, the former is used at least occasionally within Cantera, while we have no existing uses of the `filter` function.

ischoegl: :man_shrugging: any of the implementations are fine with me (obviously including the original one). I see that `map`, `filter` and `reduce` are nice features, but readability does suffer (I still should likely use it more often). `startswith` I hadn't used before. The `l[:2]` is very literal as it points towards a comparison of two letters.

ischoegl:
```python
-            all_species = list(filter(lambda x: x.startswith(prefix), labels))
+            all_species = [l for l in labels if l.startswith(prefix)]
```

ischoegl: I.e. I split the difference.

speth (approved):

stijn76:
I'm very new to this forum (and to any development forum for that matter), so my apologies in case my remark is not formulated at the right place or in the right way. I was using the latest cantera version, and trying to recreate a SolutionArray objects from stored data, but got a ValueError. The problems does not seem to be solved in the composite.py source code files which I found here at github.

The minimal code to reproduce the problem is:

```
import cantera as ct

premix = ct.Solution('gri30.yaml')
states = ct.SolutionArray(premix)
test1 = ct.SolutionArray(premix)
test2 = ct.SolutionArray(premix)

T_in = 300
P_in = 101300
air2gas = 7.6
premix.X = 'N2:'+str(air2gas*0.79) + ', O2:'+str(air2gas*0.21) + ', CH4:'+str(1.0)
premix.TP = T_in , P_in
states.append(T=premix.T, P=premix.P, X=premix.X)
states.write_csv('test1.csv', 'X')

# Reading the csv file with only 1 data line, throws an IndexError
# test1.read_csv('test1.csv')

air2gas = 8.3
premix.X = 'N2:'+str(air2gas*0.79) + ', O2:'+str(air2gas*0.21) + ', CH4:'+str(1.0)
premix.TP = T_in , P_in
states.append(T=premix.T, P=premix.P, X=premix.X)
states.write_csv('test2.csv', 'X')

# Reading the csv file with more than 1 data line, throws a ValueError
test2.read_csv('test2.csv')
```

I believe the problem is situation in de definition of the "join" function in the "composite.py" file. More specifically, within that definition, valid species is declared as:
```
valid_species = {s[2:]: labels.index(s) for s in spc
                             if s in labels}
```
but should be declared as
```
valid_species = {s[:]: labels.index(s) for s in spc
                             if s in labels}
```
or else the ValueError will always be thrown because valid_species represents species_names without prefixes ('X_', or 'Y_') while all_species represents species_names with those prefixes.
```
if len(valid_species) != len(all_species):
                    incompatible = list(set(valid_species) ^ set(all_species))
                    raise ValueError('incompatible species information for '
                                    '{}'.format(incompatible))
```

Warm regards, Stijn

stijn76:
...sorry, the large bold sentences in my previous post were meant to be comments in the python code...

speth:
I think your post from earlier today in the Cantera Users' Group is the right place to discuss this. For anyone else finding this, the post can be found at https://groups.google.com/g/cantera-users/c/AFa5u3F6JdQ.

ischoegl:
@stijn76 ... I edited your post so it shows up properly (code blocks should be between triple back ticks).

There is a minor bug in your code, as you do not export the full state to csv (only concentrations). But even fixing that, the errors you pointed out persist, and I was able to confirm what you pointed out above. 

Interestingly, things mostly work for `h2o2.yaml` (which is used for unit tests [here](https://github.com/Cantera/cantera/blob/1420a4cf2901c8ae59d895fec4ab19fc41580207/interfaces/cython/cantera/test/test_composite.py#L175-L192)), although the single line bug still persists. In other words, this works:
```python
import cantera as ct

gas = ct.Solution('h2o2.yaml')  # if changed to gri30.yaml, a ValueError is thrown for read_csv
states2 = ct.SolutionArray(gas)
gas.X = 'H2:0.5, O2:0.4'
gas.TP = 300, 101300
states2.append(T=gas.T, P=gas.P, X=gas.X)
states2.append(T=gas.T, P=gas.P, X=gas.X) # if commented out, this will throw an IndexError for read_csv

states2.write_csv('test3.csv', ('T', 'P', 'X'))
b = ct.SolutionArray(gas)
b.read_csv('test3.csv')
```

**PS:** Please create a *new* issue/bug report, rather than attaching this to this old pull request.

stijn76:
thanks for your help and advise! I've created a new bug report.

PR #688, opened 2019-08-10
Add ability to sort SolutionArray objects

### LINKED ISSUE #625: Add ability to sort SolutionArray
Right now, it is not possible to sort `SolutionArray` instances without copying everything into a new `SolutionArray` with a loop. Example, courtesy of @arghdos:

```python
import cantera as ct
gas = ct.Solution('gri30.cti')
arry = ct.SolutionArray(gas, extra=['tau'])
arry.append(gas.state, tau=1)
gas.TP = 3000, 101325
arry.append(gas.state, tau=0.5)
tau_idx = np.argsort(arry.tau)
out = ct.SolutionArray(gas, extra=['tau'])
for idx in tau_idx:
	out.append(arry._states[idx], tau=arry.tau[idx])
```

It would be useful to add a `.sort()` method that takes a column name or index as the argument and sorts based on that. To implement this will require converting the "internal" `_states` object to be consistently either a list or an array. The code below indicates (to me, anyways) that if the user provides the `states` argument, it is passed through as is (list or array), and if the shape of the `states` is going to be one-dimensional, the `_states` is a list.

https://github.com/Cantera/cantera/blob/f078f53f39118416e4d3349fce99b8e105d5b0b4/interfaces/cython/cantera/composite.py#L449-L459

ischoegl:

Fixes #625

- implements `sort` method for 1D SolutionArray objects
- adds some simple type checks for an instantiation with `states` not `None`

Intended usage:
```python
# create SolutionArray and add items with `tau` not following an increasing order
states = ct.SolutionArray(gas, extra=('tau',))
...
states.sort('tau')
```
Some observations: 
- Sorting of higher-dimensional `SolutionArray` objects is not clearly defined
- Difference/necessity of apparently duplicate `SolutionArray._extra_lists` and `SolutionArray._extra_arrays` dictionaries is not clear
- `SolutionArray._extra_lists` is apparently only used in `SolutionArray.__call__` (docstring is missing ... it works as copying constructor that extracts species, ~~but usage is not clear~~)

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -492,7 +492,12 @@ class SolutionArray:
             shape = (shape,)
 
         if states is not None:
-            self._shape = np.shape(states)[:-1]
+            if isinstance(states, list):
+                self._shape = (len(states),)
+            elif isinstance(states, np.ndarray):
+                self._shape = np.shape(states)[:-1]
+            else:
+                raise TypeError('invalid type')
             self._states = states
         else:
             self._shape = tuple(shape)
@@ -602,6 +607,25 @@ class SolutionArray:
         self._states.append(self._phase.state)
         self._indices.append(len(self._indices))
         self._shape = (len(self._indices),)
+
+    def sort(self, col, reverse=False):
+        """ 
+        Sort SolutionArray by column *col*.
+
+        :param col: Column that is used to sort the SolutionArray.
+        :param reverse: If True, the sorted list is reversed (descending order).
+        """
+        if len(self._shape) != 1:
+            raise TypeError("sort only works for 1D SolutionArray objects")
+
+        indices = np.argsort(getattr(self, col))
+        if reverse:
+            indices = indices[::-1]
+        self._states = [self._states[ix] for ix in indices]
+        for k, v in self._extra_arrays.items():
+            new = v[indices]
+            self._extra_arrays[k] = new
+            self._extra_lists[k] = list(new)
 
     def equilibrate(self, *args, **kwargs):
         """ See `ThermoPhase.equilibrate` """
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -492,12 +492,7 @@ class SolutionArray:
             shape = (shape,)
 
         if states is not None:
-            if isinstance(states, list):
-                self._shape = (len(states),)
-            elif isinstance(states, np.ndarray):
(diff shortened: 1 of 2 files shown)


bryanwweber:
Does it make sense to add a kwarg to reverse the sort? Or does the `[::-1]` indexing work to reverse the `SolutionArray`?

ischoegl:
`[::-1]` does not work, but implementing the kwarg is trivial. Can add this if there's a need.

interfaces/cython/cantera/composite.py
```diff
@@ -477,7 +477,12 @@ def __init__(self, phase, shape=(0,), states=None, extra=None):
             shape = (shape,)
 
         if states is not None:
-            self._shape = np.shape(states)[:-1]
+            if isinstance(states, list):
+                self._shape = (len(states),)
+            elif isinstance(states, np.ndarray):
+                self._shape = np.shape(states)[:-1]
+            else:
+                raise TypeError('invalid type')
```
speth: I'm not really sure what this change accomplishes -- `np.shape` works whether `states` is a `list` or an `np.array`, and in fact works better when the input is a list of lists.

speth (commented):
I realize this wasn't well documented, but the duplication of data between `_extra_arrays` and `_extra_lists` was meant as an optimization for accessing the "extra" variables. Data is initially stored in the list form so that appending to the `SolutionArray` is efficient. However, we want to return an array when those variables are accessed, and at the time I initially wrote this, I wanted to avoid the inefficiency of repeatedly converting the list to an array in the case where the data hadn't changed. 

However, with the expansion of the features of the `SolutionArray` class, I think maintaining the duplicate data isn't worth the effort, so keeping just the list-based storage makes sense. It might be worth renaming it just `_extra` in that case, since there is no need to distinguish between that and the now-removed array version.

interfaces/cython/cantera/composite.py
ischoegl: ... was overthinking this. Absolutely: list of np.arrays still works. Reverting!

ischoegl:
> Data is initially stored in the list form so that appending to the `SolutionArray` is efficient. 

That's how I interpreted this, and thought that `np.array` really only made sense for higher dimensions (which I don't believe need to be pre-empted at this point). Either way, I don't think it's necessary to keep duplicate data, and am happy that this can be simplified.

> It might be worth renaming it just `_extra` in that case, since there is no need to distinguish between that and the now-removed array version.

Done.

speth (approved):

PR #689, opened 2019-08-12
Save case sensitive species names

### LINKED ISSUE #621: Unable to import all species containing C and S
### Cantera version
2.5.0a2 (848a3bf)
### Operating System
Ubuntu 18.04

### Python/MATLAB version
Python 3.6

### Expected Behavior
Able to import all species from `nasa` containing C and S.

### Actual Behavior
CanteraError thrown: "phase gas already contains a species named Cs".

### Steps to reproduce
Run
```
import cantera as ct

gas_cti = """ideal_gas(
    name="gas",
    elements=" S C ",
    species=" nasa: all ",
    options=["skip_undeclared_elements"],
    initial_state=state(temperature=300, pressure=(1, "bar"))
)"""
gas = ct.Solution(source=gas_cti)
```

Error:
```
  File "equilibrium_modelling.py", line 20, in <module>
    gas = ct.Solution(source=gas_cti)
  File "interfaces/cython/cantera/base.pyx", line 29, in cantera._cantera._SolutionBase.__cinit__
  File "interfaces/cython/cantera/base.pyx", line 65, in cantera._cantera._SolutionBase._init_cti_xml
CanteraError thrown by Phase::addSpecies:
Phase 'reactor_gas' already contains a species named 'Cs'.
```

ischoegl:

Fixes #621

The following fails (example courtesy @Saintis).
```python
gas_cti = """ideal_gas(
    name="gas",
    elements=" S C ",
    species=" nasa: all ",
    options=["skip_undeclared_elements"],
    initial_state=state(temperature=300, pressure=(1, "bar"))
)"""
gas = ct.Solution(source=gas_cti)
```
as cantera cannot differentiate between existing species `Cs` and `CS`.

- store species names as case sensitive strings
- retain ability to look up non-case sensitive species names, e.g. `Phase::speciesIndex`
- implement flag that enforces case sensitive species names as a member variable of `Phase`
- add exception handling for species that are not uniquely defined unless case sensitive (e.g. `Cs` and `CS` in `nasa.cti` if `cs` is specified and case sensitivity is not enforced)
- ~~add deprecation warning to switch default behavior from non-enforcing to enforcing after cantera 2.5~~
- *potentially add switch to save all lower case species names* if explicitly stated *as an option for CTI/YAML import*

### DIFF
--- a/include/cantera/thermo/Phase.h
+++ b/include/cantera/thermo/Phase.h
@@ -4,7 +4,7 @@
  */
 
 // This file is part of Cantera. See License.txt in the top-level directory or
-// at http://www.cantera.org/license.txt for license and copyright information.
+// at https://cantera.org/license.txt for license and copyright information.
 
 #ifndef CT_PHASE_H
 #define CT_PHASE_H
@@ -707,6 +707,10 @@ public:
     //! Return the Species object for the named species. Changes to this object
     //! do not affect the ThermoPhase object until the #modifySpecies function
     //! is called.
+    /*!
+     * @deprecated To be removed after Cantera 2.5. Replaceable with
+     * Phase::species(size_t) via Phase::speciesIndex(std::string&).
+     */
     shared_ptr<Species> species(const std::string& name) const;
 
     //! Return the Species object for species whose index is *k*. Changes to
@@ -746,6 +750,17 @@ public:
     //! Invalidate any cached values which are normally updated only when a
     //! change in state is detected
     virtual void invalidateCache();
+
+    //! Returns `true` if case sensitive species names are enforced
+    bool caseSensitiveSpecies() const {
+        return m_caseSensitiveSpecies;
+    }
+
+    //! Set flag that determines whether case sensitive species are enforced
+    //! in look-up operations, e.g. speciesIndex
+    void setCaseSensitiveSpecies(bool cflag = true) {
+        m_caseSensitiveSpecies = cflag;
+    }
 
 protected:
     //! Cached for saved calculations within each ThermoPhase.
@@ -793,6 +808,14 @@ protected:
 
     //! Flag determining behavior when adding species with an undefined element
     UndefElement::behavior m_undefinedElementBehavior;
+
+    //! Find lowercase species name in m_speciesIndices when case sensitive
+    //! species names are not enforced. Raise exception if lowercase name
+    //! is not unique.
+    size_t findSpeciesLower(const std::string& nameStr) const;
+
+    //! Flag determining whether case sensitive species names are enforced
+    bool m_caseSensitiveSpecies;
 
 private:
     XML_Node* m_xml; //!< XML node containing the XML info for this phase
--- a/include/cantera/thermo/Phase.h
+++ b/include/cantera/thermo/Phase.h
@@ -707,10 +707,6 @@ public:
     //! Return the Species object for the named species. Changes to this object
     //! do not affect the ThermoPhase object until the #modifySpecies function
     //! is called.
-    /*!
-     * @deprecated To be removed after Cantera 2.5. Replaceable with
-     * Phase::species(size_t) via Phase::speciesIndex(std::string&).
-     */
     shared_ptr<Species> species(const std::string& name) const;
 
     //! Return the Species object for species whose index is *k*. Changes to
@@ -809,15 +805,15 @@ protected:
     //! Flag determining behavior when adding species with an undefined element
     UndefElement::behavior m_undefinedElementBehavior;
 
-    //! Find lowercase species name in m_speciesIndices when case sensitive
-    //! species names are not enforced. Raise exception if lowercase name
-    //! is not unique.
-    size_t findSpeciesLower(const std::string& nameStr) const;
-
     //! Flag determining whether case sensitive species names are enforced
     bool m_caseSensitiveSpecies;
 
 private:
+    //! Find lowercase species name in m_speciesIndices when case sensitive
+    //! species names are not enforced and a user specifies a non-canonical
+    //! species name. Raise exception if lowercase name is not unique.
+    size_t findSpeciesLower(const std::string& nameStr) const;
+
     XML_Node* m_xml; //!< XML node containing the XML info for this phase
 
     //! ID of the phase. This is the value of the ID attribute of the XML
--- a/include/cantera/thermo/Phase.h
+++ b/include/cantera/thermo/Phase.h
@@ -859,6 +859,9 @@ private:
     //! Map of species names to indices
     std::map<std::string, size_t> m_speciesIndices;
 
+    //! Map of lower-case species names to indices
+    std::map<std::string, size_t> m_speciesLower;
+
     size_t m_mm; //!< Number of elements.
     vector_fp m_atomicWeights; //!< element atomic weights (kg kmol-1)
     vector_int m_atomicNumbers; //!< element atomic numbers
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -1257,6 +1257,10 @@ class TestMisc(utilities.CanteraTest):
         gas = ct.Solution('h2o2.xml')
         self.assertFalse(gas.case_sensitive_species_names)
         self.assertTrue(gas.species_index('h2') == 0)
+        gas.X = 'h2:.5, o2:.5'
+        self.assertNear(gas.X[0], 0.5)
+        gas.Y = 'h2:.5, o2:.5'
+        self.assertNear(gas.Y[0], 0.5)
 
         gas.case_sensitive_species_names = True
         self.assertTrue(gas.case_sensitive_species_names)
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -123,7 +123,7 @@ class TestThermoPhase(utilities.CanteraTest):
         self.assertEqual(list(X[0,:,0]), list(Y))
 
     def test_setCompositionString(self):
-        self.phase.X = 'h2:1.0, o2:1.0'
+        self.phase.X = 'H2:1.0, O2:1.0'
         X = self.phase.X
         self.assertNear(X[0], 0.5)
         self.assertNear(X[3], 0.5)
@@ -161,7 +161,7 @@ class TestThermoPhase(utilities.CanteraTest):
         self.assertNear(Y[3], 0.75)
 
     def test_getCompositionDict(self):
-        self.phase.X = 'oh:1e-9, O2:0.4, AR:0.6'
+        self.phase.X = 'OH:1e-9, O2:0.4, AR:0.6'
         self.assertEqual(len(self.phase.mole_fraction_dict(1e-7)), 2)
         self.assertEqual(len(self.phase.mole_fraction_dict()), 3)
 
@@ -202,7 +202,7 @@ class TestThermoPhase(utilities.CanteraTest):
             self.phase.Y = {'H2':1.0, 'O2':'xx'}
 
     def test_setCompositionSlice(self):
-        self.phase['h2', 'o2'].X = 0.1, 0.9
+        self.phase['H2', 'O2'].X = 0.1, 0.9
         X = self.phase.X
         self.assertNear(X[0], 0.1)
         self.assertNear(X[3], 0.9)
@@ -1252,6 +1252,35 @@ class TestMisc(utilities.CanteraTest):
     def test_stringify_bad(self):
         with self.assertRaises(AttributeError):
             ct.Solution(3)
+
+    def test_case_sensitive_names(self):
+        gas = ct.Solution('h2o2.xml')
+        self.assertFalse(gas.case_sensitive_species_names)
+        self.assertTrue(gas.species_index('h2') == 0)
+        
+        gas.case_sensitive_species_names = True
+        self.assertTrue(gas.case_sensitive_species_names)
+        with self.assertRaises(ValueError):
+            gas.species_index('h2')
+        with self.assertRaisesRegex(ct.CanteraError, 'Unknown species'):
+            gas.X = 'h2:1.0, o2:1.0'
+        with self.assertRaisesRegex(ct.CanteraError, 'Unknown species'):
+            gas.Y = 'h2:1.0, o2:1.0'
+
+        gas_cti = """ideal_gas(
+            name="gas",
+            elements=" S C Cs ",
+            species=" nasa: all ",
+            options=["skip_undeclared_elements"],
+            initial_state=state(temperature=300, pressure=(1, "bar"))
+        )"""
+        ct.suppress_thermo_warnings(True)
+        gas = ct.Solution(source=gas_cti)
+        with self.assertRaisesRegex(ct.CanteraError, 'is not unique'):
+            gas.species_index('cs')
+        gas.case_sensitive_species_names = True
+        with self.assertRaises(ValueError):
+            gas.species_index('cs')
 
 
 class TestElement(utilities.CanteraTest):
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -123,7 +123,7 @@ class TestThermoPhase(utilities.CanteraTest):
         self.assertEqual(list(X[0,:,0]), list(Y))
 
     def test_setCompositionString(self):
-        self.phase.X = 'H2:1.0, O2:1.0'
+        self.phase.X = 'h2:1.0, o2:1.0'
         X = self.phase.X
         self.assertNear(X[0], 0.5)
         self.assertNear(X[3], 0.5)
@@ -161,7 +161,7 @@ class TestThermoPhase(utilities.CanteraTest):
         self.assertNear(Y[3], 0.75)
 
     def test_getCompositionDict(self):
-        self.phase.X = 'OH:1e-9, O2:0.4, AR:0.6'
+        self.phase.X = 'oh:1e-9, O2:0.4, AR:0.6'
         self.assertEqual(len(self.phase.mole_fraction_dict(1e-7)), 2)
         self.assertEqual(len(self.phase.mole_fraction_dict()), 3)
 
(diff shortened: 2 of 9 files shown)


speth (changes requested):
I'm :-1: on requiring case sensitivity, at least universally. There is no consistency between reaction mechanisms from different sources, so it's quite convenient to be able to leave compositions such as `'N2: 3.76, O2: 1.0'` alone and have code just work when switching between different mechanisms.

The case from the original issue report is fairly rare -- the only conflicts in `nasa.cti` are `CS`/`Cs` and `CS2`/`Cs2`. The one other scenario where it might be nice to have case sensitivity is if you wanted to use SMILES as identifiers, where lower case is used as an indicator of aromaticity, and so would be needed to differentiate between cyclohexane (`C1CCCCC1`) and benzene (`c1ccccc1`). I am not opposed to a flag to enable case sensitivity, with the default behavior remaining as-is.

Replacing `doublereal` with `double` throughout `Phase.h` and `Phase.cpp` makes it difficult to read the actual changes in this PR, and is the reason I suggest making that change only very locally (i.e. at most, within methods that are already being modified).

ischoegl:
Thanks @speth for the feedback.

>
> There is no consistency between reaction mechanisms from different sources, so it's quite convenient to be able to leave compositions such as `'N2: 3.76, O2: 1.0'` alone and have code just work when switching between different mechanisms.

Fair enough. I changed the logic to retain this ability and removed warnings, while still storing case sensitive species names in the C++ layer.

> 
> The case from the original issue report is fairly rare -- the only conflicts in `nasa.cti` are `CS`/`Cs` and `CS2`/`Cs2`. The one other scenario where it might be nice to have case sensitivity is if you wanted to use SMILES as identifiers, where lower case is used as an indicator of aromaticity, and so would be needed to differentiate between cyclohexane (`C1CCCCC1`) and benzene (`c1ccccc1`). I am not opposed to a flag to enable case sensitivity, with the default behavior remaining as-is.

I believe both are reasons for enabling case sensitive species? Imho it should be possible to pull rudimentary data from nasa for any of the stored species (and it is conceivable that the data base will contain more cases in the future). And, yes, the ability to use SMILES would be neat. (So would be aliases for species, but that should be kept for another PR).

> 
> Replacing `doublereal` with `double` throughout `Phase.h` and `Phase.cpp` makes it difficult to read the actual changes in this PR, and is the reason I suggest making that change only very locally (i.e. at most, within methods that are already being modified).

I kept this in a single commit, so it was quite easy to undo. It probably should nevertheless be done if the proposed PR moves forward?

The one use case this PR does not pre-empt is if one were to cut&paste bits and pieces from mechanism that use different uppercase/lowercase conventions. Since there will always be differences in nomenclature, this is imho not a major issue, as one must tread carefully in those attempts no matter what.

src/thermo/Phase.cpp
```diff
@@ -794,19 +839,39 @@ void Phase::modifySpecies(size_t k, shared_ptr<Species> spec)
...
     invalidateCache();
 }
 
 shared_ptr<Species> Phase::species(const std::string& name) const
 {
-    return m_species.at(toLowerCopy(name));
+    try {
+        return m_species.at(name);
+    } catch (std::out_of_range&) {
+        if (m_caseSensitiveSpecies) {
```
ischoegl: Not sure whether catching this (presumably rare?) case is worth the effort.

include/cantera/thermo/Phase.h
```diff
@@ -707,6 +707,10 @@ class Phase
     //! Return the Species object for the named species. Changes to this object
     //! do not affect the ThermoPhase object until the #modifySpecies function
     //! is called.
+    /*!
+     * @deprecated To be removed after Cantera 2.5. Replaceable with
+     * Phase::species(size_t) via Phase::speciesIndex(std::string&).
+     */
```
speth: I don't think this method should be deprecated.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -161,7 +161,7 @@ def test_setCompositionDict(self):
         self.assertNear(Y[3], 0.75)
 
     def test_getCompositionDict(self):
-        self.phase.X = 'oh:1e-9, O2:0.4, AR:0.6'
+        self.phase.X = 'OH:1e-9, O2:0.4, AR:0.6'
```
speth: I would prefer to leave these tests as-is -- they are written using different cases to confirm that case insensitive matching is working.

src/thermo/Phase.cpp
```diff
@@ -699,7 +743,8 @@ size_t Phase::addElement(const std::string& symbol, doublereal weight,
 }
 
 bool Phase::addSpecies(shared_ptr<Species> spec) {
-    if (m_species.find(toLowerCopy(spec->name)) != m_species.end()) {
+    // species names are case sensitive
+    if (m_species.find(spec->name) != m_species.end()) {
```
speth, line 734: This comment seems unnecessary -- it's basically describing the contrast to the previous behavior.

src/thermo/Phase.cpp
```diff
@@ -172,20 +173,47 @@ void Phase::getAtoms(size_t k, double* atomArray) const
...
 
+size_t Phase::findSpeciesLower(const std::string& name) const
+{
+    size_t loc = npos;
+    std::string nLower = toLowerCopy(name);
+    for (const auto& k : m_speciesIndices) {
+        if (toLowerCopy(k.first) == nLower) {
+            if (loc == npos) {
+                loc = k.second;
+            } else {
```
speth: Rather than falling back to an O(N) search anytime a non-canonical species name is provided, I wonder if the following might be a cleaner solution:
- Add a map of lower-case species names to species indices (say, `m_speciesIndicesLower`)
- When adding species, if there is already an entry in this map with the same name, set its value to `npos`, and possibly set the `m_caseSensitiveSpecies` flag.
- The implementation of `speciesIndex` can then use either `m_speciesIndicesLower` or `m_speciesIndices` depending on the value of `m_caseSensitiveSpecies`. If `m_speciesIndicesLower` is checked and contains `npos`, raise the exception about non-unique species identifiers.

src/thermo/Phase.cpp
```diff
@@ -294,10 +322,18 @@ void Phase::setMoleFractionsByName(const compositionMap& xMap)
...
+            size_t loc = npos;
+            if (!m_caseSensitiveSpecies) {
+                loc = findSpeciesLower(sp.first);
+            }
+            if (loc == npos) {
+                throw CanteraError("Phase::setMoleFractionsByName",
+                                   "Unknown species '{}'", sp.first);
+            } else {
+                mf[loc] = sp.second;
+            }
```
speth: Pending the suggested implementation change to `speciesIndex`, I think this could be simplified to just:
```c++
size_t k = speciesIndex(sp.first);
if (k != npos) {
    mf[loc] = sp.second;
} else {
    throw CanteraError(...);
}
```

src/thermo/RedlichKwongMFTP.cpp
```diff
@@ -710,6 +710,8 @@ vector<double> RedlichKwongMFTP::getCoeff(const std::string& iName)
     // based on crit  properties T_c and P_c:
     for (size_t isp = 0; isp < nDatabase; isp++) {
         XML_Node& acNodeDoc = doc->child(isp);
+        // not enforcing case sensitive species names as this is external
+        // to CTI or YAML input files
```
speth: I generally find comments about what isn't being done to be unhelpful after the fact.

speth (changes requested):
Thanks for the updates. I think this change will provide a concrete benefit, by allowing either the existing case-insensitive behavior in most cases but permitting case sensitivity to be activated in the instances where it is necessary.

> I kept this in a single commit, so it was quite easy to undo. It probably should nevertheless be done if the proposed PR moves forward?

As much as I wish the `doublereal` typedef had never been introduced, I think the costs of rooting it out broadly (beyond the scope of lines of code that are changing for other reasons) exceed the benefits. In addition to making diffs hard to read, it is very easy for these changes to make it hard to rebase/merge other branches that may have meaningful changes to these lines.

> The one use case this PR does not pre-empt is if one were to cut&paste bits and pieces from mechanism that use different uppercase/lowercase conventions. Since there will always be differences in nomenclature, this is imho not a major issue, as one must tread carefully in those attempts no matter what.
I'm not too worried about that, and I think at least some of those cases will work with the YAML format (less so in the CTI case).

include/cantera/thermo/Phase.h
ischoegl: As far as I recall, it's no longer used within the core, but I will leave it in.

src/thermo/Phase.cpp
ischoegl: I am not sure about this suggestion. The second (lower-case) species map would contain nearly identical information; in cases like this, I usually keep the information in a single representation, as it is easier to maintain.

While the suggested implementation would be faster *if a user opts to use species names with names other than those specified in the input file*, I doubt that it has significant real-world benefits. Assuming that those searches are typically done during problem setup, there is little computational overhead (in repeated operations, it is _always_ faster to select species by index). I am also unsure about auto-magically flipping flags, as those decisions may be best left to the user? Weighing cost and benefit, I am suggesting to leave the implementation as is.

src/thermo/Phase.cpp
ischoegl: see above: `findSpeciesLower` is used as a fallback solution.

src/thermo/Phase.cpp
```diff
@@ -794,24 +838,24 @@ void Phase::modifySpecies(size_t k, shared_ptr<Species> spec)
...
     }
-    m_species[toLowerCopy(spec->name)] = spec;
+    m_species[spec->name] = spec;
     invalidateCache();
 }
 
 shared_ptr<Species> Phase::species(const std::string& name) const
 {
-    return m_species.at(toLowerCopy(name));
+    return m_species.at(name);
```
speth: This method is currently case-sensitive, regardless of the setting of `m_caseSensitiveSpecies`. I think resolving this will require another instance of the try/catch structure that you have added elsewhere.

src/thermo/Phase.cpp
speth: True, there is a benefit to not introducing a new data structure. However, avoiding that requires several instances of this kind of convoluted try/catch construct (with I think another one needed in `Phase::species(string)`.

I think using the extra data structure would allow us to avoid adding `findSpeciesLower` as a new method -- its logic could just be in `speciesIndex`.

There's also some value in not generating exceptions for circumstances that aren't really exceptional (at least in C++). I'm less concerned about the performance implication than with the ability to run in gdb and break at an error by simply doing `catch throw`. If we use exceptions in places like this, then you end up spending a lot more time continuing past these unexceptional exceptions.

I would also note that `speciesIndex` is actually called with a name not in `m_speciesIndices` is more often than you might think -- it's not just instances of case-mismatched species names. It is used extensively in multiphase systems as a way of finding which phase a species belongs to. See for example `Kinetics::kineticsSpeciesIndex` and `Kinetics::speciesPhase`.

My thought on automatically enabling case sensitivity when there are species that can only be resolved this way was that that might be less surprising to the user than having to enable it explicitly. There may also be issues with initializing phases with species that are only distinguishable by case. For instance, what if a reaction is specified (in the input file) that contains one of these species? Adding the reaction requires looking up the indices of the involved species, and this needs to happen before the user has the chance to set the case sensitive species flag.

speth (changes requested):

src/thermo/Phase.cpp
ischoegl: @speth on second thought, the suggested solution will work regardless of the implementation. Adopting ... :thumbsup:

src/thermo/Phase.cpp
ischoegl: Similar to the other case, the logic can be delegated to `speciesIndex`, i.e.
```C++
    return speciesIndex(name);
```

ischoegl: Good catch - I overlooked this one. This involves a second map object (`m_species`), and requires an analogous handling as `m_speciesIndices`.

src/thermo/Phase.cpp
ischoegl: @speth ... thanks for your comments (plus catching the other access-species-by-name case for `Phase::species`). I.e. this involves another (almost) duplicate map. 

Your comment regarding C++ exceptions makes a lot of sense. In the current revision, I opted to replace `std::map::at` by `std::map::find`, which eliminates the `try` - `catch` blocks. 

I do realize that the kinetics managers make extensive use of `speciesIndex` (they have to), but in most cases the logic will employ case-sensitive species without additional overhead.

Regarding the automatic flags: my thought was that in most cases, case sensitive species are used by default _internally_ and the fallback solution is rarely used. I updated the error message to explicitly direct users towards the `caseSensitiveSpecies` flag, which can be employed after all species and reactions are set up: internally, everything is case sensitive, i.e. lowercase names are treated as fall-back solutions.

Finally, at this point the case-sensitive/lowercase logic is exclusively handled within `Phase::speciesIndex` and `Phase::species`, i.e. there is no longer any duplication of logic blocks.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -483,7 +490,8 @@ cdef class ThermoPhase(_SolutionBase):
 
         s = Species(init=False)
         if isinstance(k, (str, bytes)):
-            s._assign(self.thermo.species(stringify(k)))
+            kk = self.thermo.speciesIndex(stringify(k))
+            s._assign(self.thermo.species(<int>kk))
```
speth: So if we're not deprecating `ThermoPhase::species(string)`, can we change this back so that the error message when the specified species name is about the species name being invalid, rather than the index being out of bounds?

src/thermo/Phase.cpp
```diff
@@ -338,10 +370,18 @@ void Phase::setMassFractionsByName(const compositionMap& yMap)
...
     for (const auto& sp : yMap) {
         try {
-            mf[m_speciesIndices.at(toLowerCopy(sp.first))] = sp.second;
+            mf[m_speciesIndices.at(sp.first)] = sp.second;
         } catch (std::out_of_range&) {
-            throw CanteraError("Phase::setMassFractionsByName",
-                               "Unknown species '{}'", sp.first);
+            size_t loc = npos;
+            if (!m_caseSensitiveSpecies) {
+                loc = findSpeciesLower(sp.first);
```
speth: This should use `speciesIndex`, the same as in `setMoleFractionsByName`, right?

src/thermo/Phase.cpp
```diff
@@ -794,24 +834,50 @@ void Phase::modifySpecies(size_t k, shared_ptr<Species> spec)
...
+                                       "Set Phase::caseSensitiveSpecies to true to "
+                                       "enforce case sensitive species names", nLower);
+                }
+            }
+        }
+        return sp;
+    } else {
+        throw CanteraError("Phase::setMassFractionsByName",
+                           "Unknown species '{}'", name);
+    }
```
speth, line 854: Isn't there a simpler implementation of this method that leans on `speciesIndex` much like `setMoleFractionsByName`? Something like:
```c++
size_t k = speciesIndex(name);
if (k != npos) {
    return m_species.at(speciesName(k));
} else {
    throw CanteraError(...);
}
```
Or is there some other complexity here that I'm missing?

src/thermo/Phase.cpp
speth: Regarding the "automatic flags", I had to play around with this a bit to realize how this was now working - an exception is only thrown when case sensitivity is disabled and you specify a name which is not a case-sensitive match for any species. So my concern about setting up reactions is moot, and I agree that there is no need to automatically toggle this flag.

I still think the `speciesIndex` calls within the Kinetics managers for interface phases will frequently incur the O(N) overhead. For example, if you have a kinetics manager containing a surface, a gas, and a solid phase in that order, looking up the index of a species in the solid phase, even with the correct case, will cause a search through all the surface phase species, then all the gas phase species, before finally taking the easy path for the index into the solid phase.

src/thermo/Phase.cpp
```diff
@@ -172,20 +173,49 @@ void Phase::getAtoms(size_t k, double* atomArray) const
...
+}
+
 size_t Phase::speciesIndex(const std::string& nameStr) const
 {
-    size_t loc = getValue(m_speciesIndices, toLowerCopy(nameStr), npos);
+    size_t loc = npos;
+    std::map<std::string, size_t>::const_iterator it;
+
+    it = m_speciesIndices.find(nameStr);
+    if (it != m_speciesIndices.end()) {
```
speth, line 200: I think this is a good place to use `auto` instead of writing out the complex iterator type.

speth (changes requested):
Thanks for the additional updates. I like having the calls to `findSpeciesLower` isolated within `speciesIndex`, and would like to see that through to the end.

I'm still concerned about the effect on setting up interface kinetics systems -- without the auxiliary data structure, I think this turns into a O(n_species * n_reactions) computation, which can actually slow things down.

src/thermo/Phase.cpp
ischoegl: Correct. I missed this one ... resulting in a tremendous simplification :grin:

src/thermo/Phase.cpp
ischoegl: Didn't think of that one!

src/thermo/Phase.cpp
ischoegl: :sparkles: ... didn't know that one.

ischoegl:
@speth ... thanks for the additional comments. I feel this is converging (also caught another case - `setMassFractionByName` that a unit test for some reason hadn't flagged before) and applied your recommendations.

> an exception is only thrown when case sensitivity is ~dis~ _en_ abled and you specify a name which is not a case-sensitive match for any species. So my concern about setting up reactions is moot, and I agree that there is no need to automatically toggle this flag.

:thumbsup: That was exactly my intent (case sensitivity is disabled by default)

> if you have a kinetics manager containing a surface, a gas, and a solid phase in that order, looking up the index of a species in the solid phase, even with the correct case, will cause a search through all the surface phase species, then all the gas phase species, before finally taking the easy path for the index into the solid phase.

This is a legitimate point. I likewise do not see a way around the auxiliary map. I'll add a O(log(N)) fallback option in the next commit.

ischoegl:
Done. There now is a parallel `m_SpeciesLower` map, which allows for a O(log(N)) search in all cases.

speth (approved):
Looks good to me. Thanks for convincing me that this was a worthwhile change, and the willingness to iterate on the implementation details.

ischoegl:
NP. Your reviews are well worth the effort. Also good to know about #709 ...

Issue #570, opened 2018-11-07
Feature Request: support of pickle serialization of Cython objects in python API


chengdi123000:
Serialization is used to transfer objects between processes, nodes, or even between windows and linux. It is also widely used in parallel program design. However, the cython objects such as ct.Solution and ct.Reaction cannot be serialized due to lack of serialization support. 

refer to https://snorfalorpagus.net/blog/2016/04/16/pickling-cython-classes/. It seems  simple to support serialization in python using pickle.

I tried to use monkey patching to assign `__reduce__()` method to Reaction object, but it is forbidden by python. So I think it is better to be implemented in cython layer. 

Think about this scenario: I can save the Solution object or even FreeFlame object to a *.pkl file in Linux and send it to someone or forum. The receiver can simply load it in Windows and help me to check what is going wrong.

MicaelBoulet:
Interesting application example, but running parallel script might be the true motivation for this. Like large parametric study on multicore system or cluster. There is currently a way around for this, but call for more effort and that would be neat to have support for pickle.

arghdos:
@MicaelBoulet agreed -- pickle support would be huge for multiprocessing (and much more efficient than having to reinitialize your gas object on every new process).  I think the issue may lie deeper however -- wouldn't pickling require implementations of copy constructors on the underlying C++ objects (or at the very least, the implementation of a C++ `__getstate__` and `__setstate__`-like interface)?

speth:
I think serialization for storage and portability is distinct from serialization for parallel processing. In the latter case, you want to limit the data being transferred. In the instance of a `Solution` object in Python, this is already achieved by using the (temperature, density, mass fractions) vector available as `phase.state`. If you actually transmitted a more complete representation, e.g. all of the thermo and reaction data, I think that would probably be comparable to constructing the object from an input file, which you really want to do only once per thread/process.

The C++ copy constructor doesn't really provide any utility here, since it just creates another instance of an object in the same process, not one that can be saved or transmitted.

I would also say that the original suggested use of serialization here isn't a great application -- If you're having a problem, the interesting thing isn't the end state but how you got there, for which you need the code that produced that state.

bryanwweber:
[...]

PR #692, opened 2019-08-13
Pickle serialization of Solution objects

### LINKED ISSUE #10: Support of pickle serialization of Cython objects in python API
Serialization is used to transfer objects between processes, nodes, or even between windows and linux. It is also widely used in parallel program design. However, the cython objects such as ct.Solution and ct.Reaction cannot be serialized due to lack of serialization support. 

refer to https://snorfalorpagus.net/blog/2016/04/16/pickling-cython-classes/. It seems  simple to support serialization in python using pickle.

I tried to use monkey patching to assign `__reduce__()` method to Reaction object, but it is forbidden by python. So I think it is better to be implemented in cython layer. 

Think about this scenario: I can save the Solution object or even FreeFlame object to a *.pkl file in Linux and send it to someone or forum. The receiver can simply load it in Windows and help me to check what is going wrong.

ischoegl:
Pickle support for `Solution` objects would facilitate __multiprocessing applications__, where each thread/process is working with a separate copy of a Solution object. The approach is comparable to constructing the object from an input file plus setting phase and state vector.

Closes Cantera/enhancements#10

~The PR illustrates a __rudimentary implementation__ (XML/CTI support only), which *should not be considered for adoption* (at least not in its present form).~ … **now replaced by a permanent solution using YAML serialization**

It allows for the following scenario (obviously, multiprocessing applications are more relevant):
```python
import cantera as ct
import pickle

gas = ct.Solution('h2o2.cti', 'ohmech-multi')
gas.TPX = 500, 500000, 'H2:.75, O2:.25'
pickle.dump(gas, open('gas.pkl', 'wb'))
```
Recreating the object in a separate session via
```python
import pickle
gas = pickle.load(open('gas.pkl','rb'))
gas()
```
yields:
```
 ohmech-multi:

       temperature             500  K
          pressure          500000  Pa
           density         1.14397  kg/m^3
  mean mol. weight          9.5115  amu

                          1 kg            1 kmol
                       -----------      ------------
          enthalpy      6.2484e+05        5.943e+06     J
   internal energy      1.8776e+05        1.786e+06     J
           entropy           16391        1.559e+05     J/K
    Gibbs function     -7.5708e+06       -7.201e+07     J
 heat capacity c_p          3127.2        2.974e+04     J/K
 heat capacity c_v            2253        2.143e+04     J/K

                           X                 Y          Chem. Pot. / RT
                     -------------     ------------     ------------
                H2           0.75         0.158965         -14.8055
                O2           0.25         0.841035           -24.87
     [   +7 minor]              0                0

```
~The current implementation for pickling/unpickling makes use of the static XML tree that is retained after creation of a `Solution` from CTI/XML files. A more useful (but significantly more involved) implementation would generate YAML strings dynamically.~

- ~N/A: at least at this point, the PR serves for illustration purposes only~

Known issues:
- ~The current implementation will fail if the `Solution` object is changed after instantiation (e.g. species are added, etc.)~
- ~Unpickling requires that an empty constructor, i.e. `_SolutionBase()` is allowed. While this works fine for `pickle.load`, an attempt for `gas = Solution()` results in a segmentation fault.~ ... *fixed*

### DIFF
--- a/interfaces/cython/cantera/base.pyx
+++ b/interfaces/cython/cantera/base.pyx
@@ -18,6 +18,9 @@ cdef class _SolutionBase:
                         thermo=thermo, species=species, kinetics=kinetics,
                         reactions=reactions, **kwargs)
             return
+        elif any([infile, source, adjacent, origin, source, yaml,
+                  thermo, species, kinetics, reactions, kwargs]):
+            raise ValueError("Arguments are insufficient to define a phase")
 
         self._base = CxxNewSolution()
         self.base = self._base.get()
@@ -89,10 +92,8 @@ cdef class _SolutionBase:
         # Parse inputs
         if infile or source:
             self._init_cti_xml(infile, name, adjacent, source)
-        elif thermo and species:
-            self._init_parts(thermo, species, kinetics, adjacent, reactions)
         else:
-            raise ValueError("Arguments are insufficient to define a phase")
+            self._init_parts(thermo, species, kinetics, adjacent, reactions)
 
         self._selected_species = np.ndarray(0, dtype=np.uint64)
 
--- a/interfaces/cython/cantera/base.pyx
+++ b/interfaces/cython/cantera/base.pyx
@@ -283,14 +283,14 @@ cdef class _SolutionBase:
         """
         self.base.header().clear()
 
-    def write_yaml(self, filename, phases=None, units=None, precision=None,
+    def write_yaml(self, filename=None, phases=None, units=None, precision=None,
                    skip_user_defined=None, header=True):
         """
         Write the definition for this phase, any additional phases specified,
         and their species and reactions to the specified file.
 
         :param filename:
-            The name of the output file
+            The name of the output file; if ``None``, a YAML string is returned
         :param phases:
             Additional ThermoPhase / Solution objects to be included in the
             output file
@@ -328,6 +328,8 @@ cdef class _SolutionBase:
             Y.precision = precision
         if skip_user_defined is not None:
             Y.skip_user_defined = skip_user_defined
+        if filename is None:
+            return Y.to_string()
         Y.to_file(filename)
 
     def __getitem__(self, selection):
--- a/interfaces/cython/cantera/base.pyx
+++ b/interfaces/cython/cantera/base.pyx
@@ -9,6 +9,18 @@ cdef class _SolutionBase:
                   source=None, yaml=None, thermo=None, species=(),
                   kinetics=None, reactions=(), **kwargs):
 
+        # run instantiation only if valid sources are specified
+        if origin or infile or source or yaml or (thermo and species):
+
+            self._cinit(infile=infile, name=name, adjacent=adjacent,
+                        origin=origin, source=source, yaml=yaml,
+                        thermo=thermo, species=species, kinetics=kinetics,
+                        reactions=reactions, **kwargs)
+
+    def _cinit(self, infile='', name='', adjacent=(), origin=None,
+                  source=None, yaml=None, thermo=None, species=(),
+                  kinetics=None, reactions=(), **kwargs):
+
         if 'phaseid' in kwargs:
             if name is not '':
                 raise AttributeError('duplicate specification of phase name')
@@ -351,8 +363,18 @@ cdef class _SolutionBase:
             for i,spec in enumerate(species):
                 self._selected_species[i] = self.species_index(spec)
 
-    def __reduce__(self):
-        raise NotImplementedError('Solution object is not picklable')
+    def __getstate__(self):
+        """Save complete information of the current phase for pickling."""
+        if self.kinetics.nTotalSpecies() > self.thermo.nSpecies():
+            raise NotImplementedError(
+                "Pickling of Interface objects is not implemented.")
+        return self.write_yaml(), self.state
+
+    def __setstate__(self, pkl):
+        """Restore Solution from pickled information."""
+        yml, state = pkl
+        self._cinit(yaml=yml)
+        self.state = state
 
     def __copy__(self):
         raise NotImplementedError('Solution object is not copyable')
--- a/interfaces/cython/cantera/base.pyx
+++ b/interfaces/cython/cantera/base.pyx
@@ -10,6 +10,7 @@ cdef class _SolutionBase:
                   kinetics=None, reactions=(), **kwargs):
 
         # run instantiation only if valid sources are specified
+        self._references = None
         if origin or infile or source or yaml or (thermo and species):
 
             self._cinit(infile=infile, name=name, adjacent=adjacent,
--- a/interfaces/cython/cantera/base.pyx
+++ b/interfaces/cython/cantera/base.pyx
@@ -16,6 +16,19 @@ cdef class _SolutionBase:
                         origin=origin, source=source, yaml=yaml,
                         thermo=thermo, species=species, kinetics=kinetics,
                         reactions=reactions, **kwargs)
+            return
+
+        self._base = CxxNewSolution()
+        self.base = self._base.get()
+        self.base.setSource(stringify("none"))
+
+        self.base.setThermo(newThermo(stringify("none")))
+        self.thermo = self.base.thermo().get()
+        self.base.setKinetics(newKinetics(stringify("none")))
+        self.kinetics = self.base.kinetics().get()
+        self.base.setTransport(newTransport(NULL, stringify("none")))
+        self.transport = self.base.transport().get()
+        self._selected_species = np.ndarray(0, dtype=np.uint64)
 
     def _cinit(self, infile='', name='', adjacent=(), origin=None,
                   source=None, yaml=None, thermo=None, species=(),
--- a/interfaces/cython/cantera/base.pyx
+++ b/interfaces/cython/cantera/base.pyx
@@ -384,15 +384,12 @@ cdef class _SolutionBase:
         if self.kinetics.nTotalSpecies() > self.thermo.nSpecies():
             raise NotImplementedError(
                 "Pickling of Interface objects is not implemented.")
-        transport_model = pystr(self.transport.transportType())
-        return self.write_yaml(), self.state, transport_model
+        return self.write_yaml()
 
     def __setstate__(self, pkl):
         """Restore Solution from pickled information."""
-        yml, state, transport_model = pkl
-        self._cinit(yaml=yml, transport_model=transport_model)
-
-        self.state = state
+        yml = pkl
+        self._cinit(yaml=yml)
 
     def __copy__(self):
         raise NotImplementedError('Solution object is not copyable')
--- a/interfaces/cython/cantera/base.pyx
+++ b/interfaces/cython/cantera/base.pyx
@@ -177,6 +177,9 @@ cdef class _SolutionBase:
             self.base.setKinetics(newKinetics(stringify("none")))
         self.kinetics = self.base.kinetics().get()
 
+        # Transport
+        self.transport = self.base.transport().get()
+
     def _init_cti_xml(self, infile, name, adjacent, source):
         """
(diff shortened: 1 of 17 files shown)


speth:
As I mentioned in the thread on #570, implementation of YAML-based serialization is planned, which will work for objects regardless of how they were created and how they have been manipulated.

With the introduction of the YAML format, deprecation and removal of the CTI and XML formats is also planned, so I wouldn't recommend putting any significant effort into features based around these formats. Beyond that, I'm not sure whether there's a reason to keep this PR open.

ischoegl:
Agreed. I simply wanted to supply an example for pickling (which was surprisingly simple as the XML tree was available - 40ish lines of code!). As I wrote

> The PR illustrates a rudimentary implementation (XML/CTI support only), which should not be considered for adoption (at least not in its present form).

I’d still like to keep this PR open until the core of the code can be switched over to YAML. Ultimately, I hope that this feature can be implemented. See also #696, which i believe is a necessary step in that direction.

ischoegl:
Linking #984, which will create the basis for a permanent implementation.

ischoegl:
:tada: ... unsurprisingly, thanks to @speth's recent addition of #984, this became even simpler.

**PS:** Enabling pickling of `SolutionArray` looks quite feasible, but I'd rather wait for comments at the moment.

tsikes:
This would be a wonderful addition. I ran into the problem of not being able to do this when I first wanted to use multiprocessing to speed up calculations for optimization. I know that others have had the same issue based on the Cantera google group. The current method of using Cantera in a multiprocessing application is rather unintuitive and finicky. This PR would greatly simplify this task for other users.

ischoegl:
Kicking this back to draft - the implementation depends on the ability to create a `Solution` that is unconfigured, so it can be overwritten by `__setstate__`. What is implemented thus far works fine for the `pickle` case, but deactivates errors for calls of  `Solution()` without arguments. Having the `_SolutionBase` in a safe state (i.e. no Segfaults) will take some more work.

**PS:** I probably won't work on this unless I hear back from one of the @Cantera/committers, i.e. comments are explicitly welcome.

speth:
I think it should be possible to instantiate an object of the `ThermoPhase` base class if no other information is available to instantiate the phase, like we do with the `Kinetics` and `Transport` objects. In that case, you should just get a phase with no species where most of the methods raise `NotImplementedError`, which would probably be fine.

I just tried using this in a modified version of the `multiprocessing` example, and it seems to work better than I had initially expected. If you use `itertools.repeat` to pass the `Solution` object as part of the iterable used by `Pool.starmap` for example, the default chunking behavior results in the `Solution` being pickled and unpickled 4 times for each worker process, which is not as bad as I had feared (I thought it might end up doing it once for each iteration). So given the easier user experience, I'll withdraw my earlier objection (Cantera/enhancements#10) to implementing this feature at all.

I'm glad the implementation needed here for standalone `Solution` objects is pretty straightforward. I'm not sure what's going to happen when you try to serialize a an `Interface` phase where you have to correctly link the neighboring phases.

ischoegl:
> I think it should be possible to instantiate an object of the `ThermoPhase` base class if no other information is available to instantiate the phase, like we do with the `Kinetics` and `Transport` objects. 

@speth Could you clarify? None of these are currently instantiable without arguments. E.g. (using `main`)
```
In [3]: kin = ct.Kinetics()
<ipython-input-3-6f9f3674fb0c> in <module>()
----> 1 kin = ct.Kinetics()

/src/interfaces/cython/cantera/base.pyx in cantera._cantera._SolutionBase.__cinit__()
     64             self._init_parts(thermo, species, kinetics, adjacent, reactions)
     65         else:
---> 66             raise ValueError("Arguments are insufficient to define a phase")
     67 
     68         # Initialization of transport is deferred to Transport.__init__

ValueError: Arguments are insufficient to define a phase
```
As far as I can tell, all of these are limitations of the implementation of the Python interface, and not necessarily about the underlying C++ structure. The crux here is that half of the time, the required `Cxx*` references remain uninitialized, which leads to segfaults left and right.

speth:
Sorry for the confusion. What I meant was setting the `thermo` attribute of a `_SolutionBase` object to point to an instance of the base class that has been instantiated, e.g. the equivalent of `new ThermoPhase()`. This would be similar to how we never leave the `_SolutionBase.kinetics` or `_SolutionBase.transport` members as null pointers after initialization (at least for objects that derive from the corresponding Python classes), but have them point to objects of these base classes.

Probably the easiest way to do this would be to add this as a case for `ThermoFactory`, the same way that using `KineticsFactory` to create an object with the model `none` just instantiates the base class, at which point the existing `_init_parts` method would just about work (although I think we'd still need to require species input except when the thermo model is none).

ischoegl:
@speth ... now that #1129 is merged, I believe this is finally ready to be resolved (this was one of the first PR's I ever submitted!). Beyond the YAML serialization, the main sticking point was that the ability to pickle requires the creation of empty `Solution` objects (i.e. `ct.Solution()`), which needs to be safe when called outside of pickling - prior to #1129 it was hard to prevent segfaults.

At this point, it is possible to create empty `ct.ThermoPhase()`, `ct.Kinetics()` and `ct.Transport()` objects, all of which are backed by default C++ objects of type `"None"` (i.e. the base classes), which required minor updates outside of Python. One thing that I disabled is pickling of `Interface` objects, where I believe it requires a resolution of Cantera/enhancements#103 to come up with a good solution.

speth (changes requested):
I tried this for a serialized `gri30.yaml` object, and got segfaults when trying to access any transport properties -- it seems like the transport object is not being set up correctly when restoring. I was a little surprised that this didn't come up in the test suite.

ischoegl:
@speth ... interesting indeed. Could you post the code you used? (*Edit: I may have overanalyzed the 'serialized' here. If it's just accessing the restored object I can probably take it from here*)

speth:
Yes, that's all I did:
```python
gas0 = ct.Solution('gri30.yaml')
with open('test.pkl', 'wb') as out:
    pickle.dump(gas0, out)

with open('test.pkl', 'rb') as infile:
    gas1 = pickle.load(infile)

print(gas1.thermo_model)
print(gas1.kinetics_model)
print(gas1.transport_model)  # segfaults here
```

ischoegl:
@speth ... fixed. The bug was hiding in plain sight ... one of the derived class constructor that Python calls (`Transport.__init__`) usually takes care of setting the C++ transport object. That this doesn't happen here is obvious in hindsight ... it's probably a good idea to consolidate all those calls in `base.pyx` once CTI/XML are gone.

I also updated the unit test to ensure that the model is properly set (including `"None"`, which is not the default).

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -116,6 +117,95 @@ def check(a, b):
...
+        utilities.CanteraTest.setUpClass()
+        cls.gas = ct.ThermoPhase()
+
+    def test_empty_report(self):
+        with self.assertRaisesRegex(ct.CanteraError, "NotImplementedError"):
+            self.gas()
+
+    def test_empty_TPX(self):
+        with self.assertRaisesRegex(ct.CanteraError, "NotImplementedError"):
+            self.gas.TPX = 300, ct.one_atm, ""
```
speth: Setting mole fractions to the empty string wouldn't be valid even if this wasn't an empty `ThermoPhase` (though I suppose it would be a different exception)

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -116,6 +117,95 @@ def check(a, b):
...
+        utilities.CanteraTest.setUpClass()
+        cls.gas = ct.Solution()
+
+    def test_empty_composite(self):
+        self.assertEqual(self.gas.thermo_model, "None")
+        self.assertEqual(self.gas.composite, ("None", "None", "None"))
+
+    def test_empty_kinetics(self):
+        kin = ct.Kinetics()
+        # an empty kinetics object is functional
```
speth: Shouldn't this use the `self.gas` object created in `setUpClass`, rather than create a new empty `Kinetics` object? That would seem more consistent with the name of the `TestEmptySolution` class. Ditto for `test_empty_transport` below.

interfaces/cython/cantera/base.pyx
```diff
@@ -9,6 +9,31 @@ cdef class _SolutionBase:
...
+        if origin or infile or source or yaml or (thermo and species):
+
+            self._cinit(infile=infile, name=name, adjacent=adjacent,
+                        origin=origin, source=source, yaml=yaml,
+                        thermo=thermo, species=species, kinetics=kinetics,
+                        reactions=reactions, **kwargs)
+            return
+
+        self._base = CxxNewSolution()
+        self.base = self._base.get()
```
speth: I think this initialization logic misses out on some of the error checking that happened in the old `__cinit__` (and I guess is still there but inaccessible). Ideally, the _only_ new cases allowed by this PR would be:
```python
ct.Solution()
ct.ThermoPhase()
ct.Kinetics()
ct.Transport()
```
But right now, you can do things like
```python
ct.ThermoPhase(thermo='ideal-gas')
ct.Kinetics(kinetics='gas')
```
Both of which still just return objects of the base class. Previously, this would have raised an exception, namely
```
ValueError: Arguments are insufficient to define a phase
```

interfaces/cython/cantera/base.pyx
```diff
@@ -349,8 +379,20 @@ cdef class _SolutionBase:
...
 
-    def __reduce__(self):
-        raise NotImplementedError('Solution object is not picklable')
+    def __getstate__(self):
+        """Save complete information of the current phase for pickling."""
+        if self.kinetics.nTotalSpecies() > self.thermo.nSpecies():
+            raise NotImplementedError(
+                "Pickling of Interface objects is not implemented.")
+        transport_model = pystr(self.transport.transportType())
+        return self.write_yaml(), self.state, transport_model
```
speth: The YAML definition already includes the state information, and also the transport model. Is there a reason to duplicate that in the pickle? Thanks to some other changes in this PR, the following works and matches both properties:
```python
gas2 = ct.Solution(yaml=gas.write_yaml())
```

speth (changes requested):
Thanks, @ischoegl. I'm glad to see this working based on YAML serialization. I had just a few comments on the implementation.

interfaces/cython/cantera/test/test_composite.py
ischoegl: The idea was to pre-empt this with a more tailored error message here. Changed to a `TP` setter, which should clarify the intent.

interfaces/cython/cantera/test/test_composite.py
ischoegl: No. The `setUpClass` is used to override the `setUpClass` of the parent class so some tests can be 'recycled'. The other tests are added here as they only need to be run once. Tests are changed also, as empty `Kinetics` and `Transport` are no longer allowed. *edit: created a new test class to clarify*

interfaces/cython/cantera/base.pyx
ischoegl: Ha. good catch!

interfaces/cython/cantera/base.pyx
ischoegl: Another good catch. I decided to not allow for 
```python
ct.Kinetics()
ct.Transport()
```
for the time being (I cannot think of a case where `ct.Transport()` makes sense, with the same being true for the current way `ct.Kinetics()` are being handled).

I moved the old check so it again cannot be bypassed.

ischoegl:
@speth … thank you for the careful review! I believe I took care of everything. I ended up disabling `Kinetics()` and `Transport()` as I don’t see a current use case (it was surprisingly hard to catch these due to the `__slots__`). I also clarified some of the edge cases you pointed out.

speth (approved):
Thanks for the updates. This looks good to me.

Issue #652, opened 2019-06-15
Verify, catalog, and document Cantera ThermoPhase classes.


decaluwe:
Cantera's [ThermoPhase class list](https://github.com/Cantera/cantera/tree/master/src/thermo)  contains approximately 30-some classes, representing a range of phase-types. However, these classes also represent a range of utility, functionality, and documentation.  Some models are improperly or incompletely implemented, some lack coherent documentation, some lack  examples required to demonstrate use cases and implementation details, while others lack the test coverage required to maintain functionality with updates to the codebase. Beyond a handful of core models, it is at present difficult for an interested user to understand the functionality, purpose, and implementation details for many of Cantera's available models.  Lastly, there 
are a handful of classes with only marginal utility -- classes that largely duplicate functionality that is already present elsewhere in the thermophase classes or that implement models with either dubious physical validity or limited possible use cases.

Simultaneously, the ['Science' section on Cantera's website](https://cantera.org/science/) provides a logical space to document the scientific theory underpinning Cantera's various classes, but is at present underutilized. This would be a great venue for both cataloging and documenting Cantera's various thermophase classes.

This issue, therefore proposes two/three main tasks:

1. A thorough review of Cantera's thermophase classes (current as of 15 June, 2019).  The review should identify:
- Overall summary of the model, including general use cases.
- Identify governing equations for phase (p-v-T) behavior
- Identify the expressions used for activity concentration, standard concentration, and standard-state chemical potential.

2. Identify models which can be deprecated. This might be justified by any combination of: poor or incomplete implementation, a lack of documentation or test suite coverage to verify correct functionality, or functionality that is largely duplicated in other thermophase classes.  Classes which currently see moderate or heavy usage should not be candidates for deprecation, regardless of any of the above conditions (i.e. we would rather attempt to rectify the deficiencies, in place of deprecating the model).

3. After completing steps 1 and 2, Move documentation of the thermophase classes to https://cantera.org/science.  This should include a brief overview of the available classes, consisting of some subsection of the review in part 1, with links available to a more in-depth documentation of each class (i.e. duplication of the docstrings available in the class's header file).

We will use this space to document progress through this review and documentation.

PR #696, opened 2019-08-16
Create SolutionBase object in C++

### LINKED ISSUE #691: Incorrect Solution.ID for YAML generated objects
### Cantera version

2.5.0a3

### Python version

3.6

### Expected Behavior

`gas.ID` should provide the ID of the phase, with the default being taken from the input file. I.e.

```python
gas = ct.Solution('nDodecane_Reitz.yaml')
print(gas.ID)
```
should be `nDodecane_RK`. I.e. the result should be identical to the one obtained if one were to import the CTI file. Explicitly specifying the phase (i.e. `nDodecane_RK` or `nDodecane_IG`) will not change the behavior.

### Actual Behavior

```python
gas = ct.Solution('nDodecane_Reitz.yaml')
print(gas.ID)
```
yields `<phase>`, regardless of the mechanism that is imported.

### Steps to reproduce

See above.

Fix involves [ThermoFactory.cpp#L450](https://github.com/Cantera/cantera/blob/master/src/thermo/ThermoFactory.cpp#L450)

### LINKED ISSUE #695: Question: equivalent of composite Python classes in C++
### Cantera version

2.5.0a3

### Expected Behavior

It would be useful to have C++ equivalents of composite classes that are defined in Python (i.e. `Solution`, `Interface`, etc.). A replication of those classes would clarify origin and structure of thermo/kinetics/transport managers that are instantiated from the Python composite class (the same holds true for Matlab). 

This *does not mean the replication of double/triple inheritance* within C++.  Rather, the C++ classes would receive pointers to previously allocated managers, as well as hold type information (which is currently lost in C++) and a uniquely assignable name. 

The motivation is that a common root for composite objects appears to be necessary for serialization within the C++ layer. A direct mapping of Python to C++ objects would also clarify the overall structure of cantera.

### Actual Behavior

There is no replication of composite classes within C++

### Steps to reproduce

N/A

ischoegl:

Fixes #695, fixes #691 .

#### Changes proposed in this pull request:
- Create representation of `_SolutionBase` object in C++ layer (named `Solution`)
- Link object into Cython interface
- Assign phase name in YAML (see #691) and resolve conflation of gas.ID and gas.name in unit tests
- ~Replace `ID` by more meaningful `phase_id` attribute. The changed attribute is PEP8 compliant and corresponds to the YAML entry~
- Deprecate `Phase::id` and `Phase::setID`
- Details:
  - New `Solution` object is defined in `Solution.h`/`Solution.cpp`
  - New object collects `ThermoPhase`, `Kinetics` and `Transport` managers
  - ~Preserves information how managers were constructed as `type` (e.g. `Solution`, `Interface`)~
  - Move/share `name` attribute from `Phase` to `_SolutionBase` and ensure unique default names generated in Cython (anticipates uses for serialization)
  - Deprecate `ID` 

#### Illustration

There are only minor additions to the user interface; changes mostly benefit "work under the hood".
```python
In [2]: gas = ct.Solution('nDodecane_Reitz.yaml')

In [4]: gas.name # fix of #691 
Out[4]: 'nDodecane_RK'

In [5]: gas.composite
Out[5]: ('RedlichKwong', 'Gas', 'Transport')

In [6]: gas.thermo_model, gas.kinetics_model, gas.transport_model
Out[6]: ('RedlichKwong', 'Gas', 'Transport')

In [7]: gas.ID
DeprecationWarning: To be removed after Cantera 2.5. Use `name` attribute instead
Out[7]: 'nDodecane_RK'

In [9]: gas = ct.Solution('nDodecane_Reitz.yaml', phaseid='nDodecane_IG')
FutureWarning: Keyword `name` replaces `phaseid`
```

#### Comments
The following advantages are anticipated:
- clarify role of phase in #694
- create an obvious 'root' node for YAML serialization
- serve as a starting point for C++ copy constructors of `Solution` and associated managers via YAML, which would be useful for *true* parallel processing.
- facilitate pickling of `_SolutionBase` objects as envisioned by #692 

*Beyond this PR* ... moving creation of objects from `Phase`/`ThermoPhase` to `Solution` would enable switching of thermo models on the fly (similar to transport) and thus eliminate the need for multiple entries in yaml for multiple thermo models (e.g. `nDodecane_RK` and `nDodecane_IG` differ in exactly *one* entry, i.e. `thermo`, which is either `ideal-gas` or `Redlich-Kwong`)


speth:
I had some some high-level comments on this, which it might be worth discussing before going further on the implementation of this. I'll provide a more detailed review once we work out the big picture.

I'm coming around to the idea that a class of this sort would be useful, especially for the second capability you mentioned, where it would serve as the thing that can be serialized to a YAML phase definition.

Regarding the name / ID / phase attributes: Despite the description of this in `Phase.h` (and now in `Base.h`) I've never quite understood the need for two distinct identifiers for a phase. I'm pretty sure we could just get rid of "ID" and skip adding the "phase" identifier.

I don't quite understand why the `SolutionBase` object doesn't provide access to the thermo / kinetics / transport objects. Is this not something you see as useful, or just something to defer until later?

Why is this new class `SolutionBase` and not just `Solution`? Are you anticipating other classes derived from this, and what would their role be? The Python `_SolutionBase` class only for the purpose of dealing with the special inheritance limitations of Cython classes.

> serve as a starting point for C++ copy constructors of SolutionBase and associated managers via YAML, which would be useful for true parallel processing.

Perhaps I'm missing something, but I don't think C++ copy constructors are useful for parallel processing. They only provide a way of copying an object within the memory space of a single process, which doesn't help with transmitting objects to other parallel processes (but of course, YAML
serialization could be used for this).

> moving creation of objects from Phase/ThermoPhase to SolutionBase would enable switching of thermo models on the fly

I'm not sure I follow - in Python, the creation of all the underlying C++ objects is done in `_SolutionBase`. And in C++, the creation of the `ThermoPhase` / `Kinetics` / `Transport` objects is independent. There's nothing preventing you from creating a new `ThermoPhase` with a different model and linking that to a previously-existing `Kinetics` and `Transport` object.

ischoegl:
@speth ... thanks for your comments! I agree that a high-level discussion is warranted before it makes sense to start a detailed review.

> I'm coming around to the idea that a class of this sort would be useful, ...

:+1:  Really happy to hear! 

> Regarding the name / ID / phase attributes: [...] I've never quite understood the need for two distinct identifiers for a phase.

My opinion on this is that within the context of serialization, ~`name` would be the solution type (class identifier), and `phase` would be an unique identifier~ Edit: ... ahem ... `phase` would be the same as specified in YAML, and `name` would be the unique identifier (at least that's how this PR handles this distinction). 

> I don't quite understand why the SolutionBase object doesn't provide access to the thermo / kinetics / transport objects. Is this not something you see as useful, or just something to defer until later?

It's the last: defer until later. In the current implementation, there isn't a need, as I mainly looked into the plumbing from Python back to C++. I am anticipating that the object could become important in C++ as a one-stop shop access point for various managers, where this type of access would become essential.

> Why is this new class `SolutionBase` and not just `Solution`? Are you anticipating other classes derived from this [...]?

My main rationale here was to directly map python nomenclature back to C++. Honestly, I have thought about (ultimately) renaming `Base.h` to `Solution.h`, but that requires a bit of work as that name is already taken. There may be future applications for inherited C++ objects, but I haven't looked into this yet (~#652 may uncover something eventually, although I don't necessarily see a need either~ Edit: On second thought, some surface chemistry applications could benefit). I guess the `Base` in `SolutionBase` makes sense as it is defined in `Base.h`.

>  I don't think C++ copy constructors are useful for parallel processing

Could be, but we don't know how users may deploy the Cantera code in custom applications. At some point, I had started looking into wrapping Cantera into `deal.II` for 2D simulations (ultimately LaminarSmoke put that effort to an end). Based on that experience, being able to easily replicate a Cantera kernel object (i.e. `SolutionBase`) would be really convenient.

ischoegl:
PS: ad C++ ramifications. If this object were to be adopted for use within the C++ layer, additional thought needs to go into factories, etc.. I didn't want to go there with this PR (and doubt that it needs to be added immediately).

Ultimately, I think that an almost exact replication of object structures in Python and C++ would benefit maintenance and simplify overall understanding (figuring out where managers are actually instantiated took a while).

speth:
> Honestly, I have thought about (ultimately) renaming `Base.h` to `Solution.h`, but that requires a bit of work as that name is already taken.

There's no file in Cantera named `Solution.h` that I'm aware of. While there are some exceptions to this, I generally prefer the correspondence of `Foo.h` -> declaration for class `Foo`.

> Ultimately, I think that an almost exact replication of object structures in Python and C++ would benefit maintenance and simplify overall understanding (figuring out where managers are actually instantiated took a while).

I think the challenge in figuring out where things are in the C++ layer is more an indication that how Cantera objects are created in C++ is currently a bit of a mess. While some general correspondence of C++ and Python features is useful, I think we shouldn't expect one language interface to be just a transliteration of the other. Doing that just ends up with each interface being the worst of both worlds based on the limitations of each language and fitting in as part of an idiomatic Python or C++ application. And this quickly becomes unmanageable if you consider additional language interfaces.

ischoegl:
> There's no file in Cantera named Solution.h that I'm aware of. 

My bad :man_facepalming: : it's been a while since I submitted the PR. The (potential) renaming conflict I was recalling was `Phase.h`, so this one is moot. Still, `base.h` corresponds to `base.pyx`, so that's where I tried to keep consistency in the current suggestion.

> I think the challenge in figuring out where things are in the C++ layer is more an indication that how Cantera objects are created in C++ is currently a bit of a mess.

:rofl: Well, I would have called it 'challenging'. It works well, but it is not straight-forward.

> While some general correspondence of C++ and Python features is useful, I think we shouldn't expect one language interface to be just a transliteration of the other. 

I do agree on your comments about exact replications, but as far as `SolutionBase` (or whatever name we may ultimately settle on) goes, my personal inclination would be to *eventually* have a slender SolutionBase constructor called within a factory, which delegates creation of the other three managers in the current fashion. This - at least in my opinion - would allow for straight-forward serialization.

PS: Cython currently not supporting diamond inheritance (I believe this is still true) is one good example where exact replication is limiting. In this context, I also don't think it to be necessary (nor useful) to replicate the diamond structure in `composite.py` in C++, as C++ allows for much simpler handling of native objects.

ischoegl:
One last thought/comment: `Solution` object usually refer to phases. Are there exceptions? (I am not as familiar with features that aren't covered by examples; my own needs are mostly gas phase related, so I may be overlooking something).

ischoegl:
@speth ... Thought about it some and am :thumbsup: with defining `Solution` in `Solution.h` in C++ (instead of `SolutionBase` in `Base.h`). While the structure of objects is different, Python's `Solution` object has the same functionality as the C++ variant In terms of tying the managers together (the Python `Solution` is, however, much more powerful as it inherits from managers, which I don’t believe makes sense in C++). Let me know if you have a better solution in mind ...

PS: an alternative would be to just call those objects `Base`. It isn’t very informative, though. Another more consistent version would simply rename `Base.h` to `SolutionBase.h`, I.e. ignoring the analogy to Cython.

ischoegl:
Weighing all options, renaming `Base.h` to `SolutionBase.h` is probably the most consistent resolution. `SolutionBase` is a class that is does not offer inherited functions (a complete version would be `Solution`, but I do not think that it makes sense to implement this one in C++). I also added accessors for `ThermoPhase`/`Kinetics`/`Transport` managers for future use.

speth:
> > Regarding the name / ID / phase attributes: [...] I've never quite understood the need for two distinct identifiers for a phase.
> 
> My opinion on this is that within the context of serialization, ~`name` would be the solution type (class identifier), and `phase` would be an unique identifier~ Edit: ... ahem ... `phase` would be the same as specified in YAML, and `name` would be the unique identifier (at least that's how this PR handles this distinction).

I recognize the need to have unique identifiers for phases in the serialized form of say, a reactor network, but I think we could make the generation of those identifiers part of the serialization process. Otherwise, I don't know how you guarantee the uniqueness of the `name` attribute without making it something that is generated automatically and can't be set by the user. In which case, you might as well just use the object's address as the unique identifier and replace it with something more human-friendly when serializing.

> I do agree on your comments about exact replications, but as far as `SolutionBase` (or whatever name we may ultimately settle on) goes, my personal inclination would be to eventually have a slender `SolutionBase` constructor called within a factory, which delegates creation of the other three managers in the current fashion. This - at least in my opinion - would allow for straight-forward serialization.

Do we expect `SolutionBase` to be a polymorphic type, or is the polymorphism always embodied in the `ThermoPhase` / `Kinetics` / `Transport` classes? In the former case, I guess the idea would be to introduce something like the Python `Interface` class where you would be able to access the held pointers as objects of their appropriate derived types (e.g. `InterfaceKinetics`) in order to access methods which are not defined on the corresponding base class. Is this what you're thinking? Otherwise, I'm not sure how the factory pattern is useful here.

> Weighing all options, renaming `Base.h` to `SolutionBase.h` is probably the most consistent resolution. `SolutionBase` is a class that is does not offer inherited functions (a complete version would be `Solution`, but I do not think that it makes sense to implement this one in C++). I also added accessors for `ThermoPhase`/`Kinetics`/`Transport` managers for future use.

For the most part, the classes in Cantera that are named `SomethingBase` are incomplete and not directly usable (although they aren't strictly abstract base classes). So given that this class does provide a full interface, I think the name `Solution` is more appropriate. It functions much like the Python `Solution` class, just without the helpful wrapper functions.

ischoegl:
> I recognize the need to have unique identifiers for phases in the serialized form of say, a reactor network, but I think we could make the generation of those identifiers part of the serialization process. Otherwise, I don't know how you guarantee the uniqueness of the name attribute without making it something that is generated automatically and can't be set by the user. In which case, you might as well just use the object's address as the unique identifier and replace it with something more human-friendly when serializing.

I believe that human-readability is the main issue (as you mentioned, `ReactorNet` is a good example). At the same time, I do believe that it makes sense to retain information of the original `phase` name (setters should probably be deprecated/removed). For that reason, my preferred solution would be to retain both `phase` and `name`. For the latter, the auto-generation of names could be kept simplistic (pretty much like it's been handled thus far), and an exception being raised if a user insists on setting duplicate `name`s (this check can compare addresses). As this is part of serialization, I'd keep handling of duplicate `name` instances out of this PR though.

> Do we expect SolutionBase to be a polymorphic type, or is the polymorphism always embodied in the ThermoPhase / Kinetics / Transport classes? 

Excellent question. Ideally, it should not require polymorphism, but then there is `InterfaceKinetics` (which I personally would rather merge into `ReactorNet`). So it depends.

> ... classes in Cantera that are named SomethingBase are incomplete and not directly usable (although they aren't strictly abstract base classes). 

I recognize this differentiation. One argument *for* `SolutionBase` is that it is the direct equivalent of the python object, and renaming it to `Solution` could cause confusion due to the lack of wrappers that are not (and should not be) available in `C++`. I guess I'm trying to be pragmatic in terms of naming. At the same time, I'm ok with `Solution`, despite not having a preference for it.

Ad factories. Currently, they're not needed (the Cython `SolutionBase` object more or less owns the C++ object, which is where the consistency question comes in):

```python
self._base = CxxNewSolutionBase()
```
which wraps `SolutionBase::create`. For a non-polymorphic approach, this is clearly sufficient. Nevertheless, a factory class 'collecting' all objects (original copies of otherwise shared pointers) is more consistent in terms of how things are typically handled within cantera. One other use case is that external cantera applications may want to add an inherited `SolutionBase` object (e.g. `Wall` vs `WallBase`). Overall, I am not arguing to add factories at this point though (i.e. it goes beyond this PR).

speth:
> I believe that human-readability is the main issue (as you mentioned, `ReactorNet` is a good example). At the same time, I do believe that it makes sense to retain information of the original `phase` name (setters should probably be deprecated/removed). For that reason, my preferred solution would be to retain both `phase` and `name`. For the latter, the auto-generation of names could be kept simplistic (pretty much like it's been handled thus far), and an exception being raised if a user insists on setting duplicate `name`s (this check can compare addresses). As this is part of serialization, I'd keep handling of duplicate `name` instances out of this PR though.

Yes, the label used in the input file should certainly be stored somewhere -- no argument there. I think it has to remain something that can be set, though, since in the case where a user defines a phase on-the-fly and wants to save it to YAML, they will need to be able to provide a name. I don't think we need to worry about retaining the input file information exactly -- there are plenty of other operations on `Solution` objects which will make the object no longer correspond to its input file definition.

I would much rather have a single name, which is sufficient for all current purposes, with the exception of serialization. The [discussion in the documentation](https://github.com/Cantera/cantera/blob/fe7be79552433c3779ffbd5b018a493114fffdbc/include/cantera/thermo/Phase.h#L125) of why different names are needed does not seem correct to me. If the definition of "same problem" is a reactor network, there is no requirement that the contents of different reactors have different names (and they can in fact be represented by the same `Solution` object even though the states of the reactors are different). In the case of a multiphase `Mixture`, I think the problem that will arise is not the duplication of the name but the existence of multiple identical phases in the first place. And in an interface kinetics problem, the limitation will be duplication of species names in different phases.

My preference on the single attribute for holding this name is `name`. It makes more sense to me that something like `gri30` is the *name* of a gas rather than its *phase* (which I would be more likely to think of as being synonymous with the thermodynamic model, i.e. ideal gas).

> > ... classes in Cantera that are named SomethingBase are incomplete and not directly usable (although they aren't strictly abstract base classes).
> 
> I recognize this differentiation. One argument _for_ `SolutionBase` is that it is the direct equivalent of the python object, and renaming it to `Solution` could cause confusion due to the lack of wrappers that are not (and should not be) available in `C++`. I guess I'm trying to be pragmatic in terms of naming. At the same time, I'm ok with `Solution`, despite not having a preference for it.

Well, that potential for confusion arguably goes both ways -- the Python `_SolutionBase` class is not (and is not intended to be) something that can be directly used, so a C++ user moving to Python might be confused by the need to use the `Solution` class, right? My thought is that the level at which equivalence between different languages is useful is to be able to say that "`Solution` is an object that provides (through some interface) thermodynamic, kinetic, and transport properties for a phase. The exact methods that are used for working with that object depend on the language interface that you're using."

ischoegl:
@speth ... after some more thought, I'm :thumbsup: with `Solution` as there simply isn't a clear-cut definition (I do not anticipate polymorphism, and exact correspondence is not necessary).

Regarding `phase`/`ID` and `name`. It looks like there is another issue I was not considering, see `Phase.h`'s documentation:
```C++
 * A species name may be referred to via three methods:
 *
 *    -   "speciesName"
 *    -   "PhaseId:speciesName"
 *    -   "phaseName:speciesName"
```
Currently, all methods work, i.e. removal of one or the other would create a more complex PR than needed. In the Python frontend, `PhaseId` maps to `ID` (I suggested to rename it to `phase`) and `PhaseName` maps to `name`. 

I do agree that `phase` may not be the ideal name as it may be mistaken for the `type` of the phase (which I added to the interface). Overall, my thoughts on this at this point are to simply table the issue and leave `ID` for later (or simply use `phase_id` as a more meaningful name in Python). I still believe there is a place for the distinction, but this is not the main issue this PR seeks to address.

ischoegl:
Ok. Per my previous comment, I renamed the C++ object to `Solution` and clarified the identity of the `ThermoPhase` as `phase_id` (for the latter by retracing my steps). 

Further, I believe we may be able to resolve `name` vs `phase_id` as follows:
* `phase_id` is associated with the `ThermoPhase` object (I suggest to remove it from the C++ `Solution`) and matches the original label in the input file (unless reset)
* `name` is associated with the C++ `Solution` object (with the C++ `ThermoPhase` looking it up from its root object if needed)
* due to inheritance, both `name` and `phase_id` coexist in Python, but they originate from a different context.

ischoegl:
Alright. I removed `phaseID` from the C++ `Solution` object so it is clearly associated with C++ `Phase`. Conversely, the `name` is now truly owned by C++ `Solution`. 

Not unexpectedly, an attempt to deprecate `Phase::name`/`Phase::setName` (to enforce usage of `Phase::id`) wreaks havoc: e.g. `phase = ct.Solution('h2o2.xml')` runs into issues within `cantera._cantera._SolutionBase._init_cti_xml`, breaking a substantial portion of the test suite.

ischoegl:
@speth ... looked over the code again and I think the big picture concerns should be addressed.

src/base/Solution.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+std::string Solution::toYAML() const {
+    throw NotImplementedError("Solution::toYAML");
+}
+
+void Solution::setThermoPhase(shared_ptr<ThermoPhase> thermo) {
+    m_thermo = thermo;
+    if (m_thermo) {
+        m_thermo->setRoot(shared_from_this());
+    }
```
speth, line 41: This two-way link between `Solution` and its held objects makes things a bit complicated. If you want `shared_from_this` to always work, then you have to make it impossible to allocate `Solution` on the stack by making all of its constructors private, leaving only the `create` method public.

include/cantera/base/Solution.h
```diff
@@ -0,0 +1,74 @@
...
+    //! Set the Kinetics object
+    void setKinetics(shared_ptr<Kinetics> kinetics);
+
+    //! Set the Transport object
+    void setTransport(shared_ptr<Transport> transport);
+
+    //! Accessor for the ThermoPhase object
+    shared_ptr<ThermoPhase> thermo() {
+        return m_thermo;
+    }
```
speth, line 51: Does it make sense to return a `ThermoPhase&` rather than a `shared_ptr`? I'm having trouble imagining a case where you'd expect the `ThermoPhase` to outlive the `Solution`. Depending on the way that the `Solution` class is ultimately used, that could result in a lot of unnecessary copying of the shared pointers, if someone writes their code to work through mainly through the `Solution` object, e.g. `soln->thermo()->setState_TPX(...); soln->thermo()->enthalpy_mass(); ...`.

src/thermo/Phase.cpp
```diff
@@ -78,12 +79,22 @@ void Phase::setID(const std::string& id_)
 
 std::string Phase::name() const
 {
-    return m_name;
+    shared_ptr<Solution> root = m_root.lock();
+    if (root) {
+        return root->name();
```
speth: So synchronizing the name is the only use of this two-way link between `Solution` and `Phase`? Wouldn't it be easier to have `Solution::name` just return `Phase::name`?

interfaces/cython/cantera/base.pyx
```diff
@@ -21,25 +45,72 @@ cdef class _SolutionBase:
...
+            return pystr(self.base.name())
+
+        def __set__(self, name):
+            self.base.setName(stringify(name))
+
+    property composite:
+        """
+        Returns tuple of thermo/kinetics/transport models associated with
+        this SolutionBase object.
+        """
```
speth, line 92: When would this function be needed, as opposed to using the separate return values of `thermo_model`, `kinetics_model`, and `transport_model` as needed?

interfaces/cython/cantera/composite.py
```diff
@@ -23,20 +23,29 @@ class Solution(ThermoPhase, Kinetics, Transport):
...
     definition, species and reactions defined in an input file::
 
-        gas = ct.Solution('gri30.cti')
+        gas = ct.Solution('gri30.yaml')
 
-    If an input file defines multiple phases, the phase *name* (in CTI) or *id*
-    (in XML) can be used to specify the desired phase::
+    If an input file defines multiple phases, the *phases* entry (in YAML),
+    *name* (in CTI), or *id* (in XML) can be used to specify the desired
+    phase via the ``phase_id`` keyword argument of the constructor::
```
speth: This doesn't read quite right to me. I think it should be more along the lines:
```
If an input file defines multiple phases, the corresponding key in the *phases* map (in YAML),
*name* (in CTI), or *id* (in XML) can be used...
```

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -7,11 +7,29 @@
...
+        self.assertTrue(isinstance(self.phase.thermo_model, str))
+        self.assertTrue(isinstance(self.phase.kinetics_model, str))
+        self.assertTrue(isinstance(self.phase.transport_model, str))
+        self.assertTrue(isinstance(self.phase.composite, tuple))
+        self.assertTrue(len(self.phase.composite)==3)
+        self.assertTrue(
+            self.phase.composite == (self.phase.thermo_model,
+                                     self.phase.kinetics_model,
+                                     self.phase.transport_model))
+        self.phase.name = 'spam'
```
speth, line 28: It is generally preferable to use the specialized assertions defined in the `unittest` module such as `assertEqual`, `assertGreater`, `assertIsInstance`, etc. as they can provide more informative error messages when tests fail.

src/base/Solution.cpp
```diff
@@ -0,0 +1,61 @@
...
+    m_name = name;
+}
+
+std::string Solution::toYAML() const {
+    throw NotImplementedError("Solution::toYAML");
+}
+
+void Solution::setThermoPhase(shared_ptr<ThermoPhase> thermo) {
+    m_thermo = thermo;
+    if (m_thermo) {
```
speth, line 39: Should the name of the solution be update here to match that of `m_thermo`?

src/base/Solution.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "cantera/kinetics/Kinetics.h"
+#include "cantera/transport/TransportBase.h"
+
+namespace Cantera
+{
+
+Solution::Solution() :
+    m_thermo(nullptr),
+    m_kinetics(nullptr),
+    m_transport(nullptr),
```
speth: `shared_ptr`s are automatically initialized to `nullptr`, so there's no need to do this explicitly.

src/base/Solution.cpp
```diff
@@ -0,0 +1,61 @@
...
+    m_name("<Solution_name>")
+{}
+
+Solution::Solution(const std::string& infile,
+                   const std::string& phasename) :
+    Solution()
+{
+    // this *may* be a spot to load all pieces of a phase
+    throw NotImplementedError("Solution constructor from file");
```
speth: I'd rather avoid adding methods which are not implemented.

src/base/Solution.cpp
```diff
@@ -0,0 +1,61 @@
...
+std::string Solution::name() const {
+    return m_name;
+}
+
+void Solution::setName(const std::string& name){
+    m_name = name;
+}
+
+std::string Solution::toYAML() const {
+    throw NotImplementedError("Solution::toYAML");
```
speth: Ditto this one.

interfaces/cython/cantera/base.pyx
```diff
@@ -21,25 +45,72 @@ cdef class _SolutionBase:
...
 
         self._selected_species = np.ndarray(0, dtype=np.integer)
 
     def __init__(self, *args, **kwargs):
         if isinstance(self, Transport):
             assert self.transport is not NULL
 
-    def _init_yaml(self, infile, phaseid, phases, source):
+        phase_name = pystr(self.thermo.id())
+        name = kwargs.get('name', None)
```
speth: The default return value for `get` is `None`, so specifying `None` here is unnecessary.

interfaces/cython/cantera/base.pyx
```diff
@@ -62,19 +133,19 @@ cdef class _SolutionBase:
 
         # Kinetics
         cdef vector[CxxThermoPhase*] v
-        cdef _SolutionBase phase
+        cdef _SolutionBase adj
 
         if isinstance(self, Kinetics):
             v.push_back(self.thermo)
-            for phase in phases:
+            for adj in adjacent:
```
speth: I would argue that `for phase in adjacent` is more readable.

interfaces/cython/cantera/test/test_mixture.py
```diff
@@ -112,10 +112,11 @@ def test_charge(self):
 
     def test_phase_moles(self):
         M = self.mix.phase_moles()
+        name = self.phase2.name
         self.assertEqual(M[0], self.mix.phase_moles(0))
-        self.assertEqual(M[1], self.mix.phase_moles('air'))
+        self.assertEqual(M[1], self.mix.phase_moles(name))
 
-        self.mix.set_phase_moles('air', 4)
+        self.mix.set_phase_moles(name, 4)
```
speth: The fact that this no longer works as it originally did is a clear downside of making the the name globally unique.

interfaces/cython/cantera/base.pyx
```diff
@@ -21,25 +45,72 @@ cdef class _SolutionBase:
...
+            n = _phase_counts[phase_name]
+            self.name = '{0}_{1}'.format(phase_name, n)
+        else:
+            _phase_counts[phase_name] = 0
+            self.name = phase_name
+
+    property name:
+        """
+        The name assigned to this SolutionBase object. The default is
+        taken from the CTI/XML/YAML input file.
```
speth: This description isn't really accurate now, if the name is going to be replaced by a globally unique value.

speth (changes requested):
As I mentioned before, I'm satisfied that a composite class of some sort will provide a useful interface for serialization, as well as a few other purposes.

I think I'd like to defer the resolution of these issues relating to the multiple name-like attributes and whether they need to be unique until we start to flesh out the serialization interface.

speth:
> Regarding `phase`/`ID` and `name`. It looks like there is another issue I was not considering, see `Phase.h`'s documentation:
> 
> ```c++
>  * A species name may be referred to via three methods:
>  *
>  *    -   "speciesName"
>  *    -   "PhaseId:speciesName"
>  *    -   "phaseName:speciesName"
> ```
I was trying to avoid dragging this into this PR, but I was reminded of this capability while reviewing your earlier PR on enabling case sensitivity in species names. I believe that these latter two methods are not actually in use within Cantera, and I'm really not sure that we need or want them. For multiphase mixtures (C++ class `MultiPhase`, Python class `Mixture`, you always need to provide both a phase name/index as well as species name in order to look up information about a species in a particular phase. And for surfaces, the usual approach is that the species names are unique among the phases, e.g. `H2O` is a gas species and `H2O(s)` is the same species absorbed onto the surface. So at some point, I was planning to deprecate and remove this capability anyway.

src/base/Solution.cpp
ischoegl: I don’t think so. The `name` of a `Solution` should be assigned directly. The only reason I left a `m_name` with `Phase` is that otherwise all kinds of things would break for those instances when a `ThermoPhase` is instantiated without a `Solution`. I anticipate that this requirement will disappear in the long run.

src/thermo/Phase.cpp
ischoegl: `Phase::m_name` is only used when there is no `Solution` attached (and only preserved here to avoid breakage of substantial amounts of code). Once addressing via
```
"phaseName:speciesName"
```
is deprecated (which is a delicate endeavor), `Phase::name` will no longer be needed.

interfaces/cython/cantera/base.pyx
ischoegl: It is intended as a service function to provide quick access to the make-up of a Python `Solution` (which is a `composite`).

include/cantera/base/Solution.h
ischoegl: Hm. Tried this via `m_thermo.get()`, but casting a smart pointer to `ThermoPhase&` appears to involve more work than expected. Anything I'm overlooking? (beyond your question being to revert to conventional pointers?)

Regarding `ThermoPhase` outliving `Solution` ... you could reassemble a different `Solution` object from existing parts, in which case shared pointers would make sense. The scenario for unnecessary copying you describe above can be easily avoided by a local variable holding the result of `soln->thermo()`. In the case of a 0D or 1D object, the `Solution` object could be used during initialization; however, it may be more efficient to store references to `ThermoPhase`, `Kinetics`, etc. internally. In either case, the base object can always be retrieved using the `m_root` pointer (see current `Phase::name` implementation).

interfaces/cython/cantera/test/test_mixture.py
ischoegl: The test suite creates numerous `Solution` objects, each of which creates a name entry. In most use cases, the automatic renaming won't get triggered (or a user may prefer a custom name), so I don't think that there is a huge downside here.

ischoegl:
@speth ... thanks for the prompt review! I think most of your comments are addressed, but I left things that may need further discussions unresolved.

> I was trying to avoid dragging this into this PR, but I was reminded of this capability while reviewing your earlier PR on enabling case sensitivity in species names. I believe that these latter two methods are not actually in use within Cantera, and I'm really not sure that we need or want them. 

This makes sense. I wasn't really questioning the use, just trying not to break anything :wink: ... it certainly affects the PR, as both `name` and `phase_id` are directly involved.

include/cantera/base/Solution.h
speth: > Hm. Tried this via `m_thermo.get()`, but casting a smart pointer to `ThermoPhase&` appears to involve more work than expected. Anything I'm overlooking? 

All you have to do to be able to return `ThermoPhase&` is dereference the shared_ptr, i.e.
```c++
ThermoPhase& thermo() {
    return *m_thermo;
}
```
 > (beyond your question being to revert to conventional pointers?)
No, I'm definitely not suggesting that `Solution` should store a raw pointer. Thinking about it more, the scenario of concern with returning a reference would where a user:
1. Allows the `ThermoPhase` object to be held only by the Solution object
2. Stores the address of that reference (i.e. the address of the underlying `ThermoPhase`)
3. Allows the `Solution` object to be destroyed
4. Tries to continue using the pointer to the `ThermoPhase` object.

I don't know how likely this is -- if a user takes the second step, I would argue that they are deliberately subverting the fact that the interface returns a reference, but perhaps it's better to avoid the possibility in any case.

> In the case of a 0D or 1D object, the `Solution` object could be used during initialization; however, it may be more efficient to store references to `ThermoPhase`, `Kinetics`, etc. internally. In either case, the base object can always be retrieved using the `m_root` pointer (see current `Phase::name` implementation).

Are you expecting that we will eventually require 0D / 1D objects to be initialized using `Solution` objects? Otherwise, what behavior should be exhibited when the `ThermoPhase` etc. used in a reactor network don't have an `m_root` pointer when attempting to serialize?

interfaces/cython/cantera/base.pyx
speth: Yes, I understand that. I'm just wondering what the hypothetical use case is -- my intuition is that most times you want this information, you need to use the pieces separately anyway.

src/base/Solution.cpp
speth: Why should the the `name` be assigned directly rather than coming from an input file? Given that the only way to construct a `Solution` (for now) is from an already-existing `ThermoPhase` object, this would be the opportunity to set at least a default value for that name.

interfaces/cython/cantera/test/test_mixture.py
speth: I wouldn't want to presume anything about the level of complexity of a user's code, and how many objects with the same name may be initialized. I'm thinking especially of interactive use in an environment like a Jupyter notebook, where the easiest thing to do is often to re-run the initialization code, in which case this will result in the names all changing.

I do not want to enforce a requirement of globally unique names, which has a non-negligible usability cost, when it is not necessary as part of the current interface or for features (like serialization) that we want to implement.

interfaces/cython/cantera/base.pyx
ischoegl: I think it's a meaningful shortcut for people new to cantera. I added a reference to the description of the `Solution` object.

src/base/Solution.cpp
ischoegl: From my perspective, `Phase::name` is a duplication of `Phase::id` that is not necessary (see #724, which deprecates usage in Cantera internals). In making this choice, I tried to be consistent with past descriptions, which identified `phase_id` as ~non-~ *not necessarily* unique and `name` as unique (also: usage of `--id` in ck2cti, `--phase-id` in ck2yaml, and `phase_id` as keyword argument in `Solution`). 

Uniqueness is needed for serialization, which makes only sense for the entire `Solution` object, which is why I moved `name` there, i.e. `Solution::name`. At the moment, the default value set by the parsers is overwritten in Python; after #724 the `Phase::name` assigned by the parsers will not be used anywhere and can be left empty (a deprecation message can be added at that point).

interfaces/cython/cantera/test/test_mixture.py
ischoegl: I guess this boils down to the question on whether `Mixture._phases` stores `name` or `phase_id` information. Currently it is the former.

IMHO, this issue is easily addressed as `Mixture.phase_index` should probably support *both* `name` and `phase_id`? As there are some open questions in #724, I'll open an issue and defer. Edit: issue #727

bryanwweber:
Hi @ischoegl... sorry to jump in late (again), I just wanted to add my 2 cents on the distinction between `name` and `id` semantically. You said

>  I tried to be consistent with past descriptions, which identified `phase_id` as non-unique and `name` as unique (also: usage of `--id` in ck2cti, `--phase-id` in ck2yaml, and `phase_id` as keyword argument in Solution).

To me, `name` sounds more like something a user would set (people have names by which they introduce themselves and can be changed), whereas an `id` is fixed and unique, like a Social Security Number. I'm not that familiar with how these terms have been used/mixed in the past based on the existing code, but to the extent that their meanings are changing with this work (and other stuff that you're doing), to me it makes more sense to allow the user to set/change a field called `name` and let the unique one be called `id` (or `phase_id` or whatever). Note also that in YAML, the identifier for a phase has the key `name`. Of course, if this is already too far along to make this change, that's fine, it is only a small difference and both ways will work.

ischoegl:
@bryanwweber ... from `Phase.h` (on master):
```C++
    /*! @name Name and ID
     * Class Phase contains two strings that identify a phase. The ID is the
     * value of the ID attribute of the XML phase node that is used to
     * initialize a phase when it is read. The name field is also initialized
     * to the value of the ID attribute of the XML phase node.
     *
     * However, the name field may be changed to another value during the
     * course of a calculation. For example, if a phase is located in two
     * places, but has the same constitutive input, the IDs of the two phases
     * will be the same, but the names of the two phases may be different.
     *
     * It is an error to have two phases in a single problem with the same name
     * and ID (or the name from one phase being the same as the id of
     * another phase). Thus, it is expected that there is a 1-1 correspondence
     * between names and unique phases within a Cantera problem.
     */
```
> to me it makes more sense to allow the user to set/change a field called `name` and let the unique one be called `id`

This is the current (intended) behavior of this PR :wink: ... `phase_id` is automatically assigned and (while it can be changed) should be left alone. It is in a way unique, as it refers to an assembled `Phase` within a `Solution` (via YAML or CTI/XML). The non-uniqueness arises whenever a user decides to use a *duplicate* object in multiple reactors, or similar (i.e. it is a strict serialization issue).

include/cantera/base/Solution.h
ischoegl: :+1:  thanks for the code sample, I guess I was over-thinking this one! (I thought that smart pointers required something less straight-forward :man_facepalming:). Either way, I am changing the accessor as it isn't currently used (I simply added it due to being asked about it). Other accessors can be added later on as needed ...

> Are you expecting that we will eventually require 0D / 1D objects to be initialized using `Solution` objects? Otherwise, what behavior should be exhibited when the `ThermoPhase` etc. used in a reactor network don't have an m_root pointer when attempting to serialize?

Not sure where this may go in the long run: setting all in one swoop is simpler (regardless of the accessor implementation), and it's always possible to assign a root node with empty `Transport` and `Kinetics` managers (much like is already done in Python).

ischoegl:
@bryanwweber ... copying some thoughts over here: `name` does indeed indicate that the user is welcome to change it, whereas `id` should be left alone (or used with caution). Despite the YAML implementation, I believe that the `name` field in YAML is an example for the latter, as it probably should support round-trip serialization (YAML->`Solution`->YAML) where it should remain unchanged. 

I view the `name` of the `Solution` as temporary, i.e. it can be reassigned as needed by the user (unique/human-readable shorthand, ...). Those local changes shouldn't affect the original YAML entry.

Finally, I believe the differentiation suggested by this PR is consistent with other Cantera interfaces (i.e. there is a long tradition of the keyword argument `phaseid` in `ct.Solution` and `--id` in `ck2cti`/`ck2yaml`, despite renaming those to `phase_id` and `--phase-id` in this PR)

interfaces/cython/cantera/base.pyx
ischoegl: Some of my changes apparently got lost.

ischoegl:
Ugh. It looks like I force-pushed a stale branch, losing most of my updates over the last couple of days. I don't think that I have the work elsewhere, so I'll have to retrace.

ischoegl:
:tada: retraced all changes that were lost in a force-push.

src/thermo/Phase.cpp
ischoegl: Update: pushed a PR that detangles `id` and `name` (PR #724). I will add a deprecation message to `Phase::name` in #724 if this PR is accepted.

interfaces/cython/cantera/base.pyx
ischoegl: See: https://github.com/Cantera/cantera/blob/746c2532de768f8fb30d5f0af5490e01d8641a0b/interfaces/cython/cantera/composite.py#L58-L62

interfaces/cython/cantera/composite.py
ischoegl: Updated to: https://github.com/Cantera/cantera/blob/746c2532de768f8fb30d5f0af5490e01d8641a0b/interfaces/cython/cantera/composite.py#L28-L31

interfaces/cython/cantera/base.pyx
ischoegl: Fixed.

interfaces/cython/cantera/base.pyx
ischoegl: changed to: https://github.com/Cantera/cantera/blob/746c2532de768f8fb30d5f0af5490e01d8641a0b/interfaces/cython/cantera/base.pyx#L89-L91

ischoegl:
To add an alternative approach (following a discussion with @bryanwweber): instead of trying to enforce `Phase:id` following a long tradition (and the original intent of some docstrings), the introduction of YAML (which uses the field `name`) could be used as a reason to switch over to use `Phase:name` consistently and abandon `id` and associated names instead (i.e. deprecate it without a substitute).

I.e. use `name` as a key in `Solution('gri30.yaml', name='gri30')`, use `--name=...` in `ck2yaml`, etc.. It is probably not the end of the world if a user changes the name of a `Phase`/`Solution`, and the original identifier (i.e. what was used in the YAML file) gets lost. There is only one field in YAML, and everything else may be excessive ...

I guess I am finally coming around. I still would like to use this opportunity to ensure that the naming convention is being used consistently. The change is not overly complicated ...

interfaces/cython/cantera/composite.py
```diff
@@ -27,19 +27,16 @@ class Solution(ThermoPhase, Kinetics, Transport):
...
+        gas = ct.Solution('diamond.yaml', name='gas')
+        diamond = ct.Solution('diamond.yaml', name='diamond')
 
-    The name of the `Solution` object needs to be unique and defaults to the
-    *phase* specified in the input file. If another object using the same
-    constituting information already exists, the name is automatically appended
-    by a suffix. A custom name can be set via the ``name`` keyword argument of
-    the constructor, i.e.::
+    The name of the `Solution` object defaults to the *phase* specified in the 
+    input file. Once instatiated, a custom name can assigned via::
```
ischoegl: fix text/typo

ischoegl:
I went ahead as the change was quick and straight-forward. Getting rid of ID altogether certainly feels more consistent. I'll wait for a re-review before adjusting #724 (which will be switched to a - much less intrusive - deprecation of `Phase::id` and `Phase::setID`)

src/base/Solution.cpp
ischoegl: The discussion is resolved: `Solution` accesses the name via `Phase`, I.e. there is no distinction any longer.

interfaces/cython/cantera/test/test_mixture.py
ischoegl: Resolved by switching to name throughout.

interfaces/cython/cantera/test/test_mixture.py
```diff
@@ -112,10 +112,11 @@ def test_charge(self):
 
     def test_phase_moles(self):
         M = self.mix.phase_moles()
+        name = self.phase2.name
```
ischoegl: No longer needed. Switch to old behavior.

ischoegl:
I added deprecation of `Phase::id` and `Phase::setID` to this PR to illustrate that there is indeed no need to maintain two string identifiers. I also cherry-picked deprecation of name/id support for `Phase::speciesIndex` from #724 to consolidate all related efforts (i.e. #724 and #727 are now moot and closed).

I cannot think of anything else at this point, so this PR is ready for review.

include/cantera/base/Solution.h
```diff
@@ -0,0 +1,70 @@
...
+
+    //! Accessor for the Transport object
+    Transport& transport() {
+        return *m_transport;
+    }
+
+protected:
+    shared_ptr<ThermoPhase> m_thermo;  //! ThermoPhase manager
+    shared_ptr<Kinetics> m_kinetics;  //! Kinetics manager
+    shared_ptr<Transport> m_transport;  //! Transport manager
```
speth: The end-of-line Doxygen comments need to be written starting with `//!<`. Otherwise, they are seen as applying to the following entity.

include/cantera/thermo/Phase.h
```diff
@@ -66,15 +68,11 @@ namespace Cantera
...
- *
- *    -   "speciesName"
- *    -   "PhaseId:speciesName"
- *    -   "phaseName:speciesName"
- *    .
- *
- * The first two methods of naming may not yield a unique species within
- * complicated assemblies of %Cantera Phases.
+ * A species name is referred to via "speciesName", which is unique within a
+ * given phase. Note that within multiphase mixtures ("MultiPhase"), both a
```
speth: I think these will be best handled by Doxygen (autolinked to the corresponding methods / classes) if you write them as `speciesName()` and `MultiPhase` (the parens are allowable, but the quotes are unnecessary).

interfaces/cython/cantera/base.pyx
```diff
@@ -21,25 +43,67 @@ cdef class _SolutionBase:
...
         if isinstance(self, Transport):
             assert self.transport is not NULL
 
-    def _init_yaml(self, infile, phaseid, phases, source):
+        phase_name = pystr(self.base.name())
+        name = kwargs.get('name')
+        if name is not None:
+            self.name = name
+        else:
+            self.name = phase_name
```
speth: To clarify, this is handling the case for setting the name when a `Solution` is created from scratch?

interfaces/cython/cantera/base.pyx
```diff
@@ -21,25 +43,67 @@ cdef class _SolutionBase:
...
+        The name assigned to this object. The default value corresponds 
+        to the CTI/XML/YAML input file phase entry.
+        """
+        def __get__(self):
+            return pystr(self.base.name())
+
+        def __set__(self, name):
+            self.base.setName(stringify(name))
+
+    property composite:
```
speth, line 88: Can I suggest something along the lines of `composite_type` or `composite_model` for this property?

interfaces/cython/cantera/transport.pyx
```diff
@@ -144,7 +144,11 @@ cdef class Transport(_SolutionBase):
                     model = 'None'
                 self.transport = newTransportMgr(stringify(model), self.thermo)
             self._transport.reset(self.transport)
+
+        base_type = kwargs.pop('base_type', 'Transport')
+        kwargs['base_type'] = base_type
```
speth: When / where would the `base_type` be set or used? How is this different from `transport_model`?

src/thermo/Phase.cpp
```diff
@@ -206,6 +214,11 @@ size_t Phase::speciesIndex(const std::string& nameStr) const
         std::string pn;
         std::string sn = parseSpeciesName(nameStr, pn);
         if (pn == "" || pn == m_name || pn == m_id) {
+            warn_deprecated("Phase::speciesIndex()",
+                            "Retrieval of species indices via "
+                            "'PhaseId:speciesName' or 'phaseName:"
+                            "speciesName to be removed "
+                            "after Cantera 2.5.");
```
speth: A minor suggestion, but to avoid having to wrap this over so many lines, we sometimes will wrap lines in a multiline statement by just one level, rather than all the way to after the open parenthesis, e.g.
```c++
            warn_deprecated("Phase::speciesIndex()",
                "Retrieval of species indices via 'PhaseId:speciesName' or "
                "'phaseName:speciesName to be removed after Cantera 2.5.");
```
I find I end up doing this a lot with warning and error messages.

speth (changes requested):
Thanks for making these updates. I think converging on `name` as the sole phase identifier and removal of the extra `phase_name:species_name` access are both useful simplifications to the user interface, and make the introduction of the `Solution` class simpler.

I think there are just a few minor points left to address before this can be merged.

include/cantera/base/Solution.h
ischoegl: OK.

interfaces/cython/cantera/transport.pyx
ischoegl: Good catch: this one I overlooked when removing a superfluous feature. It's not needed anywhere else.

interfaces/cython/cantera/base.pyx
ischoegl: It's a tuple of types/models, so I'd prefer to leave it as is (an alternate would be `composite_composition`, which sounds redundant).

To clarify, `composite_type` (or `composite_model`) to me would indicate `Solution`, `Interface`, etc. (e.g. a duplicate of `type(gas)`, with `gas` being a composite object). I wanted to differentiate (and used the name of the declaring module as a guide). I'm open to alternatives though.

src/thermo/Phase.cpp
ischoegl: OK. Fixing instances introduced in this PR, but leaving others.

interfaces/cython/cantera/base.pyx
ischoegl: Correct. Also, I removed the `else` branch that is actually not necessary.

interfaces/cython/cantera/base.pyx
speth: Yeah, I see your point on that being a bit misleading. Just `composite` feels somehow incomplete to me, but I don't have a better suggestion, so we can leave this as is for now.

speth (approved):

PR #708, opened 2019-09-17
Add tests for ck2yaml

speth:
The test suite re-uses the tests for ck2cti.

Also fix a few small issues identified by running the tests.

### DIFF
--- a/interfaces/cython/cantera/test/test_convert.py
+++ b/interfaces/cython/cantera/test/test_convert.py
@@ -5,19 +5,30 @@ from pathlib import Path
 
 from . import utilities
 import cantera as ct
-from cantera import ck2cti, cti2yaml
+from cantera import ck2cti, ck2yaml, cti2yaml
 
+class converterTestCommon:
+    def convert(self, inputFile, thermo=None, transport=None,
+                surface=None, output=None, quiet=True, permissive=None):
+        if output is None:
+            output = inputFile[:-4]  # strip '.inp'
+        if inputFile is not None:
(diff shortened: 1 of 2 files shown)


interfaces/cython/cantera/test/test_convert.py
```diff
@@ -5,19 +5,30 @@
 
 from . import utilities
 import cantera as ct
-from cantera import ck2cti, cti2yaml
+from cantera import ck2cti, ck2yaml, cti2yaml
+
+class converterTestCommon:
```
bryanwweber, line 10: I was going to suggest using `pathlib` here, to move away from string manipulation for paths, but I'm not sure it's worth it. Since the converter functions don't work with path-types, it night as well just still be string stuff...

bryanwweber (approved):
Other than the one comment below, this all seems fine. Any reason not to use this to refactor the CTML tests too? I'm also asking because I'm wondering if I should use it for the `ctml2yaml` tests.

speth:
> Other than the one comment below, this all seems fine. Any reason not to use this to refactor the CTML tests too? I'm also asking because I'm wondering if I should use it for the `ctml2yaml` tests.

What do you mean by "the CTML tests"? The refactoring here is fairly easy since in both cases you're going from CK format to something that can be imported as a `Solution`, just with a different intermediate format. I think the tests for `ctml2yaml` are unique, since we don't have any other converter that starts from the XML format, and there should be tests of `ctml2yaml` that don't have corresponding input files in a different format (other than reference YAML files).

PR #709, opened 2019-09-19
Updated URL in references to Cantera's license.

paulblum:
- Updated URL in references to Cantera's license, from http://www.cantera.org/license.txt to https://cantera.org/license.txt

### DIFF
--- a/include/cantera/Edge.h
+++ b/include/cantera/Edge.h
@@ -1,7 +1,7 @@
 //! @file Edge.h
 
 // This file is part of Cantera. See License.txt in the top-level directory or
-// at http://www.cantera.org/license.txt for license and copyright information.
+// at https://cantera.org/license.txt for license and copyright information.
 
 #ifndef CXX_EDGE
 #define CXX_EDGE
--- a/include/cantera/IdealGasMix.h
+++ b/include/cantera/IdealGasMix.h
@@ -1,7 +1,7 @@
 //! @file IdealGasMix.h
 
 // This file is part of Cantera. See License.txt in the top-level directory or
-// at http://www.cantera.org/license.txt for license and copyright information.
+// at https://cantera.org/license.txt for license and copyright information.
 
 #ifndef CXX_IDEALGASMIX
 #define CXX_IDEALGASMIX
--- a/include/cantera/IncompressibleSolid.h
+++ b/include/cantera/IncompressibleSolid.h
@@ -1,7 +1,7 @@
 //! @file IncompressibleSolid.h
 
 // This file is part of Cantera. See License.txt in the top-level directory or
-// at http://www.cantera.org/license.txt for license and copyright information.
+// at https://cantera.org/license.txt for license and copyright information.
 
 #ifndef CXX_INCOMPRESSIBLE
 #define CXX_INCOMPRESSIBLE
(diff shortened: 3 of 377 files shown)


ischoegl:
While this PR does address a known issue, it will create quite a few merge conflicts for pending pull requests (plus complete re-builds for existing toolchains when switching between branches that are not rebased). If adopted, I'd request to make this change at the last moment prior to the official release of 2.5.

bryanwweber:
@ischoegl I hope this doesn't create many merge conflicts. We don't have many open PRs that touch the top of these files, I think yours are the only ones, so it should be pretty easy to sort out.

For the other concern, I'm afraid there's not much we can do about that... Any long-running branches will have to rebased onto master before they're merged anyways, so they might as well be rebased after (if) this change is merged so that the rebuild only happens once. I don't see why waiting until just before 2.5 would help that much, could you clarify that?

ischoegl:
@bryanwweber I haven't checked other PR's, but there are indeed quite a few files that are affected in what is currently pending for myself. It's _not a big issue_, so I"ll deal with it. Regarding my comment: a version jump appears to be the obvious time when house-keeping operations are applied.

ischoegl:
@paulblum ... thank you submitting the PR regardless of my above comment(s). Depending on the order of approval of pending PR's, we'll just have to resolve the unavoidable merge conflicts case-by-case. Overall, it's a good thing if this is taken care of once and for all.

speth:
@ischoegl Just having changes in the same file won't generate merge conflicts -- they need to be changes to the same line or the two adjacent lines in order for git's automatic merge algorithm to fail and require human intervention. If the two branches make the exact same changes, the merge/rebase will also succeed. Of your 6 open PRs, the only place where I get a merge conflict when rebasing is for one file (base.pyx) in #696, since you have additions immediately below the URL being updated here.

I understand this isn't completely painless, but I had suggested making this change en masse to avoid having to keep asking PR authors to make the piecemeal version of this change and to keep this unrelated change out of the diff of future commits / PRs.

ischoegl:
@speth ... thanks for the clarification. I apparently did not recognize that identical changes in independent branches won't cause problems, despite this being the logical behavior. I apologize if dread of having to perform rebase-acrobatics got the better of me.

bryanwweber:
@ischoegl Rebase-acrobatics-dread is certainly understandable. I'm reminded of this XKCD comic: https://xkcd.com/1597/ As @speth said, this touches areas of the files unlikely to be touched in other branches. I also wasn't aware that git would just figure things out if the same change was made twice, but I'm glad to know that's the case!

Also, full disclosure, @paulblum is one of my students and I asked him to make this PR 😸

ischoegl:
@bryanwweber ... thanks for the XKCD link - he always seems to have an appropriate comic! Glad to have learned something useful also: no more automatic alarm bells if the same lines are touched in separate branches ... (still: beware of adjacent modifications!)

@paulblum In hindsight my comments were based on a false alarm on my part. Sorry about that ...

speth:
To add a little data to the theory: After having merged #689, which makes a this same update to several files, Github still says "Rebase and merge can be performed automatically." So with that, I think this is safe to merge.

PR #693, opened 2019-08-14
Add script to convert CTML to YAML input format

bryanwweber:
The script in the PR will convert CTML to YAML input format. The first version here works for GRI-3.0 and I'm posting it to get feedback on the approach I'm taking. Obviously, there is still a lot to be done (I'll try to generate a checklist at some point).

I've attached the current `gri30.yaml` file to this message as a `.txt` file.

[gri30.txt](https://github.com/Cantera/cantera/files/3502920/gri30.txt)

Some notes about the YAML format I'm noticing as I do this:

* Should the `note` field of a species entry should be at the `species` level, not the `thermo` level? This would match the CTI and CTML locations of this data, but I do believe it is extracted from the NASA thermo entry, so maybe it does belong in the thermo field. (`ctml1yaml.py` currently puts it at the species level, while I believe `ck2yaml.py` puts it at the `thermo` level)
* The current `gri30.yaml` that is generated only has one phase entry for mixture averaged transport. This is different from `gri30.cti` and `gri30.xml` that have 3 phase entries.

PR #711, opened 2019-09-20
Python wrapper classes for NASA9 and piecewise Gibbs

### LINKED ISSUE #705: Python SpeciesThermo wrapper classes missing for NASA9 and piecewise Gibbs (Mu0Poly)
In `speciesthermo.pyx`, wrapper classes are defined for the most commonly used species thermo models (`NasaPoly2`, `ShomatePoly2`, and `ConstantCp`) but not for the other available models (`Nasa9PolyMultiTempRegion` and `Mu0Poly`).

ischoegl:

Fixes #705

- Add `Nasa9PolyMultiTempRegion` and `Mu0Poly` class wrappers to `speciesthermo.pyx`
- Add missing C++ constructor for `Nasa9PolyMultiTempRegion`
- Enable variable number of coefficients
- Add unit tests

Note: PR depends on #720 (`piecewise-Gibbs` is affected by #595)

### DIFF
--- a/include/cantera/thermo/Nasa9PolyMultiTempRegion.h
+++ b/include/cantera/thermo/Nasa9PolyMultiTempRegion.h
@@ -57,11 +57,18 @@ public:
      * @param thigh   Maximum temperature
      * @param pref    reference pressure (Pa).
      * @param coeffs  Vector of coefficients used to set the parameters for the
-     *                standard state.
+     *                standard state. The vector has 1 + 11*`nzones` elements
+     *                in the following order:
+     *                - `coeffs[0]`: Number of zones (`nzones`)
+     *                - `coeffs[1 + 11*zone]`: minimum temperature within zone
+     *                - `coeffs[2 + 11*zone]`: maximum temperature within zone
+     *                - `coeffs[3:11 + 11*zone]`: 9 coefficient parameterization
+     *                where `zone` runs from zero to `nzones`-1.
      */
-    Nasa9PolyMultiTempRegion(double tlow, double thigh, double pref, const double* coeffs);
+    Nasa9PolyMultiTempRegion(double tlow, double thigh, double pref,
+                             const double* coeffs);
 
-   //! Set the array of polynomial coefficients for each temperature region
+    //! Set the array of polynomial coefficients for each temperature region
     /*!
      *  @param regions  Map where each key is the minimum temperature for a
      *                  region and each value is the array of 9 polynomial
--- a/include/cantera/thermo/Nasa9PolyMultiTempRegion.h
+++ b/include/cantera/thermo/Nasa9PolyMultiTempRegion.h
@@ -51,7 +51,17 @@ public:
      */
     Nasa9PolyMultiTempRegion(std::vector<Nasa9Poly1*> &regionPts);
 
-    //! Set the array of polynomial coefficients for each temperature region
+    //! Constructor with all input data
+    /*!
+     * @param tlow    Minimum temperature
+     * @param thigh   Maximum temperature
+     * @param pref    reference pressure (Pa).
+     * @param coeffs  Vector of coefficients used to set the parameters for the
+     *                standard state.
+     */
+    Nasa9PolyMultiTempRegion(double tlow, double thigh, double pref, const double* coeffs);
+
+   //! Set the array of polynomial coefficients for each temperature region
     /*!
      *  @param regions  Map where each key is the minimum temperature for a
      *                  region and each value is the array of 9 polynomial
--- a/include/cantera/thermo/Nasa9PolyMultiTempRegion.h
+++ b/include/cantera/thermo/Nasa9PolyMultiTempRegion.h
@@ -75,6 +75,8 @@ public:
                                       doublereal* cp_R, doublereal* h_RT,
                                       doublereal* s_R) const;
 
+    virtual size_t nCoeffs() const;
+
     //! This utility function reports back the type of parameterization and all
     //! of the parameters for the species, index.
     /*!
--- a/include/cantera/thermo/SpeciesThermoInterpType.h
+++ b/include/cantera/thermo/SpeciesThermoInterpType.h
@@ -206,11 +206,11 @@ public:
                                       doublereal* h_RT,
                                       doublereal* s_R) const;
 
-    //! This utility function reports back the number of coefficients
+    //! This utility function returns the number of coefficients
     //! for a given type of species parameterization
     virtual size_t nCoeffs() const;
 
-    //! This utility function reports back the type of parameterization and all
+    //! This utility function returns the type of parameterization and all
     //! of the parameters for the species.
     /*!
      * All parameters are output variables
--- a/include/cantera/thermo/SpeciesThermoInterpType.h
+++ b/include/cantera/thermo/SpeciesThermoInterpType.h
@@ -206,11 +206,9 @@ public:
                                       doublereal* h_RT,
                                       doublereal* s_R) const;
 
-
     //! This utility function reports back the number of coefficients
     //! for a given type of species parameterization
     virtual size_t nCoeffs() const;
-
 
     //! This utility function reports back the type of parameterization and all
     //! of the parameters for the species.
--- a/include/cantera/thermo/SpeciesThermoInterpType.h
+++ b/include/cantera/thermo/SpeciesThermoInterpType.h
@@ -206,6 +206,12 @@ public:
                                       doublereal* h_RT,
                                       doublereal* s_R) const;
 
+
+    //! This utility function reports back the number of coefficients
+    //! for a given type of species parameterization
+    virtual size_t nCoeffs() const;
+
+
     //! This utility function reports back the type of parameterization and all
     //! of the parameters for the species.
     /*!
--- a/interfaces/cython/cantera/speciesthermo.pyx
+++ b/interfaces/cython/cantera/speciesthermo.pyx
@@ -32,7 +32,7 @@ cdef class SpeciesThermo:
         if not init:
             return
 
-        if not self.check_n_coeffs(len(coeffs)):
+        if not self._check_n_coeffs(len(coeffs)):
             raise ValueError("Coefficient array has incorrect length")
         cdef np.ndarray[np.double_t, ndim=1] data = np.ascontiguousarray(
             coeffs, dtype=np.double)
@@ -79,7 +79,7 @@ cdef class SpeciesThermo:
                                            T_high, P_ref, &data[0])
             return data
 
-    def check_n_coeffs(self, n):
+    def _check_n_coeffs(self, n):
         raise NotImplementedError('Needs to be overloaded')
 
     def cp(self, T):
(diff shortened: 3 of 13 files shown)


ischoegl:
@speth ... I wasn’t able to find a reference for a `Mu0Poly` YAML entry ... i am planning to add a test to this PR (adding the entry to `thermo-models.yaml`). I found the `AnyMap` parser, 
https://github.com/Cantera/cantera/blob/55fe2233b18d72aa60d776d2bdace14f98d94346/src/thermo/SpeciesThermoFactory.cpp#L368
but would like to avoid guessing the correct format.

ischoegl:
Added another test after locating a `piecewise-Gibbs` YAML entry, which fails with:
```
FAIL: test_piecewise_Gibbs_load (cantera.test.test_thermo.TestSpeciesThermo)
  File "/src/build/python/cantera/test/test_thermo.py", line 1204, in test_piecewise_Gibbs_load
    sol = ct.Solution('thermo-models.yaml', 'HMW-NaCl-electrolyte')
  File "interfaces/cython/cantera/thermo.pyx", line 267, in cantera._cantera.ThermoPhase.__init__
  File "interfaces/cython/cantera/transport.pyx", line 140, in cantera._cantera.Transport.__init__
CanteraError thrown by HMWSoln::setDensity:
Density is not an independent variable
```
indicating that there is a co-dependence with #720 and #595. Specifically, there is a call to `Phase::restoreState` here:
https://github.com/Cantera/cantera/blob/55fe2233b18d72aa60d776d2bdace14f98d94346/src/transport/TransportFactory.cpp#L91

Update: issue is fixed in #720

bryanwweber:
Interestingly, I ran into exactly this same error today in #693. Can you extract the fix into its own PR so it can be merged a little quicker?

ischoegl:
> Interestingly, I ran into exactly this same error today in #693. Can you extract the fix into its own PR so it can be merged a little quicker?

@bryanwweber that's indeed curious! ... I worked on #720 for a while today (streamlining & simplifying, so it covers all of `thermo-models.yaml`): it's close, and should fix all of the behavior covered in #595 for good (I expect that you'll find this pop up multiple times in #693). Only thing missing is `lattice-compound` (which disables setting of mass fractions). I.e. once this is working correctly, the entire PR should be reviewed.

Edit: #720 is ready for review

interfaces/cython/cantera/speciesthermo.pyx
```diff
@@ -130,7 +165,34 @@ cdef class NasaPoly2(SpeciesThermo):
...
+
+            - `coeffs[0]`: Number of zones (nzones)
+            - within each zone
+              - `coeffs[zone][0]`: minimum temperature
+              - `coeffs[zone][1]`: maximum temperature
+              - `coeffs[zone][2:10]`: The 9 coefficients of the parameterization
+
+        These coefficients should be provided in their customary units (i.e.
+        such that :math:`c_p^o` is in J/gmol-K and :math:`H^o` is in kJ/gmol,
+        as in the NIST Chemistry WebBook).
```
speth: Aren't the coefficients nondimensional, since the polynomials are for c_p/R and H/RT, as per https://cantera.org/science/science-species.html#the-nasa-9-coefficient-polynomial-parameterization?

interfaces/cython/cantera/speciesthermo.pyx
```diff
@@ -130,7 +165,34 @@ cdef class NasaPoly2(SpeciesThermo):
...
+    This is a wrapper for the C++ class :ct:`Nasa9PolyMultiTempRegion`.
+
+    :param coeffs:
+        An array of 1 + 11*nzones elements, in the following order:
+
+            - `coeffs[0]`: Number of zones (nzones)
+            - within each zone
+              - `coeffs[zone][0]`: minimum temperature
+              - `coeffs[zone][1]`: maximum temperature
+              - `coeffs[zone][2:10]`: The 9 coefficients of the parameterization
```
speth: This makes it seem like `coeffs` can be a nested array. I think this needs to be written out in more gorey detail, i.e.
`coeffs[1 + 11*zone]: minimum temperature` (or whatever the correct index is).

include/cantera/thermo/Nasa9PolyMultiTempRegion.h
```diff
@@ -51,7 +51,17 @@ class Nasa9PolyMultiTempRegion : public SpeciesThermoInterpType
...
-    //! Set the array of polynomial coefficients for each temperature region
+    //! Constructor with all input data
+    /*!
+     * @param tlow    Minimum temperature
+     * @param thigh   Maximum temperature
+     * @param pref    reference pressure (Pa).
+     * @param coeffs  Vector of coefficients used to set the parameters for the
+     *                standard state.
+     */
+    Nasa9PolyMultiTempRegion(double tlow, double thigh, double pref, const double* coeffs);
```
speth: The structure of `coeffs` needs to be documented here.

include/cantera/thermo/Nasa9PolyMultiTempRegion.h
```diff
@@ -51,7 +51,17 @@ class Nasa9PolyMultiTempRegion : public SpeciesThermoInterpType
...
+
+   //! Set the array of polynomial coefficients for each temperature region
```
speth: Indentation error, I think?

include/cantera/thermo/SpeciesThermoInterpType.h
```diff
@@ -206,6 +206,10 @@ class SpeciesThermoInterpType
                                       doublereal* h_RT,
                                       doublereal* s_R) const;
 
+    //! This utility function reports back the number of coefficients
```
speth:
```suggestion
    //! This utility function returns the number of coefficients
```

src/thermo/Nasa9PolyMultiTempRegion.cpp
```diff
@@ -57,6 +57,43 @@ Nasa9PolyMultiTempRegion::Nasa9PolyMultiTempRegion(vector<Nasa9Poly1*>& regionPt
     }
 }
 
+Nasa9PolyMultiTempRegion::Nasa9PolyMultiTempRegion(double tlow, double thigh, double pref,
+                                                   const double* coeffs)
+{
+    size_t regions = (size_t)coeffs[0];
```
speth: We usually try (admittedly not consistent, especially in older code) to use `static_cast` rather than the C-style casts, i.e.
```suggestion
    size_t regions = static_cast<size_t>(coeffs[0]);
```

src/thermo/Nasa9PolyMultiTempRegion.cpp
```diff
@@ -57,6 +57,43 @@ Nasa9PolyMultiTempRegion::Nasa9PolyMultiTempRegion(vector<Nasa9Poly1*>& regionPt
...
+        vector_fp cs;
+        poly->setRefPressure(pref);
+        poly->setMinTemp(coeffs[11*i+1]);
+        poly->setMaxTemp(coeffs[11*i+2]);
+        for (size_t j=11*i+3; j<=11*(i+1); j++) {
+            cs.push_back(coeffs[j]);
+        }
+        poly->setParameters(cs);
+        m_regionPts.emplace_back(poly);
+    }
```
speth: I think this can be simplified a bit by passing arguments into the constructor to `Nasa9Poly1`, e.g.
```c++
Nasa9Poly1* poly = new Nasa9Poly1(coeffs[11*i+1], coeffs[11*i+2], pref, coeffs + 11*i + 3);
```

src/thermo/Nasa9PolyMultiTempRegion.cpp
```diff
@@ -57,6 +57,43 @@ Nasa9PolyMultiTempRegion::Nasa9PolyMultiTempRegion(vector<Nasa9Poly1*>& regionPt
...
+
+    m_lowerTempBounds.resize(regions);
+    m_lowT = tlow;
+    m_highT = thigh;
+    m_Pref = pref;
```
speth: I would recommend delegating initialization of these three properties through a call to the `SpeciesThermoInterpType` constructor. For example, see the equivalent constructor for `NasaPoly2`.

src/thermo/SpeciesThermoInterpType.cpp
```diff
@@ -41,6 +41,12 @@ void SpeciesThermoInterpType::updatePropertiesTemp(const double temp,
                        "Not implemented");
 }
 
+size_t SpeciesThermoInterpType::nCoeffs() const
+{
+    throw CanteraError("SpeciesThermoInterpType::nCoeffs",
+                       "Not implemented");
```
speth: You can use `NotImplementedError("SpeciesThermoInterpType::nCoeffs")` here.

interfaces/cython/cantera/speciesthermo.pyx
```diff
@@ -72,6 +79,9 @@ cdef class SpeciesThermo:
                                            T_high, P_ref, &data[0])
             return data
 
+    def _check_n_coeffs(self, n):
+        raise NotImplementedError('Needs to be overloaded')
```
speth: A docstring for this function would be useful.

Also, we could avoid the need to override this in every child class by having `return n == self.n_coeffs` be the default implementation, which would be correct for all classes where the number of coefficients is a constant.

speth (changes requested):
Thanks, this clears up an obvious gap in the Python interface. I think all of my comments are just about implementation details and documentation.

ischoegl:
@speth: thank you for the review! I will, however, wait for #720 to make sure that things can be merged in order.

speth:
Well, I think my review of #720 is going to be a little more involved, so I wouldn't mind getting this merged first. The test errors here can be fixed quite easily, by replacing `Solution` with `ThermoPhase` in the two failing tests.

src/thermo/SpeciesThermoInterpType.cpp
ischoegl: True. Had copied this from neighboring functions. Updating all 5 instances.

interfaces/cython/cantera/speciesthermo.pyx
ischoegl: Correct. Looks like I copied a non-pertinent docstring fragment.

interfaces/cython/cantera/speciesthermo.pyx
ischoegl: Ok on the docstring.

The other suggestion won’t work all the time as the number of zones is not known a priori for the NASA9 multi temp case.

src/thermo/Nasa9PolyMultiTempRegion.cpp
ischoegl: indeed.

interfaces/cython/cantera/speciesthermo.pyx
ischoegl: Ok to docstring.

The suggestion won't work as the function is called prior to the instantiation of the C++ object. This is clarified in the docstring.

ischoegl:
NP. Adopted suggested changes and (temporarily) replaced `Solution` by `ThermoPhase` in the two failing tests. Can switch this back after rebasing #720 upon merge of this PR.

interfaces/cython/cantera/speciesthermo.pyx
speth: Ah, fair point. The documentation of the tricky behavior is appreciated.

speth (approved):

PR #714, opened 2019-09-24
Add species alias

ischoegl:
This PR adds utility functions that facilitate selection of species and isomers. Functions are defined in the C++ layer and broken out using the Cython interface.

- `add_species_alias` allows for alternate species names/shorthands
- `find_isomers` returns a list of isomers for a given composition

Usage:
```python
In [1]: import cantera as ct

In [2]: gas = ct.Solution('nDodecane_Reitz.yaml')

In [3]: gas.find_isomers({'C':4, 'H':9, 'O':2})
Out[3]: ['c4h8ooh', 'c4h9o2']

In [4]: gas.find_isomers({'C':7, 'H':16})
Out[4]: []

In [5]: gas.find_isomers('C:1, H:4')
Out[5]: ['ch4']

In [6]: gas.add_species_alias('ch4', 'Methane')

In [7]: gas.species_index('Methane') == gas.species_index('ch4')
Out[7]: True

In [8]: gas.X = 'Methane:1.'

In [9]: gas.X[gas.species_index('ch4')]
Out[9]: 1.0
```
PS: the implementation of `findIsomers` is analogous to #635; this PR explicitly does not address the overloading of `speciesIndex` (which could be simplified based on this PR).

### DIFF
--- a/include/cantera/thermo/Phase.h
+++ b/include/cantera/thermo/Phase.h
@@ -703,6 +703,19 @@ public:
      */
     virtual void modifySpecies(size_t k, shared_ptr<Species> spec);
 
+    //! Add a species alias (i.e. user-defined alternative species name).
+    //! Aliases are case-sensitive.
+    //!     @param name original species name std::string.
+    //!     @param alias alternate name std::string.
+    //!     @return `true` if the alias was successfully added
+    //!             (i.e. the original species name is found)
+    void addSpeciesAlias(const std::string& name, const std::string& alias);
+
+    //! Return a vector with isomers names matching a given composition map
+    //!     @param compMap compositionMap of the species.
+    //!     @return A vector of species names for matching species.
+    virtual std::vector<std::string> findIsomers(const compositionMap& compMap) const;
+
     //! Return the Species object for the named species. Changes to this object
     //! do not affect the ThermoPhase object until the #modifySpecies function
     //! is called.
--- a/include/cantera/thermo/Phase.h
+++ b/include/cantera/thermo/Phase.h
@@ -716,6 +716,11 @@ public:
     //!     @return A vector of species names for matching species.
     virtual std::vector<std::string> findIsomers(const compositionMap& compMap) const;
 
+    //! Return a vector with isomers names matching a given composition string
+    //!     @param comp String containing a composition map
+    //!     @return A vector of species names for matching species.
+    virtual std::vector<std::string> findIsomers(const std::string& comp) const;
+
     //! Return the Species object for the named species. Changes to this object
     //! do not affect the ThermoPhase object until the #modifySpecies function
     //! is called.
--- a/interfaces/cython/cantera/thermo.pyx
+++ b/interfaces/cython/cantera/thermo.pyx
@@ -524,6 +524,22 @@ cdef class ThermoPhase(_SolutionBase):
         if self.kinetics:
             self.kinetics.invalidateCache()
 
+    def add_species_alias(self, name, alias):
+        """
+        Add the alternate species name *alias* for an original species *name*.
+        """
+        self.thermo.addSpeciesAlias(stringify(name), stringify(alias))
+
+    def find_isomers(self, comp):
+        """
+        Find species/isomers matching a composition specified by *comp*.
+        """
+
+        assert isinstance(comp, dict), 'Composition needs to be specified as dictionary'
+        iso = self.thermo.findIsomers(comp_map(comp))
+
+        return [pystr(b) for b in iso]
+
     def n_atoms(self, species, element):
         """
         Number of atoms of element *element* in species *species*. The element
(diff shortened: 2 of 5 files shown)


ischoegl:
Comment: the 'species_alias' could be implemented as a property (dictionary), but I'm not sure whether this is overkill.

src/thermo/Phase.cpp
```diff
@@ -843,6 +843,35 @@ void Phase::modifySpecies(size_t k, shared_ptr<Species> spec)
...
 }
 
+bool Phase::addSpeciesAlias(const std::string& name, const std::string& alias)
+{
+    size_t k = speciesIndex(name);
+    if (k != npos) {
+        m_speciesIndices[alias] = k;
+        return true;
+    } else {
+        return false;
```
speth: I don't really understand the logic of failing quietly here. When would you want that behavior? And why is it specific to the C++ interface, while the Python interface raises an exception?

src/thermo/Phase.cpp
```diff
@@ -843,6 +843,35 @@ void Phase::modifySpecies(size_t k, shared_ptr<Species> spec)
...
+    }
+}
+
+vector<std::string> Phase::findIsomers(const compositionMap& compMap) const
+{
+    vector<std::string> isomersVector;
```
speth: Minor suggestion: `isomerNames` would be a more descriptive name.

include/cantera/thermo/Phase.h
```diff
@@ -704,6 +704,24 @@ class Phase
      */
     virtual void modifySpecies(size_t k, shared_ptr<Species> spec);
 
+    //! Add a species alias (i.e. user-defined alternative species name).
+    //! Aliases are case-sensitive.
+    //!     @param name original species name std::string.
+    //!     @param alias alternate name std::string.
+    //!     @return `true` if the alias was successfully added
+    //!             (i.e. the original species name is found)
+    virtual bool addSpeciesAlias(const std::string& name, const std::string& alias);
```
speth: I'm wondering whether it's necessary for these methods to be virtual. Can you think of a case where overriding them in a derived class would be necessary?

interfaces/cython/cantera/thermo.pyx
```diff
@@ -517,6 +517,29 @@ cdef class ThermoPhase(_SolutionBase):
...
+        """
+
+        if isinstance(comp, dict):
+            iso = self.thermo.findIsomers(comp_map(comp))
+        elif isinstance(comp, (str, bytes)):
+            iso = self.thermo.findIsomers(stringify(comp))
+        else:
+            raise CanteraError('Invalid composition')
+
+        return [b.decode('utf-8') for b in iso]
```
speth:
```suggestion
        return [pystr(b) for b in iso]
```

src/thermo/Phase.cpp
```diff
@@ -843,6 +843,35 @@ void Phase::modifySpecies(size_t k, shared_ptr<Species> spec)
     invalidateCache();
 }
 
+bool Phase::addSpeciesAlias(const std::string& name, const std::string& alias)
+{
+    size_t k = speciesIndex(name);
+    if (k != npos) {
+        m_speciesIndices[alias] = k;
```
speth, line 863: One issue with this implementation that I noticed is that you can overwrite the entry that corresponds to a species, leaving it without a way to be referenced by name.

speth (changes requested):
The addition of `findIsomers` makes sense to me.

I don't see much of a direct use case for `addSpeciesAlias`, but I can imagine ways of using it if you had some extra data in your input file, e.g. the SMILES corresponding to each species, so adding some underlying infrastructure for this kind of feature seems fine.

interfaces/cython/cantera/thermo.pyx
ischoegl: Ah. That's the helper function I was missing!

include/cantera/thermo/Phase.h
ischoegl: No. I.e. can be removed.

src/thermo/Phase.cpp
ischoegl: Good suggestion!

src/thermo/Phase.cpp
ischoegl: True. Moving everything to the C++ layer, and making return type void.

src/thermo/Phase.cpp
ischoegl: Good catch. This is indeed the case. I added a check to make sure that `alias` does not already exist. That way, it is guaranteed that a new entry is created in the `m_speciesIndices` map.

ischoegl:
> I don't see much of a direct use case for `addSpeciesAlias` [...]

Sometimes I find large molecules with non-intuitive species names, and keep having to remind myself what they actually are. Those names are also not consistent among groups; as `m_speciesIndices` is already a map, it's pretty simple to add this feature.

speth (approved):

PR #719, opened 2019-10-03
Differentiate quality from mole fraction in Python

ischoegl:

Fixes: N/A; but follows discussion in review of #687:

> > [...] it may make sense to use Q for quality at some point in the future. [...]

> This would also be consistent with other naming conventions that have been adopted in the Python module, e.g. D for density instead of R for Rho. [...]

- Use `Q` instead of `X` for quality (vapor fraction) 
- Deprecate all instances using `X` in `PureFluid`
- ~In `SolutionArray.collect`, the species name is appended in labels, i.e. `Q_H2O`~

I.e.
```python
In [1]: import cantera as ct

In [2]: w = ct.Water()

In [3]: w.TX = 400, .5
DeprecationWarning: To be removed after Cantera 2.5. Attribute renamed to 'TQ'

In [4]: w.TQ
Out[4]: (400.0, 0.5)
```

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -481,22 +481,12 @@ class SolutionArray:
         'transport_model',
     ]
 
-    _all_states = [
-        # all setter/getter combos defined by the ThermoPhase base class
-        'TD', 'TDX', 'TDY', 'TP', 'TPX', 'TPY', 'UV', 'UVX', 'UVY',
-        'DP', 'DPX', 'DPY', 'HP', 'HPX', 'HPY', 'SP', 'SPX', 'SPY',
-        'SV', 'SVX', 'SVY'
-    ]
-
     _interface_passthrough = ['site_density']
     _interface_n_species = ['coverages']
 
     _purefluid_scalar = ['Q']
 
     def __init__(self, phase, shape=(0,), states=None, extra=None):
-        super().__setattr__('_setters', {})
-        super().__setattr__('_getters', {})
-        super().__setattr__('_extra', {})
         self._phase = phase
 
         if isinstance(shape, int):
@@ -521,6 +511,7 @@ class SolutionArray:
             self._indices = list(np.ndindex(self._shape))
             self._output_dummy = self._states[..., 0]
 
+        self._extra = {}
         if isinstance(extra, dict):
             for name, v in extra.items():
                 if not np.shape(v):
@@ -540,45 +531,17 @@ class SolutionArray:
                 " supplied in a dict if the SolutionArray is not initially"
                 " empty")
 
-        # add properties dynamically for states not defined by ThermoPhase
-        state_sets = set(phase._full_states.values()) | set(phase._partial_states.values())
-        if isinstance(self._phase, PureFluid):
-            # add deprecated setters
-            # @todo: remove .. deprecated:: 2.5
-            state_sets |= {s.replace('Q', 'X') for s in state_sets if 'Q' in s}
-
-        state_sets = state_sets - set(self._all_states)
-        for name in state_sets:
-            ph = type(phase)
-            if len(name) == 2:
-                getter, setter = _state2_prop(name, ph)
-            elif len(name) == 3:
-                getter, setter = _state3_prop(name, ph)
-            else:
-                raise NotImplementedError("Failed adding property '{}' for "
-                                          "phase '{}'".format(name, phase))
-            self._getters[name] = getter
-            self._setters[name] = setter
-
     def __getitem__(self, index):
         states = self._states[index]
         shape = states.shape[:-1]
         return SolutionArray(self._phase, shape, states)
 
     def __getattr__(self, name):
-        if name in self._getters:
-            return self._getters[name](self)
-        elif name in self._extra:
+        if name in self._extra:
             return np.array(self._extra[name])
         else:
             raise AttributeError("'{}' object has no attribute '{}'".format(
                 self.__class__.__name__, name))
-
-    def __setattr__(self, name, value):
-        if name in self._setters:
-            self._setters[name](self, value)
-        else:
-            super().__setattr__(name, value)
 
     def __call__(self, *species):
         return SolutionArray(self._phase[species], states=self._states,
@@ -715,8 +678,7 @@ class SolutionArray:
                 break
         if mode == '':
             # concentration/quality specifier ('X' or 'Y') is not used
-            states = [st.translate({ord(i): None for i in 'XY'})
-                      for st in states]
+            states = [st.rstrip('XY') for st in states]
         elif len(valid_species) != len(all_species):
             incompatible = list(set(valid_species) ^ set(all_species))
             raise ValueError('incompatible species information for '
@@ -1009,13 +971,16 @@ def _state2_prop(name, doc_source):
     return getter, setter
 
 
-def _state3_prop(name, doc_source):
+def _state3_prop(name, doc_source, scalar=False):
     # Factory for creating properties which consist of a tuple of three
     # variables, e.g. 'TPY' or 'UVX'
     def getter(self):
         a = np.empty(self._shape)
         b = np.empty(self._shape)
-        c = np.empty(self._shape + (self._phase.n_selected_species,))
+        if scalar:
+            c = np.empty(self._shape)
+        else:
+            c = np.empty(self._shape + (self._phase.n_selected_species,))
         for index in self._indices:
             self._phase.state = self._states[index]
             a[index], b[index], c[index] = getattr(self._phase, name)
@@ -1045,15 +1010,33 @@ def _state3_prop(name, doc_source):
 def _make_functions():
     # this is wrapped in a function to avoid polluting the module namespace
 
-    # state setters
-    for name in SolutionArray._all_states:
-        if len(name) == 2:
-            getter, setter = _state2_prop(name, Solution)
-        elif len(name) == 3:
-            getter, setter = _state3_prop(name, Solution)
-        doc = getattr(Solution, name).__doc__
-        prop = property(getter, setter, doc=doc)
-        setattr(SolutionArray, name, prop)
+    names = []
+    for ph, ext in [(ThermoPhase, 'XY'), (PureFluid, 'Q')]:
+
+        # all state setters/getters are combination of letters
+        setters = 'TDPUVHS' + ext
+        scalar = ext == 'Q'
+
+        # add deprecated setters for PureFluid (e.g. PX/TX)
+        # @todo: remove .. deprecated:: 2.5
+        setters = setters.replace('Q', 'QX')
+
+        # obtain setters/getters from thermo objects
+        all_states = [k for k in ph.__dict__
+                      if not set(k) - set(setters) and len(k)>1]
+
+        # state setters (copy from ThermoPhase objects)
+        for name in all_states:
+            if name in names:
+                continue
+            names.append(name)
+            if len(name) == 2:
+                getter, setter = _state2_prop(name, ph)
+            elif len(name) == 3:
+                getter, setter = _state3_prop(name, ph, scalar)
+            doc = getattr(ph, name).__doc__
+            prop = property(getter, setter, doc=doc)
+            setattr(SolutionArray, name, prop)
 
     # Functions which define empty output arrays of an appropriate size for
     # different properties
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -491,7 +491,7 @@ class SolutionArray:
     _interface_passthrough = ['site_density']
     _interface_n_species = ['coverages']
 
-    _purefluid_n_species = ['Q']
+    _purefluid_scalar = ['Q']
 
     def __init__(self, phase, shape=(0,), states=None, extra=None):
         super().__setattr__('_setters', {})
@@ -542,6 +542,11 @@ class SolutionArray:
 
         # add properties dynamically for states not defined by ThermoPhase
         state_sets = set(phase._full_states.values()) | set(phase._partial_states.values())
+        if isinstance(self._phase, PureFluid):
+            # add deprecated setters
+            # @todo: remove .. deprecated:: 2.5
+            state_sets |= {s.replace('Q', 'X') for s in state_sets if 'Q' in s}
+
         state_sets = state_sets - set(self._all_states)
         for name in state_sets:
             ph = type(phase)
@@ -688,16 +693,13 @@ class SolutionArray:
         states = list(self._phase._full_states.values())
 
         # add partial and/or potentially non-unique state definitions
-        if isinstance(self._phase, PureFluid):
-            states += ['TPQ']
         states += list(self._phase._partial_states.values())
 
         # determine whether complete concentration is available (mass or mole)
-        # assumes that 'X', 'Y' or 'Q' is always in last place
-        # (for PureFluid, 'Q_H2O' is treated in a similar fashion)
+        # assumes that 'X' or 'Y' is always in last place
         mode = ''
         valid_species = {}
-        for prefix in ['X_', 'Y_', 'Q_']:
+        for prefix in ['X_', 'Y_']:
             if not any([prefix[0] in s for s in states]):
                 continue
             spc = ['{}{}'.format(prefix, s) for s in self.species_names]
@@ -712,8 +714,8 @@ class SolutionArray:
                 states = [v[:-1] for v in states if mode in v]
                 break
         if mode == '':
-            # concentration/quality specifier ('X', 'Y' or 'Q') is not used
-            states = [st.translate({ord(i): None for i in 'XYQ'})
+            # concentration/quality specifier ('X' or 'Y') is not used
+            states = [st.translate({ord(i): None for i in 'XY'})
                       for st in states]
         elif len(valid_species) != len(all_species):
             incompatible = list(set(valid_species) ^ set(all_species))
@@ -722,7 +724,7 @@ class SolutionArray:
 
         # determine suitable thermo properties for reconstruction
         basis = 'mass' if self.basis == 'mass' else 'mole'
-        prop = {'T': ('T'), 'P': ('P'),
+        prop = {'T': ('T'), 'P': ('P'), 'Q': ('Q'),
                 'D': ('density', 'density_{}'.format(basis)),
                 'U': ('u', 'int_energy_{}'.format(basis)),
                 'V': ('v', 'volume_{}'.format(basis)),
@@ -846,8 +848,8 @@ class SolutionArray:
             elif c in species_names:
                 single_species = True
                 collabels = ['{}_{}'.format(species, c)]
-            elif c in self._purefluid_n_species:
-                collabels = ['{}_{}'.format(c, s) for s in self.species_names]
+            elif c in self._purefluid_scalar:
+                collabels = [c]
             else:
                 raise CanteraError('property "{}" not supported'.format(c))
 
@@ -1090,8 +1092,8 @@ def _make_functions():
     for name in SolutionArray._interface_n_species:
         setattr(SolutionArray, name, make_prop(name, empty_species, Interface))
 
-    for name in SolutionArray._purefluid_n_species:
-        setattr(SolutionArray, name, make_prop(name, empty_species, PureFluid))
+    for name in SolutionArray._purefluid_scalar:
+        setattr(SolutionArray, name, make_prop(name, empty_scalar, PureFluid))
 
     for name in SolutionArray._n_total_species:
         setattr(SolutionArray, name,
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -491,6 +491,8 @@ class SolutionArray:
     _interface_passthrough = ['site_density']
     _interface_n_species = ['coverages']
 
+    _purefluid_n_species = ['Q']
+
     def __init__(self, phase, shape=(0,), states=None, extra=None):
         super().__setattr__('_setters', {})
         super().__setattr__('_getters', {})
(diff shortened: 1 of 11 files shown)


interfaces/cython/cantera/thermo.pyx
```diff
@@ -1440,40 +1453,93 @@ cdef class PureFluid(ThermoPhase):
     """
 
     _full_states = {frozenset(k): k
-                    for k in ('TD', 'TP', 'UV', 'DP', 'HP', 'SP', 'SV', 'TX',
-                              'PX', 'ST', 'TV', 'PV', 'UP', 'VH', 'TH', 'SH')}
+                    for k in ('TD', 'TP', 'UV', 'DP', 'HP', 'SP', 'SV', 'TQ',
```
ischoegl: I am wondering whether `TP`, `TQ` and `PQ` should be removed from here, as they are not necessarily full states.

interfaces/cython/cantera/composite.py
```diff
@@ -631,14 +633,15 @@ def restore_data(self, data, labels):
...
-                           if fs not in last] + ['TPX'] + last
+                           if fs not in last] + ['TPQ'] + last
 
         # determine whether complete concentration is available (mass or mole)
-        # assumes that `X` or `Y` is always in last place
+        # assumes that 'X' or 'Y' is always in last place
+        # (for PureFluid, 'Q_H2O' is treated in a similar fashion)
         mode = ''
-        for prefix in ['X_', 'Y_']:
+        for prefix in ['X_', 'Y_', 'Q_']:
```
speth: Given that PureFluid is a single-phase species, and I don't think we've worked out an equivalent to the quality for multi-species two-phase mixtures, I think the quality can just be stored as `Q` rather than `Q_speciesName`.

interfaces/cython/cantera/test/test_purefluid.py
```diff
@@ -142,25 +144,88 @@ def test_thermal_expansion_coeff_TD(self):
...
+            self.assertTrue(issubclass(w[-1].category, DeprecationWarning))
+            self.assertTrue("renamed to 'TQ'" in str(w[-1].message))
+
+            X = self.water.X
+            self.assertTrue(issubclass(w[-1].category, DeprecationWarning))
+            self.assertTrue("renamed to 'Q'" in str(w[-1].message))
+
+            self.water.X = X
+            self.assertTrue(issubclass(w[-1].category, DeprecationWarning))
+            self.assertTrue("renamed to 'Q'" in str(w[-1].message))
```
speth: This test would succeed even if the statement `self.water.X = X` did not issue a `DeprecationWarning`. Ditto for the other paired warnings. That said, I'm not sure it's really necessary to test every single deprecation warning.

interfaces/cython/cantera/test/test_purefluid.py
```diff
@@ -142,25 +144,88 @@ def test_thermal_expansion_coeff_TD(self):
...
+    def test_deprecated_X(self):
+
+        with warnings.catch_warnings(record=True) as w:
+
+            # cause all warnings to always be triggered.
+            warnings.simplefilter("always")
+
+            self.water.TX = 400, 0.8
+            self.assertTrue(issubclass(w[-1].category, DeprecationWarning))
+            self.assertTrue("renamed to 'TQ'" in str(w[-1].message))
```
speth: Please use the specialized asserts rather than `assertTrue` where possible.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -810,6 +810,19 @@ cdef class ThermoPhase(_SolutionBase):
...
+    property Q:
+        """
+        Placeholder for derived classes with phase transition
+        """
+        def __get__(self):
+            raise NotImplementedError("Attribute 'Q' (vapor fraction) is "
+                                      "not implemented/defined for this phase")
+        def __set__(self, Q):
+            raise NotImplementedError("Attribute 'Q' (vapor fraction) is "
+                                      "not implemented/defined for this phase")
```
speth: I would like to avoid introducing this kind of placeholder. I think the only reason it's needed is for how the `Q` property is handled in `SolutionArray`, but that could be treated like we do other properties that only appear in derived classes, e.g. `InterfacePhase.coverages`.

speth (changes requested):
I'm fully on board with this renaming. I think it will be worth it to avoid the confusion caused by the overloading of X as both mole fraction and vapor fraction depending on the phase type.

interfaces/cython/cantera/composite.py
ischoegl: ~With this PR, the mole fraction `X_H2O` will become accessible again (instead of being masked by the quality.~ (actually not, as exceptions are raised). Still, for consistency, I'd prefer to keep the old behavior (it's also much easier to follow the implementation as it just uses what's already there for `X_` and `Y_`). `SolutionArray` will add the species name for other single species mass/mole fractions (which are by definition 1). So if this is done, it applies to `X`, `Y` and `Q` regardless.

interfaces/cython/cantera/test/test_purefluid.py
ischoegl: Yes. Tried to compress this and allowed for that behavior. Fixed.

interfaces/cython/cantera/composite.py
ischoegl: see:
https://github.com/Cantera/cantera/blob/54c12fc59cfd10a6da2bb7cb81980cd1f7a24696/interfaces/cython/cantera/composite.py#L796-L797

speth: I think if you treat `Q` like another scalar variable (still unique to the `PureFluid` class, of course), rather than an "n_species" variable, this should work out. I don't have any problem with the per-species variables including `X` and `Y` still ending up like `X_speciesName`.

speth (commented):
While trying to experiment with getting `Q` to work as the column name for `SolutionArray` output, I ran into a problem, and haven't quite been able to figure out why it's failing:
```python
>>> import cantera as ct
>>> p = ct.Water()
>>> s = ct.SolutionArray(p, 5)
>>> data, cols = s.collect_data()
>>> s.restore_data(data, cols)
```
```
<ipython-input-5-02284ff2df78> in <module>()
----> 1 s.restore_data(data, cols)

~/src/cantera/build/python/cantera/composite.py in restore_data(self, data, labels)
    704             raise ValueError(
    705                 "invalid/incomplete state information (detected "
--> 706                 "partial information as mode='{}')".format(mode)
    707             )
    708

ValueError: invalid/incomplete state information (detected partial information as mode='Y')
```

ischoegl:
Hm, I can reproduce this. I will have to look at this, as it looks like the standard `TD` state (which is the default) is causing trouble. (Also: likely a pre-existing condition unrelated to the switch to `Q`.)
```python
In [1]: import cantera as ct
   ...: p = ct.Water()
   ...: s = ct.SolutionArray(p, 5)
   ...: data, cols = s.collect_data()

In[2]: cols
Out[2]: ['T', 'density', 'Y_H2O']
```
Will add some additional unit tests to make sure that there aren't any surprises.

ischoegl:
@speth ... on second look, the fix is already included in #720, which allows for different standard states (up until now, the standard state is defined as `TDY` regardless of the phase model; PR #720 sets it to `TD` for `PureFluid`).

So, would you mind tabling this until #720 is discussed & merged? (I recall you mentioning that there may be several issues that may need to be converged).

speth:
Ah, I didn't realize that it was a pre-existing issue, but that makes sense, and I'm happy to see that it's already fixed by #720.

ischoegl:
@speth ... I confirmed that the bug you found was indeed pre-existing, but also found a second issue that prevented #720 from fixing it directly (the additional fix is now added in ce424c9). This is added to #720, as a fix of #644 (and associated issues) is required regardless. I also added a unit test to confirm that things are working as expected.

ischoegl:
Adding an observation from #720 here (per @speth's findings)
```python
w = ct.Water()
a = ct.SolutionArray(gas, 10)
'TX' in dir(a)  # False, as expected
b = ct.SolutionArray(w)
'TX' in dir(b)  # True, as expected
'TX' in dir(a)  # True, as expected, but not desired
a.TX  # unexpectedly (but as desired) raises AttributeError
```
Once `TX` and `TQ` are differentiated, all should have the expected/desired outcome.

ischoegl:
**Remaining steps:**

 * [x] rebase to incorporate new features introduced in #720 
 * [x] ensure that `data, cols = s.collect_data()` yields `['T', 'density']` 
 * [x] enforce single-letter property acronyms by changing signature of `Phase::nativeState` from `std::map<std::string, size_t>` to `std::map<char, size_t>` 
 * [x] confirm that `SolutionArray` has correct setters (i.e. includes `TPQ` etc.)
 * [x] introduce `Phase::isFluid` flag to facilitate identification of substances with phase change

interfaces/cython/cantera/composite.py
ischoegl: I believe there is a future use case for multi-component fluids (i.e. *not* pure fluids), that will have dissimilar vapor pressures (and thus different `Q` values). From that perspective, I'd suggest to keep the species name attached and treat vapor fraction as an `_n_species` variable.

ischoegl:
@speth after rebasing to #720, `SolutionArray` appears to be working as expected.
```
In [1]: import cantera as ct
   ...: w = ct.Water()
   ...: a = ct.SolutionArray(gas, 10)
   ...: b = ct.SolutionArray(w, 2)
   ...: 

In [2]: [st in dir(w) for st in ['TX', 'TQ']] # expected
Out[2]: [True, True]

In [3]: [st in dir(a) for st in ['TX', 'TQ']] # expected
Out[3]: [False, False]

In [4]: [st in dir(b) for st in ['TX', 'TQ']] # unexpected? <<<<<<<<<<<<<<<<<
Out[4]: [False, False]

In [5]: b.TQ # expected
Out[5]: (array([ 300.,  300.]), array([ 0.,  0.]))

In [6]: b.TQ = 400, .5

In [7]: b.TQ
Out[7]: (array([ 400.,  400.]), array([ 0.5,  0.5]))
```
I.e. I believe this is ready for a review.

speth (changes requested):
Currently, this eliminates the the `TX` and `PX` attributes of `SolutionArray` without a (runtime) deprecation warning. I also noticed that the `X` and `TPX` properties of `SolutionArray` don't seem to emit any warnings for pure fluids.

I'm not really sold on making `nativeState` work with `char` instead of `std::string`. The efficiency benefit shouldn't ever matter, and I don't want to assume that there will always be a clear one-character name for any state variable. For example, what letter would you use for the electron temperature in a non-thermal plasma?

The idea of having `Q` be dimensioned as `nSpecies` to support more general multi-phase fluids is an interesting one, but it's a bit inconsistent here -- the value returned by `PureFluid.Q` is a scalar, while `SolutionArray.Q` returns an array with `nSpecies` in the last dimension, and the second value returned by `SolutionArray.TQ` is not. I realize that this is an inconsistency inherited from the previous mis-use of  `X`, but that's not a great reason to keep it. I'd like to avoid making the interface to the simple case that we already have (i.e. single-component fluids) less convenient in support of a capability that is mostly theoretical at present (would Q even be a useful state setter in such a case?).

Regarding `isFluid`, I'm not really sure I see what the use case is, but if we really want such a flag, then the name is going to need some work. `isMultiphase` would be one option, but I'd worry about distinguishing between "this model includes phase changes" and "the current state involves multiple phases".

ischoegl:
@speth ... thanks for the review. I'm responding to your points before addressing major issue.

> this eliminates the the `TX` and `PX` attributes of `SolutionArray` without a (runtime) deprecation warning. I also noticed that the `X` and `TPX` properties of `SolutionArray` don't seem to emit any warnings for pure fluids.

Yes, I just confirmed this. I hadn't anticipating this as I thought that the unit tests would catch - turns out that I was only testing deprecation of `PureFluid`. Need to pick over 750a80b to see what's going on.

> I'm not really sold on making `nativeState` work with `char` instead of `std::string`. 

I dropped the commit.

> The idea of having `Q` be dimensioned as `nSpecies` to support more general multi-phase fluids is an interesting one, but it's a bit inconsistent here [...] I'd like to avoid making the interface to the simple case that we already have (i.e. single-component fluids) less convenient [...]

I want to believe that Cantera should be able to handle realistic multi-fuel vaporization applications eventually. My thought here was to exactly replicate the behavior of mass/mole fractions. I don't see much of an inconvenience as the only instance where `Q_H2O` etc. is visible is in HDF/CSV output. (It'd still be easy to suppress it for a *pure* substance, while keeping it as `n_species`. Since `X` and `Y` are gone for pure species input/output, there isn't any inconsistency.)

> Regarding `isFluid`, I'm not really sure I see what the use case is, but if we really want such a flag, then the name is going to need some work.

I had `PureFluid` in mind, which would have both `isPure` and `isFluid` set to `true`. I do, however, see your point, and would just name it `hasPhaseTransition`. The reason why I'd like to see this flag being added is that for various equations of state, it would be immediately clear whether phase transitions are implemented or not. Some real gas EOS are capable of predicting the regions, but I wouldn't assume that all cases are covered.

speth:
> I don't see much of an inconvenience as the only instance where `Q_H2O` etc. is visible is in HDF/CSV output.

I don't think `Q_H2O` showing up in the CSV/HDF5 output is particularly problematic. What I'm more concerned about is what is returned by `Solution.Q` when the underlying phase is, say, either `pure-fluid` or `Redlich-Kwong`. Currently, this is a `float` for the former, while the latter is not implemented even at the C++ level, but if were, would presumably need to return an array of length `n_species`, But I don't think we want the return type to vary with the number of species, and I'd rather not have to write `water.Q[0]` to get the vapor fraction.

I don't think the vapor fraction of each species in a multiphase mixture can be used as part of a state setter in the same way as it can for a single-species phase, since it would over-specify the system per the Gibbs' Phase Rule, so I see this as more of a calculated property than part of the state. It could make sense to have two separate properties, one for the total vapor fraction across all species and one for the fraction of each species in the vapor phase. This would be analogous to what we do for many other properties, e.g. `partial_molar_enthalpies` versus `enthalpy_mass`, where the latter is available as part of state setters like `HP`.

> The reason why I'd like to see this flag being added is that for various equations of state, it would be immediately clear whether phase transitions are implemented or not

OK, I can see that being useful for cases like `WaterSSTP` where presumably the correct return value is `false`.

ischoegl:
> I don't think the vapor fraction of each species in a multiphase mixture can be used as part of a state setter in the same way as it can for a single-species phase, since it would over-specify the system per the Gibbs' Phase Rule, so I see this as more of a calculated property than part of the state. 

This is a valid point: a scalar it is.

> ... `WaterSSTP`, `Redlich-Kwong`, etc.

Exactly - it's often not immediately clear what's implemented. I hope we'll be able to set `hasPhaseTransition` to true for more than just `PureFluid` eventually (same for the pending `Peng-Robinson` PR).

ischoegl:
@speth ... per discussion, I switched `isFluid` to `hasPhaseTransition` and made `Q` a scalar.

~The remaining issue is that it appears that setters/getters defined by~
```
getter, setter = _state2_prop(name, Solution)
```
~don't overload in derived classes? Will get back to it next week.~

interfaces/cython/cantera/composite.py
ischoegl: resolved ... made `Q` a scalar

ischoegl:
@speth ... I believe 174c9b7 addresses setters/getters not being defined correctly, i.e. almost all points previously raised are now addressed.

The lack of warnings is curious: I verified that the functions that should raise the warnings are indeed called. I found another context where warnings (and even errors!) are not raised, which *may* be related: see #746

bryanwweber:
@ischoegl For warnings at least, the warning is only shown the first time a `method` is passed... i.e., the signature of the function is `warn(method, message)`, where the `method` is stored and the message is only shown the first time that `method` is encountered. This is done in C++ so applies to all the interfaces. Not sure if that's the issue you're encountering (and certainly, this is not the resolution to #746).

ischoegl:
>  For warnings at least, the warning is only shown the first time a method is passed...

Correct. The warnings here are, however, raised in `pyx` files, so the issue cannot be caused by the C++ implementation.

ischoegl:
@bryanwweber and @speth ... documenting the lack of warning (which I believe to be a bug that goes beyond this PR)
```
In [1]: import cantera as ct
   ...: w = ct.Water()
   ...: w.TQ = 300., .5
   ...: a = ct.SolutionArray(gas, 10)
   ...: b = ct.SolutionArray(w, 2)
   ...: 

In [2]: b.TPX # <--- no warning issued
Out[2]: 
(array([ 300.,  300.]), array([ 3528.21380229,  3528.21380229]), array([[ 0.5],
        [ 0.5]]))

In [3]: w.TPX # <--- calling the method raising the warning directly
/usr/bin/ipython3:1: DeprecationWarning: To be removed after Cantera 2.5. Attribute renamed to 'TPQ'
  #! /bin/sh
Out[3]: (300.0, 3528.213802290444, 0.5000000000000002)

In [4]: b.TPQ
Out[4]: 
(array([ 300.,  300.]),
 array([ 3528.21380229,  3528.21380229]),
 array([ 0.5,  0.5]))
```

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -36,6 +36,11 @@ def test_load_thermo_models(self):
                 self.assertEqual(sol.T, T0)
                 self.assertEqual(sol.P, p0)
 
+                if sol.thermo_model in ('PureFluid'):
```
speth: While this happens to work, `('PureFluid')` isn't a tuple -- it should be `('PureFluid',)` instead.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1533,6 +1534,20 @@ def test_get_state(self):
         self.assertEqual(P.shape, (4,))
         self.assertEqual(Y.shape, (4, self.gas.n_species))
 
+    @unittest.expectedFailure
+    def test_idealgas_getters(self):
```
speth, line 1536: I'm having trouble understanding this test. Why is it an expected failure? What behavior is this test trying to test?

interfaces/cython/cantera/composite.py
```diff
@@ -538,40 +531,20 @@ def __init__(self, phase, shape=(0,), states=None, extra=None):
...
         else:
             raise AttributeError("'{}' object has no attribute '{}'".format(
                 self.__class__.__name__, name))
 
     def __setattr__(self, name, value):
-        if name in self._setters:
-            self._setters[name](self, value)
-        else:
-            super().__setattr__(name, value)
+        super().__setattr__(name, value)
```
speth: At this point, the `__setattr__` override can just be deleted, right? Likewise, the initialization of `self._extra` can be restored to its simpler form.

interfaces/cython/cantera/composite.py
```diff
@@ -710,16 +680,17 @@ def restore_data(self, data, labels):
                 states = [v[:-1] for v in states if mode in v]
                 break
         if mode == '':
-            # concentration specifier ('X' or 'Y') is not used
-            states = [s for s in states if 'X' not in s and 'Y' not in s]
+            # concentration/quality specifier ('X' or 'Y') is not used
+            states = [st.translate({ord(i): None for i in 'XY'})
+                      for st in states]
```
speth: A simpler alternative:
```python
states = [st.rstrip('XY') for st in states]
```

interfaces/cython/cantera/thermo.pyx
```diff
@@ -1647,76 +1707,183 @@ cdef class PureFluid(ThermoPhase):
...
+        def __get__(self):
+            return self.u, self.v, self.Q
 
     property DPX:
+        """Get the density [kg/m^3], pressure [Pa], and vapor fraction.
+
+        .. deprecated:: 2.5
+
+             To be deprecated with version 2.5, and removed thereafter.
+             Renamed to `DPQ`.
```
speth: Would it be worth noting in these deprecation messages that the change in behavior after Cantera 2.5 won't be that this property doesn't exist, but that it will return the mole fraction instead of the vapor fraction? Likewise for the other properties that are shared with `ThermoPhase`.

speth (changes requested):
Thanks for the updates. This is looking pretty good to me.

I think there are some examples that need to be updated to use these new properties. `rankine.py` comes to mind immediately, but there may be others.

interfaces/cython/cantera/test/test_thermo.py
ischoegl: This ensures that all getters can actually be used, i.e.
```python
        getters = 'TDPUVHSXY'
        all_getters = [k for k in states.__dir__()
                       if not set(k) - set(getters) and len(k)>1]
```
this would not be true for a setter `TQ`, which is part of `states__dir__()`, but will fail with an attribute error. I added comments.

In a nutshell, these unit tests probe edge cases that previously caused trouble (where getters/setters did not work as expected).

interfaces/cython/cantera/composite.py
ischoegl: absolutely.

ischoegl:
@speth ... made all requested changes and updated `rankine.py` (couldn't find other examples that are affected)

interfaces/cython/cantera/test/test_thermo.py
speth: OK, that makes sense, but I don't think that's what this test achieves. What will happen here is that the first property that works will not raise an `AttributeError` and then test will fail. If you want to test all the getters that should work, you could build that list using the properties of the `ThermoPhase` class. Then you can invert the structure of the test by just eliminating `with` clause.

I think that we should reserve `expectedFailure` for tests that should pass but do not because of a known bug. The only cases where we currently use `expectedFailure` are for some of the equilibrium tests, for which I believe there are open issues in the tracker (although I realize there is unfortunately no comment linking to the relevant issue).

I would also suggest using `dir(foo)` over `foo.__dir__()` and `getattr(foo, bar)` over `foo.__getattribute__(bar)`.

ischoegl: > I think that we should reserve `expectedFailure` for tests that should pass but do not because of a known bug.

Ok. I changed the tests accordingly. 

> What will happen here is that the first property that works will not raise an `AttributeError` and then test will fail. 

Indeed! Thanks for pointing that out - on the upside, the unit test surfaced an old typo (`self.p` instead of `self.P` in `PureFluid.PV`).

speth (approved):

Issue #633, opened 2019-04-30
setMoleFractions violates model assumptions for surfPhase thermo phase


decaluwe:
### Cantera version
2.5.0a2
### Operating System
All

### Expected Behavior
The surfPhase (aka 'ideal_interface', at the cti level) model assumes a fixed site density (i.e. molar density of available sites; `m_n0` in the class, typically something like `gamma_o`, in textbooks).  

### Actual Behavior
Setting the mole fraction independently  (via <phase_name>.X at the Python level) instead assumes that the mass density is constant, and automatically re-sets the site density to maintain this assumption.  While the user can (and should be able to) change the site density by setting it directly, this kind of indirect setting of the site density would seem to violate the model assumptions, and as such seems sub-optimal.  

The method already has a `coverages` method which correctly sets the surface composition while preserving the model assumptions.  I would suggest overwriting surfPhase.setMoleFractions to either point to surfPhase.coverages, or to correctly implement a mole fraction setting routine which holds m_n0 fixed (see comment below, which weighs between these two options).

### Steps to reproduce
1. Load a surface phase `surf`.
2. Print object surf()
3. Set mole fractions using `X`, e.g. 
import numpy as np
surf.X = np.ones_like(surf.X)
4. print object surf() and note that mass density has not changed, even though it should.

decaluwe:
Separate but related question:  technically, the mole fractions are not the same as the coverages.  

- The coverage theta_k is fraction of the available lattice sites occupied by species k.
- The mole fraction X_k is the moles of species k per mole of total surface species present.  

For simple species which occupy only one site, the two are equivalent.  However, if a species is present which occupies more than one lattice site, the total moles of surface species present has decreased.

Anyway, just curious whether or not it is useful to have methods present for both quantities.  i.e. one could overwrite `setMoleFractions` so that it simply points to `setCoverages`, but this is not technically correct.  But is there a use case for setting the mole fractions?  I would suspect yes, but would love to hear input from others.

ischoegl:
This is interesting. Is there any analogy to be made between coverages / mole fractions and the way gas phase `Mixtures` are handled? For the latter, the composition of individual `Solution` components can be queried in terms of mole/mass fractions etc.; the `Mixture` object only provides cumulative`species_moles`, and avoids a composite mole fraction.

speth:
[...]

PR #720, opened 2019-10-04
Make the concept of saving state vectors more general

### LINKED ISSUE #595: save/restore ThermoPhase state assumes density is an independent variable

### LINKED ISSUE #644: Setting of density in `LatticePhase`

ischoegl:

Fixes #595, fixes #644, 

- Define valid state combinations in C++ layer
- Ensure all phases covered by `thermo-models.yaml` can be saved and restored correctly
- Fix is needed for #711 and #693
- Make `SolutionArray` (almost) phase agnostic

E.g. using the examples from #595 and #644, the behavior is now correct, and invalid setters are disabled (or deprecated).
```
In [1]: import cantera as ct
   ...: phase = ct.ThermoPhase('test_problems/cathermo/issp/IdealSolidSolnPhaseExample.xml')
   ...: T0, p0 = phase.TP # T0 = 500, p0 == 101325
   ...: z = phase.state # calls Phase::saveState
   ...: phase.TP = 600, 2*ct.one_atm
   ...: phase.state = z # calls Phase::restoreState
   ...: assert phase.T == T0, 'temperature'
   ...: assert phase.P == p0, 'pressure'
   ...: 

In [2]: phase.TP
Out[2]: (500.0, 101325.0)

In [3]: phase.TD = 500, 12
DeprecationWarning: IdealSolidSolnPhase::setDensity: Function does not change density and will be removed after Cantera 2.5.
[...]
CanteraError: 
CanteraError thrown by IdealSolidSolnPhase::setDensity:
Density is not an independent variable

In [4]: phase._native_state
Out[4]: ('T', 'P', 'Y')

In [5]: ysz = ct.Solution('sofc.cti','oxide_bulk')

In [6]: ysz.TD = 1073, 200
[...]
CanteraError: 
CanteraError thrown by Phase::assertCompressible:
Setter 'setDensity' is not available. Density is not an independent 
variable for 'oxide_bulk' ('Lattice')

In [7]: ysz._default_state
Out[7]: ('T', 'P', 'X')
```


ischoegl:
As an aside, there is one interesting aspect here: from a purely thermodynamic perspective, `P` is not part of the state (same for `IdealGasPhase`, which should only have `T` and `Y`). I.e. `P` is not needed for the reconstruction of a state, and thus should probably not be part of `phase.state`.

ischoegl:
@bryanwweber ... at this point, `Phase::saveState` and `Phase::restoreState` work for all thermo models covered in `thermo-models.yaml`, i.e. this PR is ready for review (I believe this fix addresses all potential issues connected to #595 you may run into for #693). 

__Note:__ other phases not covered by this PR can be very easily added by overloading `Phase::isIncompressible()`, `Phase::isStoichPhase()` and `Phase::defaultState()`. For the latter, I assumed that  mass base (`Y`) is the default for compositions, with exception of lattice-related phases, which use mole base (`X`).

include/cantera/thermo/FixedChemPotSSTP.h
```diff
@@ -182,6 +182,15 @@ class FixedChemPotSSTP : public SingleSpeciesTP
         return "FixedChemPot";
     }
 
+    virtual bool isStoichPhase() const {
+        return true;
+    }
+
+    //! Return whether phase represents an incompressible substance
```
ischoegl: delete comment

src/thermo/PureFluidPhase.cpp
```diff
@@ -77,6 +77,12 @@ void PureFluidPhase::setParametersFromXML(const XML_Node& eosdata)
     }
 }
 
+std::vector<std::string> PureFluidPhase::fullStates() const
+{
+    return {"TD", "TP", "UV", "DP", "HP", "SP", "SV", "TX",
+            "PX", "ST", "TV", "PV", "UP", "VH", "TH", "SH"};
```
ischoegl: Move `TP`, `TX` and `PX` to `PureFluidPhase::partialStates()`?

ischoegl (commented):
Left comments to avoid another force-push (all are taken care of now).

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -16,6 +21,46 @@ def setUp(self):
...
+
+            ph_name = ph['name']
+            try:
+                sol = ct.Solution(yml_file, ph_name)
+
+                T0, p0 = sol.TP
+                z = sol.state # calls Phase::saveState
+                sol.TP = 300, 2*ct.one_atm
+                sol.state = z # calls Phase::restoreState
+                self.assertTrue(sol.T, T0)
```
ischoegl: assertEqual

ischoegl:
Applied small tweaks. PR is ready for review.

ischoegl:
Applied small fixes.

speth (changes requested):
Conditioning `saveState` and `restoreState` on `isIncompressible` makes sense to me, and is a fine resolution for #595. Strictly speaking, we would probably like to have `saveState` use the variables in `defaultState` and get the kind of behavior that this PR gets in `SolutionArray.colect_data`, but that has the potential to get messy, and I don't think we have any phase types right now where that would really be worthwhile.

Disabling the composition setters for stoichiometric phases seems unnecessary. The mass/mole fractions get normalized anyway, so any value specified there will have no effect in any case.

I'm not sure that it's worth the extra complexity of keeping track of which phases have fixed stoichiometry in order to reduce the size of the state used by `Phase::saveState` and `Phase::restoreState` by one variable. As it is, the the Python `state` property doesn't even handle this -- we would need to add yet another method which returned the size of the state depending on the value of `isStoichPhase`.

I think the best way to disable setting density to fix #644 and other equivalent scenarios is to override the `setDensity` method of the appropriate class at the C++ level. This achieves the desired behavior in all language interfaces, not just Python, and avoids the overhead of the `_check_setter` method.

After that, the only case in which the distinction between the "partial states" and "full states" is made is within `SolutionArray`, where it's still part of a special case, so I'm hoping there's an opportunity to simplify this somewhat.

ischoegl:
@speth ... thanks for the comments!

> Strictly speaking, we would probably like to have `saveState` use the variables in `defaultState` and get the kind of behavior that this PR gets in `SolutionArray.collect_data`, but that has the potential to get messy [...]

That is exactly what this PR is trying to accomplish :grin: (it's actually less messy than I thought at first)

> I'm not sure that it's worth the extra complexity of keeping track of which phases have fixed stoichiometry in order to reduce the size of the state used by `Phase::saveState` and `Phase::restoreState` by one variable. 

That's a minor side-effect of the implementation. I mainly needed to peel off the compositions for stoichiometric phases as this makes things consistent with the `fullStates` definitions (which are essential for `SolutionArray.restore_data`).

Your comment still pointed out one case I hadn't thought off (someone could add `Y` to the output columns of a stoichiometric substance and the import would fail, see `PureFluid`). I'll fix that one and likely add more unit tests for additional phases. Nevertheless, the solution as it stands at this point should already work for most thermo phases (also: this is where the availability of setters is important).

ischoegl:
@speth ... I confirmed that this PR indeed allows `SolutionArray` objects being restored for any of the phases defined in `thermo_models.yaml` based on their `defaultState` definitions (only exception: `'ions-from-neutral-molecule'`). Also, the remaining issue with `PureFluid` is fixed.

I added more detailed unit tests to that effect, i.e. https://github.com/Cantera/cantera/blob/16c5073932005dd0d560e3cb025934e73d1558f8/interfaces/cython/cantera/test/test_models.py#L56

ischoegl:
Edit: the behavior described below is easy to fix (see f8abd72 pushed below)

There's one behavior I observed and am not sure about. For ` StoichSubstance`, the setter for `TPX` does not make sense from the perspective of recovering the state from a minimal set of states (`X` would always be set to 1 regardless).

The current behavior for `StoichSubstance` is that the setter for `TPX` is disabled (i.e. it raises an `AttributeError`). For `SolutionArray`, I followed the past practice of only adding setters that are defined for phases (via `_full_states` etc.). As stoichiometric substances have a reduced set of valid state definitions, the derived `SolutionArray` thus does not define all setters that are defined for `Solution`. In the case of `TPX`, it becomes assignable (!) as a newly created class variable.

This potentially confusing (and pre-exisiting) behavior could be avoided by adding all of the setters that are defined for `Solution` regardless, and not just the ones that are defined by `_full_states`.

ischoegl:
> I think the best way to disable setting density to fix #644 and other equivalent scenarios is to override the `setDensity` method of the appropriate class at the C++ level. 

I can look into that but with #652 in mind it's hard to keep track. As most of the setters are presumably still inherited from `Phase`/`ThermoPhase`, a C++ solution would still involve the equivalent of a `check_setter` method.

ischoegl:
@speth ... all checks are now moved to the C++ layer. 
* `Phase::isCompressible` is checked centrally in `Phase::setDensity`/`Phase::setMolarDensity` where errors are raises as needed: imho this is a more permanent fix than overloading `setDensity` in various derived classes (with ad-hoc behavior). 
* `Phase::isStoichPhase` is still included, and will now trigger a warning if someone tries to change concentrations for a stoichiometric phase (i.e. this is no longer silently ignored due to re-normalization). Unfortunately #746 gobbles up some of those warnings.

include/cantera/thermo/DebyeHuckel.h
```diff
@@ -643,6 +643,7 @@ class DebyeHuckel : public MolalityVPSSTP
      *       independent variable however.
      *
      * @param rho Input density (kg/m^3).
+     * @deprecated To be removed after Cantera 2.5.
```
speth: I think we need to be more specific in this case: that `setDensity` will be non-virtual, and therefore non-overridable, after Cantera 2.5.

src/thermo/PDSS_SSVol.cpp
```diff
@@ -174,6 +174,9 @@ void PDSS_SSVol::setState_TP(doublereal temp, doublereal pres)
 void PDSS_SSVol::setState_TR(doublereal temp, doublereal rho)
 {
     setTemperature(temp);
+    warn_deprecated("PDSS_SSVol::setState_TR",
+        "Setter only changes temperature and "
+        "will be removed after Cantera 2.5.");
```
speth, line 179: I think dealing with the fate of this particular method (and `PDSS::setState_TR` in general) is a bit different from the rest of this PR.

src/thermo/Phase.cpp
```diff
@@ -256,38 +256,149 @@ std::string Phase::speciesSPName(int k) const
...
+        setPressure(state[1]);
+    } else {
         setDensity(state[1]);
+    }
+
+    if (!isStoichPhase()) {
+        if (defaultState()[2] == "Y") {
+            setMassFractions_NoNorm(state + 2);
+        } else {
+            setMoleFractions(state + 2);
```
speth: Why not `setMoleFractions_noNorm`?

src/thermo/Phase.cpp
```diff
@@ -256,38 +256,149 @@ std::string Phase::speciesSPName(int k) const
...
+    }
+    compositionChanged();
+}
+
+void Phase::setStoichMoleFractions()
+{
+    if (isStoichPhase()) {
+        m_y[0] = 1.;
+        m_mmw = m_molwts[0];
+        m_ym[0] = 1./m_mmw;
```
speth: I don't think this method is necessary -- these are precisely the values that these variables take after the first call to `Phase::addSpecies` (see the last block in the function).

src/thermo/Phase.cpp
```diff
@@ -573,7 +689,7 @@ void Phase::setConcentrations(const doublereal* const conc)
         norm += ck;
     }
     m_mmw = sum/norm;
-    setDensity(sum);
+    setConstDensity(sum);
```
speth: Shouldn't `setConcentrations` fail for a constant density phase? Ditto for `setConcentrationsNoNorm`.

src/thermo/Phase.cpp
```diff
@@ -643,6 +761,22 @@ doublereal Phase::molarVolume() const
     return 1.0/molarDensity();
 }
 
+void Phase::setDensity(const doublereal density_)
```
speth:
```suggestion
void Phase::setDensity(const double density_)
```

src/thermo/Phase.cpp
```diff
@@ -643,6 +761,22 @@ doublereal Phase::molarVolume() const
...
+    } else {
+        throw CanteraError("Phase::setDensity()",
+            "density must be positive. density = {}", density_);
+    }
+}
+
+void Phase::setConstDensity(const doublereal density_)
+{
+    m_dens = density_;
+}
```
speth: I think this should at least do the same sanity check as `setDensity` (i.e. that the density is positive). It might also make sense to require that the phase is actually incompressible.

I'm wondering if there's a way we can make this function `protected`, given that almost all of its use is internal to Phase and derived classes -- the only exception is the `LatticeSolidPhase.fromScratch` test, which is a bit arbitrary anyway.

src/thermo/ThermoPhase.cpp
```diff
@@ -194,6 +194,7 @@ void ThermoPhase::setState_HP(double Htarget, double p, double rtol)
 
 void ThermoPhase::setState_UV(double u, double v, double rtol)
 {
+    assertCompressible("setState_UV");
```
speth, line 197: Why not just let this through? An exception will be raised once it tries to set the density inside `setState_HPorUV`, and that avoids running this check twice.

speth (changes requested):
I'm on board with introducing `isIncompressible` (although, as a minor point, I wonder if `isCompressible` would be better from the standpoint of having the normal, compressible case come first and not having to think about the double negative "not incompressible"), and having that supersede the practice of overriding of `setDensity` and `setMolarDensity`. I would suggest that we add a note to `Phase::setDensity` indicating that it will be made non-virtual in the future.

I can see that it is useful to being able to check if a set of properties constitutes a full or partial (settable) thermodynamic state for the phase. However, whether composition information is needed isn't strictly a function of the phase model, just whether or not the phase contains more than one species. For example, an ideal gas with one component can also have its state fully specified by just T and P. So the implementations of `fullStates()` and `partialStates()` could be done without introducing the `isStoichPhase()` method and requiring stoichiometric phases to overridde it.

I see a fair amount of complexity generated by the attempt to avoid setting the composition for single-species phases, and I'm really not seeing the benefit of enforcing this. We wouldn't issue a warning for a single-species ideal gas, so why worry about whether a different phase type with a single species? This makes it harder to write code that works for any phase type, as you have already seen from the special cases added in the equilibrium solvers. And `Reactor` only survives without modification due to the fact that `setMassFractions_noNorm` doesn't call `assertMultiSpecies` for some reason. Likewise, there's a fair amount of simplicity to be gained by always having the length of the state be be `2 + n_species`.

src/thermo/PDSS_SSVol.cpp
ischoegl: True. But I'd still suggest to keep the warning here as not raising a warning creates the illusion that this is a legitimate setter.

src/thermo/Phase.cpp
ischoegl: It is necessary as the standard pathway raises warnings.

src/thermo/Phase.cpp
ischoegl: No. You could argue that there are liquid reacting phases that are nearly incompressible (i.e. using the TPX setter), where an update of the density is still appropriate.

src/thermo/Phase.cpp
ischoegl: Absolutely. Didn't think of this for this PR.

include/cantera/thermo/DebyeHuckel.h
ischoegl: I updated the deprecation messge. I also made the `Phase::density` method non-virtual *now* as it is not overloaded anywhere.

**Note:** `PureFluidPhase` uses a legitimate over-load, so making this entirely non-virtual is not as straight-forward. Same for `WaterSSTP`, which I don't want to touch.

src/thermo/ThermoPhase.cpp
ischoegl: I added it here as it triggers a more meaningful error (and there's no point in going farther as `setDensity` will fail no matter what).

src/thermo/Phase.cpp
ischoegl: Done. Was able to work around the `LatticeSolidPhase.fromScratch` issue by using `setParameters` instead.

src/thermo/Phase.cpp
ischoegl: Yes! I guess some of the old calls I replaced were already superfluous.

**Note:** `PureFluidPhase` and `WaterSSTP` require legitimate updates. I removed `setStoichMoleFractions`, and used the `setMoleFractions_NoNorm` loophole instead.

src/thermo/Phase.cpp
ischoegl: On a related note: other phases (`StoichSubstance` etc.) side-step this issue by using `setParameters`

ischoegl:
@speth ... thanks for the review! I was able to take care of most issues, with the exceptions left unresolved (this PR is somewhat like playing whack-a-mole). Most of the remaining unresolved points arise due to `PureFluidPhase` having legitimate reasons to override defaults.

Renaming `isIncompressible` to `isCompressible` is a good suggestion that I adopted.

The last remaining point:

> I see a fair amount of complexity generated by the attempt to avoid setting the composition for single-species phases, and I'm really not seeing the benefit of enforcing this. We wouldn't issue a warning for a single-species ideal gas, so why worry about whether a different phase type with a single species? 

My point here is to reserve the strict treatment to phases that are stoichiometric by definition. I renamed the function issuing the warning to `assertNotStoich` to clarify. In a way, it is a self-documenting feature that ensures that phases are used according to their definition.

> This makes it harder to write code that works for any phase type, as you have already seen from the special cases added in the equilibrium solvers. And `Reactor` only survives without modification due to the fact that `setMassFractions_noNorm` doesn't call `assertMultiSpecies` for some reason. 

I disagree: Stoichiometric phases are quite different from multi-species phases, and not differentiating from the get-go may lead to sloppy implementations with unintended consequences. The equilibrium check is legitimate, and I don't foresee this as having to be done often. The single-species `Reactor` is imho curious, but also the reason for exempting `setMassFractions_noNorm` as a loophole.

> Likewise, there's a fair amount of simplicity to be gained by always having the length of the state be be 2 + n_species.

There will be more complexity once you account for saving of plasma phases, etc.. Hence the `virtual` definition of `saveState` (edit: although I'm not sure about whether this is necessary any longer). Apart from this, I view this as a minor issue. As mentioned above, the main point is to differentiate behavior and make those phases easier to recognize for what they are.

src/thermo/Phase.cpp
speth: Sorry, I should have said for an *incompressible* phase, not constant density. For a simple liquid model, with a constant molar volume for each species, setting the mole fractions is of course fine, and would result in a change in density. However, setting the concentrations overspecifies things: either the density can be calculated from the species molar volumes, or the one determined from the concentrations and molecular weights alone.

include/cantera/thermo/DebyeHuckel.h
speth: I don't think we should make the method non-virtual without first marking it as deprecated (it's not the right term, since the method will still be there, but we can clarify that in the comment), as that would break any user code which currently relies on that being a virtual method.

I think there's probably a path to removing `PureFluidPhase::setDensity`, but we don't have to pursue that now.

src/thermo/Phase.cpp
ischoegl: Of course, yes! It should definitely fail. Units are kmol/m3 for bulk phases ... 

Adding an `assertCompressible` check at the beginning.

include/cantera/thermo/DebyeHuckel.h
ischoegl: Regarding making a virtual method non-virtual: I don’t see a good way of notifying deprecation other than temporarily renaming the function?

ischoegl: @speth ... I am reversing making `Phase::density` non-virtual. It's not necessary and I don't see a good way of letting users know.

Regarding `PureFluidPhase::setDensity` there probably is a way to eliminate it, but I'm not sure that this is necessary at the moment.

src/thermo/Phase.cpp
ischoegl: @speth ... Implicitly, this appears to pull in #633. For any phase where `Phase::isCompressible` is `false` should trigger a recalculation of density, correct? Different from your suggestion in #633, this could now be addressed centrally?

speth: I'm not quite sure I see the relation. What is it that you think should trigger recalculation of density?

ischoegl: A call to `Phase::setMoleFractions`. I guess I should have been more specific ... 

Some of the updates are taken care of by `compositionChanged` callbacks, but apparently not all. As for `setDensity` it may be easier to update things centrally as it is pretty clear when it is necessary. I'd rather not attach this to this PR though.

src/thermo/Phase.cpp
ischoegl: Update: reverted to the dedicated `Phase::setStoichMoleFractions`, as it will be easier to maintain as using `setMoleFractions_NoNorm` as a loophole. 

~Edit: the function is renamed to `Phase::resetStoichPhase` to better describe what is being done (essentially, it makes sure that molecular weights are set correctly).~

The root cause is that `m_mmw` was not reset in a call of `setMolecularWeight` from `PureFluid`. I updated that function, which eliminates the requirement of calling extra functions during initialization.

include/cantera/thermo/Phase.h
```diff
@@ -276,25 +281,45 @@ class Phase
...
+    virtual std::vector<std::string> fullStates() const;
+
+    //! Return a vector of settable partial property sets within a phase
+    virtual std::vector<std::string> partialStates() const;
+
     //! Save the current internal state of the phase.
     //! Write to vector 'state' the current internal state.
     //!     @param state output vector. Will be resized to nSpecies() + 2.
-    void saveState(vector_fp& state) const;
+    virtual void saveState(vector_fp& state) const;
```
ischoegl: Not sure whether these should be made virtual.

include/cantera/thermo/Phase.h
```diff
@@ -276,25 +281,45 @@ class Phase
 
     //!@} end group Element and Species Information
 
+    //! Return whether phase represents a stoichiometric (fixed composition)
+    //! substance
+    virtual bool isStoichPhase() const {
```
ischoegl: Also not sure whether stoichiometric is the best description. Thermodynamically, those are presumably phases describing ‘pure’ substances.

speth:
> > Likewise, there's a fair amount of simplicity to be gained by always having the length of the state be be 2 + n_species.
> 
> There will be more complexity once you account for saving of plasma phases, etc.. Hence the `virtual` definition of `saveState` (edit: although I'm not sure about whether this is necessary any longer).

OK, I can see your point for cases where one might want to include additional state variables, and on that basis making `saveState` and `restoreState` virtual makes sense to me. I would suggest that the versions which take `(size_t, double*)` as their arguments be the virtual versions, with the versions that take `vector_fp&` left non-virtual, since they're just wrappers around the other one. An additional virtual function needs to be added in this case, though, to get the appropriate size of the state.

> I disagree: Stoichiometric phases are quite different from multi-species phases, and not differentiating from the get-go may lead to sloppy implementations with unintended consequences.

My understanding of the rationale for treating some single-species phases as a special case is because it allows some properties to be determined easily at that stage, since there's no difference between the properties of the one species and those of the mixture. Beyond that, I don't think they are particularly special, with the exception of the pure substance models. Can you expand on what "sloppy implementations with unintended consequences" you are anticipating, because I don't really see it? Is there some case where you think the current implementation leads to unexpected behavior, given the consistency of normalizing input mass and mole fraction vectors?

src/thermo/Phase.cpp
speth: Yes, I think that would make sense, and agree that this PR isn't the place to do it. I'll take further discussion to #633 so it doesn't get lost here.

ischoegl:
> Can you expand on what "sloppy implementations with unintended consequences" you are anticipating, because I don't really see it?

I think "sloppy" is probably the wrong adjective. There are quite a few instances where functions are called to update things that are not their primary focus, which makes it pretty hard to follow what's going on. From that perspective, this PR was pretty interesting, as it showed how everything is interconnected. There are numerous interactions that are quite hard to keep track of. 

One example I can think of is `PureFluid::initThermo` using the `Phase::setMoleFractions` method to update `m_mmw` instead of adding the update to the `Phase::setMolecularWeight` method that is called right before. It's certainly convenient to recycle code, but it does not help readability! Imho, having some road-blocks that differentiate usage will create cleaner code in the long run. 

Imho, setting some flags, - e.g. `isCompressible` or `isPure` (~which I'd suggest to use instead of the historically motivated `isStoichPhase`~ ... renamed from `isStoichPhase`) - will help sort things out quickly. This can also help avoid having to overload functions for *every* derived instance of a certain subclass. A correctly set central flag is all that's needed (see this PR's approach for `setDensity`).

As mentioned, reducing the length of the state vector from 3 to 2 is a really minor aspect of this effort that I'd be happy to remove (although I don't know why it should be necessary to 'save' non-information). Preventing users from setting mole fractions for pure species is just being consistent, and should raise an error at least from the python interface. Right now, there's a single occasion where I had to avoid `Phase::setState_TPX` (in the multi-phase equilibrium solver). Regarding the pure substance reactor exemption, the consistent approach would be to not integrate composition.

ischoegl:
@speth ... I made the change to keep some functions non-virtual (`saveState`/`restoreState`), and clarified the boolean flag as `isPure` (rather than `isStoichPhase`).

include/cantera/thermo/Phase.h
ischoegl: Resolved.

include/cantera/thermo/Phase.h
ischoegl: Renamed to `isPure`

src/equil/MultiPhase.cpp
```diff
@@ -343,7 +343,11 @@ void MultiPhase::setPhaseMoleFractions(const size_t n, const doublereal* const x
         init();
     }
     ThermoPhase* p = m_phase[n];
-    p->setState_TPX(m_temp, m_press, x);
+    if ( p->isPure() ) {
+        p->setState_TP(m_temp, m_press);
+    } else {
+        p->setState_TPX(m_temp, m_press, x);
+    }
```
bryanwweber: Ideally, this check wouldn't be necessary. I'd argue this is less readable than the existing version.

src/thermo/Phase.cpp
```diff
@@ -256,42 +256,122 @@ std::string Phase::speciesSPName(int k) const
...
+        }
+    }
+}
+
+vector<std::string> Phase::partialStates() const
+{
+    if (isPure()) {
+        return {};
+    } else {
+        if (!isCompressible()) {
```
bryanwweber: This is the only one of these functions where you negate this check, I think it'd be better to be consistent with the other functions.

src/thermo/Phase.cpp
```diff
@@ -256,42 +256,122 @@ std::string Phase::speciesSPName(int k) const
...
+}
+
 void Phase::saveState(vector_fp& state) const
 {
-    state.resize(nSpecies() + 2);
+    if (isPure()) {
+        state.resize(2);
+    } else {
+        state.resize(nSpecies() +2);
+    }
```
bryanwweber: Likewise, I don't think that this should be necessary. I don't think that having one extra column in the output is a problem.

src/thermo/Phase.cpp
```diff
@@ -256,42 +256,122 @@ std::string Phase::speciesSPName(int k) const
...
     state[0] = temperature();
-    state[1] = density();
-    getMassFractions(state + 2);
+    if (isCompressible()) {
+        state[1] = density();
+    } else {
+        state[1] = pressure();
+    }
+    if (!isPure()) {
+        if (defaultState()[2] == "Y") {
```
bryanwweber: I'm not a big fan of this "magic number" approach to finding the mass or mole fraction default. There is an implicit requirement that overriding functions keep the same order (i.e., mole or mass fraction last). However, I recognize that there is value in the simplicity of having just `std::vector<std::string>` as the type. Is there any way to remove this magic number and keep the simplicity (he asked without much hope)? If not, can this requirement at least be documented?

src/thermo/Phase.cpp
```diff
@@ -628,21 +717,40 @@ doublereal Phase::elementalMoleFraction(const size_t m) const
...
+    }
+}
+
+void Phase::setConstDensity(const double density_)
+{
+    if (density_ > 0.0) {
+        m_dens = density_;
+    }
+}
+
```
bryanwweber: I see the semantic purpose of distinguishing this function, but what is the practical purpose? Why remove the check of negative density? Presumably, someone could put a negative density in their input file...

src/thermo/Phase.cpp
```diff
@@ -256,42 +256,122 @@ std::string Phase::speciesSPName(int k) const
...
+            setMoleFractions_NoNorm(state + 2);
+        }
     }
+    compositionChanged();
 }
 
-void Phase::setMoleFractions(const doublereal* const x)
+void Phase::setMoleFractions(const double* const x)
 {
+    assertNotPure("setMoleFractions");
```
bryanwweber: I think the change to `PureFluid` and others to not need a call to `setMoleFractions` to update `m_mmw` is fine. But I don't see a reason why, at runtime, code can't use the `setMoleFractions` for a pure substance. It is effectively a no-op and allowing this to happen simplifies the mental pattern that people have to carry around about how states can be set in Cantera.

bryanwweber (commented):
@ischoegl @speth I'm thinking about the `isPure()` checks and single-species phases. I agree that there is some utility in warnings about misuse of various setters, and also in distinguishing the special case single phase species by some flag. I also agree that there's no reason for users to be warned when a phase is implemented incorrectly (presumably it should be implemented correctly in the first place!), and I don't think the code should twist itself in knots to verify that everything is just right at runtime. Is it possible to issue compiler warnings or errors if various methods are misused, rather than runtime errors/warnings? This way the information would be shown to the right people (the developers) not the users who will have no idea about such implementation details.

I also had a few (perhaps naive) questions below.

src/thermo/Phase.cpp
ischoegl: For incompressible substances, the `setDensity` path is blocked; `setConstDensity` is used instead to allow for assigning constant values.

ischoegl:
@bryanwweber  ... thanks for the feedback!

>  I don't think the code should twist itself in knots to verify that everything is just right at runtime. [...] But I don't see a reason why, at runtime, code can't use the `setMoleFractions` for a pure substance. It is effectively a no-op and allowing this to happen simplifies the mental pattern that people have to carry around about how states can be set in Cantera.

OK. I guess I'm outnumbered here! Introducing those checks did help me personally to sort out what phase is doing what (this PR felt a little bit like playing whack-a-mole). But I don't have a problem removing those checks/restrictions: *once* things are working (i.e. everything is implemented correctly), the code should indeed do the right thing.

bryanwweber:
@ischoegl Is there a way to make sure that any *future* code that we write will also do the "right thing", without making it a runtime issue? I thought of compiler warnings or errors since we have a fairly clean warning log right now but maybe there's a better way.

ischoegl:
> @ischoegl Is there a way to make sure that any _future_ code that we write will also do the "right thing", without making it a runtime issue? I thought of compiler warnings or errors since we have a fairly clean warning log right now but maybe there's a better way.

I cannot think of a good way of doing this at compile time (or may have simply not worked with compilers enough). 

I implemented `assertNotPure`, etc. to ensure that future code is compliant. While the checks are computationally light-weight (they're `inline`), I can see that it may feel overly intrusive. I'd still like to point out that the equilibrium solver differentiates between single-species phases elsewhere, so there isn't really a departure in terms of 'rethinking' the way things have been implemented.

include/cantera/thermo/Phase.h
```diff
@@ -780,6 +824,34 @@ class Phase
...
+            warn_user("Phase::assertNotPure",
+                      "Setter '{}' should not be used for pure (single-species) "
+                      "phase '{}' ('{}')", setter, name(), type());
+        }
+    }
+
+    //! Ensure that phase is compressible.
+    //! An error is raised if the state is incompressible
+    //!     @param setter  name of setter (used for exception handling)
+    inline void assertCompressible(const std::string& setter) {
```
speth: The `inline` keyword does nothing here - member functions defined inside the class definition are implicitly inline.

Also, this can be a `const` function.

src/thermo/Phase.cpp
speth: I agree that this function is needed, but the docstring could use some work to explain when it should be used. The statement there that "Note the density of a phase is not an independent variable" doesn't really clear things up.

I think the `else/throw` clause from `setDensity` belongs here as well.

include/cantera/thermo/Phase.h
```diff
@@ -780,6 +824,34 @@ class Phase
...
+                               "Setter '{}' is not available. Density is not an "
+                               "independent \nvariable for "
+                               "'{}' ('{}')", setter, name(), type());
+        }
+    }
+
+    //! Set the internally stored constant density (kg/m^3) of the phase.
+    //! Note the density of a phase is not an independent variable.
+    //!     @param[in] density_ density (kg/m^3).
+    virtual void setConstDensity(const double density_);
```
speth: Is there a need for this to be virtual? I'm having trouble thinking of when that would useful.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -290,6 +285,46 @@ cdef class ThermoPhase(_SolutionBase):
...
+        """
+        def __get__(self):
+            return self.thermo.isPure()
+
+    property _is_compressible:
+        """
+        The density of the phase is an independent variable
+        """
+        def __get__(self):
+            return self.thermo.isCompressible()
```
speth: These methods are unused. Are they needed?

src/thermo/Phase.cpp
speth: I'm not sure whether this is worth the extra complexity, but `defaultState()` could return a `std::map<std::string, size_t>` where the keys are the property names and the values are the offsets into the state vector. It wouldn't be a bad structure for the (future) case where one wants to include additional state variables.

src/thermo/Phase.cpp
```diff
@@ -256,42 +256,122 @@ std::string Phase::speciesSPName(int k) const
...
-void Phase::restoreState(size_t lenstate, const doublereal* state)
+void Phase::restoreState(size_t lenstate, const double* state)
 {
-    if (lenstate >= nSpecies() + 2) {
-        setMassFractions_NoNorm(state + 2);
-        setTemperature(state[0]);
+    size_t ls = 2;
+    if (!isPure()) {
+        ls += nSpecies();
+    }
```
speth: If the size of the state is going to be variable (beyond `2 + nSpecies()`) there needs to be a public method that returns this value.

include/cantera/thermo/Phase.h
```diff
@@ -276,6 +281,26 @@ class Phase
...
+        return false;
+    }
+
+    //! Return whether phase represents a compressible substance
+    virtual bool isCompressible() const {
+        return true;
+    }
+
+    //! Return a vector with default properties defining the state of a phase
+    virtual std::vector<std::string> defaultState() const;
```
speth: "default" doesn't feel like the right name for this to me. I would say "intrinsic", except that already has a meaning in thermo. Maybe "native"?

speth (changes requested):
I appreciate the improvement to `setMolecularWeight` and getting rid of the wart of using `setMoleFractions` for its side-effect of fixing up the value of `m_mmw`. However, I would argue that this issue was independent of whether setting the composition of a single-species phase was allowed -- calls to `setMolecularWeight` could have been used from multi-species phases as well, and would also have required some update afterwards.

I'm OK with the addition of `isPure` and using it to change the size of the state vector. While there's some increase in complexity and not really any efficiency gained in avoiding storing the obvious mass fraction of 1 for these phases, it does lead toward a more general definition of the state that hopefully will be useful for some new phase types in the future.

I'm still not sold on the need to warn about setting the composition of "pure" phases. I would prefer that we drop that portion of this PR so that the parts that we are in agreement on can be merged, especially since this is also blocking #719.

include/cantera/thermo/Phase.h
ischoegl: `nativeState` sounds good. Adopted.

interfaces/cython/cantera/thermo.pyx
ischoegl: `_is_pure` is used in unit tests. `_is_compressible` is indeed not used, but I probably should add a unit test using this also. EDIT: added.

src/thermo/Phase.cpp
ischoegl: I renamed the function to `assignDensity`, and improved the docstring.

src/equil/MultiPhase.cpp
ischoegl: It may be, but the MultiPhase solvers differentiate between pure and multi-species phases elsewhere. Suggest to keep as it is inconsequential (i.e. this PR assigns the 'native' state in all instances)

src/thermo/Phase.cpp
ischoegl: Not sure that I am following. This function returns state setters that don't affect molecular concentrations in a partial state assignment (which does not apply to pure substances, where those setters change the full state).

src/thermo/Phase.cpp
ischoegl: Adopted a `Phase::nStates` method to make future implementations more flexible, per @speth's suggestion. I.e. this particular line no longer has this check.

src/thermo/Phase.cpp
ischoegl: All checks are removed.

include/cantera/thermo/Phase.h
ischoegl: Removed `virtual`

src/thermo/Phase.cpp
ischoegl: Implemented `Phase::nStates` method.

src/thermo/Phase.cpp
ischoegl: That's a valid point. I followed @speth's idea to implement a map here, which eliminates the magic numbers and allows for future additional states.

ischoegl:
@speth and @bryanwweber ... thanks for the feedback. I eliminated the warnings and adopted suggestions. `setConstDensity` is now `assignDensity` (with an improved docstring). I liked the idea to use maps for `nativeState`, which is now implemented. Let me know if there's anything else.

include/cantera/thermo/Phase.h
```diff
@@ -276,6 +281,29 @@ class Phase
...
+    }
+
+    //! Return a vector with default properties defining the state of a phase
+    virtual std::map<std::string, size_t> nativeState() const;
+
+    //! Return a vector containing full states defining a phase
+    virtual std::vector<std::string> fullStates() const;
+
+    //! Return a vector of settable partial property sets within a phase
+    virtual std::vector<std::string> partialStates() const;
```
speth: Can you expand the documentation of these methods? If I hadn't already looked at the implementations, I don't think I'd have any idea what these methods returned.

include/cantera/thermo/Phase.h
```diff
@@ -276,6 +281,29 @@ class Phase
...
+
+    //! Return size of vector defining internal state of the phase.
+    virtual size_t nStates() const;
```
speth: `nStates` to me implies "number of states", which isn't quite right here. Maybe `stateSize` or `nStateVariables`? Also, it would be helpful to explicitly reference the `saveState` and `restoreState` methods as being the ones where this size matters.

include/cantera/thermo/Phase.h
```diff
@@ -780,6 +827,35 @@ class Phase
...
+    //! Ensure that phase has multiple species
+    //! An error is raised if the state is pure
+    //!     @param setter  name of setter (used for exception handling)
+    inline void assertNotPure(const std::string& setter) const {
+        if (isPure()) {
+            warn_user("Phase::assertNotPure",
+                      "Setter '{}' should not be used for pure (single-species) "
+                      "phase '{}' ('{}')", setter, name(), type());
+        }
+    }
```
speth: Can we remove `assertNotPure` since it isn't being used now?

include/cantera/thermo/Phase.h
```diff
@@ -780,6 +827,35 @@ class Phase
...
+
+    //! Ensure that phase is compressible.
+    //! An error is raised if the state is incompressible
+    //!     @param setter  name of setter (used for exception handling)
+    inline void assertCompressible(const std::string& setter) const {
```
speth: The `inline` keyword has no effect here.

interfaces/cython/cantera/composite.py
```diff
@@ -671,15 +694,16 @@ def restore_data(self, data, labels):
...
                 # save species mode and remaining full_state candidates
                 mode = prefix[0]
-                full_states = [v[:-1] for v in full_states if mode in v]
+                states = [v[:-1] for v in states if mode in v]
                 break
-        if len(valid_species) != len(all_species):
+        if mode == '':
+            # concentration specifier ('X' or 'Y') is not used
+            states = [s for s in states
+                      if all([v not in s for v in {'X', 'Y'}])]
```
speth: A slightly simpler alternative:
```python
states = [s for s in states if 'X' not in s and 'Y' not in s]
```

interfaces/cython/cantera/composite.py
```diff
@@ -530,6 +536,19 @@ def __init__(self, phase, shape=(0,), states=None, extra=None):
...
+        state_sets = state_sets | set(self._all_states)
+        for name in state_sets:
+            ph = type(phase)
+            if len(name) == 2:
+                setattr(SolutionArray, name, _state2_prop(name, ph))
+            elif len(name) == 3:
+                setattr(SolutionArray, name, _state3_prop(name, ph))
+            else:
+                raise NotImplementedError("Failed adding property '{}' for "
+                                          "phase '{}'".format(name, phase))
```
speth: Adding members to the class object in the constructor for a particular instance seems very strange to me. I would expect this to affect all current and future `SolutionArray` objects, but that doesn't seem to be quite what's happening here, and I'm confused. To illustrate:
```python
w = ct.Water()
a = ct.SolutionArray(gas, 10)
'TX' in dir(a)  # False, as expected
b = ct.SolutionArray(w)
'TX' in dir(b)  # True, as expected
'TX' in dir(a)  # True, as expected, but not desired
a.TX  # unexpectedly (but as desired) raises AttributeError
```

 I gather that the point of moving this here is that the phase may define state setters that are not included in `SolutionArray._all_states`, is that correct? That's not even something we have support for in class `Solution`, so solving it here is asking a bit much.

interfaces/cython/cantera/test/test_models.py
```diff
@@ -0,0 +1,275 @@
...
+                TD = sol.TD
+                z = sol.state # calls Phase::saveState
+                sol.TP = 300, 2*ct.one_atm
+                sol.state = z # calls Phase::restoreState
+                self.assertEqual(sol.T, T0)
+                self.assertEqual(sol.P, p0)
+
+                if not sol._is_compressible:
+                    self.assertRaisesRegex(ct.CanteraError,
+                                           'Density is not an independent')
```
speth: This assertion isn't actually doing anything. I think you want to write something like
```python
with self.assertRaisesRegex(ct.CanteraError, 'Density is not an independent'):
    sol.TD = TD
```

interfaces/cython/cantera/test/test_models.py
```diff
@@ -0,0 +1,275 @@
...
+        utilities.CanteraTest.setUpClass()
+        cls.yml_file = pjoin(cls.test_data_dir, "thermo-models.yaml")
+        with open(cls.yml_file, 'rt', encoding="utf-8") as stream:
+            cls.yml = yaml.safe_load(stream)
+
+    def test_load_thermo_models(self):
+
+        for ph in self.yml['phases']:
+
+            ph_name = ph['name']
```
speth: A minor style nitpick, but we normally don't use blank lines at the start of an indented block in Python, e.g. this would just be:
```python
    def test_load_thermo_models(self):
        for ph in self.yml['phases']:
            ph_name = ph['name']
```
I guess the one exception where this is OK is between the start of a `class` declaration and the first method.

interfaces/cython/cantera/test/test_models.py
```diff
@@ -0,0 +1,275 @@
...
+    from ruamel import yaml
+
+import cantera as ct
+from . import utilities
+
+
+class TestModels(utilities.CanteraTest):
+
+    @classmethod
+    def setUpClass(cls):
```
speth: I'm not sure I see the benefit of these tests being in a new file / subtest suite. What makes them different from the tests in `test_thermo`? At the very least, the name "models" is pretty non-specific and doesn't make it clear what kind of tests belong here as opposed to other locations.

interfaces/cython/cantera/thermo.pyx
speth: I guess where I was going with this is that either these methods are useful and should be "public" members (no underscore), or they aren't and they don't need to be accessible from Python in the first place.

src/equil/MultiPhase.cpp
speth: Given that we aren't adding the calls to `assertNotPure` that necessitated this change, I would suggest removing this and the other equivalent changes.

src/thermo/Phase.cpp
speth: I think what Bryan's suggesting is to just to reverse the order of the two branches here and write it as
```c++
        if (isCompressible()) {
            return {"TD", "TP", "UV", "DP", "HP", "SP", "SV"};
        } else {
            return {"TP", "HP", "SP"};
        }
```

src/thermo/Phase.cpp
```diff
@@ -256,42 +256,125 @@ std::string Phase::speciesSPName(int k) const
...
-    getMassFractions(state + 2);
+    auto native = nativeState();
+
+    // function assumes default definition of nativeState
+    state[native["T"]] = temperature();
+    if (isCompressible()) {
+        state[native["density"]] = density();
+    } else {
+        state[native["P"]] = pressure();
+    }
```
speth: I would use `native.at("T")` and similar, so that an exception is raised if `native` does not contain the key `T`. Otherwise, an element for this key will be default constructed, which means that the expression will evaluate to 0.

src/thermo/Phase.cpp
```diff
@@ -256,42 +256,125 @@ std::string Phase::speciesSPName(int k) const
...
+    if (!isPure()) {
+        if (native.find("Y") == native.end()) {
+            getMoleFractions(state + native["X"]);
+        } else {
+            getMassFractions(state + native["Y"]);
+        }
+    }
```
speth: To avoid accidentally reading 0 for non-existent keys:
```c++
        if (native.count("X")) {
            getMoleFractions(state + native["X"]);
        } else if (native.count("Y") {
            getMassFractions(state + native["Y"]);
        }
```
And actually, this structure negates the need to check on `isPure()`.

src/thermo/Phase.cpp
```diff
@@ -628,21 +715,43 @@ doublereal Phase::elementalMoleFraction(const size_t m) const
...
     return 1.0/molarDensity();
 }
 
+void Phase::setDensity(const double density_)
+{
+    assertCompressible("setDensity");
+    if (density_ > 0.0) {
+        m_dens = density_;
+    } else {
+        throw CanteraError("Phase::setDensity()",
```
speth:
```suggestion
        throw CanteraError("Phase::setDensity",
```

src/thermo/Phase.cpp
```diff
@@ -628,21 +715,43 @@ doublereal Phase::elementalMoleFraction(const size_t m) const
...
+            "density must be positive. density = {}", density_);
+    }
+}
+
+void Phase::assignDensity(const double density_)
+{
+    if (density_ >= 0.0) {
+        m_dens = density_;
+    } else {
+        throw CanteraError("Phase::assignDensity()",
```
speth:
```suggestion
        throw CanteraError("Phase::assignDensity",
```

src/thermo/Phase.cpp
```diff
@@ -628,21 +715,43 @@ doublereal Phase::elementalMoleFraction(const size_t m) const
...
+    } else {
+        throw CanteraError("Phase::setDensity()",
+            "density must be positive. density = {}", density_);
+    }
+}
+
+void Phase::assignDensity(const double density_)
+{
+    if (density_ >= 0.0) {
+        m_dens = density_;
```
speth: Why is density = 0 allowed for `assignDensity`, when `setDensity` requires a strictly positive value?

src/thermo/ThermoFactory.cpp
```diff
@@ -561,7 +561,16 @@ void setupPhase(ThermoPhase& thermo, AnyMap& phaseNode, const AnyMap& rootNode)
...
+        auto node = phaseNode["state"].as<AnyMap>();
+        if (thermo.isPure() &&
+            (node.hasKey("X") || node.hasKey("mole-fractions") ||
+             node.hasKey("Y") || node.hasKey("mass-fractions")) ) {
+            warn_user("setupPhase",
+                "Invalid specification for stoichiometric phase '{}' ('{}'). "
+                "State specifies concentrations, which for a single species "
+                "phase is unity by definition.", thermo.name(), thermo.type());
+        }
+        thermo.setState(node);
```
speth: I think this falls under the previous arguments regarding single-species phases, and that therefore this check should be removed.

speth (changes requested):
OK, I think we're making progress here!

I like the name `assignDensity` -- it clearly sets this apart from `setDensity`.

I did notice a few added blank lines containing whitespace. If you could clean those up it would be appreciated. They are unfortunately not  visible in the GitHub UI, but you can see them pretty clearly in the output of `git diff` if you have color formatting enabled.

src/thermo/Phase.cpp
ischoegl: ah. I guess I was overthinking here!

interfaces/cython/cantera/thermo.pyx
ischoegl: I am making them public as I view these 'service' functions as essential for testing purposes (i.e. debugging or unit tests).

include/cantera/thermo/Phase.h
ischoegl: I see the point. Going with `stateSize`

src/thermo/Phase.cpp
ischoegl: Nice!

src/thermo/Phase.cpp
ischoegl: Yup. One of the unit tests uses a Phase with zero density. Since I wasn't sure whether this was by design, I changed things here.

src/thermo/ThermoFactory.cpp
ischoegl: I agree, but would also say that explicitly specifying the composition of a single species is redundant. I'm suggesting to change the wording instead and leave the warning.
```python
            warn_user("setupPhase",
                "Redundant specification for stoichiometric phase '{}' ('{}'). "
                "State specifies concentrations, which for a pure "
                "phase is unity by definition.", thermo.name(), thermo.type());
```

interfaces/cython/cantera/thermo.pyx
ischoegl: I am making them public as I view these 'service' functions as essential for testing purposes (i.e. debugging or unit tests).

interfaces/cython/cantera/composite.py
ischoegl: This is indeed the point. Since the lists of setters are no longer available as static variables in python (they are now defined in the C++ layer), I didn't see an alternative. Suggestions?

Regarding the `PureFluid` issue, this is temporary and can be cleaned up once `TQ` and `TX` are clearly separated in #719.

interfaces/cython/cantera/test/test_models.py
ischoegl: Indeed.

interfaces/cython/cantera/test/test_models.py
ischoegl: Ok. Not my preference, but not a problem.

src/thermo/ThermoFactory.cpp
bryanwweber: @ischoegl But why do users need to be warned at all? Redundant is fine since it doesn't actually cause any problems, and allowing this (without warnings) means that more code can be in common for more phase types.

src/thermo/Phase.cpp
speth: Where is this test? How was it previously able to set the density to zero?

src/equil/MultiPhase.cpp
ischoegl: Reverting.

interfaces/cython/cantera/test/test_models.py
ischoegl: I view those tests as going beyond `test_thermo` as they apply to entire solution objects. To clarify, I renamed this to `test_composite`. Arguably, `test_purefluid` could be merged into this more generic case.

ischoegl: I felt that `test_thermo` wasn't appropriate as tests really apply to `Solution` objects as a whole. Clarified this as `test_composite` (arguably, it can absorb `test_purefluid`).

src/thermo/ThermoFactory.cpp
ischoegl: It can be set with 'X' being 0.5, which would be normalized to 1. I am suggesting to leave this in to avoid clutter (or misleading information) in user created input.

ischoegl: A user could set `X` to 0.5, which would be misleading (apart from being clutter). Hence my suggestion to leave this in to inform users when they create custom input.

interfaces/cython/cantera/test/test_models.py
ischoegl: I felt that `test_thermo` was not appropriate as tests involve entire `Solution` objects. Changing this to `test_composite` (as these tests pertain to `composite.py`).

src/thermo/ThermoFactory.cpp
bryanwweber: Sure, and I think the example @speth gave earlier was that a user could set `X` for an ideal gas phase with one species to 0.5, but that case won't raise a warning even though the result is the same as here. I really don't see the case for warning users about this. It seems like noise, and at some point, users have to be trusted to know what they're doing.

interfaces/cython/cantera/thermo.pyx
ischoegl: PS: while I made those methods public, a similar rationale applies to `_native_state`, `_full_state` and `_partial_states`. I'd still keep them 'hidden' though.

src/thermo/Phase.cpp
ischoegl: @speth It fails in `Kinetics.ElectrochemFromYaml`, where `LatticePhase` is instantiated with `m_site_density` equal to 0.
```C++
assignDensity(meanMolecularWeight() * m_site_density);
```
raises the error. What's the most appropriate way: changing the default `m_site_density` or using a 'small' number to avoid the error?

EDIT: suggesting
```C++
assignDensity(std::max(meanMolecularWeight() * m_site_density, SmallNumber));
```

src/thermo/ThermoFactory.cpp
ischoegl: Ok. Will take it out.

ischoegl: ok.

bryanwweber:
FYI, I just noticed in the documentation for `ThermoPhase` (since you've moved `pressure()` to `Phase`):

```
The distinction is that the methods
declared in ThermoPhase require knowing the particular equation of state of
the phase of interest, while those of class Phase do not, since they only
involve data values stored within the object.
```

ischoegl:
@bryanwweber 

> FYI, I just noticed in the documentation for `ThermoPhase` (since you've moved `pressure()` to `Phase):
> 
> ```
> The distinction is that the methods
> declared in ThermoPhase require knowing the particular equation of state of
> the phase of interest, while those of class Phase do not, since they only
> involve data values stored within the object.
> ```

Thanks for pointing this out. Will address in a final revision.

@speth ... I believe this is now converging. 

This time around I decided to shorten `density` to `D` in `Phase::nativeState` (perceived this as a *wart*: everything else is a single letter, see new documentation of `Phase::fullStates`). This required a small change in `SolutionArray.collect_data`, where the `D` is turned into `density`. 

PS: the signature of `Phase::nativeState` could be changed to `std::map<char, size_t>`, but I'll wait for feedback first.

ischoegl:
> FYI, I just noticed in the documentation for `ThermoPhase` (since you've moved `pressure()` to `Phase`):
> 
> ```
> The distinction is that the methods
> declared in ThermoPhase require knowing the particular equation of state of
> the phase of interest, while those of class Phase do not, since they only
> involve data values stored within the object.
> ```

@bryanwweber ... I figure the differentiation between `Phase` and `ThermoPhase` in the description made sense when `density` was the exclusive property used to define a state. This won't work for incompressible substances, where pressure has to be saved *and density has to be calculated based on an equation of state*. 

What would you suggest as a replacement for this description? Or would you suggest to move pressure back, let `Phase::saveState` throw a `NotImplementedError`, and add actual code in `ThermoPhase::saveState`? (same for `restoreState`). Curiously in that case `Phase::setDensity` does not make sense either so I'd suggest to just change the decription?

bryanwweber:
I think it probably makes most sense to change the description, since as you note, not every phase has density as the independent variable.

interfaces/cython/cantera/composite.py
speth: The concern isn't about `TX` in particular, since I realize that that will change with #719. I picked it because it's the one case I know of where we have a state tuple that's supported for a model that isn't present in the base `ThermoPhase` class.

You can avoid modifying the base class with the approach in 2550e4860cba217def5fe055425887731cde4ea4, where I've basically replace the properties with the use of `__getattr__` and `__setattr__` to provide the same interface. However, both this approach and the one you have implemented have the downfall of removing all of these properties from the documentation. This approach has the additional side effect that none of these properties show up in `dir` or the autocomplete options in IPython, which is unfortunate.

One alternative would be to continue with your approach, but add the base set of properties from `SolutionArray._all_states` when the module is loaded (i.e. the behavior before this commit) and then add any others implemented by a particular phase here. That way, the properties from `_all_states` would at least be documented.

interfaces/cython/cantera/test/test_models.py
speth: OK, `test_composite` makes more sense.

src/thermo/Phase.cpp
speth: Oh, that's interesting. I would say that zero site density in LatticePhase is as much of a problem as zero density in other phases, so we should probably initialize it to something like `SmallNumber` in the first place. The `surface-phases.yaml` file should also probably specify `site-density` for those phase entries.

Looking at this further, this stems from a loophole in `Phase::setMolarDensity` which sets `m_dens` directly but doesn't impose any requirements. It's always something...

speth:
I wouldn't sweat over the distinction between `Phase` and `ThermoPhase`, and I think the adjustments to which methods are where that has been made here is fine. Since I can't really imagine a case where you would have a `Phase` that isn't a `ThermoPhase`, we could theoretically merge the two classes. I've avoided doing this partly because the two header files are large enough as it is.

src/thermo/Phase.cpp
ischoegl: Yes, it’s always something with this PR 😁 . Suggesting the fix indicated above (setting density to a small number if site-density is zero), in absence of clear physical intuition here. Any solution should allow YAML not specifying a site density ...

interfaces/cython/cantera/composite.py
ischoegl: @speth ... I like your suggestion (I.e. the referenced commit). How would you go about the documentation for that case? I figure one way would be to add all and raise exceptions when they’re not available?

ischoegl:
@bryanwweber and @speth ... there were several instances where updates from this PR affect descriptions of Phase definitions. To facilitate review, I kept updates in a single commit (a582743).

src/thermo/Phase.cpp
speth: I'm happy to defer any further digging on this to later. Your fix for the immediate issue seems fine.

include/cantera/thermo/Phase.h
```diff
@@ -276,6 +292,51 @@ class Phase
 
     //!@} end group Element and Species Information
 
+    //! Return whether phase represents a stoichiometric (fixed composition)
+    //! substance
+    virtual bool isPure() const {
```
speth: Would "...represents a pure (single species) substance" be more clear?

include/cantera/thermo/Phase.h
```diff
@@ -49,23 +50,33 @@ namespace Cantera
...
  * for class ThermoPhase. It is not generally necessary to overloaded any of
- * class Phase's methods, with the exception of incompressible phases. In that
- * case, the density must be replaced by the pressure as the independent
- * variable and functions such as setMassFraction within class Phase must
- * actually now calculate the density (at constant T and P) instead of leaving
- * it alone as befits an independent variable. This also applies for nearly-
- * incompressible phases or phases which utilize standard states based on a
- * T and P, in which case they need to overload these functions too.
+ * class Phase's methods, which handles both compressible and incompressible
+ * phases. For incompressible phases, the density is replaced by the pressure
```
speth, line 54: Should this statement about it not generally being necessary to overload methods of class `Phase` be removed, given that the mechanism for making a phase incompressible is to do just that?

include/cantera/thermo/Phase.h
```diff
@@ -276,6 +292,51 @@ class Phase
...
+
+    //! Return whether phase represents a compressible substance
+    virtual bool isCompressible() const {
+        return true;
+    }
+
+    //! Return a map of properties defining the native state of a substance.
+    //! By default, entries include "T", "D", "Y" for a compressible substance
+    //! and "T", "P", "Y" for an incompressible substance, with offsets 0, 1 and
+    //! 2, respectively. Mass fractions "Y" are omitted for pure species.
```
speth, line 307: I would mention "offsets into the state vector used by the getState() and setState() methods".

include/cantera/thermo/Phase.h
```diff
@@ -276,6 +292,51 @@ class Phase
...
+    virtual std::vector<std::string> fullStates() const;
+
+    //! Return a vector of settable partial property sets within a phase.
+    //! Partial states encompass all valid combinations of properties that allow
+    //! for the specification of a state while ignoring species concentrations
+    //! (e.g. "TD", "TP", "SV").
+    virtual std::vector<std::string> partialStates() const;
+
+    //! Return size of vector defining internal state of the phase.
+    //! Used by saveState and restoreState.
```
speth: Writing "saveState() and restoreState()" will cause Doxygen to link to those methods.

include/cantera/thermo/Phase.h
```diff
@@ -285,7 +346,7 @@ class Phase
     //!     @param lenstate length of the state array. Must be >= nSpecies()+2
     //!     @param state    output vector. Must be of length nSpecies() + 2 or
     //!                     greater.
-    void saveState(size_t lenstate, doublereal* state) const;
+    virtual void saveState(size_t lenstate, doublereal* state) const;
```
speth, line 349: Documentation about the length needs to be updated.

interfaces/cython/cantera/composite.py
ischoegl: OK. I applied the suggested solution from 2550e48 *verbatim*. As #719 will provide a really good use case, could we defer the remainder of this discussion for the next PR?

ischoegl (commented):
@speth ... I believe that the `SolutionArray` setters are the only remaining issue. For a suggested resolution, I applied your suggested approach from 2550e48 verbatim, which I honestly believe is a very good solution.

interfaces/cython/cantera/composite.py
ischoegl: PS: since `SolutionArray` doesn't define these setters/getters itself, I am actually not sure that the documentation needs to be inherited from `Solution`, `PureFluid`, etc.. I believe that your suggestion may be the preferable approach anyways (i.e. it may fully resolve this remaining issue).

speth (changes requested):
Regarding the `SolutionArray` state properties, I pushed an update to my branch (08cabf5ee) which achieves what I was after with at least the properties that are defined on `ThermoPhase` maintaining the existing behavior, and then any other properties being added dynamically. You should be able to add that to your branch without affecting the authorship information using `git cherry-pick`.

Thanks for making the updates to the documentation. My remaining comments just pertain to those updates.

ischoegl:
> You should be able to add that to your branch without affecting the authorship information using `git cherry-pick`

I tried that before, unsuccessfully. Are you writing to my repo or upstream? (`git pull` is up to date and `git fetch upstream` didn't yield the desired result. I assume I'm overlooking something?)

speth:
> > You should be able to add that to your branch without affecting the authorship information using `git cherry-pick`
> 
> I tried that before, unsuccessfully. Are you writing to my repo or upstream? (`git pull` is up to date and `git fetch upstream` didn't yield the desired result. I assume I'm overlooking something?)

Neither, I'm pushing to my fork.

include/cantera/thermo/Phase.h
ischoegl: Yes.

ischoegl:
> Neither, I'm pushing to my fork.

Pretty obvious (in hindsight! you even mentioned as much). Thus far, I haven't had your fork set up as a remote. Thanks for resolving!

interfaces/cython/cantera/composite.py
ischoegl: Cherry-picked resolution. Thank you @speth

bryanwweber (approved):
I'm deferring final review approval to @speth at this point. I think we've converged to a good place.

ischoegl:
> I'm deferring final review approval to @speth at this point. I think we've converged to a good place.

Thanks, Bryan!

bryanwweber:
Thank you @ischoegl!

speth (approved):

ischoegl:
Thanks for the thorough review and discussion @speth!

PR #729, opened 2019-10-14
[Examples] add version info to ic_engines.py

### LINKED ISSUE #728: AttributeError: 'cantera._cantera.Valve' object has no attribute 'set_time_function'
### Cantera version
'2.3.0'
### Operating System
Windows7
### Python/MATLAB version
Python
### Steps to reproduce
1.
2.
3.

ischoegl:

Fixes #728

- Add comment to clarify minimum Cantera version

### DIFF
--- a/interfaces/cython/cantera/examples/reactors/ic_engine.py
+++ b/interfaces/cython/cantera/examples/reactors/ic_engine.py
@@ -5,6 +5,8 @@ Simulation of a (gaseous) Diesel-type internal combustion engine.
 The simulation uses n-Dodecane as fuel, which is injected close to top dead
 center. Note that this example uses numerous simplifying assumptions and
 thus serves for illustration purposes only.
+
+Requires: Cantera >= 2.5.0, scipy >= 0.19, matplotlib >= 2.0
 """
 
 import cantera as ct


bryanwweber:
I'm not sure if this change is needed if we merge Cantera/cantera-website#86. On one hand, it won't be necessary after the next release. On the other hand, it is phrased so that it won't really be out of date. @speth do you have any thoughts here?

ischoegl:
Based on the UG, some people still use 2.3. I think it may eliminate some questions if examples included information on a minimum cantera version. The phrasing is debatable; I simply pushed this to initiate discussion.

ischoegl:
@bryanwweber ... I clarified the added text as
```
This version of `ic_engine.py` relies on features introduced in Cantera v2.5
and will not work for earlier Cantera releases.
```
As #728 indicates, users will be tempted to run examples that float around the web regardless of the Cantera version they are running, i.e. the issue is not limited to what's linked on the website. 

Also, note that the fix on the website will break once v2.5 is released: the official examples will change, while many users will still continue to use v2.4 or even v2.3.

speth:
It might be worth documenting the requirements for all of the examples, including not just the Cantera version, but any other additional dependencies such as pandas or matplotlib. In that case, I would suggest keeping the statement simple, e.g.
```
Requires: Cantera >= 2.5.0, matplotlib >= 2.0
```
If this were done in a sufficiently consistent way, it could allow us to do something clever with the website so new examples compatible with the current release could be shown.

bryanwweber:
@speth I like this idea a lot! I think standardizing the format so we can do something neat with the website would be great.

@ischoegl I don't want to put on you to update all of the examples 😃 , so if you can update this one, that'd be a great place to start I think.

ischoegl:
@speth ... this is indeed a good suggestion!

@bryanwweber ... thanks, I just pushed the consensus solution.

ischoegl:
Done.

interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -5,6 +5,8 @@
 The simulation uses n-Dodecane as fuel, which is injected close to top dead
 center. Note that this example uses numerous simplifying assumptions and
 thus serves for illustration purposes only.
+
+Requires: Cantera >= 2.5.0, matplotlib >= 2.0
```
bryanwweber: Can you also document versions for numpy and scipy? Might as well capture all of the dependencies. You can probably just pick some arbitrary recent versions, any features we're using in NumPy and SciPy have been there for a long time and it's probably not worth going all the way back to find the exact version, since people won't even be able to get those old versions most likely.

bryanwweber: Probably `numpy >= 1.15, scipy >= 0.19` would work.

speth: Adding the SciPy version is a good idea. I think numpy can be left as an implicit dependency of Cantera unless the example happens to be using some feature that requires a particularly version.

ischoegl: Done.

speth (approved):

PR #731, opened 2019-10-22
Improve handling of max_time_step in ReactorNet

### LINKED ISSUE #730: Heat flux through the wall is not working for certain conditions
### Cantera version
2.4.0
### Operating System
macOS
### Python/MATLAB version
3.7.2
### Expected Behavior
The heat flux function defined for the wall should transfer heat from reactor 1 to reactor 2. Therefore increasing the temperature in reactor 2.
### Actual Behavior
The temperature does not change for any of the reactors. 
### Steps to reproduce
```python
import numpy as np
import cantera as ct
def qdot(t):
    t0 = 50e-3
    t1 = t0+10e-3
    return 5e5*(np.sign(t-t0)-np.sign(t-t1))
gas1 = ct.Solution('gri30.xml')
gas1.TPX = 1000.0, ct.one_atm, 'N2:1'
gas2 = ct.Solution('gri30.xml')
gas2.TPX = 1000.0, ct.one_atm, 'N2:1'
r1 = ct.IdealGasReactor(gas1)
r1.volume = 0.5
r2 = ct.IdealGasReactor(gas2)
r2.volume = 1.0
w = ct.Wall(r1, r2, A=10,Q=qdot)
sim = ct.ReactorNet([r1, r2])
states1 = ct.SolutionArray(r1.thermo, extra=['t'])
states2 = ct.SolutionArray(r2.thermo, extra=['t'])
for n in range(100):
    time = (n+1)*0.001
    sim.advance(time)
    states1.append(r1.thermo.state, t=1000*time)
    states2.append(r2.thermo.state, t=1000*time)
plt.figure(1)
plt.plot(states1.t, states1.T, 'b-',label='r1')
plt.plot(states2.t, states2.T, 'r-',label='r2')
plt.legend()
plt.xlabel('Time (ms)')
plt.ylabel('Temperature (K)')
plt.show()
```

fig1

### Extra step to find the Bug
Adding a mass flow controller seems to suddenly activate the heat flux.
By adding this line after the wall definition:
```python
mfc=ct.MassFlowController(r1, r2, mdot=0.1)
```
fig2

ischoegl:
Enable checks for `max_time_step` from Python frontend, and clarify default setting in docstring.

Fixes #730

* implement `ReactorNet::maxTimeStep` to allow for external queries
* implement getter/setter for property `ReactorNet.max_time_step`
* deprecate `ReactorNet.set_max_time_step`

Changing the default behavior would be simple by restricting `m_maxstep` here:
https://github.com/Cantera/cantera/blob/c149c4ed7399d43d4797929717399b2fe6b7e0b3/src/zeroD/ReactorNet.cpp#L22
but I am hesitant to do so.

### DIFF
--- a/include/cantera/zeroD/ReactorNet.h
+++ b/include/cantera/zeroD/ReactorNet.h
@@ -35,6 +35,11 @@ public:
     //! using the current mixture state as the initial condition.
     void setInitialTime(double time);
 
+    //! Get the maximum time step.
+    double maxTimeStep() {
+        return m_maxstep;
+    }
+
     //! Set the maximum time step.
     void setMaxTimeStep(double maxstep);
 
--- a/interfaces/cython/cantera/reactor.pyx
+++ b/interfaces/cython/cantera/reactor.pyx
@@ -761,11 +761,11 @@ cdef class MassFlowController(FlowDevice):
 
     .. math:: \dot m = \max(\dot m_0*g(t), 0.),
 
-    where :math:`\dot m_0` is a constant value and :math:`g(t)` is a function of 
+    where :math:`\dot m_0` is a constant value and :math:`g(t)` is a function of
     time. Both :math:`\dot m_0` and :math:`g(t)` can be set individually by
-    the property `mass_flow_coeff` and the method `set_time_function`, 
-    respectively. The method `set_mass_flow_rate` combines the former 
-    into a single function. Note that if :math:`\dot m_0*g(t) < 0`, the mass flow 
+    the property `mass_flow_coeff` and the method `set_time_function`,
+    respectively. The method `set_mass_flow_rate` combines the former
+    into a single function. Note that if :math:`\dot m_0*g(t) < 0`, the mass flow
     rate will be set to zero, since reversal of the flow direction is not allowed.
 
     Unlike a real mass flow controller, a MassFlowController object will
@@ -805,7 +805,7 @@ cdef class MassFlowController(FlowDevice):
         Set the mass flow rate [kg/s] through this controller to be either
         a constant or an arbitrary function of time. See `Func1`.
 
-        Note that depending on the argument type, this method either changes 
+        Note that depending on the argument type, this method either changes
         the property `mass_flow_coeff` or calls the `set_time_function` method.
 
         >>> mfc.set_mass_flow_rate(0.3)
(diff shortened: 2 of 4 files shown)


include/cantera/zeroD/ReactorNet.h
```diff
@@ -35,6 +35,9 @@ class ReactorNet : public FuncEval
     //! using the current mixture state as the initial condition.
     void setInitialTime(double time);
 
+    //! Get the maximum time step.
+    double maxTimeStep();
```
bryanwweber: I think it is typical in Cantera for very simple getter methods to implement the logic in the header file.

bryanwweber (approved):
Thanks @ischoegl. This looks good to me, I didn't even realize you couldn't query this value. One small change, I think.

ischoegl:
@bryanwweber ... before making the requested change (which is minor). Do you have any comments on the default behavior (see #730). I believe that limiting the time step may resolve some of the undesirable behavior (very large time step with CVODE interpolation missing some potential changes).

include/cantera/zeroD/ReactorNet.h
ischoegl: NP

bryanwweber:
No, I think we should leave the default as-is. CVODES has a built-in way to determine the appropriate time step. I think it is on the user to determine for their use case if that needs to be limited somehow. I don't think we can pick a value that is both 1) large enough that we don't limit the CVODES unnecessarily and 2) small enough to actually effect any change.

ischoegl:
> I think we should leave the default as-is. 

OK.

interfaces/cython/cantera/reactor.pyx
```diff
@@ -969,10 +969,10 @@ cdef class PressureController(FlowDevice):
         .. deprecated:: 2.5
 
              To be deprecated with version 2.5, and removed thereafter.
-             Replaced by property `pressure_coeff`.
+             Replaced by property 'pressure_coeff'.
```
speth: Shouldn't references in docstrings be left as backticks in order to enable linking by Sphinx (see https://cantera.org/documentation/dev/sphinx/html/cython/zerodim.html#cantera.PressureController.set_pressure_coeff)

ischoegl: Correct. I meant to replace backticks in warning/error messages (a convention that @bryanwweber pointed out at some point), not here ...

bryanwweber: The only reason for the convention is that backticks have semantic meaning to Sphinx ("link to something that looks like this string") but on the terminal, the backticks don't have any meaning. Unless you have a terminal font with non-monospace by default and a Markdown parser... I wouldn't put it past someone, I'm sure it exists...

ischoegl: That's what I recalled (and why I took the step to replace backticks). As an aside, one of my student is running an installation where non-monospaced error messages do occur.

speth (approved):

PR #733, opened 2019-10-24
Deprecate magic numbers for RxnRates.h and Falloff.h

ischoegl:
* Deprecate unused magic number `int type()` methods in `RxnRates.h`. 
* Deprecate magic numbers `int getType()` and introduce `string type()` in `Falloff.h`

I am leaving the remaining magic numbers for kinetics objects as they are, as it may make sense to convert handling of various types to factories instead of having things hard-coded (this is more elaborate and should be handled by a separate PR).

### DIFF
--- a/interfaces/cython/cantera/reaction.pyx
+++ b/interfaces/cython/cantera/reaction.pyx
@@ -10,10 +10,6 @@ cdef extern from "cantera/kinetics/reaction_defs.h" namespace "Cantera":
     cdef int CHEMACT_RXN
     cdef int INTERFACE_RXN
 
-    cdef int SIMPLE_FALLOFF
-    cdef int TROE_FALLOFF
-    cdef int SRI_FALLOFF
-
 
 cdef class Reaction:
     """
@@ -422,7 +418,7 @@ cdef class Falloff:
     :param init:
         Used internally when wrapping :ct:`Falloff` objects returned from C++.
     """
-    falloff_type = SIMPLE_FALLOFF
+    falloff_type = "Lindemann"
 
     def __cinit__(self, params=(), init=True):
         if not init:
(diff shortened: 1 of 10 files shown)


src/kinetics/FalloffFactory.cpp
```diff
@@ -34,10 +37,23 @@ Falloff* FalloffFactory::newFalloff(int type, const vector_fp& c)
...
+    return f;
+}
+
 shared_ptr<Falloff> newFalloff(int type, const vector_fp& c)
 {
     shared_ptr<Falloff> f(FalloffFactory::factory()->newFalloff(type, c));
     return f;
 }
 
+shared_ptr<Falloff> newFalloff(std::string type, const vector_fp& c)
```
speth: It should be possible to avoid an unnecessary copy with:
```suggestion
shared_ptr<Falloff> newFalloff(const std::string& type, const vector_fp& c)
```

interfaces/cython/cantera/reaction.pyx
```diff
@@ -422,7 +418,7 @@ cdef class Falloff:
     :param init:
         Used internally when wrapping :ct:`Falloff` objects returned from C++.
     """
-    falloff_type = SIMPLE_FALLOFF
+    falloff_type = "Falloff"
```
speth: I think this name needs to correspond to the name used by `newFalloff`.

src/kinetics/Reaction.cpp
```diff
@@ -359,30 +359,28 @@ void readFalloff(FalloffReaction& R, const XML_Node& rc_node)
...
     }
 
-    int falloff_type = 0;
     if (caseInsensitiveEquals(falloff["type"], "lindemann")) {
-        falloff_type = SIMPLE_FALLOFF;
         if (np != 0) {
             throw CanteraError("readFalloff", "Lindemann parameterization "
                 "takes no parameters, but {} were given", np);
         }
+        R.falloff = newFalloff("Simple", falloff_parameters);
```
speth: Does it make sense to use `Lindemann` as the name `Simple`, which follows more naturally from a reading of the [science docs](https://cantera.org/science/reactions.html#falloff-reactions)?

speth (changes requested):
I'm always glad to see magic numbers go away, and this looks pretty good to me. I had just a couple of minor comments / questions.

interfaces/cython/cantera/reaction.pyx
ischoegl: Good catch - I was thinking about the class name.

src/kinetics/Reaction.cpp
ischoegl: Changing the default name to `Lindemann` and set up a synonym for `Simple`.

ischoegl:
I switched the type string to `"Lindemann"` to make things clear for the *simple* falloff type, and also added some rudimentary tests for the falloff reaction type to the suite.

speth (approved):

PR #735, opened 2019-10-27
Replace convenience classes by C++ Solution object

ischoegl:
This PR introduces the C++ `Solution` object (previously only instantiated from the Python front-end) to the C++ test suite and examples.

- implement initialization of C++ `Solution` objects from input files
- replace convenience wrapper classes (i.e. `IdealGasMix`, `Interface`) for test_problems and samples
- deprecate convenience wrappers (instead of porting them to YAML/`AnyMap`)

### DIFF
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -67,33 +67,18 @@ public:
     //! Set the Transport object
     void setTransport(shared_ptr<Transport> transport);
 
-    //! Accessor for the ThermoPhase object
-    ThermoPhase& thermo() {
-        return *m_thermo;
-    }
-
-    //! Accessor for the Kinetics object
-    Kinetics& kinetics() {
-        return *m_kinetics;
-    }
-
-    //! Accessor for the Transport object
-    Transport& transport() {
-        return *m_transport;
-    }
-
     //! Accessor for the ThermoPhase pointer
-    shared_ptr<ThermoPhase> thermoPtr() {
+    shared_ptr<ThermoPhase> thermo() {
         return m_thermo;
     }
 
     //! Accessor for the Kinetics pointer
-    shared_ptr<Kinetics> kineticsPtr() {
+    shared_ptr<Kinetics> kinetics() {
         return m_kinetics;
     }
 
     //! Accessor for the Transport pointer
-    shared_ptr<Transport> transportPtr() {
+    shared_ptr<Transport> transport() {
         return m_transport;
     }
 
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -20,8 +20,6 @@ class Solution : public std::enable_shared_from_this<Solution>
 {
 private:
     Solution();
-    Solution(const std::string& infile, std::string name, std::string transport,
-             std::vector<shared_ptr<Solution> > adjacent);
 
 public:
     ~Solution() {}
@@ -33,25 +31,6 @@ public:
         return shared_ptr<Solution>( new Solution );
     }
 
-    //! Create and initialize a Solution object from an input file
-    /*!
-     * This constructor wraps newPhase(), newKinetics() and
-     * newTransportMgr() routines for initialization.
-     *
-     * @param infile name of the input file
-     * @param name   name of the phase in the file.
-     *               If this is blank, the first phase in the file is used.
-     * @param transport name of the transport model.
-     * @param adjacent vector containing adjacent solution objects.
-     * @returns an initialized Solution object.
-     */
-    static shared_ptr<Solution> create(const std::string& infile,
-                                       std::string name="",
-                                       std::string transport="",
-                                       std::vector<shared_ptr<Solution> > adjacent={}) {
-        return shared_ptr<Solution>( new Solution(infile, name, transport, adjacent) );
-    }
-
     //! Return the name of this Solution object
     std::string name() const;
 
@@ -59,7 +38,7 @@ public:
     void setName(const std::string& name);
 
     //! Set the ThermoPhase object
-    void setThermoPhase(shared_ptr<ThermoPhase> thermo);
+    void setThermo(shared_ptr<ThermoPhase> thermo);
 
     //! Set the Kinetics object
     void setKinetics(shared_ptr<Kinetics> kinetics);
@@ -88,14 +67,22 @@ protected:
     shared_ptr<Transport> m_transport;  //!< Transport manager
 };
 
-/**
- *  Create a new solution manager.
+//! Create and initialize a new Solution manager from an input file
+/*!
+ * This constructor wraps newPhase(), newKinetics() and
+ * newTransportMgr() routines for initialization.
+ *
+ * @param infile name of the input file
+ * @param name   name of the phase in the file.
+ *               If this is blank, the first phase in the file is used.
+ * @param transport name of the transport model.
+ * @param adjacent vector containing adjacent solution objects.
+ * @returns an initialized Solution object.
  */
-inline shared_ptr<Solution> newSolution(const std::string& infile, std::string name="",
-                                        std::string transport="",
-                                        std::vector<shared_ptr<Solution> > adjacent={}) {
-    return Solution::create(infile, name, transport, adjacent);
-}
+shared_ptr<Solution> newSolution(const std::string& infile,
+                                 const std::string& name="",
+                                 const std::string& transport="",
+                                 const std::vector<shared_ptr<Solution>>& adjacent={});
 
 }
 #endif
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -20,14 +20,36 @@ class Solution : public std::enable_shared_from_this<Solution>
 {
 private:
     Solution();
+    Solution(const std::string& infile, std::string name, std::string transport,
+             std::vector<shared_ptr<Solution> > adjacent);
 
 public:
     ~Solution() {}
     Solution(const Solution&) = delete;
     Solution& operator=(const Solution&) = delete;
 
+    //! Create an empty Solution object
     static shared_ptr<Solution> create() {
         return shared_ptr<Solution>( new Solution );
+    }
+
+    //! Create and initialize a Solution object from an input file
+    /*!
+     * This constructor wraps newPhase(), newKinetics() and
+     * newTransportMgr() routines for initialization.
+     *
+     * @param infile name of the input file
+     * @param name   name of the phase in the file.
+     *               If this is blank, the first phase in the file is used.
+     * @param transport name of the transport model.
+     * @param adjacent vector containing adjacent solution objects.
+     * @returns an initialized Solution object.
+     */
+    static shared_ptr<Solution> create(const std::string& infile,
+                                       std::string name="",
+                                       std::string transport="",
+                                       std::vector<shared_ptr<Solution> > adjacent={}) {
+        return shared_ptr<Solution>( new Solution(infile, name, transport, adjacent) );
     }
 
     //! Return the name of this Solution object
@@ -60,11 +82,35 @@ public:
         return *m_transport;
     }
 
+    //! Accessor for the ThermoPhase pointer
+    shared_ptr<ThermoPhase> thermoPtr() {
+        return m_thermo;
+    }
+
+    //! Accessor for the Kinetics pointer
+    shared_ptr<Kinetics> kineticsPtr() {
+        return m_kinetics;
+    }
+
+    //! Accessor for the Transport pointer
+    shared_ptr<Transport> transportPtr() {
+        return m_transport;
+    }
+
 protected:
     shared_ptr<ThermoPhase> m_thermo;  //!< ThermoPhase manager
     shared_ptr<Kinetics> m_kinetics;  //!< Kinetics manager
     shared_ptr<Transport> m_transport;  //!< Transport manager
 };
+
+/**
+ *  Create a new solution manager.
+ */
+inline shared_ptr<Solution> newSolution(const std::string& infile, std::string name="",
+                                        std::string transport="",
+                                        std::vector<shared_ptr<Solution> > adjacent={}) {
+    return Solution::create(infile, name, transport, adjacent);
+}
 
 }
 #endif
--- a/include/cantera/zeroD/Reactor.h
+++ b/include/cantera/zeroD/Reactor.h
@@ -65,8 +65,8 @@ public:
     }
 
     void insert(shared_ptr<Solution> sol) {
-        setThermoMgr(sol->thermo());
-        setKineticsMgr(sol->kinetics());
+        setThermoMgr(*(sol->thermo()));
+        setKineticsMgr(*(sol->kinetics()));
     }
 
     virtual void setKineticsMgr(Kinetics& kin);
--- a/include/cantera/zeroD/Reactor.h
+++ b/include/cantera/zeroD/Reactor.h
@@ -65,8 +65,8 @@ public:
     }
 
     void insert(shared_ptr<Solution> sol) {
-        setThermoMgr(*(sol->thermo()));
-        setKineticsMgr(*(sol->kinetics()));
+        setThermoMgr(*sol->thermo());
+        setKineticsMgr(*sol->kinetics());
     }
 
     virtual void setKineticsMgr(Kinetics& kin);
--- a/include/cantera/zeroD/Reactor.h
+++ b/include/cantera/zeroD/Reactor.h
@@ -64,6 +64,11 @@ public:
         setKineticsMgr(contents);
     }
 
+    void insert(shared_ptr<Solution> sol) {
+        setThermoMgr(sol->thermo());
+        setKineticsMgr(sol->kinetics());
+    }
+
     virtual void setKineticsMgr(Kinetics& kin);
 
     virtual void setChemistry(bool cflag = true) {
--- a/samples/cxx/demo.cpp
+++ b/samples/cxx/demo.cpp
@@ -66,7 +66,7 @@ void demoprog()
 
     // Reaction information
 
-    auto kin = getGasKinetics(sol);
+    auto kin = sol->kinetics();
     int irxns = kin->nReactions();
     vector_fp qf(irxns);
     vector_fp qr(irxns);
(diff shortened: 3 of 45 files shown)


interfaces/cython/cantera/examples/onedim/diffusion_flame.py
```diff
@@ -22,7 +22,7 @@
 
 # Create the gas object used to evaluate all thermodynamic, kinetic, and
 # transport properties.
-gas = ct.Solution('gri30.xml', 'gri30_mix')
+gas = ct.Solution('gri30.yaml', 'gri30')
```
bryanwweber: If I have my druthers (a big if 😄 ), the three `gri30` phases will remain in this input file, so this phase will not have a transport model enabled, so the `transport_model` kwarg should be passed here, right?

ischoegl: Correct. But I implemented this assuming that the YAML behavior was decided.

The default transport model for YAML is `Mix` (which is different from the XML version, which didn't have a default tranport model for the `gri30` phase).

ischoegl: Cherry-picked the commit replacing `gri30_mix`, which is now part of PR #737 (which is a better venue for resolution of #736). I.e. the specific change discussed here is now moved to a different PR.

include/cantera/base/Solution.h
```diff
@@ -20,16 +20,38 @@ class Solution : public std::enable_shared_from_this<Solution>
...
 
+    //! Create and initialize a Solution object from an input file
+    /*!
+     * This constructor wraps newPhase(), newKinetics() and
+     * newTransportMgr() routines for initialization.
+     *
+     * @param infile name of the input file
+     * @param name   name of the phase in the file.
+     *               If this is blank, the first phase in the file is used.
+     * @param trasnport name of the transport model.
```
speth:
```suggestion
     * @param transport name of the transport model.
```

speth (commented):
I'm very much in favor of getting rid of these supposed "convenience" classes, and I think this is a good start to extending the use of the `Solution` class.

I don't understand the need for the methods `getIdealGasPhasePtr` and `getGasKineticsPtr` - I don't think there are any important methods of these classes that aren't available from a pointer to the base class. For surface phases / kinetics, there are a few unique methods like `getCoverages`, but I'd be inclined to handle those by having a class analogous to `Solution` which always returns instances of `SurfPhase` and `InterfaceKinetics`. Naming this class `Interface` might be slightly complicated by the existing, now-deprecated `Interface` class, though.

After seeing the `Solution` class in action a bit more, I'm wondering if it would make more sense for the `thermo()` method and friends to return `shared_ptr` and avoid the need to also having a `thermoPtr()` (etc.) method. If you then wanted a reference, you could always just write `auto& gas = *sol.thermo();`.

ischoegl:
> I don't understand the need for the methods `getIdealGasPhasePtr` and `getGasKineticsPtr`

This is true, but for some of the older examples, having to type
```C++
std::dynamic_pointer_cast<IdealGasPhase>(sol->thermoPtr())
```
and
```C++
std::dynamic_pointer_cast<GasKinetics>(sol->kineticsPtr())
``` 
every time is pretty onerous. As I didn't want to introduce some of the nastier looking C++ 11 features in the simple example, I decided to create the 'convenience' accessors.

The main issue is that a more straight-forward type cast of a pointer to the base class is currently *blocked* here https://github.com/Cantera/cantera/blob/ce47733c42d0505e20adf3b2a4cff2b27eb10a24/include/cantera/thermo/Phase.h#L102-L103 

> After seeing the `Solution` class in action a bit more, I'm wondering if it would make more sense for the `thermo()` method and friends to return `shared_ptr` and avoid the need to also having a `thermoPtr()` (etc.) method.

Agreed. From what I've seen, there isn't a way around having `shared_ptr`'s. I also agree that both aren't necessary, although sometimes one or the other is more convenient. I named things such that it is immediately clear what is an object and what is a (shared) pointer. 

If you want me to switch to `shared_ptr` only, now is the time ...

ischoegl:
@speth ... regarding the types, the few instances where the base object is used are easily replaced. E.g.
```C++
double air_mw = air->thermo().meanMolecularWeight();
```
would become (with `thermo()` now returning `shared_ptr`):
```C++
double air_mw = air->(*thermo()).meanMolecularWeight();
```
Again, it may be a question of requiring more familiarity with C++ code.

ischoegl:
@speth ... alright. I consolidated everything to `shared_ptr`. Since the C++ `Solution` class is brand-new, it's probably better to do this now.

speth:
> This is true, but for some of the older examples, having to type
> 
> ```c++
> std::dynamic_pointer_cast<IdealGasPhase>(sol->thermoPtr())
> ```
> [...] every time is pretty onerous.

Under what circumstances is a `shared_ptr<IdealGasPhase>` needed rather than a `shared_ptr<ThermoPhase>`? The only case I see is as the argument to the `StFlow` constructor, and in that case I think the preferred fix might be to change it to accept a `ThermoPhase*` and then cast that to `IdealGasPhase*` internally (and throw if that fails).

> The main issue is that a more straight-forward type cast of a pointer to the base class is currently _blocked_ here
> 
> https://github.com/Cantera/cantera/blob/ce47733c42d0505e20adf3b2a4cff2b27eb10a24/include/cantera/thermo/Phase.h#L102-L103

Can you give an example of what you think is being blocked? The quoted code should not affect any pointer casting, but just disable copying and assignment.

ischoegl:
@speth ... yes, there really isn't any special feature in `IdealGasPhase` where dynamic polymorphism doesn't already do the trick. I.e. I removed all of the accessors (i.e. `getIdealGasPhase` and similar). In some instances, a `dynamic_pointer_cast` is still necessary (e.g. for the `freeflame` sample, and the `surfkin` test problem; I think it's ok to let users figure those out)

> The only case I see is as the argument to the `StFlow` constructor, and in that case I think the preferred fix might be to change it to accept a `ThermoPhase*` and then cast that to `IdealGasPhase*` internally (and throw if that fails).

~I haven't dont that yet, but it's straight-forward.~ Edit: Done. The one case where a fix is less intuitive is the `surfkin` test problem, where coverages etc. aren't defined in the base class. Not sure whether it's worth creating a virtual method for `setCoverages` in the base class.

Regarding the blocked type cast, I misspoke. It appears to be difficult to change the type once a variable is passed by reference, but casting of pointers works regardless (I guess my original point is moot). Since much of the passing happens by reference, I was overly cautious.

samples/cxx/demo.cpp
```diff
@@ -58,28 +60,29 @@ void demoprog()
...
-        gas.temperature(), gas.pressure(), gas.density(),
-        gas.enthalpy_mole(), gas.entropy_mole(), gas.cp_mole());
+        gas->temperature(), gas->pressure(), gas->density(),
+        gas->enthalpy_mole(), gas->entropy_mole(), gas->cp_mole());
 
 
     // Reaction information
 
-    int irxns = gas.nReactions();
+    auto kin = getGasKinetics(sol);
```
speth:
```suggestion
    auto kin = sol->kinetics();
```

include/cantera/base/Solution.h
```diff
@@ -20,16 +20,38 @@ class Solution : public std::enable_shared_from_this<Solution>
...
+     * @param transport name of the transport model.
+     * @param adjacent vector containing adjacent solution objects.
+     * @returns an initialized Solution object.
+     */
+    static shared_ptr<Solution> create(const std::string& infile,
+                                       std::string name="",
+                                       std::string transport="",
+                                       std::vector<shared_ptr<Solution> > adjacent={}) {
+        return shared_ptr<Solution>( new Solution(infile, name, transport, adjacent) );
+    }
```
speth: If the `newSolution` function is the preferred way to create new `Solution` objects, we could avoid the need to introduce both the constructor that takes a filename and this `create` function by implementing the logic of that constructor in `newSolution`.

include/cantera/base/Solution.h
```diff
@@ -66,5 +88,14 @@ class Solution : public std::enable_shared_from_this<Solution>
     shared_ptr<Transport> m_transport;  //!< Transport manager
 };
 
+/**
+ *  Create a new solution manager.
+ */
+inline shared_ptr<Solution> newSolution(const std::string& infile, std::string name="",
+                                        std::string transport="",
+                                        std::vector<shared_ptr<Solution> > adjacent={}) {
```
speth: These arguments can be `const std::string&`, and the vector can be a `const std::vector<shared_ptr<Solution>>&`.  Also, thanks to C++11, the space in `std::vector<shared_ptr<Solution> >` is unnecessary.

include/cantera/zeroD/Reactor.h
```diff
@@ -64,6 +64,11 @@ class Reactor : public ReactorBase
         setKineticsMgr(contents);
     }
 
+    void insert(shared_ptr<Solution> sol) {
+        setThermoMgr(*(sol->thermo()));
+        setKineticsMgr(*(sol->kinetics()));
```
speth: The extra set of parentheses is not needed here, or in other places with the same semantics, since the member access has precedence over the dereference.

samples/cxx/openmp_ignition/openmp_ignition.cpp
```diff
@@ -19,17 +19,18 @@ void run()
...
+    std::vector<std::shared_ptr<Solution>> sols;
     std::vector<std::unique_ptr<IdealGasConstPressureReactor>> reactors;
     std::vector<std::unique_ptr<ReactorNet>> nets;
 
     // Create and link the Cantera objects for each thread. This step should be
     // done in serial
     for (int i = 0; i < nThreads; i++) {
-        gases.emplace_back(new IdealGasMix("gri30.xml", "gri30"));
+        auto sol = newSolution("gri30.yaml", "gri30", "None");
+        sols.emplace_back(sol);
```
speth, line 30: Why not skip the local variable and just write `sols.emplace_back(newSolution("gri30.yaml", "gri30", "None"));`?

samples/f77/demo_ftnlib.cpp
```diff
@@ -81,38 +93,34 @@ extern "C" {
...
-            delete _trans;
-            _trans = newTransportMgr(trmodel,_gas,0);
-        } catch (CanteraError& err) {
-            _trans = newTransportMgr("",_gas,0);
-        }
     }
 
     /// integer function nElements()
-    integer nelements_()
+    int nelements_()
```
speth: Unlike `doublereal`, there is perhaps an argument for using a `typedef` for the integer type when working with Fortran, which is that someone could be using Fortran code that uses 8 byte integers and want to be able to define `integer` appropriately.

samples/f77/demo_ftnlib.cpp
```diff
@@ -81,38 +93,34 @@ extern "C" {
             string fin = string(file, lenfile);
             string fth = string(id, lenid);
             trmodel = string(transport, lentr);
-            delete _gas;
-            _gas = new IdealGasMix(fin, fth);
+            _sol = newSolution(fin, fth, trmodel);
+            _gas = std::dynamic_pointer_cast<IdealGasPhase>(_sol->thermo());
+            _kin = std::dynamic_pointer_cast<GasKinetics>(_sol->kinetics());
+            _trans = _sol->transport();
```
speth: The `dynamic_pointer_cast`s are unnecessary if you change the type of `_gas` to `shared_ptr<ThermoPhase>`, and similar with `_kin`.

src/base/Solution.cpp
```diff
@@ -8,14 +8,39 @@
...
 namespace Cantera
 {
 
 Solution::Solution() {}
 
+  Solution::Solution(const std::string& infile, std::string name, std::string transport,
+                     std::vector<shared_ptr<Solution> > adjacent) {
+
+    // thermo phase
+    m_thermo = shared_ptr<ThermoPhase>(newPhase(infile, name));
```
speth: The shorter way to write this is
```suggestion
    m_thermo.reset(newPhase(infile, name));
```

src/base/Solution.cpp
```diff
@@ -8,14 +8,39 @@
...
+
+    // kinetics
+    std::vector<ThermoPhase*> phases;
+    for (auto & adj : adjacent) {
```
speth: Given our usual style of keeping the reference with the type name:
```suggestion
    for (auto& adj : adjacent) {
```

src/base/Solution.cpp
```diff
@@ -8,14 +8,39 @@
...
+        phases.push_back(adj->thermo().get());
+    }
+    phases.push_back(m_thermo.get());
```
speth: Shouldn't `m_thermo` be first in the `phases` list, per the documentation of `newKinetics`?

src/base/Solution.cpp
```diff
@@ -8,14 +8,39 @@
...
+    m_kinetics = std::move(newKinetics(phases, infile, name));
+
+    // transport
+    if (transport=="") {
```
speth:
```suggestion
    if (transport == "") {
```

src/base/Solution.cpp
```diff
@@ -8,14 +8,39 @@
...
+        m_transport = shared_ptr<Transport>(newDefaultTransportMgr(m_thermo.get()));
```
speth:
```suggestion
        m_transport.reset(newDefaultTransportMgr(m_thermo.get()));
```

src/oneD/StFlow.cpp
```diff
@@ -30,9 +30,14 @@ StFlow::StFlow(IdealGasPhase* ph, size_t nsp, size_t points) :
     m_zfixed(Undef),
     m_tfixed(Undef)
 {
+    if (ph->type() == "IdealGas") {
+        m_thermo = (IdealGasPhase*)ph;
```
speth: We try to avoid the unchecked C-style casts:
```suggestion
        m_thermo = static_cast<IdealGasPhase*>(ph);
```

test_problems/mixGasTransport/mixGasTransport.cpp
```diff
@@ -129,17 +133,16 @@ int main(int argc, char** argv)
         grad_T[0] = (T2 - T1) / dist;
         grad_T[1] = (T3 - T1) / dist;
 
-        int log_level = 0;
-        Transport* tran = newTransportMgr("Mix", &g, log_level=0);
-        MixTransport* tranMix = dynamic_cast<MixTransport*>(tran);
-        g.setState_TPX(1500.0, pres, Xset.data());
+        auto tran = sol->transport();
+        auto tranMix = dynamic_pointer_cast<MixTransport>(tran);
```
speth: I don't think this is necessary -- what was here before was confused, to say the least -- there are no public methods of `MixTransport` that aren't just virtual methods available from a `Transport` pointer.

test_problems/multiGasTransport/multiGasTransport.cpp
```diff
@@ -138,61 +141,60 @@ int main(int argc, char** argv)
         grad_T[0] = (T2 - T1) / dist;
         grad_T[1] = (T3 - T1) / dist;
 
-        int log_level = 0;
-        Transport* tran = newTransportMgr("Multi", &g, log_level=0);
-        MultiTransport* tranMix = dynamic_cast<MultiTransport*>(tran);
-        g.setState_TPX(1500.0, pres, Xset.data());
+        auto tran = sol->transport();
+        auto tranMix = dynamic_pointer_cast<MultiTransport>(tran);
```
speth: Likewise here.

speth (changes requested):
Thanks, this is looking pretty good. I mostly have just a few suggestions for simplifications and a couple of minor formatting nitpicks.

src/base/Solution.cpp
ischoegl: Well, that means that the old class class wasn't consistent with the docs, as I kept things implemented like before. This will likely change some output_blessed files but not much else.

samples/cxx/openmp_ignition/openmp_ignition.cpp
ischoegl: it's used again a couple of lines below for `reactors.back()->insert(sol);` ... going with readability here.

include/cantera/base/Solution.h
ischoegl: Not sure. I mostly introduced the `newSolution` for readability and consistency with other code (otherwise, it would have been `Solution::create`). The function is inline, so the compiler knows what to do, i.e. I prefer to use to use a dedicated constructor to do the actual processing.

src/base/Solution.cpp
ischoegl: Tests passed even after changing the order to the one consistent with `newKinetics`.

This is what I had based things on (which is superseded by `newKinetics`): https://github.com/Cantera/cantera/blob/07e0e2fb2ed67c47a16ba11c7c435a87ea4094eb/src/kinetics/importKinetics.cpp#L155-L160

samples/f77/demo_ftnlib.cpp
ischoegl: reverting.

include/cantera/base/Solution.h
speth: I agree that the `newSolution` function is easier to use than `Solution::create` (people seem to struggle with how to call static methods in C++) and is consistent with the other methods like `newPhase`.

I'm not concerned about the performance of any particular implementation, but about simplicity. I don't the advantage to doing the initialization in the constructor, rather than the `newSolution` method. If we consolidate these methods, then there's never any question about whether to use `Solution::create` or `newSolution`, or why there's this constructor that can't be called.

speth (changes requested):

ischoegl:
> If we consolidate these methods, then there's never any question about whether to use `Solution::create` or `newSolution`, or why there's this constructor that can't be called.

That's a fair point. Will update once our PROCI manuscript is submitted (or I need to take a break) ...

ischoegl:
Done, i.e. `newSolution(infile, name, transport, adjacent)` is now the default way to instantiate.

include/cantera/base/Solution.h
ischoegl: Done.

speth (approved):

PR #737, opened 2019-11-01
Transitional handling of 'gri30_mix' and 'gri30_multi' phases in YAML

### LINKED ISSUE #736: Deprecate all instances of `gri30_mix` and `gri30_multi`
### Cantera version
2.5.0a3

### Operating System / Python/MATLAB version
all

### Expected Behavior
The `gri30_mix` and `gri30_multi` phases are deprecated and are not part of all `gri30.yaml` versions (the version in `test/work/python` defines those, the default version does not). For consistency, all instances should be removed for 2.5, as the transport model can be specified using a flag. I.e. the differentiation of transport models in phases is largely obsolete:
```python
In [1]: import cantera as ct

In [2]: gas = ct.Solution('gri30.yaml', 'gri30_multi')
CanteraError: 
InputFileError thrown by AnyValue::getMapWhere:
Error on line 16 of /usr/local/lib/python3.6/dist-packages/cantera/data/gri30.yaml:
List does not contain a map where 'name' = 'gri30_multi'
[...]

In [3]: gas = ct.Solution('gri30.xml', 'gri30_multi')

In [4]: gas.composite
Out[4]: ('IdealGas', 'Gas', 'Multi')

In [5]: gas = ct.Solution('gri30.yaml', transport_model='Multi')

In [6]: gas.composite
Out[6]: ('IdealGas', 'Gas', 'Multi')
```
### Questions

There are two questions:
- it is not straight forward to completely remove the `gri_mix` and `gri_multi` phases, as they may be widely used in user-developed code. Should there be a customized error message (or warning)?
- the default behavior should be clarified: will the transport model default to `mixture-averaged` or should no transport manager be instantiated if the `transport_model` flag is not used? (i.e. under many circumstances, the transport manager is not used)

### Actual Behavior
`gri30.yaml` are not consistent, i.e. some have hand-coded entries. 

Deprecating `gri30_mix` is straight-forward (it is replacable by `gri30`):
```
$ find . -type f | xargs grep 'gri30_mix'
./samples/cxx/flamespeed/flamespeed.cpp:        IdealGasMix gas("gri30.cti","gri30_mix");
./test_problems/clib_test/output_output.txt:  gri30_mix:
./test_problems/clib_test/output_blessed.txt:  gri30_mix:
./test_problems/clib_test/clib_test.c:    int thermo = thermo_newFromFile("gri30.xml", "gri30_mix");
./test_problems/clib_test/clib_test.c:    int kin = kin_newFromFile("gri30.xml", "gri30_mix", thermo, 0, 0, 0, 0);
./test_problems/mixGasTransport/mixGasTransport.cpp:        IdealGasMix g("gri30.xml", "gri30_mix");
./test_problems/multiGasTransport/multiGasTransport.cpp:        IdealGasMix g("gri30.xml", "gri30_mix");
./test/matlab/TestImport.m:            gas = Solution('gri30.xml', 'gri30_mix');
./test/work/python/gri30.yaml:- name: gri30_mix
./include/cantera/kinetics/importKinetics.h: *        ok =  buildSolutionFromXML(root, "gri30_mix", "phase", th, kin)
./interfaces/matlab/toolbox/GRI30.m:        s = Solution('gri30.xml', 'gri30_mix');
./interfaces/cython/cantera/examples/onedim/flamespeed_sensitivity.py:gas = ct.Solution('gri30.xml', 'gri30_mix')
./interfaces/cython/cantera/examples/onedim/flame_fixed_T.py:gas = ct.Solution('gri30.xml', 'gri30_mix')
./interfaces/cython/cantera/examples/onedim/diffusion_flame.py:gas = ct.Solution('gri30.xml', 'gri30_mix')
./interfaces/cython/cantera/data/gri30_highT.yaml:- name: gri30_mix
./interfaces/cython/cantera/data/gri30_highT.cti:ideal_gas(name = "gri30_mix",
./interfaces/cython/cantera/data/gri30.cti:ideal_gas(name = "gri30_mix",
./data/inputs/gri30_highT.cti:ideal_gas(name = "gri30_mix",
./data/inputs/gri30.cti:ideal_gas(name = "gri30_mix",
[...]
```
There aren't many instances of `gri30_multi`:
```
$ find . -type f | xargs grep 'gri30_multi'
./test/work/python/gri30.yaml:- name: gri30_multi
./interfaces/matlab/toolbox/GRI30.m:        s = Solution('gri30.xml', 'gri30_multi');
./interfaces/cython/cantera/data/gri30_highT.yaml:- name: gri30_multi
./interfaces/cython/cantera/data/gri30_highT.cti:ideal_gas(name = "gri30_multi",
./interfaces/cython/cantera/data/gri30.cti:ideal_gas(name = "gri30_multi",
./data/inputs/gri30_highT.cti:ideal_gas(name = "gri30_multi",
./data/inputs/gri30.cti:ideal_gas(name = "gri30_multi",
[...]
```
### Steps to reproduce
1. use newest development version (compiled from source)

ischoegl:

Fixes #736 

- catch `gri30_mix` and `gri30_multi` phase definitions for YAML and issue warnings
- replace occurrences of `gri30.cti` and `gri30.xml` By `gri30.yaml` in test suite

An attempt to load deprecated phases from `gri30.yaml` has the expected behavior, but issues a warning:

```
In [1]: import cantera as ct

In [2]: gas = ct.Solution('gri30.yaml', 'gri30_mix')
DeprecationWarning: The phases 'gri30_mix' and 'gri30_multi' are no longer defined in the input file 'gri30.yaml'. The YAML version uses the 'Mix' transport manager by default instead, which can be overridden by specifying a transport manager via the `transport_model' keyword argument during initialization. Support for transitional behavior will be removed after Cantera 2.5.

In [3]: gas.name, gas.composite
Out[3]: ('gri30_mix', ('IdealGas', 'Gas', 'Mix'))

In [4]: gas = ct.Solution('gri30.yaml', 'gri30_multi')
DeprecationWarning: The phases 'gri30_mix' and 'gri30_multi' are no longer defined in the input file 'gri30.yaml'. The YAML version uses the 'Mix' transport manager by default instead, which can be overridden by specifying a transport manager via the `transport_model' keyword argument during initialization. Support for transitional behavior will be removed after Cantera 2.5.

In [5]: gas.name, gas.composite
Out[5]: ('gri30_multi', ('IdealGas', 'Gas', 'Multi'))
```


interfaces/cython/cantera/base.pyx
```diff
@@ -111,8 +109,21 @@ cdef class _SolutionBase:
         elif source:
             root = AnyMapFromYamlString(stringify(source))
 
-        phaseNode = root[stringify("phases")].getMapWhere(stringify("name"),
-                                                          stringify(name))
+        if name == 'gri30_mix' or name == 'gri30_multi':
```
bryanwweber: This can be simplified to

```suggestion
        if name in ["gri30_mix", "gri30_multi"]:
```

interfaces/cython/cantera/base.pyx
```diff
@@ -111,8 +109,21 @@ cdef class _SolutionBase:
...
+            # catch transitional behavior
+            warnings.warn(
+                "The phases 'gri30_mix' and 'gri30_multi' are no longer "
+                "defined in the input file 'gri30.yaml'. The YAML version uses "
+                "the 'Mix' transport manager by default instead, which can be "
+                "overridden by specifying a transport manager via the "
+                "`transport_model' keyword argument during initialization. "
+                "Support for transitional behavior will be removed after "
+                "Cantera 2.5.", DeprecationWarning)
```
bryanwweber: I think the message is a little too broad, and I'd prefer if it was explicit about how users could fix the warning. Something like:
```
The phases named 'gri30_mix' and 'gri30_multi' are no longer defined
in the input file 'gri30.yaml'. By default, 'gri30.yaml' uses the mixture-averaged
transport model. The following options are available:

* ct.Solution('gri30.yaml')  # Mixture-averaged transport
* ct.Solution('gri30.yaml', transport_model="mix")  # Mixture-averaged transport
* ct.Solution('gri30.yaml', transport_model=None)  # No transport model, useful for 0-D simulations
* ct.Solution('gri30.yaml', transport_model="multicomponent")  # Multicomponent transport
```
This is now bordering on detailed documentation... maybe some of the examples could be eliminated in favor of a link to a page on our website (or on the GH Wiki) with a more detailed explanation. Regardless, I think at least one example of the appropriate syntax should be kept.

interfaces/cython/cantera/base.pyx
```diff
@@ -111,8 +109,21 @@ cdef class _SolutionBase:
...
+            phaseNode = root[stringify("phases")].getMapWhere(stringify("name"),
+                                                              stringify("gri30"))
```
bryanwweber: If I understand the new default correctly, this isn't needed, right?

interfaces/cython/cantera/examples/onedim/diffusion_flame.py
```diff
@@ -22,7 +22,7 @@
 
 # Create the gas object used to evaluate all thermodynamic, kinetic, and
 # transport properties.
-gas = ct.Solution('gri30.xml', 'gri30_mix')
+gas = ct.Solution('gri30.yaml', 'gri30')
```
bryanwweber: Might as well drop the phase name here, since there's only one. Likewise in the other examples.

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -81,6 +83,31 @@ def test_gri30(self):
         ref, gas = self.checkConversion('gri30.xml', 'gri30_test')
         self.checkKinetics(ref, gas, [300, 1500], [5e3, 1e5, 2e6])
 
+    def test_gri30_phases(self):
+
+        # cause all warnings to always be triggered.
+        warnings.simplefilter("always")
+
+        with warnings.catch_warnings(record=True) as w:
```
bryanwweber: I think you can use `assertWarnsRegex()` here, right?

test/matlab/TestImport.m
```diff
@@ -4,8 +4,8 @@
             self = self@TestCase(name);
         end
 
-        function testImportXML(self)
-            gas = Solution('gri30.xml', 'gri30_mix');
+        function testImportYAML(self)
```
bryanwweber, line 7: As @speth noted in #768, we shouldn't necessarily remove tests of the XML interface just yet. It's probably worth adding a separate test of the YAML interface here

bryanwweber (changes requested):
I think this solution is not surgical enough at the moment. The main problem I foresee is users who deliberately create their own version of GRI-3.0 with these phase definitions in them. I'm always surprised when I see someone on the Users' Group who has a copy of `gri30.cti` in their working directory, but as I recall it has happened multiple times. Is there any way to have this warning raised if and only if the `gri30.yaml` file comes from the installed data location?

interfaces/cython/cantera/base.pyx
ischoegl: It is still needed as the old names are processed elsewhere.

test/matlab/TestImport.m
ischoegl: Adding a separate test here.

interfaces/cython/cantera/test/test_convert.py
ischoegl: correct. Didn't know this one (https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarnsRegex)

ischoegl:
@bryanwweber ... thank you for the review!

Regarding your comment about users potentially keeping local copies of `gri30.cti`or `gri30.xml`: this is implicitly taken care of, as the warning is only issued if users attempt to load a `yaml` file. 

```
In [2]: gas = ct.Solution('gri30.yaml', 'gri30_mix')
/usr/bin/ipython3:1: DeprecationWarning: The phases named 'gri30_mix' and 'gri30_multi' are no longer defined in the input file 'gri30.yaml'. By default, 'gri30.yaml' uses the mixture-averaged transport model. The following options are available:
 * ct.Solution('gri30.yaml')  # Mixture-averaged transport
 * ct.Solution('gri30.yaml', transport_model='mix')  # Mixture-averaged transport
 * ct.Solution('gri30.yaml', transport_model=None)  # No transport model, useful for 0-D simulations
 * ct.Solution('gri30.yaml', transport_model='multicomponent')  # Multicomponent transport
Support for transitional behavior will be removed after Cantera 2.5.

In [3]: gas = ct.Solution('gri30.cti', 'gri30_mix') # does not issue a warning

In [4]: gas = ct.Solution('gri30.xml', 'gri30_mix') # does not issue a warning
```

However, you are correct that the solution needs to be more surgical: a custom `yaml` file may contain valid `gri30_mix` definitions, i.e. they should be used when present. A case in point is anything converted using `cti2yaml` or `ctml2yaml` ... which applies to `gri30_highT.yaml` etc.

~Also: unit tests appear to load from `test/work/python/`, which contains a converted `gri30.yaml` (this causes the unit test for the deprecation warning to fail: at the moment it is commented out). Any thoughts on this (the simplest solution is to eliminate the test)?~

interfaces/cython/cantera/base.pyx
```diff
@@ -111,8 +109,32 @@ cdef class _SolutionBase:
         elif source:
             root = AnyMapFromYamlString(stringify(source))
 
-        phaseNode = root[stringify("phases")].getMapWhere(stringify("name"),
-                                                          stringify(name))
+        try:
+            phaseNode = root[stringify("phases")].getMapWhere(
+                stringify("name"), stringify(name))
+        except CanteraError as err:
```
bryanwweber: I'm not really a fan of this approach either. One reason is that any errors that are generated during parsing of the file I think would have a double layer of stack trace printed:
```python
>>> try:
...     raise ValueError("Error")
... except ValueError as err:
...     raise err
...
  File "<stdin>", line 4, in <module>
  File "<stdin>", line 2, in <module>
ValueError: Error
```
(although you can get around that by
```python
>>> try:
...     raise ValueError("Error")
... except ValueError:
...     raise
...
  File "<stdin>", line 2, in <module>
ValueError: Error
```
)
I'd really prefer if there were a way to check that the file is being loaded from the Cantera data location, since from my perspective, that's the only one that we have control over and therefore the only one we should care about. The `AnyMap` has a `__file__` field, I believe, can that be checked for one of the known Cantera locations?

ischoegl:
@bryanwweber ... I believe it's now all fixed (and `scons test-matlab` is confirmed as I'm unsure that it is available on AppVeyor). FYI, on MATLAB the warning is:
```
>> gas = Solution('gri30.yaml','gri30_mix')
DeprecationWarning: newPhase: 
The phases named 'gri30_mix' and 'gri30_multi' are no longer defined in the input file 'gri30.yaml'. By default, 'gri30.yaml' uses the mixture-averaged transport model. The following options are available:
 * 'Mix': Mixture-averaged transport
 * 'Multi': Multicomponent transport
 * 'None': No transport model, useful for 0-D simulations
Support for transitional behavior will be removed after Cantera 2.5.
```
It's hard to make it more specific as the warning is raised from the C++ layer (i.e. I cannot refer to MATLAB specific syntax). As an aside, the `None` option is newly added to MATLAB ...

src/thermo/ThermoFactory.cpp
```diff
@@ -129,10 +129,32 @@ ThermoPhase* newPhase(const std::string& infile, std::string id)
...
+        } catch (CanteraError& err) {
+            if (name=="gri30_mix" || name=="gri30_multi") {
+                warn_deprecated("newPhase",
+                    "\nThe phases named 'gri30_mix' and 'gri30_multi' are no "
+                    "longer defined in the input file 'gri30.yaml'. By default, "
+                    "'gri30.yaml' uses the mixture-averaged transport model. The "
+                    "following options are available:\n"
+                    " * 'Mix': Mixture-averaged transport\n"
+                    " * 'Multi': Multicomponent transport\n"
+                    " * 'None': No transport model, useful for 0-D simulations\n"
```
bryanwweber, line 147: It would be more useful to show how users can fix this warning - where does this option get passed?

src/thermo/ThermoFactory.cpp
```diff
@@ -129,10 +129,32 @@ ThermoPhase* newPhase(const std::string& infile, std::string id)
...
+        std::string name = id;
+        try {
+            AnyMap& phase = root["phases"].getMapWhere("name", name);
+            unique_ptr<ThermoPhase> t(newThermoPhase(phase["thermo"].asString()));
+            setupPhase(*t, phase, root);
+            return t.release();
+        } catch (CanteraError& err) {
+            if (name=="gri30_mix" || name=="gri30_multi") {
+                warn_deprecated("newPhase",
+                    "\nThe phases named 'gri30_mix' and 'gri30_multi' are no "
```
bryanwweber: I noticed this when fixing up some of the output for #722... It seems less than optimal that messages have to start with a newline character. I'm not sure how many of these messages are intended to be stretched over multiple lines, but I assume it is a bunch of them. Can this be resolved in `warn_deprecated`? Maybe outside the scope of this PR, but I wanted to mention it (I wish I had seen it in #741)

bryanwweber (changes requested):
For the Python interface, at least, it doesn't appear that the appropriate transport model is set (I could be wrong, I'm not sure if there's an extra step that calls into C++ that I'm missing in the diff view here). If a user is passing `'gri30_multi'` they would expect that Multi-component transport is enabled, but they'll get an object with mixture-averaged transport. Can that be resolved?

bryanwweber:
> and `scons test-matlab` is confirmed as I'm unsure that it is available on AppVeyor

MATLAB can't be tested on any of the public, free CI services because we can't install MATLAB on them. @speth has a buildbot instance set up on some hardware at MIT that runs the (anemic) MATLAB test suite IIRC, but I can't recall the URL for that, or if it's public facing.

src/thermo/ThermoFactory.cpp
ischoegl: Let's keep this separate, but I'm ok with that change.

src/thermo/ThermoFactory.cpp
ischoegl: Will add a line.

interfaces/cython/cantera/base.pyx
ischoegl: Since this will only stay in temporarily for 2.5, I'd prefer to use the default mechanism (while adopting your suggestion).

bryanwweber: > Since this will only stay in temporarily for 2.5, I'd prefer to use the default mechanism

I'm not sure what this means? What default mechanism?

ischoegl:
> If a user is passing 'gri30_multi' they would expect that Multi-component transport is enabled, but they'll get an object with mixture-averaged transport. Can that be resolved?

I cannot reproduce this:
```
n [1]: import cantera as ct

In [2]: gas = ct.Solution('gri30.yaml', 'gri30_multi')
/usr/bin/ipython3:1: DeprecationWarning: The phases named 'gri30_mix' and 'gri30_multi' are no longer defined in the input file 'gri30.yaml'. By default, 'gri30.yaml' uses the mixture-averaged transport model. The following options are available:
 * ct.Solution('gri30.yaml')  # Mixture-averaged transport
 * ct.Solution('gri30.yaml', transport_model='mix')  # Mixture-averaged transport
 * ct.Solution('gri30.yaml', transport_model=None)  # No transport model, useful for 0-D simulations
 * ct.Solution('gri30.yaml', transport_model='multicomponent')  # Multicomponent transport
Support for transitional behavior will be removed after Cantera 2.5.
  #! /bin/sh

In [3]: gas.composite
Out[3]: ('IdealGas', 'Gas', 'Multi')
```

interfaces/cython/cantera/base.pyx
ischoegl: Stack trace is fixed.

src/thermo/ThermoFactory.cpp
ischoegl: Added examples for MATLAB and C++

interfaces/cython/cantera/base.pyx
ischoegl: Default mechanism means that I'll let let:
```
phaseNode = root[stringify("phases")].getMapWhere(
                 stringify("name"), stringify(name))
```
check whether the phase is actually defined. I.e. I referred to `getMapWhere`

ischoegl: Default mechanism means that I'll let let:
```
phaseNode = root[stringify("phases")].getMapWhere(
                 stringify("name"), stringify(name))
```
check whether the phase is actually defined. I.e. I referred to `getMapWhere`

ischoegl: Default mechanism means that I'll let let:
```
phaseNode = root[stringify("phases")].getMapWhere(
                 stringify("name"), stringify(name))
```
check whether the phase is actually defined. I.e. I referred to `getMapWhere`

ischoegl: Stack is fixed, btw.

ischoegl: Error stack is fixed btw.

ischoegl:
Regarding MATLAB:

```
>> gas = Solution('gri30.yaml','gri30_multi');
DeprecationWarning: newPhase: The phases named 'gri30_mix' and 'gri30_multi' are no longer defined in the input file 'gri30.yaml'. By default, 'gri30.yaml' uses the mixture-averaged transport model. The following options are available:
 * 'Mix': Mixture-averaged transport
 * 'Multi': Multicomponent transport
 * 'None': No transport model, useful for 0-D simulations
Examples for setting of transport model:
 - MATLAB: Solution('gri30.yaml', 'gri30', 'None')
 - C++: newSolution("gri30.yaml", "gri30", "Mix")
Support for transitional behavior will be removed after Cantera 2.5.
>> multi = multiDiffCoeffs(gas); % does not throw error
```
and, in the same session (warning is only shown the first time)
```
>> gas = Solution('gri30.yaml','gri30_mix');
>> multi = multiDiffCoeffs(gas);
Error using ctmethods
[...]
```

interfaces/cython/cantera/base.pyx
ischoegl: > 'd really prefer if there were a way to check that the file is being loaded from the Cantera data location, since from my perspective, that's the only one that we have control over and therefore the only one we should care about.

I'm not sure I agree: the (temporary) check should warn users that they are trying to load `gri30_mix` or `gri30_multi` from a `YAML` file that doesn't contain it. If they traded a *new* `gri30.yaml` from any other location than the official one, the same rationale would apply? Now that the fix is implemented, no exception is raised if the phase is actually defined and loaded as intended. 

After Cantera 2.5, users will get an error if they try to load `gri30_mix` from the official `gri30.yaml` and the rest will still work as intended.

src/thermo/ThermoFactory.cpp
ischoegl: I'm ok with the change, but let's keep it out of the PR.

PS: the intent was to be consistent with the Python `warnings.warn` interface, which does not introduce a newline.

bryanwweber: In that case, then I don't think you should introduce a newline as the first character here.

interfaces/cython/cantera/base.pyx
bryanwweber: > If they traded a new gri30.yaml from any other location than the official one, the same rationale would apply?

OK, I can see this point. If the `gri30_mix` or `gri30_multi` phases are present in a YAML file they will be used. The only time this warning is shown is if the user-specified phase name is not in the YAML file and they have passed one of those two options. This seems reasonable.

bryanwweber: However, I think the warning message needs to be improved in this case. If a user naively converts the CK version of GRI-3.0 to YAML and uses it with an example that shows the use of these phases, the error message might not make much sense to them. I'll leave a suggestion in the message.

bryanwweber:
> > If a user is passing 'gri30_multi' they would expect that Multi-component transport is enabled, but they'll get an object with mixture-averaged transport. Can that be resolved?
>
> I cannot reproduce this:

Thanks for checking! That's what I hoped would happen but couldn't tell from the diff view 😄

interfaces/cython/cantera/base.pyx
ischoegl: Hm. There won't be any difference in terms what the code does, as the `gri30.yaml` file will be (presumably) identical to the 'official' version. But I do see your point.

ischoegl:
> Thanks for checking! That's what I hoped would happen but couldn't tell from the diff view smile

You're welcome. For these exceptions, the transport model simply goes by phase name (which doesn't change), and the conversion is caught elsewhere. (Transport models get instantiated much later.) I actually added two lines to the unit tests to make sure that this behavior won't change ...

bryanwweber:
@ischoegl What about this warning message?
```python
            if name in ['gri30_mix', 'gri30_multi']:
                input_file = pystr(root[stringify("__file__")].asString())
                # catch transitional behavior
                warnings.warn(dedent("""
                    The phase named '{}' is not defined in the input file '{}'.
                    Examples that use this phase name are deprecated. Instead, you
                    should use one of the following options to choose a transport
                    model:

                    * ct.Solution("gri30.yaml")  # Mixture-averaged transport, default
                    * ct.Solution("gri30.yaml", transport_model="mix")  # Mixture-averaged transport
                    * ct.Solution("gri30.yaml", transport_model=None)  # No transport model, useful for 0-D simulations
                    * ct.Solution("gri30.yaml", transport_model="multicomponent")  # Multicomponent transport
                """.format(name, input_file)), DeprecationWarning)
```

Maybe add `This warning will convert to an error after Cantera 2.5` at the end.

Rewriting this message, I went back and forth on whether user-generated input files should be handled in this way. On the one hand, it is nice for a user if they're following an example and the end up using one of these phase names and get warned about it. On the other hand, if I were that user, I'm probably very new to Cantera, and having warnings like this thrown at me is not super helpful.

FYI, `asString()` has to be added to `_cantera.pxd` for this to compile and you have to add `from textwrap import dedent` to the top of `base.pyx`.

One edge case I just thought of (this is a little convoluted, but bear with me 😄 ):
1. User creates their own `gri30.yaml` by converting the CK files
2. User changes the default phase name in the file, either during conversion or afterwards
3. User tries to load that file and passes one of these deprecated phase names
4. Kaboom!
```python
In [3]: ct.Solution("./gri30.yaml", "gri30_mix")
<ipython-input-3-1e59bb5fc60e>:1: DeprecationWarning:
The phase named 'gri30_mix' is not defined in the input file './gri30.yaml'.
Examples that use these phases are deprecated.
Instead, you should use one of the following options:

* ct.Solution('gri30.yaml')  # Mixture-averaged transport
* ct.Solution('gri30.yaml', transport_model='mix')  # Mixture-averaged transport
* ct.Solution('gri30.yaml', transport_model=None)  # No transport model, useful for 0-D simulations
* ct.Solution('gri30.yaml', transport_model='multicomponent')  # Multicomponent transport

  ct.Solution("./gri30.yaml", "gri30_mix")
~/GitHub/cantera/interfaces/cython/cantera/base.pyx in cantera._cantera._SolutionBase._init_yaml()
    113         try:
--> 114             phaseNode = root[stringify("phases")].getMapWhere(
    115                 stringify("name"), stringify(name))

CanteraError:
InputFileError thrown by AnyValue::getMapWhere:
Error on line 16 of ./gri30.yaml:
List does not contain a map where 'name' = 'gri30_mix'
|  Line |
|    11 | date: Fri, 13 Dec 2019 16:19:11 -0500
|    12 |
|    13 | units: {length: cm, time: s, quantity: mol, activation-energy: cal/mol}
|    14 |
|    15 | phases:
>    16 > - name: gas
          ^
|    17 |   thermo: ideal-gas
|    18 |   elements: [O, H, C, N, Ar]
|    19 |   species: [H2, H, O, O2, OH, H2O, HO2, H2O2, C, CH, CH2, CH2(S), CH3, CH4,


During handling of the above exception, another exception occurred:

<ipython-input-3-1e59bb5fc60e> in <module>
----> 1 ct.Solution("./gri30.yaml", "gri30_mix")

~/GitHub/cantera/interfaces/cython/cantera/base.pyx in cantera._cantera._SolutionBase.__cinit__()
     54         # Parse inputs
     55         if infile.endswith('.yml') or infile.endswith('.yaml') or yaml:
---> 56             self._init_yaml(infile, name, adjacent, yaml)
     57         elif infile or source:
     58             self._init_cti_xml(infile, name, adjacent, source)

~/GitHub/cantera/interfaces/cython/cantera/base.pyx in cantera._cantera._SolutionBase._init_yaml()
    128                     * ct.Solution("gri30.yaml", transport_model=None)  # No transport model, useful for 0-D simulations
    129                     * ct.Solution("gri30.yaml", transport_model="multicomponent")  # Multicomponent transport
--> 130                 """.format(name, input_file)), DeprecationWarning)
    131                 phaseNode = root[stringify("phases")].getMapWhere(
    132                     stringify("name"), stringify("gri30"))

CanteraError:
InputFileError thrown by AnyValue::getMapWhere:
Error on line 16 of ./gri30.yaml:
List does not contain a map where 'name' = 'gri30'
|  Line |
|    11 | date: Fri, 13 Dec 2019 16:19:11 -0500
|    12 |
|    13 | units: {length: cm, time: s, quantity: mol, activation-energy: cal/mol}
|    14 |
|    15 | phases:
>    16 > - name: gas
          ^
|    17 |   thermo: ideal-gas
|    18 |   elements: [O, H, C, N, Ar]
|    19 |   species: [H2, H, O, O2, OH, H2O, HO2, H2O2, C, CH, CH2, CH2(S), CH3, CH4,
```
Anyways, this sequence of events should raise an error (because the name they passed really isn't defined), so I'm not sure why the user would expect it to work in the first place. And it seems pretty unlikely to happen. Just thought I'd throw it out there.

ischoegl:
@bryanwweber ... thanks for the suggestion. Will adopt. 

Otherwise, interesting scenario! Unfortunately, this is even raised if the phase name is not changed in `ck2yaml`, so it's not that unlikely. I believe there is a way to just have the default be used (the first phase specified in the file), which would be the safe thing to do. Thanks for catching this!

speth:
If we are effectively making the data files part of the API, where changes are subject to deprecation warnings, then it might be worth discussing a more general approach to making such changes, rather than a one-off for just the `gri30_multi` and `gri30_mix` phase definitions. I'm thinking of this in terms of the comment I made on #768, where I noted that I'd like to remove (or deprecate and later remove) some of the currently-included files.

One possible implementation would be to add a field to either the file or the phase definition that is checked for when the file is loaded, and used to issue a deprecation warning. This would have the benefit of not affecting copies of files (modified or not) that users have placed on their system outside the Cantera data directories. For the YAML format, a file-level deprecation could be written:
```yaml
generator: ctml2yaml
cantera-version: 2.5.0a3
date: Thu, 21 Nov 2019 19:34:25 -0500
input-files: [build\data\gri30.xml]
deprecated: >
  The input file silicon_carbide.yaml is deprecated and will be removed
  after Cantera X.Y
```
Or an individual phase definition could be deprecated by adding:
```yaml
phases:
- name: gri30_mix
  thermo: ideal-gas
  transport: mixture-averaged
  kinetics: gas
  reactions: all
  state: {T: 300.0 K, P: 1.01325e+05 Pa}
  deprecated: >
    The gri30_mix phase definition will be removed from gri30.yaml
    after Cantera X.Y.
```
In CTI, there could be a top-level `deprecated` directive that acts on the file. For individual phases, we could put this into the `options` field, e.g.
```py
ideal_gas(name = "gri30_mix",
      elements = " O  H  C  N  Ar ",
      species = """ H2  H  O  O2  OH  H2O  HO2  H2O2  C  CH 
                   CH2  CH2(S)  CH3  CH4  CO  CO2  HCO  CH2O  CH2OH  CH3O 
                   CH3OH  C2H  C2H2  C2H3  C2H4  C2H5  C2H6  HCCO  CH2CO  HCCOH 
                   N  NH  NH2  NH3  NNH  NO  NO2  N2O  HNO  CN 
                   HCN  H2CN  HCNN  HCNO  HOCN  HNCO  NCO  N2  AR  C3H7 
                   C3H8  CH2CHO  CH3CHO """,
      reactions = "all",
      transport = "Mix",
      options = ["deprecated: The gri30_mix phase definition will be removed from gri30.cti"
                 " after Cantera X.Y."]
      initial_state = state(temperature = 300.0,
                        pressure = OneAtm)    )
```

ischoegl:
@speth ... I think adding deprecation tags is a good idea. The question boils down to: Would you suggest to ditch the `ck2yaml` route for `gri30.yaml` in favor of `ctml2yaml`? It’s a legitimate approach, albeit different from this PR. 

The one issue that would remain is that the default phase for `gri30.yaml` does not have a transport model, and thus is not compatible with default behavior of auto-generated yaml files, even if deprecation tags are used to eliminate other superfluous phase entries. Overall, I’d still advocate for a consistent default (e.g. mixture-averaged transport if transport data are available), which i believe does require some input-specific transitional handling as this PR proposes.

bryanwweber:
> thus is not compatible with default behavior of auto-generated yaml files

I'm not sure what you mean by auto-generated, but it's perfectly fine to have a YAML file without transport data.

> i believe does require some input-specific transitional handling as this PR proposes.

I don't think that would be necessary. If the default changes from no transport to mixture-averaged, I don't see a problem with that, or a need to warn anyone. IMO the one advantage this approach has is showing a way to turn off transport and reduce the time it takes to load the file, especially for large mechanisms where the only real possibility is zero-D simulations anyways. But we could add that to an example

ischoegl:
@bryanwweber ... to clarify

CTI/XML offer the following phases
```
gri30: no transport # <-- loaded by default if no phase is specified
gri30_mix: mixture-averaged transport
gri30_multi:. multi-compon transport
```
The (current) gri30.yaml generated by `ck2yaml` is incompatible (and eventually led to this PR), as it has a single phase
```
gri30: mixture-averaged transport # <-- loaded by default
```
Note that any `ck2yaml` generated input file will have a transport model loaded by default if data are available (which is where `gri30.cti` differs).

As indicated, the question boils down to whether:
1. the old behavior should be preserved, i.e. do not load transport data. (Solution: create gri30 via `cti2yaml`; and use deprecation tags) ... completely separate from this PR
2. `gri30.yaml` should exhibit the regular behavior expected from input generated from `ck2yaml` if transport data are included (i.e. load transport data if no phase is specified) ... this PR in its current form. 
3. There are other options, e.g. (a) change default loading behavior to no transport model, or (b) add a flag to `ck2yaml` that specifies what transport model is attached to the single phase that is created

If the consensus is to move away from (2), I'd suggest closing this PR and starting a new one as none of this PR's code carries over.

src/kinetics/KineticsFactory.cpp
```diff
@@ -80,7 +80,12 @@ unique_ptr<Kinetics> newKinetics(std::vector<ThermoPhase*>& phases,
 
     if (extension == "yml" || extension == "yaml") {
         AnyMap root = AnyMap::fromYamlFile(filename);
-        AnyMap& phaseNode = root["phases"].getMapWhere("name", phase_name);
+        std::string name = phase_name;
+        if (name=="gri30_mix" || name=="gri30_multi") {
```
speth:
```suggestion
        if (name == "gri30_mix" || name == "gri30_multi") {
```

src/thermo/ThermoFactory.cpp
```diff
@@ -129,10 +129,36 @@ ThermoPhase* newPhase(const std::string& infile, std::string id)
...
-        setupPhase(*t, phase, root);
-        return t.release();
+        std::string name = id;
+        try {
+            AnyMap& phase = root["phases"].getMapWhere("name", name);
+            unique_ptr<ThermoPhase> t(newThermoPhase(phase["thermo"].asString()));
+            setupPhase(*t, phase, root);
+            return t.release();
+        } catch (CanteraError& err) {
+            if (name=="gri30_mix" || name=="gri30_multi") {
```
speth:
```suggestion
            if (name == "gri30_mix" || name == "gri30_multi") {
```

src/transport/TransportFactory.cpp
```diff
@@ -100,7 +100,13 @@ Transport* TransportFactory::newTransport(thermo_t* phase, int log_level)
     std::string transportModel = "None";
     XML_Node& phaseNode = phase->xml();
     AnyMap& input = phase->input();
-    if (input.hasKey("transport")) {
+    if (phase->name()=="gri30_mix") {
```
speth:
```suggestion
    if (phase->name() == "gri30_mix") {
```

src/transport/TransportFactory.cpp
```diff
@@ -100,7 +100,13 @@ Transport* TransportFactory::newTransport(thermo_t* phase, int log_level)
...
+        // handle deprecated 'gri30_mix' phase; warning is already issued
+        transportModel = "Mix";
+    } else if (phase->name()=="gri30_multi") {
```
speth:
```suggestion
    } else if (phase->name() == "gri30_multi") {
```

speth (commented):
> The one issue that would remain is that the default phase for `gri30.yaml` does not have a transport model, and thus is not compatible with default behavior of auto-generated yaml files, even if deprecation tags are used to eliminate other superfluous phase entries. Overall, I’d still advocate for a consistent default (e.g. mixture-averaged transport if transport data are available), which i believe does require some input-specific transitional handling as this PR proposes.

Making `gri30.yaml` consistent with what a user would end up generating from their own Chemkin input files was the main point of removing the custom phase definitions of `gri30.cti`. I agree that "mixture-averaged transport if transport data are available", which is the convention inherited from `ck2cti` is a reasonable default.

> However, you are correct that the solution needs to be more surgical: a custom `yaml` file may contain valid `gri30_mix` definitions, i.e. they should be used when present. A case in point is anything converted using `cti2yaml` or `ctml2yaml` ... which applies to `gri30_highT.yaml` etc.

The logic that applies eliminating the alternate phase definitions from `gri30` applies to `gri30_highT` as well. The only thing that makes this trickier is that `gri30_highT.cti` seems to be a manually-edited copy of `gri30.cti`, rather than generated Chemkin-format files. So the only way to implement that change will be to commit the desired version of `gri30_highT.yaml` directly, perhaps as part of #768.

> One edge case I just thought of (this is a little convoluted, but bear with me  ):
> 
> 1. User creates their own `gri30.yaml` by converting the CK files
> 2. User changes the default phase name in the file, either during conversion or afterwards
> 3. User tries to load that file and passes one of these deprecated phase names
> 4. Kaboom!

This does indeed seem pretty convoluted, and I think the conditions are even more specific than this: the user would have to remove both the `gri30` and (say) the `gri30_mix` phase definitions, but then try to load the `gri30_mix` phase definition. And the only thing wrong with the error message is that it specifies that the missing name is `gri30` rather than `gri30_mix`.

bryanwweber:
@speth Are you no longer in favor of the `deprecated` field in the YAML file? I think that would actually be quite useful; it could also be used to deprecate a certain, say, phase or species thermo type, if necessary:

```yaml
phases:
- name: example
  thermo: something-deprecated
  deprecated: The 'something-deprecated' phase thermo type is deprecated...
  ...
```
Obviously, this would have to be added manually to any files that used that thermo type, and presumably there would be deprecation warnings when the thermo type was instantiated, but it would still be a useful field to have IMO.

If you would still like to pursue that avenue, I agree with @ischoegl that this PR should be closed. My preference is for suggestion (1) from @ischoegl, that is, retain the `gri30`, `gri30_mix`, and `gri30_multi` phases for 2.5.0, add the deprecation notice, and change all examples to explicitly specify the desired transport model and no longer use the phase names in our examples. Regardless of the outcome here, I'm planning to add the explicit `transport_model` option to the examples anyways as part of #768.

> I think the conditions are even more specific than this

I think the more likely case is a user converts the GRI-3.0 CK files to YAML, resulting in the default name of `gas` in the file (at least, I think that would be the default). As such, it's not quite as convoluted as I originally thought (@ischoegl pointed out this case). To be clear, my concern wasn't so much with the specificity of the error message, but rather with its length (it specifies that both `gri30` and `gri30_mix` are missing at different points). The most likely people to see that particular error message are probably relatively new users, and that would not be a particularly easy one to parse without some prior experience with Python tracebacks.

ischoegl:
> Making gri30.yaml consistent with what a user would end up generating from their own Chemkin input files was the main point of removing the custom phase definitions of gri30.cti. 

That was my understanding before starting this PR, which led me to implementation (2) :wink:. I personally agree with @speth that it's preferable to have consistent behavior. The switch of formats is imho an opportunity to get rid of inconsistent implementations.

I am not opposed to the `deprecated` flag whatsoever, but I believe it should be handled in another PR. It certainly is well aligned with #768, although it doesn't necessarily have to be addressed there.

PS: the convoluted error message can be fixed/avoided with fairly little effort.

speth:
> Are you no longer in favor of the `deprecated` field in the YAML file? I think that would actually be quite useful

No, I still think it would be useful, and I'd like to see it before we act on #768. Either the addition of a `deprecated` field in the input file or this PR would allow us to make the transition to a simpler version of `gri30.yaml`. The `deprecated` field would of course also be useful for several other situations. The one advantage of this approach would be that we wouldn't need to start of by introducing some deprecated phase definitions in `gri30.yaml`.

Perhaps we should put together an implementation of the `deprecated` field first and depending on how well that works we can decide whether that feature or the approach in this PR is the best way to deal with the `gri30_mix` / `gri30_multi` issue.

ischoegl:
@speth and @bryanwweber  ... I finished implementation (2), and made sure that the awkward error Bryan pointed out will not appear (there is a unit test for it).

> @speth: Making gri30.yaml consistent with what a user would end up generating from their own Chemkin input files was the main point of removing the custom phase definitions of gri30.cti.

To reiterate: if the above is the objective (as had been my understanding), then there are two implementation paths for how users would upgrade existing custom models:

(1) the `deprecated` field:
- Cantera 2.5: shows a deprecation warning for YAML, but has behavior of old CTI files
- Cantera 2.6: default transport model loading behavior changes from `None` to `Mix`
- Cantera 3.0: all models are fixed

(2) this PR:
- Cantera 2.5: emulates old CTI behavior to facilitate transition to YAML
- Cantera 2.6: all models are fixed

Imho the main advantage of (2) is that there will be exactly one version of `gri30.yaml` (single phase), whereas for implementation (1), there will be two versions (three phase for 2.5 and single phase for 2.6). Also, while it's possible to fix official cantera examples immediately, user code may have to be changed twice for (1), whereas it needs to change once for (2). I am *very* sceptical about adopting superfluous phases in the newly created YAML format.

In summary, while I believe the `deprecated` flag makes a lot of sense *elsewhere*, a change of default behavior for `gri30` requires more effort.

ischoegl:
PS: the two version of the warnings

Python
```
In [2]: gas = ct.Solution("./gri30.yaml", "gri30_mix")
/usr/bin/ipython3:1: DeprecationWarning: 
The phase named 'gri30_mix' is not defined in the input file './gri30.yaml'.
Examples that use this phase name are deprecated. Instead, you should
use one of the following options to choose a transport model:

* ct.Solution("gri30.yaml")  # Mixture-averaged transport, default
* ct.Solution("gri30.yaml", transport_model="Mix")  # Mixture-averaged transport
* ct.Solution("gri30.yaml", transport_model=None)  # No transport model, useful for 0-D simulations
* ct.Solution("gri30.yaml", transport_model="Multi")  # Multicomponent transport

This warning will convert to an error after Cantera 2.5.
```
C++/Matlab version (uses a *temporarily* modified example for illustration purposes)
```
$ ./combustor
DeprecationWarning: newPhase: The phase named 'gri30_mix' is not defined in the input file '/usr/local/share/cantera/data/gri30.yaml'.
Examples that use this phase name are deprecated. Instead, you should specify one of the following transport models:
 * 'Mix': Mixture-averaged transport
 * 'Multi': Multicomponent transport
 * 'None': No transport model, useful for 0-D simulations
Examples:
 - MATLAB: Solution('gri30.yaml', 'gri30', 'None')
 - C++: newSolution("gri30.yaml", "gri30", "Mix")
This warning will convert to an error after Cantera 2.5.
```

bryanwweber:
> @ischoegl: have to be changed twice for (1), whereas it needs to change once for (2). ... In summary, while I believe the `deprecated` flag makes a lot of sense elsewhere, a change of default behavior for `gri30` requires more effort.

I don't necessarily think this is true. I don't see a problem with changing the default phase in `gri30.yaml` to mixture-averaged transport and also retaining the `gri30_mix` phase. I can't think of any cases where changing the default to include transport would create a problem (certainly, the reverse is not true).

> @speth: The one advantage of this approach [in PR 737, this one] would be that we wouldn't need to start of by introducing some deprecated phase definitions in `gri30.yaml`.

This is true, but I don't think it's that big of an advantage. If we are going to be removing/changing other input files after using the `deprecated` field, we might as well do all the breaking changes at once, including `gri30.yaml`.

Another thought... From the warning message:
```
* ct.Solution("gri30.yaml")  # Mixture-averaged transport, default
```
I know that I wrote this message, but I keep thinking of ways it could be misinterpreted. For this line, this is only **definitely** the default for the built-in `gri30.yaml`. If a user converts the CK GRI-3.0 files themselves and doesn't include transport (not unreasonable, I rarely include transport data if I know I'm going to just be doing 0-D simulations), the warning message will be wrong.

There are just so many edge cases with trying to guess what a user meant to do with their own input files, I just can't shake the feeling that this warning should be limited only to files distributed with Cantera. 😞  Sorry to put this PR through the ringer, @ischoegl thanks for the constructive discussion!

ischoegl:
@bryanwweber ... I’m open for suggestions/clarifications. Getting the warning message *right* now should pay off later.

ischoegl:
Another comment: rather than porting CTI to YAML and immediately adding a `deprecated` field, wouldn't it preferable to add a `deprecated` tag and warning to CTI and never even include deprecated phases/input in YAML?

interfaces/cython/cantera/base.pyx
ischoegl: Resolved.

ischoegl:
Closing as solution proposed in #768 is less intrusive and able to handle other cases as well.

PR #741, opened 2019-11-04
Improve warning messages in C++ 

### LINKED ISSUE #683: Inconsistency between TROE formula and code when T** is zero.
### Cantera version
2.3.0
### Operating System
Linux Ubuntu18.04 64bit
### Python/MATLAB version
3.7
### Expected Behavior

When T** is not given, the default value is 0.0, as description in https://cantera.org/science/reactions.html . Then the last term exp(-T**/T) ==1

### Actual Behavior

When T** is not given, the last term exp(-T**/T) is ignored so this term is actually zero.

### Steps to reproduce
1.  With a very small T**
```python
cti_def = '''
    ideal_gas(name='gas', elements='O H C',
              species='gri30:H CH2 CH3 O2 H2O CO2',
              reactions='all',
              options=['skip_undeclared_elements', 'skip_undeclared_species', 'skip_undeclared_third_bodies'],
              initial_state=state(temperature=2000, pressure=101325))
    #  Reaction 50
falloff_reaction( "H + CH2 (+ M) <=> CH3 (+ M)",
         kf = [6.00000E+14, 0, 0],
         kf0   = [1.04000E+26, -2.76, 1600],
         falloff = Troe(A = 0.562, T3 = 91, T1 = 5836, T2 = 1e-10))
'''
gas = ct.Solution(source=cti_def)
gas.TPX = 2000,1e6,[1,1,1,1,1,1]
gas.net_rates_of_progress
## output array([6.57248322e+10])
```
2. Without T**
```python
cti_def = '''
    ideal_gas(name='gas', elements='O H C',
              species='gri30:H CH2 CH3 O2 H2O CO2',
              reactions='all',
              options=['skip_undeclared_elements', 'skip_undeclared_species', 'skip_undeclared_third_bodies'],
              initial_state=state(temperature=2000, pressure=101325))
    #  Reaction 50
falloff_reaction( "H + CH2 (+ M) <=> CH3 (+ M)",
         kf = [6.00000E+14, 0, 0],
         kf0   = [1.04000E+26, -2.76, 1600],
         falloff = Troe(A = 0.562, T3 = 91, T1 = 5836))
'''
gas = ct.Solution(source=cti_def)
gas.TPX = 2000,1e6,[1,1,1,1,1,1]
gas.net_rates_of_progress
## output array([2.8516357e+10])
```

### Comment

According to CHEMKIN's source code and theory guide, this behavior of cantera is consistent with the implementation of CHEMKIN but it is inconsistent with the cantera's document, CHEMKIN's theory guide and Troe's paper.

However, I do not know which is correct. It may influence the behavior of Weiyao's 89 species, 683 reactions reduced mechanism for kerosene.

ischoegl:

Fixes #683

- introduce `warn_user` in C++ layer (in analogy to `warn_deprecated`)
- create uniform formatting for warnings
- add information on method raising the warning
- replace various occurrences of ad-hoc formatted `writelog` warnings with `warn_user`
- issue `warn_user` for Troe when T2 (i.e. T**) is zero

**Notes**

Some `ck2cti`/`ck2yaml` converters appear to create length-4 arrays for Troe parameters, which raises the warning that was suggested for #683 in the unit test suite. I.e.

```
test_surface_mech (cantera.test.test_convert.ck2ctiTest) ... UserWarning: 'Troe::init': Zero T2 value is suppressed.
ok
test_reaction_units (cantera.test.test_convert.ck2yamlTest) ... UserWarning: 'Troe::init': Zero T2 value is suppressed.
UserWarning: 'Troe::init': Zero T2 value is suppressed.
UserWarning: 'Troe::init': Zero T2 value is suppressed.
UserWarning: 'Troe::init': Zero T2 value is suppressed.
ok
test_surface_mech (cantera.test.test_convert.ck2yamlTest) ... UserWarning: 'Troe::init': Zero T2 value is suppressed.
ok
``` 
It's probably not worth fixing `ck2cti`, but it also affects `ck2yaml`. Edit: the YAML import issue is fixed.

**Other thoughts**

Long-term, it may make sense to use a logger that allows for a meaningful differentiation of levels, e.g. [Boost.Log](https://theboostcpplibraries.com/boost.log). The introduction of `warn_user` now will facilitate a central replacement in `application.h/.cpp` in the future.


src/base/global.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "cantera/base/FactoryBase.h"
 #include "cantera/base/xml.h"
+#include "cantera/base/fmt.h"
```
ischoegl: Unused

include/cantera/base/global.h
```diff
@@ -187,6 +187,22 @@ void warn_deprecated(const std::string& method, const std::string& extra="");
...
+ *
+ * @param method  method name
+ * @param msg  C format string for the following arguments
+ * @param args  arguments used to interpolate the format string
+ */
+template <typename... Args>
+void warn_user(const std::string& method, const std::string& msg,
+               const Args&... args) {
+    _warn_user(method, fmt::format(msg, args...));
+}
```
speth: I think this should do something similar to `writelog` where `fmt::format` is used only if additional arguments for substitution are provided. This way, there's no need to worry about special characters being interpreted as part of a format string if there is no formatting to be done.

src/base/application.cpp
```diff
@@ -170,7 +170,14 @@ void Application::warn_deprecated(const std::string& method,
         return;
     }
     warnings.insert(method);
-    writelog("WARNING: '" + method + "' is deprecated. " + extra);
+    writelog("DeprecationWarning: '" + method + "': " + extra);
```
speth:
```suggestion
    writelog("DeprecationWarning: '{}': {}", method, extra);
```

src/base/application.cpp
```diff
@@ -170,7 +170,14 @@ void Application::warn_deprecated(const std::string& method,
...
+    writelogendl();
+}
+
+void Application::warn_user(const std::string& method,
+                            const std::string& extra)
+{
+    writelog("UserWarning: '" + method + "': " + extra);
```
speth:
```suggestion
    writelog("UserWarning: '{}': {}", method, extra);
```

src/thermo/PDSS_HKFT.cpp
```diff
@@ -314,10 +314,11 @@ void PDSS_HKFT::initThermo()
                 " not consistent with G and S: {} vs {} cal gmol-1",
                 sname, Hcalc/toSI("cal/gmol"), m_deltaH_formation_tr_pr);
         } else {
-            writelog("PDSS_HKFT::initThermo() WARNING: DHjmol for {} is not"
-                " consistent with G and S: calculated {} vs input {} cal gmol-1",
-                sname, Hcalc/toSI("cal/gmol"), m_deltaH_formation_tr_pr);
-            writelog("                                : continuing with consistent DHjmol = {}", Hcalc/toSI("cal/gmol"));
+            warn_user("PDSS_HKFT::initThermo()",
```
speth:
```suggestion
            warn_user("PDSS_HKFT::initThermo",
```

src/transport/MMCollisionInt.cpp
```diff
@@ -428,8 +428,9 @@ void MMCollisionInt::fit_omega22(int degree, doublereal deltastar,
     w[0]= -1.0;
     double rmserr = polyfit(n, degree, logT, values.data(), w.data(), o22);
     if (m_loglevel > 0 && rmserr > 0.01) {
-        writelogf("Warning: RMS error = %12.6g in omega_22 fit"
-                  "with delta* = %12.6g\n", rmserr, deltastar);
+        warn_user("MMCollisionInt::fit_omega22",
+            "RMS error = {:12.6g} in omega_22 fit"
```
speth:
```suggestion
            "RMS error = {:12.6g} in omega_22 fit "
```
(I realize that this was already missing a space)

test_problems/cxx_ex/output_blessed.txt
```diff
@@ -40,41 +40,36 @@ Description:
 Chemical equilibrium.
 Equilibrium composition and pressure for a range of temperatures at constant density.
 
-
-**** WARNING ****
+UserWarning: 'NasaPoly2::validate':
```
speth: I'm wondering if it would look better to format the warning message without quotes around the method name.

include/cantera/base/global.h
```diff
@@ -187,6 +187,22 @@ void warn_deprecated(const std::string& method, const std::string& extra="");
...
 void suppress_deprecation_warnings();
 
+//! helper function passing user warning to global handler
+void _warn_user(const std::string& method, const std::string& extra);
+
+/*!
+ * Print a user warning raised from *method*.
+ *
+ * @param method  method name
+ * @param msg  C format string for the following arguments
```
speth: This is not a "C format string" -- it's the Python-like style used by `fmt`.

test_problems/cxx_ex/equil_example1.cpp
```diff
@@ -55,7 +55,8 @@ int equil_example1(int job)
     std::cout << "Chemical equilibrium." << std::endl;
     if (job > 0) {
         std::cout << "Equilibrium composition and pressure for a "
-                  << "range of temperatures at constant density." << std::endl;
+                  << "range of temperatures at constant density."
+                  << std::endl << std::endl;
```
speth, line 59: It feels a little bit like cheating if you have to modify the output before a warning is issued to get it formatted as desired. One possibility to make multiline warnings more readable would be to automatically add a newline before and after a warning message if it already contains a newline.

src/kinetics/Falloff.cpp
```diff
@@ -43,6 +44,9 @@ void Troe::init(const vector_fp& c)
     }
 
     if (c.size() == 4) {
+        if (std::abs(c[3]) < SmallNumber) {
+            warn_user("Troe::init", "Zero T2 value is suppressed.");
+        }
```
speth, line 55: Given the complexity of the issue, I think this warning needs to provide some additional detail. I would suggest something along the lines of: "Unexpected parameter value T2 == 0. Omitting exp(T2/T) term from falloff expression. To suppress this warning, remove value for T2 from the input file. In the unlikely case that the exp(T2/T) term should be included with T2 effectively equal to 0, set T2 to a sufficiently small value (i.e. T2 < 1e-16)."

speth (changes requested):
This looks pretty good to me. I don't think it's a full fix for #683 yet, unless you also want to make the appropriate modifications to the Chemkin conversion scripts.

I would like to avoid adding Boost.Log to our dependencies -- While I generally like Boost, this is unfortunately one of parts that has a compiled library component, and my experience has been that these are very messy dependencies to deal with when you need to support multiple operating systems and a range of different Boost versions. However, a method for raising these warnings through the Python `warnings` module would be interesting as a future extension.

ischoegl:
@speth ... thanks for the review! Will make the changes which won’t take long.

> a method for raising these warnings through the Python warnings module would be interesting as a future extension.

I’ve been wondering about this also, but cannot think of an elegant solution. Definitely another PR

test_problems/cxx_ex/equil_example1.cpp
ischoegl: I added the extra newline here to remain consistent with the existing output, so this is pretty problem specific. I really don't like adding newlines before ... it's up to the user to format things as needed *after*.

test_problems/cxx_ex/output_blessed.txt
ischoegl: No problem. Also suggesting `CanteraWarning` to make things a little clearer

src/base/application.cpp
ischoegl: this `writelog` didn't use `fmt` ... using 
```C++
    writelog(fmt::format("CanteraWarning: {}: {}", method, extra));
```
without the ticks and using `CanteraWarning`.

ischoegl:
One final comment for the record: I noticed that the following warnings depend on a log level. I don't want to change it without discussion as this PR should not change the overall behavior.
https://github.com/Cantera/cantera/blob/6e60f2f529381962a3af9174a75eb8b2d129b351/src/transport/MMCollisionInt.cpp#L473-L493

ischoegl:
> This looks pretty good to me. I don't think it's a full fix for #683 yet, unless you also want to make the appropriate modifications to the Chemkin conversion scripts.

This depends on the philosophy: if `ck2yaml` should produce an exact translation of the chemikin input, then this PR fixes #683. If the goal is to pre-empt the behavior, `ck2yaml` should be amended. I don't have a strong opinion here, but would leave this PR as a partial fix as is.

test_problems/cxx_ex/output_blessed.txt
speth: 👍 on `CanteraWarning`

src/base/application.cpp
speth: Oh, I see. That's a bit tricky -- this resolves to `Application::Messages::writelog`, not the global `writelog` function.

speth (approved):
With the change to `ck2yaml`, I agree that this resolves #683.

PR #742, opened 2019-11-04
Prevent Cantera installation to source directory

### LINKED ISSUE #710: Improve error message when prefix is set to the source directory
### Cantera version
2.4.0 and master

### Expected Behavior
Trying to run `scons install` when `prefix` is set to the same directory as the source code should produce an error message that indicates that this is the problem.

### Actual Behavior
Build fails with the confusing error:
```
scons: *** Two different builders (<class 'SCons.Builder.BuilderBase'> and InstallBuilder) were specified for the same target: include/cantera/base/config.h
File "<source directory>/cantera/site_scons/site_tools/recursiveInstall.py", line 52, in RecursiveInstall
```

See reported installation problems in this Users' Group post: https://groups.google.com/forum/?fromgroups=#!topic/cantera-users/h-TYzPMs2PU

ischoegl:

Fixes #710

- add check in `SConstruct`

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -729,6 +729,11 @@ print()
 # *** Configure system-specific properties ***
 # ********************************************
 
+# Prevent setting prefix for Cantera installation to source directory
+if os.path.abspath(env['prefix']) == Dir('.').abspath:
+    print('ERROR: cannot install Cantera into source directory.')
+    exit(1)
+
 # Copy in external environment variables
 if env['env_vars'] == 'all':
     env['ENV'].update(os.environ)


SConstruct
```diff
@@ -1212,6 +1212,11 @@ for option, default in python_options:
             print("WARNING: Ignoring option 'python3_{0}'"
                 " because 'python_{0}' was also set.".format(option))
 
+# Prevent setting prefix for Cantera installation to source directory
+if os.getcwd() == os.path.abspath(env['prefix']):
```
bryanwweber: I'm not sure that `os.getcwd()` is the right thing here, since it may be possible to run SCons from outside the source directory (I'm not sure I've never tried). If the problem is the `'prefix'` is the same as the source directory, it would be better to check whether the `'prefix'` was the same as the directory where `SConstruct` is located.

If this is an error, should the exit code be 1 instead of 0?

ischoegl: True! Both are easily addressed.

PS: I noticed that the exit code isn't consistently 1 for errors in the build script.

bryanwweber (commented):
I think you can drop the `srcnode()`, just `Dir(".").abspath` seems to do the right thing. Also, it seems like this check should be done as soon as the env file is read. There's no need to do any configuration or any other checks if this one is just going to blow everything up anyways.

ischoegl:
> I think you can drop the `srcnode()`, just `Dir(".").abspath` seems to do the right thing. Also, it seems like this check should be done as soon as the env file is read. 

True, and done.

speth (approved):
Thanks, this is simple and effective.

Issue #703, opened 2019-09-10
Problems with code coverage


bryanwweber:
### Steps to reproduce
1. Open a pull request
2. Push commits multiple times
3. Coverage changes on files that weren't modified in the pull request

This seems to be a problem with the file `src/transport/GasTransport.cpp`. It is not clear why the coverage should be changing in this file. We should look at what is changing and try to fix either the test suite or the code to resolve this issue.

bryanwweber:
The comments here https://github.com/Cantera/cantera/pull/693#issuecomment-529120245 are a good example of what can happen

speth:
The change in coverage that's being reported is in `GasTransport::getMixDiffCoeffsMole`:

```c++
if (!m_bindiff_ok) {
   updateDiff_T();
}
```

Where the conditional is apparently reached 10 times in the test suite, and the predicate is reached either 0 or 1 time. What's strange is that I can only find two calls to `getMixDiffCoeffsMole` (or the Python property `mix_diff_coeffs_mole`) in the test suite. These occur in the Python test `TestTransport.test_mixtureAveraged`:

```python
    def test_mixtureAveraged(self):
        self.assertEqual(self.phase.transport_model, 'Mix')
        Dkm1 = self.phase.mix_diff_coeffs
        Dkm1b = self.phase.mix_diff_coeffs_mole
        Dkm1c = self.phase.mix_diff_coeffs_mass
        Dbin1 = self.phase.binary_diff_coeffs

        self.phase.transport_model = 'Multi'
        Dkm2 = self.phase.mix_diff_coeffs
        Dkm2b = self.phase.mix_diff_coeffs_mole
        Dkm2c = self.phase.mix_diff_coeffs_mass
        Dbin2 = self.phase.binary_diff_coeffs
        self.assertArrayNear(Dkm1, Dkm2)
        self.assertArrayNear(Dkm1b, Dkm2b)
        self.assertArrayNear(Dkm1c, Dkm2c)
        self.assertArrayNear(Dbin1, Dbin2)
        self.assertArrayNear(Dbin1, Dbin1.T)
```

I think this test should always exercise the code path that does not update the binary diffusion coefficients (they should be updated by earlier access to `mix_diff_coeffs`).

PR #743, opened 2019-11-07
[Test] Fix coverage consistency issue in getMixDiffCoeffsMass

speth:
Apparently, the corner case in the coverage testing described in #703  also affects the `getMixDiffCoeffsMass` method. With any luck, this should fix it for good.

### DIFF
--- a/interfaces/cython/cantera/test/test_transport.py
+++ b/interfaces/cython/cantera/test/test_transport.py
@@ -46,13 +46,23 @@ class TestTransport(utilities.CanteraTest):
         self.assertArrayNear(Dbin1, Dbin2)
         self.assertArrayNear(Dbin1, Dbin1.T)
 
-    def test_mixDiffCoeffsMole(self):
+    def test_mixDiffCoeffsChange(self):
         # This test is mainly to make code coverage in GasTransport.cpp
         # consistent by always covering the path where the binary diffusion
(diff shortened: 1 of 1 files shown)


bryanwweber:
So, my guess is that it depends on the order in which the tests are run, which determines whether or not `mix_diff_coeffs_*` will have `!m_bindiff_ok` `True` or `False`. If that's the case, we probably need to test both paths of the conditional. Is there any test we have where we know `!m_bindiff_ok` is `False` (i.e., `m_bindiff_ok` is `True`)? Otherwise this could keep happening.

speth:
This (existing) test covers the case where `m_bindiff_ok` is true:
https://github.com/Cantera/cantera/blob/43d02e95e30e7414d1d836be1a48724b5f0b5a7d/interfaces/cython/cantera/test/test_transport.py#L31-L47
The binary diffusion coefficients are updated in the first call to `mix_diff_coeffs`, and should should still be OK in the subsequent calls to `mix_diff_coeffs_mass` and `mix_diff_coeffs_mole` which occur without any change in the mixture state.

The new test that I added here should always exercise the branch where `m_bindiff_ok` is `false`. I still don't think any existing test should have been executing that branch, but this at least makes sure we are now doing so consistently.

bryanwweber (approved):

PR #745, opened 2019-11-09
Add Reaction factory

ischoegl:

- Add `ReactionFactory` in C++
- Replace hard-coded instantiation and setup in C++ and Python
- Remove magic numbers from Python interface

**Approach**

The main philosophy is to replace hard-coded reaction instantiation by an extendable `ReactionFactory` approach. None of this affects calculated outcomes. I.e. this PR repackages what's already there, with the difference that it becomes very simple to add additional reaction types.

Example: `GasKinetics` becomes fully flexible, i.e. someone can define a new reaction type externally and link it against cantera (`ctwrap` approach, where new C++ lambda functions are registered for an existing cantera factory object).

**Implementation**

* Replace magic number referenced function calls by name-mapped C++ lambda functions calls (which are mainly used for setup; actual kinetics calculations involve templated functions, i.e. there is no change from the current implementation whatsoever)
* C++ lambda functions recycle existing code directly

### DIFF
--- a/include/cantera/kinetics/GasKinetics.h
+++ b/include/cantera/kinetics/GasKinetics.h
@@ -67,6 +67,18 @@ public:
     //! reactions.
     virtual void update_rates_C();
 
+    //! register a function that adds a reaction
+    void reg_addRxn(const std::string& name,
+                    std::function<void(shared_ptr<Reaction>)> f) {
+        m_addRxn[name] = f;
+    }
+
+    //! register a function that modifies a reaction
+    void reg_modRxn(const std::string& name,
+                    std::function<void(size_t, shared_ptr<Reaction>)> f) {
+        m_modRxn[name] = f;
+    }
+
 protected:
     //! Reaction index of each falloff reaction
     std::vector<size_t> m_fallindx;
@@ -117,6 +129,14 @@ protected:
 
     //! Update the equilibrium constants in molar units.
     void updateKc();
+
+    //! map functions adding reactions
+    std::unordered_map<std::string,
+        std::function<void(shared_ptr<Reaction>)>> m_addRxn;
+
+    //! map functions modifying reactions
+    std::unordered_map<std::string,
+        std::function<void(size_t, shared_ptr<Reaction>)>> m_modRxn;
 };
 
 }
--- a/include/cantera/kinetics/Reaction.h
+++ b/include/cantera/kinetics/Reaction.h
@@ -28,6 +28,12 @@ public:
              const Composition& products);
     virtual ~Reaction() {}
 
+    virtual void setup(const XML_Node& rxn_node) {
+    }
+
+    virtual void setup(const AnyMap& node, const Kinetics& kin) {
+    }
+
     //! The reactant side of the chemical equation for this reaction
     virtual std::string reactantString() const;
 
@@ -88,6 +94,9 @@ public:
                        const Arrhenius& rate);
     virtual void validate();
 
+    virtual void setup(const XML_Node& rxn_node);
+    virtual void setup(const AnyMap& node, const Kinetics& kin);
+
     Arrhenius rate;
     bool allow_negative_pre_exponential_factor;
 };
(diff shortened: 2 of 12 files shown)


ischoegl:
Stopping here as ~any additional changes would depend on #747 and #748 (and #749).~ PR is ready for feedback/review.

decaluwe:
@ischoegl Could you provide a brief bit of discussion here on what the approach/philosophy is, and how it differs from what is done (or is this already discussed, elsewhere)? 

I see some changes in the `ElectrochemicalReaction` in `76d82a3`, for instance, that don't exactly make sense to me, but I'd like to get a better sense of your overall approach here, and what the current approach is, before commenting. This obviously also relates to #749, so I'd like to get a handle on your recommended approach, here.

ischoegl:
@decaluwe ... no problem. The main philosophy is to replace hard-coded reaction instantiation by an extendable `ReactionFactory` approach. None of this affects calculated outcomes. I.e. I'm just repackaging what's already there, with the difference that it becomes very simple to add additional reaction types.

* `GasKinetics` is one case in point: it is now fully flexible, i.e. someone can define a new reaction type externally and link it against cantera (`ctwrap` approach, where new C++ lambda functions are registered for an existing cantera factory object). Same could be done for `InterfaceKinetics` & co. (although the existing code is more convoluted / harder to untangle).
* Approach: 
  * Replace magic number referenced function calls by name-mapped C++ lambda functions calls (which are mainly used for setup; actual kinetics calculations involve templated functions, i.e. there is no change from the current implementation whatsoever)
   * I ended up using C++ lambda functions to recycle existing code directly (instead of the first commit that used new virtual member functions)
   * ~Ideally, the reaction list should be non-hierarchical (i.e. no nested definitions like `falloff`)~
*  ~The three things I came across are now #747, #748, and #749, where a minor update to the existing YAML syntax could simplify things quite a bit (i.e. making the implementation simpler to maintain/understand and also more future-proof regarding possible extensions)~

**PS:** any existing XML files for additional electrochemical reaction definitions would be beneficial. From what I can tell, there are at least three different behaviors implemented (and accessible via XML), some of which are currently not reachable via YAML.

https://github.com/Cantera/cantera/blob/719cb3d2b766707aad71b18f0f8ecbbf5420d620/include/cantera/kinetics/InterfaceKinetics.h#L525-L549

ischoegl:
After looking at this some more, I think this PR is ready for review as it stands.

Other comments that go beyond this PR:
* If the general direction is ok, a next step would be #751. (It warrants discussion beyond this PR.)
* I no longer believe that a completely flat structure (I left some comments to that extent) brings significant benefits. E.g. it's obviously ok to have a `falloff` type with three variants/flavors that are based on a reaction rate type (i.e. `Lindemann`, `Troe` and `SRI`), so #748 is moot (and closed).
* The same type + variant approach could be used for `electrochemical` and `butler-volmer` flavors down the line (see #749/#750)
* One question I have is: Is there a good reason to implement a `FalloffFactory`? (it appears to be highly specialized.) I can see the point of factory classes to break things out for Python access, but it may make more sense to do this at a more generic `RxnRatesFactory` level?

speth:
> One question I have is: Is there a good reason to implement a `FalloffFactory`? (it appears to be highly specialized.) I can see the point of factory classes to break things out for Python access, but it may make more sense to do this at a more generic `RxnRatesFactory` level?

Eliminating `FalloffFactory` sounds plausible to me. I doubt anyone is going to come up with an alternative falloff function without wanting more flexibility than this functional form provides. Making the reaction rate functions properly generic and more customizable is a bit more involved, given that there are at least perceived performance reasons for the less flexible way things are currently implemented using the `Rate1` template class.

ischoegl:
@speth ... thanks for the comment: this is what I thought. I am aware of speed benefits of templates used for reaction rates (the dynamic classes appear to be used almost exclusively for setup). 

Will most certainly not include anything else in this PR: I'd like to untangle the Python front end first (which is done), before anything else can get resolved in C++.

ischoegl:
@speth ... not sure whether you're considering this for 2.5, but the PR would be ready for review.

ischoegl:
@speth ... not sure whether this has legs or not (there may be other considerations that I’m not aware of). If this is a dead end, I’ll simply close this.

ischoegl:
Closing as this has been open and pending review for more than 6 months.

ischoegl:
A rebased branch remains on my fork on `ischoegl:reaction-factories`

ischoegl:
Reopened as #982 (cannot reopen here as I rebased the branch after closing the PR).

PR #750, opened 2019-11-12
Deprecate undocumented Butler-Volmer reaction options

### LINKED ISSUE #749: Discuss deprecating Butler-Volmer reaction types
In `Reaction.cpp`, there is a bit of processing of the XML reaction entry to extract the specific type of an electrochemical reaction: https://github.com/Cantera/cantera/blob/719cb3d2b766707aad71b18f0f8ecbbf5420d620/src/kinetics/Reaction.cpp#L859-L869

I believe a relevant XML reaction entry should look like

```xml
<reaction ... type="butlervolmer">
...
</reaction>
```

and similar for the other types in the C++ code. @decaluwe has suggested that the current method of specifying these reaction rate formalisms in the XML is somewhat tortured and should possibly be replaced, especially with the move to YAML input. There are currently no XML files in the Cantera repo that exercise the functionality present in the code snippet above, so there is no guarantee that they work as advertised.

I'm creating this issue to start the discussion about possibly deprecating and re-implementing (as necessary) these reaction rate types.

speth:
As mentioned in #749, I believe that there are implementation issues that prevent most if not all of it from being used at present (for both 2.4.0 and the current master branch) due to two mutually-exclusive conditions: (a) we don't currently allow reactant orders to be specified for reversible reactions and (b) Butler-Volmer reactions are required to be reversible and involve manipulating the reaction orders.

I think the easiest path forward to supporting these Butler-Volmer and other reaction types is to deprecate and remove these non-functional versions, and re-consider implementation after the next release.

### DIFF
--- a/include/cantera/kinetics/InterfaceKinetics.h
+++ b/include/cantera/kinetics/InterfaceKinetics.h
@@ -375,6 +375,7 @@ public:
      */
     int phaseStability(const size_t iphase) const;
 
+    //! @deprecated To be removed after Cantera 2.5.
     virtual void determineFwdOrdersBV(ElectrochemicalReaction& r, vector_fp& fwdFullorders);
 
 protected:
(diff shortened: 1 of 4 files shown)


decaluwe:
Agreed, particularly as there is a bit of 'philosophical' discussion to be hashed out about how we'd like to best implement and support this functionality. No need for those discussions to hold up the release of other functionality, at present.

PR #753, opened 2019-11-16
Enforce Python style naming convention for BinarySolutionTabulatedThermo

### LINKED ISSUE #658: BinarySolutionTabulatedThermo should use snake_case in Python
### Cantera version
2.5.0a2

Some of the variables and class names associated with the new `BinarySolutionTabulatedThermo` classes use camelCase in Python, rather than the standard TitleCase (for class names) or snake_case (for variable and function names). In my opinion, we should conform to usual Python standards here. These will need to be changed in ctml_writer.py and cti2yaml.py.

From a comment by @speth 

> If you want to fix the camelCasing of the new classes / arguments associate[d] with the addition of the `BinarySolutionTabulatedThermo` class, it will have to be done both here and in ctml_writer. I think we can make this change, since the class is new with this version of Cantera, but this should probably be brought up with @decaluwe.

@decaluwe Please also provide your thoughts 😃

ischoegl:

Fixes #658

- Use Python style naming convention for `BinarySolutionTabulatedThermo` definitions, i.e. replace `moleFraction` by `mole_fractions` in CTI input files specifying table entries
- raise errors clearly describing the issue


ischoegl:
PS: I'm not sure whether this is worth 'fixing', as it appears to be limited to CTI input, which also defines other non-compliant entries, e.g. `standardState` for species.

speth:
Given that input files using the current names are included with the paper about this model (https://doi.org/10.1016/j.electacta.2019.134797) and even shown in one of the figures, I would hesitate to change this without a deprecation warning, even though the model is just being introduced in Cantera 2.5.

And given the limited remaining lifespan of the CTI format, I'm not sure it's worth the hassle. The names in the YAML format are consistent with the rest of the names used there.

ischoegl:
I’m fully onboard with leaving the CTI as is, which means that #658 should be closed also.

PR #756, opened 2019-11-20
Reactor: ensure consistent vector lengths in C++ layer

ischoegl:

- Ensure consistent vector sizes, i.e. `Reactor::m_advancelimits.size()` is `Reactor::m_nv`
- Prevents hard to track segfaults when `m_advancelimits.size()` < `m_nv` in custom additions
- Make functions related to advance limits non-virtual

### DIFF
--- a/src/zeroD/Reactor.cpp
+++ b/src/zeroD/Reactor.cpp
@@ -103,7 +103,6 @@ void Reactor::initialize(doublereal t0)
         }
     }
     m_work.resize(maxnt);
-    m_advancelimits.resize(m_nv, -1.0);
 }
 
 size_t Reactor::nSensParams()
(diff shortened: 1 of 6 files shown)


speth (commented):
I see the issue with the keeping the length of `m_advancelimits` consistent across different derived classes (including user-created ones). I wanted to propose an alternative solution, which is to simply resize it to the correct size in the few locations where it is used (all of which are within the `Reactor` base class). See bd58afa72 for the implementation of this option.

I'm :+1: on making the related methods non-virtual.

ischoegl:
@speth ... resizing at a later time is a *very* good suggestion! Not having to resize in overloaded functions is great. Along those lines, I am suggesting to not resize `m_advancelimits` at all unless it is actually being used. (which allows for simple checks that reduce overhead).

src/zeroD/Reactor.cpp
```diff
@@ -445,20 +444,25 @@ void Reactor::setAdvanceLimits(const double *limits)
         throw CanteraError("getState",
                            "Error: reactor is empty.");
     }
-    for (size_t j = 0; j < m_nv; j++) {
-        m_advancelimits[j] = limits[j];
+    m_advancelimits.resize(m_nv, -1.0);
+    std::copy(limits, limits + m_nv, m_advancelimits.begin());
```
speth: In one line:
```c++
m_advancelimits.assign(limits, limits + m_nv);
```

src/zeroD/Reactor.cpp
```diff
@@ -445,20 +444,25 @@ void Reactor::setAdvanceLimits(const double *limits)
...
+
+    // resize to zero length if no limits are set
+    if (std::none_of(m_advancelimits.begin(), m_advancelimits.end(),
+                     [](double l){return l>0;})) {
```
speth: Can we have a variable name besides `l`? Looks way too much like `1` in many fonts. Applies to both instances of this.

src/zeroD/Reactor.cpp
```diff
@@ -445,20 +444,25 @@ void Reactor::setAdvanceLimits(const double *limits)
...
 bool Reactor::getAdvanceLimits(double *limits)
 {
-    bool has_limit = false;
-
-    for (size_t j = 0; j < m_nv; j++) {
-        limits[j] = m_advancelimits[j];
-        has_limit |= (limits[j] > 0.);
+    // m_advancelimit has zero length if no limits are set
+    bool has_limit = m_advancelimits.size();
+    if (has_limit) {
```
speth, line 459: Using `hasAdvanceLimits()` here would be self-documenting.

speth (changes requested):
This looks good to me. I only have a couple of minor suggestions. In addition to these, can you squash the third commit into the first, since it reverts a number of the changes made in the first commit?

ischoegl:
Done (and squashed).

speth (approved):

PR #759, opened 2019-11-28
1D continuation with Flame Point Control Technique from axbriones

youmengtian:
Fixes #541 

**Changes proposed in this pull request**

- One dimensional continuation techniques (one- and two-point) from axbriones

PR #541, opened 2018-06-01
1 d continuation

axbriones:
Please fill in the issue number this pull request is fixing:

Fixes #534 

Changes proposed in this pull request:
- One dimensional continuation techniques (one- and two-point) 
-
-

PR #766, opened 2019-12-05
1d point control

youmengtian:
Fixes #759

- rebased with new master branch

### DIFF
--- a/include/cantera/oneD/Sim1D.h
+++ b/include/cantera/oneD/Sim1D.h
@@ -131,6 +131,12 @@ public:
     //! Add node for fixed temperature point of freely propagating flame
     int setFixedTemperature(doublereal t);
 
+    //! Set the fuel internal boundary location and tempereture
+    void setFuelSideTemperature(doublereal tFuel);
+
+    //! Set the oxidizer side internal boundary location and temperature
+    void setOxidSideTemperature(doublereal tOxid);
+
     /**
      * Set grid refinement criteria. If dom >= 0, then the settings
      * apply only to the specified domain.  If dom < 0, the settings
--- a/include/cantera/oneD/Sim1D.h
+++ b/include/cantera/oneD/Sim1D.h
@@ -131,11 +131,11 @@ public:
     //! Add node for fixed temperature point of freely propagating flame
     int setFixedTemperature(doublereal t);
 
-    //! Set the fuel internal boundary location and tempereture
-    void setFuelSideTemperature(doublereal tFuel);
+    //! Set the fuel internal boundary location
+    void setFuelSideBoundary(doublereal tFuel);
 
-    //! Set the oxidizer side internal boundary location and temperature
-    void setOxidSideTemperature(doublereal tOxid);
+    //! Set the oxidizer side internal boundary location
+    void setOxidSideBoundary(doublereal tOxid);
 
     /**
      * Set grid refinement criteria. If dom >= 0, then the settings
--- a/include/cantera/oneD/StFlow.h
+++ b/include/cantera/oneD/StFlow.h
@@ -132,6 +132,34 @@ public:
         return m_fixedtemp[j];
     }
 
+    //! The current fuel side internal boundary temperature
+    doublereal fuelSideTemperature() const {
+        if ((m_onePntCtrl || m_twoPntCtrl) && (m_zFuel != Undef)) {
+            return m_tFuel;
+        }
+    }
+
+    //! Set the temperature in the fuel side internal boundary
+    void setFuelSideTemperature(doublereal tFuel) {
+        if ((m_onePntCtrl || m_twoPntCtrl) && (m_zFuel != Undef)) {
+            m_tFuel = tFuel;
+        }
+    }
+
+    //! The current oxidizer side internal boundary temperature
+    doublereal oxidSideTemperature() const {
+        if (m_twoPntCtrl && (m_zOxid != Undef)) {
+            return m_tOxid;
+        }
+    }
+
+    //! Set the temperature in the oxidizer side internal boundary
+    void setOxidSideTemperature(doublereal tOxid) {
+        if (m_twoPntCtrl && (m_zOxid != Undef)) {
+            m_tOxid = tOxid;
+        }
+    }
+
     // @}
 
     virtual std::string componentName(size_t n) const;
--- a/include/cantera/oneD/StFlow.h
+++ b/include/cantera/oneD/StFlow.h
@@ -23,8 +23,9 @@ const size_t c_offset_U = 0; // axial velocity
 const size_t c_offset_V = 1; // strain rate
 const size_t c_offset_T = 2; // temperature
 const size_t c_offset_L = 3; // (1/r)dP/dr
-const size_t c_offset_E = 4; // electric poisson's equation
-const size_t c_offset_Y = 5; // mass fractions
+const size_t c_offset_Uo = 4; // oxidizer axial velocity
+const size_t c_offset_E = 5; // electric poisson's equation
+const size_t c_offset_Y = 6; // mass fractions
 
 class Transport;
 
@@ -229,7 +230,7 @@ public:
     }
 
     virtual bool fixed_mdot() {
-        return (domainType() != cFreeFlow);
+        return (domainType() != cFreeFlow && !m_onePntCtrl && !m_twoPntCtrl);
     }
     void setViscosityFlag(bool dovisc) {
         m_dovisc = dovisc;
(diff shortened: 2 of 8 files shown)


speth:
Thanks for updating this. I developed a more minimal example (or at least, one that runs much faster) of a case where I'm seeing solver failures:
```py
import cantera as ct
import numpy as np
from collections import defaultdict

# Input parameters
p = ct.one_atm  # pressure
tin_f = 300.0  # fuel inlet temperature
tin_o = 300.0  # oxidizer inlet temperature
mdot_f = 0.2  # kg/m^2/s
width = 0.05 # Distance between inlets

comp_o = 'O2:0.21, AR:0.78'  # oxidizer composition
comp_f = 'H2:1, AR:1'  # fuel composition

loglevel = 1  # amount of diagnostic output (0 to 5)

gas.TPX = tin_f, p, comp_f
density_f = gas.density

gas.TPX = tin_o, p, comp_o
density_o = gas.density

# Set the state of the two inlets
f.fuel_inlet.mdot = mdot_f
f.fuel_inlet.X = comp_f
f.fuel_inlet.T = tin_f

f.oxidizer_inlet.mdot = mdot_f / density_f * density_o
f.oxidizer_inlet.X = comp_o
f.oxidizer_inlet.T = tin_o

f.set_refine_criteria(ratio=4, slope=0.2, curve=0.3, prune=0.04)

# Solve the problem
f.solve(loglevel, auto=True)

Tmax = max(f.T)
print(f"The maximum temperature is {Tmax:.4g} K")

Q = [np.trapz(f.heat_release_rate, f.grid)]
T_max = [Tmax]
a_mean = [f.strain_rate('mean')]

f.twoPointControl_enabled = True

while np.max(f.T) > 300:
    f.set_fuel_side_boundary(np.min(f.T) + 0.95*(np.max(f.T) - np.min(f.T)))
    f.set_oxid_side_boundary(np.min(f.T) + 0.95*(np.max(f.T) - np.min(f.T)))
    f.tFuel -= 10
    f.tOxid = f.tFuel

    print(f'Setting temperature points to T = {f.tFuel}')

    # Use callback function to collect data after each timestep
    # This data will be available after a failed solution attempt
    grids = {}
    TT = defaultdict(list)
    YY = defaultdict(list)
    def callback(x):
        N = len(f.grid)
        grids[N] = f.grid
        TT[N].append(f.T)
        YY[N].append(f.Y)
        return 0

    callback(0)
    f.set_time_step_callback(callback)

    try:
        f.solve(loglevel)
    except Exception as err:
        print(f'error at T = {f.tFuel}')
        print(err)
        break

    Tmax = np.max(f.T)
    q = np.trapz(f.heat_release_rate, f.grid)
    a = f.strain_rate('mean')
    print(f"The maximum temperature is {Tmax:.4g} K (a = {a:.4g}, Q = {q:.4g})")
    T_max.append(Tmax)
    a_mean.append(a)
    Q.append(q)
```
This is able to find a solution for a couple of iterations, but then fails in a case where it takes too many timesteps without reaching a steady-state solution. It appears to get stuck in a small-amplitude, high-frequency oscillation between two states. This can be seen by plotting the mean value of how much the temperature changes between each iteration (not really dT/dt, since the time step is not constant):
oscillation

It seems like it's possible to get past this behavior at least some of the time by tightening the transient relative  tolerances, e.g. with something like `f.flame.set_transient_tolerances(default=(1e-6, 1e-12))`, but this doesn't seem to be entirely reliable. When it works, I am able to trace out much of the unstable branch, so this is certainly quite promising. For example:
extinction2
Even with this adjustment though, I haven't been able to get solutions for temperatures less than ~1000 K, though.

youmengtian:
Thanks for your good test! @speth   As constant `dT` may result this situation when the gradient of maxT vs. strain rate is too large. So various `dT` should be used when the simulation fails, the previous solution should be restored firstly and `dT` changes smaller to fit this situations. I'll try to change your test case with unfixed `dT` with later reply.

With the point control technology, it is may possible but difficult to achieve the exact ignition point. So the maximum temperature couldn't less than ~1000K, because the `dT` is too large or it is too near to ignition point.

speth:
I think your diagnosis is correct for the second failure mode. You can see from the second figure I posted that the strain rate and heat release rate are changing quite a bit for each temperature increment as the maximum temperature approaches 1000 K. I guess what I'm not sure about is whether or not this method should be able to trace out the second turning point onto the lower stable branch.

However, the failure mode exhibited by the script I posted in its original form occurs much earlier, when Tmax is around 2590 K, and seems to occur relatively consistently regardless of the `dT` used. In addition, if you plot the evolution of the temperature and species profiles that are captured by the script above, you will see that the solution changes rapidly in the first time steps, adapting to the changed values of the temperature fixed points before settling into the oscillation shown. So I don't think it has much to do with the size of the temperature change.

youmengtian:
@speth Actually the f.set_fuel_side_boundary(T) may have a big effection. I have  not found a uniform T for various fuel-oxidizer flame. It may affect the failure mode I think.

wandadars:
Hey guys,

I don't know how I just saw this PR. I had been following an older one from a few years ago. What is the status of the PR? I've been looking to transition from FlameMaster to Cantera, and a method for generating 1D diffusion flamelet solutions along the unstable branch(arclength continuation) was a big sticking point because Cantera lacked this ability.

speth:
The status is what's posted in this thread: The current implementation has some numerical stability issues that need to be resolved before this can be merged. To the best of my knowledge, no one is actively working on this.

wandadars:
Would you be able to share your scripts that you used to make the plots shown above?

speth:
I think I generated those plots on the fly -- they're just plots based on the variables saved in the script posted above. For example, I think the upper plot is just `np.diff(T_max)`, the lower left plot is `T_max` versus `a_mean`, and the lower right is `Q` versus `a_mean`.

ischoegl:
Could anyone expand a little on the background of this PR? I am interested in unstable branches in a different context (premixed nonadiabatic), and had always been under the impression that those were unattainable with the current numeric scheme. (It may not be comparable though as mass flux is not prescribed for the premixed case.)

wandadars:
In the Python example that @speth posted, the fuel composition is listed as having both species H2 and AR as 100% of the composition. Is that a typo?  I also tried to just change the AR to N2 and the version of Cantera in this pull request threw an error:  

CanteraError thrown by Phase::setMoleFractionsByName:
Unknown species 'N2'

When I used:  comp_o = 'O2:0.21, N2:0.78'  # oxidizer composition

Not sure if that's the correct behavior or not.

wandadars:
@ischoegl   I think I may have found a reference for what is being implemented in this pull request.

See: http://www.yang.gatech.edu/publications/Journal/C&F%20(2014,%20Huo)%20-%20S%20curve.pdf

Looks like section 2.4 might where the good details are about the method. I wish I was capable enough to drag this PR over the finish line because I've been wanting to generate the unstable branch in Cantera for years now lol.

Reference information: https://www.sciencedirect.com/science/article/abs/pii/S0010218014001825

wandadars:
I was attempting to rebase this PR to origin/main, but it seems like some organizational changes have been made to the structure of the classes that are involved in this PR (for example, the getMeta method, I can't quite figure out how to handle the merge conflicts between this PR's version of that method and the current one). Does anyone have to knowledge to rebase this PR?

Edit: I figured it out

wandadars:
I have a version of this branch that is rebased onto the current Cantera main, and I have been able to partially reproduce @speth 's plots from the script provided.  One thing I noticed in comparing the work in this pull request versus the one that it is based on (#541) is that the original one and two point control coding also introduced an option for a strain equation that updated the residual vectors of the `c_offset_U` and `c_offset_V` equations.

I'm working on try to updated the original test_onedim.py test from the author's test case for marching along the 's-curve' of the solutions. It does appear that the structure of his test case and the one in this thread are different. 

I'm looking at the oscillation issue that @speth mentioned. I also notice that the approach in the script above is quite sensitive to the parameters for adjusting the internal boundary temperatures. From the reference paper provided in a comment above, it doesn't seem like such difficulties should be expected when using the two-point approach.

speth:
Hi @wandadars, I'm glad you were able to rebase this onto the current `main` branch. Assuming you would like to investigate the robustness of this feature and get this over the finish line, I'd suggest opening a new PR from your fork, with a link back to this one.

ischoegl:
I was about to suggest the same (i.e. open a new PR). Just make sure to preserve the commits of the original author as best as you can so that credit is assigned appropriately. If you need help with that, please let us know.

wandadars:
I put my foot in my mouth. I checked my cantera development directories and just realized that I had created a new branch and manually translated the changes from that pull request into a fresh branch. I did this because I was having issues compiling Cantera from the older version (with the rebase of this PR in it) that I had because of unstaged changes in ext/sundials and ext/yaml-cpp, and I couldn't figure out how to get rid of those in the `git status` . In this situation the original author's commits are lost.  If you guys could help me to resolve the unstaged changes (from a submodule I think?) that arrived when I finished the rebase, then I can add my adjustments on top of their commits.  Otherwise my pull request will have to be updated with  to the original author and also commits that credit the author's work. This isn't the best solution I would image, so hopefully I can get this other issue resolved and have this done the proper way.

This is what I see when I do a `git status`
Screenshot from 2023-06-21 15-52-16

speth:
The way to bring the submodules up to date is by running `git submodule update`, which should hopefully just work.

As far as the commit authorship goes, I wouldn't worry about this too much -- we can fix the history to credit the correct authors for any given commit (which needs to include crediting @axbriones for their work as well). I'd be happy to take care of this myself after you create the new PR.

wandadars:
> 

When I ran that command I saw:
```

fatal: remote error: upload-pack: not our ref 887af4374af2271db9310d31eaa9b5aeff49e829
fatal: the remote end hung up unexpectedly
Fetched in submodule path '../../ext/sundials', but it did not contain 887af4374af2271db9310d31eaa9b5aeff49e829. Direct fetching of that commit failed.
```
Edit:  I had to do a `git submodule sync`  and then the command worked.

speth:
Oh, this rebase is probably starting from before Sundials moved to GitHub and we were using an unofficial mirror repository. I think you should try `git submodule sync` first, then `git submodule update`. 🤞

ischoegl:
Superseded by #1510

PR #767, opened 2019-12-11
[ext/fmtlib] Bump fmtlib to 6.1.2

bryanwweber:

- Bumps fmtlib/fmt to version 6.1.2, which was released today (Dec. 11, 2019)
- I'm expecting the tests to fail, when I was testing for #722 I was using fmtlib 6.1.0 and the output in several places had dropped a decimal point (i.e., 1.0000e+02 became 10000e+02). The changes went away when I dropped back to fmtlib 5.0.0, so I surmise the cause is an updated function in fmtlib. This PR is mainly here to develop a fix for this problem.


speth:
I noticed this as well while trying out some other new features in `fmtlib` the other day. It appears to be a problem in 6.1.0 that was fixed in 6.1.1 (see https://github.com/fmtlib/fmt/commit/8bbe76af3a5427597185ad6207ee6e588d2eb0fd).

bryanwweber:
@speth Good to know, it's listed right in the release notes. Is this bump worth merging and if so, should we have SCons produce a useful error message for fmt 6.1.0?

speth:
6.1.0 lasted as the current release for all of 4 days. A specific check for that seems unnecessary.

bryanwweber:
And somehow I managed to end up with it from Hombrew... Point taken, is this PR worth merging or should it just be closed?

speth:
🤷‍♂ Your call. We should probably update this at some point, but I'd also like to make sure that we're testing with at least one older version. A reasonable target for this would be fmt 4.0.0, which is what is in Ubuntu 18.04.

bryanwweber:
Sure. I started setting up an Azure Pipelines instance to run more of these kinds of tests in parallel.

Issue #669, opened 2019-07-08
ck2yaml and cti2yaml require Python 3, breaking builds that use system Python 2


bryanwweber:
### Cantera version
2.5.0a3

### Operating System
Any

When building from source, either Python 2 or Python 3 should be able to be used to run SCons. The new `ck2yaml` and `cti2yaml` scripts depend on Python 3 features. The build script relies on at least `ck2yaml` to convert CHEMKIN input files to the YAML format, so builds that use Python 2 to run SCons and don't specify a `python_cmd` that is a Python 3 executable result in errors. One of them is documented on the [Google Group](https://groups.google.com/forum/#!topic/cantera-users/bZr5IZTUx5g):

```
scons: Building targets ...
"/usr/bin/python" interfaces/cython/cantera/ck2yaml.py --quiet --no-validate --input=data/inputs/gri30.inp --output=build/data/gri30.yaml --transport=data/transport/gri30_tran.dat
  File "interfaces/cython/cantera/ck2yaml.py", line 177
    def __init__(self, *, Tmin, Tmax, Tmid, low_coeffs, high_coeffs, note=''):
                        ^
SyntaxError: invalid syntax
scons: *** [build/data/gri30.yaml] Error 1
scons: building terminated because of errors.
```

### Steps to reproduce
1. Run SCons with a Python 2 sys.executable and do not specify a Python 3 executable in `python_cmd`

One possible solution to this is to check the YAML files into the repository rather than creating them during the build process. Another option is to make the scripts compatible with Python 2.

ischoegl:
Python 2 EOL is January 1st, 2020. What is cantera's perspective for Python 2 support beyond this date?

bryanwweber:
For the Python interface, 2.4.x will be the last release to support Python 2. We are attempting to keep the build script Python 2 compatible for the foreseeable future, so that, e.g., `apt install scons` and `scons build` will "just work". It's still under discussion, but I'm pretty sure the fix to this issue and #662 is going to be to check the YAML input files into the repository. Before we do that, I'm working on a CTML->YAML converter.

ischoegl:
@bryanwweber ... thanks for the clarification. From a user perspective, being forced to consistently use Python 3 may have some advantages: a meaningful error message would avoid quite a bit of confusion (e.g. user forum questions). From what I can gather, a recent scons version is all that is required?

bryanwweber:
[...]

Issue #339, opened 2016-05-13
Adding citation to the chem input file in the data folder


jiweiqi:
Using the chemical mechanism file in the system data of Cantera is very convenient. However, it seems that some of the mechanism files are not properly commented with the associated citation. I would suggest adding citation in all of the mechanism files, at least the chemical format mechanism files.

bryanwweber:
Please feel free to submit a pull request, if this is something you are interested in doing.

bryanwweber:
The `h2o2.cti` file is a subset of GRI-3.0: https://groups.google.com/d/msg/cantera-users/urFfNmZ0InI/KqTb7XCiFAAJ

bryanwweber:
@jiweiqi I don't think this issue is fixed yet

speth:
I think this was resolved as part of #768.

PR #768, opened 2019-12-11
[WIP] Commit yaml files

### LINKED ISSUE #662: The build script requires NumPy and ruamel_yaml for YAML conversions
### Cantera version
2.5.0a3 (with YAML PR merged)

### Operating System
Any

### Python/MATLAB version
Any

With the new YAML input files, we are converting CK input files to YAML with the ck2yaml.py script during the build process. This script requires NumPy and ruamel_yaml (or ruamel.yaml), even if the Python interface is not being built. These packages should be checked for during the build process and a useful error message should be shown to the user. The current behavior is a `ModuleNotFoundError` on the first attempted conversion.

bryanwweber:

Fixes #662, #669

- Commit YAML input files into the repository

I also plan to convert all the tests that are currently using CTI/XML input files to use the equivalent YAML version, except of course tests that are testing the behavior of one of those two formats, or the conversion of them.

It occurred to me that `ruamel.yaml` will still be required to run the tests, so the benefit of removing the build-time dependency is less than I hoped. I still think this is the appropriate resolution of #662 and #669 though.


ischoegl:
Some xml/cti occurrences are also replaced in #737 (addressing #736), which is pending resolution.

speth:
I would suggest limiting the replacement of CTI/XML with YAML in other tests. CTI and XML are still fully supported for Cantera 2.5, and I'd hate to have a bug creep in because we stopped testing these formats.

If you really want to include all of the converted input files, I think this would be a good opportunity to think about which of the existing ones we should actually keep. There are quite a few that are either not used in the test suite or barely used and that I don't think are all that useful to users. A few examples are `silicon.cti`, `silicon_carbide.cti`, `argon.cti`, and `nasa.cti` (just a combination of `nasa_gas` and `nasa_condensed`, as far as I can tell)

ischoegl:
> If you really want to include all of the converted input files, I think this would be a good opportunity to think about which of the existing ones we should actually keep. There are quite a few that are either not used in the test suite or barely used and that I don't think are all that useful to users.

What is the benefit of no longer supporting those input files (there is next to no development effort involved)? Also, it is hard to know what people are using. speaking for myself, I have been using argon and silicon carbide, mostly when I need a phase that just acts as a cheap reservoir for heat transfer purposes. It’s a minor issue, but I wanted to put it out there.

bryanwweber:
I think one primary reason is that we don't want to implicitly endorse one set of thermokinetics/transport data over another. By including the file with a Cantera distribution, we're kind of doing that. Obviously, there are several files we use in examples (`nDodecane_reitz` for one) that aren't generic, and GRI-3.0 seems like an historical exception, but I think in general we'd like to distribute fewer input files.

ischoegl:
This is understandable. However, the message that users should use 'modern' input files is already muddled as cantera de facto endorses `gri30`, `nDodecane_reitz`, and the `h2o2` portion of gri30 (for the latter, there are better h2o2 models around). Are there better ways to weaken this message than making cantera usage less convenient? An alternative would be to expand usage examples that have clear disclaimers regarding models used?

bryanwweber:
Personally, I see the distinction as related to the example. If the example is associated with a published work, then the example should more than likely use the same input file as the publication (I believe this is the case for the RK phase example that introduced the Reitz file).

As for GRI-3.0, I personally give that an historical exception. At this point, everyone *should* know not to use GRI-3.0 for any publication-quality work, so I don't mind including it. We need to have some sort of input file for our examples, and having a somewhat realistic (if outdated) file seems like a benefit over having a trivial input file (e.g., 1-2 species, 1 reaction, etc.).

The provenance of many of the other files is questionable, to me. To my understanding, many of them were invented for the purpose of the related example and haven't been peer-reviewed, and we should probably do a better job of clarifying where that's the case. One problem I have is that many of them were added before I started contributing, so I'm not sure what's good and what isn't.

ischoegl:
> As for GRI-3.0, I personally give that an historical exception. At this point, everyone should know not to use GRI-3.0 for any publication-quality work, so I don't mind including it. 

I am less optimistic in this regard, and believe that this needs to be pointed out especially to new users that may not be as familiar with the current state of affairs. 

> The provenance of many of the other files is questionable, to me. To my understanding, many of them were invented for the purpose of the related example and haven't been peer-reviewed, and we should probably do a better job of clarifying where that's the case. 

This is my understanding also.

I agree that some simple, semi-realistic example input files are needed (whether or not those are `argon` or `silicon-carbide` is beside the point). (EDIT: I guess I misread, and am actually advocating for more scaled down examples). I personally hope that the NSF project will result in improvements here.

One suggestion I'd have for 2.5 is to add fields to all cantera-supplied YAML (and CTI?)  files that include a disclaimer and/or clarify provenance (similar to the `deprecated` field that @speth suggested elsewhere). To that end, `ck2yaml` etc. could take an optional argument `--comment` that adds an entry so files do not have to be modified by hand.

ischoegl:
Adding/referencing an open issue that could be laid to rest here: #339

bryanwweber:
> I am less optimistic in this regard, and believe that this needs to be pointed out especially to new users that may not be as familiar with the current state of affairs.

I'm not sure where the best place to do this is - include a disclaimer in every example that uses GRI-3.0? I guess another question is, does it matter? There's no way anyone could publish a paper these days using GRI-3.0. Are there other outputs from using Cantera? (That question is clearly unanswerable 😄 ) Maybe someone uses it at an engineering firm, wouldn't they validate with experimental results for a design?

I guess, it seems like adding a disclaimer for all the examples that use GRI-3.0 seems like adding quite a bit of noise 🤷‍♂ 

> (EDIT: I guess I misread, and am actually advocating for more scaled down examples)

What do you mean by "scaled down"?

> I personally hope that the NSF project will result in improvements here.

As far as I know, we're not planning any work specifically on existing examples. I'm also not sure what you mean by improvements - improvements where? If you have any ideas here, feel free to put an issue at Cantera/improvements.

> One suggestion I'd have for 2.5 is to add fields to all cantera-supplied YAML (and CTI?) files that include a disclaimer and/or clarify provenance (similar to the deprecated field that @speth suggested elsewhere)

This is already done, see
https://github.com/Cantera/cantera/blob/6b9fe8af162c69852ab649ddd38a2c2bc88c11a6/data/gri30.yaml#L1-L6

> To that end, `ck2yaml` etc. could take an optional argument `--comment` that adds an entry so files do not have to be modified by hand.

I believe that `ck2yaml` currently reads any comments in the CK files and adds them to the `description` key. I'm not sure that adding a command line flag to read in a long text blob is a great UI, maybe it could be either a short message or a file name. PRs welcome!

> Adding/referencing an open issue that could be laid to rest here: #339

I am not planning to address #339 in this pull request.

ischoegl:
> I'm not sure that adding a command line flag to read in a long text blob is a great UI, maybe it could be either a short message or a file name

The `description` field is clearly different as it pertains to what the original authors of the mechanism had to say. All I was pointing to here is to add a short *official comment by cantera developers* to clarify what those input files represent and/or how they should be used (how this gets implemented is secondary). 

This suggestion would partially address #339, especially as the future fate of some of the distributed input files is being actively discussed at this point. I am writing this in an attempt to provide constructive suggestions. I am not planning a PR myself if this is just a shot in the dark; this is clearly a higher level decision. It’s not up to me ...

Ad scaled down Input / improvements: I don’t think that existing examples need work, but I was hoping for a more comprehensive suite of examples, some of which could be really basic ... :wink:

ischoegl:
@bryanwweber ... rather than adding a PR, I posted Cantera/enhancements#15

bryanwweber:
> The `description` field is clearly different as it pertains to what the original authors of the mechanism had to say.

I don't see the `description` field in this way. I think it is a more general field that can contain whatever we want. Thank you for adding the enhancement issue for references though, I think that will be valuable if we can come up with a convention there.

> This suggestion would partially address #339, especially as the future fate of some of the distributed input files is being actively discussed at this point.

Well, I ended up partially addressed #339 in https://github.com/Cantera/cantera/pull/768/commits/89157a481ad5bca90be1925d5adfb41beb7698db at least identifying when I didn't know the source of the data.

> Ad scaled down Input / improvements: I don’t think that existing examples need work, but I was hoping for a more comprehensive suite of examples, some of which could be really basic ... 😉

If you have something specific in mind (or even if not), feel free to open an issue on Cantera/enhancements. We do have the tutorials on the website which are pretty basic.

ischoegl:
> Well, I ended up partially addressed #339 in 89157a4 at least identifying when I didn't know the source of the data.

Looks good! Thanks for addressing.

bryanwweber:
@speth @ischoegl I added a possible `deprecated` implementation for the YAML files in https://github.com/Cantera/cantera/pull/768/commits/546223a85000ee38ebde95dc5260198a05cf1169 with two sample messages in `argon.yaml` and `gri30.yaml`.

I don't necessarily think it's worth adding to CTI given the impending deprecation of the whole format, plus I don't know where to add that 😂

ischoegl:
@bryanwweber ... the way I see this alternative:

Pro:

This implementation is short and won't require further code modifcations.

Con:

> @speth: The one advantage of approach [#737] would be that we wouldn't need to start of by introducing some deprecated phase definitions in gri30.yaml.

* The `gri30.yaml` file *cannot* be generated from source via `ck2yaml`, which is imho a serious drawback.
* The default loading behavior changes silently / without warning (which is inconsequential)

bryanwweber:
> @ischoegl: The `gri30.yaml` file cannot be generated from source via ck2yaml, which is imho a serious drawback.

Until we remove those phases after Cantera 2.5.0, i.e., for a limited amount of time. After Cantera 2.5.0, the file will be able to be generated directly from the CK input. As such, I don't see this as a drawback for this approach.

ischoegl:
I'm ok with this solution (as long as the `ck2yaml` generation is preserved). The warning message could be more explicit about how transport models are handled in general (which was a nice feature of the warning messages of the alternative approach). 

Edit: closed #737

bryanwweber:
> The warning message could be more explicit about how transport models are handled in general (which was a nice feature of the warning messages of the alternative approach).

Yes, these are just a message for the sake of seeing some output is generated. I'm interested in some feedback on the actual implementation for now, and I'll keep working on the messages. I was actually surprised at how easy this was to do, at least to the extent that the feature is available here.

ischoegl:
> I was actually surprised at how easy this was to do, at least to the extent that the feature is available here.

Yup. Seems like your solution hit a sweet spot - short and simple. Wouldn't have closed my PR otherwise! 

> I'm interested in some feedback on the actual implementation for now, and I'll keep working on the messages. 

Sure. I'm wondering whether it wouldn't make sense to move all warnings to `AnyMap`? You can probably raise warnings for all kinds of other mapped objects (thermo phases, species thermo, etc.) with very little effort using the same idea.

bryanwweber:
The reason I included it here is that it requires a YAML file with this field to actually function, and this is the only branch that has YAML files committed to the repo.

data/ohn.yaml
```diff
@@ -0,0 +1,18 @@
+description: This definition extracts the O/H/N submechanism from GRI-Mech 3.0
```
ischoegl: This is hand-copied from `nasa.cti` in 89157a4 (similar for many other files) ... would it make sense to add parsing of the header block and conversion to the `description` field to `ctml2yaml`? (this would also benefit users converting their own files)

data/air.yaml
```diff
@@ -0,0 +1,177 @@
+description: |-
```
ischoegl: Following up on discussion in commit 89157a4 (which is not displayed outside of the commit):

I believe it would be preferable to add the description to the initial source `air.inp` instead of adding it by hand after conversion. I.e. `ck2yaml` will generate this field automatically (I did as much in a small addition added to #777).

bryanwweber:
@ischoegl I feel very strongly that we should not be converting input files to YAML during the build process. To summarize my comments on #662 and #669, doing the conversion during build requires having a Python 3 interpreter with `ruamel.yaml` installed, and I don't think that's a requirement we should enforce at build time.

data/ohn.yaml
bryanwweber: Yes, I think it would be good to add this functionality... I think it should be possible in `ctml2yaml`, I believe comments are accessible in the tree. I'm not sure about `cti2yaml`, because that actually `exec`s the input file, so I'm not sure if comments are available. I believe a docstring would be available and could be added though...

ischoegl: From what I recall, CTI files typically have a commented `# ` header. The simplest way would be just to parse this block using `with open(...) as cti_file:` etc.?

ischoegl:
@bryanwweber ...

> I feel very strongly that we should not be converting input files to YAML during the build process. To summarize my comments on #662 and #669, doing the conversion during build requires having a Python 3 interpreter with ruamel.yaml installed, and I don't think that's a requirement we should enforce at build time.

I recall that conversation, but happen to disagree (in a nutshell: let Python 2 die, and there are more conservative approaches, like optionally loading input from a backup if `ruamel.yaml` isn't available). As some of the GRI3.0 issues have shown, it is good to be able to trace the provenance of files and compare sources whenever possible.

Not my decision though :wink:

**Edit:** Reposting @speth's comments on this issue from #736 
> Removing these files is an interesting suggestion. It would be consistent with the fact that we don't currently keep the original input files for other models (e.g. `nDodecane_Reitz.cti`). I'm wondering if that practice would have made it easier or more difficult to avoid any of the issues with `gri30.cti` that have cropped up. For the most recent one (the two changed rate constants), I think not having the original file in the repo would have made it almost impossible to understand how the error came about -- the header of the original input file with its different date was the necessary clue. I'm not sure I see the difference in having these original input files in the data directory versus `test/data`, given that they are only installed into the Python modules `test/data` directory in any case.

data/ohn.yaml
speth: `ctml2yaml` doesn't really stand a chance here of having a relevant description to copy, since the comment header in the CTI file would have been lost when running `ctml_writer`, and there are very few XML files with any header comment blocks, at least in the `cantera` repository.

I spent a little time trying to get access to the comments in a CTI file when writing `cti2yaml`, but couldn't find a way to get at them in a way that provided correspondence to the code itself -- they don't seem to be available even through the `ast` module. If we're only after the header comment to use for this field, then I think @ischoegl's suggestion is probably the best we can do.

speth:
So, I'm slowly coming around to agree that we should just commit files to the repo in the preferred Cantera-native format(s), given that most of the Chemkin-format files in the repository are not pristine originals in the first place. I would further suggest that we remove most of those after introducing the new YAML versions (except for any that we want to keep for testing purposes). I think that would relieve this tension among multiple, possibly inconsistent, versions of the same mechanism in the repo (well, at least once we can remove all of the CTI and XML files).

By committing the files directly, that gives us the option to modify them, which includes adding reference information in a standardized format (rather than just copying a text block as the "description") as well as allowing things like `deprecated` flags.

I would suggest leaving the info added by the conversion script (i.e. `generator`, `cantera-version`, etc.) so that we have some historical information in case there are ever any issues with the contents of these files. Given the history with `gri30.cti`, I would like to be prepared for that scenario.

ischoegl:
> So, I'm slowly coming around to agree that we should just commit files to the repo in the preferred Cantera-native format(s), given that most of the Chemkin-format files in the repository are not pristine originals in the first place.

I agree that *cantera-specific* `.inp` files can go. Beyond, one way or laying this controversy to rest would be to move `ck2yaml` conversions to examples (beyond retaining CK input for testing purposes). 

I still feel very strongly about providing clear guidance on how CK input is converted (presumably, this is what most users will have to do). This even has the potential of improving insights into the provenance of `h2o2.yaml`, `argon.yaml`, and `airNASA9.yaml`. Either way, I believe using the CK header as a default `description` is useful, and also hope to see the `references` field adopted. I.e. I hope for a programmatic way of generating *adequate* input files, which can be improved upon by hand-editing (I fully support the `deprecated` flag).

interfaces/cython/cantera/examples/onedim/flame_fixed_T.py
```diff
@@ -25,7 +27,7 @@
 # transport properties. It is created with two transport managers, to enable
 # switching from mixture-averaged to multicomponent transport on the last
 # solution.
-gas = ct.Solution('gri30.xml', 'gri30_mix')
+gas = ct.Solution("gri30.yaml", "gas", "Mix")
```
speth: The phase name in `gri30.yaml` is still `gri30` (which I'd support leaving as-is), so `gas` won't work here. I think providing the `transport_model` argument as a keyword is preferable, where it's needed. Also, if the default phase uses mixture-averaged transport, can't this just be `gas = ct.Solution("gri30.yaml")`?

interfaces/cython/cantera/examples/onedim/flamespeed_sensitivity.py
```diff
@@ -15,7 +16,7 @@
 width = 0.03  # m
 
 # IdealGasMix object used to compute mixture properties
-gas = ct.Solution('gri30.xml', 'gri30_mix')
+gas = ct.Solution("gri30.yaml", "gas", "Mix")
```
speth: Ditto here.

interfaces/matlab/toolbox/GRI30.m
```diff
@@ -12,30 +12,33 @@
...
 %     Instance of class :mat:func:`Solution`
 %
 
 if nargin == 0
-    s = Solution('gri30.xml', 'gri30');
+    s = Solution('gri30.yaml','gas');
 elseif nargin == 1
     if strcmp(tr, 'Mix')
-        s = Solution('gri30.xml', 'gri30_mix');
+        s = Solution('gri30.yaml', 'gas', 'Mix');
```
speth: Echoing previous comment about the phase named `gas`

samples/f77/isentropic.f
```diff
@@ -8,7 +8,7 @@ program isentropic
       double precision a(NPTS), dmach(NPTS), t(NPTS),
      $     ratio(NPTS)
 
-      call newIdealGasMix('gri30.cti','gri30','')
+      call newIdealGasMix("gri30.yaml",'gri30','')
```
speth: I'd suggest leaving the Fortran code to consistently use single quotes.

samples/matlab/flame2.m
```diff
@@ -34,7 +31,7 @@
 % This object will be used to evaluate all thermodynamic, kinetic,
 % and transport properties
 %
-gas = GRI30('Mix'); %IdealGasMix(rxnmech, transport);
+gas = Solution(rxnmech,'gas','None');
```
speth: Same comment as before on the `gas` phase name.

Also, what is the purpose of setting the transport model to `None` here, if it clearly needs to end up as something else later on?

src/thermo/ThermoFactory.cpp
```diff
@@ -451,6 +451,12 @@ void setupPhase(ThermoPhase& thermo, AnyMap& phaseNode, const AnyMap& rootNode)
         phaseNode["__file__"] = rootNode["__file__"];
     }
 
+    if (phaseNode.hasKey("deprecated")) {
+        string msg = phaseNode["deprecated"].asString();
+        string method = fmt::format("Phase deprecation: '{}'", phaseNode["name"].asString());
+        warn_deprecated(method, msg);
```
speth, line 458: I would suggest adding the file name to the warning message here, rather than in the text in the file itself. Also, given that the message will already start with `DeprecationWarning`, the message could be further simplified.

samples/matlab/lithium_ion_battery.m
```diff
@@ -50,10 +50,10 @@
...
 
 % Import all Cantera phases
-anode = Solution(inputCTI, 'anode');
-cathode = Solution(inputCTI, 'cathode');
-elde = Solution(inputCTI, 'electron');
-elyt = Solution(inputCTI, 'electrolyte');
+anode = Solution(inputCTI, 'anode', 'None');
+cathode = Solution(inputCTI, 'cathode', 'None');
+elde = Solution(inputCTI, 'electron', 'None');
+elyt = Solution(inputCTI, 'electrolyte', 'None');
```
speth: Given that there is no transport data in `lithium_ion_battery.yaml` (and no transport model specified in any of the phase definitions, it shouldn't be necessary to specify that the transport model is `None`.

samples/matlab/reactor1.m
```diff
@@ -4,7 +4,7 @@ function reactor1(g)
 %    >>>>  For a simpler way to carry out a constant-pressure simulation,
 %    see example reactor3.m <<<<<
 %
-%    This example illustrates how to use class 'Reactor' for
+%    This example illustrates how to use class '`Reactor' for
```
speth: I assume this is a typo?

data/nasa_condensed.yaml
```diff
@@ -0,0 +1,3466 @@
+description: |-
+  Condensed phases species. Source unknown, data originally added by
+  Dave Goodwin for a very early version of Cantera.
```
speth: These are the condensed-phase species from `nasathermo.dat`. Similarly, `nasa_gas` is all of the gas phase species. I would like to suggest deprecating `nasa.yaml` in favor of these two files.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -344,7 +344,9 @@ def to_yaml(cls, representer, node):
         if node.transport:
             out['transport'] = node.transport
         if node.comment:
-            out['note'] = node.comment
+            comment = node.comment.split("\n")
+            comment = " ".join([c.strip() for c in comment])
```
speth:
```suggestion
            comment = " ".join(c.strip() for c in comment)
```

interfaces/cython/cantera/examples/multiphase/adiabatic.py
```diff
@@ -51,15 +53,15 @@
 
     tad[i] = mix.T
     print('At phi = {0:12.4g}, Tad = {1:12.4g}'.format(phi[i], tad[i]))
-    xeq[:,i] = mix.species_moles
+    xeq[:, i] = mix.species_moles
 
 # write output CSV file for importing into Excel
 csv_file = 'adiabatic.csv'
-with open(csv_file, 'w') as outfile:
+with open(csv_file, 'w', newline='') as outfile:
```
speth, line 60: What's the purpose of the `newline=''` argument here?

interfaces/cython/cantera/examples/onedim/diffusion_flame.py
```diff
@@ -22,7 +23,7 @@
 
 # Create the gas object used to evaluate all thermodynamic, kinetic, and
 # transport properties.
-gas = ct.Solution('gri30.xml', 'gri30_mix')
+gas = ct.Solution("gri30.yaml", "gas", transport_model="Mix")
```
speth: Shouldn't this just be 
```suggestion
gas = ct.Solution("gri30.yaml")
```

interfaces/cython/cantera/examples/onedim/premixed_counterflow_twin_flame.py
```diff
@@ -73,30 +79,31 @@ def solveOpposedFlame(oppFlame, massFlux=0.12, loglevel=1,
 
     return np.max(oppFlame.T), K, strainRatePoint
 
+
 # Select the reaction mechanism
-gas = ct.Solution('gri30.cti')
+gas = ct.Solution("gri30.yaml", transport_model='Mix')
```
speth:
```suggestion
gas = ct.Solution("gri30.yaml")
```

interfaces/cython/cantera/examples/reactors/NonIdealShockTube.py
```diff
@@ -90,33 +94,39 @@ def ignitionDelay(states, species):
...
 
 # Set the state of the gas object:
 ideal_gas.TP = reactorTemperature, reactorPressure
 
 # Define the fuel, oxidizer and set the stoichiometry:
-ideal_gas.set_equivalence_ratio(phi=1.0, fuel='c12h26',
-                                oxidizer={'o2':1.0, 'n2':3.76})
+ideal_gas.set_equivalence_ratio(
+    phi=1.0, fuel='c12h26', oxidizer={'o2': 1.0, 'n2': 3.76}
+)
```
speth: I don't see how this is an improvement over the existing formatting.

interfaces/cython/cantera/examples/reactors/NonIdealShockTube.py
```diff
@@ -127,22 +137,25 @@ def ignitionDelay(states, species):
 
 t = 0
 
-counter = 1;
+counter = 1
 while(t < estimatedIgnitionDelayTime):
     t = reactorNetwork.step()
-    if (counter%20 == 0):
+    if (counter % 20 == 0):
```
speth: If you're fixing up the formatting, the first thing I'd get rid of here would be the superfluous parentheses.

interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -77,20 +77,25 @@
...
+
 #####################################################################
 # Set up Reactor Network
 #####################################################################
 
 # load reaction mechanism
-gas = ct.Solution(reaction_mechanism, phase_name)
+gas = ct.Solution(
+    reaction_mechanism, phase_name, transport_model=None
+)
```
speth: Why isn't this on one line?

interfaces/cython/cantera/examples/reactors/ic_engine.py
```diff
@@ -106,7 +111,8 @@ def piston_speed(t):
 inlet_delta = np.mod(inlet_close - inlet_open, 4 * np.pi)
 inlet_valve.valve_coeff = inlet_valve_coeff
 inlet_valve.set_time_function(
-    lambda t: np.mod(crank_angle(t) - inlet_open, 4 * np.pi) < inlet_delta)
+    lambda t: np.mod(crank_angle(t) - inlet_open, 4 * np.pi) < inlet_delta
+)
```
speth: Some of the changes like this one seem unnecessary, when what was here was already PEP8-compliant (at least, as far as I can tell). Otherwise, if we're going to change our expectations for Python code formatting, then we need to document that somewhere if we don't want to constantly be reformatting code.

interfaces/cython/cantera/examples/reactors/reactor2.py
```diff
@@ -1,60 +1,58 @@
...
+with 1000 K argon at 20 atm, and the other with a combustible 500 K
+methane/air mixture at 0.1 atm (phi = 1.1). At t = 0 the piston is
+released and begins to move due to the large pressure difference,
+compressing and heating the methane/air mixture, which eventually
+explodes. At the same time, the argon cools as it expands. The piston
+allows heat transfer between the reactors and some heat is lost through
+the outer cylinder walls to the environment.
+
+Note that this simulation, being zero-dimensional, takes no account of shock wave
+propagation. It is somewhat artifical, but nevertheless instructive.
```
speth: I really can't tell what column width is being enforced here. What was here seemed to be wrapped at 79 or 80, but now the first paragraph seems to be wrapped at 72 and the second one at something greater than 80.

interfaces/cython/cantera/examples/reactors/reactor1.py
```diff
@@ -24,12 +23,17 @@
 t_end = 100 * dt_max
 states = ct.SolutionArray(gas, extra=['t'])
 
-print('{:10s} {:10s} {:10s} {:14s}'.format('t [s]','T [K]','P [Pa]','u [J/kg]'))
+print('{:10s} {:10s} {:10s} {:14s}'.format(
+    't [s]', 'T [K]', 'P [Pa]', 'u [J/kg]'
+))
```
speth: So is two closing delimiters on the same line OK or not?

interfaces/cython/cantera/examples/surface_chemistry/sofc.py
```diff
@@ -84,20 +84,22 @@ def NewtonSolver(f, xstart, C=0.0):
...
-                                              ['gas', 'metal', 'oxide_bulk',])
+gas_a, anode_bulk, oxide_a = ct.import_phases(
+    "sofc.yaml", ['gas', 'metal', 'oxide_bulk'],
+)
 
 # import the surfaces on the anode side
-anode_surf = ct.Interface('sofc.cti', 'metal_surface', [gas_a])
-oxide_surf_a = ct.Interface('sofc.cti', 'oxide_surface', [gas_a, oxide_a])
+anode_surf = ct.Interface("sofc.yaml", 'metal_surface', [gas_a])
+oxide_surf_a = ct.Interface("sofc.yaml", 'oxide_surface', [gas_a, oxide_a])
```
speth: Mixing and matching `"` and `'` on the same line of code doesn't seem like an improvement in style consistency.

src/base/AnyMap.cpp
```diff
@@ -867,6 +867,12 @@ AnyMap AnyMap::fromYamlFile(const std::string& name,
     }
     cache_item.first["__file__"] = fullName;
 
+    if (cache_item.first.hasKey("deprecated")){
```
speth:
```suggestion
    if (cache_item.first.hasKey("deprecated")) {
```

data/graphite.yaml
```diff
@@ -0,0 +1,27 @@
+description: |-
+  Thermodynamic properties of graphite. Source unknown, this data was
+  included in the first available versions of Cantera by Dave Goodwin.
```
speth: The NASA polynomial coefficients are the same as those from `nasathermo.dat`.

data/gri30_highT.yaml
```diff
@@ -0,0 +1,1746 @@
+generator: cti2yaml
+cantera-version: 2.5.0a3
+date: Wed, 11 Dec 2019 16:59:05 -0500
+input-files: [gri30_highT.cti]
```
speth, line 17: I think the `gri30_mix` and `gri30_multi` phases in this file should be deprecated for the same reasons as those in `gri30.yaml`.

interfaces/cython/cantera/examples/reactors/fuel_injection.py
```diff
@@ -5,14 +5,18 @@
...
 
 import numpy as np
 import matplotlib.pyplot as plt
 import cantera as ct
 
 # Use a reduced n-dodecane mechanism with PAH formation pathways
-gas = ct.Solution('nDodecane_Reitz.cti', 'nDodecane_IG')
+gas = ct.Solution(
+    "nDodecane_Reitz.yaml", "nDodecane_IG", transport_model=None
+)
```
speth: This file doesn't include a transport model, so specifying `transport_model=None` is unnecessary.

speth (changes requested):
I didn't mark all of these instances but many of the examples are currently broken because the explicitly specify `gas` as the phase name, but the first phase name is `gri30`.

I wonder whether overriding the default transport model in almost every example is what we want to encourage users to do. It's a good capability to have some examples of, but I don't think the impact of setting `transport_model=None` is that significant except for fairly large models (much larger than GRI 3.0).

I appreciate the effort put into making `ck2yaml` callable from C++ and Matlab, but this is something that I would really like to avoid introducing. I see the implicit conversion of CK input files to CTI in the Matlab toolbox as a wart that I'd prefer to get rid of, rather than a feature to retain as we switch to the YAML format. We don't provide this capability in any of the other language interfaces. Indirectly calling Python to run conversions has been a persistent source of trouble for users (think about all of the Users' Group issues related to the setting of `PYTHON_CMD`), and I would really like to see it go away with the XML and CTI formats.

I'm a little ambivalent about some of the formatting changes in the Python files, where this PR appears to be imposing conventions that are not part of PEP8 or anything that has been clearly articulated.

One meta-comment: It would have been helpful to break this up into more than one pull request. All of the changes to the examples could have been done separately, especially those relating to formatting and other issues that are unrelated to the primary purpose of this PR. This would have made reviewing this a little more managable. I know I'm also guilty of some mega-PRs, so feel free to take this with a grain of salt.

src/thermo/ThermoFactory.cpp
ischoegl: Up to this point, deprecation warnings have been used for deprecated methods only (the output should also have consistent formatting which is similar to errors). After #741, there is now `warn_user`

interfaces/cython/cantera/examples/onedim/flame_fixed_T.py
bryanwweber: > The phase name in gri30.yaml is still gri30 (which I'd support leaving as-is), so gas won't work here

Yup, brain fart on my part 🤦‍♂️

> Also, if the default phase uses mixture-averaged transport, can't this just be gas = ct.Solution("gri30.yaml")?

My thought was to be explicit, both to show that the option was available (and I agree that the keyword should be included) as well as to avoid any issues in the future if the default transport model should change again.

samples/matlab/flame2.m
bryanwweber: > Also, what is the purpose of setting the transport model to None here, if it clearly needs to end up as something else later on?

A typo 🤦‍♂️

interfaces/cython/cantera/examples/reactors/ic_engine.py
bryanwweber: I don't think the original version was PEP8 compliant. My interpretation of it is that the closing paren can be on the same line as a multi-line construct when the text is aligned with the first line:
```
foo = func(var1, var2,
           var3, var4)
```
but if the construct starts on the next line, the closing paren should be in its own line:
```
foo = func(
    var1, var2,
    var3, var4,
)
Or
foo = func(
    var1, var2,
    var3, var4,
    )
```
There's nothing explicit about this in the PEP, but there aren't any examples in the whitespace section of the PEP that use the format that was originally on this line.

Aside from appealing to the PEP, I personally find the new version much easier to read, because the closing paren is much easier to find. The original way I find it easy to read past that paren and interpret some following code as still belonging to the function.

speth:
> I agree that _cantera-specific_ `.inp` files can go. 

So that's basically everything except GRI 3.0 and the McBride/Gordon thermo, right?

> Beyond, one way or laying this controversy to rest would be to move `ck2yaml` conversions to examples (beyond retaining CK input for testing purposes).

Can you elaborate on this suggestion? I'm not sure I understand what you mean.

> Either way, I believe using the CK header as a default `description` is useful, and also hope to see the `references` field adopted. I.e. I hope for a programmatic way of generating _adequate_ input files, which can be improved upon by hand-editing (I fully support the `deprecated` flag).

I'm fully supportive of introducing some version of a structured `references` field to the input files, in addition to pulling the `description` from the header of the source input file. I think it will be easier to work on that in a subsequent, less monstrously large PR, where we can actually see just the changes to the YAML headers instead of the rest of the newly-added YAML files.

ischoegl:
> So that's basically everything except GRI 3.0 and the McBride/Gordon thermo, right?

correct, with the caveat below

> Can you elaborate on this suggestion? I'm not sure I understand what you mean.

*all* compile-time conversions could be eliminated (`gri30.yaml` needs hand-editing at least for 2.5), as long as `ck2yaml` is used/documented *somewhere*, i.e. unit tests are necessary, and an example folder could improve documentation. For example, I could see how a simple script re-creates `gri30.yaml`, `h2o2.yaml`, `argon.yaml`, and `airNASA9.yaml`. For other code, I have done this sort of documentation in form of a Jupyter file (where console commands can be run and documented quite easily)

> I'm fully supportive of introducing some version of a structured `references` field to the input files, in addition to pulling the `description` from the header of the source input file. I think it will be easier to work on that in a subsequent, less monstrously large PR, where we can actually see just the changes to the YAML headers instead of the rest of the newly-added YAML files.

already done, see #777

speth:
> _all_ compile-time conversions could be eliminated

I think that's Bryan's primary point here.

> as long as `ck2yaml` is used/documented _somewhere_, i.e. unit tests are necessary, and an example folder could improve documentation.

There is a fairly comprehensive test suite in `test_convert.py`. The documentation of `ck2cti` at https://testing.cantera.org/tutorials/ck2cti-tutorial.html is almost identical to what we need to say about `ck2yaml` (I realized that I missed this in https://github.com/Cantera/cantera-website/pull/91). I'm not sure where else I would think to put examples besides there. They're more generic than any one language interface, and in most cases all a conversion should consist of is a single command line statement. 

> already done, see #777
 
Yes, I've seen #777, I just want to wrap this PR up before doing anything further in this area.

ischoegl:
>  I just want to wrap this PR up before doing anything further in this area.

Agreed. As long as the original CK input files that are needed for the unit tests (i.e. gri30 plus airNASA9) remain available, things can be pieced back together as needed. I.e. I’m supporting the direction this is going and have no further comments.

PS: ad examples, I will likely suggest something in [Cantera/cantera-jupyter](https://github.com/cantera/cantera-jupyter)

src/thermo/ThermoFactory.cpp
bryanwweber: @ischoegl IMO using `warn_deprecated` is fine here. It gives a nice control over whether the warnings should be fatal (e.g., making sure we don't use deprecated files/phases in the test suite) and gives users the ability to silence them.

@speth I added the filename and the phase name. If only the filename is added, then additional warnings from other phases in the same file would be suppressed.

samples/matlab/reactor1.m
bryanwweber: Yep. Thanks!

interfaces/cython/cantera/examples/multiphase/adiabatic.py
bryanwweber: This is recommended in the [Python docs](https://docs.python.org/3/library/csv.html#id3). I believe the reason is because the CSV writer will handle newlines itself and this prevents newlines in the values from being interpreted incorrectly.

data/gri30_highT.yaml
bryanwweber: Done.

data/ohn.yaml
bryanwweber: @speth I think it's probably worth seeing how hard this would be to add to `ctml2yaml`, in case anyone has hand-edited XML files that include comments, at least for a top-level comment. Not a high priority, I'll add an issue at Cantera/enhancements eventually.

interfaces/cython/cantera/examples/onedim/diffusion_flame.py
bryanwweber: I removed the phase specification here. I addressed the `transport_model` argument in another comment.

samples/matlab/lithium_ion_battery.m
bryanwweber: As I've noted elsewhere, this is future-proofing in case a transport model is added in that file.

interfaces/cython/cantera/examples/reactors/NonIdealShockTube.py
bryanwweber: In my opinion, this is much easier to read than the previous formatting (for reference, the previous formatting was:
```python
ideal_gas.set_equivalence_ratio(phi=1.0, fuel='c12h26',
                                oxidizer={'o2':1.0, 'n2':3.76})
```
)
The main reason I find it easier to read is that all of the arguments end up on the same line, so you can just read them across. This does separate things from the function name being called, so I can see the other side. @ischoegl Do you have an opinion here?

interfaces/cython/cantera/examples/reactors/reactor1.py
bryanwweber: My interpretation is that since both delimiters open on the same line, they should close on the same line. How would you format this?

interfaces/cython/cantera/examples/surface_chemistry/sofc.py
bryanwweber: Agreed, I reverted all of these `'` to `"` changes back to just `'` since that seems to be the style in the examples.

bryanwweber:
I haven't committed and pushed the changes I've made yet, but I want to respond to the comments.

> I didn't mark all of these instances but many of the examples are currently broken because the explicitly specify `gas` as the phase name, but the first phase name is `gri30`

I just ended up removing all of the phase names. See the next response, though.

> I wonder whether overriding the default transport model in almost every example is what we want to encourage users to do. It's a good capability to have some examples of, but I don't think the impact of setting `transport_model=None` is that significant except for fairly large models (much larger than GRI 3.0).

I had 2 purposes to doing this. One was, as you note, to demonstrate the presence of this argument. The other was to be explicit about the transport model that's loaded to guard against any future changes in the phases and to encourage users to do the same thing. That said, I'm not sure how to strike a balance where we don't imply that this argument is *required* (and likewise for the phase name argument, balancing including it vs. not). Suggestions welcome.

> but this is something that I would really like to avoid introducing. I see the implicit conversion of CK input files to CTI in the Matlab toolbox as a wart that I'd prefer to get rid of, rather than a feature to retain as we switch to the YAML format. We don't provide this capability in any of the other language interfaces. 

I think the only implicit place this happens is in `IdealGasMix.m`, which I would also like to get rid of. ~IMO, most of the problem with the previous situation was that `PYTHON_CMD` needed to be set for every call to `Solution.m` to convert the CTI files, whereas the YAML conversion only needs to be done once. Users have to convert their input files somehow, and this would need to use Python in any case, so whether they do it from the terminal or within MATLAB seems more-or-less immaterial to me, with the exception that the PATH is easier to control from the terminal.~ Well, rethought that and I agree it's best to have a consistent set of command line instructions that works for everyone. I'll remove this commit, it was mostly copy-paste anyways.

FWIW, we do provide this capability from the Python interface (`python -m cantera.ck2yaml` or `import cantera.ck2yaml`), but I suppose that's mostly an accident of the fact that the conversion scripts are written in Python.

> I'm a little ambivalent about some of the formatting changes in the Python files, where this PR appears to be imposing conventions that are not part of PEP8 or anything that has been clearly articulated. ... It would have been helpful to break this up into more than one pull request.

Yeah, this got away from me a little bit. I figured since I was touching the examples anyways to change the input files, I might as well change the formatting and, well, things snowballed from there. I'll split those changes into a separate commit that can be put onto a new branch and reviewed separately, if you want.

As far as the CK input files, I think those changes should be done in a separate PR, perhaps one that updates the tests as appropriate.

Last, on the deprecation warnings for files/phases... I'm happy to improve the warning messages for `gri30*` and `argon.yaml` here, but I think deprecating other files/phases should be done/discussed in another PR to help avoid even more scope creep on this PR.

src/thermo/ThermoFactory.cpp
ischoegl: @bryanwweber ... no problem, just wanted to point this out. The other observation remains: the `method` string in its present form is not consistent with usage elsewhere (a consistent method string would be simply `"setupPhase"`, with the rest of your message being merged with `msg`)

bryanwweber: @ischoegl yes, that would be consistent with the documentation and usage elsewhere but if the `method` is not unique, the message will be suppressed when it shouldn't be, e.g., multiple deprecated phases in the same file. Is there a particular solution to this you had in mind? I could update the documentation or the argument name

ischoegl: I don’t have a particular solution, but would maintain the current convention of error/warning signatures whatever the final solution may be. To that end, I’d still suggest `warn_user`, as the deprecation involves user input. Presumably, the code base does not change, so I’m not sure that this situation warrants a redefinition of `warn_deprecated`. Just 2 cents, and I won’t press this further.

interfaces/cython/cantera/examples/reactors/NonIdealShockTube.py
ischoegl: @bryanwweber ... since you’re asking: I tend to write my own code in the style of the pre-existing format, as I personally don’t like single parentheses on a line (it simply doesn’t look like python code to me). At the same time, this is form and not function so it ultimately makes little difference: both are perfectly readable.

data/graphite.yaml
```diff
@@ -0,0 +1,38 @@
+description: |-
+  This is the NASA thermodynamic database, which is available for download
+  from http://www.galcit.caltech.edu/EDL/public/thermo.html. The original
```
speth: The current URL for this is https://shepherd.caltech.edu/EDL/PublicResources/sdt/thermo.html.

I think the description should start with saying that the file is for graphite, with data from blah blah blah, rather than just a duplicate of the header from `nasa.yaml`. Similarly for all of the other files that are just subsets of `nasa.yaml`.

interfaces/cython/cantera/examples/multiphase/adiabatic.py
speth: So this should only matter if you have newline characters in your input, then, right? Good luck to anyone who manages to get newlines into their species names...

interfaces/cython/cantera/examples/reactors/ic_engine.py
speth: The three "yes"examples under [Indentation](https://www.python.org/dev/peps/pep-0008/#indentation) in PEP8 all have the close parenthesis on the same line as the last argument. The style you are advocating is shown in the "optional" section. So I think both versions are PEP8 compliant.

interfaces/cython/cantera/examples/reactors/reactor1.py
speth: Well, this is formatted differently from your formatting for the `print` call in the loop, which is almost identical in terms of the form.

samples/matlab/lithium_ion_battery.m
```diff
@@ -28,7 +28,7 @@
 P = oneatm; % [Pa] Pressure
 
 % Cell properties
-inputCTI = 'lithium_ion_battery.cti'; % Cantera input file name
+inputYAML = 'lithium_ion_battery.yaml'; % Cantera input file name
```
speth: Better to just call it `inputFile`?

src/thermo/ThermoFactory.cpp
speth: I think `warn_deprecated` is still the right function to call here, since we are warning about a deprecation. It's just that the argument name `method` is not quite generic enough -- it should just be "the name of the deprecated thing", so in this case, "*foo*.yaml/*bar*" makes sense.

For `warn_user`, the purpose of the `method` argument is to tell you where the warning is coming from, but in this case that's not really what's important.

speth (commented):
> > I wonder whether overriding the default transport model in almost every example is what we want to encourage users to do. It's a good capability to have some examples of, but I don't think the impact of setting `transport_model=None` is that significant except for fairly large models (much larger than GRI 3.0).
> 
> I had 2 purposes to doing this. One was, as you note, to demonstrate the presence of this argument. The other was to be explicit about the transport model that's loaded to guard against any future changes in the phases and to encourage users to do the same thing. That said, I'm not sure how to strike a balance where we don't imply that this argument is _required_ (and likewise for the phase name argument, balancing including it vs. not). Suggestions welcome.

If we always specify the transport model where we call `Solution`, what's the point of specifying it in the input file, then? Why the transport model and not the thermo / kinetics models? And what would be the downside if a phase that currently does not specify a transport model gains one in a future Cantera version, besides a small increase in load time?

Changes in the models specified in the input file are pretty rare. Even with all of the changes coming in 2.5.0, I think the only case where the transport models are changing is the default phase in `gri30.yaml` and `gri30_highT.yaml`, which are arguably new input files anyway.

> As far as the CK input files, I think those changes should be done in a separate PR, perhaps one that updates the tests as appropriate.
> 
> Last, on the deprecation warnings for files/phases... I'm happy to improve the warning messages for `gri30*` and `argon.yaml` here, but I think deprecating other files/phases should be done/discussed in another PR to help avoid even more scope creep on this PR.

Agreed. Let's get this finished without increasing its size any further -- My browser is choking on this PR as it is.

interfaces/cython/cantera/examples/multiphase/adiabatic.py
bryanwweber: I suppose... On the other hand, every example in the Python docs uses this form, so I'm inclined to include it here. Do you have a strong opinion that it should be removed?

speth: No, it can stay. I'm just being cranky about having to figure out the details of the Python `csv` module in order to review a PR that is supposedly about YAML files.

samples/matlab/lithium_ion_battery.m
bryanwweber: Removed these specifications.

interfaces/cython/cantera/examples/multiphase/adiabatic.py
bryanwweber: Yeah, that's fair. In the future, feel free to tell me to split the PR into several

interfaces/cython/cantera/examples/reactors/NonIdealShockTube.py
bryanwweber: I reverted the change that resulted in the closing paren on its own line, but retained adding the spaces after the colons in the dictionary.

interfaces/cython/cantera/examples/reactors/ic_engine.py
bryanwweber: I reverted this change.

interfaces/cython/cantera/examples/reactors/reactor2.py
bryanwweber: I reverted this change so it is easier to see the text that's actually updated. I'm not sure why the line lengths changed as well... I must have autoformatted it 🤷‍♂

interfaces/cython/cantera/examples/reactors/reactor1.py
bryanwweber: I made the two calls to `print` more consistent.

interfaces/cython/cantera/examples/reactors/fuel_injection.py
bryanwweber: Removed this, and in other places as well.

bryanwweber:
@speth I think this is finished now (finally). I reverted most of the controversial formatting changes, keeping mostly the ones adding spaces and the version requirements to the docstring.

data/argon.yaml
```diff
@@ -0,0 +1,40 @@
+generator: cti2yaml
+cantera-version: 2.5.0a3
+date: Wed, 11 Dec 2019 16:59:05 -0500
+input-files: [argon.cti]
+description: >-
+  Pure argon phase. Thermo and transport properties are taken
+  from GRI-Mech 3.0.
```
speth: For these files where the description is added manually, should we put the `description` field first, for consistency with the files generated by `ck2yaml`? I think it's more important for the user than the other metadata.

data/lithium_ion_battery.yaml
```diff
@@ -0,0 +1,319 @@
...
+  Cantera input file for an LCO/graphite lithium-ion battery
+
+  This file includes a full set of thermodynamic and kinetic parameters of a
+  lithium-ion battery, in particular:
+  - Active materials: LiCoO2 (LCO) and LiC6 (graphite)
+  - Organic electrolyte: EC/PC with 1M LiPF6
+  - Interfaces: LCO/electrolyte and LiC6/electrolyte
+  - Charge-transfer reactions at the two interfaces
+
+  We use Butler-Volmer kinetics by setting rate_coeff_type = "exchangecurrentdensity".
```
speth:
```suggestion
  We use Butler-Volmer kinetics by setting `exchange-current-density-formulation` to `true`.
```

test/general/test_containers.cpp
```diff
@@ -205,3 +205,9 @@ TEST(AnyMap, loadYaml)
     EXPECT_EQ(coeffs[0].size(), (size_t) 7);
     EXPECT_DOUBLE_EQ(coeffs[1][2], -8.280690600E-07);
 }
+
+TEST(AnyMap, loadDeprecatedYaml)
+{
+    make_deprecation_warnings_fatal();
```
speth: This is called automatically at the start of the test suite (see `main` as defined in `test/general/string_processing.cpp`).

test/thermo/thermoFromYaml.cpp
```diff
@@ -387,3 +387,9 @@ TEST(ThermoFromYaml, BinarySolutionTabulatedThermo)
     thermo->setMoleFractionsByName("Li[anode]: 0.55, V[anode]: 0.45");
     EXPECT_NEAR(thermo->gibbs_mass(), -87066.246182649265, 1e-9);
 }
+
+TEST(ThermoFromYaml, DeprecatedPhase)
+{
+    make_deprecation_warnings_fatal();
```
speth: Ditto this one, with `main` defined in `test/thermo/nasapoly.cpp`.

speth (changes requested):
Just a few final suggestions.

bryanwweber:
@speth Thanks for all the comments! Updated again.

speth (approved):

ischoegl:
@bryanwweber ... it looks like the chemkin to yaml conversion is no longer used in `SConstruct`, but the chemkin inputs persist in the data folders. Was this by intent or is it an oversight?

bryanwweber:
@ischoegl By intent, we decided to handle/discuss other input files in a separate PR to avoid extending the scope of this one even further.

ischoegl:
@bryanwweber ... as #777 depends on chemkin input, I (temporarily?) added the chemkin->yaml conversion back into `SConstruct`.

bryanwweber:
@ischoegl OK, we can discuss that in #777

PR #776, opened 2019-12-19
Fix cxx combustor example

ischoegl:
Thanks for contributing code! Please include a description of your change and check your PR against the list below (for further questions, refer to the [contributing guide](https://github.com/Cantera/cantera/blob/master/CONTRIBUTING.md)).

Partial fix of #774

- Fix issue in `combustor.cpp`
- Update warn_deprecated message formatting

### DIFF
--- a/include/cantera/zeroD/FlowDevice.h
+++ b/include/cantera/zeroD/FlowDevice.h
@@ -135,7 +135,7 @@ public:
         warn_deprecated("FlowDevice::setMassFlowRate",
                         "To be removed after Cantera 2.5. "
                         "Use device-specific functions (e.g. "
-                        "MassFlowController::setMassFlowCoeff or "
+                        "MassFlowController::setMassFlowRate or "
                         "Valve::setValveCoeff) instead.");
         m_mdot = mdot;
     }
--- a/include/cantera/zeroD/FlowDevice.h
+++ b/include/cantera/zeroD/FlowDevice.h
@@ -46,10 +46,10 @@ public:
      * @deprecated To be changed after Cantera 2.5.
      */
     virtual int type() const {
-        warn_deprecated("FlowDevice::type()",
+        warn_deprecated("FlowDevice::type",
                         "To be changed after Cantera 2.5. "
                         "Return string instead of magic number; use "
-                        "FlowDevice::typeStr() during transition.");
+                        "FlowDevice::typeStr during transition.");
         return m_type;
     }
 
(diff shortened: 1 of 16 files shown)


bryanwweber:
@ischoegl Just FYI, the file `Wall.cpp` claims (for doxygen purposes) that it is `WallBase.cpp`. Can you clean that up here, since you're fixing a few other formatting things anyways?

ischoegl:
> @bryanwweber  Just FYI, the file Wall.cpp claims (for doxygen purposes) that it is WallBase.cpp. Can you clean that up here, since you're fixing a few other formatting things anyways?

Done.

samples/cxx/combustor/combustor.cpp
```diff
@@ -68,15 +68,15 @@ void runexample()
     // a short Gaussian pulse only to ignite the mixture
     MassFlowController m1;
     m1.install(fuel_in, combustor);
-    m1.setMassFlowRate(fuel_mdot);
+    m1.setMassFlowCoeff(fuel_mdot);
```
speth: I hadn't noticed this when bc8b4be6542a625a641a29169d308973de3a2c63 was added, but I think it would make sense for `MassFlowController` to override `setMassFlowRate`, since that is explicitly what is meant for this flow device -- the flow rate isn't this coefficient times some other quantity.

speth (changes requested):

samples/cxx/combustor/combustor.cpp
ischoegl: Sure. This wasn't as apparent from the cython interface, but under the hood the parameter `m_coeff` was introduced to handle both constant and time-varying cases. From C++, I agree that `setMassFlowRate` is more intuitive, and thus re-introduced it for the device-specific interface (i.e. `MassFlowController::setMassFlowRate` overloads the deprecated method).

I also tweaked the cython interface, and introduced `mass_flow_rate` as a property (to replace the `set_mass_flow_rate` method)

include/cantera/zeroD/FlowDevice.h
```diff
@@ -131,10 +131,12 @@ class FlowDevice
 
     //! Set the fixed mass flow rate (kg/s) through the flow device.
     //! @deprecated To be removed after Cantera 2.5.
-    void setMassFlowRate(double mdot) {
-        warn_deprecated("FlowDevice::setMassFlowRate()",
+    //! @todo make MassFlowController::setMassFlowRate non-virtual after 2.5
+    virtual void setMassFlowRate(double mdot) {
```
speth: I don't think this method should be `virtual`.

With this method marked `virtual`, use of `setMassFlowRate` on a `MassFlowController` accessed as a `FlowDevice*` will succeed without a deprecation warning, but then become an error when we remove the base class method.

If this method is non-virtual, use of `setMassFlowRate` on a `MassFlowController` accessed as a `FlowDevice*` will raise a deprecation warning which then becomes an error when we remove the base class method.

speth (changes requested):
The `mass_flow_rate` property looks good to me. I think the only remaining issue is one of the "fun" subtleties of virtual functions.

include/cantera/zeroD/FlowDevice.h
ischoegl: Good catch! (And easily fixed)

speth (approved):

PR #781, opened 2019-12-25
Fix broken links to website in docstrings.

rwest:
The examples were reorganized and there are a couple of links to them in the docstrings.
This tiny PR just fixes the links that I came across.


speth:
You mentioned "a couple", but there's only one change here. Was this the only one you found?

EDIT: Oh, I guess this is two links in one file.

rwest:
That one change fixes two links. I'm afraid I didn't grep for others.

speth:
A quick `git grep` suggests that these are the only two direct links to Python examples in the repo, so I guess you did find them all.

PR #780, opened 2019-12-20
Add support for Sundials 5.0.0 & and update submodule version

speth:
- [x] There is a clear use-case for this code change
- [x] The commit message has a short title & references relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage

System installations of Sundials 5.0.0 work without any internal changes to Cantera. Some changes to the build system are required to use it from the git submodule, which is also updated here to point to the official LLNL repository.

PR #782, opened 2019-12-26
Simple fix for Sundials 5.0 compatibility

band-a-prend:
Sundials 5.0.0 was released on 21 oct 2019 and has the same API for CVODES
as Sundials 4.x so the compatibility fix is trivial by addition of '5.0'
to the supported `sundials_version` list within SConstruct.


bryanwweber:
Thank you! This appears to be a duplicate of #780. Can you confirm, and if so, close this PR?

band-a-prend:
Yes, the mentioned pr is more extended in comparison with mine.
I missed to check the existing requests. Thank you.

PR #783, opened 2020-01-01
Update License.txt

ischoegl:
Happy 2020, everyone!

### DIFF
--- a/License.txt
+++ b/License.txt
@@ -6,7 +6,7 @@ Copyright (c) 2009 Sandia Corporation. Under the terms of
 Contract AC04-94AL85000 with Sandia Corporation, the U.S. Government
 retains certain rights in this software.
 
-Copyright (c) 2011-2018, Cantera Developers.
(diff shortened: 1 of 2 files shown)


speth:
:tada: The copyright in `doc/sphinx/conf.py` also needs to be updated. I think those are the only two in this repo.

ischoegl:
@bryanwweber ... this code coverage diff is even more interesting. I have no idea what's happening here.

bryanwweber:
I think maybe what happens is that the comparison is done to `master` at the time when the PR is opened and then that same commit is used even if the base commit of the branch is updated (e.g., rebase after master is updated) 🤷‍♂️ that would hypothetically explain the behavior here I think

speth:
Funny, it just updated itself again, and now the coverage looks fine (the fix for the oscillation in `HMWSoln.cpp` is in #773.

speth (approved):

PR #784, opened 2020-01-03
Add vapor fraction as option for setting state in YAML files

speth:

Addressing a question from @ischoegl in on https://github.com/Cantera/cantera-website/pull/91.

* Allows (T, Q), (P, Q), and (T, P, Q) as `state` definitions for pure fluid phases. For the last one, requires the three variables to be self-consistent.
* Fixes an issue that I ran into where leaving the `pure-fluid-name` out of the phase definition gave you a `water` object instead of throwing an exception.

### DIFF
--- a/src/thermo/ThermoPhase.cpp
+++ b/src/thermo/ThermoPhase.cpp
@@ -235,6 +235,9 @@ void ThermoPhase::setState(const AnyMap& input_state)
     if (state.hasKey("density")) {
         state["D"] = state["density"];
     }
+    if (state.hasKey("vapor-fraction")) {
+        state["Q"] = state["vapor-fraction"];
+    }
 
     // Set composition
     if (state.hasKey("X")) {
@@ -252,11 +255,28 @@ void ThermoPhase::setState(const AnyMap& input_state)
         }
         state.erase("Y");
     }
-    // set thermodynamic state using whichever property pair is found
+    // set thermodynamic state using whichever property set is found
     if (state.size() == 0) {
         setState_TP(298.15, OneAtm);
     } else if (state.hasKey("T") && state.hasKey("P")) {
-        setState_TP(state.convert("T", "K"), state.convert("P", "Pa"));
+        double T = state.convert("T", "K");
+        double P = state.convert("P", "Pa");
+        if (state.hasKey("Q")) {
+            double Q = state["Q"].asDouble();
+            double Psat = satPressure(T);
+            if (std::abs(Psat / P - 1) < 1e-6) {
+                setState_Tsat(T, Q);
+            } else if ((Q == 0 && P >= Psat) || (Q == 1 && P <= Psat)) {
+                setState_TP(T, P);
+            } else {
+                throw InputFileError("ThermoPhase::setState", state,
+                    "Temperature ({}), pressure ({}) and vapor fraction ({}) "
+                    "are inconsistent.\nPsat at this T: {}",
+                    T, P, Q, satPressure(T));
+            }
+        } else {
+            setState_TP(T, P);
+        }
     } else if (state.hasKey("T") && state.hasKey("D")) {
         setState_TR(state.convert("T", "K"), state.convert("D", "kg/m^3"));
     } else if (state.hasKey("T") && state.hasKey("V")) {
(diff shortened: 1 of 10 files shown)


bryanwweber:
Any reason not to cover the error messages and the `vapor-fraction` option with a test?

src/thermo/ThermoPhase.cpp
```diff
@@ -252,11 +255,28 @@ void ThermoPhase::setState(const AnyMap& input_state)
...
     if (state.size() == 0) {
         setState_TP(298.15, OneAtm);
     } else if (state.hasKey("T") && state.hasKey("P")) {
-        setState_TP(state.convert("T", "K"), state.convert("P", "Pa"));
+        double T = state.convert("T", "K");
+        double P = state.convert("P", "Pa");
+        if (state.hasKey("Q")) {
+            double Q = state["Q"].asDouble();
+            double Psat = satPressure(T);
+            if (std::abs(Psat / P - 1) < 1e-6) {
```
ischoegl: The cython version 
https://github.com/Cantera/cantera/blob/358195ea9eb99ccbf356338b7d31fac55fc0eb01/interfaces/cython/cantera/thermo.pyx#L1762-L1792
uses slightly different tolerances here (via [np.isclose](https://docs.scipy.org/doc/numpy/reference/generated/numpy.isclose.html)). Would it make sense to move all of this to a single spot in the C++ layer? (i.e. a new `setState_TPQ`; that way, all unit tests probing `Q` would apply universally). 

**PS:** I do realize that `Q` is still `X` in the C++ layer, which should be addressed eventually (together with `D` vs `R`).

ischoegl (commented):
Everything else looks good to me.

src/thermo/ThermoPhase.cpp
speth: Yeah, that's a good point on just having a single implementation of this procedure.

I'm not really convinced on the cost/benefit of deprecating and renaming all the relevant C++ functions.

ischoegl: limited cost / limited benefit = not a high priority. 

However, in this case, `setState_TPX` is already taken and we'd be starting to mix nomenclatures (i.e. `Q` and `D` are starting to bleed more and more into the C++ layer).

speth: One fun question that hasn't been resolved in either implementation: what should happen if T is above the critical point? Right now, it throws an exception while trying to evaluate the saturation pressure.

ischoegl: Ha! `Q` would be immaterial, but it's definitely a bug! I.e. it would be a user error to provide a value for `Q` in this case.

speth: I would certainly agree for any value 0 < Q < 1, but if a user were to specify water at T = 700 K, 1000 Pa and Q = 1, it doesn't seem that different from any other superheated state, which we do allow.

ischoegl: I guess it depends on whether you describe a supercritical fluid as liquid or vapor ... from a strict thermo perspective, it is neither.

speth: At least some definitions of a supercritical fluid are that both temperature and pressure need to be above the critical point. In this case, P > Pc with T < Tc is just a liquid and should be allowed with Q = 0, and correspondingly, P < Pc with T > Tc is just a gas / vapor and should be allowed with Q = 1.

ischoegl: I agree on what you said above. Above the vapor dome Q is somewhat murky (in particular, above the critical point at equal v, properties of a 'liquid' and a 'vapor' are the same; I checked that vc is *suggested* as a demarcation although I'd rather not use different labels). From an implementation viewpoint, your suggestion makes sense. :+1:

speth: Yes, I agree that it's murky in the supercritical region, but I think we should accept some value for Q there so that you can, for example, round-trip `SolutionArray` using T-P-Q as the columns even in the supercritical region. I opted for just ignoring the value of Q when P > Pc and T > Tc, rather than imposing a particular dividing line. This is implemented in the latest commit.

bryanwweber:
@speth I added one more test of the failure condition where the substance is unknown, to make sure that change isn't reverted. Feel free to drop it if you don't think it's necessary. This just seemed like the easiest way to add that change without a bunch of back and forth about what I meant 😄

speth:
The new phase definition you added breaks the test `python:test_composite.TestModels.test_load_thermo_models`, because it tries to instantiate every phase defined in `thermo-models.yaml`. If you really want to have a test for this, I would suggest putting the phase definition string in-line, rather than creating yet another input file just for this.

bryanwweber (approved):

PR #785, opened 2020-01-03
[Test] Include non-numeric text in test_problems comparisons

speth:

The comparison code for the output of "test_problems" tests had an error in its logic where lines where all of the numerical values were nearly identical would get counted as equal, even if other text on those lines was not the same. This even applied to lines with no numerical values.


ischoegl (commented):
Looks good to me.

PR #787, opened 2020-01-04
Deprecate redundant ThermoPhase classes

speth:

- Deprecate the "gas" mode of `IdealSolnGasVPSS` - this mode duplicates the more widely-used `IdealGasPhase` class.
- Deprecate `FixedChemPotSSTP` - this class can be easily replaced using the `StoichSubstance` class with a `ConstCpPoly` species thermo object.

### DIFF
--- a/src/thermo/FixedChemPotSSTP.cpp
+++ b/src/thermo/FixedChemPotSSTP.cpp
@@ -23,22 +23,34 @@ namespace Cantera
 FixedChemPotSSTP::FixedChemPotSSTP() :
     chemPot_(0.0)
 {
+    warn_deprecated("class FixedChemPotSSTP", "To be removed after Cantera 2.5. "
+        "Use class StoichSubstance with a constant-cp species thermo model, "
+        "with 'h0' set to the desired chemical potential and 's0' set to 0.");
 }
 
 FixedChemPotSSTP::FixedChemPotSSTP(const std::string& infile, const std::string& id_) :
     chemPot_(0.0)
 {
+    warn_deprecated("class FixedChemPotSSTP", "To be removed after Cantera 2.5. "
+        "Use class StoichSubstance with a constant-cp species thermo model, "
+        "with 'h0' set to the desired chemical potential and 's0' set to 0.");
     initThermoFile(infile, id_);
 }
 FixedChemPotSSTP::FixedChemPotSSTP(XML_Node& xmlphase, const std::string& id_) :
     chemPot_(0.0)
 {
+    warn_deprecated("class FixedChemPotSSTP", "To be removed after Cantera 2.5. "
+        "Use class StoichSubstance with a constant-cp species thermo model, "
+        "with 'h0' set to the desired chemical potential and 's0' set to 0.");
     importPhase(xmlphase, this);
 }
 
 FixedChemPotSSTP::FixedChemPotSSTP(const std::string& Ename, doublereal val) :
     chemPot_(0.0)
 {
+    warn_deprecated("class FixedChemPotSSTP", "To be removed after Cantera 2.5. "
+        "Use class StoichSubstance with a constant-cp species thermo model, "
+        "with 'h0' set to the desired chemical potential and 's0' set to 0.");
     std::string pname = Ename + "Fixed";
     setName(pname);
     setNDim(3);
--- a/doc/sphinx/yaml/phases.rst
+++ b/doc/sphinx/yaml/phases.rst
@@ -378,6 +378,10 @@ Example::
 A phase defined by a fixed value of the chemical potential, as
 `described here <https://cantera.org/documentation/dev/doxygen/html/d6/db0/classCantera_1_1FixedChemPotSSTP.html#details>`__.
 
+.. deprecated:: 2.5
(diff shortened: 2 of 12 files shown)


src/thermo/FixedChemPotSSTP.cpp
```diff
@@ -23,22 +23,34 @@ namespace Cantera
...
+        "with 'h0' set to the desired chemical potential and 's0' set to 0.");
     importPhase(xmlphase, this);
 }
 
 FixedChemPotSSTP::FixedChemPotSSTP(const std::string& Ename, doublereal val) :
     chemPot_(0.0)
 {
+    warn_deprecated("classFixedChemPotSSTP", "To be removed after Cantera 2.5. "
+        "Use class StoichSubstance with a Mu0Poly (piecewise-Gibbs) species "
+        "thermo model");
```
decaluwe: Why is this `deprecated` message different from the others?

decaluwe:
Hi @speth - sorry for the time it took me to look at this.  I feel like we discussed this at the steering committee meeting, and from what I remember these classes are indeed largely redundant.

The specific changes all look good to me.
 - I noticed one place where the deprecation message was different from the others, possibly a copy-paste error?  Noted above.
- I largely ignored close examination of the test results, but did notice that the internal energy values, for example, differed between the old and new implementations.  Not sure whether this might matter for a user, but I suppose a `FixedChemPotential` model is not discriminating based on internal energy, anyway.
- Out of curiosity, what does `make_deprecation_warnings_fatal` do?

src/thermo/FixedChemPotSSTP.cpp
speth: Not a copy paste error, but still incorrect. I had initially thought that `Mu0Poly` was the species thermo model that could be used here, but the `constant-cp` model is actually the one that has the same behavior as this class.

speth:
The internal energy values change because the `FixedChemPotSSTP` class didn't actually set the density, leaving it at the nonsense default value of 0.001 kg/m^3. The `StoichSubstance` class in contrast actually uses the specified density value. For both classes, the enthalpy is the same but since the internal energy is calculated as u = h - p/rho, it has to change.

`make_deprecation_warnings_fatal` changes deprecation warnings into exceptions. The purpose of using this is so that use of deprecated methods in the test suite causes the test to fail, rather than generating warnings that might never be noticed.

decaluwe (approved):
This looks good to me.  A few general questions, but I'm guessing there are no real issues, here.

1. So a general question: am I correct that VPSS stands for "variable pressure standard state" (i.e. the standard state depends on both temperature and pressure? If so, 

    a. This seems coincidentally relevant to issue #841 that I added yesterday (i.e. this PR provides insight on that issue, not vice-versa), and 

    b. I wonder if there is any reason a user would want to have this as a separate option?  The actual phase behavior between `IdealGasPhase` and the `gas` option for `IdealSolnGasVPSS` would be exactly the same, it seems; is there any benefit to the user to having the standard state be pressure sensitive?  Would they ever want to interrogate that property?  My guess:

     - No.
     - Even if so, the benefit and degree of interest in this capability would be so minimal that this PR is still the right move.

2. Now that I'm thinking about this a little more, are the solid and liquid options here also redundant?  It seems like they might duplicate `IdealSolidSolution`?  Just a hunch. Side note: I'll also note that there's no real reason to have the `Solid` as part of this latter class's name.  I've found it just as useful for liquid solutions.  But I digress...

Assuming the answers to the above are what I think they are, this looks like it's ready to go.

Thanks!

speth:
> So a general question: am I correct that VPSS stands for "variable pressure standard state"

Yes. I think this name is a bit of a misnomer, however, compared to what the class actually does.

> a. This seems coincidentally relevant to issue #841 that I added yesterday (i.e. this PR provides insight on that issue, not vice-versa), and

Let's deal with that elsewhere.

> b. I wonder if there is any reason a user would want to have this as a separate option? The actual phase behavior between IdealGasPhase and the gas option for IdealSolnGasVPSS would be exactly the same, it seems; is there any benefit to the user to having the standard state be pressure sensitive? Would they ever want to interrogate that property? 

Well, as far as I can tell, both `IdealGasPhase` and `IdealSolnGasVPSS` return the same values (to within rounding error) even for the standard state properties, such as  `getGibbs_RT`. So I'm not sure that there is an observable difference between these classes.

> 2\. Now that I'm thinking about this a little more, are the solid and liquid options here also redundant?  It seems like they might duplicate `IdealSolidSolution`?  Just a hunch. Side note: I'll also note that there's no real reason to have the `Solid` as part of this latter class's name.  I've found it just as useful for liquid solutions.

The `IdealSolidSolution` class isn't a full replacement for `IdealSolnGasVPSS` in "soln" mode because you can't use the `PDSS_SSVol` (density as a function of temperature) species thermo model with `IdealSolidSolution`. So what I was thinking is that longer term, it might make sense to rename `IdealSolnGasVPSS` to `IdealCondensed` and deprecate `IdealSolidSolution` in favor of that.

bryanwweber:
It seems like this PR reverted to SUNDIALS 5.0.0. Was that intentional?

speth:
No, obviously not. Fix pushed directly to master (5ae37adba).

PR #795, opened 2020-01-14
Prelude to YAML serialization

speth:
**Checklist**

To make the upcoming PR which implements serialization to the YAML format a bit smaller and more manageable, I have extracted a number of commits which resolve some issues that I encountered and interface additions that I found were needed while implementing serialization, which can be considered separately.

- Some additions to the `AnyMap` class, including:
  - equality comparisons
  - improved handling for entries which can be either single maps or lists of maps
  - a `clear()` method
  - automatic skipping of "hidden" fields (e.g. `__file__`) when iterating
  - generalization of handling for metadata applicable to an `AnyMap` tree
- Fix a bug in the pure fluid models where calculating c_p and other similar properties slightly changes the state
- Make instantiating phase types which reference other phase definitions possible when using a YAML input string (rather than an input file)
- Implement Option 4 in Cantera/enhancements#20
  - Reorganize Debye-Huckel input data
  - Allow the `equation-of-state` field to allow multiple model options
- Provide a way of determining the "canonical" name used by factories for instantiating an object, and make the names used in the YAML format canonical
- Update pure fluid names to match YAML naming style
- Fix adding YAML-defined reactions to CTI/XML-defined surface phases


include/cantera/base/FactoryBase.h
```diff
@@ -74,28 +74,56 @@ class Factory : public FactoryBase {
...
+    void addAlias(const std::string& original, const std::string& alias) {
+        if (!m_creators.count(original)) {
+            throw CanteraError("Factory::addAlias",
+                "Name '{}' not registered", original);
+        }
+        m_synonyms[alias] = original;
+    }
+
+    //! Get the canonical name registered for a type
+    std::string canonicalize(const std::string& name) {
```
bryanwweber, line 96: Would a better name for this function be `getCanonicalName()`? I guess Wikipedia [says](https://en.wikipedia.org/wiki/Canonicalization) that this name is the correct one, but I sure can't pronounce it 😄

data/liquidvapor.yaml
```diff
@@ -57,14 +57,22 @@ phases:
...
     P: 1.01325e+05
   pure-fluid-name: oxygen
+- name: carbon-dioxide
+  thermo: pure-fluid
+  elements: [C, O]
+  species: [CO2]
+  state:
+    T: 280.0
+    P: 1.01325e+05
+  pure-fluid-name: carbon-dioxide
```
bryanwweber, line 67: Is this intentionally a duplicate of the next entry (except the `name`) or is it a goof in the diff view?

speth: Deliberate. I want `carbon-dioxide` to be a valid name, but I don't want to to deprecate the old name yet because it would screw with the tests (specifically, `cti2yamlTest.test_liquidvapor` and `ctml2yamlTest.test_liquidvapor`).

include/cantera/base/FactoryBase.h
speth: Maybe, but it conflicts with the convention in our C++ code that `getFoo` names are mainly used for methods where something is being "returned" in a reference/pointer argument to the function. I guess "normalize" would be a simpler word, even if this particular usage is just as esoteric.

data/liquidvapor.yaml
bryanwweber: But now we have no tests of the old name, is that what you wanted? I guess I'm confused why we wouldn't test something if it's not deprecated yet?

include/cantera/base/FactoryBase.h
bryanwweber: OK, seems fine then.

data/liquidvapor.yaml
speth: Actually, looking at this again, I realized that deprecating these phase names doesn't interfere with those tests after all, so we can do so now. I've just pushed a revised commit for this.

src/base/AnyMap.cpp
```diff
@@ -189,7 +189,14 @@ struct convert<Cantera::AnyValue> {
...
             if (isInt(nodestr)) {
-                target = intValue(nodestr);
+                try {
+                    target = node.as<long int>();
+                } catch (YAML::BadConversion&) {
+                    // This exception is raised if the value doesn't fit in a
+                    // long int, in which case we would rather store it
+                    // (possibly inexactly) as a double.
+                    target = node.as<double>();
+                }
```
bryanwweber, line 201: In `ctml2yaml.py`, I had to add some code to remove integer-like reaction `id` fields (e.g., `0001`), because they raised an error in the YAML somewhere (sorry, I don't recall the exact message). Does this fix that problem, and can that code be removed from `ctml2yaml.py`?

data/liquidvapor.yaml
bryanwweber: Oh brother, I thought this was part of the test suite, not that it was the main `liquidvapor.yaml` input file 🤦‍♂ Well, I'm glad it worked out!

src/base/AnyMap.cpp
speth: I don't think this fixes that problem. Some resolution to that is on my list of rough edges in the YAML format that need to be fixed.

bryanwweber (dismissed):

doc/sphinx/yaml/phases.rst
```diff
@@ -316,6 +316,39 @@ Example::
...
+properties are:
+
+``ionic-radius``
+    Size of the species.
+
+``electrolyte-species-type``
+    One of ``solvent``, ``charged-species``, ``weak-acid-associated``,
+    ``strong-acid-associated``, ``polar-neutral``, or ``nonpolar-neutral``.
+    The types ``solvent``, ``charged-species``, and ``nonpolar-neutral`` can be
+    inferred automatically.
```
bryanwweber: Can you add something here about how these types are automatically inferred?

doc/sphinx/yaml/phases.rst
```diff
@@ -316,6 +316,39 @@ Example::
...
+
+``weak-acid-charge``
+    Charge to use for species can break apart into charged species.
```
bryanwweber:
```suggestion
    Charge to use for species that can break apart into charged species.
```

doc/sphinx/yaml/species.rst
```diff
@@ -20,10 +20,11 @@ The fields of a ``species`` entry are:
     and parameters. See :ref:`sec-yaml-species-thermo`.
 
 ``equation-of-state``
-    Mapping containing the equation of state model specification for the
-    species, any parameters for that model, and any parameters for interactions
-    with other species. :ref:`sec-yaml-species-eos`. If this field is absent,
-    the ``ideal-gas`` model is assumed.
+    A mapping or list of mappings. Each mapping contains an equation of state
+    model specification for the species, any parameters for that model, and any
+    parameters for interactions with other species. :ref:`sec-yaml-species-eos`.
```
bryanwweber: Is the reference intended to be a complete sentence here?

bryanwweber (approved):
Looks good to me, other than a few copy-editing things.

PR #797, opened 2020-01-22
Implement Func1-derived tabulated functions

### LINKED ISSUE #17: Allow arrays as input to time-varying properties
**Abstract**

Allow two-column arrays to be used to specify time-varying properties of the `Reactor` class.

**Motivation**

Currently, we have to use a function to calculate things like velocity, etc. that can vary with time. (I'm thinking here mostly of the Python interface but the idea may be appropriate for others as well.) This requires calling out to Python on every time step. It might be faster to pass a two-column array with time and values and have this done in C++ for cases where this is possible. Certainly, adding this kind of capability would make it much easier for users to specify this kind of data, but I'm not sure how common this use case is. 

This is inspired by the need for RCM simulations to specify pre-calculated volume (i.e., velocity) as a function of time. Of course, this should be benchmarked for speed to see if there is actually any improvement.

**Possible Solutions**

The way this could work from Python is something like:
```python
import numpy as np
vel_array = np.genfromtxt('velocity-trace.txt')
# vel_array looks like [[0.0, 1.0], [1.0, 2.0], ...]
# Normal Cantera Reactor or IdealGasReactor setup
reac = ct.IdealGasReactor(gas)
w = ct.Wall(reac, env, velocity=vel_array)
# Rest of the ReactorNetwork simulation
```

Something like this would also work for the heat transfer function for `Reactor`, as well as the mass flow rate for flow devices (and possibly other cases I'm missing).

One question is how should time values between values actually specified should be handled. In other words, given `vel_array = [[0.0, 1.0], [1.0, 2.0]]`, if `t = 0.5`, what velocity should be used? The implementation of this in CHEMKIN uses the velocity from the nearest previous time value, i.e., for `t = 0.5`, `v = 1.0`.

Other possible ideas would be to fit a function (linear, quadratic, etc.) between the data points and interpolate. I'm not in favor of this idea, since we don't know what type of function should be used, and if a user wants something like this they can fall back to the existing `Func1`-type interface.

**References**

This Google Group post has one possible (typo-filled) implementation of a function to read a volume-vs-time curve into an array and return the appropriate velocity: https://groups.google.com/forum/#!topic/cantera-users/5yoAUvAdOLs

Another, class-based, implementation of the same idea is here: https://github.com/bryanwweber/CanSen/blob/43ffe2f2f6959d9208a4a44d508866e029eb567b/cansen/profiles.py#L7-L59

ischoegl:
**Checklist**

Fixes Cantera/enhancements#17

- use existing C++ `Func1.h` framework
- create Python interface for new tabulated `Func1` object and add unit tests

Implementing a tabulated function as a class derived from `Func1` allows for a straight-forward integration into the existing framework without further code changes. An implementation in the C++ layer has the advantage that this solution is usable from (or at least can be extended to) all platforms.

**Use case and benchmark**

```
In [1]: import cantera as ct
   ...: import numpy as np
   ...: arr = np.array([[0, 2], [1, 1], [2, 0]])

In [2]: fcn1 = ct.Func1(arr)
   ...: fcn1(1.3)
Out[2]: 0.7

In [3]: fcn2 = ct.Func1(arr, interpolation='previous')
   ...: fcn2(1.3)
Out[3]: 1.0

In [4]: %timeit fcn1(1.3)
The slowest run took 29.33 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 122 ns per loop

In [5]: time = arr[:, 0]
   ...: fval = arr[:, 1]
   ...: fcn3 = lambda t: np.interp(t, time, fval)

In [6]: %timeit fcn3(1.3)
The slowest run took 9.63 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 5.46 us per loop

In [7]: fcn4 = ct.Func1(fcn3)

In [8]: %timeit fcn4(1.3)
The slowest run took 13.10 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 5.76 us per loop
```

### DIFF
--- a/include/cantera/numerics/Func1.h
+++ b/include/cantera/numerics/Func1.h
@@ -275,45 +275,18 @@ public:
 };
 
 
-//! implements a tabulated function
+//! The Tabulated1 class implements a tabulated function
 class Tabulated1 : public Func1
 {
 public:
-    Tabulated1(const std::vector<double>& tvec, const std::vector<double>& fvec,
-               const std::string& method = "linear") :
-        Func1() {
-        if (tvec.size() != fvec.size()) {
-            throw CanteraError("Tabulated1::Tabulated1",
-                               "sizes of vectors do not match ({} vs {}).",
-                               tvec.size(), fvec.size());
-        } else if (tvec.empty()) {
-            throw CanteraError("Tabulated1::Tabulated1",
-                               "vectors must not be empty.");
-        }
-        m_tvec = tvec;
-        m_fvec = fvec;
-        if (method == "linear") {
-            m_isLinear = true;
-        } else if (method == "previous") {
-            m_isLinear = false;
-        } else {
-            throw CanteraError("Tabulated1::Tabulated1",
-                               "interpolation method '{}' is not implemented",
-                               method);
-        }
-    }
-
-    Tabulated1(const Tabulated1& b) :
-        Func1(b) {
-    }
-
-    Tabulated1& operator=(const Tabulated1& right) {
-        if (&right == this) {
-            return *this;
-        }
-        Func1::operator=(right);
-        return *this;
-    }
+    //! Constructor.
+    /*!
+     * @param tvec   Vector of time values
+     * @param fvec   Vector of values
+     * @param method Interpolation method ('linear' or 'previous')
+     */
+    Tabulated1(const vector_fp& tvec, const vector_fp& fvec,
+               const std::string& method = "linear");
 
     virtual std::string write(const std::string& arg) const;
     virtual int ID() const {
@@ -330,19 +303,21 @@ public:
 
     virtual Func1& derivative() const;
 private:
-    std::vector<double> m_tvec;
-    std::vector<double> m_fvec;
-    bool m_isLinear;
+    vector_fp m_tvec; //!< Vector of time values
+    vector_fp m_fvec; //!< Vector of function values
+    bool m_isLinear; //!< Boolean indicating interpolation method
 };
 
 
-/**
- * Constant.
- */
+//! The Const1 class implements a constant
 class Const1 : public Func1
 {
 public:
-    Const1(doublereal A) :
+    //! Constructor.
+    /*!
+     * @param A   Constant
+     */
+    Const1(double A) :
         Func1() {
         m_c = A;
     }
--- a/include/cantera/numerics/Func1.h
+++ b/include/cantera/numerics/Func1.h
@@ -9,6 +9,7 @@
 #define CT_FUNC1_H
 
 #include "cantera/base/ct_defs.h"
+#include "cantera/base/ctexceptions.h"
 
 #include <iostream>
 
@@ -32,6 +33,7 @@ const int CosFuncType = 102;
 const int ExpFuncType = 104;
 const int PowFuncType = 106;
 const int ConstFuncType = 110;
+const int TabulatedFuncType = 120;
 
 class TimesConstant1;
 
@@ -123,6 +125,7 @@ Func1& newCompositeFunction(Func1& f1, Func1& f2);
 Func1& newTimesConstFunction(Func1& f1, doublereal c);
 Func1& newPlusConstFunction(Func1& f1, doublereal c);
 
+
 //! implements the sin() function
 /*!
  * The argument to sin() is in radians
@@ -165,7 +168,11 @@ public:
     virtual Func1& derivative() const;
 };
 
-/// cos
+
+//! implements the cos() function
+/*!
+ * The argument to cos() is in radians
+ */
 class Cos1 : public Func1
 {
 public:
@@ -200,7 +207,8 @@ public:
     virtual Func1& derivative() const;
 };
 
-/// exp
+
+//! implements the exponential function
 class Exp1 : public Func1
 {
 public:
@@ -233,7 +241,8 @@ public:
     virtual Func1& derivative() const;
 };
 
-/// pow
+
+//! implements the power function (pow)
 class Pow1 : public Func1
 {
 public:
@@ -264,6 +273,53 @@ public:
     }
     virtual Func1& derivative() const;
 };
+
+
+//! implements a tabulated function
+class Tabulated1 : public Func1
+{
+public:
+    Tabulated1(const std::vector<double>& tvec, const std::vector<double>& fvec) :
+        Func1() {
+        if (tvec.size() != fvec.size()) {
+            throw CanteraError("Tabulated1::Tabulated1",
+                               "sizes of vectors do not match ({} vs {}).",
+                               tvec.size(), fvec.size());
+        } else if (tvec.empty()) {
+            throw CanteraError("Tabulated1::Tabulated1",
+                               "vectors must not be empty.");
+        }
+        m_tvec = tvec;
+        m_fvec = fvec;
+    }
+
+    Tabulated1(const Tabulated1& b) :
+        Func1(b) {
+    }
+
+    Tabulated1& operator=(const Tabulated1& right) {
+        if (&right == this) {
+            return *this;
+        }
+        Func1::operator=(right);
+        return *this;
+    }
+
+    virtual std::string write(const std::string& arg) const;
+    virtual int ID() const {
+        return TabulatedFuncType;
+    }
+    virtual double eval(double t) const;
+    virtual Func1& duplicate() const {
+        return *(new Tabulated1(m_tvec, m_fvec));
+    }
+
+    virtual Func1& derivative() const;
+private:
+    std::vector<double> m_tvec;
+    std::vector<double> m_fvec;
+};
+
 
 /**
  * Constant.
--- a/include/cantera/numerics/Func1.h
+++ b/include/cantera/numerics/Func1.h
@@ -279,7 +279,8 @@ public:
 class Tabulated1 : public Func1
 {
 public:
-    Tabulated1(const std::vector<double>& tvec, const std::vector<double>& fvec) :
+    Tabulated1(const std::vector<double>& tvec, const std::vector<double>& fvec,
+               const std::string& method = "linear") :
         Func1() {
         if (tvec.size() != fvec.size()) {
             throw CanteraError("Tabulated1::Tabulated1",
@@ -291,6 +292,15 @@ public:
         }
         m_tvec = tvec;
         m_fvec = fvec;
+        if (method == "linear") {
+            m_isLinear = true;
+        } else if (method == "previous") {
+            m_isLinear = false;
+        } else {
+            throw CanteraError("Tabulated1::Tabulated1",
+                               "interpolation method '{}' is not implemented",
+                               method);
+        }
     }
 
     Tabulated1(const Tabulated1& b) :
@@ -311,13 +321,18 @@ public:
     }
     virtual double eval(double t) const;
     virtual Func1& duplicate() const {
-        return *(new Tabulated1(m_tvec, m_fvec));
+        if (m_isLinear) {
+            return *(new Tabulated1(m_tvec, m_fvec, "linear"));
+        } else {
+            return *(new Tabulated1(m_tvec, m_fvec, "previous"));
+        }
     }
 
     virtual Func1& derivative() const;
 private:
     std::vector<double> m_tvec;
     std::vector<double> m_fvec;
+    bool m_isLinear;
 };
 
 
--- a/include/cantera/numerics/Func1.h
+++ b/include/cantera/numerics/Func1.h
@@ -281,11 +281,12 @@ class Tabulated1 : public Func1
 public:
     //! Constructor.
     /*!
-     * @param tvec   Vector of time values
-     * @param fvec   Vector of values
+     * @param n      Size of tabulated value arrays
+     * @param tvals   Pointer to time value array
+     * @param fvals   Pointer to function value array
      * @param method Interpolation method ('linear' or 'previous')
      */
-    Tabulated1(const vector_fp& tvec, const vector_fp& fvec,
+    Tabulated1(size_t n, const double* tvals, const double* fvals,
                const std::string& method = "linear");
 
     virtual std::string write(const std::string& arg) const;
@@ -295,9 +296,11 @@ public:
     virtual double eval(double t) const;
     virtual Func1& duplicate() const {
         if (m_isLinear) {
-            return *(new Tabulated1(m_tvec, m_fvec, "linear"));
+            return *(new Tabulated1(m_tvec.size(), &m_tvec[0], &m_fvec[0],
+                                    "linear"));
         } else {
-            return *(new Tabulated1(m_tvec, m_fvec, "previous"));
+            return *(new Tabulated1(m_tvec.size(), &m_tvec[0], &m_fvec[0],
+                                    "previous"));
         }
     }
 
--- a/interfaces/cython/cantera/func1.pyx
+++ b/interfaces/cython/cantera/func1.pyx
@@ -2,6 +2,7 @@
 # at https://cantera.org/license.txt for license and copyright information.
 
 import sys
+from numbers import Real as _real
 
 cdef double func_callback(double t, void* obj, void** err):
     """
@@ -46,35 +47,97 @@ cdef class Func1:
         >>> f3(6)
         30.0
 
-    For simplicity, constant functions can be defined by passing the constant
+    For simplicity, constant functions can be defined by passing a constant
     value directly::
 
         >>> f4 = Func1(2.5)
         >>> f4(0.1)
         2.5
 
+    A `Func1` object representing a tabulated function is defined by sample
+    points and corresponding function values. Inputs are specified either by
+    two iterable objects containing sample point location and function values,
+    or a single array that concatenates those inputs in two columns. Between
+    sample points, values are evaluated using a linear interpolation, whereas
+    outside the sample interval, the value at the closest end point is returned.
+    Examples for tabulated `Func1` object are::
+
+        >>> t1 = Func1([[0, 2], [1, 1], [2, 0]])
+        >>> [t1(v) for v in [-0.5, 0, 0.5, 1.5, 2, 2.5]]
+        [2.0, 2.0, 1.5, 0.5, 0.0, 0.0]
+
+        >>> t2 = Func1([0, 1, 2], [2, 1, 0])
+        >>> [t2(v) for v in [-0.5, 0, 0.5, 1.5, 2, 2.5]]
+        [2.0, 2.0, 1.5, 0.5, 0.0, 0.0]
+
+        >>> t3 = Func1(np.array([0, 1, 2]), (2, 1, 0))
+        >>> [t3(v) for v in [-0.5, 0, 0.5, 1.5, 2, 2.5]]
+        [2.0, 2.0, 1.5, 0.5, 0.0, 0.0]
+
     Note that all methods which accept `Func1` objects will also accept the
     callable object and create the wrapper on their own, so it is generally
     unnecessary to explicitly create a `Func1` object.
     """
-    def __cinit__(self, c):
+    def __cinit__(self, *args, **kwargs):
         self.exception = None
-        if hasattr(c, '__call__'):
-            self.callable = c
-        else:
-            try:
-                # calling float() converts numpy arrays of size 1 to scalars
-                k = float(c)
-            except TypeError:
-                if hasattr(c, '__len__') and len(c) == 1:
-                    # Handle lists or tuples with a single element
-                    k = float(c[0])
-                else:
-                    raise TypeError('Func1 must be constructed from a number or'
-                                    ' a callable object')
-            self.callable = lambda t: k
+        self.callable = None
 
+    def __init__(self, *args):
+        if len(args) == 1:
+            c = args[0]
+            if hasattr(c, '__call__'):
+                # callback function
+                self.__set_callback(c)
+            else:
+                arr = np.array(c)
+                try:
+                    if arr.ndim == 0:
+                        # handle constants or unsized numpy arrays
+                        k = float(c)
+                        self.__set_callback(lambda t: k)
+                    elif arr.size == 1:
+                        # handle lists, tuples or numpy arrays with a single element
+                        k = float(c[0])
+                        self.__set_callback(lambda t: k)
+                    elif arr.ndim == 2:
+                        # tabulated function (single argument)
+                        if arr.shape[1] == 2:
+                            time = arr[:, 0]
+                            fval = arr[:, 1]
+                            self.__set_tables(time, fval)
+                        else:
+                            raise ValueError(
+                                "Invalid dimensions: specification of "
+                                "tabulated function with a single array "
+                                "requires two columns")
+                    else:
+                        raise TypeError
+
+                except TypeError:
+                    raise TypeError(
+                        "Func1 must be constructed from a callable object, "
+                        "a number or a numeric array with two columns")
+
+        elif len(args) == 2:
+            # tabulated function (two arguments mimic C++ interface)
+            time, fval = args
+            self.__set_tables(time, fval)
+
+        else:
+            raise ValueError("Invalid number of arguments")
+
+
+    cpdef void __set_callback(self, c) except *:
+        self.callable = c
         self.func = new CxxFunc1(func_callback, <void*>self)
+
+    cpdef void __set_tables(self, time, fval) except *:
+        cdef vector[double] tvec, fvec
+        for t in time:
+            tvec.push_back(t)
+        for f in fval:
+            fvec.push_back(f)
+        self.func = <CxxFunc1*>(new CxxTabulated1(tvec, fvec))
 
     def __dealloc__(self):
         del self.func
--- a/interfaces/cython/cantera/func1.pyx
+++ b/interfaces/cython/cantera/func1.pyx
@@ -110,25 +110,32 @@ cdef class TabulatedFunction(Func1):
     by two iterable objects containing sample point location and function
     values, or a single array that concatenates those inputs in two rows or
     columns. Between sample points, values are evaluated based on the optional
-    keyword argument 'method'; options are 'linear' (linear interpolation,
-    default) or 'previous' (nearest previous value). Outside the sample
-    interval, the value at the closest end point is returned.
+    argument ``method``, which has to be supplied as a keyword; options are 
+    ``'linear'`` (linear interpolation, default) or ``'previous'`` (nearest 
+    previous value). Outside the sample interval, the value at the closest end 
+    point is returned.
 
-    Examples for `TabulatedFunction` objects are::
+    Examples for `TabulatedFunction` objects using a single (two-dimensional)
+    array as input are::
 
         >>> t1 = TabulatedFunction([[0, 2], [1, 1], [2, 0]])
         >>> [t1(v) for v in [-0.5, 0, 0.5, 1.5, 2, 2.5]]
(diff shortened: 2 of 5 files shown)


bryanwweber:
This is great, and I'm glad the implementation turned out to be so simple. I disagree, however, that interpolation should be done between values. In my opinion, the nearest previous value should be returned. If we want to include interpolation, it should be an option that can be passed, IMO, e.g.,

```python
ct.Func1(arr, interpolation=None)  # or 'None'?
ct.Func1(arr, interpolation='linear')
ct.Func1(arr, interpolation='quadratic')
```
etc.

ischoegl:
@bryanwweber ... happy that this may be useful.

The main reason why I implemented a linear interpolation is that it is more physical/flexible, and already contains the previous value case via:
```python
ct.Func1([0, 1, 1, 2], [0, 0, 1, 1]) # using the two argument version for clarity
```
Beyond, I'm unsure about the fate of `Func1.h`, which defines methods for derivatives, etc. that are not broken out to Python. As I'm working from within that framework, things aren't straight-forward, and I'd rather not get into this deeper than absolutely necessary.

**PS:** while I have a strong preference for interpolation (we don't necessarily have to mimic CHEMKIN here), adding the 'previous value' alternative is still relatively simple. I do not think that it's worth going into quadratic or spline variants in this PR.

ischoegl:
@bryanwweber ... I ended up adding the non-interpolating option per your suggestion, i.e.
```python
ct.Func1([0, 1, 1, 2], [0, 0, 1, 1], interpolation='previous') # CHEMKIN style
ct.Func1([0, 1, 1, 2], [0, 0, 1, 1], interpolation='linear')
ct.Func1([0, 1, 1, 2], [0, 0, 1, 1]) # defaults to 'linear'
```
I am proposing to leave the default as `linear` as it is more intuitive. The interpolation method strings match [scipy.interpolate.interp1d](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html) to future-proof potential additions beyond this PR.

bryanwweber:
@ischoegl Thanks for adding the "previous" method! I appreciate you taking the time to consider my suggestion and add that. I want to advocate for having the "previous" method be the default, let me know what you think.

I think there is also a physical justification for the "previous" method being the default, beyond compatibility with CHEMKIN (which I agree is not strictly necessary). The justification I'm thinking of is that by assuming a linear variation, we are assuming more about the input data than the user has provided us. We are assuming that if the user were to add additional values in between, they would fall on a straight line between the two surrounding points. However, there is no particular reason we should assume this; the data could just as easily vary quadratically, sinusoidally, etc (note, I'm not suggesting these other options be implemented here). So on that basis, the case with the fewest assumptions on our part is that the data do not vary between the provided points. IMO this is the more reasonable assumption to make. What do you think?

ischoegl:
> I want to advocate for having the "previous" method be the default, let me know what you think.

Sure! I am mainly following precedent for default options of [scipy.interpolation.interp1d](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html) or [MATLAB interp1d](https://www.mathworks.com/help/matlab/ref/interp1.html), i.e. imho 'linear' is the expected behavior. There are also some practical reasons: this approach avoids discontinuous inputs when  integrating ODE's. As a physical example, positions resulting from velocity inputs become C2 continuously differentiable, which avoids unphysical 'dirac'-type accelerations.

bryanwweber:
> As a physical example, positions resulting from velocity inputs become C2 continuously differentiable, which avoids unphysical 'dirac'-type accelerations.

Hmm, this makes a lot of sense as well. In the past, I've always set the maximum time step in the integration to be the smallest time interval between data points, which I think helps with this problem. I wonder how much difference it would make practically.

> I am mainly following precedent for default options of scipy.interpolation.interp1d or MATLAB interp1d, i.e. imho 'linear' is the expected behavior.

I think the expected behavior would be linear if the user is expecting interpolation to happen. I don't think it is obvious that interpolation should happen for tabular data input.

ischoegl:
> I think the expected behavior would be linear if the user is expecting interpolation to happen. I don't think it is obvious that interpolation should happen for tabular data input.

Hmm. Based on past experience (using MATLAB/Scipy/etc.) I would automatically assume that tabulated input involves interpolation. I guess I've never used CHEMKIN much, so I'm not used to their convention. I.e. there may be no clear answer here.

I am still partial to the physics argument: if given the choice of C1 vs C0 input, I'd stick with the more 'natural' choice. My intuition would be that the system is less stiff for C1, as the ODE solver is unable to predict the effect of discontinuous input for C0. (Ps: similar arguments hold for higher derivatives, but C0 is imho the worst-case scenario for ODE input.)

Considering the alternatives, I'm still advocating for `linear`.

ischoegl:
@speth ... while this is motivated by RCM work, would you mind having a quick look? Parts of this are reviving C++ Func1 objects where I’m not sure what their fate was intended to be.

include/cantera/numerics/Func1.h
```diff
@@ -265,6 +274,68 @@ class Pow1 : public Func1
...
 
+
+//! implements a tabulated function
+class Tabulated1 : public Func1
+{
+public:
+    Tabulated1(const std::vector<double>& tvec, const std::vector<double>& fvec,
+               const std::string& method = "linear") :
+        Func1() {
+        if (tvec.size() != fvec.size()) {
```
speth: This implementation of this method is sufficiently involved that it should probably be moved to the `.cpp` file.

include/cantera/numerics/Func1.h
```diff
@@ -265,6 +274,68 @@ class Pow1 : public Func1
     virtual Func1& derivative() const;
 };
 
+
+//! implements a tabulated function
+class Tabulated1 : public Func1
+{
+public:
+    Tabulated1(const std::vector<double>& tvec, const std::vector<double>& fvec,
```
speth: Cantera generally uses simple array pointers here to avoid forcing users to use specific container types. Following the examples of the `Poly1` and `Fourier1` classes, this would then be
```c++
Tabulated1(size_t n, const double* tvec, const double* fvec, const std::string& method="linear");
```
Also, can you add a docstring for the constructor (even if none of the other `Func1` classes have a meaningful one).

include/cantera/numerics/Func1.h
```diff
@@ -265,6 +274,68 @@ class Pow1 : public Func1
...
+        } else {
+            throw CanteraError("Tabulated1::Tabulated1",
+                               "interpolation method '{}' is not implemented",
+                               method);
+        }
+    }
+
+    Tabulated1(const Tabulated1& b) :
+        Func1(b) {
+    }
```
speth: This implementation of the copy constructor can't be correct -- the values of `m_tvec`, `m_fvec`, and `m_isLinear` won't be assigned. I think you can actually just omit this, as the automatically generated copy constructor will do just fine.

include/cantera/numerics/Func1.h
```diff
@@ -265,6 +274,68 @@ class Pow1 : public Func1
...
+
+    Tabulated1& operator=(const Tabulated1& right) {
+        if (&right == this) {
+            return *this;
+        }
+        Func1::operator=(right);
+        return *this;
+    }
```
speth: Likewise, this doesn't set any of the data specific to `Tabulated1`, and the compiler-generated default implementation should actually do the correct thing.

I think you're being misled by some of the other classes here where there are no member variables that need to be assigned because they just use the constant `m_c` that is part of the base class.

interfaces/cython/cantera/func1.pyx
```diff
@@ -46,36 +47,110 @@ cdef class Func1:
...
+    sample points, values are evaluated based on the keyword argument
+    'interpolation'; options are 'linear' (linear interpolation, default) or
+    'previous' (nearest previous value). Outside the sample interval, the value
+    at the closest end point is returned.
+
+    Examples for tabulated `Func1` object are::
+
+        >>> t1 = Func1([[0, 2], [1, 1], [2, 0]])
+        >>> [t1(v) for v in [-0.5, 0, 0.5, 1.5, 2, 2.5]]
+        [2.0, 2.0, 1.5, 0.5, 0.0, 0.0]
```
speth: I think it might make more sense to introduce a new `TabulatedFunc1` class in Python rather than overloading the constructor for `Func1` based on argument types. That would leave the door open for introducing other function types, and if I'm following this implementation correctly, remove the need to expose the `Const1` class.

include/cantera/numerics/Func1.h
```diff
@@ -265,6 +274,68 @@ class Pow1 : public Func1
...
+            return *(new Tabulated1(m_tvec, m_fvec, "linear"));
+        } else {
+            return *(new Tabulated1(m_tvec, m_fvec, "previous"));
+        }
+    }
+
+    virtual Func1& derivative() const;
+private:
+    std::vector<double> m_tvec;
+    std::vector<double> m_fvec;
```
speth:
```suggestion
    vector_fp m_fvec;
```
Here and anywhere else this appears.

include/cantera/numerics/Func1.h
```diff
@@ -265,6 +274,68 @@ class Pow1 : public Func1
...
+        if (tvec.size() != fvec.size()) {
+            throw CanteraError("Tabulated1::Tabulated1",
+                               "sizes of vectors do not match ({} vs {}).",
+                               tvec.size(), fvec.size());
+        } else if (tvec.empty()) {
+            throw CanteraError("Tabulated1::Tabulated1",
+                               "vectors must not be empty.");
+        }
+        m_tvec = tvec;
+        m_fvec = fvec;
```
speth: Presumably this should do some check to make sure that the time values are monotonically increasing, right?

speth (changes requested):
I have plenty of misgivings about the current implementation of the `Func1` set of classes, but this barely touches on the problematic parts of that interface, so I think it should be okay and won't require too many changes if/when I get around to a more complete overhaul of `Func1`.

include/cantera/numerics/Func1.h
ischoegl: I am really conflicted about using raw pointers here. For a future overhaul of the `Func1` set of classes, I would advocate for an implementation where `vector_fp` is used for instances like that (I did change `std::vector<double>` to `vector_fp` here).

:thumbsup: for the docstring.

interfaces/cython/cantera/func1.pyx
ischoegl: I exposed the `Const1` class as the C++ implementation is presumably more efficient (and it was already implemented).

Further, the reason for keeping everything in a single constructor was inspired by the pre-existing exemption of constants (which are now handled by the C++ layer). I'd either change both (defining new `Tabulated1` and `Const1` classes, deprecating the old behavior for the latter), or leave things as they are.

As an aside: if we start to introduce derived classes in Python, I may add a `Func1` derived function implementing a crank function, as this is frequently needed for engine-related applications.

include/cantera/numerics/Func1.h
speth: It's not just the `Func1` family of classes, though. This applies to most Cantera classes, including `ThermoPhase`, `Kinetics`, `Transport`, and others. Switching broadly to `vector<double>` would incur a lot of extra temporary array allocations and copy operations, especially when using Cantera from languages other than C++. The performance impact for this class is almost certainly negligible, but in some other cases it will be more significant and I think the consistency is worthwhile.

interfaces/cython/cantera/func1.pyx
speth: I guess my point is that currently, `Func1` objects in Python are almost always created automatically as a wrapper when the user provides either a Python function object or a constant to a method like `Wall.set_velocity`. With the tabulated formulation, I would expect users to write something more like:
```python
wall.set_velocity(TabulatedFunc([0,1,2], [0.1, 0.2, 0.3]))
```
rather than simply:
```py
wall.set_velocity([[0,1,2], [0.1, 0.2, 0.3]])
```
I'm fine with splitting out `Const1` as long as the existing behavior for functions like `Wall.set_velocity` is preserved.

ischoegl: Makes sense. Hadn't really considered this auto-generated wrapper case.

ischoegl: Alright. The `Tabulated1` Python class is now introduced (nomenclature matches C++), i.e.

```python
wall.set_velocity(Tabulated1([0,1,2], [0.1, 0.2, 0.3]))
```

After running some tests, it turns out the the C++ version of `Const1` is actually 50% slower (!) than the Python `lambda` function (likely Cython overhead?); accordingly, there's no point in exposing and/or introducing an interface for C++ `Const1`.

interfaces/cython/cantera/func1.pyx
```diff
@@ -87,3 +100,78 @@ cdef class Func1:
...
+    def __init__(self, *args, **kwargs):
+        if len(args) == 1:
+            # tabulated function (single argument)
+            arr = np.array(args[0])
+            if arr.ndim == 2:
+                if arr.shape[1] == 2:
+                    time = arr[:, 0]
+                    fval = arr[:, 1]
+                    method = kwargs.get('interpolation', 'linear')
+                    self._set_tables(time, fval, stringify(method))
```
bryanwweber: Any reason not to allow row-based input here as well? It seems as long as `ndim` is 2, it doesn't matter whether it is rows or columns.

interfaces/cython/cantera/func1.pyx
```diff
@@ -46,34 +47,46 @@ cdef class Func1:
...
                 else:
-                    raise TypeError('Func1 must be constructed from a number or'
-                                    ' a callable object')
-            self.callable = lambda t: k
+                    raise TypeError
+
+            except TypeError:
+                raise TypeError(
+                    "Func1 must be constructed from a callable object "
+                    "or a number")
```
bryanwweber: I think here you'll want to add `raise TypeError(...) from None` to avoid the double traceback. See here: https://docs.python.org/3/library/exceptions.html

interfaces/cython/cantera/func1.pyx
```diff
@@ -87,3 +100,78 @@ cdef class Func1:
 
     def __copy__(self):
         raise NotImplementedError('Func1 object is not copyable')
+
+
+cdef class Tabulated1(Func1):
```
bryanwweber: Can we come up with another, more descriptive, name for this than `Tabulated1`? I assume the base class is called `Func1` because it takes one argument, but I don't think we necessarily need to carry that convention into derived classes. In most cases, it is not necessary to explicitly call `Func1`, so I don't think most users will see that name. However, this class must be called by a user to set it up, so a better name would help with discoverability.

interfaces/cython/cantera/func1.pyx
```diff
@@ -87,3 +100,78 @@ cdef class Func1:
...
+        >>> t3 = Tabulated1([0, 1, 2], [2, 1, 0])
+        >>> [t3(v) for v in [-0.5, 0, 0.5, 1.5, 2, 2.5]]
+        [2.0, 2.0, 1.5, 0.5, 0.0, 0.0]
+
+        >>> t4 = Tabulated1(np.array([0, 1, 2]), (2, 1, 0))
+        >>> [t4(v) for v in [-0.5, 0, 0.5, 1.5, 2, 2.5]]
+        [2.0, 2.0, 1.5, 0.5, 0.0, 0.0]
+    """
+
+    def __init__(self, *args, **kwargs):
```
bryanwweber: There's an implicit requirement here that `interpolation` be a keyword argument, but it's not clear from the docstring that this is a *requirement*. I also wonder whether `method` would be a better name? Especially because this is a keyword argument, `interpolation` is pretty long and probably prone to typos that won't be caught as an argument error (!), leading to really hard to debug behavior.

I think possibly an easier-to-use implementation of this might be:
```python
def __init__(self, arr1=None, arr2=None, *, method='linear'):
```
Then the first two arguments could be provided as positional or keyword, and `method` is checked properly (avoids typos). Probably need to come up with better names than `arr1` and `arr2` though.

interfaces/cython/cantera/func1.pyx
```diff
@@ -87,3 +100,78 @@ cdef class Func1:
...
+                    "a numeric array with two columns")
+
+        elif len(args) == 2:
+            # tabulated function (two arrays mimic C++ interface)
+            time, fval = args
+            method = kwargs.get('interpolation', 'linear')
+            self._set_tables(time, fval, stringify(method))
+
+        else:
+            raise ValueError("Invalid number of arguments")
```
bryanwweber: Can you improve this error message? What is the appropriate number of arguments?

interfaces/cython/cantera/func1.pyx
```diff
@@ -87,3 +100,78 @@ cdef class Func1:
...
+        fsiz = np.asarray(fval).size
+        if tsiz != fsiz:
+            raise ValueError("Sizes of arrays do not match "
+                             "({} vs {}).".format(tsiz, fsiz))
+        elif tsiz == 0:
+            raise ValueError("Arrays must not be empty.")
+        cdef np.ndarray[np.double_t, ndim=1] tvec = np.asarray(time,
+                                                               dtype=np.double)
+        cdef np.ndarray[np.double_t, ndim=1] fvec = np.asarray(fval,
+                                                               dtype=np.double)
```
bryanwweber: Since you turn these into arrays on the first two lines of this function, why not just do that once?

interfaces/cython/cantera/func1.pyx
```diff
@@ -2,6 +2,7 @@
 # at https://cantera.org/license.txt for license and copyright information.
 
 import sys
+from numbers import Real as _real
```
bryanwweber: This appears not to be used anywhere in this file.

bryanwweber (changes requested):
Thanks @ischoegl! I have a few comments and questions about the Python interface here.

interfaces/cython/cantera/func1.pyx
ischoegl: I changed this to `TabulatedFunction`.

interfaces/cython/cantera/func1.pyx
ischoegl: Not sure about the named `arr1` and `arr2` arguments (leaving `*args`), but :+1: for `method`.

interfaces/cython/cantera/func1.pyx
ischoegl: The `tsiz = 0` exception was necessary to handle empty arrays. I can simplify though ...

interfaces/cython/cantera/func1.pyx
```diff
@@ -83,7 +95,83 @@ cdef class Func1:
...
+cdef class TabulatedFunction(Func1):
+    """
+    A `TabulatedFunction` object representing a tabulated function is defined by
+    sample points and corresponding function values. Inputs are specified either
+    by two iterable objects containing sample point location and function
+    values, or a single array that concatenates those inputs in two rows or
+    columns. Between sample points, values are evaluated based on the optional
+    keyword argument 'method'; options are 'linear' (linear interpolation,
+    default) or 'previous' (nearest previous value). Outside the sample
+    interval, the value at the closest end point is returned.
```
bryanwweber:
```
    keyword argument ``method``; options are ``'linear'`` (linear interpolation,
    default) or ``'previous'`` (nearest previous value). Outside the sample
    interval, the value at the closest end point is returned. ``method`` must be
    provided as a keyword argument.
```

interfaces/cython/cantera/func1.pyx
```diff
@@ -83,7 +95,83 @@ cdef class Func1:
...
+        >>> [t2(v) for v in [-0.5, 0, 0.5, 1.5, 2, 2.5]]
+        [2.0, 2.0, 2.0, 1.0, 0.0, 0.0]
+
+        >>> t3 = TabulatedFunction([0, 1, 2], [2, 1, 0])
+        >>> [t3(v) for v in [-0.5, 0, 0.5, 1.5, 2, 2.5]]
+        [2.0, 2.0, 1.5, 0.5, 0.0, 0.0]
+
+        >>> t4 = TabulatedFunction(np.array([0, 1, 2]), (2, 1, 0))
+        >>> [t4(v) for v in [-0.5, 0, 0.5, 1.5, 2, 2.5]]
+        [2.0, 2.0, 1.5, 0.5, 0.0, 0.0]
```
bryanwweber: Can you provide prose explanations of what these are all doing, and what the difference between them are? 

Technically, the first and second cases here don't match the docstring which states that a single argument must be an array. Not sure there's a better way to phrase that though, just wanted to point it out in case you had any ideas.

interfaces/cython/cantera/func1.pyx
```diff
@@ -83,7 +95,83 @@ cdef class Func1:
...
+                    fval = arr[1, :]
+                else:
+                    raise ValueError("Invalid dimensions: specification of "
+                                     "tabulated function with a single array "
+                                     "requires two rows or columns")
+                self._set_tables(time, fval, stringify(method))
+            else:
+                raise TypeError(
+                    "'TabulatedFunction' must be constructed from "
+                    "a numeric array with two columns")
```
bryanwweber:
```suggestion
                    "a numeric array with two dimensions")
```

interfaces/cython/cantera/test/test_func1.py
```diff
@@ -70,3 +70,53 @@ def test_uncopyable(self):
...
+        with self.assertRaisesRegex(ValueError, 'Invalid dimensions'):
+            ct.TabulatedFunction(np.zeros((3, 3)))
+        with self.assertRaisesRegex(ValueError, 'do not match'):
+            ct.TabulatedFunction(range(2), range(3))
+        with self.assertRaisesRegex(ValueError, 'must not be empty'):
+            ct.TabulatedFunction([], [])
+        with self.assertRaisesRegex(ct.CanteraError, 'monotonically'):
+            ct.TabulatedFunction((0, 1, 0.5, 2), (2, 1, 1, 0))
+        with self.assertRaisesRegex(ct.CanteraError, 'not implemented'):
+            ct.TabulatedFunction((0, 1, 1, 2), (2, 1, 1, 0), method='quadratic')
```
bryanwweber:
```suggestion
            ct.TabulatedFunction((0, 1, 1, 2), (2, 1, 1, 0), method='not implemented')
```
Just in case we ever implement `'quadratic'` one day 😄

bryanwweber (commented):
Some copy editing changes this time around. Thanks for the quick turnaround!

interfaces/cython/cantera/func1.pyx
ischoegl: I guess I use *array* loosely here, as it can be a `numpy.array` (otherwise it's technically a `list` of `list`s?). It also supports tuples ... also not sure, but I believe the examples are clearer now.

ischoegl:
@bryanwweber / @speth ... thanks for the comments/recommendations!

interfaces/cython/cantera/func1.pyx
```diff
@@ -46,34 +46,46 @@ cdef class Func1:
...
-                                    ' a callable object')
-            self.callable = lambda t: k
+                    raise TypeError
+
+            except TypeError:
+                raise TypeError(
+                    "'Func1' objects must be constructed from a number or "
+                    "a callable object") from None
 
+    cpdef void _set_callback(self, c) except *:
```
bryanwweber, line 87: Out of curiosity, what is the purpose of adding this method?

interfaces/cython/cantera/func1.pyx
```diff
@@ -83,7 +95,88 @@ cdef class Func1:
...
+cdef class TabulatedFunction(Func1):
+    """
+    A `TabulatedFunction` object representing a tabulated function is defined by
+    sample points and corresponding function values. Inputs are specified either
+    by two iterable objects containing sample point location and function
+    values, or a single array that concatenates those inputs in two rows or
+    columns. Between sample points, values are evaluated based on the optional
+    keyword argument ``method``; options are ``'linear'`` (linear interpolation,
+    default) or ``'previous'`` (nearest previous value). Outside the sample
+    interval, the value at the closest end point is returned.
```
bryanwweber: One more change here to note the `method` is a kwarg only. I had included some text in my previous comment, but it wasn't obvious 😊

bryanwweber (approved):
Looks good, one minor change, and a curiosity question

interfaces/cython/cantera/func1.pyx
ischoegl: It's impossible to interact with the C++ layer from a regular python function, and the `__cinit__` cannot be overloaded. Hence the need to introduce this helper function.

interfaces/cython/cantera/func1.pyx
ischoegl: I moved it a little further down - it's still there; the docstring mentions that `method` is a keyword argument twice. I.e. I *think* this is resolved?

bryanwweber: I think it's important to specifically clarify that it is a keyword _only_ to avoid confusion, since `*args` will take in any number of given positional arguments, leading to the error of too many arguments here.

ischoegl: :thumbsup: I see what you're saying. Personally, a keyword *is* a keyword (I usually avoid skipped keyword names, although it's a legitimate use - well, I realize that `ct.Solution` has everything defined as kwargs, but I've never seen `infile` actually used). I changed the wording, as the interface indeed requires a kwarg.

interfaces/cython/cantera/func1.pyx
speth: What about the ambiguity of the case where a 2x2 array is provided? I sort of prefer just having the two-argument form. That way, you can use your input if it's either rows or columns. Assuming numpy arrays, that would just be one of
```python
TabulatedFunction(x[0], x[1])
TabulatedFunction(x[:,0], x[:,1])
```

ischoegl: That’s a valid point. I don’t have any objections to just requiring the two argument form. @bryanwweber ... any thoughts?

ischoegl: Going back to Cantera/enhancements#17, I believe there was the `vel_array = np.genfromtxt('velocity-trace.txt')` case, with the trace having two columns. I believe it would make sense to allow for a shorthand
```
TabulatedFunction(vel_array)
```
I.e. I'm tempted to just revert to the original implementation. I also think it may make sense to support something like `[(0, 1), (1, 1.5), (2, 2)]` where tuples are used to specify points.

bryanwweber: I personally prefer the two argument form that @speth suggested. A two column or row array is easy enough for the user to specify as two arguments.

interfaces/cython/cantera/func1.pyx
bryanwweber: I guess the question was really, why is this needed _now_ when it wasn't before? Again, just for my understanding

ischoegl: Previously, there was no `__init__`, only a `__cinit__`. Now that we're starting to overload, some helper functions are needed (`__cinit__` cannot be overloaded).

interfaces/cython/cantera/func1.pyx
ischoegl: Done - two arguments it shall be ...

speth: If a user is coming into this with list of tuples, they can always transform it with something like `TabulatedFunction(*zip(*vel_array))`.

speth (approved):

PR #803, opened 2020-02-10
General cleanup of tests, SCons building, and submodules

bryanwweber:
**Checklist**

- Fix a few conversion script tests that were loading the wrong files
- Have SConstruct check `instRoot` instead of `env['prefix']` to see if this is the source directory

This branch is rebased on top of #796 so I can continue to use it to test the Conda packages


speth:
Can you rebase this on current master and fix the merge conflicts?

bryanwweber:
@speth Done!

speth (commented):
I see the logic of fully specifying the path to the newly-converted input file, but I'm a little confused by a couple of things here.

First, what's wrong with loading the reference file from the Cantera data path? This is presumably what's making it necessary to add another copy of `gri30.xml` to that location?

Second, doesn't this logic also apply to the other sets of tests in `test_convert.py`, i.e. `cti2yamlTest` and `ctml2yamlTest`. Right now, this PR is affects the `ck2cti` and `ck2yaml` tests. And in that case, we definitely wouldn't want to have to copy all of the reference files that are currently in the common data directory to `test/data`.

bryanwweber:
> First, what's wrong with loading the reference file from the Cantera data path? This is presumably what's making it necessary to add another copy of gri30.xml to that location?

If I recall correctly, the reason I added the test data directory explicitly was to avoid loading a file with the same name from the data directory. It may have been the `liquidvapor.yaml` file, which is now fixed by #826. I'm going to temporarily comment out the change there and see if the test failures return in the Conda recipe.

> Second, doesn't this logic also apply to the other sets of tests in `test_convert.py`, i.e. `cti2yamlTest` and `ctml2yamlTest`. Right now, this PR is affects the `ck2cti` and `ck2yaml` tests. 

Yes, I was trying to limit the fix to the one that would solve the problem I was having.

> And in that case, we definitely wouldn't want to have to copy all of the reference files that are currently in the common data directory to `test/data`

Eventually, those files are going to go away, at least my understanding of the plan is that we're going to stop distributing all CTI and XML files with 3.0. So they will have to be copied at some point. I think the change to that line can be deferred until that happens.

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -85,30 +86,35 @@ def checkKinetics(self, ref, gas, temperatures, pressures, tol=1e-8):
...
         self.convert(None, thermo='dummy-thermo.dat', output='dummy-thermo')
-        cti = "ideal_gas(elements='C H', species='dummy-thermo:R1A R1B P1')"
-        gas = ct.Solution(source=cti)
+        if self.ext == ".cti":
+            cti = "ideal_gas(elements='C H', species='dummy-thermo:R1A R1B P1')"
+            gas = ct.Solution(source=cti)
+        elif self.ext == ".yaml":
+            yaml = ("phases:\n- name: gas\n  elements: [C, H]\n  "
+                    "species: [dummy-thermo.yaml/species: [R1A, R1B, P1]]\n  "
+                    "thermo: ideal-gas")
```
speth: This might be a good place to use the YAML flow style.

Also, the `elements` declaration is almost never required in YAML.

speth:
I'm not able to compile the version of Googletest introduced in this PR branch using MSVC 14.0 (aka Visual Studio 2015). I get the following error:
```
cl.exe /Fobuild\ext\googletest\googletest\src\gtest-all.obj /c ext\googletest\googletest\src\gtest-all.cc /EHsc /MD /nologo /D_SCL_SECURE_NO_WARNINGS /D_CRT_SECURE_NO_WARNINGS /O2 /Zi /Fdbuild\ext\googletest\googletest\src\gtest-all.obj.pdb /DGTEST_HAS_PTHREAD=0 /DNDEBUG /Iext\googletest\googletest /Iext\googletest\googletest\include /IC:\src\boost_1_71_0
gtest-all.cc
C:\src\cantera\ext\googletest\googletest\include\gtest/internal/gtest-internal.h(1188): error C2039: 'FlatTupleBase<testing::internal::FlatTuple<bool,bool>,testing::internal::IndexSequence<0,1> >': is not a member of 'testing::internal::FlatTuple<bool,bool>'
C:\src\cantera\ext\googletest\googletest\include\gtest/internal/gtest-param-util.h(747): note: see reference to class template instantiation 'testing::internal::FlatTuple<bool,bool>' being compiled
C:\src\cantera\ext\googletest\googletest\include\gtest/gtest-param-test.h(360): note: see reference to class template instantiation 'testing::internal::ValueArray<bool,bool>' being compiled
```
I haven't been able to find anything searching for this error, and I'm not sure what versions of MSVC this version of Googletest is meant to support.

bryanwweber:
Interestingly, the builds pass with MSVC 14.0 on the Windows GitHub actions builds: https://github.com/Cantera/cantera/pull/775/checks?check_run_id=713201352 Those have VS 2019 with the 14.0 build tools installed.

speth (changes requested):
Most of this seems fine to me, except I think there is a better solution to the issue of compiler intermediates that aren't being cleaned up.

I pushed an alternative commit that fixes this for more compilers and without requiring additional arguments to the test cases to my version of this branch (https://github.com/speth/cantera/commits/fix-python-tests), along with a commit refactoring of how the classes used for building these test cases. I can push these changes to this PR branch if you want to consider them as a friendly amendment.

bryanwweber:
@speth Thanks! Feel free to push the changes for the test file cleanup.

interfaces/cython/cantera/test/test_convert.py
bryanwweber: Done! I included `elements` since the CTI test did as well 🤷‍♂️

speth (approved):

bryanwweber:
Thanks @speth!

PR #805, opened 2020-02-11
SolutionArray/HDF interface for 1D FlameBase

ischoegl:
**Checklist**

Partially addresses “big” project ideas (see [1D Flame Development Ideas](https://github.com/Cantera/cantera/projects/2)):

> Introduce HDF5 or CSV or JSON or YAML or some other output format for 1D flames to replace XML

> Add ability to directly restore 1D flame objects from a saved solution (need to embed info about flame configuration and mechanism). 

- Add `SolutionArray` input/output for 1D `FlameBase` objects (i.e. `FreeFlame`, etc.)
- HDF5 input/output
- Small fixes of `SolutionArray` glitches (i.e. slicing, extra column import)
- Restart via `set_initial_guess`

E.g. change to `cantera/examples/onedim` folder, and run (within `ipython`):
```
In [1]: %run adiabatic_flame.py
[...]
multicomponent flamespeed = 0.727008 m/s
Solution saved to file h2_adiabatic.xml as solution multi.
Solution saved to 'h2_adiabatic.csv'.

In [2]: arr = f.to_solution_array()
  ... : arr
Out[3]: <cantera.composite.SolutionArray at 0x7f9464f827f0>

In [3]: f1 = ct.FreeFlame(gas)
  ... : f1.from_solution_array(arr)

In [4]: all(f.grid == f1.grid)
Out[4]: True

In [5]: all(f.T == f1.T)
Out[5]: True

In [6]: f.to_pandas()
Out[6]: 
      grid  velocity            T   density      X_H2           X_H       X_O      X_O2      X_OH         X_H2O         X_HO2        X_H2O2      X_AR
0     -inf  0.727008   300.000000  1.338612  0.154930  0.000000e+00  0.000000  0.140845  0.000000  0.000000e+00  0.000000e+00  0.000000e+00  0.704225
1    0.000  0.727008   300.000000  1.338612  0.154930  1.407594e-16  0.000000  0.140845  0.000000  0.000000e+00  0.000000e+00  8.663079e-17  0.704225
2    0.006  0.727008   300.000000  1.338612  0.154930  0.000000e+00  0.000000  0.140845  0.000000  0.000000e+00  0.000000e+00  8.422789e-17  0.704225
3    0.012  0.727008   300.000000  1.338612  0.154930  0.000000e+00  0.000000  0.140845  0.000000  0.000000e+00  0.000000e+00  8.866101e-17  0.704225
4    0.015  0.727008   300.000000  1.338612  0.154930  0.000000e+00  0.000000  0.140845  0.000000  1.130882e-16  0.000000e+00  9.622272e-17  0.704225
..     ...       ...          ...       ...       ...           ...       ...       ...       ...           ...           ...           ...       ...
136  0.033  4.148451  1858.587531  0.234589  0.000072  6.253460e-06  0.000071  0.070944  0.001071  1.631748e-01  9.233693e-07  5.341914e-08  0.764660
137  0.036  4.149534  1859.115819  0.234528  0.000064  5.258929e-06  0.000063  0.070960  0.001012  1.632172e-01  8.979551e-07  5.130584e-08  0.764678
138  0.042  4.150244  1859.462329  0.234488  0.000059  4.657906e-06  0.000058  0.070971  0.000972  1.632451e-01  8.831509e-07  4.998669e-08  0.764689
139  0.048  4.150480  1859.577354  0.234474  0.000057  4.467670e-06  0.000057  0.070974  0.000959  1.632544e-01  8.785943e-07  4.956310e-08  0.764693
140  0.060  4.150480  1859.577354  0.234474  0.000057  4.467670e-06  0.000057  0.070974  0.000959  1.632544e-01  8.785943e-07  4.956310e-08  0.764693

[141 rows x 13 columns]

In [7]: f.write_hdf('solution.h5')

In [8]: f2 = ct.FreeFlame(gas)
   ...: f2.set_initial_guess(data='solution.h5')

In [9]: all(f.grid == f2.grid)
Out[9]: True
```


ischoegl:
The currently failing build is fixed by #808 (i.e. I will rebase as soon as it is merged).

ischoegl:
@speth (and @bryanwweber)... I am stopping here for feedback before adding a couple more tweaks (e.g. adding HDF-based `set_initial_guess` to other `FlameBase` objects).

interfaces/cython/cantera/composite.py
```diff
@@ -715,11 +719,12 @@ def restore_data(self, data, labels):
...
         # labels may include calculated properties that must not be restored
-        calculated = self._scalar + self._n_species + self._n_reactions
-        exclude = [l for l in labels
-                   if any([v in l for v in calculated])]
-        extra = {l: list(data[:, i]) for i, l in enumerate(labels)
-                 if l not in exclude}
+        exclude = [lab for lab in labels
+                   if any([lab in self._scalar,
+                           '_'.join(lab.split('_')[:-1]) in self._n_species,
+                           lab.split(' ')[0] in self._n_reactions])]
```
speth, line 741: I think a little more documentation here would be helpful to explain what these patterns are trying to find.

interfaces/cython/cantera/onedim.py
```diff
@@ -3,14 +3,15 @@
...
-from .composite import Solution
+from .composite import Solution, SolutionArray
 import csv as _csv
 from math import erf
 
 
 class FlameBase(Sim1D):
     """ Base class for flames with a single flow domain """
     __slots__ = ('gas',)
+    _extra = ()
```
speth: A comment explaining what this variable is and how it's used in derived classes would be helpful.

interfaces/cython/cantera/onedim.py
```diff
@@ -299,6 +300,147 @@ def write_csv(self, filename, species='X', quiet=True):
         if not quiet:
             print("Solution saved to '{0}'.".format(filename))
 
+    def to_solution_array(self):
+        """Return the solution vector as a Cantera SolutionArray object.
+
+        The SolutionArray has the following ``extra`` entries:
+         * `grid`: grid point positions along the flame [m]
+         * `velocity`: normal velocity [m/s]
+         * `gradient`: tangential velocity gradient [1/s] (if applicable)
```
speth: I understand not wanting to stick to the variable names `u` and `V`, but `gradient` doesn't seem much better. Maybe it would make sense to go with the more verbose `velocity_gradient`?

interfaces/cython/cantera/onedim.py
```diff
@@ -299,6 +300,147 @@ def write_csv(self, filename, species='X', quiet=True):
...
+        extra = {}
+        for e in self._extra:
+            if e == 'grid':
+                extra[e] = self.grid
+            elif e == 'velocity':
+                extra[e] = self.profile(dom, 'u')
+            elif e == 'gradient':
+                extra[e] = self.profile(dom, 'V')
+            else:
+                extra[e] = self.profile(dom, e)
```
speth: I'd suggest ditching the loop and just writing each case as something like
```python
if 'grid' in self._extra:
    extra[e] = self.grid
```

interfaces/cython/cantera/onedim.py
```diff
@@ -299,6 +300,147 @@ def write_csv(self, filename, species='X', quiet=True):
...
+
+        # create solution array object
+        arr = SolutionArray(self.gas, self.flame.n_points, extra=extra)
+
+        # retrieve species concentrations and set states
+        X = []
+        for n in range(self.flame.n_points):
+            self.set_gas_state(n)
+            X.append(self.gas.X)
+        arr.TPX = self.T, self.gas.P, np.vstack(X)
```
speth: Slightly simpler alternative:
```python
for n in range(self.flame.n_points):
    self.set_gas_state(n)
    arr[n].TPY = gas.T, gas.P, gas.Y
```
I suggest `Y` instead of `X` since that's the native state variable.

interfaces/cython/cantera/onedim.py
```diff
@@ -299,6 +300,147 @@ def write_csv(self, filename, species='X', quiet=True):
...
+
+        return arr
+
+    def from_solution_array(self, arr):
+        """Restore the solution vector from a Cantera SolutionArray object.
```
speth: Except when the whole docstring fits on one line, I prefer to just leave the `"""` on a line by itself.

interfaces/cython/cantera/onedim.py
```diff
@@ -299,6 +300,147 @@ def write_csv(self, filename, species='X', quiet=True):
         if not quiet:
             print("Solution saved to '{0}'.".format(filename))
 
+    def to_solution_array(self):
+        """Return the solution vector as a Cantera SolutionArray object.
+
+        The SolutionArray has the following ``extra`` entries:
```
speth: Class names like `SolutionArray` should be quoted with backticks when they appear in Python docstrings so that they will be linked by Sphinx.

interfaces/cython/cantera/onedim.py
```diff
@@ -299,6 +300,147 @@ def write_csv(self, filename, species='X', quiet=True):
...
+        The SolutionArray requires the following ``extra`` entries:
+         * `grid`: grid point positions along the flame [m]
+         * `velocity`: normal velocity [m/s]
+         * `gradient`: tangential velocity gradient [1/s] (if applicable)
+         * `lambda`: radial pressure gradient [N/m^4] (if applicable)
+         * `eField`: electric field strength (if applicable)
+        """
+        # restore grid
+        self.domains[1].grid = arr.grid
+        self._get_initial_solution()
```
speth, line 476: I don't think this call is necessary.

interfaces/cython/cantera/onedim.py
```diff
@@ -299,6 +300,147 @@ def write_csv(self, filename, species='X', quiet=True):
...
+        # restore species profiles
+        X = arr.X
+        for i, spc in enumerate(self.gas.species_names):
+            self.set_profile(spc, xi, X[:, i])
+
+        # restore pressure
+        self.P = arr.P[0]
+
+    def to_pandas(self):
+        """Return the solution vector as a pandas DataFrame.
```
speth: This can be linked directly to the `pandas` docs by writing ``` `pandas.DataFrame` ```.

speth (changes requested):
This is certainly interesting, and a good use of the work that you put into the HDF5 / pandas interface to the `SolutionArray` class. For the Python interface, this seems like it may be the best way to provide a more useful alternative to the XML output format. Ultimately, we still need some output format that can be used directly from C++, but this might save us from having to do anything annoying like adding a dependency on the HDF5 library.

I did have a few questions / comments:

* I'm not sure I understand the need to extend the interface of `FreeFlame.set_initial_guess`. Why is this different from just using `from_solution_array`?
* I think it's counterintuitive that `from_solution_array` only sets data for the 1D domain but does not set any of the boundary values. I'm not sure what the best way to handle this is. While in many cases, these values are the same as the first/last grid points, this is not always true (e.g. in the case of the mass fractions in `BurnerFlame`).
* Is there any way to extend this to include so that the saved data is as complete as what gets written to the XML output files? Some of these properties are per grid point, so the mapping is straightforward, but there are others that don't, e.g. the grid refinement tolerances and the `species_enabled` flag.

interfaces/cython/cantera/onedim.py
ischoegl: It actually is necessary as there’s a segfault otherwise (I realize that this used to work several releases ago). I’m likely putting a band-aid here for something that should be fixed elsewhere.

ischoegl:
> Is there any way to extend this to include so that the saved data is as complete as what gets written to the XML output files? Some of these properties are per grid point, so the mapping is straightforward, but there are others that don't, e.g. the grid refinement tolerances and the `species_enabled` flag.

Absolutely - at least for HDF (and minus the mechanism which imho should be dealt with separately). I mainly wanted to get feedback before addressing minor issues that will take some time and thought to implement.

interfaces/cython/cantera/onedim.py
ischoegl: I opted to use the loop as `_extra` typically matches the C++ *name* of a given `FlameBase` solution variable, i.e. I do not want to hard code except for those instances where the names are replaced (i.e. `grid` which doesn't have a profile, and where nomenclature clashes, i.e. `u` and `V`).

interfaces/cython/cantera/onedim.py
ischoegl: Well - I looked into this and the reason is that the `Sim1D` object doesn't know that the vector sizes have been updated for one of its constituting `Domain1D`'s. There are probably ways to circumnavigate this issue, but calling `_get_initial_solution` here is *much* simpler: i.e. I am suggesting to leave the call ...

interfaces/cython/cantera/onedim.py
ischoegl: Apart from clarifying `dom` as `self.flame` I am leaving this as is (edit: this also simplified after subsequent commits).

interfaces/cython/cantera/onedim.py
ischoegl: `velocity_gradient` imho isn't great either as it could be mistaken for the normal velocity gradient. I initially had `tangential_velocity_gradient`, which is a mouthful. Any alternatives??

ischoegl:
@speth ... I addressed most of your comments (except for nomenclature I believe all are resolved).

Per your questions:

> I'm not sure I understand the need to extend the interface of `FreeFlame.set_initial_guess`. Why is this different from just using `from_solution_array`?

The difference is motivated by restart applications which (1) avoid poor convergence for marginal conditions (i.e. the usual initial guess performs poorly), or (2) quickly fill flame data bases. Based on past experience, it is much quicker to assign a pruned solution from a related condition as an initial guess. Obviously, some small adjustments are needed (e.g. initial temperature), which is where differences arise. 

> I think it's counterintuitive that `from_solution_array` only sets data for the 1D domain but does not set any of the boundary values. I'm not sure what the best way to handle this is. While in many cases, these values are the same as the first/last grid points, this is not always true (e.g. in the case of the mass fractions in `BurnerFlame`).

This is an excellent point. My motivation was mainly *restart* via `set_initial_guess`, but I agree that everything should be restored as anticipated for `from_solution_array` as well. I believe this is best handled by a flag where the default behavior restores boundary conditions, while the `set_initial_guess` route leaves them alone.

ischoegl:
@speth ... quick question regarding storage of simulation settings: I went through parameters that are accessible from Python, and came up with the following list:
```
['transport_model',
 'energy_enabled',
 'soret_enabled',
 'radiation_enabled',
 'ratio',
 'slope',
 'curve',
 'prune',
 'max_time_step_count',
 'max_grid_points',
 'steady_abstol',
 'steady_reltol',
 'transient_abstol',
 'transient_reltol']
```
where for tolerances it is straight-forward to store mode and deviating values to avoid clutter. It is straight-forward to save this in tabular form to the same HDF container that also holds the data.

Things that I anticipate to be missing at the moment (a few require the implementation of property getters):
 * [x] `z_fixed` / `t_fixed` 
 * [x] boundary emissivities
 * [x] radiation `qdot` if `radiation_enabled`
 * [ ] `species_enabled` ... missing in Python - is this relevant for `FlameBase` simulations?
 * [x] time stamp.

ischoegl:
Here is a brief summary about `settings` output to HDF:

```
In [1]: %run adiabatic_flame.py  
[...]
multicomponent flamespeed = 0.727008 m/s
Solution saved to file h2_adiabatic.xml as solution multi.
Solution saved to 'h2_adiabatic.csv'.

In [2]: f.write_hdf('solution.h5')

In [3]: import pandas as pd

In [4]: with pd.HDFStore('solution.h5') as hdf:
    ...:     keys = hdf.keys()
    ...:     

In [5]: keys
Out[5]: ['/data', '/settings']

In [6]: s = pd.read_hdf('solution.h5', key='settings')

In [7]: s
Out[7]: 
    key                             date configuration transport_model  ...  steady_abstol  steady_reltol  transient_abstol  transient_reltol
0  data  Thu, 27 Feb 2020 21:00:23 +0000     FreeFlame           Multi  ...   1.000000e-09         0.0001      1.000000e-11            0.0001

[1 rows x 20 columns]

In [8]: list(s.columns)
Out[8]: 
['key',
 'date',
 'configuration',
 'transport_model',
 'energy_enabled',
 'soret_enabled',
 'radiation_enabled',
 'emissivity_left',
 'emissivity_right',
 'fixed_temperature',
 'ratio',
 'slope',
 'curve',
 'prune',
 'max_time_step_count',
 'max_grid_points',
 'steady_abstol',
 'steady_reltol',
 'transient_abstol',
 'transient_reltol']
```

ischoegl:
@speth ... ~I will stop at this point for further feedback.~ At this point, I believe all that is left is:

1. ~settle on a replacement for the component name `V` (I am advocating for deprecation rather than local renaming - see `velocity` replacing `u`).~ *Edit: I am proposing `vGradient`*... This should eventually resolve the pre-existing band-aids
https://github.com/Cantera/cantera/blob/c16c4b0b0f3de6bb9e5366db5d6154ab2a9e0189/interfaces/cython/cantera/onedim.py#L341-L342
2. follow up on discussion of restart via `set_initial_guess`

**PS:** ~Not sure what's up with the code coverage (again) ... the issue is not with files that I touched ...~

interfaces/cython/cantera/onedim.py
ischoegl: Resolving this via the suggestion `vGradient`; this follows the CamelCase precedent `eField`.

ischoegl:
Rebased to current master, and took care of a *small project idea* (renaming `Inlet1D.h` as well as the `Bdry1D` class).

ischoegl:
@speth ... your comment:

> Is there any way to extend this to include so that the saved data is as complete as what gets written to the XML output files? Some of these properties are per grid point, so the mapping is straightforward, but there are others that don't, e.g. the grid refinement tolerances and the `species_enabled` flag.

prompted me to add this feature to this PR (this includes restoring of settings from HDF data). Regarding the `species_enabled` flag, note that the interface handles `FlameBase`, which is a subset of `Domain1D` where the flag is not being used. It's certainly possible to extend this idea to more generic cases, but I'd prefer to leave this for another PR.

I'd appreciate if you could have a look before this gets larger. Most of what I can think of is implemented at this point; what I anticipate beyond is mostly implementing `set_initial_guess` for `FlameBase` classes other than `FreeFlame`, and some tinkering with examples.

**PS:** I will not add anything beyond this point until I hear back.

~**PPS:** Regarding the `Boundary1D.h` diff that doesn't show up in the overview, please refer to 6ea933f8ab2cacee07158f0e498c0bd7048db213 (which renames `Brdy1D` to `Boundary1D`, etc.), ec63aa2a9e45be7c06661c68719509c00d8d5189 (handling of deprecation) and f61592f38ba58bcd042bbcbc0c7684b011c4ae18 (which replaces `doublereal` by `double`; since I'm touching the entire file, I felt like this was an appropriate opportunity).~

ischoegl:
**Update:** While I'm not adding anything, I moved unrelated renaming changes (`Inlet1D.h` / `Bdry1D`) to a separate PR #815 to avoid clutter.

include/cantera/oneD/Sim1D.h
```diff
@@ -129,15 +129,22 @@ class Sim1D : public OneDim
     int refine(int loglevel=0);
 
     //! Add node for fixed temperature point of freely propagating flame
-    int setFixedTemperature(doublereal t);
+    int setFixedTemperature(double t);
+
+    //! Return temperature at the point used to fix the flame location
+    double getFixedTemperature();
```
speth: For methods that simply return the specified value (rather than say, filling values in an array that is passed as an argument), we usually omit the `get` prefix.

include/cantera/oneD/StFlow.h
```diff
@@ -215,7 +226,14 @@ class StFlow : public Domain1D
...
 
-    virtual void setFixedPoint(int j0, doublereal t0) {}
+    /*!
+     * @deprecated To be removed after Cantera 2.5.
+     */
+    virtual void setFixedPoint(int j0, doublereal t0) {
+        // this does nothing and does not appear to be overloaded
+        warn_deprecated("StFlow::setFixedPoint",
+                        "To be removed after Cantera 2.5.");
+    }
```
speth, line 236: Interesting find.

include/cantera/oneD/StFlow.h
```diff
@@ -198,13 +198,24 @@ class StFlow : public Domain1D
...
      * calculation.
      */
-    void setBoundaryEmissivities(doublereal e_left, doublereal e_right);
+    void setBoundaryEmissivities(double e_left, double e_right);
+
+    //! Return emissivitiy at left boundary
+    double getLeftEmissivity() const { return m_epsilon_left; }
+
+    //! Return emissivitiy at right boundary
+    double getRightEmissivity() const { return m_epsilon_right; }
```
speth: Likewise, would suggest simply `leftEmissivity` and `rightEmissivity` here.

interfaces/cython/cantera/onedim.py
```diff
@@ -162,15 +171,45 @@ def T(self):
...
+        .. deprecated:: 2.5
+
+             To be deprecated with version 2.5, and removed thereafter.
+             Replaced by property `velocity`.
         """
-        return self.profile(self.flame, 'u')
+        warnings.warn("To be removed after Cantera 2.5. "
+                      "Replaced by property 'velocity'",
+                      DeprecationWarning)
+        return self.profile(self.flame, 'velocity')
```
speth, line 263: I think these deprecation warnings need to specify the name of the deprecated method. It isn't particularly clear from the message printed at runtime. For example:
```
path/to/onedim.py:180: DeprecationWarning: To be removed after Cantera 2.5. Replaced by property 'velocity'
  warnings.warn("To be removed after Cantera 2.5. "
```

src/oneD/StFlow.cpp
```diff
@@ -574,9 +574,9 @@ string StFlow::componentName(size_t n) const
 {
     switch (n) {
     case 0:
-        return "u";
+        return "velocity";
     case 1:
-        return "V";
+        return "vGradient";
```
speth: Would it be better to just use `tangential_velocity_gradient` here as well? I think the consistency would be beneficial, and the name isn't used so often that the verbosity becomes cumbersome.

interfaces/cython/cantera/onedim.pyx
```diff
@@ -532,6 +562,10 @@ cdef class IdealGasFlow(_FlowBase):
         gas = getIdealGasPhase(thermo)
         self.flow = new CxxStFlow(gas, thermo.n_species, 2)
 
+    def component_index(self, str name):
+        """Index of the component with name 'name'"""
+        return self.flow.componentIndex(stringify(name))
+
```
speth: I think this implementation becomes unnecessary if we make `Domain1D::componentIndex` virtual. It's actually fairly surprising for it to have a non-virtual override.

interfaces/cython/cantera/onedim.py
```diff
@@ -279,6 +318,360 @@ def write_csv(self, filename, species='X', quiet=True):
...
+                extra[e] = extra[e][:-1]
+
+        # create solution array object
+        arr = SolutionArray(self.gas, n_arr, extra=extra)
+
+        # add left boundary
+        self.gas.TPY = left.T, self.P, left.Y
+        arr[0].TPY = self.gas.TPY
+        arr._extra['grid'][0] = -np.inf
+        arr._extra['velocity'][0] = left.mdot/self.gas.density
```
speth, line 433: I hadn't realized that there wasn't a cleaner way to modify elements of `extra` arrays in the `SolutionArray` class. That may be worth revisiting at some point.

interfaces/cython/cantera/examples/onedim/adiabatic_flame.py
```diff
@@ -6,6 +6,10 @@
 """
 
 import cantera as ct
+try:
+    import pandas as pd
```
speth, line 10: I think this should also try importing `tables`, since that is also a requirement for `write_hdf` to work.

interfaces/cython/cantera/onedim.py
```diff
@@ -744,6 +1164,7 @@ def __init__(self, gas, grid=None, width=None):
 class CounterflowDiffusionFlame(FlameBase):
     """ A counterflow diffusion flame """
     __slots__ = ('fuel_inlet', 'flame', 'oxidizer_inlet')
+    _extra = ('grid', 'velocity', 'gradient', 'lambda')
```
speth: The third element should be `vGradient`, right? (or `tangential_velocity_gradient` pending the suggestion on that name)

interfaces/cython/cantera/onedim.py
```diff
@@ -1022,6 +1443,7 @@ def mixture_fraction(self, m):
 class ImpingingJet(FlameBase):
     """An axisymmetric flow impinging on a surface at normal incidence."""
     __slots__ = ('inlet', 'flame', 'surface')
+    _extra = ('grid', 'velocity', 'gradient', 'lambda')
```
speth: Same comment as before regarding `gradient`

interfaces/cython/cantera/onedim.py
```diff
@@ -1095,13 +1517,14 @@ def set_initial_guess(self, products='inlet'):
...
-        self.set_profile('u', locs, [u0, 0.0])
-        self.set_profile('V', locs, [0.0, 0.0])
+        self.set_profile('velocity', locs, [u0, 0.0])
+        self.set_profile('vGradient', locs, [0.0, 0.0])
 
 
 class CounterflowPremixedFlame(FlameBase):
     """ A premixed counterflow flame """
     __slots__ = ('reactants', 'flame', 'products')
+    _extra = ('grid', 'velocity', 'gradient', 'lambda')
```
speth: Same comment as before regarding `gradient`.

interfaces/cython/cantera/onedim.py
```diff
@@ -1195,6 +1618,7 @@ class CounterflowTwinPremixedFlame(FlameBase):
     shooting into each other.
     """
     __slots__ = ('reactants', 'flame', 'products')
+    _extra = ('grid', 'velocity', 'gradient', 'lambda')
```
speth: Same comment as before regarding `gradient`.

src/oneD/StFlow.cpp
```diff
@@ -610,6 +620,8 @@ size_t StFlow::componentIndex(const std::string& name) const
                 return n;
             }
         }
+        throw CanteraError("StFlow1D::componentIndex",
+                           "no component named " + name);
```
speth, line 624: With this addition, you can remove the `return npos;` statement below.

speth (changes requested):
I was looking at the structure of the HDF5 files that this generates, and I'm kind of disappointed by what pandas is doing. They've layered their own structure on top of the HDF5 layout in a way that makes it basically useless except for being read by pandas. If you open the file using pytables or another library (or if you are working in a different programming language) the resulting structure is a total mess.

I realize that this also applies to the HDF5 files generated from the SolutionArray class, and I think that we should reconsider that format. All of this seems basically fine as a way of generating a pandas object / file, but this doesn't generate what I would expect from a `write_hdf` method.

I do like the approach to capturing all of the "settings" that are necessary to fully recreate a `FlameBase` object. We still need some lower-level interface for at least saving 1D flame output from C++ before we can remove the XML output format, but that's a different project.

src/oneD/StFlow.cpp
ischoegl: @speth, ... this is what I had initially, but I'm not really happy with that nomenclature, either ... see docstring 
```
cdef class IdealGasFlow(_FlowBase):
"""
    An ideal gas flow domain. Functions `set_free_flow` and
    `set_axisymmetric_flow` can be used to set different type of flow.

    For the type of axisymmetric flow, the equations solved are the similarity
    equations for the flow in a finite-height gap of infinite radial extent.
    The solution variables are:

    *velocity*
        axial velocity
    *tangential_velocity_gradient*
        radial velocity divided by radius
    *T*
        temperature
    *lambda*
        (1/r)(dP/dr)
    *Y_k*
        species mass fractions
[...]
```
i.e. the property name is a misnomer as the quantity is actually just normalized, and not a gradient? As an aside, `Inlet1D` uses `spread_rate` ... which may be more accurate?

interfaces/cython/cantera/onedim.py
ischoegl: Agreed.

interfaces/cython/cantera/examples/onedim/adiabatic_flame.py
ischoegl: ~Not necessarily. Someone may just want to use the `pandas` interface. For HDF output, `pandas` raises an appropriate error.~

interfaces/cython/cantera/onedim.pyx
ischoegl: Funny ... `StFlow::componentIndex` was already `virtual`, so I guess that was just a minor glitch.

src/oneD/StFlow.cpp
ischoegl: Marking this as resolved by proposing `spread_rate`.

ischoegl:
> @speth ... I was looking at the structure of the HDF5 files that this generates, and I'm kind of disappointed by what pandas is doing. They've layered their own structure on top of the HDF5 layout in a way that makes it basically useless except for being read by pandas. If you open the file using pytables or another library (or if you are working in a different programming language) the resulting structure is a total mess.

I completely agree. While I was considering a `h5py` route, I ended up opting for the `pandas` route anyways, as `SolutionArray` doesn't have a C++ equivalent, and I do consider the `pandas` generated HDF output more natural within a Python environment (i.e. the exact data structure is not apparent to the user, and there are benefits of a light-weight interface). 

**PS:** I do realize that there's a benefit of writing output in C++ but that runs into a larger discussion about why `SolutionArray` is not defined in C++ and thus goes far beyond this PR.

ischoegl:
Finally, is there a need to expand *restart* beyond `FreeFlame` (see 8e5bfd2b683edd175d48b77a46fd33cf3c02ae1d)? 

**Edit:** remaining restarts are now added.

ischoegl:
@speth ... stopping again for feedback: I think this should be close.

Ad HDF: I’m overall open for a discussion on revisiting the HDF structure, but view it as a separate issue. The implementation is easily changed in a follow-up if so desired, as storage ‘under the hood’ is not apparent to the user. None of the interfaces/methods introduced here (or in #680 or #687) would need to be changed if the HDF data structure were to be replaced. It hasn’t made it to stable yet so it’s not too late to be amended before 2.5.0.

ischoegl:
@speth ... stopping again for feedback: I think this should be close (after yet another update).

Ad HDF: I’m overall open to a discussion on revisiting the HDF structure, but view it as a separate issue. The implementation is easily changed in a follow-up if so desired, as storage ‘under the hood’ is not apparent to the user. None of the interfaces/methods introduced here (or in #680 or #687) would need to be changed if the HDF data structure were to be replaced. It hasn’t made it to stable yet so it’s not too late to be amended before 2.5.0.

interfaces/cython/cantera/onedim.pyx
```diff
@@ -958,7 +956,13 @@ cdef class Sim1D:
         flow_domains = [D for D in self.domains if isinstance(D, _FlowBase)]
         zmin = [D.grid[0] for D in flow_domains]
         zmax = [D.grid[-1] for D in flow_domains]
-        nPoints = [len(flow_domains[0].grid), 12, 24, 48]
+
+	# 'data' entry is used for restart
```
ischoegl: replace tab.

ischoegl:
Rebased as newly merged #816 pre-empted fixes within this PR.

PS: one potential addition would be to replace `u` and `V` by `velocity` and `spread_rate` in various CSV output scenarios for consistency. ~I am somewhat reluctant to touch this without prior discussion though.~

PPS: the existing `FlameBase.write_csv` routines could be re-routed through `SolutionArray.write_csv` (see last commit). The advantage is to have a uniform interface that allows for re-importing of (reformatted) CSV output, a (minor) disadvantage is that units are lost from the column headers. For older Python versions, the order of onedim-specific columns may not be preserved, see #832.

interfaces/cython/cantera/examples/onedim/adiabatic_flame.py
ischoegl: Reconsidering ... for the example, it does make sense to change (to avoid confusion).

On the other hand, anticipating that HDF output will be superseded by a resolution of #830, I am marking this as resolved within this PR.

interfaces/cython/cantera/onedim.py
```diff
@@ -63,6 +72,68 @@ def get_refine_criteria(self):
         """
         return super().get_refine_criteria(self.flame)
 
+    def set_initial_guess(self, *args, **kwargs):
+        """
+        Set the initial guess for the solution, and load restart data if
+        provided. Derived classes extend this function to set approximations
+        for the temperature and composition profiles.
+        """
```
speth, line 91: Since this is the place where the `data` and `key` arguments are actually used, I would suggest listing those arguments explicitly here, along with `*args` and `**kwargs` to capture any other arguments. I would also like to see the documentation for the `data` and `key` arguments here, rather than duplicated in the 6 other implementations of `set_initial_guess`, with say a one-sentence mention and link to `FlameBase.set_initial_guess`.

interfaces/cython/cantera/onedim.py
```diff
@@ -63,6 +72,68 @@ def get_refine_criteria(self):
...
+
+                # adjust velocities
+                if self.flame.flow_type != "Free Flame":
+                    self.gas.TPY = left.T, self.P, left.Y
+                    u0 = left.mdot/self.gas.density
+                    arr.velocity *= u0 / arr.velocity[0]
+
+            self.from_solution_array(arr, restore_boundaries=False)
+        else:
+            return None
```
speth: The `else` clause here is unnecessary.

interfaces/cython/cantera/onedim.py
```diff
@@ -63,6 +72,68 @@ def get_refine_criteria(self):
...
 
+    def set_initial_guess(self, *args, **kwargs):
+        """
+        Set the initial guess for the solution, and load restart data if
+        provided. Derived classes extend this function to set approximations
+        for the temperature and composition profiles.
+        """
+        self._set_initial_guess(*args, **kwargs)
+        data = kwargs.get('data')
+        if data:
```
speth: Given that the rest of the method is in this `if` block, I would suggest writing:
```py
if data is None:
    return
<rest of method>
```
which lets you dedent everything that follows by a level.

ischoegl:
This is odd ... the previous build failure on Windows was due to `SolutionArray.write_csv` adding two newlines (imho a bug, which is now fixed). I cannot reproduce the current build error on my own machine (which runs ubuntu bionic, rather than xenial)

interfaces/cython/cantera/onedim.py
```diff
@@ -162,15 +232,45 @@ def T(self):
...
-        return self.profile(self.flame, 'V')
+        warnings.warn("Property 'V' to be removed after Cantera 2.5. "
+                      "Replaced by property 'spread_rate'",
+                      DeprecationWarning)
+        return self.profile(self.flame, 'spread_rate')
+
+    @property
+    def spread_rate(self):
+        """
+        Array containing the radial velocity divided by radius [1/s] at each point.
```
speth: I'm think the name `spread_rate` is a good choice, but I would prefer to retain the description as the "tangential velocity gradient" rather than "radial velocity divided by radius", in anticipation of supporting the planar geometry at some point in the future.

interfaces/cython/cantera/onedim.py
```diff
@@ -279,6 +380,344 @@ def write_csv(self, filename, species='X', quiet=True):
...
+         * ``eField``: electric field strength (if applicable)
+        """
+        # create extra columns
+        extra = {}
+        for e in self._extra:
+            if e == 'grid':
+                val = self.grid
+            else:
+                val = self.profile(self.flame, e)
+            extra[e] = np.hstack([np.nan, val, np.nan])
```
speth, line 410: I can't remember whether this had been my suggestion for storing the boundary values, but I'm wondering if it would be easier to store them as part of the "settings" rather than by introducing these extra first and last points with fictitious locations. I think that would make the resulting `SolutionArray` easier to use, as well as simplify handling of the data if users try to use it somewhere that doesn't handle the `inf` and `nan` values gracefully.

interfaces/cython/cantera/onedim.py
```diff
@@ -279,6 +380,344 @@ def write_csv(self, filename, species='X', quiet=True):
...
+        for tname, tol in tols.items():
+            # add mode (most frequent tolerance setting)
+            values, counts = np.unique(tol, return_counts=True)
+            ix = np.argmax(counts)
+            out.update({tname: values[ix]})
+
+            # add values deviating from mode
+            ix = np.logical_not(np.isclose(tol, values[ix]))
+            out.update({'{}_{}'.format(tname, c)
+                        for c, t in zip(comp[ix], tol[ix])})
```
speth, line 687: This feature doesn't seem to work. If I create a flame and call:
```py
f.flame.set_steady_tolerances(T=(2e-7, 4e-8))
```
I get the error:
```
ValueError: dictionary update sequence element #0 has length 15; 2 is required
```
If I change this to
```py
            out.update({'{}_{}'.format(tname, c): t
                        for c, t in zip(comp[ix], tol[ix])})
```
It outputs the relevant tolerance in what I think is the intended fashion, but restoring the settings this way doesn't appear to have any effect.

interfaces/cython/cantera/onedim.py
```diff
@@ -279,6 +380,344 @@ def write_csv(self, filename, species='X', quiet=True):
...
+    @settings.setter
+    def settings(self, s):
+        # simple setters
+        attr = {'transport_model',
+                'energy_enabled', 'soret_enabled', 'radiation_enabled',
+                'fixed_temperature',
+                'max_time_step_count', 'max_grid_points'}
+        attr = attr & set(s.keys())
+        for key in attr:
+            self.__setattr__(key, s[key])
```
```suggestion
            setattr(self, key, s[key])
```

interfaces/cython/cantera/onedim.pyx
```diff
@@ -467,7 +467,26 @@ cdef class _FlowBase(Domain1D):
...
-        self.flow.setBoundaryEmissivities(e_left, e_right)
+        """
+        .. deprecated:: 2.5
+
+             To be deprecated with version 2.5, and removed thereafter.
+             Replaced by property `boundary_emissivities`.
+        """
+        warnings.warn("To be removed after Cantera 2.5. "
+                      "Replaced by property 'boundary_emissivities'",
+                      DeprecationWarning)
```
speth: The warning message should include the name of the deprecated method.

interfaces/cython/cantera/onedim.pyx
```diff
@@ -830,11 +868,9 @@ cdef class Sim1D:
         def __set__(self, nmax):
             self.sim.setMaxTimeStepCount(nmax)
 
-    def set_initial_guess(self, *args, **kwargs):
+    def _set_initial_guess(self, *args, **kwargs):
         """
-        Set the initial guess for the solution. Derived classes extend this
-        function to set approximations for the temperature and composition
-        profiles.
+        Store arguments for initial guess and prepare storage for solution.
```
speth, line 873: Why does this need to be renamed? It could still be called from `FlameBase.set_initial_guess` as `super().set_initial_guess`.

interfaces/cython/cantera/onedim.pyx
```diff
@@ -1124,8 +1167,34 @@ cdef class Sim1D:
...
         propagating flame.
+
+        .. deprecated:: 2.5
+
+             To be deprecated with version 2.5, and removed thereafter.
+             Replaced by property `fixed_temperature`.
+        """
+        warnings.warn("To be removed after Cantera 2.5. "
+                      "Replaced by property 'fixed_temperature'",
+                      DeprecationWarning)
```
speth, line 1178: Warning method should include the name of the deprecated method.

interfaces/cython/cantera/test/test_onedim.py
```diff
@@ -157,6 +157,45 @@ def solve_multi(self):
 
         self.assertEqual(self.sim.transport_model, 'Multi')
 
+    def test_flow_type(self):
+        # Solve with the energy equation disabled
```
speth: Comment doesn't seem to be accurate

src/oneD/Sim1D.cpp
```diff
@@ -511,8 +511,34 @@ int Sim1D::setFixedTemperature(doublereal t)
...
+            t_fixed = d->m_tfixed;
+            break;
+        }
+    }
+    return t_fixed;
+}
+
+double Sim1D::fixedTemperatureLocation()
+{
+    double z_fixed = -1.;
```
speth: Is -1 the best value to return if no fixed temperature location has been set? Should this return NaN instead? Or raise an exception? There's a similar question for the temperature, but at least -1 is clearly invalid there.

speth (changes requested):

src/oneD/Sim1D.cpp
ischoegl: opting for NaN without raising an exception.

interfaces/cython/cantera/onedim.py
ischoegl: Good catch; the error is fixed. The issue that the settings cannot be set is due to a pre-existing bug (see #833).

interfaces/cython/cantera/onedim.py
ischoegl: You mentioned that restoring of boundaries should be feasible:

> I think it's counterintuitive that `from_solution_array` only sets data for the 1D domain but does not set any of the boundary values. I'm not sure what the best way to handle this is. While in many cases, these values are the same as the first/last grid points, this is not always true (e.g. in the case of the mass fractions in `BurnerFlame`).

I chose this solution as this appeared to be the most natural way to implement this feature within `pandas` HDF. The `h5py` route is significantly more flexible, so I'd suggest to table this for the moment, and revisit this when addressing #830.

**PS:** I now added restoring from CSV output also (since it is now compatible with `SolutionArray`), which may play into this discussion.

speth (approved):

PR #810, opened 2020-02-17
[Kinetics] Move heat_release_rate to Python Kinetics class.

ischoegl:
While `heat_release_rate` and `heat_production_rates` are defined for 1D objects, no corresponding calls are available for 0D simulations. Moving the definition to `kinetics.pyx` enables these methods to be used for all `Solution` (or `SolutionArray`) objects.

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/kinetics.pyx
+++ b/interfaces/cython/cantera/kinetics.pyx
@@ -365,6 +365,26 @@ cdef class Kinetics(_SolutionBase):
         def __get__(self):
             return get_reaction_array(self, kin_getDeltaSSEntropy)
 
+    property heat_release_rate:
+        """
+        Get the total volumetric heat release rate [W/m^3].
+        """
+        def __get__(self):
+            return - np.sum(self.partial_molar_enthalpies *
+                            self.net_production_rates, 0)
+
+    property heat_production_rates:
+        """
+        Get the volumetric heat production rates [W/m^3] on a per-reaction
+        basis. The sum over all reactions results in the total volumetric heat
+        release rate.
+        Example: C. K. Law: Combustion Physics (2006), Fig. 7.8.6
+
+        >>> gas.heat_production_rates[1]  # heat production rate of the 2nd reaction
+        """
+        def __get__(self):
+            return - self.net_rates_of_progress * self.delta_enthalpy
+
 
 cdef class InterfaceKinetics(Kinetics):
     """
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -430,6 +430,8 @@ class SolutionArray:
         'critical_temperature', 'critical_pressure', 'critical_density',
         'P_sat', 'T_sat', 'isothermal_compressibility',
         'thermal_expansion_coeff', 'electric_potential',
+        # From Kinetics
+        'heat_release_rate',
         # From Transport
         'viscosity', 'electrical_conductivity', 'thermal_conductivity',
     ]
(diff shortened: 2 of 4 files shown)


interfaces/cython/cantera/kinetics.pyx
```diff
@@ -365,6 +365,26 @@ cdef class Kinetics(_SolutionBase):
...
+                            self.net_production_rates, 0)
+
+    property heat_production_rates:
+        """
+        Get the volumetric heat production rates [W/m^3] on a per-reaction
+        basis. The sum over all reactions results in the total volumetric heat
+        release rate.
+        Example: C. K. Law: Combustion Physics (2006), Fig. 7.8.6
+
+        >>> gas.heat_production_rates[2]  # heat production rate of the 2nd reaction
```
bryanwweber: I believe this should be the third reaction, or else the index should be 1

ischoegl: Good catch - you are correct! I had just cut&paste what was there already and had not noticed this issue.

ischoegl: fixed.

bryanwweber (commented):
This seems like a good generalization of the feature. I'm not sure why it wouldn't have been added to `Solution` in the first place, so I'm a little hesitant to approve this without thinking a bit. Can a user still do
```
f.heat_release_rate
```
for a flame simulation? If not, that should be added back, potentially by reference to the new methods.

ischoegl:
> Can a user still do `f.heat_release_rate` for a flame simulation?

absolutely (I preserved that feature) - it is done in this line: https://github.com/Cantera/cantera/blob/511f96df535e32d68e472bc04b9e554eb75af408/interfaces/cython/cantera/onedim.py#L339

I likewise was surprised that these methods were missing from the general interface, which is why I am proposing to add them.

interfaces/cython/cantera/kinetics.pyx
```diff
@@ -365,6 +365,26 @@ cdef class Kinetics(_SolutionBase):
...
+        """
+        Get the volumetric heat production rates [W/m^3] on a per-reaction
+        basis. The sum over all reactions results in the total volumetric heat
+        release rate.
+        Example: C. K. Law: Combustion Physics (2006), Fig. 7.8.6
+
+        >>> gas.heat_production_rates[1]  # heat production rate of the 2nd reaction
+        """
+        def __get__(self):
+            return - self.net_rates_of_progress * self.delta_standard_enthalpy
```
speth: Shouldn't this use `delta_enthalpy` instead of `delta_standard_enthalpy`? They're the same in the ideal gas case, but otherwise I think this would be inconsistent with the implementation of `heat_release_rate`.

speth (changes requested):

speth:
I don't think it's necessary to second-guess the original implementation too much. It served the purpose for which it was intended, but the change in this PR is clearly a useful generalization.

You can find where this was added with `git blame` -- it came from commit 6b5b8e7735. There's no PR or discussion to go along with it since this is from the old days of Google Code.

interfaces/cython/cantera/kinetics.pyx
ischoegl: This was based on a simple cut&paste job :man_shrugging: ... I'm used to gas phase chemistry where I typically define everything based on standard enthalpies, so what's there is consistent with my understanding (see also my unit test). Of course, I may be missing something as not everything is (ideal) gas phase ...

ischoegl: Be it as it may, I changed it ... (reading from context, the proposed change makes sense).

speth: In the case of a real gas, using the "standard" enthalpy (per Cantera's definition, however problematic it may be) would miss any impact of the interaction terms. So from a consistency standpoint, I think you want to also use the full version of the enthalpy (`partial_molar_enthalpies`) in your test case.

bryanwweber (approved):

interfaces/cython/cantera/kinetics.pyx
ischoegl: Done.

speth (approved):

PR #811, opened 2020-02-19
Add get/set equivalenceRatio/mixtureFraction functions

g3bk47:

- bring get/set equivalence ratio functions to C++ interface
- add get/set (Bilger) mixture fraction function and stoichiometric air to fuel ratio functions to C++ and python interface (mixture fraction is commonly used in combustion modeling and has been requested in the past https://groups.google.com/forum/#!searchin/cantera-users/bilger|sort:date/cantera-users/VenkSlGjnb4/PRFCLKcJBAAJ)
- fix current behavior of get_equivalence_ratio: it is now consistent for arbitrary fuel and oxidizer compositions and is independent of reaction progress (e.g. it is the same in the unburnt and burnt state).

In the C++ interface, all functions take the fuel and oxidizder compositions as const double*, vector_fp, std::string and compositionMap. All functions have a "*_X" and "*_Y" suffix version for providing the fuel and oxidizer composition either as mole or mass fractions. The "getEquivalenceRatio" function has a special case where the fuel and oxidizer composition is not specified: this computes the equivalence ratio purely from the elemental composition.

At the moment, get_equivalence_ratio and set_equivalence_ratio functions from the python interface default to the *_X versions to keep the current behaviour the same.

For further discussion, see:
https://github.com/Cantera/enhancements/issues/28
https://groups.google.com/forum/embed/#!topic/cantera-users/owoAebI6rr8
https://groups.google.com/forum/embed/#!topic/cantera-users/gs-wHjo18Do

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,7 +18,7 @@ Steven DeCaluwe (@decaluwe), Colorado School of Mines
 Vishesh Devgan (@vdevgan)
 Thomas Fiala (@thomasfiala), Technische Universität München
 David Fronczek
-@g3bk47
+Thorsten Zirwes, Karlsruhe Institute of Technology
 Matteo Giani (@MarcDuQuesne)
 Dave Goodwin, California Institute of Technology
 John Hewson (@jchewson), Sandia National Laboratory
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,7 +18,6 @@ Steven DeCaluwe (@decaluwe), Colorado School of Mines
 Vishesh Devgan (@vdevgan)
 Thomas Fiala (@thomasfiala), Technische Universität München
 David Fronczek
-Thorsten Zirwes, Karlsruhe Institute of Technology
 Matteo Giani (@MarcDuQuesne)
 Dave Goodwin, California Institute of Technology
 John Hewson (@jchewson), Sandia National Laboratory
@@ -51,3 +50,4 @@ Laurien Vandewalle (@lavdwall)
 Bryan Weber (@bryanwweber), University of Connecticut
 Armin Wehrfritz (@awehrfritz)
 Richard West (@rwest), Northeastern University
+Thorsten Zirwes (@g3bk47), Karlsruhe Institute of Technology
--- a/include/cantera/thermo/ThermoPhase.h
+++ b/include/cantera/thermo/ThermoPhase.h
@@ -1401,6 +1401,8 @@ public:
      * @internal
      */
     virtual MultiSpeciesThermo& speciesThermo(int k = -1);
+    
+    virtual const MultiSpeciesThermo& speciesThermo(int k = -1) const;
 
     /**
      * @internal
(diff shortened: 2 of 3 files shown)


ischoegl:
@g3bk47 ... Thank you for finishing what you had posted elsewhere. While I do not want to pre-empt what others will add, I have a brief suggestion: would it make sense to merge the three different versions of Python calls to a single method that takes a keyword argument, i.e.
```python
gas.set_equivalence_ratio(1, fuel, oxidizer, basis='mole')
gas.set_equivalence_ratio(1, fuel, oxidizer, basis='mass')
```
with the default being `mole`, instead of
```python
gas.set_equivalence_ratio(1, fuel, oxidizer)
gas.set_equivalence_ratio_X(1, fuel, oxidizer)
gas.set_equivalence_ratio_Y(1, fuel, oxidizer)
```
etc.? Imho this would cut back on some redundancies and facilitate long-term maintenance.

g3bk47:
I introduced the basis parameter in the python interface. I think it is a good idea as it keeps the interface cleaner.

ischoegl:
I'm glad you liked the suggestion. On further thought - the default behavior should likely be governed by the `ThermoPhase.basis` property as done elsewhere. Sorry I didn't think of this earlier. 

As another (minor) comment ... the usual Cantera-preferred convention is to omit keyword arguments if they use the default value (see your example). I can provide further feedback later on if you'd like.

g3bk47:
I removed the default "basis" argument from the example. Regarding the thermo_basis: I'm not sure if this expresses the same thing: from how I understand it, the thermo_basis is a property of the thermo object and affects how intensive properties are returned (https://github.com/Cantera/cantera/blob/01a4e904be2aa3bfbccb594c93ee0916f6ce3a6c/interfaces/cython/cantera/thermo.pyx#L365-L369). The "basis" parameter in the new python functions expresses how the fuel and oxidizer compositions in the arguments should be interpreted regardless of how the thermo object is configured. Feel free to give any feedback.

ischoegl:
Re: `basis` ... the `ThermoPhase.basis` specifies the default setting for various thermodynamic properties, and indeed accomplishes a similar task. My suggestion to use this setting as the default for your setters/getters is based on the scenario where a user changes the default basis to `'mass'` and subsequently calls `set_equivalence_ratio` (or similar) without using the `basis` keyword. In this case, I believe the expected behavior would be that the `basis` argument default follows `ThermoBase.basis`. Obviously, a user can always specify a basis that does not match the default by using the keyword. It's admittedly a really minor issue.

I will probably go through your PR in more detail before long, but will leave this as is for the time being so others can comment as well.

bryanwweber:
@g3bk47 Any reason to close this pull request? It looks like you completely overwrote these changes with your new pull request.

g3bk47:
Sorry, this was on accident. Should I create this as a new pull request again?

bryanwweber:
Yes, please do. You'll need to put the changes on a branch other than `master`. Let me know if you need help to do that.

PR #814, opened 2020-02-28
Simple fix for Sundials 5.1 compatibility

band-a-prend:
**Checklist**

Fixes #

The Sundials 5.1 has no any changes in CVODES API in comparison with sundials 4.0, 4.1, 5.0 therefore the compatibility configuration fix is trivial.

```
*****************************
***    Testing Summary    ***
*****************************

Tests passed: 1046
Up-to-date tests skipped: 0
Tests failed: 0

*****************************
```

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1108,7 +1108,7 @@ if env['system_sundials'] == 'y':
 
     # Ignore the minor version, e.g. 2.4.x -> 2.4
     env['sundials_version'] = '.'.join(sundials_version.split('.')[:2])
-    if env['sundials_version'] not in ('2.4','2.5','2.6','2.7','3.0','3.1','3.2','4.0','4.1','5.0'):
+    if env['sundials_version'] not in ('2.4','2.5','2.6','2.7','3.0','3.1','3.2','4.0','4.1','5.0','5.1'):
         print("""ERROR: Sundials version %r is not supported.""" % env['sundials_version'])
         sys.exit(1)
     print("""INFO: Using system installation of Sundials version %s.""" % sundials_version)
--- a/SConstruct
+++ b/SConstruct
@@ -1108,13 +1108,14 @@ if env['system_sundials'] == 'y':
 
     # Ignore the minor version, e.g. 2.4.x -> 2.4
     env['sundials_version'] = '.'.join(sundials_version.split('.')[:2])
-    if env['sundials_version'] not in ('2.4','2.5','2.6','2.7','3.0','3.1','3.2','4.0','4.1','5.0','5.1'):
+    sundials_ver = LooseVersion(env['sundials_version'])
+    if sundials_ver < LooseVersion('2.4') or sundials_ver >= LooseVersion('6.0'):
         print("""ERROR: Sundials version %r is not supported.""" % env['sundials_version'])
         sys.exit(1)
     print("""INFO: Using system installation of Sundials version %s.""" % sundials_version)
 
     #Determine whether or not Sundials was built with BLAS/LAPACK
-    if LooseVersion(env['sundials_version']) < LooseVersion('2.6'):
+    if sundials_ver < LooseVersion('2.6'):
         # In Sundials 2.4 / 2.5, SUNDIALS_BLAS_LAPACK is either 0 or 1
         sundials_blas_lapack = get_expression_value(['"sundials/sundials_config.h"'],
                                                        'SUNDIALS_BLAS_LAPACK')
@@ -1606,7 +1607,7 @@ linkSharedLibs = ['cantera_shared']
 
 if env['system_sundials'] == 'y':
     env['sundials_libs'] = ['sundials_cvodes', 'sundials_ida', 'sundials_nvecserial']
-    if env['use_lapack'] and LooseVersion(env['sundials_version']) >= LooseVersion('3.0'):
+    if env['use_lapack'] and sundials_ver >= LooseVersion('3.0'):
         if env.get('has_sundials_lapack'):
             env['sundials_libs'].extend(('sundials_sunlinsollapackdense',
                                          'sundials_sunlinsollapackband'))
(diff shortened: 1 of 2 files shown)


band-a-prend:
P.s.

Tested for system installed sundials.

I tried to make changes like in https://github.com/Cantera/cantera/pull/780/commits/94a7003ea272cec8b34ba9d59aa6f388a49ee24e
 ,but the submodule fetch sundials major release 5.0 instead of 5.1.

bryanwweber:
@band-a-prend Thanks for submitting this! You can update the submodule by
```
cd ext/sundials
git fetch
git checkout **tag for 5.1.0**
cd ../..
git add .
git commit -m 'Update Sundials submodule'
```
After that, assuming no code in SUNDIALS has moved around, the other changes in 94a7003 should be straightforward.

band-a-prend:
Thank you for the clue. The command `git fetch` in submodule directory didn't result in any thing for me and checkout didn't work after it. I have to use `git submodule update --remote -- .` to clone submodule repo and then `git checkout v5.1.0`. After that and going to root directory of cantera project the commited diff noticed a change in submodule sha1.

I push changes now and hope to check build process on my local machine this evening.

bryanwweber:
Thanks @band-a-prend perhaps the remote was not configured for the submodule or something. Anyways, the change seems to be appropriate

band-a-prend:
I recheck currently build and tests, please wait.

band-a-prend:
Appropriate sha1-snapshot (b78e575639babe99aea9a0558d0f64732d6d729e) of sundials tag v5.1.0 is fetched now into `ext/sundials` by `SConstruct` script. The build with options `python_package = 'full' f90_interface = 'y' system_eigen = 'y' system_sundials = 'n' extra_inc_dirs = '/usr/include/eigen3'` is successfully done. Tests are passed:
```
*****************************
***    Testing Summary    ***
*****************************

Tests passed: 1046
Up-to-date tests skipped: 0
Tests failed: 0

*****************************
```

speth:
Given the increased rate of releases of SUNDIALS, and the fact that we haven't seen breaking changes (at least, that affect Cantera) in several recent minor releases, I think we should assume compatibility with at least minor versions, i.e. 5.x. A build-time warning for versions newer than the latest we have tested would probably still be useful.

bryanwweber:
Also, since we now support every released version back to 2.4, I think we should turn this into a `LooseVersion` check, rather than explicitly listing all of the satisfactory values. That will simplify that rather unwieldy line, I think. It could be something like

```python
if LooseVersion(env['sundials_version']) >= LooseVersion('6.0'):
    error...
```

band-a-prend:
@bryanwweber,
I replaced version list with LooseVersion conditions

SConstruct
```diff
@@ -1048,6 +1048,7 @@ env['has_demangle'] = conf.CheckDeclaration("boost::core::demangle",
                                 '#include <boost/core/demangle.hpp>', 'C++')
 
 import SCons.Conftest, SCons.SConf
+from distutils.version import LooseVersion
```
bryanwweber: You don't need to import this, `LooseVersion` is imported somewhere in the `builtutils`, which is `*` imported in `SConstruct`.

SConstruct
```diff
@@ -1108,7 +1109,7 @@ if env['system_sundials'] == 'y':
 
     # Ignore the minor version, e.g. 2.4.x -> 2.4
     env['sundials_version'] = '.'.join(sundials_version.split('.')[:2])
-    if env['sundials_version'] not in ('2.4','2.5','2.6','2.7','3.0','3.1','3.2','4.0','4.1','5.0'):
+    if LooseVersion(env['sundials_version']) < LooseVersion('2.4') or LooseVersion(env['sundials_version']) >= LooseVersion('6.0'):
```
bryanwweber: I'd suggest creating a variable for `LooseVersion(env['sundials_version'])` here to shorten this line... the whole point was to make it shorter :joy: You could then reuse that variable on line 1118 in that other `if` statement comparing to `2.4`.

bryanwweber (changes requested):
Couple of small changes

SConstruct
band-a-prend: I think I then also could safely reuse this shorter variable at line 1610 because it's within  another one `if env['system_sundials'] == 'y'` condition check.

`system_sundials_version` seems as reasonable name this case: although this will not make expression significantly shorter but it will avoid evaluation of LooseVersion() again and again for  `if env['system_sundials'] == 'y'` blocks.

> the whole point was to make it shorter 😂

Yes, at this point of view this attempt certainly has failed. But it then should become more readable in comparison with rapidly expanding list of numbers.

band-a-prend: The last commit is updated.

SConstruct
band-a-prend: Removed.

SConstruct
```diff
@@ -1108,13 +1108,14 @@ if env['system_sundials'] == 'y':
 
     # Ignore the minor version, e.g. 2.4.x -> 2.4
     env['sundials_version'] = '.'.join(sundials_version.split('.')[:2])
-    if env['sundials_version'] not in ('2.4','2.5','2.6','2.7','3.0','3.1','3.2','4.0','4.1','5.0'):
+    system_sundials_version = LooseVersion(env['sundials_version'])
```
bryanwweber: This isn't necessarily associated with the system SUNDIALS though, so I think just `sundials_ver` would be sufficient and shorter

bryanwweber (changes requested):

SConstruct
band-a-prend: Done

bryanwweber (approved):

bryanwweber:
I think the version check will have to go in `ck2yaml.py`, `cti2yaml.py`, and `ctml2yaml.py`. I wonder if that can somehow be centrally managed... Probably not

speth (approved):

PR #817, opened 2020-03-06
Added access functions to polynomial fits of gas transport properties

lavdwall:
**Checklist**

Fixes issue 40 in Cantera/enhancements repo

- Added const access functions to the polynomial fits of viscosity, conductivity and binary diffusivity

### DIFF
--- a/include/cantera/transport/GasTransport.h
+++ b/include/cantera/transport/GasTransport.h
@@ -115,6 +115,24 @@ public:
      */
     virtual void getMixDiffCoeffsMass(doublereal* const d);
 
+    //! Return the polynomial fits to the viscosity of each species
+    const std::vector<vector_fp> viscosityPolynomials() const
+    {
+        return m_visccoeffs;
+    }
+
+    //! Return the temperature fits of the heat conduction
+    const std::vector<vector_fp> conductivityPolynomials() const
+    {
+        return m_condcoeffs;
+    }
+
+    //! Return the polynomial fits to the binary diffusivity of each species
+    const std::vector<vector_fp> binDiffusivityPolynomials() const
+    {
+        return m_diffcoeffs;
+    }
+
     virtual void init(ThermoPhase* thermo, int mode=0, int log_level=0);
 
 protected:
--- a/include/cantera/transport/GasTransport.h
+++ b/include/cantera/transport/GasTransport.h
@@ -116,12 +116,15 @@ public:
     virtual void getMixDiffCoeffsMass(doublereal* const d);
 
     //! Return the polynomial fits to the viscosity of species i
+    //! @see fitProperties()
     virtual void getViscosityPolynomials(size_t i, double* coeffs) const;
 
     //! Return the temperature fits of the heat conductivity of species i
+    //! @see fitProperties()
     virtual void getConductivityPolynomials(size_t i, double* coeffs) const;
 
     //! Return the polynomial fits to the binary diffusivity of species pair (i, j)
+    //! @see fitDiffCoeffs()
     virtual void getBinDiffusivityPolynomials(size_t i, size_t j, double* coeffs) const;
 
     virtual void init(ThermoPhase* thermo, int mode=0, int log_level=0);
--- a/include/cantera/transport/GasTransport.h
+++ b/include/cantera/transport/GasTransport.h
@@ -115,23 +115,14 @@ public:
      */
     virtual void getMixDiffCoeffsMass(doublereal* const d);
 
-    //! Return the polynomial fits to the viscosity of each species
-    const std::vector<vector_fp> viscosityPolynomials() const
-    {
-        return m_visccoeffs;
-    }
+    //! Return the polynomial fits to the viscosity of species i
+    virtual void getViscosityPolynomials(size_t i, double* coeffs) const;
 
-    //! Return the temperature fits of the heat conduction
-    const std::vector<vector_fp> conductivityPolynomials() const
-    {
-        return m_condcoeffs;
-    }
+    //! Return the temperature fits of the heat conductivity of species i
+    virtual void getConductivityPolynomials(size_t i, double* coeffs) const;
 
-    //! Return the polynomial fits to the binary diffusivity of each species
-    const std::vector<vector_fp> binDiffusivityPolynomials() const
-    {
-        return m_diffcoeffs;
-    }
+    //! Return the polynomial fits to the binary diffusivity of species pair (i, j)
+    virtual void getBinDiffusivityPolynomials(size_t i, size_t j, double* coeffs) const;
 
     virtual void init(ThermoPhase* thermo, int mode=0, int log_level=0);
 
--- a/src/transport/GasTransport.cpp
+++ b/src/transport/GasTransport.cpp
@@ -828,9 +828,9 @@ void GasTransport::getBinDiffusivityPolynomials(size_t i, size_t j, double* coef
 {
     size_t mi = (j >= i? i : j);
     size_t mj = (j >= i? j : i);
-    int ic = 0;
+    size_t ic = 0;
     for (size_t ii = 0; ii < mi; ii++) {
-        ic += m_nsp-ii;
+        ic += m_nsp - ii;
     }
     ic += mj - mi;
 
--- a/src/transport/GasTransport.cpp
+++ b/src/transport/GasTransport.cpp
@@ -810,4 +810,33 @@ void GasTransport::getBinDiffCorrection(double t, MMCollisionInt& integrals,
           (q2*xk*xk + q1*xj*xj + q12*xk*xj);
 }
 
+void GasTransport::getViscosityPolynomials(size_t i, double* coeffs) const
+{
+    for (size_t k = 0; k < (m_mode == CK_Mode ? 4 : 5); k++) {
+        coeffs[k] = m_visccoeffs[i][k];
+    }
+}
+
+void GasTransport::getConductivityPolynomials(size_t i, double* coeffs) const
+{
+    for (size_t k = 0; k < (m_mode == CK_Mode ? 4 : 5); k++) {
+        coeffs[k] = m_condcoeffs[i][k];
+    }
+}
+
+void GasTransport::getBinDiffusivityPolynomials(size_t i, size_t j, double* coeffs) const
+{
+    size_t mi = (j >= i? i : j);
+    size_t mj = (j >= i? j : i);
+    int ic = 0;
+    for (size_t ii = 0; ii < mi; ii++) {
+        ic += m_nsp-ii;
+    }
+    ic += mj - mi;
+
+    for (size_t k = 0; k < (m_mode == CK_Mode ? 4 : 5); k++) {
+        coeffs[k] = m_diffcoeffs[ic][k];
+    }
+}
+
 }
--- a/test/transport/transportPolynomials.cpp
+++ b/test/transport/transportPolynomials.cpp
@@ -10,15 +10,19 @@ class TransportPolynomialsTest : public testing::Test
 {
 public:
     TransportPolynomialsTest() {
-        phase.reset(newPhase("gri30.yaml"));
+        phase.reset(newPhase("h2o2.yaml"));
+        tran.init(phase.get(), 0);
+        ck_tran.init(phase.get(), CK_Mode);
     }
 
     void check_viscosity_poly(const std::string& speciek, const vector_fp& visc_coeff_expected, int cmode) {
-        MixTransport tran;
-        tran.init(phase.get(), cmode);
         size_t k = phase->speciesIndex(speciek);
         vector_fp coeffs (cmode == CK_Mode ? 4 : 5);
-        tran.getViscosityPolynomials(k, &coeffs[0]);
+        if (cmode == CK_Mode) {
+            ck_tran.getViscosityPolynomials(k, &coeffs[0]);
+        } else {
+            tran.getViscosityPolynomials(k, &coeffs[0]);
+        }
         for (size_t i = 0; i < visc_coeff_expected.size(); i++) {
             EXPECT_NEAR(coeffs[i], visc_coeff_expected[i], 1e-5);
         }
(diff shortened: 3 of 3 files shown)


bryanwweber:
Hi @lavdwall Thanks for the contribution! Would it be possible to add some documentation here about the specific polynomials that are used for the fits? I don't think we have that anywhere on the website yet, but if we do, that's fine you don't need to change anything here. Also, can you add some tests that these functions are returning the expected result?

Just for linking: Cantera/enhancements#40

speth:
If these are going to be public member functions, then I think in addition to documenting the polynomial form, the structure of the returned array should be clearly described.

lavdwall:
OK, I will add some tests and documentation.

About the documentation: 
- The polynomials that are used for the fits are briefly described with the definition of fitProperties() and fitDiffCoeffs() in GasTransport.H, and can be found in the Doxygen documentation: https://cantera.org/documentation/docs-2.4/doxygen/html/d8/d58/classCantera_1_1GasTransport.html#a267d20cdea7486fe84e722ee82d84b20
- The structure of the returned arrays is briefly described with the definition of m_diffcoeffs, m_condcoeffs and m_visccoeffs, also in GasTransport.H

So, how much more detail should I add?

test/transport/transportPolynomials.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+
+TEST_F(TransportPolynomialsTest, viscosityPolynomials)
+{
+    check_viscosity_poly("H2", {-0.00044135741,0.00054143229,-0.00010356811,9.67388e-06,-3.32304e-07}, 0);
+    check_viscosity_poly("O2", {-0.00628086,0.0036755258,-0.00069891557,6.0422681e-05,-1.9517881e-06}, 0);
+    check_viscosity_poly("H2O", {0.01283557,-0.0070486222,0.0014476475,-0.00012428231,3.86979e-06}, 0);
+    check_viscosity_poly("H2", {-15.80391,0.853653,-0.028234165,0.00126934}, 10);
+    check_viscosity_poly("O2", {-19.450466,2.6764033,-0.27227259,0.012162949}, 10);
+    check_viscosity_poly("H2O", {-14.902224,-0.528783,0.3032122,-0.018641952}, 10);
```
speth: Where do these reference values come from? This should be documented.

Also, the constant `CK_Mode` should be used rather than the magic number 10.

include/cantera/transport/GasTransport.h
```diff
@@ -115,6 +115,24 @@ class GasTransport : public Transport
      */
     virtual void getMixDiffCoeffsMass(doublereal* const d);
 
+    //! Return the polynomial fits to the viscosity of each species
+    const std::vector<vector_fp> viscosityPolynomials() const
```
speth: I assume these functions should be returning references, i.e. `const std::vector<vector_fp>&`. There's no reason to return a copy, right?

test/transport/transportPolynomials.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "cantera/thermo/ThermoFactory.h"
+#include "cantera/transport/MixTransport.h"
+
+using namespace Cantera;
+
+class TransportPolynomialsTest : public testing::Test
+{
+public:
+    TransportPolynomialsTest() {
+        phase = newPhase("gri30.cti");
```
speth: Just to shave off a little bit of run-time for the tests, I would suggest using `h2o2.yaml` as the input file.

test/transport/transportPolynomials.cpp
```diff
@@ -0,0 +1,87 @@
...
+                for (size_t kk = jj; kk < phase->nSpecies(); kk++) { ic++;}
+            }
+            for (size_t kk = j; kk < k; kk++) { ic++;}
+        }    
+        for (size_t i = 0; i<bindiff_coeff_expected.size(); i++) {
+            EXPECT_FLOAT_EQ(tran.binDiffusivityPolynomials()[ic][i], bindiff_coeff_expected[i]);
+        }
+    }
+
+    ThermoPhase* phase;
```
speth: Use `std::unique_ptr<ThermoPhase>` to avoid creating a memory leak.

include/cantera/transport/GasTransport.h
```diff
@@ -115,6 +115,24 @@ class GasTransport : public Transport
...
+    }
+
+    //! Return the temperature fits of the heat conduction
+    const std::vector<vector_fp> conductivityPolynomials() const
+    {
+        return m_condcoeffs;
+    }
+
+    //! Return the polynomial fits to the binary diffusivity of each species
+    const std::vector<vector_fp> binDiffusivityPolynomials() const
```
speth: Given the complexity of getting the polynomial for a species pair _(i, j)_, I think it would make more sense for this method to return the polynomial for just a single species given the indices so we can do that work for the user. That would also simplify documenting the return type.

I also think we should implement an interface that is similar to the way that Cantera generally handles "returning" array data without exposing implementation details about the underlying data structure, which is to fill in a user-provided array. In this case, the signature for this function might be
```c++
void getBinDiffusivityPolynomials(size_t i, size_t k, double* coeffs);
```

The other methods could be defined similarly.

test/transport/transportPolynomials.cpp
```diff
@@ -0,0 +1,87 @@
...
+using namespace Cantera;
+
+class TransportPolynomialsTest : public testing::Test
+{
+public:
+    TransportPolynomialsTest() {
+        phase = newPhase("gri30.cti");
+    }
+
+    void check_viscosity_poly(const std::string speciek, vector_fp visc_coeff_expected, int mode) {
```
speth: This is a very minor optimization, but for consistency with what we do generally, non-trivial types should be passed by const reference. I'd also suggest `species_k` as the variable name:
```suggestion
    void check_viscosity_poly(const std::string& species_k, const vector_fp& visc_coeff_expected, int mode) {
```

test/transport/transportPolynomials.cpp
```diff
@@ -0,0 +1,87 @@
...
+        MixTransport tran;
+        tran.init(phase, mode);
+        size_t k = phase->speciesIndex(speciek);
+        for (size_t i = 0; i<visc_coeff_expected.size(); i++) {
```
speth: Minor formatting quibble (here and elsewhere):
```suggestion
        for (size_t i = 0; i < visc_coeff_expected.size(); i++) { 
```

test/transport/transportPolynomials.cpp
```diff
@@ -0,0 +1,87 @@
...
+    TransportPolynomialsTest() {
+        phase = newPhase("gri30.cti");
+    }
+
+    void check_viscosity_poly(const std::string speciek, vector_fp visc_coeff_expected, int mode) {
+        MixTransport tran;
+        tran.init(phase, mode);
+        size_t k = phase->speciesIndex(speciek);
+        for (size_t i = 0; i<visc_coeff_expected.size(); i++) { 
+            EXPECT_FLOAT_EQ(tran.viscosityPolynomials()[k][i], visc_coeff_expected[i]);
```
speth: This should always be:
```suggestion
            EXPECT_DOUBLE_EQ(tran.viscosityPolynomials()[k][i], visc_coeff_expected[i]);
```

speth (changes requested):
If you're going to rely on the documentation for other methods / member variables, you should reference those items in the documentation for each method, preferably in a way that generates a link within the Doxygen documentation. For member variables, you can link to them using syntax `#variableName`, and member functions `methodName()`.

Also, you may want to set your email address on the commits to an address that is known to GitHub so your commits will be attributed correctly.

ecisneros8:
Hello all,

What's the status of this PR? My group is interested in this, and we'd like to make these polynomials available through the Python interface too. I'm hoping this is the right channel for this question-- let me know otherwise. Thanks in advance.

bryanwweber:
Hi @ecisneros8 Thanks for your interest! I'm not sure of the status here after @speth made some comments to @lavdwall that needed to be fixed. If you can determine how to complete those suggestions, please feel free to start from this branch and create a new pull request. Otherwise, feel free to start over since there hasn't been any activity in some time.

ecisneros8:
Thanks @bryanwweber, I'll definitely pursue this, though I'll weigh my options first. Best!

ischoegl:
Hi @lavdwall (or @ecisneros8) ... what is the current status of this PR? I believe the review comments should be relatively straight-forward to address.

lavdwall:
Hi @ischoegl,
To be honest, I have been using my implementation as such for a while now without any problems so I didn't really feel the urge to do it the "proper way" and incorporate your comments. But I agree that it should be done properly if it is to end up in the official repo. I will look at the comments later this month and address them accordingly.
Laurien

ischoegl:
Hi @lavdwall, that would be terrific! (I believe providing interfaces to parameters that are useful to the community is important.) Credit for reviews goes to @speth, of course. From my own perspective, it would be nice to break things out to Python also, but that can always be added at a later point.

lavdwall:
Hi,

I made some changes in GasTransport.h and GasTransport.cpp.
Regarding the reference values for the tests, I actually let the code print those values once from within the fitProperties function (where they are originally calculated). I assumed that the values there are the correct ones, so I took those as "expected values' in the tests. I get that this might not be the best way. But I also don't really see what can go wrong in the new access functions, so perhaps it is not really necessary to have those tests (so transportPolynomials.cpp can just be omitted?).

Laurien

ischoegl:
Hi @lavdwall  ... thanks for updating. I think there are some minor points left to be addressed (also: some windows tests fail). Regarding tests, it's always good to have those included, and adding a comment on how values were originally chosen would be good. In effect, you just ensure that unit tests catch a change in behavior, with is legitimate.

lavdwall:
Hi @ischoegl,
I am afraid I don't really understand what goes wrong in the Windows tests...Is there a way I can test/debug this before committing any changes?  The "scons test" only tells me something about the test for my installation (so the "ubuntu-20.04-python-3.8" check in the list). How can I check, before committing, if it works on Windows?  Preferably without actually having to build/test it in MS Visual Studio...

ischoegl:
Hi @lavdwall … you can see details on the error stack on the ‘checks’ tab on top (or by clicking ‘details’ next to the failing tests). From what I can tell, MSVC may not like statements like `double coeffs[(cmode == CK_Mode ? 4 : 5)];`

speth:
Variable-length arrays are a C99 extension, not part of the C++11 standard, and not supported by MSVC. The C++ way to do this is using `std::vector`, e.g. `vector_fp coeffs((cmode == CK_Mode) ? 4 : 5);`.

lavdwall:
Thanks! That explains....
Let's wait 15 minutes for the checks, but I guess it should be OK now :)

lavdwall:
OK, so there is one check still failing: "CI / Test pre-release version of Cython (pull_request)"
But as far as I can see this has nothing to do with the changes I made... I guess I can ignore it then?

bryanwweber:
Thanks @lavdwall that check is failing due to a change in Cython, and we've fixed it elsewhere.

src/transport/GasTransport.cpp
```diff
@@ -810,4 +810,33 @@ void GasTransport::getBinDiffCorrection(double t, MMCollisionInt& integrals,
...
+    }
+}
+
+void GasTransport::getBinDiffusivityPolynomials(size_t i, size_t j, double* coeffs) const
+{
+    size_t mi = (j >= i? i : j);
+    size_t mj = (j >= i? j : i);
+    int ic = 0;
+    for (size_t ii = 0; ii < mi; ii++) {
+        ic += m_nsp-ii;
```
ischoegl: Cantera usually uses a whitespace here, i.e. 
```C++
ic += m_nsp - ii;
```

src/transport/GasTransport.cpp
```diff
@@ -810,4 +810,33 @@ void GasTransport::getBinDiffCorrection(double t, MMCollisionInt& integrals,
...
+    for (size_t k = 0; k < (m_mode == CK_Mode ? 4 : 5); k++) {
+        coeffs[k] = m_condcoeffs[i][k];
+    }
+}
+
+void GasTransport::getBinDiffusivityPolynomials(size_t i, size_t j, double* coeffs) const
+{
+    size_t mi = (j >= i? i : j);
+    size_t mj = (j >= i? j : i);
+    int ic = 0;
```
ischoegl: Should this `int` be a `size_t`? I believe the way this is coded, `ic` cannot become negative. 

As an aside, there should be a way to eliminate the loop that calculates `ic`, but I don't think that this operation is time critical. Either way, it would be good to add a comment about how coefficients are being accessed within the storage structure.

test/transport/transportPolynomials.cpp
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    ThermoPhase* phase;
+};
+
+
+
+TEST_F(TransportPolynomialsTest, viscosityPolynomials)
+{
+    check_viscosity_poly("H2", {-0.00044135741,0.00054143229,-0.00010356811,9.67388e-06,-3.32304e-07}, 0);
```
ischoegl: please add white-spaces after the commas here and where it applies below.

test/transport/transportPolynomials.cpp
```diff
@@ -0,0 +1,81 @@
...
+public:
+    TransportPolynomialsTest() {
+        phase = newPhase("gri30.cti");
+    }
+
+    void check_viscosity_poly(const std::string& speciek, const vector_fp& visc_coeff_expected, int cmode) {
+        MixTransport tran;
+        tran.init(phase, cmode);
+        size_t k = phase->speciesIndex(speciek);
+        vector_fp coeffs (cmode == CK_Mode ? 4 : 5);
```
ischoegl, line 20: No whitespace after `coeffs` (here and where it applies below) as this is a call of the constructor with parameters in parentheses.

ischoegl (changes requested):
This looks pretty close, although I added a couple of really minor formatting comments.

Beyond, the only other things left are `std::unique_ptr<ThermoPhase>` and `gri30.cti` replaced by `h2o2.yaml`, as pointed out in an earlier review.

**PS:** I also noticed that there a couple of commits that have identical messages, as well as commits that may result from a merge. It would be great if you could clarify the commit history.

include/cantera/transport/GasTransport.h
```diff
@@ -115,8 +115,18 @@ class GasTransport : public Transport
...
     virtual void getMixDiffCoeffsMass(doublereal* const d);
 
+    //! Return the polynomial fits to the viscosity of species i
+    virtual void getViscosityPolynomials(size_t i, double* coeffs) const;
+
+    //! Return the temperature fits of the heat conductivity of species i
+    virtual void getConductivityPolynomials(size_t i, double* coeffs) const;
+
+    //! Return the polynomial fits to the binary diffusivity of species pair (i, j)
+    virtual void getBinDiffusivityPolynomials(size_t i, size_t j, double* coeffs) const;
```
speth: It would be good to document the length of the `coeffs` vector, especially given that it is of variable size depending on the transport model variant. Also, links to the `fitProperties` and `fitDiffCoeffs` methods would be useful, since the forms of the fits are documented there.

speth (changes requested):

test/transport/transportPolynomials.cpp
ischoegl: this is actually fine.

lavdwall:
> This looks pretty close, although I added a couple of really minor formatting comments.
> 
> Beyond, the only other things left are `std::unique_ptr<ThermoPhase>` and `gri30.cti` replaced by `h2o2.yaml`, as pointed out in an earlier review.
> 
> **PS:** I also noticed that there a couple of commits that have identical messages, as well as commits that may result from a merge. It would be great if you could clarify the commit history.

I resolved the formatting comments, but I didn't change gri30 into h2o2.yaml, because the values for the tests are not exactly the same with both mechanisms so I would need to change all the values...
How do you want me to "clarify the commit history"?

speth:
Regarding the commit history, @ischoegl may have some additional suggestions but I think the biggest issue is getting rid of the merge commits, since we keep a linear history on Cantera's `main` branch. I tested rebasing on to `main`, and found that it generated some merge conflicts, so I resolved them and pushed the resulting branch to my fork: https://github.com/speth/cantera/tree/pr-817. I would suggest resetting your branch to that match that (you probably need to add my fork as an additional "remote" in git, and fetch that first) as a starting point.

Beyond that, I think you may want to update the author information on your git commits. If you look at the [history page](https://github.com/Cantera/cantera/pull/817/commits) here on Github you can see that some of the commits have a different name listed, and many of them do not seem to have an email address specified that is linked to your GitHub account (since they don't show your user icon in the preview, just the GitHub octocat). You can reset this information as part of an interactive rebase with git. This [StackOverflow post](https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit) provides some helpful commands.

ischoegl:
@lavdwall … regarding the commit history, I don’t have any suggestions beyond @speth’s recommendations (and what was already said before: renaming or squashing commits that have the same descriptions). Rebasing is one procedure where I yet have to make up my mind about how best to explain the process, so I believe Ray’s helpful suggestion to reset to the branch he straightened up would be the way to go. The rest should be straightforward.

lavdwall:
I think it worked...
Thanks for all the help!

speth:
Looks like the author reset worked well. The fact that the "committer" information is not recognized by GitHub suggests that you might want to update your Git settings, e.g.
```bash
git config --global user.name "John Doe"
git config --global user.email johndoe@example.com
```
So that both the author and committer will be correct automatically. It doesn't affect this PR, though, since the "committer" will be set to whoever merges the PR in any case.

lavdwall:
Thanks!
I had actually done that, but I had set my UGent email address (which is used in github enterprise) instead of the private email address that is apparently linked to the "regular" github. I updated the list of email addresses in my account ;)

speth:
Also, I'll reiterate the suggestion to use `h2o2.yaml`. I know it's annoying to have to update all of the coefficients (I guess, given that we have the demonstration that it works here, you can just use the new methods to extract them), but it has a pretty significant impact on the time to run the tests, in cases like the "coverage" build where we have to run in debug mode. Here, these tests end up taking over 30 seconds. There's also room to speed this up by only creating two `MixTransport` objects - one for the "regular" mode, and one for `CK_Mode`. Right now, the tests end up creating 8 identical `MixTransport` objects with each setting.

(The coefficients are because the transport properties end up being fit over a different temperature range because some of the species that are only in GRI-3.0 have a max temperature listed at 3000 K, while all of the species in the H2/O2 submechanism have max temperature of 3500, and we do the transport fits to the same (smaller) temperature range for all species).

ischoegl (approved):
Thanks!

speth (approved):
Looks good to me. Thank you!

lavdwall:
> Hi @lavdwall, that would be terrific! (I believe providing interfaces to parameters that are useful to the community is important.) Credit for reviews goes to @speth, of course. From my own perspective, it would be nice to break things out to Python also, but that can always be added at a later point.

I have made these functions available in the Python interface.
https://github.com/lavdwall/cantera/tree/transport_properties

Next to access to these polynomial fits (both in C++ and Python), I also added the possibility to modify the polynomial coefficients manually. Reason: there is literature that reports the polynomial fits for binary diffusion coefficients and collision integrals explicitly, rather than in terms of the classic L-J parameters. If this is important, the user now has the option to change the polynomial fits for some species or species pairs. 

If the option to modify the polynomial fits is not OK for the public repo, I can remove those things and just keep the python access functions. Let me know.

Grtz,
Laurien

lavdwall:
@ischoegl 
Not sure if you get notified of comments to closed pull request. See my latest comment. As per your earlier suggestion, I added the access functions in the python interface. Should I open a new pull request for this?

ischoegl:
@lavdwall … yes, please do!

PR #820, opened 2020-03-11
[samples] Fix heat release in ic_engine.py

### LINKED ISSUE #819: Example ic_engine.py heat release rate far too high
**System information**

- Cantera version: 2.5 (current dev version, pulled from git yesterday)
- OS: debian10 
- Python/MATLAB version: python3.7

**Actual behavior**

I was running the ic_engine.py and the output of
'Heat release rate per cylinder (estimate):'
is far too high, something like 1.3 GW!
That's, considering the amount of fuel injected, impossible.
The work is less than 20kW and seems to be reasonable.

The time resolved heat release is computed during simulation as:
```python
heat_release_rate = - cyl.volume * ct.gas_constant * cyl.T * \
        np.sum(gas.standard_enthalpies_RT * cyl.thermo.net_production_rates, 0)
```
It's always positive, whereas the second line is always negative, probably the walls are adiabatic, and hence there's no quenching in the model for that. I assume that should not be of a big issue.
But I didn't understand completely what's done here, especially not in the second line (what's the zero in the end of the line, supposed to do?). I also checked the units and it should be result in J/s, and be correct. But the heat release during combustion should be the integral c_v dT from beginning to end for constant volume combustion (I know that this is not the ideal case here). But this seems to be only the time revolved chemical energy, but the actual integration is done over time:
```python
# heat release
Q = trapz(states.heat_release_rate, t)
print('{:45s}{:>4.1f} kW'.format('Heat release rate per cylinder (estimate):',
                                 Q / t[-1] / 1000.))
```
I didn't check the math here, if you can easily exchange coordinates from integration over temperature with over time. Anyway the result is far too high, and it comes mainly from the sum over the production rates.

**Additional context**

I also computed the AFR, just to check if something is wrong in general (because the fuel was changed from propane to C12H26 in the example a few time ago, but the fuel amount remained), but it's about 16. So seemed to be ok:
```python
# AFR
airIn = trapz(states.mdot_in,t)
AFR = airIn/(injector_mass*sim_n_revolutions*0.5)
print("AFR",AFR)
```
Please let me know, if you want to add these lines to the example, and if you need a merge request or something like that, for it.

ischoegl:
**Checklist**

Fixes #819

This commit fixes a bug caused by the recently introduced `heat_release_rate` method of `SolutionArray` objects (PR #810): the new method masked the pre-existing homonymous `extra` column.

The behavior is now corrected
```
In [1]: %run ic_engine.py
Heat release rate per cylinder (estimate):   35.4 kW
Expansion power per cylinder (estimate):     18.5 kW
Efficiency (estimate):                       52.3 %
CO emission (estimate):                       8.8 ppm
```

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -513,9 +513,15 @@ class SolutionArray:
             self._indices = list(np.ndindex(self._shape))
             self._output_dummy = self._states[..., 0]
 
+        reserved = self.__dir__()
+
         self._extra = {}
         if isinstance(extra, dict):
             for name, v in extra.items():
+                if name in reserved:
+                    raise ValueError(
+                        "Unable to create extra column '{}': name is already "
+                        "used by SolutionArray objects.".format(name))
                 if not np.shape(v):
                     self._extra[name] = [v]*self._shape[0]
                 elif len(v) == self._shape[0]:
@@ -526,6 +532,10 @@ class SolutionArray:
 
         elif extra and self._shape == (0,):
             for name in extra:
+                if name in reserved:
+                    raise ValueError(
+                        "Unable to create extra column '{}': name is already "
+                        "used by SolutionArray objects.".format(name))
                 self._extra[name] = []
 
         elif extra:
--- a/interfaces/cython/cantera/examples/reactors/ic_engine.py
+++ b/interfaces/cython/cantera/examples/reactors/ic_engine.py
@@ -155,7 +155,7 @@ cyl.set_advance_limit('temperature', delta_T_max)
 # set up output data arrays
 states = ct.SolutionArray(
     cyl.thermo,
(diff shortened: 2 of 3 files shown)


speth (commented):
This change looks fine as a fix for the specific issue at hand. The one thing I'm wondering is whether we should prevent the use of `extra` columns that would be masked in this way. It seems like a bit of surprising behavior that may catch users off guard.

ischoegl:
> The one thing I'm wondering is whether we should prevent the use of extra columns that would be masked in this way. It seems like a bit of surprising behavior that may catch users off guard.

Sounds reasonable and *should* be relatively straight-forward to implement. Should I attach it to this PR or open another one?

speth:
I'm thinking it should be fairly simple as well. If it is, feel free to add it to this PR.

ischoegl:
@speth ... easy enough, no surprises.

ischoegl:
@speth .. not sure what the build error is.

speth (approved):

PR #822, opened 2020-03-13
Add 'negative_A' option for third body reactions

jsantner:
Previously, a set of reactions in chemkin format such as the following...
`R1A+R1B+M=>R6+M              2e11  0.0   20000.0`
`  DUPLICATE`
` `  
`R1A+R1B+M=>R6+M              -1e11  0.0   20000.0`
`DUPLICATE`
...would cause the error below when validating the mechanism.
`***********************************************************************`
`CanteraError thrown by ElementaryReaction::validate:`
`Undeclared negative pre-exponential factor found in reaction 'R1A + R1B + M => R6 + M'`
`***********************************************************************`

This pull request automatically adds the 'negative_A' options to these third body reactions after conversion to cantera format, similar to the implementation with Arrhenius reactions. This does not modify falloff third-body reactions, for which the validator doesn't support negative A-factors. A test is also added.

### DIFF
--- a/interfaces/cython/cantera/test/test_convert.py
+++ b/interfaces/cython/cantera/test/test_convert.py
@@ -264,6 +264,13 @@ class converterTestCommon:
         self.checkKinetics(ref, gas, [300, 800, 1450, 2800], [5e3, 1e5, 2e6],
                            tol=2e-7)
 
+    def test_negative_A_factor(self):
+        self.convert('negative-rate.inp', thermo='dummy-thermo.dat')
+        gas = ct.Solution('negative-rate.cti')  # Validate the mechanism
+        self.assertLess(gas.reaction(4).rate.pre_exponential_factor, 0)
+        self.assertLess(gas.reaction(1).rate.pre_exponential_factor, 0)
+        self.assertLess(gas.reaction(2).rate.pre_exponential_factor, 0)
+
     def test_bad_troe_value(self):
         with self.assertRaises(ValueError):
             self.convert('bad-troe.inp', thermo='dummy-thermo.dat')
--- a/interfaces/cython/cantera/ck2cti.py
+++ b/interfaces/cython/cantera/ck2cti.py
@@ -669,6 +669,12 @@ class ThirdBody(KineticsModel):
         lines[-1] = lines[-1][:-1] + ')'
         return '\n'.join(lines)
 
+    def options(self):
+        if self.arrheniusHigh.A[0] < 0:
+            return ['negative_A']
+        else:
+            return []
+
 
 class Falloff(ThirdBody):
     """
--- a/test/data/negative-rate.inp
+++ b/test/data/negative-rate.inp
@@ -1,0 +1,27 @@
+ELEMENTS
+H  C   AR
+END
+
+SPECIES
+H
+AR
+R1A R1B P1 R6
+END
+
+REACTIONS
+
+R1A+R1B=>H+P1              1e12  0.0   20000.0
+  DUPLICATE
+
+H+P1=>R1A+R1B             -5e13 -2.0 5000.0
+
+R1A+R1B=>H+P1              -1e11  0.0   20000.0
+  DUPLICATE
+
+R1A+R1B+M=>R6+M              2e11  0.0   20000.0
+  DUPLICATE
+
+R1A+R1B+M=>R6+M              -1e11  0.0   20000.0
+  DUPLICATE
+
+END


bryanwweber (changes requested):
Thanks @jsantner! Good catch. I have a few questions and suggestions:

1. Is the `negative_A` only allowed for duplicate reactions? If so, is that another check that should be added in the `options()` method?
2. Why is the option removed for `Falloff` and `ChemicallyActivated` reactions?
3. Can you please also add this fix to `ck2yaml.py`, if it isn't there?
4. In the test, can you please add some assertions about the `Solution`? Perhaps that the A-factors for the reactions are indeed negative as expected.

Thanks again!

jsantner:
@bryanwweber Good suggestions, I'll respond below.
1. I'm not sure. I think it only makes sense if it's a duplicate reaction such that the total rate of all duplicates summed together is positive. But, that isn't checked for two-body reactions so I didn't add it for three-body reactions. Should I add a check to ensure that all reaction types that include `negative_A` also include `duplicate`? To check that the total rate is positive over all duplicates, I'd need to specify a temperature range. I'm guessing that a negative total reaction rate would give some errors when a chemistry solver (equilibrium, reactor, or flame) is run. Is that correct?
2. `Falloff` reactions don't allow negative A-factors based on the lines I copied below from Reaction.cpp:
`void FalloffReaction::validate() {`
`    Reaction::validate();`
`    if (low_rate.preExponentialFactor() < 0 ||`
`        high_rate.preExponentialFactor() < 0) {`
`        throw CanteraError("FalloffReaction::validate", "Negative "`
`            "pre-exponential factor found for reaction '" + equation() + "'");`
`    }`
`}`
I've never written C++ code, so I don't want to modify that. I also figured there may be some reason that check is there, so I was worried about removing it. Instead, I only applied my change to "simple" three-body reactions, instead of `Falloff` and `ChemicallyActivated` reactions. Do you think I should apply the change to those reaction types? I can modify the lines in Reaction.cpp above, but I'm worried I'd break something because I'm not experienced in C++
3. It looks like it's already in `ck2yaml.py` for `Falloff`, `ThreeBody`, and `ChemicallyActivated`. Since all three of these reaction types allow negative a-factors, I guess I can add them in `ck2cti.py`, as you suggested in point number 2. Do you agree?
4. Yes, I'll do this and push the changes after I hear from you about 1-3 and make those changes too.

bryanwweber:
Thanks @jsantner I'll take a look at `ck2yaml.py` when I get a few minutes and see what's going on there. If the negative A is forbidden by the check in the C++ class, then it wouldn't make sense to include that in the input file. Most likely, there is something about the equation that prevents use of the negative A factor for those types of reactions. For instance (and this is a separate case, but gives the idea) PLOG reactions cannot have negative A factors at a given pressure because of the log-log interpolation. It's unlikely that an overall negative rate would cause errors for the solvers, etc. though, I think, only if it was a problem to compute the rate.

My first point about the `duplicates` was related to the PLOG case, where duplicate reactions that have the same pressure value are allowed to have negative A factors if, and only if, the sum is positive. At least, I think that's how it works.

speth:
I forget where this came up, but for falloff / chemically-activated reactions, there is no mathematical problem with negative rate constants if both the high and low rate constants are negative. This is the case that `ck2yaml.py` is trying to set the flag for.

However, you're right that this is still blocked by at the C++ level by `FalloffReaction::validate`. I think it would make sense to change that behavior to be allowable if the `negative-A` option is set in the input file, but that will require more modification in C++ to parse and store that setting for falloff and chemically-activated reactions. I don't think it's necessary to handle this case in this PR if you don't feel comfortable making these changes.

jsantner:
In this commit, I made the requested changes. I didn't modify the C++ file as speth mentioned I can leave that out of this PR.

bryanwweber (approved):
Thanks @jsantner!

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -263,6 +263,13 @@ def test_negative_order_permissive(self):
         self.checkKinetics(ref, gas, [300, 800, 1450, 2800], [5e3, 1e5, 2e6],
                            tol=2e-7)
 
+    def test_negative_A_factor(self):
+        self.convert('negative-rate.inp', thermo='dummy-thermo.dat')
+        gas = ct.Solution('negative-rate.cti')  # Validate the mechanism
+        assert gas.reaction(4).rate.pre_exponential_factor < 0
+        assert gas.reaction(1).rate.pre_exponential_factor < 0
+        assert gas.reaction(2).rate.pre_exponential_factor < 0
```
speth: Test assertions should be done using the functions from the Python `unittest` module, i.e.
```python
    self.assertLess(gas.reaction(1).rate.pre_exponential_factor, 0)
```

speth (changes requested):
Looks fine, with just a minor suggestion on how to write the test cases.

speth (approved):
These changes look good. I rebased this onto the current master and undid the merge commit to clean up the history. I'll merge this after the tests run again, just to make sure I didn't break anything.

PR #823, opened 2020-03-21
Created a sample ThermoPhase.pyx

Arsh0508:
**Checklist**

Fixes #

-
-
-


Arsh0508:
@bryanwweber sir, I have wrote a demo wrapper for method/function "setState_PX" in class "ThermoPhase" in namespace "Cantera" supporting Pa and atm.
Please kindly review it once and if this is fine,  I will extend it to other methods, units and classes.

Regards,
Arsh

bryanwweber:
@Arsh0508 I'm not sure this is a good approach. The problem is, what if a user wants to input pounds per square inch? Or pounds per square foot? Or kPa? Or... Hopefully you can see how treating each of these individually will be unsustainable. Moreover, this would have to be done for every method that requires units, which would be a lot of work.

Instead, I think a better approach will be to use something like Pint (as mentioned on Cantera/enhancements#3) and its `ureg.wraps()` method. In addition, I think it would be best to automatically generate the appropriate Python code that wraps each method in the .pyx files.

Hope it helps!

Arsh0508:
@bryanwweber sir, though I haven't really worked with Pint in the past, but I sure will check that out right away and then write a demo cython code using the ureg.wraps() method, if you prefer it that way and get back to you.

Thanks for taking time off your schedule to review it!

bryanwweber:
Superseded by #825

PR #824, opened 2020-03-22
[SCons] Fix 'KeyError' for ' --prefix=${python_prefix}'

band-a-prend:
**Checklist**

Fixes # https://github.com/Cantera/cantera/commit/582eb42#commitcomment-37960551

The `.format()` method is applied after concatenation of two parts of string
in `localenv.subst()` function so the `{python_prefix}` in the first sunstring
was assumed as `format` mapping key instead of env variable `${python_prefix}`.

### DIFF
--- a/interfaces/cython/SConscript
+++ b/interfaces/cython/SConscript
@@ -109,7 +109,7 @@ elif localenv['python_prefix']:
     elif localenv['libdirname'] != 'lib':
         # 64-bit RHEL / Fedora etc. or e.g. x32 Gentoo profile
         extra = localenv.subst(
-            ' --prefix=${python_prefix}'
+            ' --prefix=${{python_prefix}}'
             ' --install-lib=${{python_prefix}}/${{libdirname}}/python{}/site-packages'.format(py_version))
     else:
         extra = '--user'


interfaces/cython/SConscript
```diff
@@ -109,8 +109,8 @@ elif localenv['python_prefix']:
     elif localenv['libdirname'] != 'lib':
         # 64-bit RHEL / Fedora etc. or e.g. x32 Gentoo profile
         extra = localenv.subst(
-            ' --prefix=${python_prefix}'
-            ' --install-lib=${{python_prefix}}/${{libdirname}}/python{}/site-packages'.format(py_version))
```
speth, line 113: The addition of the extra set of parentheses around the whole string should not be necessary.

speth (commented):
I'm pretty sure this should have always generated an exception -- the implicit string concatenation takes place before the method call in any version of Python, and this has nothing to do with SCons.

interfaces/cython/SConscript
band-a-prend: But without parentheses there will be KeyError as .format() method is applied after concatenation and encounter the {} expression. I wonder why the previous time I tested early patch didn't result in the issue observed therefore I use additional () here this time.

The KeyError points to next string so initially I was confused what is happened.

Do you mean I need remove [SCons] from pr title?

The easy way to reproduce this problem if assumed we want print `{test_1} {test_2} test_value` is run
```
value = 'test_value'
print(
    '{test_1}
    ' {{test_2}} {}'.format(value)
    )
```
KeyError pointed to second string of print() appeared. While
```
value = 'test_value'
print(
    ('{{test_1}}
    ' {{test_2}} {}').format(value)
    )
```
prints the desired string.

bryanwweber: @speth is referring to the extra set of round brackets `()` that you added. The difference between your two code samples is due to the extra set of curly brackets `{}` that you added around `test_1`. For instance, this works, without the extra parentheses:
```python
>>> print(
... '{{test_1}}'
... ' {{test_2}} {}'.format(value)
... )
{test_1} {test_2} test_value
```

band-a-prend: Yes, I unferstanded.  I just tried to be too careful here. I'll remove parenthesis.

speth (approved):

PR #825, opened 2020-03-22
Added a sample wrapper using Pint library, UnitRegistry()

Arsh0508:
**Checklist**

Fixes #

-
-
-

### DIFF
--- a/interfaces/cython/cantera/sample_wrapper/ThermoPhase.pyx
+++ b/interfaces/cython/cantera/sample_wrapper/ThermoPhase.pyx
@@ -1,6 +1,10 @@
 # This is a sample cython code that contains a sample wrapper for the function "setState_PX" in class "ThermoPhase" in namespace "Cantera".
 # I have made this wrapper assuming we would pass a third parameter (datatype = string) in the function containing which units we want, which by default would be SI unit(Pascal(Pa) in this case).
-# Right now I have added support for only atm and Pa and will extend to other Methods, Classes and Units once this sample is approved
+# Right now I have added support for only for one Method/Function. I will extend to other Methods, Classes and Units once this sample is approved
+
+from pint import UnitRegistry
+
+ureg = UnitRegistry()
 
 cdef extern from "ThermoPhase.h" namespace "Cantera":
     cdef cppclass ThermoPhase:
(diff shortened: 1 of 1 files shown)


Arsh0508:
@bryanwweber sir, sorry for the delay I had not used Pint before and hence took me sometime to figure out how to use it.
But now, I have written a demo wrapper using UnitRegistry from pint library.
Is this method good enough?
If yes, then I can proceed and start working on all Classes and Methods.

bryanwweber:
This is closer to what I had in mind. However, I don't think this will work as written because the type of `p` is not `doublereal` any longer. I think this had better be a regular Python file that imports the relevant functions it needs from the existing interface to be able to wrap them. You also have the extra `units` argument still there, but that's not hurting anything.

Arsh0508:
@bryanwweber sir, thanks for reviewing my work, but I'm afraid couldn't really understand what changes are you suggesting me to make.
Do you want me to implement the same functionality using a python file instead of a Cython file??
Also sir was the data type of p modified?

bryanwweber:
Yes, I think a regular Python file is the best way to go. I may be wrong though. What is your reasoning to use a Cython file?

What is the type of `p` that this code would expect? First of all, `doublereal` is not a Cython type. Second, a `double` does not have a `.to()` attribute. Anyways, the type will be irrelevant if you use a regular Python file.

Arsh0508:
No no sir, Python indeed seems better to me too, actually I had used Cython because when I asked you on Gmail, "whether I should just make the needful changes in cpp code itself", you told that my code would go in interfaces/Cython, so I thought maybe you intended me to write a Cython code afterall and said it to be Python because Cython has similar syntax and libraries.

Anyways sir, really sorry for the confusion on my end.

I'll implement the wrapper as a python code and make the Pull Request on the earliest possible.

bryanwweber:
OK, I can see why there was some confusion there! We used to have an older interface in a folder called "interfaces/Python", so when we made the newer interface, we had to put it in the folder called "Cython". Nonetheless, regular Python files can go in that folder and will be included.

There's no need to make a pull request, if you create a branch on your fork you can send me a link to that branch in an email or include it in your draft proposal and I can look at it. If you want to make another pull request here, please set it as "Draft" status, so we can sort them out.

I'm going to close this pull request in the meantime. Feel free to reopen it if you add some more changes to this branch.

Arsh0508:
Oh yes sir, that exactly was the reason for the confusion and sure I'll make the needful changes and get back to you soon!

PR #826, opened 2020-03-23
Fix test failure when re-running purefluid tests 

speth:

The `liquidvapor.yaml` generated by the test suite from the XML or CTI input files is not the same as the one in the Cantera data directory because new names for some of the phase names have been added (such as `carbon-dioxide` instead of `carbondioxide`) which are not in the old input files.

Testing errors would occur when the generated `liquidvapor.yaml` was then read instead of the version from the Cantera data directory, which would happen when re-running the test suite, since normally the pure fluid tests run before the conversion tests.

Also, update Travis badge to reflect migration from .org to .com.

**Checklist**

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -157,8 +157,8 @@ development of Cantera through NumFOCUS.
     :width: 675px
     :align: middle
 
-.. |travisci| image:: https://travis-ci.org/Cantera/cantera.svg?branch=master
-    :target: https://travis-ci.org/Cantera/cantera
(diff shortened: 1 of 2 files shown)


bryanwweber:
I added a variable for the output file (which may or may not be helpful, I couldn't decide and just went for it) and a comment indicating why these tests are different from all the other ones. Feel free to force push back to your original state, but since these were really small changes I just went ahead with them.

bryanwweber (approved):

PR #828, opened 2020-03-25
Updates to fortran interface

dholland88:
**Checklist**

Fixes #

- Add `getPartialMolarIntEnergies`, `setState_TRX`, and `setState_TPY` to the Fortran
interface.

- Allow the option of compiling with the PGI fortran compiler pgfortran. This requires unsetting the thread_flags option e.g. on the command line using the option `thread_flags='' `.

### DIFF
--- a/src/fortran/fct.cpp
+++ b/src/fortran/fct.cpp
@@ -635,6 +635,17 @@ extern "C" {
         return 0;
     }
 
+    status_t th_getpartialmolarintenergies_r_(const integer* n, doublereal* ie)
+    {
+        try {
+            thermo_t* thrm = _fth(n);
+            thrm->getPartialMolarIntEnergies(ie);
+        } catch (...) {
+            return handleAllExceptions(-1, ERR);
+        }
+        return 0;
+    }
+
     //-------------- Kinetics ------------------//
 
     integer kin_newfromfile_(const char* filename, const char* phasename,
--- a/SConstruct
+++ b/SConstruct
@@ -412,7 +412,7 @@ config_options = [
     PathVariable(
         'FORTRAN',
         """The Fortran (90) compiler. If unspecified, the builder will look for
-           a compatible compiler (gfortran, ifort, g95) in the 'PATH' environment
+           a compatible compiler (pgfortran, gfortran, ifort, g95) in the 'PATH' environment
            variable. Used only for compiling the Fortran 90 interface.""",
         '', PathVariable.PathAccept),
     ('FORTRANFLAGS',
@@ -1171,7 +1171,7 @@ if env['f90_interface'] in ('y','default'):
     if env['FORTRAN']:
         foundF90 = check_fortran(env['FORTRAN'], True)
 
-    for compiler in ('gfortran', 'ifort', 'g95'):
+    for compiler in ('pgfortran', 'gfortran', 'ifort', 'g95'):
         if foundF90:
             break
         foundF90 = check_fortran(compiler)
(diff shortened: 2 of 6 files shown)


src/fortran/fct.cpp
```diff
@@ -354,6 +354,17 @@ extern "C" {
         }
     }
 
+    status_t phase_getpartialmolarintenergies_(const integer* n, doublereal* ie)
```
speth: Shouldn't this method be named `th_getpartialmolarintenergies_` for consistency with methods like `th_getentropies_r_`?

speth (changes requested):
Looks okay to me, with just one small suggestion for consistency with existing methods.

src/fortran/fct.cpp
dholland88: Agreed. I have modified this function name.

dholland88:
I could also do with compiling the fortran interface as a shared library but when I do this I get an error in building the static library:
`scons: *** [build/src/fortran/cantera_funcs.o] AttributeError : 'NoneType' object has no attribute 'get_build_env'`

Do you have any ideas as to what could be going wrong? I don't know much about scons.

bryanwweber:
@dholland88 Can you post your `cantera.conf` file?

dholland88:
Here are my cantera.conf and the modified src/fortran/Sconscript

[cantera.conf.txt](https://github.com/Cantera/cantera/files/4438721/cantera.conf.txt)
[SConscript.txt](https://github.com/Cantera/cantera/files/4438757/SConscript.txt)

bryanwweber:
I'm able to get the compiling to work with this script:
```python
from buildutils import *

Import('env', 'build', 'install')

localenv = env.Clone()
localenv.Prepend(CPPPATH=['#include', '#src'])

f90_src = mglob(localenv, '.', 'f90', 'cpp')

f90_artifacts = localenv.SharedObject(f90_src,
                            F90PATH=['#build/src/fortran'])
mods = [o for o in f90_artifacts if o.path.endswith('.mod')]
objects = [o for o in f90_artifacts if not o.path.endswith('.mod')]
localenv.Depends(objects, localenv['config_h_target'])

lib = build(localenv.StaticLibrary(target='../../lib/cantera_fortran',
                             source=objects))

install('$inst_libdir', lib)
install('$inst_incdir', mods)

# Build the Cantera Fortran interface shared library
if localenv['layout'] != 'debian':
    if localenv['renamed_shared_libraries']:
        sharedName = '../../lib/cantera_fortran_shared'
    else:
        sharedName = '../../lib/cantera_fortran'

    if localenv['versioned_shared_library']:
        shlib = build(localenv.SharedLibrary(sharedName, objects,
                                           SPAWN=getSpawn(localenv),
                                           SHLIBVERSION=localenv['cantera_pure_version'], LIBS=list(env['cantera_libs'])))
        install(localenv.InstallVersionedLib, '$inst_libdir', shlib)
    else:
        shlib = build(localenv.SharedLibrary(target=sharedName, source=objects,
                                           SPAWN=getSpawn(localenv), LIBS=list(env['cantera_libs'])))
        install('$inst_libdir', shlib)
#    env['cantera_shlib'] = shlib
    localenv.Depends(shlib, localenv['config_h_target'])
```
However, it warns me that the static library has no symbols, so I'm not sure that's working properly. I'm not familiar enough with Fortran compiling to know how to fix that. SCons also warns that the shared library is linked using the C++ compiler, and this may cause problems.

dholland88:
Thank you! I cannot see the warning for the static library has no symbols on my set up with the intel compiler, which compiler did you use?

bryanwweber:
`g++` and `gfortran` on macOS, with both installed by Homebrew.

dholland88:
So I have built cantera with intel and gcc on my linux machine, neither gave the warning you mentioned and I have been able to compile my own code against both the static and shared libraries that were created.

A quick google suggests that this might be a building in parallel on Mac thing:
https://stackoverflow.com/questions/4929255/building-static-libraries-on-mac-using-cmake-and-gcc#answer-4953904

speth (approved):

speth:
I modified the last commit here to include linking the rest of the Cantera libraries with the shared fortran library, as in the example `SConscript` that Bryan gave. This means that it is possible to link a program while specifying just `-lcantera_fortran_shared`, without the need to link `cantera_shared` or any of its dependencies like Sundials and BLAS/LAPACK.

I tested this on macOS and did not run into the problem that Bryan mentioned, and was able to link a fortran program using either the static or shared libraries, so I think this is good enough for now.

PR #837, opened 2020-03-30
Fixed slicing update

### LINKED ISSUE #804: Slicing of SolutionArray raises error
**System information**

- Cantera version: 2.5.0a4
- OS: Ubuntu
- Python/MATLAB version: Python 3.6.8

**Expected behavior**

Slicing of `SolutionArray` objects should work as usual (as indexing is implemented).

**Actual behavior**

Slicing does not work.

**To Reproduce**

```
In [1]: import cantera as ct
   ...: arr = ct.SolutionArray(gas, 10)

In [2]: arr[0]
Out[2]: <cantera.composite.SolutionArray at 0x7ffa61a364e0>

In [3]: arr[:]
<ipython-input-5-47ed00660ce3> in <module>()
----> 1 arr[:]

/usr/local/lib/python3.6/dist-packages/cantera/composite.py in __getitem__(self, index)
    534     def __getitem__(self, index):
    535         states = self._states[index]
--> 536         shape = states.shape[:-1]
    537         return SolutionArray(self._phase, shape, states)
    538 

AttributeError: 'list' object has no attribute 'shape'
```

**Additional context**

The error message clearly points at the issue.

sin-ha:
**Checklist**

Fixes #804 
The changes allow the required update in edge cases of list slicing in SolutionArray after squish

-Changes to accommodate list slicing by making two cases of the SolutionArray

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -547,6 +547,8 @@ class SolutionArray:
         states = self._states[index]
         if(isinstance(states, list)):
             num_rows = len(states)
+            if num_rows == 0:
+                states = None
             return SolutionArray(self._phase, num_rows, states)
         else:
             shape = states.shape[:-1]
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -1804,7 +1804,7 @@ class TestSolutionArray(utilities.CanteraTest):
         arr = soln[2:9:3]
         self.assertEqual(len(arr.T), 3)
 
-        #tests for edge cases in sliced solutionArray
+    def test_zero_length_slice_SolutionArray(self):
         states = ct.SolutionArray(self.gas, 4)
         arr1 = states[3:3]
         self.assertEqual(len(arr1.T), 0)
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -1803,3 +1803,14 @@ class TestSolutionArray(utilities.CanteraTest):
         soln = ct.SolutionArray(self.gas, 10)
         arr = soln[2:9:3]
         self.assertEqual(len(arr.T), 3)
+
+        #tests for edge cases in sliced solutionArray
+        states = ct.SolutionArray(self.gas, 4)
+        arr1 = states[3:3]
+        self.assertEqual(len(arr1.T), 0)
+        self.assertEqual(arr1.X.shape, (0,9))
+        self.assertEqual(arr1.n_reactions, 28)
+
+        states.TP = [100,300,900,323.23], ct.one_atm
+        arr2 = states[slice(0)]
+        self.assertEqual(len(arr2.T), 0)


interfaces/cython/cantera/composite.py
```diff
@@ -509,6 +509,9 @@ def __init__(self, phase, shape=(0,), states=None, extra=None):
         if len(self._shape) == 1:
             self._indices = list(range(self._shape[0]))
             self._output_dummy = self._indices
+        elif (len(self._shape) == 0 and len(states) == 0):
```
bryanwweber: Why did you add the `and len(states) == 0` clause to the check here? Just the check on `self._shape` is sufficient in my testing to resolve the bug this is intended to fix. To be clear, I'm not asking you to delete the `len(states)` check, I'm wondering why it is necessary.

You also do not need the parentheses around the statement.

bryanwweber (changes requested):
Another way to fix this is to change line 497 to say `if states:` instead of `if states is not None:`. I'm not sure what other implications that would have though. Do you have any idea which would be better?

bryanwweber:
Also, can you please add some tests that check that this fix is not reverted by accident in the future? Same for your previous fix too.

interfaces/cython/cantera/composite.py
sin-ha: arr[1] would have been treated the same way but we wish to depend on arr
for instance, this part of the test_slicing_onedim should fail.

```
test_slicing_onedim(self):
        states = ct.SolutionArray(self.gas, 5)
        states.TPX = np.linspace(500, 1000, 5), 2e5, 'H2:0.5, O2:0.4'
        T0 = states.T
        H0 = states.enthalpy_mass

        # Verify that original object is updated when slices change
        state = states[1]
        state.TD = 300, 0.5
        self.assertNear(states.T[0], 500)
        self.assertNear(states.T[1], 300)
```
I would remove the Paranthesis.

sin-ha:
> Another way to fix this is to change line 497 to say `if states:` instead of `if states is not None:`. I'm not sure what other implications that would have though. Do you have any idea which would be better?

the same test fails for this it would need more tweaks I guess.


```
>>> states[1]    #arr[1] in the issue
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/cantera/composite.py", line 553, in __getitem__
    return SolutionArray(self._phase, shape, states)
  File "/usr/local/lib/python3.7/site-packages/cantera/composite.py", line 497, in __init__
    if states:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

> Also, can you please add some tests that check that this fix is not reverted by accident in the future? Same for your previous fix too.

I thought there was one written and this was a subcase hence skipped this, Would write one now

bryanwweber:
Thanks for the explanations @sin-ha! Can you edit the commit messages to make them something more meaningful? Thank you!

sin-ha:
@bryanwweber  yes sure, I just observed there is a typo in the comment would mend that too with it.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1803,3 +1803,13 @@ def test_slice_SolutionArray(self):
...
+
+        #tests for edge cases in sliced solution array
+        arr1 = soln[1:1]
+        self.assertEqual(arr1.n_elements, 3)
+
+        arr2 = soln[slice(0)]
+        self.assertEqual(arr2.n_species, 9)
+
+        arr3 = soln[None:0]
+        self.assertEqual(arr3.n_reactions, 28)
```
speth: I think the tests need to cover properties other than scalars that are independent of the slice. For instance, shouldn't all of these cases generate zero-length slices? I get some worrying output from taking such slices:
```python
>>> gas = ct.Solution('h2o2.yaml')
>>> s = ct.SolutionArray(gas, 5)
>>> s.TP = [400,500,600,700,800], ct.one_atm
>>> s[:0].T
array(101325.)
>>> s[1:1].P
array(101325.)
>>> s[3:3].X
array([0.00e+000, 0.00e+000, 0.00e+000, 0.00e+000, 0.00e+000, 9.88e-321,
       0.00e+000, 0.00e+000, 4.94e-324])
```

speth (changes requested):

interfaces/cython/cantera/test/test_thermo.py
sin-ha: @speth I guess the problem was with the default value of states. I guess the update should work.

bryanwweber:
@sin-ha can you please rebase this branch onto the current tip of the `master` branch, and fix any conflicts? If you need some help with that, please let me know!

sin-ha:
Yes I am getting some conflicts in Sconstruct which shouldn't be there while trying to rebase.
edit:
I think I merged the commits instead of rebasing the branch, apologies

sin-ha:
@bryanwweber i tried but i cant find why would the travis ci build fail for Mac OS i guess my branch is properly rebased isn't it as i don't think the changes are critical to the build to collapse before starting as in the logs

bryanwweber:
@sin-ha I restarted the job, it looked like it failed to install conda.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1803,3 +1803,14 @@ def test_slice_SolutionArray(self):
         soln = ct.SolutionArray(self.gas, 10)
         arr = soln[2:9:3]
         self.assertEqual(len(arr.T), 3)
+
+        #tests for edge cases in sliced solutionArray
```
bryanwweber: This would probably be better as its own test function, `test_zero_length_slice_SolutionArray()`

bryanwweber (changes requested):
Thanks @sin-ha! One small change here, then it's good from me.

interfaces/cython/cantera/test/test_thermo.py
sin-ha: done @bryanwweber

bryanwweber (approved):

speth (approved):

PR #838, opened 2020-04-01
Fixes Indexed assignment of extra columns in SolutionArray

### LINKED ISSUE #829: Indexed assignment of extra columns in SolutionArray fails
**System information**

- Cantera version: 2.5.0a4
- OS: ubuntu
- Python/MATLAB version: Python 3.6.9

**Expected behavior**

The following works
```
In [1]: import cantera as ct
   ...: arr = ct.SolutionArray(gas, 10, extra={'new_column': range(10)})

In [2]: arr.new_column[0] = 10
```

**Actual behavior**

The assignment does not work, i.e.
```
In [3]: arr.new_column
Out[3]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
```

**To Reproduce**

See above.

**Additional context**

Comments in #805.

sin-ha:
**Checklist**

Fixes #829 

-Earlier the self._extra[name] was initialised as a list but when the set item method was called it returned a `np.array()` copy of it 
-So I propose to initialise the init method with the np.array and amended other methods to abide with it
-I think that this would be a better implementation.
- This solves item assignment as the original object gets passed instead of the copy

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -624,7 +624,9 @@ class SolutionArray:
             raise IndexError("Can only append to 1D SolutionArray")
 
         for name, value in self._extra.items():
-            np.append(value, kwargs.pop(name))
+            value = list(value)
+            value.append(kwargs.pop(name))
+            value = np.array(value)
 
         if state is not None:
             self._phase.state = state
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -574,7 +574,7 @@ class SolutionArray:
 
     def __getattr__(self, name):
         if name in self._extra:
-            return np.array(self._extra[name])
+            return self._extra[name]
         else:
             raise AttributeError("'{}' object has no attribute '{}'".format(
                 self.__class__.__name__, name))
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -632,6 +632,16 @@ class SolutionArray:
                 "'{}'".format(", ".join(missing_extra_kwargs))
             )
 
+        # For the checks of the state below, the kwargs dictionary can
+        # only contain keywords that match properties of the state. Here
+        # we pop any kwargs that have to do with the extra items so they
+        # aren't included in that check. They are put into a temporary
+        # storage so that appending can be done at the end of the function
+        # all at once.
+        extra_temp = {}
+        for name in self._extra:
+            extra_temp[name] = kwargs.pop(name)
+
         if state is not None:
             self._phase.state = state
 
@@ -655,6 +665,12 @@ class SolutionArray:
 
         self._states.append(self._phase.state)
         self._indices.append(len(self._indices))
+        for name, value in self._extra.items():
+            # Casting to a list before appending is ~5x faster than using
+            # np.append when appending a single item.
+            v = value.tolist()
+            v.append(extra_temp.pop(name))
+            self._extra[name] = np.array(v)
         self._shape = (len(self._indices),)
 
     @property
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -529,6 +529,10 @@ class SolutionArray:
         reserved = self.__dir__()
 
         self._extra = OrderedDict()
+
+        if isinstance(extra, str):
+            extra = [extra]
+
         if isinstance(extra, dict):
             for name, v in extra.items():
                 if name in reserved:
@@ -536,25 +540,36 @@ class SolutionArray:
                         "Unable to create extra column '{}': name is already "
                         "used by SolutionArray objects.".format(name))
                 if not np.shape(v):
-                    self._extra[name] = [v]*self._shape[0]
-                elif len(v) == self._shape[0]:
-                    self._extra[name] = list(v)
+                    self._extra[name] = np.array([v]).reshape(self._shape)
+                elif len(v) == self._shape or np.array(v).shape == self._shape:
+                    self._extra[name] = np.array(v)
                 else:
                     raise ValueError("Unable to map extra SolutionArray"
                                      "input for named {!r}".format(name))
 
-        elif extra and self._shape == (0,):
-            for name in extra:
-                if name in reserved:
-                    raise ValueError(
-                        "Unable to create extra column '{}': name is already "
-                        "used by SolutionArray objects.".format(name))
-                self._extra[name] = []
+        elif extra is not None:
+            try:
+                iterator = iter(extra)
+            except TypeError :
+                raise ValueError(
+                    "Extra properties can be created by passing an iterable"
+                    " of names for the properties, if the SolutionArray is not initially"
+                    " empty. If you want to supply initial values for the properties, use"
+                    " a dictionary whose keys are the names of the properties and values "
+                    " are the initial values.") from None
 
-        elif extra:
-            raise ValueError("Initial values for extra properties must be "
-                             "supplied in a dict if the SolutionArray is not "
-                             "initially empty")
+            for name in extra:
+                if isinstance(name, str):
+                    if name in reserved:
+                        raise ValueError(
+                            "Unable to create extra column '{}': name is already "
+                            "used by SolutionArray objects.".format(name))
+                    self._extra[name] = np.empty(self._shape)
+
+                else:
+                    raise ValueError(
+                        "Unable to create extra column, passed value {}: "
+                        "is not a string".format(name))
 
         if meta is None:
             self._meta = {}
@@ -609,7 +624,7 @@ class SolutionArray:
             raise IndexError("Can only append to 1D SolutionArray")
 
         for name, value in self._extra.items():
-            value.append(kwargs.pop(name))
+            np.append(value, kwargs.pop(name))
 
         if state is not None:
             self._phase.state = state
@@ -655,7 +670,7 @@ class SolutionArray:
             indices = indices[::-1]
         self._states = [self._states[ix] for ix in indices]
         for k, v in self._extra.items():
-            self._extra[k] = list(np.array(v)[indices])
+            self._extra[k] = np.array(v)[indices]
 
     def equilibrate(self, *args, **kwargs):
         """ See `ThermoPhase.equilibrate` """
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -540,13 +540,12 @@ class SolutionArray:
                         "Unable to create extra column '{}': name is already "
                         "used by SolutionArray objects.".format(name))
                 if not np.shape(v):
-                    self._extra[name] = np.array([v]).reshape(self._shape)
-                elif len(v) == self._shape or np.array(v).shape == self._shape:
+                    self._extra[name] = np.full(self._shape, v)
+                elif np.array(v).shape == self._shape:
                     self._extra[name] = np.array(v)
                 else:
                     raise ValueError("Unable to map extra SolutionArray "
-                                     "input for named {!r}".format(name))
-
+                                     "input named {!r}".format(name))
         elif extra is not None:
             try:
                 iter_extra = iter(extra)
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -544,21 +544,21 @@ class SolutionArray:
                 elif len(v) == self._shape or np.array(v).shape == self._shape:
                     self._extra[name] = np.array(v)
                 else:
-                    raise ValueError("Unable to map extra SolutionArray"
+                    raise ValueError("Unable to map extra SolutionArray "
                                      "input for named {!r}".format(name))
 
         elif extra is not None:
             try:
-                iterator = iter(extra)
+                iter_extra = iter(extra)
             except TypeError :
                 raise ValueError(
-                    "Extra properties can be created by passing an iterable"
-                    " of names for the properties, if the SolutionArray is not initially"
-                    " empty. If you want to supply initial values for the properties, use"
-                    " a dictionary whose keys are the names of the properties and values "
-                    " are the initial values.") from None
+                    "Extra properties can be created by passing an iterable "
+                    "of names for the properties, if the SolutionArray is not initially "
+                    "empty. If you want to supply initial values for the properties, use "
+                    "a dictionary whose keys are the names of the properties and values "
+                    "are the initial values.") from None
 
-            for name in extra:
+            for name in iter_extra:
                 if isinstance(name, str):
                     if name in reserved:
                         raise ValueError(
@@ -567,7 +567,7 @@ class SolutionArray:
                     self._extra[name] = np.empty(self._shape)
 
                 else:
-                    raise ValueError(
+                    raise TypeError(
                         "Unable to create extra column, passed value {}: "
                         "is not a string".format(name))
 
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -905,7 +905,6 @@ class SolutionArray:
         if tabular and len(self._shape) != 1:
             raise AttributeError("Tabular output of collect_data only works "
                                  "for 1D SolutionArray")
-        out = OrderedDict()
 
         # Create default columns (including complete state information)
         if cols is None:
@@ -929,7 +928,6 @@ class SolutionArray:
 
         def split(c, d):
             """ Split attribute arrays into columns for tabular output """
-            single_species = False
             # Determine labels for the items in the current group of columns
             if c in self._n_species:
                 collabels = ['{}_{}'.format(c, s) for s in self.species_names]
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -634,15 +634,19 @@ class SolutionArray:
         elif len(kwargs) == 1:
             attr, value = next(iter(kwargs.items()))
             if frozenset(attr) not in self._phase._full_states:
-                raise KeyError("{} does not specify a full thermodynamic state")
+                raise KeyError(
+                    "'{}' does not specify a full thermodynamic state".format(attr)
+                )
             setattr(self._phase, attr, value)
 
         else:
             try:
                 attr = self._phase._full_states[frozenset(kwargs)]
             except KeyError:
-                raise KeyError("{} is not a valid combination of properties "
-                    "for setting the thermodynamic state".format(tuple(kwargs)))
+                raise KeyError(
+                    "{} is not a valid combination of properties for setting "
+                    "the thermodynamic state".format(tuple(kwargs))
+                ) from None
             setattr(self._phase, attr, [kwargs[a] for a in attr])
 
         self._states.append(self._phase.state)
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -623,10 +623,14 @@ class SolutionArray:
         if len(self._shape) != 1:
             raise IndexError("Can only append to 1D SolutionArray")
 
-        for name, value in self._extra.items():
-            value = list(value)
-            value.append(kwargs.pop(name))
-            value = np.array(value)
+        # This check must go before we start appending to any arrays so that
+        # array lengths don't get out of sync.
+        missing_extra_kwargs = self._extra.keys() - kwargs.keys()
+        if missing_extra_kwargs:
+            raise TypeError(
+                "Missing keyword arguments for extra values: "
+                "'{}'".format(", ".join(missing_extra_kwargs))
+            )
 
         if state is not None:
             self._phase.state = state
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -544,6 +544,10 @@ class SolutionArray:
                     raise ValueError("Unable to map extra SolutionArray "
                                      "input named {!r}".format(name))
         elif extra is not None:
+            if self._shape != (0,):
+                raise ValueError("Initial values for extra properties must be "
+                                 "supplied in a dictionary if the SolutionArray "
+                                 "is not initially empty.")
             if isinstance(extra, np.ndarray):
                 extra = extra.flatten()
             elif isinstance(extra, str):
@@ -551,7 +555,7 @@ class SolutionArray:
 
             try:
                 iter_extra = iter(extra)
-            except TypeError :
+            except TypeError:
                 raise ValueError(
                     "Extra properties can be created by passing an iterable "
                     "of names for the properties. If you want to supply initial "
@@ -568,7 +572,7 @@ class SolutionArray:
                     raise ValueError(
                         "Unable to create extra column '{}': name is already "
                         "used by SolutionArray objects.".format(name))
-                self._extra[name] = np.empty(self._shape)
+                self._extra[name] = np.empty(shape=(0,))
 
         if meta is None:
             self._meta = {}
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -670,7 +670,7 @@ class SolutionArray:
             indices = indices[::-1]
         self._states = [self._states[ix] for ix in indices]
         for k, v in self._extra.items():
-            self._extra[k] = np.array(v)[indices]
+            self._extra[k] = v[indices]
 
     def equilibrate(self, *args, **kwargs):
         """ See `ThermoPhase.equilibrate` """
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -530,9 +530,6 @@ class SolutionArray:
 
         self._extra = OrderedDict()
 
-        if isinstance(extra, str):
-            extra = [extra]
-
         if isinstance(extra, dict):
             for name, v in extra.items():
                 if name in reserved:
@@ -547,28 +544,31 @@ class SolutionArray:
                     raise ValueError("Unable to map extra SolutionArray "
                                      "input named {!r}".format(name))
         elif extra is not None:
+            if isinstance(extra, np.ndarray):
+                extra = extra.flatten()
+            elif isinstance(extra, str):
+                extra = [extra]
+
             try:
                 iter_extra = iter(extra)
             except TypeError :
                 raise ValueError(
                     "Extra properties can be created by passing an iterable "
-                    "of names for the properties, if the SolutionArray is not initially "
-                    "empty. If you want to supply initial values for the properties, use "
-                    "a dictionary whose keys are the names of the properties and values "
-                    "are the initial values.") from None
+                    "of names for the properties. If you want to supply initial "
+                    "values for the properties, use a dictionary whose keys are "
+                    "the names of the properties and values are the initial "
+                    "values.") from None
 
             for name in iter_extra:
-                if isinstance(name, str):
-                    if name in reserved:
-                        raise ValueError(
-                            "Unable to create extra column '{}': name is already "
-                            "used by SolutionArray objects.".format(name))
-                    self._extra[name] = np.empty(self._shape)
-
-                else:
+                if not isinstance(name, str):
                     raise TypeError(
-                        "Unable to create extra column, passed value {}: "
+                        "Unable to create extra column, passed value '{!r}' "
                         "is not a string".format(name))
+                if name in reserved:
+                    raise ValueError(
+                        "Unable to create extra column '{}': name is already "
+                        "used by SolutionArray objects.".format(name))
+                self._extra[name] = np.empty(self._shape)
 
         if meta is None:
             self._meta = {}
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -1745,6 +1745,15 @@ class TestSolutionArray(utilities.CanteraTest):
         self.assertEqual(states.prop.shape, (2, 2))
         self.assertArrayNear(states.prop, np.array(((3, 3), (3, 3))))
 
+    def test_extra_not_empty(self):
+        """Test that a non-empty SolutionArray raises a ValueError if
+           initial values for properties are not supplied.
+        """
+        with self.assertRaisesRegex(ValueError, "Initial values for extra properties"):
+            ct.SolutionArray(self.gas, 3, extra=["prop"])
+        with self.assertRaisesRegex(ValueError, "Initial values for extra properties"):
+            ct.SolutionArray(self.gas, 3, extra=np.array(["prop", "prop2"]))
+
     def test_extra_wrong_shape(self):
         with self.assertRaisesRegex(ValueError, "Unable to map"):
             ct.SolutionArray(self.gas, (3, 3), extra={"prop": np.arange(3)})
@@ -1784,10 +1793,10 @@ class TestSolutionArray(utilities.CanteraTest):
 
     def test_extra_create_by_ndarray(self):
         properties_array = np.array(["prop1", "prop2", "prop3"])
-        states = ct.SolutionArray(self.gas, shape=(2, 6, 9), extra=properties_array)
-        self.assertEqual(states.prop1.shape, (2, 6, 9))
-        self.assertEqual(states.prop2.shape, (2, 6, 9))
-        self.assertEqual(states.prop3.shape, (2, 6, 9))
+        states = ct.SolutionArray(self.gas, shape=(0,), extra=properties_array)
+        self.assertEqual(states.prop1.shape, (0,))
+        self.assertEqual(states.prop2.shape, (0,))
+        self.assertEqual(states.prop3.shape, (0,))
         # Ensure that a 2-dimensional array is flattened
         properties_array = np.array((["prop1"], ["prop2"]))
         states = ct.SolutionArray(self.gas, extra=properties_array)
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -1734,6 +1734,36 @@ class TestSolutionArray(utilities.CanteraTest):
 
         with self.assertRaises(ValueError):
             states = ct.SolutionArray(self.gas, extra=['creation_rates'])
+        states = ct.SolutionArray(self.gas, shape=(0, 0), extra=("prop1"))
+        self.assertEqual(states.prop1.shape, (0, 0))
+
+    def test_extra1(self):
+        states = ct.SolutionArray(self.gas, 7, extra={'prop': range(7)})
+        array = np.array(range(7))
+        self.assertArrayNear(states.prop, array)
+        states.prop[1] = -5
+        states.prop[3:5] = [0, 1]
+        array_mod = np.array([0, -5,  2,  0,  1,  5,  6])
+        self.assertArrayNear(states.prop, array_mod)
+
+    def test_extra2(self):
+        states = ct.SolutionArray(self.gas, shape=(5, 9), extra=["prop1", "prop2"])
+        self.assertEqual(states.prop1.shape, (5, 9))
+        states1 = ct.SolutionArray(self.gas, shape=(0, 0, 0), extra="prop1")
+        self.assertEqual(states1.prop1.shape, (0, 0, 0))
+
+    def test_extra3(self):
+        states2 = ct.SolutionArray(self.gas, shape=(2, 6, 9), extra="prop1")
+        self.assertEqual(states2.prop1.shape, (2, 6, 9))
+
+    def test_extra4(self):
+        states3 = ct.SolutionArray(self.gas, shape=(2, 6, 9), extra=("prop1", "prop2", "prop3"))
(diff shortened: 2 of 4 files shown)


interfaces/cython/cantera/composite.py
```diff
@@ -523,9 +523,9 @@ def __init__(self, phase, shape=(0,), states=None, extra=None):
                         "Unable to create extra column '{}': name is already "
                         "used by SolutionArray objects.".format(name))
                 if not np.shape(v):
-                    self._extra[name] = [v]*self._shape[0]
+                    self._extra[name] = np.array([v]*self._shape[0])
```
ischoegl: I know this is a pre-existing choice, but It may make sense to allow for more dimensions? I.e. this would be a case where we have something like
```
   >>> states = ct.SolutionArray(gas, (6, 10), extra=[‘foo’, ‘bar’])
```

sin-ha: I can't clearly understand what you want to say. I guess you want to say foo and bar to be extra arrays of size (6,10)?
like 
```
>>> states = ct.SolutionArray(gas, (6,10), extra={'foo':sd,"bar":sd})
```
where sd is any (6,10) list/np array
??

ischoegl: @sin-ha ... what I wrote was deliberate - no typo there. It currently throws a `ValueError`, but you could argue that it should initialize `np.ndarray`'s with the full `_shape`, not just the first dimension. From my perspective, the shape of the `_extra` entries should be consistent with other data.

bryanwweber: @ischoegl I'm also not sure what you're suggesting, but it seems like it would be a new feature? In that case, I think it seems outside the scope of this fix.

sin-ha: @ischoegl  As of now extra expects dicts unless the shape of SolutionArray is trivial
- If the user passes a list we should be able to initiate the extra array with same shape.
- else the user will pass that as a dict , in that case we should validate extra with the complete SolutionArray shape instead of ```self._shape[0]```

sin-ha: I also do think this is more of adding another functionality, seems a bit off of the issue so should it be pulled here or open another issue?

ischoegl: I think the point I raised is a trivial addition, and the application was stated in my first post: it’s a reasonable assumption that the arrays are instantiated with correct dimensions based on the information already provided. My guess is that the former single-dimensioned approach was based on `_extra` formerly being list-based. Now that this is no longer the case, putting in a more consistent approach is trivial.

What I am talking about affects some of the same lines that are changed already and won’t take more than a couple of additional lines. Hence, I’d like to see it considered as a friendly amendment of the original issue 😉

interfaces/cython/cantera/composite.py
```diff
@@ -516,21 +516,34 @@ def __init__(self, phase, shape=(0,), states=None, extra=None):
         reserved = self.__dir__()
 
         self._extra = {}
-        if isinstance(extra, dict):
+
+        if isinstance(extra,str):
+            extra = [extra]
+
+        if isinstance(extra, list) and all(isinstance(name, str) for name in extra):
```
ischoegl: That's captures almost everything! However, `extra` could also be an instance of `tuple` or `ndarray` of strings? I believe all of these are supported for when no shape is used for the instantiation of `SolutionArray` (i.e. I am talking about the former `elif extra and self._shape == (0,):` case, in which case any of the above can be iterated over).

sin-ha: that'd be a quick fix but I don't think np.arrays are important?. 
Besides I believe the other part should be removed where `self._shape` is trivial as after this commit it would only be a desperate attempt to map something in extra, probably not useful (is there a particular use case?). it lacks checks and something weird like integers, for instance can pass which can't be used back.e.g.

`>>>>states3 = ct.SolutionArray(gas, shape=(0,0,0), extra=[1])`

ischoegl: I don’t think that the alternative case can be executed for useful content either (you already caught all useful cases), I.e. I believe it can be deleted. If I’m not mistaken, the unit tests should still pass.

On a formatting note, there are a couple of spaces missing after commas (e.g. it should be `shape=(0, 0, 0)` in the previous post; same for some other lines you edited). It’s minor, but the code should remain in compliance with PEP 8.

**PS:** ad ndarrays ... again it’s minor but it should be included for consistency.

sin-ha: Yes the unit tests pass.
sorry for the formatting though.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1643,37 +1643,37 @@ def test_slicing_ndim(self):
...
-            states = ct.SolutionArray(self.gas, extra=['creation_rates'])
-        states = ct.SolutionArray(self.gas, shape=(0,0), extra=("prop1"))
-        self.assertEqual(states.prop1.shape,(0,0))
+            states = ct.SolutionArray(self.gas, extra = ['creation_rates'])
+        states = ct.SolutionArray(self.gas, shape = (0, 0), extra = ("prop1"))
+        self.assertEqual(states.prop1.shape, (0, 0))
 
         # test indexed operations on extra arrays
-        states = ct.SolutionArray(self.gas, 7, extra={'n': range(7)})
+        states = ct.SolutionArray(self.gas, 7, extra = {'prop': range(7)})
```
bryanwweber: I have other thoughts about these changes which I'll add later when I have some time, but as a formatting note: for Python, `=` in keyword arguments should not have a space around them.

interfaces/cython/cantera/composite.py
```diff
@@ -516,32 +516,38 @@ def __init__(self, phase, shape=(0,), states=None, extra=None):
         reserved = self.__dir__()
 
         self._extra = {}
-        if isinstance(extra, dict):
-            for name, v in extra.items():
+
+        if isinstance(extra, str):
+            extra = [extra]
+
+        if isinstance(extra, (list, tuple, np.ndarray)) and all(isinstance(name, str) for name in extra):
```
bryanwweber: I'm not a fan of this approach of trying to exhaustively check the types of an object (referring to the first `isinstance()` check). For instance, I may want to supply the keys or items of a dictionary as an iterable to this function, but:

```python
>>> isinstance({'a': 'b'}.keys(), (list, tuple))
False
```

You might want to do this if you already have a dictionary of data, and you want to reuse just the keys but replace the data. Another example here would be the columns of a Pandas DataFrame, which may be more likely since we have the ability to convert `SolutionArray`s to Pandas DataFrames.

The solution to this problem that I see online is to do:

```python
try:
    iterator = iter(extra)
except TypeError:
    raise ValueError(...) from None
for name in iterator:
    ...
```

Unfortunately, dictionaries would pass this test, but can't be operated on in the same way as a simple sequence like a list (hence the existing separated handling). So here's what I propose:

1. Move the `isinstance(extra, dict)` check to be the first check. I can't think of other objects in Python that behave like dictionaries (i.e., provide mappings) other than perhaps DataFrames. Adding support for DataFrames seems like a sufficiently separate thing that I think it should not be included here, so we can leave the `dict` check as-is (other than my comment on the line below).
2. Have an `else` that does the `try...except...` that I mentioned above. Just watch out that if you do the `all(isinstance(name, str)...)` check on the iterator object that you create from `iter()` that the iterator will be exhausted and you'll have to make a new one, which might be expensive. That said, you could maybe add a separate `assert` like this:
   ```python
   try:
       iterator = iter(extra)
       assert all(isinstance(name, str) for name in extra)
   except TypeError, AssertionError:
       raise ValueError(...) from None
   ```
   In the first line of the `try`, you've already confirmed that the object is, in fact, iterable, so the generator on the assert line should be fine.

Another check that might be good to add somewhere here is if the type is an `np.ndarray`, it is one-dimensional.

bryanwweber: BTW, the `raise ValueError(...) from None`, the `from None` is necessary to avoid a double traceback, since we don't actually care about the `TypeError` or `AssertionError`, which would otherwise be raised in addition to the `ValueError`.

interfaces/cython/cantera/composite.py
```diff
@@ -516,32 +516,38 @@ def __init__(self, phase, shape=(0,), states=None, extra=None):
...
+            for name, v in extra.items():
                 if name in reserved:
                     raise ValueError(
                         "Unable to create extra column '{}': name is already "
                         "used by SolutionArray objects.".format(name))
-                self._extra[name] = []
+                if not np.shape(v):
+                    self._extra[name] = np.array([v]*self._shape)
+                elif len(v) == self._shape or np.array(v).shape == self._shape:
+                    self._extra[name] = np.array(list(v))
```
bryanwweber: I don't think you need the `list(v)` here, since you're converting it to an `np.array()` in the check, `v` should already be able to be directly converted. Plus, converting to a list may change the dimensions.

interfaces/cython/cantera/composite.py
```diff
@@ -516,32 +516,38 @@ def __init__(self, phase, shape=(0,), states=None, extra=None):
...
 
         self._extra = {}
-        if isinstance(extra, dict):
-            for name, v in extra.items():
+
+        if isinstance(extra, str):
+            extra = [extra]
+
+        if isinstance(extra, (list, tuple, np.ndarray)) and all(isinstance(name, str) for name in extra):
+            #if extra is a list and all elements in extra are of type str
```
bryanwweber: Can you delete this comment? Well, given the changes I suggested above it won't be relevant anymore. But even if that check were staying, this comment is explaining **what** rather than **why**. People who read the code on the previous line will already be able to tell that this branch will be executed if extra is a list-like and all elements are of type str. But notice that this comment has already become out-of-date, since `extra` could be more than just a `list` based on the check. You can read more about commenting for why instead of what in many places online, e.g., https://dev.to/andreasklinger/comments-explain-why-not-what-and-2-more-rules-on-writing-good-comments Just leaving this comment for learning purposes :smiley:

interfaces/cython/cantera/composite.py
```diff
@@ -516,32 +516,38 @@ def __init__(self, phase, shape=(0,), states=None, extra=None):
...
+                    self._extra[name] = np.array(list(v))
+                else:
+                    raise ValueError("Unable to map extra SolutionArray"
+                                     "input for named {!r}".format(name))
 
         elif extra:
             raise ValueError("Initial values for extra properties must be"
-                " supplied in a dict if the SolutionArray is not initially"
-                " empty")
+                " supplied in a dict or can be initialised with the property string"
```
bryanwweber: I'd like to improve this error message. What about:

> Extra properties can be created by passing a sequence of names for the properties, if the SolutionArray is not initially empty. If you want to supply initial values for the properties, use a dictionary whose keys are the names of the properties and values are the initial values.

Does that make sense and still capture what we're going for here?

interfaces/cython/cantera/composite.py
```diff
@@ -590,7 +596,7 @@ def append(self, state=None, **kwargs):
             raise IndexError("Can only append to 1D SolutionArray")
 
         for name, value in self._extra.items():
-            value.append(kwargs.pop(name))
+            np.append(value,kwargs.pop(name))
```
```suggestion
            np.append(value, kwargs.pop(name))
```

interfaces/cython/cantera/composite.py
```diff
@@ -628,7 +634,7 @@ def sort(self, col, reverse=False):
             indices = indices[::-1]
         self._states = [self._states[ix] for ix in indices]
         for k, v in self._extra.items():
-            self._extra[k] = list(np.array(v)[indices])
+            self._extra[k] = np.array(v)[indices]
```
bryanwweber: These should already all be arrays, right? So no need to do the `np.array(v)` conversion?

interfaces/cython/cantera/composite.py
bryanwweber: But this isn't correct, `np.append` does not work in-place. See the documentation: https://numpy.org/doc/1.18/reference/generated/numpy.append.html This will most likely be very slow, since NumPy allocates a new array every time `np.append()` is called. I wonder if it would be better to cast the values in `_extra` to lists and use the list append method, and then cast them back to arrays? Maybe the complexity isn't worth it though.

bryanwweber (changes requested):
Hi @sin-ha! Thanks for taking this on. I've got several comments below, I think we need to change how this is implemented. Let me know your thoughts, or if you have any questions/concerns!

bryanwweber:
I forgot to review the test file as well! Sorry, my main comment there was that I think each time you have a comment that delimits a different set of tests, that should be a new test function.

bryanwweber:
@sin-ha Are you going to be able to address my comments? Thanks 😄

sin-ha:
Yes @bryanwweber, actually i was a bit busy with my never ending university exams and forgot about it 
Sorry for the late reply. I will look into everything asap.

bryanwweber:
Thanks @sin-ha I certainly understand having lots of exams 😊

interfaces/cython/cantera/composite.py
sin-ha: This is certainly insightful. Here is a bit I think about the second part though.

> 2\. Have an `else` that does the `try...except...` that I mentioned above. Just watch out that if you do the `all(isinstance(name, str)...)` check on the iterator object that you create from `iter()` that the iterator will be exhausted and you'll have to make a new one, which might be expensive

we are iterating twice over the entire the iterable. I think instead of first checking and then allocating extra properties we can do that simultaneously and raise an error if that is not a string? something like
```
try:
    iter(extra)
    for name in extra:
        if (isinstance(name, str):
        #allocate extra array
       else:
       #raise error
except TypeError, AssertionError:
    raise ValueError(...) from None

```

interfaces/cython/cantera/composite.py
sin-ha: yes, this was relevant more to the original issue #829 .I guess not anymore as `extra` beyond that.

will remove this.

interfaces/cython/cantera/composite.py
sin-ha: This was quite resourceful thank you!! 😄😄.

The entire block could be deleted with the new suggestions

interfaces/cython/cantera/composite.py
sin-ha: this is certainly better but does `sequence` imply an iterable of strings? I don't know about this

interfaces/cython/cantera/composite.py
bryanwweber: OK, I like that suggestion! In that case, you can `raise` a `TypeError` in the `else` clause, and you don't need the `AssertionError`.

interfaces/cython/cantera/composite.py
bryanwweber: A sequence implies any iterable, at least to me, but maybe `iterable` would be a better word there instead of `sequence`.

ischoegl:
@sin-ha ... you may want to rebase on the current version of `main`. I.e. attach your new commits after what has been merged since you last worked on this.

interfaces/cython/cantera/composite.py
sin-ha: I also can't find any other way around [see](https://stackoverflow.com/questions/13215525/how-to-extend-an-array-in-place-in-numpy)

interfaces/cython/cantera/composite.py
sin-ha: yes, this is redundant. was relevant to the original issue somehow I guess

interfaces/cython/cantera/composite.py
bryanwweber: What do you mean, any other way around? You'll have to find some way, since this won't work as expected. I think the simplest method is just to reassign `value`: `value = np.append(value, kwargs.pop(name))` That might be slow, since NumPy has to create a new array every time, but hopefully not *too* slow.

interfaces/cython/cantera/composite.py
```diff
@@ -528,32 +528,47 @@ def __init__(self, phase, shape=(0,), states=None, extra=None, meta=None):
...
                         "used by SolutionArray objects.".format(name))
                 if not np.shape(v):
-                    self._extra[name] = [v]*self._shape[0]
-                elif len(v) == self._shape[0]:
-                    self._extra[name] = list(v)
+                    self._extra[name] = np.array([v]).reshape(self._shape)
+                elif len(v) == self._shape or np.array(v).shape == self._shape:
+                    self._extra[name] = np.array(v)
                 else:
                     raise ValueError("Unable to map extra SolutionArray"
```
bryanwweber: I know you are not responsible for this line, but can you add a space after `SolutionArray` here to fix the formatting? Thanks :smile:

interfaces/cython/cantera/composite.py
```diff
@@ -528,32 +528,47 @@ def __init__(self, phase, shape=(0,), states=None, extra=None, meta=None):
...
                                      "input for named {!r}".format(name))
 
-        elif extra and self._shape == (0,):
+        elif extra is not None:
```
bryanwweber, line 546: Good catch here that `extra` defaults to `None`!

interfaces/cython/cantera/composite.py
```diff
@@ -528,32 +528,47 @@ def __init__(self, phase, shape=(0,), states=None, extra=None, meta=None):
...
+            try:
+                iterator = iter(extra)
+            except TypeError :
+                raise ValueError(
+                    "Extra properties can be created by passing an iterable"
+                    " of names for the properties, if the SolutionArray is not initially"
+                    " empty. If you want to supply initial values for the properties, use"
+                    " a dictionary whose keys are the names of the properties and values "
+                    " are the initial values.") from None
```
bryanwweber: Can you move the space here to the end of the line instead of the start of the line?

interfaces/cython/cantera/composite.py
```diff
@@ -528,32 +528,47 @@ def __init__(self, phase, shape=(0,), states=None, extra=None, meta=None):
...
+
             for name in extra:
```
bryanwweber: You might as well use the iterator you created here, probably best to call it something more specific, like `extra_iter` or something like that.

interfaces/cython/cantera/composite.py
```diff
@@ -528,32 +528,47 @@ def __init__(self, phase, shape=(0,), states=None, extra=None, meta=None):
...
+                            "Unable to create extra column '{}': name is already "
+                            "used by SolutionArray objects.".format(name))
+                    self._extra[name] = np.empty(self._shape)
 
-        elif extra:
-            raise ValueError("Initial values for extra properties must be "
-                             "supplied in a dict if the SolutionArray is not "
-                             "initially empty")
+                else:
+                    raise ValueError(
```
bryanwweber: I think this should be a `TypeError`, since the `name` is of the wrong type.

bryanwweber (changes requested):
Thanks for the changes @sin-ha, just a few things left!

sin-ha:
done @bryanwweber .
Meanwhile, may I also add my name to `Cantera/AUTHORS` ?. I don't know if I qualify though 😅

interfaces/cython/cantera/composite.py
```diff
@@ -608,7 +623,7 @@ def append(self, state=None, **kwargs):
             raise IndexError("Can only append to 1D SolutionArray")
 
         for name, value in self._extra.items():
-            value.append(kwargs.pop(name))
+            np.append(value, kwargs.pop(name))
```
bryanwweber: This line still isn't fixed.

bryanwweber (changes requested):
Thanks @sin-ha. One change still to do, and please feel free to add yourself to `AUTHORS`!

interfaces/cython/cantera/composite.py
sin-ha: I didn't think this through before, Anyways

> I think the simplest method is just to reassign `value`: `value = np.append(value, kwargs.pop(name))

Appending `np.ndarrays` is not as fast 

Turns out converting them to lists and appending (as you said before). does a better job!!,

see https://stackoverflow.com/a/22394181 and https://stackoverflow.com/questions/7133885/fastest-way-to-grow-a-numpy-numeric-array

```
In [2]: %%timeit 
   ...: a = np.empty((0,3), int) 
   ...: l = list(a) 
   ...: for i in range(1000): 
   ...:     l.append([3*i+1,3*i+2,3*i+3]) 
   ...: l = np.array(l)                                                                                                                                                                                                   
467 µs ± 6.32 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [3]: %%timeit 
   ...: a = np.array([3,4,3]) 
   ...: for i in range(1000): 
   ...:     l = list(a) 
   ...:     l.append([3*i+1,3*i+2,3*i+3]) 
   ...:     l = np.array(l) 
   ...:                                                                                                                                                                                                    
2.27 ms ± 531 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [4]: %%timeit 
   ...: a = np.empty((0,3), int) 
   ...: for i in range(1000): 
   ...:     a = np.append(a, 3*i+np.array([[1,2,3]]), 0) 
   ...:                                                                                                                                                                                                    
3.88 ms ± 21.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

```

bryanwweber:
@sin-ha I pushed a few more commits to this branch to fix a few edge cases I found in some testing. For your reference, the changes are in commits b86786b, 1cc53b4, and 8786b51. The other commits fix a few other things I found while working on this code.

The main things to watch out for are:

1) Make sure you test all the branches of an if/else conditional. There was a `ValueError` if you tried to create an extra item from only a single value (for example, `extra={"prop1": 1}`) due to the way you were using `.reshape()`.
2) Similarly, if you make a change, make sure to add a test to make sure that it works. There was no test of the `append()` for extra items. To be fair, there wasn't one before either... Adding a test caused me to add the fixes in 5524e99, 477cbed, and ef1fa99
3) It's helpful to name test functions descriptively. You never actually have to call them anywhere, so using up the entire line with the function name isn't really a problem (sorry, I should have pointed this out before).

bryanwweber:
The test failures are due to using NumPy 1.11. The docs for `.full()` are: https://docs.scipy.org/doc/numpy-1.11.0/reference/generated/numpy.full.html#numpy.full

The default `dtype` for `.full()` in 1.11 is `float`, which means string values can't be included in extra arrays without specifying a different `dtype` on creating. NumPy 1.12 and up use `np.array(fill_value).dtype` as the default, so don't have this problem. I guess if we want to support NumPy 1.11 (which is the version available in the Ubuntu 16.04 repositories), we'll have to specify the `dtype` for all the extra arrays...

interfaces/cython/cantera/composite.py
```diff
@@ -528,32 +528,46 @@ def __init__(self, phase, shape=(0,), states=None, extra=None, meta=None):
...
+                    self._extra[name] = np.array(v)
                 else:
-                    raise ValueError("Unable to map extra SolutionArray"
-                                     "input for named {!r}".format(name))
+                    raise ValueError("Unable to map extra SolutionArray "
+                                     "input named {!r}".format(name))
+        elif extra is not None:
+            if isinstance(extra, np.ndarray):
+                extra = extra.flatten()
+            elif isinstance(extra, str):
```
ischoegl, line 553: I just noticed the possibility of enabling `str` being explicitly considered: please be aware that this may break things elsewhere (e.g. restoration of CSV data to `SolutionArray` etc.). I.e. while I haven’t run any tests, specifically planning for `str` input may require additional ‘fixes’)

Imho, it would be better to raise an error here, and leave data as strictly numeric (which had been my assumption throughout)

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1676,15 +1676,74 @@ def test_slicing_ndim(self):
...
+            ct.SolutionArray(self.gas, extra=[1])
+
+    def test_extra_reserved_names(self):
+        with self.assertRaisesRegex(ValueError, "name is already used"):
+            ct.SolutionArray(self.gas, extra=["creation_rates"])
+
+        with self.assertRaisesRegex(ValueError, "name is already used"):
+            ct.SolutionArray(self.gas, extra={"creation_rates": 0})
+
+    def test_extra_create_by_string(self):
```
ischoegl, line 1781: Same here. I believe allowing for this may create a slew of problems elsewhere.

ischoegl:
@bryanwweber ... while I’m not aware of this, are there instances of `str` assigned to `extra` in current use? If not, I believe it may be worth moving the discussion over to Cantera/enhancements, while wrapping this PR up with an error raised.

interfaces/cython/cantera/composite.py
bryanwweber: Other than possibly breaking CSV/HDF input/output, is there a reason to prohibit strings? Would it be better to raise that error in the CSV/HDF input/output methods, if there are no other problems with using strings as extra items?

interfaces/cython/cantera/test/test_thermo.py
bryanwweber: This creates one empty `extra` item by passing a single string, the values in the array are not strings. I believe this should be left.

bryanwweber:
@ischoegl 

> while I’m not aware of this, are there instances of `str` assigned to extra in current use?

There's no way to know the answer to this question... There are none in our examples (AFAIK), but there may be in user code. If you determine there are problems with having strings in the `extra` arrays, I would say that'd be a bug and you should file an issue here. I don't have time at this exact moment to check the potential spots you mentioned, but I will check them if I don't see any updates from you when I can come back to this.

interfaces/cython/cantera/test/test_thermo.py
ischoegl: Correct - this is resolved.

interfaces/cython/cantera/composite.py
bryanwweber: Actually, this is just checking the name of the extra items, not the values. So this code won't need to change at all.

ischoegl: As mentioned, I had never considered str to be a sanctioned use in the first place.

bryanwweber: For the names of the columns? In that case, I'd say that the HDF/CSV code needs to change, if it is broken.

ischoegl:
> @ischoegl
> 
> > while I’m not aware of this, are there instances of `str` assigned to extra in current use?
> 
> There's no way to know the answer to this question... There are none in our examples (AFAIK), but there may be in user code. If you determine there are problems with having strings in the `extra` arrays, I would say that'd be a bug and you should file an issue here. I don't have time at this exact moment to check the potential spots you mentioned, but I will check them if I don't see any updates from you when I can come back to this.

I can look into this, but would rather confirm, before settling on not raising an error here.

I am frankly not sure whether it’s worth supporting a use case that is (presumably?) rare while complicating implementation.

bryanwweber:
As I said above, supporting _names_ of extra "columns" being strings is critical IMO, and there's no reason to expect they won't be strings. I believe most of our examples use strings as names of the columns.

For _values_ in the column, that is definitely more rare. Nonetheless, we do have the `phase_of_matter` attribute, which returns a string, so there's at least some possibility for using strings as values with built-in functionality (although not as `extra`).

ischoegl:
OK, here are the results:

```
In [1]: import cantera as ct
    ...: arr = ct.SolutionArray(gas, 3, extra={'spam': 'eggs'})
    ...: 

In [2]: arr._extra
Out[2]: OrderedDict([('spam', ['eggs', 'eggs', 'eggs'])])

In [3]: !cat test.csv
spam,T,density,Y_H2,Y_H,Y_O,Y_O2,Y_OH,Y_H2O,Y_HO2,Y_H2O2,Y_AR
eggs,300.0,0.08189392763801234,1.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
eggs,300.0,0.08189392763801234,1.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
eggs,300.0,0.08189392763801234,1.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0

In [4]: arr2 = ct.SolutionArray(gas, extra={'spam'})

In [4]: arr2.read_csv('test.csv')

In [5]: arr2._extra
Out[5]: {'spam': array([ nan,  nan,  nan])}

In [6]: arr.write_hdf('test.h5')
<ipython-input-26-d708cd336db5> in <module>()
----> 1 arr.write_hdf('test.h5')

/usr/local/lib/python3.6/dist-packages/cantera/composite.py in write_hdf(self, filename, cols, group, subgroup, attrs, mode, append, compression, compression_opts, *args, **kwargs)
   1117                 dgroup.attrs[key] = val
   1118             for header, col in data.items():
-> 1119                 dgroup.create_dataset(header, data=col, **hdf_kwargs)
   1120 
   1121         return group

/home/docker/.local/lib/python3.6/site-packages/h5py/_hl/group.py in create_dataset(self, name, shape, dtype, data, **kwds)
    134 
    135         with phil:
--> 136             dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
    137             dset = dataset.Dataset(dsid)
    138             if name is not None:

/home/docker/.local/lib/python3.6/site-packages/h5py/_hl/dataset.py in make_new_dset(parent, shape, dtype, data, chunks, compression, shuffle, fletcher32, maxshape, compression_opts, fillvalue, scaleoffset, track_times, external, track_order, dcpl)
    116         else:
    117             dtype = numpy.dtype(dtype)
--> 118         tid = h5t.py_create(dtype, logical=1)
    119 
    120     # Legacy

h5py/h5t.pyx in h5py.h5t.py_create()

h5py/h5t.pyx in h5py.h5t.py_create()

h5py/h5t.pyx in h5py.h5t.py_create()

TypeError: No conversion path for dtype: dtype('<U4')
```
I.e. the pre-existing `write_csv` supports `str`, the newly introduced `read_csv` fails to import it, and `h5py` has issues.

ischoegl:
> As I said above, supporting names of extra "columns" being strings is critical IMO, and there's no reason to expect they won't be strings. I believe most of our examples use strings as names of the columns.

That was my bad, :cry:

> For values in the column, that is definitely more rare. Nonetheless, we do have the phase_of_matter attribute, which returns a string, so there's at least some possibility for using strings as values with built-in functionality (although not as extra).

It's not written to the file, unless forced as `col` output ... certainly an edge case?

ischoegl:
Concluding, there are two options:

1. Disallow `str` here (assuming that there is no user code), or
2. Open an issue to fix the export/import issue

Further discussion probably needs to be moved (either to an issue or `Cantera/enhancements`)

**PS:** I filed #896 as the `str` issue is tangential to this PR.

bryanwweber:
Thanks for investigating @ischoegl! I think a new issue is the right call here 👍😊 I marked all the related comments in this thread as resolved to avoid distracting from the main point of this PR.

interfaces/cython/cantera/composite.py
ischoegl: No, values - at this point, this can be considered resolved here, as I opened an issue in #896

ischoegl:
> The test failures are due to using NumPy 1.11. The docs for `.full()` are: https://docs.scipy.org/doc/numpy-1.11.0/reference/generated/numpy.full.html#numpy.full
> 
> The default `dtype` for `.full()` in 1.11 is `float`, which means string values can't be included in extra arrays without specifying a different `dtype` on creating. NumPy 1.12 and up use `np.array(fill_value).dtype` as the default, so don't have this problem. I guess if we want to support NumPy 1.11 (which is the version available in the Ubuntu 16.04 repositories), we'll have to specify the `dtype` for all the extra arrays...

Ubuntu 16.04 support ends April 2021, which is less than a year out. Citing a conservative (albeit current) distro, CentOS8 uses `numpy` 1.14. GitHub's CI still uses 18.04 as `ubuntu-latest`, but it likely won't be long before it is upgraded to 20.04. I guess it boils down to a similar question as #891, where it's only a system-installed version that will cause issues?

**PS:** It appears that numpy requirements are actually not strictly enforced in `SConstruct`, and there’s only a warning issued?

bryanwweber:
Thanks for the thoughts on NumPy @ischoegl. It is really good to know that a conservative distribution like CentOS includes 1.14. Until this PR, there has not been a required minimum version of NumPy, we have been able to be flexible since we hadn't used very many of the user-interface features (basically, we just used the C headers to provide arrays in Cython). Hence, it was only necessary to provide a warning for versions that we hadn't tried of NumPy. Going forward, it seems it may be necessary to enforce an actual minimum version.

ischoegl:
ad `numpy` ... Ubuntu 18.04 is still on 1.13.3, so this still puts some restrictions on new features, e.g. #896. I ended up testing for `distutils.version.LooseVersion` in #900 ...

interfaces/cython/cantera/composite.py
```diff
@@ -528,32 +528,46 @@ def __init__(self, phase, shape=(0,), states=None, extra=None, meta=None):
...
                     raise ValueError(
                         "Unable to create extra column '{}': name is already "
                         "used by SolutionArray objects.".format(name))
-                self._extra[name] = []
-
-        elif extra:
-            raise ValueError("Initial values for extra properties must be "
-                             "supplied in a dict if the SolutionArray is not "
-                             "initially empty")
+                self._extra[name] = np.empty(self._shape)
```
speth: I don't the 'extra' arrays should be allowed to contain uninitialized values. The existing behavior of requiring initial values if the `SolutionArray` is not empty should be preserved.

speth (changes requested):
From reading the discussion, I think all of the previously-mentioned issues have been hashed out at this point, right? I just had one case where I think the behavior should change.

bryanwweber (approved):

interfaces/cython/cantera/composite.py
ischoegl: @speth ... I’d ask to consider retaining this behavior (while potentially avoiding `np.empty`).

Think of a scenario where `extra` is calculated based on the content of the `SolutionArray` itself (which typically receives the actual value using some setter only after it is instantiated, so the content of extra is not known a priori). For this case, you’d have to provide dummy values to create `extra` columns, which would be subsequently overwritten. I believe that automatically initializing values as zeros would be a lot more convenient (and avoid `np.empty`); columns that contain strings are presumably rare, so having to provide a value there would be ok.

bryanwweber: @ischoegl I originally had the same objection as you, but I considered it for a while and I ended up agreeing with @speth. There are two main reasons for this, as I see it:

1. It is trivially easy to specify a dummy value, since you do not need to specify the shape. For instance:

   ```python
   >>> arr = ct.SolutionArray(gas, (3, 10), extra={"prop": 0, "prop2": 0.0})
   >>> arr.prop
   array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
          [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
          [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]])
   >>> arr.prop2
   array([[0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
          [0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]])
   >>> # This is not allowed right now, but would be almost the same, except the dtype
   >>> arr = ct.SolutionArray(gas, (3, 10), extra=["prop", "prop2"])
   ```

   Since I've used `np.full()` to fill the initial value, there's no need to give an array as the initial value.

2. It isn't clear what `dtype` should be used - it is not only strings, but whether one wants to use floats vs. ints or any other dtype as well. I suppose floats could be used in place of ints, but there are legitimate use cases for the int dtype, for instance, as a categorical or level variable. I think it would also not be clear why one would have to specify an initial value for some dtypes but not for others.

ischoegl: @bryanwweber ... I didn’t think of the scalar route, so your argument is convincing.

speth: The availability of the scalar initializer is a nice compromise for the case where the value is going to be calculated anyway.

If I'm reading this block of code correctly now, the only time that this line can be called is if `self._shape` is `(0,)`, right? It might be more clear if the value was just set to `self._extra[name] = np.empty(0)` then.

bryanwweber: Yes, that's correct, it only runs if `self._shape == (0,)`. I'm not sure it's clearer to just pass `0` though. Shapes of NumPy arrays are always returned as tuples, and just writing `np.empty(0)` implies to me that `0` will be the value in the array (notwithstanding the name `empty`), not that there's a shape of `0`. So I think it would be better to use `np.empty(shape=(0,))`, which is then just `self._shape`, and putting in `(0,)` I thought would be more confusing than using `self._shape`, because the question arises, if these are equal why not just use the existing variable name?

speth: I think explicitly showing the size as zero (with whatever notation you like) is preferable because it makes it clear that this doesn't result in an array with uninitialized values. Tracing the logic back to see that `self._shape` is `(0,)` if we're in this branch is not really that simple.

bryanwweber: OK, this is fixed now with `shape=(0,)` 👍

speth (approved):

PR #836, opened 2020-03-29
[1D] Make tolerance setters more flexible

ischoegl:
**Checklist**

- [x] There is a clear use-case for this code change
- [x] The commit message has a short title & references relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] The pull request is ready for review

**If applicable, fill in the issue number this pull request is fixing**

Fixes #833

**Changes proposed in this pull request**

The docstrings of `Domain1D.set_transient_tolerances` and `Domain1D.set_steady_tolerances` state: [...] keywords *abs* and *rel* can be used to specify arrays for the absolute and relative tolerances for each solution component. The wording of the docstring is ambiguous, as in practice, no action is taken if only one of the keyword arguments is provided (also, no warning is issued).

This commit ensures the expected behavior if only one of the two arguments (abs or rel) is supplied.

PR #834, opened 2020-03-29
[thermo] Ensure ordered extra columns in SolutionArray CSV output

ischoegl:
**Checklist**

- [x] There is a clear use-case for this code change
- [x] The commit message has a short title & references relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] The pull request is ready for review

**If applicable, fill in the issue number this pull request is fixing**

Fixes #832

**Changes proposed in this pull request**

Ensure that extra columns of `SolutionArray` CSV output preserve order (resolves issue on Python systems where `dict` does not enforce order).

PR #840, opened 2020-04-03
Implement custom HDF storage structure

### LINKED ISSUE #830: Make HDF5 output format usable outside of pandas
The HDF5 output option added to the `SolutionArray` class currently uses the `pandas` library to write the HDF5 files. However, `pandas`-generated HDF5 files introduce an additional layer of abstraction beyond the native structure of HDF5 datasets which makes them difficult to use except with `pandas`. For example, saving a `SolutionArray` generates an HDF5 file with keys:
```
df/axis0
df/axis1
df/block0_items
df/block0_values
```
where `df` is the given name of the dataset, `df/axis0` contains the column names, `df/axis1` is a list of integers, `df/block0_items` contains the column names again, and `df/block0_values` contains the 2D array of actual values.

Using a simpler, more direct structure would allow these files to be easily used with other libraries in Python (e.g. `h5py` and `pytables`) as well as in other languages such as Matlab or C++. For example, this could simply be
```
df/T
df/density
df/Y_H2
df/Y_H
...
```
where each dataset was the corresponding array of data. One question is whether it would make sense to store `df/Y` as a 2D array with an additional array for the species names. That would make it easier to do some operations when working with the same mechanism file, but make loading data for a different mechanism (with different species) more difficult.

Since the HDF5 output is new for version 2.5, we can still make changes without having to worry about backwards compatibility, and I think we should do so while we can.

This also applies to the HDF5 output proposed in #805.

### LINKED ISSUE #832: Order of `extra` columns in SolutionArray is not preserved
**System information**

- Cantera version: 2.5.0a4 (and earlier?)
- OS: Ubuntu Xenial (16.04)
- Python/MATLAB version: Python 3.5.2

**Expected behavior**

`SolutionArray.write_csv` writes `extra` columns in correct order regardless of OS / Python version.

**Actual behavior**

Some columns are swapped for some OS / Python versions.

**To Reproduce**

```
In [1]: import cantera as ct
  ... : gas = ct.Solution('h2o2.yaml')
  ... : arr = ct.SolutionArray(gas, 10, extra={'grid': range(10), 'velocity': range(10)})

In [2]: arr._extra
Out[2]: 
{'velocity': [0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
 'grid': [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]}

In [4]: !head test.csv --lines=3
velocity,grid,T,density,Y_H2,Y_H,Y_O,Y_O2,Y_OH,Y_H2O,Y_HO2,Y_H2O2,Y_AR
0.0,0.0,300.0,0.081893927638,1.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
1.0,1.0,300.0,0.081893927638,1.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
```

**Thoughts:**

* This is obviously linked to Python dictionaries not (necessarily) preserving order
* An option that preserves an `OrderedDict` would require a tweak of internal storage, where the following should result in the expected behavior:
```python
arr = ct.SolutionArray(gas, 10, extra=OrderedDict([('grid', range(10)), 
                                                   ('velocity', range(10))]))
```

### LINKED ISSUE #833: Domain1D does not allow for tolerances to be set by arrays
**System information**

- Cantera version: 2.5.0a4
- OS: Ubuntu Xenial (16.04)
- Python/MATLAB version: 3.5.2

**Expected behavior**

In `Domain1D.set_transient_tolerances` and `Domain1D.set_steady_tolerances`, the keywords *abs* and *rel* can be used to specify arrays for the absolute and relative tolerances for each solution component. (per docstring)

**Actual behavior**

The setters do not work as specified, i.e.
```
In [1]: import cantera as ct
   ...: f = ct.BurnerFlame(gas)

In [2]: sr = f.flame.steady_reltol()
   ...: sr
Out[2]: 
array([  1.00000000e-04,   1.00000000e-04,   2.00000000e-07,
         1.00000000e-04,   1.00000000e-04,   1.00000000e-04,
         1.00000000e-04,   1.00000000e-04,   1.00000000e-04,
         1.00000000e-04,   1.00000000e-04,   1.00000000e-04,
         1.00000000e-04,   1.00000000e-04])

In [3]: sr[0] *= 10
   ...: sr
Out[3]:
array([  1.00000000e-03,   1.00000000e-04,   2.00000000e-07,
         1.00000000e-04,   1.00000000e-04,   1.00000000e-04,
         1.00000000e-04,   1.00000000e-04,   1.00000000e-04,
         1.00000000e-04,   1.00000000e-04,   1.00000000e-04,
         1.00000000e-04,   1.00000000e-04])

In [4]: f.flame.set_steady_tolerances(rel=sr)
   ...: f.flame.steady_reltol()
Out[4]: 
array([  1.00000000e-04,   1.00000000e-04,   2.00000000e-07,
         1.00000000e-04,   1.00000000e-04,   1.00000000e-04,
         1.00000000e-04,   1.00000000e-04,   1.00000000e-04,
         1.00000000e-04,   1.00000000e-04,   1.00000000e-04,
         1.00000000e-04,   1.00000000e-04])

```
The same issue occurs for the keyword `abs` in both `set_steady_tolerances` and `set_transient_tolerances`.

**To Reproduce**
See above.

**Thoughts**

~A fix could introduce more Pythonic properties, with getters and setters for `steady_reltol` etc. (although different names are required so the old methods can be deprecated).~

ischoegl:
**Checklist**

Fixes #830, fixes #832 (PR supersedes #834), fixes #833 (PR incorporates #836).

- The previously implemented `SolutionArray` HDF output option uses the `pandas` library to write HDF5 files. However, `pandas` introduces an additional layer of abstraction beyond the native structure of HDF5 datasets.
- A simpler, more direct structure ensures HDF output portability to other languages such as Matlab or C++.
- Add `meta` attribute to `SolutionArray` to facilitate storage and restoration of meta data (e.g. settings of `Domain1D` objects).

This PR proposes the following implementation for a single `SolutionArray`:
```
some_group_name/d0/solution/attrs['name'] = 'gri30'
some_group_name/d0/solution/attrs['source'] = 'gri30.xml
some_group_name/d0/T
some_group_name/d0/density
some_group_name/d0/Y_H2
some_group_name/d0/Y_H
...
```
For `FlameBase` output, the intermediate layer `d{}` allows for multiple `SolutionArrays` to be stored in a single group. 
```
some_group_name/attrs['type`] = 'FreeFlame'
some_group_name/attrs[<setting1>] = ...
some_group_name/attrs[<setting2>] = ...
... 
some_group_name/d0/attrs['type'] = 'Inlet1D'
some_group_name/d0/solution/attrs['name'] = 'ohmech'
some_group_name/d0/solution/attrs['source'] = 'h2o2.yaml'
some_group_name/d0/T
some_group_name/d0/density
some_group_name/d0/Y_H2
some_group_name/d0/Y_H
...
some_group_name/d1/attrs['type'] = 'Domain1D'
some_group_name/d1/solution/attrs['name'] = 'ohmech'
some_group_name/d1/solution/attrs['source'] = 'h2o2.yaml'
some_group_name/d1/T
some_group_name/d1/density
some_group_name/d1/Y_H2
some_group_name/d1/Y_H
...
some_group_name/d2/attrs['type'] = 'Outlet1D'
some_group_name/d2/solution/attrs['name'] = 'ohmech'
some_group_name/d2/solution/attrs['source'] = 'h2o2.yaml'
```

### DIFF
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -37,6 +37,12 @@ public:
     //! Set the name of this Solution object
     void setName(const std::string& name);
 
+    //! Return the source of this Solution object
+    std::string source() const { return m_source; }
+
+    //! Set the source of this Solution object
+    void setSource(const std::string& source) { m_source = source; }
+
     //! Set the ThermoPhase object
     void setThermo(shared_ptr<ThermoPhase> thermo);
 
@@ -62,6 +68,8 @@ public:
     }
 
 protected:
+    std::string m_source; //!< Source of Solution (e.g. file name)
+
     shared_ptr<ThermoPhase> m_thermo;  //!< ThermoPhase manager
     shared_ptr<Kinetics> m_kinetics;  //!< Kinetics manager
     shared_ptr<Transport> m_transport;  //!< Transport manager
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -37,12 +37,6 @@ public:
     //! Set the name of this Solution object
     void setName(const std::string& name);
 
-    //! Return the source of this Solution object
-    std::string source() const { return m_source; }
-
-    //! Set the source of this Solution object
-    void setSource(const std::string& source) { m_source = source; }
-
     //! Set the ThermoPhase object
     void setThermo(shared_ptr<ThermoPhase> thermo);
 
@@ -68,8 +62,6 @@ public:
     }
 
 protected:
-    std::string m_source; //!< Source of Solution (e.g. file name)
-
     shared_ptr<ThermoPhase> m_thermo;  //!< ThermoPhase manager
     shared_ptr<Kinetics> m_kinetics;  //!< Kinetics manager
     shared_ptr<Transport> m_transport;  //!< Transport manager
--- a/interfaces/cython/cantera/base.pyx
+++ b/interfaces/cython/cantera/base.pyx
@@ -85,6 +85,13 @@ cdef class _SolutionBase:
         def __set__(self, name):
             self.base.setName(stringify(name))
 
+    property source:
+        """
+        The source of this object (e.g. a file name)
+        """
+        def __get__(self):
+            return pystr(self.base.source())
+
     property composite:
         """
         Returns tuple of thermo/kinetics/transport models associated with
@@ -108,8 +115,10 @@ cdef class _SolutionBase:
         cdef CxxAnyMap root
         if infile:
             root = AnyMapFromYamlFile(stringify(infile))
+            self.base.setSource(stringify(infile))
         elif source:
             root = AnyMapFromYamlString(stringify(source))
+            self.base.setSource(stringify('custom YAML'))
 
         phaseNode = root[stringify("phases")].getMapWhere(stringify("name"),
                                                           stringify(name))
@@ -142,8 +151,10 @@ cdef class _SolutionBase:
         """
         if infile:
             rootNode = CxxGetXmlFile(stringify(infile))
+            self.base.setSource(stringify(infile))
         elif source:
             rootNode = CxxGetXmlFromString(stringify(source))
+            self.base.setSource(stringify('custom CTI/XML'))
 
         # Get XML data
         cdef XML_Node* phaseNode
@@ -180,6 +191,7 @@ cdef class _SolutionBase:
         Instantiate a set of new Cantera C++ objects based on a string defining
         the model type and a list of Species objects.
         """
+        self.base.setSource(stringify('custom parts'))
         self.thermo = newThermoPhase(stringify(thermo))
         self._thermo.reset(self.thermo)
         self.thermo.addUndefinedElements()
--- a/interfaces/cython/cantera/base.pyx
+++ b/interfaces/cython/cantera/base.pyx
@@ -35,6 +35,7 @@ cdef class _SolutionBase:
             self.kinetics = other.kinetics
             self.transport = other.transport
             self._base = other._base
+            self._source = other._source
             self._thermo = other._thermo
             self._kinetics = other._kinetics
             self._transport = other._transport
@@ -45,6 +46,7 @@ cdef class _SolutionBase:
 
         # Assign base and set managers to NULL
         self._base = CxxNewSolution()
+        self._source = None
         self.base = self._base.get()
         self.thermo = NULL
         self.kinetics = NULL
@@ -87,10 +89,10 @@ cdef class _SolutionBase:
 
     property source:
         """
-        The source of this object (e.g. a file name)
+        The source of this object (such as a file name).
         """
         def __get__(self):
-            return pystr(self.base.source())
+            return self._source
 
     property composite:
         """
@@ -115,10 +117,10 @@ cdef class _SolutionBase:
         cdef CxxAnyMap root
         if infile:
             root = AnyMapFromYamlFile(stringify(infile))
-            self.base.setSource(stringify(infile))
+            self._source = infile
         elif source:
             root = AnyMapFromYamlString(stringify(source))
-            self.base.setSource(stringify('custom YAML'))
+            self._source = 'custom YAML'
 
         phaseNode = root[stringify("phases")].getMapWhere(stringify("name"),
                                                           stringify(name))
@@ -151,10 +153,10 @@ cdef class _SolutionBase:
         """
         if infile:
             rootNode = CxxGetXmlFile(stringify(infile))
-            self.base.setSource(stringify(infile))
+            self._source = infile
         elif source:
             rootNode = CxxGetXmlFromString(stringify(source))
-            self.base.setSource(stringify('custom CTI/XML'))
+            self._source = 'custom CTI/XML'
 
         # Get XML data
         cdef XML_Node* phaseNode
@@ -191,7 +193,7 @@ cdef class _SolutionBase:
         Instantiate a set of new Cantera C++ objects based on a string defining
         the model type and a list of Species objects.
         """
-        self.base.setSource(stringify('custom parts'))
+        self._source = 'custom parts'
         self.thermo = newThermoPhase(stringify(thermo))
         self._thermo.reset(self.thermo)
         self.thermo.addUndefinedElements()
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -500,7 +500,7 @@ class SolutionArray:
 
     _purefluid_scalar = ['Q']
 
-    def __init__(self, phase, shape=(0,), states=None, extra=None, meta={}):
+    def __init__(self, phase, shape=(0,), states=None, extra=None, meta=None):
         self._phase = phase
 
         if isinstance(shape, int):
@@ -555,7 +555,10 @@ class SolutionArray:
                              "supplied in a dict if the SolutionArray is not "
                              "initially empty")
 
-        self._meta = meta
+        if meta is None:
+            self._meta = {}
+        else:
+            self._meta = meta
 
     def __getitem__(self, index):
         states = self._states[index]
@@ -629,7 +632,8 @@ class SolutionArray:
     @property
     def meta(self):
         """
-        Metadata
+        Dictionary holding information describing the `SolutionArray`. Metadata
+        should be provided for the creation of `SolutionArray` objects.
         """
         return self._meta
 
@@ -663,7 +667,9 @@ class SolutionArray:
         dictionary. Thus, this method allows to restore data exported by
         `collect_data`.
 
-        :param data: dictionary holding data to be restored.
+        :param data: Dictionary holding data to be restored, where keys
+            refer to thermodynamic states (e.g. ``T``, ``P``) or extra
+            entries, and values contain corresponding data.
 
         The receiving `SolutionArray` either has to be empty or should have
         matching dimensions. Essential state properties and extra entries
@@ -831,7 +837,7 @@ class SolutionArray:
             self._phase.set_equivalence_ratio(phi[index], *args, **kwargs)
             self._states[index][:] = self._phase.state
 
-    def collect_data(self, cols=None, tabular=False, threshold=0, species='Y'):
+    def collect_data(self, cols=None, tabular=False, threshold=0, species=None):
         """
         Returns the data specified by *cols* in an ordered dictionary, where
         keys correspond to SolutionArray attributes to be exported.
@@ -843,7 +849,8 @@ class SolutionArray:
             include any arrays which were specified as 'extra' variables when
             defining the `SolutionArray` object. The special value 'extra' can
             be used to include all 'extra' variables.
-        :param tabular: Split 2D data into 1D columns
+        :param tabular: Split 2D data into separate 1D columns for each
+            species / reaction
         :param threshold: Relative tolerance for including a particular column
             if tabular output is enabled. The tolerance is applied by comparing
             the maximum absolute value for a particular column to the maximum
@@ -860,6 +867,11 @@ class SolutionArray:
         # Create default columns (including complete state information)
         if cols is None:
             cols = ('extra',) + self._phase._native_state
+
+        if species is None:
+            species = cols[-1]
+        else:
+            cols = cols[:-1] + (species,)
 
         # Expand cols to include the individual items in 'extra'
         expanded_cols = []
@@ -887,7 +899,7 @@ class SolutionArray:
                 collabels = [c]
 
             if d.ndim > 1:
-                if threshold:
+                if threshold and d.size:
                     # Determine threshold value and select columns to keep
                     maxval = abs(d).max()
                     keep = (abs(d) > threshold * maxval).any(axis=0)
@@ -903,6 +915,7 @@ class SolutionArray:
         species_names = set(self.species_names)
         for c in expanded_cols:
             if c in species_names:
+
                 d = getattr(self(c), species)
             elif c in attrs:
                 d = getattr(self, c)
@@ -1009,7 +1022,7 @@ class SolutionArray:
         :param attrs:
             Dictionary of user-defined group attributes.
         :param mode:
-            Mode to open the file {'a' (default), 'w', 'r+}.
+            Mode to open the file {'a' (default), 'w', 'r+'}.
         :param append:
             If False, the content of a pre-existing group is deleted before
             writing the `SolutionArray` in the first position. If True, the
@@ -1020,6 +1033,8 @@ class SolutionArray:
         :param compression_opts:
             Options for the h5py compression filter; for 'gzip', this
             corresponds to the compression level {None, 0-9}.
+        :return:
+            Identifiers for group and name used for storing HDF data.
 
         Arguments *compression*, and *compression_opts* are mapped to parameters
         for `h5py.create_dataset`; in both cases, the choices of `None` results
@@ -1044,7 +1059,6 @@ class SolutionArray:
         with _h5py.File(filename, mode) as hdf:
 
             # check existence of tagged item
-            msg = "HDF group with group identifier '{}' exists in '{}': {}"
             if not group:
                 # add group with default name
                 group = 'group{}'.format(len(hdf.keys()))
@@ -1103,6 +1117,8 @@ class SolutionArray:
             `Solution` object), with an error being raised if the current source
             does not match the original source. If True, the error is
             suppressed.
+        :return: User-defined attributes provided to describe the group holding
+            the `SolutionArray` information.
 
         The method imports data using `restore_data` and requires a working
         installation of h5py (`h5py` can be installed using pip or conda).
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -984,8 +984,8 @@ class SolutionArray:
             data_dict[label] = data[:, i]
         self.restore_data(data_dict)
 
-    def write_hdf(self, filename, *args, cols=None, group=None, attrs={},
-                  mode='a', append=False,
+    def write_hdf(self, filename, *args, cols=None, group=None, name=None,
+                  attrs={}, mode='a', append=False,
                   compression=None, compression_opts=None, **kwargs):
         """
         Write data specified by *cols* to a HDF container file named *filename*.
@@ -999,7 +999,13 @@ class SolutionArray:
             A list of any properties of the solution being exported.
         :param group:
             Identifier for the group in the container file. A group may contain
-            multiple `SolutionArray` objects.
+            multiple `SolutionArray` objects. If 'None', group names default
+            to 'groupN', with N being the number of pre-existing groups within
+            the HDF container file.
+        :param name:
+            Name identifier for a subgroup representing the `SolutionArray`
+            object to be added. If 'None', subgroup names default to 'arrM'
+            with M being the number of pre-existing subgroups within the group.
         :param attrs:
             Dictionary of user-defined group attributes.
         :param mode:
@@ -1064,21 +1070,22 @@ class SolutionArray:
                 root.attrs[attr] = value
 
             # add subgroup containing data
-            sub_name = 'd{}'.format(count)
-            sub = root.create_group(sub_name)
+            if name is None:
+                name = 'arr{}'.format(count)
+            sub = root.create_group(name)
             for key, val in self._meta.items():
                 sub.attrs[key] = val
             for header, col in data.items():
                 sub.create_dataset(header, data=col, **hdf_kwargs)
 
             # add subgroup containing information on gas
-            sol = sub.create_group('solution')
+            sol = sub.create_group('phase')
             sol.attrs['name'] = self.name
             sol.attrs['source'] = self.source
 
-        return group, sub_name
+        return group, name
 
-    def read_hdf(self, filename, group=None, index=0, force=False):
+    def read_hdf(self, filename, group=None, name=None, force=False):
         """
         Read a dataset from a HDF container file and restore data to the
         `SolutionArray` object. This method allows for recreation of data
@@ -1088,8 +1095,9 @@ class SolutionArray:
             are `.hdf`, `.hdf5` or `.h5`.
         :param group: Identifier for the group in the container file. A group
             may contain multiple `SolutionArray` objects.
-        :param index: If data was previously written using the `append` option,
-            the option *index* specifies the location within the group.
+        :param name:
+            Name identifier for a subgroup representing the `SolutionArray`
+            object to be read. If 'None', the first subgroup added.
         :param force: If False, matching `SolutionArray` source identifiers are
             enforced (e.g. input file used for the creation of the underlying
             `Solution` object), with an error being raised if the current source
@@ -1111,23 +1119,26 @@ class SolutionArray:
                 group = groups[0]
 
             if not (group in hdf.keys()):
-                msg = "HDF file does not contain group '{}'"
-                raise IOError(msg.format(group))
+                msg = ("HDF file does not contain group '{}'; "
+                       "available groups are: {}")
+                raise IOError(msg.format(group, list(hdf.keys())))
 
             # load root and attributes
             root = hdf[group]
             root_attrs = {attr: value for attr, value in root.attrs.items()}
 
             # identify subgroup
-            sub_names = ['d{}'.format(i) for i, key in enumerate(root.keys())]
+            sub_names = list(root.keys())
             if not len(sub_names):
                 msg = "HDF group '{}' does not contain valid data"
                 raise IOError(msg.format(group))
-            elif index > len(sub_names) - 1:
-                msg = ("Index '{}' exceeds the number of data sets stored "
-                       "within HDF group '{}' ('{}')")
-                raise IOError(msg.format(index, group, len(sub_names)))
-            sub = root[sub_names[index]]
+            elif name is None:
+                name = sub_names[0]
+            elif name not in sub_names:
+                msg = ("HDF file does not contain data set '{}' within "
+                       "group '{}'; available data sets are: {}")
+                raise IOError(msg.format(name, group, sub_names))
+            sub = root[name]
 
             def strip_ext(source):
                 """Strip extension if source identifies a file name"""
@@ -1139,7 +1150,7 @@ class SolutionArray:
                 return out
 
             # ensure that mechanisms are matching
-            sol_source = strip_ext(sub['solution'].attrs['source'])
+            sol_source = strip_ext(sub['phase'].attrs['source'])
             source = strip_ext(self.source)
             if sol_source != source and not force:
                 msg = ("Sources of thermodynamic phases do not match: '{}' vs "
@@ -1152,7 +1163,7 @@ class SolutionArray:
             # load data
             data = OrderedDict()
             for name, value in sub.items():
-                if name != 'solution':
+                if name != 'phase':
                     data[name] = np.array(value)
 
         self.restore_data(data)
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -668,7 +668,7 @@ class SolutionArray:
         `collect_data`.
 
         :param data: Dictionary holding data to be restored, where keys
-            refer to thermodynamic states (e.g. ``T``, ``P``) or extra
+            refer to thermodynamic states (e.g. ``T``, ``density``) or extra
             entries, and values contain corresponding data.
 
         The receiving `SolutionArray` either has to be empty or should have
@@ -694,7 +694,7 @@ class SolutionArray:
             if col.shape[0] != rows:
                 raise ValueError("'SolutionArray.restore_data' requires "
                                  "all data entries to have a consistent "
-                                 "first dimenstion")
+                                 "first dimension")
 
         if self._shape != (0,) and self._shape != (rows,):
             raise ValueError(
@@ -997,7 +997,7 @@ class SolutionArray:
             data_dict[label] = data[:, i]
         self.restore_data(data_dict)
 
-    def write_hdf(self, filename, *args, cols=None, group=None, name=None,
+    def write_hdf(self, filename, *args, cols=None, group=None, subgroup=None,
                   attrs={}, mode='a', append=False,
                   compression=None, compression_opts=None, **kwargs):
         """
@@ -1005,24 +1005,42 @@ class SolutionArray:
         Note that it is possible to write multiple data entries to a single HDF
         container file, where *group* is used to differentiate data.
 
+        An example for the default HDF file structure is:::
+
+            /                        Group
+            /group0                  Group
+            /group0/some_attr        Attribute
+            ...
+            /group0/T                Dataset
+            ...
+            /group0/phase            Group
+            /group0/phase/name       Attribute
+            /group0/phase/source     Attribute
+
+        where ``group0`` is the default name for the top level HDF entry. In
+        addition to datasets, information stored in `SolutionArray.meta` is
+        saved in form of HDF attributes. An additional intermediate layer may
+        be created using the *subgroup* argument.
+
         :param filename:
-            Name of the HDF container file; typical file extensions are `.hdf`,
-            `.hdf5` or `.h5`.
+            Name of the HDF container file; typical file extensions are
+            ``.hdf``, ``.hdf5`` or ``.h5``.
         :param cols:
             A list of any properties of the solution being exported.
         :param group:
-            Identifier for the group in the container file. A group may contain
-            multiple `SolutionArray` objects. If 'None', group names default
-            to 'groupN', with N being the number of pre-existing groups within
-            the HDF container file.
-        :param name:
-            Name identifier for a subgroup representing the `SolutionArray`
-            object to be added. If 'None', subgroup names default to 'arrM'
-            with M being the number of pre-existing subgroups within the group.
+            Identifier for the group in the container file. If no subgroup is
+            specified, a group represents a `SolutionArray`. If 'None', group
+            names default to 'groupN', with N being the number of pre-existing
+            groups within the HDF container file.
+        :param subgroup:
+            Name identifier for an optional subgroup, with subgroups
+            representing individual `SolutionArray` objects. If 'None', no
+            subgroup is created.
         :param attrs:
-            Dictionary of user-defined group attributes.
+            Dictionary of user-defined attributes added at the group level
+            (typically used in conjunction with a subgroup argument).
         :param mode:
-            Mode to open the file {'a' (default), 'w', 'r+'}.
+            Mode used by h5py to open the file {'a' (default), 'w', 'r+'}.
         :param append:
             If False, the content of a pre-existing group is deleted before
             writing the `SolutionArray` in the first position. If True, the
@@ -1034,7 +1052,7 @@ class SolutionArray:
             Options for the h5py compression filter; for 'gzip', this
             corresponds to the compression level {None, 0-9}.
         :return:
-            Identifiers for group and name used for storing HDF data.
+            Group identifier used for storing HDF data.
 
         Arguments *compression*, and *compression_opts* are mapped to parameters
         for `h5py.create_dataset`; in both cases, the choices of `None` results
@@ -1042,8 +1060,8 @@ class SolutionArray:
 
         Additional arguments (i.e. *args* and *kwargs*) are passed on to
         `collect_data`; see `collect_data` for further information. This method
-        works only with 1D `SolutionArray` objects and requires a working
-        installation of h5py (`h5py` can be installed using pip or conda).
+        requires a working installation of h5py (`h5py` can be installed using
+        pip or conda).
         """
         if isinstance(_h5py, ImportError):
             raise _h5py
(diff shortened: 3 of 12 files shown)


ischoegl:
@speth and @bryanwweber ... I believe this is a decent draft, which includes HDF output from `SolutionArray` and `FlameBase` (with settings saved as HDF attributes).

Some minor issues that remain to be fixed (will have to tend to some other issues before being able to get back to this):
 - [x] HDF export/import of `ReactingSurface1D`
 - [x] HDF export of `Y` / `X` fractions as 2-D array rather than column-wise (by species)

**Other thoughts:** Axes of `FlameBase.X` and `SolutionArray.X` are flipped ... while this is not consistent, I am not sure that it is worth addressing.

ischoegl:
@speth / @bryanwweber ... done.

Here's the full structure produced by `adiabatic_flame.py` (i.e. both `mix` and `multi` solutions as HDF):
```
In [1]: %run adiabatic_flame.py

[...]
multicomponent flamespeed = 0.727008 m/s
Solution saved to 'h2_adiabatic.csv'.

In [2]: !h5ls -r h2_adiabatic.h5
/                        Group
/mix                     Group
/mix/flame               Group
/mix/flame/T             Dataset {140}
/mix/flame/X             Dataset {140, 9}
/mix/flame/density       Dataset {140}
/mix/flame/grid          Dataset {140}
/mix/flame/phase         Group
/mix/flame/velocity      Dataset {140}
/mix/products            Group
/mix/products/T          Dataset {0}
/mix/products/X          Dataset {0, 9}
/mix/products/density    Dataset {0}
/mix/products/phase      Group
/mix/reactants           Group
/mix/reactants/T         Dataset {1}
/mix/reactants/X         Dataset {1, 9}
/mix/reactants/density   Dataset {1}
/mix/reactants/phase     Group
/mix/reactants/velocity  Dataset {1}
/multi                   Group
/multi/flame             Group
/multi/flame/T           Dataset {140}
/multi/flame/X           Dataset {140, 9}
/multi/flame/density     Dataset {140}
/multi/flame/grid        Dataset {140}
/multi/flame/phase       Group
/multi/flame/velocity    Dataset {140}
/multi/products          Group
/multi/products/T        Dataset {0}
/multi/products/X        Dataset {0, 9}
/multi/products/density  Dataset {0}
/multi/products/phase    Group
/multi/reactants         Group
/multi/reactants/T       Dataset {1}
/multi/reactants/X       Dataset {1, 9}
/multi/reactants/density Dataset {1}
/multi/reactants/phase   Group
/multi/reactants/velocity Dataset {1}

In [3]: !h5dump -n 1 h2_adiabatic.h5
HDF5 "h2_adiabatic.h5" {
FILE_CONTENTS {
 group      /
 group      /mix
 attribute  /mix/Sim1D_type
 attribute  /mix/curve
 attribute  /mix/date
 attribute  /mix/description
 attribute  /mix/energy_enabled
 attribute  /mix/fixed_temperature
 attribute  /mix/max_grid_points
 attribute  /mix/max_time_step_count
 attribute  /mix/prune
 attribute  /mix/radiation_enabled
 attribute  /mix/ratio
 attribute  /mix/slope
 attribute  /mix/soret_enabled
 attribute  /mix/transport_model
 group      /mix/flame
 attribute  /mix/flame/Domain1D_type
 attribute  /mix/flame/name
 attribute  /mix/flame/emissivity_left
 attribute  /mix/flame/emissivity_right
 attribute  /mix/flame/steady_abstol
 attribute  /mix/flame/steady_reltol
 attribute  /mix/flame/transient_abstol
 attribute  /mix/flame/transient_reltol
 dataset    /mix/flame/T
 dataset    /mix/flame/X
 dataset    /mix/flame/density
 dataset    /mix/flame/grid
 group      /mix/flame/phase
 attribute  /mix/flame/phase/name
 attribute  /mix/flame/phase/source
 dataset    /mix/flame/velocity
 group      /mix/products
 attribute  /mix/products/Domain1D_type
 attribute  /mix/products/name
 dataset    /mix/products/T
 dataset    /mix/products/X
 dataset    /mix/products/density
 group      /mix/products/phase
 attribute  /mix/products/phase/name
 attribute  /mix/products/phase/source
 group      /mix/reactants
 attribute  /mix/reactants/Domain1D_type
 attribute  /mix/reactants/name
 dataset    /mix/reactants/T
 dataset    /mix/reactants/X
 dataset    /mix/reactants/density
 group      /mix/reactants/phase
 attribute  /mix/reactants/phase/name
 attribute  /mix/reactants/phase/source
 dataset    /mix/reactants/velocity
 group      /multi
 attribute  /multi/Sim1D_type
 attribute  /multi/curve
 attribute  /multi/date
 attribute  /multi/description
 attribute  /multi/energy_enabled
 attribute  /multi/fixed_temperature
 attribute  /multi/max_grid_points
 attribute  /multi/max_time_step_count
 attribute  /multi/prune
 attribute  /multi/radiation_enabled
 attribute  /multi/ratio
 attribute  /multi/slope
 attribute  /multi/soret_enabled
 attribute  /multi/transport_model
 group      /multi/flame
 attribute  /multi/flame/Domain1D_type
 attribute  /multi/flame/name
 attribute  /multi/flame/emissivity_left
 attribute  /multi/flame/emissivity_right
 attribute  /multi/flame/steady_abstol
 attribute  /multi/flame/steady_reltol
 attribute  /multi/flame/transient_abstol
 attribute  /multi/flame/transient_reltol
 dataset    /multi/flame/T
 dataset    /multi/flame/X
 dataset    /multi/flame/density
 dataset    /multi/flame/grid
 group      /multi/flame/phase
 attribute  /multi/flame/phase/name
 attribute  /multi/flame/phase/source
 dataset    /multi/flame/velocity
 group      /multi/products
 attribute  /multi/products/Domain1D_type
 attribute  /multi/products/name
 dataset    /multi/products/T
 dataset    /multi/products/X
 dataset    /multi/products/density
 group      /multi/products/phase
 attribute  /multi/products/phase/name
 attribute  /multi/products/phase/source
 group      /multi/reactants
 attribute  /multi/reactants/Domain1D_type
 attribute  /multi/reactants/name
 dataset    /multi/reactants/T
 dataset    /multi/reactants/X
 dataset    /multi/reactants/density
 group      /multi/reactants/phase
 attribute  /multi/reactants/phase/name
 attribute  /multi/reactants/phase/source
 dataset    /multi/reactants/velocity
 }
}
```

ischoegl:
After some small edits I think I’m happy with the structure, so I think this is ready for review ...

include/cantera/base/Solution.h
```diff
@@ -37,6 +37,12 @@ class Solution : public std::enable_shared_from_this<Solution>
     //! Set the name of this Solution object
     void setName(const std::string& name);
 
+    //! Return the source of this Solution object
+    std::string source() const { return m_source; }
+
+    //! Set the source of this Solution object
+    void setSource(const std::string& source) { m_source = source; }
```
speth: If the "source" is only going to be set when a `Solution` is constructed in Python, it would probably be simpler to just make it a property of the Python `Solution` class rather than storing it at the C++ level.

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -2,13 +2,32 @@
...
+    import h5py as _h5py
+
+# avoid explicit dependence of cantera on pandas
+try:
+    pkg_resources.get_distribution('pandas')
+except pkg_resources.DistributionNotFound:
+    _pandas = ImportError('Method requires a working pandas installation.')
+else:
+    import pandas as _pandas
+
```
speth, line 13: You could avoid repeating this code from `composite.py` by just importing `cantera.composite` and using the values of `_h5py` and `_pandas` from there.

interfaces/cython/cantera/composite.py
```diff
@@ -488,7 +500,7 @@ class SolutionArray:
 
     _purefluid_scalar = ['Q']
 
-    def __init__(self, phase, shape=(0,), states=None, extra=None):
+    def __init__(self, phase, shape=(0,), states=None, extra=None, meta={}):
```
speth: I think the default value for `meta` needs to be `None`, and a new `dict` should be created if that is the case. Otherwise, all `SolutionArray` objects where no initial value for `meta` is supplied will share the same `dict`.

interfaces/cython/cantera/composite.py
```diff
@@ -613,6 +626,13 @@ def append(self, state=None, **kwargs):
         self._indices.append(len(self._indices))
         self._shape = (len(self._indices),)
 
+    @property
+    def meta(self):
+        """
+        Metadata
+        """
+        return self._meta
```
speth, line 638: Does this need to be a property? I guess that makes it easier to provide Sphinx documentation, but the documentation here is a bit terse.

interfaces/cython/cantera/composite.py
```diff
@@ -637,14 +657,13 @@ def equilibrate(self, *args, **kwargs):
...
 
-    def restore_data(self, data, labels):
+    def restore_data(self, data):
         """
-        Restores a `SolutionArray` based on *data* specified in a single
-        2D Numpy array and a list of corresponding column *labels*. Thus,
-        this method allows to restore data exported by `collect_data`.
+        Restores a `SolutionArray` based on *data* specified in an ordered
+        dictionary. Thus, this method allows to restore data exported by
+        `collect_data`.
```
speth, line 668: It would help to describe the structure of this dictionary.

interfaces/cython/cantera/composite.py
```diff
@@ -788,17 +843,19 @@ def collect_data(self, cols=None, threshold=0, species='Y'):
             include any arrays which were specified as 'extra' variables when
             defining the `SolutionArray` object. The special value 'extra' can
             be used to include all 'extra' variables.
-        :param threshold: Relative tolerance for including a particular column.
-            The tolerance is applied by comparing the maximum absolute value for
-            a particular column to the maximum absolute value in all columns for
-            the same variable (e.g. mass fraction).
+        :param tabular: Split 2D data into 1D columns
```
```suggestion
        :param tabular: Split 2D data into separate 1D columns for each species / reaction
```

interfaces/cython/cantera/composite.py
```diff
@@ -910,64 +979,196 @@ def from_pandas(self, df):
...
+            to 'groupN', with N being the number of pre-existing groups within
+            the HDF container file.
+        :param name:
+            Name identifier for a subgroup representing the `SolutionArray`
+            object to be added. If 'None', subgroup names default to 'arrM'
+            with M being the number of pre-existing subgroups within the group.
+        :param attrs:
+            Dictionary of user-defined group attributes.
+        :param mode:
+            Mode to open the file {'a' (default), 'w', 'r+}.
```
```suggestion
            Mode to open the file {'a' (default), 'w', 'r+'}.
```
Also, we should probably explain what these modes mean, or at least refer to the documentation of `h5py.File`

interfaces/cython/cantera/composite.py
```diff
@@ -910,64 +979,196 @@ def from_pandas(self, df):
...
+            A list of any properties of the solution being exported.
+        :param group:
+            Identifier for the group in the container file. A group may contain
+            multiple `SolutionArray` objects. If 'None', group names default
+            to 'groupN', with N being the number of pre-existing groups within
+            the HDF container file.
+        :param name:
+            Name identifier for a subgroup representing the `SolutionArray`
+            object to be added. If 'None', subgroup names default to 'arrM'
+            with M being the number of pre-existing subgroups within the group.
```
speth: Do we need to always have two levels of identifiers here? I'm glad to see that you can create whatever structure you like within the HDF5 file by specifying `group` as a pathlike-name (e.g. `foo/bar/baz`) which will currently create the group `foo/bar/baz/arr0`, but it seems like this would simpler and more flexible if we just dropped the `name` argument.

interfaces/cython/cantera/composite.py
```diff
@@ -910,64 +979,196 @@ def from_pandas(self, df):
...
+
+        hdf_kwargs = {'compression': compression,
+                      'compression_opts': compression_opts}
+        hdf_kwargs = {k: v for k, v in hdf_kwargs.items() if v is not None}
+
+        # save to container file
+        with _h5py.File(filename, mode) as hdf:
+
+            # check existence of tagged item
+            msg = "HDF group with group identifier '{}' exists in '{}': {}"
```
speth: Is this message used anywhere?

interfaces/cython/cantera/onedim.py
```diff
@@ -545,13 +496,14 @@ def from_pandas(self, df, restore_boundaries=True, settings=None):
...
+        arr = SolutionArray(self.gas, extra=self.extra())
         arr.from_pandas(df)
         self.from_solution_array(arr, restore_boundaries=restore_boundaries,
                                  settings=settings)
 
-    def write_hdf(self, filename, key=None, species='X',
-                  mode=None, complevel=None, quiet=True):
+    def write_hdf(self, filename, group=None, species='X', mode='a',
+                  description=None,
+                  compression=None, compression_opts=None, quiet=True):
```
speth: Can this be modified to pass `**kwargs` to `SolutionArray.collect_data`, the same way that `SolutionArray.write_hdf` does? That would allow using the `tabular` and `threshold` arguments, which would be nice additions. I tried modifying it, and it almost works except for some gotchas with "empty" domains like the free flame outlet domain.

interfaces/cython/cantera/composite.py
```diff
@@ -910,64 +979,196 @@ def from_pandas(self, df):
...
+                sub.attrs[key] = val
+            for header, col in data.items():
+                sub.create_dataset(header, data=col, **hdf_kwargs)
+
+            # add subgroup containing information on gas
+            sol = sub.create_group('phase')
+            sol.attrs['name'] = self.name
+            sol.attrs['source'] = self.source
+
+        return group, name
```
speth: This return value should to be documented.

interfaces/cython/cantera/composite.py
```diff
@@ -910,64 +979,196 @@ def from_pandas(self, df):
...
+
+            # load data
+            data = OrderedDict()
+            for name, value in sub.items():
+                if name != 'phase':
+                    data[name] = np.array(value)
+
+        self.restore_data(data)
+
+        return root_attrs
```
speth, line 1210: This return value needs to be documented.

interfaces/cython/cantera/onedim.py
```diff
@@ -34,6 +28,41 @@ def __init__(self, domains, gas, grid=None):
         self.gas = gas
         self.flame.P = gas.P
 
+    def extra(self, domain=None):
+        """
+        Extra columns used for saving/restoring of simulation data that are
```
speth: Can you document the meaning of the `domain` argument (in particular, what the default of `None` means) for this and the other new methods that take this argument?

interfaces/cython/cantera/onedim.py
```diff
@@ -34,6 +28,41 @@ def __init__(self, domains, gas, grid=None):
...
+        specific to a class derived from `FlameBase`. Entries may include:
+         * ``grid``: grid point positions along the flame [m]
+         * ``velocity``: normal velocity [m/s]
+         * ``spread_rate``: tangential velocity gradient [1/s]
+         * ``lambda``: radial pressure gradient [N/m^4]
+         * ``eField``: electric field strength
```
speth, line 40: While these variables are "extra" for the `SolutionArray`, beyond what is needed to specify the thermodynamic state, they're not really "extra" here. Is there a more descriptive name we could use? Admittedly, I'm not sure what to call it. With the exception of the grid, they're all state variables for the system, but that's a big exception.

interfaces/cython/cantera/onedim.py
```diff
@@ -34,6 +28,41 @@ def __init__(self, domains, gas, grid=None):
...
+                          if e not in {'grid', 'lambda', 'eField'}])
+        elif isinstance(dom, IdealGasFlow):
+            return self._extra
+        else:
+            return ()
+
+    def phase(self, domain=None):
+        """
+        Return phase describing the domain (i.e. gas phase or surface phase).
+        """
```
speth: Would it make more sense for this to be a method on class `Domain1D`, where the default is to return `dom.gas` and it is overridden in `ReactingSurf1D` to return `dom.surface`?

interfaces/cython/cantera/onedim.py
```diff
@@ -389,128 +423,45 @@ def write_csv(self, filename, species='X', quiet=True):
...
-        if self.radiation_enabled:
-            qdot = self.flame.radiative_heat_loss
-            extra['qdot'] = np.hstack([np.nan, qdot, np.nan])
-
-        # consider inlet boundaries
-        left = self.domains[0]  # left boundary is always an inlet
-        if isinstance(self.domains[2], Inlet1D):
-            right = self.domains[2]
-            n_arr = self.flame.n_points + 2
+        Derived classes define default values for *extra*.
```
speth: Would the simpler way to handle turning any domain into a `SolutionArray` be to implement `to_solution_array` on `Domain1D`, and just have a thin convenience wrapper on `FlameBase` that calls `self.flame.to_solution_array`? This would mirror what we do with many of the other methods of the `FlameBase` class, but it may not be worth the effort right now.

interfaces/cython/cantera/onedim.py
```diff
@@ -560,132 +512,90 @@ def write_hdf(self, filename, key=None, species='X',
...
+            Identifier for the group in the container file. A group may contain
+            multiple `SolutionArray` objects.
         :param species:
             Attribute to use obtaining species profiles, e.g. ``X`` for
             mole fractions or ``Y`` for mass fractions.
         :param mode:
-            Mode to open file (see `pandas.DataFrame.to_hdf`)
-        :param complevel:
-            Compression level (see `pandas.DataFrame.to_hdf`)
+            Mode to open the file {'a' (default), 'w', 'r+}.
```
```suggestion
            Mode to open the file {'a' (default), 'w', 'r+'}.
```

interfaces/cython/cantera/onedim.py
speth: The `quiet` option isn't documented (is anyone likely to use `quiet=False`?)

Also, I think the structure of the resulting file needs to be described at least in general terms here or somewhere else suitable. This is especially true for the "attributes" of the datasets, which can be easy to overlook, at least when exploring the dataset using `h5py`.

interfaces/cython/cantera/onedim.pyx
```diff
@@ -239,6 +250,18 @@ cdef class Domain1D:
     def __copy__(self):
         raise NotImplementedError('Domain1D object is not copyable')
 
+    property settings:
+        """
+        Metadata
```
speth: This docstring could be a bit more descriptive...

interfaces/cython/cantera/onedim.pyx
```diff
@@ -376,6 +399,11 @@ cdef class ReactingSurface1D(Boundary1D):
         self.surf = new CxxReactingSurf1D()
         self.boundary = <CxxBdry1D*>(self.surf)
 
+    def __init__(self, *args, **kwargs):
+        self._weakref_proxy2 = _WeakrefProxy()
```
speth: I think you can just make multiple weakrefs to the original `_weakref_proxy` object.

interfaces/cython/cantera/onedim.pyx
```diff
@@ -385,12 +413,18 @@ cdef class ReactingSurface1D(Boundary1D):
...
 
     property coverage_enabled:
         """Controls whether or not to solve the surface coverage equations."""
         def __set__(self, value):
             self.surf.enableCoverageEquations(<cbool>value)
 
+    property surface:
+        def __get__(self):
+            return self._surface
+
```
speth: If this is just going to be a simple passthrough, you can just declare it as `cdef public Kinetics surface` in `_cantera.pxd`.

interfaces/cython/cantera/onedim.pyx
```diff
@@ -466,6 +500,62 @@ cdef class _FlowBase(Domain1D):
     def __dealloc__(self):
         del self.flow
 
+    property settings:
+        """
+        Metadata
```
speth: Please add some documentation here.

interfaces/cython/cantera/onedim.pyx
```diff
@@ -837,6 +928,121 @@ cdef class Sim1D:
...
+            states = T, P, np.vstack(Y).T
+            return states, extra_cols, meta
+
+        elif isinstance(dom, Inlet1D):
+            self.gas.TPY = dom.T, self.P, dom.Y
+
+            # create extra columns
+            for e in extra:
+               if e == 'velocity':
+                   extra_cols[e] = dom.mdot/self.gas.density
```
speth:
```suggestion
                   extra_cols[e] = dom.mdot / self.gas.density
```

interfaces/cython/cantera/onedim.pyx
```diff
@@ -837,6 +928,121 @@ cdef class Sim1D:
...
+
+        if isinstance(dom, _FlowBase):
+            n_points = dom.n_points
+
+            # retrieve gas state
+            T = self.profile(self.flame, 'T')
+            P = dom.P
+            k0 = self.flame.component_index(self.gas.species_name(0))
+            Y = [self.profile(self.flame, k)
+                 for k in range(k0, k0 + self.gas.n_species)]
```
speth: Can't this be replaced with
```suggestion
            T = self.T
            P = self.P
            Y = self.Y
```
Or just write `states = self.T, self.P, self.Y.T` below?

interfaces/cython/cantera/onedim.pyx
```diff
@@ -837,6 +928,121 @@ cdef class Sim1D:
...
+        elif isinstance(dom, Boundary1D):
+            return ([], [], []), extra_cols, meta
+
+        else:
+            msg = ("Export of '{}' is not implemented")
+            raise NotImplementedError(msg.format(type(self).__name__))
+
+    def restore_data(self, domain, states, extra_cols, meta):
+        """
+        Restore data vector of domain *domain* from `SolutionArray` *arr*.
```
speth: What is `arr`? Is this supposed to be `states`?

interfaces/cython/cantera/onedim.pyx
```diff
@@ -837,6 +928,121 @@ cdef class Sim1D:
...
+            dom.mdot = extra_cols['velocity'] * self.gas.density
+
+        elif isinstance(dom, ReactingSurface1D):
+            dom.surface.TPY = T, P, Y
+
+        elif isinstance(dom, Boundary1D):
+            pass
+
+        else:
+            msg = ("Export of '{}' is not implemented")
```
speth:
```suggestion
            msg = ("Import of '{}' is not implemented")
```

interfaces/cython/cantera/onedim.pyx
```diff
@@ -837,6 +928,121 @@ cdef class Sim1D:
...
+            dom.T = T
+            self.P = P
+            dom.Y = Y
+            dom.mdot = extra_cols['velocity'] * self.gas.density
+
+        elif isinstance(dom, ReactingSurface1D):
+            dom.surface.TPY = T, P, Y
+
+        elif isinstance(dom, Boundary1D):
+            pass
```
speth: It might be better to be explicit here about the `Boundary1D` classes that do work without any additional state. Otherwise, this would inadvertently let through any new `Boundary1D` derived classes that should require some state variables to be set.

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -119,6 +138,90 @@ def check(a, b):
...
+        b = ct.SolutionArray(self.gas)
+        b.read_csv(outfile)
+        self.assertTrue(np.allclose(states.T, b.T))
+        self.assertTrue(np.allclose(states.P, b.P))
+        self.assertTrue(np.allclose(states.X, b.X))
+
+    def test_to_pandas(self):
+        if isinstance(_pandas, ImportError):
+            return
```
speth: Rather than giving the impression that this test passes when `pandas` is not installed, I would suggest conditionally skipping the test instead, using the `skipIf` decorator. I think it should be something like:
```suggestion
    @unittest.skipIf(isinstance(_pandas, ImportError), "pandas is not installed")
    def test_to_pandas(self):
```
And likewise for the tests requiring h5py.

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -148,84 +251,86 @@ def check(a, b, atol=None):
...
         b = ct.SolutionArray(self.gas)
-        b.restore_data(data, labels)
+        b.restore_data(data)
         check(a, b)
 
         # skip concentrations
         b = ct.SolutionArray(self.gas)
-        b.restore_data(data[:, :2], labels[:2])
+        b.restore_data(OrderedDict([tup for i, tup in enumerate(data.items())
+                                    if i < 2]))
```
speth: Wouldn't it be a little more clear to just write:
```suggestion
        b.restore_data({'T': data['T'], 'density': data['density']})
```

interfaces/cython/cantera/test/test_onedim.py
```diff
@@ -1069,18 +1134,45 @@ def run_reacting_surface(self, xch4, tsurf, mdot, width):
...
-        self.assertTrue(all(np.diff(sim.Y[gas.species_index('CH4')]) < 0))
-        self.assertTrue(all(np.diff(sim.Y[gas.species_index('CO2')]) > 0))
+        self.assertTrue(all(np.diff(sim.Y[sim.gas.species_index('CH4')]) < 0))
+        self.assertTrue(all(np.diff(sim.Y[sim.gas.species_index('CO2')]) > 0))
+        self.sim = sim
 
     def test_reacting_surface_case1(self):
         self.run_reacting_surface(xch4=0.095, tsurf=900.0, mdot=0.06, width=0.1)
 
+
```
speth: extraneous blank line.

interfaces/cython/cantera/test/test_onedim.py
```diff
@@ -604,10 +638,38 @@ def test_write_csv(self):
...
         data.read_csv(filename)
-        self.assertArrayNear(data.grid[1:], self.sim.grid)
-        self.assertArrayNear(data.T[1:], self.sim.T)
+        self.assertArrayNear(data.grid, self.sim.grid)
+        self.assertArrayNear(data.T, self.sim.T)
         k = self.gas.species_index('H2')
-        self.assertArrayNear(data.X[1:,k], self.sim.X[k,:])
+        self.assertArrayNear(data.X[:, k], self.sim.X[k, :])
+
+    def test_write_hdf(self):
```
speth, line 639: These tests need to be conditioned on `h5py` being available, don't they?

interfaces/cython/cantera/test/test_onedim.py
```diff
@@ -1043,22 +1105,25 @@ def test_restart(self):
...
         # to generate a good starting estimate for the coverages.
         surf_phase.advance_coverages(1.)
 
-        sim = ct.ImpingingJet(gas=gas, width=width, surface=surf_phase)
+        return ct.ImpingingJet(gas=gas, width=width, surface=surf_phase)
+
+    def run_reacting_surface(self, xch4, tsurf, mdot, width):
+        # Simplified version of the example 'catalytic_combustion.py'
+        tinlet = 300.0  # inlet temperature
+        comp = {'CH4': xch4, 'O2':0.21, 'N2':0.79}
```
speth:
```suggestion
        comp = {'CH4': xch4, 'O2': 0.21, 'N2': 0.79}
```

speth (changes requested):
Thanks for putting this together, @ischoegl, and I'm sorry it took me so long to provide a review. I think the structure of the HDF5 output here looks very good, and should hopefully be easy for people to work with.

ischoegl:
@speth ... thanks for the detailed review! Interesting observation/suggestion to use entire paths rather than my baked-in two-layered approach. Should be able to turn this around next week, as I hope to see this in beta ... which will finally give me the ability to settle on a HDF format for derivative code I’ve been working on.

interfaces/cython/cantera/test/test_composite.py
ischoegl: As far as I remember, this is mainly because of pesky Xenial and Python 3.5, which do not preserve order of dictionary items. After that, regular dictionaries are sufficient. I suppose Python 3.5 will be supported for some more time ...

I no longer have a xenial toolchain, so I'd prefer to leave it as is until cantera requires Python 3.6 as a minimum version.

interfaces/cython/cantera/onedim.pyx
ischoegl: deleted the docstring here as it overloads `Domain1D.settings` (which is updated).

include/cantera/base/Solution.h
ischoegl: It may be simpler, but in many instances PR's eventually end up creating entries in the C++ layer. What I did for Python can easily be expanded to save source information for objects instantiated in C++ directly.

interfaces/cython/cantera/composite.py
ischoegl: I'd like to keep it that way as `meta` should be set when the `SolutionArray` is generated, and shouldn't necessarily be overwritten at a later point. Docstring is updated.

interfaces/cython/cantera/onedim.py
ischoegl: I interpreted `extra` as being non-thermodynamic information in the strict sense. While what is referred to here as `extra` is part of the *state* of a 1D object, it's not thermodynamic, so I'd prefer to keep the nomenclature.

interfaces/cython/cantera/onedim.py
ischoegl: Well, it's actually used to access domains within the `Sim1D.domains` list. It can be relocated to `Sim1D` (which I did); also, `phase` can be defined within `Domain1D` and `ReactingSurf1D` as you suggested (which I also implemented).

interfaces/cython/cantera/onedim.pyx
ischoegl: Yes, you are correct that this is overly complicated now - this is probably a left-over from the previous HDF version where I had to build arrays from scratch. Only thing is that `self.Y` needs to be transposed.

interfaces/cython/cantera/onedim.py
ischoegl: Unfortunately, this is a no ... I would have preferred it this way also, but I don't think that this is possible at the moment. `Domain1D`'s do not know how the solution vector is stored/accessed, so this needs to go through `Sim1D` (the critical step is `Sim1D.collect_data`).

PS: this is also the reason for introducing `phase` and `extra` methods on `Sim1D`, rather than limiting them to `Domain1D`.

interfaces/cython/cantera/onedim.py
ischoegl: Good suggestion, and the gotcha was straightforward to resolve.

ischoegl: Added a description to the docstring.

interfaces/cython/cantera/composite.py
ischoegl: OK, this was a legitimate concern. I renamed the second level to `subgroup` and made it optional. The two levels are needed for `Sim1D` output, but it can be flattened for conventional `SolutionArray` output (i.e. the `arrX` is now gone). It's indeed neat that HDF allows for a direct creation of deeper levels - I added a unit test for your `foo/bar/baz` case ...

ischoegl:
@speth ... I believe this is mostly done - I adopted most of your suggestions. Apart from minor tweaks here and there I made the second HDF output level optional.

The main thing i was not able to address is to delegate `to_solutionarray` to `Domain1D`, as some critical data handling access points only exist for `Sim1D` (and an alternative implementation would go way beyond the scope of this PR). 

I’ll probably read over the docstrings one more time, but the code portion runs.

interfaces/cython/cantera/onedim.py
ischoegl: Changed my mind here: proposing `other` instead of `extra` ... changes are applied in the last commit.

ischoegl:
Alright - had a last look over the docstrings and addressed one final comment (suggesting to replace `extra` by `other`). I believe this is done!

interfaces/cython/cantera/onedim.py
```diff
@@ -34,6 +28,32 @@ def __init__(self, domains, gas, grid=None):
         self.gas = gas
         self.flame.P = gas.P
 
+    def other(self, domain=None):
+        """
+        The method returns simulation componentes that are specific to a class
```
```suggestion
        The method returns simulation components that are specific to a class
```

interfaces/cython/cantera/onedim.py
speth: I think the confusion, at least for me, comes from the incompleteness of the name. How about `other_components`?

interfaces/cython/cantera/test/test_composite.py
speth: Sorry, I wasn't concerned about the use of `OrderedDict` here, which I don't think has any impact here regardless of Python version. I was looking at this as a somewhat roundabout way of getting a dict with these two items from `data`.

In general, even with the change in `dict` behavior from Python 3.6 on, I've seen the argument that using `OrderedDict` is still a good practice when you want to emphasize the fact that the order is important.

interfaces/cython/cantera/onedim.py
```diff
@@ -545,147 +487,142 @@ def from_pandas(self, df, restore_boundaries=True, settings=None):
...
+            /group0/products/phase           Group
+            /group0/reactants                Group
+            /group0/reactants/Domain1D_type  Attribute
+            ...
+            /group0/reactants/T              Dataset
+            ...
+            /group0/reactants/phase          Group
+
+        where ``group0`` is the default name for the top level HDF entry, and
+        ``reactants``, ``flame`` and ``reactants`` correspond to domain names.
```
```suggestion
        ``reactants``, ``flame`` and ``products`` correspond to domain names.
```

interfaces/cython/cantera/onedim.py
```diff
@@ -545,147 +487,142 @@ def from_pandas(self, df, restore_boundaries=True, settings=None):
...
-        Simulation settings are stored in tabular form as a separate HDF group
-        named ``settings``.
+    def write_hdf(self, filename, *args, group=None, species='X', mode='a',
+                  description=None, compression=None, compression_opts=None,
+                  quiet=True, **kwargs):
+        """
+        Write the solution vector to a HDF container file.
+
+        The `write_hdf` method preserves the stucture of a `FlameBase`-derived
+        object (i.e. `FreeFlame`, etc.). Each simulation is saved as a *group*,
```
speth: A minor style suggestion that @bryanwweber has gotten me started on is to avoid the use of these Latin abbreviations as a way of making things more accessible to non-fluent English speakers.
```suggestion
        object (such as `FreeFlame`). Each simulation is saved as a *group*,
```

include/cantera/base/Solution.h
speth: I think it introduces confusion for this value to only be set correctly for the case of a `Solution` instantiated in Python, though. Rather than suggesting that this needs to be populated in those other instances though, I thought having this available only in Python was the simpler fix.

The longer-term option for making this information accessible is that the filename is in the metadata of the `AnyMap` available as `ThermoPhase.input()` for objects created from YAML files. So, once the CTI and XML formats are dead, we can provide this information that way.

interfaces/cython/cantera/composite.py
```diff
@@ -910,64 +992,220 @@ def from_pandas(self, df):
...
+            groups within the HDF container file.
+        :param subgroup:
+            Name identifier for an optional subgroup, with subgroups
+            representing individual `SolutionArray` objects. If 'None', no
+            subgroup is created.
+        :param attrs:
+            Dictionary of user-defined attributes added at the group level
+            (typically used in conjunction with a subgroup argument).
+        :param mode:
+            Mode used by h5py to open the file {'a' (default), 'w', 'r+'}.
```
speth: Can you document what the mode options mean, or reference the relevant `h5py` documentation?

speth (changes requested):
Thanks for the changes, I think this is pretty close. Just a few (hopefully) final comments from me.

interfaces/cython/cantera/composite.py
ischoegl: ok. The best link isn't part of the API documentation, so I'd rather not use a web link (also not sure where you set up InterSphinx). 

I.e. I added documention in plain text.

interfaces/cython/cantera/onedim.py
ischoegl: `other_components` it is ...

interfaces/cython/cantera/test/test_composite.py
ischoegl: Alright, I see the point. I guess using a regular dictionary here shouldn't matter anyways, as order won't matter in this case.

include/cantera/base/Solution.h
ischoegl: I think I'm not absolutely certain here: even if it's part of `AnyMap`/YAML, the point here is that neither is saved to HDF - rather, the point of saving `source` in HDF metadata is that things are properly re-created.

Regardless, it is correct that it's only used from the Python interface at the moment, so I'll remove it from the C++ base.

ischoegl:
@speth ... I think everything is resolved now.

interfaces/cython/cantera/base.pyx
```diff
@@ -108,8 +117,10 @@ cdef class _SolutionBase:
         cdef CxxAnyMap root
         if infile:
             root = AnyMapFromYamlFile(stringify(infile))
+            self._source = infile
         elif source:
             root = AnyMapFromYamlString(stringify(source))
+            self_source = 'custom YAML'
```
speth: This is supposed to be
```suggestion
            self._source = 'custom YAML'
```
right?

speth (changes requested):
One last typo and then we'll be done.

interfaces/cython/cantera/base.pyx
ischoegl: absolutely! 

out of curiosity ... there appear to be some linters that work on git diffs only - in case you're using those, do you have any recommendations?

speth: No, no recommendations there, but it sounds like a nice feature. I use a combination of Visual Studio Code, the GitHub web interface, and `git diff` when reviewing PRs. I noticed this mainly because the syntax highlighting of `self` compared to a couple of lines above caught my attention.

speth (approved):

interfaces/cython/cantera/base.pyx
ischoegl: Thanks, I thought I’d ask - visual studio code is indeed nice (started using it recently, despite having been on emacs for what feels like eternities).

Anyhow, thanks for the review!

PR #843, opened 2020-04-09
Fix error in exchange current density conversion

### LINKED ISSUE #841: IdealGasPhase getStandardChemPotentials
**System information**

- Cantera version: 2.4
- OS: all (C++ source code)

**Expected behavior**
The standard state definition for an ideal gas, typically, is pressure-independent.  All pressure dependence in thermodynamic definitions is generally handled in the activity term, where the activity of a species k is equal to `pressure*molefraction[k]/standardPressure`.

More directly, this should always result in reverse reaction rate coefficients, calculated via k_rev = k_fwd*exp(-deltaG^o/RT)*prod(standardConcentrations^nu_k` being pressure independent (so long as f_fwd is pressure-independent).

**Actual behavior (EDITED FROM ORIGINAL POST)**

- For the `IdealGasPhase`, the standard state is defined as pressure-dependent (cf `IdealGasPhase::getStandardChemPotentials`, which includes `muStar + log(pressure() / refPressure()) * RT();`)
- For reactions where the forward rate `k_fwd` is supplied and the thermo is used to calculate `k_ref` (i.e. standard kinetics), this causes no problems. When it comes to calculating the reverse rate, the `GasKinetics::updateKc` routine includes the term `m_rkcn[irxn]*rrt - m_dn[irxn]*m_logStandConc)`.  In short, the logStandardConc term compensates for any pressure dependence in delta G^0. 
- However, for interface reactions where `rate_coeff_type = exchangecurrentdensity` is used to supply the composition- and pressure-independent part of the exchange current density (let's call it i_o^*), which Cantera converts into a `k_fwd` and `k_rev`, we end up with reverse rate coefficients which are pressure dependent.  @wbessler first noticed this when comparing charge-transfer reactions using  forward and reverse rate constants (k_fwd and k_rev), versus using `rate_coeff_type = exchangecurrentdensity`.  The former gives the expected behavior, whereas the latter does not.

**Additional context**
See discussions below.

decaluwe:
-Fixes issue #841, wherein
`InterfaceKinetics`::`convertExchangeCurrentDensityFormulation` incorrectly
returned pressure-dependent rate constants, due to a mising `m_beta`
exponent in the standard concentration product.
-Also adds a test to the python test suite to ensure that a specified
rate constant remains appropriately pressure-independent.
-Caught an error in the conversion from rate constant back to exchange
current, wherein dG^0 was multiplied rather than divided by RT.

**Checklist**

Fixes #841

- Raise `m_ProdStanConcReac` to the power `m_beta` in `InterfaceKinetics`::`convertExchangeCurrentDensityFormulation` when converting between exchange current density and k_fwd.
-Adds a test to test the pressure dependende of f_fwd as converted from exchange current density.
-Corrects a separate typo in `InterfaceKinetics`::`convertExchangeCurrentDensityFormulation`.

### DIFF
--- a/interfaces/cython/cantera/test/test_kinetics.py
+++ b/interfaces/cython/cantera/test/test_kinetics.py
@@ -1215,3 +1215,29 @@ class TestReaction(utilities.CanteraTest):
 
         # M&W toggled on (locally) for reaction 9
         self.assertNear(2.0 * k1[9], k2[9]) # sticking coefficient = 1.0
+
+class TestButlerVolmer(utilities.CanteraTest):
+
+    def test_pressure_dependence(self):
+        infile = 'butler-volmer-test.yaml'
+        
+        gas = ct.Solution(infile,'gas')
+        Pt = ct.Solution(infile,'metal')
+        Naf = ct.Solution(infile,'electrolyte')
+        tpb = ct.Interface(infile,'tpb',[gas,Pt,Naf])
+        phases = {gas,Pt,Naf,tpb}
+
+        TP1 = 300, ct.one_atm
+        for ph in phases:
+            ph.TP = TP1
+
+        k1 = tpb.forward_rate_constants[0]
+
+        TP2 = 300., 10*ct.one_atm
+        for ph in phases:
+            ph.TP = TP2
+        
+        k2 = tpb.forward_rate_constants[0]
+
+        self.assertNear(k1,k2)
+
--- a/src/kinetics/InterfaceKinetics.cpp
+++ b/src/kinetics/InterfaceKinetics.cpp
@@ -269,7 +269,7 @@ void InterfaceKinetics::convertExchangeCurrentDensityFormulation(doublereal* con
                 //  Calculate the term and modify the forward reaction
                 double tmp = exp(- m_beta[i] * m_deltaG0[irxn]
                                  / thermo(reactionPhaseIndex()).RT());
-                tmp *= 1.0 / m_ProdStanConcReac[irxn] / Faraday;
+                tmp *= 1.0 / pow(m_ProdStanConcReac[irxn],m_beta[i]) / Faraday;
                 kfwd[irxn] *= tmp;
             }
             //  If BVform is nonzero we don't need to do anything.
@@ -284,8 +284,8 @@ void InterfaceKinetics::convertExchangeCurrentDensityFormulation(doublereal* con
                 // constant so that it's in the exchange current density
                 // formulation format
                 double tmp = exp(m_beta[i] * m_deltaG0[irxn]
-                                 * thermo(reactionPhaseIndex()).RT());
-                tmp *= Faraday * m_ProdStanConcReac[irxn];
+                                 / thermo(reactionPhaseIndex()).RT());
+                tmp *= Faraday * pow(m_ProdStanConcReac[irxn],m_beta[i]);
                 kfwd[irxn] *= tmp;
             }
         }
--- a/test/data/butler-volmer-test.yaml
+++ b/test/data/butler-volmer-test.yaml
@@ -1,0 +1,81 @@
+
+units: {length: cm, time: s, quantity: mol, activation-energy: J/mol}
+
+phases:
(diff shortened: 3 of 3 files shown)


interfaces/cython/cantera/test/test_kinetics.py
```diff
@@ -1215,3 +1215,29 @@ def test_motz_wise(self):
 
         # M&W toggled on (locally) for reaction 9
         self.assertNear(2.0 * k1[9], k2[9]) # sticking coefficient = 1.0
+
+class TestButlerVolmer(utilities.CanteraTest):
+
+    def test_pressure_dependence(self):
+        infile = 'butler-volmer-test.yaml'
+        
+        gas = ct.Solution(infile,'gas')
```
bryanwweber, line 1224: The typical Python convention is one space after a comma

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -284,8 +284,8 @@ void InterfaceKinetics::convertExchangeCurrentDensityFormulation(doublereal* con
                 // constant so that it's in the exchange current density
                 // formulation format
                 double tmp = exp(m_beta[i] * m_deltaG0[irxn]
-                                 * thermo(reactionPhaseIndex()).RT());
-                tmp *= Faraday * m_ProdStanConcReac[irxn];
+                                 / thermo(reactionPhaseIndex()).RT());
+                tmp *= Faraday * pow(m_ProdStanConcReac[irxn],m_beta[i]);
```
bryanwweber, line 288: Can you add a test that covers this line? Without that, there's no guarantee that this turned out right 😊

test/data/butler-volmer-test.yaml
```diff
@@ -0,0 +1,81 @@
...
+    X: {electron: 1.0}
+  density: 9.0 kg/m^3
+- name: electrolyte
+  thermo: ideal-condensed
+  elements: [H, E]
+  species: [H(elyte), Vac(elyte)]
+  state:
+    T: 300.
+    P: 1.01325e+05
+    X: {H(elyte): 1.0, Vac(elyte): 1e-5}
```
bryanwweber, line 30: These don't sum to 1.0. I know they'll be normalized but it seems like it'd be better to do that ahead of time here since it's a test.

interfaces/cython/cantera/test/test_kinetics.py
```diff
@@ -1215,3 +1215,29 @@ def test_motz_wise(self):
 
         # M&W toggled on (locally) for reaction 9
         self.assertNear(2.0 * k1[9], k2[9]) # sticking coefficient = 1.0
+
+class TestButlerVolmer(utilities.CanteraTest):
+
+    def test_pressure_dependence(self):
```
bryanwweber, line 1221: A better name for this might be `test_rate_pressure_independence` to clarify what it's actually testing

bryanwweber (approved):
Just some minor comments. I wonder if it makes sense to post your derivations that led to this fix somewhere?

decaluwe:
Iterating with @wbessler on this, a bit more.  WIll re-open PR when ready.  :(

interfaces/cython/cantera/test/test_kinetics.py
decaluwe: Good idea.  Done.

src/kinetics/InterfaceKinetics.cpp
decaluwe: Yeah, we've got some of that worked up.  Will add.

test/data/butler-volmer-test.yaml
decaluwe: Not sure I see why it matters, but ok.

bryanwweber: I guess the way I see it is that you're implicitly relying on some behavior here which is hidden from anyone looking at just the test. It's a minor thing, but especially for tests I like to be as explicit as possible so we only test what we intend to test.

PR #850, opened 2020-05-08
[Thermo] Provide const access for speciesThermo()

g3bk47:
Provide const access for speciesThermo() member function in case the in case the actual thermo object is only available as const reference and the user wants information about the species thermo.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,7 +18,7 @@ Steven DeCaluwe (@decaluwe), Colorado School of Mines
 Vishesh Devgan (@vdevgan)
 Thomas Fiala (@thomasfiala), Technische Universität München
 David Fronczek
-@g3bk47
+Thorsten Zirwes, Karlsruhe Institute of Technology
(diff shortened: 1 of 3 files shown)


bryanwweber:
You may want to move this change off your master branch as well, considering what happened to #811 😄

AUTHORS
```diff
@@ -18,7 +18,7 @@ Steven DeCaluwe (@decaluwe), Colorado School of Mines
 Vishesh Devgan (@vdevgan)
 Thomas Fiala (@thomasfiala), Technische Universität München
 David Fronczek
-@g3bk47
+Thorsten Zirwes, Karlsruhe Institute of Technology
```
ischoegl: Quick comment: authors are sorted by last name. Also, you may want to retain your GitHub handle.

g3bk47:
Thanks for the hint. I created a new commit to fix the order of authors

speth (approved):

PR #851, opened 2020-05-09
Add get/set equivalenceRatio/mixtureFraction functions

g3bk47:
I accidentally deleted the original pull request. The exact same changes are provided again in this PR. The original discussion can be found here: https://github.com/Cantera/cantera/pull/811


AUTHORS
```diff
@@ -18,7 +18,7 @@ Steven DeCaluwe (@decaluwe), Colorado School of Mines
 Vishesh Devgan (@vdevgan)
 Thomas Fiala (@thomasfiala), Technische Universität München
 David Fronczek
-@g3bk47
+Thorsten Zirwes, Karlsruhe Institute of Technology
```
speth: We generally include the GitHub handle for contributors who have one, i.e.
```suggestion
Thorsten Zirwes (@g3bk47), Karlsruhe Institute of Technology
```
but it's up to you if you don't want to include it.

src/thermo/ThermoPhase.cpp
```diff
@@ -689,6 +689,617 @@ void ThermoPhase::setState_SPorSV(double Starget, double p,
...
+}
+
+double ThermoPhase::getStoichAirFuelRatio_X(const double* fuelComp,
+                                            const double* oxComp) const
+{
+    // indices of elements
+    auto iC = elementIndex("C");
+    auto iS = elementIndex("S");
+    auto iO = elementIndex("O");
+    auto iH = elementIndex("H");
```
speth: For very simple types like `size_t` and `double`, I would suggest using the type names explicitly rather than `auto`.

src/thermo/ThermoPhase.cpp
```diff
@@ -689,6 +689,617 @@ void ThermoPhase::setState_SPorSV(double Starget, double p,
...
+
+    // non-normalized element mole fractions
+    auto Co = 0.0, Cf = 0.0;
+    auto So = 0.0, Sf = 0.0;
+    auto Oo = 0.0, Of = 0.0;
+    auto Ho = 0.0, Hf = 0.0;
+    auto sum_Xf = 0.0, sum_Xo = 0.0;
+
+    for (size_t i=0; i!=m_kk; ++i)
+    {
```
speth: A couple of style notes applicable here and elsewhere:
```suggestion
    for (size_t i = 0; i != m_kk; ++i) {
```
The opening brace should be on the same line for `if` statements as well.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1171,8 +1171,246 @@ class ThermoPhase : public Phase
...
+    //! Fuel and oxidizer compositions are specified by their mass fractions (do
+    //! not need to be normalized). Pressure is kept constant.
+    //! Elements C, S, H and O are considered for the oxidation. 
+    // 
+    //
+    /*!
+     * @param phi        equivalence ratio
+     * @param fuelComp   mass fractions of species in the fuel
+     * @param oxComp     mass fractions of species in the oxidizer 
+     */
```
speth: To get Doxygen to do the right thing, the formatting for the longer comment blocks like this should be
```c++
//! One or two line
//! summary.
/*!
 *! (extended description, including parameters)
 */
```

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1171,8 +1171,246 @@ class ThermoPhase : public Phase
...
+    //
+    /*!
+     * @param mixFrac    mixture fraction (between 0 and 1)
+     * @param fuelComp   mole fractions of species in the fuel
+     * @param oxComp     mole fractions of species in the oxidizer 
+     */
+    void setMixtureFraction_X(double mixFrac, const double* fuelComp, const double* oxComp);
+    
+    //! @copydoc ThermoPhase::setMixtureFraction_X(double mixFrac, const double* fuelComp, const double* oxComp)
+    void setMixtureFraction_X(double mixFrac, const vector_fp& fuelComp, const vector_fp& oxComp);
```
speth: Cantera doesn't generally provide functions which take `vector_fp&` as arguments -- users are expected to use the functions which take simple `double*`. So I think this overload and the other similar ones should be removed. The `string` and `compsitionMap` ones are of course fine.

src/thermo/Phase.cpp
```diff
@@ -1042,4 +1042,18 @@ void Phase::compositionChanged() {
...
 
+vector_fp Phase::getCompositionFromMap(const compositionMap& comp) const
+{
+    vector_fp X(m_kk);
+    for (const auto& sp : comp)
+    {
+        auto loc = speciesIndex(sp.first);
+        if (loc == npos)
+            throw CanteraError("Phase::getCompositionFromMap",
+                               "Unknown species '{}'", sp.first);
```
speth: Here and elsewhere:
```suggestion
        if (loc == npos) {
            throw CanteraError("Phase::getCompositionFromMap",
                               "Unknown species '{}'", sp.first);
        }
```

src/thermo/Phase.cpp
```diff
@@ -1042,4 +1042,18 @@ void Phase::compositionChanged() {
     m_stateNum++;
 }
 
+vector_fp Phase::getCompositionFromMap(const compositionMap& comp) const
```
speth, line 1031: This function could be used to provide most of the implementation for the `Phase::setMoleFractionsByName(const compositionMap& xMap)` and `Phase::setMassFractionsByName(const compositionMap& yMap)` functions.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -745,91 +747,250 @@ cdef class ThermoPhase(_SolutionBase):
...
+        
+        self.state = original_state 
+        
+        self.thermo.setMixtureFraction_X(mixFrac, &f[0], &o[0])
+
+
+    def get_equivalence_ratio(self, fuel=None, oxidizer=None, basis='mole'):
+        """
+        Get the equivalence ratio of the current mixture. This is a quantity
+        that is conserved after oxidation. Considers the oxidation of C to CO2,
```
speth: As can be surmised from my comments on #521, I agree that this is a more useful definition of the equivalence ratio. However, I think that we need to provide a definition of this method that is backwards-compatible for the next Cantera release.

This could be done by adding a keyword argument such as `behavior='old'` where 'old' is the default that uses the current method but triggers a deprecation warning, and setting `behavior='new'` is required to get the new behavior for now.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1171,8 +1171,246 @@ class ThermoPhase : public Phase
...
+    //! @copydoc ThermoPhase::getEquivalenceRatio_Y(const double* fuelComp, const double* oxComp) 
+    double getEquivalenceRatio_Y(const vector_fp& fuelComp, const vector_fp& oxComp) const;
+    //! @copydoc ThermoPhase::getEquivalenceRatio_Y(const double* fuelComp, const double* oxComp) 
+    double getEquivalenceRatio_Y(const std::string& fuelComp, const std::string& oxComp) const;
+    //! @copydoc ThermoPhase::getEquivalenceRatio_Y(const double* fuelComp, const double* oxComp) 
+    double getEquivalenceRatio_Y(const compositionMap& fuelComp, const compositionMap& oxComp) const;
+ 
+    //! Compute the equivalence ratio for the current mixture from elemental
+    //! composition assuming that the fuel only contains the elements C, S, H
+    //! (and elements that do not oxidize) and the fuel only contains O (and
```
speth:
```suggestion
    //! (and elements that do not oxidize) and the oxidizer only contains O (and
```

interfaces/cython/cantera/thermo.pyx
```diff
@@ -745,91 +747,250 @@ cdef class ThermoPhase(_SolutionBase):
...
+            Xo = self.X
+
+        cdef np.ndarray[np.double_t, ndim=1] f = \
+                np.ascontiguousarray(Xf, dtype=np.double)
+        cdef np.ndarray[np.double_t, ndim=1] o = \
+                np.ascontiguousarray(Xo, dtype=np.double)
+        
+        self.state = original_state 
+        
+        self.thermo.setMixtureFraction_X(mixFrac, &f[0], &o[0])
```
speth: Watch out for end-of-line whitespace, here and in several other places. I recommend configuring your editor to highlight it so it can be avoided.
```suggestion

        self.state = original_state

        self.thermo.setMixtureFraction_X(mixFrac, &f[0], &o[0])
```

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -287,23 +287,72 @@ def test_get_equivalence_ratio(self):
...
+
+        fuel = "CH4:0.2,O2:0.02,N2:0.1,CO:0.05,CO2:0.02"
+        ox = "O2:0.21,N2:0.79,CO:0.04,CH4:0.01,CO2:0.03"
+
+        gas =  ct.Solution("gri30.yaml")
+        gas.set_equivalence_ratio(1.3, fuel, ox)
+        gas.TP = 300, 1e5
+
+        # set mixture to  burnt state to make sure that equivalence ratio and
+        # mixture fraction are indipendent of reaction progress
```
speth:
```suggestion
        # mixture fraction are independent of reaction progress
```

test/thermo/ThermoPhase_Test.cpp
```diff
@@ -109,4 +109,80 @@ TEST_F(TestThermoMethods, setState_AnyMap)
...
+    ox["N2"] = 0.79;
+    ox["CO"] = 0.04;
+    ox["CH4"] = 0.01;
+    ox["CO2"] = 0.03;
+
+    // set equivalence ratio to 1.3
+    gas.setEquivalenceRatio_X(1.3, fuel, ox);
+
+    // set mixture to  burnt state to make sure that equivalence ratio and
+    // mixture fraction are indipendent of reaction progress
```
speth:
```suggestion
    // mixture fraction are independent of reaction progress
```

test/thermo/ThermoPhase_Test.cpp
```diff
@@ -109,4 +109,80 @@ TEST_F(TestThermoMethods, setState_AnyMap)
...
+
+    EXPECT_NEAR(phi, 1.3, 1e-4);
+    EXPECT_NEAR(phi2, 1.3, 1e-4);
+    EXPECT_NEAR(phi_loc, 1.1726068608, 1e-4);
+    EXPECT_NEAR(mf, 0.13415725911, 1e-4);
+    EXPECT_NEAR(l, 6.5972850678733, 1e-4);
+
+    // do the same for *_Y functions
+
+    // convert fuel and oxdizer compositions to mass fractions
```
speth:
```suggestion
    // convert fuel and oxidizer compositions to mass fractions
```

src/thermo/ThermoPhase.cpp
```diff
@@ -689,6 +689,617 @@ void ThermoPhase::setState_SPorSV(double Starget, double p,
...
+
+
+    auto denominator = 2.0*(Z_Cf-Z_Co) + 0.5*(Z_Hf-Z_Ho) 
+                     + 2.0*(Z_Sf-Z_So) - (Z_Of-Z_Oo);
+
+    if (denominator == 0.0)
+        throw CanteraError("ThermoPhase::getMixtureFraction_Y",
+                "fuel and oxidizer have the same composition");
+
+    auto mf = ( 2.0*(Z_C -Z_Co) + 0.5*(Z_H -Z_Ho) + 2.0*(Z_S -Z_So) - (Z_O -Z_Oo) )
```
speth:
```suggestion
    auto mf = (2.0*(Z_C - Z_Co) + 0.5*(Z_H - Z_Ho) + 2.0*(Z_S - Z_So) - (Z_O - Z_Oo))
```

src/thermo/ThermoPhase.cpp
```diff
@@ -689,6 +689,617 @@ void ThermoPhase::setState_SPorSV(double Starget, double p,
...
+                                         const vector_fp& oxComp) const
+{
+    return getMixtureFraction_X(fuelComp.data(), oxComp.data());
+}
+
+double ThermoPhase::getMixtureFraction_X(const double* fuelComp,
+                                        const double* oxComp) const
+{
+    // mass fractions of fuel and oxidizer
+    vector_fp massFractions(m_kk * 2);
```
speth: Using a single vector to store both the fuel and oxidizer compositions is an unnecessary complication.

src/thermo/ThermoPhase.cpp
```diff
@@ -689,6 +689,617 @@ void ThermoPhase::setState_SPorSV(double Starget, double p,
...
+double ThermoPhase::getEquivalenceRatio_Y(const double* fuelComp,
+                                          const double* oxComp) const
+{
+    vector_fp X(2*m_kk); // mole fractions of fuel and oxidizer
+
+    // mean molecular weights of fuel and oxidizer
+    auto rmmw_f = std::inner_product(fuelComp, fuelComp+m_kk, molecularWeights().data(),
+            0.0, std::plus<double>(), std::divides<double>());
+    auto rmmw_o = std::inner_product(oxComp, oxComp+m_kk, molecularWeights().data(), 0.0,
+            std::plus<double>(), std::divides<double>());
```
speth: I think I would prefer the readability of the straightforward loop for doing this calculation.

src/thermo/ThermoPhase.cpp
```diff
@@ -689,6 +689,617 @@ void ThermoPhase::setState_SPorSV(double Starget, double p,
...
+
+    if (rmmw_f == 0.0 || rmmw_o == 0.0)
+        throw CanteraError("ThermoPhase::getEquivalenceRatio_Y",
+                           "no fuel and/or oxidizer composition specified");
+
+    auto YtoX = [=](const double& Xk, const double& Mk){return Xk/(rmmw_f*Mk);};
+
+    // compute mole fractions for fuel and oxidizer
+    std::transform(fuelComp, fuelComp+m_kk, molecularWeights().data(), X.data(), YtoX);
+    std::transform(oxComp, oxComp+m_kk, molecularWeights().data(), X.data()+m_kk, YtoX);
```
speth: I think I would prefer the readability of the straightforward loop for doing this calculation.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1171,8 +1171,246 @@ class ThermoPhase : public Phase
...
+    //! elements that do not oxidize)
+    // 
+    //
+    /*!
+     * @param returns    equivalence ratio 
+     */
+
+    double getEquivalenceRatio() const;
```
speth: So this is essentially an implementation of the previous definition of `getEquivalenceRatio`, is that correct? In that case, is it necessary or useful?

speth (changes requested):
Thanks for submitting this pull request. In general, I think this will be a useful addition.

One concern I have is the sheer number of new functions introduced here, especially in the C++ interface. What I think would help would be to use an argument to switch between molar and mass basis, rather than having separate function names for each. This could just be an optional `basis` argument, which would mirror what you've done in the Python interface.

include/cantera/thermo/ThermoPhase.h
g3bk47: This function is the special case of assuming that fuel and oxidizer are pure (what is sometimes called "local equivalence ratio", because it only considers the locally available oxygen and required oxygen and is therefore independent of fuel or oxidizer compositions). I introduced this function for convenience, because you can just write `gas.getEquivalenceRatio()` instead of specifying a concrete oxidizer and fuel composition like `gas.getEquivalenceRatio("H2:1","O2:1")`, where the actual values do not matter as long as the compositions only contain fuel and oxidizer elements, respectively. 
In the python interface, I solved this by defaulting the oxidizer and fuel compositions to `None`. However, defaulting the `compositionMap& oxComp` to `compositionMap{}` or `double* oxComp` to `nullptr` feels a but clunky in the C++ interface. I could rename the function to `getLocalEquivalenceRatio`, but this might be more confusing. What do you think?

g3bk47:
Thanks for the comments. I will make the requested changes. Just a few questions:

1. About the `behavior='old'`: This would mean that the default behavior of the python interface is different from default behavior in the C++ interface (I think it doesn't make sense to bring the incorrect "old" python implementation to C++). And should the use of the old behavior be marked as deprecated?

2. Regarding the `bases="mole/mass"`: Do you want to specify this as a std::string argument or should I introduce an enum to the C++ side similar to https://github.com/Cantera/cantera/blob/01a4e904be2aa3bfbccb594c93ee0916f6ce3a6c/interfaces/cython/cantera/thermo.pyx#L8-L10

3. I saw that there is already a mixture_fraction function for couterflow flames:
https://github.com/Cantera/cantera/blob/113bd4b89ea9094dea013f5e57efe612018b1e51/interfaces/cython/cantera/onedim.py#L1358-L1378
This is based on a single element, whereas the new functions are technically the Bilger mixture fraction based on all elements. I could rename the new functions to BilgerMixtureFraction. Or maybe there could be another argument like `element="O"`, `element="C"` or `element="Bilger"` (where `"Bilger"` is the default), so the user can switch between the single element mode (which might be useful for tracking differential diffusion effects) and the Bilger mode, which considers all elements.

speth:
> Thanks for the comments. I will make the requested changes. Just a few questions:
> 
> 1. About the `behavior='old'`: This would mean that the default behavior of the python interface is different from default behavior in the C++ interface (I think it doesn't make sense to bring the incorrect "old" python implementation to C++). And should the use of the old behavior be marked as deprecated?

Yes, I think the "old" definition can be deprecated. I agree that bringing the "old" definition into C++ is not worthwhile, so we will just have to deal with the discrepancy for this version.

> 2. Regarding the `bases="mole/mass"`: Do you want to specify this as a std::string argument or should I introduce an enum to the C++ side similar to https://github.com/Cantera/cantera/blob/01a4e904be2aa3bfbccb594c93ee0916f6ce3a6c/interfaces/cython/cantera/thermo.pyx#L8-L10

For the C++ interface, probably an `enum class` would be the most idiomatic, something like
```c++
enum class ThermoBasis
{
    mass,
    molar
};
```

> 3. I saw that there is already a mixture_fraction function for couterflow flames:
>    https://github.com/Cantera/cantera/blob/113bd4b89ea9094dea013f5e57efe612018b1e51/interfaces/cython/cantera/onedim.py#L1358-L1378
>    
>    This is based on a single element, whereas the new functions are technically the Bilger mixture fraction based on all elements. I could rename the new functions to BilgerMixtureFraction. Or maybe there could be another argument like `element="O"`, `element="C"` or `element="Bilger"` (where `"Bilger"` is the default), so the user can switch between the single element mode (which might be useful for tracking differential diffusion effects) and the Bilger mode, which considers all elements.

That sounds like an interesting generalization, and having the option of getting the Bilger mixture fraction for the counterflow flame would be nice. I like the idea of doing this with an extra, optional argument as you've suggested.

g3bk47:
I think I addressed all requested changes in the most recent commits. Just one comment:
I changed the name from `ThermoBasis `to `ThermoBasisType `for the enum class in the C++ interface because the name clashed with the enum in the python interface.
https://github.com/Cantera/cantera/pull/851/files#diff-6e2333a075896e676562c6cacb247be3R142-R146
and
https://github.com/Cantera/cantera/blob/d02f2a728290387a78fe27feaa81f12382d2e1cf/interfaces/cython/cantera/thermo.pyx#L8-L10 One is an enum and the other an enum class, so there is no implicit conversion between them. I guess the `ThermoBasis `could be changed to use the `ThermoBasisType `definition so that everything is named `ThermoBasis`. But I didn't want to change the existing code in this PR.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -40,6 +40,19 @@ const int cSS_CONVENTION_VPSS = 1;
 const int cSS_CONVENTION_SLAVE = 2;
 //@}
 
+/*!
+ * @name CONSTANTS - Specification of the input mixture composition 
+ */
+//@{
```
speth: Despite the error in the existing code, there should only be one block (delimited by `@{` and `@}`) with the name "constants", and this enum should go inside that block.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -40,6 +40,19 @@ const int cSS_CONVENTION_VPSS = 1;
...
+//! Differentiate between mole fractions and mass fractions for input mixture composition
+enum class ThermoBasisType : int
```
speth: Since the `ThermoBasis` enum in `thermo.pyx` is effectively just an implementation detail and the name `ThermoBasis` itself is never used again, I'd be comfortable renaming that one and having this one get the shorter name `ThermoBasis`.

It appears that support for `enum class` in Cython is still a pending feature (https://github.com/cython/cython/pull/3640) so we shouldn't worry about trying making these two implementations use the same underlying class.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1172,7 +1185,179 @@ class ThermoPhase : public Phase
...
+    //! mixture fraction = kg fuel / (kg oxidizer + kg fuel)
+    /*!
+     * Fuel and oxidizer compositions are given either as
+     * mole fractions or mass fractions (specified by `basis`)
+     * and do not need to be normalized. Pressure is kept constant.
+     * Elements C, S, H and O are considered for the oxidation.
+     *
+     * @param mixFrac    mixture fraction (between 0 and 1)
+     * @param fuelComp   composition of the fuel
+     * @param oxComp     composition of the oxidizer
```
speth: There are trailing spaces on this line and a few others. Can you clean these up? They're pretty easy to spot if you use `git log --patch --color`.
```suggestion
     * @param oxComp     composition of the oxidizer
```

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1172,7 +1185,179 @@ class ThermoPhase : public Phase
...
+     *                   in terms of the Bilger mixture fraction, or 
+     *                   an element name, to compute the mixture fraction
+     *                   bsaed on a single element (default: "Bilger")
+     */
+    double getMixtureFraction(const double* fuelComp, const double* oxComp,
+                              ThermoBasisType basis = ThermoBasisType::molar, const std::string& element = "Bilger") const;
+    //! @copydoc ThermoPhase::getMixtureFraction_X(const double* fuelComp, const double* oxComp, ThermoBasisType basis = ThermoBasisType::molar, const std::string& element = "Bilger")
+    double getMixtureFraction(const std::string& fuelComp, const std::string& oxComp,
+                              ThermoBasisType basis = ThermoBasisType::molar, const std::string& element = "Bilger") const;
+    //! @copydoc ThermoPhase::getMixtureFraction_X(const double* fuelComp, const double* oxComp, ThermoBasisType basis = ThermoBasisType::molar, const std::string& element = "Bilger")
```
speth: These two methods reference the old method name `getMixtureFraction_X`. This and a number of other problems with the Doxygen docstrings are visible if you look at the output of building the Doxygen docs with `scons doxygen`. Apologies that the output of that command isn't completely clean on the master branch, but you can of course ignore any warnings that aren't associated with the changes in this PR.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1172,7 +1185,179 @@ class ThermoPhase : public Phase
...
+     * @param basis      either ThermoPhase::mole or ThermoPhase::mass
+     *                   fuel and oxidizer composition are interpreted
+     *                   as mole or mass fractions (default: mole)
+     * @param element    either "Bilger" to compute the mixture fraction
+     *                   in terms of the Bilger mixture fraction, or 
+     *                   an element name, to compute the mixture fraction
+     *                   bsaed on a single element (default: "Bilger")
+     */
+    double getMixtureFraction(const double* fuelComp, const double* oxComp,
+                              ThermoBasisType basis = ThermoBasisType::molar, const std::string& element = "Bilger") const;
```
speth: I apologize for not catching this before, but I realized that the names of these methods are somewhat inconsistent with the convention that we usually follow in Cantera, where the "getter" methods don't have the "get" prefix except when the computed value is "returned" in an array argument, e.g. `double pressure()` versus `getMassFractions(double*)`.

For the Python interface, this has I think a small advantage of allowing us to simply deprecate the old `get_equivalence_ratio` method and implement the `equivalence_ratio` method using the new definition without having to support the old method with the same function signature.

Would you mind changing the names so that this is more consistent? If not, I can handle that refactoring.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1172,7 +1185,179 @@ class ThermoPhase : public Phase
...
+    //@}
+
+    //! @name Set Mixture Composition by Equivalence Ratio
+    //! @{
+
+    //! Set the mixture composition according to the equivalence ratio.
+    /*!
+     * Fuel and oxidizer compositions are given either as
+     * mole fractions or mass fractions (specified by `basis`)
+     * and do not need to be normalized. Pressure is kept constant.
```
speth: Pressure and temperature, right?

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1172,7 +1185,179 @@ class ThermoPhase : public Phase
...
+     * Elements C, S, H and O are considered for the oxidation.
+     *
+     * @param phi        equivalence ratio
+     * @param fuelComp   composition of the fuel
+     * @param oxComp     composition of the oxidizer 
+     * @param basis      either ThermoPhase::mole or ThermoPhase::mass
```
speth: `ThermoBasisType::mole` and `ThermoBasisType::mass`, right? (or just `ThermoBasis` pending my other suggestion)

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1172,7 +1185,179 @@ class ThermoPhase : public Phase
...
+    /*!
+     * Computes the equivalence ratio \f$ \phi \f$ from 
+     * \f[ \phi = \frac{Z_{\mathrm{mole},C} + Z_{\mathrm{mole},S} + \frac{1}{4}Z_{\mathrm{mole},H}}
+     * {\frac{1}{2}Z_{\mathrm{mole},O}} \f]
+     * where \f$ Z_{\mathrm{mole},m} \f$ is the elemental mole fraction
+     * of element \f$ m \f$. In this special case, the equivalence ratio
+     * is independent of a fuel or oxidizer composition  because it only
+     * considers the locally available oxygen compared to the required oxygen
+     * for complete oxidation. It is the same as assuming that the oxidizer
+     * only contains O(and inert elements) and the fuel contains only
```
speth:
```suggestion
     * only contains O (and inert elements) and the fuel contains only
```

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1172,7 +1185,179 @@ class ThermoPhase : public Phase
...
+
+    //! Compute the equivalence ratio for the current mixture
+    //! from available oxygen and required oxygen
+    /*!
+     * Computes the equivalence ratio \f$ \phi \f$ from 
+     * \f[ \phi = \frac{Z_{\mathrm{mole},C} + Z_{\mathrm{mole},S} + \frac{1}{4}Z_{\mathrm{mole},H}}
+     * {\frac{1}{2}Z_{\mathrm{mole},O}} \f]
+     * where \f$ Z_{\mathrm{mole},m} \f$ is the elemental mole fraction
+     * of element \f$ m \f$. In this special case, the equivalence ratio
+     * is independent of a fuel or oxidizer composition  because it only
```
speth:
```suggestion
     * is independent of a fuel or oxidizer composition because it only
```

include/cantera/thermo/ThermoPhase.h
speth: I think what you have here makes sense. Maybe it would help to add a suggestion to use the versions that take fuel and oxidizer compositions if either of these conditions for the fuel or oxidizer isn't met.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1172,7 +1185,179 @@ class ThermoPhase : public Phase
...
+
+    //! Compute the equivalence ratio for the current mixture
+    //! given the compositions of fuel and oxidizer
+    /*!
+     * The equivalence ratio \f$ \phi \f$ is computed from
+     * \f[ \phi = \frac{Z}{1-Z}\frac{1-Z_{\mathrm{st}}}{Z_{\mathrm{st}}} \f]
+     * where \f$ Z \f$ is the Bilger mixture fraction of the mixture
+     * given the specified fuel and oxidizer compositions and 
+     * \f$ Z_{\mathrm{st}} \f$ is the mixture fraction at stoichiometric
+     * conditions. Fuel and oxidizer compositions are given either as
```
speth, line 1291: It might be helpful to add a `@see getMixtureFraction(...)` here to provide a reference to the definition of the Bilger mixture fraction.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1172,7 +1185,179 @@ class ThermoPhase : public Phase
...
+    //! @name Compute Stoichiometric Air to Fuel Ratio
+    //! @{
+
+    //! Compute the stoichiometric air to fuel ratio (kmol oxidizer to kmol fuel)
+    //! given fuel and oxidizer compositions.
+    /*!
+     * Fuel and oxidizer compositions are given either as
+     * mole fractions or mass fractions (specified by `basis`)
+     * and do not need to be normalized.
+     * Elements C, S, H and O are considered for the oxidation.
```
speth, line 1345: Perhaps it would be worth noting that the stoichiometric air fuel ratio does not depend on the current mixture composition, and that the method to get the current air fuel ratio would be to take FA_st / phi.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -287,23 +287,90 @@ def test_get_equivalence_ratio(self):
...
+        # make sure the pressure was held constant
+        p = gas.P
+
+        self.assertNear(phi, 1.3)
+        self.assertNear(phi2, 1.3)
+        self.assertNear(phi_loc, 1.1726068608)
+        self.assertNear(mf, 0.13415725911)
+        self.assertNear(mf_C, (gas.elemental_mass_fraction("C")-Y_Co)/(Y_Cf-Y_Co))
+        self.assertNear(mf_O, (gas.elemental_mass_fraction("O")-Y_Oo)/(Y_Of-Y_Oo))
+        self.assertNear(l, 6.5972850678733)
```
speth: I don't think this should be the same as the stoichiometric fuel air ratio on a molar basis (line 336). Checking the value for a very simple case gives a result that suggests that it's still using a molar basis when 'mass' is specified:
```python
>>> gas.get_stoich_air_fuel_ratio('CH4', 'O2', 'mass')
2.0
```

interfaces/cython/cantera/thermo.pyx
```diff
@@ -714,28 +713,30 @@ cdef class ThermoPhase(_SolutionBase):
...
+            >>> gas.set_equivalence_ratio(0.5, 'CH4', 'O2:1.0, N2:3.76', basis='mole')
             >>> gas.mole_fraction_dict()
-            {'CH4': 0.049900199, 'N2': 0.750499001, 'O2': 0.199600798}
-
-            >>> gas.set_equivalence_ratio(1.2, {'NH3':0.8, 'CO':0.2}, 'O2:1.0')
+            {'CH4': 0.028376330528516805, 'N2': 0.7452356312613028, 'O2': 0.22638803821018033}
+            >>> gas.set_equivalence_ratio(1.2, {'NH3':0.8, 'CO':0.2}, 'O2:1.0', basis='mole')
             >>> gas.mole_fraction_dict()
-            {'CO': 0.1263157894, 'NH3': 0.505263157, 'O2': 0.36842105}
+            {'CO': 0.14784006249290751, 'NH3': 0.3595664554540104, 'O2': 0.49259348205308207}
```
speth: I don't follow why the expected output for these examples would change. Shouldn't they be the same? I still get the original output when testing both of these.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -745,91 +746,317 @@ cdef class ThermoPhase(_SolutionBase):
...
+            Only for behavior='new'. Determines if *fuel* and *oxidizer* are
+            given in mole fractions (basis='mole') or mass fractions (basis='mass')
+
+            >>> gas.set_equivalence_ratio(0.5, 'CH3:0.5, CH3OH:.5, N2:0.125', 'O2:0.21, N2:0.79, NO:0.01')
+            >>> gas.get_equivalence_ratio('CH3:0.5, CH3OH:.5, N2:0.125', 'O2:0.21, N2:0.79, NO:0.01', behavior='new')
+            0.5
+        """
+
+        if behavior == 'old':
+
```
speth: I prefer not to have blank line at start of a block.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -745,91 +746,317 @@ cdef class ThermoPhase(_SolutionBase):
...
-        and S to SO2. Other elements are assumed not to participate in oxidation
-        (i.e. N ends up as N2).
+            raise ValueError("basis must either be 'mass' or mole'.")
+
+        cdef np.ndarray[np.double_t, ndim=1] f = \
+                np.ascontiguousarray(fu, dtype=np.double)
+        cdef np.ndarray[np.double_t, ndim=1] o = \
+                np.ascontiguousarray(ox, dtype=np.double)
+
+        self.state = original_state
```
speth: Given that this block now appears 5 times in `thermo.pyx`, it's probably worth factoring out into a private helper method.

src/thermo/ThermoPhase.cpp
```diff
@@ -689,6 +689,438 @@ void ThermoPhase::setState_SPorSV(double Starget, double p,
...
+                               "fuel and oxidizer have the same composition");
+        }
+
+        double mf = (2.0*(Z_C - Z_Co) + 0.5*(Z_H - Z_Ho) + 2.0*(Z_S - Z_So) - (Z_O - Z_Oo))
+               / denominator;
+
+        return std::min(std::max(mf, 0.0), 1.0);
+    }
+    else // compute the mixture fraction from the mass fractions of a single element
+    {
```
speth:
```suggestion
    } else {
        // compute the mixture fraction from the mass fractions of a single element
```

test/thermo/ThermoPhase_Test.cpp
```diff
@@ -109,4 +109,226 @@ TEST_F(TestThermoMethods, setState_AnyMap)
     EXPECT_NEAR(thermo->temperature(), 298.15, 1e-6);
 }
 
+TEST(TestMixtureMethods, getSet_EquilRatio_MixtureFraction)
```
speth: It's probably worth breaking this up into a few separate test cases -- basically, one for each of the places where you have a comment. You can put any parts that get reused between tests in a fixture class and then use the `TEST_F(TestClassName, testName)` style that you see in some other tests.

speth (changes requested):
Thanks for providing the updates, I think they have helped a lot. I have some more comments, but I think most of these are relatively minor issues.

g3bk47:
There are still some inconsistencies, good catch. I will take care of making all requested changes ASAP. To address your questions:
1. regarding the output of the stoichiometric air fuel ratio
`gas.get_stoich_air_fuel_ratio('CH4', 'O2', 'mass') -> 2.0`
This is technically a minor inconsistency, which I briefly mentioned before (https://github.com/Cantera/cantera/pull/811#issuecomment-588295033)
The ThermoBasis in the python interface affects either the output of some functions (like density) or the input of the setters (like `HP`). In the functions I introduced, the ThermoBasis only affects how the inputs (fuel and oxidizer concentrations) are interpreted, never the output. There are basically three functions I introduced:

- get equivalence ratio: doesn't the matter if user specifies mass or mole because the concept of equivalence ratio pretty independent of it
- get mixture fraction: returns always kg fuel / kg mixture, regardless of ThermoBase. I virtually do not know anyone who uses a mixture fraction in terms of mol fuel / mol mixture so I think a user might expect this behavior, therefore my choice (which makes it technically inconsistent to the python interface).
- get stoich air to fuel ratio: always returns mol oxidizer / mol air, regardless of ThermoBase, just because it was convenient for my implementation of the other functions. Therefore, the function above treats the input as 1 kg CH4/1 kg fuel and 1 kg O2 /1 kg oxidizer, but returns the stoichiometric air to fuel ratio as mol oxidizer / mol fuel. If you want, I can change the output when the ThermoBasis is fuel, but then what about the mixture fraction function? Or maybe the name should be changed to something like "molarStoichiometricAirToFuelRatio"? Or alternatively, only the mixture fraction function could be renamed to massMixtureFraction, but I think this might be more confusing.

2. About the changed output `{'CO': 0.14784006249290751, 'NH3': 0.3595664554540104, 'O2': `0.49259348205308207} in the `gas.mole_fraction_dict()` example: I will double check it.

speth:
Ah yes, I see your point about the stoichiometric air fuel ratio, and I realize now that the documentation does indeed say that it provides the AFR on a molar basis. Like the mixture fraction, though, I think that the mass-based definition is the one more often used. 

If you switched this to use the mass-based definition, you would just need to swap mole fractions for mass fractions in the equivalence ratio methods, and they will work the same way.

g3bk47:
In addition to the requested changes, I introduced two helper functions on the C++ side (`o2Required `and `o2Present`). Together with the change, that `stoichAirToFuelRatio `returns kg oxidizer/kg fuel, this led to significant simplification of the code.
As a nice side effect, all functions now work with mass fractions internally, just like the thermo object itself. And an additional error is detected in some cases, where fuel and oxygen compositions are mixed up.

Regarding the doxygen comment for `ThermoBasis`: I removed the open block `//@{`, but left everything else in place.

g3bk47:
One python test in the CI has failed ("test_ignition_delay_sensitivity").
However, I am unable to reproduce this on my system. When I check the output, I get:

`assertNear: -1.26076767538e-09 vs. -1.26487098268e-09 (OK)`

When I look at the result of

`gas.set_equivalence_ratio(0.4, 'H2', 'O2:1.0, AR:4.0')`

I get e.g. for the mass fraction of H2:

`y(H2) = 0.00833872772358`

When I calculate this by hand, the correct mass fraction should be

`y(H2) = 4032/483527 = 0.0083387277235811...`

which is pretty much the same. So maybe someone else should double check why this test fails, before merging the PR.

speth:
I just rebased this onto the current main branch, and didn't get any test failures locally. Hopefully, it will pass the checks on the CI servers as well.

bryanwweber:
The failures seem to be on macOS only... I can take a look tomorrow now that the conda packages seem sorted out

g3bk47:
The test failure persists on macOS. At least, the "wrong" result seems so to be always the same numerical value (across all different python versions and also from the last CI run), which might make debugging easier. Is there a way to get an image of the macOS VM (if it is a VM), so I can help debugging?

bryanwweber:
I can't reproduce this locally on my mac. I'm going to push some debugging statements to this branch to see what the CI is doing.

speth:
I pushed a commit to tighten the tolerances on the sensitivity variables, which at least on my computer gives better agreement between the finite difference and sensitivity coefficient-based calculations in this test (worst relative agreement is around 1e-3, which is much less than the test tolerance of 5e-2). Hopefully this will be enough to pass the CI tests.

The problem seems to stem from the fact that the default tolerances on the sensitivity variables is fairly loose (1e-4), but tightening them has a tendency to lead to integration failures. I suspect this is another reminder that we need to take another look at the details of how we solve the equations for the sensitivities, since it's not as efficient or robust as I think should be possible. But that's obviously well outside the scope of this PR.

bryanwweber:
Thanks @speth! Feel free to drop my commit with the debug statements (either you or @g3bk47).

> I suspect this is another reminder that we need to take another look at the details of how we solve the equations for the sensitivities, since it's not as efficient or robust as I think should be possible.

Can you file an issue about this so we don't forget?

speth (approved):

speth:
@bryanwweber I documented the issues I've observed with sensitivity analysis in Cantera/enhancements#55.

PR #857, opened 2020-06-03
Fix #853 raise an error when there is number in an element symbol during ck2yaml conversion

### LINKED ISSUE #853: Error when parsing ck2cti or ck2yaml with large element number(more than 3 digits)
**System information**

- Cantera version: 2.5.0
- OS: Ubuntu 18.0.4
- Python/MATLAB version: 3.8.0

**Expected behavior**

Regarding to the [issue](https://groups.google.com/forum/?fromgroups=#!topic/cantera-users/JK5HM5wTgys) on Cantera google user group.  The expected behavior is a successful conversion.
**Actual behavior**

It was not clear where is the error position when the user used ck2cti in Cantera 2.4.0 and python 3.7, so I tried Cantera 2.5.0 and python 3.8. The reason was that there are several species has element numbers more than 3 digits, for example, `BIN7AJ                 C 1250H 812          G    300.00   3500.00 1260.00      1`.
The function in ck2cti,  [readThermoEntry](https://github.com/Cantera/cantera/blob/7d87534bb7ef0a55bca2ab3dc4282a2ae7ddd8d5/interfaces/cython/cantera/ck2cti.py#L1015)(line 1032 `composition = self.parseComposition(lines[0][24:44], 4, 5) ` ) does not split the string correctly, `[24:44] `only covers part of the element symbols and numbers, especially when the element number is very large (species `BIN12AJ ` in the thermo file). Besides, the input value to function [parseComposition](https://github.com/Cantera/cantera/blob/7d87534bb7ef0a55bca2ab3dc4282a2ae7ddd8d5/interfaces/cython/cantera/ck2cti.py#L966) on line 1032 only successfully converts when the element number has less than 3 digits.

I am not sure if these large number are added in the file accidentally, but should Cantera add some feature to enable conversion for species with large element number?

**To Reproduce**

1. Open the directory where the chemkin files are saved
2. Run` ck2cti --input=input.CKI --thermo=thermo.CKT --transport=trans.TRC --output=new.cti --debug --permissive`
3. See error 
```
  File "/home/chao/anaconda3/envs/cantera38/bin/ck2cti", line 11, in <module>
    load_entry_point('Cantera==2.5.0a4', 'console_scripts', 'ck2cti')()
  File "/home/chao/anaconda3/envs/cantera38/lib/python3.8/site-packages/cantera/ck2cti.py", line 2352, in script_entry_point
    main(sys.argv[1:])
  File "/home/chao/anaconda3/envs/cantera38/lib/python3.8/site-packages/cantera/ck2cti.py", line 2323, in main
    surfaces = Parser.convertMech(inputFile, thermoFile, transportFile,
  File "/home/chao/anaconda3/envs/cantera38/lib/python3.8/site-packages/cantera/ck2cti.py", line 2264, in convertMech
    surface_names = parser.writeCTI(name=phaseName, outName=outName)
  File "/home/chao/anaconda3/envs/cantera38/lib/python3.8/site-packages/cantera/ck2cti.py", line 2068, in writeCTI
    speciesNames = self.getSpeciesString(self.speciesList, 21)
  File "/home/chao/anaconda3/envs/cantera38/lib/python3.8/site-packages/cantera/ck2cti.py", line 2012, in getSpeciesString
    raise InputParseError("Undefined elements in species '{}':"
cantera.ck2cti.InputParseError: Undefined elements in species 'BIN7AJ': '1'
```

**Attachments**

[input.txt](https://github.com/Cantera/cantera/files/4690917/input.txt)
[thermo.txt](https://github.com/Cantera/cantera/files/4690919/thermo.txt)
[trans.txt](https://github.com/Cantera/cantera/files/4690920/trans.txt)

**Additional context**

12Chao:
### Problem:  
Regarding to #853,  When there is large element number(more than 3 digits) in the thermo file trying to fit everything in one line, instead writing an additional line with element numbers like this: https://github.com/Cantera/cantera/blob/7d87534bb7ef0a55bca2ab3dc4282a2ae7ddd8d5/test/data/soot-therm.dat#L39-L49
The ck2yaml file could accidentally convert an incorrect element symbol in species composition, which includes part of the element number in the element symbol string for example: The composition for  
`BIN7AJ                 C 1250H 812          G    300.00   3500.00 1260.00      1` is parsed as `{'1': 250, H: 812}` It won't pass the validation, but raise an error before a yaml file gets created maybe more helpful.

### Fix: 
Check if there is any number in the symbol and if the symbol is in the elements list when creating the element symbol. If the check failed, raise the error thermo entry and mention the first line has an incorrect format.
**Checklist**

Fixes #853 

 raise an error including the entire thermo entry when parsing a species thermo data with big element number in incorrect format


interfaces/cython/cantera/ck2yaml.py
```diff
@@ -865,6 +865,14 @@ def read_NASA7_entry(self, lines, TintDefault, comments):
         if not composition:
             raise InputError("Error parsing elemental composition for "
                              "species '{}'", species)
+        else:
+            for symbol in composition.keys():
+                # Here it checks if there is a number in the symbol string and if this is
+                # in the element list as a customized element
```
bryanwweber: The comment here explains *what* the line below it is doing, but not *why*. Most programmers could understand the *what* by just reading the code, so the comment isn't helping in that sense. Can you add a comment about *why* this change is needed instead?

interfaces/cython/cantera/ck2yaml.py
```diff
@@ -865,6 +865,14 @@ def read_NASA7_entry(self, lines, TintDefault, comments):
...
+                if any(map(str.isdigit, symbol)) == True and symbol not in self.elements:
```
bryanwweber: The way Python works, you don't need the `== True`, just `any(map(str.isdigit, symbol))` should be enough.

interfaces/cython/cantera/ck2yaml.py
```diff
@@ -865,6 +865,14 @@ def read_NASA7_entry(self, lines, TintDefault, comments):
...
+                    raise InputError("Error parsing elemental composition for "
+                             "species thermo entry\n{}"
+                             "\n the first line has an incorrect format", "".join(lines))
```
bryanwweber: Can you modify the formatting of these lines so that they conform to PEP 8? Thanks 😄

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -468,7 +468,13 @@ def test_extra(self):
...
         for key in ['foo', 'bar']:
             self.assertIn(key, yml.keys())
-
+    
+    def test_error_for_big_element_number(self):
+        try:
+            self.convert('big_ele_num_err_mech.txt', thermo='big_ele_num_err_thermo.txt',
+                        output='big_ele_num_err.yaml')
+        except Exception as e:
+            self.assertIn('the first line has an incorrect format', str(e))
```
bryanwweber: You can use the `self.assertRaisesRegex` context manager here, rather than the try/except block. Check out line 443 in this file for an example.

bryanwweber: You can also leave the extension off of the output file and the test class should be able to handle that.

bryanwweber (changes requested):
Thanks for this change @12Chao! It's a great start. I have a few comments below. One additional comment is, rather than including a large mechanism, we'd rather have the smallest possible mechanism that can reproduce the error condition. You can have it as small as one species, and even put the thermo data into the input file.

12Chao:
Hi @bryanwweber, thanks for the comments! I rebased the commit, and I think I have got all the changes you mentioned earlier in the comments. Please let me know if I should change anything.

bryanwweber:
Hi @12Chao I made a few more changes to simplify things a bit and change the comment to clarify what I meant about explaining why 😄  Let me know what you think!

12Chao:
Thanks @bryanwweber! I think these changes really help make thinge clear. One thing I am not sure is that should we mention that there is a specified chemkin format about how to add large element number in the thermo entry. I feel like that is helpful but it will probably affect the simplicity of the error message.

bryanwweber:
@12Chao That's a good point. I can't quite think of a better way to phrase it to point out there is an alternate syntax, without, as you say, removing the simplicity of the error message. It's not that elements can't have more than 3 digits, it's that they can't have more than three digits when specified in the default location. But I don't know how to point that out in a way that is useful for the end user.

speth:
Would it be helpful to include the URL to our help on debugging issues in Chemkin files (https://cantera.org/tutorials/ck2yaml-tutorial.html#debugging-common-errors-in-ck-files)? One flaw with this option is that our instructions there don't currently provide any help for this type of error, and don't describe the supported syntax for extended composition information, but those are things that you could fix.

12Chao:
I think it is a good option, and the syntax for extended composition information is something good to have on the Cantera website no matter if we add url to the error message or not. I can work on adding the instruction about that syntax on Cantera website, and it would be nice to add the url to other error messages raised due to the indentation error or other format errors so that people can compare the format there as a template. 

I am thinking that would it be helpful if we add a flag like `--verbose` to enable an explaination about an error and the possible fix?

speth:
Thanks, @12Chao, adding that documentation would be helpful. And adding links to other error messages would be helpful as well.

If you want to avoid repeating the URL in many places, you could just assign it to a global variable and substitute that into the error messages.

I wouldn't worry too much about adding a `--verbose` flag -- I think it would be appropriate to provide this information by default.

speth:
@12Chao, Can you rebase this on to the current `main` branch so it will run the new CI tests (added in #775)?

12Chao:
> @12Chao, Can you rebase this on to the current `main` branch so it will run the new CI tests (added in #775)?

Sorry, I guess that I just messed it up, and  I will open a new pull request about this change

speth:
I don't think you need to close this PR. All you need to do is push a branch that contains the current `main` plus your 1 commit to this same branch name.

12Chao:
Thanks @speth, I just pulled branch 'main' and pushed it with the commit to the same b
ranch again

bryanwweber:
I believe I've fixed the merge conflicts that were happening and reset this branch to the correct state, rebased onto the current tip of `main`.

I have one suggestion here, which is that since the same message is added to every `InputError`, why not just add it to that class? Something like:
```
class InputError(Exception):
    def __init__(self, message, *args, **kwargs):
        message += "\nPlease check https://cantera.org/... for the correct Chemkin syntax."
        if args or kwargs:
            # the rest of the method is the same
```

Then, each time `InputError` is called, just make sure the existing message ends with a period so that the grammar works out.

bryanwweber:
By the way @12Chao, to get your branch the same as this one, do

```
git fetch origin
git reset --hard origin/improve_ck2yaml
```

The first command updates the reference for this branch, and the second one resets your working directory and history to match the updated reference. Just make sure you there are no uncommitted changes in your folder before you run the second one.

12Chao:
Thanks for the fix @bryanwweber, that's really helpful. I have added the message with url to the `InputError`.

interfaces/cython/cantera/ck2yaml.py
```diff
@@ -151,6 +154,7 @@ class InputError(Exception):
     the exceptional behavior.
     """
     def __init__(self, message, *args, **kwargs):
+        message += "\nPlease check {} for the correct Chemkin syntax.".format(url)
```
speth: Since `url` is now only used here, I think Bryan's suggestion was to just fold it into the message directly, that is:
```suggestion
        message += ("\nPlease check https://cantera.org/tutorials/ck2yaml-tutorial.html#debugging-common-errors-in-ck-files"
            "\nfor the correct Chemkin syntax.")
```
and delete the definition of `url` above.

12Chao: Thanks @speth for the revision. I think that I have pulled the latest changes from `main` branch and made the suggested change, but it seems like ruamel-yaml could not be installed successfully on the builds for all the systems with python3.8.  I am not sure what caused that, could you please give me some hints about how to fix it?

12Chao:
Thanks @speth for the revision. I have pulled the latest version of Cantera and made the suggested changes, but it seems ruamel-yaml cannot be built successfully on system with Python3.8 on Travis. I am not sure how to fix that, and it would be really appreciated if you can give me some hints about how to fix that

speth:
@12Chao, the revision I made was mostly to get rid of the merge commit and to combine a few commits which were just modifying the same bits of code. As a reminder for the future, please do not *merge* changes from the `main` branch into your PR branch. If you want to update your PR, *rebase* your branch onto the head of the current `main` branch.

The test failures have nothing to do with any of your code, so it isn't necessary for those to pass for this PR to be merged. See #924 for what we currently know about those test failures.

speth (approved):

12Chao:
Thanks @speth, I will do what you said on another pull request I made.

bryanwweber (approved):

PR #859, opened 2020-06-05
Update find_isomers to return dictionary

ischoegl:
This simple update addresses a comment by @bryanwweber (see #635)

>  ... `find_isomers` returns species names, not indices, resulting in an additional step for users. I suppose users would have to look up the species name for the indices returned by `species_index` anyways, since the index by itself is not very meaningful

As `find_isomers` was introduced recently and only exists in the development branch, the change of the interface is inconsequential.

**New behavior:**

```
In [1]: import cantera as ct

In [2]: gas = ct.Solution('gri30.yaml')

In [3]: gas.find_isomers('C:1, H:2')
Out[3]: {'CH2': 10, 'CH2(S)': 11}
```

**Checklist**

### DIFF
--- a/include/cantera/thermo/Phase.h
+++ b/include/cantera/thermo/Phase.h
@@ -800,13 +800,13 @@ public:
 
     //! Return a vector with isomers names matching a given composition map
     //!     @param compMap compositionMap of the species.
-    //!     @return A vector of species names for matching species.
-    virtual std::vector<std::string> findIsomers(const compositionMap& compMap) const;
+    //!     @return A map of species names and indices for matching species.
+    virtual std::map<std::string, size_t> findIsomers(const compositionMap& compMap) const;
 
     //! Return a vector with isomers names matching a given composition string
     //!     @param comp String containing a composition map
-    //!     @return A vector of species names for matching species.
-    virtual std::vector<std::string> findIsomers(const std::string& comp) const;
+    //!     @return A map of species names and indices for matching species.
+    virtual std::map<std::string, size_t> findIsomers(const std::string& comp) const;
 
     //! Return the Species object for the named species. Changes to this object
     //! do not affect the ThermoPhase object until the #modifySpecies function
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -197,8 +197,8 @@ cdef extern from "cantera/thermo/ThermoPhase.h" namespace "Cantera":
         cbool caseSensitiveSpecies()
         void setCaseSensitiveSpecies(cbool)
         void addSpeciesAlias(string, string) except +translate_exception
-        vector[string] findIsomers(Composition&) except +translate_exception
-        vector[string] findIsomers(string) except +translate_exception
+        stdmap[string, size_t] findIsomers(Composition&) except +translate_exception
+        stdmap[string, size_t] findIsomers(string) except +translate_exception
 
         double molecularWeight(size_t) except +translate_exception
         double meanMolecularWeight()
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -1158,15 +1158,27 @@ class TestSpecies(utilities.CanteraTest):
             self.gas.add_species_alias('spam', 'eggs')
 
     def test_isomers(self):
+        def check(iso):
+            self.assertIsInstance(iso, dict)
+            for name, index in iso.items():
+                self.assertIn(name, gas.species_names)
+                self.assertEqual(gas.species_index(name), index)
+
         gas = ct.Solution('nDodecane_Reitz.yaml')
         iso = gas.find_isomers({'C':4, 'H':9, 'O':2})
+        check(iso)
         self.assertTrue(len(iso) == 2)
         iso = gas.find_isomers('C:4, H:9, O:2')
         self.assertTrue(len(iso) == 2)
+        check(iso)
         iso = gas.find_isomers({'C':7, 'H':15})
         self.assertTrue(len(iso) == 1)
+        check(iso)
         iso = gas.find_isomers({'C':7, 'H':16})
         self.assertTrue(len(iso) == 0)
+        check(iso)
+        with self.assertRaisesRegex(ct.CanteraError, 'Invalid'):
+            gas.find_isomers(3)
 
 
 class TestSpeciesThermo(utilities.CanteraTest):
@@ -1644,7 +1656,7 @@ class TestSolutionArray(utilities.CanteraTest):
     def test_extra(self):
         with self.assertRaises(ValueError):
             states = ct.SolutionArray(self.gas, extra=['creation_rates'])
-        
+
     def test_append(self):
         states = ct.SolutionArray(self.gas, 5)
         states.TPX = np.linspace(500, 1000, 5), 2e5, 'H2:0.5, O2:0.4'
--- a/interfaces/cython/cantera/thermo.pyx
+++ b/interfaces/cython/cantera/thermo.pyx
@@ -604,17 +604,20 @@ cdef class ThermoPhase(_SolutionBase):
 
     def find_isomers(self, comp):
         """
-        Find species/isomers matching a composition specified by *comp*.
+        Find species/isomers matching a composition specified by *comp*;
+        returns a dictionary of species names and indices.
+
+        >>> phase.find_isomers('C:1, H:2')
+        {'CH2': 10, 'CH2(S)': 11}
         """
-
         if isinstance(comp, dict):
-            iso = self.thermo.findIsomers(comp_map(comp))
-        elif isinstance(comp, (str, bytes)):
-            iso = self.thermo.findIsomers(stringify(comp))
-        else:
-            raise CanteraError('Invalid composition')
+            return {pystr(item.first): item.second
+                    for item in self.thermo.findIsomers(comp_map(comp))}
+        if isinstance(comp, (str, bytes)):
+            return {pystr(item.first): item.second
+                    for item in self.thermo.findIsomers(stringify(comp))}
 
-        return [pystr(b) for b in iso]
+        raise CanteraError('Invalid composition')
 
     def n_atoms(self, species, element):
         """
--- a/src/thermo/Phase.cpp
+++ b/src/thermo/Phase.cpp
@@ -973,20 +973,20 @@ void Phase::addSpeciesAlias(const std::string& name, const std::string& alias)
     }
 }
 
-vector<std::string> Phase::findIsomers(const compositionMap& compMap) const
+map<std::string, size_t> Phase::findIsomers(const compositionMap& compMap) const
 {
-    vector<std::string> isomerNames;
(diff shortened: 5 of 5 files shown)


bryanwweber:
Thinking out loud here... I think the list of string names is useful, but if a user wanted a list of strings of isomer names, they could do `list(find_isomers(...).keys())`, and likewise for indices. So that's not an issue.

I guess maybe that it's not clear what the value of the dictionary corresponds to, unless you read the docstring... I wonder if a `NamedTuple` interface would be better? You could return a list of `NamedTuple`s with attributes `name` and `index`? `dataclasses` might also be a good fit, but they're only in Python 3.7 and up. The downside of this approach is that one has to access an individual isomer by index in the resulting list, and the index in that list does not correspond to anything useful... OK I talked myself out of this idea :joy:

OK, another question, related to the last - why return the index when you could return the `Species` instance as the value in the dictionary? I suppose it depends on what you want to use this for. If you want to fill an array for, say, setting mole fractions then the species index might be useful. Or for some of the internal uses mentioned in #603. But since the Solution object can be "indexed" by species names, and then you can *get* mole fractions like:

```
gas["CH4"].X
```

You don't actually need the species index for a lot of operations... Can we think about whether it makes sense to return the index or the Species instance? Maybe it makes sense to do different things in Python (user-facing) vs. C++ (more internally facing)?

ischoegl:
@bryanwweber ... I think there are many variations here. I am ultimately addressing what I consider a “paper cut”, where I thought it to be somewhat nicer to get both names and indices - the latter are crucial when slicing, although there are alternatives as you point out. I don’t think that reading a docstrings to get the meaning of the value is a disadvantage ... this is what docstrings are for and the keys are self-explanatory.

What I am proposing here could also help address #635 implicitly: I agree with your comment that returning a list is not consistent with other implementations. The crux is that isomers are inherently ill defined, so pointing users to the updated `find_isomers` method when raising an error would result in what I consider a more consistent solution.

bryanwweber:
@ischoegl can you clarify what you mean by isomers being ill-defined? As far as I know, they have a very specific definition, the same elemental composition with varying physical structure.

As far as this PR, I would prefer to get it right this time, rather than finding out we need/want to change it again later. 

> That's what docstrings are for

Yes, but if the code can be self-documenting for the user (i.e., the "type" is obvious from printing it) that's even better IMO.

I'm not trying to say that returning `Species` instances is definitely better, but let's at least think about it for a bit. Can you think of a reason _not_ to return the `Species`?

Last, I suspect that with the change here `findIsomers()` (in C++) can replace the need for #635 entirely, with the possible exception that it is kind of orthogonally named to what is envisioned in #603, i.e., finding the index of a species for other uses within Cantera.

ischoegl:
@bryanwweber  ... I'd be happy to have a discussion.

> can you clarify what you mean by isomers being ill-defined? As far as I know, they have a very specific definition, the same elemental composition with varying physical structure.

... in terms of inherent unique-ness, the elemental composition allows for various constellations. My comment was not meant as not being able to define what isomers are, so we're in full agreement.

> As far as this PR, I would prefer to get it right this time, rather than finding out we need/want to change it again later.

... not sure that 'getting it right' is what this PR had in mind. In terms of my original goals (find species names based on composition, which was a task that I frequently ran into), the current implementation is perfectly fine. I was mainly trying to address your comment in #635 in a constructive way.

> [...] code can be self-documenting for the user 

Absolutely, but it has to be balanced with complexity, e.g. `NamedTuples` may be useful, but they're not commonly used. I have no reservations against using these 'under the hood', but I honestly think that keeping things simple for the end-user needs to be another objective. Imho, standard objects (`list`, `tuple` and `dict`) should be used to interface with users whenever possible, as these can be assumed to be familiar to most.

> I'm not trying to say that returning `Species` instances is definitely better, but let's at least think about it for a bit. Can you think of a reason *not* to return the `Species`?

... when slicing a numpy (or MATLAB) array, the species object is not useful. For `Species`, the cython interface may be tricky but can probably be worked out. Regardless, the question is whether this fits with the overarching philosophy of how species are accessed. It's still a discussion worth having, but I'm somewhat partial to the slicing argument.

> I suspect that with the change here `findIsomers()` (in C++) can replace the need for #635 entirely, with the possible exception that it is kind of orthogonally named to what is envisioned in #603, ...

I'm not going to second-guess here :man_shrugging: ... I still think that a method that finds species indices by composition is useful. The only thing I'd lobby for is to keep interfaces consistent, i.e. a function `species_index` should probably return an integer value in all of its variations. 

So, please don't take it the wrong way ... this is a friendly suggestion, not more.

**PS:** one option addressing self-documentation would be to have `find_isomers('C:1, H:2')` return a `list` (i.e. no change from current behavior), and `find_isomers('C:1, H:2', index=True)` return the `dict` that includes indices.

bryanwweber:
To say up front: I like this change, I think the dictionary interface is the right interface for returning multiple values to the user in this case, and I'd like to see this resolved so it can be merged!

> not sure that 'getting it right' is what this PR had in mind.

If 'getting it right' wasn't in mind, then I think it should have been an issue first to discuss different options? This is not meant as a dig or insult to you, but I believe that we should do our best to get the interface right before we merge a new feature. Absent that, if it comes up again then something wasn't fully right the first time (by definition, otherwise would not have come up again) so it should be discussed. Since this is the second time we're returning to this function (well, first time returning I suppose...), IMO we should do a better job this time of thinking through the options in front of us.

> I was mainly trying to address your comment in #635 in a constructive way.

Yes, I think the change here is definitely an improvement, I like the dictionary interface it seems very useful (I'm not in favor of `NamedTuple` and only included it in my last comment to document my train of thought so _I_ don't forget what I've already thought of 😄 ).

Given that the dictionary is the appropriate interface, I agree that the keys should be the string names of the species. The main question to be resolved here is what the values of the dictionary should be.

> when slicing a numpy (or MATLAB) array, the species object is not useful.

I guess this was the crux of my last comment about returning `Species` instances. How often does a user need to slice NumPy arrays, vs. access the properties of the species? Since at least the `Solution` object can be "sliced" by providing species names (i.e., the keys of the dictionary), it is still possible to do a certain amount of these operations even if the `Species` instance is returned.

I don't think it's possible to answer the question I just posed in a general sense. However, my personal feeling is that the `Species` instance would be more useful to the end user in Python. But this is just a gut feeling, I don't have any evidence for it, and there's something to be said (as you note) for an integer value there. 😕 That said...

> PS: one option addressing self-documentation would be to have `find_isomers('C:1, H:2')` return a list (i.e., no change from current behavior), and `find_isomers('C:1, H:2', index=True)` return the dict that includes indices.

I like this option, I wonder if it makes sense to extend it to a second option that returns the `Species` instance as the values? (This is a question for discussion, not an implicit suggestion to implement this right away) This does mean the function will return different values depending on the kwargs passed, but I suppose that's not all that uncommon and at least it is an explicit kwarg, not implicit based on the input type. 

I think in C++ it makes sense for the value of the `map` to be the index.

(I'm going to leave aside MATLAB because this PR doesn't change the MATLAB interface and I think having different return values in different interfaces is more-or-less fine, especially when MATLAB doesn't have quite the same thing as a dictionary from Python.)

I think it'd be good to get perspective from other people as well. @decaluwe @speth Do you have any thoughts here?

ischoegl:
@bryanwweber ... 

> [...] but I believe that we should do our best to get the interface right before we merge a new feature. 

Please rest assured that I’m having exactly this in mind whenever I add something. No need to second-guess my motivation here. Please keep in mind that I’m volunteering my time here, so let’s please keep the discussion civil.

ischoegl:
I am closing this for technical reasons motivated by PEP 20. The current implementation of `find_isomers` is simple and does exactly what one may expect. I don’t think there’s any need to make this more complex and/or add clutter to the Python interface.

Further, I believe that the C++ underpinnings of the existing implementation de facto already address most of #603, as the species index can be easily found In C++ based on the content of the map `Phase.m_speciesIndices` once the name is identified.

decaluwe:
So, not exactly sure why this issue got closed, but to weigh in with my 2 cents: I find myself agreeing with @ischoegl on two fronts:

- In terms of isomers being "ill defined": my read here of what @ischoegl intended (correct me if wrong) is that the output from this function, by nature, requires some "operator" in the loop. i.e. there is no automated way, in most cases, to discriminate between the various isomers. In most cases, I would think the user is in search of one particular isomer, and wants to interrogate the returned list and select one in particular.
- Thinking about how I use Cantera, I need species indices for any number of operations not related to Cantera objects and slicing. If I need to handle one particular species differently from others (the reason I'd be looking for a species index in the first place), I might need to access it in an any number of arrays that I have declared, external to Cantera, including arrays of species fluxes, chemical production rates, and in derivative/residual arrays. The index is much more useful to me, here, because the slicing can be applied uniformly to slice my arrays and to any arrays produced by Cantera.
- That said, it might be ultimately better to just return the string, and have the user take the one extra step to find the species index for that string, dynamically. In the case I've mentioned above, my implementation would be to find the particular isomer I want, and identify it somehow in my model input file, where I typically store it in a `dict` of parameters. In this case, having the parameter be the string is both clearer/more meaningful for anyone following behind and trying to understand my code, relative to an arbitrary species index, and is also safer.  It is possible, though perhaps not terribly likely, that someone changes the species orders in the input file, in which case the index is no longer valid.  This could happen, for example, if they add a new species to the phase, and add it to the front of the list (again, doesn't seem likely, but is not impossible).  On the other hand, the chance that they would go through and change the species name (which requires changes in multiple places) seems less likely.  Furthermore, if the species name does change, the user likely gets an error (a species under the old name can no longer be located), whereas an incorrect index will likely go undetected.  

Thinking some more, this last bullet point is exactly what my current approach is.  When needed, I specify the species name in my input file, then run routines in my model to find the index to manipulate how that species is handled, in a generalized way.  At present, I interrogate the input file manually to find this species name, which is okay since my mechanisms are small.  But if I had a large mechanism and had to discriminate between isomers, even if the name strings don't tell me exactly what I need to know, I can use the names to search the input file and read more.

So I guess I've come around to where I do understand why @ischoegl closed the PR 😁.

@bryanwweber - can you explain what extra use cases would be enabled by returning the `species` instance?  Could the user interrogate this instance to help get extra information to help discriminate between the two?

ischoegl:
@decaluwe ... while I tend to use large mechanisms, my mode of operation is very similar to yours. I returned to PEP20 as I'm afraid that this discussion is getting carried away in terms of trying to anticipate what an end-user wants. As this is impossible (all of us have different coding styles and expectations), defaulting to the simplest possible interface is probably the best solution.

bryanwweber:
@decaluwe The `Species` instance contains all the information that Cantera needs to know about a given species, such as its thermodynamic data, charge, transport parameters, etc. Eventually, I imagine that will grow to include data specified in YAML input files as well (notes, etc.). So I don't think there's anything in there which would allow discrimination between different isomers at the moment, but most likely someday.

That said, constructing a dictionary in this way would be pretty easy for the end user with a comprehension:

```python
{name: gas.species(name) for name in gas.find_isomers({'C': 1, 'H': 2})}
```
So it seems like leaving the return value from `.find_isomers()` as a list is fine, and closing the PR seems to be the consensus.

PR #862, opened 2020-06-08
Make species charge an accessible phase property

ischoegl:
**Checklist**

- [x] There is a clear use-case for this code change
- [x] The commit message has a short title & references relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [ ] The pull request is ready for review

**If applicable, fill in the issue number this pull request is fixing**

Fixes Cantera/enhancements#50 

**Changes proposed in this pull request**

- Add `ThermoPhase.electrical_charges` property to Python interface
- Consistent with `molecular_weights`, the property returns a `np.ndarray`

**Thoughts**

The initial version of this PR is meant to be a starting point for discussion. 

* The array is already defined in C++ as `Phase::m_speciesCharge` ... it is straight-forward to thread this array through the Cython interface (I used the comprehension for the time being).
* It is also debatable whether to use `electrical_charges`, `species_charges` or other for the property name (on further thought, `species_charges` may be preferable, but I'll leave things as they are for discussion).

I am leaving the box 'ready for review' unchecked until the discussion has converged.

PR #863, opened 2020-06-08
[Thermo] Add property to call species charges in phase

### LINKED ISSUE #50: Make species charge an accessible phase property
Make the array of species elementary charges an accessible phase property, e.g. via the Python or Matlab interfaces

At present a user must run through the species and save this variable array external to cantera, similar to: 

```
for j in phase.species_names:
    params['Z_k'][phase.species_index(j)] = phase.species(j).charge
```
It should be relatively straightforward to make this array accessible, similar to `phase.molecular_weights`.

**Motivation**
This has utility in a number of electrochemical applications, where the species charge number is used to relate a species flux to a current density.

korffdm:
This commit adds the ability for the user to call the charge
 of all species within a phase. This required updates to the
phase source code and the Python wrapper files.

**Checklist**

Fixes https://github.com/Cantera/enhancements/issues/50
Similar to #862 

- Add `ThermoPhase.species_charges` as a Property in the Python wrapper
- Follows format and method for `molecular_weights` code

**Notes**

In #862 there is already a test set up, is it possible to incorporate that test into this code or would a separate test need to be written for the approach used here?


ischoegl:
> In #862 there is already a test set up, is it possible to incorporate that test into this code or would a separate test need to be written for the approach used here?

@korffdm  ... Feel free to recylce the test! Simply replace `electrical_charges` with `species_charges` and it should run. PR #862 is closed, so simply copy the section of `test_thermo.py` over to your PR.

include/cantera/thermo/Phase.h
```diff
@@ -471,6 +471,23 @@ class Phase
     //! units = kg / kmol
     const vector_fp& molecularWeights() const;
 
+    //! Charge of species \c k.
+    //!     @param k index of species \c k
+    //!     @returns the charge of species \c k.
+    doublereal speciesCharge(size_t k) const;
```
speth: This is a duplicate of `Phase::charge`, isn't it?

include/cantera/thermo/Phase.h
```diff
@@ -471,6 +471,23 @@ class Phase
...
+
+    //! Copy the vector of species charges into vector charges.
+    //!     @param charges Output vector of species charges (-)
+    void getSpeciesCharges(vector_fp& charges) const;
```
speth: Cantera does not generally provide overloads specialized for `vector_fp`, since the more flexible `double*` version can easily be used with a `vector_fp` argument via the `.data()` method.

include/cantera/thermo/Phase.h
```diff
@@ -471,6 +471,23 @@ class Phase
...
+
+    //! Copy the vector of species charges into array weights.
+    //!     @param weights Output array of species charges (-)
+    void getSpeciesCharges(doublereal* charges) const;
```
speth: Please use `double` rather than `doublereal` in any new code (but don't change the use of the `doublereal` macro on any lines that would otherwise be unmodified).

Presumably `weights` in the summary line should be `charges`?

Is the notation `(-)` meant to suggest that this is dimensionless? I think it would be more correct to say that this is in units of the elementary charge.

src/matlab/phasemethods.cpp
```diff
@@ -196,6 +196,9 @@ void phasemethods(int nlhs, mxArray* plhs[],
         case 22:
             iok = thermo_getMolecularWeights(ph,nsp, &x[0]);
             break;
+        case 23:
+            iok = thermo_getSpeciesCharges(ph, nsp, &x[0]);
+            break;
```
speth, line 201: Are you planning on adding the `.m` file needed to make this available in Matlab?

include/cantera/thermo/Phase.h
```diff
@@ -471,6 +471,23 @@ class Phase
...
+
+    //! Return a const reference to the internal vector of molecular weights.
+    //! units = (-)
+    const vector_fp& speciesCharges() const;
```
speth: I know that this is being included as a parallel to the equivalent `molecularWeights()` method, but I don't think this method is necessary. In order to implement the other methods, you can use the `m_speciesCharge` array directly.

src/clib/ct.cpp
```diff
@@ -284,6 +284,19 @@ extern "C" {
         }
     }
 
+    int thermo_getSpeciesCharges(int n, size_t lenm, double* sc)
+    {
+        try {
+            ThermoPhase& p = ThermoCabinet::item(n);
+            p.checkSpeciesArraySize(lenm);
+            const vector_fp& ch = p.speciesCharges();
+            copy(ch.begin(), ch.end(), sc);
```
speth: A simpler implementation of this would be simply
```c++
p.getSpeciesCharges(sc);
```

speth (changes requested):
Congrats on your first Cantera PR, @korffdm! I just have a few suggestions to improve this which I realized are mostly related to warts in the implementation of `getMolecularWeights`, which this parallels.

One question I have which others may want to weigh in on is whether the names used here should just be `charges` for the Python property and `getCharges` for the C++ method, rather than `species_charges`. There isn't anything else associated with a `ThermoPhase` that can have a charge, right?

include/cantera/thermo/Phase.h
korffdm: Ah, yes it is a duplicate, I didn't see that previously. I don't see that method actually implemented in the .cpp file though.

include/cantera/thermo/Phase.h
korffdm: I will update to `double` in the new code, thanks.

Yes, that's a leftover from copying the code segment, I'll fix that.

That makes sense, I wasn't sure how best to express the units.

src/matlab/phasemethods.cpp
korffdm: I haven't really interacted with the Matlab interface much, but I could look at adding this to it.

include/cantera/thermo/Phase.h
korffdm: Ah so just use `m_speciesCharge` directly in `void getSpeciesCharges(double* charges) const;`?

src/clib/ct.cpp
korffdm: In place of line `292`? 

`const vector_fp& ch = p.getSpeciesCharges(sc);`

include/cantera/thermo/Phase.h
korffdm: Ah, so the `vector_fp&` overload is unnecessary if the other method is changed to 

`void getSpeciesCharges(double* charges) const;`?

include/cantera/thermo/Phase.h
speth: It's not -- the implementation is in `Phase.h`, since it's just a one-liner.

src/matlab/phasemethods.cpp
speth: Well, you've already done the hard part by adding this and C wrapper in `ct.cpp`.

include/cantera/thermo/Phase.h
speth: Correct.

src/clib/ct.cpp
speth: No, that would replace both lines. `sc` is a pointer to an the array where you want to store the charges, and `Phase::getSpeciesCharges(double* charges)` does exactly that.

korffdm: Ah I see, that change is now up to date in this PR.

src/matlab/phasemethods.cpp
korffdm: Okay, I'm working on this now, but I've pushed the other change requests in the meantime.

include/cantera/thermo/Phase.h
korffdm: Gotcha, that makes sense. I've removed my duplicate in the new push.

include/cantera/thermo/Phase.h
```diff
@@ -471,6 +471,10 @@ class Phase
     //! units = kg / kmol
     const vector_fp& molecularWeights() const;
 
+    //! Copy the vector of species charges into array charges.
+    //!     @param charges Output array of species charges (elem. charge)
+    const vector_fp& getSpeciesCharges(double* charges) const;
```
speth: I think this function signature should be `void getSpeciesCharges(double* charges) const;`. There's no need to both copy the charges into `charges` and return a reference to the internal data structure.

src/thermo/Phase.cpp
```diff
@@ -559,6 +559,13 @@ const vector_fp& Phase::molecularWeights() const
     return m_molwts;
 }
 
+const vector_fp& Phase::getSpeciesCharges(double* charges) const
+{
+    const vector_fp& ch = m_speciesCharge;
+    copy(ch.begin(), ch.end(), charges);
```
speth: There's no need to make a reference to `m_speciesCharge`. You can just write
```suggestion
    copy(m_speciesCharge.begin(), m_speciesCharge.end(), charges);
```

interfaces/cython/cantera/thermo.pyx
```diff
@@ -660,6 +660,11 @@ cdef class ThermoPhase(_SolutionBase):
         def __get__(self):
             return self._getArray1(thermo_getMolecularWeights)
 
+    property species_charges:
+        """Array of species charges [-]."""
```
speth: The description of units should be updated to reflect that they are in units of the elementary charge.

interfaces/matlab/toolbox/@ThermoPhase/speciesCharges.m
```diff
@@ -0,0 +1,12 @@
+function ch = speciesCharges(tp)
+% SPECIESCHARGES  Get the elementary charge of the species.
```
speth: I would suggest using essentially the same description as you have for the Python property, e.g. "Get the array of species charges".

speth (changes requested):
Thanks for making these updates. I'd like to reiterate my earlier suggestion of removing the adjective "species" from the method/property names, since I think it should be clear that it's the species charges that are being returned.

interfaces/matlab/toolbox/@ThermoPhase/charges.m
```diff
@@ -0,0 +1,12 @@
...
+% x = speciesCharges(tp)
+%
+% :param tp:
+%     Instance of class :mat:func:`ThermoPhase` (or another
+%     object that derives from ThermoPhase)
+% :return:
+%     Vector of species charges. Units: elem. charge
+%
+
+ch = phase_get(tp.tp_id, 23);
```
bryanwweber: You need a newline at the end of this file

bryanwweber (approved):
One small change here to the `.m` file, but otherwise looks good to me.

interfaces/matlab/toolbox/@ThermoPhase/charges.m
korffdm: Just an extra line below line 12? I added that and pushed, but I'm not seeing it on my github repo.

bryanwweber: At least in my view (I think I'm using the new interface) it shows as a red circle with a line in it when there's no new line in the Files Changed view on the PR. The problem seems to be fixed by 32ec0f5 so it's all good. See here for a brief reasoning on why files should end with a blank line: https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

speth (approved):

PR #868, opened 2020-06-18
[Kinetics] Allow negative pre-exponential factor for falloff reactions

speth:
If both the high and low rate coefficients for a falloff or chemically activated reaction are negative, then there is no problem evaluating the rate, so it can be allowed in the same cases where a negative rate would be allowed for an elementary reaction, i.e. if the presence of the negative pre-exponential factors has been explicitly declared.

This complements the change made for three-body reactions in #822.

**Checklist**


bryanwweber (approved):
I pushed a small change for the test function to use the return value from the `convert()` method. Otherwise, this looks good to me.

PR #869, opened 2020-06-19
[Input] Make accessing input file data outside of phase node easier

speth:
Some phase types utilize information in the input file that is contained outside the phase definition (e.g. phase definitions for nested phases). This change to `ThermoPhase::setParameters` makes it easier for those phases to have access to the full input
file definition without re-reading and reparsing the YAML file.

While I had originally thought that these cases were only in some of the dustier corners of Cantera, this also comes up in the new `PlasmaPhase` type being introduced in #700, so I think it's worth having a cleaner interface for this capability.

### DIFF
--- a/include/cantera/thermo/ThermoPhase.h
+++ b/include/cantera/thermo/ThermoPhase.h
@@ -1495,8 +1495,11 @@ public:
     virtual void getParameters(int& n, doublereal* const c) const {
     }
 
-    //! Set equation of state parameters from an AnyMap phase description
-    void setParameters(const AnyMap& phaseNode);
+    //! Set equation of state parameters from an AnyMap phase description.
+    //! Phases that need additional parameters from the root node should
+    //! override this method.
+    virtual void setParameters(const AnyMap& phaseNode,
+                               const AnyMap& rootNode=AnyMap());
 
     //! Access input data associated with the phase description
     const AnyMap& input() const;
--- a/include/cantera/thermo/IonsFromNeutralVPSSTP.h
+++ b/include/cantera/thermo/IonsFromNeutralVPSSTP.h
@@ -273,6 +273,8 @@ public:
     void setNeutralMoleculePhase(shared_ptr<ThermoPhase> neutral);
     shared_ptr<ThermoPhase> getNeutralMoleculePhase();
 
+    virtual void setParameters(const AnyMap& phaseNode,
+                               const AnyMap& rootNode=AnyMap());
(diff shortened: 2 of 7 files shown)


include/cantera/thermo/ThermoPhase.h
```diff
@@ -1495,8 +1495,11 @@ class ThermoPhase : public Phase
     virtual void getParameters(int& n, doublereal* const c) const {
     }
 
-    //! Set equation of state parameters from an AnyMap phase description
-    void setParameters(const AnyMap& phaseNode);
+    //! Set equation of state parameters from an AnyMap phase description.
```
bryanwweber, line 1498: Can this link to the `AnyMap` docs?

bryanwweber (approved):
Small improvement here, otherwise seems like a good generalization.

include/cantera/thermo/ThermoPhase.h
speth: I think Doxygen should do this automatically. See for example the docs for [`ThermoPhase::setState`](https://cantera.org/documentation/dev/doxygen/html/dc/d38/classCantera_1_1ThermoPhase.html#a310b4b2821895a9e11d8b98a949459c9) which contains automatically generated links to several classes:
https://github.com/Cantera/cantera/blob/8a591b4920e8ed525d41e94576f9241f5a890551/include/cantera/thermo/ThermoPhase.h#L1148-L1171
(and one spurious one to `namespace Cantera`.

bryanwweber: Great, I didn't know that Doxygen will autolink arbitrary words that it finds in the code. Guess you'd better not name a method with a common word :joy:

decaluwe:
Thanks, @speth.  I looked over the code changes and don't have too much input, but agree that this looks really useful.

One quasi-related question: could this generalization be useful when importing/initializing interface reactions, where access to phase properties is useful but a little convoluted (at least last time I looked at it, admittedly before yaml serialization)?

speth:
@decaluwe, I don't think there's an obvious parallel with interface reactions, although there have been some changes with the way those are handled since introducing the YAML format. Namely, we now have made explicit the requirement that you have to have a `Kinetics` object (and its associated `ThermoPhase` objects) in order to create a `Reaction` object from an input file, since you can't figure out what the units of the rate constant are without knowing what the units of the standard concentration are for each species in the reaction, and that information is carried by the `ThermoPhase`.

PR #871, opened 2020-06-19
Add HDF output to 1D examples

ischoegl:
HDF output for 1D simulations has been added in #840. This PR adds this output option to the Python examples.

Approach: if `h5py` is installed, output is saved to HDF; if `h5py` is not installed, retain old XML output.

**Checklist**

Fixes: N/A

**Changes proposed in this pull request:**

- use HDF output as long as `h5py` is installed
- remove duplicate `burner_ion_flame.py` example (retain `ion_burner_flame`)
- make output data file names consistent with example names
- use container format (i.e. internal data structure) instead of data folders for batch jobs `diffusion_flame_batch.py` and `diffusion_flame_extinction.py`

**Note:** `diffusion_flame_extinction.py` appears to be broken (loop does not terminate), see #872


interfaces/cython/cantera/examples/onedim/adiabatic_flame.py
```diff
@@ -33,11 +29,12 @@
...
 # Solve with the energy equation enabled
-if h5:
-    f.write_hdf('h2_adiabatic.h5', group='mix', mode='w',
+try:
+    # save to HDF container file if h5py is installed
+    f.write_hdf('adiabatic_flame.h5', group='mix', mode='w',
                 description='solution with mixture-averaged transport')
-else:
-    f.save('h2_adiabatic.xml', 'mix',
+except:
```
speth: Here and in similar cases. The use of the bare `except` clause is generally discouraged.
```suggestion
except ImportError:
```

interfaces/cython/cantera/examples/onedim/diffusion_flame_batch.py
```diff
@@ -27,10 +28,17 @@ class FlameExtinguished(Exception):
     pass
 
 
-# Create directory for output data files
-data_directory = 'diffusion_flame_batch_data/'
-if not os.path.exists(data_directory):
-    os.makedirs(data_directory)
+hdf_output = find_spec('h5py') is not None
```
speth: For a function that's only going to be used once, I think it's more clear to just write `import importlib` above and use the fully qualified function name here.

interfaces/cython/cantera/examples/onedim/diffusion_flame_batch.py
```diff
@@ -27,10 +28,17 @@ class FlameExtinguished(Exception):
...
+
+if not hdf_output:
+    # Create directory for output data files
+    data_directory = 'diffusion_flame_batch_data/'
+    if not os.path.exists(data_directory):
+        os.makedirs(data_directory)
+    fig_name = '{}figure_{{0}}.png'.format(data_directory)
```
speth: In this case, I find this simpler:
```suggestion
    fig_name = data_directory + 'figure_{0}.png'
```

interfaces/cython/cantera/examples/onedim/diffusion_flame_batch.py
```diff
@@ -122,16 +137,24 @@ def interrupt_extinction(t):
     try:
         # Try solving the flame
         f.solve(loglevel=0)
-        file_name = 'pressure_loop_' + format(p, '05.1f') + '.xml'
-        f.save(data_directory + file_name, name='solution', loglevel=1,
-               description='Cantera version ' + ct.__version__ +
-               ', reaction mechanism ' + reaction_mechanism)
+        if hdf_output:
+            group = 'pressure_loop_{:05.1f}'.format(p)
```
speth: Maybe a nice place to use nested HDF5 groups?
```suggestion
            group = 'pressure_loop/{:05.1f}'.format(p)
```
Likewise with `strain_loop`.

speth (changes requested):
Looks good overall. I just had a couple of minor comments.

interfaces/cython/cantera/examples/onedim/diffusion_flame_batch.py
bryanwweber: This may also be a good place to use `pathlib`, there might be a few places you have to cast the `Path` instance to a `str()` though, in Cantera's methods. I haven't tried it though, depending on how things are done internally, it may "just work", but then again, maybe only on Python 3.6+ which is I believe when they switched over to the built-in `open()` allowing `pathlib`.

interfaces/cython/cantera/examples/onedim/diffusion_flame_extinction.py
```diff
@@ -14,14 +14,19 @@
...
-data_directory = 'diffusion_flame_extinction_data/'
-if not os.path.exists(data_directory):
-    os.makedirs(data_directory)
+hdf_output = find_spec('h5py') is not None
+
+if not hdf_output:
+    # Create directory for output data files
+    data_directory = 'diffusion_flame_extinction_data/'
+    if not os.path.exists(data_directory):
+        os.makedirs(data_directory)
```
bryanwweber: Same comment here about `pathlib`

bryanwweber:
Thanks for making these changes @ischoegl! I also had one comment in addition to @speth's

interfaces/cython/cantera/examples/onedim/diffusion_flame_batch.py
ischoegl: I opted to replace the path concatenations by `os.path.join`, i.e. keeping strings albeit in a 'cleaner' approach. Refactoring how paths are handled isn't the main priority here, especially as there's the possibility of dropping XML from Python altogether.

interfaces/cython/cantera/examples/onedim/diffusion_flame_extinction.py
ischoegl: Again, not sure. I can replace this using `pathlib` but still believe that XML output should be dropped here eventually, which would render this moot. Let me know ...

speth (approved):

PR #874, opened 2020-06-20
Add tests of the Fortran module

speth:

- Add tests of the Fortran module, using the existing F77 and F90 "samples"
- Fix some issues to enable compilation using Flang on Windows (installed from conda-forge)

**Checklist**


bryanwweber:
Woo! Fortran! A small comment before my brain turns off completely - the change in 9e4ba1c about path separators appears as though it will be backwards-incompatible with `cantera.conf` files that people might already have. Is there any way to issue warnings/errors, particularly on Windows?

speth:
I don't think it should be backwards-incompatible where it is currently works (Linux / macOS), and on Windows the old behavior was broken, so we don't need to worry about the change in behavior there, right?

bryanwweber:
Yeah, I agree about macOS and Linux. Was it broken on Windows such that the user would get an error message and the build would stop?  In that case, we don't need to worry I guess. But if it was able to work with colons instead of semicolons at all (in other words, no error message even if it ended up doing something incorrect), then I think it would be breaking in the sense that there is now a behavior change for the user (to the correct behavior, but still...).

speth:
I added an error message to what I think is the only case where using a colon as a path separator in `extra_lib_dirs` would have previously worked. The behavior I was seeing is that if you have given a path like `c:\some\path`, SCons was adding `<current directory>\c` and `<current drive>\some\path` as library directories. Which "works" as long as everything is on the same drive.

bryanwweber (approved):
Thanks for adding the warning

PR #880, opened 2020-06-25
Improve error message for missing phases key

### LINKED ISSUE #812: Improve the error message when the phases key is missing from a YAML input file
As reported on the User's Group (https://groups.google.com/d/msg/cantera-users/2sNq2p-fVKM/-fSXMWA9CAAJ), a YAML input file that is missing a top-level `phases` key has a misleading error message:
```
CanteraError:
InputFileError thrown by AnyValue::getMapWhere:
Error on line 1 of /Users/bryan/Downloads/cantera.yaml:
Element is not a mapping or list of mappings
|  Line |
>     1 > species:
          ^
|     2 | - name: H
|     3 |   composition: {H: 1}
|     4 |   thermo:
```
It would be nice if this could be improved to point out the actual problem. Minimal YAML file:

```
species:
- name: H
  composition: {H: 1}
  thermo:
    model: NASA7
    temperature-ranges: [200.0, 1000.0, 6000.0]
    data:
    - [2.5, 0.0, 0.0, 0.0, 0.0, 2.547366e+04, -0.44668285]
    - [2.5, 0.0, 0.0, 0.0, 0.0, 2.547366e+04, -0.44668285]
    note: L6/94
```

ischoegl:
**Checklist**

Fixes #812

- Cython `CxxAnyValue& operator[](string)` uses the operator version that adds a key if it is not found (instead of the `const` version that raises an error in C++)
- Use `AnyMap::at` instead (Cython support for C++ `const` Is incomplete)
- Pre-emptively replace operator `[]` by `AnyMap::at` in other instances

The error now points to the actual problem, i.e.
```
In [1]: import cantera as ct

In [2]: yaml = '''
   ...:         species:
   ...:         - name: H
   ...:           composition: {H: 1}
   ...:           thermo:
   ...:             model: NASA7
   ...:             temperature-ranges: [200.0, 1000.0, 6000.0]
   ...:             data:
   ...:             - [2.5, 0.0, 0.0, 0.0, 0.0, 2.547366e+04, -0.44668285]
   ...:             - [2.5, 0.0, 0.0, 0.0, 0.0, 2.547366e+04, -0.44668285]
   ...:             note: L6/94
   ...:         '''

In [3]: ct.Solution(yaml=yaml)
<ipython-input-11-496cc576f3d1> in <module>()
----> 1 ct.Solution(yaml=yaml)

interfaces/cython/cantera/base.pyx in cantera._cantera._SolutionBase.__cinit__()

interfaces/cython/cantera/base.pyx in cantera._cantera._SolutionBase._init_yaml()

CanteraError: 
InputFileError thrown by AnyMap::at:
Error on line 2 of input string:
Key 'phases' not found.
Existing keys: species
|  Line |
|     1 | 
>     2 >         species:
                  ^
|     3 |         - name: H
|     4 |           composition: {H: 1}
|     5 |           thermo:
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,14 @@ doc/ctdeploy_key
 *.exe.manifest
 build/
 test/work/
+test_problems/ChemEquil_ionizedGas/ChemEquil_ionizedGas
+test_problems/cathermo/testWaterPDSS/WaterPDSS
+test_problems/cathermo/testWaterTP/WaterSSTP
+test_problems/cxx_ex/cxx_ex
+test_problems/diamondSurf/diamondSurf
+test_problems/pureFluidTest/pureFluid
+test_problems/rankine_democxx/rankine_democxx
+test_problems/surfkin/surfkin
 interfaces/cython/cantera/_cantera.h
 include/cantera/base/config.h
 include/cantera/base/config.h.build
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -17,6 +17,22 @@ class TestThermoPhase(utilities.CanteraTest):
     def test_source(self):
         self.assertEqual(self.phase.source, 'h2o2.xml')
 
+    def test_missing_phases_key(self):
+        yaml = '''
+        species:
+        - name: H
+          composition: {H: 1}
+          thermo:
+            model: NASA7
+            temperature-ranges: [200.0, 1000.0, 6000.0]
+            data:
+            - [2.5, 0.0, 0.0, 0.0, 0.0, 2.547366e+04, -0.44668285]
+            - [2.5, 0.0, 0.0, 0.0, 0.0, 2.547366e+04, -0.44668285]
+            note: L6/94
+        '''
+        with self.assertRaisesRegex(ct.CanteraError, "Key 'phases' not found"):
+            _ = ct.Solution(yaml=yaml)
+
     def test_base_attributes(self):
         self.assertIsInstance(self.phase.name, str)
         self.assertIsInstance(self.phase.phase_of_matter, str)
--- a/src/base/AnyMap.cpp
+++ b/src/base/AnyMap.cpp
@@ -656,6 +656,9 @@ const AnyMap& AnyValue::getMapWhere(const std::string& key, const std::string& v
             throw InputFileError("AnyValue::getMapWhere", *this,
                 "Map does not contain a key where '{}' = '{}'", key, value);
         }
+    } else if (is<void>()) {
+        throw InputFileError("AnyValue::getMapWhere", *this,
+            "Key '{}' not found", m_key);
     } else {
         throw InputFileError("AnyValue::getMapWhere", *this,
             "Element is not a mapping or list of mappings");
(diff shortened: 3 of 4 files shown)


ischoegl:
**Comment:** One curious finding I stumbled across is that it's not possible to load `nasa.yaml` into a `Solution` (same goes for `nasa.cti` since at least 2.4).

speth (commented):
I think the changes here look good as far as fixing this for the Python module goes. Could you make the corresponding change for the C++ interface (in the `newPhase` function)?

ischoegl:
@speth ... oh, for some reason I had assumed that this was Python specific. I guess the same rationale applies in C++ when you call a method on something returned by the `[]` operator?

speth:
Yes, and I suspect the fix is as simple as what you did here (or simpler, since you won't get into a fight with proper `const` support).

ischoegl:
@speth ... I verified that the same error occurs in C++. So far, so good; after adding a non-const version of `AnyMap::at` I can apply the same fix as in Cython.

However, similar errors can occur for every single instance where the `[]` operator is used on the right-hand side (as far as I can tell, the obvious intended use for the non-`const` version is for creation of a new field on the left-hand side). There are *numerous* instances (based on searching for `].` instances) for RHS uses. I see two options:

1. Eliminate all RHS uses of the operator `[]`
2. Update the error logic for `AnyMap::getMapWhere` (and other methods) to catch and replace the misleading error. It looks like you'd have to check whether the `AnyMap` instance is empty.
3. Apply both, especially where (2) may have an error message that is insufficiently clear.

speth:
Yes, you're right, there are many places where this can occur. I think option (1) would be annoying to police, so I'd rather avoid that. Option (2) is actually not too difficult for `AnyMap::getMapWhere` -- it's just one additional check in the `if...else` chains that looks like:
```c++
    } else if (is<void>()) {
        throw InputFileError("AnyValue::getMapWhere", *this,
            "Key '{}' not found", m_key);
```
We already do a similar check in `AnyValue::as`, which is the other case that came to mind where this would be an issue, and these two might actually be the end of it.

With this change, I'm wondering whether any of the other changes are even necessary.

ischoegl:
@speth ... this is indeed a workable solution! It throws almost the exact same error (minus the list of available keys).

I'll update accordingly.

ischoegl:
@speth ... done (undid most of the changes, so the PR is much simpler now). 

PS: commit bd8ffd6b94bff804bf6d6afc70ed9162b7c7949a may be superseded by changes in #883 ... 

PPS: the only difference is that empty fields are now created, but I don't think that this matters much.

speth (approved):
Looks good. I think the fix for the test files that needed to be added to `.gitignore` is good - this was the result of changes made in #803, which resulted in the names of some of the regression test executables changing. I didn't notice it at the time because I made those changes on a Windows machine, where all of the test executables are excluded by the `*.exe` filter in `test_problems/.gitignore`.

ischoegl:
@speth ... didn't realized that there was a separate `test_problems/.gitignore` until you mentioned it. I pushed #888 to make this more consistent (I also decluttered and alphabetized).

PR #883, opened 2020-06-27
Add regression tests for C++ sample programs

### LINKED ISSUE #774: Fix examples in `samples` folder

speth:
Similar to the tests added in #874 for the Fortran sample program, this PR adds tests for the C++ samples so they won't break without us knowing.

- Add regression tests to run all of the C++ "sample" programs and compare their output against nominal output files
- Remove redundant regression tests which were essentially duplicates of the sample programs
- Add ability to have data files in subdirectories of the Cantera data directory (see Cantera/enhancements#22, #774)
- Make some improvements to test file comparisons for the C++ regression tests
- Fix issues with default composition of `air.yaml` and use of that file in the `combustor.cpp` sample
- Deprecate the argument `ndim` to `newTransportMgr` which is currently ignored / unused
- Fix erroneous deprecation warnings issued by including `onedim.h` (diagnosed thanks to running the `bvp` sample)

**Checklist**

Fixes #774


ischoegl:
> Fix erroneous deprecation warnings issued by including onedim.h (diagnosed thanks to running the bvp sample)

Missed that one ... thanks for fixing!

speth:
Rebasing this after merging #886 allows the tests to run reliably on all platforms, so I think this is ready now.

samples/cxx/gas_transport/gas_transport.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    // Get multicomponent properties at several temperatures
+    for (int i = 0; i < ntemps; i++) {
+        temp = 500.0 + 100.0*i;
+        gas->setState_TP(temp, pres);
+        output(0,i) = temp;
+        output(1,i) = multi->viscosity();
+        output(2,i) = multi->thermalConductivity();
+        multi->getThermalDiffCoeffs(&output(3,i));
+    }
```
bryanwweber, line 71: Any reason to have 2 separate loops here?

src/base/application.cpp
```diff
@@ -421,36 +422,54 @@ std::string Application::findInputFile(const std::string& name)
...
-            string inname = dirs[i] + "/" + name;
-            std::ifstream fin(inname);
+            string full_name = home + name.substr(1, npos);
+            std::ifstream fin(full_name);
             if (fin) {
-                return inname;
+                return full_name;
+            } else {
+                throw CanteraError("Application::findInputFile",
+                                "Input file '{}' not found", name);
```
bryanwweber: I think the indentation here should line up
```suggestion
                                   "Input file '{}' not found", name);
```

bryanwweber (changes requested):
Looks good, just a few small things. This also means that we no longer need the "Build Samples" step in the CI, right? Since the samples will actually be run during `scons test` now?

samples/cxx/gas_transport/gas_transport.cpp
speth: The loops don't have to be separate, but it seems like a simpler example that way, e.g. if you only care about mixture-averaged properties, you can just delete everything after `writePlotFile("transport_mix.csv", ...)`, and then it's easier to see how the second part is just a variation on that theme.

bryanwweber: Yeah, that makes sense. I was kind of torn myself for that reason. Let's leave it as 2 loops, I think that's better as you say.

speth:
> This also means that we no longer need the "Build Samples" step in the CI, right? Since the samples will actually be run during `scons test` now?

Correct. Just pushed a commit that removes these build steps.

bryanwweber (approved):

PR #886, opened 2020-06-29
Fix reactor mass flow rate

speth:

This PR comes from my attempts to get repeatable output from the `combustor.cpp` sample in #883. I separated these changes out to avoid bloating that PR.

- Ensure that the mass flow rate returned by `FlowDevice` objects is correct after a call to `ReactorNet::advance`, and not the value at some later time that the integrator reached.
- Deprecate the `time` argument to `FlowDevice::massFlowRate` (and equivalent methods in the various language interfaces), since getting the value at any time other than the current integrator time may not be correct.
- Remove a few variables / functions that are no longer necessary as a result of the above changes.

**Checklist**

### DIFF
--- a/include/cantera/clib/ctreactor.h
+++ b/include/cantera/clib/ctreactor.h
@@ -53,7 +53,8 @@ extern "C" {
     CANTERA_CAPI int flowdev_del(int i);
     CANTERA_CAPI int flowdev_install(int i, int n, int m);
     CANTERA_CAPI int flowdev_setMaster(int i, int n);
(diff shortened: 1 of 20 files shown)


bryanwweber (approved):
This looks good to me. One small point for discussion is whether this would be a good time to change the references to the `master` flow devices, or whether that would be better in another PR. I'm fine either way.

@paulblum these changes will affect what you're working on.

speth:
I was going to save that for another PR. I'm *trying* to keep my PRs to a manageable size. 😂

bryanwweber:
The foreberance is appreciated 🤣

PR #891, opened 2020-07-02
[SCons] Check for minimum version of ruamel.yaml

### LINKED ISSUE #831: Insufficient ruamel.yaml dependencies on older systems
**System information**

- Cantera version: 2.5.0a4
- OS: Ubuntu Xenial with default packages
- Python/MATLAB version: 3.5.2

**Expected behavior**

`scons test` works out of the box, after successful `scons build` 

**Actual behavior**

Test fails
```
  File "/src/test/python/runCythonTests.py", line 35, in <module>
    from cantera.test.utilities import unittest
  File "/src/build/python/cantera/test/__init__.py", line 13, in <module>
    from .test_convert import *
  File "/src/build/python/cantera/test/test_convert.py", line 8, in <module>
    from cantera import ck2cti, ck2yaml, cti2yaml, ctml2yaml
  File "/src/build/python/cantera/ck2yaml.py", line 100, in <module>
    yaml.RoundTripRepresenter.add_representer(float, represent_float)
AttributeError: module 'ruamel.yaml' has no attribute 'RoundTripRepresenter'
```

**To Reproduce**

1. Install Xenial (ubuntu 16.04 - same as current Travis) and update `apt-get update`
2. Install default `ruamel-yaml` via `apt-get install python3-ruamel.yaml` (which has version '0.10.23')
3. See above failure

**Additional context**

The alternate `pip3 install ruamel-yaml` works (version '0.16.10') ...

To resolve, a minimum version for `ruamel.yaml` appears to be necessary.

bryanwweber:
To run the tests of the converter scripts, ruamel.yaml must be installed
with a minumum version at least 0.15.94. We aren't sure what the true
minimum version is, but 0.15.94 was released in April 2019 so is
probably widely released enough to count. Note that the version
available from the Ubuntu 16.04 repositories is not new

Fixes #831

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1233,6 +1233,21 @@ python_min_version = LooseVersion('3.5')
 cython_min_version = LooseVersion('0.23')
 numpy_min_test_version = LooseVersion('1.8.1')
 
+# We choose ruamel.yaml 0.15.34 as the minimum version
+# since it is the highest version available in the Ubuntu
+# 18.04 repositories and seems to work. Older versions such as
+# 0.13.14 on CentOS7 and 0.10.23 on Ubuntu 16.04 raise an exception
+# that they are missing the RoundTripRepresenter
+ruamel_min_version = LooseVersion('0.15.34')
+
+# Check for the minimum ruamel.yaml version, 0.15.34, at install and test
+# time. The check happens at install and test time because ruamel.yaml is
+# only required to run the Python interface, not to build it.
+check_for_ruamel_yaml = any(
+    target in COMMAND_LINE_TARGETS
+    for target in ["install", "test", "test-python-convert"]
+)
+
 # Handle the version-specific Python package options
 python_options = (('package', 'default'),
                   ('cmd', sys.executable),
@@ -1281,6 +1296,25 @@ if env['python_package'] != 'none':
         if err:
             print(err)
     """)
+    expected_output_lines = 3
+    if check_for_ruamel_yaml:
+        ru_script = textwrap.dedent("""\
+            try:
+                import ruamel_yaml as yaml
+                print(yaml.__version__)
+            except ImportError as ru_err:
+                try:
+                    from ruamel import yaml
+                    print(yaml.__version__)
+                except ImportError as ru_err_2:
+                    print('0.0.0')
+                    err += str(ru_err) + '\\n'
+                    err += str(ru_err_2) + '\\n'
+        """).splitlines()
+        s = script.splitlines()
+        s[-2:-2] = ru_script
+        script = "\n".join(s)
+        expected_output_lines = 4
 
     try:
         info = getCommandOutput(env['python_cmd'], '-c', script).splitlines()
@@ -1299,6 +1333,17 @@ if env['python_package'] != 'none':
         python_version = LooseVersion(info[0])
         numpy_version = LooseVersion(info[1])
         cython_version = LooseVersion(info[2])
+        if check_for_ruamel_yaml:
+            ruamel_yaml_version = LooseVersion(info[3])
+            if ruamel_yaml_version == LooseVersion("0.0.0"):
+                print("ERROR: ruamel.yaml was not found. {} or newer is "
+                    "required".format(ruamel_min_version))
+                sys.exit(1)
+            elif ruamel_yaml_version < ruamel_min_version:
+                print("ERROR: ruamel.yaml is an incompatible version: Found "
+                    "{}, but {} or newer is required.".format(
+                        ruamel_yaml_version, ruamel_min_version))
+                sys.exit(1)
 
     if warn_no_python:
         if env['python_package'] == 'default':
@@ -1313,9 +1358,10 @@ if env['python_package'] != 'none':
         print('INFO: Building the minimal Python package for Python {}'.format(python_version))
     else:
         warn_no_full_package = False
-        if len(info) > 3:
-            print("WARNING: Unexpected output while checking Python / Numpy / Cython versions:")
-            print('| ' + '\n| '.join(info[3:]))
+        if len(info) > expected_output_lines:
+            print("WARNING: Unexpected output while checking Python "
+                  "dependency versions:")
+            print('| ' + '\n| '.join(info[expected_output_lines:]))
 
         if python_version < python_min_version:
             print("WARNING: Python version is incompatible with the full Python module: "
--- a/SConstruct
+++ b/SConstruct
@@ -1345,6 +1345,11 @@ if env['python_package'] != 'none':
                         ruamel_yaml_version, ruamel_min_version))
                 sys.exit(1)
 
+        if len(info) > expected_output_lines:
+            print("WARNING: Unexpected output while checking Python "
+                  "dependency versions:")
+            print('| ' + '\n| '.join(info[expected_output_lines:]))
+
     if warn_no_python:
         if env['python_package'] == 'default':
             print('WARNING: Not building the Python package because the Python '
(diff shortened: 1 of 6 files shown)


test/SConscript
```diff
@@ -268,6 +268,33 @@ test_root = '#interfaces/cython/cantera/test'
 for f in mglob(localenv, test_root, '^test_*.py'):
     python_subtests.append(f.name[5:-3])
 
+if localenv['python_package'] != 'none':
+    # We choose ruamel.yaml 0.15.94 as the minimum version,
```
ischoegl: 0.15.87 ... ?

ischoegl: Actually
```
 ruamel_yaml-0.15.80        |py37h8f50634_1001         267 KB  conda-forge
```
on several of the runners.

ischoegl: Update: Ubuntu 18.04 uses 0.15.34 [see link](https://packages.ubuntu.com/source/bionic/ruamel.yaml), and I recall that it worked out of the box. Same minimum version will work for [debian](https://packages.debian.org/search?keywords=python-ruamel.yaml). CentOS7 is older (0.13.14, although CentOS8 has 0.15.41) and like Xenial (0.10.23) throws the error.

bryanwweber: Ok, thank you for helping triangulate this!

speth (commented):
I'm wondering if it would make more sense to do this check in `SConstruct`, at the same time we do the checks for Numpy and Cython. Otherwise, we create the scenario where the user just runs `scons build && scons install`, never triggering this check and ending up with a broken Cantera installation.

bryanwweber:
The reservation I have about putting this check in `SConstruct` instead of `test/SConscript` is that `ruamel.yaml` is not a build requirement, it's a runtime requirement. In the case you suggest, the user would only hit an error message when they tried to convert a mechanism, and the fix would be that `ruamel.yaml` isn't installed (or the version is too old). They could perfectly well use the library with any existing YAML file with no problems. Another case where I don't want to introduce another build-time requirement is for packaging.

Maybe if we put the check in `SConstruct`, it could be a warning for the build/install stage and an error during the test stage?

speth:
Hm, I didn't realize that we didn't run the tests in the packaging environments. But is there any environment we have set up where installing ruamel.yaml is any more difficult than installing Numpy or Cython? I feel like people usually ignore warnings that pop up in `scons build` (which isn't unreasonable in 10000 lines of output).

Despite what I said earlier, maybe we should check for the minimum ruamel.yaml version in the conversion scripts, given that we know that there are still-supported operating systems which provide older versions. And it would be nice to provide a more helpful error message than `AttributeError: module 'ruamel.yaml' has no attribute 'RoundTripRepresenter'`.

At that point, checking for ruamel.yaml at the `scons test` stage isn't much different from just letting those particular tests fail. I think the point of checking at the start of `scons build` is to fail fast if possible.

bryanwweber:
Thanks for the thoughts @speth. I implemented the version check in the `*2yaml.py` scripts for now, to improve the error message as you say.

> Hm, I didn't realize that we didn't run the tests in the packaging environments.

It depends on the package, but for instance, for the macOS MATLAB package, we don't build the Python module at all, so there's no reason to have `ruamel.yaml` installed.

> But is there any environment we have set up where installing ruamel.yaml is any more difficult than installing Numpy or Cython?

Not particularly, at least, not that I can think of. Apologies for being dense, I probably should have caught this before, but are you suggesting that the check for `ruamel.yaml` only be done if `python_package != 'none'`? I have a few conflicting thoughts here, let me write them all out and see if things start to clarify.

First, think that conditioning the check on actually building the Python interface could make sense. On the other hand, I have a gut feeling that we shouldn't restrict the user's environment unless it's really necessary.

What if we check for a good version if a user tries to install the Python interface? A user running `scons install` will presumably want to actually use the interface, so that'd be a good time to check without restricting the `build` step. And fixing the error wouldn't require rebuilding anything, so it wouldn't be a case where the user would save time by seeing the error in the `build` step.

bryanwweber:
OK, I implemented a check of `ruamel.yaml` at install time. I also slightly redid the check for Python version to make the required version for both the full and minimal interfaces the same. I'm not sure about this change, because it's possible in the future that the minimum version for the minimal interface could be lower than for the full interface (which is how it is implemented right now). So thoughts about that change would be appreciated, and its in its own commit so can be dropped easily.

SConstruct
```diff
@@ -1233,6 +1233,13 @@ python_min_version = LooseVersion('3.5')
 cython_min_version = LooseVersion('0.23')
 numpy_min_test_version = LooseVersion('1.8.1')
 
+# We choose ruamel.yaml 0.15.34 as the minimum version
+# since it is the highest version available in the Ubuntu
+# 18.04 repositories and seems to work. Older versions such as
+# 0.13.14 on CentOS and 0.10.23 on Ubuntu 16.04 raise an exception
```
ischoegl: This is actually only for CentOS7. CentOS8 has 0.15.41 and will work.

speth:
> are you suggesting that the check for `ruamel.yaml` only be done if `python_package != 'none'`? I have a few conflicting thoughts here, let me write them all out and see if things start to clarify.

Yes, that was my thought. If you aren't installing at least the "minimal" Python module, it doesn't matter whether or not you have `ruamel.yaml` installed.

> First, think that conditioning the check on actually building the Python interface could make sense. On the other hand, I have a gut feeling that we shouldn't restrict the user's environment unless it's really necessary.

I'm confused - what does "restricting the user's environment" mean?
 
> What if we check for a good version if a user tries to install the Python interface? A user running `scons install` will presumably want to actually use the interface, so that'd be a good time to check without restricting the `build` step. And fixing the error wouldn't require rebuilding anything, so it wouldn't be a case where the user would save time by seeing the error in the `build` step.

Checking before `install` or `test` makes sense. Currently, you're only doing it before `install`, right?

bryanwweber:
> what does "restricting the user's environment" mean?

Requiring them to install `ruamel.yaml` when they don't need it.

> Checking before `install` or `test` makes sense. Currently, you're only doing it before install, right?

In `SConstruct`, the only check is before `install`. The test phase is de facto handled by the checks in the scripts themselves. It would be possible to expand the `SConstruct` check to include the test phase, but that would preclude running subsets of the test suite (for example, `scons test-thermo`) if the Python interface had been built - perhaps a minor consideration, but since it isn't necessary to check at test time in `SConstruct` given the check in the script, I don't think its necessary to add it.

speth:
If I'm running `scons test`, I'd rather it fail immediately than run for several minutes before telling me that I need to install `ruamel.yaml`. What about writing something like:
```py
check_for_ruamel_yaml = any(target in COMMAND_LINE_TARGETS for target in ['install', 'test', 'test-python-convert'])
```
and then using this variable in the various `if` statements? I think this catches all of the cases where you would need `ruamel.yaml`.

SConstruct
bryanwweber: Thanks for the note! I added it in the most recent update.

bryanwweber:
@speth I implemented that suggestion, thanks. Any thoughts about synchronizing the minimum Python versions?

ischoegl:
FWIW, I think every time SCons can be simplified, it should be. In that vein, I’d argue that keeping Python versions synchronized is a plus.

SConstruct
```diff
@@ -1299,6 +1331,26 @@ if env['python_package'] != 'none':
...
+                        ruamel_yaml_version, ruamel_min_version))
+                sys.exit(1)
+
+        if len(info) > expected_output_lines:
+            msg = "WARNING: Unexpected output while checking Python / NumPy / Cython versions:"
+            if check_for_ruamel_yaml:
+                msg = msg.split()
+                msg.insert(-1, "/ ruamel.yaml")
+                msg = " ".join(msg)
+            print(msg)
```
speth: This might be a bit excessive -- we could just go with
```py
print("WARNING: Unexpected output while checking Python dependency versions")
```

SConstruct
```diff
@@ -1229,9 +1229,22 @@ env['python_cmd_esc'] = quoted(env['python_cmd'])
...
+# We choose ruamel.yaml 0.15.34 as the minimum version
+# since it is the highest version available in the Ubuntu
+# 18.04 repositories and seems to work. Older versions such as
+# 0.13.14 on CentOS7 and 0.10.23 on Ubuntu 16.04 raise an exception
+# that they are missing the RoundTripRepresenter
+ruamel_min_version = LooseVersion('0.15.34')
+check_for_ruamel_yaml = any(
+    target in COMMAND_LINE_TARGETS
+    for target in ["install", "test", "test-python-convert"]
+)
```
speth, line 1251: I think the explanation for why we make this more complex and don't always check the `ruamel.yaml` version needs to be documented here.

interfaces/cython/cantera/ck2yaml.py
```diff
@@ -59,6 +57,21 @@
...
+yaml_version = yaml.version_info
+# We choose ruamel.yaml 0.15.34 as the minimum version
+# since it is the highest version available in the Ubuntu
+# 18.04 repositories and seems to work. Older versions such as
+# 0.13.14 on CentOS7 and 0.10.23 on Ubuntu 16.04 raise an exception
+# that they are missing the RoundTripRepresenter
+yaml_min_version = (0, 15, 34)
+if yaml_version < yaml_min_version:
+    raise RuntimeError(
+        "The minimum version of ruamel.yaml is 0.15.34. If you installed "
```
speth:
```suggestion
        "The minimum supported version of ruamel.yaml is 0.15.34. If you installed "
```
And ditto in the other scripts.

speth (commented):
I think this looks good, generally. Just had a couple of minor suggestions.

bryanwweber:
@speth Thanks for the comments! Updated now.

speth (approved):
Thanks! This looks good now.

Issue #877, opened 2020-06-21
Broken surf_pfr.py example


ischoegl:
**System information**

- Cantera version: development version
- OS: Ubuntu 18.04
- Python/MATLAB version: 3.6.9

**Expected behavior**

Example runs without error.

**Actual behavior**

Example errors with
```
In [1]: %run surf_pfr.py
    distance       X_CH4        X_H2        X_CO
---------------------------------------------------------------------------
CanteraError                              Traceback (most recent call last)
/src/interfaces/cython/cantera/examples/reactors/surf_pfr.py in <module>()
    108     upstream.syncState()
    109     sim.reinitialize()
--> 110     sim.advance_to_steady_state()
    111     dist = n * rlen * 1.0e3  # distance in mm
    112 

interfaces/cython/cantera/reactor.pyx in cantera._cantera.ReactorNet.advance_to_steady_state()

interfaces/cython/cantera/reactor.pyx in cantera._cantera.ReactorNet.step()

CanteraError: 
***********************************************************************
CanteraError thrown by Phase::setDensity:
density must be positive. density = -1.3738258649251213e-27
***********************************************************************
```

**To Reproduce**

See above.

**Additional context**

~Error potentially introduced in the context of #720~

A switch back from YAML to CTI input removes the issue, i.e. the error appears to be due to a bug in YAML conversion or parsing, see #902.

ischoegl:
Closing this issue as it is superseded by #902.

PR #876, opened 2020-06-21
Minor reactor example updates

ischoegl:
**Checklist**

- [x] There is a clear use-case for this code change
- [x] The commit message has a short title & references relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] The pull request is ready for review

**If applicable, fill in the issue number this pull request is fixing**

Fixes #875

**Changes proposed in this pull request**

- Use recently implemented methods to simplify examples
- Small fixes

**PS:** The broken AppVeyor build appears to be unrelated to changes in this PR, see #878.

Issue #875, opened 2020-06-21
Broken piston.py example


ischoegl:
**System information**

- Cantera version: development version
- OS: ubuntu 18.04
- Python/MATLAB version: 3.6.9

**Expected behavior**

Example runs without errors

**Actual behavior**

Example errors with
```
In [14]: %run piston.py
time [s]    T1 [K]      T2 [K]      V1 [m^3]    V2 [m^3]    V1+V2 [m^3]  X(CO)     
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/src/interfaces/cython/cantera/examples/reactors/piston.py in <module>()
     54 net = ct.ReactorNet([r1, r2])
     55 
---> 56 states1 = ct.SolutionArray(r1.thermo, extra=['t', 'v'])
     57 states2 = ct.SolutionArray(r2.thermo, extra=['t', 'v'])
     58 

/usr/local/lib/python3.6/dist-packages/cantera/composite.py in __init__(self, phase, shape, states, extra, meta)
    548                     raise ValueError(
    549                         "Unable to create extra column '{}': name is already "
--> 550                         "used by SolutionArray objects.".format(name))
    551                 self._extra[name] = []
    552 

ValueError: Unable to create extra column 'v': name is already used by SolutionArray objects.
```

**To Reproduce**

Steps to reproduce the behavior: see above

**Additional context**

The extra column `v` masks a pre-existing thermo state, which previously did not raise an error. The fix is trivial.

PR #892, opened 2020-07-03
[CI] Run python examples from scons

### LINKED ISSUE #60: Run Python examples during CI to catch breakages
**Abstract**

Work on 2.5 resulted in multiple Python examples being broken. It would be of value to add running of Python examples during CI (see also #37) to avoid inadvertent breakage or deprecation of functionality.

**Possible Solutions**

Simply running examples while filtering `DeprecationWarnings` as errors will reveal most issues (see closed PR Cantera/cantera#892). An alternative would be to build full regression tests comparing output to validated results.

**References**

Improvements to CI: #37
Limited implementation: Cantera/cantera#892
Regression tests for C++ Cantera/cantera#883 and FORTRAN Cantera/cantera#874

ischoegl:

- Add running of Python examples to CI

Fixes Cantera/enhancements#60
Addresses Cantera/enhancements#37

**Checklist**


bryanwweber:
Is this superseded by https://github.com/Cantera/cantera/pull/883#issuecomment-653606768 ?

ischoegl:
@bryanwweber ... I don't think so, as this is (thus far) for Python. 

Edit: I haven't looked into running (as opposed to building) of cxx yet.

bryanwweber:
@ischoegl Ray's work in #883 runs the C++ samples and I think #874 did Fortran, so no need to worry there. We usually refer to the Python code as examples rather than samples (hence my earlier confusion), do you mind renaming this?

ischoegl:
> We usually refer to the Python code as examples rather than samples (hence my earlier confusion), do you mind renaming this?

No problem.

ischoegl:
@bryanwweber ... ~will stop here.~ Ubuntu works, and there are some environmental variables left on windows that I need to work out. OSX runners currently won’t leave the queue, so I disabled the new runs. `matplotlib` causes some issues for sundials and Xenial runners, so those won’t work.

**PS:** `matplotlib` causes issues on OSX runners also
```
ImportError: Python is not installed as a framework. The Mac OS X backend will not be able to 
function correctly if Python is not installed as a framework. See the Python documentation for 
more information on installing Python as a framework on Mac OS X. Please either reinstall 
Python as a framework, or try one of the other backends. If you are using (Ana)Conda please 
install python.app and replace the use of 'python' with 'pythonw'. See 'Working with Matplotlib 
on OSX' in the Matplotlib FAQ for more information.
```

ischoegl:
Alright. At this point I believe I have come to a stopping point. 

* `check-examples` works on Linux and Windows; I cannot test on OSX (**Edit:** the CI runner finally works, so it presumably is ok)
* `matplotlib` (which is used by many examples) causes issues on some CI environments (e.g. version requirements)
* CI runner works on Ubuntu, Windows, and macOS, as long as the `Agg` backend for `matplotlib` is used (this is essential for Windows and macOS)
* ~`matplotlib` isn't compatible with the OSX CI environment~ **Edit:** this is a Python 3.5 issue

**Caveats:** `scons check-examples` runs in the build folder, and some outputs will get copied during installation, which I'm not sure is an issue. Further, examples with known issues are skipped.

ischoegl:
:tada: ... finally the CI runners work on all OS's - it was mainly a function of `matplotlib` backends, and avoiding Python 3.5.

speth:
Running the samples as part of the test process certainly makes sense. I was thinking about this myself as a follow on to #874 and #883. The approach that I had been considering was to run the Python examples as part of the Python test suite.

The advantage of that approach is that we would be able to use the `unittest` machinery to do some checks that the output of these samples is as expected. Perhaps not all of the detailed numerical value comparisons that we do for some other tests, but at least we could check that expected output files are generated, and that there are no deprecation warnings issued.

ischoegl:
@speth ... thanks for the feedback.

I think the important thing is that examples are run, which is an infinite improvement over the status quo where examples can break without notice, despite being careful when closing out a PR. Going back to fix issues several months later is a painful exercise (thinking about #877 here; #875 was thankfully trivial, and the others I don't think were due to my work).

From a philosophical viewpoint, I view unit tests as sanity checks on small test problems, whereas examples are coding examples that are provided to end users. The handful of python examples currently take up about the same amount of run time as the entire test suite, so it's not particularly efficient to automatically add the examples. One other caveat is that writing examples that include regression tests will require a lot more developer input as it's not as straight forward as simply contributing a script.

Regarding my proposed implementation, I think that its simplicity is a feature: it's clear and simple to maintain (`scons` by itself is not for the faint of heart). I really think that just checking that examples don't break during modifications is a beyond 90% solution. Getting to a higher percentage will complicate things a lot more, and I am not sure that it is worth the effort. I honestly believe that beefing up on unit tests instead is preferable (Edit: probably there should be more that compare to a reference/'blessed' solution).

~Ad warnings: I agree on `DeprecationWarnings` (or warnings in general). In this regard, however, not all are caught by Python, as warnings raised in C++ may 'escape'.  One option would be to ` | tee log`  terminal output to a file, and simply search for deprecations. I also looked into the `PYTHONWARNINGS=error` environment variable, but it's failing due to an `ImportWarning` right from the get-go (a known issue caused by Cython); otherwise making sure that no warnings are raised would be an option.~ Edit: found a solution - see below.

ischoegl:
**PS:** I added the option `-W error::DeprecationWarning:` which will throw errors. (Edit: already found a `DeprecationWarning`, but still need to circumnavigate the Windows issue with an imported library throwing a DeprecationWarning).

ischoegl:
@speth and @bryanwweber ... with the latest force-push, the `DeprecationWarning`s are now raised as errors on all OS's (had to filter out an `imp` related warning that is specific to Windows). I could also include an additional test (and remove one redundant commit) once #876 gets merged.

Any feedback is appreciated at this point; if you'd rather pursue a different route, I'll simply close this PR.

ischoegl:
Closing. While it was interesting to figure out `scons` for some simple checks, adding more comprehensive regression tests is not my cup of tea. I added fixes that deserve to survive to #876 and #899.

ischoegl:
Reopening as an alternative to the CI runners proposed in #905 - while this implementation is still a stop-gap, it runs examples on all OS's, can be invoked via `scons`, and checks for `DeprecationWarning`'s.

The PR needs to be rebased once #899 is resolved.

ischoegl:
Closed again as there appears to be strong opposition to this suggestion.

PR #897, opened 2020-07-09
[Input] Improve handling of YAML null values

speth:

- YAML documents containing null values no longer generate `bad conversion` errors while parsing.
- Modify error messages generated by null values to indicate that the error can be caused by either a missing key or a key with no value.

**Checklist**

### DIFF
--- a/test/general/test_containers.cpp
+++ b/test/general/test_containers.cpp
@@ -292,6 +292,23 @@ TEST(AnyMap, iterators)
     EXPECT_TRUE(std::find(keys.begin(), keys.end(), "bar") != keys.end());
 }
 
+TEST(AnyMap, null_values)
+{
+    AnyMap m = AnyMap::fromYamlString(
+        "{a: 1, b: ~, c: , d: 5}"
+    );
+    EXPECT_EQ(m.size(), (size_t) 4);
+    EXPECT_TRUE(m.at("c").is<void>());
+
+    try {
+        m.at("b").asString();
+        FAIL();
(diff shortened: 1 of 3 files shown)


test/general/test_containers.cpp
```diff
@@ -292,6 +292,23 @@ TEST(AnyMap, iterators)
...
+    );
+    EXPECT_EQ(m.size(), (size_t) 4);
+    EXPECT_TRUE(m.at("c").is<void>());
+
+    try {
+        m.at("b").asString();
+        FAIL();
+    } catch (CanteraError& err) {
+        EXPECT_THAT(err.what(),
+                    testing::HasSubstr("Key 'b' not found or contains no value"));
```
bryanwweber, line 308: I see that the tests pass, but I'm surprised this isn't raised for key `'c'`. Is that a limitation of `yaml-cpp`?

speth: `yaml-cpp` correctly converts both the blank value and `~` to null. The difference here is that the exception is raised by `asString()`, not `at("b")`. I guess in this sense, the `AnyMap` class isn't quite a faithful representation of YAML in that in `AnyMap`, a key with a null value is equivalent to the key not being present, while YAML makes a distinction. Since we don't use null values anywhere in the Cantera YAML format, I'm not sure whether it's worth the extra effort.

bryanwweber: Makes sense, thanks!

bryanwweber (approved):

PR #899, opened 2020-07-10
Minor fixes in AnyValue and scons

ischoegl:

- ensure that a YAML field: `some-field: '12345'` is read as a string by the C++ parser (which currently returns an integer)
- make C++ parser consistent with `ruamel.yaml`
- Also: minor formatting update in scons output (mixed windows/linux style `os.path.sep`s in install message on windows)

This is a minor fix that is left over after closing #881. Also cherry-picking minor fixes from #892 to ensure that fixes of existing features survive.

**Checklist**

### DIFF
--- a/src/base/AnyMap.cpp
+++ b/src/base/AnyMap.cpp
@@ -190,7 +190,12 @@ struct convert<Cantera::AnyValue> {
         if (node.IsScalar()) {
             // Scalar nodes are int, doubles, or strings
             std::string nodestr = node.as<std::string>();
-            if (isInt(nodestr)) {
+            if (node.Tag() == "!") {
+                // Prevent quoted strings from being implicitly converted to
+                // numeric types, e.g. the quoted YAML string '12345' should not
+                // be interpreted as an integer
+                target = nodestr;
+            } else if (isInt(nodestr)) {
                 try {
                     target = node.as<long int>();
                 } catch (YAML::BadConversion&) {
--- a/SConstruct
+++ b/SConstruct
@@ -1410,6 +1410,7 @@ env['debian'] = any(name.endswith('dist-packages') for name in sys.path)
 
 # Directories where things will be after actually being installed. These
 # variables are the ones that are used to populate header files, scripts, etc.
+env['prefix'] = os.path.normpath(env['prefix'])
 env['ct_installroot'] = env['prefix']
 env['ct_libdir'] = pjoin(env['prefix'], env['libdirname'])
 env['ct_bindir'] = pjoin(env['prefix'], 'bin')
@@ -1781,15 +1782,18 @@ def postInstallMessage(target, source, env):
         """.format(**env_dict))
 
     if os.name != 'nt':
+        env['setup_cantera'] = pjoin(env['ct_bindir'], 'setup_cantera')
+        env['setup_cantera_csh'] = pjoin(env['ct_bindir'], 'setup_cantera.csh')
         install_message += textwrap.dedent("""
+
             Setup scripts to configure the environment for Cantera are at:
 
-              setup script (bash)         {ct_bindir!s}/setup_cantera
-              setup script (csh/tcsh)     {ct_bindir!s}/setup_cantera.csh
+              setup script (bash)         {setup_cantera!s}
+              setup script (csh/tcsh)     {setup_cantera_csh!s}
 
             It is recommended that you run the script for your shell by typing:
 
-              source {ct_bindir!s}/setup_cantera
+              source {setup_cantera!s}
 
             before using Cantera, or else include its contents in your shell login script.
         """.format(**env_dict))
--- a/interfaces/cython/SConscript
+++ b/interfaces/cython/SConscript
@@ -83,14 +83,13 @@ localenv.Depends(ext, localenv['cantera_staticlib'])
 
 for f in (mglob(localenv, 'cantera', 'py') +
           mglob(localenv, 'cantera/test', 'py') +
-          mglob(localenv, 'cantera/examples/tutorial', 'py') +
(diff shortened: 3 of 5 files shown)


src/base/AnyMap.cpp
```diff
@@ -190,7 +190,11 @@ struct convert<Cantera::AnyValue> {
         if (node.IsScalar()) {
             // Scalar nodes are int, doubles, or strings
             std::string nodestr = node.as<std::string>();
-            if (isInt(nodestr)) {
+            if (node.Tag() == "!") {
+                // Undo conversion from string to other formats (done
+                // automatically by yaml-cpp, e.g. '12345' is converted to int)
+                target = nodestr;
+            } else if (isInt(nodestr)) {
```
bryanwweber: I believe this change will allow resolution of this comment in `ctml2yaml.py`: https://github.com/Cantera/cantera/blob/773523a4e314e4ce4bfec379b438c5ca55cebf5c/interfaces/cython/cantera/ctml2yaml.py#L1947-L1950 🎉  (I'm not suggesting that be added to this PR)

ischoegl: Ok. Happy that it may help elsewhere!

src/base/AnyMap.cpp
```diff
@@ -190,7 +190,11 @@ struct convert<Cantera::AnyValue> {
         if (node.IsScalar()) {
             // Scalar nodes are int, doubles, or strings
             std::string nodestr = node.as<std::string>();
-            if (isInt(nodestr)) {
+            if (node.Tag() == "!") {
+                // Undo conversion from string to other formats (done
+                // automatically by yaml-cpp, e.g. '12345' is converted to int)
```
speth: I think the fix you've added here, at least as far as I understand `yaml-cpp`, makes sense, but the comment is not correct -- the automatic conversions are being done here in this method, based on whether the string value looks like an integer or floating point value.

speth (commented):
Looks basically fine to me. Could you update the commit "[scons] Small updates/fixes" to have a more descriptive message? It's not immediately obvious what the purpose of these changes is.

src/base/AnyMap.cpp
ischoegl: `yaml-cpp` erroneously strips the `'` from the string (which causes the issue), so yes, this needs updating.

ischoegl:
Sorry for pushing again, but I noticed that the note in `ch4_ion.yaml` was not indented correctly (!). I stumbled over this when I tried to verify that the missing quotes I corrected here weren't an artifact produced by `cti2yaml` and/or `ctml2yaml` as the note happens to be a valid numeric value (result: both converters create the quotes, and indent correctly).

ischoegl:
Will probably add another update for the formatting, as I just found a snippet for Anaconda on windows documenting the status before this PR
```
[...]

File locations:

  applications                C:/Users/<user>/cantera\bin
  library files               C:/Users/<user>/cantera\lib
  C++ headers                 C:/Users/<user>/cantera\include
  samples                     C:/Users/<user>/cantera\samples
  data files                  C:/Users/<user>/cantera\data
  Python package (cantera)    C:\Users\<user>\anaconda3\envs\cantera-dev\Lib\site-packages
  Python samples              C:\Users\<user>\anaconda3\envs\cantera-dev\Lib\site-packages\cantera\examples
  Matlab toolbox              C:/Users/<user>/cantera\matlab\toolbox
  Matlab samples              C:/Users/<user>/cantera\samples\matlab

An m-file to set the correct matlab path for Cantera is at:

  C:/Users/<user>/cantera\matlab\toolbox\ctpath.m

scons: done building targets.
```
All this takes is a couple of `os.path.normpath`s ...

ischoegl:
:tada: ... better:
```
[...]

File locations:

  applications                C:\Users\ischo\cantera\bin
  library files               C:\Users\ischo\cantera\lib
  C++ headers                 C:\Users\ischo\cantera\include
  samples                     C:\Users\ischo\cantera\samples
  data files                  C:\Users\ischo\cantera\data
  Python package (cantera)    C:\Users\ischo\miniconda3\envs\cantera-dev\Lib\site-packages
  Python samples              C:\Users\ischo\miniconda3\envs\cantera-dev\Lib\site-packages\cantera\examples
  Matlab toolbox              C:\Users\ischo\cantera\matlab\toolbox
  Matlab samples              C:\Users\ischo\cantera\samples\matlab

An m-file to set the correct matlab path for Cantera is at:

  C:\Users\ischo\cantera\matlab\toolbox\ctpath.m

scons: done building targets.
```

src/base/AnyMap.cpp
speth: I don't think it's correct to say that what `yaml-cpp` is doing is erroneous. If I have the YAML `{a: 12345, b: '12345', c: hello, d: 'hello'}`, then the string values of (a) and (b) should be equal, the same as the string values of (c) and (d) are equal. The only distinction is that (b) and (d) were indicated as being strings by being quoted, which is captured in the value of the "tag". The comment I would suggest is just "Prevent quoted strings from being implicitly converted to numeric types".

ischoegl: Correct; shouldn’t have inferred as much as I did from `ruamel.yaml`; `yaml-cpp` clearly works at a lower level. Resolving as
```
                // Prevent quoted strings from being implicitly converted to
                // numeric types, e.g. the quoted YAML string '12345' should not
                // be interpreted as an integer
```

ischoegl:
@speth ... I think I took care of the requested change of the comment. If it’s adequate, I think this should be ready to merge.

speth (approved):
Looks good. Sorry for the delayed review.

ischoegl:
NP - I figure you took a well-deserved break :wink:

PR #900, opened 2020-07-10
Support file I/O for string columns in SolutionArray

### LINKED ISSUE #896: What to do about `str` columns in `SolutionArray`
**Problem description**

The current implementation of `SolutionArray` allows for `extra` columns containing `str` (see discussion in #838), which may or may not be an intended use case: built-in attributes are almost exclusively numeric, with the notable exception of `state_of_matter`. Other implementations did not anticipate non-numeric data, e.g. CSV import and HDF export/import, so `str` support either needs to be deprecated/disabled or consistently implemented.

One interpretation is that `str` input was never intended, but also not explicitly checked for. I.e. it just happens to be supported by `numpy` (similar to sequences not being checked, see #895). In this case, the most appropriate resolution may be to catch and deprecate `str` columns, while warning that it is not fully supported for CSV/HDF.

**Steps to reproduce**

```
In [1]: import cantera as ct
    ...: arr = ct.SolutionArray(gas, 3, extra={'spam': 'eggs'})
    ...: 

In [2]: arr._extra
Out[2]: OrderedDict([('spam', ['eggs', 'eggs', 'eggs'])])

In [3]: !cat test.csv
spam,T,density,Y_H2,Y_H,Y_O,Y_O2,Y_OH,Y_H2O,Y_HO2,Y_H2O2,Y_AR
eggs,300.0,0.08189392763801234,1.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
eggs,300.0,0.08189392763801234,1.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
eggs,300.0,0.08189392763801234,1.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0

In [4]: arr2 = ct.SolutionArray(gas, extra={'spam'})

In [4]: arr2.read_csv('test.csv')

In [5]: arr2._extra
Out[5]: {'spam': array([ nan,  nan,  nan])}

In [6]: arr.write_hdf('test.h5')
<ipython-input-26-d708cd336db5> in <module>()
----> 1 arr.write_hdf('test.h5')

/usr/local/lib/python3.6/dist-packages/cantera/composite.py in write_hdf(self, filename, cols, group, subgroup, attrs, mode, append, compression, compression_opts, *args, **kwargs)
   1117                 dgroup.attrs[key] = val
   1118             for header, col in data.items():
-> 1119                 dgroup.create_dataset(header, data=col, **hdf_kwargs)
   1120 
   1121         return group

/home/docker/.local/lib/python3.6/site-packages/h5py/_hl/group.py in create_dataset(self, name, shape, dtype, data, **kwds)
    134 
    135         with phil:
--> 136             dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
    137             dset = dataset.Dataset(dsid)
    138             if name is not None:

/home/docker/.local/lib/python3.6/site-packages/h5py/_hl/dataset.py in make_new_dset(parent, shape, dtype, data, chunks, compression, shuffle, fletcher32, maxshape, compression_opts, fillvalue, scaleoffset, track_times, external, track_order, dcpl)
    116         else:
    117             dtype = numpy.dtype(dtype)
--> 118         tid = h5t.py_create(dtype, logical=1)
    119 
    120     # Legacy

h5py/h5t.pyx in h5py.h5t.py_create()

h5py/h5t.pyx in h5py.h5t.py_create()

h5py/h5t.pyx in h5py.h5t.py_create()

TypeError: No conversion path for dtype: dtype('<U4')

In [7]: arr.spam
Out[7]: 
array(['eggs', 'eggs', 'eggs'],
      dtype='<U4')
```

**Behavior**

The pre-existing `write_csv` supports `str`, the newly introduced `read_csv` fails to import it, and `h5py` has issues. 

I am filing this as a bug report as this presumably needs to be fixed prior to the release of 2.5.

**System information**

- Cantera version: 2.5.a4
- OS: Ubuntu 18.04
- Python 3.6

**Additional context**

#838, #895

ischoegl:

- Enable CSV import of string columns
- Enable HDF export/import of string columns
- Add `phase_of_matter` to `SolutionArray`

Fixes #896

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -963,8 +963,7 @@ class SolutionArray:
             data_dict = OrderedDict()
             for label in data.dtype.names:
                 if data[label].dtype.type == np.bytes_:
-                    dtype = '<U' + str(data[label].dtype)[2:]
-                    data_dict[label] = data[label].astype(dtype)
+                    data_dict[label] = data[label].astype('U')
                 else:
                     data_dict[label] = data[label]
         else:
@@ -1127,8 +1126,7 @@ class SolutionArray:
                 dgroup.attrs[key] = val
             for header, value in data.items():
                 if value.dtype.type == np.str_:
-                    dtype = '|S' + str(value.dtype)[2:]
-                    dgroup.create_dataset(header, data=value.astype(dtype),
+                    dgroup.create_dataset(header, data=value.astype('S'),
                                           **hdf_kwargs)
                 else:
                     dgroup.create_dataset(header, data=value, **hdf_kwargs)
@@ -1222,8 +1220,7 @@ class SolutionArray:
                 if name == 'phase':
                     continue
                 elif value.dtype.type == np.bytes_:
-                    dtype = '<U' + str(value.dtype)[2:]
-                    data[name] = np.array(value).astype(dtype)
+                    data[name] = np.array(value).astype('U')
                 else:
                     data[name] = np.array(value)
 
@@ -1327,7 +1324,10 @@ def _make_functions():
         return np.empty(self._shape)
 
     def empty_strings(self):
-        return np.empty(self._shape, dtype='<U50')
+        # The maximum length of strings assigned by built-in methods is
+        # currently limited to 50 characters; an attempt to assign longer
+        # character arrays will result in truncated strings.
+        return np.empty(self._shape, dtype='U50')
 
     def empty_species(self):
         return np.empty(self._shape + (self._phase.n_selected_species,))
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -963,8 +963,7 @@ class SolutionArray:
             data_dict = OrderedDict()
             for label in data.dtype.names:
                 if data[label].dtype.type == np.bytes_:
-                    dtype = '{}'.format(data[label].dtype)
-                    dtype = dtype.replace('S', 'U').replace('|', '<')
+                    dtype = '<U' + data[label].dtype.__str__()[2:]
                     data_dict[label] = data[label].astype(dtype)
                 else:
                     data_dict[label] = data[label]
@@ -973,7 +972,7 @@ class SolutionArray:
             data = np.genfromtxt(filename, delimiter=',',
                                  dtype=None, names=True, encoding=None)
             data_dict = OrderedDict({label: data[label]
-                                    for label in data.dtype.names})
+                                     for label in data.dtype.names})
         self.restore_data(data_dict)
 
     def to_pandas(self, cols=None, *args, **kwargs):
@@ -1126,8 +1125,13 @@ class SolutionArray:
             # store SolutionArray data
             for key, val in self._meta.items():
                 dgroup.attrs[key] = val
-            for header, col in data.items():
-                dgroup.create_dataset(header, data=col, **hdf_kwargs)
+            for header, value in data.items():
+                if value.dtype.type == np.str_:
+                    dtype = '|S' + value.dtype.__str__()[2:]
+                    dgroup.create_dataset(header, data=value.astype(dtype),
+                                          **hdf_kwargs)
+                else:
+                    dgroup.create_dataset(header, data=value, **hdf_kwargs)
 
         return group
 
@@ -1215,7 +1219,12 @@ class SolutionArray:
             # load data
             data = OrderedDict()
             for name, value in dgroup.items():
-                if name != 'phase':
+                if name == 'phase':
+                    pass
+                elif value.dtype.type == np.bytes_:
+                    dtype = '<U' + value.dtype.__str__()[2:]
+                    data[name] = np.array(value).astype(dtype)
+                else:
                     data[name] = np.array(value)
 
         self.restore_data(data)
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -4,6 +4,7 @@
 from ._cantera import *
 import numpy as np
 from collections import OrderedDict
+from distutils import version
 import csv as _csv
 
 import pkg_resources
@@ -955,14 +956,24 @@ class SolutionArray:
         using `restore_data`. This method allows for recreation of data
         previously exported by `write_csv`.
         """
-        # read data block and header separately
-        data = np.genfromtxt(filename, skip_header=1, delimiter=',')
-        labels = np.genfromtxt(filename, max_rows=1, delimiter=',', dtype=str)
-
-        data_dict = OrderedDict()
-        for i, label in enumerate(labels):
-            data_dict[label] = data[:, i]
-
+        if version.LooseVersion(np.__version__) < version.LooseVersion("1.14"):
+            # bytestring needs to be converted for columns containing strings
+            data = np.genfromtxt(filename, delimiter=',',
+                                 dtype=None, names=True)
+            data_dict = OrderedDict()
+            for label in data.dtype.names:
+                if data[label].dtype.type == np.bytes_:
+                    dtype = '{}'.format(data[label].dtype)
+                    dtype = dtype.replace('S', 'U').replace('|', '<')
+                    data_dict[label] = data[label].astype(dtype)
+                else:
+                    data_dict[label] = data[label]
+        else:
+            # the 'encoding' parameter introduced with NumPy 1.14 simplifies import
+            data = np.genfromtxt(filename, delimiter=',',
+                                 dtype=None, names=True, encoding=None)
+            data_dict = OrderedDict({label: data[label]
+                                    for label in data.dtype.names})
         self.restore_data(data_dict)
 
     def to_pandas(self, cols=None, *args, **kwargs):
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -448,6 +448,7 @@ class SolutionArray:
         # From Transport
         'viscosity', 'electrical_conductivity', 'thermal_conductivity',
     ]
+    _strings = ['phase_of_matter']
     _n_species = [
         # from ThermoPhase
         'Y', 'X', 'concentrations', 'partial_molar_enthalpies',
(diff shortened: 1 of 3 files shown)


ischoegl:
@bryanwweber ... enabling strings wasn’t difficult after all. However, any comments (or better ideas) on the numpy version if/else condition here? (Alternative would be a less specific `try`/`except`)

 It’s pretty quiet these days otherwise: I hope that the work on beta is coming along.

interfaces/cython/cantera/composite.py
```diff
@@ -955,14 +957,23 @@ def read_csv(self, filename):
...
         """
-        # read data block and header separately
-        data = np.genfromtxt(filename, skip_header=1, delimiter=',')
-        labels = np.genfromtxt(filename, max_rows=1, delimiter=',', dtype=str)
-
-        data_dict = OrderedDict()
-        for i, label in enumerate(labels):
-            data_dict[label] = data[:, i]
-
+        if version.LooseVersion(np.__version__) < version.LooseVersion("1.14"):
```
bryanwweber: It looks like NumPy has the built-in capability to do this kind of comparison. I went searching because I was wondering if they had a tuple we could use instead of a string... anyhow, this should work:
```suggestion
        if np.lib.NumpyVersion(np.__version__) < "1.14.0":
```
and avoids the `distutils` import.

interfaces/cython/cantera/composite.py
```diff
@@ -955,14 +957,23 @@ def read_csv(self, filename):
...
+            # bytestring needs to be converted for columns containing strings
+            data = np.genfromtxt(filename, delimiter=',',
+                                 dtype=None, names=True)
+            data_dict = OrderedDict()
+            for label in data.dtype.names:
+                if data[label].dtype.type == np.bytes_:
+                    dtype = '<U' + data[label].dtype.__str__()[2:]
```
```suggestion
                    dtype = '<U' + str(data[label].dtype)[2:]
```
Just for clarity

interfaces/cython/cantera/composite.py
```diff
@@ -1204,7 +1220,12 @@ def strip_ext(source):
             # load data
             data = OrderedDict()
             for name, value in dgroup.items():
-                if name != 'phase':
+                if name == 'phase':
+                    pass
+                elif value.dtype.type == np.bytes_:
+                    dtype = '<U' + value.dtype.__str__()[2:]
```
```suggestion
                    dtype = '<U' + str(value.dtype)[2:]
```

interfaces/cython/cantera/composite.py
```diff
@@ -1204,7 +1220,12 @@ def strip_ext(source):
             # load data
             data = OrderedDict()
             for name, value in dgroup.items():
-                if name != 'phase':
+                if name == 'phase':
+                    pass
```
bryanwweber: I think it would be more typical to use `continue` here rather than `pass`. They're equivalent in this context, but if any code is added below the `else` but still inside the loop, that would be executed when it might not be expected to.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1787,6 +1787,16 @@ def test_purefluid(self):
...
 
+    def test_phase_of_matter(self):
+        water = ct.Water()
+        states = ct.SolutionArray(water, 5)
+        T = [300, 500, water.critical_temperature*2, 300]
+        P = [101325, 101325, 101325, water.critical_pressure*2]
+        states[:4].TP = T, P
+        states[4].TQ = 300, .4
+        pom = np.array(['liquid', 'gas', 'supercritical', 'supercritical', 'liquid-gas-mix'])
+        self.assertTrue((states.phase_of_matter == pom).all())
```
bryanwweber:
```suggestion
        pom = ['liquid', 'gas', 'supercritical', 'supercritical', 'liquid-gas-mix']
        self.assertEqual(list(states.phase_of_matter), pom)
```
This gives a much more useful failure message, showing the offending elements.

interfaces/cython/cantera/composite.py
```diff
@@ -1306,6 +1327,9 @@ def _make_functions():
     def empty_scalar(self):
         return np.empty(self._shape)
 
+    def empty_strings(self):
+        return np.empty(self._shape, dtype='<U14')
```
bryanwweber: With this value hardcoded to 14 characters, anything additional is truncated off the value. I'm not sure if there's a way to work around that, but this would be an extremely tricky bug to track down if we ever have strings longer than 14 characters for phases of matter. If there's not an easy way to resolve it, then maybe just setting it to a much larger number would be sufficient (say 50?).

I also wonder how this would affect `extra` columns that are longer than 14 characters?

interfaces/cython/cantera/composite.py
```diff
@@ -1115,8 +1126,13 @@ def write_hdf(self, filename, *args, cols=None, group=None, subgroup=None,
             # store SolutionArray data
             for key, val in self._meta.items():
                 dgroup.attrs[key] = val
-            for header, col in data.items():
-                dgroup.create_dataset(header, data=col, **hdf_kwargs)
+            for header, value in data.items():
+                if value.dtype.type == np.str_:
+                    dtype = '|S' + value.dtype.__str__()[2:]
```
```suggestion
                    dtype = '|S' + str(value.dtype)[2:]
```

bryanwweber (changes requested):
Thanks @ischoegl. In https://github.com/Cantera/cantera/commit/6edb133c89ab3d209d99fd4d3b7b63a3dce4f8c0 we set the minimum NumPy version to 1.12, since that's required for #838 to work properly. Ubuntu 18.04 provides 1.13.x in the system packages, and we didn't want to break compatibility with that version, so that's how I picked 1.12. Anyhow, I left a comment below about the version comparison.
Too many other things going on right now, with the semester approaching. It is my goal to get 2.5.0 out before the semester (i.e., before August 31) but... you know :smile:

interfaces/cython/cantera/composite.py
ischoegl: Was wondering about this myself (but settled on the longest string currently returned by phase-of-matter). I don’t think that this line affects `extra`, but it would affect potential future string attributes. I don’t have a problem making it longer ...

interfaces/cython/cantera/composite.py
ischoegl: Nice - I wasn’t aware of this!

interfaces/cython/cantera/composite.py
ischoegl: I increased the length to 50; there would a work-around suggested [here](https://stackoverflow.com/questions/9108837/numpy-recarray-strings-of-variable-length) - i.e. use an array of `object` type, rendering string length flexible - but that will open a can of worms elsewhere when it comes to saving, restoring, etc.. I have a strong preference for keeping this simple.

ischoegl:
@bryanwweber ... thanks for the review: your recommendations were useful as always and were easily addressed. I'm happy about the suggestion for `np.lib.NumpyVersion`, as I didn't like the `distutils` dependency (which I however still preferred to a `try` ... `except` structure).

Definitely looking forward to the release of 2.5.0 - I appreciate your and @speth's efforts on getting this out! I won't teach combustion again until next spring though :stuck_out_tongue_winking_eye: 

**PS:** failing tests are fixed by #905

interfaces/cython/cantera/composite.py
```diff
@@ -1306,6 +1326,9 @@ def _make_functions():
     def empty_scalar(self):
         return np.empty(self._shape)
 
+    def empty_strings(self):
+        return np.empty(self._shape, dtype='<U50')
```
bryanwweber: I looked into this a little more, what do you think of `dtype=object`? That uses a generic Python object as the element in the array, and as such has no limits on the string length. My understanding of the tradeoff is that there is additional memory overhead and processing required to use `object` instead of `U` type. However, in my limited testing, since we are over-provisioning the string, the memory use of `U50` is 25x that of `object` (!), and it takes no longer to create the arrays or assign states.
```python
import cantera as ct
import sys
w = ct.Water()
arr = ct.SolutionArray(w, int(5e5))
arr.TP = [300]*int(5e5), [101325]*int(5e5)  # This takes a while
sys.getsizeof(arr.phase_of_matter)
```
gives me about 4MiB for `object` and 95 MiB for `U50`, and roughly the same amount of time to create `arr` for both cases.

bryanwweber: Oh! I missed your earlier comment about `object`. I didn't see anything on that StackOverflow link about problems with saving, etc. and the test suite passes when using `object`. Is there something specific you're concerned about?

ischoegl: I ran a couple of quick tests, and would say that these are the things to consider:

* `dtype=object` is not the default behavior of strings handled by `numpy.ndarray`, i.e. every time something has to be loaded (e.g. using `genfromtxt`) *latin1* encoded fixed column strings have to be detected and a conversion needs to take place. I confirmed that saving to CSV isn't affected though.
* the change will affect the handling of `extra` columns, making some changes necessary (meaning that there may be a risk of merge conflicts due to #838)
* it certainly appears to be doable, but any future modification / feature enhancement has to consider the non-default behavior, as I don't think that it's good to randomly mix *latin1* fixed column arrays and `object`s pointing to strings.

Let me know about your thoughts. The alternative would be to leave a detailed comment where `empty_strings` is defined in `composite.py`

bryanwweber: I agree with your concern about bifurcating the behavior between `extra` columns and built in columns, and that reading from data sources would require extra processing. I hadn't thought of that. The balance is against how likely it is to truncate a built-in string (that is, a bug in the built-in column), and I think the balance falls in favor of NumPy's default behavior, i.e., I agree with the choice of `U50` and adding a comment to hopefully avoid the possible bug.

The last concern I have is with respect to setting the endian-ness of the string, I think it should be possible to leave that off and let the implementation decide what to use.

ischoegl: Ok - just to test, I pushed a version where `object` is used instead of `np.str_`. It's not *too* difficult but doesn't resolve my reservations (I had left the `extra` out though). The endian-ness comment is well taken - it's indeed not necessary (same for lengths), and going with defaults appears to be safe.

ischoegl: **PS:** the extra processing for loading of data isn't really an issue (there are only minor modifications necessary), so it's really the `extra` columns and potential future modifications that give me reason to pause. Let me know in case you should reconsider.

ischoegl:
@bryanwweber ... thanks again for your feedback - the comment is now added; I believe everything is taken care of at this point.

bryanwweber (approved):
Thanks @ischoegl!

ischoegl:
Thanks @bryanwweber!

PR #903, opened 2020-07-12
[Kinetics] Change error handling about invalid reactions

12Chao:
Current version of Cantera only throw `CanteraError` for one invalid Plog reaction and its invalid reaction condition(temperature, pressure) when loading input file by `ct.Solution()`; however, multiple Plog reactions with invalid Arrhenius parameters may exist in the input file. This pull request helps raise `CanteraError` for all invalid Plog reactions in the input file and their invalid conditions.
For example, there are 2 Plog reactions shown below with invalid Arrhenius parameters:
```
- equation: CH2CHOO <=> CO + CH3O  
  type: pressure-dependent-Arrhenius
  rate-constants:
  - {P: 0.316 atm, A: -1.8e+33, b: -7.27, Ea: 3.376e+04}
  - {P: 0.316 atm, A: 8.69e-50, b: 16.63, Ea: -3900.0}
  - {P: 1.0 atm, A: -3.83e+33, b: -7.2, Ea: 3.51e+04}
  - {P: 1.0 atm, A: 1.19e-39, b: 13.61, Ea: -1317.0}

- equation: CH2CHOO <=> CO2 + CH3 
  type: pressure-dependent-Arrhenius
  rate-constants:
  - {P: 100.0 atm, A: -2510.0, b: 1.41, Ea: 1.442e+04} 
  - {P: 100.0 atm, A: 4.05e-09, b: 5.14, Ea: 1.048e+04}

```
Current version of Cantera only throw the error message for the first reactions at first invalid presssure point(0.316 atm) and first invalid temperature point(200K, defined by Cantera) as shown below.
```
CanteraError thrown by Plog::validate:
Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'
at P = 32018.699999999983, T = 500.0
```
This pull request will change Cantera to raise error showing, in this case, first reaction is invalid at (0.316 atm, 500K; 0.316 atm, 1000K; 1 atm 500K, 1atm, 1000K); second reaction is invalid at 100 atm, 500K.
```
CanteraError thrown by Plog::validate:




Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'
at P = 32018.699999999983 , T = 500.0
Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'
at P = 32018.699999999983 , T = 1000.0
Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'
at P = 101324.99999999999 , T = 500.0
Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'
at P = 101324.99999999999 , T = 1000.0




Invalid rate coefficient for reaction 'CH2CHOO <=> CH3 + CO2'
at P = 10132499.999999985 , T = 500.0

```

Related to [enhancement proposal](https://github.com/Cantera/enhancements/issues/42#issuecomment-641341008)

**Checklist**


include/cantera/kinetics/RxnRates.h
```diff
@@ -290,8 +290,11 @@ class Plog
             }
             log_k2 = std::log(k);
         }
-
-        return std::exp(log_k1 + (log_k2-log_k1) * (logP_-logP1_) * rDeltaP_);
+        if (logP_ == logP1_) {
+            return std::exp(log_k1);
+        } else {
+            return std::exp(log_k1 + (log_k2-log_k1) * (logP_-logP1_) * rDeltaP_);
+        }
```
bryanwweber: Is this change necessary? Since `logP_` and `logP1_` are both `double` numbers, the chance that they are identical is very small. The previous line accounted for this by taking the difference, so that if it was almost zero, the offset from `log_k1` would be also nearly zero.

interfaces/cython/cantera/test/test_kinetics.py
```diff
@@ -404,7 +404,21 @@ def test_modify_thermo(self):
...
+                  "Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'"
+                  "\nat P = 101324.99999999999 , T = 500.0\n"
+                  "Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'"
+                  "\nat P = 101324.99999999999 , T = 1000.0\n\n\n\n\n"
+                  "Invalid rate coefficient for reaction 'CH2CHOO <=> CH3 + CO2'"
+                  "\nat P = 10132499.999999985 , T = 500.0")
+        try:
+            ct.Solution('pdep_err_test.yaml')
+        except ct.CanteraError as e:
+            self.assertEqual(str(e).find(err_msg), 116)
```
bryanwweber: This check finds the index that starts the error message, right? That seems a little bit fragile to me, in the sense that totally unrelated changes might change the index where this error message starts and would cause this test to fail. I don't think you want to use `self.assertRaisesRegex()` here, so I agree with the general structure. Can we try to find another way to check the error message? For instance, can we split the error message by newlines and assert that all of the lines in `err_msg` are present in the raised error?

src/kinetics/KineticsFactory.cpp
```diff
@@ -94,6 +96,7 @@ unique_ptr<Kinetics> newKinetics(std::vector<ThermoPhase*>& phases,
 
 void addReactions(Kinetics& kin, const AnyMap& phaseNode, const AnyMap& rootNode)
 {
+    fmt::memory_buffer err_Plog_reactions;
```
bryanwweber: I think it would be better if this were defined closer to the first place its used.

src/kinetics/KineticsFactory.cpp
```diff
@@ -168,17 +171,44 @@ void addReactions(Kinetics& kin, const AnyMap& phaseNode, const AnyMap& rootNode
             AnyMap reactions = AnyMap::fromYamlFile(fileName,
                 rootNode.getString("__file__", ""));
             for (const auto& R : reactions[node].asVector<AnyMap>()) {
-                kin.addReaction(newReaction(R, kin));
+                try{
+                    kin.addReaction(newReaction(R, kin));
+                }
+                catch(CanteraError& err){
+                    format_to(err_Plog_reactions, "{}", err.what());
```
bryanwweber: It seems like this will catch any `CanteraError`s that are generated by trying to add the reaction, not just errors related to Plog reactions. In that case, the name of the variable seems to be a bit of a misnomer. If you only want to capture the Plog reactions here, then I think you need some of the same string processing as you've got in the next section.

To be clear though, I think it might be a good thing to catch all the errors and report them all at once back to the user, not just the errors related to adding PLOG reactions. I'd like to try it out and see how that works from the user side though, first. I also realize that might be expanding the scope of this change quite a bit, so this is more of a "thinking-out-loud" paragraph 😄

src/kinetics/KineticsFactory.cpp
```diff
@@ -168,17 +171,44 @@ void addReactions(Kinetics& kin, const AnyMap& phaseNode, const AnyMap& rootNode
...
+                        captured_err.erase(0, 40);
+                        format_to(err_Plog_reactions, "{}", captured_err);
+                    } else {
+                        std::regex re("(\\w+)::(\\w+)");
+                        std::smatch match;
+                        regex_search(captured_err, match, re);
+                        std::string procedure = match[0];
+                        std::string msg = captured_err.substr(25+procedure.length(), -1);
+                        throw CanteraError(procedure,
+                            "{}", msg);
```
bryanwweber: Why is this processing necessary? What are you erasing here that wasn't here before? Processing based on character lengths is quite fragile, as I mentioned earlier, since changes in other places can totally break this without any warning.

src/kinetics/KineticsFactory.cpp
```diff
@@ -168,17 +171,44 @@ void addReactions(Kinetics& kin, const AnyMap& phaseNode, const AnyMap& rootNode
...
+                    }
+                }
             }
         }
     }
-
     kin.checkDuplicates();
+    if (!(to_string(err_Plog_reactions).empty())){
```
bryanwweber: This can be simplified to (and likewise in the other similar check):
```suggestion
    if (err_Plog_reactions.size()){
```

test/data/pdep_err_test.yaml
```diff
@@ -0,0 +1,263 @@
+description: |-
```
bryanwweber: Can you use the absolute minimum file that produces the condition you want to test, including removing all the comments that aren't related to the test?

src/kinetics/RxnRates.cpp
```diff
@@ -99,11 +102,15 @@ void Plog::validate(const std::string& equation)
                 // k is NaN. Increment the iterator so that the error
                 // message will correctly indicate that the problematic rate
                 // expression is at the higher of the adjacent pressures.
-                throw CanteraError("Plog::validate",
-                    "Invalid rate coefficient for reaction '{}'\nat P = {}, T = {}",
-                    equation, std::exp((++iter)->first), T[i]);
-            }
-        }
+                 // expression is at the higher of the adjacent pressures.
```
bryanwweber: This line is duplicated

src/kinetics/KineticsFactory.cpp
```diff
@@ -168,17 +171,44 @@ void addReactions(Kinetics& kin, const AnyMap& phaseNode, const AnyMap& rootNode
...
+        std::string Plog_err = to_string(err_Plog_reactions);
+        throw CanteraError("Plog::validate",
+                        "\n{}", Plog_err);
```
bryanwweber:
```suggestion
        throw CanteraError("Plog::validate",
                           "\n{}", to_string(err_Plog_reactions));
```

src/kinetics/RxnRates.cpp
```diff
@@ -99,11 +102,15 @@ void Plog::validate(const std::string& equation)
...
+                format_to(err_reactions, "\nInvalid rate coefficient for reaction '{}'\nat P = {} , T = {}",
+                    equation, std::exp((iter)->first), T[i]);
+             }
+         }
+     }
+    if (!(to_string(err_reactions).empty())){
+        throw CanteraError("Plog::validate",
+                        "\n{}", to_string(err_reactions));
```
bryanwweber:
```suggestion
                           "\n{}", to_string(err_reactions));
```

src/kinetics/RxnRates.cpp
```diff
@@ -99,11 +102,15 @@ void Plog::validate(const std::string& equation)
...
                 // message will correctly indicate that the problematic rate
                 // expression is at the higher of the adjacent pressures.
-                throw CanteraError("Plog::validate",
-                    "Invalid rate coefficient for reaction '{}'\nat P = {}, T = {}",
-                    equation, std::exp((++iter)->first), T[i]);
-            }
-        }
+                 // expression is at the higher of the adjacent pressures.
+                format_to(err_reactions, "\nInvalid rate coefficient for reaction '{}'\nat P = {} , T = {}",
+                    equation, std::exp((iter)->first), T[i]);
```
bryanwweber: Do you still need to increment the iterator here?

bryanwweber (changes requested):
Thanks @12Chao I think this is a really good start. As I mentioned in one of the comments, I wonder if it makes sense to process all the reactions and collect all the errors to report them all at once (duplicates, errors with Chebychev coefficients, etc.). That would be a pretty significant expansion of the scope of this PR though, so I don't want to put that on you until we decide whether it makes sense to go for that. Regardless, there are a few comments that you can fix up here that will help the final outcome anyways, plus a couple of questions.

include/cantera/kinetics/RxnRates.h
12Chao: Yes, I believe it is necessary(or at least there should be applied some change). I think in the validation case, since the input value of `update_C` is taken from `pressures_` in `Plog::validate`, so the lower reference pressure `log_P1`  should be the exact same value with `logP_` because the logarithm of lower reference pressure(`log_P1`) is assigned from `pressures_`  as well.  I add `if` here because I want to differentiate the validation case from the normal case(case that input pressure does not equal to any refference pressure). `std::exp(log_k1 + (log_k2-log_k1) * (logP_-logP1_) * rDeltaP_)` involves `log_k2` at higher reference pressure and `log_k1` at lower reference pressure, in validation case, every reference pressure point will be called twice as higher pressure point first and lower pressure point when the input changes to next reference pressure point. In this case, if the rate constant at a reference pressure point is invalid, the error will be raised twice as it involves in formula`std::exp(log_k1 + (log_k2-log_k1) * (logP_-logP1_) * rDeltaP_)` twice. For example, in reaction 
```
- equation: CH2CHOO <=> CO + CH3O  
  type: pressure-dependent-Arrhenius
  rate-constants:
  - {P: 1.0 atm, A: 1.8e+33, b: -7.27, Ea: 3.376e+04}
  - {P: 1.0 atm, A: 8.69e-50, b: 16.63, Ea: -3900.0}  
  - {P: 10.0 atm, A: -3.83e+33, b: -7.2, Ea: 3.51e+04}
  - {P: 10.0 atm, A: 1.19e-39, b: 13.61, Ea: -1317.0}
```
reaction rate constant at `P = 1.0 atm` is  valid (positive),  when validate the pressure point P = 1.0 atm, `log_k1` at lower reference pressure 1.0 atm is valid and it should pass the validation, but the `log_k2` at higher reference pressure 10.0 atm is invalid, so the formula `std::exp(log_k1 + (log_k2-log_k1) * (logP_-logP1_) * rDeltaP_)` will return `NaN` , which should actually be log_k1 because `logP_ - logP1_` equal to zero in the validation case.  In the original code, it is not a problem because is throws the next pressure point 10 atm in the error message and stops. However, to check all the reference pressure point in validation case, It is not necessary to involve `log_k2` when calculating `log_k1` .

src/kinetics/KineticsFactory.cpp
12Chao: Here I am trying to raise Plog error while not affecting raising other `CanteraError`, so I  distinguish the `Plog::validation` error and other `CanteraError`,  the first erase line(line189) is for erasing all the '*' sign automatically added in each CanteraError. `captured_err.erase(0, 40)` erases string `CanteraError thrown by Plog::validation`. If the captured `CanteraError` is not a `Plog::validation` error, it uses regex to extract the procedure(the function that raise the error) and message in `CanteraError` and use `throw CanteraError` to raise it again. I forgot to do the same in line 178.  In this case, as long as we do not change the `CanteraError` format like `CanteraError thrown by (\\w+)::(\\w+)` and `Plog::validation` in Plog error message, it should be ok. However, I agree it is fragile to process string based on character length or regex, and I would like to take any suggestion about how to improve this part. I think maybe a function to separate `CanteraError` into predure and message would be good?

src/kinetics/RxnRates.cpp
12Chao: Because of the change in `RxnRate.h` line 293, the iterator does not have to be incremented here

12Chao:
Hi @bryanwweber, thanks for the review, I have added some changes according to your comments. As I replied in the comment, I am trying to catch all the errors and only add Plog validation error to the buffer while raise other CanteraError again as the way it should be raised. I am not sure if capture all the errors and raise them all at once would be more helpful as it depends on users preference and the hardness to trace all the problems, but I personally like to visualize all the errors at once to avoid the fraustration of seeing them one by one.

interfaces/cython/cantera/test/test_kinetics.py
```diff
@@ -405,20 +405,30 @@ def test_modify_thermo(self):
...
+        err_msg1 = "Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'"
+        err_msg2 = "at P = 32018.699999999983 , T = 500.0"
+        err_msg3 = "Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'"
+        err_msg4 = "at P = 32018.699999999983 , T = 1000.0"
+        err_msg5 = "Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'"
+        err_msg6 = "at P = 101324.99999999999 , T = 500.0"
+        err_msg7 = "Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'"
+        err_msg8 = "at P = 101324.99999999999 , T = 1000.0"
+        err_msg9 = "Invalid rate coefficient for reaction 'CH2CHOO <=> CH3 + CO2'"
+        err_msg10 = "at P = 10132499.999999985 , T = 500.0"
```
bryanwweber: Rather than 10 variables, you can make the same tuple as before (although you have to remove the `\n` from the strings), and then loop for the asserts:

```python
err_msg_list = str(e).splitlines()
for msg in err_msg:
    self.assertIn(msg, err_msg_list)
```

12Chao: Thanks for the instruction, I just applied the change

include/cantera/kinetics/RxnRates.h
speth: Rather than changing this, which introduces an extra branch point every time these rates are evaluated, I would suggest replacing the call to `updateRC` in the `validate` method and just calculating the value of `k` directly at that point, which is just
```c++
double k = 0;
for (size_t i = ilow1_; i < ilow2_; i++) {
    k += rates_[i].updateRC(logT, recipT);
}
```
And then all you have to do is check that `k` is positive.

src/kinetics/RxnRates.cpp
```diff
@@ -90,20 +91,26 @@ Plog::Plog(const std::multimap<double, Arrhenius>& rates)
...
+    fmt::memory_buffer err_reactions;
     double T[] = {200.0, 500.0, 1000.0, 2000.0, 5000.0, 10000.0};
+
     for (auto iter = pressures_.begin(); iter->first < 1000; iter++) {
         update_C(&iter->first);
         for (size_t i=0; i < 6; i++) {
             double k = updateRC(log(T[i]), 1.0/T[i]);
-            if (!(k >= 0)) {
+            // avoid to add invalid k at cantera inserted pressure(P=0) to error message
+            if (!(k >= 0) && std::exp((iter)->first) != 0) {
```
speth: A cleaner approach would be to just skip over the first value in `pressures_` by incrementing the loop variable as it's initialized, e.g.
```c++
for (auto iter = ++pressures_.begin(); iter->first < 1000; iter++) {
```

src/kinetics/RxnRates.cpp
```diff
@@ -6,6 +6,7 @@
 #include "cantera/kinetics/RxnRates.h"
 #include "cantera/base/Array.h"
 
+using std::string;
```
speth: It doesn't look like you use an unqualified `string` declaration anywhere in this file, so I think this can be removed.

src/kinetics/RxnRates.cpp
```diff
@@ -90,20 +91,26 @@ Plog::Plog(const std::multimap<double, Arrhenius>& rates)
...
-                    "Invalid rate coefficient for reaction '{}'\nat P = {}, T = {}",
-                    equation, std::exp((++iter)->first), T[i]);
-            }
-        }
+                format_to(err_reactions, "\nInvalid rate coefficient for reaction '{}'\nat P = {} , T = {}",
+                          equation, std::exp((iter)->first), T[i]);
+             }
+         }
+     }
+    if (err_reactions.size()){
```
speth:
```suggestion
    if (err_reactions.size()) {
```

src/kinetics/RxnRates.cpp
```diff
@@ -90,20 +91,26 @@ Plog::Plog(const std::multimap<double, Arrhenius>& rates)
...
+            if (!(k >= 0) && std::exp((iter)->first) != 0) {
                 // k is NaN. Increment the iterator so that the error
                 // message will correctly indicate that the problematic rate
                 // expression is at the higher of the adjacent pressures.
-                throw CanteraError("Plog::validate",
-                    "Invalid rate coefficient for reaction '{}'\nat P = {}, T = {}",
-                    equation, std::exp((++iter)->first), T[i]);
-            }
-        }
+                format_to(err_reactions, "\nInvalid rate coefficient for reaction '{}'\nat P = {} , T = {}",
```
speth: The error message output would be a little cleaner if you added a format specifier for the pressure, e.g. `P = {:.5g}` so that some of rounding errors in unit conversion won't be as visible, e.g. you'll see `P = 1.0132e+05` instead of `P = 101324.99999999999`.

src/kinetics/KineticsFactory.cpp
```diff
@@ -168,17 +171,61 @@ void addReactions(Kinetics& kin, const AnyMap& phaseNode, const AnyMap& rootNode
...
                 rootNode.getString("__file__", ""));
             for (const auto& R : reactions[node].asVector<AnyMap>()) {
-                kin.addReaction(newReaction(R, kin));
+                try{
+                    kin.addReaction(newReaction(R, kin));
+                }
+                catch (CanteraError& err){
+                    std::string captured_err = err.what();
+                    captured_err.erase(std::remove(captured_err.begin(), captured_err.end(), '*'), captured_err.end());
+                    std::string str_to_erase = "CanteraError thrown by Plog::validate:";
```
speth: Rather than trying to reformat the output of `err.what()`, you can use the `err.getClass()` and the `err.getMessage()` methods, which should simplify things a bit.

src/kinetics/KineticsFactory.cpp
```diff
@@ -168,17 +171,61 @@ void addReactions(Kinetics& kin, const AnyMap& phaseNode, const AnyMap& rootNode
...
+                        captured_err.erase(pos, str_to_erase.length());
+                        format_to(err_Plog_reactions, "{}", captured_err);
+                    } else {
+                        std::regex re("(\\w+)::(\\w+)");
+                        std::smatch match;
+                        regex_search(captured_err, match, re);
+                        std::string procedure = match[0];
+                        std::string msg = captured_err.substr(25+procedure.length(), -1);
+                        throw CanteraError(procedure,
+                                           "{}", msg);
```
speth: I think this block can be replaced by just re-throwing the existing exception, i.e. `throw err`. But even beyond that, I'm not sure why you wouldn't want to collect all reaction errors in the same way that you aggregate the Plog errors.

src/kinetics/KineticsFactory.cpp
```diff
@@ -168,17 +171,61 @@ void addReactions(Kinetics& kin, const AnyMap& phaseNode, const AnyMap& rootNode
             AnyMap reactions = AnyMap::fromYamlFile(fileName,
                 rootNode.getString("__file__", ""));
             for (const auto& R : reactions[node].asVector<AnyMap>()) {
-                kin.addReaction(newReaction(R, kin));
+                try{
+                    kin.addReaction(newReaction(R, kin));
+                }
+                catch (CanteraError& err){
```
speth: These formatting suggestions apply here and elsewhere:
```suggestion
                try {
                    kin.addReaction(newReaction(R, kin));
                } catch (CanteraError& err) {
```

speth (changes requested):
Thanks for working on this, @12Chao. Besides the specific comments I've made here, I think this is actually really close to a more general improvement to the way we report errors in reactions. Instead of applying this approach of collecting the error messages from just Plog reactions, wouldn't it be simpler and more helpful to the user to do this for all of the errors that can occur when calling `addReaction`?

12Chao:
Hi @speth, thanks for the suggested changes. I was not sure about if it was helpful to capture all the errors and throw them all at once, so I only kept PLOG errors. I have changed the code to capture the error when add reactions and throw them at last.  In the test case the error message is:
```
**********************************************************************
CanteraError thrown by KineticsFactory::addReactions:


CanteraError thrown by Plog::validate:


Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'
at P = 32019 , T = 500.0
Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'
at P = 32019 , T = 1000.0
Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'
at P = 1.0132e+05 , T = 500.0
Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'
at P = 1.0132e+05 , T = 1000.0

CanteraError thrown by Plog::validate:


Invalid rate coefficient for reaction 'CH2CHOO <=> CH3 + CO2'
at P = 1.0132e+07 , T = 500.0
```

interfaces/cython/cantera/test/test_kinetics.py
```diff
@@ -404,7 +404,23 @@ def test_modify_thermo(self):
...
+        err_msg = ("Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'",
+                    "at P = 32019 , T = 500.0",
+                    "Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'",
+                    "at P = 32019 , T = 1000.0",
+                    "Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'",
+                    "at P = 1.0132e+05 , T = 500.0",
+                    "Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'",
+                    "at P = 1.0132e+05 , T = 1000.0",
+                    "Invalid rate coefficient for reaction 'CH2CHOO <=> CH3 + CO2'",
+                    "at P = 1.0132e+05 , T = 500.0")
```
speth: You can delete the repeated lines from this set of strings -- you only need to check for `Invalid rate coefficient for reaction '<reaction string>'` once for each reaction.

src/kinetics/KineticsFactory.cpp
```diff
@@ -11,6 +11,8 @@
 #include "cantera/kinetics/EdgeKinetics.h"
 #include "cantera/kinetics/importKinetics.h"
 #include "cantera/base/xml.h"
+#include <regex>
+
```
speth: This include can be removed, since you aren't using it anymore.

src/kinetics/KineticsFactory.cpp
```diff
@@ -168,17 +171,31 @@ void addReactions(Kinetics& kin, const AnyMap& phaseNode, const AnyMap& rootNode
             AnyMap reactions = AnyMap::fromYamlFile(fileName,
                 rootNode.getString("__file__", ""));
             for (const auto& R : reactions[node].asVector<AnyMap>()) {
-                kin.addReaction(newReaction(R, kin));
+                try {
+                    kin.addReaction(newReaction(R, kin));
+                } catch (CanteraError& err) {
+                    std::string captured_err = err.what();
+                    format_to(add_rxn_err, captured_err);
```
speth: * You can just write `err.what()` directly in the call to `format_to`. There's no need to create a variable for a string that's only used once.
* This is kind of a subtle point, but the signature of `format_to` is actually `format_to(buffer, format string, arguments...)`. What you've written is only valid in the special case where `captured_err` doesn't contain anything that might be interpreted as a format group, i.e. something containing `{` and `}`. For some validation errors, this definitely happens, which results in an error here. The correct syntax for this is actually `format_to(add_rxn_err, "{}", captured_err)`.

To catch this mistake, I would suggest adding a test for a different kind of reaction error, such as an unbalanced reaction equation.

src/kinetics/KineticsFactory.cpp
```diff
@@ -168,17 +171,31 @@ void addReactions(Kinetics& kin, const AnyMap& phaseNode, const AnyMap& rootNode
...
             }
         }
     }
-
     kin.checkDuplicates();
+    if (add_rxn_err.size()) {
+    std::string reaction_err = to_string(add_rxn_err);
+    throw CanteraError("KineticsFactory::addReactions",
+                        "\n{}", reaction_err);
+    }
```
speth: Please fix the indentation of the body of the `if` block.

Also, this code is in the function `addReactions`, which is not actually part of the `KineticsFactory` class, so the method name here should just be `addReactions`.

src/kinetics/RxnRates.cpp
```diff
@@ -90,21 +90,30 @@ Plog::Plog(const std::multimap<double, Arrhenius>& rates)
...
-                    "Invalid rate coefficient for reaction '{}'\nat P = {}, T = {}",
-                    equation, std::exp((++iter)->first), T[i]);
+                format_to(err_reactions, "\nInvalid rate coefficient for reaction '{}'\nat P = {:.5g} , T = {}",
+                          equation, std::exp((iter)->first), T[i]);
             }
         }
     }
+    if (err_reactions.size()) {
+        throw CanteraError("Plog::validate",
+                           "\n{}", to_string(err_reactions));
```
speth: I think the formatted error message ends up containing more blank lines than it needs to, which you can improve by removing the one that's added on this line. The `CanteraError` constructor is a bit more clever than `format_to`, so you can just remove the argument `"\n{}"` and get the behavior you want.

speth (changes requested):
Thanks for the updates, @12Chao, I think this has become both useful with a simpler implementation. I just had a few additional suggestions.

speth (approved):
Thanks, @12Chao. I squashed the commits to reduce some of the churn and make it easier to see what changes were being made here, and fixed a couple of small formatting issues that had snuck in.

interfaces/cython/cantera/test/test_kinetics.py
```diff
@@ -405,6 +405,23 @@ def test_modify_thermo(self):
...
+    def test_pdep_err(self):
+        err_msg = ("Invalid rate coefficient for reaction 'CH2CHOO <=> CH3O + CO'",
+                    "at P = 32019, T = 500.0",
+                    "at P = 32019, T = 1000.0",
+                    "at P = 1.0132e+05, T = 500.0",
+                    "at P = 1.0132e+05, T = 1000.0",
+                    "Invalid rate coefficient for reaction 'CH2CHOO <=> CH3 + CO2'",
+                    "at P = 1.0132e+05, T = 500.0",
+                    "InputFileError thrown by Kinetics::checkReactionBalance:",
+                    "The following reaction is unbalanced: H2O2 + OH <=> 2 H2O + HO2")
```
bryanwweber: These lines are indented by one too many spaces

test/data/addReactions_err_test.yaml
```diff
@@ -0,0 +1,169 @@
+description: |-
+  This is the file for testing Plog errors
+
+generator: ck2yaml
+input-files: [mech.inp, therm.DAT, trans.dat]
+cantera-version: 2.5.0a2
+date: Mon, 29 Jun 2020 14:13:58 +0000
+
```
bryanwweber: I would suggest deleting these lines to keep the test file as small as possible.

test/data/addReactions_err_test.yaml
```diff
@@ -0,0 +1,169 @@
...
+      -4.8371971e+04, 9.9009035]
+    - [4.6365111, 2.7414569e-03, -9.9589759e-07, 1.6038666e-10, -9.1619857e-15,
+      -4.9024904e+04, -1.9348955]
+  transport:
+    model: gas
+    geometry: linear
+    well-depth: 244.0
+    diameter: 3.763
+    polarizability: 2.65
+    rotational-relaxation: 2.1
```
bryanwweber: Also this, and the other transport properties as well. I'm actually surprised you can specify a transport model in the phase, since a few species don't have transport parameters.

bryanwweber (changes requested):
A few small formatting changes. I would advocate for this change to be merged after the release of 2.5.0, since it's a somewhat large new feature, and the beta for 2.5.0 has already been published. Either that, or creating a release branch for 2.5 at this point before merging this PR.

12Chao:
Thanks for the review, @bryanwweber, I have made the changes according to the suggested changes.

speth:
@12Chao I'm not sure exactly what led to this, but your latest changes have reverted the cleanup work that I did and introduced a new, undesirable merge commit that prevents this PR from being integrated. I would suggest that you reset your branch to match the version I last updated, which I have pushed to my fork (https://github.com/speth/cantera/tree/pr/12Chao/903) and make the updates to address @bryanwweber's review starting from that point.

12Chao:
@speth sorry about that, I pulled the changes from my repo trying to sync the changes you made, and it messed up the commit history. I think I have fixed it.

test/data/addReactions_err_test.yaml
speth: The reason this doesn't cause a problem is because loading the file results in an error in creating the `Kinetics` object before it gets to trying to create a `Transport` object.

bryanwweber (approved):

Issue #902, opened 2020-07-11
Bug in handling of surface reactions imported from YAML


ischoegl:
**Problem description**

The switch from `methane_pox_on_pt.cti` to `methane_pox_on_pt.yaml` input apparently broke the Python example `surf_pfr.py` (see original issue report #877). 

**Edit:** The example `catalytic_combustion.py` is affected by a closely related issue (see #873).

**Steps to reproduce**

A run of the Python example `surf_pfr.py` currently fails (see #877). Switching from YAML back to CTI input restores the old behavior.  It can be easily verified that re-creating YAML using `cti2yaml` does not fix the issue. The same issue also arises if the original file is converted from XML (via `ctml_writer` and `ctml2yaml`).

**Behavior**

Time integration reduces CO levels to negative values until the density itself becomes negative and an error is raised (see #877), which appears to be either due to a faulty `cti2yaml` / `ctml2yaml` conversion of reaction parameters, or a faulty initialization of a cantera object from YAML. The context points at an issue with surface reactions.

**System information**

- Cantera version: latest development (2.5.0a4)
- OS: Ubuntu 18.04
- Python/MATLAB/other software versions: Python 3.6.9

**Additional context**

* #877 (superseded by this issue report)
* #873 (superseded by this issue report)

The behavior became apparent after 1866e35 which switched example input files to YAML.

bryanwweber:
Spent parts of three days debugging this... man, this was tricky. The intermediate cause appears to be because the thermo phases are stored in a different order inside the kinetics object for YAML vs. CTI files. This means that the order of species in the internal arrays is different, which results in a different rates of production. This cause can be verified by changing the order of the `ThermoPhase` arrays in the `_SolutionBase.__cinit__()` method:

```python
        if isinstance(self, Kinetics):
            for phase in adjacent:
                # adjacent bulk phases for a surface phase
                v.push_back(phase.thermo)
            v.push_back(self.thermo)  # Move this after the adjacent phases to fix the example
```

With that change, both `surf_pfr.py` and `catalytic_combustor.py` work again.

I don't believe this is the root cause, or the appropriate fix, because the underlying calculation of the rates of production is intended to be independent of the order that phases are added. More digging is required to find where this assumption is violated in the current code.

Debugging details follow, just in case anyone finds it interesting, and also because I'm so frustrated at how long this took that I want to write it all out!
<details>
[...]

Issue #873, opened 2020-06-19
Broken catalytic_combustion.py example


ischoegl:
**System information**

- Cantera version: development version
- OS: Ubuntu 18.04
- Python/MATLAB version: 3.6.9

**Expected behavior**

The Cantera example `catalytic_combustion.py` works without error

**Actual behavior**

The Cantera example `catalytic_combustion.py` failsr

**To Reproduce**

1. Run: `python catalytic_combustion.py`
2. See error
```
.............................................................................
Attempt Newton solution of steady-state problem...    failure. 
Take 10 timesteps   Traceback (most recent call last):
  File "catalytic_combustion.py", line 97, in <module>
    sim.solve(loglevel)
  File "interfaces/cython/cantera/onedim.pyx", line 1137, in cantera._cantera.Sim1D.solve
cantera._cantera.CanteraError: 
***********************************************************************
CanteraError thrown by OneDim::timeStep:
Time integration failed.
***********************************************************************
```
**Other:**

The example works as expected on 2.4.

ischoegl:
PS: the problematic behavior was introduced in 1866e35 which switched the input file to YAML, i.e. it is a manifestation of an underlying bug reported in #902; the old CTI input still works.

Issue #872, opened 2020-06-19
Broken diffusion_flame_extinction.py example


ischoegl:
**System information**

- Cantera version: current development branch
- OS: Ubuntu 18.04
- Python/MATLAB version: 3.6.9

**Expected behavior**

`diffusion_flame_extinction.py` terminates and produces plot.

**Actual behavior**

`diffusion_flame_extinction.py` appears to be stuck in indefinite loop.

**To Reproduce**

1. Run `python `
2. After the temperature drops below `temperature_limit_extinction`, the loop gets stuck:
```
Flame extinguished at n = 284. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 285. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 286. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 287. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 288. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 289. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 290. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 291. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 292. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 293. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 294. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 295. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 296. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 297. Restoring n = 283 with alpha = 284.0
Flame extinguished at n = 298. Restoring n = 283 with alpha = 284.0
[...]
```
3. The above keeps running beyond `n = 2000`

**Additional Context:**

Issue was found in the context of #871.

**PS:** It is relatively straight-forward to ensure that the loop exits, but there appear to be some issues with the implementation of the termination criterion itself.

bryanwweber:
Thanks for reporting this! Can you locate when this problem started happening? Is it a problem for the 2.4.0 release?

ischoegl:
@bryanwweber ... example works on 2.4.

ischoegl:
Recent commits appear to have resolved the issue - I no longer can reproduce it.

PR #905, opened 2020-07-29
Fix examples

### LINKED ISSUE #821: Broken extract_submechanism.py example
**System information**

- Cantera version: 2.5.0a4 (current master / commit c16c4b0b0f3de6bb9e5366db5d6154ab2a9e0189)
- OS: Ubuntu
- Python/MATLAB version: Python 3.6.8

**Expected behavior**

The example `kinetics/extract_submechanism.py` runs.

**Actual behavior**

```
In [1]: %run extract_submechanism.py
Species: H2, H, O, O2, OH, H2O, HO2, H2O2, C, CO, CO2, N2
/src/build/python/cantera/examples/kinetics/extract_submechanism.py in <module>()
     37 
     38 # Filter reactions, keeping only those that only involve the selected species
---> 39 all_reactions = ct.Reaction.listFromFile('gri30.yaml')
     40 reactions = []
     41 

interfaces/cython/cantera/reaction.pyx in cantera._cantera.Reaction.listFromFile()

ValueError: A Kinetics object is required.
```

**To Reproduce**

See above.

bryanwweber:

- Fixes the Python examples for any changes since they were last run
- Adds a CI job using `find ... -exec python` to run the examples and just double check that they all work. I added this because I expect that doing real regression testing in SCons will require quite a bit of effort and so will be some way off. This stop gap will be easy to delete or modify when the time comes. It is not intended to be comprehensive or something a developer can run on their own machine (although copying the command is effective). It is intended as a double check on my work here and a hopeful hedge against future breakage.

Copies of commit messages for reference:

## [[Cython/Examples] Fix extract_submechanism example](https://github.com/Cantera/cantera/pull/905/commits/048890775201e008191dd68ca3c201454d7972e0)

The Reaction.listFromFile() method requires a Kinetics instance when
reading a YAML file. In this case, the appropriate Kinetics instance
could be created by Solution('gri30.yaml'). Since we'd have to create
that object anyways to be able to use Reaction.listFromFile(), we might
as well use the Solution.reactions() method, which returns the same
list as Reaction.listFromFile() for this case where the species
definitions are in the same file as the reaction definitions.

## [Fix infinite loop in diffusion_flame_extinction.py](https://github.com/Cantera/cantera/pull/905/commits/0cb09eddda0e4cb5ed9286d4e46e60368b582a47)

This change reworks the abortion criteria for the diffusion flame
extinction example. Previously, the loop was only aborted when the flame
temperature was above the inlet temperature, the difference of the flame
temperature in the prior two iterations was less than the threshold, AND
the change in strain rate parameter was less than the threshold.
However, I found that this combination of conditions was never reached,
because the only way to change the strain rate delta was to have an
extinguished flame, and eventually the flame remains extinguished even
if the delta_alpha is reduced. This leads to an infinite loop, since the
abortion criteria check are only done if the flame is burning. This
change moves the abortion criteria check into the case where the flame
is not burning.

In addition, the flame temperature check is modified to use
np.isclose(). I found that on the last burning iteration, the maximum
flame temperature was only just barely above the oxidizer inlet
temperature. Thus, the result printed to the screen was not very useful.
The change to np.isclose() produces a more useful result for the final
burning iteration before extinction. The last non-burning iteration is
also saved to the output file(s) for reference.

## [Fix boundary emissivities deprecation warning](https://github.com/Cantera/cantera/pull/905/commits/7357c34d2e4a3c49b9891c41414693b644f15f70)

The set_boundary_emissivities method is deprecated and will be removed
after Cantera 2.5.0. This change switches one example to use the new
property. The property was also missing from the FlameBase class.

## [[Example] Specify data file path in flame_fixed_T](https://github.com/Cantera/cantera/pull/905/commits/ece2b5eeb19279d9d359a04dd4a9c4a6d5a64296)

This allows the script to be run from the installed location, or from
the root of the source tree.

## [[Examples] Replace argon.yaml with air.yaml](https://github.com/Cantera/cantera/pull/905/commits/b0591844ea193406c8b6527f43862af68182a93c)

argon.yaml is deprecated, use air.yaml and specify the composition
instead.

## Remaining Fixes

Per #902, we need to determine a final solution for the loading of YAML files with surface mechanisms. I expect the new job to fail until I figure out how to fix that.

Fixes #821, #902, #877, #873, #872 

**Checklist**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -174,6 +174,44 @@ jobs:
       - name: Test Cantera
         run: scons test
 
+  run-examples:
+    name: Run the Python examples using bash
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+        name: Checkout the repository
+        with:
+          submodules: recursive
+      - name: Setup Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: '3.8'
+          architecture: x64
+      - name: Install Apt dependencies
+        run: sudo apt-get install libboost-dev gfortran graphviz liblapack-dev libblas-dev
+      - name: Upgrade pip
+        run: python3 -m pip install -U pip setuptools wheel
+      - name: Install Python dependencies
+        run: python3 -m pip install ruamel.yaml scons numpy cython h5py pandas matplotlib scipy
+      - name: Build Cantera
+        run: python3 `which scons` build -j2
+      - name: Run the examples
+        # See https://unix.stackexchange.com/a/392973 for an explanation of the -exec part
+        run: |
+          find interfaces/cython/cantera/examples -type f -iname "*.py" \
+          -exec sh -c 'for n; do echo "$n" | tee -a results.txt && python3 "$n" >> results.txt || exit 1; done' sh {} +
+        env:
+          PYTHONPATH: build/python
+          PYTHONWARNINGS: error
+          MPLBACKEND: Agg
+      - name: Save the results file for inspection
+        uses: actions/upload-artifact@v2
+        with:
+          path: results.txt
+        # Run this step if the job was successful or failed, but not if it was cancelled
+        # Using always() would run this step if the job was cancelled as well.
+        if: failure() || success()
+
   multiple-sundials:
     name: Sundials ${{ matrix.sundials-ver }}
     runs-on: ubuntu-latest
--- a/data/sofc.yaml
+++ b/data/sofc.yaml
@@ -5,7 +5,7 @@ description: |-
   species thermochemistry ARE NOT REAL VALUES - they are chosen only
   for the purposes of this example.
 
-  Defines bulk (i.e., 3D) phases - a gas, a metal, and an oxide.
+  Defines bulk (that is, 3D) phases - a gas, a metal, and an oxide.
   The gas contains only the minimum number of species needed to model
   operation on hydrogen. The species definitions are imported from
   gri30.yaml. The initial composition is set to hydrogen + 5% water, but
@@ -26,14 +26,14 @@ description: |-
   charge balances on the metal, then we would require a more complex
   model. Note that there is no work function for this metal.
 
-  Note: the "const_cp" species thermo model is used throughout this
+  Note: the 'const_cp' species thermo model is used throughout this
   file (with the exception of the gaseous species, which use NASA
-  polynomials imported from gri30.cti). The const_cp model assumes a
+  polynomials imported from gri30.yaml). The const_cp model assumes a
   constant specific heat, which by default is zero. Parameters that
-  can be specified are cp0, t0, h0, and s0. If omitted, t0 = 300 K, h0
-  = 0, and s0 = 0. The thermo properties are computed as follows: h =
-  h0 + cp0*(t - t0), s = s0 + cp0*ln(t/t0). For work at a single
-  temperature, it is sufficient to specify only h0.
+  can be specified are cp0, t0, h0, and s0. If omitted, t0 = 300 K,
+  h0 = 0, and s0 = 0. The thermo properties are computed as follows:
+  h = h0 + cp0*(t - t0), s = s0 + cp0*ln(t/t0).
+  For work at a single temperature, it is sufficient to specify only h0.
 
   The 'oxide_bulk' phase is a very simple model for the bulk phase. We only
   consider the oxygen sublattice. The only species we define are a
@@ -41,7 +41,7 @@ description: |-
   required input, but is not used here, so may be set arbitrarily.
 
   The vacancy will be modeled as truly vacant - it contains no atoms,
-  has no charge, and has zero enthalpy and entropy.  This is different
+  has no charge, and has zero enthalpy and entropy. This is different
   from the usual convention in which the vacancy properties are are
   expressed relative to the perfect crystal lattice. For example, in
   the usual convention, an oxygen vacancy has charge +2. But the
@@ -65,7 +65,7 @@ description: |-
   'mol' and 'cm' were specified in the units directive below as the
   units for quantity and length, respectively.
   2. The 'reactions' field specifies that all reaction entries in this file
-  that have are in the 'metal_surface-reactions' field are reactions belonging
+  that are in the 'metal_surface-reactions' field are reactions belonging
   to this surface mechanism.
 
   On the oxide surface, we consider four species:
@@ -74,11 +74,11 @@ description: |-
       3. OH'(ox)  - a surface hydroxyl with charge -1
       4. H2O(ox)  - physisorbed neutral water
 
-  The 'tpb' phase is th etriple phase boundary between the metal, oxide, and gas. A
-  single species is specified, but it is not used, since all reactions
-  only involve species on either side of the tpb. Note that the site
-  density is in mol/cm. But since no reactions involve TPB species,
-  this parameter is unused.
+  The triple phase boundary (TPB) between the metal, oxide, and gas is
+  specified in the 'tpb' phase. A single species is specified, but it
+  is not used, since all reactions only involve species on either side
+  of the TPB. Note that the site density is in mol/cm. But since no
+  reactions involve TPB species, this parameter is unused.
 
 generator: cti2yaml
 cantera-version: 2.5.0a3
--- a/include/cantera/thermo/EdgePhase.h
+++ b/include/cantera/thermo/EdgePhase.h
@@ -22,8 +22,8 @@ namespace Cantera
  * thermodynamic object.
  *
  * All of the equations and formulations carry through from SurfPhase to this
- * EdgePhase object. It should be noted however, that dimensional object with
- * length dimensions, have their dimensions reduced by one.
+ * EdgePhase object. It should be noted that dimensional quantities with
+ * dimensions including a length have their dimensions reduced by one.
  *
  * @ingroup thermoprops
  */
--- a/interfaces/cython/cantera/examples/kinetics/extract_submechanism.py
+++ b/interfaces/cython/cantera/examples/kinetics/extract_submechanism.py
@@ -14,7 +14,8 @@ from timeit import default_timer
 import cantera as ct
 import matplotlib.pyplot as plt
 
-all_species = ct.Species.listFromFile('gri30.yaml')
+input_file = 'gri30.yaml'
+all_species = ct.Species.listFromFile(input_file)
 species = []
 
 # Filter species
@@ -36,7 +37,8 @@ species_names = {S.name for S in species}
 print('Species: {0}'.format(', '.join(S.name for S in species)))
 
 # Filter reactions, keeping only those that only involve the selected species
-all_reactions = ct.Reaction.listFromFile('gri30.yaml')
+ref_phase = ct.Solution(thermo='ideal-gas', kinetics='gas', species=all_species)
+all_reactions = ct.Reaction.listFromFile(input_file, ref_phase)
 reactions = []
 
 print('\nReactions:')
@@ -51,8 +53,8 @@ for R in all_reactions:
     print(R.equation)
 print('\n')
 
-gas1 = ct.Solution('gri30.yaml')
-gas2 = ct.Solution(thermo='IdealGas', kinetics='GasKinetics',
+gas1 = ct.Solution(input_file)
+gas2 = ct.Solution(thermo='ideal-gas', kinetics='gas',
                    species=species, reactions=reactions)
 
 
--- a/interfaces/cython/cantera/examples/surface_chemistry/sofc.py
+++ b/interfaces/cython/cantera/examples/surface_chemistry/sofc.py
@@ -120,14 +120,16 @@ def anode_curr(E):
     # get the species net production rates due to the anode-side TPB reaction
     # mechanism. The production rate array has the values for the neighbor
     # species in the order listed in the .yaml file, followed by the tpb phase.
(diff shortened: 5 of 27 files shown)


ischoegl:
@bryanwweber ... thanks for fixing.

I'm happy to see running of examples added to CI. I'm still not sure why I got so much blow-back when I suggested the same in #892 (which I closed and moved to Cantera/enhancements#60 as it appeared that full regression tests were the preferred option). Doing the same from bash is probably less comprehensive than my `scons` approach, but ultimately accomplishes the same goal. ~_Edit: I reopened #892 for discussion._~

**PS:** the one issue that came up with #892 are `DeprecationWarning`s, which would be good to add (all it takes is to add the flag `-W error::DeprecationWarning:` when running examples).

ischoegl:
**PPS:** one other thing that came up in my own tests is poor convergence of `'ion_burner_flame.py` - mainly wanted to mention it here also as it took a long time to run.

.github/workflows/main.yml
```diff
@@ -174,6 +174,33 @@ jobs:
...
+        run: sudo apt-get install libboost-dev gfortran
+      - name: Upgrade pip
+        run: python3 -m pip install -U pip setuptools wheel
+      - name: Install Python dependencies
+        run: python3 -m pip install ruamel.yaml scons numpy cython h5py pandas
+      - name: Build Cantera
+        run: python3 `which scons` build -j2
+      - name: Run the examples
+        run: |
+          find interfaces/cython/cantera/examples -type f -name '*.py' -exec python '{}' \;
```
ischoegl: To raise `DeprecationWarnings` as errors, the line could be replaced by:
```
        run: |
          find interfaces/cython/cantera/examples -type f -name '*.py' -exec python -W error::DeprecationWarning: '{}' \;
```
(which works on Linux / OSX, but won't work on windows due to an upstream issue with `imp`)

bryanwweber: Yes, I was actually planning to set the `PYTHONWARNINGS` environment variable, but accomplishing the same thing. Thanks for the suggestion!

bryanwweber:
Thanks for the comments @ischoegl! The concern that I had with your previous approach was exactly that it was adding commands to SCons (making that more complicated) without adding what we want to eventually do. As I said, adding the running here is in part to make sure that I've actually fixed what I set out to fix and avoid "It works on my machine!" problems. I would be fine with deleting that CI run before this PR is merged, once all the examples are fixed up, if we think it adds too much overhead. It will be as easy as deleting the lines from the workflow, without having to modify several of the SConstruct/SConscript files. It was an intentional choice not to try to make something that works on all the platforms, since my idea was to have the simplest possible thing to add and the simplest possible thing to remove when the time comes.

The `ion_burner_flame.py` example does take a *very* long time to run, and I've also been having some convergence problems.

ischoegl:
@bryanwweber ... you're welcome. 

I am strongly in favor of running examples during CI in one form or another. Ultimately it makes little difference whether this is done using your approach here or what I had put together in #892. Regarding `scons`, I tried my best to not make it overly complicated; the advantage would be that it can be run on developer's (or less experienced contributor's!) machines without relying on CI, which may be helpful.

.github/workflows/main.yml
ischoegl: Not sure whether `PYTHONWARNINGS` has the same granularity as the warnings filter - simply setting it to `error` may catch too many exceptions (which is problematic if warnings are raised in external packages - see current `imp` issue on Windows).

bryanwweber: According to the docs, `PYTHONWARNINGS` is equivalent to the `-W` switch, and has the same syntax except that a comma-separated list of conditions is passed instead of multiple `-W` options. In any case, this is only ever intended to run on Ubuntu, but I do want to catch deprecation warnings in third-party software (e.g., NumPy), so I think catching everything until we find some exceptions is the right thing to do here.

bryanwweber:
The examples appear to be working now, with the exception of `ion_burner_flame.py` which I don't know how to address. Otherwise, I think this PR is done.

.github/workflows/main.yml
ischoegl: In the newest version of the `PYTHONWARNINGS` filter you implemented (`error,error::DeprecationWarning`), doesn't the first `error` already turn all warnings into exceptions? (any subsequent filter adds upon previous specifications, so there's nothing new added?) Regarding the environment variable, I wasn't sure as my attempts to use it with scons were not successful (likely my fault), but the [docs](https://docs.python.org/3/library/warnings.html) are indeed quite clear ...

bryanwweber: > In the newest version of the `PYTHONWARNINGS` filter you implemented (`error,error::DeprecationWarning`), doesn't the first `error` already turn all warnings into exceptions?

I think it should, that was my attempt to debug something where I thought there was a deprecation warning that wasn't being converted. It could probably be removed.

ischoegl:
@bryanwweber ... it may make sense to move 529b799 to its own PR so the CI runners can be fixed without being affected by the unrelated review of examples?

ischoegl:
@bryanwweber ... I would advocate for a temporary removal of the `ion_burner_flame.py` example for 2.5, as I do have concerns that a fix would not be simple (see #909, despite not being a root cause). The example appears to have had poor convergence since 2.4 (which is the release where it was introduced).

bryanwweber:
@ischoegl I'm not in favor of removing the example wholesale. Given the response from Bang-Shiuh in #908, I wonder if issuing a runtime warning and putting a note in the docstring would be more appropriate.

ischoegl:
~@bryanwweber ... I don't think that the intent should be to remove the example, just to not package it for 2.5, as the fix may be more involved (unless @BangShiuh has an idea for how to fix it before the release, which is presumably three weeks out). I believe it to be more important to have the ability to run examples for CI; including an example that is known to fail feels awkward.~

~**PS:** A run-time warning is one way to go about it, but the outcome would be to always have a failing test in CI.~

ischoegl:
@bryanwweber ... nevermind, I found a band-aid fix for `ion_burner_flame.py` - a reduction of domain size makes things converge - see #912. This is clearly a band-aid; after reviewing the code, there appear to be several issues with `IonFlow` (I opened #913 to document). However, the fix in #912 should be good enough for 2.5.

bryanwweber:
Thanks, I saw that change. Does that have any other effect on the solution, are the boundary conditions satisfied, etc.?

ischoegl:
> Does that have any other effect on the solution, are the boundary conditions satisfied, etc.?

The boundary conditions - as implemented - are satisfied by the solver; the physicality is a different issue. Curiously, @BangShiuh removed the offending `dYdz` condition in #700 - which is still pending. I just verified that this works for both examples; I just added this to #912 so #909 can be closed. I opened a separate issue report #913 summarizing non-trivial items I found while looking over the code.

.github/workflows/main.yml
```diff
@@ -174,6 +174,43 @@ jobs:
...
+      - name: Install Python dependencies
+        run: python3 -m pip install ruamel.yaml scons numpy cython h5py pandas matplotlib scipy
+      - name: Build Cantera
+        run: python3 `which scons` build -j2
+      - name: Run the examples
+        # See https://unix.stackexchange.com/a/392973 for an explanation of the -exec part
+        run: |
+          find interfaces/cython/cantera/examples -type f -iname "*.py" ! -iname "*ion_burner_flame.py" \
+          -exec sh -c 'for n; do echo "$n" | tee -a results.txt && python3 "$n" >> results.txt || exit 1; done' sh {} +
+        env:
```
ischoegl, line 203: I'd also recommend adding the environmental variable
```
        MPLBACKEND: Agg
```
(see [docs](https://matplotlib.org/3.3.0/faq/environment_variables_faq.html)) to ensure matplotlib won't act up when run on a headless server. It is currently optional on Linux (which has better backend support), but leaving it out on other OS's will wreak havoc (based on my experience in #892).

ischoegl:
@bryanwweber ... I saw that you cherry-picked the commits from #912 as discussed :+1:. Beyond, I *think* #899 is ready to be merged as long as my updated comment is ok.

data/sofc.yaml
```diff
@@ -74,10 +74,10 @@ description: |-
...
       4. H2O(ox)  - physisorbed neutral water
 
-  The 'tpb' phase is th etriple phase boundary between the metal, oxide, and gas. A
-  single species is specified, but it is not used, since all reactions
-  only involve species on either side of the tpb. Note that the site
-  density is in mol/cm. But since no reactions involve TPB species,
+  The 'tpb' phase is the triple phase boundary between the metal, oxide,
+  and gas. A single species is specified, but it is not used, since all
+  reactions only involve species on either side of the tpb. Note that the
+  site density is in mol/cm. But since no reactions involve tpb species,
```
speth: I'd suggest going the other way around and capitalize the 'TPB' initialism when it's used in text. (not suggesting changing the phase name).

include/cantera/thermo/EdgePhase.h
```diff
@@ -22,8 +22,8 @@ namespace Cantera
  * thermodynamic object.
  *
  * All of the equations and formulations carry through from SurfPhase to this
- * EdgePhase object. It should be noted however, that dimensional object with
- * length dimensions, have their dimensions reduced by one.
+ * EdgePhase object. It should be noted that dimensional objects with
+ * dimensions including a length have their dimensions reduced by one.
```
speth, line 26: Would this be more if we said dimensional *quantities*?

interfaces/cython/cantera/examples/kinetics/extract_submechanism.py
```diff
@@ -36,7 +37,7 @@
 print('Species: {0}'.format(', '.join(S.name for S in species)))
 
 # Filter reactions, keeping only those that only involve the selected species
-all_reactions = ct.Reaction.listFromFile('gri30.yaml')
+all_reactions = ct.Solution(input_file).reactions()
```
speth: My only concern about the proposed fix is that it leaves us without an example showing how to use `Reaction.listFromFile`. I think the alternative which shows what is actually needed to use the method now is to write:
```py
ref_phase = ct.Solution(thermo='ideal-gas', kinetics='gas', species=all_species)
all_reactions = ct.Reaction.listFromFile('gri30.yaml', ref_phase)
```
This avoids the redundancy of instantiating all of the `Reaction` objects twice, but does give you a `Solution` object that defines all of the `Species` which need to be present in order to instantiate the reactions.

interfaces/cython/cantera/examples/surface_chemistry/sofc.py
```diff
@@ -123,11 +123,12 @@ def anode_curr(E):
     # Since the first neighbor phase is the bulk metal, species 0 is the
     # electron.
     w = tpb_a.net_production_rates
+    electron_index = tpb_a.kinetics_species_index('electron')
```
speth, line 127: Comment above needs to be fixed as well.

src/oneD/Boundary1D.cpp
```diff
@@ -707,9 +705,17 @@ void ReactingSurf1D::eval(size_t jg, double* xg, double* rg,
         double* xb = x - nc;
         rb[c_offset_T] = xb[c_offset_T] - m_temp; // specified T
         size_t nSkip = m_flow_left->rightExcessSpecies();
+        size_t l_offset = 0;
+        thermo_t* left_thermo = &m_flow_left->phase();
```
speth:
```suggestion
        ThermoPhase* left_thermo = &m_flow_left->phase();
```
I would eventually like to eliminate these typedefs for the core Cantera types -- they just introduce unnecessary confusion.

speth (changes requested):
I think this is mostly fine, with a few small changes.

speth:
@ischoegl: I didn't intend my comments on #892 to be a wholesale rejection of the approach. I do want to eventually have the ability to do some more refined testing of the Python examples, but I realize that I'm not going to have that ready before the current release, so a stop-gap of one sort or another is useful. I don't have a strong preference for Bryan's approach here over yours for this purpose, but I think it will be easier to take this version along with the other fixes Bryan's added here.

ischoegl:
@speth ... thanks for the comment - I was mainly surprised as the initial push-back was stronger than expected. As mentioned having the ability to test examples resolves 90% of the issues. @bryanwweber's stop-gap is simpler than #892 , and testing from scons would likely not be used often anyways. The important thing is that issues with examples won't go unnoticed once this stop-gap is adopted.

ischoegl:
@bryanwweber ... the conflict that just popped up is due to 7c4240230bb6fbd26b5 (where a smaller fix was just merged in #899). I didn't realize that I found issues twice with the same file.

bryanwweber:
Thanks @ischoegl I'll sort out the conflict and amend the commit

data/sofc.yaml
bryanwweber: I do not recall making that change to lowercase `tpb`... In any case, I defined the acronym at the first use and uppercase-d both times in the text.

bryanwweber:
@speth Thanks for the comments, I believe I addressed all of them. The only other change I was considering was to switch how `Reactor.cpp` finds the bulk phase offset from the current approach that assumes that species names are unique among all phases to the pointer comparison, as is done in the 1-D code. Any thoughts?

speth:
Thanks, that looks good to me. I'm not too concerned about that assumption -- I think you'd have a whole bunch of other problems if you re-used species names between phases in an `InterfaceKinetics` object.

speth (approved):

PR #904, opened 2020-07-13
Limit ci runtime / cancel redundant jobs

ischoegl:
**Changes proposed in this pull request**

Same as #894 (after force-pushing while PR was closed I can no longer reopen)

- reduce time-out limit
- cancel redundant jobs

**If applicable, fill in the issue number this pull request is fixing**

Fixes #893 

**Checklist**

- [x] There is a clear use-case for this code change
- [x] The commit message has a short title & references relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] The pull request is ready for review

**Thoughts**

* https://github.com/technote-space/auto-cancel-redundant-job
* https://docs.github.com/en/actions/reference/events-that-trigger-workflows#repository_dispatch

Per @bryanwweber's comment 

> One of the things I've seen in some of the Issues on these actions is is that the person doing the push has to have enough rights to cancel the workflow run in the GUI for them to work correctly. I saw a few comments recommending alternate actions that appeared to work better, but I can't recall where anymore

Issue #605, opened 2019-03-04
Setting T and P for PureFluid heptane fails at low temperatures


speth:
### Cantera version
2.4.0 and 2.5.0a2

### Steps to reproduce
Run the following in Python:

```py
p = ct.Heptane()
p.TP = 204.4, ct.one_atm
```
Generates the exception:
```
CanteraError thrown by Substance::set_TPp:
no convergence for P* = 6.19745e-07, V* = 0.257196
```
Setting the state to a temperature of 204.5 or higher works fine. The minimum temperature for the model is supposed to be 182.56 K.

ischoegl:
I eventually ran into this issue while working on #907 (`Substance::Tsat` should raise an error for a pressure lower than the triple point pressure, which can be found by probing the triple point temperature); curiously, all other representations of `tpx` objects work down to the minimum temperature (I.e. the triple point temperature). 

The `Substance::set_TPp` method could likely be replaced by a more conventional root finder (there appear to be quite a few parameters with ‘tuned’ switches that are difficult to interpret and may cause hung loops), but there’s also the possibility of an issue with property data (e.g. a typo) causing the failure documented in this issue.

ischoegl:
Fwiw, I checked the values of numerical constants against the original source, and couldn't find a discrepancy (other than a minute difference for the value of Pc used in equation S-2, which appears to be inconsequential).

ischoegl:
Making a note here that #907 deferred some fixes due to this issue (marked as `@TODO`):

* `interfaces/cython/cantera/test/test_purefluid.py`: in `test_saturation_near_limits(self)`
* `src/tpx/Sub.cpp`: in `double Substance::Tsat(double p)`

PR #907, opened 2020-07-30
Fix edge cases for saturated PureFluid states / illustrate saturated water properties

### LINKED ISSUE #906: Assigning of critical point via TQ is not consistent (e.g. ct.Water)
**Problem description**

Cantera does not allow for setting of the state to the critical point using the `TQ` setter (which ~is~ *could be* legitimate: `Q` does not matter at the critical point). The error thrown is clearly not factual as the state is not supercritical (yet).

**Steps to reproduce**

```
In [1]: import cantera as ct
   ...: w = ct.Water()
   ...: w.TQ = w.critical_temperature, 0
   ...: 
<ipython-input-1-2b22a034afd6> in <module>()
      1 import cantera as ct
      2 w = ct.Water()
----> 3 w.TQ = w.critical_temperature, 0

/src/interfaces/cython/cantera/thermo.pyx in cantera._cantera.PureFluid.TQ.__set__()
   1757             T = values[0] if values[0] is not None else self.T
   1758             Q = values[1] if values[1] is not None else self.Q
-> 1759             self.thermo.setState_Tsat(T, Q)
   1760 
   1761     property PX:

CanteraError: 
CanteraError thrown by Substance::Set:
Can't set vapor fraction above the critical point

In [2]: w.TQ = w.critical_temperature - 1.e-9, 0

In [3]: w.TQ = w.min_temp, 0
```
I.e. a marginal reduction of temperature works. The issue does not exist at the low end (i.e. the triple point).

**System information**

- Cantera version: 2.5.0b1
- OS: ubuntu 18.04
- Python/MATLAB/other software versions: 3.6.9

### LINKED ISSUE #915: Bug for saturation pressure at critical point
**Problem description**

The calculation of saturation temperature/pressure has bugs/raises errors at the critical point.

**Steps to reproduce**

```
In [1]: import cantera as ct
   ...: w = ct.Water()

In [2]: w.T_sat
Out[2]: 373.1772329354826

In [3]: w.P_sat
Out[3]: 3528.213802290444

In [4]: w.TP = w.critical_temperature, w.critical_pressure

In [5]: w.TP
Out[5]: (647.286, 22088999.992234096)

In [6]: w.P_sat # <--- produces incorrect result (equal to last calculation)
Out[6]: 3528.213802290444

In [7]: w.T_sat # <--- raises error (instead of returning critical temperature)
<ipython-input-7-1b90fdda2181> in <module>()
----> 1 w.T_sat

/src/interfaces/cython/cantera/thermo.pyx in cantera._cantera.ThermoPhase.T_sat.__get__()
   1223         """Saturation temperature [K] at the current pressure."""
   1224         def __get__(self):
-> 1225             return self.thermo.satTemperature(self.P)
   1226 
   1227     ######## Methods to get/set the complete thermodynamic state ########

CanteraError: 
CanteraError thrown by Substance::Ps:
illegal temperature value 647.2860009464163

In [8]: w.P_sat # <--- different from previous call (!), and very close
Out[8]: 22088999.992961578
```

**System information**

- Cantera version: 2.5.0b1
- OS: Ubuntu 18.04
- Python version: 3.6.9

**Additional context**

Bug surfaced while finalizing #907; it appears to be related to the behavior that @speth encountered when investigating the initial version of the PR.

### LINKED ISSUE #917: PureFluid TP setter does not detect saturated mixtures
**Problem description**

Using a `TP` setter for a saturated mixture should throw an exception, but does not.

**Steps to reproduce**

```
In [1]: import cantera as ct
   ...: w = ct.Water()
   ...: w.TQ = 400, .5

In [2]: psat = w.P_sat

In [3]: w.TP = 400, psat # <--- this should raise an exception

In [4]: w.Q, w.phase_of_matter # <--- quality is changed from previous state
Out[4]: (0.0, 'gas')
```

There's a secondary issue above where `phase_of_matter` is clearly incorrect (see #919).

**Behavior**

**System information**

- Cantera version: 2.5.0b1
- OS: Ubuntu 18.04
- Python: 3.6.9

**Additional context**

Issue surfaced while resolving #907, but goes beyond.

### LINKED ISSUE #919: Bug in phase_of_matter output for saturated vapor/liquid
**Problem description**

The output of `phase_of_matter` is incorrect for saturated vapor and liquid.

**Steps to reproduce / Behavior**

```
In [1]: import cantera as ct
   ...: w = ct.Water()
   ...: w.TP
Out[1]: (300.0, 101324.99996785645)

In [2]: w.phase_of_matter, w.Q
Out[2]: ('liquid', 0.0)

In [3]: w.TQ = 400, 0 # saturated liquid
   ...: w.phase_of_matter, w.Q # <--- yields wrong phase of matter
Out[3]: ('gas', 0.0)

In [4]: w.TQ = 400, 1 # saturated vapor
   ...: w.phase_of_matter, w.Q # <--- yields wrong phase of matter
Out[4]: ('liquid', 1.0)

In [5]: w.TQ = 400, .5 # saturated mixture
   ...: w.phase_of_matter, w.Q # <--- expected output
Out[5]: ('liquid-gas-mix', 0.5)

```

**System information**

- Cantera version: 2.5.0b1
- OS: Ubuntu 18.04
- Python/MATLAB/other software versions: 3.6.9

**Additional context**

Bug also visible in #917; fix (and initial discussion) included in #907

The issue is created to facilitate discussion whether `phase_of_matter` should be set to `liquid-gas-mix` for edge cases (see discussion in #907), or `liquid` for `Q=0` and `gas` for `Q=1`. My personal interpretation is the latter, which is based on the definition of `Q` itself.

One additional comment pertains to the current implementation where various properties (e.g. `cp_mass`, `thermal_expansion_coeff`, `isothermal_compressibility`) are defined for edge cases based on a single phase, whereas they return `inf` for the `liquid-gas-mix` region.

### LINKED ISSUE #920: Bug in property calculations for saturated liquid/vapor
**Problem description**

The implementation of property calculations for edge cases (saturated liquid / vapor) is incorrect.

**Steps to reproduce / Behavior**

```
In [1]: import cantera as ct
   ...: w = ct.Water()
   ...: w.TP
Out[1]: (300.0, 101324.99996785645)

In [2]: w.cp_mass, w.thermal_expansion_coeff, w.isothermal_compressibility
Out[2]: (4180.788860505055, 0.0002739534956286111, 4.5370965768089256e-10)

In [3]: w.TP = 300, w.P_sat + 1 # compressed liquid
   ...: w.cp_mass, w.thermal_expansion_coeff, w.isothermal_compressibility
Out[19]: (4181.302071450938, 0.0002736968713794185, 4.5385398006526666e-10)

In [4]: w.TQ = 300, 0 # saturated liquid
   ...: w.cp_mass, w.thermal_expansion_coeff, w.isothermal_compressibility
Out[4]: (81268775.17346017, 66.66325231711855, 110675.13197217113)

In [5]: w.TQ = 300, 1 # saturated vapor
   ...: w.cp_mass, w.thermal_expansion_coeff, w.isothermal_compressibility
Out[5]: (81266465.29549846, 66.66325263277206, 2.834506650014112)

In [6]: w.TP = 300, w.P_sat - 1 # superheated vapor
   ...: w.cp_mass, w.thermal_expansion_coeff, w.isothermal_compressibility
Out[6]: (1871.4780865974458, 0.0033559522559820127, 0.00028393341634016384)
```
I.e. values jump to unreasonable values for edge cases (where properties are calculated based on one-sided differentials).

**System information**

- Cantera version: 2.5.0b1
- OS: Ubuntu 18.04
- Python/MATLAB/other software versions: 3.6.9

**Additional context**

Documenting another bug found while investigating #907. There is a question whether these properties should be calculated at all (see #919)..

**PS:** The `cv` calculation is likewise affected (calculations for saturated mixtures yield unreasonable results).

ischoegl:

- fix `TQ`/`PQ` assignment at critical point
- add example illustrating saturated water properties / vapor dome
- fix various bugs for edge case calculations (critical point / triple point)
- disable `TP` setter for saturated mixtures

Fixes #906, fixes #915, fixes #917, fixes #919, fixes #920

**Checklist**

**Thoughts**

I generated a digital version of a saturated steam table for an exam; i am including it here as it could be considered as an example and potentially expanded upon (otherwise I’ll simply drop it).

### DIFF
--- a/interfaces/cython/cantera/test/test_purefluid.py
+++ b/interfaces/cython/cantera/test/test_purefluid.py
@@ -180,9 +180,9 @@ class TestPureFluid(utilities.CanteraTest):
         self.assertNear(self.water.T, self.water.critical_temperature)
 
         # Supercritical
-        with self.assertRaisesRegex(ct.CanteraError, 'above the critical point'):
+        with self.assertRaisesRegex(ct.CanteraError, 'supercritical'):
             self.water.TQ = 1.001 * self.water.critical_temperature, 0.
-        with self.assertRaisesRegex(ct.CanteraError, 'above the critical point'):
+        with self.assertRaisesRegex(ct.CanteraError, 'supercritical'):
             self.water.PQ = 1.001 * self.water.critical_pressure, 0.
 
         # Q negative
@@ -201,6 +201,27 @@ class TestPureFluid(utilities.CanteraTest):
         self.water.TP = 300, ct.one_atm
         with self.assertRaisesRegex(ct.CanteraError, 'Saturated mixture detected'):
             self.water.TP = 300, self.water.P_sat
+
+        # Saturated vapor
+        self.water.TQ = 373.15, 1.
+        self.assertEqual(self.water.phase_of_matter, 'gas')
+        self.assertFalse(np.isinf(self.water.cp_mass))
+        self.assertFalse(np.isinf(self.water.thermal_expansion_coeff))
+        self.assertFalse(np.isinf(self.water.isothermal_compressibility))
+
+        # Saturated mixture
+        self.water.TQ = 373.15, .5
+        self.assertEqual(self.water.phase_of_matter, 'liquid-gas-mix')
+        self.assertTrue(np.isinf(self.water.cp_mass))
+        self.assertTrue(np.isinf(self.water.thermal_expansion_coeff))
+        self.assertTrue(np.isinf(self.water.isothermal_compressibility))
+
+        # Saturated liquid
+        self.water.TQ = 373.15, 0.
+        self.assertEqual(self.water.phase_of_matter, 'liquid')
+        self.assertFalse(np.isinf(self.water.cp_mass))
+        self.assertFalse(np.isinf(self.water.thermal_expansion_coeff))
+        self.assertFalse(np.isinf(self.water.isothermal_compressibility))
 
     def test_saturation_near_limits(self):
         # Low temperature limit (triple point)
--- a/interfaces/cython/cantera/test/test_purefluid.py
+++ b/interfaces/cython/cantera/test/test_purefluid.py
@@ -259,9 +259,10 @@ class TestPureFluid(utilities.CanteraTest):
         with self.assertRaisesRegex(ct.CanteraError, 'Illegal temperature'):
             self.water.TP = .999 * self.water.min_temp, ct.one_atm
             self.water.P_sat
-        with self.assertRaisesRegex(ct.CanteraError, 'Illegal pressure value'):
-            self.water.TP = 300, .999 * psat
-            self.water.T_sat
+        # Test disabled pending fix of GitHub issue #605
+        # with self.assertRaisesRegex(ct.CanteraError, 'Illegal pressure value'):
+        #     self.water.TP = 300, .999 * psat
+        #     self.water.T_sat
 
     def test_TPQ(self):
         self.water.TQ = 400, 0.8
--- a/interfaces/cython/cantera/test/test_purefluid.py
+++ b/interfaces/cython/cantera/test/test_purefluid.py
@@ -197,6 +197,11 @@ class TestPureFluid(utilities.CanteraTest):
         with self.assertRaisesRegex(ct.CanteraError, 'Invalid vapor fraction'):
             self.water.PQ = ct.one_atm, 1.001
 
+    def test_saturated_mixture(self):
+        self.water.TP = 300, ct.one_atm
+        with self.assertRaisesRegex(ct.CanteraError, 'Saturated mixture detected'):
+            self.water.TP = 300, self.water.P_sat
+
     def test_saturation_near_limits(self):
         # Low temperature limit (triple point)
         self.water.TP = 300, ct.one_atm
--- a/interfaces/cython/cantera/test/test_purefluid.py
+++ b/interfaces/cython/cantera/test/test_purefluid.py
@@ -164,12 +164,6 @@ class TestPureFluid(utilities.CanteraTest):
             self.water.TD = T, 0.1
             self.assertNear(T * self.water.thermal_expansion_coeff, 1.0, 1e-2)
 
-    def test_fd_properties_twophase(self):
-        self.water.TQ = 400, 0.1
-        self.assertEqual(self.water.cp, np.inf)
-        self.assertEqual(self.water.isothermal_compressibility, np.inf)
-        self.assertEqual(self.water.thermal_expansion_coeff, np.inf)
-
     def test_quality_exceptions(self):
         # Critical point
         self.water.TP = 300, ct.one_atm
@@ -202,26 +196,33 @@ class TestPureFluid(utilities.CanteraTest):
         with self.assertRaisesRegex(ct.CanteraError, 'Saturated mixture detected'):
             self.water.TP = 300, self.water.P_sat
 
+        w = ct.Water()
+
         # Saturated vapor
         self.water.TQ = 373.15, 1.
         self.assertEqual(self.water.phase_of_matter, 'gas')
-        self.assertFalse(np.isinf(self.water.cp_mass))
-        self.assertFalse(np.isinf(self.water.thermal_expansion_coeff))
-        self.assertFalse(np.isinf(self.water.isothermal_compressibility))
+        w.TP = self.water.T, .999 * self.water.P_sat
+        self.assertNear(self.water.cp, w.cp, 1.e-3)
+        self.assertNear(self.water.cv, w.cv, 1.e-3)
+        self.assertNear(self.water.thermal_expansion_coeff, w.thermal_expansion_coeff, 1.e-3)
+        self.assertNear(self.water.isothermal_compressibility, w.isothermal_compressibility, 1.e-3)
 
         # Saturated mixture
         self.water.TQ = 373.15, .5
         self.assertEqual(self.water.phase_of_matter, 'liquid-gas-mix')
-        self.assertTrue(np.isinf(self.water.cp_mass))
-        self.assertTrue(np.isinf(self.water.thermal_expansion_coeff))
-        self.assertTrue(np.isinf(self.water.isothermal_compressibility))
+        self.assertEqual(self.water.cp, np.inf)
+        self.assertTrue(np.isnan(self.water.cv))
+        self.assertEqual(self.water.isothermal_compressibility, np.inf)
+        self.assertEqual(self.water.thermal_expansion_coeff, np.inf)
 
         # Saturated liquid
         self.water.TQ = 373.15, 0.
         self.assertEqual(self.water.phase_of_matter, 'liquid')
-        self.assertFalse(np.isinf(self.water.cp_mass))
-        self.assertFalse(np.isinf(self.water.thermal_expansion_coeff))
-        self.assertFalse(np.isinf(self.water.isothermal_compressibility))
+        w.TP = self.water.T, 1.001 * self.water.P_sat
+        self.assertNear(self.water.cp, w.cp, 1.e-3)
+        self.assertNear(self.water.cv, w.cv, 1.e-3)
+        self.assertNear(self.water.thermal_expansion_coeff, w.thermal_expansion_coeff, 1.e-3)
+        self.assertNear(self.water.isothermal_compressibility, w.isothermal_compressibility, 1.e-3)
 
     def test_saturation_near_limits(self):
         # Low temperature limit (triple point)
--- a/interfaces/cython/cantera/test/test_purefluid.py
+++ b/interfaces/cython/cantera/test/test_purefluid.py
@@ -118,10 +118,10 @@ class TestPureFluid(utilities.CanteraTest):
         self.assertNear(k1, k2, tol)
         self.assertNear(alpha1, alpha2, tol)
 
-        # calculating these finite difference properties should not perturbe the
-        # state of the object
-        self.assertEqual(h1a, h1b)
-        self.assertEqual(h2a, h2b)
+        # calculating these finite difference properties should not perturb the
+        # state of the object (except for checks on edge cases)
+        self.assertNear(h1a, h1b, 1e-9)
+        self.assertNear(h2a, h2b, 1e-9)
 
     def test_properties_near_min(self):
         self.check_fd_properties(self.water.min_temp*(1+1e-5), 101325,
@@ -163,6 +163,21 @@ class TestPureFluid(utilities.CanteraTest):
         for T in [440, 550, 660]:
             self.water.TD = T, 0.1
             self.assertNear(T * self.water.thermal_expansion_coeff, 1.0, 1e-2)
+
+    def test_pq_setter_triple_check(self):
+        self.water.PQ = 101325, .2
+        T = self.water.T
+        # change T such that it would result in a Psat larger than P
+        self.water.TP = 400, 101325
+        # ensure that correct triple point pressure is recalculated
+        # (necessary as this value is not stored by the C++ base class)
+        self.water.PQ = 101325, .2
+        self.assertNear(T, self.water.T, 1e-9)
+        with self.assertRaisesRegex(ct.CanteraError, 'below triple point'):
+            # min_temp is triple point temperature
+            self.water.TP = self.water.min_temp, 101325
+            P = self.water.P_sat # triple-point pressure
+            self.water.PQ = .999*P, .2
 
     def test_quality_exceptions(self):
         # Critical point
--- a/interfaces/cython/cantera/test/test_purefluid.py
+++ b/interfaces/cython/cantera/test/test_purefluid.py
@@ -274,7 +274,7 @@ class TestPureFluid(utilities.CanteraTest):
         with self.assertRaisesRegex(ct.CanteraError, 'Illegal temperature'):
             self.water.TP = .999 * self.water.min_temp, ct.one_atm
             self.water.P_sat
-        # Test disabled pending fix of GitHub issue #605
+        # @TODO: test disabled pending fix of GitHub issue #605
         # with self.assertRaisesRegex(ct.CanteraError, 'Illegal pressure value'):
         #     self.water.TP = 300, .999 * psat
         #     self.water.T_sat
--- a/interfaces/cython/cantera/test/test_purefluid.py
+++ b/interfaces/cython/cantera/test/test_purefluid.py
@@ -34,6 +34,8 @@ class TestPureFluid(utilities.CanteraTest):
         self.water.TV = 400, 1.45
         self.assertNear(self.water.T, 400)
         self.assertNear(self.water.v, 1.45)
+        with self.assertRaisesRegex(ct.CanteraError, 'Negative specific volume'):
+            self.water.TV = 300, -1.
 
         self.water.PV = 101325, 1.45
         self.assertNear(self.water.P, 101325)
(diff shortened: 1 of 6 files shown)


bryanwweber:
FYI the failures in the `multiple-sundials` job are fixed by https://github.com/Cantera/cantera/pull/905/commits/529b7990987ca2e5cfc8fc168e6b5d47845683a1 in #905

speth (changes requested):
Unfortunately, I think fixing this problem is more complicated than just changing the comparisons in the state setters. With this change, the state of the mixture when setting the state to be a saturated state at the critical temperature does not end up at the critical point, and the resulting state depends on the previous state of the mixture. For example:
```py
>>> w = ct.Water()
>>> w.TP = 300, 101325
>>> w.TQ = w.critical_temperature, 0
>>> w.TP
(647.286, 1089571723.7984445) # not even close
>>> w.TQ = w.critical_temperature, 1
>>> w.TP
(647.286, 7623.02783606933) # wrong, and different
>>> w.TP = 645, 101325
>>> w.TQ = w.critical_temperature, 0
(647.286, 22304442.572488125) # closer, but still wrong
```

ischoegl:
That’s actually much worse than what I recall for the tests I ran - I saw an insignificant discrepancy depending on whether I set Q equal to 0 or 1, but nothing of that magnitude. Based on what you are showing, this is likely due to having marched along saturated liquid and vapor lines, meaning that I was reasonably close to the correct result, which led me to believe that the simple fix worked as expected. *Update: located the bug - see #915 - and added the fix to this PR.*

~Since this is just a papercut for an edge case, I’d suggest to close both issue and PR~ - if there’s interest in the example, I could still modify it as it’s not too difficult to circumnavigate. There’s also the question about the proposed change making physical sense, while being mathematically ill defined (there is no unique value for Q).

ischoegl:
**PS:** it may still make sense to change the error message, as it clearly doesn’t apply to the critical point itself; I’ll likely push a suggestion before long.

bryanwweber:
> if there’s interest in the example, I could still modify it as it’s not too difficult to circumnavigate.

I think the example is valuable, so I'd be in favor of fixing it up.

> There’s also the question about the proposed change making physical sense, while being mathematically ill defined (there is no unique value for Q).

This was also my concern about the proposed change that I never got around to writing up.

ischoegl:
@speth / @bryanwweber ... I replaced the 'fix' by just changing the error message, and updated the example.

On a separate note, I think that this is the last pending PR from my side - unless there's interest in resurrecting #888 (a simplification of `.gitignore` that has no other technical contribution). #904 is something that I likely cannot push further, and the two 'stubs' depend on other work (or clarifications).

**PS:** I haven't forgotten about Cantera/cantera-website#104 and Cantera/cantera-jupyter#27, which I'll get to one of these days.

**Update:** Adding a unit test surfaced a previously unknown issue for the `PQ` setter (the exception was actually not reachable), which is now fixed. *Edit: another issue surfaced with #915.*

ischoegl:
Added fixes for #915.

This should open the option to allow for `TQ`/`PQ` convenience setters at the critical point after all (with the mathematical caveat that the value of `Q` would not matter). At this point, I am leaning against it.

ischoegl:
I added a couple of tests to the suite. ~There is one last edge case left for discussion: at the critical point, `Q` can be read, although it cannot be specified in setters and a unique value cannot be defined.~ *Nevermind: this is a larger issue, to be discussed in #916.*

Work on this PR is done, except for review.

speth:
> There’s also the question about the proposed change making physical sense, while being mathematically ill defined (there is no unique value for Q).

I'm not convinced that it's mathematically ill defined just because any value of Q corresponds to the same state. I think it's just that there's no longer a natural inverse (that is, a way to determine Q from the intrinsic state) based on the physical definition of Q. That doesn't preclude defining Q in a consistent way at the critical point or even outside the vapor dome.

Of course, we wouldn't want to add this complication unless it were useful, but I think the `vapordome.py` example makes a case for being able to set the state to the Tcrit and any value of Q. Otherwise, you're forced to jump through hoops in order to traverse the whole saturation curve.

ischoegl:
@speth ... the convenience argument is what made me file the issue and this PR in the first place, so it’s not difficult to persuade me here :wink: ... I’ll push an update shortly.

Regarding definitions, it’s no longer possible to differentiate liquid from vapor at the critical point, so it actually becomes impossible to define Q. It’s a different discussion (#916) though and won’t affect the implementation here.

ischoegl:
@speth ... `TQ`/`PQ` setters are now enabled for the critical point. 

I found quite a few irregularities of various solvers for edge cases while working on this PR, and fixed whatever popped up (it didn’t quite extend to #605 though). I also added unit tests to ensure proper behavior.

ischoegl:
First time the example runner revealed an issue :tada: (`rankine.py` failed as `cp` couldn't be evaluated for a saturated liquid/vapor). Disabling the `TP` setters made some minor changes of the logic necessary, but I believe it's clearer now. 

There also was a glitch in `phase_of_matter`, which returned 'liquid' for saturated vapor. Also fixed ...

ischoegl:
@speth / @bryanwweber ... I think this is (finally) done. Not sure about the MSVC runner failing, but it doesn't look related to work on this PR.

src/thermo/PureFluidPhase.cpp
```diff
@@ -90,7 +90,7 @@ std::string PureFluidPhase::phaseOfMatter() const
         return "supercritical";
     } else if (m_sub->TwoPhase() == 1) {
         return "liquid-gas-mix";
-    } else if (pressure() < m_sub->Ps()) {
+    } else if (m_sub->x() == 1) {
```
bryanwweber: Just curious how this change is any different than the existing condition. What was the problem with the existing code? I guess it was returning the wrong value in a few cases, but which cases and why does this fix it? I also think that the routine should return `liquid-gas-mix` if the state is exactly the saturated vapor.

ischoegl: @bryanwweber ... if you set the state to saturated vapor, i.e. `Q = 1`, the old implementation returned `liquid` as the pressure is still `Ps()`. I didn’t feel like opening another issue, and decided to just fix it. There's also a unit test now, just to make sure.

Regarding your comment, this was my initial reaction/fix also, but I then realized that several properties still make sense as being defined for saturated vapor or liquid (e.g. `cp`), but not the saturated mixture. I guess this is debatable, but at least this is how things were implemented; I didn't want to change this in the end as someone in the past obviously made a point of making those calculations possible.

bryanwweber: > if you set the state to saturated vapor, I.e. `Q = 1`, the old implementation returned `liquid`

Yes, but why? :smile: Like I said, my preference would be to return `liquid-gas-mix` for the saturated vapor, I think that's the typical definition. The change here would rely on the quality being defined as equal to 1 for gases, which your other issue #916 would change...

ischoegl: Correct - #916 would change it, in which case the correct check would be
``` C++
} else if (pressure() < m_sub->Ps() ||  m_sub->x() == 1) {
```

Regarding the definition, my first reaction was the same as yours, until the existing implementation reminded me of being able to define `cp`, etc. for the saturated lines based on one-sided differentials. Q=0 and 1 *are* liquid and vapor, after all, so it's not a clear cut case.

ischoegl: **PS:** this is not a major sticking point for me. If we go with `liquid-gas-mix`, `cp` and a couple of other properties should likely return `nan` for the edge cases for consistency's sake?

ischoegl: **PPS:** On further consideration, it would be odd to declare a "saturated *liquid*" with `Q=0` to be a `liquid-vapor-mix` instead of the `liquid` that defines the term; same goes for a "saturated *vapor*" ... so I *think* the current implementation makes the most sense?

ischoegl: @bryanwweber ... I created #919 to facilitate the discussion.

ischoegl:
This PR keeps giving, as there's another pre-existing issue that came up (#920).

~#920 remains to be addressed. I just checked and the current implementation has to be revised somewhat more (there should also be unit tests). Will wait for clarification of #919.~

ischoegl:
#920 is fixed now (#605 required a workaround). ~Another fix may be needed as `Heptane.T_sat` is still broken as a side-effect of #605).~

ischoegl:
@speth / @bryanwweber ... I’m finally happy with most of this PR, although some minor warts remain due to #605 (which I’m not volunteering for at the moment).

src/thermo/PureFluidPhase.cpp
ischoegl: Reverting to the old implementation, while updating `TwoPhase` to include boundaries (a new flag `strict` excludes boundaries to enable property calculations).

ischoegl:
Minor change to the behavior of saturated liquid and vapor lines - I added them to the `liquid-vapor-mix` per @bryanwweber's recommendation. The behavior is now consistent with CoolProp.

ischoegl:
Rebased, now that 2.5.0b1 is out :tada:

src/tpx/Sub.cpp
```diff
@@ -387,9 +449,10 @@ void Substance::Set(PropertyPair::type XY, double x0, double y0)
         if (y0 > 1.0 || y0 < 0.0) {
             throw CanteraError("Substance::Set",
                                "Invalid vapor fraction, {}", y0);
-        } else if (x0 >= Tcrit()) {
+        } else if (x0 < Tmin() || x0 > Tcrit()) {
             throw CanteraError("Substance::Set",
-                               "Can't set vapor fraction above the critical point");
+                               "Illegal temperature value: {} "
+                               "(supercritical of below triple line)", x0);
```
speth:
```suggestion
                               "(supercritical or below triple point)", x0);
```

src/tpx/Sub.cpp
```diff
@@ -370,14 +430,16 @@ void Substance::Set(PropertyPair::type XY, double x0, double y0)
...
+        temp = Tmin();
         if (y0 > 1.0 || y0 < 0.0) {
             throw CanteraError("Substance::Set",
                                "Invalid vapor fraction, {}", y0);
-        } else if (temp >= Tcrit()) {
+        } else if (x0 < Ps() || x0 > Pcrit()) {
             throw CanteraError("Substance::Set",
-                               "Can't set vapor fraction above the critical point");
+                               "Illegal pressure value: {} (supercritical "
+                               "or below triple line)", x0);
```
speth:
```suggestion
                               "or below triple point)", x0);
```

src/tpx/Sub.cpp
```diff
@@ -370,14 +430,16 @@ void Substance::Set(PropertyPair::type XY, double x0, double y0)
                x0, y0, TolAbsS, TolAbsH, TolRel, TolRel);
         break;
     case PropertyPair::PX:
-        temp = Tsat(x0);
+        temp = Tmin();
         if (y0 > 1.0 || y0 < 0.0) {
             throw CanteraError("Substance::Set",
                                "Invalid vapor fraction, {}", y0);
-        } else if (temp >= Tcrit()) {
+        } else if (x0 < Ps() || x0 > Pcrit()) {
```
speth: This condition isn't checking the right value of `P_sat` - I think you need to calculate `P_sat` at `Tmin()`. Example:
```python
In [37]: w = ct.Water()

In [38]: w.PQ = 101325, 0.2

In [39]: w.TP = 400, 101325

In [40]: w.PQ = 101325, 0.2
<ipython-input-40-71304f9c35a7> in <module>
----> 1 w.PQ = 101325, 0.2

C:\src\cantera\interfaces\cython\cantera\thermo.pyx in cantera._cantera.PureFluid.PQ.__set__()
   1783             P = values[0] if values[0] is not None else self.P
   1784             Q = values[1] if values[1] is not None else self.Q
-> 1785             self.thermo.setState_Psat(P, Q)
   1786
   1787     property ST:

CanteraError:
CanteraError thrown by Substance::Set:
Illegal pressure value: 101325.0 (supercritical or below triple line)
```
To avoid recalculating that value every time the `PX` setter is used, we could calculated it once and save it when the substance is instantiated.

ischoegl:
Inactive for 3 months - closing.

ischoegl:
Based on a UG post there may be value in wrapping this up after all

speth (changes requested):
Besides a few minor suggestions for the error messages, I think there's just one case where these new checks aren't doing the right thing.

src/tpx/Sub.cpp
ischoegl: Absolutely ... while under the vapor dome it's a line, the term 'triple line' is not a thing ...

src/tpx/Sub.cpp
ischoegl: Ah ... the temperature is set to `Tmin()` on line 433 right above, which in theory would ensure a correct execution, but I figure I forgot to call `set_T` on that.

Regarding pre-calculation: the constructor isn't really used at the moment, so there's not a good 'hook'. I'm also afraid that it will make #605 rear its head again (currently I don't have spare time to address).

ischoegl:
@speth ... thanks for the review, I believe I addressed everything.

Since triple point properties are currently not broken out (Python's current `min_temp` is actually `T_triple`, and there isn't an equivalent `min_pressure`), I'd suggest leaving a pre-calculation for another time. If you agree, I will open a feature request on Cantera/enhancements ...

**PS:** I can rebase once #946 is merged, so CI will give expected results.

ischoegl:
close/reopen to trigger checks

interfaces/cython/cantera/test/test_purefluid.py
```diff
@@ -162,12 +164,121 @@ def test_thermal_expansion_coeff_TD(self):
...
+            self.water.T_sat
+
+        # Below triple point
+        with self.assertRaisesRegex(ct.CanteraError, 'Illegal temperature'):
+            self.water.TP = .999 * self.water.min_temp, ct.one_atm
+            self.water.P_sat
+        # Test disabled pending fix of GitHub issue #605
+        # with self.assertRaisesRegex(ct.CanteraError, 'Illegal pressure value'):
+        #     self.water.TP = 300, .999 * psat
+        #     self.water.T_sat
```
speth: Can you mark this with `@todo`? It makes it easier to find things like this when searching for the code, even if the `@todo` label is only really treated specially in documentation read by Doxygen.

src/tpx/Sub.cpp
```diff
@@ -237,26 +292,29 @@ double Substance::x()
 
 double Substance::Tsat(double p)
 {
-    if (p <= 0.0 || p > Pcrit()) {
-        throw CanteraError("Substance::Tsat", "illegal pressure value");
+    double Tsave = T;
+    if (p <= 0. || p > Pcrit()) {
+        // TODO: this check should also fail below the triple point pressure
```
speth:
```suggestion
        // @TODO: this check should also fail below the triple point pressure
```

speth (approved):
I think this looks fine, and your suggestions for a few fine points that can resolved later are reasonable. My only suggestion is to flag a couple of these things in the code with `@todo` so they're easier to find later, although keeping track of them in the related Github issue is probably also a good idea.

ischoegl:
> I think this looks fine, and your suggestions for a few fine points that can resolved later are reasonable. My only suggestion is to flag a couple of these things in the code with `@todo` so they're easier to find later, although keeping track of them in the related Github issue is probably also a good idea.

@speth ... thanks! Your suggestions are straightforward (as long as there isn't a preference for upper/lowercase - I used `@TODO`).

bryanwweber (approved):

Issue #911, opened 2020-08-09
Poisson solver not activated in `IonFlameBase`


ischoegl:
**Problem description**

When running `IonFlameBase`-derived simulations, the solution involves two stages. In `IonFlameBase.solve`, the second stage involves 
https://github.com/Cantera/cantera/blob/29f6b94ed3ba490fba1b74e6c6e0cd328a02c917/interfaces/cython/cantera/onedim.py#L940-L941

However, `poisson_enabled` is not a valid attribute - presumably, the attribute was renamed to `electric_field_enabled` at some point.

**Steps to reproduce**

1. In iPython, run 
```
In [1]: %run ion_free_flame.py  
```
3. Check attribute
```
In [2]: f.electric_field_enabled
Out[2]: False
```

**Behavior**

The Poisson solver is apparently never used, see docstring
https://github.com/Cantera/cantera/blob/29f6b94ed3ba490fba1b74e6c6e0cd328a02c917/interfaces/cython/cantera/onedim.py#L920-L923

**System information**

- Cantera version: 2.5.0b4
- OS: all
- Python/MATLAB/other software versions: all

ischoegl:
Superseded by #913

PR #912, opened 2020-08-09
Some fixes for IonFlame issues

### LINKED ISSUE #909: Flawed upstream boundary condition in 1D IonFlow?
**Problem description**

There appears to be an inadequate upstream boundary condition in `IonFlow`. For charged species, the upstream boundary condition uses a von Neumann (zero gradient!) assumption, where the following rationale is provided:
https://github.com/Cantera/cantera/blob/ba13c652b45e74acac1daa929f66a3b6b3f92a63/src/oneD/IonFlow.cpp#L178-L185

There are obvious concerns about the physicality of this boundary condition; mathematically, this appears to impose von Neumann conditions for both inlet and outlet, which is inherently ill defined. Finally, the short-cut does not accommodate a case where a user could specify a non-zero upstream concentration of a charged species (where a convective term added in `Boundaries1D::eval` would be quite problematic).

While upstream gradients may be sufficiently small for this ad-hoc assumption to work for a free flame (i.e. `ion_free_flame.py`), it appears that this assumption is inadequate for burner-stabilized flames. However, the assumption does not appear to be a root cause for poor convergence (see _Additional Context_).

**System information**

- Cantera version: 2.5
- OS: Ubuntu 18.04
- Python/MATLAB/other software versions: 3.6

**Additional context**

The example `ion_burner_flame.py` exhibits extremely poor convergence with multiple errors thrown (recovered in `auto` solution mode)
```
CanteraError thrown by OneDim::timeStep:
Took maximum number of timesteps allowed (500) without reaching steady-state solution.
```

As this happens during stage 1 of the solution (i.e. using `IonFlow::frozenIonMethod`) where `m_flux` of charged species is set to zero (and the code section above is bypassed), the boundary condition is not a root cause. Commenting out the lines where the BC residual is set in `IonFlow::evalResidual` does indeed not change the behavior. 

Poor convergence is also observed for Cantera 2.4, i.e. the version where the example was initially introduced.

### LINKED ISSUE #910: Couldn't read phase node from XML file when loading NASA cti and xml files packaged with Cantera
**Problem description**

Creating a Solution using any of the NASA cti or xml files, nasa, nasa_gas, nasa_condensed, raises:
```
File "interfaces\cython\cantera\base.pyx", line 29, in cantera._cantera._SolutionBase.__cinit__
File "interfaces\cython\cantera\base.pyx", line 61, in cantera._cantera._SolutionBase._init_cti_xml
ValueError: Couldn't read phase node from XML file
```

**Steps to reproduce**

```
import cantera as ct
ct.Solution('nasa_gas.xml')
```

**Behavior**

Running 
```
import cantera as ct
ct.Solution('gri30.xml')
```
works as expected according to the Python tutorial (https://cantera.org/tutorials/python-tutorial.html). Similarly, using KOH, silane, or water also works, so I would expect to be able to do the same with any of the nasa files that come prepackaged with Cantera on install. I haven't altered any of the files in any way since installing Cantera.

**System information**

- Cantera version: 2.4.0
- OS: Windows 10
- Python: 3.7.6
- VS Code: 1.47.3

ischoegl:

- ~Apply band-aid fix to `ion_burner_flame.py` - a reduction of domain size makes the example more stable; there is no apparent reason for simulations to fail on a larger domain, which points at underlying issues that should be fixed at a later point~ __cherry-picked to #905__
- ~Fix `ch4_ion.yaml` that did not match `ch4_ion.cti`~ __cherry-picked to #905__
- Fix apparent glitch that results in the Poisson solver not being activated (Edit: *the fix has no effect as it affects C++ code that is not used*)
- Remove upstream von Neumann boundary condition for ionized species (see also #700)
- Add optional plots to `IonFlow`-based examples
- ~Fix export of `_other` columns from `IonFlow` (i.e. grid, etc.)~ __cherry-picked to #905__

Additional issues that surfaced while reviewing code are listed in #913 - changes are more extensive and should be handled by a separate PR (potentially after the release of 2.5).

**PS:** After cherry-picking 3 fixes into #905, the remaining changes are not worth keeping this PR open.

Fixes #909

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/examples/onedim/ion_burner_flame.py
+++ b/interfaces/cython/cantera/examples/onedim/ion_burner_flame.py
@@ -1,9 +1,10 @@
 """
 A burner-stabilized premixed methane-air flame with charged species.
 
-Requires: cantera >= 2.5.0
+Requires: cantera >= 2.5.0, matplotlib >= 2.0
 """
 
+import sys
 import cantera as ct
 
 p = ct.one_atm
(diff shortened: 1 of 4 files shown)


ischoegl:
@BangShiuh ... could you have a brief look at this (and the corresponding bug report in #911)?

ischoegl:
~On further inspection, the Python attribute `electric_field_enabled` (as well as the C++ counterparts) has no effect on the actual solution - the implementation is incomplete as the C++ vector `m_do_electric_field` is set but never used.~ **Edit:** this behavior is now reported as part of #913 - I am not planning to address it here.

ischoegl:
@bryanwweber ... I added plots to the examples, and fixed another glitch for the `SolutionArray` export of `IonFlow`. At this point, I think this PR ready for review.

ischoegl:
After cherry-picking 3 fixes into #905, the remaining changes are not worth keeping this PR open.

PR #809, opened 2020-02-16
Improve WaterSSTP helper class documentation.

24sharkS:
**Checklist**

- [x] There is a clear use-case for this code change
- [x] The commit message has a short title & references relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] The pull request is ready for review

**If applicable, fill in the issue number this pull request is fixing**

Fixes #646 

**Changes proposed in this pull request**

- Added non-instantiability property in WaterSSTP helper class documentation.

PR #921, opened 2020-08-27
Fix open WaterPropsIAPWS issues / add WaterSSTP as alternative Water backend

### LINKED ISSUE #577: PDSS Water calculation error at 287.9K
### Cantera version
2.4.0
### Operating System
Windows 7 (Linux also tested)
### Python/MATLAB version
3.6.5 (Anaconda distribution)

### Description of the problem
When trying to use the PDSS module for non-ideal mixtures, the calculation of the water density crashes at exactly 287.9 K regardless of the pressure used. Other temperatures work just fine.

A small model to check this is to get the xml file for the solution of NaCl from the repository:
[https://github.com/Cantera/cantera/blob/master/test_problems/VCSnonideal/NaCl_equil/HMW_NaCl.xml](url)
and changing a few lines of code (18-23) to disable the solute molalities. We change from:
```
      <soluteMolalities>
             Na+:6.0954
             Cl-:6.0954
             H+:2.1628E-9
             OH-:1.3977E-6
      </soluteMolalities>
```
to
```
      <!--soluteMolalities>
             Na+:6.0954
             Cl-:6.0954
             H+:2.1628E-9
             OH-:1.3977E-6
      </soluteMolalities-->
```

and we run this small code under jupyter notebook:
````
import numpy as np
import cantera as ct

liq = ct.Solution('HMW_NaCl.xml', 'NaCl_electrolyte') 
liq.TP = 287.9, 1e5
print(liq.report())
````
Then, the following error occurs:

> CanteraError                              Traceback (most recent call last)
> <ipython-input-9-4c3c5efe80a2> in <module>()
>       1 liq = ct.Solution('HMW_NaCl.xml', 'NaCl_electrolyte')
> ----> 2 liq.TP = 287.9, 1e5
>       3 print(liq.report())
> 
> interfaces\cython\cantera\thermo.pyx in cantera._cantera.ThermoPhase.TP.__set__()
> 
> CanteraError: 
> ***********************************************************************
> CanteraError thrown by WaterPropsIAPWS::corr:
> Error occurred trying to find liquid density at (T,P) = 287.9  1678.12
> ***********************************************************************
> 
We tried to track the function which is causing the error by comparing inputs and outputs of the failed iteration (which would work if we call the function separately). We tracked the error to the following files/classes/functions:
1)WaterPropsIAPWS.cpp -> void WaterPropsIAPWS::corr ->densLiq = _density_(temperature, pressure, WATER_LIQUID, densLiq)
2)WaterPropsIAPWS.cpp -> doublereal WaterPropsIAPWS::density ->_m_phi.dfind_(p_red, tau, deltaGuess)
3)WaterPropsIAPWSphi.cpp->doublereal WaterPropsIAPWSphi::dfind(doublereal p_red, doublereal tau, doublereal deltaGuess)

And there we did not go deeper. It would give no convergence when called with the previous code, but it would work if we did the calculation alone with the supposed input data of the corresponding failed iteration.

For more data or questions, we would be glad to give/answer them, respectively.

### LINKED ISSUE #646: Improve WaterSSTP helper class documentation
### Cantera version
2.4.0

### Operating System
All

### Python/MATLAB version
All

### Actual Behavior
At present, there are a number of classes in the `thermo` directory which provide supporting calculations to the `WaterSSTP` ThermoPhase class.  These classes:

- `WaterProps`
- `WaterPropsIAPWS`
- `WaterPropsIAPWSphi`

Are not instantiatable on their own, but provide low-level functionality necessary for property calculations in the aforementioned `WaterSSTP` class.

This aspect of these three files should be stated more explicitly, near the top of each of the relefvant `.cpp` and `.h` files, so that interested users can more easily locate the relevant thermoPhase class, and avoid wasting time trying to understand these helper classes, which are most likely not what they are interested in (directly) implementing.

### LINKED ISSUE #721: Setting the state of PureLiquidWater with T,P raises a CanteraError

### PR DESCRIPTION
**Pull request by ischoegl**
There are some open issues with `WaterPropsIAWSS`; adding access to `WaterSSTP` via an alternative backend to the `Water` wrapper facilitates interactions with underlying C++ objects.

The `WaterPropsIAPWS` helper classes implement a highly accurate Industrial Formulation for the equation of state for pure water, which is currently used by `WaterSSTP` and `PDSS_Water` classes. The former is loadable via YAML as part of the `ThermoFactory` approach, but only implements the liquid portion. Per @bryanwweber ’s observations (in #721):

> The `WaterSSTP` phase, as implied by the name in the XML file (`PureLiquidWater`), is intended to represent only the liquid states of water. It is based on the `WaterPropsIAPWS` class, which can represent liquid to vapor states, but the `WaterSSTP` class is used in a few places where liquids only are the assumption. Hypothetically, a different class that implements the full behavior of the EOS could be written using `WaterPropsIAPWS`, but this ain't it.

The current loader is cumbersome and involves a custom input file, e.g.
```
water = ct.PureFluid('test/data/liquid-water.xml')
```
from within the cantera source folder (see #721), i.e. suitable input files aren’t distributed outside the test suite. The proposed change enables loading of the `WaterSSTP` model via
```
water = ct.Water(backend='IAPWS95')
```
while the existing behavior remains the default. Beyond, this PR also fixes some open issues.

- Add phase using `liquid-water-IAPWS95` to `liquidvapor.yaml`
- Add an alternative `backend=iawps95` switch to the existing Python `Water` interface to facilitate usage to the `liquid-water-IAPWS95` thermo model; the existing behavior corresponds to `backend=default`
- Update Python docstrings with information already available in MATLAB wrapper classes
- Clarify role of `WaterSSTP` helper classes (based on @24sharkS 's solution in #809)
- Ensure that state of `WaterSSTP` remains liquid (or supercritical) - this checks the classical criterion for saturated conditions.

Fixes #577, fixes #646, fixes #721

**Related Issues**

Cantera/enhancements#6, #809 

**Checklist**

**Other thoughts**

1. Short-term, this fixes some issues and (hopefully) helps clarify the IAPWS formulation.
2. Long-term, it may make sense to add [CoolProp](https://github.com/CoolProp/CoolProp) objects as additional alternative `backend` s for existing wrapper objects, see Cantera/enhancements#35.

### DIFF
--- a/include/cantera/thermo/WaterProps.h
+++ b/include/cantera/thermo/WaterProps.h
@@ -83,8 +83,10 @@ class PDSS_Water;
 //! The WaterProps class is used to house several approximation routines for
 //! properties of water.
 /*!
- * This is a helper class for WaterSSTP and is not instantiable on its own
- * (see \ref thermoprops and class \link Cantera::WaterSSTP WaterSSTP\endlink).
+ * This is a helper class for WaterSSTP and PDSS_Water and does not constitute
+ * a complete implementation of a thermo phase by itself (see \ref thermoprops
+ * and classes \link Cantera::WaterSSTP WaterSSTP\endlink and
+ * \link Cantera::PDSS_Water PDSS_Water\endlink).
  *
  * The class is also a wrapper around the WaterPropsIAPWS class which provides
  * the calculations for the equation of state properties for water.
--- a/include/cantera/thermo/WaterProps.h
+++ b/include/cantera/thermo/WaterProps.h
@@ -83,6 +83,9 @@ class PDSS_Water;
 //! The WaterProps class is used to house several approximation routines for
 //! properties of water.
 /*!
+ * This is a helper class for WaterSSTP and is not instantiable on its own
+ * (see \ref thermoprops and class \link Cantera::WaterSSTP WaterSSTP\endlink).
+ *
  * The class is also a wrapper around the WaterPropsIAPWS class which provides
  * the calculations for the equation of state properties for water.
  *
--- a/include/cantera/thermo/WaterSSTP.h
+++ b/include/cantera/thermo/WaterSSTP.h
@@ -140,7 +140,7 @@ public:
     explicit WaterSSTP(XML_Node& phaseRef, const std::string& id = "");
 
     virtual std::string type() const {
-        return "WaterSSTP";
+        return "liquid-water-IAPWS95";
     }
 
     virtual std::string phaseOfMatter() const;
@@ -241,6 +241,12 @@ public:
     }
 
     //! Switch that enables calculations in the gas phase
+    /**
+     *  Since this phase represents a liquid (or supercritical) phase, it is an
+     *  error to return a gas-phase answer. The sole intended use for this
+     *  member function is to check the thermodynamic consistency of the
+     *  underlying WaterProps class with ideal-gas thermo functions.
+     */
     void allowGasPhase(bool flag) { m_allowGasPhase = flag; }
 
 protected:
--- a/include/cantera/thermo/WaterSSTP.h
+++ b/include/cantera/thermo/WaterSSTP.h
@@ -140,7 +140,7 @@ public:
     explicit WaterSSTP(XML_Node& phaseRef, const std::string& id = "");
 
     virtual std::string type() const {
-        return "Water";
+        return "WaterSSTP";
     }
 
     virtual std::string phaseOfMatter() const;
--- a/include/cantera/thermo/WaterSSTP.h
+++ b/include/cantera/thermo/WaterSSTP.h
@@ -238,6 +238,9 @@ public:
         return m_waterProps.get();
     }
 
+    //! Switch that enables calculations in the gas phase
+    void allowGasPhase(bool flag) { m_allowGasPhase = flag; }
+
 protected:
     /**
      * @internal This internal routine must be overridden because it is not
--- a/include/cantera/thermo/WaterSSTP.h
+++ b/include/cantera/thermo/WaterSSTP.h
@@ -20,7 +20,7 @@ namespace Cantera
 class WaterPropsIAPWS;
 class WaterProps;
 //! Class for single-component water. This is designed to cover just the liquid
-//! part of water.
+//! and supercritical phases of water.
 /*!
  * The reference is W. Wagner, A. Pruss, "The IAPWS Formulation 1995 for the
  * Thermodynamic Properties of Ordinary Water Substance for General and
@@ -142,6 +142,8 @@ public:
     virtual std::string type() const {
         return "Water";
     }
+
+    virtual std::string phaseOfMatter() const;
 
     //! @name  Molar Thermodynamic Properties of the Solution
     //! @{
@@ -281,8 +283,8 @@ private:
     bool m_ready;
 
     /**
-     *  Since this phase represents a liquid phase, it's an error to
-     *  return a gas-phase answer. However, if the below is true, then
+     *  Since this phase represents a liquid (or supercritical) phase, it is an
+     *  error to return a gas-phase answer. However, if the below is true, then
      *  a gas-phase answer is allowed. This is used to check the thermodynamic
      *  consistency with ideal-gas thermo functions for example.
      */
--- a/include/cantera/thermo/WaterSSTP.h
+++ b/include/cantera/thermo/WaterSSTP.h
@@ -247,7 +247,7 @@ public:
      *  member function is to check the thermodynamic consistency of the
      *  underlying WaterProps class with ideal-gas thermo functions.
      */
-    void allowGasPhase(bool flag) { m_allowGasPhase = flag; }
+    void _allowGasPhase(bool flag) { m_allowGasPhase = flag; }
 
 protected:
     /**
--- a/interfaces/cython/cantera/liquidvapor.py
+++ b/interfaces/cython/cantera/liquidvapor.py
@@ -15,6 +15,19 @@ def Water():
     return WaterWithTransport('liquidvapor.yaml', 'water', transport_model='Water')
 
 
+def LiquidWater():
+    """
+    Create a `PureFluid` object using the IAPWS Formulation 1995 for
+    thermodynamic properties and the `WaterTransport` class for viscosity
+    and thermal conductivity.
+    """
+    class WaterWithTransport(PureFluid, _cantera.Transport):
+        __slots__ = ()
+
+    return WaterWithTransport('liquidvapor.yaml', 'liquid-water',
+                              transport_model='Water')
+
+
 def Nitrogen():
     """Create a `PureFluid` object using the equation of state for nitrogen."""
     return PureFluid('liquidvapor.yaml', 'nitrogen')
--- a/interfaces/cython/cantera/liquidvapor.py
+++ b/interfaces/cython/cantera/liquidvapor.py
@@ -8,11 +8,30 @@ def Water():
     """
     Create a `PureFluid` object using the equation of state for water and the
     `WaterTransport` class for viscosity and thermal conductivity.
+
+    The object returned by this method implements an accurate equation of
+    state for water that can be used in the liquid, vapor, saturated
+    liquid/vapor, and supercritical regions of the phase diagram. The
+    equation of state is taken from
+
+    W. C. Reynolds, *Thermodynamic Properties in SI: graphs, tables, and
+    computational equations for forty substances.* Stanford: Stanford
+    University, 1979. Print.
+
+    whereas formulas for transport are taken from
+
+    J. V. Sengers, J. T. R. Watson, *Improved International Formulations for
+    the Viscosity and Thermal Conductivity of Water Substance,* J. Phys. Chem.
+    Ref. Data, 15, 1291, 1986.
+
+    For more details, see classes Cantera::PureFluid, tpx::water and
+    Cantera::WaterTransport in the Cantera C++ source code documentation.
     """
     class WaterWithTransport(PureFluid, _cantera.Transport):
         __slots__ = ()
 
-    return WaterWithTransport('liquidvapor.yaml', 'water', transport_model='Water')
+    return WaterWithTransport('liquidvapor.yaml', 'water',
+                              transport_model='Water')
 
 
 def LiquidWater():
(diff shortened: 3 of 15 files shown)

### CODE REVIEW

**Comment by ischoegl** [comment]
**Other thoughts:** it may make sense to raise an error for `Q` (vapor fraction), unless `WaterSSTP::m_allowGasPhase` is `True`.

File: include/cantera/thermo/WaterSSTP.h [hunk 1]
```diff
@@ -140,9 +140,11 @@ class WaterSSTP : public SingleSpeciesTP
     explicit WaterSSTP(XML_Node& phaseRef, const std::string& id = "");
 
     virtual std::string type() const {
-        return "Water";
+        return "WaterSSTP";
```
speth: If anything, this should be `liquid-water-IAPWS95`, no? I've been trying to (slowly) kill off these obscure initialisms, especially in the most user-accessible parts of the code.

include/cantera/thermo/WaterProps.h
```diff
@@ -83,6 +83,9 @@ class PDSS_Water;
 //! The WaterProps class is used to house several approximation routines for
 //! properties of water.
 /*!
+ * This is a helper class for WaterSSTP and is not instantiable on its own
+ * (see \ref thermoprops and class \link Cantera::WaterSSTP WaterSSTP\endlink).
```
speth: Technically, you can instantiate these classes on their own. I think the confusion is more about the fact that they aren't instances of `ThermoPhase`. So maybe something like "not instantiable as a ThermoPhase object on its own"?

interfaces/cython/cantera/liquidvapor.py
```diff
@@ -4,47 +4,180 @@
...
+    which can be used in the liquid, vapor, saturated liquid/vapor, and
+    supercritical regions of the phase diagram.
+
+    The ``iapws95`` backend implements an IAPWS (International Association for
+    the Properties of Water and Steam) formulation for thermodynamic properties
+    taken from
+
+        J. V. Sengers, J. T. R. Watson, *Improved International Formulations for
+        the Viscosity and Thermal Conductivity of Water Substance,* J. Phys.
+        Chem. Ref. Data, 15, 1291, 1986.
```
speth, line 38: I don't think this is the correct reference for the thermodynamic properties -- as the paper title says, this for the viscosity and thermal conductivity. The one that should be listed here is the Wagner and Pruss paper (which is what you added to `liquidvapor.yaml`).

interfaces/cython/cantera/liquidvapor.py
```diff
@@ -4,47 +4,180 @@
...
+
+    which currently only implements liquid and supercritical regions.
+
+    For more details, see classes Cantera::PureFluid, tpx::water,
+    Cantera::WaterSSTP and Cantera::WaterTransport in the Cantera C++ source
```
speth: To generate links to the C++ documentation, write these as ``` :ct:`PureFluid` ``` etc. I'm not actually sure what to do about `tpx::water`, though.

interfaces/cython/cantera/liquidvapor.py
```diff
@@ -4,47 +4,180 @@
 from . import PureFluid, _cantera
 
 
-def Water():
+def Water(backend='default'):
```
speth: Perhaps `Reynolds` would be a better name for this backend?

include/cantera/thermo/WaterSSTP.h
```diff
@@ -238,6 +240,9 @@ class WaterSSTP : public SingleSpeciesTP
         return m_waterProps.get();
     }
 
+    //! Switch that enables calculations in the gas phase
+    void allowGasPhase(bool flag) { m_allowGasPhase = flag; }
```
speth: I'm worried about the addition of this member function, because it implies that the `m_allowGasPhase` flag actually works, which I think isn't true. As an example, if this flag is set to `true`, setting T and P to (300, 101325) and then (400, 101325) leads to a density of 937.411 kg/m^3, which is clearly wrong.

speth (commented):
While I appreciate some of the improvements here, and think they're worth including, I am concerned about making the current implementation of the IAPWS95 equation of state more user accessible because it is so incomplete when it comes to the two phase and vapor regions. Without being able to handle those portions of the phase diagram, I don't think it makes sense to include a phase definition in `liquidvapor.yaml`.

And given the prospect of using CoolProp (Cantera/enhancements#35), I'm not sure it would be worth extending the implementation of this phase definition to handle those regions.

include/cantera/thermo/WaterProps.h
ischoegl: I mainly copied from #809 here (and elsewhere) but will update.

include/cantera/thermo/WaterSSTP.h
ischoegl: I really didn’t want to implement this either, but an example in the test suite (`test_problems/cathermo/testWaterTP/testWaterSSTP.cpp`) depends on it. I’d certainly refuse to add it to Python ...

include/cantera/thermo/WaterSSTP.h
ischoegl: I'm very much in favor of going away from acronyms also, although I'm not sure that this clearly differentiates from `PDSS_Water` (the `TP` portion did point at one feature of this class that makes it suited for the two-phase region).

Going forward, I do not know the prospects of `iawps-95` - if it's useful enough to keep around I hope to see it as a more accurate and fully featured implementation than `reynolds`, - in which case the `liquid` is somewhat limiting. Since the `TP` approach pre-empts extensions to two-phase it will require a different approach with a different class heritage. So I am adopting the suggestion, as this is consistent with the label used for the thermo factory.

interfaces/cython/cantera/liquidvapor.py
ischoegl: Correct. Apparently made a mistake when copy-pasting (I initially had a separate class for IAPWS).

interfaces/cython/cantera/liquidvapor.py
ischoegl: Updated, but left the `tpx::water` etc for lack of better options.

interfaces/cython/cantera/liquidvapor.py
ischoegl: (although opting for lower-case)

include/cantera/thermo/WaterProps.h
ischoegl: Updated this to
```
 * This is a helper class for WaterSSTP and PDSS_Water and does not constitute
 * a complete implementation of a thermo phase by itself (see \ref thermoprops
 * and classes \link Cantera::WaterSSTP WaterSSTP\endlink and
 * \link Cantera::PDSS_Water PDSS_Water\endlink).
```
to reflect that `WaterSSTP` is not the only class using this.

include/cantera/thermo/WaterSSTP.h
ischoegl: Beyond adding a longer docstring:
```
    //! Switch that enables calculations in the gas phase
    /**
     *  Since this phase represents a liquid (or supercritical) phase, it is an
     *  error to return a gas-phase answer. The sole intended use for this
     *  member function is to check the thermodynamic consistency of the
     *  underlying WaterProps class with ideal-gas thermo functions.
     */
    void allowGasPhase(bool flag) { m_allowGasPhase = flag; }
```
 I don't have a good alternative idea, as I really do not want to break the test suite (which really targets the implementation of the underlying `WaterProps` class).

ischoegl:
@speth ... thank you for the review! I believe all is taken care of ... see 1069692 

> While I appreciate some of the improvements here, and think they're worth including, I am concerned about making the current implementation of the IAPWS95 equation of state more user accessible because it is so incomplete when it comes to the two phase and vapor regions. Without being able to handle those portions of the phase diagram, I don't think it makes sense to include a phase definition in liquidvapor.yaml.

With the changes of this PR, appropriate errors are raised if a user attempts to set a state outside of the permissible region (and it's only possible to deactivate in C++, where caveats are now clearly stated). Since `IAPWS` is presumably more accurate, I believe it's in the interest of users to be able to access what's already implemented. Perhaps someone will find this of use and take initiative to expand further: the Python front-end is certainly a more convenient starting point.

> And given the prospect of using CoolProp (Cantera/enhancements#35), I'm not sure it would be worth extending the implementation of this phase definition to handle those regions.

I personally agree, but am not sure what will become of `PDSS_Water` (as much of the code is shared with `WaterSSTP`, the Python access would still be useful). My obvious recommendation would be to simply allow for different backends for all wrappers once the `CoolProp` connection is implemented.

ischoegl:
A small change takes care of #577. The updated Python `Water` wrapper was convenient to track this issue down :wink:

ischoegl:
Inactive for 3 months - closing.

ischoegl:
Based on a UG post there may be value in wrapping this up after all

ischoegl:
@bryanwweber and @speth ... let me know if this is of interest for 2.5 (I believe all feedback is taken care of). It’s the last of my pending PRs.

ischoegl:
.... Triggering CI

include/cantera/thermo/WaterSSTP.h
speth: Taking a second look at this, I think I'd settle for naming this function `_allowGasPhase`, using the leading underscore to indicate that this function is in some sense internal and not to be used.

speth (changes requested):
Overall, I think this looks pretty good. I'm happy to merge this pending my one suggestion below.

include/cantera/thermo/WaterSSTP.h
ischoegl: done!

speth (approved):

ischoegl:
@speth ... thanks for the review (and the ones that came before) as well as all the work towards 2.5 ...

speth:
No problem. As you can tell from some of the recent activity, @bryanwweber and I are just working out the last few things that need to be taken care of for the 2.5.0 release.

PR #922, opened 2020-08-28
New C++ custom ODEs example

paulblum:
Proposing the addition of a new example for Cantera's C++ sample code, `custom.cpp`. This example solves a closed-system constant-pressure ignition problem using custom-implemented ODE governing equations, which are solved by CVODES. The actual problem is equivalent to the one solved in the existing [`custom.py`](https://cantera.org/examples/python/reactors/custom.py.html) example, and produces the same results.

This example was created as part of my GSoC project, *Developing a 0-D Steady-State Combustion Solver for Cantera*. The project is ongoing, and progress updates can be found in the [enhancements repository](https://github.com/Cantera/enhancements/issues/31).


samples/cxx/custom/custom.cpp
```diff
@@ -0,0 +1,191 @@
...
+    shared_ptr<ThermoPhase> gas;
+    shared_ptr<Kinetics> kinetics;
+    double pressure;
+    int nSpecies;
+    int nEqs;
+};
+
+int main() {
+    /* -------------------- CREATE GAS & SPECIFY STATE -------------------- */
+    auto sol = newSolution("gri30.yaml");
```
ischoegl, line 127: Minor comment: you could use a smaller mechanism for this H2-O2 system? The same is apparently true for `custom.py` (In both instances, `N2` would have to be replaced by `AR` if `H2O2.yaml` was used instead).

ischoegl:
Two more tangential comments that don't really involve the implementation itself ... :wink:

The sample should probably be added to automatic builds here:
https://github.com/Cantera/cantera/blob/329aa3fb606f2c89627a3993f398ccb56c6cf42c/samples/cxx/SConscript#L5-L15

and there *could* be a regression test, i.e.:
https://github.com/Cantera/cantera/blob/329aa3fb606f2c89627a3993f398ccb56c6cf42c/test_problems/SConscript#L227

ischoegl:
@paulblum ... great to see a new example! The current CI failures should be relatively easy to fix by specifying thresholds (just had to fix a similar thing for #927 - I admittedly didn't investigate and just copied values).

Beyond, I think it may make sense to switch the mechanism from `gri30.yaml` to `h2o2.yaml` here and for `custom.py` (other than not having `N2` - which is replaceable by `AR` for the purpose of an example - this shouldn't change much).

bryanwweber:
@ischoegl Any reason to prefer `h2o2.yaml` over `gri30.yaml`? I doubt there's any difference significant difference in how long they take to run on modern processors.

ischoegl:
@bryanwweber ... it's admittedly a minor (philosophical?) point, but things do add up. I would generally advocate for using the smallest possible mechanism that will do the job especially when there is absolutely no benefit of using a larger mechanism: as `h2o2.yaml` is available (as GRI minus carbon/nitrogen), it's the logical candidate for probing a H2-O2 system for the sake of an example.

**PS:** As another benefit, end-users (who are probably interested in some C species) will have to actually *think* about what mechanism to use instead of always defaulting to `gri30.yaml` - I believe we all agree that there are better options out there.

paulblum:
Just updated the regression test tolerances from the defaults to `threshold=1e-10` and `tolerance=2e-4`, which match the ones specified for `cxx-combustor`.

paulblum:
@ischoegl thanks for the feedback! I'll look into switching this mechanism next.

bryanwweber:
@paulblum @ischoegl I'd be in favor of deferring the switch to `h2o2.yaml` to a separate pull request, since it could be done for several examples, and leaving this with `gri30.yaml` means it'll match the Python example until such time as we switch any other examples. Doing a PR to change over the example input files would be a good first issue for any new contributors as well.

ischoegl:
🤷‍♂️ ... makes no difference as long as it’ll get done. I agree that this is an easy fix ... 5ish lines spread over a couple of files? Not sure it’s worth its own PR ...

**PS:** #932

samples/cxx/custom/custom.cpp
```diff
@@ -0,0 +1,197 @@
+/* @file custom.cpp
+ * Solve a closed-system constant pressure ignition problem where the governing equations
+ * are custom-implemented.
+ */
```
speth: If the goal is for this to show up in the Doxygen docs, the comment should to start with `/*!` or `/**`. Although I'll note that Doxygen isn't currently configured to scan the `samples/cxx` directory, so this doesn't really matter.

samples/cxx/custom/custom.cpp
```diff
@@ -0,0 +1,197 @@
...
+
```
speth, line 6: I think the Cantera copyright statement should be added here:
```
```

samples/cxx/custom/custom.cpp
```diff
@@ -0,0 +1,197 @@
...
+        wdot.resize(nSpecies);
+
+        // number of equations in the ODE system. a conservation equation for each
+        // species, plus a single energy conservation equation for the system.
+        nEqs = nSpecies + 1;
+    }
+
+    /**
+     * Evaluate the ODE right-hand-side function, ydot = f(t,y).
+     *   - overrided from FuncEval, called by the integrator during simulation.
```
speth:
```suggestion
     *   - overridden from FuncEval, called by the integrator during simulation.
```
This appears in several locations.

samples/cxx/custom/custom.cpp
```diff
@@ -0,0 +1,197 @@
...
+     * @param[in] t time.
+     * @param[in] y solution vector, length neq()
+     * @param[out] ydot rate of change of solution vector, length neq()
+     * @param[in] p sensitivity parameter vector, length nparams()
```
speth, line 58: The ability to use sensitivity analysis isn't implemented here, right? That's fine, but it's probably worth noting that this is the case.

samples/cxx/custom/custom.cpp
```diff
@@ -0,0 +1,197 @@
...
+private:
+    // private member variables, to be used internally.
+    shared_ptr<ThermoPhase> gas;
+    shared_ptr<Kinetics> kinetics;
+    vector_fp hbar;
+    vector_fp wdot;
+    double pressure;
+    int nSpecies;
+    int nEqs;
+};
```
speth: This isn't critical because this is a sample rather than core Cantera code, but we usually prefix member variables with `m_`, which makes it easier to differentiate between members and locals when reading the code in member functions.

samples/cxx/custom/custom.cpp
```diff
@@ -0,0 +1,197 @@
...
+        kinetics->getNetProductionRates(&wdot[0]);
+        
+        /* -------------------------- ENERGY EQUATION ------------------------- */
+        // the rate of change of the system temperature is found using the energy
+        // equation for a closed-system constant pressure ideal gas:
+        //     m*cp*dT/dt = - sum[h(k) * dm(k)/dt]
+        // or equivalently:
+        //     dT/dt = - sum[hbar(k) * dw(k)/dt] / (rho * cp)
+        double hdot_vol = 0;
+        for (int k = 0; k < nSpecies; k++)
```
speth: I like to recommend the use of `size_t` for these index variables (and for the type of `nSpecies` as well), since that's the type used for sizes in the C++ standard library.

samples/cxx/custom/custom.cpp
```diff
@@ -0,0 +1,197 @@
...
+            hdot_vol += hbar[k] * wdot[k];
```
speth: I strongly recommend against using the unbraced (one line) `for` loop construct, here and elsewhere.
```suggestion
        for (int k = 0; k < nSpecies; k++) {
            hdot_vol += hbar[k] * wdot[k];
        }
```

samples/cxx/custom/custom.cpp
```diff
@@ -0,0 +1,197 @@
...
+    // create an ODE integrator object, which will be used to solve the system of ODES defined
+    // in the ReactorODEs class. a C++ interface to the C-implemented SUNDIALS CVODES integrator
+    // (CVodesIntegrator) is built into Cantera, and will be used to solve this example.
+    //  - the default settings for CVodesIntegrator are used:
+    //     solution method: BDF_Method
+    //     problem type: DENSE + NOJAC
+    //     relative tolerance: 1.0e-9
+    //     absolute tolerance: 1.0e-15
+    //     max step size: +inf
+    Integrator *integrator = newIntegrator("CVODE");
```
speth: To avoid a memory leak here, write
```suggestion
    std::unique_ptr<Integrator> integrator(newIntegrator("CVODE"));
```

samples/cxx/custom/custom.cpp
```diff
@@ -0,0 +1,197 @@
...
+        integrator->integrate(tnow);
+        double *solution = integrator->solution();
+
+        // output the current absolute time and solution state vector to the csv file. you can view
+        // these results by opening the "custom_cxx.csv" file that appears in this program's directory
+        // after compiling and running.
+        outputFile << tnow << ", ";
+        for (int i = 0; i < odes.neq(); i++)
+            outputFile << solution[i] << ", ";
+        outputFile << std::endl;
```
speth: In this case, there's a fairly easy way to avoid the trailing comma after the last variable on each line:
```suggestion
        outputFile << tnow;
        for (int i = 0; i < odes.neq(); i++) {
            outputFile << ", " << solution[i];
        }
        outputFile << std::endl;
```

speth (changes requested):
Thanks a lot for providing this example, @paulblum! I think it looks very good, and provides a C++ example that's non-trivial but still fairly clean.

I think there are just a handful of issues to tweak, along with rebasing this and fixing the conflicts.

paulblum:
Thanks for the review @speth, these were all really good suggestions! Just pushed a commit with the updates, please let me know if there's anything else you think can be improved 😄

paulblum:
Failed CI test due to regression test tolerances, I had fixed that in an earlier version but looks like it got removed by a conflict... latest commit should be all set!

speth (approved):

speth:
Thanks, @paulblum, this looks great. I think the only thing I'd like to do before merging is to squash the last couple of commits to avoid the large diff associated with the changes to the `custom_cxx_blessed.csv` file. If you're comfortable doing that, please go ahead, otherwise I can do it.

paulblum:
@speth that makes sense to me, although I'm not familiar with how to squash commits... probably better for you to do it if it's not too much trouble. Thanks again for the help!

PR #931, opened 2020-09-05
Ensure consistent extra columns in SolutionArray

### LINKED ISSUE #895: Sequences can be appended as extra items on a SolutionArray
When appending a state to a `SolutionArray`, an item of length greater than 1 can be passed for any `extra` items. This raises a `DeprecationWarning` from NumPy, and gives a... less than useful result:

```python
>>> gas = ct.Solution("air.yaml")
>>> states = ct.SolutionArray(gas, 3, extra={"prop": [5, 6, 7]}
>>> states.append(TPX=(1100, 3e5, "AR:1.0"), prop=[1, 2, 3])
>>> states.prop
/Users/bryan/GitHub/cantera/build/python/cantera/composite.py:576: VisibleDeprecationWarning:
Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays
with different lengths or shapes) is deprecated. If you meant to do this, you must specify 
'dtype=object' when creating the ndarray
  return np.array(self._extra[name])
array([5, 6, 7, list([1, 2, 3])], dtype=object)
```

Yes, that's a list as an item in a NumPy array.

Fixing this is not super straightforward, because we may need to also account for the possibility of strings as values, so a simple `if len(item) > 1` check is not sufficient, see #896.

### LINKED ISSUE #930: It is possible to create inconsistent extra SolutionArray entries
**Problem description**

Inconsistent extra entries for `SolutionArray` are not caught.

**Steps to reproduce / Behavior**

```
>>> gas = ct.Solution("air.yaml")
>>> states = ct.SolutionArray(gas, 3, extra={"prop": [5, 6, 7]}
>>> states.prop = "hello world"
>>> states.prop
'hello world'
```
I.e. `states.prop` is neither of correct type (should be `np.ndarray`) nor consistent in terms of dimensions.

**System information**

- Cantera version: 2.5.0b1
- OS: Ubuntu 18.04
- Python/MATLAB/other software versions: 3.6.9

ischoegl:

- Fix bugs in beta release of 2.5.0b1
- Ensure that dimensions of `extra` columns are consistent
- Enable multi-dimensional `extra` 'columns' (e.g. 2D arrays similar to what already exists for other `SolutionArray` attributes)
- Like the existing code, this PR takes advantage of NumPy for on-the-fly adjustments of data types.

Fixes #895, fixes #930

**Illustration**

```
In [2]: gas = ct.Solution("air.yaml")
   ...: states = ct.SolutionArray(gas, 3, extra={"prop": [5, 6, 7]})
   ...: states.append(TPX=(1100, 3e5, "AR:1.0"), prop=[1, 2, 3]) # <--- Fix #895
   ...: 
[...]
ValueError: Encountered incompatible value '[1, 2, 3]' for extra column 'prop'.

In [3]: states.prop = ‘spam’ # <--- overwrite individual entries / Fix #930
   ...: states.prop
   ...: 
Out[3]: array([‘spam’, ‘spam’, ‘spam’], dtype=‘<U4’)

In [4]: states.prop = np.ones((3,4)) # <--- enable multi-dimensional extra 'columns'
   ...: states.prop
   ...: 
Out[4]: 
array([[1., 1., 1., 1.],
       [1., 1., 1., 1.],
       [1., 1., 1., 1.]])

In [5]: states.prop = [1, 2, 3, 4] # <--- ensure that dimensions are consistent
[...]
ValueError: Incompatible shapes for extra column 'prop': cannot assign value with shape (4,) to SolutionArray with shape (3,)
```

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -598,8 +598,8 @@ class SolutionArray:
             return SolutionArray(self._phase, shape, states)
 
     def __getattr__(self, name):
-        if name in self.__dict__['_extra']:
-            return self.__dict__['_extra'][name]
+        if name in self._extra:
+            return self._extra[name]
         elif name in self.__dict__:
             super().__getattr__(name)
         else:
@@ -607,11 +607,11 @@ class SolutionArray:
                 self.__class__.__name__, name))
 
     def __setattr__(self, name, value):
-        if name in self.__dict__['_extra']:
+        if name in self._extra:
             new = np.array(value)
             if not new.shape:
                 # maintain shape of extra entry
-                new = np.full(self.__dict__['_extra'][name].shape, value)
+                new = np.full(self._extra[name].shape, value)
             elif new.shape[:len(self._shape)] != self._shape:
                 raise ValueError(
                     "Incompatible shapes for extra column '{}': cannot assign "
@@ -690,20 +690,15 @@ class SolutionArray:
                 ) from None
             setattr(self._phase, attr, [kwargs[a] for a in attr])
 
-        extra = self._extra
         for name, value in self._extra.items():
-            new = extra_temp.pop(name)
+            new = extra_temp[name]
             if len(value):
                 if (value.ndim == 1 and hasattr(new, '__len__') and
                     not isinstance(new, str)):
-                    # revert changes to extra before raising error
-                    self._extra = extra
                     raise ValueError(
                         "Encountered incompatible value '{}' for extra column '{}'."
                         "".format(new, name))
                 elif value.ndim > 1 and value.shape[1:] != np.array(new).shape:
-                    # revert changes to extra before raising error
-                    self._extra = extra
                     raise ValueError(
                         "Shape of new element does not match existing extra "
                         "column '{}'".format(name))
@@ -711,7 +706,11 @@ class SolutionArray:
             # np.append when appending a single item.
             v = value.tolist()
             v.append(new)
-            self._extra[name] = np.array(v)
+            extra_temp[name] = np.array(v)
+
+        for name, value in extra_temp.items():
+            self._extra[name] = value
+
         self._states.append(self._phase.state)
         self._indices.append(len(self._indices))
         self._shape = (len(self._indices),)
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -502,6 +502,7 @@ class SolutionArray:
     _purefluid_scalar = ['Q']
 
     def __init__(self, phase, shape=(0,), states=None, extra=None, meta=None):
+        self.__dict__['_extra'] = OrderedDict()
         self._phase = phase
 
         if isinstance(shape, int):
@@ -528,8 +529,6 @@ class SolutionArray:
 
         reserved = self.__dir__()
 
-        self._extra = OrderedDict()
-
         if isinstance(extra, dict):
             for name, v in extra.items():
                 if name in reserved:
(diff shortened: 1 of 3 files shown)


ischoegl:
@speth / @bryanwweber ... I cannot request a review, but I’d appreciate if those fixes could make it into 2.5.

interfaces/cython/cantera/composite.py
```diff
@@ -591,12 +598,28 @@ def __getitem__(self, index):
...
         else:
             raise AttributeError("'{}' object has no attribute '{}'".format(
                 self.__class__.__name__, name))
 
+    def __setattr__(self, name, value):
+        if name in self.__dict__['_extra']:
+            new = np.array(value)
+            if not new.shape:
+                new = np.full(self._shape, value)
+            elif new.shape[0] != self._shape[0]:
```
ischoegl, line 614: Needs to check all of `self._shape`

interfaces/cython/cantera/composite.py
```diff
@@ -591,12 +598,28 @@ def __getitem__(self, index):
...
+            super().__getattr__(name)
         else:
             raise AttributeError("'{}' object has no attribute '{}'".format(
                 self.__class__.__name__, name))
 
+    def __setattr__(self, name, value):
+        if name in self.__dict__['_extra']:
+            new = np.array(value)
+            if not new.shape:
+                new = np.full(self._shape, value)
```
ischoegl, line 613: Should preserve shape of extra column.

ischoegl (commented):
[minor edits are taken care of]

interfaces/cython/cantera/composite.py
```diff
@@ -591,12 +598,29 @@ def __getitem__(self, index):
             return SolutionArray(self._phase, shape, states)
 
     def __getattr__(self, name):
-        if name in self._extra:
-            return self._extra[name]
+        if name in self.__dict__['_extra']:
+            return self.__dict__['_extra'][name]
+        elif name in self.__dict__:
+            super().__getattr__(name)
```
speth: I don't think either of the changes here are necessary. The only place it is necessary to avoid writing `self._extra` due to the addition of a `__setattr__` method is when setting it in `__init__`, since the initial `if` condition in `__setattr__` can't be evaluated until the `_extra` attribute exists. This applies to the other places where `self.__dict__['_extra']` is written as well.

Unlike `__setattr__`, `__getattr__` is only called when the default mechanism (i.e. accessing an instance attribute) fails, so the second case here will never be satisfied.

interfaces/cython/cantera/composite.py
```diff
@@ -666,14 +690,30 @@ def append(self, state=None, **kwargs):
                 ) from None
             setattr(self._phase, attr, [kwargs[a] for a in attr])
 
-        self._states.append(self._phase.state)
-        self._indices.append(len(self._indices))
+        extra = self._extra
```
speth: This doesn't actually create a copy of `self._extra`, so using this to revert changes on erroneous input doesn't work. For example:
```python
>>> gas = ct.Solution('air.yaml')
>>> states = ct.SolutionArray(gas, 3, extra={"p1": [5, 6, 7], "p2": [1, 2, 3]})
>>> states.append(TPX=(300, 101325, 'O2:1.0'), p1=5, p2=[7,8])
ValueError: Encountered incompatible value '[7, 8]' for extra column 'p2'.
>>> print(states._extra)
OrderedDict([('p1', array([5, 6, 7, 5])), ('p2', array([1, 2, 3]))])
```

speth (changes requested):
I think these are useful improvements to eliminate some confusing behavior in the `SolutionArray` class. Just noticed a couple of things to fix up further.

interfaces/cython/cantera/composite.py
ischoegl: Good point - it's all by reference in Python, of course.

ischoegl:
@speth ... thanks for the review. I believe it's all addressed.

As for the other PR's, I can rebase once #946 is merged.

speth (approved):
Looks good, thanks.

PR #938, opened 2020-09-23
FIX: Update isentropic.py - thermo example

farisachugthai:
The ImportError doesn't notify the user that matplotlib is missing.

It simply prints debugging information which may be more confusing than not catching the error altogether.

This modification notifies the user and then prints debugging info in case of any other exceptions.

`noqa` is used for flake8 if the repository uses it to verify code quality.

-
-
-

Fixes #

**Checklist**

- [ ] There is a clear use-case for this code change
- [ ] The commit message has a short title & references relevant issues

### DIFF
--- a/interfaces/cython/cantera/examples/thermo/isentropic.py
+++ b/interfaces/cython/cantera/examples/thermo/isentropic.py
@@ -69,5 +69,7 @@ if __name__ == "__main__":
         plt.show()
 
     except ImportError:
+        print('Matplotlib not found. Unable to plot results.')
+    except:  # noqa
         print('area ratio,   Mach number,   temperature,   pressure ratio')
         print(data)


interfaces/cython/cantera/examples/thermo/isentropic.py
```diff
@@ -69,5 +69,7 @@ def isentropic(gas=None):
         plt.show()
 
     except ImportError:
+        print('Matplotlib not found. Unable to plot results.')
+    except:  # noqa
```
ischoegl, line 73: The current modification would result in data no longer being displayed, i.e. it breaks current behavior.

farisachugthai: It only falls back to that behavior if an error outside of ImportError is encountered. The current behavior needs to be more specific if Matplotlib isnt installed.

ischoegl: I believe it was the intent of the original implementation to display results if matplotlib is not installed. The way you changed this, results no longer get displayed. I.e. things would now get displayed if there is another error. Before your change, other errors are raised (which is imho the correct implementation).

bryanwweber: I agree with @ischoegl's interpretation - there's no need to add the extra bare `except` statement here. Adding the `print()` call is fine though.

farisachugthai: Ah I see. Okay well I can close this PR then.

PR #950, opened 2020-12-16
Loosen tolerances in PureFluidTestCases.test_consistency_volume

inducer:
This seems to be required to get this test to pass on the FPUs on Power9 and ARM. See
https://github.com/conda-forge/cantera-feedstock/pull/1
specifically the build failure for
https://github.com/conda-forge/cantera-feedstock/pull/1/commits/22bc536d649d6f8ba392dd83f1e289c84cf37a3e
and the successful build for
https://github.com/conda-forge/cantera-feedstock/pull/1/commits/472b4eb76d3abae6e8d4fb3eaf461e2babc968a4

### DIFF
--- a/interfaces/cython/cantera/test/test_purefluid.py
+++ b/interfaces/cython/cantera/test/test_purefluid.py
@@ -330,7 +330,7 @@ class PureFluidTestCases:
             a2 = self.a(state.T, 1/(V+0.5*dV))
 
             # dP/drho is high for liquids, so relax tolerances
-            tol = 100*self.tol.dAdV if state.phase == 'liquid' else self.tol.dAdV
+            tol = 300*self.tol.dAdV if state.phase == 'liquid' else self.tol.dAdV
 
             # At constant temperature, dA = - p dV
             msg = 'At state: T=%s, rho=%s' % (state.T, state.rho)


inducer:
Ready for review. The one CI failure seems spurious.

bryanwweber:
Interestingly, I can reproduce the exact value (accurate to 1e-12) by reducing the finite difference of the volume to `dV = 1e-6 * V` instead of `5e-6` on my Intel Macbook Pro.

```python
import cantera as ct
def a(w, T, rho):
    """ Helmholtz free energy """
    w.TD = T, rho
    return w.u - T * w.s
w = ct.Water()
T = 295
v = 0.0010025
w.TD = T, 1.0/v
# Mimic the calculations done in the test,
# although it doesn't appear to matter
rho = 1.0/v
V = 1 / rho
dV = 1e-6 * V
a1 = a(w, T, 1/(V-0.5*dV))
a2 = a(w, T, 1/(V+0.5*dV))
failure = 2616.0642095932
failure + (a2-a1)/dV
```

speth:
@bryanwweber, I think the reason that you see the same (imprecise) value for dA/dV in other cases is that we're just crossing a threshold for the tolerance in one of the nested iterative algorithms that the `Substance` class uses in the two-phase region, and the difference in the values is what you would get from taking one more iteration. This particular test point is in a place where the vapor fraction is something like 5e-9, so it's not surprising that some of these quantities have steep slopes.

This seems to be the minimally-invasive fix for the test failure - I think we can save improvements on the accuracy of the `Substance` solution algorithm for another day.

speth (approved):

PR #955, opened 2021-01-09
Enabling beta = 0.

### LINKED ISSUE #954: ctml_writer.py ignores beta=0 for electrochemical reactions
**Problem description**
If I define an electrochemical surface_reaction in my .cti file and specify beta=0, the ctml_writer does not write a `<electrochem beta="0">` node but treats it as a non-electrochemical reaction. The problem is that "importKinetics" then recognizes this reaction as electrochemical and assigns a default value of beta=0.5. 
This makes it impossible to define an electrochemical reaction that has a potential-independent energy barrier without either manually manipulating the XML file or setting beta to a very small value (e.g. 1e-20) as workaround. While these are easy fixes, the main problem is that the described behaviour is not expected.

**Steps to reproduce**

1. Create CTI file with electrochemical surface reaction where beta=0
2. Run ctml_writer.py on it to create XML input file
3. Import surface phase via "get_XML_File" and "findXMLPhase"
4. Create "InterfaceKinetics" object via "importKinetics"

**Behavior**

The ctml_writer seems to consider beta=0 as indicative of a reaction without charge transfer and therefore does not write a <electrochem> node to the XML file. The C++ import functions then recognize the reaction as indeed having charge transfer and apply the "standard" value of 0.5.

The main problem is line 132 in "ctml_writer.py"

>         if self._type == 'edge' or self._type == 'surface':
>            if self._beta > 0:
>                electro = kfnode.addChild('electrochem')
>                electro['beta'] = repr(self._beta)

**System information**

- Cantera version: 2.4.0
- OS: Windows 10
- Python/MATLAB/other software versions: Python/C++

**Attachments**

Example CTI file for reproduction:

> t_0 = 298.15	 # standard temperature in K
> p_0 = 101325	 # standard pressure in Pa
> 
> incompressible_solid(
> 	 name='liquid',
> 	 elements='H E',	 
> 	 species=('H+'),
> 	 density=(997,'kg/m3')
> ) 
> 
> species( 
> 	 name='H+', 
> 	 atoms='H:1 E:-1'
> )
> 
> metal(
> 	 name='solid',
> 	 elements='E',	 species=('e-')
> )
> 
> species( 
> 	 name='e-', 
> 	 atoms='E:1'
> ) 
> 
> ideal_interface(
> 	 name='surface',
> 	 elements='H',	 
> 	 species=('(Cu)','H(Cu)')
> )
> 
> species( 
> 	 name='(Cu)', 
> 	 atoms='H:0'
> ) 
> 
> species( 
> 	 name='H(Cu)', 
> 	 atoms='H:1'
> )
> 
> surface_reaction(
> 	id='H2_ads',
> 	equation='1 H+ + 1 e- + 1 (Cu) <=> 1 H(Cu)',
> 	kf=(6.214e+12,0,1), 
> 	beta=0
> )

**Additional context**

Unfortunately, just changing line 1322 to `if self._beta > 0:` will result in another problem: Now if we do not specify a beta for an electrochemical reaction, it will be automatically set to zero and written as such to the XML. This defies the idea of having beta=0.5 as a default value. Maybe the `__init__` in the ctml_writer should set beta to a negative value. This way we can distinguish between a non-specified beta and a beta=0.

decaluwe:

- For electrochemical reactions, changes the default `beta` value in `ctml_writer.py` to `None`, and checks to see if `beta` has been set to a value that is not `None` to detect an electrochemical reaction.
- Previously, the default was `beta = 0.0` which resulted in electrochemical reactions with `beta = 0.0` not being detected, and therefore handled incorrectly.

Fixes #954 

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/ctml_writer.py
+++ b/interfaces/cython/cantera/ctml_writer.py
@@ -1319,9 +1319,11 @@ class reaction(object):
             self._kf = []
 
         if self._type == 'edge' or self._type == 'surface':
-            if self._beta > 0:
+            if self._beta is not None:
                 electro = kfnode.addChild('electrochem')
                 electro['beta'] = repr(self._beta)
+            else:
+                self._beta = 0.0
 
         for kf in self._kf:
             if isinstance(kf, rate_expression):
@@ -1640,7 +1642,7 @@ class surface_reaction(reaction):
                  kf=None,
                  id='',
                  order='',
-                 beta = 0.0,
+                 beta = None,
                  options=[],
                  rate_coeff_type = ''):
         """
@@ -1686,7 +1688,7 @@ class edge_reaction(reaction):
                  kf = None,
                  id = '',
                  order = '',
-                 beta = 0.0,
+                 beta = None,
                  options = [],
                  rate_coeff_type = ''):
         reaction.__init__(self, equation, kf, id, order, options)
--- a/interfaces/cython/cantera/ctml_writer.py
+++ b/interfaces/cython/cantera/ctml_writer.py
@@ -1322,8 +1322,6 @@ class reaction(object):
             if self._beta is not None:
                 electro = kfnode.addChild('electrochem')
                 electro['beta'] = repr(self._beta)
-            else:
-                self._beta = 0.0
 
         for kf in self._kf:
             if isinstance(kf, rate_expression):


interfaces/cython/cantera/ctml_writer.py
```diff
@@ -1319,9 +1319,11 @@ def build(self, p):
             self._kf = []
 
         if self._type == 'edge' or self._type == 'surface':
-            if self._beta > 0:
+            if self._beta is not None:
                 electro = kfnode.addChild('electrochem')
                 electro['beta'] = repr(self._beta)
+            else:
+                self._beta = 0.0
```
speth: I don't think it's necessary to set a value for `_beta` here. This would just have the awkward effect of having the result of `build` change if it was called a second time (not that this actually happens in practice).

speth (changes requested):

interfaces/cython/cantera/ctml_writer.py
decaluwe: Okay, thanks @speth. Just went in and corrected that.

speth (approved):

PR #957, opened 2021-01-14
[CI] Fix 404 for python3-lxml dependency

bryanwweber:
The version of the python3-lxml package stored in the default apt
database on the Ubuntu 18.04 runner is not available from the package
server. This update finds the new copy to resolve the problem.

**Checklist**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,10 +49,18 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [ '3.5', '3.6', '3.7', '3.8' ]
+        python-version: [ '3.5', '3.6', '3.7', '3.8', '3.9' ]
         os: ['macos-10.15', 'macos-11.0']
       fail-fast: false
     steps:
+    # Attempt to fix intermittent cloning errors. The error message says something like
+    # error: RPC failed; curl 18 transfer closed with outstanding read data remaining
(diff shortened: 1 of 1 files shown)


ischoegl:
Good to see the Ubuntu issue fixed. Otherwise, it looks like the macOS failures started with 37ab749ebb595ae61cbd651abd4af1169fef1758. Are the failing runs macOS 10.5 or 11? It’s odd that there isn’t a trace back.

bryanwweber:
It seems like it's failing for different version of Python on both 10.15 and 11.0, so both? Very annoying there's no log messages. And there's the run that just never starts, it stays queued forever... very strange.

ischoegl:
Funny. Everything passed now! Looks like a GitHub Actions (macOS capacity?) issue ... . Per your changes, these look straightforward. It always appears to be the little things ...

speth (approved):

ischoegl:
@bryanwweber ... great to see this fixed!

Only comment I have is that GH Actions limits MacOS runners to 5 concurrent instances (see [here](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration)), so having 10 jobs does creates a bottleneck.

bryanwweber:
@ischoegl Hmmm good point. We probably don't need to run 3.5 and 3.6 on macOS, 3.7-3.9 is probably sufficient since we have the other Pythons covered on other jobs.

PR #968, opened 2021-01-27
[Kinetics] Fix third body concentrations for non-ideal gases

### LINKED ISSUE #967: Bug with the computation of [M] with real gases
We have been looking through the Cantera source code to examine how the activity concentrations are implemented in preparation for us to add a real-gas equation of state model. 

When looking through `GasKinetics.cpp` and `ThirdBodyCalc.h` we noticed what could be a bug when computing the real-gas third body concentration.  On line 63 of `GasKinetics.cpp` the activity concentrations are stored in `m_conc` and on line 64 the normal total concentration is stored in `ctot`.  The activity concentrations, `m_conc`, and total molar density, `ctot`, is fed into 3-body update function.  

On lines 36-44 of `ThirdBodyCalc.h`, [M] is calculated in the update function.  Note: `m_conc` is named `conc` in this function.  These lines of code perform:

`[M] = m_default*ctot + sum(m_eff*conc_i)`,

where (presumably) `m_default = 1` and `m_eff = actual efficiency factor - 1`.  The portion of [M] computed using ctot does not include real gas effects, whereas the other portion includes the real gas effects.  Written another way using `c_i` for molar concentration and `a_i` for the activity concentration for species i:

`[M] = sum(c_i + (efficiency_factor_i - 1)*a_i)`

This seems inconsistent with real gas thermodynamics.  I am also pretty sure that, in general, `sum(c_i) != sum(a_i)`. 

It seems to me that ctot needs to be redefined in the computation of [M] as `sum(a_i)`. 

Perhaps this is a bug that is left over from optimizing the computation of [M] for ideal gas mixtures?

speth:

- Consistently use physical concentrations when calculating third body and falloff reaction rates, for both specifically enumerated efficiencies and generic third body colliders.

Partial fix for #967

**Checklist**

### DIFF
--- a/src/kinetics/GasKinetics.cpp
+++ b/src/kinetics/GasKinetics.cpp
@@ -68,17 +68,18 @@ void GasKinetics::update_rates_T()
 
 void GasKinetics::update_rates_C()
 {
-    thermo().getActivityConcentrations(m_conc.data());
+    thermo().getActivityConcentrations(m_act_conc.data());
+    thermo().getConcentrations(m_phys_conc.data());
     doublereal ctot = thermo().molarDensity();
 
     // 3-body reactions
     if (!concm_3b_values.empty()) {
-        m_3b_concm.update(m_conc, ctot, concm_3b_values.data());
+        m_3b_concm.update(m_phys_conc, ctot, concm_3b_values.data());
     }
 
     // Falloff reactions
     if (!concm_falloff_values.empty()) {
-        m_falloff_concm.update(m_conc, ctot, concm_falloff_values.data());
+        m_falloff_concm.update(m_phys_conc, ctot, concm_falloff_values.data());
     }
 
     // P-log reactions
@@ -187,10 +188,10 @@ void GasKinetics::updateROP()
     }
 
     // multiply ropf by concentration products
-    m_reactantStoich.multiply(m_conc.data(), m_ropf.data());
+    m_reactantStoich.multiply(m_act_conc.data(), m_ropf.data());
 
     // for reversible reactions, multiply ropr by concentration products
-    m_revProductStoich.multiply(m_conc.data(), m_ropr.data());
+    m_revProductStoich.multiply(m_act_conc.data(), m_ropr.data());
 
     for (size_t j = 0; j != nReactions(); ++j) {
         m_ropnet[j] = m_ropf[j] - m_ropr[j];
--- a/include/cantera/kinetics/BulkKinetics.h
+++ b/include/cantera/kinetics/BulkKinetics.h
@@ -62,7 +62,12 @@ protected:
     //! valued stoichiometries.
     vector_fp m_dn;
 
-    vector_fp m_conc;
+    //! Activity concentrations, as calculated by
+    //! ThermoPhase::getActivityConcentrations
+    vector_fp m_act_conc;
+
+    //! Physical concentrations, as calculated by ThermoPhase::getConcentrations
+    vector_fp m_phys_conc;
     vector_fp m_grt;
 
     bool m_ROP_ok;
--- a/include/cantera/kinetics/GasKinetics.h
+++ b/include/cantera/kinetics/GasKinetics.h
@@ -101,6 +101,7 @@ protected:
     vector_fp falloff_work;
     vector_fp concm_3b_values;
     vector_fp concm_falloff_values;
+
     //!@}
 
     void processFalloffReactions();
--- a/src/kinetics/BulkKinetics.cpp
+++ b/src/kinetics/BulkKinetics.cpp
@@ -178,7 +178,8 @@ void BulkKinetics::modifyElementaryReaction(size_t i, ElementaryReaction& rNew)
 void BulkKinetics::resizeSpecies()
 {
     Kinetics::resizeSpecies();
-    m_conc.resize(m_kk);
+    m_act_conc.resize(m_kk);
+    m_phys_conc.resize(m_kk);
     m_grt.resize(m_kk);
 }
 


decaluwe:
This fix looks straightforward and correct, but I have a question about whether it is necessary.  I'll post my thoughts over on the thread for #967 though.

src/kinetics/GasKinetics.cpp
```diff
@@ -61,17 +61,18 @@ void GasKinetics::update_rates_T()
...
     // 3-body reactions
     if (!concm_3b_values.empty()) {
-        m_3b_concm.update(m_conc, ctot, concm_3b_values.data());
+        m_3b_concm.update(m_phys_conc, ctot, concm_3b_values.data());
     }
 
     // Falloff reactions
     if (!concm_falloff_values.empty()) {
-        m_falloff_concm.update(m_conc, ctot, concm_falloff_values.data());
+        m_falloff_concm.update(m_phys_conc, ctot, concm_falloff_values.data());
```
decaluwe, line 82: Just want to verify that for a `falloff` reaction we do in fact want the physical concentrations (this would also be a departure from how we were previously handling this, if I am reading correctly.

decaluwe (dismissed):
Hi @speth this looks good.  Just one question below, to make sure that we are intentionally also changing the falloff reaction formulation to use the physical concentrations, rather than the previously-used activity concentrations.

src/kinetics/GasKinetics.cpp
speth: Yes, that was intentional on my part. I was interpreting the role of the third-body collider in `falloff` reactions to be analogous to its role in `three-body` reactions. I would appreciate @rwest or @cfgoldsmith weighing in one way or the other.

speth: In either case, the inconsistency between `m_conc` and `ctot` in the above expression exists for falloff reactions in the current implementation, the same as for `three-body` reactions.

rwest: I agree that falloff and three-body should be treated the same way. The most detailed discussion I've found of these issues is in https://doi.org/10.1016/j.combustflame.2019.11.041 and I think my reading of it (or my intuition) would suggest we use the actual concentration for both, but I guess the important thing is to do the same as whatever the rate coefficient's "author" did when fitting it.  That article also throws some shade at "Most combustion codes" for not yet implementing recently proposed mixture rules. 
We could ask @The-Burke-Lab to comment.

decaluwe: > the important thing is to do the same as whatever the rate coefficient's "author" did when fitting it.

Ah, a very good point.  Which is making this start to sound more like our discussions of thermo and the "reference" vs. "standard" vs. whatever state... 😓

My _guess_ would be that most reaction fitting is done using an ideal gas assumption, which would be convenient, since the actual physical concentration and activity concentration are the same.

ischoegl:
Reading through the comments, it appears that there is a consensus that concentrations of third body collision partners should use physical concentrations as implemented in this PR. However, there is a second point that was raised in #967, which pertains to reactions that explicitly state the colliding species, which would be treated in an inconsistent manner, i.e. 

```H+O2+N2<=>HO2+N2```

As an alternative to @speth and @decaluwe's suggestion (i.e. rewrite YAML), the detection of those reactions without modification of YAML input would be relatively straight-forward (stoich coefficient equal to one on both sides). As long as this behavior is clearly documented on the science pages, I believe an 'auto-detection' mechanism as is done elsewhere may be the simplest solution.

One way forward would be to merge the current PR but leave the issue open, as there is some overlap with other ongoing work (#995). If a consensus can be found, I'd be happy to add a fix there.

decaluwe:
Thanks, @ischoegl . One comment, one question:
1. I think we would need to search for a reaction where a species has _any_ stoich coefficient on both sides, not a coefficient of one on both sides--see the example that @speth posted on #967 (`nu_fwd` is 2 and `nu_rev` is 1 for species `O2`).  Am I thinking about this correctly?
2. Regardless, agreed that an automatic detection would be preferable, if it can be robustly implemented.  Regarding @speth's suggestion (make the user specify `type: three-body`, but automatically detect the collider species).  Would there ever be a reason where a species is both a reactant and product, but _does not_ serve as a third-body collider?

ischoegl:
@decaluwe thanks ... absolutely true about `H + 2 O2 <=> HO2 + O2`! (although the implementation wouldn’t be much different). I believe that specifying `three-body` would generally not be necessary unless the efficiency is not equal to ‘1’ (similar rationale as for electrochemical reactions, which do not need to be labeled as such - you were able to convince me on that one). Per your question, I’m curious to hear other opinions, as I am not sure myself that my intuition - that this won’t happen - is actually correct at all times.

ischoegl (commented):
I'm adding my approval here (with the caveat that this does not completely resolve #967).

speth:
I just pushed a revised commit message so this won't automatically close #967. GitHub's notation that I "dismissed decaluwe’s stale review" sounds way too judgemental 🤣 .

decaluwe:
😤😤😤 My reviews are as fresh as they get, GitHub!

🤣 

Okay, so the remaining issue, @ischoegl is the auto-detction issue for where the third body collider is explicitly listed, is that correct?

ischoegl:
@decaluwe ... correct. Would create a short PR so that won’t get buried in 3.5k lines in #995, although it neatly fits with the work there :wink: ... science docs are another remaining issue.

*Edit: as an alternative, I could create a new PR building on @speths's commits here, which will avoid leaving code in flux (although it presumably only affects real gas calculations). Presumably pushing here would work, but I haven't done that before.*

ischoegl (dismissed):

ischoegl:
@speth ... looks like pushing to this PR may be simpler than expected. Any objections if I add something here?

speth:
Pushing to PR branches is indeed pretty straightforward. If you have an implementation for this idea ready, please go ahead and push it here.

ischoegl:
@speth ... don’t have anything ready (yet), but I’d rather tackle this separate from (and before) #995, as merge conflicts will arise. This PR already clashes as is.

ischoegl:
@speth, @decaluwe and @rwest 

I pushed a draft for detection of three-body reactions (which mostly works). Fixing the few failing tests is simple, but I wanted to get some second opinions first:
* For automatically detected three-body reactions, some of the reaction equations are currently rewritten: the third body now is always in last place, and collision partners appear with the same coefficient on both sides. E.g. `2 O2 + H <=> HO2 + O2` would become `H + O2 + O2 <=> HO2 + O2` ... this could either be an annoyance or a feature, and is obviously picked up by some regression tests.
* Some of the detection has to be limited (likely requiring a coefficient of one on one side, and a coefficient larger than one on the other). As an example, the CI dug up a soot mechanism with the reaction `C6H5 + BIN5 => .9846153846 BIN5 + 1.53846154E-2 BIN6`, which certainly shouldn't qualify.

speth:
On the second point, would it make sense to just skip this for reactions where there are non-integer stoichiometries? I think that since they are not elementary reactions, all of the careful considerations here about what the third body means have likely gone out the window.

ischoegl:
> On the second point, would it make sense to just skip this for reactions where there are non-integer stoichiometries? I think that since they are not elementary reactions, all of the careful considerations here about what the third body means have likely gone out the window.

I considered this, but there may be the occasional equation with 1.5 O2 on one side (but I guess that would be far-fetched also) - I'm happy to implement any suggestion that finds a consensus.

ischoegl:
Nevermind. On second thought, the argument was pretty compelling already. Which leaves the formatting question ...

ischoegl:
~Looking over regression tests, I noticed that `H + HCNO <=> H + HNCO` is flagged based on the current criterion (non-zero integer stoich coeff on both sides). What kinetic considerations apply for this type of reaction? From an algorithmic viewpoint this may not matter much, but the current implementation would assign the type `three-body` ... comments?~

[*Edit: one obvious solution would be to simply apply the `three-body` label, and require three molecules on at least one side*]

ischoegl:
This is quite odd. All tests pass locally on my machine (on the non-rebased branch) - could there be interactions with #984? The only thing I can think of is the detection of units as some species are handled as third bodies rather than regular species. Before the merge, my test runs were consistent with the CI checks.

**PS:** looking through the code, I believe it’s due to the new `Reaction::calculateRateCoeffUnits`, which can no longer rely on third bodies being clearly marked.

[*Edit: **PPS:** @speth ... I verified that there has been a functional clash with #984 (I rebased on a separate branch - ischoegl:fix-967-detect-thirdbodies). I am very hesitant to rebase here as it's not my PR, but a fix would require changing code that is not available on this branch (yet).*]

speth:
Yes, I'd be happy to take a look.

ischoegl:
~@speth ... I parked a fix for the issue on `ischoegl:fix-967-detect-thirdbodies` (which builds and passes all tests) ... had to rebase to include changes introduced by #984 and tweak `ThreeBodyReaction::calculateRateCoeffUnits` (as expected). With those fixes, #967 should be mostly taken care of.~

*Edit: moot ... I created a dedicated PR for the detection of unspecified third-body collision partners.*

ischoegl (approved):
As the detection of third-bodies ended up to be more comprehensive than this original PR while not depending on the proposed change, I decided to propose it as a separate PR (#1015).

PR #973, opened 2021-02-04
Fix some compilation issues and bump to 2.5.0rc1

speth:

- Skip compilation of Sundials Fortran interface files to fix errors about duplicate symbols encountered on macOS and with MinGW
- Fix an issue compiling Fortran samples with gfortran when Cantera is compiled with clang
- Fix some compiler warnings
- Bump version for release candidate

**Checklist**

### DIFF
--- a/interfaces/cython/SConscript
+++ b/interfaces/cython/SConscript
@@ -39,17 +39,16 @@ info = getCommandOutput(localenv['python_cmd'], '-c', script)
 module_ext, inc, pylib, prefix, py_version, numpy_include = info.splitlines()[-6:]
 localenv.Prepend(CPPPATH=[Dir('#include'), inc, numpy_include])
 localenv.Prepend(LIBS=localenv['cantera_libs'])
+
+# Don't print deprecation warnings for internal Python changes.
+# Only applies to Python 3.8. The field that is deprecated in Python 3.8
+# and causes the warnings to appear will be removed in Python 3.9 so no
+# further warnings should be issued.
+if localenv['HAS_CLANG'] and parse_version(py_version) == parse_version('3.8'):
+    localenv.Append(CXXFLAGS='-Wno-deprecated-declarations')
+
 if localenv['OS'] == 'Darwin':
     localenv.Append(LINKFLAGS='-undefined dynamic_lookup')
-
-    # Don't print deprecation warnings for internal Python changes.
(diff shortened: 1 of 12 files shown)


interfaces/cython/SConscript
```diff
@@ -39,17 +39,16 @@ info = getCommandOutput(localenv['python_cmd'], '-c', script)
 module_ext, inc, pylib, prefix, py_version, numpy_include = info.splitlines()[-6:]
 localenv.Prepend(CPPPATH=[Dir('#include'), inc, numpy_include])
 localenv.Prepend(LIBS=localenv['cantera_libs'])
+
+# Don't print deprecation warnings for internal Python changes.
+# Only applies to Python 3.8. The field that is deprecated in Python 3.8
+# and causes the warnings to appear will be removed in Python 3.9 so no
+# further warnings should be issued.
+if localenv['HAS_CLANG'] and localenv['python_version'] == parse_version('3.8'):
```
bryanwweber: If we're using the `python_version` from the original `env` here, it will be set based on the value of `python_cmd` (or the Python running SCons if `python_cmd` isn't given). In that case, do we need to set `py_version` on Line 39 of this file? They should be exactly the same version.

speth: Ah, I missed that we were reading the Python version in this `SConscript` file. I agree that they should be exactly the same. For the sake of making this a more minimal patch, I'll update this to use the local `py_version` variable.

bryanwweber: Sounds good. Just make sure to wrap it in `parse_version()`. That line got missed when we did that change earlier, so now it's reporting the deprecation warnings for me on my Mac again 😄

speth: Are you saying that the current `main` branch is giving you warnings? Because that would be consistent with what I saw while testing this on Ubuntu, where `py_version` is the string `3.8` and whatever `parse_version` returns returns `False` for comparisons to string values.

My fix (at the moment) is to just compare the string values, since it's just the `X.Y` version string, but we could wrap both of them in `parse_version` if there's a reason to.

bryanwweber: Yes, that exactly what I mean. I don't see why not use `parse_version()`, that's what it's there for 😁

bryanwweber (approved):

PR #976, opened 2021-02-11
Docs updates for 2.5.1

bryanwweber:

- Update doxygen configuration file
- Fix all of the Sphinx warnings from Cython docstrings :tada: (Next is the doxygen warnings 😭 )
- Add SCons configuration to enable additional options to be passed to the Sphinx builder. The default turns warnings into errors (because all the Sphinx warnings are fixed!)
- Deprecate the `mixFrac` argument to the `set_mixture_fraction` function, replacing it with `mixture_fraction`.

**Checklist**

### DIFF
--- a/doc/SConscript
+++ b/doc/SConscript
@@ -69,17 +69,18 @@ def extract_matlab_docstring(mfile, level):
     return docstring + '\n'
 
 
-def get_function_name(str):
+def get_function_name(function_string):
     """
         Return the Matlab function or classdef signature, assuming that
         the string starts with either 'function ' or 'classdef '.
     """
-    if str.startswith('function '):
-        sig = str[len('function '):]
-    elif str.startswith('classdef '):
-        sig = str[len('classdef '):]
+    if function_string.startswith('function '):
+        sig = function_string[len('function '):]
+    elif function_string.startswith('classdef '):
+        sig = function_string[len('classdef '):]
     else:
-        print("Unknown function declaration in MATLAB document", str)
+        print("Unknown function declaration in MATLAB document", function_string)
+        sys.exit(1)
 
     # Split the function signature on the equals sign, if it exists.
     # We don't care about what comes before the equals sign, since
--- a/doc/SConscript
+++ b/doc/SConscript
@@ -112,6 +112,7 @@ if localenv['sphinx_docs']:
         'sphinx/conf.py',
         '${sphinx_cmd} -b html -d ${SPHINXBUILD}/doctrees ${SPHINXSRC} ${SPHINXBUILD}/html'))
     env.Alias('sphinx', sphinxdocs)
+    env.Depends(sphinxdocs, env['python_module'])
 
     # Create a list of MATLAB classes to document. This uses the NamedTuple
     # structure defined at the top of the file. The @Data and @Utilities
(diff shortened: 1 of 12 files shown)


doc/SConscript
```diff
@@ -106,11 +107,13 @@ if localenv['doxygen_docs']:
 if localenv['sphinx_docs']:
     localenv['SPHINXBUILD'] = Dir('#build/docs/sphinx')
     localenv['SPHINXSRC'] = Dir('sphinx')
+    localenv['SPHINXOPTS'] = localenv['sphinx_options']
 
     sphinxdocs = build(localenv.Command('${SPHINXBUILD}/html/index.html',
         'sphinx/conf.py',
-        '${sphinx_cmd} -b html -d ${SPHINXBUILD}/doctrees ${SPHINXSRC} ${SPHINXBUILD}/html'))
+        '${sphinx_cmd} ${SPHINXOPTS} -b html -d ${SPHINXBUILD}/doctrees ${SPHINXSRC} ${SPHINXBUILD}/html'))
```
speth: Can't you just use `${sphinx_options}` here, and avoid introducing the `SPHINXOPTS` variable?

interfaces/cython/cantera/thermo.pyx
```diff
@@ -782,8 +783,8 @@ cdef class ThermoPhase(_SolutionBase):
         :param oxidizer:
             Oxidizer species name or mole/mass fractions as a string, array, or
             dict.
-        :param basis: determines if *fuel* and *oxidizer* are given in mole
-            fractions (basis='mole') or mass fractions (basis='mass')
+        :param basis: determines if ``fuel`` and ``oxidizer`` are given in mole
+            fractions (``basis='mole'``) or mass fractions (``basis='mass'``)
```
speth: There are a couple more instances of these parameters that should be formatted consistently in the first couple lines of `set_equivalence_ratio`'s docstring.

speth (changes requested):
Looks pretty good to me. I think there is one more function, `get_equivalence_ratio`, that has messed up formatting similar to the other instances that you are fixing here.

interfaces/cython/cantera/thermo.pyx
bryanwweber: I didn't bother to change these for deprecated functions. Do you think I should?

speth: `set_equivalence_ratio` (the function these changes are in) isn't deprecated, though.

bryanwweber: Thanks, dunno how I missed those!

speth:
I'm still not 100% sold on the immediate bump to 2.5.1, but can you at least make the commit that does that *only* change version numbers? One reason for this is that I find looking at the old "version bump" commits to be helpful in finding all the places that need to be changed when making later "version bump" commits.

bryanwweber:
> but can you at least make the commit that does that _only_ change version numbers?

Done! I looked at your commit for 2.5.0 to see what needed to be bumped here, so I get that 😃  I decided to leave the version numbers in the YAML data files at 2.5.0, since this is only a patch and doesn't have a chance of affecting any of them anyways.

speth:
> I decided to leave the version numbers in the YAML data files at 2.5.0, since this is only a patch and doesn't have a chance of affecting any of them anyways.

Yes, I agree that those version numbers should not change. My thought on this was that we should only change the version numbers in these files either to eliminate pre-release version suffixes or if the file actually changes. So if we're at Cantera 3.0 and `gri30.yaml` hasn't changed, I think it's fine for that file to still say 2.5.0, because it's an indicator of the version of Cantera used to create the file, not the version of Cantera that the file ships with. And in the case of removing the pre-release suffixes, we should verify that the files haven't actually changed (which you did for 2.5.0).

speth (approved):

PR #979, opened 2021-02-18
Compile speedup

speth:
The main goal of this PR is to reduce compile times, which have been creeping upward. It does not address the single biggest single build step, compiling the Cython-generated `_cantera.cpp` file, which is a project unto itself.

- Stop using `fmt` in header-only mode, which speeds up the compilation of each compilation unit that includes `fmt`
- Combine most of the source files defining functions in the `VCS_SOLVE` class. The whole class is a bit too long for one source file, but 2 compilation units is better than 12.
- Use forward declarations to devolve some includes so that fewer header files are linked to each other. In addition offering a small speedup for a full rebuild, this helps limit the number of source files that have to be rebuilt after a change to any header file.
- Clean up the constructors for `ThermoPhase` derivatives. Deprecate a couple of cases which take extra arguments that should be provided using the normal methods for setting parameters (function calls or input files), and eliminate the need for a separate default constructor.
- Stop using and deprecate the confusing `thermo_t` typedef
- Drop CI builds using Python 3.5, which is now end of life
- Disable debug symbols in CI builds
- Tweak the "coverage" build to run a little faster by adding the `NDEBUG` preprocessor define

**Results**

For the builds using GitHub Actions,

- The standard Ubuntu builds are ~40% faster
- The macOS builds are ~16% faster 
- The "docs" build is ~56% faster
- The runs of the examples are ~16% faster
- The Windows builds are ~13% faster, though this is somewhat variable
- The coverage build is ~15% faster

**Checklist**


ischoegl:
👍 for getting rid of `thermo_t`

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -113,20 +113,17 @@ namespace Cantera
...
-    BinarySolutionTabulatedThermo();
-
     //! Construct and initialize an BinarySolutionTabulatedThermo ThermoPhase object
     //! directly from an input file
     /*!
      * This constructor will also fully initialize the object.
      *
      * @param infile File name for the input file containing information
-     *               for this phase
+     *               for this phase. If not specified, and empty phase will be created.
```
bryanwweber:
```suggestion
     *               for this phase. If not specified, an empty phase will be created.
```

bryanwweber:
> * Stop using `fmt` in header-only mode, which speeds up the compilation of each compilation unit that includes `fmt`

The upshot of the header-only mode is that we don't have to add a runtime dependency on `fmt`. Is it feasible to let this be a configurable option? I think for compiled packages/installers (which I guess would just include the Conda `libcantera` package and the Ubuntu package?), it would be an advantage to use the header only mode because the compilation time is less important.

As a counterpoint, there is an overhead associated with maintaining the separate option. I'm not sure which is better/worse!

speth:
> 
> 
> > * Stop using `fmt` in header-only mode, which speeds up the compilation of each compilation unit that includes `fmt`
> 
> The upshot of the header-only mode is that we don't have to add a runtime dependency on `fmt`. Is it feasible to let this be a configurable option? I think for compiled packages/installers (which I guess would just include the Conda `libcantera` package and the Ubuntu package?), it would be an advantage to use the header only mode because the compilation time is less important.
> 
> As a counterpoint, there is an overhead associated with maintaining the separate option. I'm not sure which is better/worse!

If you compile using the git submodule, the `fmt` library code still ends up embedded in the Cantera library, so it doesn't add a runtime dependency unless you build using an external copy of the `fmt` library.

For the Ubuntu package, where we don't (and shouldn't) use our submodules, having a dependency on `fmt` is pretty simple, and I don't think something we should shy away from. I'd argue the same is true for Conda, where `fmt` is available in the main conda channel.

bryanwweber:
> If you compile using the git submodule, the `fmt` library code still ends up embedded in the Cantera library

Even for the shared library? I guess the answer is yes, since there's not a `libsundials.so` file in the build directory. In that case, the change seems fine to me! We'll need to update the conda recipes in parallel with this change. In fact, I wonder if it'd be better to rely on Conda for more of the dependencies...

speth:
> 
> 
> > If you compile using the git submodule, the `fmt` library code still ends up embedded in the Cantera library
> 
> Even for the shared library? I guess the answer is yes, since there's not a `libsundials.so` file in the build directory. In that case, the change seems fine to me! 

Yes, even for the shared library. We bundle everything that we compile into one library file. For things that we don't compile (specified by `system_blah=y`), then we link externally.

> We'll need to update the conda recipes in parallel with this change. In fact, I wonder if it'd be better to rely on Conda for more of the dependencies...

I'd agree that it's probably a good idea, but I don't see why we have to. Doesn't the conda recipe currently use the submodule for `fmt` and `yaml-cpp`? If so, I don't think anything needs to change immediately.

include/cantera/transport/MixTransport.h
```diff
@@ -149,15 +149,9 @@ class MixTransport : public GasTransport
...
-    virtual void init(thermo_t* thermo, int mode=0, int log_level=0);
+    virtual void init(ThermoPhase* thermo, int mode=0, int log_level=0);
 
 protected:
-    //! Calculate the pressure from the ideal gas law
-    doublereal pressure_ig() const {
-        return (m_thermo->molarDensity() * GasConstant *
-                m_thermo->temperature());
-    }
-
```
bryanwweber, line 160: Just double checking that removing this entirely is the intention, since you created a new function in `MultiTransport.h`.

include/cantera/zeroD/ReactorNet.h
```diff
@@ -228,22 +229,18 @@ class ReactorNet : public FuncEval
     //! integration.
     void setNeedsReinit() {
         m_integrator_init = false;
-    }
+}
```
bryanwweber: I think this curly brace should still be indented.

src/kinetics/Kinetics.cpp
```diff
@@ -267,7 +269,7 @@ void Kinetics::checkReactionBalance(const Reaction& R)
     }
 }
 
-void Kinetics::selectPhase(const doublereal* data, const thermo_t* phase,
+void Kinetics::selectPhase(const doublereal* data, const ThermoPhase* phase,
```
bryanwweber: A chance here to change this to `double*` instead of `doublereal*`.

src/zeroD/FlowReactor.cpp
```diff
@@ -66,6 +68,17 @@ void FlowReactor::updateState(doublereal* y)
...
+{
+    m_rho0 = m_thermo->density();
+    m_speed = mdot/m_rho0;
+    m_speed0 = m_speed;
+    m_T = m_thermo->temperature();
+    m_P0 = m_thermo->pressure() + m_rho0*m_speed*m_speed;
+    m_h0 = m_thermo->enthalpy_mass() + 0.5*m_speed*m_speed;
+}
+
+
```
bryanwweber: Looks like an extra line here

bryanwweber (changes requested):
Some minor formatting suggestions/questions. Looks good otherwise, refactoring is nice. Thank you for doing this!

bryanwweber:
One other comment, can you add a "why/how it works briefly" to the commit messages for a045740 and 88ea0f3? I think that'll help if anyone goes to look why those arguments are there.

include/cantera/transport/MixTransport.h
speth: Yes, in the `MixTransport` class, it was a `protected` member of that was not actually used.

In the case of `MultiTransport`, the method is used, so I moved the definition to the `.cpp` file because doing so eliminates the need to have the full definition of `ThermoPhase` available in `MultiTransport.h`.

src/kinetics/Kinetics.cpp
speth: I took the opportunity to update this in a few other places as well.

bryanwweber (approved):

PR #951, opened 2020-12-16
Adding adaptive preconditioner functionality to Cantera (draft)

anthony-walker:
**Changes proposed in this pull request**

This code is adding preconditioning capabilities to Cantera/CVODEs solver interface for reactor networks. It is focused on a specific type of preconditioning but the code is being written with extensibility in mind. It adds a "Preconditioners" file to the numerics portion of cantera which contains a "PreconditionerBase" that is used to write future preconditioners. It also contains a class derived from this called "AdaptivePreconditioner" which is the aforementioned specific preconditioner. 

There are also a couple of functions added to specific reactors. IdealGasReactor is one example. Currently, I added a function to evaluate the energy equation which is also done inside of evalEqs but I wanted to avoid touching core Cantera code as much as possible. In this future, I would like to collaborate with the appropriate people to stream line this. 

**Checklist**

- [x] There is a clear use-case for this code change
- [x] The commit message has a short title & references relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] The pull request is ready for review

PR #982, opened 2021-02-20
Make reactions extendable

### LINKED ISSUE #63: Make reactions extendable
**Abstract**

At the moment, reactions are hard-coded and do not make use of 'factory' creation, i.e. they cannot be easily extended by user code compiled against core Cantera. Adding a `ReactionFactory` would introduce more flexibility. Further, reactions still use *magic numbers* that have been largely removed elsewhere.

**Motivation**

The main motivation would be to replace hard-coded reaction instantiation by an extendable `ReactionFactory` approach. 

Example: `GasKinetics` would become fully flexible, i.e. someone can define a new reaction type externally and link it against cantera (`ctapp` approach, where new C++ lambda functions are registered for an existing Cantera factory object).

**Possible Solutions**

Replace magic number referenced function calls by name-mapped C++ lambda functions calls (which are mainly used for setup; actual kinetics calculations involve templated functions, i.e. there would be no change from the current implementation whatsoever). Use C++ lambda functions to recycle existing code directly.

**References**

* Cantera/cantera#982 
* Cantera/cantera#745 (now closed), and corresponding [`ischoegl:add-reaction-factories`](https://github.com/ischoegl/cantera/tree/reaction-factories) branch
* [ctapp](https://github.com/ischoegl/ctapp) - illustration on how to create external objects and register them in existing Cantera class factories.
* Cantera/cantera#597 (no longer pursued) .. an example for a reaction type that is currently unavailable, and could be added in user code instead of core Cantera.

ischoegl:

-   Add `ReactionFactory` in C++
-   Replace hard-coded instantiation and setup in C++ and Python
-   Remove magic numbers from Python interface
-   **A new abstract base class `ReactionRateBase` allows for flexible implementations** (derived classes are de-virtualized)
-   **A new `MultiRateBase` class is introduced as an alternative to `Rate1` templates**
-   **A new `CustomReaction` in Python uses `Func1` objects**

**Approach**

The main philosophy is to replace hard-coded reaction instantiation by an extendable `ReactionFactory` approach. None of this affects calculated outcomes. I.e. this PR repackages what's already there, with the difference that it becomes very simple to add additional reaction types.

*Example:* `GasKinetics` becomes fully flexible, i.e. someone can define a new reaction type externally and link it against cantera (`ctwrap` approach, where new C++ lambda functions are registered for an existing cantera factory object).

**Implementation**

-   Replace magic number referenced function calls by name-mapped C++ lambda functions calls (which are mainly used for setup; actual kinetics calculations involve templated functions, i.e. there is no change from the current implementation whatsoever). 
-    C++ lambda functions recycle existing code directly

Supersedes #745 (can no longer be opened as it was rebased)

Fixes Cantera/enhancements#63 (a new framework is introduced, but not fully implementated)

*A full replacement of `Rate1` goes beyond the scope of this PR.*

**References:**

- Cantera/enhancements#79
- Cantera/enhancements#84

**Checklist**


ischoegl:
kicking this back to draft - I am planning to extend the framework by a custom Python class to work out some of the issues mentioned in Cantera/enhancements#63 (there are some bits in `GasKinetics.h` that need to be worked out).

ischoegl:
@speth ... I believe this is a pretty decent draft: I'd appreciate if you could provide some feedback at this point.

In addition to `ReactionFactory`, I now have added a Python `CustomReaction`, which is instantiated as
```Python
rxn = CustomReaction(equation='H2 + O <=> H + OH',
                     rate=lambda T: 38.7 * T**2.7 * exp(-3150.15428/T),
                     kinetics=gas)
```
At this point, other reactions can be added from C++, as long as they have a non-null `shared_ptr<RxnRate>` (which is how they are recognized). `RxnRate` itself is an abstract base class that replaces templates, and are loaded into `GasKinetics`.

**PS:** Some of the docstrings could likely be improved, but for the time being I mostly wanted to create a functional framework.

ischoegl:
@speth ... for benchmarking purposes, I added a new `TestReaction` class, which is identical to `ElementaryReaction`, but uses the new abstract base class `RxnRate` instead of `Rate1` templates for reaction rate evaluations. 

Further, a new example (`interfaces/cython/cantera/examples/kinetics/custom_reactions.py`) can be used to illustrate the following comparisons:
1. Base case uses original mechanism
2. One `ElementaryReaction` is replaced by a Python `CustomReaction` (test penalty for Python call)
3. All `ElementaryReaction`s are replaced by `TestReaction`s (test penalty for using `virtual` calls)
```
$ python custom_reactions.py
Average time of 100 simulation runs:
- Original mechanism: 9.84 ms (T_final=2674.97)
- One Python reaction: 11.75 ms (T_final=2674.97)
- Alternative reactions: 10.78 ms (T_final=2674.97)
```
As expected, there is a small penalty for `virtual` functions; the implementation is significantly simpler though.

**PS:** ~There may be a way to split the difference using static polymorphism (curiously recurring template patterns? - which still need an abstract base class for storage purposes) for built-in reactions to recover the (modest) speed penalty. Haven't tested this as I honestly believe that other approaches (e.g. #951) are preferable as expected speed-ups are orders of magnitudes.~ *Update: Don't think that CRTP's will work.*

ischoegl:
Stopping here for feedback. Per Cantera/enhancements#84, some implementations are incomplete, but I'd rather have feedback at this point before creating some large diffs that make code review more difficult.

ischoegl:
Stopping here for feedback. Per Cantera/enhancements#84, some implementations are incomplete, but I'd rather have feedback at this point before creating some large diffs that make code review more difficult.

include/cantera/kinetics/RxnRates.h
```diff
@@ -429,6 +430,133 @@ class ChebyshevRate
...
+    static double m_recipT;
+
+    //! Pressure used for reaction rate evaluation
+    static double m_pressure;
+
+    //! Logarithm of pressure used for reaction rate evaluation
+    static double m_logP;
+
+    //! 10-base logarithm of pressure used for reaction rate evaluation
+    static double m_log10P;
```
speth: Mutable `static` class members are not thread safe.

include/cantera/kinetics/GasKinetics.h
```diff
@@ -67,6 +67,18 @@ class GasKinetics : public BulkKinetics
...
+                    std::function<void(shared_ptr<Reaction>)> f) {
+        m_addRxn[name] = f;
+    }
+
+    //! register a function that modifies a reaction
+    void reg_modRxn(const std::string& name,
+                    std::function<void(size_t, shared_ptr<Reaction>)> f) {
+        m_modRxn[name] = f;
+    }
+
```
speth: Why are these being introduced only to `GasKinetics` class? I'd think that this feature ought to apply at the most general level.

include/cantera/kinetics/Reaction.h
```diff
@@ -37,12 +43,39 @@ class Reaction
...
 
+    //! Get validity flag of reaction
+    bool valid() const {
+        return m_valid;
+    }
+
+    //! Set validity flag of reaction
+    void set_valid(bool valid) {
+        m_valid = valid;
+    }
```
speth: I'd like to avoid introducing a new property on `Reaction` objects just for the purpose of supporting the "skip if undeclared species" option. I do see how continuing to use the `reaction_type` member to carry that information is no longer viable, though.

include/cantera/kinetics/Reaction.h
```diff
@@ -75,6 +108,13 @@ class Reaction
 
     //! Input data used for specific models
     AnyMap input;
+
+protected:
+    //! Flag indicating whether reaction is set up correctly
+    bool m_valid;
+
+    //! Reaction rate used by generic reactions
+    shared_ptr<RxnRate> m_rate;
```
speth: Is the idea that the other reaction types will be eventually be changed to use this structure?

include/cantera/kinetics/Reaction.h
```diff
@@ -286,6 +386,49 @@ std::vector<shared_ptr<Reaction> > getReactions(const XML_Node& node);
...
+                           const Kinetics& kin);
+
+// declarations of setup functions
+void setupElementaryReaction(ElementaryReaction&, const XML_Node&);
+void setupElementaryReaction(ElementaryReaction&, const AnyMap&,
+                             const Kinetics&);
+
+void setupThreeBodyReaction(ThreeBodyReaction&, const XML_Node&);
+void setupThreeBodyReaction(ThreeBodyReaction&, const AnyMap&,
+                            const Kinetics&);
```
speth, line 428: If these functions are going to become public (as opposed to only defined in `Reaction.cpp`) they may as well become member functions of the `Reaction` classes, right?

speth (commented):
Thank you for assembling this PR, as I think it helps clarify some of the challenges and opportunities in how we handle reactions and reaction rates.

Adding the ReactionFactory class and retiring the magic numbers for Reaction types both make sense to me. I would kind of like to leave everything to do with XML alone. Modifying or introducing new code related to the XML input files seems like needless churn at this point.

I appreciate the experiment with evaluating the performance of handling rates with both virtual function calls and doing the rate evaluation in Python. I was curious about the behavior for larger mechanisms, so I tried a modified version of your example using the `nDodecane_Reitz.yaml` mechanism, and found that the slowdown for using `TestReaction` was 15-20%. I think there is room for some optimization (profiling suggest that looping over the `std::map` of reactions takes a noticeable amount of time) but there may still be performance reasons for having some kind of specialization for at least the simplest / most common reaction types. The overhead of a single Python rate is a bit more than I had hoped for, as well.

I think the `reg_addRxn` and `reg_modRxn` functions are still unnecessary. Unless a user introduces a new Kinetics class, I'm having trouble seeing how they could use a new reaction type that was meaningfully different from the base `Reaction` class with a `rxnRate` class of their creation, which is the one case that circumvents the `reg_addRxn` functions here.

The generalization of `RxnRate` introduced here is much more restrictive than the suggestion I outlined in Cantera/enhancements#63. The only data available to calculate a reaction rate in this case are the temperature and pressure. Notwithstanding the thread safety issues introduced by `static` member variables, this structure would be insufficient to implement the kinds of rates that are not currently available, for example those using the Blowers-Masel formulation (Cantera/enhancements#66) or rates using a partial equilibrium assumption as in the Users' Group post you noted in Cantera/enhancements#63. For users interested only in modified functions of T and P, we already the Chebyshev rate type which is able to represent arbitrary smooth functions of T and P.

With that said, I'm not sure I see merging this PR in its current form as actually getting us closer to the goal of Cantera/enhancements#63 (not to mention Cantera/enhancements#79). I'd like to suggest leaving it open for the time being as a point of reference for the discussion on Cantera/enhancements#63. I'll try to find time soon to post some of the alternative structures that I think might actually achieve the level of extensibility that we want.

include/cantera/kinetics/RxnRates.h
ischoegl: That's a fair point ... but I believe that there are versions that are thread safe (`std::atomic<>`) or other implementations that accomplish the same thing.

include/cantera/kinetics/GasKinetics.h
ischoegl: These are mainly introduced to recycle existing code and would be eliminated eventually (they are not necessary if `RxnRate` were to be implemented).

include/cantera/kinetics/Reaction.h
ischoegl: Not having the magic number any longer is exactly the reason for that introduction. There are ways around that (like an *invalid* reaction rate type), which i would address if this were to be merged.

include/cantera/kinetics/Reaction.h
ischoegl: Exactly. This is the main difference to the current approach ... the new framework would create a one-to-one mapping of reaction and reaction rate objects.

include/cantera/kinetics/Reaction.h
ischoegl: These are just place holders to recycle existing code. For XML I wouldn't want to change anything beyond, but for YAML you are correct. It would just create a much larger diff if these were moved already and imho only makes sense for reaction types that are replaced by new versions.

ischoegl:
@speth ... thank you for your feedback. I'll respond to some of those, without taking further action for the time being:

> Modifying or introducing new code related to the XML input files seems like needless churn at this point.

Anything 'changed' for XML was cut & paste to recycle existing code. _PS: shuffling of code was necessary for the creation of a consistent Python interface._

> I tried a modified version of your example using the `nDodecane_Reitz.yaml` mechanism, and found that the slowdown for using TestReaction was 15-20%.  [...] profiling suggest that looping over the std::map of reactions takes a noticeable amount of time

Interesting. ~Presumably storing things in a vector would be faster, which I haven't tested.~ _Edit: the same 15-20% penalty existed for `gri30.yaml`, and `std::vector` is slightly faster._

>  having some kind of specialization for at least the simplest / most common reaction types.  

~I tried keeping `elementary` reactions separate with a `std::map<int, ArrheniusRate>`, but didn't see any difference (for this version, all methods should be defined at compilation time).~ ... _Edit: enforcing `final` derived objects gave small performance gains._

> I think the `reg_addRxn` and `reg_modRxn` functions are still unnecessary. 

~Agreed - I opted for this to recycle existing code (I implemented that early on). As you probably noticed, the newly introduced classes bypass those calls.~ _Edit: removed those as this was clearly a distraction._

> The only data available to calculate a reaction rate in this case are the temperature and pressure. 

~You are correct, but concentrations are easily added. I actually anticipated including concentrations, but thought that it wasn't needed at the moment and I would add those methods at a later point.~ _Edit: clarified the interface to accommodate rates that depend on the entire gas state_

> [...] thread safety issues introduced by `static` member variables [...]

~That is absolutely true - for this approach, the entire implementation hinges on having to copy this only once without the current external loop of the templated approach in `Rate1`. There are, however, thread-safe ways of doing the same thing.~ _Edit: implemented a version that is thread-safe._

include/cantera/kinetics/GasKinetics.h
ischoegl: Removed them as they detract from the substance.

include/cantera/kinetics/RxnRates.h
ischoegl: Thread-safety is a major issue - I replaced the implementation by passing a data container.

ischoegl:
@speth ... looked at this some more and was able to figure out how to recover most of the speed penalties. The biggest difference was to store specialized objects as vectors directly (e.g. a specialized `std::vector<ArrheniusRate>` is much faster than a generic `std::vector<shared_ptr<RxnRate>>`, which itself is somewhat faster than `std::map<size_t, shared_ptr<RxnRate>>`).

The new scores are:
```
- Original mechanism: 148.14 ms (T_final=2736.60)
- One Python reaction: 156.36 ms (T_final=2736.60) ... +5.55%
- Alternative reactions: 150.46 ms (T_final=2736.60) ... +1.56%
```
(Alternate reactions use specialized vectors). For Reitz, it is
```
Average time of 50 simulation runs for 'nDodecane_Reitz.yaml' (c12h26)
- Original mechanism: 481.02 ms (T_final=2823.43)
- Alternative reactions: 511.15 ms (T_final=2823.43) ... +6.26%
```

From my perspective, I believe that this is a workable basis for a truly flexible approach that would (if fully implemented) resolve Cantera/enhancements#63. The main difference to the current framework is that `RxnRate` objects are no longer wrapped into templates as before, and thus become directly accessible while being polymorphic. I expect that it may be possible to shave off more time, but I am stopping here.

**PS:** I believe I took care of most of the issues that were pointed out in initial feedback.

ischoegl:
To summarize the approach needed to add a new reaction type using the framework proposed in this PR:

1. Define a new reaction in the C++ layer (derived from `Reaction`), and register it in the (new) `ReactionFactory`; 
2. The reaction rate uses a shared `m_rxnRate` pointer to an object that is derived from the (new) `RxnRate` 
3. [optional] expose the new reaction class to Python by adding a new Python class. Specifying a `reaction_type` will instantiate the correct object based on the `ReactionFactory` using the existing framework.
4. [optional] if applicable, expose a new reaction rate class to Python. A `RxnRateFactory` that parallels `ReactionFactory` would be a logical extension of the framework, but goes beyond the scope of the current implementation.

The approach outlined above is fully implemented for a `TestReaction` object (based on `ArrheniusRate`) that was created for testing purposes. It bypasses the existing `Rate1` templates.

include/cantera/kinetics/Reaction.h
speth: I'm happy to defer doing that to later (post-XML removal). Maybe the `AnyMap` versions should be marked as "internal; may be changed without notice in future versions"?

include/cantera/kinetics/ReactionFactory.h
```diff
@@ -0,0 +1,123 @@
...
+ * Factory class to construct reaction function calculators.
+ * The reaction factory is accessed through static method factory:
+ *
+ * @code
+ * Reaction* f = ReactionFactory::factory()->newReaction(type, c)
+ * @endcode
+ *
+ * @ingroup reactionGroup
+ */
+class ReactionFactory : public Factory<Reaction>
```
speth, line 35: I think if you make the base class type `Factory<Reaction, const AnyMap&, const Kinetics&>`, you'll be able to use the `reg` and `create` functions provided by the base class and won't have to implement `setup_AnyMap` etc. Or at least, this was the intention behind allowing additional template parameters to the `Factory` class.

include/cantera/kinetics/GasKinetics.h
```diff
@@ -89,6 +89,15 @@ class GasKinetics : public BulkKinetics
...
 
+    //! Vector of generic reaction rates
+    std::vector<shared_ptr<RxnRate>> m_rxn_rates;
+
+    //! Vector of specialized arrhenius reaction rates
+    std::vector<ArrheniusRate> m_arrhenius_rates;
+
+    std::map<size_t, size_t> m_rxn_indices;
+    std::map<size_t, size_t> m_arrhenius_indices;
+
```
speth: Can we move all of this up to the base `Kinetics` class? I don't see anything that makes this idea particular to gas kinetics. It would be nice to be able to represent some of the more complex / ad hoc surface rate formulations with this generic framework, and I know that there are people who would like to eventually add liquid and solid phase kinetics, which would presumably come with new rate types as well.

speth:
Thanks for the continued work on this, @ischoegl. I think this is moving toward being a very robust generalization of handling reaction rates. The one part that I think still needs some additional flexibility is handling data that should be available for updating the rate constants. I've been contemplating a number of different structures, and here's the one that I think provides the most flexibility without introducing too much new complexity.

```c++
// A base class for evaluating all reactions of a particular type.
// Because this class has no template parameters, the Kinetics object can store
// all of these rate coefficient evaluators as a vector<shared_ptr<MultiRateBase>>.
class MultiRateBase
{
public:
    virtual ~MultiRateBase() {}
    virtual void getRateConstants(const ThermoPhase& reacting_phase,
                                  const Kinetics& kin, double* kf) const = 0;
    virtual void update_T(const ThermoPhase& reacting_phase, const Kinetics& kin) = 0;
    virtual void update_P(const ThermoPhase& reacting_phase, const Kinetics& kin) = 0;
    virtual void update_C(const ThermoPhase& reacting_phase, const Kinetics& kin) = 0;
};

// This class is analogous to Rate1, but uses a generic data structure "m_shared"
// to collect common information used by all instances of a particular rate type
// coefficient. I've left out a bunch of the infrastructure here that should be
// basically the same as what is already in Rate1.
template <class RateType, class DataType>
class MultiRate : public MultiRateBase
{
public:
    virtual void getRateConstants(const ThermoPhase& reacting_phase,
                                  const Kinetics& kin, double* kf) const {
        for (size_t i = 0; i < m_rates.size(); i++) {
            kf[m_rxn[i]] = m_rates[i].rateConstant(m_shared);
        }
    }

    virtual void update_T(const ThermoPhase& reacting_phase, const Kinetics& kin) {
        // update common data once for each reaction type
        m_shared.update_T(reacting_phase, kin);
        if (RateType::uses_update_T()) {
            // update reaction-specific data for each reaction. Like in Rate1, this loop 
            // should be efficient because it involves no virtual function calls, and all of
            // the rate objects are contiguous in memory
            for (auto& rate : m_rates) {
                rate.update_T(m_shared, reacting_phase, kin);
            }
        }
    }

    virtual void update_P(const ThermoPhase& reacting_phase, const Kinetics& kin) {
        m_shared.update_P(reacting_phase, kin);
        if (RateType::uses_update_P()) {
            for (auto& rate : m_rates) {
                rate.update_P(m_shared, reacting_phase, kin);
            }
        }
    }

    virtual void update_C(const ThermoPhase& reacting_phase, const Kinetics& kin) {
        m_shared.update_C(reacting_phase, kin);
        if (RateType::uses_update_C()) {
            for (auto& rate : m_rates) {
                rate.update_C(m_shared, reacting_phase, kin);
            }
        }
    }

protected:
    std::vector<RateType> m_rates;
    std::vector<size_t> m_rxn;
    DataType m_shared;
};

// For each rate coefficient parameterization, there are two classes to implement

class CustomRateData
{
public:
    void update_T(const ThermoPhase& reacting_phase, const Kinetics& kin) {
        m_logT = std::log10(reacting_phase.temperature());
        m_recipT = 1 / reacting_phase.temperature();
    }

    void update_P(const ThermoPhase& reacting_phase, const Kinetics& kin) {}
    void update_C(const ThermoPhase& reacting_phase, const Kinetics& kin) {}

    double m_logT;
    double m_recipT;
};

// Base class for reaction rates. The Reaction class would hold a shared_ptr<ReactionRate>.
class ReactionRate
{
public:
    virtual shared_ptr<MultiRateBase> getMultiRate() const = 0;
};

// Same basic idea as existing "RxnRate" classes; no virtual methods except for
// getMultiRate
class CustomRate : public ReactionRate
{
public:
    double rateConstant(CustomRateData& shared_data) const {
        return m_A * std::exp(m_b*shared_data.m_logT - m_Ea*shared_data.m_recipT);
    }

    void update_T(const CustomRateData& shared_data,
                  const ThermoPhase& reacting_phase, const Kinetics& kin) {}
    void update_P(const CustomRateData& shared_data,
                  const ThermoPhase& reacting_phase, const Kinetics& kin) {}
    void update_C(const CustomRateData& shared_data,
                  const ThermoPhase& reacting_phase, const Kinetics& kin) {}

    // At the individual rate level, I'm not sure how often the update_T method
    // will be used, but I put it in for the sake of generality
    static bool uses_update_T() { return false; }
    static bool uses_update_P() { return false; }
    static bool uses_update_C() { return false; }

    // This should be called within Kinetics once for each new type of reaction
    // that gets added
    virtual shared_ptr<MultiRateBase> getMultiRate() const {
        return make_shared<MultiRate<CustomRate, CustomRateData>>();
    }

protected:
    double m_A, m_b, m_Ea;
};

```

This is definitely not complete, but I figured it would be better to just share this structure and let you run with it, since you seem to be eager to move forward with this project, and I don't want to make you wait for me to find the time to implement this myself. While there is admittedly some complexity here, I think it will pay off in the complexity that it removes from the Kinetics  classes by eliminating the current glaring abstraction violation where `Kinetics` needs to know about specific reaction types. Here, it would be able to operate entirely at the level of the `ReactionRate` and `MultiRateBase` classes.

One wrinkle that I don't have a resolution for here is a way to calculate k(T, P) in the absence of the `ThermoPhase` / `Kinetics` classes, in the cases where that's logically possible, and which I know is something you were trying to achieve.

ischoegl:
@speth ... thanks for typing up your thoughts! If I'm not mistaken, one main difference is that you have generalized the `State` data container concept to a `CustomRateData` class, which makes sense. Having dual `update_T` routes (either via `CustomRateData` or `CustomRate`) is something I'll think about, although the motivation is quite clear. I have more thoughts that I'll share at a later point, but I have one central question about the use of templates in this context. 

While I believe I understand your setup (it's still resembles `Rate1`), the implementation doesn't look like it's as flexible as what I had proposed. You will need to know both `RateType` and `DataType` used in `template <class RateType, class DataType> class MultiRate` a priori (i.e. known at compile time as the various `Kinetics` objects need to know what they're evaluating). 

Unless I'm overlooking something, this means that this approach won't extend to a `ReactionFactory`: imagine that someone may want to add a new reaction type in external C++ code that compiles *against* stock Cantera (I have shown that this approach works for `Reactor` and `IdealGasFlow` objects here: https://github.com/ischoegl/ctapp). While it's possible to compile new `MultiRate` objects externally, there wouldn't be a way to make the `Kinetics` handler aware of these externally defined classes, which means that corresponding reactions would be unusable, correct?

include/cantera/kinetics/Reaction.h
```diff
@@ -18,12 +18,18 @@ class Kinetics;
 class Falloff;
 class XML_Node;
 
-//! Intermediate class which stores data about a reaction and its rate
+//! Abstract base class which stores data about a reaction and its rate
 //! parameterization so that it can be added to a Kinetics object.
 class Reaction
 {
 public:
+    explicit Reaction();
```
speth: `explicit` is mostly useful only for preventing surprising conversions for single-argument constructors. I don't think it has any meaning for the default constructor.

include/cantera/kinetics/Reaction.h
```diff
@@ -37,12 +43,39 @@ class Reaction
...
+
+    //! Set validity flag of reaction
+    void set_valid(bool valid) {
+        m_valid = valid;
+    }
+
+    //! Get reaction rate pointer
+    shared_ptr<RxnRate> rxnRate() {
+        return m_rate;
+    }
```
speth: Broadly speaking, I think I'd like to avoid using the tortured abbreviation `RxnRate`, even if it does derive from the existing `RxnRate.h` header file. Maybe `rateConstant()` for the accessor method, and `RateConstant` for the class?

include/cantera/kinetics/RxnRates.h
```diff
@@ -429,6 +430,187 @@ class ChebyshevRate
...
+        return eval(State(T, P, conc));
+    }
+
+    // [...] other overrides are not created (yet)
+
+    //! Evaluate reaction rate derivative (with respect to temperature)
+    virtual double ddT(const State& state) const {
+        throw CanteraError("RxnRate::ddT",
+                           "Not (yet) implemented by derived RxnRate object.");
+    }
```
speth: I'd recommend leaving this business of derivatives for another time, and unused stub methods don't really add much clarity to understanding how this would work or where the real difficulties are.

include/cantera/kinetics/ReactionFactory.h
```diff
@@ -0,0 +1,123 @@
+/**
+ *  @file ReactionFactory.h
+ *  Factory class for reaction functions. Used by classes
+ *  that implement kinetics
+ *  (see \ref reactionGroup and class \link Cantera::Reaction Reaction\endlink).
```
speth, line 5: I think this header file should be added to the broader `cantera/kinetics.h` file, especially since the declarations for the existing `newReaction` functions got moved here.

include/cantera/kinetics/RxnRates.h
```diff
@@ -429,6 +430,187 @@ class ChebyshevRate
...
+
+//! Custom Python reaction rate depending only on temperature
+/**
+ * The rate expression is provided by external Python code taking a single
+ * argument (temperature) and does not use a formalized parameterization.
+ *
+ * @warning This class is an experimental part of the %Cantera API and
+ *    may be changed or removed without notice.
+ */
+class CustomPyRate final : public RxnRate
```
speth: There isn't really anything specific to Python about this class, right? All that matters is that it uses a `Func1` object to define k(T), so maybe there's a more suitable name.

interfaces/cython/cantera/reaction.pyx
```diff
@@ -61,20 +56,53 @@ cdef class Reaction:
...
+        Wrap a C++ Reaction object with a Python object of the correct derived type.
+        """
+        # ensure all reaction types are registered
+        if not(_reaction_class_registry):
+            def all_subclasses(cls):
+                return set(cls.__subclasses__()).union(
+                    [s for c in cls.__subclasses__() for s in all_subclasses(c)])
+            # update global reaction class registry
+            _reaction_class_registry.update({getattr(c, 'reaction_type'): c
+                                             for c in all_subclasses(Reaction)})
```
speth: Shouldn't this use the `_all_reaction_objects` method defined above? Or that unused method should be removed. Also, I have an alternative that doesn't involve as many complicated list/dict comprehensions to suggest:
```python
def update_registry(cls):
    for c in cls.__subclasses__():
        _reaction_class_registry[getattr(c, 'reaction_type')] = c
        update_registry(c)

**Comment by ischoegl** [comment]
@speth ... thanks again for your comments (ended up deleting my initial response as I had overlooked a critical detail).

So the main difference between our approaches is an intermediate object (`MultiRate`) between rate objects and kinetics manager, where the overhead of `virtual` methods is more manageable as all rates of a given reaction rate type are managed within `MultiRate`. The different `CustomRate::update_T`'s etc. are mostly bypassed but can be used as needed (I agree that `CustomRate::update_T` is unlikely to be used). Most repetitive calculations are handled by `CustomRateData::update_T`, which corresponds to the `State` data container that I used in the last incarnation of my draft.

Regarding the `k(T)`/`k(T,P)`/`k(T,P,C)`, it's definitely possible to handle things (and relatively straight-forward *if* de-virtualization is used): I'm pretty sure that `CustomRate::rateConstant` could be `virtual` as long as the class is `final`. Same for some of the other service functions, which would simplify the code a bit.

Overall, this certainly looks doable.

File: interfaces/cython/cantera/reaction.pyx [hunk 14]
```diff
@@ -353,7 +381,7 @@ cdef class Arrhenius:
             return self.rate.activationEnergy_R() * gas_constant
 
     def __repr__(self):
-        return 'Arrhenius(A={:g}, b={:g}, E={:g})'.format(
+        return '<Arrhenius: A={:g}, b={:g}, E={:g}>'.format(
```
speth: What's the motivation for this change? Per the [Python documentation](https://docs.python.org/3.9/library/functions.html?highlight=repr#repr), the angle bracket notation is normally used when the returned string doesn't provide the full information needed to reconstruct the object.

interfaces/cython/cantera/reaction.pyx
```diff
@@ -375,12 +403,155 @@ cdef copyArrhenius(CxxArrhenius* rate):
     return r
 
 
+cdef class _RxnRate:
+
+    def __repr__(self):
+        return "<{}>".format(pystr(self.base.type()))
```
speth: I'd recommend still including the `id` of the object in this string, e.g. 
```suggestion
    def __repr__(self):
        return "<{} at {:0x}>".format(pystr(self.base.type()), id(self))
```

src/kinetics/KineticsFactory.cpp
```diff
@@ -10,6 +10,7 @@
 #include "cantera/kinetics/InterfaceKinetics.h"
 #include "cantera/kinetics/EdgeKinetics.h"
 #include "cantera/kinetics/importKinetics.h"
+#include "cantera/kinetics/ReactionFactory.h"
```
speth: Why is this needed?

src/kinetics/ReactionFactory.cpp
```diff
@@ -0,0 +1,245 @@
...
+    reg_AnyMap("interface",
+               [](Reaction* R, const AnyMap& node, const Kinetics& kin) {
+                   setupInterfaceReaction(*(InterfaceReaction*)R, node, kin);
+               });
+
+    // register electrochemical reactions
+    reg("electrochemical", []() { return new ElectrochemicalReaction(); });
+    addAlias("electrochemical", "butlervolmer_noactivitycoeffs");
+    addAlias("electrochemical", "butlervolmer");
+    addAlias("electrochemical", "surfaceaffinity");
```
speth: I'd really like to avoid dragging some of these dead ends into any new code. These reaction types don't exist in YAML, and I'm pretty sure they never worked correctly even in the XML format.

More broadly, this kind of thing is my main motivation for suggesting that all the XML stuff should just be left as-is. I guess the one concern is whether that gets in the way of converting the existing rate types to use the new, generic formulation or not. It's not clear to me whether we would be able to change the existing classes without a round of deprecations, so if those deprecated classes were still used with XML for Cantera 2.6 and then both XML and the old rate classes go away together in 3.0, it would be fine.

speth (changes requested):
I think this is looking interesting. My main suggestion is about really generalizing the state information available to custom rate classes, as described in my comment above. The comments here are mostly more minor manners.

include/cantera/kinetics/Reaction.h
ischoegl: will be replaced by `MultiRateBase` - moot

include/cantera/kinetics/GasKinetics.h
ischoegl: These would be replaced by `MultiRate` objects; but no objection to moving those to `Kinetics`.

include/cantera/kinetics/Reaction.h
ischoegl: Np about avoiding tortured abbreviations! I'd suggest `ReactionRate` with derived names `ArrheniusRate` etc. instead though as I have a pretty strong aversion to perpetuating the misnomer 'constant' more than necessary.

src/kinetics/ReactionFactory.cpp
ischoegl: No problem removing those aliases - I created those a year ago (prior to the discussion and ensuing deprecation of some of these).

src/kinetics/KineticsFactory.cpp
ischoegl: it requires `newReaction` ... adding `ReactionFactory.h` to `Kinetics.h` per your suggestion makes this no longer necessary.

include/cantera/kinetics/RxnRates.h
ischoegl: Correct. Renaming to `CustomFunc1Rate` (same for `CustomFunc1Reaction`).

I guess this is actually more powerful than anticipated! ... :tada:

interfaces/cython/cantera/reaction.pyx
ischoegl: This was mostly for aesthetics ... reactions use angle brackets. This specific object will disappear and new objects will use angle brackets (see other comment).

Reverting as it will be moot eventually.

include/cantera/kinetics/ReactionFactory.h
ischoegl: I admit that there may be opportunity to streamline further, but while XML is still around, I'm treading carefully. This doesn't mean that this shouldn't be taken care of eventually.

interfaces/cython/cantera/reaction.pyx
ischoegl: Good catch. This is superseded by the `wrap` method and there's no longer a need for it ... 

I may incorporate the suggestion into `wrap` at a later time - an initial attempt broke CI.

include/cantera/kinetics/GasKinetics.h
speth: Great. I think doing that would (eventually) eliminate the need for most specializations of `Kinetics` other than `BulkKinetics` and `InterfaceKinetics`.

include/cantera/kinetics/Reaction.h
speth: Yeah, I agree that the use of the word 'constant' is not great. I was trying to avoid the redundancy of `Reaction.reactionRate`, but it's not that bad, and `ReactionRate` is a good name for the class if not the method of `Reaction`. We unfortunately can't just use `rate` since `ElementaryReaction` already has a member with that name. I guess we'll eventually be able to deprecate that and name this getter `rate`.

I'm 👍 on `FooRate` for the various types, although we may run into a little bit of trouble there since we already have a `ChebyshevRate` class.

interfaces/cython/cantera/reaction.pyx
ischoegl: Keeping this open as a reminder.

include/cantera/kinetics/ReactionFactory.h
ischoegl: I thought about it some more and don't see a clean way to handle this at the moment (XML creates some wrinkles). As the factory approach needs to work for both XML and AnyMap, I decided to add a `@todo` in the header rather than solving this right now.

include/cantera/kinetics/GasKinetics.h
ischoegl: I believe it may make sense to keep two separate templated `MultiRateBase` classes that handle `BulkKinetics` and `InterfaceKinetics`, respectively. As such, I added a `m_bulk_rates` to `BulkKinetics`, and left interface-specific routines for a later time.

include/cantera/kinetics/RxnRates.h
ischoegl: Derivatives are the main incentive for me to work on this - it's hard to justify spending the time otherwise ...

ischoegl:
@speth ... at this point, the `MultiRateBase` framework you outlined is implemented (it turned out to be a hybrid of my former de-virtualized methods and the templates you suggested). The implementation was relatively straight-forward as much of the existing plumbing didn't have to change. For the time being, I focused on `MultiBulkRates` that should be flexible enough to handle all reaction types within `BulkKinetics` (I'd suggest to create a separate `MultiInterfaceRates` template class to allow for simpler signatures). Also, I am not separating `update_T` etc. for the time being as I mainly wanted to get to the next phase for this proof of concept.

The framework is now handled by two new header files (the old `RxnRates.h` still exists, but became too busy):
* `include/cantera/kinetics/ReactionRate.h` ... new reaction rate objects (as a variation of my old approach)
* `include/cantera/kinetics/MultiRate.h` ...  templated reaction rate evaluators

Benchmarks are more or less unchanged from before:

```
- Original mechanism: 143.00 ms (T_final=2736.60)
- One Python reaction: 151.73 ms (T_final=2736.60) ... +6.10%
- Alternative reactions: 147.98 ms (T_final=2736.60) ... +3.48%
```

and

```
Average time of 50 simulation runs for 'nDodecane_Reitz.yaml' (c12h26)
- Original mechanism: 455.79 ms (T_final=2823.43)
- One Python reaction: [not run]
- Alternative reactions: 487.60 ms (T_final=2823.43) ... +6.98%
```

**PS:** ~It looks like GitHub Actions is currently hung/stuck, so I'm leaving this comment before CI finishes (`scons test` was successful on my machine).~ *resolved*

interfaces/cython/cantera/func1.pyx
```diff
@@ -63,6 +63,9 @@ cdef class Func1:
 
     def __init__(self, c):
         if hasattr(c, '__call__'):
+            if type(c).__name__ == 'method':
+                raise TypeError("Unable to use callback from 'method'.")
+
```
speth: What's the problem here? Using bound member functions as `Func1` objects works just fine, to the best of my knowledge.

ischoegl: Have a look at `interfaces/cython/cantera/test/test_reaction.py` (specifically `TestCustom::setUp`). For whatever reason the type is modified when you assign `Func1` to a static member variable. Once it's a `method`, a nasty error ensues (which the code above catches). Overwriting the static variable curiously fixes the behavior.

include/cantera/kinetics/Reaction.h
ischoegl: Interesting comment. If things proceed as anticipated (i.e. framework is adopted), `rate` and various other member variables of other reaction classes would become part of a `ReactionRate` object handled by `MultiRateBase` and thus are no longer attached to the reaction object itself. As they're member variables, it's impossible to deprecate them - thoughts?

One way would be to rename `ElementaryReaction::rate` to `ElementaryReaction::rate2` and introduce the shorter getter name right away (which would avoid renaming/deprecation at a later point).

ischoegl: ~Update: added a commit that renames member variables to `rate2` in order to pre-empt a future change.~

Adding an intermediate `Reaction2` object defining `Reaction2::rate` avoids breakages while keeping the shorter name.

ischoegl:
Force-pushed as I found a better solution to allow for `Reaction2::rate`/`Reaction2::setRate` getters/setters (the intermediate object `Reaction2` can be eliminated once the new framework is fully adopted and old objects are deprecated). At this point I'll stop for comments.

**PS:** As of this morning, it looks like boost is - at least partially - broken (#990)

**PPS:** Overall, I'd propose to limit the current PR to the creation of the `MultiRateBase` framework (which should be more or less done - I'll finish doc strings once things look like they are converged). I assume once everything is updated this will end up around 2k lines. Subsequent PR's can transition existing reaction objects to the new framework (I envision 2 PR's for remaining `BulkKinetics`, - one for everything except `Falloff`, and one for `Falloff`, - and eventually a third PR for `InterfaceKinetics`). Breaking this up should also help to limit conflicts with other ongoing efforts (blowers-masel, electrochemistry, etc.).

include/cantera/kinetics/Reaction.h
ischoegl: For the time being, I cannot think of a better solution (would have to change the reaction type *after* creation).

interfaces/cython/cantera/func1.pyx
speth: If I'm correctly interpreting what you had initially tried in that test correctly, it was this, right?
```python
class TestCustom(TestElementary):
    _rate = lambda T: 38.7 * T**2.7 * exp(-3150.15428/T)
```
The resulting behavior and error is indeed a somewhat surprising, but I guess it makes sense if you consider that this is basically the equivalent of writing
```python
class TestCustom(TestElementary):
    def _rate(T):
        return 38.7 * T**2.7 * exp(-3150.15428/T)
```
where it's more clear that the method ought to have a `self` parameter, or be declared with the `@staticmethod` decorator.

Disabling the use of `method` for `Func1` as above prevents the following use case, which I think we should still allow:
```python
class MyRate:
    def __init__(self, A, b, Ta):
        self.A = A
        self.b = b
        self.Ta = Ta
    def eval(self, T):
        return self.A * T ** self.b * np.exp(-self.Ta/T)

rate = MyRate(2e8, 0.1, 2100)
rxn = ct.CustomReaction(equation=..., rate=rate.eval, kinetics=gas)
```

ischoegl: Not a problem to roll it back - I 'circumnavigated' the issue in the end so there's no clear need for the exception. 

Your comments make a lot of sense, so there's nothing wrong with `func1` itself.

PS: *rolled back changes*

src/kinetics/ReactionRate.cpp
```diff
@@ -0,0 +1,44 @@
+//! @file RxnRates.cpp
```
speth:
```suggestion
//! @file ReactionRate.cpp
```

include/cantera/kinetics/ReactionRate.h
```diff
@@ -0,0 +1,255 @@
...
+    virtual double eval(double T, double P) const override {
+        return eval(DataType(T, P));
+    }
+
+    virtual double eval(const ThermoPhase& bulk) const override {
+        return eval(DataType(bulk));
+    }
+
+    //! Evaluate derivative of reaction rate with respect to temperature
+    virtual double ddT(const DataType& shared_data) const {
```
speth, line 111: Clang gives me lots of warning about this kind of thing -- if you use `override` once in a class, you need to use it everywhere else in the class where it applies.
```suggestion
    virtual double ddT(const DataType& shared_data) const override {
```

include/cantera/kinetics/MultiRate.h
```diff
@@ -0,0 +1,115 @@
...
+    virtual void add(const size_t rxn_index,
+                     const shared_ptr<ReactionRateBase> rate) override {
+        m_indices[rxn_index] = m_rates.size();
+        m_rates.push_back(*std::dynamic_pointer_cast<RateType>(rate));
+        m_rxn.push_back(rxn_index);
+    }
+
+    virtual bool replace(const size_t rxn_index,
+                         const shared_ptr<ReactionRateBase> rate) override {
+        if (rate->type() != RateType::staticType()) {
```
speth: Would writing this as follows eliminate the need for the `staticType()` function?
```suggestion
        if (typeid(rate) != typeid(RateType)) {
```

include/cantera/kinetics/MultiRate.h
```diff
@@ -0,0 +1,115 @@
...
+ *     may be changed or removed without notice.
+ */
+template <class RateType, class DataType>
+class MultiBulkRates final : public MultiRateBase
+{
+public:
+    virtual void add(const size_t rxn_index,
+                     const shared_ptr<ReactionRateBase> rate) override {
+        m_indices[rxn_index] = m_rates.size();
+        m_rates.push_back(*std::dynamic_pointer_cast<RateType>(rate));
```
speth: Since this class is just making a copy of the `RateType` object, there's not much benefit to having it take a `shared_ptr` argument. The following modification would also cause a type mismatch to raise an exception rather than resulting in a null pointer dereference as is currently the case:
```suggestion
    virtual void add(const size_t rxn_index,
                     const ReactionRateBase* rate) override {
        m_indices[rxn_index] = m_rates.size();
        m_rates.push_back(dynamic_cast<RateType&>(*rate));
```

include/cantera/kinetics/ReactionRate.h
```diff
@@ -0,0 +1,255 @@
...
+
+    //! Evaluate reaction rate derivative based on bulk phase
+    virtual double ddT(const ThermoPhase& bulk) const = 0;
+};
+
+
+//! Class template for reaction rate definitions with specialized DataType
+/**
+ * This class template ensures that derived objects are aware of specialized
+ * data types, which are passed by `MultiRateBase` evaluators.
```
speth:
```suggestion
 * data types, which are passed by MultiRateBase evaluators.
```

include/cantera/kinetics/ReactionRate.h
```diff
@@ -0,0 +1,255 @@
...
+ *
+ * @warning This class is an experimental part of the %Cantera API and
+ *    may be changed or removed without notice.
+ */
+template <class DataType>
+class ReactionRate : public ReactionRateBase
+{
+public:
+    //! Constructor
+    ReactionRate() = default;
```
speth: I don't think this is necessary, is it?

interfaces/cython/cantera/reaction.pyx
```diff
@@ -741,17 +901,114 @@ cdef class ChebyshevReaction(Reaction):
...
+    reaction_type = "interface"
 
     property coverage_deps:
         """
         Get/Set a dict containing adjustments to the Arrhenius rate expression
         dependent on surface species coverages. The keys of the dict are species
         names, and the values are tuples specifying the three coverage
         parameters ``(a, m, E)`` which are the modifiers for the pre-exponential
-        factor [m, kmol, s units], the temperature exponent [nondimensional],
+        factor [m, kmol, s units], The temperature exponent [nondimensional],
```
speth: Why?

src/kinetics/BulkKinetics.cpp
```diff
@@ -120,6 +120,28 @@ bool BulkKinetics::addReaction(shared_ptr<Reaction> r)
     } else {
         m_irrev.push_back(nReactions()-1);
     }
+
+    if (r->newFramework()) {
+        shared_ptr<ReactionRateBase> rate;
+        rate = std::dynamic_pointer_cast<Reaction2>(r)->rate();
```
speth: The `newFramework` method isn't really necessary. You can just check the result of the `dynamic_pointer_cast` and if it's 0, then `r` isn't holding a `Reaction2`.

src/kinetics/ReactionFactory.cpp
```diff
@@ -0,0 +1,242 @@
...
+    reg_AnyMap("elementary-new",
+               [](Reaction* R, const AnyMap& node, const Kinetics& kin) {
+                   R->setup(node, kin);
+               });
+
+    // register interface reactions
+    reg("interface", []() { return new InterfaceReaction(); });
+    addAlias("interface", "surface");
+    addAlias("interface", "edge");
+    addAlias("interface", "global");
```
speth: I think `global` is another of those unimplemented vestigial rate types.

src/kinetics/ReactionRate.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+void CustomFunc1Rate::setRateFunction(shared_ptr<Func1> f) {
+    m_ratefunc = f;
+}
+
+double CustomFunc1Rate::eval(const CustomFunc1Data& shared_data) const {
+    if (m_ratefunc) {
+        return m_ratefunc->eval(shared_data.m_temperature);
+    }
+    return 0.;
```
speth: Should this throw if no rate function is set? Just returning 0 if you forget to set the rate function seems a little too error-prone.

speth (changes requested):
Thanks, @ischoegl, I think this is getting very close. Most of my comments below are fairly minor implementation details. I have no problem leaving the follow-on work you've mentioned to later PRs. The only thing I'm not looking forward to is fixing the merge conflicts this will introduce to #984 😬.

One thing I'm still a little iffy on is all the constructors that have to be implemented for each `DataType` class in order to support evaluation of rates for reactions not associated with any `Kinetics` object. While I recognize that T and P are sufficient for quite a few cases, what are you supposed to do for the cases where that doesn't work? Having to implement all of those constructors and throw an exception in them seems kind of tedious, and I don't think we should require users to do that. Would it make sense to have a common base class for the `DataType` classes, with `update_T` and `update_TP` methods that throw `NotImplementedError` by default?

interfaces/cython/cantera/reaction.pyx
ischoegl: :man_shrugging: - no idea myself. Not sure how this snuck in!

include/cantera/kinetics/ReactionRate.h
ischoegl: This method is actually not available to the base class (i.e. it's not an `override`), as `DataType` only becomes available in the template. I'll make sure that docstrings make this sufficiently clear. I did overlook one elsewhere though.

include/cantera/kinetics/MultiRate.h
ischoegl: :tada: indeed!

src/kinetics/BulkKinetics.cpp
ischoegl: :tada: nice!

include/cantera/kinetics/MultiRate.h
ischoegl: There indeed isn't much of a point in the shared pointer. Had to tweak that a little bit (`dynamic_cast` causes issues for `const`), and passing the actual object by reference works just fine.
```C++
    virtual void add(const size_t rxn_index,
                     ReactionRateBase& rate) override {
        m_indices[rxn_index] = m_rates.size();
        m_rates.push_back(dynamic_cast<RateType&>(rate));
        m_rxn.push_back(rxn_index);
    }
```

src/kinetics/ReactionRate.cpp
ischoegl: I chose that behavior to be consistent with other default constructor behavior (e.g. the original `Arrhenius` will create zero rates if no arguments are passed). I.e. there could be an argument made to raise an error in all cases.

src/kinetics/ReactionFactory.cpp
ischoegl: Since this is based on existing code, I am opting for:
```
addDeprecatedAlias("interface", "global");
```

ischoegl: Nevermind ... couldn't locate it. Deleting ...

interfaces/cython/cantera/reaction.pyx
ischoegl: simplified.

ischoegl:
@speth ... done. I actually looked at #984 and there are surprisingly few (and simple) merge conflicts. I ended up renaming `Reaction2::setup` to `Reaction2::setParameters` to mirror `getParameters` you are defining there (also moved it to `Reaction2` as I don't have plans to backport that functionality).

I also looked into your comment on `DataType`. I tried a base class, but couldn't really find much in terms of simplifications. I did, however, implement multiple `update` versions, which ended up making things somewhat simpler to follow. I also moved all of the data types to `ReactionData.h/.cpp` for clarity. Docstrings are updated as well - I'll add more for the follow-up PR's. There may be some additional minor items, but otherwise I believe it's ready to be merged.

Regarding next steps, I may wait for #984 to be merged, as there are some synergies. Once that is done, I am planning to create new classes mirroring the old ones (e.g. `ElementaryReaction2`, etc.) so things can be cleanly deprecated.

src/kinetics/ReactionRate.cpp
speth: Fair enough.

include/cantera/kinetics/ReactionRate.h
speth: Ah, sorry, I guess I commented on the wrong one. I can confirm the warnings are now gone, though.

speth (approved):
Thanks for the quick revisions. This all looks good to me. 

I think these changes to the `ReactionData` interface will work fine, at least for the time being. I imagine that as we exercise this capability on the existing rate classes, we'll be able to get a better feel for possible improvements / simplifications.

PR #984, opened 2021-02-25
Enable serialization of Cantera objects

speth:

- Adds `getParameters(AnyMap&)` methods to ThermoPhase, Kinetics, Reaction, etc. classes to collect data needed for serialization
- The data in these fields is populated based on the current state of the objects, so it can be used for objects created not only for YAML files, but also objects created programmatically from scratch or from the legacy CTI and XML input files.
- For objects that originate from YAML input files, "extra" YAML fields not used by Cantera are preserved and returned
- Add the `input_data` property to the corresponding Python objects, which returns this information converted to native Python types, e.g. `dict`s and `list`s
- Introduce the `YamlWriter` class for creating input files defining phases, species, and reactions. Generated output files can contain multiple phase definitions and reaction sections, and have use user-specified unit systems.

**Examples**

Python `input_data` property:
```python
>>> gas = ct.Solution('gri30.yaml')
>>> gas.input_data
{'elements': ['O', 'H', 'C', 'N', 'Ar'],
 'kinetics': 'gas',
 'name': 'gri30',
 'species': ['H2',
  'H',
  ...
  'CH2CHO',
  'CH3CHO'],
 'state': {'T': 300.0, 'Y': {'H2': 1.0}, 'density': 0.08189392763801234},
 'thermo': 'ideal-gas',
 'transport': 'mixture-averaged'}

>>> gas.species(1).input_data
{'composition': {'H': 1.0},
 'name': 'H',
 'thermo': {'data': [[2.5,
    ...
    -0.446682853],
   [2.50000001,
   ...
    -0.446682914]],
  'model': 'NASA7',
  'note': 'L7/88',
  'temperature-ranges': [200.0, 1000.0, 3500.0]},
 'transport': {'diameter': 2.05,
  'geometry': 'atom',
  'model': 'gas',
  'well-depth': 145.0}}

>>> gas.reaction(1).input_data
{'efficiencies': {'AR': 0.7,
  'C2H6': 3.0,
  'CH4': 2.0,
  'CO': 1.5,
  'CO2': 2.0,
  'H2': 2.0,
  'H2O': 6.0},
 'equation': 'H + O + M <=> OH + M',
 'rate-constant': {'A': 500000000000.0001, 'Ea': 0.0, 'b': -1.0},
 'type': 'three-body'}
```

`YamlWriter` class in C++:
```c++
auto original = newSolution("gri30.yaml");
// Add or modify species reactions at will
YamlWriter writer;
writer.addPhase(original);
writer.setUnits({{"activation-energy", "K"}});
writer.toYamlFile("modified-mechanism.yaml");
```

From a Python `Solution` or `Interface`:
```python
gas = ct.Solution('ptcombust.yaml', 'gas', transport_model='mixture-averaged')
surf = ct.Interface('ptcombust.yaml', 'Pt_surf', [gas])
surf.write_yaml('output.yaml', units={'length': 'cm', 'quantity': 'mol'})
```

**Remaining issues**
- I have not figured out how to get Sphinx to include documentation for the `_SolutionBase.input_data` property and `_SolutionBase.write_yaml` method with the `Solution` class. I tried adding a `:members: input_data, write_yaml` directive in `importing.rst`,  but it didn't seem to do anything.

**Potential future improvements**
- YamlWriter could use some options to specify additional metadata fields, perhaps something akin to the the `--extra` option of `ck2yaml`. This probably requires adding a converter for Python data structures to `AnyMap`.
- Units of "extra" fields not interpreted by Cantera are not subject to any unit conversions. For now, if you specify the same output units as the input file, it should be fine.
- There are some cases where floating point rounding issues result in kind of ugly values when round-tripping input data, e.g. reactions with pre-exponential factors of 7.000000000000001e+11. The best workaround I have for this is judicious use of the `precision` option.

Addresses Cantera/enhancements#11.

**Checklist**


ischoegl:
@speth ... this is great news! While I haven't looked at details (beyond `_cantera.pxd`), I had a quick question. The new method `anymapToPython` will translate `AnyMap`s to a dictionary. Would it be possible to add a corresponding `python_to_anymap` (using snake case instead of camel-case)? I've come across this in unrelated work (where the currently most efficient way is to write a YAML string from a dictionary, which is then loaded with existing methods).

include/cantera/kinetics/Falloff.h
```diff
@@ -88,6 +90,10 @@ class Falloff
     //! Get the values of the parameters for this object. *params* must be an
     //! array of at least nParameters() elements.
     virtual void getParameters(double* params) const {}
+
+    //! Store the falloff-related parameters needed to reconstruct an identical
+    //! Reaction using the newReaction(AnyMap&, Kinetics&) function.
+    virtual void getParameters(AnyMap& reactionNode) const {}
```
ischoegl, line 96: an alternative (here and elsewhere) would be
```C++
virtual unique_ptr<AnyMap> parameters() const {}
```
(or `shared_ptr`). I have another comment that complements this further below.

interfaces/cython/cantera/utils.pyx
```diff
@@ -72,3 +72,78 @@ class CanteraError(RuntimeError):
     pass
 
 cdef public PyObject* pyCanteraError = <PyObject*>CanteraError
+
+cdef anyvalueToPython(string name, CxxAnyValue& v):
```
ischoegl: names in `utils.pyx` are not very 'pythonic' (although it's not exposed to Python) - `anyvalue_to_python` would be (same for `anymap_to_dict` below - would suggest `dict` as this is what it returns)

interfaces/cython/cantera/utils.pyx
```diff
@@ -72,3 +72,78 @@ class CanteraError(RuntimeError):
...
+    m.applyUnits()
+    for item in m:
+        py_items.append((item.second.order(),
+                         pystr(item.first),
+                         anyvalueToPython(item.first, item.second)))
+    py_items.sort()
+    return {key: value for (_, key, value) in py_items}
+
+
+cdef mergeAnyMap(CxxAnyMap& primary, CxxAnyMap& extra):
```
ischoegl: Would it make sense to handle that within the C++ layer? Also, as a comment, the `anymapToPython` (or `anymap_to_dict`) applies units whereas `mergeAnyMap` does not.

~Regarding my other comment about `getParameters`, a C++ implementation of `merge` would go in hand with having `unique_ptr<AnyMap> parameters() const` returned throughout, which then would mean that you can `AnyMap::merge(const AnyMap& other)` (or similar).~

~I assume that there may be a small performance penalty for smart pointers (and potentially a larger one if merging requires copying), but it would be easier to track where objects are initially created (while this may be subjective, it is often difficult to track how things are assembled when the receiving function creates the object which is subsequently added to as in your `getParameters` approach).~

src/kinetics/Reaction.cpp
```diff
@@ -257,6 +348,40 @@ ChebyshevReaction::ChebyshevReaction(const Composition& reactants_,
 {
 }
 
+void ChebyshevReaction::getParameters(AnyMap& reactionNode) const
```
ischoegl, line 512: Would it make sense to pass the unit system as a parameter? Presumably the destination is known at the time when the process starts?

src/kinetics/Reaction.cpp
```diff
@@ -244,6 +321,20 @@ PlogReaction::PlogReaction(const Composition& reactants_,
...
 }
 
+void PlogReaction::getParameters(AnyMap& reactionNode) const
+{
+    Reaction::getParameters(reactionNode);
+    reactionNode["type"] = "pressure-dependent-Arrhenius";
+    std::vector<AnyMap> rateList;
+    for (const auto& r : rate.rates()) {
+        AnyMap rateNode;
+        rateNode["P"].setQuantity(r.first, "Pa");
```
ischoegl, line 438: Should this inherit from a unit system rather than defaulting to `"Pa"`?

ischoegl (commented):
@speth ... as indicated earlier, it is great that this is finally close to becoming reality (:tada:). I haven't looked at a very detailed level (I assume that the unit test suite is able to ensure that the emitted YAML code can be re-imported without issues), but do have some minor comments.

ischoegl:
@speth ... if you get a chance, could you rebase? After that, I'll go over the code in more detail so this can get merged.

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -422,3 +422,103 @@ def check(a, b):
...
+
+        with open('ptcombust-generated.yaml') as infile:
+            generated = yaml.safe_load(infile)
+        for key in ('phases', 'species', 'gas-reactions', 'Pt_surf-reactions'):
+            self.assertIn(key, generated)
+        self.assertEqual(len(generated['gas-reactions']), gas.n_reactions)
+        self.assertEqual(len(generated['Pt_surf-reactions']), surf.n_reactions)
+        self.assertEqual(len(generated['species']), surf.n_total_species)
+
+
```
ischoegl, line 547: It may make sense to expand tests some, e.g. generate yaml and benchmark test problems for both original and generated yaml input (along the lines of `test_convert` for CTI/XML->YAML). Including larger mechanisms (Reitz?) may be interesting also.

Individual objects are well covered by C++ tests, so this is mainly testing the `YamlWriter`'s front end, correct?

test/general/test_serialization.cpp
```diff
@@ -0,0 +1,349 @@
...
+    auto original = newSolution("h2o2.xml");
+    YamlWriter writer;
+    writer.addPhase(original);
+    writer.setPrecision(14);
+    writer.setUnits({
+        {"activation-energy", "K"},
+        {"quantity", "mol"},
+        {"length", "cm"}
+    });
+    // Should fail because pre-exponential factors from XML can't be converted
```
ischoegl: Not sure that I am following based on what this comment states - while XML may have limitations for *import*, shouldn't this become moot once data are read and available to the C++ objects? (at which point everything should be standardized, i.e. everything is known implicitly)

src/base/YamlWriter.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+            if (speciesDefIndex.count(name) == 0) {
+                speciesDefs.emplace_back(speciesDef);
+                speciesDefIndex[name] = speciesDefs.size() - 1;
+            } else if (speciesDefs[speciesDefIndex[name]] != speciesDef) {
+                throw CanteraError("YamlWriter::toYamlString",
+                    "Multiple species with different definitions are not "
+                    "supported:\n>>>>>>\n{}\n======\n{}\n<<<<<<\n",
+                    speciesDef.toYamlString(),
+                    speciesDefs[speciesDefIndex[name]].toYamlString());
```
ischoegl, line 80: As `YamlWriter` is central to serialization, it may make sense to add a unit test for special cases  (here and elsewhere as marked by CodeCov).

src/base/AnyMap.cpp
```diff
@@ -168,22 +235,241 @@ struct convert<Cantera::AnyMap> {
...
+            if (foundType) {
+                if (width + name.size() + valueStr.size() + 4 > 79) {
+                    out << YAML::Newline;
+                    width = 15;
+                }
+                out << name;
+                out << valueStr;
+                width += name.size() + valueStr.size() + 4;
+            } else {
+                // Put items of an unknown (compound) type on a line alone
```
ischoegl, line 284: As `AnyMap` is central to serialization, it may make sense to add a unit test for special cases (here and elsewhere as marked by CodeCov).

interfaces/cython/cantera/yamlwriter.pyx
```diff
@@ -0,0 +1,67 @@
...
+        """
+        def __set__(self, pybool skip):
+            self.writer.skipUserDefined(skip)
+
+    property output_units:
+        """
+        Set the units to be used in the output file. Dimensions not specified
+        will use Cantera's defaults.
+
+        :param units:
```
ischoegl, line 52: There appears to be a potential synergy with #991 ... would it make sense to allow for `pint` units?

interfaces/cython/cantera/_cantera.pxd
```diff
@@ -120,6 +136,8 @@ cdef extern from "cantera/thermo/Species.h" namespace "Cantera":
         Composition composition
         double charge
         double size
+        void getParameters(CxxAnyMap&) except +translate_exception
+        CxxAnyMap input
```
ischoegl: C++ signatures are not consistent: some expose member variables directly 
```C++
CxxAnyMap input
```
while others use a getter
```C++
CxxAnyMap& input()
```
I believe using the latter consistently is easier to maintain long-term. Performance should be the same as those functions will get inlined by the compiler?

interfaces/cython/cantera/reaction.pyx
```diff
@@ -302,6 +302,17 @@ cdef class Reaction:
...
 
+    property input_data:
+        """
+        Get input data for this reaction with its current parameter values,
+        along with any user-specified data provided with its input (YAML)
+        definition.
+        """
+        def __get__(self):
+            cdef CxxAnyMap params
+            self.reaction.getParameters(params)
```
ischoegl: Here and elsewhere: rather than calling into C++ twice (once for `getParameters`, once for `input`, followed by a merge), wouldn't it make sense to add an optional argument, i.e.
```C++
getParameters(AnyMap&  params, bool merge_input=true)
```
Which (presumably) eliminates the need for `mergeAnyMap`?

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1709,6 +1709,20 @@ class ThermoPhase : public Phase
...
+    //! does not include user-defined fields available in input().
+    virtual void getParameters(AnyMap& phaseNode) const;
+
+    //! Get phase-specific parameters of a Species object such that an
+    //! identical one could be reconstructed and added to this phase.
+    /*!
+     * @param name     Name of the species
+     * @param eosNode  Mapping to be populated with parameters
+     */
+    virtual void getSpeciesParameters(const std::string& name,
```
ischoegl, line 1725: Is a dedicated `getSpeciesParameters` necessary? Presumably parameters can be accessed via `species[name].getParameters(eosNode)` or similar. The implementation seems like an outlier, as there is no corresponding `getReactionParameters` in `Kinetics.h`.

include/cantera/kinetics/Falloff.h
ischoegl: I believe the current implementation is both simpler and faster.

src/kinetics/Reaction.cpp
ischoegl: mostly moot - presumably this is updated by `YamlWriter` ...

src/kinetics/Reaction.cpp
ischoegl: mostly moot as I believe YamlWriter converts later.

interfaces/cython/cantera/utils.pyx
ischoegl: while my initial comment is mostly moot, I believe this function is mainly used to merge output from `getParameters` and `input`? I believe that it would be nice to merge this by passing an optional flag to `getParameters` where it applies (and handle the merge in C++). This may eliminate the need for this specific Cython function ...

ischoegl (changes requested):
@speth ... I rescinded some of my initial comments as I believe your implementation is more efficient (despite being harder to track). 

Beyond, I think adding an optional flag to `getParameters` to add `input` where it applies (and handling a merge within C++) may be more consistent and easier long-term. I'm also not sure about `getSpeciesParameters`, as it seems somewhat superfluous. Spot-wise benchmark tests that illustrate that a re-imported solution preserves results may add another level of security.

**PS:** I believe it would make sense to finalize this before #995 (and rebase the latter, as this would allow me to port `getParameters` immediately, rather than requiring another go-around).

**PPS:** Adding this capability to the [extract_submechanism.py](https://github.com/Cantera/cantera/blob/main/interfaces/cython/cantera/examples/kinetics/extract_submechanism.py) example would be neat.

test/general/test_serialization.cpp
speth: This is definitely a thorny issue. Because reactions are instantiated from XML without having a `Kinetics` object available, there is no way to determine the dimensions of the rate constant. All you know is that the value itself has already been converted to Cantera's mks+kmol system. I guess you _could_ set the value of `Reaction.rate_units` when calling `Kinetics.addReaction`, which would resolve this for what will presumably be the most common use case, serializing a complete phase definition. But a bare Reaction object created from XML does not contain the information needed to serialize it to a non-default unit system.

ischoegl: From what I understand, the reaction order provides most of the missing information, so the main issues are non-standard units as well as volumetric vs surface geometry. Am I seeing this correctly? Regardless, I believe export should work *if* associated kinetics objects are defined (as you mention, this can be done implicitly and is likely the most common use case); this would leave an exception if the reaction isn’t fully set up.

include/cantera/thermo/ThermoPhase.h
speth: The reason for this method is that the behavior of `getSpeciesParameters` is specialized for different `ThermoPhase` models, and it is collecting data stored in the `ThermoPhase` object that applies to the specified species. Most of the time, this is really just populating the `equation-of-state` child node, but I think there were a couple of cases where there were fields outside that that needed to be set as well.

This is somewhat different from the case of reactions, where much of the specialized behavior is handled by the different `Reaction` classes, while the `GasKinetics` class is mostly generic (and getting more generic thanks to your work on #995).

ischoegl: Makes sense.

test/general/test_serialization.cpp
speth: Correct, the units are determined by the dimensionality of the various phases, plus the units of the standard concentration for each phase. And while kmol/m^3 or kmol/m^2 are probably the most common, there are several phases where the standard concentration is dimensionless (see the various implementations of `standardConcentrationUnits`).

interfaces/cython/cantera/yamlwriter.pyx
speth: Perhaps, but I'd like to suggest deferring such an enhancement until both of these PRs are merged.

interfaces/cython/cantera/_cantera.pxd
speth: Please correct me if I'm getting this wrong, but I think the variable is only public in cases like the `Species` class that are mostly just structs where all of the member data is already `public`. So the consistency is with the parent class, not across implementations of the `input` member variable / property.

interfaces/cython/cantera/yamlwriter.pyx
ischoegl: Absolutely - wasn’t necessarily advocating for adopting things now, so this was mostly a note.

interfaces/cython/cantera/_cantera.pxd
ischoegl: This is mostly a stylistic matter ... per code styling guide, *For most classes, member variables should not be public.*. As `input` is newly introduced, it may make sense to just use the setter methods and default to having `m_input` protected. That way, it's always accessed using a getter method. There shouldn't be a difference in performance ...

**PS:** hadn’t picked up on the struct vs class  portion, but the former can have protected members, which would create a uniform interface.

interfaces/cython/cantera/yamlwriter.pyx
ischoegl: I left a comment in #991, which resolves this conversation.

test/general/test_serialization.cpp
ischoegl: Thanks for the clarification; I believe I’m on the same page now - units do pop up in #995.

Regarding the XML conversion, I still believe that implicitly setting units would make sense.

interfaces/cython/cantera/_cantera.pxd
speth: I take your point, but while looking through this PR for cases to fix, I realized that the only new instance of this is for the `SpeciesThermoInterpType` class, where it is actually implemented as getter (with the underlying member variable `m_input`). The other cases that you see here are just making existing member variables accessible within Cython.

ischoegl: It’s admittedly a pretty minor point, so it’s ok to leave as is ... depending on how the other comment is addressed (`getParameters`), it may not be used by itself very often. 

**PS:** from my experiences thus far, getters are much easier to work with whenever things are changed compared to member variables, so overall I think the preference for getters in new code makes sense. For some reason I had assumed that `input` was new, while it actually is already introduced in 2.5.1 ... changing this now is a major headache of course, and it’s certainly not worth the trouble.

ischoegl: Marking this as resolved as this was pre-existing.

interfaces/cython/cantera/utils.pyx
speth: The initial reason for this came from the desire to control the order of the output keys. I want the user-defined keys in `input` to generally come last, and the ordering mechanism mostly relies on preserving insertion order. If the base `Reaction` class were responsible for adding in the user-defined keys, then they would come before any keys added by child classes.

I actually ended up finding a hybrid solution that both gets rid of the need to merge the fields defining object with the extra input data at the point of use, and avoids the requirement for the user-accessible function to take an `AnyMap` as it's input, by making the virtual `getParameters` method `protected`, and introducing a non-virtual method with the signature `AnyMap parameters(bool withInput=true)` on the base classes only. This way, the `parameters` method can initialize the `AnyMap`, use the virtual `getParameters` method to populate all of the relevant fields, and then add the extra keys last before returning it.

interfaces/cython/cantera/test/test_composite.py
speth: Yes, these tests were mainly just focused on the `YamlWriter` Python interface, but I did add a couple of additional checks here to show that whole `Solution` is being restored in a consistent state. I think the consistency of all the different phase and reaction types is reasonably-well covered by the `ThermoYamlRoundTrip` test suite (in `test/thermo/thermoToYaml.cpp`), and the checks done in the `ReactionToYaml` test suite (in `test/kinetics/kineticsFromYaml.cpp`) by the `compareReactions` function.

test/general/test_serialization.cpp
speth: I realized while thinking about this that CTI/XML isn't the only case where you can manage to create a reaction without needing to set it's rate units -- this is also true if you just create a `Reaction` object from scratch. Setting `Reaction.rate_units` now occurs as part of `Kinetics.addReaction`, if it doesn't get set before then. I also ended up refactoring up the `rateCoeffUnits` function and making it a member function of the `Reaction` class (as `calculateRateCoeffUnits`).

interfaces/cython/cantera/reaction.pyx
speth: The new `Reaction.parameters(bool withInput)` method has indeed made `mergeAnyMap` unnecessary.

speth:
@ischoegl, thank you for the good suggestions. I think I've implemented most of them, with the exception of some improvements to the test coverage, which I'll work on after the CI runs.

test/general/test_serialization.cpp
ischoegl: That’s what had been my suspicion all along. If you look at #995, I have new constructors for `ReactionRate` from `AnyMap` where this will become relevant. Obviously things need to be consolidated as I didn’t have `rate_units` at my disposal before.

ischoegl:
:+1: ... looks really good! Pretty happy about `reaction::calculateRateCoeffUnits` ... this will help a bit going forward.

ischoegl (dismissed):

speth:
I think I'm happy with the code coverage provided by the tests at this point.

ischoegl (approved):

ischoegl:
@speth ... thanks for going into more detail with test coverage. As mentioned above, this looks good to go.

bryanwweber:
@ischoegl Feel free to hit the merge button 😄

speth:
🎉

PR #988, opened 2021-03-06
Add thermal conductivity to DustyGas in python.

chinahg:

- Thermal conductivity can now be calculated using the DustyGas class in python
-
-

Fixes #

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/transport.pyx
+++ b/interfaces/cython/cantera/transport.pyx
@@ -182,7 +182,11 @@ cdef class Transport(_SolutionBase):
             return self.transport.electricalConductivity()
 
     property thermal_conductivity:
-        """Thermal conductivity. [W/m/K]."""
+        """
+        Thermal conductivity. [W/m/K]
+        Returns thermal conductivity of the ideal gas object using the multicomponent
+        model. The value is not specific to the dusty gas model.
+        """
         def __get__(self):
             return self.transport.thermalConductivity()
 
--- a/interfaces/cython/cantera/transport.pyx
+++ b/interfaces/cython/cantera/transport.pyx
@@ -276,6 +276,11 @@ cdef class DustyGasTransport(Transport):
         def __set__(self, value):
             (<CxxDustyGasTransport*>self.transport).setPermeability(value)
 
+    property thermal_conductivity:
+        def __get__(self):
+            return (<CxxDustyGasTransport*>self.transport).gasTransport().thermalConductivity()
+
+
     def molar_fluxes(self, T1, T2, rho1, rho2, Y1, Y2, delta):
         """
         Get the molar fluxes [kmol/m^2/s], given the thermodynamic state at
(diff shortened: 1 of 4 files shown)


decaluwe (changes requested):
Thanks, @chinahg. This looks great.  Coding looks straightforward, and I pulled the code to my machine and verified that it works in Python.

Two questions, both regarding demonstration/testing:
1. Can we add some code demonstrating that this is functional in the [dusty_gas.py example](https://github.com/Cantera/cantera/blob/main/interfaces/cython/cantera/examples/transport/dusty_gas.py)?
2. It looks like test coverage increases as a result of this change, but unless I'm mistaken, it doesn't look like this specific capability is tested.  Can you add something to the [dustyGasTransport test problem](https://github.com/Cantera/cantera/tree/main/test_problems/dustyGasTransport) to test that this function works? 

Great job!

decaluwe:
Thanks, @chinahg -- change to the example file looks simple but effective (both of those are good things, for the record 😉).

Now if we can get something testing this in the associated test problem, this will be all ready to merge. 🎉

include/cantera/transport/DustyGasTransport.h
```diff
@@ -19,8 +19,9 @@ namespace Cantera
 {
 //! Class DustyGasTransport implements the Dusty Gas model for transport in porous media.
 /*!
- * As implemented here, only species transport is handled. The viscosity,
- * thermal conductivity, and thermal diffusion coefficients are not implemented.
+ * As implemented here, only species transport is handled. The viscosity
+ * and thermal diffusion coefficients are not implemented. Thermal conductivity
+ * is calculated as the conductivity of the gas phase.
```
speth: There were no changes here to the `thermalConductivity` method of the C++ `DustyGasTransport` class. That still raises a `NotImplementedError`, which I think is correct In that case, this comment was more correct as it originally was.

interfaces/cython/cantera/transport.pyx
```diff
@@ -272,6 +272,11 @@ cdef class DustyGasTransport(Transport):
         def __set__(self, value):
             (<CxxDustyGasTransport*>self.transport).setPermeability(value)
 
+    property thermal_conductivity:
+        def __get__(self):
+            return (<CxxDustyGasTransport*>self.transport).gasTransport().thermalConductivity()
```
speth, line 285: Here I think we need to document that this is returning the thermal conductivity of the embedded gas object, as opposed to a value specific to the dusty gas model.

speth (changes requested):
@decaluwe, there are no code changes to the C++ implementation here, so I don't think there's anything to test in the C++ test. I think the coverage changes are all spurious, and just an indication that this should be rebased onto the current `main` branch.

@chinahg, can you add a test to the Python test suite? You can add it to the `TestDustyGas` class in `interfaces/cython/cantera/test/test_transport.py`.

decaluwe:
Good point, @speth -- thanks.

speth (approved):

decaluwe (approved):

decaluwe:
Many thanks and excellent work, @chinahg!

PR #1002, opened 2021-03-26
Fix documentation bug for YAML format

### LINKED ISSUE #1001: Documentation bug for YAML format
One of the documentation examples in Solution ([link to code](https://github.com/Cantera/cantera/blob/main/interfaces/cython/cantera/composite.py#L66) | [link to site](https://cantera.org/documentation/docs-2.5/sphinx/html/cython/importing.html?highlight=solution#cantera.Solution)) demonstrating that Species and Reactions can be imported from input files to construct Solutions appears to be invalid in 2.5. When using yaml inputs, `ct.Reaction.listFromFile` [needs a Kinetics object](https://github.com/Cantera/cantera/blob/main/interfaces/cython/cantera/reaction.pyx#L143-L144) as the second argument.

> Solution objects can also be constructed using Species and Reaction objects which can themselves either be imported from input files or defined directly in Python:
>
> ```python
> spec = ct.Species.listFromFile('gri30.yaml')
> rxns = ct.Reaction.listFromFile('gri30.yaml')  ## broken ##
> gas = ct.Solution(thermo='IdealGas', kinetics='GasKinetics',
>                   species=spec, reactions=rxns, name='my_custom_name')
> ```

The example should probably be corrected (either adding the second argument or switching back to cti/xml) or removed. The issue was introduced in #696, which updated a bunch of `*.cti` to `*.yaml` in the docstrings.

leesharma:

- Fixes the `Reaction.listFromFile` example syntax in the `Solution` docs by adding the required second argument (using the `gas` object created in a previous code example.)

  I'm not sure if this is the best solution: it has the downside of requiring `gas` to already be initialized to a `Kinetics` object with the appropriate species, which doesn't seem super useful. I'm new to Cantera, but I couldn't find an easy way to make a "dummy" Kinetics object. Is there a better way to do this?

  Maybe the example should just be removed? 

Fixes #1001

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -63,7 +63,7 @@ class Solution(ThermoPhase, Kinetics, Transport):
     directly in Python::
 
         spec = ct.Species.listFromFile('gri30.yaml')
-        rxns = ct.Reaction.listFromFile('gri30.yaml')
+        rxns = ct.Reaction.listFromFile('gri30.yaml', gas)
         gas = ct.Solution(thermo='IdealGas', kinetics='GasKinetics',
                           species=spec, reactions=rxns, name='my_custom_name')
 
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -63,14 +63,17 @@ class Solution(ThermoPhase, Kinetics, Transport):
     directly in Python::
 
         spec = ct.Species.listFromFile('gri30.yaml')
-        rxns = ct.Reaction.listFromFile('gri30.yaml', gas)
+        spec_gas = ct.Solution(thermo='IdealGas', species=spec)
+        rxns = ct.Reaction.listFromFile('gri30.yaml', spec_gas)
         gas = ct.Solution(thermo='IdealGas', kinetics='GasKinetics',
                           species=spec, reactions=rxns, name='my_custom_name')
 
     where the ``thermo`` and ``kinetics`` keyword arguments are strings
     specifying the thermodynamic and kinetics model, respectively, and
     ``species`` and ``reactions`` keyword arguments are lists of `Species` and
-    `Reaction` objects, respectively.
+    `Reaction` objects, respectively. Note that importing the reactions from a
+    YAML input file requires a `Kinetics` object containing the species, as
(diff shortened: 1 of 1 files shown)


interfaces/cython/cantera/composite.py
```diff
@@ -63,7 +63,7 @@ class Solution(ThermoPhase, Kinetics, Transport):
     directly in Python::
 
         spec = ct.Species.listFromFile('gri30.yaml')
-        rxns = ct.Reaction.listFromFile('gri30.yaml')
+        rxns = ct.Reaction.listFromFile('gri30.yaml', gas)
```
speth: Can you update this to give a complete example, where you first initialize a gas with the species but no reactions, then create the list of reactions? Right now, `gas` is used before it is defined.

speth (changes requested):
Thanks for pointing this out. While this requirement is a little confusing, you need to know the thermo model of each reactant species in a reaction in order to convert the rate constant to Cantera's internal unit system of mks+kmol. This wasn't necessary before because the XML format already had the rate constants converted to those units, and the CTI sort of cheated and already knew the right units to use for the handful of thermo models that it supported.

ischoegl:
One alternative would be to provide a base unit system as a second argument, which however would require the C++ unit system being exposed to Python. Another issue are electrochemical reactions, where the current YAML import requires associated thermo models.

interfaces/cython/cantera/composite.py
leesharma: I pushed an update defining `spec_gas` explicitly from the species (plus a sentence stating the requirement). It looks a bit contrived, but the example runs without error. What do you think?

speth:
@ischoegl, the unit system isn't really the issue. What is needed to convert the coefficients of the rate constant are the dimensions of the standard concentration for each reactant, along with the dimensions of the phase where the reaction occurs.

speth (approved):
Looks good to me. Maybe an option for making this less awkward would be to get rid of the `Reaction.listFromFile` constructor and add methods to the `Kinetics` class for creating a list of `Reaction`s objects from a file, and a method for adding a list of `Reaction` objects to that `Kinetics` object, rather than specifying the reactions in the constructor.

PR #1004, opened 2021-03-31
Add support for SUNDIALS 5.5-5.7 compiled with LAPACK

bryanwweber:

- Support for system SUNDIALS 5.5-5.7 with LAPACK. The preprocessor define indicating LAPACK support in SUNDIALS was changed with version 5.5.
- Fix that the Boost version message is printed even if Boost isn't found
- Simplify preprocessor logic in `BandMatrix.cpp`
- Fix deprecation warning about SunLinSol_SPGMR
- Fix error when `omp.h` is installed, but the library cannot be found by the linker (for example, omp installed by Homebrew, the Homebrew path on the search path, but using Apple's compilers).

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1034,10 +1034,11 @@ env['HAS_OPENMP'] = conf.CheckCXXHeader('omp.h', '""')
 boost_version_source = get_expression_value(['<boost/version.hpp>'], 'BOOST_LIB_VERSION')
 retcode, boost_lib_version = conf.TryRun(boost_version_source, '.cpp')
 env['BOOST_LIB_VERSION'] = '.'.join(boost_lib_version.strip().split('_'))
-print('INFO: Found Boost version {0}'.format(env['BOOST_LIB_VERSION']))
 if not env['BOOST_LIB_VERSION']:
     config_error("Boost could not be found. Install Boost headers or set"
                  " 'boost_inc_dir' to point to the boost headers.")
+else:
+    print('INFO: Found Boost version {0}'.format(env['BOOST_LIB_VERSION']))
 # demangle is availble in Boost 1.55 or newer
 env['has_demangle'] = conf.CheckDeclaration("boost::core::demangle",
                                 '#include <boost/core/demangle.hpp>', 'C++')
--- a/SConstruct
+++ b/SConstruct
@@ -1112,7 +1112,7 @@ if env['system_sundials'] == 'y':
 
     print("""INFO: Using system installation of Sundials version %s.""" % sundials_version)
 
-    #Determine whether or not Sundials was built with BLAS/LAPACK
+    # Determine whether or not Sundials was built with BLAS/LAPACK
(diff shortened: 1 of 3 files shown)


bryanwweber:
By the way, I think it'd be good to backport at least ~64c0736~ fc93ece to the 2.5 branch

src/numerics/BandMatrix.cpp
```diff
@@ -10,19 +10,11 @@
 #if CT_USE_LAPACK
     #include "cantera/numerics/ctlapack.h"
 #else
-    #if CT_SUNDIALS_USE_LAPACK
-        #if CT_SUNDIALS_VERSION >= 30
-            #include "sunlinsol/sunlinsol_lapackband.h"
-        #else
-            #include "cvodes/cvodes_lapack.h"
-        #endif
+    #if CT_SUNDIALS_VERSION >= 30
```
ischoegl, line 13: the logic appears to be different here. The PR comment states that it was simplified ...

ischoegl (dismissed):
Per my comments elsewhere, this isn't necessarily my strong suit (and I don't have the toolchains to verify this installed). On the other hand, both PR rationale and code comments are clear and the changes appear to be straightforward. Testing every combination in GH Actions may be overkill, but if there are candidates of particular concern this may be warranted. Overall, no concerns from my side ...

src/numerics/BandMatrix.cpp
bryanwweber: Hmmm good point. It's simplified in the sense that if `CT_USE_LAPACK` is false, then `CT_SUNDIALS_USE_LAPACK` is also always false, which means the first branch in the `# else` clause could never be followed. I'll clarify the commit message.

ischoegl: Beyond, please feel free to ignore my 'Accept' - per discussion elsewhere, this is probably more a comment. *Edit: the commit message is likely adequate ...*

SConstruct
```diff
@@ -1121,10 +1122,22 @@ if env['system_sundials'] == 'y':
...
+        # SUNLINSOL packages indicating whether SUNDIALS has been built with LAPACK
+        env["has_sundials_lapack"] = conf.CheckDeclaration(
+            "SUNDIALS_SUNLINSOL_LAPACKBAND",
+            '# include "sundials/sundials_config.h"',
+            "C++",
+        ) and conf.CheckDeclaration(
+            "SUNDIALS_SUNLINSOL_LAPACKDENSE",
+            '# include "sundials/sundials_config.h"',
+            "C++",
+        )
```
speth: Two formatting quibbles here:
* extraneous space between the `#` and `include` in each check
* I find this long `z = x and y` statement pretty hard to parse. I think it would be better to just break it up.

speth (commented):
This looks good from a functional standpoint. I just have a couple of requests on the formatting.

speth (approved):

PR #1007, opened 2021-04-07
Update to 88 character maximum per line

ischoegl:

- Update coding style guidelines to 88 characters per line limit (up from 80)

N/A

I realize that this is a departure from long-standing practice (and am fine with immediately closing the PR). However, I am still suggesting this as it may prevent some line breaks that I believe make code less readable. Especially with vectors/maps of templated classes in C++ forced line breaks tend to get unwieldy; also, doxygen's inline comments (`//!<`) become more feasible.

The number of characters is arbitrarily inspired by 'black', which I am otherwise not advocating for.

**PS:** I was mulling over whether to open this as an issue (which it really isn't) or a discussion (which appears somewhat off-topic). I ended up with a PR, as this may be the most efficient way to either adopt, reject or tweak the suggestion.

**Checklist**

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,8 +33,8 @@
 * Push the changes on your new feature branch to your forked copy of the
   `Cantera/cantera` repository on GitHub.
 
-* Submit a Pull Request on Github, from your forked copy. Check the results 
-  of the continuous-integration tests run using Travis and AppVeyor and resolve 
+* Submit a Pull Request on Github, from your forked copy. Check the results
+  of the continuous-integration tests run using Travis and AppVeyor and resolve
   any issues that arise.
 * Additional discussion of good Git & Github workflow is provided at
   http://matplotlib.org/devel/gitwash/development_workflow.html and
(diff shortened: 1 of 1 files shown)


bryanwweber (approved):
I'm :+1: on this change. It'd be really nice to have an automated linter that could check these things for us, but I haven't found one that we can sufficiently configure and which will ignore lines that aren't changed in a particular PR. Oh well.

ischoegl:
> It'd be really nice to have an automated linter that could check these things for us, but I haven't found one that we can sufficiently configure and which will ignore lines that aren't changed in a particular PR. 

@bryanwweber ... interesting thought. There may be a way to use some bash magic in a GH Actions. As a starting point, this works
```
git diff HEAD..main | awk '{ print length }'
```
but there's obviously a little more to that. I'm not that great on bash, so this would certainly be a different PR ...

speth (approved):
It's taken me a while to come around to the idea of longer line lengths, but I think this is a reasonable value that preserves most of the benefits of keeping line lengths relatively short.

I don't think I'd want to rigidly enforce this particular style guideline with an automated linter, for what that's worth.

PR #1012, opened 2021-04-12
Simplify/Refactor buildutils.py

bryanwweber:

- Simplify/Refactor the `site_scons/buildutils.py` file. As mentioned in the [discussion post](https://github.com/Cantera/enhancements/discussions/93), this is going to be a long-running project. I'm planning to do the scripts/files in batches, so `buildutils.py` was the first.

**Checklist**


ischoegl:
:tada: ... as I mentioned elsewhere, this is not necessarily my strength. From what I can tell, this is largely to improve formatting, with a couple of more substantial changes. Big :+1: for `logger`, but `os.path` is very much alive. Snake case, quotes, etc. are as you mentioned elsewhere. Would you mind commenting on other parts you tried to tackle in this PR? Besides `build_utils.py` of course ...

bryanwweber:
Thanks @ischoegl!

* My goal in general is to simplify the code, in part by using Python 3.6 and up features (f-strings and pathlib are the two that come immediately to mind). I'm also trying to update the code style to reflect modern Python practices and PEP-8. I guess there wasn't a whole lot of simplification that was able to be done to `buildutils.py`, but I think the changes here will make it easier to simplify other things later, or possibly add features. This is really the tip of the iceberg, so to speak.
* The changes in the `SCons*` files are almost entirely due to changing/removing functions in `buildutils.py`. There are a few places where there are unrelated changes to the `SCons*` files, but I think they're pretty small changes for now that I made incidentally while making other changes.
* I'm a big fan of `pathlib` and a big anti-fan of `os.path`. Operating purely on strings when it comes to paths is (can be) really tricky, especially when working between platforms. `pathlib` automatically handles things like different slash conventions, trailing slashes, etc. (Edit to add: the intention with this PR was not to replace `os.path` everywhere, yet. I'm going to go through the files in batches to simplify PR review).

site_scons/buildutils.py
```diff
@@ -1,78 +1,138 @@
...
+
+from pathlib import Path
+import logging
+
+from typing import Union
+
+
+__all__ = ("logger", "remove_directory", "remove_file", "test_results",
+           "add_RegressionTest", "get_command_output", "listify", "which",
+           "ConfigBuilder", "mglob", "get_spawn")
```
bryanwweber: I need to add `"help"` here.

ischoegl (commented):
No objections from my side - the CI appears to be happy, so I am happy (I don't have enough insights to go beyond).

**PS:** Regarding my comment about `os.path` - I noticed a couple of new imports for the 'un'preferred library, hence my comment. ... overall I second your preference for `pathlib`, but it appears the transition will come later, correct? From personal experience, replacing paths is laborious, and the diff is already quite busy because of the formatting changes. So leaving it for later is likely preferable.

bryanwweber:
> the CI appears to be happy, so I am happy (I don't have enough insights to go beyond).

Thanks @ischoegl If there are any places where you see possible improvements in the Python, I'd be happy to know them!

> I noticed a couple of new imports for the 'un'preferred library, hence my comment. ... So leaving it for later is likely preferable.

Yes, that's my plan 😄 I started here by limiting the imports that `from buildutils import *` takes, which meant that all the stdlib imports in `buildutils` that were carried into other files were removed. And then did the minimum necessary in each of the `SCons*` files to get them working again.

I should also say that I'm trying to keep exact feature parity as I make these changes, including exactly the same output as before. So far, I've been successful 😃

site_scons/buildutils.py
```diff
@@ -1,78 +1,138 @@
...
+
+
+class OutputFormatter(logging.Formatter):
```
ischoegl, line 83: docstring - I know it's inheriting from logging, but what is the objective?

site_scons/buildutils.py
```diff
@@ -1,78 +1,138 @@
...
+        record.message = record.getMessage()
+        if getattr(record, "print_level", True):
+            s = level_style.format(**record.__dict__)
+        else:
+            s = no_level_style.format(**record.__dict__)
+        return s
+
+
+# Modified from https://stackoverflow.com/a/36338212
+class LevelFilter(logging.Filter):
```
ischoegl, line 114: docstring - I know it's inheriting from logging, but what is the objective?

site_scons/buildutils.py
```diff
@@ -308,89 +405,85 @@ def compareProfiles(env, ref_file, sample_file):
...
-    # Patterns to consider:
-    # 123
-    # 123.45
-    # 123.45e6
-    # 123e4
-    if 'e' in x:
-        x, exponent = x.split('e')
-        exponent = int(exponent)
+        header = ["Failed series comparisons:"]
+        header.append("coordinate  comp.  reference val    test value    "
```
ischoegl: here I'd almost recommend an exception to the line length, or to put things on their own line with `"""` (i.e. the old formatting was clearer)

site_scons/buildutils.py
```diff
@@ -407,61 +500,68 @@ def compareCsvFiles(env, file1, file2):
     files are dissimilar. Lines containing non-numeric data are
     automatically ignored.
     """
-    try:
-        import numpy as np
-    except ImportError:
-        print('WARNING: skipping .csv diff because numpy is not installed')
-        return 0
+    if not np:
+        logger.warning(
```
ischoegl: does this fit on one line with 88 characters? This is one of the `black` formatting choices I don't understand.

ischoegl (commented):
Alright :wink: ... I looked over `buildutils.py` and have a couple of minor points. Can give it a closer look later on.

Understood about `from xyz import *` ... linters really don't like it.

site_scons/buildutils.py
bryanwweber: I reformatted this, hopefully it's clearer now? The advantage is that the field width is a little more obvious now that I'm using `str.format`, the disadvantage is, why are you using `str.format` with string literal arguments? Ah well.

site_scons/buildutils.py
bryanwweber: Yes, this was this way from a previous point in the history when it was called `build_logger` rather than `logger`, and the `build_` tipped it over 88 characters. I'm not actually using Black to format the file 😄

site_scons/buildutils.py
```diff
@@ -1,78 +1,160 @@
...
+
+
+if sys.version_info[:2] < (3, 6):
+    logger.error(
+            "Cantera must be built using Python 3.6 or higher. If you are using "
+            "a system install of SCons, you can invoke SCons by writing\n"
+            "    python3 `which scons`\n"
+            "followed by any desired options."
+        )
+    sys.exit(1)
```
ischoegl: You could invoke `sys.exit(TestState.FAIL)` here, which relates to the class below.

site_scons/buildutils.py
```diff
@@ -1,78 +1,160 @@
...
     """
     Used along with DefineDict to generate a customized config.h file
     from a config.h.in file using the variables given in 'defines'.
     """
-    def __init__(self, defines):
+    def __init__(self, defines: dict) -> None:
         self.defines = DefineDict(defines)
 
-    def __call__(self, source, target, env):
+    def __call__(self, target, source, env):
```
ischoegl: You start with PEP 484 above, but didn't continue ...

site_scons/buildutils.py
```diff
@@ -617,92 +704,95 @@ def listify(value):
...
+    from IPython.core import getipython
 
-    ip = ipapi.get()
+    ip = getipython.get_ipython()
     def_colors = ip.colors
     Pdb(def_colors).set_trace(sys._getframe().f_back)
 
 
-def getSpawn(env):
+def get_spawn(env):
```
ischoegl: PEP 484 ... I believe `env` is specific to `scons`, but there are some ways to define custom types.

site_scons/buildutils.py
```diff
@@ -472,137 +591,105 @@ def add_RegressionTest(env):
...
+            exe_file = Path(path).joinpath(fpath)
+            if exe_file.exists() and os.access(exe_file, os.X_OK):
+                return True
+    return False
 
-optionWrapper = textwrap.TextWrapper(initial_indent='    ',
-                                   subsequent_indent='    ',
-                                   width=72)
 
+def help(env, options):
```
ischoegl: PEP 484. Docstring is not clear: what options are there?

site_scons/buildutils.py
```diff
@@ -472,137 +591,105 @@ def add_RegressionTest(env):
         action=env.Action(regression_test, regression_test_message))
 
 
-def quoted(s):
-    """ Returns the given string wrapped in double quotes."""
-    return '"%s"' % s
-
-
 def mglob(env, subdir, *args):
```
ischoegl: `mglob` could use a better name; I recall that at some point it wasn't obvious that this was a cantera-specific command. And PEP 484, of course ...

site_scons/buildutils.py
```diff
@@ -248,48 +333,82 @@ def compareTextFiles(env, file1, file2):
...
+    if "." in number:
+        digits = -len(number.split(".")[1])
+    else:
+        digits = 0
+
+    return exponent + digits
 
 
-def compareProfiles(env, ref_file, sample_file):
+def compare_profiles(env, ref_file, sample_file):
```
ischoegl: Same, PEP 484 is used intermittently, but not here.

site_scons/buildutils.py
```diff
@@ -617,92 +704,95 @@ def listify(value):
...
-    try:
-        useShowWindow = subprocess.STARTF_USESHOWWINDOW
-    except AttributeError:
-        useShowWindow = subprocess._subprocess.STARTF_USESHOWWINDOW
+    if "cmd.exe" not in env["SHELL"] or env.subst("$CXX") == "cl":
+        return env["SPAWN"]
 
-    def ourSpawn(sh, escape, cmd, args, environ):
-        newargs = ' '.join(args[1:])
+    def our_spawn(sh, escape, cmd, args, environ):
```
ischoegl: while this is internal, it is complex enough to use some comments

ischoegl (commented):
While I believe the Python code overall looks good, some of the docstrings could use improvements. Also, you started applying PEP 484, but didn't follow through with all signatures. While I know that some don't like the appearance, I find those annotations tremendously helpful even when looking at code that I wrote myself (especially once short-term memory no longer kicks in).

site_scons/buildutils.py
ischoegl: Somewhat yes, although the original formatting wasn’t too bad either (only issue was the indent with the triple quotes). Your call ...

site_scons/buildutils.py
bryanwweber: Good point... I'm interested in your opinion about this. The reason I left them out of the functions that I did here is that they seemed more like noise and not so helpful for future use. It would require importing several libraries under the type hinting check, and I didn't think it would add enough value to include the types in the signature. That may change as I go through the other files, though. What do you think?

ischoegl: I *think* you can keep this relatively simple by using type aliases, i.e. `TypeVar` (see [here](https://www.python.org/dev/peps/pep-0484/#id17))
```
T = TypeVar('T', bound=...)
```
I.e. you just locally define an alias that documents what this is? I had to dig this up from something one of my students found at some point.

bryanwweber: I don't think it's necessary to use `TypeVar` here, a simple alias would be enough. I'll add types in a single commit, and let's see if it doesn't make the code illegible. If it looks terrible, I can just drop that commit.

site_scons/buildutils.py
bryanwweber: The intention of the `TestState` enum is to explicitly report the status of tests, not general exit status. I added a docstring to clarify.

site_scons/buildutils.py
bryanwweber: The options are defined in `SConstruct`. They are the build options, not options for this function.

site_scons/buildutils.py
bryanwweber: What comments do you think would help improve, beyond the top-level docstring?

site_scons/buildutils.py
bryanwweber: Alright, it didn't come out too bad. I didn't add types for unused arguments, but everything else should have one now.

site_scons/buildutils.py
ischoegl: that was clear :wink:, but PEP 484 by itself clarifies things here. (if this weren't an interface only seen by people digging deep under the hood, I would recommend to italicize *options* in the docstring)

site_scons/buildutils.py
ischoegl: It looks great :+1:, within the context that the aesthetics of PEP 484 may not be universally liked (I myself don't mind the annotations; on the contrary, I like the fact that it really helps when looking at an interface I am not familiar with).

Regarding unused arguments (e.g. 'target' and 'source' in `regression_test_message`), it may still help to know why they remain there (and what types they would be). You mentioned this in `print_report`, but it still is unclear what those required arguments represent.

site_scons/buildutils.py
ischoegl: The point is well taken, but it may not be universally known what the magic number `1` means for the exit status.

site_scons/buildutils.py
ischoegl: Looking over this again, I believe it's fine. Some of the context doesn't show up on when looking at the PR on the web interface (and trusty emacs isn't too helpful here either - although someone probably has written an extension). 

Regardless, opening the PR in VS code was a revelation ... :wink:

ischoegl (commented):
@bryanwweber ... feel free to consider all of my comments as resolved. I'll leave a final approval to those with more intricate knowledge of the internal workings of `scons` (myself, I'd have to rely on GH Actions passing). Apart from that, I am pretty happy that I can make more sense of `buildutils.py` after the refactoring.

SConstruct
```diff
@@ -1218,7 +1198,7 @@ if env['VERBOSE']:
     print(open('config.log').read())
     print('--------------------- end config.log ---------------------')
 
-env['python_cmd_esc'] = quoted(env['python_cmd'])
+env['python_cmd_esc'] = f'"{env["python_cmd"]!s}"'
```
speth: While I generally agree that f-strings are great, I feel like the `quoted` function was easier to read, and there are multiple gotchas to writing this out the next time one needs the behavior of `quoted`, like needing to remember to use `!s`, and understanding how nesting of the different quote types works with the f-string braces.

SConstruct
```diff
@@ -37,97 +37,100 @@ Basic usage:
...
+import textwrap
+from os.path import join as pjoin
+from pkg_resources import parse_version
+import SCons
 from buildutils import *
 
 if not COMMAND_LINE_TARGETS:
     # Print usage help
-    print(__doc__)
+    logger.info(__doc__, extra={"print_level": False})
```
speth: Are you trying to get rid of all usage of `print`? Why?

SConstruct
```diff
@@ -1479,16 +1459,23 @@ addInstallActions = ('install' in COMMAND_LINE_TARGETS or
...
-    if env['python_prefix']:
-        env['python_prefix'] = pjoin(os.getcwd(), env['stage_dir'],
-                       stripDrive(env['python_prefix']).strip('/\\'))
+if env["stage_dir"]:
+    stage_prefix = Path(env["prefix"])
+    # Strip the root off the prefix if it's absolute
+    if stage_prefix.is_absolute():
+        stage_prefix = Path(*stage_prefix.parts[1:])
+
+    instRoot = Path.cwd().joinpath(env["stage_dir"], stage_prefix)
```
speth, line 1478: While you're renaming everything?
```suggestion
    install_root = Path.cwd().joinpath(env["stage_dir"], stage_prefix)
```

SConstruct
```diff
@@ -1630,7 +1617,7 @@ env.Prepend(CPPPATH=[],
 
 # preprocess input files (cti -> xml)
 convertedInputFiles = set()
-for cti in mglob(env, 'data/inputs', 'cti'):
+for cti in multi_glob(env, 'data/inputs', 'cti'):
```
speth, line 1630: So are you replacing all of my beautiful single quotes with double quotes, or not? 😁

site_scons/buildutils.py
```diff
@@ -138,108 +248,107 @@ def regression_test(target, source, env):
...
 
-def compareFiles(env, file1, file2):
+def compare_files(env: "SCEnvironment", file1: Path, file2: Path) -> TestState:
     """
     Compare the contents of two files, using a method chosen based on
     their file extensions.
     """
-    if file1.endswith('csv') and file2.endswith('csv'):
-        return compareCsvFiles(env, file1, file2)
+    if all(f.suffix == ".csv" for f in (file1, file2)):
```
speth: I understand that need for some change because of `pathlib`, but I think the boolean expression is easier to read than the all/generator expression format when there are only two items.

site_scons/buildutils.py
```diff
@@ -1,78 +1,179 @@
...
+            "Cantera must be built using Python 3.6 or higher. If you are using "
+            "a system install of SCons, you can invoke SCons by writing\n"
+            "    python3 `which scons`\n"
+            "followed by any desired options."
+        )
+    sys.exit(1)
+
+
+class TestState(enum.IntEnum):
+    """Represent the passing state of a test.
```
speth: Maybe `TestResult` would be more clear? `TestState` makes me think about whether or not the test has run.

site_scons/buildutils.py
```diff
@@ -407,205 +545,182 @@ def compareCsvFiles(env, file1, file2):
...
+        if actual != default:
+            lines.append(f"    - actual: {actual!r}")
+        message.append("\n".join(lines))
 
-    return lines
+    logger.info("\n\n".join(message), extra={"print_level": False})
 
 
-def listify(value):
+def listify(value: "TListify") -> "List[str]":
```
speth: The alias `TListify` is pretty inscrutable, and only used in this one place. Can you just write the actual set of viable types here?

speth (changes requested):
I think most of my in-line comments are fairly minor.

The one slightly larger concern I have is that I don't really see the benefit to using the `logging` module here. I don't think it's possible to use any of this except when run as a script by SCons, so there's not really a case where this would need to interact with other logging activities, which is where the more complex logging machinery makes sense. This takes a chunk of additional setup in `build_utils.py`, plus has several end-use points where you are forced to do convoluted things like `logger.info(__doc__, extra={"print_level": False})` instead of the simple `print(__doc__)`.

SConstruct
bryanwweber: Yeah, I was thinking about this as well. Giving something a definite name (that is, `quoted`) has some value in indicating the intentionality of the change. I'll change it back.

SConstruct
bryanwweber: This is a great question, I'll address it in a top-level comment addressing the use of `logging` alltogether.

SConstruct
bryanwweber: Renaming this here would require renaming it in other places in this file, which I was trying to avoid (to keep this PR focused on changing `buildutils.py` and the minimum set of changes to keep things working). A few things slipped through that don't strictly follow that, like adding the `logger` in a few places, but that was my general thought.

SConstruct
bryanwweber: Same as the previous comment, this was a simple find-replace change for the `mglob` name. Don't you worry, those single quotes _must go_! 😄

bryanwweber:
> The one slightly larger concern I have is that I don't really see the benefit to using the `logging` module here. I don't think it's possible to use any of this except when run as a script by SCons, so there's not really a case where this would need to interact with other logging activities, which is where the more complex logging machinery makes sense. This takes a chunk of additional setup in `build_utils.py`, plus has several end-use points where you are forced to do convoluted things like `logger.info(__doc__, extra={"print_level": False})` instead of the simple `print(__doc__)`.

> Are you trying to get rid of all usage of `print`? Why?

@speth I considered a few things in choosing the `logging` module here. I'm not totally sold that it's the right choice, so I'd be happy to discuss. My reasoning was:

1. The `logging` module allows us (as developers) to communicate intent by the method name instead of with the first few words of the message.
2. It makes it easier to be consistent with message formatting. `info` messages will always say `INFO` unless you explicitly disable that, likewise `warning`, `error`, and `debug`. Combined with the first reason above, this means no more forgetting to add context about how serious a problem you think this output is.
3. It should allow users to have finer control over the output. SCons is really noisy IMO and our configuration and various other output doesn't help that. It'd be nice to have a setting for "Yes, I'm pretty sure everything is going to work, just give me a progress bar" or even no output at all. I think it will be possible, with `logging`, for a user to set a variable in `cantera.conf` (or on the command line) and select the level of output, and I don't think that'd be possible (or at least much more difficult) just by using `print` calls.
4. It's more of a "set-it-and-forget-it" approach to output, in terms of where the output is sent. Right now, I've configured it so that `debug`, `info`, and `warning` output is sent to `stdout` and `error` and higher are sent to `stderr`. Using `print` calls would require setting the `file` kwarg for error messages if we want to send error output to `stderr` (which I believe would be an improvement).

Now, the counterarguments (which you've noted, and which I also thought of in working on this and balanced in favor of `logging`):

1. It's a lot of configuration.
2. One of the big advantages of `logging` is collecting messages from long-running processes when you need the program to keep running, but want a log of any output; or, interacting with/combining several libraries and combining their output. Obviously, as you note, SCons scripts are basically standalone, run-once scripts, so these advantages are moot.
3. The convoluted method of printing "regular" output instead of log messages, via the `print_level` Boolean.

So, those are the thoughts that I had. I did consider a couple of other options for terminal output as well... One was, if you look at the commit history, I'm pretty sure it's still there where I started with two `Logger` instances which avoided the problem of the weird Boolean hack. On the other hand, it wasn't clear when to use each logger, so I didn't like it ☹️ 

I similarly thought about using a `Logger` for INFO/WARNING/ERROR messages, and `print()` for "output", but again, it wasn't clear to me how I would explain to someone when to use which, so I didn't do it. I suppose there's a similar problem with "when to use `print_level=False`, but at least that's self-documenting.

Let me know what you think! Like I said, I'm not totally sold on using `logging` here, so I'd be happy to hear any further thoughts you have.

ischoegl:
As a comment, I just checked out the branch to see what the new output would look like. On my Python 3.6.9 tool-chain (mostly vanilla Ubuntu 18.04), `scons` fails with the unhelpful error
```
docker@xps13:/src$ scons build
  File "/src/SConstruct", line 62

    logger.error(f"Unrecognized command line target: {command!r}")

                                                                ^

SyntaxError: invalid syntax

docker@xps13:/src$ python --version
Python 3.6.9
```

**PS:** I did run `scons clean` before (from `main`) ... and here's my `SCons` version

```
docker@xps13:/src$ scons --version
SCons by Steven Knight et al.:
	script: v3.0.1.74b2c53bc42290e911b334a6b44f187da698a668, 2017/11/14 13:16:53, by bdbaddog on hpmicrodog
	engine: v3.0.1.74b2c53bc42290e911b334a6b44f187da698a668, 2017/11/14 13:16:53, by bdbaddog on hpmicrodog
	engine path: ['/usr/lib/scons/SCons']
Copyright (c) 2001 - 2017 The SCons Foundation
```
which I believe is still the system's version (see toolchain configuration [here](https://github.com/ischoegl/cantera-docker/blob/master/ubuntu-toolchain/Dockerfile))

bryanwweber:
Thanks @ischoegl can you try running python3 ``which scons`` Sometimes, especially for the system SCons, it pulls the system Python 2 because of the shebang in the script

ischoegl:
I have `alias python=python3` set in `.bashrc` (which scons happily ignores, so I have another one in `cantera.conf`). Here's the output you requested:
```
docker@xps13:/src$ python3 which scons
python3: can't open file 'which': [Errno 2] No such file or directory
docker@xps13:/src$ which scons
/usr/bin/scons
```
The error for scons itself is somewhat surprising, as I've been using the toolchain without issues for a while ...

bryanwweber:
@ischoegl  Sorry, that's a failure of the GitHub Markdown formatting. You need backticks around "which scons", so that the shell runs it as a command and inserts the output after the `python`. So (trying again): ``python3 `which scons` ``

I think the problem is with the Python that's running SCons being the system Python (that is, Python 2), not the Python used to build the interface (`python_cmd`). Unfortunately, these are distinct things for the system-installed SCons...

ischoegl:
Alright, I was wondering! Here's the output ...
```
docker@xps13:/src$ python3 `which scons`

SCons build script for Cantera

Basic usage:
    'scons help' - print a description of user-specifiable options.

[...]
```

bryanwweber:
Thanks! Yeah, so I need to find some way to interrupt SCons and produce a useful error message when it's using the system Python 2. I'm not sure how to do that, it seems that SCons tries to import all the SConstruct files first, and finds that syntax error... I'm not sure why the check I put into `buildutils.py` isn't being triggered.

ischoegl:
Sorry, this may be a silly question, but this is a breaking change - how do I run this now? Imho, the expectation to type `scons build` as before and have things working isn't too far fetched. If it's the f-strings that cause this error, then they probably have to go.

.github/workflows/main.yml
```diff
@@ -163,27 +163,6 @@ jobs:
           --exclude='/doxygen/xml' --delete --delete-excluded \
           "${DOCS_OUTPUT_DIR}" ${RSYNC_USER}@${RSYNC_SERVER}:${RSYNC_DEST}
 
-  ubuntu-1804-py2:
```
ischoegl, line 174: I believe this has to remain

ischoegl (dismissed):
I believe CI should check that this will continue to work. Ubuntu 18.04 LTS will be supported for another 2 years.

speth:
> I'm not sure why the check I put into buildutils.py isn't being triggered.

The `SConstruct` file has to be *parsed* before it can get as far as trying to execute the `import` statement for `buildutils`. The best way of providing information about the problem that I can think of would be to stick an f-string near the top of `SConstruct` where the string itself contains the information telling the user what to do to fix it.

.github/workflows/main.yml
bryanwweber: Ubuntu 18.04 using Python 3 to run SCons is tested in the first build job. This only removes the case where Python 2 (that is, the system Python) is used to run SCons, which is no longer possible with this PR.

bryanwweber:
> > I'm not sure why the check I put into buildutils.py isn't being triggered.
> 
> The `SConstruct` file has to be _parsed_ before it can get as far as trying to execute the `import` statement for `buildutils`. The best way of providing information about the problem that I can think of would be to stick an f-string near the top of `SConstruct` where the string itself contains the information telling the user what to do to fix it.

Alternatively, I could move that check from `buildutils.py` back into `SConstruct` at the top of the file, and raise something more useful than a `SyntaxError`, right?

ischoegl:
@speth ... not necessarily elegant but likely very effective!

To document how this can be run anyways, it is (thanks @bryanwweber )
```
python3 `which scons` build
```

bryanwweber:
@ischoegl @speth Here's the error message I came up with (from `buildutils.py`), let me know if it can be improved:

```python
if sys.version_info[:2] < (3, 6):
    logger.error(
        "Cantera must be built using Python 3.6 or higher. If you are using "
        "a system install of SCons, you can invoke SCons by writing\n"
        "    python3 `which scons`\n"
        "followed by any desired options."
    )
    sys.exit(1)
```

ischoegl:
@bryanwweber ... error looks quite good but the system should know what the build option is so it *could* be more explicit?

bryanwweber:
@ischoegl That same command must be used for any SCons things that one wants to do... `scons build`, `scons test`, `scons install`, etc. So I didn't want to be specific about any one option.

speth:
> Alternatively, I could move that check from `buildutils.py` back into `SConstruct` at the top of the file, and raise something more useful than a `SyntaxError`, right?

No, the same thing applies. The whole file has to be parsed first before any statements are actually evaluated.

ischoegl:
> The whole file has to be parsed first before any statements are actually evaluated.

Not using f-strings would resolve the syntax errors. There's another issue with `pathlib` (per discussion with @bryanwweber), but that's likely easier to resolve.

ischoegl:
>  That same command must be used for any SCons things that one wants to do... scons build, scons test, scons install, etc. So I didn't want to be specific about any one option.

What I mean is that the build option added by the user should be known to SCons, and can potentially be inserted into the error message by constructing the string?

bryanwweber:
> > That same command must be used for any SCons things that one wants to do... scons build, scons test, scons install, etc. So I didn't want to be specific about any one option.
> 
> What I mean is that the build option should be known, and can potentially be inserted into the error message by constructing the string?

Ah, I see. Based on what @speth said, I don't think that will be possible, since the options won't have been set up until the file is parsed (which is interrupted by the `SyntaxError`).

ischoegl:
@bryanwweber ... I believe resolving the `SyntaxError` (by going back to older-style string formatting) would be the way to go, as everything else becomes much easier.

.github/workflows/main.yml
ischoegl: True. However, is there a way to check that the correct error is thrown in a GH action? I believe that's still worthwhile.

bryanwweber:
> @bryanwweber ... I believe resolving the `SyntaxError` (by going back to older-style string formatting) would be the way to go, as everything else becomes much easier.

@ischoegl I've been thinking about this a lot over the weekend. I don't agree this is the case, because it then limits the entire file to be compatible syntactically with Python 2 and Python 3, which defeats some decent percentage of the point of this rewrite. For instance, we can't use `f`-strings, we can't use keyword-only arguments, we can't use modern-style type annotations, and probably more. Moreover, that caveat would only apply to `SConstruct`, and I know that I'd forget to only use compatible syntax in that one file.

So, after trying many alternatives, I've landed on deliberately throwing a `SyntaxError` that gives a useful message, as @speth suggested, as the most robust and the easiest to implement solution to this issue.

@speth Do you have any more comments/thoughts about using logging here, as a followup to [my comment](https://github.com/Cantera/cantera/pull/1012#issuecomment-825326643)?

ischoegl:
@bryanwweber ... thanks for your response. It had been my assumption that the main issue with compatibility would be f-strings, hence my comment. I believe that there is a [work-around](https://stackoverflow.com/questions/35230635/type-hinting-in-python-2) for type hinting (PEP 484 is important to keep); regarding keywords-only, I'm not sure how much of a restriction it is, but I'm overall not sure that it's worth discussing further, as keeping Python 2 compatibility for the sake of having a better formatted error message is somewhat awkward.

Going forward, I believe there are two main things I hope this can accomplish:
* Make packaging/maintenance easier (which I believe is the point of this exercise)
* Ensure that things are reasonably robust in terms of older or deliberately stable systems (Ubuntu LTS, CentOS)

I'm definitely *not* arguing to leave Python 2 on life support ... . 

Regarding `logging`, I believe most of that information needs to be hidden (or redirected to a log file) by default, as showing some trivial logging info (e.g. `cti2yaml`) will only create questions on the UG even when there is no problem to start out with.

bryanwweber:
> Regarding `logging`, I believe most of that information needs to be hidden (or redirected to a log file) by default, as showing some trivial logging info (e.g. `cti2yaml`) will only create questions on the UG even when there is no problem to start out with.

Thanks @ischoegl! Yes, I hope that for the `scons` scripts, adding `logging` will make redirecting/hiding output easier. I'm not planning any changes to the converter scripts, though, just the build scripts.

ischoegl:
> I'm not planning any changes to the converter scripts, though, just the build scripts.

Agreed! However, `test_convert.py` appears to create a lot of logging output (at least on Ubuntu 18.04, *after* #1020 was merged) that I hope can be redirected one way or another.

bryanwweber:
@speth in d9c4bf1 I think I found a better solution to the `print_level` weirdness. It can now be passed directly as a keyword argument to any of the logging functions. Let me know what you think.

ischoegl:
Failing check is fixed by 0cc81add42976c4c3f654daae379ebf2d22285eb (per #1025) ... it should probably be applied to whatever is merged next as this part of CI is currently broken.

bryanwweber:
> Failing check is fixed by [0cc81ad](https://github.com/Cantera/cantera/commit/0cc81add42976c4c3f654daae379ebf2d22285eb) (per #1025) ... it should probably be applied to whatever is merged next as this part of CI is currently broken.

Thanks @ischoegl My philosophy is that anyone who would run into that particular issue would be a pretty experienced developer and could either file an issue or see that the commit was in a pull request. Since that's a non-essential job, and #1025 won't take that much longer to get merged, I don't see a problem with leaving it on that PR.

site_scons/buildutils.py
```diff
@@ -1,118 +1,280 @@
...
+
+
+if sys.version_info[:2] < (3, 6):
+    logger.error(
+        "Cantera must be built using Python 3.6 or higher. If you are using "
+        "a system install of SCons, you can invoke SCons by writing\n"
+        "    python3 `which scons`\n"
+        "followed by any desired options."
+    )
+    sys.exit(1)
```
speth: Given the `SyntaxError`-generating f-string in `SConstruct`, this is no longer necessary, right?

site_scons/buildutils.py
```diff
@@ -407,302 +597,289 @@ def compareCsvFiles(env, file1, file2):
...
-    # Fix the representation of boolean options
-    if actual == 'True':
-        actual = 'yes'
-    elif actual == 'False':
-        actual = 'no'
+        # Get the actual value in the current environment
+        if opt.key in env:
+            actual = env.subst(f"${opt.key!s}")
+            if actual in ("True", "False"):
+                actual = bool_options.get(actual == "True", actual)
```
speth: I'm not convinced this is more readable than what was here.

speth (changes requested):
Overall, I think I'm satisfied with the direction this has gone. There were just two minor suggested changes below.

I agree that there's some benefit to using `logging.info("...")` instead of `print("INFO: ...")`, and the `print_level` argument is not that bad for the handful of cases where we want to print something without a prefix.

Regarding a few of the comments, I'm not convinced that trying to reduce the verbosity of the build system would be productive. There are just too many cases where the details of the build log are the only thing that helps identify a problem. But this PR doesn't actually move in that direction, so this is kind of a moot point.

site_scons/buildutils.py
bryanwweber: Agreed!

speth (approved):

PR #995, opened 2021-03-19
Refactor reaction rate evaluators

ischoegl:
**Changes proposed in this pull request**

Use newly introduced `MultiBulkRates` rate evaluator to handle replacement classes, and deprecate current approaches. This PR will replace everything except `FalloffReaction` and introduce the following new classes:

- [x] `ElementaryReaction3`
- [x] `ThreeBodyReaction3`
- [x] `PlogReaction3`
- [x] `ChebyshevReaction3`

These classes will be used by default by YAML importers, although the old classes remain accessible from YAML. CTI and XML file import will use the old classes. The `3` will be dropped from class names after 2.6, and old class names will have `2` added (*Edit: likely the old classes will be removed*).

Items that will be addressed in subsequent PRs are:

* Factories for `ReactionRate` and templated `MultiRate` objects.
* Various implementations of `FalloffReaction`s
* Optimizations of `ThirdBody` handlers
* Reactions used by `InterfaceKinetics`
* Create infrastructure that allows for modifications of reaction rate parameters in memory for *new* reaction implementations

**If applicable, fill in the issue number this pull request is fixing**

Addresses Cantera/enhancements#84 and Cantera/enhancements#87

**Checklist**

- [x] There is a clear use-case for this code change
- [x] The commit message has a short title & references relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] The pull request is ready for review

PR #1014, opened 2021-04-17
Implement Python dict to C++ AnyMap conversion

ischoegl:

This PR adds the ability to create C++ `AnyMap` and `AnyValue` objects from the Python interface, where `dict_to_anymap`/`python_to_anyvalue` mirrors recently introduced `anymap_to_dict`/`anyvalue_to_python` (see PR #984). The following uses are anticipated:

* Instantiation of core objects from the Python API, e.g. reactions, see Cantera/enhancements#84
* Serialization of `--extra` fields in YAML serialization, see #1013
* Creation of user-data attached to various Cantera objects, see Cantera/enhancements#56

The new service functions enable the use of C++ constructors (or other C++ functions) that take `AnyMap` or `AnyValue` as an argument from the Cython layer. Round-trip conversion is covered by unit tests via a hidden helper function `cantera._cantera._py_to_any_to_py` in a new `test_utils.py` job. Beyond, the functions are not accessible from the Python API, as they are internal service functions that facilitate the transfer of data from Python to the C++ layer.

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/test/test_utils.py
+++ b/interfaces/cython/cantera/test/test_utils.py
@@ -29,7 +29,7 @@ class TestPyToAnyValue(utilities.CanteraTest):
         self.assertEqual(out, [])
 
     def test_set(self):
-        self.check_raises({'a', 'b'}, NotImplementedError, "Python sets")
+        self.check_raises({'a', 'b'}, NotImplementedError, "Python set")
 
     def test_empty_list(self):
         self.check_conversion([])
@@ -65,7 +65,7 @@ class TestPyToAnyValue(utilities.CanteraTest):
         self.check_conversion([True, False])
 
     def test_list_various(self):
-        self.check_conversion([True, 'spam', 3, 4.])
+        self.check_conversion([True, 'spam', 3, 4., {'foo': 'bar'}])
 
     def test_tuple(self):
         self.check_inexact_conversion((True, 'spam', 3, 4.))
@@ -78,7 +78,7 @@ class TestPyToAnyValue(utilities.CanteraTest):
 
     def test_ndarray3(self):
         self.check_raises(np.random.randn(3, 2, 4),
-            NotImplementedError, 'Cannot process float')
+            NotImplementedError, 'cannot process float')
 
     def test_nested_string(self):
         self.check_conversion([['spam', 'eggs'], ['foo', 'bar']])
@@ -94,19 +94,27 @@ class TestPyToAnyValue(utilities.CanteraTest):
 
     def test_raises_string(self):
         self.check_raises([[['spam', 'eggs'], ['foo', 'bar']]],
-            NotImplementedError, 'Cannot process string')
+            NotImplementedError, 'cannot process string')
+
+    def test_raises_named(self):
+        self.check_raises({'abcd': [[['spam', 'eggs'], ['foo', 'bar']]]},
+            NotImplementedError, "value with key 'abcd'")
 
     def test_raises_int(self):
         self.check_raises([[[1, 2, 3], [4, 5, 6]]],
-            NotImplementedError, 'Cannot process integer')
+            NotImplementedError, 'cannot process integer')
 
     def test_raises_float(self):
         self.check_raises([[[1., 2., 3.], [4., 5., 6.]]],
-            NotImplementedError, 'Cannot process float')
+            NotImplementedError, 'cannot process float')
 
     def test_raises_bool(self):
         self.check_raises([[[True, False], [False, True]]],
-            NotImplementedError, 'Cannot process boolean')
+            NotImplementedError, 'cannot process boolean')
+
+    def test_multi_dict(self):
+        vv = [{'a': [['spam', 'eggs'], ['foo', 'bar']], 'b': {'c': 4}}, {'d': 3}]
+        self.check_conversion(vv)
 
     def test_inhomogeneous(self):
         self.check_raises([[1, 2], ['a', 'b']], NotImplementedError, 'inhomogeneous')
--- a/interfaces/cython/cantera/test/test_utils.py
+++ b/interfaces/cython/cantera/test/test_utils.py
@@ -1,0 +1,124 @@
+import numpy as np
+
+import cantera as ct
+from . import utilities
+from .utilities import unittest
+
+from cantera._cantera import _py_to_any_to_py, _dict_to_any_to_dict
+
+class TestPyToAnyValue(utilities.CanteraTest):
+
+    def check_conversion(self, value):
+        out = _py_to_any_to_py(value)
+        self.assertEqual(out, value)
+
+    def check_inexact_conversion(self, value):
+        out = _py_to_any_to_py(value)
+        if isinstance(value, np.ndarray):
+            self.assertEqual(out, value.tolist())
+        else:
+            self.assertEqual(out, list(value))
+
+    def check_raises(self, value, ee, regex):
+        with self.assertRaisesRegex(ee, regex):
+            _py_to_any_to_py(value)
+
+    def test_none(self):
+        self.check_raises(None, NotImplementedError, "'None'")
+
+    def test_set(self):
+        self.check_raises({'a', 'b'}, NotImplementedError, "Python sets")
+
+    def test_empty1(self):
+        self.check_raises([], NotImplementedError, "empty sequences")
+
+    def test_empty2(self):
+        self.check_raises(np.ndarray((0,)), NotImplementedError, "empty sequences")
+
+    def test_scalar_string(self):
+        self.check_conversion('spam')
+
+    def test_scalar_int(self):
+        self.check_conversion(3)
+
+    def test_scalar_float(self):
+        self.check_conversion(3.1415)
+
+    def test_scalar_bool(self):
+        self.check_conversion(True)
+
+    def test_list_string(self):
+        self.check_conversion(['spam', 'eggs'])
+
+    def test_list_int(self):
+        self.check_conversion([1, 2, 3])
+
+    def test_list_float(self):
+        self.check_conversion([1., 2., 3.])
+
+    def test_list_bool(self):
+        self.check_conversion([True, False])
+
+    def test_list_various(self):
+        self.check_conversion([True, 'spam', 3, 4.])
+
+    def test_tuple(self):
+        self.check_inexact_conversion((True, 'spam', 3, 4.))
+
+    def test_ndarray1(self):
+        self.check_inexact_conversion(np.random.randn(10))
+
+    def test_ndarray2(self):
+        self.check_inexact_conversion(np.random.randn(3, 2))
+
+    def test_ndarray3(self):
+        self.check_raises(np.random.randn(3, 2, 4),
+            NotImplementedError, 'Cannot process float')
+
+    def test_nested_string(self):
+        self.check_conversion([['spam', 'eggs'], ['foo', 'bar']])
+
+    def test_nested_int(self):
+        self.check_conversion([[1, 2, 3], [4, 5, 6]])
+
+    def test_nested_float(self):
+        self.check_conversion([[1., 2., 3.], [4., 5., 6.]])
+
+    def test_nested_bool(self):
+        self.check_conversion([[True, False], [False, True]])
+
+    def test_raises_string(self):
+        self.check_raises([[['spam', 'eggs'], ['foo', 'bar']]],
+            NotImplementedError, 'Cannot process string')
+
+    def test_raises_int(self):
+        self.check_raises([[[1, 2, 3], [4, 5, 6]]],
+            NotImplementedError, 'Cannot process integer')
+
+    def test_raises_float(self):
+        self.check_raises([[[1., 2., 3.], [4., 5., 6.]]],
+            NotImplementedError, 'Cannot process float')
+
+    def test_raises_bool(self):
+        self.check_raises([[[True, False], [False, True]]],
+            NotImplementedError, 'Cannot process boolean')
+
+    def test_inhomogeneous(self):
+        self.check_raises([[1, 2], ['a', 'b']], NotImplementedError, 'inhomogeneous')
+
+    def test_ragged(self):
+        self.check_raises([[1, 2, 3], [4]], NotImplementedError, 'ragged')
+
+    def test_dict1(self):
+        self.check_conversion({'a': 1, 'b': 2., 'c': 'eggs', 'd': True})
+
+    def test_dict2(self):
+        dd = {'a': 1, 'b': 2., 'c': 'eggs', 'd': True}
+        self.assertEqual(dd, _dict_to_any_to_dict(dd))
+
+    def test_nested_dict1(self):
+        self.check_conversion({'a': 1, 'b': 2., 'c': {'d': 'eggs'}})
+
+    def test_nested_dict2(self):
+        dd = {'a': 1, 'b': 2., 'c': {'d': [1, 2, 3]}}
+        self.assertEqual(dd, _dict_to_any_to_dict(dd))
--- a/interfaces/cython/cantera/test/test_utils.py
+++ b/interfaces/cython/cantera/test/test_utils.py
@@ -98,7 +98,7 @@ class TestPyToAnyValue(utilities.CanteraTest):
 
     def test_raises_named(self):
         self.check_raises({'abcd': [[['spam', 'eggs'], ['foo', 'bar']]]},
-            NotImplementedError, "value with key 'abcd'")
+            NotImplementedError, "with key 'abcd'")
 
     def test_raises_int(self):
         self.check_raises([[[1, 2, 3], [4, 5, 6]]],
--- a/interfaces/cython/cantera/test/test_utils.py
+++ b/interfaces/cython/cantera/test/test_utils.py
@@ -2,7 +2,6 @@ import numpy as np
 
 import cantera as ct
 from . import utilities
-from .utilities import unittest
 
 from cantera._cantera import _py_to_any_to_py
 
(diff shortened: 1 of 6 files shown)


speth:
Well, I had a good portion of this implemented, as it is the logical extension to #984, but I guess you beat me to opening a PR. I pushed my work-in-progress on this to my [set-user-data](https://github.com/speth/cantera/tree/set-user-data) branch in case it's of any use.

ischoegl:
@speth ... sorry, didn’t realize. While there is some redundancy, my PR doesn’t include user data. The reason I implemented it was that not being able to move data from Python to AnyMap forced me to use awkward workarounds, and I wanted to have this available for #995.

ischoegl:
@speth ... I believe some of your ideas made the code simpler (especially using `set` operations). Shaved some 20 lines off, and the logic is clearer.

speth:
I updated the implementation on my branch, following on from what you have here. This version achieves a couple of things:
* It works for any combination of nested containers (where the base types are convertible), including "ragged" nested arrays and higher-dimensional arrays.
* The use of helper functions for populating the containers of like items avoids creating a bunch of unused containers on every call to `python_to_anyvalue`.

I hope you're willing to take this as a friendly amendment. If so, you can push those commits to this branch and we can continue the review from there (and of course you or @bryanwweber should provide a review of my changes).

ischoegl:
@speth ... I did a hard reset to your branch, so I believe at this point both of our ideas are blended. 

Overall, I have no objections to adopting the composite implementation, ~despite being somewhat sad to see `isEmpty` go, as I believe it to be more intuitive than `isType[void]`, but that's not worth haggling over. Regarding 3-dim (and higher) arrays I had blocked those as I didn't see a need, and treating them as `vector<AnyValue>>` is different from the logical extension of `vector<vector<vector<...>>>`; I guess it's a theoretical question at the moment and would only have repercussions in the C++ layer anyways. Of course, they are legitimate values (just like ragged), but my `NotImplementedError` would have tabled these question. Overall, I'm generally more concerned about what features and interfaces are available than how things are implemented. To qualify the last statement: performance and code readability do matter, but that's not a concern here.~ *Edit: those minor points are all resolved.*

@bryanwweber ... it's up to you to review I guess, as @speth's friendly amendments are an improvement over my initial code, and I'm ok with them (*Edit: my comments below are intended to clarify my points above*). As an aside: @speth in case any changes come up, feel free to push directly to my branch.

interfaces/cython/cantera/utils.pyx
```diff
@@ -76,6 +77,9 @@ cdef public PyObject* pyCanteraError = <PyObject*>CanteraError
 cdef anyvalue_to_python(string name, CxxAnyValue& v):
     cdef CxxAnyMap a
     cdef CxxAnyValue b
+    if v.isType[void]():
```
ischoegl: I had introduced `AnyValue::isEmpty` here as I felt that this was missing from the C++ API (was not the first time I was looking for this).

interfaces/cython/cantera/test/test_utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def test_ndarray1(self):
+        self.check_inexact_conversion(np.random.randn(10), 'vector<double>')
+
+    def test_ndarray2(self):
+        self.check_inexact_conversion(np.random.randn(3, 2), 'vector<vector<double>>')
+
+    def test_ndarray3(self):
+        # Each inner AnyValue holds a vector<vector<double>>
+        self.check_inexact_conversion(np.random.randn(3, 2, 4), 'vector<AnyValue>')
```
ischoegl: I had blocked these conversions as 'not implemented', as the logical extension of `test_ndarray2` would be to produce a `vector<vector<vector<double>>>` here. It won't be felt in the Python API, but any (future) C++ implementation would be affected.

bryanwweber:
I'll take a look later this week after I finish this round of grading.

interfaces/cython/cantera/utils.pyx
speth: You can drop that removal if you like (the C++ version is `v.is<void>()`, which looks slightly less weird, but that clashes with `is` being a keyword in Python). I don't think it necessarily needs to be exposed in Python, though. I'm hoping these conversion functions will pretty much be the boundary for the `AnyMap`/`AnyValue` classes in Python.

interfaces/cython/cantera/test/test_utils.py
speth: What I was aiming for was consistency with the YAML to `AnyMap` conversion: for an n-dimensional array, the last two dimensions are converted to `vector<vector<double>>`, and then stored in nested `vector<AnyValue>`.

In C++, we have to draw the line somewhere, since you can't generate these new types dynamically. Based on the additional complexity it leads to and the limited ways in which it's used, I'm sort of wondering if we should just do this for the last dimension only, both here and when reading from YAML.

interfaces/cython/cantera/utils.pyx
ischoegl: I'd be :+1: if we could keep `isEmpty` (and use it here) as I was hunting for this before in the context of `setParameters` within C++, and `v.is<void>()` definitely didn't come to mind when I was looking over the header file :wink:

interfaces/cython/cantera/test/test_utils.py
ischoegl: I completely agree that the line has to be drawn (hence my `NotImplementedError` in lieu of an implementation). ~To my knowledge, it isn't used at the moment, and anyone needing it in the future would have to be sufficiently familiar with C++ to make a decision on how to implement this based on an actual use case. That said, leaving the implementation as suggested is completely legitimate.~

Since YAML to `AnyMap` already solves this in the same fashion, this is obviously the way to go.

interfaces/cython/cantera/utils.pyx
ischoegl: Regarding the boundary for the Python API, I believe getting access to the unit system may be good, but there isn't a pressing need at the moment (and the Python unit PR's have to be resolved first anyways).

interfaces/cython/cantera/test/test_utils.py
ischoegl: Decided to 'hide' my last comment as it really doesn't address this properly. Since there's a YAML precedent, I have to reassess - I hadn't noticed that possibility in the reverse `anyvalue_to_python`, which I why I decided to leave this unimplemented.

Regarding the specific implementation: `vector<vector<double>>` isn't ideal, but it maps to 2D arrays that are commonly used. Another alternative would be to use a boost array (I guess), although I'm not sure I want to go there. So I'm happy to stick with `<vector<vector<...>>` as I cannot think of any use cases that go beyond at the moment.

interfaces/cython/cantera/utils.pyx
speth: Please go ahead and drop the last commit on this branch, then.

ischoegl: done.

interfaces/cython/cantera/test/test_utils.py
```diff
@@ -0,0 +1,117 @@
+import numpy as np
+
+import cantera as ct
+from . import utilities
+from .utilities import unittest
```
bryanwweber: This import appears to be unused

interfaces/cython/cantera/utils.pyx
```diff
@@ -117,5 +125,209 @@ cdef anyvalue_to_python(string name, CxxAnyValue& v):
...
+                itype.add(bool)  # otherwise bools will get counted as integers
+            elif isinstance(i, numbers.Integral):
+                itype.add(numbers.Integral)
+            elif isinstance(i, numbers.Real):
+                itype.add(numbers.Real)
+            else:
+                itype.add(type(i))
+
+        if itype == {numbers.Real, numbers.Integral}:
+            itype = {numbers.Real}
```
bryanwweber: 2 questions here:

1. Are the first two conditions needed? It seems as though `itype.add(type(i))` would do well enough for those.
2. For the numbers, could you combine that condition? Since you reduce the set to `numbers.Real` at the end anyways. Or is there a use to having just `numbers.Integral` as the type?

It might be worth adding comments that answer these questions, for any future readers.

interfaces/cython/cantera/utils.pyx
```diff
@@ -117,5 +125,209 @@ cdef anyvalue_to_python(string name, CxxAnyValue& v):
...
+
+        if itype == {list}:
+            inner_types = set()
+            ndim_inner = set()
+            for j in item:
+                type_j, ndim_j = get_types(j)
+                inner_types.add(type_j)
+                ndim_inner.add(ndim_j)
+            if len(inner_types) == 1 and len(ndim_inner) == 1:
+                return inner_types.pop(), ndim_inner.pop() + 1
```
bryanwweber, line 190: If `item` is a triply-nested list, will this return early?

interfaces/cython/cantera/utils.pyx
```diff
@@ -117,5 +125,209 @@ cdef anyvalue_to_python(string name, CxxAnyValue& v):
...
+    for i, item in enumerate(data):
+        v[i] = <double>item
+    return v
+
+cdef vector[long] list_int_to_anyvalue(data):
+    cdef vector[long] v
+    v.resize(len(data))
+    cdef size_t i
+    for i, item in enumerate(data):
+        v[i] = <long>item
```
bryanwweber: If an `int` is passted to `python_to_anyvalue`, it gets type `long int`, but if a list of `int`s is passed, it ends up just being `long`. Is there a difference?

interfaces/cython/cantera/utils.pyx
bryanwweber: I can see from code below that the answer to question 2 is, yes there is a use to having `numbers.Integral` as the type.

interfaces/cython/cantera/utils.pyx
```diff
@@ -117,5 +125,209 @@ cdef anyvalue_to_python(string name, CxxAnyValue& v):
...
+    for i, item in enumerate(data):
+        v[i] = dict_to_anymap(item)
+    return v
+
+cdef vector[vector[double]] list2_double_to_anyvalue(data):
+    cdef vector[vector[double]] v
+    v.resize(len(data))
+    cdef size_t i, j
+    for i, item in enumerate(data):
+        v[i].resize(len(item))
```
bryanwweber: My first thought here was there must be a way to optimize this, but I think this is this way because of the possibility of ragged arrays. Can you add a comment to that effect?

interfaces/cython/cantera/test/test_utils.py
```diff
@@ -0,0 +1,117 @@
...
+            self.assertEqual(held_type, check_type)
+
+    def check_raises(self, value, ee, regex):
+        with self.assertRaisesRegex(ee, regex):
+            _py_to_any_to_py(value)
+
+    def test_none(self):
+        out, held_type = _py_to_any_to_py(None)
+        self.assertEqual(out, None)
+        self.assertEqual(held_type, "void")
```
bryanwweber:
```suggestion
        self.check_conversion(None, "void")
```

interfaces/cython/cantera/test/test_utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def test_set(self):
+        # Sets are converted to lists
+        self.check_inexact_conversion({'a', 'b'}, 'vector<string>')
```
bryanwweber: Double quotes in the Python files please 😃

interfaces/cython/cantera/test/test_utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def test_dict(self):
+        self.check_conversion({'a': 1, 'b': 2., 'c': 'eggs', 'd': True}, 'AnyMap')
+
+    def test_nested_dict(self):
+        self.check_conversion({'a': 1, 'b': 2., 'c': {'d': 'eggs'}}, 'AnyMap')
+
+    def test_unconvertible(self):
+        class Foo: pass
+        self.check_raises(Foo(), ct.CanteraError, "Unable to convert")
```
bryanwweber: You can eliminate this class declaration by instantiating `object()` in the `check_raises()` call.

bryanwweber (changes requested):
Relatively minor changes here, thanks @ischoegl!

interfaces/cython/cantera/utils.pyx
ischoegl: They're both the same. As there was only one occurrence of `long int`, I replaced it by `long`.

interfaces/cython/cantera/utils.pyx
ischoegl: This will indeed return early, but triply-nested lists are converted to `vector<AnyValue>` with `AnyValue` subsequently converted to `vector<vector<doubl>>`; per (resolved) conversation above, 3D lists are not handled as `vector<vector<vector<double>>>`. There's a precedent for this behavior in the YAML importer, so skipping higher dimensions does not matter. I added a comment ...

interfaces/cython/cantera/utils.pyx
ischoegl: Ad (1) ... I don't believe that the two conditions can be merged. Of course the second one can be simplified to `isinstance(i, (list, tuple))`.

ischoegl:
@bryanwweber ... thanks for the review! I believe I resolved everything, so this should be ready to merge. The two comments where I left responses 'unresolved' pertain to code @speth (re-)wrote, so if there are remaining questions, I'll defer to him.

interfaces/cython/cantera/test/test_utils.py
```diff
@@ -0,0 +1,113 @@
...
+        self.check_inexact_conversion(np.ndarray((0,)))
+
+    def test_empty_dict(self):
+        self.check_conversion({})
+
+    def test_scalar_string(self):
+        self.check_conversion("spam", "string")
+
+    def test_scalar_int(self):
+        self.check_conversion(3, "long int")
```
bryanwweber, line 49: This should be `long`, right?

interfaces/cython/cantera/test/test_utils.py
```diff
@@ -0,0 +1,113 @@
...
+        self.check_conversion(3.1415, "double")
+
+    def test_scalar_bool(self):
+        self.check_conversion(True, "bool")
+
+    def test_list_string(self):
+        self.check_conversion(["spam", "eggs"], "vector<string>")
+
+    def test_list_int(self):
+        self.check_conversion([1, 2, 3], "vector<long int>")
```
bryanwweber, line 61: Likewise here

interfaces/cython/cantera/test/test_utils.py
ischoegl: Unfortunately not, as this comes from the C++ core. @speth consistently used `long int` in the C++ core in `AnyMap.h`, but in some instances, it is not possible to use `long int` in Cython: e.g. the cast `<long int>` is possible, whereas `v.isType[long int]()` throws an error. At this point, I believe it's most consistent to use `<long int>` for the type cast, and leave `long` in the other instances.

So I guess the difference is based on coding style (my personal preference would be to just use `long`).

interfaces/cython/cantera/test/test_utils.py
ischoegl: see above.

ischoegl:
@bryanwweber ... I hope I was able to clarify the discrepancy between `long int` and `long`.

ischoegl:
@bryanwweber ... any additional questions? I am mostly done rebasing #995 to include serialization (i.e. assuming that this would get merged before).

bryanwweber:
@ischoegl Sorry, I was waiting to see if @speth wanted to respond to the other unresolved questions.

interfaces/cython/cantera/utils.pyx
speth: @ischoegl is right that `list` + `tuple` check can be combined.

The reason for not just using `itype.add(type(i))` is that I want to treat any type that inherits from `dict` as a `dict`, and likewise any types inheriting from `list` or `tuple` are essentially `list`-like for the purpose of serialization. I added a few comments to this effect.

interfaces/cython/cantera/test/test_utils.py
speth: While I've been consistent about writing `long int`, I don't have that strong a preference for it. `long int` / `long` are barely used in Cantera outside of `AnyMap.h`/`AnyMap.cpp`, so it wouldn't be that invasive to just change it if other have a particular preference for `long`.

speth:
I responded to two of the open threads. I didn't have anything to add to the third -- I think it's all set now, right?

ischoegl:
@speth … thank you for your comments! All set from my side too.

interfaces/cython/cantera/test/test_utils.py
bryanwweber: I also don't have a preference one way or the other, except a preference for consistency. Whether we do it here or elsewhere, whatever the consistent choice is should be documented somewhere.

ischoegl: I looked over the code and as @speth indicated, there are probably less than 50 occurrences of `long int` in all of cantera. It's widely used in `ext` dependencies, so I believe there is some consistency and would propose to leave as is.

Regarding the shortened, `long` occurrence, it is necessary for
```
cbool isType "is" [T]()
```
where I haven't found a way to make it work for a two-word C++ type. I added a comment to document this exception.

bryanwweber (approved):

ischoegl:
:tada: ... thanks, @bryanwweber (and @speth)

PR #1015, opened 2021-04-19
Detect unspecified thirdbody collision partners

### LINKED ISSUE #967: Bug with the computation of [M] with real gases

ischoegl:

This PR is the second half of a fix for #967 (with the first half being taken care of by #968). I decided to post this as a dedicated PR instead of attaching the commits to #968, as review policies weren't clear, and this addition doesn't depend on the solution proposed in #968.

I believe the question on how to treat concentration has been settled in discussions in #967 and #968, but a process to convert reactions that explicitly specify a collision partner was still missing (per @decaluwe's suggestion). As an alternative to explicitly relabeling all affected reactions, it is possible to leave the existing syntax
```yaml
- equation: H + 2 O2 <=> HO2 + O2
  rate-constant: {A: 2.08e+19, b: -1.24, Ea: 0.0}
```
and use simple rules to identify this reaction as a `three-body` reaction in a pre-processing step (similar to what is already done for the detection of electrochemical reactions). I.e. `O2` would be detected as the collision partner, and treated as if one would manually specify
```yaml
- equation: H + O2 + M <=> HO2 + M
  rate-constant: {A: 2.08e+19, b: -1.24, Ea: 0.0}
  type: three-body
  default-efficiency: 0
  efficiencies: {O2: 1.0}
```

Specifically, I am suggesting the following four rules for converting an `elementary` to a `three-body` reaction:
1. one species has to occur on both sides
2. reaction `type` must *not* be specified
3. integer stoichiometric coefficients for all participating species
4. one side has to have exactly 3 participating species to qualify

Fixes #967 (as long as #968 is merged also)

**Checklist**


ischoegl:
@speth, @decaluwe and @rwest ... after making this its own PR - and adding unit tests - I believe this is ready for review and/or comments. If the approach is acceptable and finds approval, it should probably be merged at the same time as #968 so #967 can be resolved.

rwest (commented):
Looks OK to me! Thanks Ingmar

ischoegl:
@decaluwe ... I saw that you merged #968 (which is the companion PR). I resolved the merge conflict created by Blowers-Masel, so this is ready for review again.

include/cantera/kinetics/Reaction.h
```diff
@@ -212,6 +212,8 @@ class ThreeBodyReaction : public ElementaryReaction
     //! Relative efficiencies of third-body species in enhancing the reaction
     //! rate.
     ThirdBody third_body;
+
+    bool specified = false; //!< Explicitly specified collision partner
```
speth: I think this name is a little confusing, given that it ends up as an equal to the other members of the `Reaction` class like `reactants` and `reversible`. To borrow from the comment, maybe `specified_collision_partner` would be more clear?

interfaces/cython/cantera/test/test_kinetics.py
```diff
@@ -1221,9 +1221,11 @@ def test_modify_invalid(self):
             self.gas.modify_reaction(0, R2)
 
         # different reactants
-        R = self.gas.reaction(7)
+        R = self.gas.reaction(7) # implicitly defined three-body reaction
+        R2 = ct.ElementaryReaction(R.reactants, R.products)
+        R2.rate = R.rate
         with self.assertRaisesRegex(ct.CanteraError, 'Reactants are different'):
-            self.gas.modify_reaction(23, R)
+            self.gas.modify_reaction(23, R2)
```
speth: I think the original intent of this test could be satisfied by using any reaction that is still an `ElementaryReaction`. The use of reaction 7 here is arbitrary.

interfaces/cython/cantera/test/test_reaction.py
```diff
@@ -4,6 +4,58 @@
...
+        utilities.CanteraTest.setUpClass()
+        cls.gas = ct.Solution('gri30.yaml')
+
+    def test_implicit_three_body(self):
+        yaml1 = """
+            equation: H + 2 O2 <=> HO2 + O2
+            rate-constant: {A: 2.08e+19, b: -1.24, Ea: 0.0}
+            """
+        rxn1 = ct.Reaction.fromYaml(yaml1, self.gas)
+        self.assertEqual(rxn1.reaction_type, 'three-body')
```
speth: Shouldn't this also test that the default efficiency is zero and that efficiencies is equal to `{"O2": 1}`?

src/kinetics/ReactionFactory.cpp
```diff
@@ -159,6 +159,44 @@ ReactionFactory::ReactionFactory()
                });
 }
 
+bool isThreeBody(Reaction& R)
```
speth:
```suggestion
bool isThreeBody(const Reaction& R)
```

src/kinetics/Kinetics.cpp
```diff
@@ -143,8 +143,13 @@ std::pair<size_t, size_t> Kinetics::checkDuplicates(bool throw_err) const
...
             } else if (R.type() == "three-body") {
-                ThirdBody& tb1 = dynamic_cast<ThreeBodyReaction&>(R).third_body;
-                ThirdBody& tb2 = dynamic_cast<ThreeBodyReaction&>(other).third_body;
+                auto& R1 = dynamic_cast<ThreeBodyReaction&>(R);
+                auto& R2 = dynamic_cast<ThreeBodyReaction&>(other);
+                if (R1.specified != R2.specified) {
+                    continue; // No overlap as one third body is explicitly specified
+                }
+                ThirdBody& tb1 = R1.third_body;
+                ThirdBody& tb2 = R2.third_body;
```
speth: Does the duplicate reaction handling actually need to change? I guess I would be inclined to regard the two reactions below as unmarked duplicates, and I think that without this change, that would already be the case.
```yaml
- equation: H + O2 + H2O <=> HO2 + H2O
  rate-constant: {A: 1.126e+19, b: -0.76, Ea: 0.0}
- equation: H + O2 + M <=> HO2 + M
  rate-constant: {A: 1.126e+19, b: -0.76, Ea: 0.0}
  type: three-body
  default-efficiency: 0
  efficiencies: {H2O: 1}
```

src/kinetics/ReactionFactory.cpp
```diff
@@ -212,6 +250,16 @@ unique_ptr<Reaction> newReaction(const XML_Node& rxn_node)
         throw CanteraError("newReaction",
             "Unknown reaction type '" + rxn_node["type"] + "'");
     }
+    if (type == "") {
```
speth:
```suggestion
    if (type.empty()) {
```

test_problems/fortran/f77_ctlib_blessed.txt
```diff
@@ -1,53 +1,53 @@
...
-           46   1.8867924809455872E-002    3674698.329455073     
-           47   1.8867924809455872E-002   -2633122.304698291     
-           48   1.8867924809455872E-002    31625354.14713462     
-           49   1.8867924809455872E-002  -2.3592239273284576E-013
-           50   1.8867924809455872E-002    2079545.452265354     
-           51   1.8867924809455872E-002   -6636261.786209078     
-           52   1.8867924809455872E-002   -2239463.603775544     
-           53   1.8867924809455872E-002   -528516.5887338896     
+           1   1.8867924809455872E-002   107579642.89353865     
+           2   1.8867924809455872E-002  -127616418.89938366
```
speth, line 2: All these changes seem like they should be below the test tolerances. Was there something that was actually causing a failure?

src/kinetics/Reaction.cpp
```diff
@@ -858,6 +877,46 @@ BlowersMasel readBlowersMasel(const Reaction& R, const AnyValue& rate,
...
+            "in reaction '{}'.", R.equation());
+    }
+
+    R.third_body.default_efficiency = 0.;
+    R.specified = true;
+    auto sp = R.third_body.efficiencies.begin();
+
+    // adjust reactant coefficients
+    auto reac = R.reactants.find(sp->first);
+    if (abs(reac->second - 1) > SmallNumber) {
```
speth: I get a compiler warning about this: `warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]`. I think this (and the below instance) should be `std::abs`.

speth (changes requested):
This looks pretty good to me, @ischoegl. All of my comments are on fairly minor items. Beyond the in-line comments, there are just a couple of other things:
* Can you add a test for a reaction created via XML, given that you implemented this logic for both YAML and CTI/XML derived reactions?
* Have you thought about how these reactions should be serialized? Right now, they are converted to `three-body` reactions with the `default-efficiency` and `efficiencies` fields set as one might expect. I guess the question is whether or not that extra information should be deleted.

Lastly, I just wanted to note that I like that this ends up moving the third body to the end of the reactant / product strings for these reactions.

test_problems/fortran/f77_ctlib_blessed.txt
ischoegl: There was a change in `AR`, which previously had a spurious value but now is exactly `0`, which is the expected value but also causes a failure. I.e. I replaced the blessed output. ... this is line 49 btw. and all the other values just have one digit added!

ischoegl:
@speth ... thanks for your review. Hadn't thought about serialization as this is still a brand-new addition, but your point is well taken. I had to change duplicate checks but will look at this again.

> I just wanted to note that I like that this ends up moving the third body to the end of the reactant / product strings for these reactions.

Likewise. Have to admit that this was an unintended 'feature', and would probably have fought to keep the new behavior (as it is both more instructive and easier to implement).

**PS:** Regarding XML tests, those are implicitly included by `test_convert.py`.

src/kinetics/Kinetics.cpp
bryanwweber: This was also my main comment in looking at this PR. I agree that those reactions should be treated as unmarked duplicates; likewise something like `H + O2 + O2` and `H + 2 O2` should be treated as a duplicate (although I'd be surprised if that particular case wasn't already treated as a duplicate).

ischoegl: I think that’s a valid point. Regarding how this is treated internally, third body `O2`‘S are already detected, so how this is written doesn’t matter. My initial thought had mainly been that a user may *want* to keep those as unmarked duplicates, but it’s certainly safer the other way.

src/kinetics/Reaction.cpp
ischoegl: Replaced by `trunc(reac->second) != 1`, which may be clearer anyways.

interfaces/cython/cantera/test/test_reaction.py
ischoegl: it does this implicitly with the checks right below, but it won't hurt to add it explicitly.

src/kinetics/Kinetics.cpp
ischoegl: Done. Took me a while, as I didn't realize that a corresponding unit test would need to write a file (!) ... see #1019

ischoegl:
@speth (and @bryanwweber) ... I think this should take care of the review comments. I did not feel the need to implement XML unit tests, as those are (implicitly) covered by `test_convert.py`: it is already verified that equation types created from CTI and YAML correspond.

I did, however, implement a shortened serialization output, which I had not thought about.

src/kinetics/Reaction.cpp
```diff
@@ -276,28 +276,49 @@ ThreeBodyReaction::ThreeBodyReaction(const Composition& reactants_,
...
+    for (const auto& reac : reactants) {
+        if (reac.first != "M" && products.count(reac.first)) {
+            // detected specified third-body collision partner
+            specified_collision_partner_ = true;
+        }
+    }
+    if (!specified_collision_partner_) {
+        const ThermoPhase& rxn_phase = kin.thermo(kin.reactionPhaseIndex());
+        rate_units *= rxn_phase.standardConcentrationUnits().pow(-1);
+    }
```
speth: Sorry for not mentioning this before, but could you add some comments here to help explain the somewhat tricky behavior? I assume it all has to do with the fact that `calculateRateCoeffUnits` is often called before the rest of the setup for the specific reaction type, which would do things like remove `M` from the reactants and products list.

ischoegl: No problem. The logic implemented here is indeed to catch these cases. I added the comment as requested.

speth (approved):

ischoegl:
Thanks @speth!

PR #1020, opened 2021-04-25
Collect build/test timing information and fix some bugs

speth:

- Use `pytest` to run Python tests if it is installed
- Collect timing information for SCons commands and Python tests
- Start coverage tests immediately, rather than waiting for Ubuntu tests to finish
- Fix some diffusion flame extinction cases that weren't using the correct inlet temperature
- Fix `set_initial_guess` for `CounterFlowTwinPremixedFlame` so it doesn't overwrite deliberately-set profiles
- Disable the `test_ignition_delay_sensitivity` - it's a never-ending source of trouble, and the problems with it are already documented in Cantera/enhancements#55
- Speed up compilation a bit by adding `AnyMap.h` to the precompiled header.

*Note: I removed a bunch of changes from this PR that started to make material changes to the test suite, for the sake of getting something we can merge sooner and get back to having the CI servers working.*

**Checklist**

### DIFF
--- a/src/base/units.h
+++ b/src/base/units.h
@@ -10,8 +10,8 @@
 // This file is part of Cantera. See License.txt in the top-level directory or
 // at https://cantera.org/license.txt for license and copyright information.
 
-#ifndef CT_UNITS_H
-#define CT_UNITS_H
+#ifndef CT_LEGACY_UNITS_H
+#define CT_LEGACY_UNITS_H
 
 #include "cantera/base/ct_defs.h"
 #include "cantera/base/ctexceptions.h"
--- a/src/base/units.h
+++ b/src/base/units.h
@@ -3,8 +3,6 @@
  * Header for units conversion utilities, which are used to translate
  * user input from input files (See \ref inputfiles and
  * class \link Cantera::Unit Unit\endlink).
- *
- * This header is included only by file misc.cpp.
  */
 
 // This file is part of Cantera. See License.txt in the top-level directory or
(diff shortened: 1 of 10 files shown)


speth:
Closed and re-opened in the hope that this would get the coverage report to actually show up.

bryanwweber:
@speth Coverage is not uploaded if the tests fail

speth:
Yes, I know that, but it was missing before after several force-pushes where all of the tests did succeed. Right now, I'm trying to get the annotations for failing tests generated by https://pypi.org/project/pytest-github-actions-annotate-failures/ to show up.

speth:
Well, the annotations show up here: https://github.com/Cantera/cantera/actions/runs/787611863, which is somewhat helpful, but they don't show up in the PR under "Files Changed". I'm guessing the reason is because `pytest` sees the test's path as something like `../../build/python/cantera/test/test_kinetics.py`, rather than `interfaces/cython/cantera/test/test_kinetics.py`, and can't find the right file in the repository to annotate.

ischoegl (commented):
I believe this looks good.

src/base/units.h
```diff
@@ -10,8 +10,8 @@
 // This file is part of Cantera. See License.txt in the top-level directory or
 // at https://cantera.org/license.txt for license and copyright information.
 
-#ifndef CT_UNITS_H
-#define CT_UNITS_H
+#ifndef CT_LEGACY_UNITS_H
```
ischoegl, line 11: Not ever having noticed this file made me curious. It appears that the file docstring above is incorrect: `misc.cpp` no longer exists. At this point, this header is included in `application.cpp` and `global.cpp`.

ischoegl (changes requested):
I guess I did find something after all.

src/base/units.h
speth: I deleted that part of the docstring.

ischoegl: Out of curiosity - is this only used by CTI/XML or is it more pervasive?

ischoegl (approved):

PR #1010, opened 2021-04-09
Adding adaptive preconditioner functionality to Cantera

anthony-walker:
This code is adding preconditioning capabilities to Cantera/CVODEs solver interface for reactor networks. It is focused on a specific type of preconditioning but the code is being written with extensibility in mind. It adds Preconditioning capabilities to the numerics portion of cantera which contains a "PreconditionerBase" that is used to write future preconditioners. It also contains a class derived from this called "AdaptivePreconditioner" which is the aforementioned specific preconditioner.

There are also a couple of functions added to specific reactors. IdealGasReactor is one example. Currently, I added a function to evaluate the energy equation which is also done inside of evalEqs but I wanted to avoid touching core Cantera code as much as possible. In the future, I would like to collaborate with the appropriate people to stream line this.

**Checklist**

- [x] There is a clear use-case for this code change
- [x] The commit message has a short title & references relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] The pull request is ready for review

Issue #654, opened 2019-06-16
Fix implementation of advance_to_steady_state


ischoegl:
### Cantera version

all

### Operating System

all

### Python/MATLAB version

all

### Expected Behavior

The choice of the Feature Scaling approach in `advance_to_steady_state` to determine convergence should use an actual time derivative. 

### Actual Behavior

In the current implementation, convergence is checked for an absolute difference between integrator states. The step size is, however, internally chosen by CVODE and thus neither constant nor defined. As a consequence, convergence is poorly defined in the current code, as it depends on the CVODE integrator state as much as it does on `residual_threshold`.

PR #629 introduces an interface to time derivatives calculated by CVODE, which can be used to address this issue.

rwest:
It's now apparent this could use some work.  As Referenced by @12Chao in https://github.com/Cantera/cantera/pull/1305#issuecomment-1146313290 the answer may be to revive the 0D solver in #1021, but @ischoegl could you elaborate a little on how one could use #629 to address this?

ischoegl:
> @ischoegl could you elaborate a little on how one could use https://github.com/Cantera/cantera/pull/629 to address this?

The 0D solver in #1021 and solver derivatives obtained from #629 are orthogonal approaches. What I had in mind here was to use a convergence criterion based on vanishing derivatives (above ignition) rather than the 'feature scaling' (where I was never able to locate documentation that was sufficiently clear about the background).

**PS:** I don't think that this would help with the integration failures in https://github.com/Cantera/cantera/pull/1305#issuecomment-1146313290 though ...

12Chao:
> PS: I don't think that this would help with the integration failures in https://github.com/Cantera/cantera/pull/1305#issuecomment-1146313290 though ...

Could the `solve_steady` functionality mentioned in https://github.com/Cantera/cantera/pull/1021 be a substitute of the `advance_to_steady_state`?

ischoegl:
> Could the `solve_steady` functionality mentioned in #1021 be a substitute of the `advance_to_steady_state`?

in general, I believe so. But it’s a very different approach.

bryanwweber:
[...]

PR #1021, opened 2021-04-26
Steady-State Solver for ZeroD Reactors

paulblum:
Cantera can solve systems of equations representing physically zero-dimensional (i.e., time dependent-only) or one-dimensional (i.e., time + space) systems. The solution of the 1-D problems is done under the assumption of steady state, resulting in a differential-algebraic system of equations to solve. There is a solver implemented in Cantera as part of the 1-D code that solves these problems. 0-D systems can either be transient or steady state, represented by an ODE or algebraic equation respectively; at the moment, the only way to achieve solutions of the steady state problem is to integrate the transient problem until properties stop changing. We would like to modify or implement the existing steady-state 1-D solver for 0-D systems.

- Introduce `Newton`, a simplified version of Cantera's 1-D solver for the solution of 0-D steady-state systems.
- Add `solveSteady()` functionality to `ReactorNet` in C++ and `solve_steady()` in Python.
- Modify `DenseMatrix` for optimized Jacobian computation.

Fixes #

**Checklist**

- [ ] There is a clear use-case for this code change
- [ ] The commit message has a short title & references relevant issues


bryanwweber:
@gkogekar FYI, this is the work that @paulblum has been doing on the steady state solver.

ischoegl:
@paulblum ... as this just came up on the UG, what is the current status of this PR? - and to confirm: this should close Cantera/enhancements/issues/31, correct?

ischoegl:
@paulblum and @bryanwweber … this PR came up again in discussions recently. Are there any plans to push this over the finishing line?

bryanwweber:
@ischoegl Thanks for following up. I don't personally have any plans to work on this, and @paulblum has moved on to a PhD program in a different area, unfortunately, so he probably won't have time either. If you want to pick up the ball from here, please feel free to reuse this code/commits as necessary, or ditch it altogether.

ischoegl:
@bryanwweber ... thanks for the clarification and too bad indeed. It's not something I was intending to work on (and don't have any immediate plans at least at the moment).

src/numerics/Newton.cpp
```diff
@@ -0,0 +1,264 @@
...
+            if (solve(m_x.data()) < 0) {
+                writelog("\nTimestep failure after {} newton solves, {} timesteps.", newtonsolves, timesteps);
+                return -1;
+            }
+            timesteps++;
+        }
+        copy(m_x.begin(), m_x.end(), m_xsave.begin());
+    }
+    writelog("Failure to converge in {} steps.", MAX);
+    return 0;
```
speth, line 168: I think this solver loop only does a much more limited version of the hybrid time-stepping / steady solver structure that's implemented by the `MultiNewton` class, and is only alternating between solving the steady state and taking time steps of a fixed size. Given the sophistication of the `MultiNewton` algorithm, it might be worth seeing if we can pose this problem in a way that that solver can use directly.

src/zeroD/ReactorNet.cpp
```diff
@@ -430,4 +431,47 @@ size_t ReactorNet::registerSensitivityParameter(
...
+    for (int i = 0; i < m_nv; i++)
+    {
+        m_newt->setBounds(i, -1.0e-3, 1.01);
+    }
+    m_newt->setBounds(0, 1.0e-3, 100);
+    m_newt->setBounds(1, 0, 5);
+    m_newt->setBounds(2, -10000000, 10000000);
+
+    m_newt->setConstants({1});
+    // m_newt->setConstants({0,1,2,11,12});
```
speth, line 472: Settings bounds definitely needs to be moved to the individual reactor classes. I think these are meant to apply only to the base `Reactor` class, but even then the values for the first (mass) and second (volume) components don't quite make sense.

speth (commented):
I rebased this onto the current `main` branch so it doesn't get left too far behind. The comments below are partly just to remind me (or whoever finds time to pick this up again) of a couple things that I saw that need some work in the implementation.

Currently, there are no tests, and the simplest thing I could think of to test didn't actually work. This was to use the `mix1.py` example and replace the call to `sim.advance_to_steady_state()` with a call to `sim.solve_steady()`. The solver prints out the message "Failure to converge in 17 steps.", and the solution is clearly not the right steady-state.

ischoegl:
I have a relatively broad question here that follows up on  https://github.com/Cantera/cantera/issues/654#issuecomment-1146637653, reproduced here:

> #1021 and `advance_to_steady_state` are very different approaches to achieve the same goal. The proposal in https://github.com/Cantera/cantera/pull/1021 uses the existing steady-state solver for 1-D problems, which is a hybrid transient solver. The idea is, in steady state, the problem reduces to a set of equations that can be solved by Newton iteration... however, Newton iteration can be unstable, in the sense that it is very sensitive to the initial guess, and it can diverge if the initial guess isn't "good" enough. We know that time stepping will always bring the solution closer to the steady state, so https://github.com/Cantera/cantera/pull/1021 tries a direct Newton solution for the equations, and if it fails, takes a few timesteps, then tries the Newton solution again. This repeats until the Newton solver converges. As it happens, the accuracy of the timesteps doesn't have a strong effect on the Newton solver's final solution, so this solver doesn't use CVODES because it doesn't need all the accuracy that CVODES is capable of providing.
>
> On the other hand, `advance_to_steady_state` relies on CVODES and repeatedly calls the `advance` method until some stopping criteria are satisfied. This works because, as I said, timestepping always goes to the steady state (if there is one), but it is potentially much slower than the Newton solution, because time steps take a long time to solve, and you need to take a lot of them to maintain accuracy.

Fundamentally, there's nothing wrong with the 1D approach, but at least from my experience, it is prone to getting stuck for problems that involve marginal (low heating value) mixtures. 

But if I am informed correctly, the basic idea for the hybrid solver predates Cantera, and is essentially what was implemented for the original CHEMKIN. My main concern here is that we may be implementing 1980's ideas (which still may be worth keeping, but a lot has happened since) ... why not use CVODES for the transient solver, especially if we will be able to precondition soon? Are there better approaches than a standard newton solver? (e.g. CG etc.). Can we leverage analytical jacobians (which we are close to finally having after #1089 and #1010)? Can we leverage packaged root finders? Can we use some findings here for updated approaches to the 1D solver down the road?

bryanwweber:
> why not use CVODES for the transient solver, especially if we will be able to precondition soon? ... Can we leverage analytical jacobians (which we are close to finally having after #1089 and #1010)?

My understanding is that we simply don't need all the overhead of creating CVODES stuff, either the time or memory overhead, since the controls that CVODES provides are simply unnecessary. Likewise for analytical Jacobians.

> Can we leverage packaged root finders? 

Paul originally attempted KINSOL here, but packaged solvers don't tend to provide control on constraints (e.g., mass fractions) that are required for this application. They also don't provide the hybrid approach, and a Newton algorithm is fairly simple to implement, and it's already done for us and been validated for ~20 years... Again, just my understanding.

> Are there better approaches than a standard newton solver? (e.g. CG etc.).

This looks like great enhancement request if you can provide some more details 😄 As this implementation is relatively close to done, I don't think we should close this, unless replacement features are merged first though.

> Can we use some findings here for updated approaches to the 1D solver down the road?

Sure, that seems natural...

bryanwweber:
> it is prone to getting stuck for problems that involve marginal (low heating value) mixtures.

This sounds like a bug, or something that could be improved more generally. Opening an issue with a sample script that fails would be most welcome!

ischoegl:
@bryanwweber … would Cantera/enhancements#154 help with the KINSOL issues?

bryanwweber:
@ischoegl I don't think so. You still have constraints on the state variables to be positive, so that wouldn't be resolved. You'd also have to run the hybrid loop separately still, KINSOL can't handle that no matter what the state variables are 😞

ischoegl:
@bryanwweber ... I believe KINSOL does support constraints, see one of the examples 
https://github.com/LLNL/sundials/blob/e2f29c34f324829302037a1492db480be8828084/examples/kinsol/serial/kinFerTron_dns.c#L47-L48

and you can put constraints on number of iterations, see https://sundials.readthedocs.io/en/latest/kinsol/Usage/index.html#c.KINSetNumMaxIters

One big caveat is that I haven't spent any time with implementations here/

bryanwweber:
@ischoegl I haven't tried it personally either 🤷‍♂️ all I can say is that the people who have tried it (@paulblum and @speth IIRC) have noted that there are enough restrictions on using it that it isn't worth it for our case.

speth:
I think this is the wrong place for this general discussion on the relative merits of different approaches to solving the steady-state problem -- we already have an enhancement issue here: https://github.com/Cantera/enhancements/issues/31.

ischoegl:
Looks like this PR got an ‘honorable’ mention [here](https://chemrxiv.org/engage/chemrxiv/article-details/643b20dc73c6563f14f20f73) …

speth:
Superseded by #1907.

PR #1027, opened 2021-04-29
Update detection of skipped/undeclared species

### LINKED ISSUE #1024: Skipping undeclared species does not work with specific third body detection
**Problem description**

A phase definition specifying a subset of species from another mechanism, using the qualifier `declared-species` to select reactions and with the `skip-undeclared-third-bodies` option enabled, fails when there are third body reactions for specific colliders if that collider is not in the phase. This seems to be a corner case for the new handling introduced in #1015. 

**Steps to reproduce**

```python
import cantera as ct

gas_def = """
    phases:
    - name: gas
      species:
      - h2o2.yaml/species: [H, O2, H2O, HO2]
      thermo: ideal-gas
      kinetics: gas
      reactions:
      - h2o2.yaml/reactions: declared-species
      skip-undeclared-third-bodies: true
    """

gas = ct.Solution(yaml=gas_def)
gas.X = [1] * gas.n_species
for i in range(gas.n_reactions):
    R = gas.reaction(i)
    if isinstance(R, ct.ThreeBodyReaction):
        print(f"i = {i}: {R!s:30} {R.rate}")
```

**Behavior**

```
i = 0: H + O2 + M <=> HO2 + M         Arrhenius(A=2.8e+12, b=-0.86, E=0)
i = 1: H + O2 + O2 <=> HO2 + O2       Arrhenius(A=2.08e+13, b=-1.24, E=0)
i = 2: H + O2 + H2O <=> HO2 + H2O     Arrhenius(A=1.126e+13, b=-0.76, E=0)
i = 3: H + O2 + N2 <=> HO2 + N2       Arrhenius(A=inf, b=-1.24, E=0)
i = 4: H + O2 + AR <=> HO2 + AR       Arrhenius(A=inf, b=-0.8, E=0)
[20448590.35034667  8714720.31109497 72687079.49225374               nan
               nan]
```

For the last two reactions, which have `N2` and `AR` as third bodies, the pre-exponential factor ends up as infinity, and the reaction rate ends up being NaN. I think we need to find a way to skip adding these reactions, but it's not clear where to do so. `GasKinetics::addThreeBodyReaction` has access to the necessary information, but doesn't have a mechanism to make the determination, but by the point that it's called, all of the setup at the level of the parent `Kinetics` and `BulkKinetics` classes has already been completed.

I haven't looked yet, but I wonder if the refactoring in #995 provides a more obvious location to handle this case.

**System information**

- Cantera version: 2.6.0a2 / `main` at b753dd5e4.

ischoegl:

- Fix skipping of undeclared species with specific third body detection
- Streamline detection of undeclared species in other cases 
- Output all missing species, not just first
- Update formatting for `Reaction::validate`

~While `IonBurnerFlame` does act up, this is not the first time. I have skipped those tests for the time being.~

Fixes #1024

E.g. if `skip-undeclared-third-bodies=false`, the error is formatted as
```
InputFileError thrown by Reaction::checkSpecies:
Error on line 215 of /src/build/python/cantera/data/h2o2.yaml:
Reaction 'H + O2 + N2 <=> HO2 + N2'
is a three-body reaction with undeclared species: 'N2'
|  Line |
|   210 |   efficiencies: {O2: 0.0, H2O: 0.0, N2: 0.0, AR: 0.0}
|   211 | - equation: H + 2 O2 <=> HO2 + O2  # Reaction 7
|   212 |   rate-constant: {A: 2.08e+19, b: -1.24, Ea: 0.0}
|   213 | - equation: H + O2 + H2O <=> HO2 + H2O  # Reaction 8
|   214 |   rate-constant: {A: 1.126e+19, b: -0.76, Ea: 0.0}
>   215 > - equation: H + O2 + N2 <=> HO2 + N2  # Reaction 9
            ^
|   216 |   rate-constant: {A: 2.6e+19, b: -1.24, Ea: 0.0}
|   217 | - equation: H + O2 + AR <=> HO2 + AR  # Reaction 10
|   218 |   rate-constant: {A: 7.0e+17, b: -0.8, Ea: 0.0}
```

**Checklist**

### DIFF
--- a/src/kinetics/Reaction.cpp
+++ b/src/kinetics/Reaction.cpp
@@ -102,6 +102,15 @@ void Reaction::validate()
             }
         }
     }
+
+    // If reaction orders are specified, then this reaction does not follow
+    // mass-action kinetics, and is not an elementary reaction. So check that it
+    // is not reversible, since computing the reverse rate from thermochemistry
+    // only works for elementary reactions.
+    if (reversible && !orders.empty()) {
+        throw InputFileError("Reaction::validate", input,
+            "Reaction orders may only be given for irreversible reactions");
+    }
 }
 
 AnyMap Reaction::parameters(bool withInput) const
@@ -222,22 +231,64 @@ void updateUndeclared(std::vector<std::string>& undeclared,
     }
 }
 
-std::vector<std::string> Reaction::undeclaredSpecies(const Kinetics& kin) const {
-    std::vector<std::string> undeclared;
-    updateUndeclared(undeclared, reactants, kin);
-    updateUndeclared(undeclared, products, kin);
-    return undeclared;
-}
-
 std::vector<std::string> Reaction::undeclaredThirdBodies(const Kinetics& kin) const {
     std::vector<std::string> undeclared;
     return undeclared;
 }
 
-std::vector<std::string> Reaction::undeclaredOrders(const Kinetics& kin) const {
+bool Reaction::checkSpecies(const Kinetics& kin) const {
+
+    // Check for undeclared species
     std::vector<std::string> undeclared;
+    updateUndeclared(undeclared, reactants, kin);
+    updateUndeclared(undeclared, products, kin);
+    if (!undeclared.empty()) {
+        if (kin.skipUndeclaredSpecies()) {
+            return false;
+        } else {
+            throw InputFileError("Reaction::checkSpecies", input, "Reaction '{}'\n"
+                "contains undeclared species: '{}'",
+                equation(), boost::algorithm::join(undeclared, "', '"));
+        }
+    }
+
+    undeclared.clear();
     updateUndeclared(undeclared, orders, kin);
-    return undeclared;
+    if (!undeclared.empty()) {
+        if (kin.skipUndeclaredSpecies()) {
+            return false;
+        } else {
+            if (input.hasKey("orders")) {
+                throw InputFileError("Reaction::checkSpecies", input["orders"],
+                    "Reaction '{}'\n"
+                    "defines reaction orders for undeclared species: '{}'",
+                    equation(), boost::algorithm::join(undeclared, "', '"));
+            }
+            // Error for empty input AnyMap (e.g. XML)
+            throw InputFileError("Reaction::checkSpecies", input, "Reaction '{}'\n"
+                "defines reaction orders for undeclared species: '{}'",
+                equation(), boost::algorithm::join(undeclared, "', '"));
+        }
+    }
+
+    // Use helper function while there is no uniform handling of third bodies
+    undeclared = undeclaredThirdBodies(kin);
+    if (!undeclared.empty()) {
+        if (!kin.skipUndeclaredThirdBodies()) {
+            if (input.hasKey("efficiencies")) {
+                throw InputFileError("Reaction::checkSpecies", input["efficiencies"],
+                    "Reaction '{}'\n"
+                    "defines third-body efficiencies for undeclared species: '{}'",
+                    equation(), boost::algorithm::join(undeclared, "', '"));
+            }
+            // Error for specified ThirdBody or empty input AnyMap
+            throw InputFileError("Reaction::checkSpecies", input, "Reaction '{}'\n"
+                "is a three-body reaction with undeclared species: '{}'",
+                equation(), boost::algorithm::join(undeclared, "', '"));
+        }
+    }
+
+    return true;
 }
 
 ElementaryReaction::ElementaryReaction(const Composition& reactants_,
@@ -262,7 +313,7 @@ void ElementaryReaction::validate()
     Reaction::validate();
     if (!allow_negative_pre_exponential_factor &&
         rate.preExponentialFactor() < 0) {
-        throw CanteraError("ElementaryReaction::validate",
+        throw InputFileError("ElementaryReaction::validate", input,
             "Undeclared negative pre-exponential factor found in reaction '"
             + equation() + "'");
     }
@@ -412,11 +463,11 @@ void FalloffReaction::validate() {
     if (!allow_negative_pre_exponential_factor &&
         (low_rate.preExponentialFactor() < 0 ||
          high_rate.preExponentialFactor() < 0)) {
-        throw CanteraError("FalloffReaction::validate", "Negative "
+        throw InputFileError("FalloffReaction::validate", input, "Negative "
             "pre-exponential factor found for reaction '" + equation() + "'");
     }
     if (low_rate.preExponentialFactor() * high_rate.preExponentialFactor() < 0) {
-        throw CanteraError("FalloffReaction::validate", "High and "
+        throw InputFileError("FalloffReaction::validate", input, "High and "
             "low rate pre-exponential factors must have the same sign."
             "Reaction: '{}'", equation());
     }
@@ -700,7 +751,7 @@ void BlowersMaselReaction::validate()
     Reaction::validate();
     if (!allow_negative_pre_exponential_factor &&
         rate.preExponentialFactor() < 0) {
-        throw CanteraError("BlowersMaselReaction::validate",
+        throw InputFileError("BlowersMaselReaction::validate", input,
             "Undeclared negative pre-exponential factor found in reaction '"
             + equation() + "'");
     }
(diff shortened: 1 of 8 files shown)


speth (changes requested):
The failures in the ionized flame tests aren't trivial. It looks like the logic here is causing all third body and falloff reactions containing "M" as the third body to be left out. The `ch4_ion` input file should create a phase with 110 reactions, but it only contains 96 with the changes in this PR.

I'm a little surprised that these are the only tests that catches this problem.

As far as the implementation here, I wonder if it would make sense to move more of this checking into the `Reaction` classes, rather than having different implementations of this decision tree in `Kinetics::addReaction` and `getReactions`. It could almost be part of `Reaction::validate`, with the wrinkle that `validate` doesn't currently take the `Kinetics` object as an argument.

ischoegl:
> The failures in the ionized flame tests aren't trivial. 

Huh?! ... I remember checking on the number of reactions earlier also, but must have overlooked something here. It's pretty surprising that none of the other CI tests failed, and I had issues with `IonFlame` before.

Regarding the implementation: I retained the *old* philosophy where `Reaction::validate` checks for syntax, whereas `Kinetics::addReaction` ensures that all species are present and the reaction can actually be added. I believe it makes sense to keep this separation ...

speth:
Given that the only place where `Reaction::validate` is called is in `Kinetics::addReaction`, I'm not sure the distinction of what kind of checks belong where is that important. I guess there is a distinction between whether a `Reaction` is generically valid, and whether it is fully defined for a specific `Kinetics` object, so maybe this doesn't belong in the `validate` function specifically.

But I think we could avoid having to add three new functions to the `Reaction` interface, and avoid the growth in complexity of both `Kinetics::addReaction` and `getReactions`, especially since the two implementations differ in ways that don't entirely make sense to me. For instance, the version in `Kinetics::addReaction` goes through the effort of explaining which species are undeclared, while the version in `getReactions` does not provide this information. Plus, that way it's only one implementation that has to be debugged to fix the current issue.

ischoegl:
> [...] I guess there is a distinction between whether a `Reaction` is generically valid, and whether it is fully defined for a specific `Kinetics` object, so maybe this doesn't belong in the `validate` function specifically.

This is pretty much my point, and `validate` should be imho called after the `Reaction` setup outside of `Kinetics` (e.g. it could be called by the factory methods `newReaction`). That way that separation would be much clearer - it's a little more of a change than what I had intended to do, but it would make sense.

I see your point of not wanting to add more methods to `Reaction` than necessary. A middle ground would be to create a new `Reaction::checkSpecies` method, that gets called by `Kinetics::addReaction`. Note that once #995 is fully implemented, there wouldn't be a need to overload, as the handling of third bodies will be more generic and everything can be handled by a single method. One advantage would be that the same logic can be used by `getReactions`, where I did improve output, but not to the same extent as what is done elsewhere.

speth:
> This is pretty much my point, and `validate` should be imho called after the `Reaction` setup outside of `Kinetics` (e.g. it could be called by the factory methods `newReaction`). That way that separation would be much clearer - it's a little more of a change than what I had intended to do, but it would make sense.

You could do the validation in `newReaction` for reactions created from YAML, but this validation also applies for reactions that are built up purely in code, in which case it's sort of hard to do it before you get to `addReaction`. And we wouldn't want to do the validation twice for the most common case where the reactions are generated from an input file.

> I see your point of not wanting to add more methods to `Reaction` than necessary. A middle ground would be to create a new `Reaction::checkSpecies` method, that gets called by `Kinetics::addReaction`. Note that once #995 is fully implemented, there wouldn't be a need to overload, as the handling of third bodies will be more generic and everything can be handled by a single method. One advantage would be that the same logic can be used by `getReactions`, where I did improve output, but not to the same extent as what is done elsewhere.

I'm 👍 on `Reaction::checkSpecies`, and if it doesn't require any overloads once we get through #995, that's even better.

ischoegl:
> You could do the validation in `newReaction` for reactions created from YAML, but this validation also applies for reactions that are built up purely in code, in which case it's sort of hard to do it before you get to `addReaction`. And we wouldn't want to do the validation twice for the most common case where the reactions are generated from an input file.

Fair point. For the new structure, I started to differentiate between validation of rate parameters and reaction equations, which would make that part easier, as each can be tied to calling specific functions with known behavior. From my perspective, it would still be worthwhile to run the check immediately (i.e. prior to `addReaction`), no matter whether a reaction is constructed from YAML or built in code. I believe I have an idea and will suggest that once I get back to updating this PR.

> [...] if it doesn't require any overloads once we get through #995, that's even better.

Unfortunately overloads will be necessary until CTI/XML support is removed as I do not anticipate porting the old loaders to the new reaction framework.

ischoegl:
@speth ... I fixed the logic, and moved the tests as discussed (I also moved the `checkReactionBalance` check so it will cover `getReactions`). I ended up not moving the `Reaction::validate` calls to avoid conflicts with other ongoing work.

speth:
For the case of `ch4_ion.yaml`, I'm still seeing one issue, which is that the reaction `H + O2 + AR <=> HO2 + AR` is still getting added, even though the phase doesn't define `AR` as a species. This results in a reaction where the `forward_rates_of_progress` entry is `nan`, as are the `net_production_rates` for `H', `O2` and `HO2`.

Interestingly, the same reaction is also added when creating a phase from `ch4_ion.cti`, but in that case the rate for this reaction is just zero, and the net production rates are fine. This is presumably why the ionized flame tests are working.

ischoegl:
@speth ... thanks for catching this: my earlier fix of the logic glitch broke something that had worked before ... turns out that the logic needs to work differently for reactions with specified collision partners. It should all be good now - I also switched the unit tests to `ch4_ion.yaml` to make sure.

src/kinetics/Reaction.cpp
```diff
@@ -86,20 +86,31 @@ void Reaction::validate()
     if (!allow_nonreactant_orders) {
         for (const auto& order : orders) {
             if (reactants.find(order.first) == reactants.end()) {
-                throw CanteraError("Reaction::validate", "Reaction order "
-                    "specified for non-reactant species '" + order.first + "'");
-            }
+                throw InputFileError("Reaction::validate", input,
+                    "Reaction order specified for non-reactant species '{}'",
+                    order.first);
+           }
```
speth, line 92: ❤️ the expanding use of `InputFileError`

src/kinetics/Reaction.cpp
```diff
@@ -211,6 +222,126 @@ void Reaction::calculateRateCoeffUnits(const Kinetics& kin)
...
+
+std::pair<std::vector<std::string>, bool> Reaction::undeclaredThirdBodies(
+        const Kinetics& kin) const {
+    std::vector<std::string> undeclared;
+    return std::make_pair(undeclared, false);
+}
+
+void Reaction::checkBalance(const Kinetics& kin) const {
+
+    Composition balr, balp;
```
speth: If you want the visual space here, I'd suggest
```suggestion
void Reaction::checkBalance(const Kinetics& kin) const
{
    Composition balr, balp;
```
Cantera's code isn't particularly consistent about this minor formatting detail, so I think it's fine with the brace in either place, but that might be a way of solving our opposite preferences about blank lines at the start of a block. The one case where I do think having the brace on a new line by itself is especially useful is when the function signature end up being split across multiple lines so you don't end up with a weird looking dedent.

src/kinetics/Reaction.cpp
```diff
@@ -211,6 +222,126 @@ void Reaction::calculateRateCoeffUnits(const Kinetics& kin)
...
+            throw InputFileError("Reaction::checkSpecies", input, "Reaction '{}'\n"
+                "is a three-body reaction with undeclared species: '{}'",
+                equation(), boost::algorithm::join(undeclared, "', '"));
+        } else if (kin.skipUndeclaredSpecies() && specified_collision_partner_) {
+            return false;
+        }
+    }
+
+    //
+    checkBalance(kin);
```
speth, line 341: Rogue comment line

speth (commented):
Thanks, @ischoegl, this looks good. I have only a couple of incredibly minor comments below.

src/kinetics/Reaction.cpp
ischoegl: Of course ... it provides superior feedback.

src/kinetics/Reaction.cpp
ischoegl: Not a problem. I don't have hard preferences when it comes to coding style.

speth (approved):

PR #1030, opened 2021-05-08
Fix memory error when deallocating Python reaction rate wrappers

speth:
Using the value of member variables that are Python objects in `__dealloc__` is not reliable, because they may have already been
deleted by the time `__dealloc__` is called [[Cython docs]](https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#finalization-method-dealloc). Instead, we need a C data member to check for whether or not the rate object owns the underlying C++ object and should delete it.

The only way I've been able to trigger this bug is interactively in IPython, when trying to access members of the rate object with tab autocompletion. I think the reason for this is that this scenario ends up invoking the garbage collector, rather than having objects just get deleted when their reference counts go to zero, and that ends up behaving differently. I'm honestly surprised I didn't run into this before -- this is the way this has worked as long as this information has been accessible in Cython.

- Add C++ boolean `own_rate` to Cython `Arrhenius` and `BlowersMasel` classes and use this when checking whether to delete the pointer to the corresponding C++ objects.

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -1130,10 +1130,12 @@ cdef class CustomReaction(Reaction):
 
 cdef class Arrhenius:
     cdef CxxArrhenius* rate
+    cdef cbool own_rate
     cdef Reaction reaction # parent reaction, to prevent garbage collection
 
 cdef class BlowersMasel:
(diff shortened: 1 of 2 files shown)


ischoegl (approved):
Interesting - fix looks good to me.

PR #1031, opened 2021-05-09
Enable modification of user input data fields in Python

speth:

- Adds `update_user_data` and `clear_user_data` methods to Python objects with an `input_data` property

This allows custom data to be stored when writing to YAML files. Implementation of this was, as expected, 🍰 with the `dict_to_anymap` method already in place thanks to #1014. 

**Checklist**


interfaces/cython/cantera/base.pyx
```diff
@@ -232,6 +232,21 @@ cdef class _SolutionBase:
         def __get__(self):
             return anymap_to_dict(self.base.parameters(True))
 
+    def update_user_data(self, data):
+        """
+        Add the contents of the provided `dict` as additional fields when generating
+        YAML phase definition files with `write_yaml` or in the data returned by
+        `input_data`.
+        """
```
bryanwweber: We should document what happens when duplicate keys are present. Presumably, the key is updated with the new value, but it'd be good to be explicit about explaining that.

bryanwweber: Also, whether this is a recursive update, or if it will overwrite sub-fields

ischoegl (commented):
I may have some thoughts later this week.

ischoegl:
OK ... I mostly have a philosophical question about the implementation, which boils down to what `input`/`input_data` actually means. 

When I first saw the term `input` in #984, I thought of it as being 'original input data', but going through I quickly realized that as it (still) stands it is 'input data for a future incarnation'. Prior to this PR, there was relatively little impact, as once instantiated, there wasn't much of a way of adding/changing the parameterization. I.e. thus far, it was indeed `input`.

Going forward, I can see two options for how users interact with objects:
1. Users change objects as they go, and would have to reload the original `Solution` if they want to reset to the original *input*
2. Users have the ability to walk back changes at any time (e.g. by calling a `Solution.reset()` or `Reaction.reset()`, etc. to restore original data).

The PR as it stands is (1), and is a legitimate 'solution'; I mainly want to make sure that there may not be a potential use case for (2) as this would be the time to start. The implementation would be quite simple, where all that is needed is a second `AnyMap` member variable that holds information added after the objects are instantiated, plus some tweaks for logic in some of the getters.

speth:
It occurred to me recently that with the addition of YAML serialization, even the phrase "input file" is a little bit weird, since now you can output an input file. I'm certainly open to suggestions for alternative terminology here, as I do see why the name `input_data` might have led you down that path.

I don't quite know what a potential use case for (2) would be, but as you said, this information can be retrieved by loading another `Solution` (or whatever) from the input file. So I don't think that we're precluding any possible uses by not differentiating between the original input data and ways that the object has been modified since the input file was loaded. Realistically, I think that `Solution.reset()` would be as expensive as instantiating a new `Solution` in many instances.

decaluwe:
Re "input file" vs. some other name, I've been using "mechanism file" in my own code, which I think works.  

The rebuttal question that immediately comes to mind, though: even if we are "outputting" a YAML file, is is only useful/a thing we care about to the extent that it will one day be imported as an "input" for some other, future simulation/numerical experiment?

I actually think the answer is "no - it has other uses." i.e if you are fitting experimental data and adjusting the mechanism details on the fly, those rates, thermo values, etc, would be the result of the fitting, and something you would report out.

interfaces/cython/cantera/_cantera.pxd
```diff
@@ -187,6 +191,7 @@ cdef extern from "cantera/thermo/ThermoPhase.h" namespace "Cantera":
         string type()
         string phaseOfMatter() except +translate_exception
         void getSpeciesParameters(string, CxxAnyMap&) except +translate_exception
+        CxxAnyMap& input()
```
bryanwweber, line 194: This maybe is a little late to catch this, but why do some classes use a getter method while others just expose the `AnyMap` directly?

interfaces/cython/cantera/base.pyx
```diff
@@ -232,6 +232,21 @@ cdef class _SolutionBase:
         def __get__(self):
             return anymap_to_dict(self.base.parameters(True))
 
+    def update_user_data(self, data):
+        """
+        Add the contents of the provided `dict` as additional fields when generating
+        YAML phase definition files with `write_yaml` or in the data returned by
+        `input_data`. Existing keys with matching names are overwritten.
```
bryanwweber, line 239: The docstring for `AnyMap::update` says that

> If keys in `other` also exist in this AnyMap, the `keepExisting` option determines which item is used.

By my reading, setting `keepExisting` as `True` means that the existing keys are kept and duplicate keys are _not_ overwritten. Can you clarify?

interfaces/cython/cantera/base.pyx
```diff
@@ -232,6 +232,21 @@ cdef class _SolutionBase:
...
+        """
+        self.thermo.input().update(dict_to_anymap(data), True)
+
+    def clear_user_data(self):
+        """
+        Clear any user-provided auxillary input data, so that the data given by
+        `input_data` or `write_yaml` will only include values generated by Cantera.
+        """
+        self.thermo.input().clear()
```
bryanwweber, line 249: I don't see how this distinguishes between data that the user has added using `update_user_data` and the data coming from the YAML file that was used to create the `Solution`. Can you clarify that for me? Thanks :smile:

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -569,6 +581,27 @@ def test_yaml_inconsistent_species(self):
         with self.assertRaisesRegex(ct.CanteraError, "different definitions"):
             gas.write_yaml('h2o2-error.yaml', phases=gas2)
 
+    def test_yaml_user_data(self):
+        gas = ct.Solution('h2o2.yaml')
```
bryanwweber:
```suggestion
        gas = ct.Solution("h2o2.yaml")
```

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -569,6 +581,27 @@ def test_yaml_inconsistent_species(self):
...
+        extra = {"spam": {"A": 1, "B": 2}, "eggs": [1, 2.3, 4.5]}
+        gas.update_user_data(extra)
+        S = gas.species(2)
+        S.update_user_data({"foo": "bar"})
+        S.transport.update_user_data({"baz": 1234.5})
+        S.thermo.update_user_data({"something": (False, True)})
+        gas.reaction(5).update_user_data({"baked-beans": True})
+
+        gas.write_yaml("h2o2-generated-user-data.yaml")
+        gas2 = ct.Solution("h2o2-generated-user-data.yaml")
```
bryanwweber, line 596: This may be more of a feature request not appropriate to add here, but it'd be nice to be able to "write" the output into a `StringIO` buffer to avoid the filesystem entirely. Alternatively, to have a method that just returns a string of the generated YAML, which you can pass to the `yaml` kwarg of `Solution`.

bryanwweber (changes requested):
Thanks @speth. I don't have any strong opinions about the naming distinction, since as @decaluwe points out, these files can be used in many different contexts. "Mechanism file" implies to me that it involves some reaction mechanism, which also may not always be the case. "Input file" seems to be the best choice to me, but no strong opinion.

I do have several questions here about the implementation, probably because this is my first time really digging into the details. So if my questions/suggestions don't make any sense, please take them in the "please teach me something" spirit 😄

interfaces/cython/cantera/_cantera.pxd
ischoegl: I actually had the same question when this was first introduced (and posted as much). If I remember the justification correctly, in some instances, the `AnyMap` holds static information and is accessed directly. In other instances, it needs to be assembled, and thus uses a getter.

speth: The `input()` getter is just a simple getter in all cases where it's used -- the `getParameters(AnyMap&)` or `AnyMap parameters()` methods are the ones that do work to determine their output.

The cases where `input` is just a public member variable arose in classes that were (at least originally) basically data containers where all of the members were public and there were few if any member functions. The `GasTransportData` class still mostly follows this model, for example. For other classes that don't expose any public member data, e.g. `ThermoPhase`, the getter function is used.

I think debating whether the classes with public data members were a good idea is a ship that sailed a while ago. I'm pretty sure the `input` member for most if not all of these is already present as part of the 2.5 release.

ischoegl: Thanks for reiterating that this dates back to before the recent addition of YAML serialization. I agree that this is now de facto part of the released API.

bryanwweber: Although this has been released already, I think considering whether we can improve the consistency of the API is worthwhile, at least briefly. My concern is that anyone new coming into this code (myself, for example) will look at this and have no idea why some classes have public member data and some have getters, and since there's not really an apparent pattern as to which classes have which, it makes understanding the code that much more difficult. I'm trying to keep in mind that small inconsistencies and gotchas like this are things that we have to teach to the next generation, and there's already more than enough of that to go around 😄 

The simplest solution is probably to add getter methods to classes with public member data and prefer using those, even if the member data remains public. That way we can document the behavior as semi-intentional and have consistence of interfaces. I'm not sure if this would be consistent with C++ coding style, though (I imagine not).

ischoegl: FWIW, the coding guide does not explicitly disallow public member variables, as it states: "Protected and private member variable names are generally prefixed with `m_`. For most classes, member variables should not be public."

The way I view it, `input` (as getter or member variable), is mostly for internal use while the `getParameters(AnyMap&)` or `AnyMap parameters()` methods provide a consistent interface. From that perspective, adding the clarification `@internal` to all `input` docstrings may sufficiently resolve this controversy. Having multiple interfaces for a rarely used method is imho not necessary (neither is the significant effort for deprecation of a public member variable, which is a major undertaking).

**PS:** that said, I generally agree that consistency should be attempted whenever possible. But once something is part of the released API, it's tough.

bryanwweber: OK, this has evolved into much more discussion than I intended. I think at this point, the conversation has resolved itself and no changes to the code are needed in this PR. In the future, we may consider some way to resolve the consistency issue that I see here, to avoid confusing future developers, since as I said earlier, there's no pattern that is obvious to me which classes have direct access and which use a getter.

speth: If you want to see how difficult it is to change the _type_ of an existing member (which is roughly equivalent to replacing a public member variable into a getter) without breaking things for existing users, take a look at #995. I think it's probably worth just living with the inconsistency.

The suggestion of marking the `input` member as internal is an interesting one. The problem with the Doxygen `@internal` command is that its only effect is to simply exclude the documentation for that member from the generated docs depending on a specific flag. Right now, we have it set to include all documented members, given that the API docs are a useful resource for developers who do want to be able to use these internal members.

bryanwweber: > If you want to see how difficult it is to change the _type_ of an existing member (which is roughly equivalent to replacing a public member variable into a getter)

Just to be clear, I'm not suggesting _replacing_ the public member variable with a getter, but adding a getter _in addition_ to the public access, with a consistent and documented preference for using the getter _within Cantera's code_. There are (at least) 2 caveats to that: 1) I'm not sure whether the compiler would allow that and 2) I'm not sure how bad a violation (if at all) of implicit C++ code style guides it would be (that is, if an experienced C++ programmer looked at that, would they insist on removing it). I know that this is suboptimal, in terms of wanting to have one way to do things. However, I think it would be a benefit for Cantera developers (us, future us, and other future developers) to have a consistent way to access this data, since there isn't an obvious (to me) pattern that governs when a public member is used vs. when a getter is used.

interfaces/cython/cantera/base.pyx
speth: Good catch -- I had this backwards. The tests have been updated to verify this as well.

interfaces/cython/cantera/base.pyx
speth: It doesn't distinguish - both are erased here. But once the object's been constructed, Cantera doesn't need a pristine copy of the data from the input file, so clearing this doesn't have any noticeable effect besides removing the auxiliary data, since the necessary fields will be calculated through the calls to `getParameters`.

bryanwweber: So in principle, this does not support true round-tripping of an input file, right? If all the data are cleared, and any data computed by Cantera is not quite identical to the original file, the resulting generated file wouldn't be identical?

This may be a pedantic thing, but since the docstring says one thing, and the code actually does something else, I foresee potential problems as a result.

interfaces/cython/cantera/test/test_composite.py
speth: Given that the `write_yaml` method was introduced in #984 and not here, I think this is a separate feature request.

interfaces/cython/cantera/base.pyx
speth: Again, I think we're getting pretty far off the purpose of _this_ PR, as this basically all about features that were introduced in #984 if not earlier, but I can try to provide a bit of an explanation.

We don't keep track of how an object has changed since it was instantiated from the input file, so any time you ask for the `input_data` or call `write_yaml`, that's being calculated on the fly. I don't think there's anything that promises to precisely round-trip an input file, and there are currently a variety of ways in which that isn't currently possible. Some of these will probably be addressed relatively soon, like the serialization of the file-level metadata (as requested in #1013), while others may not (like retaining comments in YAML files). The only promise that I think we should be trying to make regarding serialization is that you can generate an input file that will recreate an object in its current state, and I think that is currently the case.

ischoegl: > So in principle, this does not support true round-tripping of an input file, right? If all the data are cleared, and any data computed by Cantera is not quite identical to the original file, the resulting generated file wouldn't be identical?
>
> This may be a pedantic thing, but since the docstring says one thing, and the code actually does something else, I foresee potential problems as a result.

This is quite similar to what I had mentioned in one of my comments above (what does *input data* mean?). Either path forward is fine; I believe round-trip conversion would be simple to implement as stated earlier:

> The implementation would be quite simple, where all that is needed is a second `AnyMap` member variable that holds information added after the objects are instantiated, plus some tweaks for logic in some of the getters.

bryanwweber: I think it's fine if we consider changing/discussing the round-trip behavior out of scope for this PR. That said, I'd prefer if this docstring were a little more accurate to reflect what's actually happening. Since this very explicitly says that user-provided data will be cleared, it implies that's the only thing that's cleared. Maybe something like:

> Clear auxiliary input data, including user-provided values. When `write_yaml` or `input_data` are called, Cantera will regenerate the input data set to reproduce the current object state.

I'm not sure if this is accurate though; I'm thinking of the case where a user may have manually, e.g., changed a species thermo entry (for example, to debug a discontinuous polynomial) and _that_ data will be written out instead of the data that was originally used to create the `Solution()`. Is that correct?

speth: I had been using "auxiliary" to mean data that is not used by Cantera, so there is no auxiliary input data that isn't user-provided. I guess this would be more precise:
> Clear all saved input data, so that the data given by `input_data` or `write_yaml` will only include values generated by Cantera based on the current object state.

bryanwweber: That sounds great to me! Thanks!

ischoegl (commented):
Looks good to me.

bryanwweber (approved):
Thanks @speth!

PR #1035, opened 2021-05-10
Fix test_mass_flow_controller failures

### LINKED ISSUE #1033: Sporadic failures of CI in test_mass_flow_controller
**Problem description**

Recent CI runs show spurious failures of `test_reactor.TestIdealGasReactor.test_mass_flow_controller` in the Sundials suite. It appears that failures do not occur for every run: thus far, they most often occurred for Sundials 5.7, but at least one failure occurred for Sundials 3.

**Steps to reproduce**

E.g. see recent GH Actions failures in #1032

**Behavior**

Test fails with 
```
FAIL: test_mass_flow_controller (cantera.test.test_reactor.TestIdealGasReactor)
  File "/home/runner/work/cantera/cantera/build/python/cantera/test/test_reactor.py", line 478, in test_mass_flow_controller
    self.assertArrayNear(ma * Ya + 0.1 * gas2.Y, mb * Yb)
  File "/home/runner/work/cantera/cantera/build/python/cantera/test/utilities.py", line 82, in assertArrayNear
    self.fail(worst[1])
AssertionError: AssertNear: 0.1 - 0.10000000665769 = -6.6576915647154e-09
Relative error for element (0,) of 6.651040e-08 exceeds rtol = 1.000000e-08

```

**System information**

- Cantera version: latest `main` (19f1593476112279d2d8045ba09f193038335bc1)
- OS: GH Actions

speth:

- Disable multithreaded OpenBLAS in in tests of different Sundials versions

Closes #1033 

**Checklist**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -216,6 +216,8 @@ jobs:
   multiple-sundials:
     name: Sundials ${{ matrix.sundials-ver }}
     runs-on: ubuntu-latest
+    env:
+      OPENBLAS_NUM_THREADS: "1"
     defaults:
       run:
         shell: bash -l {0}
@@ -244,7 +246,7 @@ jobs:
         run: |
           scons build extra_inc_dirs=$CONDA_PREFIX/include:$CONDA_PREFIX/include/eigen3 \
           extra_lib_dirs=$CONDA_PREFIX/lib system_fmt=y system_eigen=y system_yamlcpp=y \
-          system_sundials=y blas_lapack_libs='lapack,blas' -j2 VERBOSE=True debug=n
+          system_sundials=y blas_lapack_libs='lapack,blas' -j2 VERBOSE=True debug=n env_vars=PATH,LD_LIBRARY_PATH,PYTHONPATH,OPENBLAS_NUM_THREADS
       - name: Test Cantera
         run: scons test
 


bryanwweber:
@speth The OpenBLAS we're getting here uses the `pthread` library. From the SUNDIALS 5.7 action:

```
openblas-0.3.15            |pthreads_h4748800_0         9.8 MB  conda-forge
```

I wonder if setting `OMP_NUM_THREADS` actually did anything in that case?

speth:
I had no idea that was a thing. This game of whack-a-mole is getting a little ridiculous.

bryanwweber:
Resetting this to OpenBLAS 0.3.13 seems to have fixed the error (the SUNDIALS 2 failure is due to a dependency resolution failure).

speth:
OK, I forced the tests to run again, and I believe you that it's something to do with the change in the OpenBLAS version.

bryanwweber:
I can't tell from the changelogs what would have affected this test: https://github.com/xianyi/OpenBLAS/releases

ischoegl (commented):
Excellent, but I’ll leave it to @bryanwweber to review & merge.

.github/workflows/main.yml
```diff
@@ -244,7 +246,7 @@ jobs:
         run: |
           scons build extra_inc_dirs=$CONDA_PREFIX/include:$CONDA_PREFIX/include/eigen3 \
           extra_lib_dirs=$CONDA_PREFIX/lib system_fmt=y system_eigen=y system_yamlcpp=y \
-          system_sundials=y blas_lapack_libs='lapack,blas' -j2 VERBOSE=True debug=n
+          system_sundials=y blas_lapack_libs='lapack,blas' -j2 VERBOSE=True debug=n env_vars=PATH,LD_LIBRARY_PATH,PYTHONPATH,OPENBLAS_NUM_THREADS
```
bryanwweber, line 249: Is it necessary to include all these `env_vars` in the passthrough? If yes, can they go on their own line?

bryanwweber (changes requested):
This is simple enough, just one question.

.github/workflows/main.yml
speth: That's the default list (which would be superseded by any setting for `env_vars`), plus the one new one, so this is the minimal perturbation to the build environment. I haven't really given consideration to whether any of the others could be left off for this CI job.

bryanwweber: Would it make more sense to add `*_NUM_THREADS` to the environment variable passthrough in SConstruct? I foresee this as being something that would be useful in other testing scenarios, even locally.

ischoegl: I believe having more thorough tests for multi-threaded operation is an interesting thought. Thus far, it was my impression that multi-threading worked for reactor networks automatically and reliably, which this episode somehow puts in question. That said, I believe the fix for CI is ok as it stands, but it may make sense to address multi-threading in a more strategic fashion in a follow up feature PR.

speth: the `env_vars` option doesn't support wildcards, so unless you mean listing all of half dozen or more options, not really. There is the option of passing through all environment variables (`env_vars=all`), but that kind of goes against the SCons philosophy of not passing this information through to keep the build repeatable.

I don't think it would be productive to do any specific testing of multithreaded code that is two layers abstracted from Cantera (with SUNDIALS sitting in the middle, at least in this case), and where we have basically no control over what library the user may ultimately have linked to. I have trouble imagining a scenario in which a problem that arises in such a scenario would be something that could be addressed within Cantera.

bryanwweber: > the `env_vars` option doesn't support wildcards

We already do some processing of the `env_vars` option, though... We could do something like:

```python
import re
import os
r = re.compile(".*_NUM_THREADS")
num_threads = list(filter(r.match, os.environ.keys()))
for n in num_threads:
    # add the options to env_vars
```

speth: You want to include these variables all the time, not just when they've been specified as part of `env_vars`? If you want to open a different PR, be my guest.

speth:
Apparently, based on #1039, fiddling around with multithreading isn't actually a reliable solution to this problem.

Issue #390, opened 2016-11-15
Improve handling of un-normalized mass/mole fractions


speth:
When setting composition using un-normalized mass fractions, the mole fractions are still normalized, and vice versa. This is problematic for the normal use case of un-normalized mass fractions of calculating Jacobians because now perturbing the mass fraction for species *k* perturbs the mole fractions for all species, resulting in an incorrect Jacobian. For instance:
```py
>>> import cantera as ct
>>> gas = ct.Solution('gri30.xml')
>>> Y = np.zeros(53)
>>> Y[0] = 0.4
>>> Y[5] = 0.6
>>> gas.set_unnormalized_mass_fractions(Y)
>>> print(gas.mole_fraction_dict())
{'H2': 0.8562761715092376, 'H2O': 0.14372382849076246}
>>> Y[5] = 0.61
>>> gas.set_unnormalized_mass_fractions(Y)
>>> print(gas.mole_fraction_dict())
{'H2': 0.8542299515251598, 'H2O': 0.14577004847484037}
```

Instead, what we would like is for only the mole fraction of species *k* to change.

@decaluwe has suggested the following algorithm:

> If I set Y_k to something where sum(Y_k) =/= 1, then X_k should be calculated as Rho*Y_k/(C*W_k), where Rho is the mass density (kg/m3), C is the molar density (kmol/m3), and W_k is the molar mass of species k (kg/kmol). A similar equation could be derived for the opposite (Y_k based on a non-normalized X_k).

Originally discussed here: https://groups.google.com/forum/#!topic/cantera-users/qEgtxkn9P6o

ischoegl:
I had a look at this, and it appears somewhat murky (unless I am overlooking something). In the current implementation, C uses $\rho/W_{mean}$, where $W_{mean}$ (aka `m_mmw`) implicitly assumes that the mass fractions sum up to unity, i.e.
```C++
 m_mmw = dot(x, x + m_kk, m_molwts.begin());
``` 
in `setMoleFractions_NoNorm` (or the less straight-forward version in `setMassFractions_NoNorm`)

The suggested solution would require that C is known, with $C=\rho/W_{mean}$. With density being straight forward, this still requires knowledge of $W_{mean}$, which is impossible to calculate if constituents are partially unknown (i.e. it's a circular problem). 

I do not see an obvious way around assuming that fractions sum up sufficiently _close to_ unity. For analytical Jacobians, there is obviously no issue.

[...]

PR #1037, opened 2021-05-10
SolutionArray/ flamebase importers allow unnormalized and/or negative mass/mole fractions

### LINKED ISSUE #1016: SolutionArray truncates negative values
For `SolutionArray`/`FlameBase`, importers are not currently creating the same states. Negative values are truncated to zero which makes it impossible to have the same states initially stored in the `SolutionArray` or `FlameBase` object. 

Quick example:

Operating system: Windows 10
Python Version: 3.9.2
Cantera version: 2.6.0a1
```
In [1]: import cantera as ct
   ...: gas.set_unnormalized_mole_fractions(gas.X - 1e-16)
   ...: arr = ct.SolutionArray(gas, 5)
   
In [2]: arr.X
Out[2]: 
array([[ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16],
       [ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16],
       [ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16],
       [ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16],
       [ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16]])
In [3]: arr.write_hdf('test.h5') # save to HDF
Out[3]: 'group0'
   
In [4]: gas_new = ct.Solution('h2o2.yaml')
   ...: arr_new = ct.SolutionArray(gas_new)
In [5]: arr_new.read_hdf('test.h5') # load from HDF
   ...: arr_new.X
Out[5]: 
array([[1., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [1., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [1., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [1., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [1., 0., 0., 0., 0., 0., 0., 0., 0., 0.]])
```
Also, `SolutionArray` does not have `set_unnormalized_mole_fractions()` or `set_unnormalized_mass_fractions()` methods like  a `Solution` object which I think should be applicable to `SolutionArray` too.

```
In [6]: arr_new.set_unnormalized_mole_fractions(arr.X - 1e-16)
<ipython-input-14-22b735859e09> in <module>
----> 1 arr_new.set_unnormalized_mole_fractions(arr.X - 1e-16)
~/.pyenv/versions/miniconda3-latest/envs/cantera-2.5.1/lib/python3.8/site-packages/cantera/composite.py in __getattr__(self, name)
    604             super().__getattr__(name)
    605         else:
--> 606             raise AttributeError("'{}' object has no attribute '{}'".format(
    607                 self.__class__.__name__, name))
    608 
AttributeError: 'SolutionArray' object has no attribute 'set_unnormalized_mole_fractions'
```

A proposed fix for the two issues raised above would be to implement `set_unnormalized_mass_fractions` and `set_unnormalized_mole_fractions` for `SolutionArray` objects.

Another related issue which I think would be fixed together with the above raise issues is

```
In [7]: states = ct.SolutionArray(gas)
   ...: states.append(T=600., P=ct.one_atm, X=[0.1, 0.3, 0.3, -0.4, 0.25, 0.15, -0.2, 0.1, 0, 0])
   ...: states.X            
Out[7]:
    array([[0.08333333, 0.25      , 0.25      , 0.        , 0.20833333,
            0.125     , 0.        , 0.08333333, 0.        , 0.        ]])
```

DavidAkinpelu:

-  `SolutionArray`/ `Flambase` importers allow negative values for mass and mole fractions
- `SolutionArray` appends `Solution` object with unnormalized mass/mole fractions

Closes #1016 

Operating system: Windows 10
Python Version: 3.9.2
Cantera version: 2.6.0a1
```
In [1]: import cantera as ct
   ...: gas.set_unnormalized_mole_fractions(gas.X - 1e-16)
   ...: arr = ct.SolutionArray(gas, 5)

In [2]: arr.X
Out[2]: 
array([[ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16],
       [ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16],
       [ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16],
       [ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16],
       [ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16]])
In [3]: arr.write_hdf('test.h5') # save to HDF
Out[3]: 'group0'

In [4]: gas_new = ct.Solution('h2o2.yaml')
   ...: arr_new = ct.SolutionArray(gas_new)
In [5]: arr_new.read_hdf('test.h5') # load from HDF
   ...: arr_new.X
Out[5]: 
array([[ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16],
       [ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16],
       [ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16],
       [ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16],
       [ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16]])

In[6]: gas1 = ct.Solution('h2o2.yaml')
   ...: states = ct.SolutionArray(gas1)    
   ...: states.append(T=300., P=ct.one_atm, X = gas.X -1.e-16, normalize = False)    
In[7]: states.X
Out[7]:
array([[ 1.e+00, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16, -1.e-16,
        -1.e-16, -1.e-16, -1.e-16]])
```

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -682,8 +682,8 @@ class SolutionArray:
                     "'{}' does not specify a full thermodynamic state".format(attr)
                 )
             if not normalize:
-                attr_nonorm = attr + "_NoNorm"
-                setattr(self._phase, attr_nonorm, value)
+                setattr(self._phase, attr[-1] + "_no_norm", value[-1])
+                setattr(self._phase, attr[:-1], value[:-1])
             else:
                 setattr(self._phase, attr, value)
         else:
@@ -695,11 +695,11 @@ class SolutionArray:
                     "the thermodynamic state".format(tuple(kwargs))
                 ) from None
             if not normalize:
-                attr_nonorm = attr + "_NoNorm"
-                setattr(self._phase, attr_nonorm, [kwargs[a] for a in attr])
+                setattr(self._phase, attr[-1] + "_no_norm", kwargs[attr[-1]])
+                setattr(self._phase, attr[:-1], [kwargs[a] for a in attr[:-1]])
             else:
                 setattr(self._phase, attr, [kwargs[a] for a in attr])
-        
+
         for name, value in self._extra.items():
             new = extra_temp[name]
             if len(value):
@@ -719,12 +719,12 @@ class SolutionArray:
             extra_temp[name] = np.array(v)
 
         for name, value in extra_temp.items():
-            self._extra[name] = value  
+            self._extra[name] = value
 
         self._states.append(self._phase.state)
         self._indices.append(len(self._indices))
         self._shape = (len(self._indices),)
-        
+
     @property
     def meta(self):
         """
@@ -910,14 +910,23 @@ class SolutionArray:
             self._output_dummy = self._indices
             self._shape = (rows,)
 
-        # use unnormalized state setters
-        if mode in ('TDY', 'TPX', "TPY"):
-            mode = mode + "_NoNorm"
+        # use unnormalized state setters for Three property setter only
+        if mode in {"TPX", "TPY", "HPX", "HPY", "SPX", "SPY",
+                    "TDX", "TDY", "TPX", "TPY", "UVX", "UVY",
+                    "DPX", "DPY","HPX", "HPY", "SPX", "SPY",
+                    "SVX", "SVY"}:
 
-        # restore data
-        for i in self._indices:
-            setattr(self._phase, mode, [st[i, ...] for st in state_data])
-            self._states[i] = self._phase.state
+                for i in self._indices:
+                    setattr(self._phase, mode[-1] + "_no_norm", [st[i, ...] for st in state_data][2])
+                    setattr(self._phase, mode[:-1], [st[i, ...] for st in \
+                            state_data[:-1]])
+                    self._states[i] = self._phase.state
+
+        else:
+             # restore data
+            for i in self._indices:
+                setattr(self._phase, mode, [st[i, ...] for st in state_data])
+                self._states[i] = self._phase.state
 
         self._extra = extra_lists
 
@@ -937,39 +946,6 @@ class SolutionArray:
             self._phase.state = self._states[index]
             self._phase.set_equivalence_ratio(phi[index], *args, **kwargs)
             self._states[index][:] = self._phase.state
-
-    def set_unnormalized_mass_fractions(self, Y, *args, **kwargs):
-        """
-        See `ThermoPhase.set_unnormalized_mass_fractions
-
-        Note that *Y* needs to be a list with length matching the number of species or 
-        multidimensional list with dimension matching the SolutionArray and the length 
-        of the inner lists matching the number of species.
-        """
-
-        Y = np.tile(Y, (len(self._indices), 1))
-
-        for index in self._indices:
-            self._phase.state = self._states[index]
-            self._phase.set_unnormalized_mass_fractions(Y[index])
-            self._states[index][:] = self._phase.state
-
-    def set_unnormalized_mole_fractions(self, X, *args, **kwargs):
-        """
-        See `ThermoPhase.set_unnormalized_mole_fractions
-
-        Note that *X* needs to be a list with length matching the number of species or 
-        multidimensional list with dimension matching the SolutionArray and the length 
-        of the inner lists matching the number of species.
-        """
-
-        X = np.tile(X, (len(self._indices), 1))
-
-        for index in self._indices:
-            self._phase.state = self._states[index]
-            self._phase.set_unnormalized_mole_fractions(X[index])
-            self._states[index][:] = self._phase.state
-
 
     def collect_data(self, cols=None, tabular=False, threshold=0, species=None):
         """
@@ -1417,7 +1393,7 @@ def _state3_prop(name, doc_source, scalar=False):
                 self._phase.state = self._states[index]
                 setattr(self._phase, name, (A[index], B[index], C[index]))
                 self._states[index][:] = self._phase.state
-    
+
     return getter, setter
 
 def _make_functions():
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -628,7 +628,7 @@ class SolutionArray:
         return SolutionArray(self._phase[species], states=self._states,
                              extra=self._extra)
 
-    def append(self, state=None, **kwargs):
+    def append(self, state=None, normalize=True, **kwargs):
         """
         Append an element to the array with the specified state. Elements can
         only be appended in cases where the array of states is one-dimensional.
@@ -681,8 +681,11 @@ class SolutionArray:
                 raise KeyError(
                     "'{}' does not specify a full thermodynamic state".format(attr)
                 )
-            setattr(self._phase, attr, value)
-
+            if not normalize:
+                attr_nonorm = attr + "_NoNorm"
+                setattr(self._phase, attr_nonorm, value)
+            else:
+                setattr(self._phase, attr, value)
         else:
             try:
                 attr = self._phase._full_states[frozenset(kwargs)]
@@ -691,8 +694,12 @@ class SolutionArray:
                     "{} is not a valid combination of properties for setting "
                     "the thermodynamic state".format(tuple(kwargs))
                 ) from None
-            setattr(self._phase, attr, [kwargs[a] for a in attr])
-
+            if not normalize:
+                attr_nonorm = attr + "_NoNorm"
+                setattr(self._phase, attr_nonorm, [kwargs[a] for a in attr])
+            else:
+                setattr(self._phase, attr, [kwargs[a] for a in attr])
+        
         for name, value in self._extra.items():
             new = extra_temp[name]
             if len(value):
@@ -712,12 +719,12 @@ class SolutionArray:
             extra_temp[name] = np.array(v)
 
         for name, value in extra_temp.items():
-            self._extra[name] = value
+            self._extra[name] = value  
 
         self._states.append(self._phase.state)
         self._indices.append(len(self._indices))
         self._shape = (len(self._indices),)
-
+        
     @property
     def meta(self):
         """
@@ -903,10 +910,15 @@ class SolutionArray:
             self._output_dummy = self._indices
             self._shape = (rows,)
 
+        # use unnormalized state setters
+        if mode in ('TDY', 'TPX', "TPY"):
+            mode = mode + "_NoNorm"
+
         # restore data
         for i in self._indices:
             setattr(self._phase, mode, [st[i, ...] for st in state_data])
             self._states[i] = self._phase.state
+
         self._extra = extra_lists
 
     def set_equivalence_ratio(self, phi, *args, **kwargs):
@@ -925,6 +937,39 @@ class SolutionArray:
             self._phase.state = self._states[index]
             self._phase.set_equivalence_ratio(phi[index], *args, **kwargs)
             self._states[index][:] = self._phase.state
+
+    def set_unnormalized_mass_fractions(self, Y, *args, **kwargs):
+        """
+        See `ThermoPhase.set_unnormalized_mass_fractions
+
+        Note that *Y* needs to be a list with length matching the number of species or 
+        multidimensional list with dimension matching the SolutionArray and the length 
+        of the inner lists matching the number of species.
+        """
+
+        Y = np.tile(Y, (len(self._indices), 1))
+
+        for index in self._indices:
+            self._phase.state = self._states[index]
+            self._phase.set_unnormalized_mass_fractions(Y[index])
+            self._states[index][:] = self._phase.state
+
+    def set_unnormalized_mole_fractions(self, X, *args, **kwargs):
+        """
+        See `ThermoPhase.set_unnormalized_mole_fractions
+
+        Note that *X* needs to be a list with length matching the number of species or 
+        multidimensional list with dimension matching the SolutionArray and the length 
+        of the inner lists matching the number of species.
+        """
+
+        X = np.tile(X, (len(self._indices), 1))
+
+        for index in self._indices:
+            self._phase.state = self._states[index]
+            self._phase.set_unnormalized_mole_fractions(X[index])
+            self._states[index][:] = self._phase.state
+
 
     def collect_data(self, cols=None, tabular=False, threshold=0, species=None):
         """
@@ -1372,7 +1417,7 @@ def _state3_prop(name, doc_source, scalar=False):
                 self._phase.state = self._states[index]
                 setattr(self._phase, name, (A[index], B[index], C[index]))
                 self._states[index][:] = self._phase.state
-
+    
     return getter, setter
 
 def _make_functions():
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -650,10 +650,11 @@ class SolutionArray:
 
               mystates.append(T=300, P=101325, X={'O2':1.0, 'N2':3.76})
 
-        By default, setters with `moleFractions` are normalized to sum to 1.0 in all
-        cases. while element appended using ``state`` or the setter contains
-        `massFractions` specified as an array, are non-normalized. If this is
-        not desired, the ``normalize`` argument can be set to ``True``.
+        By default, the mass or mole fractions will be non-normalized
+        if appending a `state` or numpy array specifying `X` or `Y`.
+        When appending a state, the mass or mole fractions are always
+        non-normalized. For states setters, if the non-normalized
+        form is desired, the ``normalize`` argument can be set to ``True``.
         For example::
 
             mystates.append(T=300, P=101325, Y=gas.Y - 0.1, normalize=True)
@@ -681,48 +682,30 @@ class SolutionArray:
             extra_temp[name] = kwargs.pop(name)
 
         if state is not None:
-            mode = "".join(self._phase._native_state)
-            if normalize:
-                if len(mode) == 3:
-                    state_data = (state[0], state[1], state[2:])
-                elif len(mode) == 2:
-                    state_data = (state[0], state[1:])
-                else:
-                    state_data = state
-                setattr(self._phase, mode, state_data)
-            else:
-                self._phase.state = state
+            self._phase.state = state
 
-        # Non normalize form must be set using the mass fractions as the
-        # mole fractions are not in the native_state setter.
         elif len(kwargs) == 1:
             attr, value = kwargs.popitem()
             if frozenset(attr) not in self._phase._full_states:
                 raise KeyError(
                     "'{}' does not specify a full thermodynamic state".format(attr)
                 )
-            if attr[-1] == "Y":
-                if normalize:
+            if attr[-1] == "X":
+                if not normalize and isinstance(value[-1], np.ndarray):
+                    self._phase.set_unnormalized_mole_fractions(value[-1])
+                    attr = attr[:-1]
+                    value = value[:-1]
                     setattr(self._phase, attr, value)
-                elif normalize is None:
-                    if isinstance(value[-1], (list, tuple, np.ndarray)):
-                        self._phase.set_unnormalized_mass_fractions(value[-1])
-                        attr = attr[:-1]
-                        value = value[:-1]
-                        setattr(self._phase, attr, value)
-                    else:
-                        setattr(self._phase, attr, list(kwargs.values()))
                 else:
-                    if isinstance(value[-1], (list, tuple, np.ndarray)):
-                        self._phase.set_unnormalized_mass_fractions(value[-1])
-                        attr = attr[:-1]
-                        value = value[:-1]
-                        setattr(self._phase, attr, value)
-                    else:
-                        raise ValueError(
-                           "Mass fraction must be specified as an array if "
-                           "`normalize` is True"
-                        )
+                    setattr(self._phase, attr, value)
+            elif attr[-1] == "Y":
+                if not normalize and isinstance(value[-1], np.ndarray):
+                    self._phase.set_unnormalized_mass_fractions(value[-1])
+                    attr = attr[:-1]
+                    value = value[:-1]
+                    setattr(self._phase, attr, value)
+                else:
+                    setattr(self._phase, attr, value)
             else:
                 setattr(self._phase, attr, value)
 
@@ -734,26 +717,20 @@ class SolutionArray:
                     "{} is not a valid combination of properties for setting "
                     "the thermodynamic state".format(tuple(kwargs))
                 ) from None
-            if attr[-1] == "Y":
-                if normalize:
+            if attr[-1] == "X":
+                if not normalize and isinstance(kwargs["X"], np.ndarray):
+                    self._phase.set_unnormalized_mole_fractions(kwargs.pop("X"))
+                    attr = attr[:-1]
                     setattr(self._phase, attr, list(kwargs.values()))
-                elif normalize is None:
-                    if isinstance(kwargs["Y"], (list, tuple, np.ndarray)):
-                        self._phase.set_unnormalized_mass_fractions(kwargs.pop("Y"))
-                        attr = attr[:-1]
-                        setattr(self._phase, attr, list(kwargs.values()))
-                    else:
-                        setattr(self._phase, attr, list(kwargs.values()))
                 else:
-                    if isinstance(kwargs["Y"], (list, tuple, np.ndarray)):
-                        self._phase.set_unnormalized_mass_fractions(kwargs.pop("Y"))
-                        attr = attr[:-1]
-                        setattr(self._phase, attr, list(kwargs.values()))
-                    else:
-                        raise ValueError(
-                           "Mass fraction must be specified as an array if "
-                           "`normalize` is True"
-                        )
+                    setattr(self._phase, attr, list(kwargs.values()))
+            elif attr[-1] == "Y":
+                if not normalize and isinstance(kwargs["Y"], np.ndarray):
+                    self._phase.set_unnormalized_mass_fractions(kwargs.pop("Y"))
+                    attr = attr[:-1]
+                    setattr(self._phase, attr, list(kwargs.values()))
+                else:
+                    setattr(self._phase, attr, list(kwargs.values()))
             else:
                 setattr(self._phase, attr, list(kwargs.values()))
 
@@ -832,9 +809,9 @@ class SolutionArray:
         does not contain those entries, an error is raised.
 
         If the data contains mass or mole fractions they will be set
-        without normalizing their sum to 1.0 by default. If this is not desired,
-        the ``normalize`` argument can be set to ``True`` to force the data
-        to sum to 1.0.
+        without normalizing their sum to 1.0 by default. If this is
+        not desired, the ``normalize`` argument can be set to ``True``
+        to force the data to sum to 1.0.
         """
 
         # check arguments
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -628,7 +628,7 @@ class SolutionArray:
         return SolutionArray(self._phase[species], states=self._states,
                              extra=self._extra)
 
-    def append(self, state=None, normalize=True, **kwargs):
+    def append(self, state=None, normalize=None, **kwargs):
         """
         Append an element to the array with the specified state. Elements can
         only be appended in cases where the array of states is one-dimensional.
@@ -650,14 +650,13 @@ class SolutionArray:
 
               mystates.append(T=300, P=101325, X={'O2':1.0, 'N2':3.76})
 
-        By default, the mass or mole fractions will be normalized so they sum to 1.0.
-        If this is not desired, the ``normalize`` argument can be set to ``False``.
-        In this case, a tuple of three elements that corresponds to the
-        full state setters of the `Solution` must be in terms of the mass fraction
-        e.g. `TPY` or `HPY` and the mass fractions must be specified as an array.
+        By default, setters with `moleFractions` are normalized to sum to 1.0 in all
+        cases. while element appended using ``state`` or the setter contains
+        `massFractions` specified as an array, are non-normalized. If this is
+        not desired, the ``normalize`` argument can be set to ``True``.
         For example::
 
-            mystates.append(T=300, P=101325, Y=gas.Y - 0.1, normalize=False)
+            mystates.append(T=300, P=101325, Y=gas.Y - 0.1, normalize=True)
         """
         if len(self._shape) != 1:
             raise IndexError("Can only append to 1D SolutionArray")
@@ -683,7 +682,7 @@ class SolutionArray:
 
         if state is not None:
             mode = "".join(self._phase._native_state)
-            if normalize or mode[-1] == "Q":
+            if normalize:
                 if len(mode) == 3:
                     state_data = (state[0], state[1], state[2:])
                 elif len(mode) == 2:
@@ -702,18 +701,30 @@ class SolutionArray:
                 raise KeyError(
                     "'{}' does not specify a full thermodynamic state".format(attr)
                 )
-            if normalize or attr[-1] == "Q":
-                setattr(self._phase, attr, value)
-            else:
-                if attr[-1] == "Y":
-                    self._phase.set_unnormalized_mass_fractions(value[-1])
-                    attr = attr[:-1]
-                    value = value[:-1]
+            if attr[-1] == "Y":
+                if normalize:
                     setattr(self._phase, attr, value)
+                elif normalize is None:
+                    if isinstance(value[-1], (list, tuple, np.ndarray)):
+                        self._phase.set_unnormalized_mass_fractions(value[-1])
+                        attr = attr[:-1]
+                        value = value[:-1]
+                        setattr(self._phase, attr, value)
+                    else:
+                        setattr(self._phase, attr, list(kwargs.values()))
                 else:
-                    raise KeyError(
-                       "Non-normalize only works with mass fractions"
-                    )
+                    if isinstance(value[-1], (list, tuple, np.ndarray)):
+                        self._phase.set_unnormalized_mass_fractions(value[-1])
+                        attr = attr[:-1]
+                        value = value[:-1]
+                        setattr(self._phase, attr, value)
+                    else:
+                        raise ValueError(
+                           "Mass fraction must be specified as an array if "
+                           "`normalize` is True"
+                        )
+            else:
+                setattr(self._phase, attr, value)
 
         else:
             try:
@@ -723,17 +734,28 @@ class SolutionArray:
                     "{} is not a valid combination of properties for setting "
                     "the thermodynamic state".format(tuple(kwargs))
                 ) from None
-            if normalize or attr[-1] == "Q":
-                setattr(self._phase, attr, list(kwargs.values()))
-            else:
-                if attr[-1] == "Y":
-                    self._phase.set_unnormalized_mass_fractions(kwargs.pop("Y"))
-                    attr = attr[:-1]
+            if attr[-1] == "Y":
+                if normalize:
                     setattr(self._phase, attr, list(kwargs.values()))
+                elif normalize is None:
+                    if isinstance(kwargs["Y"], (list, tuple, np.ndarray)):
+                        self._phase.set_unnormalized_mass_fractions(kwargs.pop("Y"))
+                        attr = attr[:-1]
+                        setattr(self._phase, attr, list(kwargs.values()))
+                    else:
+                        setattr(self._phase, attr, list(kwargs.values()))
                 else:
-                    raise KeyError(
-                       "Non-normalize only works with mass fractions"
-                    )
+                    if isinstance(kwargs["Y"], (list, tuple, np.ndarray)):
+                        self._phase.set_unnormalized_mass_fractions(kwargs.pop("Y"))
+                        attr = attr[:-1]
+                        setattr(self._phase, attr, list(kwargs.values()))
+                    else:
+                        raise ValueError(
+                           "Mass fraction must be specified as an array if "
+                           "`normalize` is True"
+                        )
+            else:
+                setattr(self._phase, attr, list(kwargs.values()))
 
         for name, value in self._extra.items():
             new = extra_temp[name]
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -652,7 +652,8 @@ class SolutionArray:
 
         By default, the mass or mole fractions will be normalized so they sum to 1.0.
         If this is not desired, the ``normalize`` argument can be set to ``False``.
-        In this case, the mass or mole fractions must be specified as an array::
+        In this case, the mass or mole fractions must be specified as an array.
+        For example::
 
             mystates.append(T=300, P=101325, X=gas.X - 1e-16, normalize=False)
         """
@@ -689,7 +690,7 @@ class SolutionArray:
                 )
             if normalize or attr[-1] == "Q":
                 setattr(self._phase, attr, value)
-            if not normalize:
+            elif not normalize:
                 if attr[-1] == "X":
                     self._phase.set_unnormalized_mole_fractions(value[-1])
                     attr = attr[:-1]
@@ -710,7 +711,7 @@ class SolutionArray:
                 ) from None
             if normalize or attr[-1] == "Q":
                 setattr(self._phase, attr, list(kwargs.values()))
-            if not normalize:
+            elif not normalize:
                 if attr[-1] == "X":
                     self._phase.set_unnormalized_mole_fractions(kwargs.pop("X"))
                     attr = attr[:-1]
@@ -939,7 +940,7 @@ class SolutionArray:
             for i in self._indices:
                 setattr(self._phase, mode, [st[i, ...] for st in state_data])
                 self._states[i] = self._phase.state
-        if not normalize:
+        elif not normalize:
             for i in self._indices:
                 if mode[-1] == "X":
                     self._phase.set_unnormalized_mole_fractions([st[i, ...] for \
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -368,12 +368,12 @@ class SolutionArray:
         >>> states.equilibrate('HP')
         >>> states.T # -> adiabatic flame temperature at various equivalence ratios
 
-    To append a state with unnormalized and/or negative mass/mole fractions, the mole/mass fractions
+    To append a state with non-normalized and/or negative mass/mole fractions, the mole/mass fractions
     must be specified as an array::
 
         >>> states = ct.SolutionArray(gas)
         >>> states.append(T=300., P=ct.one_atm, X = gas.X -1.e-16, normalize = False)
-        >>> states.X # -> unnormalized mole fractions
+        >>> states.X # -> non-normalized mole fractions
 
     `SolutionArray` objects can also be 'sliced' like Numpy arrays, which can be
     used both for accessing and setting properties::
@@ -657,7 +657,7 @@ class SolutionArray:
 
               mystates.append(T=300, P=101325, X={'O2':1.0, 'N2':3.76})
 
-        To specify unnormalize mass/mole fractions, the normalize argument should
+        To specify non-normalized mass/mole fractions, the normalize argument should
         be set to false. This is only applicable when the mole/mass fractions are
         specified as an array::
 
@@ -923,7 +923,7 @@ class SolutionArray:
             self._output_dummy = self._indices
             self._shape = (rows,)
 
-        # use unnormalized state setters for Three property setter only
+        # use non-normalized state setters for Three property setter only
         if mode in {"TPX", "TPY", "HPX", "HPY", "SPX", "SPY",
                     "TDX", "TDY", "TPX", "TPY", "UVX", "UVY",
                     "DPX", "DPY","HPX", "HPY", "SPX", "SPY",
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -652,10 +652,12 @@ class SolutionArray:
 
         By default, the mass or mole fractions will be normalized so they sum to 1.0.
         If this is not desired, the ``normalize`` argument can be set to ``False``.
-        In this case, the mass or mole fractions must be specified as an array.
+        In this case, a tuple of three elements that corresponds to the
+        full state setters of the `Solution` must be in terms of the mass fraction
+        e.g. `TPY` or `HPY` and the mass fractions must be specified as an array.
         For example::
 
-            mystates.append(T=300, P=101325, X=gas.X - 1e-16, normalize=False)
+            mystates.append(T=300, P=101325, Y=gas.Y - 0.1, normalize=False)
         """
         if len(self._shape) != 1:
             raise IndexError("Can only append to 1D SolutionArray")
@@ -682,6 +684,8 @@ class SolutionArray:
         if state is not None:
             self._phase.state = state
 
+        # Non normalize form must be set using the mass fractions as the
+        # mole fractions are not in the native_state setter.
         elif len(kwargs) == 1:
             attr, value = kwargs.popitem()
             if frozenset(attr) not in self._phase._full_states:
@@ -690,16 +694,16 @@ class SolutionArray:
                 )
             if normalize or attr[-1] == "Q":
                 setattr(self._phase, attr, value)
-            elif not normalize:
-                if attr[-1] == "X":
-                    self._phase.set_unnormalized_mole_fractions(value[-1])
-                    attr = attr[:-1]
-                    value = value[:-1]
-                elif attr[-1] == "Y":
+            else:
+                if attr[-1] == "Y":
                     self._phase.set_unnormalized_mass_fractions(value[-1])
                     attr = attr[:-1]
                     value = value[:-1]
-                setattr(self._phase, attr, value)
+                    setattr(self._phase, attr, value)
+                else:
+                    raise KeyError(
+                       "Non-normalize only works with mass fractions"
+                    )
 
         else:
             try:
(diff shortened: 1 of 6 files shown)


ischoegl:
Thanks, @DavidAkinpelu. As I am involved with this, I’ll leave it to @speth and/or @bryanwweber to review.

interfaces/cython/cantera/composite.py
```diff
@@ -368,6 +368,13 @@ class SolutionArray:
         >>> states.equilibrate('HP')
         >>> states.T # -> adiabatic flame temperature at various equivalence ratios
 
+    To append a state with non-normalized and/or negative mass/mole fractions, the mole/mass fractions
+    must be specified as an array::
+
+        >>> states = ct.SolutionArray(gas)
+        >>> states.append(T=300., P=ct.one_atm, X = gas.X -1.e-16, normalize = False)
+        >>> states.X # -> non-normalized mole fractions
+
```
bryanwweber: Since this is probably not going to be an option that people use frequently, I think the documentation associated with the `append()` method is enough, and you can remove this.

interfaces/cython/cantera/composite.py
```diff
@@ -649,6 +656,12 @@ def append(self, state=None, **kwargs):
           the full-state setters::
 
               mystates.append(T=300, P=101325, X={'O2':1.0, 'N2':3.76})
+
+        To specify non-normalized mass/mole fractions, the normalize argument should
+        be set to false. This is only applicable when the mole/mass fractions are
+        specified as an array::
+
+            mystates.append(T=300, P=101325, X=gas.X - 1e-16, normalize=False)
```
bryanwweber: I think this could use a little more explanation of the default case. Maybe something like:

> By default, the mass or mole fractions will be normalized so they sum to 1.0. If this is not desired, the ``normalize`` argument can be set to ``False``. In this case, the mass or mole fractions must be specified as an array.

interfaces/cython/cantera/composite.py
```diff
@@ -681,8 +694,11 @@ def append(self, state=None, **kwargs):
                 raise KeyError(
                     "'{}' does not specify a full thermodynamic state".format(attr)
                 )
-            setattr(self._phase, attr, value)
-
+            if not normalize:
+                setattr(self._phase, attr[-1] + "_no_norm", value[-1])
+                setattr(self._phase, attr[:-1], value[:-1])
+            else:
+                setattr(self._phase, attr, value)
```
bryanwweber: The logic here would be a little easier to understand if it was reversed, in my opinion. That's for 2 reasons: 1) `normalize == True` is the default case, so it should come first and 2) dealing with negation is always trickier in the conditionals. So, it could be:

```python
if normalize:
    ...
else:
    ...
```

In addition, and related to my other comments about `*_no_norm`, I think the non-normalized case is a little bit confusing because it is using string indexing and concatenation of the `attr`. So I wonder if something like:

```python
if normalize:
    ...
else:
    if attr[-1] == "X":
        setattr(self._phase, "no norm mole fractions", value[-1])
    elif attr[-1] == "Y":
        setattr(...)
    setattr(self._phase, attr[:2], value[:2])
```
To me, this is a little bit clearer because it explicitly lists the permitted values for the last character, and also explicitly says to take the first 2 characters to set the rest of the state. It makes it easier for the next developer to come along and see what is being done here and is maybe marginally slower to execute.

interfaces/cython/cantera/composite.py
```diff
@@ -681,8 +694,11 @@ def append(self, state=None, **kwargs):
                 raise KeyError(
```
bryanwweber, line 685: This isn't related to your changes, but I find the

```python
attr, value = next(iter(kwargs.items()))
```

a few lines up quite confusing. I suspect it was needed at some point for Python 2 compatibility, but there is a [`dict.popitem()`](https://docs.python.org/3/library/stdtypes.html#dict.popitem) method that returns the last item inserted. In this case, since the length is only one, it should be much simpler:

```python
attr, value = kwargs.popitem()
```

I hate to expand the scope of PRs, but I hope this change is simple enough... if it doesn't work immediately, don't worry about it.

interfaces/cython/cantera/composite.py
```diff
@@ -691,7 +707,11 @@ def append(self, state=None, **kwargs):
                     "{} is not a valid combination of properties for setting "
                     "the thermodynamic state".format(tuple(kwargs))
                 ) from None
-            setattr(self._phase, attr, [kwargs[a] for a in attr])
+            if not normalize:
+                setattr(self._phase, attr[-1] + "_no_norm", kwargs[attr[-1]])
+                setattr(self._phase, attr[:-1], [kwargs[a] for a in attr[:-1]])
+            else:
+                setattr(self._phase, attr, [kwargs[a] for a in attr])
```
bryanwweber: Likewise here about the reversed logic and the `attr` checking. Another potential simplification here, specifically, might be:

```python
if attr[-1] == "X":
    setattr(self._phase, "no norm mole fractions", kwargs.pop("X"))
elif attr[-1] == "Y":
    setattr(...)
setattr(self._phase, attr[:2], list(kwargs.values()))
```
and a similar change for the normalized case. I'm not sure if that'll pull the right values though. You might even be able to get away with not casting the `kwargs.values()` to a `list`, depending on how that ends up being used in the setters, but I'm not sure about that.

interfaces/cython/cantera/composite.py
```diff
@@ -903,10 +923,24 @@ def join(species):
...
 
-        # restore data
-        for i in self._indices:
-            setattr(self._phase, mode, [st[i, ...] for st in state_data])
-            self._states[i] = self._phase.state
+        # use non-normalized state setters for Three property setter only
+        if mode in {"TPX", "TPY", "HPX", "HPY", "SPX", "SPY",
+                    "TDX", "TDY", "TPX", "TPY", "UVX", "UVY",
+                    "DPX", "DPY","HPX", "HPY", "SPX", "SPY",
+                    "SVX", "SVY"}:
```
bryanwweber: This change seems like the default will be to do non-normalized restoring of states, is that correct? Assuming that is correct, I'm not sure I agree that should be the default, I'd be interested to hear why you're making this change.

I guess the expectation for users is that the data they provide is exactly used to set the state, which is fair enough. I wonder whether another keyword argument should be used to control this behavior?

Last, I don't think it's necessary to provide an exhaustive list of the three-property setters, it should be sufficient to check that `len(mode) == 3`, right?

interfaces/cython/cantera/thermo.pyx
```diff
@@ -678,6 +678,19 @@ cdef class ThermoPhase(_SolutionBase):
             else:
                 self._setArray1(thermo_setMassFractions, Y)
 
+    property Y_no_norm:
```
bryanwweber: As I said in another comment, I think you can be more explicit with the code and get around needing to add these two properties. I don't think that these will be used very much in general, and I don't particularly want to make it easier for people to set un-normalized mass or mole fractions. It is almost never what the user actually wants, so making it easier to access doesn't seem great to me. Finally, this is totally personal preference, but all of our single-letter setters are only available as the single letter (I'm pretty sure), so this would break that paradigm.

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -145,6 +145,33 @@ def test_collect_data(self):
...
+        self.assertEqual(states[0].T, gas.T)
+        self.assertEqual(states[0].P, gas.P)
+        self.assertArrayNear(states[0].X, gas.X)
+
+    def test_import_no_norm_data(self):
+        outfile = pjoin(self.test_work_dir, 'solutionarray.h5')
+        if os.path.exists(outfile):
+            os.remove(outfile)
+
+        gas = ct.Solution('h2o2.yaml')
```
bryanwweber: Please use double quotes in all new Python code. Thank you 😄

bryanwweber (changes requested):
Hi @DavidAkinpelu! Thank you for contributing this change, and welcome to contributing to Cantera! I've made a few comments below. I think overall the change is a great addition and it'll certainly make the `SolutionArray` more useful. The comments are mainly aimed at simplifying the code and hopefully making it a little easier for the next person to understand!

interfaces/cython/cantera/composite.py
DavidAkinpelu: From the documentation, the method `restore_data` is called internally to restored data saved to different file types( "csv", "h5" etc). We want the same exact state stored to be restored. That is why I think non-normalized should be the default.
There are some other three-property setters such as `TPQ`, `TDQ` that are not applicable in this context and that is what I am trying to catch. I agree with you that providing an exhaustive list of the three-property setters might not be the best approach. Doing this might be better ``` if len(mode) == 3 and "Q" not in mode```

bryanwweber: Yeah, I think it's a fair expectation on the part of the user to have exactly the input data restored. I think my nervous feeling is because this would be the only case that I can think of where a user action triggers non-normalized mole fraction setting by default. So in that sense it may be doing something a different user doesn't expect.

Whichever becomes the default here, it should definitely be possible to "restore" and have Cantera normalize mole fractions, for instance, if a set of initial conditions for a set of simulations is read from CSV, and the user specifies mole numbers instead of fractions.

DavidAkinpelu: The `restore_data` method is called internally on previously saved `SolutionArray` object from Cantera and not on some user-defined set of initial conditions or a random "csv" or "h5" file format. We want the `SolutionArray` object data to be reloaded in the same format it was saved initially that is why I thought making non-normalized form the default would be applicable here. It gives the user a lot of flexibility as the user can now decide to go with the non-normalized or normalized form depending on the application. i.e, the `restore_data` is used to restore previously saved simulation results and we want the result to be consistent with the initial result saved.

interfaces/cython/cantera/thermo.pyx
ischoegl: This setter was based on my suggestion, where the naming is loosely based on C++'s `setMassFractions_NoNorm`. The rationale here was to implement an efficient way to reach the C++ layer for the non-normalized setters that this PR seeks to implement.

Regarding single-letter (or multi-letter) uppercase setters (or getters), I wasn't aware of a paradigm (other than them not being pythonic, although the rationale is obvious). There isn't a `setMassFractions` setter in Python, so the suggested property is imho closest to what is already there. Regarding the concern of making this available to users, it would be simple to hide (via `_Y_no_norm` etc.), but at the same time there may be an argument to allow this for those users who need it and locate it in the documentation.

bryanwweber: > This setter was based on my suggestion, where the naming is loosely based on C++'s `setMassFractions_NoNorm`. The rationale here was to implement an efficient way to reach the C++ layer for the non-normalized setters that this PR seeks to implement.

The non-normalized setter is already implemented in Python, https://github.com/Cantera/cantera/blob/627e21e95e2dc534dabf9fb8fda1e4093177704e/interfaces/cython/cantera/thermo.pyx#L944-L970, which this implementation just calls. So I don't think these new properties are needed, and the existing functions can be used directly.

ischoegl: Good point - this was quite obviously hiding in plain sight.

interfaces/cython/cantera/composite.py
ischoegl: The signature of `restore_data` is currently
```Python
    def restore_data(self, data):
```
but it is also true that it is an internal function that is typically called by `read_csv`, `read_hdf` and `from_pandas`.

I believe it would make sense to add a keyword argument to `restore_data` with the default set to `normalize=False` (for reasons pointed out by @DavidAkinpelu). At least for `read_csv` and `from_pandas`, I believe that adding a similar keyword would allow a user to override the default behavior for cases as pointed out by @bryanwweber. Obviously those selections should be threaded through to `restore_data`.

interfaces/cython/cantera/composite.py
```diff
@@ -903,10 +923,24 @@ def join(species):
...
+
+                for i in self._indices:
+                    setattr(self._phase, mode[-1] + "_no_norm", [st[i, ...] for st in state_data][2])
+                    setattr(self._phase, mode[:-1], [st[i, ...] for st in \
+                            state_data[:-1]])
+                    self._states[i] = self._phase.state
+
+        else:
+             # restore data
```
ischoegl: Careful with indents - there's an extra whitespace here (before `# restore data`)

interfaces/cython/cantera/test/test_composite.py
ischoegl: @bryanwweber: On this note, it may make sense to update `CONTRIBUTING.md` ... there are also references to Python versions that are no longer relevant there. As I believe that you have a better overview of everything Python, I won't create a PR myself.

interfaces/cython/cantera/composite.py
```diff
@@ -657,9 +650,9 @@ def append(self, state=None, normalize=True, **kwargs):
 
               mystates.append(T=300, P=101325, X={'O2':1.0, 'N2':3.76})
 
-        To specify non-normalized mass/mole fractions, the normalize argument should
-        be set to false. This is only applicable when the mole/mass fractions are
-        specified as an array::
+        By default, the mass or mole fractions will be normalized so they sum to 1.0.
+        If this is not desired, the normalize argument can be set to False.
```
```suggestion
        If this is not desired, the ``normalize`` argument can be set to ``False``.
```

interfaces/cython/cantera/composite.py
```diff
@@ -689,16 +682,21 @@ def append(self, state=None, normalize=True, **kwargs):
...
                 raise KeyError(
                     "'{}' does not specify a full thermodynamic state".format(attr)
                 )
-            if not normalize:
-                setattr(self._phase, attr[-1] + "_no_norm", value[-1])
-                setattr(self._phase, attr[:-1], value[:-1])
-            else:
+            if normalize:
                 setattr(self._phase, attr, value)
+
```
```suggestion
```

interfaces/cython/cantera/composite.py
```diff
@@ -689,16 +682,21 @@ def append(self, state=None, normalize=True, **kwargs):
...
+            else:
+                if attr[-1] == "X":
+                    self._phase.set_unnormalized_mole_fractions(value[-1])
+                elif attr[-1] == "Y":
+                    self._phase.set_unnormalized_mass_fractions(value[-1])
+                setattr(self._phase, attr[:2], value[:2])
```
bryanwweber: Awesome, thank you for changing this!  I think the only remaining thing to consider is the `TPQ` setter, does that apply here? If it doesn't, you can just ignore the rest of this comment 😄 

I think there's a couple ways of approaching that, which would be to add a second `elif`, but I think `TPQ` needs to be set all at once to trigger the appropriate error checking. Maybe set the first condition to `if normalize or attr  == "TPQ"`? Another way would be more drastic relative to what's here, something like this could work:

```python
if not normalize:
    if attr[-1] == "X":
        self._phase.set_unnormalized_mole...
        attr = attr[:-1]
        value = value[:-1]
    elif attr[-1] == "Y":
        self._phase.set...
        attr = attr[:-1]
        value = value[:-1]
setattr(self._phase, attr, value)
```
This is just a suggestion, please don't take it as a requirement to change it just like this! I'll leave it up to you to do something to handle the TPQ case (if it's relevant).

interfaces/cython/cantera/composite.py
```diff
@@ -707,11 +705,14 @@ def append(self, state=None, normalize=True, **kwargs):
                     "{} is not a valid combination of properties for setting "
                     "the thermodynamic state".format(tuple(kwargs))
                 ) from None
-            if not normalize:
-                setattr(self._phase, attr[-1] + "_no_norm", kwargs[attr[-1]])
-                setattr(self._phase, attr[:-1], [kwargs[a] for a in attr[:-1]])
-            else:
+            if normalize:
                 setattr(self._phase, attr, [kwargs[a] for a in attr])
```
```suggestion
            if normalize:
                setattr(self._phase, attr, kwargs.values())
```

bryanwweber: If it works 😉 If it doesn't work, then casting it to a `list` (or maybe a `tuple`?) like below would be good.

interfaces/cython/cantera/composite.py
```diff
@@ -707,11 +705,14 @@ def append(self, state=None, normalize=True, **kwargs):
...
+            else:
+                if attr[-1] == "X":
```
bryanwweber: I think this is the same as my previous comment about `TPQ` and can probably be resolved in a similar way.

interfaces/cython/cantera/composite.py
```diff
@@ -786,6 +787,10 @@ def restore_data(self, data):
         properties is omitted. If the receiving `SolutionArray` has extra
         entries already specified, only those will be imported; if *labels*
         does not contain those entries, an error is raised.
+
+        By default, if the data contains mass or mole fractions,
+        they will be non-normalized. If this is not desired,
+        the normalize argument can be set to True, so they sum to 1.0.
```
```suggestion
        If the data contains mass or mole fractions they will be set
        without normalizing their sum to 1.0 by default. If this is not desired,
        the ``normalize`` argument can be set to ``True`` to force the data
        to sum to 1.0.
```

interfaces/cython/cantera/composite.py
```diff
@@ -1072,11 +1080,15 @@ def write_csv(self, filename, cols=None, *args, **kwargs):
...
-    def read_csv(self, filename):
+    def read_csv(self, filename, normalize=False):
         """
         Read a CSV file named *filename* and restore data to the `SolutionArray`
         using `restore_data`. This method allows for recreation of data
         previously exported by `write_csv`.
+
+        The normalize argument is passed on to `restore_data` to normalize
+        mole or mass fractions if desired. By default, the normalize
+        argument is set to False.
```
```suggestion
        The ``normalize`` argument is passed on to `restore_data` to normalize
        mole or mass fractions, if desired. By default, ``normalize`` is ``False``.
```

bryanwweber: And similar for a few cases below.

bryanwweber (changes requested):
Thanks @DavidAkinpelu! These changes are looking great. I have a few more comments, mostly to clarify wording and text formatting in a few places.

interfaces/cython/cantera/composite.py
DavidAkinpelu: I forgot to modify this line in plainsight

interfaces/cython/cantera/composite.py
```diff
@@ -676,12 +682,23 @@ def append(self, state=None, **kwargs):
...
-            attr, value = next(iter(kwargs.items()))
+            attr, value = kwargs.popitem()
             if frozenset(attr) not in self._phase._full_states:
                 raise KeyError(
                     "'{}' does not specify a full thermodynamic state".format(attr)
                 )
-            setattr(self._phase, attr, value)
+            if normalize or attr[-1] == "Q":
+                setattr(self._phase, attr, value)
+            if not normalize:
```
ischoegl: Should this be else if?

interfaces/cython/cantera/composite.py
```diff
@@ -904,9 +935,26 @@ def join(species):
...
 
         # restore data
-        for i in self._indices:
-            setattr(self._phase, mode, [st[i, ...] for st in state_data])
-            self._states[i] = self._phase.state
+        if normalize or mode[-1] == "Q":
+            for i in self._indices:
+                setattr(self._phase, mode, [st[i, ...] for st in state_data])
+                self._states[i] = self._phase.state
+        if not normalize:
```
ischoegl: Same here.

interfaces/cython/cantera/composite.py
```diff
@@ -649,6 +649,12 @@ def append(self, state=None, **kwargs):
           the full-state setters::
 
               mystates.append(T=300, P=101325, X={'O2':1.0, 'N2':3.76})
+
+        By default, the mass or mole fractions will be normalized so they sum to 1.0.
+        If this is not desired, the ``normalize`` argument can be set to ``False``.
+        In this case, the mass or mole fractions must be specified as an array::
```
ischoegl: Add: ‘, for example’

interfaces/cython/cantera/composite.py
```diff
@@ -691,7 +708,16 @@ def append(self, state=None, **kwargs):
                     "{} is not a valid combination of properties for setting "
                     "the thermodynamic state".format(tuple(kwargs))
                 ) from None
-            setattr(self._phase, attr, [kwargs[a] for a in attr])
+            if normalize or attr[-1] == "Q":
+                setattr(self._phase, attr, list(kwargs.values()))
+            if not normalize:
```
ischoegl: Same. 

PS: it may be good to cover those `Q` cases in unit tests

ischoegl (changes requested):
Minor things.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -699,6 +712,19 @@ cdef class ThermoPhase(_SolutionBase):
             else:
                 self._setArray1(thermo_setMoleFractions, X)
 
+    property X_no_norm:
```
ischoegl: Looks like these properties came back when rebasing.

ischoegl (dismissed):

interfaces/cython/cantera/composite.py
```diff
@@ -676,12 +683,23 @@ def append(self, state=None, **kwargs):
...
-            attr, value = next(iter(kwargs.items()))
+            attr, value = kwargs.popitem()
             if frozenset(attr) not in self._phase._full_states:
                 raise KeyError(
                     "'{}' does not specify a full thermodynamic state".format(attr)
                 )
-            setattr(self._phase, attr, value)
+            if normalize or attr[-1] == "Q":
+                setattr(self._phase, attr, value)
+            elif not normalize:
```
bryanwweber: I think this should just be `else`... What do you think @ischoegl, since you suggested `elif`?

interfaces/cython/cantera/composite.py
```diff
@@ -691,7 +709,16 @@ def append(self, state=None, **kwargs):
                     "{} is not a valid combination of properties for setting "
                     "the thermodynamic state".format(tuple(kwargs))
                 ) from None
-            setattr(self._phase, attr, [kwargs[a] for a in attr])
+            if normalize or attr[-1] == "Q":
+                setattr(self._phase, attr, list(kwargs.values()))
+            elif not normalize:
```
bryanwweber: Same here

interfaces/cython/cantera/composite.py
```diff
@@ -904,9 +936,26 @@ def join(species):
...
 
         # restore data
-        for i in self._indices:
-            setattr(self._phase, mode, [st[i, ...] for st in state_data])
-            self._states[i] = self._phase.state
+        if normalize or mode[-1] == "Q":
+            for i in self._indices:
+                setattr(self._phase, mode, [st[i, ...] for st in state_data])
+                self._states[i] = self._phase.state
+        elif not normalize:
```
bryanwweber: Same here

interfaces/cython/cantera/composite.py
```diff
@@ -904,9 +936,26 @@ def join(species):
...
+            for i in self._indices:
+                if mode[-1] == "X":
+                    self._phase.set_unnormalized_mole_fractions([st[i, ...] for \
+                                                                 st in state_data][2])
```
bryanwweber: I think PEP 8 specifically dis-recommends using backslash line continuation. I think you can format this like:

```python
                    self._phase.set_unnormalized_mole_fractions(
                        [st[i, ...] for st in state_data][2]
                    )
```

interfaces/cython/cantera/composite.py
```diff
@@ -904,9 +936,26 @@ def join(species):
...
+                    setattr(self._phase, mode[:2], [st[i, ...] for st in \
+                                                    state_data[:2]])
```
bryanwweber: I'd suggest

```python
                    setattr(self._phase, mode[:2],
                            [st[i, ...] for st in state_data[:2]])
```

interfaces/cython/cantera/composite.py
```diff
@@ -904,9 +936,26 @@ def join(species):
...
+                elif mode[-1] == "Y":
+                    self._phase.set_unnormalized_mass_fractions([st[i, ...] for \
+                                                                 st in state_data][2])
+                    setattr(self._phase, mode[:2], [st[i, ...] for st in \
+                                                    state_data[:2]])
```
bryanwweber: Same suggestions here to eliminate the backslash.

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -145,6 +145,53 @@ def test_collect_data(self):
         self.assertIn('X', collected)
         self.assertEqual(collected['X'].shape, (0, self.gas.n_species))
 
+    def test_append_no_norm_data(self):
+        gas = ct.Solution("h2o2.yaml")
+        gas.TP = 300, ct.one_atm
+        gas.X = gas.X - 1.e-16
+        states = ct.SolutionArray(gas)
+        states.append(T=gas.T, P=gas.P, X = gas.X, normalize = False)
```
bryanwweber: Small nitpick, there shouldn't be spaces around `=` in function calls.

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -145,6 +145,53 @@ def test_collect_data(self):
         self.assertIn('X', collected)
         self.assertEqual(collected['X'].shape, (0, self.gas.n_species))
 
+    def test_append_no_norm_data(self):
+        gas = ct.Solution("h2o2.yaml")
+        gas.TP = 300, ct.one_atm
+        gas.X = gas.X - 1.e-16
```
bryanwweber: By default, this will normalize the mole fractions and you'll end up with a sum equal to 1.0 still. Then, setting `normalize=False` in the `append()` below doesn't actually do anything.

As another note, changing the mole fractions by 1E-16 here and elsewhere is not a big enough change. 1E-16 is approximately at the precision of a double floating point number, so this change may or may not actually be changing the numbers in the array. It'd be a more stringent test to set all the un-normalized mole fractions to 0.1 or something, since the phase has more than 10 species. That way you're sure that everything is working properly. Apologies that I didn't notice this before...

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -145,6 +145,53 @@ def test_collect_data(self):
...
+        states = ct.SolutionArray(gas)
+        states.append(T=gas.T, P=gas.P, X = gas.X, normalize = False)
+        self.assertEqual(states[0].T, gas.T)
+        self.assertEqual(states[0].P, gas.P)
+        self.assertArrayNear(states[0].X, gas.X)
+
+        w = ct.Water()
```
bryanwweber: This should be a new test, IMO, to limit the number of assertions per test. It's usually best to keep the scope of the test to one class or function or behavior.

bryanwweber (dismissed):
Thanks @DavidAkinpelu! This is very close, great work so far.

interfaces/cython/cantera/test/test_composite.py
bryanwweber: Actually, I think setting a value of 0.3 (or something else that won't sum to 1.0) for all the mole fractions would be more robust.

bryanwweber:
Sorry, two more comments:

1. It'd be good to add a test for the `normalize=True` case of `restore_data()` (although I guess that may be covered by other tests?)
2. I'm still conflicted about the choice of `normalize=False` as the default value for `restore_data`, but especially for `read_csv` and `read_hdf`. The reasons I'm conflicted are: 1) it'd be the only setter (I think) that works this way by default, which increases the difficulty of using Cantera slightly; 2) it's a breaking change in behavior; but on the other hand 3) users who save data out from a `SolutionArray` would expect to restore exactly that data back if the use a function called `restore_data`. Maybe a compromise is to set the default to `True` for the two`read_*` functions, and leave it `False` for `restore_data`? In my head, that resolves some of the conflict in terms of user expectations based on the function names, but then you get two different default values for very similar functions.

Since I really can't decide, I'm going to ping @speth for a third opinion.

interfaces/cython/cantera/composite.py
ischoegl: Yup. While `elif` works, the cases are mutually exclusive, so you’re correct; my main objection was against the `if` that was here before as it was possible to enter both clauses.

ischoegl:
> Maybe a compromise is to set the default to `True` for the two `read_*` functions, and leave it `False` for `restore_data`?

I’d be ok with that compromise. For reading from file, not getting back what was saved was the starting point for this PR, and it certainly should not be normalized in that case by default. Normalize is only enforced by default when composition is explicitly set; otherwise Cantera allows for deviations.

interfaces/cython/cantera/composite.py
```diff
@@ -682,6 +684,8 @@ def append(self, state=None, normalize=True, **kwargs):
         if state is not None:
             self._phase.state = state
 
+        # Non normalize form must be set using the mass fractions as the
```
DavidAkinpelu: @bryanwweber, thank you for pointing the `Near` issue. It uncovered another, more fundamental issue. Since "Y" is the native state parameter, `non-normalized` forms should be appended using that.

ischoegl: Thanks for pointing that out. Did you check what happens if you append using the `state` keyword, where concentrations are not normalized?

DavidAkinpelu: Appending with `state`,  `Mass fractions`  are not normalized while `mole fractions` are, because of the above-mentioned reason.

ischoegl: Thanks for the clarification! I'm not sure what's the default, but I believe append with `state` should *not* normalize as this would preserve the true value by default (i.e. `normalize=False` should be default, where a user would have to override to enforce normalization).

ischoegl: Looks like the logic actually has to be a little more complex, as a user would expect `normalize=True` if a composition is appended, whereas `normalize=False` would be the expectation if `state` or a full numpy array is used for either `X` or `Y`. It would be good to establish a couple of 'rules' here.

bryanwweber: I don't understand the need for this change... `set_mole_fractions_unnormalized` sets the mass fractions to the values corresponding to the appropriate unnormalized mole fractions. Can you clarify?

ischoegl: I believe values were sufficiently close in the `assertNear` unit test that it masked the fact that the test actually failed. @DavidAkinpelu ... could you confirm?

DavidAkinpelu: > I don't understand the need for this change... `set_mole_fractions_unnormalized` sets the mass fractions to the values corresponding to the appropriate unnormalized mole fractions. Can you clarify?

`SolutionArray` append method works with `state` by default. The issue here is that if we used `set_mole_fractions_unnormalized`, the `SolutionArray` appends the `state` values which will have a different `mean_molecular_weight` compare to the `phase` `mean_molecular_weight` because `setMassFractions` and `setMoleFractions` calculates `mean_molecular_weight` differently.

bryanwweber:
@DavidAkinpelu I'm going to reply in a top level comment here so that I can have some more space. I guess what you mean is this:

```python
import cantera as ct
gas
In [2]: gas = ct.Solution("h2o2.yaml")
In [3]: import numpy as np
In [5]: x = np.ones(gas.n_species) * 0.3
In [6]: gas.set_unnormalized_mole_fractions(x)
In [9]: gas.X
Out[9]: array([0.3, 0.3, 0.3, 0.3, 0.3, 0.3, 0.3, 0.3, 0.3, 0.3])
In [10]: gas.Y
Out[10]:
array([0.00912106, 0.00456053, 0.07238482, 0.14476964, 0.07694535,
       0.08150588, 0.14933017, 0.1538907 , 0.18074715, 0.1267447 ])
In [11]: gas.mean_molecular_weight
Out[11]: 66.3081
In [13]: np.dot(gas.X, gas.molecular_weights)
Out[13]: 66.3081
In [16]: 1/np.dot(gas.Y, 1/gas.molecular_weights)
Out[16]: 22.102699999999995
```

If the system were in a consistent state, those would be the same values, but they're not. As far as I can see, this kind of transformation is inevitable since the conversion from mole to mass fraction relies on the sum of them being equal to 1.0.

In the end, I don't think the right procedure is to simply disable the non-normalized mole fraction setter when `Y` is in the natural states set (at least, I think that's happening here. I'm having a hard time following the code). One should not expect the mass fractions to also be non-normalized if non-normalized mole fractions are set, and vice versa, for the reason I stated above. IMO, the code you had here before was much simpler and more robust (other than the test case).

In addition, I think it would be a better user experience if the `normalize` and `states` keyword-arguments were mutually exclusive in `append()`. I don't see any cases for using both together, and if it's needed, simply getting `TPX` and appending that would work just as well.

DavidAkinpelu:
Thanks for the clarification, you couldn't have explained it better. You are right. in the previous case, appending using `TPX` failed, the tests were only able to pass through because 1.e-16 is within the tolerance.

DavidAkinpelu:
> @DavidAkinpelu I'm going to reply in a top level comment here so that I can have some more space. I guess what you mean is this:
> 
> ```python
> import cantera as ct
> gas
> In [2]: gas = ct.Solution("h2o2.yaml")
> In [3]: import numpy as np
> In [5]: x = np.ones(gas.n_species) * 0.3
> In [6]: gas.set_unnormalized_mole_fractions(x)
> In [9]: gas.X
> Out[9]: array([0.3, 0.3, 0.3, 0.3, 0.3, 0.3, 0.3, 0.3, 0.3, 0.3])
> In [10]: gas.Y
> Out[10]:
> array([0.00912106, 0.00456053, 0.07238482, 0.14476964, 0.07694535,
>        0.08150588, 0.14933017, 0.1538907 , 0.18074715, 0.1267447 ])
> In [11]: gas.mean_molecular_weight
> Out[11]: 66.3081
> In [13]: np.dot(gas.X, gas.molecular_weights)
> Out[13]: 66.3081
> In [16]: 1/np.dot(gas.Y, 1/gas.molecular_weights)
> Out[16]: 22.102699999999995
> ```
> 
> If the system were in a consistent state, those would be the same values, but they're not. As far as I can see, this kind of transformation is inevitable since the conversion from mole to mass fraction relies on the sum of them being equal to 1.0.
> 
> In the end, I don't think the right procedure is to simply disable the non-normalized mole fraction setter when `Y` is in the natural states set (at least, I think that's happening here. I'm having a hard time following the code). One should not expect the mass fractions to also be non-normalized if non-normalized mole fractions are set, and vice versa, for the reason I stated above. IMO, the code you had here before was much simpler and more robust (other than the test case).
> 
> In addition, I think it would be a better user experience if the `normalize` and `states` keyword-arguments were mutually exclusive in `append()`. I don't see any cases for using both together, and if it's needed, simply getting `TPX` and appending that would work just as well.

ischoegl:
Linking #390 as this is related.

interfaces/cython/cantera/composite.py
```diff
@@ -649,6 +649,13 @@ def append(self, state=None, **kwargs):
           the full-state setters::
 
               mystates.append(T=300, P=101325, X={'O2':1.0, 'N2':3.76})
+
+        By default, the mass or mole fractions will be normalized so they
+        sum to 1.0. If this is not desired, the ``normalize`` argument can
+        be set to ``False``. In this case, the mass or mole fractions must
+        be specified as an array. For example::
```
ischoegl: Per discussions elsewhere, I believe the intuitive approach is to not normalize when appending a state or a numpy array specifying X or Y. If the composition is provided using a composition, then normalization should be the default.

I am suggesting this, as the state holds the 'true' information (as `Y`), and thus should *not* be changed. Further, a user presumably expects setters for numpy arrays to remain unchanged (even if `X` cannot be retrieved without being normalized for the output).

I am mainly commenting on the docstring here, and haven't checked the code, as I am starting to lose track on the overarching philosophy. Could you prepare a post (regular comment, not a reply here) summarizing the overall strategy when to normalize and when not to? At this point it may be good to see this documented in a single point.

DavidAkinpelu:
These are the conditions for SolutionArray.append:
Appending with ``state`` is mutually exclusive with the ``normalize`` argument. It is non-normalized all the time.
**Default**/ **None**:  If `X` and `Y` are specified as arrays, don't normalize , otherwise, normalize.
**True**: Normalize irrespective of how `X` and `Y` are given.
**False**: If `X` and `Y` are specified as arrays, don't normalize, otherwise, normalize.

ischoegl:
> These are the conditions for SolutionArray.append: [...]

Thanks, @DavidAkinpelu! 

... @bryanwweber are there comments/concerns? (I view the default for `X`/`Y` to not normalize for arrays as a potential sticking point. Myself, I believe the intent of the user here would be to preserve the numeric values, though).

DavidAkinpelu:
... @bryanwweber, this is ready for review.

interfaces/cython/cantera/composite.py
```diff
@@ -676,12 +685,29 @@ def append(self, state=None, **kwargs):
...
         elif len(kwargs) == 1:
-            attr, value = next(iter(kwargs.items()))
+            attr, value = kwargs.popitem()
             if frozenset(attr) not in self._phase._full_states:
                 raise KeyError(
                     "'{}' does not specify a full thermodynamic state".format(attr)
                 )
-            setattr(self._phase, attr, value)
+            if attr[-1] == "X":
+                if  isinstance(value[-1], np.ndarray) and normalize is not True:
```
ischoegl: You can use `not normalize` (Here and elsewhere)

interfaces/cython/cantera/composite.py
```diff
@@ -1221,7 +1293,7 @@ def write_hdf(self, filename, *args, cols=None, group=None, subgroup=None,
 
         return group
 
-    def read_hdf(self, filename, group=None, subgroup=None, force=False):
+    def read_hdf(self, filename, group=None, subgroup=None, force=False, normalize=True):
```
ischoegl, line 1273: I believe this should not normalize by default, as `hdf` is a custom container format where you presumably do not want to change anything.

interfaces/cython/cantera/onedim.py
```diff
@@ -436,20 +458,23 @@ def from_solution_array(self, arr, domain=None):
         meta = arr.meta
         super().restore_data(domain, states, other_cols, meta)
 
-    def to_pandas(self, species='X'):
+    def to_pandas(self, species='X', normalize=False):
```
ischoegl: I noticed that you don’t normalize by default for writing, but do normalize for reading? I think the consistent way would be to not normalize on both operations. You could argue that you don’t lose information when exporting, and leave it to the user to avoid losing information when importing if you prefer to leave as is.

interfaces/cython/cantera/onedim.py
```diff
@@ -565,7 +593,7 @@ def write_hdf(self, filename, *args, group=None, species='X', mode='a',
             msg = "Solution saved to '{0}' as group '{1}'."
             print(msg.format(filename, group))
 
-    def read_hdf(self, filename, group=None, restore_boundaries=True):
+    def read_hdf(self, filename, group=None, restore_boundaries=True, normalize=True):
```
ischoegl, line 590: Same: HDF readers should not normalize by default.

ischoegl (commented):
Some comments mostly on choices of default values. Here I believe that an exact recreation of the original state would be the expectation for default behavior of HDF, which is not as easily created by external tools as the other formats.

interfaces/cython/cantera/composite.py
DavidAkinpelu: I believe this is consistent with @bryanwweber 's position. You can check the previous conversations.

bryanwweber: Before we make any further changes, I'd like @speth to weigh in on the default values here and elsewhere.

interfaces/cython/cantera/composite.py
speth: I think that would be a pretty substantial inconsistency with the behavior of the C++ `Phase` class, which has always normalized mass and mole fractions, even when they are provided as arrays.

interfaces/cython/cantera/onedim.py
```diff
@@ -413,7 +421,21 @@ def to_solution_array(self, domain=None):
...
+                        arr._phase.set_unnormalized_mass_fractions(states[2])
+                        arr._phase.TP = states[0], states[1]
+                        arr._states[0] = arr._phase.state
+                    else:
+                        for i in range(n_points):
+                            arr._phase.set_unnormalized_mass_fractions(states[2][i])
+                            arr._phase.TP = states[0][i], states[1]
+                            arr._states[i] = arr._phase.state
+                else:
+                    arr.TPY = states
```
speth: Can this setter actually succeed? Based on the `if` condition above, `len(states) != 3` here, so this ought to fail, right?

interfaces/cython/cantera/onedim.py
```diff
@@ -387,20 +387,28 @@ def write_csv(self, filename, species='X', quiet=True):
...
         # save data
         cols = ('extra', 'T', 'D', species)
-        self.to_solution_array().write_csv(filename, cols=cols)
+        self.to_solution_array(normalize=normalize).write_csv(filename, cols=cols)
 
         if not quiet:
             print("Solution saved to '{0}'.".format(filename))
 
-    def to_solution_array(self, domain=None):
+    def to_solution_array(self, domain=None, normalize=False):
```
speth: While I agree that having the option to get the unnormalized composition is useful in certain circumstances, I'm really not convinced that it should be the default for extracting information from the 1D flame solver. Any deviations from a sum of 1 are just numerical noise, not anything deliberately set by the user, and the most notable feature will  be negative mass fractions below the absolute tolerance of the solver, which I don't think are that interesting to most users.

speth (changes requested):
Allowing users the option to preserve mass fraction arrays that do not sum to 1 and contain negative values makes sense to me, but I think we should be careful about when that makes sense as a default behavior. I view working with such arrays as something that only a fairly experienced user should be trying to do, given the numerous pitfalls, and it's not unreasonable to expect such users to be able to set this option when they need it.

The case where I can see doing this by default is when the source of the mass fractions is a `ThermoPhase` object, where the only way that such a mass fraction vector could have been set in the first place is with one of the `noNorm` setters.

The case of restoring data from a file is a bit fuzzier. If you assume that the data came from Cantera, it could make sense, but I don't think that's an assumption that's universally true, and I think from the viewpoint of seeing this as an advanced option, I would recommend defaulting to the current behavior of normalizing the mass fractions.

When setting the state from a list or Numpy array, deciding not to normalize by default would be a significant departure from the behavior for the `Solution` object and the underlying `Phase` class, and I don't think the `SolutionArray` class should be a special case here.

DavidAkinpelu:
@bryanwweber, @speth... The PR is updated and is ready for further review.

interfaces/cython/cantera/composite.py
ischoegl: This point is moot.

interfaces/cython/cantera/onedim.py
ischoegl: Same here.

interfaces/cython/cantera/onedim.py
ischoegl: And here

interfaces/cython/cantera/composite.py
```diff
@@ -676,12 +680,24 @@ def append(self, state=None, **kwargs):
...
+            if normalize or attr.endswith("Q"):
+                setattr(self._phase, attr, value)
+            else:
+                if attr.endswith("X"):
+                    self._phase.set_unnormalized_mole_fractions(value[-1])
+                    attr = attr[:-1]
+                    value = value[:-1]
+                    setattr(self._phase, attr, value)
+                elif attr.endswith("X"):
+                    self._phase.set_unnormalized_mass_fractions(value[-1])
```
```suggestion
                elif attr.endswith("Y"):
                    self._phase.set_unnormalized_mass_fractions(value[-1])
```

interfaces/cython/cantera/composite.py
```diff
@@ -676,12 +680,24 @@ def append(self, state=None, **kwargs):
...
+                    attr = attr[:-1]
+                    value = value[:-1]
+                    setattr(self._phase, attr, value)
```
speth: These three lines can be moved out of the inner `if` block, since they're repeated. I think the same applies to several other blocks with a similar structure.

interfaces/cython/cantera/composite.py
```diff
@@ -649,6 +649,10 @@ def append(self, state=None, **kwargs):
           the full-state setters::
 
               mystates.append(T=300, P=101325, X={'O2':1.0, 'N2':3.76})
+
+        By default, the mass or mole fractions will be normalized i.e they
+        sum up to 1.0. If this is not desired, the ``normalize`` argument
+        can be set to ``False``.
```
speth: You might want to also document that `normalize=True` truncates negative values, given that that was one of the concerns that prompted this PR.

interfaces/cython/cantera/composite.py
```diff
@@ -759,13 +785,17 @@ def restore_data(self, data):
         :param data: Dictionary holding data to be restored, where keys
             refer to thermodynamic states (e.g. ``T``, ``density``) or extra
             entries, and values contain corresponding data.
+        :param normalize: If True, mole or mass fractions are normalized
+            i.e. they sum up to 1.0. If False, mole or mass fractions
+            are non-normalized.
```
```suggestion
            so that they sum up to 1.0. If False, mole or mass fractions
            are not normalized.
```

interfaces/cython/cantera/composite.py
```diff
@@ -759,13 +785,17 @@ def restore_data(self, data):
...
 
         The receiving `SolutionArray` either has to be empty or should have
         matching dimensions. Essential state properties and extra entries
         are detected automatically whereas stored information of calculated
         properties is omitted. If the receiving `SolutionArray` has extra
         entries already specified, only those will be imported; if *labels*
         does not contain those entries, an error is raised.
+
```
speth: Avoid adding a blank line here.

interfaces/cython/cantera/composite.py
```diff
@@ -1239,11 +1295,15 @@ def read_hdf(self, filename, group=None, subgroup=None, force=False):
...
             suppressed.
+        :param normalize: Passed on to `restore_data`. If True, mole or mass
+            fractions are normalized i.e. they sum up to 1.0. If False, mole
+            or mass fractions are non-normalized.
         :return: User-defined attributes provided to describe the group holding
             the `SolutionArray` information.
 
         The method imports data using `restore_data` and requires a working
         installation of h5py (`h5py` can be installed using pip or conda).
+
```
speth: Avoid adding the blank line here.

interfaces/cython/cantera/composite.py
```diff
@@ -1239,11 +1295,15 @@ def read_hdf(self, filename, group=None, subgroup=None, force=False):
             `Solution` object), with an error being raised if the current source
             does not match the original source. If True, the error is
             suppressed.
+        :param normalize: Passed on to `restore_data`. If True, mole or mass
+            fractions are normalized i.e. they sum up to 1.0. If False, mole
+            or mass fractions are non-normalized.
```
```suggestion
            fractions are normalized so that they sum up to 1.0. If False, mole
            or mass fractions are not normalized.
```

interfaces/cython/cantera/onedim.py
```diff
@@ -387,20 +387,27 @@ def write_csv(self, filename, species='X', quiet=True):
         :param species:
             Attribute to use obtaining species profiles, e.g. ``X`` for
             mole fractions or ``Y`` for mass fractions.
+        :param normalize:
+            Boolean flag to indicate whether the mole/mass fractions should
+            be normalized (default is ``True``)
```
speth: In general, I don't think it's necessary to document the default value here -- it's visible in the code, as well as the generated Sphinx documentation.

interfaces/cython/cantera/onedim.py
```diff
@@ -413,7 +420,21 @@ def to_solution_array(self, domain=None):
         if n_points:
             arr = SolutionArray(self.phase(domain), n_points,
                                 extra=other_cols, meta=meta)
-            arr.TPY = states
+            if normalize:
+                arr.TPY = states
+            else:
+                if len(states) == 3:
+                    if n_points==1:
+                        arr._phase.set_unnormalized_mass_fractions(states[2])
```
speth: I wonder if using `np.atleast_2d` on the `states` variable would allow us to avoid this special case for `n_points == 1`.

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -417,6 +454,33 @@ def check(a, b):
...
+    def test_import_no_norm_water(self):
+        outfile = self.test_work_path / "solutionarray.h5"
+        # In Python >= 3.8, this can be replaced by the missing_ok argument
+        if outfile.is_file():
+            outfile.unlink()
+
+        w = ct.Water()
+        w.TQ = 300, 0.5
+        states = ct.SolutionArray(w, 5)
+        states.write_hdf(outfile)
```
speth, line 466: With the change to the default value for the `normalize` argument, this no longer tests the case where `normalize=False`, right?

speth (changes requested):
Thanks, @DavidAkinpelu. This largely looks good to me. With the exception of the one case where mass and mole fractions are mixed up, most of my comments are just a few points where the code might be simplified a bit and some suggestions for the documentation.

DavidAkinpelu:
@speth... The PR is updated and is ready for further review.

interfaces/cython/cantera/composite.py
```diff
@@ -649,6 +649,10 @@ def append(self, state=None, **kwargs):
           the full-state setters::
 
               mystates.append(T=300, P=101325, X={'O2':1.0, 'N2':3.76})
+
+        By default, the mass or mole fractions will be normalized i.e negative values
+        are trumncated and the mass or mole fractions are sum up to 1.0. If this
```
speth: Sorry, just noticed another typo here:
```suggestion
        are truncated and the mass or mole fractions sum up to 1.0. If this
```

speth (changes requested):
Thanks for the updates, @davidAkinpelu. I just saw one last typo that can be fixed. Otherwise, I think this is good to go.

DavidAkinpelu:
@speth.. The PR has been updated.

speth (approved):
Looks good. Thank you!

PR #1038, opened 2021-05-12
[Kinetics] Fix issue 717, Add check for sticking coefficient more than 1

### LINKED ISSUE #717: Nothing to check or ensure that Sticking Coefficients are in the range [0,1]
### Summary
Cantera doesn't check or ensure that Sticking Coefficients are in the range [0,1].

The documentation both at
at https://cantera.github.io/docs/sphinx/html/cti/reactions.html#sticking-coefficients
and https://cantera.org/science/reactions.html#sticking-coefficients
states

> "The values of these constants must be specified so that the sticking coefficient γ is between 0 and 1 for all temperatures.“

but the code does nothing to enforce or check this.

### Cantera version
2.3.0 (and earlier)

### Operating System
All

### Expected Behavior

Copying the CHEMKIN approach would be one option, and may be what some users expect.

The 1996 Sandia technical report manual to SURFACE CHEMKIN-III  https://doi.org/10.2172/481906 page 38 has a footnote
“CAUTION: Because γi is defined as a probability, it must lie between 0 and 1 to make physical sense. Therefore, SURFACE CHEMKIN-III checks the value of γi, and an unphysical sticking coefficient greater than 1 is changed to the value 1. Some earlier versions of SURFACE CHEMKIN did not truncate the values at 1."
It seems this note is also in [more recent versions](http://www.cvd.louisville.edu/Course/Chemical%20Vapour%20Deposition/Manuals/chemkin/chemkin11surfacechemkin.pdf) of the commercial Chemkin manual.

The CHEMKIN definition is thus
Screen Shot 2019-10-01 at 4 43 39 PM
whereas the Cantera definition is this
Screen Shot 2019-10-01 at 4 58 41 PM

Another option would be to include a sanity/consistency/validation check, a bit like when negative A factors are encountered.  This could at very least give the user a warning. Or it could raise an error unless users add a special flag to "allow greater than 1".

Or some hybrid: allow greater than 1 if the user specifies it with a flag, else truncate it. 
Or truncate it if the user specifies with a flag. 
Probably the code that translates from CHEMKIN files (ck2cti or ck2yaml) should set whatever flags reproduce the CHEMKIN behaviour?

### Actual Behavior

As far as I can tell from [this](https://github.com/Cantera/cantera/blob/c4aff04418851d159ae9ca7709f823b3969b15d7/src/kinetics/InterfaceKinetics.cpp#L619) bit of code, Cantera does nothing to enforce or check that sticking coefficients are between 0 and 1. If parameters are provided that evaluate to 1e8 at some temperature, then very non-physical (and probably stiff) rates will result.

12Chao:

- This pull request fixes issue #717, which checks if the sticking coefficient gamma is more than 1.  The check is conducted at
6 different temperature points from 200K to 10000K for `InterfaceReaction` and `BlowersMaselInterfaceReaction`. Any temperature point with sticking coefficient more than 1 will be added into error message. 

Closes #717 

**Checklist**

### DIFF
--- a/include/cantera/base/global.h
+++ b/include/cantera/base/global.h
@@ -221,6 +221,9 @@ void warn_user(const std::string& method, const std::string& msg,
 //! @copydoc Application::make_deprecation_warnings_fatal
 void make_deprecation_warnings_fatal();
 
+//! @copydoc Application::make_warnings_fatal
+void make_warnings_fatal();
+
 //! @copydoc Application::suppress_thermo_warnings
 void suppress_thermo_warnings(bool suppress=true);
 
--- a/include/cantera/kinetics/Kinetics.h
+++ b/include/cantera/kinetics/Kinetics.h
@@ -893,6 +893,11 @@ public:
         m_root = root;
     }
 
+    //! Calculate the reaction enthalpy of a reaction which
+    //! has not necessarily been added into the Kinetics object
+    virtual double reactionEnthalpy(const Composition& reactants,
+                                    const Composition& products);
+
 protected:
     //! Cache for saved calculations within each Kinetics object.
     ValueCache m_cache;
--- a/include/cantera/kinetics/Reaction.h
+++ b/include/cantera/kinetics/Reaction.h
@@ -67,6 +67,10 @@ public:
     //! valid.
     virtual void validate();
 
+    //! Perform validation checks that need access to a complete Kinetics objects, for
+    // example to retrieve information about reactant / product species.
+    virtual void validate(Kinetics& kin) {}
+
     //! Return the parameters such that an identical Reaction could be reconstructed
     //! using the newReaction() function. Behavior specific to derived classes is
     //! handled by the getParameters() method.
@@ -197,7 +201,9 @@ public:
     ElementaryReaction2();
     ElementaryReaction2(const Composition& reactants, const Composition products,
                         const Arrhenius& rate);
+
     virtual void validate();
+    using Reaction::validate;
     virtual void getParameters(AnyMap& reactionNode) const;
 
     virtual std::string type() const {
@@ -283,7 +289,9 @@ public:
 
     virtual std::string reactantString() const;
     virtual std::string productString() const;
+
     virtual void validate();
+    using Reaction::validate;
     virtual void calculateRateCoeffUnits(const Kinetics& kin);
     virtual void getParameters(AnyMap& reactionNode) const;
 
@@ -347,6 +355,7 @@ public:
     }
 
     virtual void validate();
+    using Reaction::validate;
     virtual void getParameters(AnyMap& reactionNode) const;
 
     Plog rate;
@@ -396,6 +405,9 @@ public:
                       const Arrhenius& rate, bool isStick=false);
     virtual void calculateRateCoeffUnits(const Kinetics& kin);
     virtual void getParameters(AnyMap& reactionNode) const;
+
+    virtual void validate(Kinetics& kin);
+    using Reaction::validate;
 
     virtual std::string type() const {
         return "interface";
@@ -450,6 +462,7 @@ public:
     virtual void getParameters(AnyMap& reactionNode) const;
     virtual void validate();
     virtual void calculateRateCoeffUnits(const Kinetics& kin);
+    virtual void validate(Kinetics& kin);
 
     virtual std::string type() const {
         return "surface-Blowers-Masel";
@@ -717,5 +730,4 @@ void setupElectrochemicalReaction(ElectrochemicalReaction&,
 void setupBlowersMaselInterfaceReaction(BlowersMaselInterfaceReaction&,
                                         const AnyMap&, const Kinetics&);
 }
-
 #endif
--- a/interfaces/cython/cantera/test/test_kinetics.py
+++ b/interfaces/cython/cantera/test/test_kinetics.py
@@ -470,6 +470,23 @@ class KineticsRepeatability(utilities.CanteraTest):
             for msg in err_msg:
                 self.assertIn(msg, err_msg_list)
 
+    def test_sticking_coeff_err(self):
+        err_msg = ("Sticking coefficient is greater than 1 for reaction \'O2 \+ 2 PT\(S\) => 2 O\(S\)\'",
+                   "at T = 200.0",
+                   "at T = 500.0",
+                   "at T = 1000.0",
+                   "at T = 2000.0",
+                   "at T = 5000.0",
+                   "at T = 10000.0",
+                   "Sticking coefficient is greater than 1 for reaction",
+                   "CanteraError thrown by InterfaceReaction::validate:",
+                   "CanteraError thrown by BlowersMaselInterfaceReaction::validate:"
+                   )
+        ct.make_warnings_fatal()
+        for err in err_msg:
+            with self.assertRaisesRegex(ct.CanteraError, err):
+                gas = ct.Solution('sticking_coeff_check.yaml')
+                surface = ct.Interface('sticking_coeff_check.yaml', 'Pt_surf', [gas])
 
 class TestUndeclared(utilities.CanteraTest):
 
--- a/interfaces/cython/cantera/utils.pyx
+++ b/interfaces/cython/cantera/utils.pyx
@@ -70,6 +70,9 @@ def make_deprecation_warnings_fatal():
                             message='.*Cantera.*')  # for warnings in Cython code
     Cxx_make_deprecation_warnings_fatal()
 
+def make_warnings_fatal():# for warnings in Cython code
+    Cxx_make_warnings_fatal()
+
 def suppress_deprecation_warnings():
     warnings.filterwarnings('ignore', category=DeprecationWarning,
                             module='cantera')  # for warnings in Python code
--- a/src/kinetics/Kinetics.cpp
+++ b/src/kinetics/Kinetics.cpp
@@ -528,6 +528,8 @@ void Kinetics::resizeSpecies()
 bool Kinetics::addReaction(shared_ptr<Reaction> r, bool resize)
 {
     r->validate();
+    r->validate(*this);
+
     if (m_kk == 0) {
         init();
     }
@@ -645,6 +647,26 @@ shared_ptr<const Reaction> Kinetics::reaction(size_t i) const
 {
     checkReactionIndex(i);
     return m_reactions[i];
+}
+
+double Kinetics::reactionEnthalpy(const Composition& reactants, const Composition& products)
+{
+    vector_fp hk(nTotalSpecies());
+    for (size_t n = 0; n < nPhases(); n++) {
+        thermo(n).getPartialMolarEnthalpies(&hk[m_start[n]]);
+    }
+    double rxn_deltaH = 0;
+    for (const auto& product : products) {
+        size_t k = kineticsSpeciesIndex(product.first);
+        double stoich = product.second;
+        rxn_deltaH += hk[k] * stoich;
+    }
+    for (const auto& reactant : reactants) {
+            size_t k = kineticsSpeciesIndex(reactant.first);
+            double stoich = reactant.second;
+            rxn_deltaH -= hk[k] * stoich;
+        }
+    return rxn_deltaH;
 }
 
 }
--- a/src/kinetics/Reaction.cpp
+++ b/src/kinetics/Reaction.cpp
@@ -801,6 +801,26 @@ void InterfaceReaction::getParameters(AnyMap& reactionNode) const
     }
 }
 
+void InterfaceReaction::validate(Kinetics& kin)
+{
+    if (is_sticking_coefficient) {
+        fmt::memory_buffer err_reactions;
+        double T[] = {200.0, 500.0, 1000.0, 2000.0, 5000.0, 10000.0};
+        for (size_t i=0; i < 6; i++) {
+            double k = rate.updateRC(log(T[i]), 1/T[i]);
+            if (k > 1) {
+                fmt_append(err_reactions,
+                    "\n Sticking coefficient is greater than 1 for reaction '{}'\n"
+                    " at T = {:.1f}\n",
+                    equation(), T[i]);
+            }
+        }
+        if (err_reactions.size()) {
+            warn_user("InterfaceReaction::validate", to_string(err_reactions));
+        }
+    }
+}
+
 ElectrochemicalReaction::ElectrochemicalReaction()
     : beta(0.5)
     , exchange_current_density_formulation(false)
@@ -898,6 +918,33 @@ void BlowersMaselInterfaceReaction::validate()
         throw InputFileError("BlowersMaselInterfaceReaction::validate", input,
             "Undeclared negative pre-exponential factor found in reaction '"
             + equation() + "'");
+        }
+}
+
+void BlowersMaselInterfaceReaction::validate(Kinetics& kin)
+{
+    if (is_sticking_coefficient) {
+        double original_T = kin.thermo().temperature();
+        double original_P = kin.thermo().pressure();
+        fmt::memory_buffer err_reactions;
+        double T[] = {200.0, 500.0, 1000.0, 2000.0, 5000.0, 10000.0};
+        size_t length_T = sizeof(T) / sizeof(T[0]);
+        for (size_t i=0; i < length_T; i++) {
+            kin.thermo().setState_TP(T[i], original_P);
+            double rxn_deltaH = kin.reactionEnthalpy(reactants, products);
+            double k = rate.updateRC(log(T[i]), 1/T[i], rxn_deltaH);
+            if (k > 1) {
+                fmt_append(err_reactions,
+                          "\n Sticking coefficient is bigger than 1 for reaction '{}'\n"
+                          " at T = {:.1f}\n",
+                          equation(), T[i]);
+            }
+
+        }
+        kin.thermo().setState_TP(original_T, original_P);
+        if (err_reactions.size()) {
+            warn_user("BlowersMaselInterfaceReaction::validate", to_string(err_reactions));
+        }
     }
 }
 
@@ -2003,6 +2050,7 @@ std::vector<shared_ptr<Reaction>> getReactions(const AnyValue& items,
     for (const auto& node : items.asVector<AnyMap>()) {
         shared_ptr<Reaction> R(newReaction(node, kinetics));
         R->validate();
+        R->validate(kinetics);
         if (R->valid() && R->checkSpecies(kinetics)) {
             all_reactions.emplace_back(R);
         }
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -131,6 +131,7 @@ cdef extern from "cantera/base/global.h" namespace "Cantera":
     cdef XML_Node* CxxGetXmlFile "Cantera::get_XML_File" (string) except +translate_exception
     cdef XML_Node* CxxGetXmlFromString "Cantera::get_XML_from_string" (string) except +translate_exception
     cdef void Cxx_make_deprecation_warnings_fatal "Cantera::make_deprecation_warnings_fatal" ()
+    cdef void Cxx_make_warnings_fatal "Cantera::make_warnings_fatal" ()
     cdef void Cxx_suppress_deprecation_warnings "Cantera::suppress_deprecation_warnings" ()
     cdef void Cxx_suppress_thermo_warnings "Cantera::suppress_thermo_warnings" (cbool)
     cdef void Cxx_use_legacy_rate_constants "Cantera::use_legacy_rate_constants" (cbool)
--- a/src/base/application.cpp
+++ b/src/base/application.cpp
@@ -127,6 +127,7 @@ Application::Application() :
     m_suppress_deprecation_warnings(false),
     m_fatal_deprecation_warnings(false),
     m_suppress_thermo_warnings(false),
+    m_fatal_warnings(false),
 #if CT_LEGACY_RATE_CONSTANTS
     m_use_legacy_rate_constants(true)
 #else
@@ -182,6 +183,9 @@ void Application::warn_deprecated(const std::string& method,
 void Application::warn_user(const std::string& method,
                             const std::string& extra)
 {
+    if (m_fatal_warnings) {
+        throw CanteraError(method, extra);
+    }
     writelog(fmt::format("CanteraWarning: {}: {}", method, extra));
     writelogendl();
 }
--- a/src/base/application.h
+++ b/src/base/application.h
@@ -354,6 +354,12 @@ public:
     //! information can be specified in *extra*.
     void warn_user(const std::string& method, const std::string& extra="");
 
+    //! Turns Cantera warnings into exceptions. Activated within the test
+    //! suite to make sure that your warning message are being raised.
+    void make_warnings_fatal() {
+        m_fatal_warnings = true;
+    }
+
     //! Globally disable printing of warnings about problematic thermo data,
     //! e.g. NASA polynomials with discontinuities at the midpoint temperature.
     void suppress_thermo_warnings(bool suppress=true) {
@@ -453,6 +459,7 @@ protected:
     bool m_suppress_deprecation_warnings;
     bool m_fatal_deprecation_warnings;
     bool m_suppress_thermo_warnings;
+    bool m_fatal_warnings;
     bool m_use_legacy_rate_constants;
 
     ThreadMessages pMessenger;
--- a/src/base/global.cpp
+++ b/src/base/global.cpp
@@ -71,6 +71,11 @@ void make_deprecation_warnings_fatal()
     app()->make_deprecation_warnings_fatal();
 }
 
+void make_warnings_fatal()
+{
+    app()->make_warnings_fatal();
+}
+
 void suppress_thermo_warnings(bool suppress)
 {
     app()->suppress_thermo_warnings(suppress);
--- a/test/data/sticking_coeff_check.yaml
+++ b/test/data/sticking_coeff_check.yaml
@@ -1,0 +1,56 @@
+units: {length: cm, quantity: mol, activation-energy: J/mol}
+
+phases:
+- name: gas
+  thermo: ideal-gas
+  elements: [O, H, C, N, Ar]
+  species:
+  - gri30.yaml/species: [ O2, OH]
+  skip-undeclared-elements: true
+  kinetics: gas
+  reactions:
+  - gri30.yaml/reactions: declared-species
+  transport: mixture-averaged
+- name: Pt_surf
+  thermo: ideal-surface
+  elements: [Pt, H, O, C]
+  species: [{Pt-surf-species: all}]
+  kinetics: surface
+  reactions: [Pt-reactions]
+
+Pt-surf-species:
+- name: PT(S)
+  composition: {Pt: 1}
+  thermo:
+    model: NASA7
+    temperature-ranges: [300.0, 1000.0, 3000.0]
+    data:
+    - [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
+    - [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
+- name: OH(S)
+  composition: {O: 1, H: 1, Pt: 1}
+  thermo:
+    model: NASA7
+    temperature-ranges: [300.0, 1000.0, 3000.0]
+    data:
+    - [-2.0340881, 9.3662683e-03, 6.6275214e-07, -5.2074887e-09, 1.7088735e-12,
+      -2.5319949e+04, 8.9863186]
+    - [1.8249973, 3.2501565e-03, -3.1197541e-07, -3.4603206e-10, 7.9171472e-14,
+      -2.6685492e+04, -12.280891]
+- name: O(S)
+  composition: {O: 1, Pt: 1}
+  thermo:
+    model: NASA7
+    temperature-ranges: [300.0, 1000.0, 3000.0]
+    data:
+    - [-0.94986904, 7.4042305e-03, -1.0451424e-06, -6.112042e-09, 3.3787992e-12,
+      -1.3209912e+04, 3.6137905]
+    - [1.945418, 9.1761647e-04, -1.1226719e-07, -9.9099624e-11, 2.4307699e-14,
+      -1.4005187e+04, -11.531663]
+
+Pt-reactions:
+- equation: O2 + 2 PT(S) => 2 O(S)  # Reaction 1
+  sticking-coefficient: {A: 2.0, b: 0, Ea: 0}
+- equation: OH + PT(S) => OH(S)  # Reaction 2
+  type: Blowers-Masel
+  sticking-coefficient: {A: 2.0, b: 0, Ea0: 0, w: 100000}


src/kinetics/Reaction.cpp
```diff
@@ -773,6 +773,14 @@ void InterfaceReaction::getParameters(AnyMap& reactionNode) const
     }
 }
 
+void InterfaceReaction::validate() {
+    ElementaryReaction::validate();
+    if (is_sticking_coefficient != false) {
```
ischoegl: can be simplified to `if (is_sticking_coefficient) {`

include/cantera/kinetics/RxnRates.h
```diff
@@ -91,6 +91,10 @@ class Arrhenius
         return m_E;
     }
 
+    //! Check to make sure that the sticking coefficient is
+    //! less than 1 over a range of temperatures
+    void validate(const std::string& equation);
```
ischoegl: I agree that sticking coefficients are related to rate expressions. Nevertheless, I am not convinced that this is the correct place to add this check, as for the vast majority of `Arrhenius` rates the sticking coefficient does not apply. With the upcoming new framework, it may be possible to clarify this ambiguity eventually.

src/kinetics/RxnRates.cpp
```diff
@@ -68,6 +69,23 @@ void Arrhenius::getParameters(AnyMap& rateNode, const Units& rate_units) const
...
+        double gamma = m_A * std::exp(m_b*log(T[i]) - m_E*(1/T[i]));
+        if (gamma > 1) {
+             format_to(err_reactions,
+                       "\n Sticking coefficient is more than 1 for reaction '{}'\n"
+                       " at T = {:.1f}\n",
+                       equation, T[i]);
+        }
+    }
+    if (err_reactions.size()) {
+        throw CanteraError("Arrhenius::validate", to_string(err_reactions));
```
ischoegl: With the introduction of YAML, `InputFileError` provides a mechanism to point directly to the YAML section where the issue occurs (although the exception still works for CTI). It provides superior feedback, and the function call is almost identical (it's already in use in `Reaction.cpp` for examples). Note that `RxnRates.cpp` does not have the required information available, but `Reaction.cpp` does.

ischoegl:
While this looks pretty good, I believe that the validation should be implemented strictly within `InterfaceReaction::validate()` / `BlowersMaselInterfaceReaction::validate()` as the checks you implemented are specific to those types of reactions. (*Edit: I am actually torn on this issue, as within the new framework I am currently implementing, - see Cantera/enhancements#87, - handling things within `ReactionRate.cpp` would make sense. A lot of this is still under development, so this won't become clear until #995 is merged.*)

> 6 different temperature points from 200K to 10000K

For temperatures, it may make sense to just check the range where species thermo information of reactants and products is valid?

ischoegl (changes requested):

bryanwweber:
@12Chao Have you been able to update this PR in response to the review? Thanks!

12Chao:
Sorry for the late response, I agree that there could be a better place for validate functions, but I cannot figure out where is more reasonable for these functions given that an `Arrhenius` class is created before a `SurfaceArrhenius` class.

speth:
You can just evaluate the rate constant at the specified set of temperatures in `InterfaceReaction::validate` and `BlowersMaselInterfaceReaction::validate`, right? I think that would alleviate the concern about having a strange `Arrhenius::validate` method that would likely fail if you ran it for a non-sticking reaction.

12Chao:
> You can just evaluate the rate constant at the specified set of temperatures in `InterfaceReaction::validate` and `BlowersMaselInterfaceReaction::validate`, right? I think that would alleviate the concern about having a strange `Arrhenius::validate` method that would likely fail if you ran it for a non-sticking reaction.

Yes, `Arrhenius::validate` and `BlowersMasel::validate` functions are only called by `InterfaceReaction::validate` and `BlowersMaselInterfaceReaction::validate` for sticking reactions, but I think @ischoegl means that maybe the rate validate function could be placed at some less confused place with the new framework coming up? I agree that adding the validate function into `SurfaceArrhenius` and `BMSurfaceArrhenius` classes makes more sense, but I could not get it work because the `Arrhenius` and `BlowersMasel ` rate classes are created first when adding the surface reactions, maybe this can be achieved with the new reaction framework?

speth:
> Yes, `Arrhenius::validate` and `BlowersMasel::validate` functions are only called by `InterfaceReaction::validate` and `BlowersMaselInterfaceReaction::validate` for sticking reactions, but I think @ischoegl means that maybe the rate validate function could be placed at some less confused place with the new framework coming up? 

I'm not sure there's that much reason to worry about future changes to use the new `ReactionRate` classes for interface reactions. If this feature is implemented purely on the `InterfaceReaction` and `BlowersMaselInterfaceReaction` classes, it will be easier to make the transition than if this validation is done in the `Arrhenius` class that doesn't (and shouldn't) know anything about sticking coefficients in the first place.

> I agree that adding the validate function into `SurfaceArrhenius` and `BMSurfaceArrhenius` classes makes more sense, but I could not get it work because the `Arrhenius` and `BlowersMasel ` rate classes are created first when adding the surface reactions, maybe this can be achieved with the new reaction framework?

I am not suggesting moving the validation to the `SurfaceArrhenius` or `BMSurfaceArrhenius` classes - I'm suggesting that the validation should happen entirely in the corresponding `Reaction` class's `validate` method, since this is where the check can be conditional based on the `is_sticking_coefficient` variable. You already have half of the logic placed there. For example, something like the following pseudo-code:
```c++
void InterfaceReaction::validate() {
    ElementaryReaction::validate();
    if (is_sticking_coefficient) {
        for (double T : temperatureList) {
            k = rate.updateRC(log(T), 1.0/T);
            if (k > 1) {
                throw InputFileError("InterfaceReaction::validate", input, "bad sticking factor");
            }
        }
    }
}
```

I think you'll need to do something a little more complex to validate this for Blowers-Masel reactions, given your parameter space now has two dimensions.

12Chao:
@speth Thanks for the instruction, Could you please give me a hint about how to call the `ThermoPhase` object in `reaction.cpp` file so that I can use the delta enthalpy for Blowers-Masel reaction? I failed trying to use `m_thermo` and `thermo_t` objects to represent the `ThermoPhase` .

speth:
A `Reaction` object doesn't inherently have access to a `ThermoPhase` object to use for this purpose. Even if it did, I think it wouldn't be sufficient to validate using the enthalpy of reaction at the current state of the `ThermoPhase` (when the reaction is being added) - I think you'd need to consider the enthalpy of reaction at any temperature that's valid for the phase.

Doing that would complicate things a bit -- you'd probably need to pass the `Kinetics` object as an argument to the `Reaction::validate` method (from which you can get access to the `ThermoPhase` object(s), and that would require shuffling some things around to deprecate the current `validate` method as well.

12Chao:
Thanks for all the help, I have addressed all the comments for this PR and would be appreciated if anyone could review the new changes

ischoegl:
@12Chao ... thanks for your work on this. But I do have a philosophical question - checking the sticking coefficient here goes in the direction of a mechanism check, which is historically not really done within Cantera but relegated to external tools (e.g. ChemCheck, see Cantera/enhancements#42). In another context, you'll see some warnings (e.g. thermo data with discontinuities), but that's about that. 

Throwing an error means that a (potentially inexperienced) user would have to fix the YAML file in order to proceed. If they just happened to convert a (buggy) CHEMKIN file from some web repository, this would mean that in most cases those users would be stuck and post on the user group, etc.. So the main question here (which goes beyond syntax) would be:
* Would it make sense to just limit the sticking coefficient to 1 by a simple `min` operation? (Just like rate-of-progress is automatically set to 0 when concentrations become negative).
* Could you run the check you implemented just once at the beginning to trigger the warning? Also, if a user won't ever run at temperatures high enough to be of concern, does there even have to be a warning?

Regarding changing the signature of `Reaction::validate()` to include `Kinetics` I do have some reservations ... current checks mostly involve simple syntax/formatting checks, and not more comprehensive kinetic validations. In general, there is a chicken-and-egg issue when defining reactions: you build `Kinetics` from `Reactions`, but at the same time need `Kinetics` to validate a reaction. Keep in mind that there is the possibility of building kinetics from 'scratch' (individual reactions), where a completely coherent approach would be to run the check you're envisioning when a reaction is added to `Kinetics` (instead of when it is originally created). One step in this direction would be to simply use a different name and call it later, while leaving the original `Reaction::validate()` as is - as it is really meant for a different kind of 'validation'.

*PS: Sorry if this goes beyond my original feedback (which was strictly on code/implementation); I meant to post some update since, but haven't gotten around that until now.*

12Chao:
> checking the sticking coefficient here goes in the direction of a mechanism check, which is historically not really done within Cantera but relegated to external tools (e.g. ChemCheck, see [Cantera/enhancements#42](https://github.com/Cantera/enhancements/issues/42)). In another context, you'll see some warnings (e.g. thermo data with discontinuities), but that's about that.

@ischoegl Thanks for the comments. Moving the check to ChemCheck is not a hard task, and I can definately work on that in the future.
> * Would it make sense to just limit the sticking coefficient to 1 by a simple `min` operation? (Just like rate-of-progress is automatically set to 0 when concentrations become negative)

 I'm not sure how to deal with the problematic sticking coefficient except raising an error message, but I think at least we need to warn the user if their sticking coefficient exceeds the limit.
 
> Regarding changing the signature of `Reaction::validate()` to include `Kinetics` I do have some reservations ... current checks mostly involve simple syntax/formatting checks, and not more comprehensive kinetic validations. In general, there is a chicken-and-egg issue when defining reactions: you build `Kinetics` from `Reactions`, but at the same time need `Kinetics` to validate a reaction. Keep in mind that there is the possibility of building kinetics from 'scratch' (individual reactions), where a completely coherent approach would be to run the check you're envisioning when a reaction is added to `Kinetics` (instead of when it is originally created). One step in this direction would be to simply use a different name and call it later, while leaving the original `Reaction::validate()` as is - as it is really meant for a different kind of 'validation'.

I'm a little confused about this comment. The `Kinetics`  input is not used for the validate of other types of reactions but only `BlowersMaselInterfaceReactions`. `Kinetics` is used because  `ThermoPhase` is needed in the sticking coefficient calculation for `BlowersMaselInterfaceReactions` , there is no reaction data involved. Even a new reaction is added, it has to go through the same validation as well, so there should be no problem of validating the new reaction right(because it only needs the `Kinetics` to obtain species thermo data from the `ThermoPhase`, sorry if I was wrong)?



> * Could you run the check you implemented just once at the beginning to trigger the warning? Also, if a user won't ever run at temperatures high enough to be of concern, does there even have to be a warning?

This is a good point,  I am currently just doing the same check like the PLOG reaction validation, but I can try to move the validate functions to `KineticsFactory::addReactions` and only validate the current temperature point, in that way I don't have to add `Kinetics` input to all the validate functions  either.

ischoegl:
> I'm a little confused about this comment. [...]

It's a pretty subtle point. A lot of the `Reaction` derived objects use overloaded methods, so they can be called from the base without having to think about a specific implementation (meaning that all methods are called in the same way). @speth mentioned that the current (argument-less) `validate` method would have to be deprecated in favor of one that takes `Kinetics` as an argument to expose thermo data. Extrapolating, this would mean to add `Kinetics` everywhere even if this is ultimately only needed for a select few cases. 

My (admittedly personal) opinion here is to use arguments as frugally as possible (i.e. don't add them unless absolutely needed), which would favor the case against changing the current `validate` signature. The other subtle point here is that checking things as late as possible gets around the chicken-and-egg problem I previously mentioned. If your check were to be executed when the reaction is added to `Kinetics`, things flow naturally. As things stand now, `Reaction::validate` is called when the object is instantiated. All of this is admittedly subtle - so take it with a grain of salt.

12Chao:
Thanks for the suggestion! I'll try to come up with a better way to avoid the changes to these `validate` functions.

ischoegl:
> Thanks for the suggestion! I'll try to come up with a better way to avoid the changes to these `validate` functions.

Sure. But wait for others to weigh in also - whatever you do, these subtle points shouldn't keep you from wrapping this work up. Passing `Kinetics` isn't a problem with the code as it stands, and my preference to run some of those checks as late as possible may not be universally shared. I am just hoping to convince everyone eventually ...

My main point is that you should be careful raising an error where other codes (chemkin?) may not find an issue. Having a converted buggy mechanism fail would create some unnecessary repercussions on the user group where simply limiting the output to 1 would be easier. Just looking at the code, having `k = std::min(1., updateRC(logT, recipT));` (if I understand how this limit is applied correctly) would sidestep things for the most part (this could even be checked within `updateRC` itself). Issuing a warning would certainly be a benefit of course.

speth:
> My (admittedly personal) opinion here is to use arguments as frugally as possible (i.e. don't add them unless absolutely needed), which would favor the case against changing the current `validate` signature.

I agree in general, but I think that this isn't actually the only case where having the `Kinetics` (and by extension, `ThermoPhase`) object present for validating a reaction would be helpful. For instance, it could also be used to select the temperatures used for validating P-log rates based on the valid temperature range for the phase.

> If your check were to be executed when the reaction is added to `Kinetics`, things flow naturally. As things stand now, `Reaction::validate` is called when the object is instantiated. All of this is admittedly subtle - so take it with a grain of salt.

The point where validation is called isn't changed here, is it? Generally, it will be called at the point when a reaction is added to the `Kinetics` object. The only case where it happens when a `Reaction` is first created is when the `getReactions` function is called. Whether that makes sense or not seems orthogonal to this PR.

> But I do have a philosophical question - checking the sticking coefficient here goes in the direction of a mechanism check, which is historically not really done within Cantera but relegated to external tools (e.g. ChemCheck, see [Cantera/enhancements#42](https://github.com/Cantera/enhancements/issues/42)). In another context, you'll see some warnings (e.g. thermo data with discontinuities), but that's about that.
>
> Throwing an error means that a (potentially inexperienced) user would have to fix the YAML file in order to proceed. If they just happened to convert a (buggy) CHEMKIN file from some web repository, this would mean that in most cases those users would be stuck and post on the user group, etc..

I think it's possible to make a useful distinction between checking for issues in an input file and a tool for helping figure out what's actually wrong with the input file and how to fix it. I'd say the first part is something that is good to build directly into Cantera. Otherwise, people are going to just blithely use mechanisms with these problems and are not going to do the due diligence of running them through `ChemCheck`.

I do agree that automatically treating this as an error might be a bit extreme, and by analogy with how Cantera treats discontinuous thermo data, it might make sense to have the default behavior here be to just issue a warning. I'm sure we'll still get some posts to the Users' Group, though.

ischoegl:
> I think that this isn't actually the only case where having the`Kinetics` (and by extension, `ThermoPhase`) object present for validating a reaction would be helpful.

and 

> I think it's possible to make a useful distinction between checking for issues in an input file and a tool for helping figure out what's actually wrong with the input file and how to fix it.

:+1: I agree ... 

> The point where validation is called isn't changed here, is it? Generally, it will be called at the point when a reaction is added to the `Kinetics` object. 

This is technically true for the YAML route, but only as things happen in a direct sequence. As you mentioned, this is not universally true: `validate` is typically called from a `Reaction` constructor, which doesn't necessarily require an immediate addition to `Kinetics` for some other creation routes. 

My main point is that I hope to avoid an across-the-board `void validate(const Kinetics& kin)` signature ... I'd much prefer to have one set of checks that do not rely on `Kinetics` (i.e. these would trigger failures at the `Reaction` level), and a second set of checks that require more in-depth information from `Kinetics` (which would fail when those reactions are added to `Kinetics`). Much of this can be differentiated in separate PR's, as this goes well beyond of what was originally proposed here. Having separate `void validate()` and `void validate(const Kinetics& kin)` is all I am asking for at the moment (it would help to have a more meaningful nomenclature *eventually*). 

In other words, I am asking for
```C++
//! @deprecated To be removed after Cantera 2.6
virtual void validate();
```
to be reconsidered.

12Chao:
I can keep the validation function as is, if we all agree on throw a warning instead of an error.

12Chao:
Another way to raise the warning for problematic sticking coefficient at the current temperature is to add 
```
    for (size_t n = 0; n < m_stickingData.size(); n++) {
        const StickData& item = m_stickingData[n];
        if (kf[item.index] > 1) {
            double T = thermo(surfacePhaseIndex()).temperature();
            warn_user("InterfaceKinetics::applyStickingCorrection",
                      "\nSticking coefficient of reaction {} is bigger than 1 at {} K.",
                      m_reactions[item.index]->equation(), T);
        }
        if (item.use_motz_wise) {
            kf[item.index] /= 1 - 0.5 * kf[item.index];
        }
        kf[item.index] *= factors[n] * sqrt(T) * item.multiplier;
```
in `InterfaceKinetics::applyStickingCorrection`, it will check if a forward rate constant is more than 1 before applying the correction. Since `InterfaceKinetics::applyStickingCorrection` is called in `InterfaceKinetics::_update_rates_T`, so it will throw a warning if there is a problem with sticking coefficients every time the `forward_rate_constants` function is called. This way the code modification is the least, but it may affect the code efficiency as it does extra checks when `forward_rate_constants` is called. Any suggestion on which way I should go for would be appreciated! @rwest @bryanwweber @speth @ischoegl

speth:
> Having separate `void validate()` and `void validate(const Kinetics& kin)` is all I am asking for at the moment (it would help to have a more meaningful nomenclature _eventually_).

OK, that makes sense to me. Being able to catch issues that don't need the `Kinetics` object when the `Reaction` is first instantiated would have some benefits.

@12Chao, can you update to reflect this consensus? I hope you still have a branch with the version where you added the `void validate(const Kinetics& kin)` function?

12Chao:
I have updated the code, but I didn't figure out how to write a test to catch the `CanteraWarning` .  I tried to catch is as a `CanteraError`, but it didn't work. Is there another way that I could try?

ischoegl:
@12Chao ... The way I've seen this done is to make Cantera warnings fatal (`Cantera::make_deprecation_warnings_fatal();` in C++, or `ct.make_deprecation_warnings_fatal()` in Python). If this flag is set, warnings are converted to errors and you can test for correct behavior.

*Caveat: This is for deprecation errors and thus is not necessarily the easiest approach ... I guess I responded too fast.*

speth:
Yeah, I don't think we currently have a way of testing warnings issued by the C++ `warn_user` command.

If you want to be able to test the warning message being generated here, I guess you could add a similar global flag like we use for deprecation warnings to temporarily make the warnings into errors and catch that exception -- see the implementation of `Application::warn_deprecated` and the functions related to the `Application::m_fatal_deprecation_warnings` variable as a kind of template.

This is related to https://github.com/Cantera/enhancements/discussions/112, where the feature described there would make it easier to make trap the warning in a Python test.

One short-term option is to just have a test that issues this warning and leave testing whether that is actually the case as a "TODO" item, and we can accept that the fact that the test coverage will show this bit of code being called as good enough for now.

12Chao:
Thanks! I have made a new function to convert `CanteraWarning` into `CanteraError` and used in the test.

include/cantera/base/global.h
```diff
@@ -215,6 +215,9 @@ void warn_user(const std::string& method, const std::string& msg,
 //! @copydoc Application::make_deprecation_warnings_fatal
 void make_deprecation_warnings_fatal();
 
+//! @copydoc Application::make_cantera_warnings_fatal
+void make_cantera_warnings_fatal();
```
speth: I think this function can just be named `make_warnings_fatal`. Likewise for the Python function.

include/cantera/kinetics/InterfaceKinetics.h
```diff
@@ -376,6 +376,8 @@ class InterfaceKinetics : public Kinetics
      */
     int phaseStability(const size_t iphase) const;
 
+    virtual double rxn_enthalpy(const Composition& reactants, const Composition& products);
+
```
speth: This function needs a docstring. I'd also suggest `reactionEnthalpy` as a name that's more consistent with other function names in Cantera's C++ code.

include/cantera/kinetics/Kinetics.h
```diff
@@ -873,6 +873,9 @@ class Kinetics
         m_root = root;
     }
 
+    virtual double rxn_enthalpy(const Composition& reactants, const Composition& products){
+        return 0;
+    };
```
speth: I understand that this function only gets used for `InterfaceKinetics`, but is there a reason for it to be implemented only for that class? Would the implementation for `Kinetics` be any different? Having it return 0 here seems not ideal.

include/cantera/kinetics/Reaction.h
```diff
@@ -54,10 +54,11 @@ class Reaction
     //! where the reaction occurs. Sets the value of #rate_units.
     virtual void calculateRateCoeffUnits(const Kinetics& kin);
 
-    //! Ensure that the rate constant and other parameters for this reaction are
-    //! valid.
     virtual void validate();
 
+    //! New validate function takes kinetics object as the input
+    virtual void validate(Kinetics& kin);
```
speth: For the docstring, I'd suggest something along the lines of "Perform validation checks that need access to a complete `Kinetics` objects, for example to retrieve information about reactant / product species. Also, the existing docstring for the `validate()` method shouldn't be removed.

include/cantera/kinetics/Reaction.h
```diff
@@ -188,7 +189,10 @@ class ElementaryReaction2 : public Reaction
     ElementaryReaction2();
     ElementaryReaction2(const Composition& reactants, const Composition products,
                         const Arrhenius& rate);
+
     virtual void validate();
+    //! New validate function takes kinetics object as the input
+    virtual void validate(Kinetics& kin);
```
speth: You should only add a docstring to the base class's declaration, unless there is something different about this class's implementation that needs to be noted.

interfaces/cython/cantera/utils.pyx
```diff
@@ -67,6 +67,11 @@ def make_deprecation_warnings_fatal():
                             message='.*Cantera.*')  # for warnings in Cython code
     Cxx_make_deprecation_warnings_fatal()
 
+def make_cantera_warnings_fatal():
+    warnings.filterwarnings('error', category=Warning, module='cantera') # for warnings in Python code
+    warnings.filterwarnings('error', category=Warning, message=".*Cantera.*") # for warnings in Cython code
+    Cxx_make_cantera_warnings_fatal()
```
speth: If the call to `Cxx_make_cantera_warnings_fatal` does as it says, what is the effect of the two calls to `filterwarnings`?

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -953,4 +953,23 @@ void InterfaceKinetics::applyStickingCorrection(double T, double* kf)
...
+{
+    for (size_t n = 0; n < nPhases(); n++) {
+        thermo(n).getPartialMolarEnthalpies(m_grt.data() + m_start[n]);
+    }
+    double rxn_deltaH = 0;
+    for (auto iter = products.begin(); iter != products.end(); ++iter) {
+        int k = kineticsSpeciesIndex(iter->first);
+        double stoich = iter->second;
+        rxn_deltaH += m_grt.data()[k] * stoich;
+    }
```
speth: A few suggestions here, which also apply to the loop below.
```suggestion
    for (const auto& product : products) {
        size_t k = kineticsSpeciesIndex(product.first);
        double stoich = product.second;
        rxn_deltaH += m_grt[k] * stoich;
    }
```

src/kinetics/Reaction.cpp
```diff
@@ -119,6 +119,42 @@ void Reaction::validate()
     }
 }
 
+void Reaction::validate(Kinetics& kin)
+{
+    if (!allow_nonreactant_orders) {
```
speth: I don't think the `validate(Kinetics&)` functions should duplicate checks done by the `validate()` (no args) functions, since you are keeping the latter functions in place. Anywhere you call the former, you also call the latter, but not necessarily the other way around, which I think was the reason @ischoegl suggested keeping them separate.

src/kinetics/Reaction.cpp
```diff
@@ -883,6 +982,34 @@ void BlowersMaselInterfaceReaction::getParameters(AnyMap& reactionNode) const
     }
 }
 
+void BlowersMaselInterfaceReaction::validate(Kinetics& kin)
+{
+    BlowersMaselReaction::validate(kin);
+    if (is_sticking_coefficient) {
+        doublereal original_T = kin.thermo().temperature();
+        doublereal original_P = kin.thermo().pressure();
```
speth: Please don't use `doublereal` in any new code -- just write `double`.

src/kinetics/RxnRates.cpp
```diff
@@ -88,6 +90,7 @@ void Arrhenius::getParameters(AnyMap& rateNode, const Units& rate_units) const
     rateNode.setFlowStyle();
 }
 
+
```
speth: Please revert changes that are just changing whitespace unrelated to your actual updates (ideally, by doing a fixup commit so that there isn't a commit adding this blank line and then another removing it).

src/kinetics/RxnRates.cpp
```diff
@@ -5,6 +5,8 @@
 
 #include "cantera/kinetics/RxnRates.h"
 #include "cantera/base/AnyMap.h"
+#include "cantera/base/global.h"
+#include <math.h>
```
speth: Are these includes needed? especially the latter seems suspect in light of #1110.

interfaces/cython/cantera/test/test_kinetics.py
```diff
@@ -466,6 +466,27 @@ def test_pdep_err(self):
...
+                   )
+        try:
+            ct.make_cantera_warnings_fatal()
+            gas = ct.Solution('sticking_coeff_check.yaml')
+            surface = ct.Interface('sticking_coeff_check.yaml', 'Pt_surf', [gas])
+            self.fail('CanteraError not raised')
+        except ct.CanteraError as e:
+            err_msg_list = str(e).splitlines()
+            for msg in err_msg:
+                self.assertIn(msg, err_msg_list)
```
speth: Check some of the other code that uses the `with self.assertRaises` construct for a nicer way of doing this.

include/cantera/kinetics/RxnRates.h
```diff
@@ -212,6 +216,7 @@ class BlowersMasel
         return m_w;
     }
 
+
```
speth: Please avoid the extraneous whitespace changes.

speth (changes requested):
Thanks for the updates, @12Chao. I think we'll find being able to test warning messages handy in some other cases as well.

I have some suggested revisions that I think will help simplify the implementation of the new validation methods.

include/cantera/kinetics/Kinetics.h
```diff
@@ -873,6 +873,7 @@ class Kinetics
         m_root = root;
     }
 
+    virtual double reactionEnthalpy(const Composition& reactants, const Composition& products);
```
ischoegl: I believe this needs a docstring (if this function is needed at all). You may also want to include an empty line below.

include/cantera/kinetics/Reaction.h
```diff
@@ -54,10 +54,14 @@ class Reaction
     //! where the reaction occurs. Sets the value of #rate_units.
     virtual void calculateRateCoeffUnits(const Kinetics& kin);
 
-    //! Ensure that the rate constant and other parameters for this reaction are
-    //! valid.
     virtual void validate();
```
ischoegl, line 68: Per @speth’s comment, please leave the old docstring in place.

interfaces/cython/cantera/examples/kinetics/blowers_masel.py
```diff
@@ -109,6 +109,6 @@ def change_species_enthalpy(gas, species_name, dH):
 plt.plot(deltaH_list, Ea_list)
 plt.xlabel("Enthalpy Change (J/kmol)")
 plt.ylabel("Activation Energy Change (J/kmol)")
-plt.title(r"$E_a$ vs. $\Delta H$ For O+H2<=>H+OH", y=1.1)
+plt.title(r"$E_a$ vs. $\Delta{H}$ For O+H2<=>H+OH", y=1.1)
```
ischoegl: While this fixes the failing tests, @speth mentioned that this will resolve itself eventually (per #1132), so no action is needed here.

src/kinetics/Kinetics.cpp
```diff
@@ -631,4 +633,24 @@ shared_ptr<const Reaction> Kinetics::reaction(size_t i) const
...
+        size_t k = kineticsSpeciesIndex(product.first);
+        double stoich = product.second;
+        rxn_deltaH += hk[k] * stoich;
+    }
+    for (const auto& reactant : reactants) {
+            size_t k = kineticsSpeciesIndex(reactant.first);
+            double stoich = reactant.second;
+            rxn_deltaH -= hk[k] * stoich;
+        }
+    return rxn_deltaH;
```
ischoegl, line 669: Is there a way to use https://github.com/Cantera/cantera/blob/4570f3e6b0c7e964181ffb4ac2486b223cf80cde/include/cantera/kinetics/Kinetics.h#L392-L407
here? I have recently used 
```c++
bulk.getPartialMolarEnthalpies(m_grt.data());
getReactionDelta(m_grt.data(), dH.data());
```
(this used just one phase, with `dH` being a vector with the length of the number of reactions); I *believe* this may extend to multiple phases as long as the partial molar enthalpies are in the correct order. You mainly need to pick out the correct reaction index (while this calculates things for all reactions, it does not require lookup; it's a tradeoff, but as it is called during validation speed may not be as critical).

ischoegl (changes requested):
@12Chao ... thanks! As far as I can tell, you addressed most of @speth's comments. I also appreciate that you left the original `validate` in place, and implemented the `make_warnings_fatal` capability (which I believe will be useful). I noticed one instance where you may be able to leverage existing code as commented on below.

src/kinetics/Reaction.cpp
```diff
@@ -883,6 +903,34 @@ void BlowersMaselInterfaceReaction::getParameters(AnyMap& reactionNode) const
...
+        fmt::memory_buffer err_reactions;
+        doublereal T[] = {200.0, 500.0, 1000.0, 2000.0, 5000.0, 10000.0};
+        size_t length_T = sizeof(T) / sizeof(T[0]);
+        // auto& interface_kin = dynamic_cast<InterfaceKinetics&>(kin);
+        for (size_t i=0; i < length_T; i++) {
+            kin.thermo().setState_TP(T[i], original_P);
+            double rxn_deltaH = kin.reactionEnthalpy(reactants, products);
+            double k = rate.updateRC(log(T[i]), 1/T[i], rxn_deltaH);
+            if (k > 1) {
+                format_to(err_reactions,
```
ischoegl: Please use `fmt_append` here as there are some recent deprecation warnings triggered by `format_to` (see #1122). Also, there are a `doublereal` and a stray comment left a couple of lines above.
(there is a second instance of `format_to` on line 776 further above)

src/kinetics/Kinetics.cpp
12Chao: Thanks for the review, I think I have addressed all the comments except this one. I would rather not change this part if it is not going to significantly speed things up

ischoegl: @12Chao … no problem. I’ll leave the final approval to @speth. From my perspective, what you have here may be redundant code that could be avoided, as the same may be achieved with 5 lines of code within your validation functions. Another option would be to make this a local function within `Reaction.cpp`, as it is not being used elsewhere and I do not think that it is a good fit for `Kinetics`.

interfaces/cython/cantera/examples/kinetics/blowers_masel.py
ischoegl: Rather than undoing the edit, could you just drop the previous commit? (05ab00102089752ab4a728544805c620ce33de6f)

src/kinetics/Kinetics.cpp
speth: I thought that the reason for adding the `deltaEnthalpy` function was because you need to be able to evaluate it for a reaction that isn't already part of the `Kinetics` object. Correct?

ischoegl: @speth … valid point: in this case `getReactionDelta` does not apply. It still doesn’t quite fit with `Kinetics` and I believe it would be better to have it as a helper function within `Reaction`. But it’s not a major sticking point for me.

ischoegl (dismissed):

src/kinetics/Kinetics.cpp
speth: I agree, it does feel like an awkward addition to the Kinetics class. Given that it relies almost entirely on public member functions of `Kinetics` (and the one exception, the use of `m_start`, can be fixed), I can see it more naturally being a member of the `Reaction` class.

speth (approved):

speth:
Thanks, @12Chao. To fix some merge conflicts that arose as a result of #1101 (I think), I ended up having to squash most of the commits here down into one.

PR #1049, opened 2021-05-31
Replace deprecated yaml.safe_load / simplify paths

### LINKED ISSUE #1048: yaml.safe_load broke in CI
**Problem description**

After throwing `DeprecationWarnings` for a while, `yaml.safe_load` finally broke.

**Steps to reproduce**

Run GitHub Actions on a PR.

**Behavior**

The old deprecation warning was:
```
build/python/cantera/test/test_composite.py::TestSolutionSerialization::test_yaml_outunits
  /home/runner/work/cantera/cantera/build/python/cantera/test/test_composite.py:579: PendingDeprecationWarning: 
  safe_load will be removed, use
  
    yaml=YAML(typ='safe', pure=True)
    yaml.load(...)
  
  instead
    generated = yaml.safe_load(infile)
```
The error now is:
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ruamel.yaml.loader.SafeLoader object at 0x7f536a988208>, block = True
indentless_sequence = True

    def parse_node(self, block=False, indentless_sequence=False):
        # type: (bool, bool) -> Any
        if self.scanner.check_token(AliasToken):
            token = self.scanner.get_token()
            event = AliasEvent(token.value, token.start_mark, token.end_mark)  # type: Any
            self.state = self.states.pop()
            return event
    
        anchor = None
        tag = None
        start_mark = end_mark = tag_mark = None
        if self.scanner.check_token(AnchorToken):
            token = self.scanner.get_token()
            self.move_token_comment(token)
            start_mark = token.start_mark
            end_mark = token.end_mark
            anchor = token.value
            if self.scanner.check_token(TagToken):
                token = self.scanner.get_token()
                tag_mark = token.start_mark
                end_mark = token.end_mark
                tag = token.value
        elif self.scanner.check_token(TagToken):
            token = self.scanner.get_token()
            start_mark = tag_mark = token.start_mark
            end_mark = token.end_mark
            tag = token.value
            if self.scanner.check_token(AnchorToken):
                token = self.scanner.get_token()
                start_mark = tag_mark = token.start_mark
                end_mark = token.end_mark
                anchor = token.value
        if tag is not None:
            handle, suffix = tag
            if handle is not None:
                if handle not in self.tag_handles:
                    raise ParserError(
                        'while parsing a node',
                        start_mark,
                        _F('found undefined tag handle {handle!r}', handle=handle),
                        tag_mark,
                    )
                tag = self.transform_tag(handle, suffix)
            else:
                tag = suffix
        # if tag == '!':
        #     raise ParserError("while parsing a node", start_mark,
        #             "found non-specific tag '!'", tag_mark,
        #      "Please check 'http://pyyaml.org/wiki/YAMLNonSpecificTag'
        #     and share your opinion.")
        if start_mark is None:
            start_mark = end_mark = self.scanner.peek_token().start_mark
        event = None
        implicit = tag is None or tag == '!'
        if indentless_sequence and self.scanner.check_token(BlockEntryToken):
            comment = None
            pt = self.scanner.peek_token()
>           if self.loader and self.loader.comment_handling is None:
E           AttributeError: 'SafeLoader' object has no attribute 'comment_handling'

/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/ruamel/yaml/parser.py:414: AttributeError
___________ ERROR at setup of TestModels.test_restore_thermo_models ____________

cls = <class 'cantera.test.test_composite.TestModels'>

    @classmethod
    def setUpClass(cls):
        utilities.CanteraTest.setUpClass()
        cls.yml_file = pjoin(cls.test_data_dir, "thermo-models.yaml")
        with open(cls.yml_file, 'rt', encoding="utf-8") as stream:
>           cls.yml = yaml.safe_load(stream)

../../build/python/cantera/test/test_composite.py:27: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/ruamel/yaml/main.py:1118: in safe_load
    return load(stream, SafeLoader, version)
/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/ruamel/yaml/main.py:1071: in load
    return loader._constructor.get_single_data()
```

**System information**

- Cantera version: `main`
- OS: GH Actions

**Affected Tests**

```
*****************************
***    Testing Summary    ***
*****************************

Tests passed: 1439
Up-to-date tests skipped: 0
Tests failed: 3
Failed tests:
    - python: test_composite.TestSolutionSerialization.test_yaml_outunits
    - python: test_convert.ck2yamlTest.test_extra
    - python: test_convert.ck2yamlTest.test_sri_zero
*****************************
```

ischoegl:

- Replace deprecated `yaml.safe_load`
- Implement dedicated loader in `test/utilities.py`
- Allow for loading of `Solution` from `Path`
- Transition ~some~ unit tests to `pathlib`

Fixes #1048

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/base.pyx
+++ b/interfaces/cython/cantera/base.pyx
@@ -2,6 +2,7 @@
 # at https://cantera.org/license.txt for license and copyright information.
 
 from collections import defaultdict as _defaultdict
+from pathlib import PurePath
 
 cdef class _SolutionBase:
     def __cinit__(self, infile='', name='', adjacent=(), origin=None,
@@ -53,6 +54,11 @@ cdef class _SolutionBase:
         self.transport = NULL
 
         # Parse inputs
+        if isinstance(infile, PurePath):
+            if infile.name:
+                infile = str(infile)
+            else:
+                infile = ''
         if infile.endswith('.yml') or infile.endswith('.yaml') or yaml:
             self._init_yaml(infile, name, adjacent, yaml)
         elif infile or source:
--- a/interfaces/cython/cantera/base.pyx
+++ b/interfaces/cython/cantera/base.pyx
@@ -55,10 +55,7 @@ cdef class _SolutionBase:
 
         # Parse inputs
         if isinstance(infile, PurePath):
-            if infile.name:
-                infile = str(infile)
-            else:
-                infile = ''
+            infile = str(infile)
         if infile.endswith('.yml') or infile.endswith('.yaml') or yaml:
             self._init_yaml(infile, name, adjacent, yaml)
         elif infile or source:
--- a/interfaces/cython/cantera/test/test_composite.py
+++ b/interfaces/cython/cantera/test/test_composite.py
@@ -6,12 +6,6 @@ import numpy as np
 from collections import OrderedDict
 import warnings
 
-try:
-    import ruamel_yaml as yaml
-except ImportError:
-    from ruamel import yaml
-
-
 import cantera as ct
 from cantera.composite import _h5py, _pandas
 from . import utilities
@@ -23,8 +17,7 @@ class TestModels(utilities.CanteraTest):
     def setUpClass(cls):
         utilities.CanteraTest.setUpClass()
         cls.yml_file = pjoin(cls.test_data_dir, "thermo-models.yaml")
-        with open(cls.yml_file, 'rt', encoding="utf-8") as stream:
-            cls.yml = yaml.safe_load(stream)
+        cls.yml = utilities.load_yaml(cls.yml_file)
 
     def test_load_thermo_models(self):
         for ph in self.yml['phases']:
@@ -478,8 +471,7 @@ class TestSolutionSerialization(utilities.CanteraTest):
         gas.equilibrate('HP')
         gas.TP = 1500, ct.one_atm
         gas.write_yaml('h2o2-generated.yaml')
-        with open('h2o2-generated.yaml', 'r') as infile:
-            generated = yaml.safe_load(infile)
+        generated = utilities.load_yaml("h2o2-generated.yaml")
         for key in ('generator', 'date', 'phases', 'species', 'reactions'):
             self.assertIn(key, generated)
         self.assertEqual(generated['phases'][0]['transport'], 'mixture-averaged')
@@ -503,10 +495,8 @@ class TestSolutionSerialization(utilities.CanteraTest):
         gas.TP = 1500, ct.one_atm
         units = {'length': 'cm', 'quantity': 'mol', 'energy': 'cal'}
         gas.write_yaml('h2o2-generated.yaml', units=units)
-        with open('h2o2-generated.yaml') as infile:
-            generated = yaml.safe_load(infile)
-        with open(pjoin(self.cantera_data, "h2o2.yaml")) as infile:
-            original = yaml.safe_load(infile)
+        generated = utilities.load_yaml("h2o2-generated.yaml")
+        original = utilities.load_yaml(pjoin(self.cantera_data, "h2o2.yaml"))
         self.assertEqual(generated['units'], units)
 
         for r1, r2 in zip(original['reactions'], generated['reactions']):
@@ -529,8 +519,7 @@ class TestSolutionSerialization(utilities.CanteraTest):
         surf.coverages = np.ones(surf.n_species)
         surf.write_yaml('ptcombust-generated.yaml')
 
-        with open('ptcombust-generated.yaml') as infile:
-            generated = yaml.safe_load(infile)
+        generated = utilities.load_yaml("ptcombust-generated.yaml")
         for key in ('phases', 'species', 'gas-reactions', 'Pt_surf-reactions'):
             self.assertIn(key, generated)
         self.assertEqual(len(generated['gas-reactions']), gas.n_reactions)
--- a/interfaces/cython/cantera/test/test_composite.py
+++ b/interfaces/cython/cantera/test/test_composite.py
@@ -1,5 +1,4 @@
 import sys
-from pathlib import Path
 
 import numpy as np
 from collections import OrderedDict
@@ -15,7 +14,7 @@ class TestModels(utilities.CanteraTest):
     @classmethod
     def setUpClass(cls):
         utilities.CanteraTest.setUpClass()
-        cls.yml_file = Path(cls.test_data_dir).joinpath("thermo-models.yaml")
+        cls.yml_file = cls.test_data_path / "thermo-models.yaml"
         cls.yml = utilities.load_yaml(cls.yml_file)
 
     def test_load_thermo_models(self):
@@ -125,13 +124,6 @@ class TestSolutionArrayIO(utilities.CanteraTest):
         utilities.CanteraTest.setUpClass()
         cls.gas = ct.Solution('h2o2.yaml', transport_model=None)
 
-    def tearDown(self):
-        temp_files = ["solutionarray.csv", "solutionarray.h5"]
-        for f in temp_files:
-            fpath = Path(self.test_work_dir).joinpath(f)
-            if fpath.is_file():
-                fpath.unlink()
-
     def test_collect_data(self):
         states = ct.SolutionArray(self.gas)
         collected = states.collect_data(tabular=True)
@@ -149,7 +141,7 @@ class TestSolutionArrayIO(utilities.CanteraTest):
         states.TPX = np.linspace(300, 1000, 7), 2e5, 'H2:0.5, O2:0.4'
         states.equilibrate('HP')
 
-        outfile = Path(self.test_work_dir).joinpath("solutionarray.csv")
+        outfile = self.test_work_path / "solutionarray.csv"
         states.write_csv(outfile)
 
         data = np.genfromtxt(outfile, names=True, delimiter=',')
@@ -166,7 +158,7 @@ class TestSolutionArrayIO(utilities.CanteraTest):
     def test_write_csv_str_column(self):
         states = ct.SolutionArray(self.gas, 3, extra={'spam': 'eggs'})
 
-        outfile = Path(self.test_work_dir).joinpath("solutionarray.csv")
+        outfile = self.test_work_path / "solutionarray.csv"
         states.write_csv(outfile)
 
         b = ct.SolutionArray(self.gas, extra={'spam'})
@@ -176,7 +168,7 @@ class TestSolutionArrayIO(utilities.CanteraTest):
     def test_write_csv_multidim_column(self):
         states = ct.SolutionArray(self.gas, 3, extra={'spam': np.zeros((3, 5,))})
 
-        outfile = Path(self.test_work_dir).joinpath("solutionarray.csv")
+        outfile = self.test_work_path / "solutionarray.csv"
         with self.assertRaisesRegex(NotImplementedError, 'not supported'):
             states.write_csv(outfile)
 
@@ -193,7 +185,9 @@ class TestSolutionArrayIO(utilities.CanteraTest):
 
     @utilities.unittest.skipIf(isinstance(_h5py, ImportError), "h5py is not installed")
     def test_write_hdf(self):
-        outfile = Path(self.test_work_dir).joinpath("solutionarray.h5")
+        outfile = self.test_work_path / "solutionarray.h5"
+        if outfile.is_file():
+            outfile.unlink()
 
         extra = {'foo': range(7), 'bar': range(7)}
         meta = {'spam': 'eggs', 'hello': 'world'}
@@ -236,7 +230,9 @@ class TestSolutionArrayIO(utilities.CanteraTest):
 
     @utilities.unittest.skipIf(isinstance(_h5py, ImportError), "h5py is not installed")
     def test_write_hdf_str_column(self):
-        outfile = Path(self.test_work_dir).joinpath("solutionarray.h5")
+        outfile = self.test_work_path / "solutionarray.h5"
+        if outfile.is_file():
+            outfile.unlink()
 
         states = ct.SolutionArray(self.gas, 3, extra={'spam': 'eggs'})
         states.write_hdf(outfile, mode='w')
@@ -247,7 +243,9 @@ class TestSolutionArrayIO(utilities.CanteraTest):
 
     @utilities.unittest.skipIf(isinstance(_h5py, ImportError), "h5py is not installed")
     def test_write_hdf_multidim_column(self):
-        outfile = Path(self.test_work_dir).joinpath("solutionarray.h5")
+        outfile = self.test_work_path / "solutionarray.h5"
+        if outfile.is_file():
+            outfile.unlink()
 
         states = ct.SolutionArray(self.gas, 3, extra={'spam': [[1, 2], [3, 4], [5, 6]]})
         states.write_hdf(outfile, mode='w')
@@ -496,7 +494,7 @@ class TestSolutionSerialization(utilities.CanteraTest):
         units = {'length': 'cm', 'quantity': 'mol', 'energy': 'cal'}
         gas.write_yaml('h2o2-generated.yaml', units=units)
         generated = utilities.load_yaml("h2o2-generated.yaml")
-        original = utilities.load_yaml(Path(self.cantera_data).joinpath("h2o2.yaml"))
+        original = utilities.load_yaml(self.cantera_data_path / "h2o2.yaml")
         self.assertEqual(generated['units'], units)
 
         for r1, r2 in zip(original['reactions'], generated['reactions']):
--- a/interfaces/cython/cantera/test/test_composite.py
+++ b/interfaces/cython/cantera/test/test_composite.py
@@ -186,6 +186,7 @@ class TestSolutionArrayIO(utilities.CanteraTest):
     @utilities.unittest.skipIf(isinstance(_h5py, ImportError), "h5py is not installed")
     def test_write_hdf(self):
         outfile = self.test_work_path / "solutionarray.h5"
+        # In Python >= 3.8, this can be replaced by the missing_ok argument
         if outfile.is_file():
             outfile.unlink()
 
@@ -231,6 +232,7 @@ class TestSolutionArrayIO(utilities.CanteraTest):
     @utilities.unittest.skipIf(isinstance(_h5py, ImportError), "h5py is not installed")
     def test_write_hdf_str_column(self):
         outfile = self.test_work_path / "solutionarray.h5"
+        # In Python >= 3.8, this can be replaced by the missing_ok argument
         if outfile.is_file():
             outfile.unlink()
 
@@ -244,6 +246,8 @@ class TestSolutionArrayIO(utilities.CanteraTest):
     @utilities.unittest.skipIf(isinstance(_h5py, ImportError), "h5py is not installed")
     def test_write_hdf_multidim_column(self):
         outfile = self.test_work_path / "solutionarray.h5"
+        # In Python >= 3.8, this can be replaced by the missing_ok argument
+
         if outfile.is_file():
             outfile.unlink()
 
--- a/interfaces/cython/cantera/test/test_composite.py
+++ b/interfaces/cython/cantera/test/test_composite.py
@@ -2,7 +2,6 @@ import sys
 
 import numpy as np
 from collections import OrderedDict
-import warnings
 
 import cantera as ct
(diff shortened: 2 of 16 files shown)


ischoegl:
Fixed a couple of `"` that used to be `'`

interfaces/cython/cantera/base.pyx
```diff
@@ -53,6 +54,11 @@ cdef class _SolutionBase:
         self.transport = NULL
 
         # Parse inputs
+        if isinstance(infile, Path):
```
bryanwweber: I think this should be `pathlib.PurePath`, which is at the root of the inheritance tree.

ischoegl:
Ended up removing references to `os.path` in favor of `pathlib` for the Python test suite.

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -22,9 +14,8 @@ class TestModels(utilities.CanteraTest):
     @classmethod
     def setUpClass(cls):
         utilities.CanteraTest.setUpClass()
-        cls.yml_file = pjoin(cls.test_data_dir, "thermo-models.yaml")
-        with open(cls.yml_file, 'rt', encoding="utf-8") as stream:
-            cls.yml = yaml.safe_load(stream)
+        cls.yml_file = cls.test_data_path / "thermo-models.yaml"
+        cls.yml = utilities.load_yaml(cls.yml_file)
```
bryanwweber, line 17: I'm wondering when it makes sense to turn `cls.test_data_dir` into a `Path`, I think we were waiting to drop 3.5 support because 3.6 was the first version to support `os.Pathlike` everywhere.

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -133,6 +124,13 @@ def setUpClass(cls):
         utilities.CanteraTest.setUpClass()
         cls.gas = ct.Solution('h2o2.yaml', transport_model=None)
 
+    def tearDown(self):
+        temp_files = ["solutionarray.csv", "solutionarray.h5"]
```
bryanwweber: I think you can simplify this as:

```python
for p in self.test_work_path.glob("solutionarray.*"):
    p.unlink()
```

assuming that `self.test_work_path` is a `PurePath` instance.

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -150,7 +148,7 @@ def test_write_csv(self):
         states.TPX = np.linspace(300, 1000, 7), 2e5, 'H2:0.5, O2:0.4'
         states.equilibrate('HP')
 
-        outfile = pjoin(self.test_work_dir, 'solutionarray.csv')
+        outfile = self.test_work_path / "solutionarray.csv"
```
bryanwweber, line 143: I go back and forth about using the `/` operator to combine paths. On one hand, it's very short. On the other hand, `joinpath()` is a little clearer for anyone less familiar with `pathlib`. 🤷

interfaces/cython/cantera/test/test_kinetics.py
```diff
@@ -925,17 +922,15 @@ def test_listFromFile(self):
 
     def test_listFromCti(self):
         gas = ct.Solution("h2o2.xml", transport_model=None)
-        p = os.path.dirname(__file__)
-        with open(pjoin(p, '..', 'data', 'h2o2.cti')) as f:
+        with open(self.cantera_data_path / "h2o2.cti") as f:
             R = ct.Reaction.listFromCti(f.read())
```
bryanwweber: I think you can use `Path.read_text()` here, if I'm interpreting the documentation of `read()` correctly.

interfaces/cython/cantera/test/test_onedim.py
```diff
@@ -510,9 +508,9 @@ def test_save_restore(self):
 
         self.create_sim(p, Tin, reactants)
         self.solve_fixed_T()
-        filename = pjoin(self.test_work_dir, 'onedim-fixed-T.xml')
-        if os.path.exists(filename):
-            os.remove(filename)
+        filename = self.test_work_path / "onedim-fixed-T.xml"
+        if filename.is_file():
+            filename.unlink()
```
bryanwweber, line 514:
```suggestion
        # In Python >= 3.8, this can be replaced by the missing_ok argument
        if filename.is_file():
            filename.unlink()
```
So that we have a note that this can be simplified in the future

interfaces/cython/cantera/test/test_reactor.py
```diff
@@ -1516,7 +1516,7 @@ class CombustorTestImplementation:
     """
 
     def setUp(self):
-        self.referenceFile = pjoin(os.path.dirname(__file__), 'data', 'CombustorTest-integrateWithAdvance.csv')
+        self.referenceFile = Path(__file__).parent / "data" / "CombustorTest-integrateWithAdvance.csv"
```
bryanwweber: IMO in cases where more than one path is being added, `joinpath()` is clearer.

bryanwweber: But isn't this the `test_data_dir`?

interfaces/cython/cantera/test/utilities.py
```diff
@@ -6,47 +6,66 @@
...
-            except OSError as e:
-                if e.errno == errno.EEXIST:
-                    pass
-                elif e.errno == errno.EACCES:
-                    cls.test_work_dir = tempfile.mkdtemp()
-                else:
-                    raise
+                cls.test_work_path.mkdir()
+            except FileExistsError:
+                pass
```
bryanwweber: I think you can use the `exist_ok=True` argument here to simplify this.

interfaces/cython/cantera/test/utilities.py
```diff
@@ -6,47 +6,66 @@
...
+        cls.cantera_data = str(cls.cantera_data_path)
+
 
     @classmethod
     def tearDownClass(cls):
         # Remove the working directory after testing, but only if its a temp directory
         if getattr(cls, "using_tempfile", False):
             try:
-                shutil.rmtree(cls.test_work_dir)
+                shutil.rmtree(str(cls.test_work_path))
```
bryanwweber: I don't think this has to be cast to a string, but please correct me if that's wrong 😄

interfaces/cython/cantera/utils.pyx
```diff
@@ -23,7 +24,8 @@ cdef pystr(string x):
 
 def add_directory(directory):
     """ Add a directory to search for Cantera data files. """
-    CxxAddDirectory(stringify(directory))
+    directory = Path(directory)
+    CxxAddDirectory(stringify(str(directory)))
```
bryanwweber: What's the purpose in casting to a `Path` and then immediately back to a string?

interfaces/cython/cantera/base.pyx
bryanwweber: Likewise elsewhere.

interfaces/cython/cantera/test/test_composite.py
bryanwweber: I made the comment above before the changes were finished, so I'm not sure if it applies. Feel free to ignore/resolve if it's done now.

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -478,26 +470,24 @@ def test_extra(self):
...
-        with open(output, 'rt', encoding="utf-8") as stream:
-            yml = yaml.safe_load(stream)
+        output = self.test_work_path / ("gri30_extra" + self.ext)
+        yml = utilities.load_yaml(output)
 
         desc = yml['description'].split('\n')[-1]
         self.assertEqual(desc, 'This is an alternative description.')
         for key in ['foo', 'bar']:
             self.assertIn(key, yml.keys())
         # This test tests it can convert the SRI parameters when D or E equal to 0
```
bryanwweber: I think this comment is too far indented, and I think it should be a docstring rather than a comment.

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -579,7 +569,7 @@ def test_long_source_input(self):
 
         gas = ct.Solution('pdep-test.yaml')
 
-        with open(pjoin(self.test_data_dir, 'pdep-test.cti'), 'r') as f:
+        with open(self.test_data_path / "pdep-test.cti", "r") as f:
             data = f.read()
```
bryanwweber: I think you can replace this with `read_text()`.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1113,8 +1110,7 @@ def test_fromCti(self):
 
     def test_fromXml(self):
         import xml.etree.ElementTree as ET
-        p = os.path.dirname(__file__)
-        root = ET.parse(pjoin(p, '..', 'data', 'h2o2.xml')).getroot()
+        root = ET.parse(str(self.cantera_data_path / "h2o2.xml")).getroot()
```
bryanwweber: Elsewhere, you didn't cast the argument to a string, so I don't think you have to here either.

interfaces/cython/cantera/test/test_kinetics.py
bryanwweber: This occurs in several places.

interfaces/cython/cantera/test/utilities.py
```diff
@@ -6,47 +6,66 @@
...
-                else:
-                    raise
+                cls.test_work_path.mkdir()
+            except FileExistsError:
+                pass
+            except FileNotFoundError:
+                cls.test_work_path = Path(tempfile.mkdtemp())
+                cls.using_tempfile = True
+            except:
+                raise
```
bryanwweber: This isn't needed, if the exception doesn't match it will automatically be raised.

bryanwweber (changes requested):
Thanks for taking this on @ischoegl, I've been wanting to switch more stuff to `pathlib` for a while!

interfaces/cython/cantera/test/test_composite.py
ischoegl: I actually just changed this one to deleting the old files at the last possible moment (to make sure that files exist after running a single test). I.e. `tearDown` is no longer used.

interfaces/cython/cantera/test/test_composite.py
ischoegl: Sorry for incrementally updating things: I left `cls.test_data_dir` as there would be a log of churn in `test_convert.py`. At this point, `CONTRIBUTING.md` assumes Python 3.6, so I believe turning everything into paths would be ok otherwise. Let me know whether to apply the changes ... which would add another 100 or so lines to thie PR.

interfaces/cython/cantera/test/test_reactor.py
ischoegl: It is, but it doesn't inherit from `utilities.CanteraTest` ... (unfortunately)

interfaces/cython/cantera/test/test_composite.py
ischoegl: I (mostly) left the `joinpath` where it existed, but am otherwise partial to the shorter `/` ... it's neat as it resembles the command line (and thus should be intuitive).

interfaces/cython/cantera/test/utilities.py
ischoegl: True, should have been introduced in 3.5.

interfaces/cython/cantera/test/test_composite.py
bryanwweber: I say let's just go for the total change, no reason to leave old code lying around.

interfaces/cython/cantera/utils.pyx
ischoegl: That way, you can add a `Path` in `cantera.add_directory` ... see `__init__.py`.

bryanwweber: But if it's passed in as a `Path`, you don't need to convert it to a `Path`, and `str` on a string is a no-op, right?

ischoegl: `stringify` on a `Path` raises an error, hence the need to convert `Path` to `str`

bryanwweber: No, I get that. The question is about why `directory = Path(directory)`.

interfaces/cython/cantera/test/test_convert.py
ischoegl: Looks like it is the docstring for the next test - this was not changed, but I think it's worth fixing.

interfaces/cython/cantera/test/test_thermo.py
ischoegl: True. wasn't sure that `ElementTree` took `Path` as an argument (but you're correct, it there is another instance that doesn't cast).

interfaces/cython/cantera/test/utilities.py
ischoegl: Replaced it with a solution that only uses `Path`.

interfaces/cython/cantera/test/test_composite.py
ischoegl: Per discussion, I am consistently using the `/` operator for brevity.

interfaces/cython/cantera/utils.pyx
ischoegl: I guess a simple `directory = str(directory)` would suffice. Ugh!

ischoegl:
@bryanwweber ... replacing the remaining path strings created some more churn (as expected), but now everything is consistent.

interfaces/cython/cantera/test/test_reactor.py
ischoegl: I added a module constant `utilities.TEST_DATA_PATH` that is also assigned to `utilities.CanteraTest.test_data_path`. I am using the latter throughout, as it's consistent with the original `utilities.CanteraTest.test_data_dir`.

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -478,26 +471,24 @@ def test_extra(self):
...
 
         desc = yml['description'].split('\n')[-1]
         self.assertEqual(desc, 'This is an alternative description.')
         for key in ['foo', 'bar']:
             self.assertIn(key, yml.keys())
-        # This test tests it can convert the SRI parameters when D or E equal to 0
-   
+
     def test_sri_zero(self):
+        # This test tests it can convert the SRI parameters when D or E equal to 0
```
bryanwweber, line 483: Can you make this a docstring (in triple quotes)? IIRC, pytest is able to display docstrings in certain modes.

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -517,8 +508,8 @@ def test_duplicate_reactions(self):
 
         with self.assertRaises(SystemExit):
             ck2yaml.main([
-                '--input={}/undeclared-duplicate-reactions.inp'.format(self.test_data_dir),
-                '--thermo={}/dummy-thermo.dat'.format(self.test_data_dir)])
+                '--input={}/undeclared-duplicate-reactions.inp'.format(self.test_data_path),
+                '--thermo={}/dummy-thermo.dat'.format(self.test_data_path)])
```
bryanwweber: May as well use `f` strings here :smile:

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -674,8 +663,10 @@ def checkTransport(self, ctiPhase, yamlPhase, temperatures,
 
     @utilities.slow_test
     def test_gri30(self):
-        cti2yaml.convert(Path(self.cantera_data).joinpath('gri30.cti'),
-                         Path(self.test_work_dir).joinpath('gri30.yaml'))
+        cti2yaml.convert(
+            self.cantera_data_path / "gri30.cti",
+            self.test_work_path / "gri30.yaml"
```
bryanwweber:
```suggestion
            self.test_work_path / "gri30.yaml",
```

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -685,15 +676,19 @@ def test_gri30(self):
         self.checkTransport(ctiPhase, yamlPhase, [298, 1001, 2400])
 
     def test_pdep(self):
-        cti2yaml.convert(Path(self.test_data_dir).joinpath('pdep-test.cti'),
-                         Path(self.test_work_dir).joinpath('pdep-test.yaml'))
+        cti2yaml.convert(
+            self.test_data_path / "pdep-test.cti",
+            self.test_work_path / "pdep-test.yaml"
```
bryanwweber:
```suggestion
            self.test_work_path / "pdep-test.yaml",
```

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -685,15 +676,19 @@ def test_gri30(self):
...
         ctiPhase, yamlPhase = self.checkConversion('pdep-test')
         self.checkKinetics(ctiPhase, yamlPhase, [300, 1000, 2200],
                            [100, ct.one_atm, 2e5, 2e6, 9.9e6])
 
     def test_ptcombust(self):
-        cti2yaml.convert(Path(self.cantera_data).joinpath('ptcombust.cti'),
-                         Path(self.test_work_dir).joinpath('ptcombust.yaml'))
+        cti2yaml.convert(
+            self.cantera_data_path / "ptcombust.cti",
+            self.test_work_path / "ptcombust.yaml"
```
bryanwweber:
```suggestion
            self.test_work_path / "ptcombust.yaml",
```

bryanwweber: And likewise elsewhere, since the closing paren is on the next line.

interfaces/cython/cantera/test/utilities.py
```diff
@@ -117,12 +129,12 @@ def compareProfiles(reference, sample, rtol=1e-5, atol=1e-12, xtol=1e-5):
...
     fails, a formatted report of the differing elements is returned.
     """
-    if isinstance(reference, str):
+    if isinstance(reference, (str, Path)):
         reference = np.genfromtxt(reference, delimiter=',').T
     else:
         reference = np.asarray(reference).T
 
-    if isinstance(sample, str):
+    if isinstance(sample, (str, Path)):
```
bryanwweber: I think these should both be `PurePath` as well.

bryanwweber (changes requested):
Just a few minor changes from the updates. Thanks again @ischoegl!

interfaces/cython/cantera/test/test_convert.py
ischoegl: I am hesitant, as I have seen docstrings replace the actual stack information (i.e. what file / test, which imho is not ideal). Also, I didn't really touch this line other than moving it to where it belongs.

interfaces/cython/cantera/test/test_convert.py
ischoegl: Had already added this in a couple of instances, but absolutely!

interfaces/cython/cantera/test/test_convert.py
ischoegl: Sure. FYI, first `f` strings I've ever typed! :tada:

interfaces/cython/cantera/test/utilities.py
bryanwweber: This seems like it got missed along the way.

ischoegl: Not sure - from what I can tell, these are already changed to `PurePath`?

bryanwweber: Whoops thanks!

interfaces/cython/cantera/test/test_convert.py
ischoegl: Marking as resolved as docstrings for unit tests is a larger discussion (that goes beyond the scope here: the comment belongs to a test that was not modified in this PR)

ischoegl:
@bryanwweber ... I believe everything is addressed. Unless there's anything else, this may be ready for a merge.

bryanwweber (approved):

ischoegl:
Thanks @bryanwweber!

PR #1052, opened 2021-06-09
[SCons] Fix type reference and missing import

bryanwweber:
In #1012, normapth was no longer exported from buildutils, but I missed
adding that import in the builder for python_minimal. In addition, Dict
is only available while type checking in buildutils, so it should be
quoted as a forward reference.

**Checklist**

### DIFF
--- a/interfaces/python_minimal/SConscript
+++ b/interfaces/python_minimal/SConscript
@@ -1,5 +1,5 @@
 """Minimal Python Module"""
-from os.path import join as pjoin
+from os.path import join as pjoin, normpath
 from buildutils import *
 
 Import('env', 'build', 'install')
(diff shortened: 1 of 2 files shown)


ischoegl:
See [bug report on UG](https://groups.google.com/g/cantera-users/c/iLEUj8A-u8E).

speth:
I'm confused about the issue with `Dict` - why doesn't this show up as an error all the time?

bryanwweber:
> I'm confused about the issue with `Dict` - why doesn't this show up as an error all the time?

@speth `ourSpawn()` is only called when an alternative compiler (that is, not VC++) is used on Windows. I guess since the type annotations are only evaluated at runtime, it's a `NameError` and not a `SyntaxError`, and it only comes up when the function is actually called.

speth (approved):

Issue #1053, opened 2021-06-09
Paper cuts in YAML serialization API


ischoegl:
**Problem description**

After the introduction of new YAML features (especially in Python), there has been some discussion of minor 'paper cuts'. Below is a list of issues that could/should be addressed:

* [x] Unpythonic method names `Reaction.fromYaml`, `Reaction.listFromFile`, `Reaction.listFromYaml` ... *fix proposed in #1061* (now merged)
* [x] Similar for `Species.fromYaml`, etc. … *fix proposed in #1129* (now merged)
* [x] Per comment in #1031, it would be nice to have `Solution.write_yaml` be able to write into a string … *added to #692* (now merged)
* [x] ~It would be nice to have constructors that use keyword names that correspond to YAML field names. This is somewhat complicated by the occurrence of hyphens in many top-level fields (e.g. `rate-constant` for reaction definitions, see #995);~ an alternative `@classmethod` `from_dict` is an approach that does not have this limitation.
* [ ] With the introduction of `dict_to_anymap` in #1014, it is now possible to pass dictionaries that are equivalent to YAML definitions to the Python API. In #995, a new `Reaction.from_dict` is proposed (now merged). The same is possible for various other imports. ... *fix proposed in #1179*
* [x] ~With #984, the `input_data` field was introduced, which corresponds to the dictionary that can be used by `from_dict`. A consistent naming scheme may be beneficial, especially as the field does not preserve the original parameterization (see #1031).~ ... probably not an issue
* [x] ~It would be nice to allow for mixed `dict`/YAML string imports for `dict_to_anymap` (see #995).~ ... *implementation proposed in #1179*, but dropped 
* [x] ~At this point, YAML may refer to a file, string, or even a corresponding Python dictionary, where the context isn't always clear.~  ... probably not an issue
* [x] Root level information of the YAML parameterization is not accessible, see also #1013 … *fix proposed in #1129* (now merged)
* [ ] Per comments in #1031 and #984, there is some inconsistency of C++ definitions for `input`, which either refers to a member variable or a class method. As this was initially introduced in #584 (and/or related PR's), this part of the API is already released in 2.5.1 and thus is difficult to address. ... *not sure that this should be addressed*
* [x] There are some imperfections for YAML emitted output, see #1128 … *fixed in #1133*

**Additional context**

Some of the unpythonic method names are based on historic equivalents for XML and CTI. As those are to be deprecated, a renaming of legacy methods appears to be unwarranted.

PR #1061, opened 2021-06-14
Reaction rate factories

### LINKED ISSUE #1071: Inconsistency in ReactionRateCoeff units calculation
**Problem description**

Since the introduction of YAML, reaction constructors ensure that units are provided in a consistent manner. Units are, however, not consistently handled when constructing reactions from scratch, i.e. 
```C++
ElementaryReaction3(const Composition& reactants, const Composition& products, const ArrheniusRate& rate);
```
or
```C++
ElementaryReaction2(const Composition& reactants_, const Composition products_, const Arrhenius& rate_);
```
as well as all other corresponding constructor instances.

While it is (at least for the newer framework) possible to associate units with `ReactionRate` objects, there never is a call to `Reaction::calculateRateCoeffUnits` that would otherwise handle units.

Per @speth's comments ...

> the units for the rate constant are a combination of the reacting phase standard concentration units and the concentration units for the reactants, as in `Reaction::calculateRateCoeffUnits`.

**Steps to reproduce**

N/A ... although these C++ constructors are widely used in the C++ test suite.

**Behavior**

Units are never calculated (no call to `Reaction::calculateRateCoeffUnits`).

**System information**

- Cantera version: current development version
- OS: all
- Python/MATLAB/other software versions: N/A (it is part of the C++ core)

**Additional context**

This issue surfaced when discussing PR #1061

ischoegl:

Create factory constructors in anticipation of refactoring of `Falloff` reactions and corresponding rates

- [x] Implement C++ `ReactionRateBase::newMultiRate` instantiation
- [x] Implement C++ `newRate` factory constructor
- [x] Generalize Python `ReactionRate` interface

The Interface now mostly replicates what already exists for `Reaction` objects. Addresses Cantera/enhancements#87.

Closes #1071 (associated discussion)

**Example**

YAML input for `ReactionRate` uses the same information as `Reaction`.  Non-rate specific information is ignored, and rate types are recognized using aliases. 
```
In [1]: import cantera as ct
   ...:  gas = ct.Solution('kineticsfromscratch.yaml')

In [2]: yaml = """
   ...:     equation: H2 + O2 <=> 2 OH
   ...:     type: pressure-dependent-Arrhenius
   ...:     rate-constants:
   ...:     - {P: 0.01 atm, A: 1.2124e+16, b: -0.5779, Ea: 1.08727e+04 cal/mol}
   ...:     - {P: 1.0 atm, A: 4.9108e+31, b: -4.8507, Ea: 2.47728e+04 cal/mol}
   ...:     - {P: 10.0 atm, A: 1.2866e+47, b: -9.0246, Ea: 3.97965e+04 cal/mol}
   ...:     - {P: 100.0 atm, A: 5.9632e+56, b: -11.529, Ea: 5.25996e+04 cal/mol}
   ...:     """

In [3]: rr = ct.ReactionRate.fromYaml(yaml, gas)
   ...: rr
Out[3]: <PlogRate at 7f73992fa060>

In [4]: dd = rr.input_data
   ...: dd
Out[4]: 
{'rate-constants': [{'A': 1.2124e+16,
   'Ea': 45491376.80000001,
   'P': 1013.2500000000003,
   'b': -0.5779},
  {'A': 4.9108e+31, 'Ea': 103649395.2, 'P': 101324.99999999999, 'b': -4.8507},
  {'A': 1.2866e+47, 'Ea': 166508556.0, 'P': 1013249.9999999992, 'b': -9.0246},
  {'A': 5.9632e+56, 'Ea': 220076726.4, 'P': 10132499.999999985, 'b': -11.529}],
 'type': 'PlogRate'}

In [5]: ct.ReactionRate.from_dict(dd, gas)
Out[5]: <PlogRate at 7f73ce3d7660>
```

For the time being, I am sticking with the usual nomenclature (see #1053)

**Checklist**

### DIFF
--- a/include/cantera/kinetics/RateFactory.h
+++ b/include/cantera/kinetics/RateFactory.h
@@ -66,7 +66,7 @@ private:
 
 //! Create a new empty ReactionRateBase object
 /*!
- * @param type string identifying type of reaction.
+ * @param type string identifying type of reaction rate.
  */
 shared_ptr<ReactionRateBase> newRate(const std::string& type);
 
--- a/include/cantera/kinetics/RateFactory.h
+++ b/include/cantera/kinetics/RateFactory.h
@@ -1,0 +1,88 @@
+/**
+ *  @file RateFactory.h
+ *  Factory class for reaction rate objects. Used by classes
+ *  that implement kinetics
+ *  (see \ref reactionGroup and class \link Cantera::Rate Rate\endlink).
+ */
+
+// This file is part of Cantera. See License.txt in the top-level directory or
+// at https://cantera.org/license.txt for license and copyright information.
+
+#ifndef CT_NEWRATE_H
+#define CT_NEWRATE_H
+
+#include "cantera/base/FactoryBase.h"
+#include "cantera/kinetics/ReactionRate.h"
+
+namespace Cantera
+{
+
+class Kinetics;
+class Units;
+
+/**
+ * Factory class to construct reaction rate calculators.
+ * The reaction factory is accessed through the static method factory:
+ *
+ * @code
+ * Rate* f = RateFactory::factory()->newRate(type, c)
+ * @endcode
+ *
+ * @ingroup reactionGroup
+ */
+class RateFactory : public Factory<ReactionRateBase, const AnyMap&, const Units&>
+{
+public:
+    /**
+     * Return a pointer to the factory. On the first call, a new instance is
+     * created. Since there is no need to instantiate more than one factory,
+     * on all subsequent calls, a pointer to the existing factory is returned.
+     */
+    static RateFactory* factory() {
+        std::unique_lock<std::mutex> lock(rate_mutex);
+        if (!s_factory) {
+            s_factory = new RateFactory;
+        }
+        return s_factory;
+    }
+
+    virtual void deleteFactory() {
+        std::unique_lock<std::mutex> lock(rate_mutex);
+        delete s_factory;
+        s_factory = 0;
+    }
+
+private:
+    //! Pointer to the single instance of the factory
+    static RateFactory* s_factory;
+
+    //! default constructor, which is defined as private
+    RateFactory();
+
+    //!  Mutex for use when calling the factory
+    static std::mutex rate_mutex;
+};
+
+
+//! Create a new empty ReactionRateBase object
+/*!
+ * @param type string identifying type of reaction.
+ */
+shared_ptr<ReactionRateBase> newRate(const std::string& type);
+
+//! Create a new Rate object using the specified parameters
+/*!
+ * @param rate_node AnyMap node describing reaction rate.
+ * @param rate_units Unit system of the reaction rate
+ */
+shared_ptr<ReactionRateBase> newRate(const AnyMap& rate_node, const Units& rate_units);
+
+//! Create a new Rate object using the specified parameters
+/*!
+ * @param rate_node AnyMap node describing reaction rate.
+ * @param kin kinetics manager
+ */
+shared_ptr<ReactionRateBase> newRate(const AnyMap& rate_node, const Kinetics& kin);
+
+}
+#endif
--- a/include/cantera/kinetics/RateFactory.h
+++ b/include/cantera/kinetics/RateFactory.h
@@ -84,5 +84,11 @@ shared_ptr<ReactionRateBase> newRate(const AnyMap& rate_node, const Units& rate_
  */
 shared_ptr<ReactionRateBase> newRate(const AnyMap& rate_node, const Kinetics& kin);
 
+//! Retrieve the canoncial rate object name
+/*!
+ * @param type string identifying alternate reaction rate name.
+ */
+std::string canonicalRateName(const std::string& type);
+
 }
 #endif
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -341,9 +341,9 @@ cdef extern from "cantera/kinetics/ReactionFactory.h" namespace "Cantera":
     cdef shared_ptr[CxxReaction] CxxNewReaction "newReaction" (XML_Node&) except +translate_exception
     cdef shared_ptr[CxxReaction] CxxNewReaction "newReaction" (CxxAnyMap&, CxxKinetics&) except +translate_exception
 
-cdef extern from "cantera/kinetics/RateFactory.h" namespace "Cantera":
-    cdef shared_ptr[CxxReactionRateBase] CxxNewRate "newRate" (string) except +translate_exception
-    cdef shared_ptr[CxxReactionRateBase] CxxNewRate "newRate" (CxxAnyMap&, CxxKinetics&) except +translate_exception
+cdef extern from "cantera/kinetics/ReactionRateFactory.h" namespace "Cantera":
+    cdef shared_ptr[CxxReactionRateBase] CxxNewReactionRate "newReactionRate" (string) except +translate_exception
+    cdef shared_ptr[CxxReactionRateBase] CxxNewReactionRate "newReactionRate" (CxxAnyMap&, CxxKinetics&) except +translate_exception
 
 cdef extern from "cantera/kinetics/Reaction.h" namespace "Cantera":
     cdef vector[shared_ptr[CxxReaction]] CxxGetReactions "getReactions" (XML_Node&) except +translate_exception
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -1157,7 +1157,7 @@ cdef class ReactionRate:
     cdef wrap(shared_ptr[CxxReactionRateBase])
 
 cdef class CustomRate(ReactionRate):
-    cdef CxxCustomFunc1Rate* cfr(self)
+    cdef CxxCustomFunc1Rate* cxx_object(self)
     cdef Func1 _rate_func
 
 cdef class Reaction:
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -343,7 +343,7 @@ cdef extern from "cantera/kinetics/ReactionFactory.h" namespace "Cantera":
 
 cdef extern from "cantera/kinetics/ReactionRateFactory.h" namespace "Cantera":
     cdef shared_ptr[CxxReactionRateBase] CxxNewReactionRate "newReactionRate" (string) except +translate_exception
-    cdef shared_ptr[CxxReactionRateBase] CxxNewReactionRate "newReactionRate" (CxxAnyMap&, CxxKinetics&) except +translate_exception
+    cdef shared_ptr[CxxReactionRateBase] CxxNewReactionRate "newReactionRate" (CxxAnyMap&) except +translate_exception
 
 cdef extern from "cantera/kinetics/Reaction.h" namespace "Cantera":
     cdef vector[shared_ptr[CxxReaction]] CxxGetReactions "getReactions" (XML_Node&) except +translate_exception
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -342,7 +342,7 @@ cdef extern from "cantera/kinetics/ReactionFactory.h" namespace "Cantera":
     cdef shared_ptr[CxxReaction] CxxNewReaction "newReaction" (CxxAnyMap&, CxxKinetics&) except +translate_exception
 
 cdef extern from "cantera/kinetics/RateFactory.h" namespace "Cantera":
-    cdef shared_ptr[CxxReactionRateBase] CxxNewRate "Cantera::newRate" (string) except +translate_exception
+    cdef shared_ptr[CxxReactionRateBase] CxxNewRate "newRate" (string) except +translate_exception
     cdef shared_ptr[CxxReactionRateBase] CxxNewRate "newRate" (CxxAnyMap&, CxxKinetics&) except +translate_exception
 
 cdef extern from "cantera/kinetics/Reaction.h" namespace "Cantera":
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -341,6 +341,10 @@ cdef extern from "cantera/kinetics/ReactionFactory.h" namespace "Cantera":
     cdef shared_ptr[CxxReaction] CxxNewReaction "newReaction" (XML_Node&) except +translate_exception
     cdef shared_ptr[CxxReaction] CxxNewReaction "newReaction" (CxxAnyMap&, CxxKinetics&) except +translate_exception
 
+cdef extern from "cantera/kinetics/RateFactory.h" namespace "Cantera":
+    cdef shared_ptr[CxxReactionRateBase] CxxNewRate "Cantera::newRate" (string) except +translate_exception
+    cdef shared_ptr[CxxReactionRateBase] CxxNewRate "newRate" (CxxAnyMap&, CxxKinetics&) except +translate_exception
+
 cdef extern from "cantera/kinetics/Reaction.h" namespace "Cantera":
     cdef vector[shared_ptr[CxxReaction]] CxxGetReactions "getReactions" (XML_Node&) except +translate_exception
     cdef vector[shared_ptr[CxxReaction]] CxxGetReactions "getReactions" (CxxAnyValue&, CxxKinetics&) except +translate_exception
@@ -1146,27 +1150,14 @@ cdef class ThermoPhase(_SolutionBase):
 cdef class InterfacePhase(ThermoPhase):
     cdef CxxSurfPhase* surf
 
-cdef class _ReactionRate:
-    cdef shared_ptr[CxxReactionRateBase] _base
-    cdef CxxReactionRateBase* base
-
-cdef class ArrheniusRate(_ReactionRate):
-    cdef CxxArrheniusRate* rate
+cdef class ReactionRate:
+    cdef shared_ptr[CxxReactionRateBase] _rate
+    cdef CxxReactionRateBase* rate
     @staticmethod
     cdef wrap(shared_ptr[CxxReactionRateBase])
 
-cdef class PlogRate(_ReactionRate):
-    cdef CxxPlogRate* rate
-    @staticmethod
-    cdef wrap(shared_ptr[CxxReactionRateBase])
-
-cdef class ChebyshevRate(_ReactionRate):
-    cdef CxxChebyshevRate3* rate
-    @staticmethod
-    cdef wrap(shared_ptr[CxxReactionRateBase])
-
-cdef class CustomRate(_ReactionRate):
-    cdef CxxCustomFunc1Rate* rate
+cdef class CustomRate(ReactionRate):
+    cdef CxxCustomFunc1Rate* cfr(self)
     cdef Func1 _rate_func
 
 cdef class Reaction:
--- a/interfaces/cython/cantera/reaction.pyx
+++ b/interfaces/cython/cantera/reaction.pyx
@@ -63,7 +63,7 @@ cdef class ReactionRate:
         return rr
 
     @classmethod
-    def from_dict(cls, data, Kinetics kinetics=None):
+    def from_dict(cls, data, Kinetics kinetics):
         """
         Create a `ReactionRate` object from a dictionary corresponding to its YAML
         representation.
@@ -86,15 +86,13 @@ cdef class ReactionRate:
             raise TypeError(
                 f"Class method 'from_dict' was invoked from '{cls.__name__}' but "
                 "should be called from base class 'ReactionRate'")
-        if kinetics is None:
-            raise ValueError("A Kinetics object is required.")
 
         cdef CxxAnyMap any_map = dict_to_anymap(data)
         cxx_rate = CxxNewReactionRate(any_map, deref(kinetics.kinetics))
         return ReactionRate.wrap(cxx_rate)
 
     @classmethod
-    def from_yaml(cls, text, Kinetics kinetics=None):
+    def from_yaml(cls, text, Kinetics kinetics):
         """
         Create a `ReactionRate` object from its YAML string representation.
 
@@ -116,8 +114,6 @@ cdef class ReactionRate:
             raise TypeError(
                 f"Class method 'from_yaml' was invoked from '{cls.__name__}' but "
                 "should be called from base class 'ReactionRate'")
-        if kinetics is None:
-            raise ValueError("A Kinetics object is required.")
 
         cdef CxxAnyMap any_map
         any_map = AnyMapFromYamlString(stringify(text))
@@ -172,7 +168,7 @@ cdef class ArrheniusRate(ReactionRate):
                 raise TypeError("Invalid parameters 'A', 'b' or 'Ea'")
             self.rate = self._rate.get()
 
-    cdef CxxArrheniusRate* arr(self):
+    cdef CxxArrheniusRate* cxx_object(self):
         return <CxxArrheniusRate*>self.rate
 
     property pre_exponential_factor:
@@ -181,21 +177,21 @@ cdef class ArrheniusRate(ReactionRate):
         powers depending on the reaction order.
         """
         def __get__(self):
-            return self.arr().preExponentialFactor()
+            return self.cxx_object().preExponentialFactor()
 
     property temperature_exponent:
         """
         The temperature exponent ``b``.
         """
         def __get__(self):
-            return self.arr().temperatureExponent()
+            return self.cxx_object().temperatureExponent()
 
     property activation_energy:
         """
         The activation energy ``E`` [J/kmol].
         """
         def __get__(self):
-            return self.arr().activationEnergy()
+            return self.cxx_object().activationEnergy()
 
     property allow_negative_pre_exponential_factor:
         """
@@ -203,9 +199,9 @@ cdef class ArrheniusRate(ReactionRate):
         pre-exponential factor.
         """
         def __get__(self):
-            return self.arr().allow_negative_pre_exponential_factor
+            return self.cxx_object().allow_negative_pre_exponential_factor
         def __set__(self, allow):
-            self.arr().allow_negative_pre_exponential_factor = allow
+            self.cxx_object().allow_negative_pre_exponential_factor = allow
 
 
 cdef class PlogRate(ReactionRate):
@@ -231,7 +227,7 @@ cdef class PlogRate(ReactionRate):
                 raise TypeError("Invalid parameter 'rates'")
             self.rate = self._rate.get()
 
-    cdef CxxPlogRate* pr(self):
+    cdef CxxPlogRate* cxx_object(self):
         return <CxxPlogRate*>self.rate
 
     property rates:
@@ -241,8 +237,9 @@ cdef class PlogRate(ReactionRate):
         """
         def __get__(self):
             rates = []
-            cdef vector[pair[double, CxxArrhenius]] cxxrates = self.pr().rates()
+            cdef vector[pair[double, CxxArrhenius]] cxxrates
             cdef pair[double, CxxArrhenius] p_rate
+            cxxrates = self.cxx_object().rates()
             for p_rate in cxxrates:
                 rates.append((p_rate.first, copyArrhenius(&p_rate.second)))
             return rates
@@ -302,46 +299,47 @@ cdef class ChebyshevRate(ReactionRate):
         self._rate.reset(new CxxChebyshevRate3(Tmin, Tmax, Pmin, Pmax, data))
         self.rate = self._rate.get()
 
-    cdef CxxChebyshevRate3* chr(self):
+    cdef CxxChebyshevRate3* cxx_object(self):
         return <CxxChebyshevRate3*>self.rate
 
     property Tmin:
         """ Minimum temperature [K] for the Chebyshev fit """
         def __get__(self):
-            return self.chr().Tmin()
+            return self.cxx_object().Tmin()
 
     property Tmax:
         """ Maximum temperature [K] for the Chebyshev fit """
         def __get__(self):
-            return self.chr().Tmax()
+            return self.cxx_object().Tmax()
 
     property Pmin:
         """ Minimum pressure [Pa] for the Chebyshev fit """
         def __get__(self):
-            return self.chr().Pmin()
+            return self.cxx_object().Pmin()
 
     property Pmax:
         """ Maximum pressure [Pa] for the Chebyshev fit """
         def __get__(self):
-            return self.chr().Pmax()
+            return self.cxx_object().Pmax()
 
     property n_pressure:
         """ Number of pressures over which the Chebyshev fit is computed """
         def __get__(self):
-            return self.chr().nPressure()
+            return self.cxx_object().nPressure()
 
     property n_temperature:
         """ Number of temperatures over which the Chebyshev fit is computed """
         def __get__(self):
-            return self.chr().nTemperature()
+            return self.cxx_object().nTemperature()
 
     property coeffs:
         """
         2D array of Chebyshev coefficients of size `(n_temperature, n_pressure)`.
         """
         def __get__(self):
-            c = np.fromiter(self.chr().coeffs(), np.double)
-            return c.reshape((self.chr().nTemperature(), self.chr().nPressure()))
+            c = np.fromiter(self.cxx_object().coeffs(), np.double)
+            return c.reshape(
+                (self.cxx_object().nTemperature(), self.cxx_object().nPressure()))
 
 
 cdef class CustomRate(ReactionRate):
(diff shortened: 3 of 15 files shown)


speth:
We can avoid the need for the `MultiRateFactory` class by adding a member function to the `ReactionRate` class, as [suggested](https://github.com/Cantera/cantera/pull/982#issuecomment-796917368) in my original outline for the `MultiRate` structure in #982. Namely, for each type of `ReactionRate`, all that's needed is a function:
```c++
virtual shared_ptr<MultiRateBase> newMultiRate() const {
    return make_shared<MultiBulkRate<ArrheniusRate, ArrheniusData>>();
}
```
Then, the replacement else-if tree in `addReaction` would work by just calling `rate->newMultiRate()`.

ischoegl:
@speth ... thanks for the suggestion - this is indeed simpler.

src/kinetics/RateFactory.cpp
```diff
@@ -0,0 +1,102 @@
...
+    // ArrheniusRate evaluator
+    reg("ArrheniusRate", [](const AnyMap& node, const Units& rate_units) {
+        return new ArrheniusRate(node, rate_units);
+    });
+    addAlias("ArrheniusRate", "");
+    addAlias("ArrheniusRate", "elementary");
+    addAlias("ArrheniusRate", "arrhenius");
+    addAlias("ArrheniusRate", "three-body");
+    addAlias("ArrheniusRate", "threebody");
+    addAlias("ArrheniusRate", "three_body");
```
speth: A number of these aliases are leftovers from the XML format. Given that importing from XML doesn't use this factory, can we do away with these extra names?

include/cantera/kinetics/RateFactory.h
```diff
@@ -0,0 +1,94 @@
...
+    //!  Mutex for use when calling the factory
+    static std::mutex rate_mutex;
+};
+
+
+//! Create a new empty ReactionRateBase object
+/*!
+ * @param type string identifying type of reaction rate.
+ */
+shared_ptr<ReactionRateBase> newRate(const std::string& type);
```
speth: Should this function be called `newReactionRate` (and the factory class named `ReactionRateFactory`)?

src/kinetics/RateFactory.cpp
```diff
@@ -0,0 +1,102 @@
...
+    addAlias("custom-function", "custom-rate-function");
+}
+
+shared_ptr<ReactionRateBase> newRate(const std::string& type)
+{
+    if (RateFactory::factory()->exists(type)) {
+        return shared_ptr<ReactionRateBase> (
+            RateFactory::factory()->create(type, AnyMap(), Units(0.0)));
+    }
+    return shared_ptr<ReactionRateBase> ();
```
speth: This behavior of silently returning a null object when the object type isn't found is a bit surprising, and different from how our other factories work. What's the motivation?

src/kinetics/ReactionRate.cpp
```diff
@@ -59,6 +60,11 @@ ArrheniusRate::ArrheniusRate(const Arrhenius& arr, bool allow_negative_A)
 {
 }
 
+unique_ptr<MultiRateBase> ArrheniusRate::newMultiRate() const
+{
+    return unique_ptr<MultiRateBase> (new MultiBulkRate<ArrheniusRate, ArrheniusData>);
```
speth: I don't think there should be a space between the end of the type name and the parentheses that start the constructor argument list.

src/kinetics/Reaction.cpp
```diff
@@ -190,6 +191,21 @@ void Reaction::setParameters(const AnyMap& node, const Kinetics& kin)
     input = node;
 }
 
+void Reaction::setRate(shared_ptr<ReactionRateBase> rate)
+{
+    if (!rate) {
+        // null pointer
+        m_rate.reset();
+    } else if (canonicalRateName(type()) == rate->type()) {
```
speth: I'm a little confused by the need for the `canonicalRateName` function. Is there a way that `type()` or `rate->type()` could return anything other than the (canonical) name as defined in the various `ReactionRate` implementations?

src/kinetics/Reaction.cpp
```diff
@@ -1041,20 +1061,22 @@ PlogReaction3::PlogReaction3(const Composition& reactants,
...
+    }
 }
 
 void PlogReaction3::getParameters(AnyMap& reactionNode) const
 {
+    reactionNode.update(m_rate->parameters(rate_units));
     Reaction::getParameters(reactionNode);
     reactionNode["type"] = "pressure-dependent-Arrhenius";
-    reactionNode.update(m_rate->parameters(rate_units));
 }
```
speth: What's the reason for changing the order of these steps? It doesn't seem to affect the output ordering of the YAML fields.

interfaces/cython/cantera/_cantera.pxd
```diff
@@ -341,6 +341,10 @@ cdef extern from "cantera/kinetics/ReactionFactory.h" namespace "Cantera":
     cdef shared_ptr[CxxReaction] CxxNewReaction "newReaction" (XML_Node&) except +translate_exception
     cdef shared_ptr[CxxReaction] CxxNewReaction "newReaction" (CxxAnyMap&, CxxKinetics&) except +translate_exception
 
+cdef extern from "cantera/kinetics/RateFactory.h" namespace "Cantera":
+    cdef shared_ptr[CxxReactionRateBase] CxxNewRate "Cantera::newRate" (string) except +translate_exception
+    cdef shared_ptr[CxxReactionRateBase] CxxNewRate "newRate" (CxxAnyMap&, CxxKinetics&) except +translate_exception
```
speth: If the `Cantera::` prefix isn't required for the second one, it's probably not necessary for the first either, right? I think there was a point where the `namespace "Cantera"` bit in the `cdef extern from` line didn't actually have the desired effect, which is why we have a lot of usage of the fully-qualified names, but that seems not to be necessary any more.

interfaces/cython/cantera/reaction.pyx
```diff
@@ -24,32 +29,121 @@ cdef class _ReactionRate:
...
+            self.rate.update(temperature)
+            return self.rate.eval(temperature)
+
+    @staticmethod
+    cdef wrap(shared_ptr[CxxReactionRateBase] rate):
+        """
+        Wrap a C++ Reaction object with a Python object of the correct derived type.
+        """
+        # ensure all reaction types are registered
+        if not(_reaction_rate_class_registry):
```
speth:
```suggestion
        if not _reaction_rate_class_registry:
```

interfaces/cython/cantera/reaction.pyx
```diff
@@ -24,32 +29,121 @@ cdef class _ReactionRate:
...
+        :param kinetics:
+            A `Kinetics` object whose associated phase(s) contain the species
+            involved in the reaction.
+        """
+        if cls._reaction_rate_type != "":
+            raise TypeError(
+                "Class method 'from_dict' was invoked from '{}' but should "
+                "be called from base class 'Reaction'".format(cls.__name__))
+        if kinetics is None:
+            raise ValueError("A Kinetics object is required.")
```
speth: You could just omit the default value if this is a required argument, right?

interfaces/cython/cantera/reaction.pyx
```diff
@@ -24,32 +29,121 @@ cdef class _ReactionRate:
...
+
+        cdef CxxAnyMap any_map = dict_to_anymap(data)
+        cxx_rate = CxxNewRate(any_map, deref(kinetics.kinetics))
+        return ReactionRate.wrap(cxx_rate)
+
+    @classmethod
+    def fromYaml(cls, text, Kinetics kinetics=None):
```
speth: Would it be better to just name this `from_yaml` now, and save ourselves the effort of deprecating and replacing it later?

interfaces/cython/cantera/reaction.pyx
```diff
@@ -24,32 +29,121 @@ cdef class _ReactionRate:
...
+        :param data:
+            A dictionary corresponding to the YAML representation.
+        :param kinetics:
+            A `Kinetics` object whose associated phase(s) contain the species
+            involved in the reaction.
+        """
+        if cls._reaction_rate_type != "":
+            raise TypeError(
+                "Class method 'from_dict' was invoked from '{}' but should "
+                "be called from base class 'Reaction'".format(cls.__name__))
```
speth: This should say `ReactionRate`, I think? Likewise for the same check in `fromYaml`.

interfaces/cython/cantera/reaction.pyx
```diff
@@ -165,7 +241,7 @@ cdef class PlogRate(_ReactionRate):
         """
         def __get__(self):
             rates = []
-            cdef vector[pair[double, CxxArrhenius]] cxxrates = self.rate.rates()
+            cdef vector[pair[double, CxxArrhenius]] cxxrates = self.pr().rates()
```
speth: I think this would be simpler without the `pr` helper method if this is the only place it's used.

interfaces/cython/cantera/reaction.pyx
```diff
@@ -223,63 +299,52 @@ cdef class ChebyshevRate(_ReactionRate):
...
 
     property coeffs:
         """
         2D array of Chebyshev coefficients of size `(n_temperature, n_pressure)`.
         """
         def __get__(self):
-            c = np.fromiter(self.rate.coeffs(), np.double)
-            return c.reshape((self.rate.nTemperature(), self.rate.nPressure()))
+            c = np.fromiter(self.chr().coeffs(), np.double)
+            return c.reshape((self.chr().nTemperature(), self.chr().nPressure()))
```
speth: I understand not wanting to re-write the cast operation, but these heavily abbreviated function names like `chr` and `arr` are a little difficult to parse. Maybe a consistent name across all of the places where this style is used would be more useful. For example, maybe something like `derived`?

src/kinetics/RateFactory.cpp
```diff
@@ -0,0 +1,102 @@
...
+        RateFactory::factory()->create(type, rate_node, rate_units));
+}
+
+shared_ptr<ReactionRateBase> newRate(const AnyMap& rate_node, const Kinetics& kin)
+{
+    if (rate_node.empty()) {
+        return newRate(AnyMap(), Units(0.0));
+    }
+    size_t idx = kin.reactionPhaseIndex();
+    Units rate_units = kin.thermo(idx).standardConcentrationUnits();
```
speth: This doesn't seem right to me - the units for the rate constant are a combination of the reacting phase standard concentration units and the concentration units for the reactants, as in `Reaction::calculateRateCoeffUnits`.

Is creating a `ReactionRate` object in the absence of a `Reaction` like this actually useful?

speth (changes requested):
Thanks, @ischoegl, most of this looks good to me. I think the only substantial concern I have is with figuring out the rate constant units in the `newRate(AnyMap, Kinetics)` function.

interfaces/cython/cantera/reaction.pyx
bryanwweber: This is not related to Ray's comment, but it applies to the same places :smile: I think we should standardize on a formatting method for strings in new code (similar to how we prefer double quotes in general). I think we can use `f` strings for most cases, rather than `.format()`. They tend to be shorter and the value to be included is clearer, since it's placed directly in the braces.

The exception is that `f` strings can't be assigned to a variable and reused. In cases where you want to reuse a string, the `.format()` method would have to be used (or the built-in string interpolation in classes derived from `Exception`).

interfaces/cython/cantera/reaction.pyx
```diff
@@ -24,32 +29,121 @@ cdef class _ReactionRate:
...
+        Create a `ReactionRate` object from a dictionary corresponding to its YAML
+        representation.
+
+        An example for the creation of a ReactionRate from a dictionary is::
+
+            rate = ReactionRate.from_dict(
+                {"rate-constant": {"A": 38.7, "b": 2.7, "Ea": 26191840.0}},
+                kinetics=gas)
+
+        In the example, *gas* is a Kinetics (or Solution) object.
```
bryanwweber: My preference is to use monospace formatting for function arguments in the docstring, but this is not consistently applied across the codebase. We should again try to develop a standard here, IMO. Also, `Kinetics` and `Solution` can be auto-linked to their docs. Likewise, `ReactionRate` can be linked in the sentence before the example. The same comment applies in `fromYaml`.

interfaces/cython/cantera/test/test_reaction.py
```diff
@@ -164,7 +164,8 @@ def test_unconfigured(self):
         self.assertTrue(np.isnan(rate(self.gas.T, self.gas.P)))
         input_data = rate.input_data
         self.assertIsInstance(input_data, dict)
-        self.assertEqual(input_data, {})
+        if input_data:
+            self.assertEqual(input_data, {'type': self._type})
```
bryanwweber: Double quotes on `type`, please 😄

bryanwweber (dismissed):
A few small style comments, thanks @ischoegl

interfaces/cython/cantera/reaction.pyx
ischoegl: No problem about changing this to `monospace`. Based on what I had seen, I thought the preference was to use *italics* for variable names.

interfaces/cython/cantera/reaction.pyx
ischoegl: @speth - absolutely, thanks!

@bryanwweber  ... hadn't used f-strings much until recently, but tend to agree that they're more succinct.

interfaces/cython/cantera/reaction.pyx
ischoegl: My thought was that it should be consistent, but I agree that it can use the future form right away.

interfaces/cython/cantera/reaction.pyx
bryanwweber: Thank you! Do you have a preference for italics vs. monospace, as one of the maintainers?

interfaces/cython/cantera/reaction.pyx
ischoegl: Correct, but on the other hand, the (currently) required `Kinetics` object is a wart that I hope to address at some point. It would be great to specify `Unit` from Python eventually. In other words, I hope to make it optional eventually.

interfaces/cython/cantera/_cantera.pxd
ischoegl: copy-pasta from `newReaction` above. Still not sure why it's apparently required for `newReaction` but not here.

src/kinetics/Reaction.cpp
ischoegl: It was a (failed) attempt to affect output ordering, that I never reverted.

include/cantera/kinetics/RateFactory.h
ischoegl: :+1: ... thought to use a shorter name, but it is admittedly not consistent. I am also renaming the files to `ReactionRateFactory.h/.cpp`.

src/kinetics/RateFactory.cpp
ischoegl: Excellent point. It turns out to not be necessary (at least at this point).

src/kinetics/Reaction.cpp
ischoegl: There is a need here: names of `ReactionRate` and `Reaction` objects do not match. In order to make this work for YAML input,  the `ReactionRateFactory` uses aliases, see, for example, `ArrheniusRate`.
```C++
    // ArrheniusRate evaluator
    reg("ArrheniusRate", [](const AnyMap& node, const Units& rate_units) {
        return new ArrheniusRate(node, rate_units);
    });
    addAlias("ArrheniusRate", "");
    addAlias("ArrheniusRate", "elementary");
    addAlias("ArrheniusRate", "three-body");
```
(I removed other - no longer necessary - aliases per request.)

In other words, the `canonicalRateName` function just looks up the rate type based on a reaction type.

interfaces/cython/cantera/reaction.pyx
ischoegl: With the factories, some things changed so there would be a need to cast (`self.rate` is now always the base class). Since I'm using the `cxx` helper functions elsewhere, I'd like to leave this in for consistency.

interfaces/cython/cantera/reaction.pyx
ischoegl: Since those `cdef` functions are not accessible from the outside, keeping things short is imho important. To make things more consistent, I changed the various implementations to `cxx`.

~I am applying the same rationale to the helper functions for `Reaction`, where I am now using `cxx` and `cxx2`.~ *undid that as class inheritance prevents a uniform naming scheme*

interfaces/cython/cantera/reaction.pyx
ischoegl: I don't, honestly. I think monospace is somewhat easier to discern, so I went ahead and replaced all italics in `reaction.pyx`

interfaces/cython/cantera/reaction.pyx
bryanwweber: > Since those `cdef` functions are not accessible from the outside, keeping things short is imho important.

To me, it is more important to be descriptive. This code will be read many more times by future developers, so having the names be clear is important in all the code, not just in userland-API code.

interfaces/cython/cantera/reaction.pyx
bryanwweber: > In other words, I hope to make it optional eventually.

Then, IMO, we should make it optional at that point. This would lead to pretty confusing documentation, where the value would appear to be optional but then actually it's required.

ischoegl: Interestingly, some of this turns out to be pre-existing (see `Reaction.fromYaml`. I don't have a problem making all those required.

interfaces/cython/cantera/reaction.pyx
ischoegl: Since we're now allowing a mix of `from_yaml` and `fromYaml`, I went ahead and deprecated `Reaction.fromYaml`, so `reaction.pyx` is consistent.

interfaces/cython/cantera/reaction.pyx
ischoegl: I believe that `cxx` is descriptive, as it refers to the actual C++ object?

src/kinetics/RateFactory.cpp
ischoegl: I believe that from an instructional viewpoint, stand-alone `ReactionRate` objects do make sense (beyond obvious benefits for the testing suite). 

As there is no information about species concentrations here, the only units available are what is attached to `Kinetics` (which imho is the most consistent assumption). ~As `Reaction::calculateRateCoeffUnits` is called after the creation of a `ReactionRate`, this gets overwritten.~ In most cases, unit information is passed in the alternate constructor (`newReactionRate(rate_node, rate_units)`), which is consistent.

**PS:** This comment actually points at an inconsistency in some of the constructors: take
```C++
ElementaryReaction3::ElementaryReaction3(const Composition& reactants,
                                         const Composition& products,
                                         const ArrheniusRate& rate);
```
and same for 
```C++
ElementaryReaction2::ElementaryReaction2(const Composition& reactants_,
                                         const Composition products_,
                                         const Arrhenius& rate_);
```
as well as all other corresponding instances.

In these cases there is no `Kinetics` object available to actually set consistent units? (`Kinetics` is needed for `Reaction::calculateRateCoeffUnits`).

interfaces/cython/cantera/reaction.pyx
bryanwweber: I think something along the lines of `cxx_object` or `cxx_pointer` or `rate_pointer` or `rate_cxx_object` would be more descriptive. I know that takes more space, but especially since this is not going to be a user-facing API, it won't have to be typed out a lot, but it will be read much more frequently, and having clarity without having to go searching for what the `cxx` means would be a benefit.

One philosophical point I've come around to the longer I've been working on Cantera is to try and avoid layers of indirection and abstraction, because there's so much going on already that it's nearly impossible to keep all of it in your brain to be able to do any work on Cantera. If it does make sense to add a layer, it should be named super clearly so that you don't need to keep referring to the implementation to know what it is.

ischoegl: NP. I now have named most of the helper functions `cxx_object` (for both `ReactionRate` and `Reaction`),

FWIW, I had adopted a pre-existing approach here (which I liked and found easy to follow), but I do agree that readability is important.

ischoegl:
@speth and @bryanwweber  - thank you for your reviews ... I addressed / responded to all comments at this point.

Regarding the unit controversy, I believe that there are some inconsistencies that go beyond this PR (see new issue #1071) - the constructor in question is useful for the creation of `Reaction` objects from scratch, where unit handling is not clear.

src/kinetics/RateFactory.cpp
ischoegl: As mentioned below, I believe that the constructor is actually critical for the creation of reactions from scratch, where - at least at this moment - unit handling is not clear. I opened #1071 as this goes well beyond the scope of this PR.

ischoegl:
Thanks, @bryanwweber!

src/kinetics/RateFactory.cpp
ischoegl: @speth ... based on #1071, I believe this should be resolved by not requiring a `Kinetics` object *at all*, and letting this be updated when a corresponding reaction is added to `Kinetics`?

ischoegl:
@speth ... I believe I resolved the issue with units by removing the `Kinetics` parameter. Regarding the question about where the creation of a stand-alone `ReactionRate` object could be useful, I believe it is critical for the creation of a `Reaction` object from scratch, e.g.
```
In [1]: import cantera as ct
   ...: 
   ...: reactants = {'O': 1., 'H2': 1}
   ...: products = {'H': 1., 'OH': 1.}
   ...: rxn = ct.ElementaryReaction(reactants, products)
   ...: rxn
   ...: 
Out[1]: <ElementaryReaction: H2 + O <=> H + OH>

In [2]: rxn.rate # reaction rate is not set up (yet)
Out[2]: <ArrheniusRate at 7f38eb99ef00>

In [3]: rxn.rate(300.)
Out[3]: nan

In [4]: rxn.rate = ct.ReactionRate.from_yaml("rate-constant: {A: 38.7, b: 2.7, Ea: 6260.0 cal/mol}")
   ...: rxn.rate
   ...: 
Out[4]: <ArrheniusRate at 7f38eb99ef60>

In [5]: rxn.rate(300.)
Out[5]: 5195.44737317942
```

src/kinetics/RateFactory.cpp
ischoegl: I believe that eliminating the `Kinetics` arguments resolves the issue here - I agree that the units wouldn't be correct in the original implementation.

ischoegl:
@speth ... the last commit blocks an edge case where a user would specify a unit system that would produce `rate_units` that have a factor other than `1.0` (which is probably the main take-away of the discussion in #1071):

```
In [2]: rxn.rate = ct.ReactionRate.from_yaml("""
   ...: rate-constant: {A: 38.7, b: 2.7, Ea: 6260.0 cal/mol}
   ...: units: {length: cm, quantity: mol}
   ...: """)

[...]

CanteraError: 
InputFileError thrown by ReactionRateFactory::newReactionRate:
Error on line 2 of input string:
Alternative units for 'length' or 'quantity` are not supported when creating
a standalone 'ReactionRate' object.
|  Line |
|     1 | 
|     2 | rate-constant: {A: 38.7 kmol/cm^3, b: 2.7, Ea: 6260.0 cal/mol}
>     3 > units: {length: cm, quantity: mol}
                 ^
```

With this update, I believe that all comments are taken care of.

**PS:** there is an exception though:
```
ct.ReactionRate.from_yaml("rate-constant: {A: 4.0e+21 cm^6/mol^2/s, b: 0.0, Ea: 1207.72688}")
```
which is a bit harder to catch (see `units-custom.yaml`).

ischoegl:
Not too hard after all. I believe this is done now ...
```
In [2]: rate = ct.ReactionRate.from_yaml("rate-constant: {A: 4.0e+21 cm^6/mol^2/s, b: 0.0, Ea: 1207.72688}")
<ipython-input-2-d77e48358ab6> in <module>()
----> 1 rate = ct.ReactionRate.from_yaml("rate-constant: {A: 4.0e+21 cm^6/mol^2/s, b: 0.0, Ea: 1207.72688}")

/src/interfaces/cython/cantera/reaction.pyx in cantera._cantera.ReactionRate.from_yaml()
    109         cdef CxxAnyMap any_map
    110         any_map = AnyMapFromYamlString(stringify(text))
--> 111         cxx_rate = CxxNewReactionRate(any_map)
    112         return ReactionRate.wrap(cxx_rate)
    113 

CanteraError: 
InputFileError thrown by Arrhenius::setParameters:
Error on line 1 of input string:
Specification of units is not supported for pre-exponential factor when
creating a standalone 'ReactionRate' object.
|  Line |
>     1 > rate-constant: {A: 4.0e+21 cm^6/mol^2/s, b: 0.0, Ea: 1207.72688}
                                   ^
```

src/kinetics/Reaction.cpp
speth: Sorry, I was missing the fact that this was comparing the rate type to the reaction type. But I don't think this is a check that we should be making in the first place -- one of the key benefits of the generic `ReactionRate` class is breaking this requirement that there is a specific rate parameterization that goes along with a given `Reaction` class. And if we don't need this check, we don't need the `canonicalRateName` function, either.

src/kinetics/Reaction.cpp
```diff
@@ -865,7 +886,7 @@ void BlowersMaselInterfaceReaction::getParameters(AnyMap& reactionNode) const
 
 ElementaryReaction3::ElementaryReaction3()
 {
-    m_rate.reset(new ArrheniusRate);
+    setRate(newReactionRate(type()));
```
speth, line 888: Do we need to create a `ReactionRate` object here? In the most common use case, where the reaction is initialized from `AnyMap` input, it's just going to get overwritten almost instantly. And this is another place where the rate type is unnecessarily determined by the `Reaction` class.

src/kinetics/Reaction.cpp
```diff
@@ -190,6 +191,21 @@ void Reaction::setParameters(const AnyMap& node, const Kinetics& kin)
...
+{
+    if (!rate) {
+        // null pointer
+        m_rate.reset();
+    } else if (canonicalRateName(type()) == rate->type()) {
+        m_rate = rate;
+    } else {
+        throw CanteraError("Reaction::setRate",
+            "Mismatched reaction rate types: reaction type '{}' requires\n"
+            "rate with canononical type '{}' but received rate object with\n"
```
speth: "canonical" is misspelled (though I think this should just go away).

src/kinetics/ReactionRateFactory.cpp
```diff
@@ -0,0 +1,99 @@
...
+    UnitSystem system = rate_node.units();
+    if (system.convertTo(1., "m") != 1. || system.convertTo(1., "kmol") != 1.) {
+        throw InputFileError("ReactionRateFactory::newReactionRate",
+            rate_node.at("__units__"),
+            "Alternative units for 'length' or 'quantity` are not supported "
+            "when creating\na standalone 'ReactionRate' object.");
+    }
+    AnyMap node(rate_node);
+    node["__standalone__"] = true;
+    return newReactionRate(node, Units(1.));
```
speth: I think most of this business with the `__standalone__` flag can be handled more simply by setting creating the rate with `Units(0)`, and then checking whether `rate_units.factor() != 0`. This is the sentinel value that I used to indicate undetermined units elsewhere.

src/kinetics/ReactionRateFactory.cpp
```diff
@@ -0,0 +1,99 @@
...
+            "Unknown reaction rate type '{}'", type);
+    }
+
+    return shared_ptr<ReactionRateBase> (
+        ReactionRateFactory::factory()->create(type, rate_node, rate_units));
+}
+
+shared_ptr<ReactionRateBase> newReactionRate(const AnyMap& rate_node)
+{
+    UnitSystem system = rate_node.units();
```
speth: This can be a reference.

src/kinetics/ReactionRateFactory.cpp
```diff
@@ -0,0 +1,99 @@
...
+    reg("PlogRate", [](const AnyMap& node, const Units& rate_units) {
+        return new PlogRate(node, rate_units);
+    });
+    addAlias("PlogRate", "pressure-dependent-Arrhenius");
+
+    // ChebyshevRate evaluator
+    reg("ChebyshevRate", [](const AnyMap& node, const Units& rate_units) {
+        return new ChebyshevRate3(node, rate_units);
+    });
+    addAlias("ChebyshevRate", "Chebyshev");
```
speth: Do we need to introduce these new names like `ChebyshevRate` and `PlogRate`, or can we just use the names defined in the YAML format? I can see a case for handling `Arrhenius` differently, as having the canonical name be the empty string would be a little weird.

speth (changes requested):
Thanks for the updates to this, @ischoegl. After the extended discussion on #1071, I think most of the complicated issues here have already been worked out without introducing too much new complexity. My remaining comments are mostly places on things that should either simplify the current implementation slightly or make some further generalizations that I suspect we both have in mind easier.

src/kinetics/Reaction.cpp
ischoegl: Hm. Interesting point. It is true that once `legacy` is removed, there won’t be much left to handle for different reaction types. But does it really make sense to have a Fall-off reaction paired with, say, Chebyshev rates? The checks here are mostly for the ‘from-scratch’ route, where YAML definitions don’t apply.

I agree that there may be some simplifications possible eventually, but I’d prefer to wait until legacy rate objects are removed (or at least all replacements are fully implemented). So could we table this?

src/kinetics/Reaction.cpp
ischoegl: typo is fixed.

src/kinetics/Reaction.cpp
ischoegl: I think it's easier to create an 'empty' rate for the default constructor, but it's straight-forward to not call this from the constructor that takes `AnyMap`. That way those 'throw-away' instantiations are avoided.

src/kinetics/ReactionRateFactory.cpp
ischoegl: Indeed much easier - thanks for the suggestion!

src/kinetics/ReactionRateFactory.cpp
ischoegl: This is an interesting proposition. While it muddles things a little short-term (no clear differentiation of names for `ReactionRate` and `Reaction` objects), it mostly avoids the `canonicalName` issue. Long-term, I don't believe that there will be a need to have different reaction types, as everything is determined by the attached rate object.

So yes, this is a great way to avoid naming issues down the road ...  :+1:

src/kinetics/Reaction.cpp
ischoegl: Thinking about this more, I removed the `canonicalName` check as suggested. It opens quite a few interesting opportunities down the line.

ischoegl:
@speth ... thank you for your re-review and the interesting comments. I ended up adopting all your suggested changes.

> [...] one of the key benefits of the generic `ReactionRate` class is breaking this requirement that there is a specific rate parameterization that goes along with a given `Reaction` class.
 
I honestly was still thinking along the old lines too much here. I certainly realized that the revised `Reaction3` specializations were pretty much hollowed out (i.e. not many differences), but following that thought it will become possible to at least partially ditch having specialized `Reaction` classes altogether. As an example, there's nothing wrong with an `elementary` reaction having a `Chebyshev` rate object. This opens up a lot of interesting possibilities and simplifications ...

ischoegl:
Realized that shifting the emphasis away from `Reaction` (using consistent naming) makes specialized `getParameters` mostly unnecessary (except for `ThreeBodyReaction3`, at least for the moment). Which means that `ElementaryReaction3`, `PlogReaction3` and `ChebyshevReaction3` are essentially the same (except for minor differences in constructors). Not bad ... :tada:

I believe this is ready to merge now. Getting rid of some `Reaction` specializations (while possible) is likely a longer discussion for another PR.

*PS: Changing the `type` string output created a minor paper cut, which is also fixed.*

src/kinetics/ReactionRate.cpp
```diff
@@ -67,7 +78,9 @@ void ArrheniusRate::setParameters(const AnyMap& node, const Units& rate_units)
         Arrhenius::setParameters(AnyValue(), node.units(), rate_units);
         return;
     }
+
     Arrhenius::setParameters(node["rate-constant"], node.units(), rate_units);
+    return;
```
speth: The bare `return` at the end of the function can be removed. Also applies to `PlogRate::setParameters`.

speth (changes requested):
Thanks, @ischoegl, this looks good to me. I had just one very minor nitpick, commented on below.

Regarding the fate of the `Reaction` class, I expect that the only distinction ultimately worth keeping will be between bulk phase reactions and interface/electrochemical reactions. But even then, some of the extra parameters and functionality may fit better within the `ReactionRate` classes.

ischoegl:
@speth ... thanks for the reviews! I removed the redundant `return`s with the last commit.

> Regarding the fate of the `Reaction` class, I expect that the only distinction ultimately worth keeping will be between bulk phase reactions and interface/electrochemical reactions. But even then, some of the extra parameters and functionality may fit better within the `ReactionRate` classes.

:+1: ... I see it the same way - the YAML `type` field will essentially be pushed through to the `ReactionRate` object, which should be able to take care of (almost) everything.

speth (approved):

Issue #267, opened 2015-06-03
Improve test suite coverage or remove unused code


speth:
The Cantera codebase contains a number of classes that are potentially unused. They are not used internally within Cantera, have no test coverage, and have no examples  available, making them essentially useless to any potential users. These classes are identified below, with notes from relevant commits. The status of each of these classes should be individually resolved in one of two ways:
- Add examples and tests
- Deprecate and remove

The current state of affairs is not sustainable, making it difficult for users to understand the code and creating unnecessary barriers to effectively maintaining and improving Cantera. Specific details about these classes are described below, based on looking through the commit history. This list excludes classes which have already been deprecated in Cantera 2.2. 
- **`ThermoPhase` derivatives**:
  - [x] `SemiconductorPhase`: Added in 8ee5b76fc9 (2007). Untouched since then. No tests, and unconstructible by `ThermoFactory`. _Deprecated in 111b490. Removed in 2.3._
  - [x] `MixtureFugacityTP` / `RedlichKwongMFTP` - _Tests added by @decaluwe in f9d5f16b._
  - [x] `MineralEQ3` Added in 3afa91395 (2008) no test cases or examples *deprecated in 6d22be2*
  - [x] `IonsFromNeutralVPSSTP`,  Added in 06276fba6d (2009) _Regression tests added in  338b216 and 3a11938_
  - [x] `MetalSHEelectrons` Added in 321d387930 (2009) no test cases or examples *deprecated in 6d22be2*
  - [x] `PhaseCombo_Interaction` Added in 30d59b8c49 (2011) no test cases or examples *deprecated in 6d22be2*
  - [x] `MixedSolventElectrolyte` Added in ff7e66e5a34 (2011) no test cases or examples. Not constructible by `ThermoFactory`. _Deprecated in 2b73fe2_.
  - [x] `MolarityIonicVPSSTP` Added in b9c9f817615 (2011) no test cases or examples *deprecated in 6d22be2*
  - [x] `RedlichKisterVPSSTP` Added in 5d1e55596d (2011) _Test case added in 1183479_
- **Other thermo classes**
  - [x] `PDSS_SSVol` Added in 48261703d4 (2009) _regression test added in a520f782_
  - [x] `PDSS_IonsFromNeutral` Added in 06276fba6da (2009)  _Regression tests added in  338b216 and 3a11938_
  - [x] `VPSSMgr_IdealGas` Added in e46b6f2b2d (2008) _eliminated by refactoring in b39537b_
[...]

PR #1062, opened 2021-06-16
Migrate "test_problems" tests to use YAML input files

speth:
In preparation for the removal of the CTI and XML formats, this updates all of the tests in the `test_problems` directory to use the YAML format (with the exception of a few tests that are run for each input format).

- Update all `test_problems` tests to use YAML input
- Fix a handful of small issues discovered while making this transition
  - Use `InputFileError` to improve error messages related to sticking reactions
  - Fix the phase dimension checks in `newReaction` to always use the correct phase
  - Fix CTI to YAML conversion for files with no reactions
  - Fix XML to YAML conversion of HMW phases with an implicit temperature model
  - Handle the case of a missing `reactionArray` node in XML files
  - Update the phase name aliases handled by `ctml2yaml`
 - Remove the `surfkin` and `silane_equil` tests, which weren't adding anything that wasn't better covered by other tests.

**Checklist**


test_problems/surfSolverTest/surfaceSolver.cpp
```diff
@@ -6,12 +6,6 @@
 // This file is part of Cantera. See License.txt in the top-level directory or
 // at https://cantera.org/license.txt for license and copyright information.
 
-//  Example
-//
-//  Read a surface growth mechanism and calculate the solution
```
ischoegl, line 11: While the comments at the top were certainly not extremely helpful, why remove them?

test_problems/surfSolverTest/surfaceSolver2.cpp
```diff
@@ -6,21 +6,8 @@
 // This file is part of Cantera. See License.txt in the top-level directory or
 // at https://cantera.org/license.txt for license and copyright information.
 
-//  Example
-//
-//  Read a surface growth mechanism and calculate the solution
```
ischoegl, line 11: Same here.

ischoegl (commented):
Thanks for taking this on - this takes the transition from CTI/XML to YAML a big step forward. I took a cursory look at the changes, and most of them are mainly replacing input files (i.e. cumbersome, but straight-forward).

I do have two comments, though: documentation on some of the tests is relatively sparse. While it's certainly beyond the scope of this PR to make major changes, leaving very brief statements at the top may be helpful in the future (you are probably the one person who has the best overview).

There was a comment about `Interface` in one of the test problems - this is obviously deprecated in 2.5 and currently removed in 2.6. There were also recent questions about this on the UG. While I am not advocating on a reversal of this particular deprecation, do you see a way to introduce a new class that represents a collection of multiple C++ `SolutionBase` objects? (I don't think it's needed, but an inherited class that simplifies loading may be nice).

test_problems/surfSolverTest/surfaceSolver.cpp
speth: The original comment was "Read a surface growth mechanism and calculate the solution using Placid". I don't know what "Placid" is, so my assumption is that this comment is just copy-pasta garbage and that's why I removed it.

ischoegl: I read that :laughing: ... but you probably know what this test actually does. Apart from that, all others are left with is purely based on the name `surfSolverTest/surfaceSolver.cpp `

speth:
> I do have two comments, though: documentation on some of the tests is relatively sparse. While it's certainly beyond the scope of this PR to make major changes, leaving very brief statements at the top may be helpful in the future (you are probably the one person who has the best overview).

Poor/incorrect documentation is one of many faults of the `test_problems` test suite. Ultimately, I think it should be eliminated in favor of using the more explicit and expressive GTest and Python test suites. Progress towards this end has been very slow, though, in part because it's laborious, and in part because it's not always easy to make sure that the replacement is actually catching all the important behavior from the original test. I can create a catch-all issue to document some of the things that ought to be done and track progress, as in #267, but I don't think trying to document these tests is a very good use of time.

ischoegl:
> … I don't think trying to document these tests is a very good use of time.

Fair point. I think getting the `Interfaces` issue worked out is certainly much more impactful. Your enhancement post is really appreciated.

test_problems/surfSolverTest/surfaceSolver.cpp
ischoegl: Closing.

ischoegl (commented):
Overall, those changes all look good to me. The only things I am not absolutely sure about are  the surface solver tests where you _”removed unused tests”_ (ed00efe72472ea3a2de886df021f121ad41d680c), with associated code removals. If these indeed weren’t used at all, I’m ok to approve this PR.

On the other hand, there’s Cantera/enhancements#103, where a resolution may simplify some of the work here. (But I’m not a fan of leaving large PR’s lingering).

speth:
> The only things I am not absolutely sure about are the surface solver tests where you _”removed unused tests”_ ([ed00efe](https://github.com/Cantera/cantera/commit/ed00efe72472ea3a2de886df021f121ad41d680c)), with associated code removals. If these indeed weren’t used at all, I’m ok to approve this PR.

The code removed in that commit falls into two categories:
1. Messy code for processing command line options that are never passed in as part of the test suite.
2. Code that is in unreachable branches, particularly where the variable `ProblemNumber` is hardcoded and then there are different `if` conditions based on the value of `ProblemNumber`, as well as some other simplifications that resulted from removing those bits of dead code.

You can tell that these changes didn't have an impact on what's actually tested from the fact that the "blessed" output files for these tests haven't changed.

ischoegl (approved):
> Poor/incorrect documentation is one of many faults of the test_problems test suite. Ultimately, I think it should be eliminated in favor of using the more explicit and expressive GTest and Python test suites. 

I agree that it would be preferable to eliminate those tests long-term. In the meantime, I am approving these changes.

bryanwweber:
@ischoegl If you approve on behalf of @Cantera/committers then feel free to merge after ~24-48 hours from PR creation.

PR #1063, opened 2021-06-16
Individual Reactor Initialization - Segmentation Fault Fix.

anthony-walker:
There was a segmentation fault when calling initialize on an individual reactor, this is a fix for that by setting it to the specified time if a ReactorNet has not been provided or used.

I specifically changed `ReactorNet::updateConnected` in both the .cpp and .h files to accept a time arguement. If the reactor network is not available, it uses the specified or default time value of the reactor specific initialize function.

**Checklist**

### DIFF
--- a/include/cantera/zeroD/Reactor.h
+++ b/include/cantera/zeroD/Reactor.h
@@ -194,7 +194,7 @@ protected:
     //!     calculated from the state, and `false` when the pressure is constant
     //!     or an independent variable.
     //! @param t0 initialization time for the reactor if not obtained from the network
-    virtual void updateConnected(bool updatePressure, double t0 = 0.0);
+    virtual void updateConnected(bool updatePressure, double t0=0.0);
 
     //! Get initial conditions for SurfPhase objects attached to this reactor
     virtual void getSurfaceInitialConditions(double* y);
--- a/include/cantera/zeroD/Reactor.h
+++ b/include/cantera/zeroD/Reactor.h
@@ -193,8 +193,7 @@ protected:
     //!     `true` for reactors where the pressure is a dependent property,
     //!     calculated from the state, and `false` when the pressure is constant
     //!     or an independent variable.
-    //! @param t0 initialization time for the reactor if not obtained from the network
-    virtual void updateConnected(bool updatePressure, double t0=0.0);
+    virtual void updateConnected(bool updatePressure);
 
     //! Get initial conditions for SurfPhase objects attached to this reactor
     virtual void getSurfaceInitialConditions(double* y);
--- a/include/cantera/zeroD/Reactor.h
+++ b/include/cantera/zeroD/Reactor.h
@@ -193,7 +193,8 @@ protected:
     //!     `true` for reactors where the pressure is a dependent property,
     //!     calculated from the state, and `false` when the pressure is constant
     //!     or an independent variable.
-    virtual void updateConnected(bool updatePressure);
+    //! @param t0 initialization time for the reactor if not obtained from the network
+    virtual void updateConnected(bool updatePressure, double t0 = 0.0);
 
     //! Get initial conditions for SurfPhase objects attached to this reactor
     virtual void getSurfaceInitialConditions(double* y);
--- a/src/zeroD/Reactor.cpp
+++ b/src/zeroD/Reactor.cpp
@@ -88,7 +88,7 @@ void Reactor::initialize(doublereal t0)
     m_thermo->restoreState(m_state);
     m_sdot.resize(m_nsp, 0.0);
     m_wdot.resize(m_nsp, 0.0);
-    updateConnected(true, t0);
+    updateConnected(true);
 
     for (size_t n = 0; n < m_wall.size(); n++) {
         WallBase* W = m_wall[n];
@@ -177,7 +177,7 @@ void Reactor::updateSurfaceState(double* y)
     }
 }
 
-void Reactor::updateConnected(bool updatePressure, double t0) {
+void Reactor::updateConnected(bool updatePressure) {
     // save parameters needed by other connected reactors
     m_enthalpy = m_thermo->enthalpy_mass();
     if (updatePressure) {
@@ -187,7 +187,7 @@ void Reactor::updateConnected(bool updatePressure, double t0) {
     m_thermo->saveState(m_state);
 
     // Update the mass flow rate of connected flow devices
-    double time = (m_net != nullptr) ? m_net->time() : t0;
+    double time = (m_net != nullptr) ? m_net->time() : 0.0;
     for (size_t i = 0; i < m_outlet.size(); i++) {
         m_outlet[i]->updateMassFlowRate(time);
     }
--- a/src/zeroD/Reactor.cpp
+++ b/src/zeroD/Reactor.cpp
@@ -88,7 +88,7 @@ void Reactor::initialize(doublereal t0)
     m_thermo->restoreState(m_state);
     m_sdot.resize(m_nsp, 0.0);
     m_wdot.resize(m_nsp, 0.0);
-    updateConnected(true,t0);
+    updateConnected(true, t0);
 
     for (size_t n = 0; n < m_wall.size(); n++) {
         WallBase* W = m_wall[n];
@@ -187,7 +187,7 @@ void Reactor::updateConnected(bool updatePressure, double t0) {
     m_thermo->saveState(m_state);
 
     // Update the mass flow rate of connected flow devices
-    double time = (m_net!=NULL) ? m_net->time() : t0;
+    double time = (m_net != nullptr) ? m_net->time() : t0;
     for (size_t i = 0; i < m_outlet.size(); i++) {
         m_outlet[i]->updateMassFlowRate(time);
     }
--- a/src/zeroD/Reactor.cpp
+++ b/src/zeroD/Reactor.cpp
@@ -88,7 +88,7 @@ void Reactor::initialize(doublereal t0)
     m_thermo->restoreState(m_state);
     m_sdot.resize(m_nsp, 0.0);
     m_wdot.resize(m_nsp, 0.0);
-    updateConnected(true);
+    updateConnected(true,t0);
 
     for (size_t n = 0; n < m_wall.size(); n++) {
         WallBase* W = m_wall[n];
@@ -177,7 +177,7 @@ void Reactor::updateSurfaceState(double* y)
     }
 }
 
-void Reactor::updateConnected(bool updatePressure) {
+void Reactor::updateConnected(bool updatePressure, double t0) {
     // save parameters needed by other connected reactors
     m_enthalpy = m_thermo->enthalpy_mass();
     if (updatePressure) {
@@ -187,7 +187,7 @@ void Reactor::updateConnected(bool updatePressure) {
     m_thermo->saveState(m_state);
 
     // Update the mass flow rate of connected flow devices
-    double time = m_net->time();
+    double time = (m_net!=NULL) ? m_net->time() : t0;
     for (size_t i = 0; i < m_outlet.size(); i++) {
         m_outlet[i]->updateMassFlowRate(time);
     }
--- a/test/zeroD/test_zeroD.cpp
+++ b/test/zeroD/test_zeroD.cpp
@@ -43,10 +43,10 @@ TEST(ZeroDim, test_individual_reactor_initialization)
     reactor2.setThermoMgr(*gas2);
     reactor2.initialize(0);
     // Get state of reactors
-    double *state1 = new double[reactor1.neq()];
-    double *state2 = new double[reactor2.neq()];
-    reactor1.getState(state1);
-    reactor2.getState(state2);
+    std::vector<double> state1 (reactor1.neq());
+    std::vector<double> state2 (reactor2.neq());
+    reactor1.getState(state1.data());
+    reactor2.getState(state2.data());
     // Compare the reactors.
     EXPECT_EQ(reactor1.neq(), reactor2.neq());
     double tol = 1e-14;
@@ -56,8 +56,6 @@ TEST(ZeroDim, test_individual_reactor_initialization)
     {
         EXPECT_NEAR(state1[i], state2[i], tol);
     }
-    delete[] state1;
-    delete[] state2;
 }
 
 int main(int argc, char** argv)
--- a/test/zeroD/test_zeroD.cpp
+++ b/test/zeroD/test_zeroD.cpp
@@ -14,7 +14,8 @@ TEST(ZeroDim, test_individual_reactor_initialization)
 {
     // Initial conditions
     double T0 = 1100.0;
-    double P0 = 10*OneAtm;
+    double P0 = 10 * OneAtm;
+    double tol = 1e-7;
     std::string X0 = "H2:1.0, O2:0.5, AR:8.0";
     // Reactor solution, phase, and kinetics objects
     std::shared_ptr<Solution> sol1 = newSolution("h2o2.yaml");
@@ -43,10 +44,7 @@ TEST(ZeroDim, test_individual_reactor_initialization)
     reactor2.getState(state2.data());
     // Compare the reactors.
     EXPECT_EQ(reactor1.neq(), reactor2.neq());
-    double tol = 1e-14;
-    EXPECT_NEAR(state1[0], state2[0], tol);
-    EXPECT_NEAR(state1[1], state2[1], tol);
-    for(size_t i = 3; i < reactor1.neq(); i++)
+    for (size_t i = 0; i < reactor1.neq(); i++)
     {
         EXPECT_NEAR(state1[i], state2[i], tol);
     }
--- a/test/zeroD/test_zeroD.cpp
+++ b/test/zeroD/test_zeroD.cpp
@@ -14,17 +14,14 @@ TEST(ZeroDim, test_individual_reactor_initialization)
 {
     // Initial conditions
     double T0 = 1100.0;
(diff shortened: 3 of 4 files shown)


src/zeroD/Reactor.cpp
```diff
@@ -187,7 +187,7 @@ void Reactor::updateConnected(bool updatePressure) {
     m_thermo->saveState(m_state);
 
     // Update the mass flow rate of connected flow devices
-    double time = m_net->time();
+    double time = (m_net!=NULL) ? m_net->time() : t0;
```
bryanwweber:
```suggestion
    double time = (m_net != NULL) ? m_net->time() : 0.0;
```

Cantera's style is to have space around the operators.

bryanwweber (changes requested):
Thanks for the quick fix @anthony-walker! Is it necessary to pass in `t0` to `updateConnected`? If a reactor is not connected to a Network, any times greater than zero aren't really meaningful, and if it is connected to a network this won't be a problem. I suggested just setting the value to zero in the change below.

Can you please also add a test for this, so we don't re-introduce the bug by accident, and we can make sure that it's fixed?

bryanwweber:
Actually, thinking about it a little more... The other option here is to throw an error if the network is not connected. Did you consider that option? If you did, and you rejected it, can you explain why? Thanks!

anthony-walker:
@bryanwweber I considered both using strictly zero and throwing an error. I chose to pass the argument along because it is presently used in all of the `initialize` functions. I suspect you are correct that it will be zero most of the time as if folks are integrating they will use a network. I rejected throwing the error because it will be useful in my case for writing some unit tests where I strictly need values from the reactor.  I also believe that in lieu of throwing an error it should just be a protected function and ReactorNet can be made a friend of it to allow it to be called in the conventional way. 

I will also add a test for which ever direction you feel we should pursue.

src/zeroD/Reactor.cpp
```diff
@@ -88,7 +88,7 @@ void Reactor::initialize(doublereal t0)
     m_thermo->restoreState(m_state);
     m_sdot.resize(m_nsp, 0.0);
     m_wdot.resize(m_nsp, 0.0);
-    updateConnected(true);
+    updateConnected(true,t0);
```
ischoegl: Another missing space.

ischoegl (commented):
I think the proposed fix looks good other than the usual formatting quibbles.

Edit: 👍 on @bryanwweber ’s request to add a unit test. Perhaps you can add the one you already started? (Minus stuff that is specific to the adaptive preconditioner).

include/cantera/zeroD/Reactor.h
```diff
@@ -193,7 +193,8 @@ class Reactor : public ReactorBase
     //!     `true` for reactors where the pressure is a dependent property,
     //!     calculated from the state, and `false` when the pressure is constant
     //!     or an independent variable.
-    virtual void updateConnected(bool updatePressure);
+    //! @param t0 initialization time for the reactor if not obtained from the network
+    virtual void updateConnected(bool updatePressure, double t0 = 0.0);
```
speth: No spaces around the `=` when it's defining a default value.

src/zeroD/Reactor.cpp
speth: Also, you should compare to `nullptr` rather than `NULL` to be more in line with C++11.

speth (commented):
Seems like a pretty straightforward fix.

One suggestion for the test case would be to make sure that the reactor still works correctly if it's initialized without a `ReactorNet` and then later added to one. I don't see a reason why this wouldn't work, but it would be good to check.

test/zeroD/test_zeroD.cpp
```diff
@@ -0,0 +1,57 @@
...
+    gas2->equilibrate("HP");
+    //Compare the two gases.
+    double tol = 1e-8; //relative tolerance from python assertNear test
+    EXPECT_NEAR(gas1->temperature(), gas2->temperature(), tol);
+    EXPECT_NEAR(gas1->density(), gas2->density(), tol);
+    EXPECT_NEAR(gas1->pressure(), P0, tol);
+    for(size_t i = 0; i < gas1->stateSize()-2; i++)
+    {
+        EXPECT_NEAR(gas1->massFraction(i), gas2->massFraction(i), tol);
+    }
```
bryanwweber: Since the test is meant to check that initializing the reactor outside a network doesn't segfault, I don't think you necessarily need the integration steps. IMO, it makes more sense to compare to things like the number of components in the reactor, etc. IMO it also makes more sense to initialize a plain `Reactor` to check the parent class directly, rather than relying on the derived classes to inherit properly and not override anything.

bryanwweber (commented):
Thanks @anthony-walker! These changes look good to me now, with the exception of the comment below about the test.

test/zeroD/test_zeroD.cpp
bryanwweber: I see now that @speth suggested adding it to a `ReactorNetwork` after initializing, so you should add a comment to that effect (and ignore the first part of my comment above 😝 ).

As one more suggestion, please limit line lengths to 88 characters (I think, see `CONTRIBUTING.md`) for the comment at the top of the test and there should be one space after the `//` for an inline comment.

anthony-walker: I agree that a `Reactor` class will be better. I think the integration may be valuable to address @speth 's concern here.

> One suggestion for the test case would be to make sure that the reactor still works correctly if it's initialized without a `ReactorNet` and then later added to one. I don't see a reason why this wouldn't work, but it would be good to check.

ischoegl (commented):
One more quibble … your commit messages don’t follow guidelines (per CONTRIBUTING.md):

> Use descriptive commit messages (summary line of no more than 72 characters, followed by a blank line and a more detailed summary, if any)

Would you mind updating those messages? Sorry about being nitpicky … it’s simple to fix using ‘reword’ when rebasing.

test/zeroD/test_zeroD.cpp
bryanwweber: Yep @anthony-walker, I think our comments crossed in the wires :joy:

ischoegl: @anthony-walker ... thanks for updating! Could you address the one remaining comment? (see above):

> As one more suggestion, please limit line lengths to 88 characters (I think, see CONTRIBUTING.md) for the comment at the top of the test and there should be one space after the // for an inline comment.

anthony-walker: @bryanwweber I have limited the length in my vs-code to 88 lines, will this take care of it or is there something else you had in mind?

test/zeroD/test_zeroD.cpp
```diff
@@ -0,0 +1,67 @@
+#include <cstdio>
+#include <string>
+#include "gtest/gtest.h"
+#include "cantera/thermo.h"
+#include "cantera/kinetics.h"
+#include "cantera/zerodim.h"
+
+using namespace Cantera;
+
+//This test ensures that prior reactor initialization of a reactor does not affect later integration within a network. This example was adapted from test_reactor.py::test_equilibrium_HP.
```
bryanwweber: @anthony-walker this still looks longer than 88 characters. Also, it doesn't look like you added a space after the comment slashes.

anthony-walker: @bryanwweber I addressed the inline comments on my local, just wanted to make sure I have the line length correct before I push again. It is showing mine as limited but I will look again to make sure.

bryanwweber: Make sure you don't have soft wrapping turned on

ischoegl (changes requested):
Quick comment: the new test failed on the windows runners and needs tweaks, see
```
cl /Fobuild\test\zeroD\test_zeroD.obj /c test\zeroD\test_zeroD.cpp /EHsc /MD /nologo /D_SCL_SECURE_NO_WARNINGS /D_CRT_SECURE_NO_WARNINGS /W3 /DNDEBUG /Iext\googletest\googletest\include /Iext\googletest\googlemock\include /Iinclude /I3rdparty\boost test_zeroD.cpp
test\zeroD\test_zeroD.cpp(44): error C2131: expression did not evaluate to a constant
test\zeroD\test_zeroD.cpp(44): note: failure was caused by call of undefined function or one not declared 'constexpr'
test\zeroD\test_zeroD.cpp(44): note: see usage of 'Cantera::Reactor::neq'
test\zeroD\test_zeroD.cpp(45): error C2131: expression did not evaluate to a constant
test\zeroD\test_zeroD.cpp(45): note: failure was caused by call of undefined function or one not declared 'constexpr'
test\zeroD\test_zeroD.cpp(45): note: see usage of 'Cantera::Reactor::neq'
scons: *** [build\test\zeroD\test_zeroD.obj] Error 2
```

test/zeroD/test_zeroD.cpp
```diff
@@ -7,53 +7,57 @@
...
+    // Secondary reactor for comparison
     Reactor reactor2;
     reactor2.setKineticsMgr(*kin2);
     reactor2.setThermoMgr(*gas2);
     reactor2.initialize(0);
-    //Get state of reactors
-    double state1 [reactor1.neq()];
-    double state2 [reactor2.neq()];
+    // Get state of reactors
+    double *state1 = new double[reactor1.neq()];
```
ischoegl: Avoid raw pointers (use `std::uniqe_ptr` or `std::shared_ptr` instead) ... although it looks like you may want to use `vector_fp` or `std::vector<double>` actually?

ischoegl (commented):
Thanks for the CI fix suggestion. While CI has not finished at the moment, here's a comment.

test/zeroD/test_zeroD.cpp
anthony-walker: Why avoid raw pointers? The method takes a `double` pointer and fills that array with the current state of the reactor, so I don't believe `vector` will work here.

ischoegl: Good question … it follows coding guidelines (`CONTRIBUTING.md`):

> Avoid manual memory management (i.e. `new` and `delete`), preferring to use standard library containers, as well as `std::unique_ptr` and `std::shared_ptr` when dynamic allocation is required.

For `std::vector`, just pass the memory it owns (see [here](https://www.cplusplus.com/reference/vector/vector/data/)). That approach is used extensively elsewhere in Cantera. You mainly see `vector_fp` which is an internal shortcut.

ischoegl: @anthony-walker … thanks for taking care of this comment. For my own feedback, that’s about it (although the `EXPECT_NEAR` checks could be simplified). But that’s really super minor.

ischoegl (dismissed):

test/zeroD/test_zeroD.cpp
anthony-walker: @ischoegl I appreciate the constructive feedback. It has been a learning experience that I can apply to my larger pull request.

test/zeroD/test_zeroD.cpp
```diff
@@ -0,0 +1,69 @@
...
+    std::vector<double> state1 (reactor1.neq());
+    std::vector<double> state2 (reactor2.neq());
+    reactor1.getState(state1.data());
+    reactor2.getState(state2.data());
+    // Compare the reactors.
+    EXPECT_EQ(reactor1.neq(), reactor2.neq());
+    double tol = 1e-14;
+    EXPECT_NEAR(state1[0], state2[0], tol);
+    EXPECT_NEAR(state1[1], state2[1], tol);
+    for(size_t i = 3; i < reactor1.neq(); i++)
```
ischoegl: One thing I hadn't noticed before: why are you skipping `i=2`? Shouldn't all state variables be comparable?

Also, as minor formatting suggestions: usually Cantera doesn't use a line break before `{` in a `for` loop (they're typically on the same line). There definitely should be a space after the `for`, though. Beyond, it may also help to add a blank line before each of your comments, as it makes your code easier to parse visually. 

I'll let @bryanwweber make the final approval, as he was the one who started the review.

test/zeroD/test_zeroD.cpp
```diff
@@ -0,0 +1,69 @@
...
+    double P0 = 1013250;
+    std::string X0 = "H2:1.0, O2:0.5, AR:8.0";
+    // Reactor solution, phase, and kinetics objects
+    std::shared_ptr<Solution> sol1 = newSolution("h2o2.yaml");
+    std::shared_ptr<ThermoPhase> gas1 = sol1->thermo();
+    std::shared_ptr<Kinetics> kin1 = sol1->kinetics();
+    gas1->setState_TPX(T0, P0, X0);
+    // Set up reactor object
+    Reactor reactor1;
+    reactor1.setKineticsMgr(*kin1);
```
ischoegl: You could use `reactor1.insert(sol1);` to simplify the code here and below.

test/zeroD/test_zeroD.cpp
anthony-walker: So, I adapted this example from `test_equilibrium_HP` where pressure is held as a constant and not compared. When it is compared it causes the test to fail based on the specified tolerance in the test. I am not entirely certain why but I assumed is was due to the methods used and was justified at the time. This example is the same in that `U` is held constant but varies slightly on the order of 1e-8. I will look further into it though for a more thorough explanation.

test/zeroD/test_zeroD.cpp
anthony-walker: I was actually able to update this in multiple places. Thank you for the suggestion.

test/zeroD/test_zeroD.cpp
ischoegl: It’s actually interesting that this unit test works to the specified tolerances at all: the test compares a result that involves reaction kinetics with one that uses equilibrium. While there shouldn’t be much of a difference for these simple mixtures in theory, it’s not completely intuitive that comparisons are that close. I’m saying this as I’ve seen significant differences between equilibrium results and kinetics calculations for some more complex mixtures.

On the other hand, I f this is the exact equivalent to a Python example, why don’t you just use the ‘hp’ equilibrium case, plus `IdealGasConstPressureReactor` objects for this unit test? That it works for HP and not for UV could just be a quirk of some solvers.

anthony-walker: @ischoegl I could reduce the tolerance and the unit test would work. I believe it passes for 1e-7 but I didn't think that was a good answer. It doesn't work for HP either though if that was unclear. Also, `Reactor` was selected to be the test reactor because it is the base class and not a derived class.

bryanwweber: IMO, 1e-7 is a pretty good tolerance on the in internal energy... We're talking about a value on the order of 1e6 right?

ischoegl: > It doesn't work for HP either though if that was unclear. 

no, I guess it wasn’t clear to me … in that case, I’d agree with @bryanwweber and just reduce the tolerance. 1e-14 is really tight if you’re not comparing apples to apples. 

Otherwise, please have another look at the formatting as suggested. Also, note that there are some additional white spaces missing on line 17, I.e. `double P0 = 10*OneAtm;` should be `double P0 = 10 * OneAtm;`

anthony-walker: @ischoegl I will fix the formatting. I always forget that. Hopefully I will catch on to that eventually. I will also make the change to 1e-7. @bryanwweber the values are pretty large, 971244.30 to be more specific.

bryanwweber: @anthony-walker Right, so that's ≈1e6. A tolerance of 1e-14 is on the order of machine precision, and applying several arithmetic operations (integrating or solving equilibrium) is going to decrease the precision of the answer. Since there's different methods being compared here, a tolerance down at machine precision is too tight. On top of that, since the values are so large, at some point the number of digits is no longer physically meaningful. If you look at some of the tests in the regression tests, the tolerances are on the order of 1e-2 to account for differences between compilers, platforms, etc.

Anyhow, all of which is to say that sometimes, changing the tolerance is a legitimate answer 😊

anthony-walker: @bryanwweber Thank you for the explanation, it makes more sense now. I will have to look through those tests. Have all concerns been addressed with the PR then?

bryanwweber (approved):
Looks good to me, thanks @anthony-walker

test/zeroD/test_zeroD.cpp
speth: > It’s actually interesting that this unit test works to the specified tolerances at all: the test compares a result that involves reaction kinetics with one that uses equilibrium. While there shouldn’t be much of a difference for these simple mixtures in theory, it’s not completely intuitive that comparisons are that close. I’m saying this as I’ve seen significant differences between equilibrium results and kinetics calculations for some more complex mixtures.

Since Cantera computes reverse reaction rates from equilibrium constants, if you have a reaction mechanism with at least as many independent, reversible reactions as species, I think you should always get "exactly" the same result from both a kinetic simulation out to steady state and the equilibrium solver. Given that both the equilibrium and ODE solvers are only accurate to within certain bounds, the level of exactness depends on the tolerances of these solvers, not to something approximating accumulated floating point roundoff.


> So, I adapted this example from `test_equilibrium_HP` where pressure is held as a constant and not compared. When it is compared it causes the test to fail based on the specified tolerance in the test.

For future reference, note that the Python `assertNear` function is applying a combined relative and absolute tolerance. The GTest `ASSERT_NEAR` is just an absolute tolerance. To get a relative tolerance, which is generally more useful, we often end up writing something like `ASSERT_NEAR(x0, x1, 1e-7 * std::abs(x0))`.

PR #1066, opened 2021-06-25
Fix extinction temperature definition in diffusion_flame_extinction.py

### LINKED ISSUE #1064: Extinction criteria in diffusion_flame_extinction.py
The extinction criteria in the current Cantera 2.5 version of [diffusion_flame_extinction.py](https://cantera.org/examples/python/onedim/diffusion_flame_extinction.py.html) only works if `temperature_limit_extinction` is set equal to the higher gas inlet temperature.  In the current version both the oxidizer inlet and extinction temperature are set to 500 K.  The code would be improved if the extinction limit were defined based on the inlet temperatures, such as changing line 67 to:

`temperature_limit_extinction = max(f.oxidizer_inlet.T, f.fuel_inlet.T)`

Alternatively, temperature_limit_extinction could be set using a dT above the max inlet temperature, and then extinction checked for with `np.max(f.T) < temperature_limit_extinction`, as in the <= 2.4 version of the example.  (Instead of the current `np.isclose()` in line 144)

The stopping criteria for the example was changed with [commit f13bbb9c](https://github.com/Cantera/cantera/commit/f13bbb9c82921b48227a6340ceaa7bf16720d799), and this issue is discussed in the User Group [here](https://groups.google.com/g/cantera-users/c/YU48mP0Ka5g/m/QaI-VQpEAAAJ).

d-e-t:
Assigning extinction temperature based on inlet temperatures, for consistancy with the use of np.isclose() for the extinction criteria in Cantera 2.5 version of this Python example.  Otherwise the extinction criteria only works if temperature_limit_extinction is set equal to the higher gas inlet temperature.

Closes #1064

### DIFF
--- a/interfaces/cython/cantera/examples/onedim/diffusion_flame_extinction.py
+++ b/interfaces/cython/cantera/examples/onedim/diffusion_flame_extinction.py
@@ -53,7 +53,7 @@ f.set_refine_criteria(ratio=3.0, slope=0.1, curve=0.2, prune=0.03)
 
 # Define a limit for the maximum temperature below which the flame is
 # considered as extinguished and the computation is aborted
-temperature_limit_extinction = 500  # K
+temperature_limit_extinction = max(f.oxidizer_inlet.T, f.fuel_inlet.T)
 
 # Initialize and solve
 print('Creating the initial solution')
@@ -114,6 +114,7 @@ while True:
     strain_factor = alpha[-1] / alpha[n_last_burning]
     # Create an initial guess based on the previous solution
     # Update grid
+    # NB: Grid scaling changes the diffusion flame width
     f.flame.grid *= strain_factor ** exp_d_a
     normalized_grid = f.grid / (f.grid[-1] - f.grid[0])
     # Update mass fluxes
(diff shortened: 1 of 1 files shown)


interfaces/cython/cantera/examples/onedim/diffusion_flame_extinction.py
```diff
@@ -114,6 +114,7 @@
     strain_factor = alpha[-1] / alpha[n_last_burning]
     # Create an initial guess based on the previous solution
     # Update grid
+    # NB: Grid scaling changes the diffusion flame width
```
speth: As a minor suggestion to improve readability, we're trying to avoid use of some more obscure abbreviations, especially these Latin ones, so I would just write "note" here.

speth (changes requested):
One small suggestion, otherwise this looks fine to me.

interfaces/cython/cantera/examples/onedim/diffusion_flame_extinction.py
d-e-t: I guess 'NB' is a habit I picked up in British boarding school that I use without thinking about much :-)   I will note well that it is now arcane and obscure!

ischoegl (approved):

speth (approved):

PR #1069, opened 2021-07-11
Bug fix: CK_Multi recognized as multicomponent transport method in fl…

lavdwall:
This is a minor fix to make sure that the Chemkin-type multi-component transport type (CK_Multi) is treated properly by the 1D flame models.

Best regards,
Laurien

### DIFF
--- a/src/oneD/StFlow.cpp
+++ b/src/oneD/StFlow.cpp
@@ -153,7 +153,7 @@ void StFlow::resetBadValues(double* xg)
 void StFlow::setTransport(Transport& trans)
 {
     m_trans = &trans;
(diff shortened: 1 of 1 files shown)


ischoegl:
@lavdwall ... thank you! The fix looks straight-forward. Per https://github.com/Cantera/cantera/blob/1420a4cf2901c8ae59d895fec4ab19fc41580207/include/cantera/transport/MultiTransport.h#L35-L37
there are two model names, with `CK_Mode` meaning Chemkin mode for the type of polynomial fits to temperature.

ischoegl (approved):

PR #1072, opened 2021-07-16
Extend BinarySolutionTabulatedThermo by tabulated partial molar volumes

dschmider-HSOG:
Enable to consider non-constant partial molar volumes provided by tabular data, since the partial molar volume of typical battery intercalation materials shows strong dependency on e.g. lithium stoichiometry. The tabular data is included in the same table as enthalpy and entropy data.

-implement optional partialMolarVolume keyword in table(thermo)
-BinarySolutionTabulatedThermo::interpolate is adjusted to be more general and also interpolate partial molar volume
-BinarySolutionTabulatedThermo::integrate is added since partial molar volumes are often needed in integrated form (integrated over lithium stoichiometry)
-BinarySolutionTabulatedThermo::calcDensity is adjusted since density is not constant anymore and changes with varying partial molar volumes

**Checklist**


include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -147,12 +147,19 @@ class BinarySolutionTabulatedThermo : public IdealSolidSolnPhase
     virtual void getParameters(AnyMap& phaseNode) const;
     virtual void initThermoXML(XML_Node& phaseNode, const std::string& id_);
 
+    void getIntegratedPartialMolarVolumes(doublereal* vbar);
+    void getPartialMolarVolumes(doublereal* vbar) const;
```
bryanwweber: Please avoid using `doublereal` in new code, preferring the standard `double` instead.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -147,12 +147,19 @@ class BinarySolutionTabulatedThermo : public IdealSolidSolnPhase
...
+    virtual void calcDensity();
+
 protected:
     //! If the compositions have changed, update the tabulated thermo lookup
     virtual void compositionChanged();
 
     //! Species thermodynamics interpolation functions
-    std::pair<double,double> interpolate(double x) const;
+    double interpolate(double x,const vector_fp& molefrac,const vector_fp& inputData) const;
```
bryanwweber: Function definitions should please have one space after a comma. I'm not sure exactly how long this line is, but we prefer to keep line lengths to 88 characters or less, if possible.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1357,18 +1357,21 @@ def get_yaml(self, out):
 
 class table:
     """User provided thermo table for BinarySolutionTabulatedThermo"""
-    def __init__(self, moleFraction=([],''), enthalpy=([],''), entropy=([],'')):
+    def __init__(self, moleFraction=([],''), enthalpy=([],''), entropy=([],''), partialMolarVolume=([],'')):
```
bryanwweber: Since the CTI/XML formats are going to be removed in the next few versions, we are preferring to add new features only to the YAML format. Unless there is already a significant userbase for this feature who are using CTI/XML files, we'd prefer to not change any of the converters.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -90,26 +91,50 @@ void BinarySolutionTabulatedThermo::initThermo()
...
+        if (table.hasKey("partialMolarVolume")){
+        	vbar = table.convertVector("partialMolarVolume", "m3/kmol", N);
+        	for(size_t i=0; i<N; i++){
+        		vbar[i]+=m_speciesMolarVolume[1];
+        	}
+        }
+        else{
+        	vbar.resize(x.size());
+        	for(size_t i=0; i<N; i++){
+        		vbar[i]=m_speciesMolarVolume[0];
```
bryanwweber: Please only use spaces for indentation, with 4 spaces per indent.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -90,26 +91,50 @@ void BinarySolutionTabulatedThermo::initThermo()
         size_t N = x.size();
         vector_fp h = table.convertVector("enthalpy", "J/kmol", N);
         vector_fp s = table.convertVector("entropy", "J/kmol/K", N);
+        vector_fp vbar;
+        // Check for partialMolarVolume tag in tabulatedThermo table, take m_speciesMolarVolume[0] if false (for backward compatibility)
+        if (table.hasKey("partialMolarVolume")){
+        	vbar = table.convertVector("partialMolarVolume", "m3/kmol", N);
+        	for(size_t i=0; i<N; i++){
```
bryanwweber: Following the style elsewhere, there should be spaces around the `=` and `<`. Also, one space before `{`. This occurs in a few places.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -48,7 +48,8 @@ void BinarySolutionTabulatedThermo::_updateThermo() const
     bool x_changed = (m_xlast != xnow);
 
     if (x_changed) {
-        std::tie(m_h0_tab, m_s0_tab) = interpolate(xnow);
+        m_h0_tab  = interpolate(xnow,m_molefrac_tab,m_enthalpy_tab);
+        m_s0_tab  = interpolate(xnow,m_molefrac_tab,m_entropy_tab);
```
bryanwweber: Please have only one space on the left of the `=`.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -90,26 +91,50 @@ void BinarySolutionTabulatedThermo::initThermo()
         size_t N = x.size();
         vector_fp h = table.convertVector("enthalpy", "J/kmol", N);
         vector_fp s = table.convertVector("entropy", "J/kmol/K", N);
+        vector_fp vbar;
+        // Check for partialMolarVolume tag in tabulatedThermo table, take m_speciesMolarVolume[0] if false (for backward compatibility)
+        if (table.hasKey("partialMolarVolume")){
```
bryanwweber:
```suggestion
        if (table.hasKey("partialMolarVolume")) {
```

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -90,26 +91,50 @@ void BinarySolutionTabulatedThermo::initThermo()
...
+        	vbar = table.convertVector("partialMolarVolume", "m3/kmol", N);
+        	for(size_t i=0; i<N; i++){
+        		vbar[i]+=m_speciesMolarVolume[1];
```
bryanwweber: Please make sure there is one space on each side of the `+=`.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -90,26 +91,50 @@ void BinarySolutionTabulatedThermo::initThermo()
...
+        	}
+        }
+        else{
+        	vbar.resize(x.size());
```
bryanwweber: Why not use `N` here?

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -161,29 +186,51 @@ void BinarySolutionTabulatedThermo::initThermoXML(XML_Node& phaseNode, const std
             getFloatArray(dataNode, h, true, "J/kmol", "enthalpy");
             getFloatArray(dataNode, s, true, "J/kmol/K", "entropy");
 
+            // Check for partialMolarVolume tag in tabulatedThermo table, take m_speciesMolarVolume[0] if false (for backward compatibility)
```
bryanwweber: As I mentioned earlier, unless there is a substantial existing userbase for this feature, there's no need to add it to the XML reader.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -211,27 +258,72 @@ void BinarySolutionTabulatedThermo::initThermoXML(XML_Node& phaseNode, const std
...
+void BinarySolutionTabulatedThermo::integrate(vector_fp& inputData,vector_fp& integratedData,vector_fp& moleFraction)
+{
+	integratedData.resize(inputData.size());
+	for(size_t i = 0; i < integratedData.size(); i++){
+		if (i==0 and moleFraction[0]==0){
+			integratedData[i]=0;
+		}
+		else if (i==0 and moleFraction[0]>0){
+			integratedData[i]=moleFraction[i]*inputData[0];
+		}
```
bryanwweber: I think you can avoid this conditional by doing the first element outside the loop and starting the loop at 1 instead of 0. Also, these two conditions are identical, since if `moleFraction[i] == 0` is true, then `moleFraction[i]*inputData[i]` will also be 0. There's no need to have the condition, I think you can just do the multiplication.

bryanwweber (dismissed):
Hi! Thank you for submitting these changes, the functionality looks like it will be useful. I have several formatting/style comments, if you can please implement them while the technical review is ongoing. Can you also please add a test for this new functionality? If any of my comments are not clear, please post a comment and ask about it, I am more than happy to help. Thank you!

src/thermo/BinarySolutionTabulatedThermo.cpp
decaluwe: @dschmider-HSOG Depending on what code editor you are using, you can likely set this as a preference, so that a tab automatically inserts 4 spaces.

decaluwe:
Hi, @dschmider-HSOG , I'll echo @bryanwweber's comments - many thanks for this excellent contribution!

In addition to @bryanwweber's comments on the coding style, there are a few places where a little bit more documentation will help a user understand some of this functionality.  
- To second @bryanwweber's comment, please add a test to `test/thermo/BinarySolutionTabulatedThermo_Test.cpp` which will test to make sure that no future changes to Cantera break this functionality.  Ideally, load the phase and test the results at a few compositions, comparing to hand-calculations of what you know the answer should be.
- The documentation at the top of the header file `include/cantera/thermo/BinarySolutionTabulatedThermo.h` should also be edited to reflect (and explain) this new capability.
- In the same file, a small comment explaining for users what getIntegratedPartialMolarVolumes is for, either in-line where the function is declared, or also at the top of this file.

Other than that, I'll leave a few small comments here and there, and like Bryan, am happy to help get this across the finish line, as needed.  It is really quite close to ready, though.  Thanks again!

include/cantera/thermo/BinarySolutionTabulatedThermo.h
decaluwe: In other words:

```suggestion
    double interpolate(double x, const vector_fp& molefrac,
        const vector_fp& inputData) const;
```

There is a similar issue on line 162 (next declaration)

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -211,27 +258,72 @@ void BinarySolutionTabulatedThermo::initThermoXML(XML_Node& phaseNode, const std
...
+{
+	double x[2];
+	IdealSolidSolnPhase::getPartialMolarVolumes(vbar);
+	Phase::getMoleFractions(x);
+	vbar[0]=interpolate(x[0],m_molefrac_tab,m_partial_molar_volume_tab);
+}
+
+void BinarySolutionTabulatedThermo::calcDensity()
+{
+	// Fallback to IdealSolidSolnPhase::calcDensity() if called before m_integrated_partial_molar_volume_tab is initialized
```
decaluwe: I understand why you cannot use the other routine if `m_integrated_partial_molar_volume_tab` is initialized, but wouldn't this still be an erroneous value?  

Is there a specific case where this will be called before the table is initialized, during setup of the class, or is it possible that a user will call it before the initialization is complete?  If the latter, I'd rather find a way for this call to trigger the initialization, or throw an informative error message, rather than returning an incorrect value.

dschmider-HSOG:
Hi,
thank you very much @decaluwe and @bryanwweber, I really appreciate your comments. 
I have just returned from holiday and will be working on the requested changes throughout this week. Since this is my first ever pull request, I have a rather general question:
Should I amend the previous commit to include all the changes or do you prefer follow-up commits to address the requested changes?

bryanwweber:
HI @dschmider-HSOG Welcome back! It's up to you whether to amend the previous commit or add new ones. I usually try to keep a logical unit of work per commit.

dschmider-HSOG:
Hi,
I hope I got all the formatting issues fixed now.
Regarding the test I might need some help. Can I just extend the existing BinarySolutionTabulatedThermo.yaml in the test data by a partial molar volume entry in the table and add test cases in BinarySolutionTabulatedThermo_Test.cpp or should I generate a separate data file?
Furthermore I will edit the documentation in the header file and also add some comments on the functions in a follow-up commit.

src/thermo/BinarySolutionTabulatedThermo.cpp
dschmider-HSOG: That is sort of a workaround. When I implemented the calcDensity() function, the code crashed because the calcDensity() routine got invoked when m_integrated_partial_molar_volume_tab was still empty. 
I wondered how this is possible because actually m_integrated_partial_molar_volume_tab is initialized at the end of void BinarySolutionTabulatedThermo::initThermo(), but I could not track down, which function called calcDensity() before initThermo() initialized the phase. So that was the only solution I could come up with.

bryanwweber:
> Regarding the test I might need some help. Can I just extend the existing BinarySolutionTabulatedThermo.yaml in the test data by a partial molar volume entry in the table and add test cases in BinarySolutionTabulatedThermo_Test.cpp or should I generate a separate data file?

Hi @dschmider-HSOG Sure, editing that same file should be fine. Thanks!

bryanwweber:
@dschmider-HSOG are you able to finish this pull request up? It looks like a neat function that'd be great to have merged. Please let me know if you have any questions!

dschmider-HSOG:
Hi @bryanwweber, thanks for asking. I was occupied with other topics in the meantime, but should soon be able to complete the commit.

dschmider-HSOG:
Hi,
finally I pushed my recent changes. Since it has been quite some time since the last commit I also rebased my changes on the latest version of the main branch. I hope this was the correct way of doing it.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -143,16 +191,58 @@ class BinarySolutionTabulatedThermo : public IdealSolidSolnPhase
...
+                       const vector_fp& inputData) const;
+
+    //! Numerical derivation of the molar volume table
+    /*!
+     *  Tabulated values are only interpolated within the limits of the provided mole
+     *  fraction. If these limits are exceeded, the values are capped at the lower or
+     *  the upper limit.
+     */
+    void derive(vector_fp& inputData, vector_fp& derivedData,
+                vector_fp& moleFraction);
```
speth: A few comments on this function:
- This is the numerical _derivative_ of the molar volume
- I think `diff` would be a better name for this function is used, the mole fraction vector is `m_molefrac_tab`
- I don't think the `moleFraction` argument is needed -- the only way this function is called, 
- The usual order of arguments would be to have the input arguments (`inputData` and `moleFraction` first, followed by the output argument.
- The input arguments should be marked as `const`
- The function itself is can be `const` as well.

include/cantera/thermo/BinarySolutionTabulatedThermo.h
```diff
@@ -143,16 +191,58 @@ class BinarySolutionTabulatedThermo : public IdealSolidSolnPhase
...
-    //! Species thermodynamics interpolation functions
-    std::pair<double,double> interpolate(double x) const;
+    //! Species thermodynamics linear interpolation function
+    /*!
+     *  Tabulated values are only interpolated within the limits of the provided mole
+     *  fraction. If these limits are exceeded, the values are capped at the lower or
+     *  the upper limit.
+     */
+    double interpolate(double x, const vector_fp& molefrac,
+                       const vector_fp& inputData) const;
```
speth: I don't think the `molefrac` vector needs to be an input argument -- it is always the `m_molefrac_tab` variable which is available in this function as well.

It would be useful to document the input and output arguments in the method docstring, e.g. with `@param x  (some description` and `@returns  the value of ...`.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -563,9 +563,9 @@ class ThermoPhase : public Phase
         throw NotImplementedError("ThermoPhase::getPartialMolarVolumes");
     }
 
-    //! @}
-    //! @name Properties of the Standard State of the Species in the Solution
-    //! @{
+    //@}
+    /// @name Properties of the Standard State of the Species in the Solution
+    //@{
```
speth: This change should be reverted - the lines containing `@{` and `@}` need to be Doxygen comment lines (starting with `//!` or `///`, as they were.

test/data/BinarySolutionTabulatedThermo.yaml
```diff
@@ -43,6 +43,16 @@ phases:
       19.2971, 19.2977, 16.2213, 19.298, 19.2978, 19.2945, 19.2899, 19.2877,
       19.2882, 19.2882, 19.2882, 19.2882, 19.2882, 19.2885, 19.2876, 19.2837,
       19.2769, 19.285, 19.31, 19.3514]
+    molarVolume: [34.582988, 34.677730, 34.776275, 34.875760, 34.973980, 35.069268,
```
speth: In the YAML format, we generally use lowercase words separated by spaces, so this should be `molar-volume`.

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -211,27 +316,66 @@ void BinarySolutionTabulatedThermo::initThermoXML(XML_Node& phaseNode, const std
...
+    vbar[m_kk_tab] = interpolate(Phase::moleFraction(m_kk_tab), m_molefrac_tab,
+                                 m_partial_molar_volume_1_tab);
+    vbar[1-m_kk_tab] = interpolate(Phase::moleFraction(m_kk_tab), m_molefrac_tab,
+                                   m_partial_molar_volume_2_tab);
+}
+
+void BinarySolutionTabulatedThermo::calcDensity()
+{
+    {
+        double vmol = interpolate(Phase::moleFraction(m_kk_tab), m_molefrac_tab,
```
speth: There's an extra, unnecessary layer of braces and indentation around the body of this function.

Also, I don't think the `Phase::` qualifier is needed before calling `moleFraction` (here or anywhere else in this class).

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -211,27 +316,66 @@ void BinarySolutionTabulatedThermo::initThermoXML(XML_Node& phaseNode, const std
...
+                        (moleFraction[i+1] - moleFraction[i-1]);
+            }
+        }
+    }
+    else {
+        derivedData.front() = 0;
+    }
+}
+
+void BinarySolutionTabulatedThermo::getPartialMolarVolumes(doublereal* vbar) const
```
speth: There's a rogue `doublereal` here that should just be `double` (likewise for this method in the header file).

src/thermo/BinarySolutionTabulatedThermo.cpp
```diff
@@ -211,27 +316,66 @@ void BinarySolutionTabulatedThermo::initThermoXML(XML_Node& phaseNode, const std
...
+                (moleFraction.back() - moleFraction[moleFraction.size()-2]);
+
+        if (inputData.size() > 2) {
+            for (size_t i = 1; i < inputData.size()-1; i++) {
+                derivedData[i] = (inputData[i+1] - inputData[i-1]) /
+                        (moleFraction[i+1] - moleFraction[i-1]);
+            }
+        }
+    }
+    else {
```
speth: A minor formatting suggestion:
```suggestion
    } else {
```

speth (changes requested):
Thanks for providing these updates. I appreciated the detailed description of the additions in the class docstring. 

Besides the handful of comments above and below*, I found an alternative to duplicating the implementation of `addSpecies` from `IdealSolidSolnPhase`, which I've pushed to my fork (https://github.com/speth/cantera/tree/tabulated_partial_molar_volumes). You can either cherry-pick that commit onto your branch here, or if you want I can push that change to this branch.

\*sorry, something funny happened with the GitHub plugin for VS Code and it posted part of this before I was finished reviewing -- please take a look at the comments from the prior review as well.

src/thermo/BinarySolutionTabulatedThermo.cpp
dschmider-HSOG: I saw that in the generation of the doxygen file, the information that this function is reimplemented from the parent class is lost if I change to double. Therefore I thought it would be better to keep doublereal. But sure I can change this to double.

dschmider-HSOG:
Thanks @speth for reviewing. I've implemented your suggestions.

Regarding `addSpecies`: I think it should be possible to drop the call for `calcDensity` in `IdealSolidSolnPhase::addSpecies` completely. I think there is no need to calculate the density at this point, since `calcDensity` is invoked in the end of `setupPhase` or `importPhase`. All tests pass without errors if I simply remove it.

src/thermo/BinarySolutionTabulatedThermo.cpp
speth: Interesting. I thought Doxygen was smart enough to see through the `typedef` and know that these methods were the same. I'd suggest changing it to `double` anyways. We'll get around to eliminating `doublereal` on a wider scale eventually.

speth:
> Regarding `addSpecies`: I think it should be possible to drop the call for `calcDensity` in `IdealSolidSolnPhase::addSpecies` completely. I think there is no need to calculate the density at this point, since `calcDensity` is invoked in the end of `setupPhase` or `importPhase`. All tests pass without errors if I simply remove it.

That's true in most cases, but not if you add species to the phase after the fact (which is of course not allowed in the case of `BinarySolutionTabulatedThermo`, but could be for `IdealSolidSolnPhase`. So I think this call needs to remain to keep the phase in a consistent state.

speth:
Hi @dschmider-HSOG, I think you haven't yet pushed the updates to this branch.

dschmider-HSOG:
> That's true in most cases, but not if you add species to the phase after the fact (which is of course not allowed in the case of `BinarySolutionTabulatedThermo`, but could be for `IdealSolidSolnPhase`. So I think this call needs to remain to keep the phase in a consistent state.

I see. But wouldn't the implementation with the `ready()`  function that you proposed in your fork also result in the fact that `calcDensity()` will never get called from within `IdealSolidSolnPhase::addSpecies`? `m_molefrac_tab` would always be empty for an `IdealSolidSolnPhase`.
If it is possible to add species after the fact, I guess it is even more important that `BinarySolutionTabulatedThermo` gets its own `addSpecies` method two prevent adding more than two species. This is checked during `initThermo()` but shouldn't it then also be checked in `addSpecies`?

speth:
> I see. But wouldn't the implementation with the `ready()` function that you proposed in your fork also result in the fact that `calcDensity()` will never get called from within `IdealSolidSolnPhase::addSpecies`? `m_molefrac_tab` would always be empty for an `IdealSolidSolnPhase`.

`ready()` is a virtual function, which means that the implementation in `BinarySolutionTabulatedThermo` will be called only if the object is actually a `BinarySolutionTabulatedThermo`. Otherwise, the base class method (in this case, `Phase::ready()`) will be called, and that returns `true` as long as there are any species in the phase.

> If it is possible to add species after the fact, I guess it is even more important that `BinarySolutionTabulatedThermo` gets its own `addSpecies` method two prevent adding more than two species. This is checked during `initThermo()` but shouldn't it then also be checked in `addSpecies`?

Yes, that's a good point. But this can be a pretty minimal implementation, which just checks that condition an then calls `IdealSolidSolnPhase::addSpecies` to avoid any code duplication.

dschmider-HSOG:
> `ready()` is a virtual function, which means that the implementation in `BinarySolutionTabulatedThermo` will be called only if the object is actually a `BinarySolutionTabulatedThermo`. Otherwise, the base class method (in this case, `Phase::ready()`) will be called, and that returns `true` as long as there are any species in the phase.

Sorry, I somehow overlooked the fact that `Phase::ready()` exists. You're absolutely right.

> Yes, that's a good point. But this can be a pretty minimal implementation, which just checks that condition an then calls `IdealSolidSolnPhase::addSpecies` to avoid any code duplication.

I changed my last commit to incorporate the `ready()` method and also check the condition of max. 2 species.

speth (commented):
Thanks, all of these changes look good to me. One other small thing that I forgot to mention is that there should be an update to `doc/sphinx/yaml/phases.rst` to describe the additional parameter in the YAML API documentation. This file is used to populate the information that ends up [here](https://cantera.org/documentation/dev/sphinx/html/yaml/phases.html#binary-solution-tabulated) on the Cantera website.

speth (approved):

dschmider-HSOG:
Thank you @decaluwe, @bryanwweber and @speth for helping me finish my very first pull request. I really appreciated your constructive feedback.

Best regards,
David

decaluwe:
Congrats, @dschmider-HSOG - so excited for your contribution!

PR #1076, opened 2021-07-17
Expose C++ Units to Python

ischoegl:

- Shorten `Units::str()` output
- Expose `CxxUnits` to Cython interface
- Implement Python API for `Units`
- Expose `Reaction.rate_coeff_units` and `ThermoPhase.standard_concentration_units`
- Expose `UnitSystem` to Python

The new feature can be used as follows:
```
In [1]: import cantera as ct
   ...: gas = ct.Solution("gri30.yaml")
   ...: gas.reaction(0)
   ...:
Out[1]: <ThreeBodyReaction: 2 O + M <=> O2 + M>

In [2]: gas.reaction(0).rate_coeff_units
Out[2]: <Units(1.0 m^6 / kmol^2 / s) at 7f928bee69f0>

In [3]: gas.standard_concentration_units
Out[3]: <Units(1.0 kmol / m^3) at 7f928bee6ab0>

In [4]: ct.Units("Pa")
Out[4]: <Units(1.0 kg / m / s^2) at 7f928bee6cf0>

In [5]: ct.UnitSystem().defaults
Out[5]: 
{'activation-energy': 'J/kmol',
 'current': 'A',
 'energy': 'J',
 'length': 'm',
 'mass': 'kg',
 'pressure': 'Pa',
 'quantity': 'kmol',
 'temperature': 'K',
 'time': 's'}
```
**Checklist**

**Other thoughts**

This PR seeks to expose units used in the C++ layer, rather than facilitating conversion of alternative units as proposed in #991.

### DIFF
--- a/include/cantera/base/Units.h
+++ b/include/cantera/base/Units.h
@@ -109,6 +109,9 @@ public:
     //! recognize an optional argument with a default value)
     UnitSystem() : UnitSystem({}) {}
 
+    //! Return default units used by the unit system
+    std::map<std::string, std::string> defaults() const;
+
     //! Set the default units to convert from when explicit units are not
     //! provided. Defaults can be set for mass, length, time, quantity, energy,
     //! and pressure. Conversion using the pressure or energy units is done only
--- a/include/cantera/base/Units.h
+++ b/include/cantera/base/Units.h
@@ -49,6 +49,11 @@ public:
     Units& operator*=(const Units& other);
 
     //! Provide a string representation of these Units
+    //! @param leading_one  print '1' if no units are in the numerator
+    std::string unit_str(bool leading_one=true) const;
+
+    //! Provide a string representation of these Units that includes the
+    //! conversion factor
     std::string str() const;
 
     //! Raise these Units to a power, changing both the conversion factor and
--- a/include/cantera/base/Units.h
+++ b/include/cantera/base/Units.h
@@ -35,7 +35,8 @@ public:
                    double quantity=0);
 
     //! Create an object with the specified dimensions
-    explicit Units(const std::string& name);
+    //! @param force_unity  ensure that conversion factor is equal to one
+    explicit Units(const std::string& name, bool force_unity=false);
 
     //! Returns `true` if the specified Units are dimensionally consistent
     bool convertible(const Units& other) const;
@@ -49,12 +50,8 @@ public:
     Units& operator*=(const Units& other);
 
     //! Provide a string representation of these Units
-    //! @param leading_one  print '1' if no units are in the numerator
-    std::string unit_str(bool leading_one=true) const;
-
-    //! Provide a string representation of these Units that includes the
-    //! conversion factor
-    std::string str() const;
+    //! @param skip_unity  do not print '1' if conversion factor is equal to one
+    std::string str(bool skip_unity=true) const;
 
     //! Raise these Units to a power, changing both the conversion factor and
     //! the dimensions of these Units.
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -53,8 +53,7 @@ cdef extern from "cantera/base/Units.h" namespace "Cantera":
     cdef cppclass CxxUnits "Cantera::Units":
         CxxUnits()
         CxxUnits(CxxUnits)
-        CxxUnits(string) except +translate_exception
-        string unit_str()
+        CxxUnits(string, cbool) except +translate_exception
         string str()
         double factor()
 
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -647,7 +647,7 @@ cdef extern from "cantera/base/YamlWriter.h" namespace "Cantera":
         void toYamlFile(string&) except +translate_exception
         void setPrecision(int)
         void skipUserDefined(cbool)
-        void setUnitSystem(CxxUnitSystem) except +translate_exception
+        void setUnitSystem(CxxUnitSystem&) except +translate_exception
 
 cdef extern from "cantera/equil/MultiPhase.h" namespace "Cantera":
     cdef cppclass CxxMultiPhase "Cantera::MultiPhase":
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -54,6 +54,7 @@ cdef extern from "cantera/base/Units.h" namespace "Cantera":
         CxxUnits()
         CxxUnits(CxxUnits)
         CxxUnits(string) except +translate_exception
+        string unit_str()
         string str()
         double factor()
 
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -49,6 +49,10 @@ cdef extern from "cantera/base/xml.h" namespace "Cantera":
         XML_Node* findID(string)
         int nChildren()
 
+cdef extern from "cantera/base/Units.h" namespace "Cantera":
+    cdef cppclass CxxUnits "Cantera::Units":
+        string str()
+
 cdef extern from "cantera/base/AnyMap.h" namespace "Cantera":
     cdef cppclass CxxAnyValue "Cantera::AnyValue"
 
@@ -208,6 +212,7 @@ cdef extern from "cantera/thermo/ThermoPhase.h" namespace "Cantera":
         size_t stateSize()
         void saveState(size_t, double*) except +translate_exception
         void restoreState(size_t, double*) except +translate_exception
+        CxxUnits standardConcentrationUnits() except +translate_exception
 
         # initialization
         void addUndefinedElements() except +translate_exception
@@ -420,6 +425,7 @@ cdef extern from "cantera/kinetics/Reaction.h" namespace "Cantera":
         cbool allow_nonreactant_orders
         cbool allow_negative_orders
         cbool usesLegacy()
+        CxxUnits rate_units
 
     cdef cppclass CxxElementaryReaction2 "Cantera::ElementaryReaction2" (CxxReaction):
         CxxElementaryReaction2()
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -51,6 +51,9 @@ cdef extern from "cantera/base/xml.h" namespace "Cantera":
 
 cdef extern from "cantera/base/Units.h" namespace "Cantera":
     cdef cppclass CxxUnits "Cantera::Units":
+        CxxUnits()
+        CxxUnits(CxxUnits)
+        CxxUnits(string) except +translate_exception
         string str()
 
 cdef extern from "cantera/base/AnyMap.h" namespace "Cantera":
@@ -1114,6 +1117,11 @@ ctypedef void (*transportMethod2d)(CxxTransport*, size_t, double*) except +trans
 ctypedef void (*kineticsMethod1d)(CxxKinetics*, double*) except +translate_exception
 
 # classes
+cdef class Units:
+    cdef CxxUnits units
+    @staticmethod
+    cdef copy(CxxUnits)
+
 cdef class SpeciesThermo:
     cdef shared_ptr[CxxSpeciesThermo] _spthermo
     cdef CxxSpeciesThermo* spthermo
--- a/interfaces/cython/cantera/_cantera.pxd
+++ b/interfaces/cython/cantera/_cantera.pxd
@@ -57,6 +57,11 @@ cdef extern from "cantera/base/Units.h" namespace "Cantera":
         string str()
         double factor()
 
+    cdef cppclass CxxUnitSystem "Cantera::UnitSystem":
+        CxxUnitSystem()
+        stdmap[string, string] defaults()
+        void setDefaults(stdmap[string, string]&) except +translate_exception
+
 cdef extern from "cantera/base/AnyMap.h" namespace "Cantera":
     cdef cppclass CxxAnyValue "Cantera::AnyValue"
 
@@ -1122,6 +1127,9 @@ cdef class Units:
     cdef CxxUnits units
     @staticmethod
     cdef copy(CxxUnits)
+
+cdef class UnitSystem:
+    cdef CxxUnitSystem unitsystem
 
 cdef class SpeciesThermo:
     cdef shared_ptr[CxxSpeciesThermo] _spthermo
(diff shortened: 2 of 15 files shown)


speth (commented):
I think making information about the units of reaction rate coefficients and phase standard concentrations available is a valuable addition to the Python interface, and a light interface to the `Units` class seems to serve this purpose, though I think there are currently a few quirks.

While I understand that the string representation of `Units` objects draws from the existing implementation (with some cleanup), I think that this output is kind of confusing when the units aren't in Cantera's base system, e.g. in the case of
```python
>>> ct.Units("mol/cm^3")
<Units(999.9999999999999 kmol / m^3) at 7f6ee85c93f0>
```
it's not really clear anymore that what this is intended to mean is that 1000 is the conversion factor between mol/cm^3 and kmol/m^3, rather than representing a particular quantity of 1000 kmol/m^3. This distinction gets even muddier with newly-allowed expressions like `Units("4 atm")`.

Assuming the goal here is just to provide unit information for these special cases where the dimensions vary in different instances, rather than presenting a full unit conversion capability (which I don't think we really want or need) I'd suggest that maybe Python `Units` objects should only be constructible in the default unit system, and that they could be represented without the leading 1.0 factor, e.g.
```python
>>> ct.Units("J/kmol")
<Units(kg * m^2 / kmol / s^2) at 7f6ef085c030>
```
and noting that practically, there's really no purpose for a user-constructed `Units` object.

I'm not clear on what the benefit of exposing the `UnitSystem` class here is. Is the purpose just to make information about the (completely fixed) default unit system available?

ischoegl:
@speth ... thanks for your comments!

> I think making information about the units of reaction rate coefficients and phase standard concentrations available is a valuable addition to the Python interface,  and a light interface to the `Units` class seems to serve this purpose, though I think there are currently a few quirks. [...] 
> 
> This distinction gets even muddier with newly-allowed expressions like `Units("4 atm")`.

Interesting observation. I definitely didn't have that in mind when I created the constructor! While I did code this up to finally wrap my head around the internal unit conversions (it *did* help), my main concern was to display what's done internally. 

> I'm not clear on what the benefit of exposing the `UnitSystem` class here is. Is the purpose just to make information about the (completely fixed) default unit system available?

Displaying the unit system was certainly one of my main concerns here (it would clear up things once and for all for any user). But at the same time, it would serve a purpose to have it as a stand-alone object that is passed to `YamlWriter` (see 5751633), where `Units`with conversion factors do make some sense.

I agree that making the Python `Units` constructor less *'useful'* makes sense. This should be a relatively easy tweak to this PR.

ischoegl:
@speth ... I reworked some of the interfaces to make things more intuitive. For unit output, the factor is suppressed:
```
In [1]: import cantera as ct
   ...:    ...: gas = ct.Solution("gri30.yaml")
   ...:    ...: gas.reaction(0)
   ...: 
Out[1]: <ThreeBodyReaction: 2 O + M <=> O2 + M>

In [2]: gas.reaction(0).rate_coeff_units
Out[2]: <Units(m^6 / kmol^2 / s) at 7fd73eb86f30>

In [3]: gas.standard_concentration_units
Out[3]: <Units(kmol / m^3) at 7fd73eb86c90>
```

The `UnitSystem` now ~displays units automatically, and~ has a more intuitive `units` property (instead of the internally used C++ `defaults`, which originates in the pre-existing `Units::setDefaults`):
```
In [4]: ct.UnitSystem().units # default unit system
Out[4]: 
{'activation-energy': 'J / kmol',
 'current': 'A',
 'energy': 'J',
 'length': 'm',
 'mass': 'kg',
 'pressure': 'Pa',
 'quantity': 'kmol',
 'temperature': 'K',
 'time': 's'}

In [5]: units = ct.UnitSystem({"quantity": "mol", "length": "cm"}) 

In [6]: units.units # alternate unit system (e.g. input to YamlWriter)
Out[6]:
{'activation-energy': 'J / mol',
 'current': 'A',
 'energy': 'J',
 'length': 'cm',
 'mass': 'kg',
 'pressure': 'Pa',
 'quantity': 'mol',
 'temperature': 'K',
 'time': 's'}
```

Finally, unit conversions are no longer enabled for unit strings with non-unity conversion factors:
```
In [6]: ct.Units('Pa')
Out[6]: <Units(kg / m / s^2) at 7f0275f42c90>

In [7]: ct.Units('3 Pa')
<ipython-input-3-f54d3ab3f8e5> in <module>
----> 1 ct.Units('3 Pa')

/src/interfaces/cython/cantera/units.pyx in cantera._cantera.Units.__cinit__()
     19             self.units = CxxUnits(stringify(name))
     20             if abs(self.units.factor() - 1.) > np.nextafter(0, 1):
---> 21                 raise ValueError(
     22                     "The creation of 'Units' objects that require a conversion "
     23                     "factor\nwith respect to Cantera's default unit system is not "

ValueError: The creation of 'Units' objects that require a conversion factor
with respect to Cantera's default unit system is not supported:
input '3 Pa' is equivalent to '3.0 kg / m / s^2' where the conversion factor is '3.0'
```

*PS: I added a direct `UnitSystem` based setter `YamlWriter::setUnitSystem` to clarify my intent.*

```
In [8]: writer = ct.YamlWriter()
   ...: writer.add_solution(ct.Solution("h2o2.yaml"))
   ...: writer.output_units = ct.UnitSystem({"quantity": "mol", "length": "cm"})
```
*(this was possible even before the latest commits, but now there's a direct path for this)*

interfaces/cython/cantera/units.pyx
```diff
@@ -0,0 +1,93 @@
...
+            if abs(self.units.factor() - 1.) > np.nextafter(0, 1):
+                raise ValueError(
+                    "The creation of 'Units' objects that require a conversion "
+                    "factor\nwith respect to Cantera's default unit system is not "
+                    f"supported:\ninput '{name}' is equivalent to "
+                    f"'{pystr(self.units.str())}' where the conversion factor is "
+                    f"'{self.units.factor()}'")
+
+    def __repr__(self):
+        if abs(self.units.factor() - 1.) < np.nextafter(0, 1):
```
speth: Is there any case where this is different from a precise equality check of `self.units.factor() == 1`? `np.nextafter(0, 1)` gives a value of 5e-324, and if the `factor` is near 1, the smallest non-zero value that `abs(self.units.factor() - 1.)` can take on is ~2e-16.

include/cantera/base/Units.h
```diff
@@ -49,6 +49,11 @@ class Units
     Units& operator*=(const Units& other);
 
     //! Provide a string representation of these Units
+    //! @param leading_one  print '1' if no units are in the numerator
+    std::string unit_str(bool leading_one=true) const;
+
+    //! Provide a string representation of these Units that includes the
+    //! conversion factor
```
speth: If we're willing to be a little flexible on the output of `Units::str`, I think it would be all-around simpler to just have the result of `str` include the factor if it's non-unity, and skip it otherwise.

interfaces/cython/cantera/_cantera.pxd
```diff
@@ -632,7 +648,7 @@ cdef extern from "cantera/base/YamlWriter.h" namespace "Cantera":
         void toYamlFile(string&) except +translate_exception
         void setPrecision(int)
         void skipUserDefined(cbool)
-        void setUnits(stdmap[string, string]&) except +translate_exception
+        void setUnitSystem(CxxUnitSystem) except +translate_exception
```
speth: I'm not sure if it actually matters to Cython, but I think this should be 
```suggestion
        void setUnitSystem(CxxUnitSystem&) except +translate_exception
```

interfaces/cython/cantera/units.pyx
```diff
@@ -0,0 +1,93 @@
...
+    Unit conversion utility
+
+    Provides functions for converting dimensional values from a given unit system.
+    The main use is for converting values specified in input files to Cantera's
+    native unit system, which is SI units except for the use of kmol as the base
+    unit of quantity, i.e. kilogram, meter, second, kelvin, ampere, and kmol.
+
+    Special functions for converting activation energies allow these values to be
+    expressed as either energy per quantity, energy (e.g. eV), or temperature by
+    applying a factor of the Avogadro number or the gas constant where needed.
```
speth: I think this docstring needs to be updated given this class won't be providing a lot of the functionality of its C++ counterpart.

src/base/Units.cpp
```diff
@@ -23,6 +24,7 @@ const std::map<std::string, Units> knownUnits{
 
     // Length [L]
     {"m", Units(1.0, 0, 1, 0)},
+    {"cm", Units(.01, 0, 1, 0)},
```
speth: What's the reason for adding this (and cal/mol) explicitly? Both are already handled by the existing conversion and prefix parsing.

test/general/test_units.cpp
```diff
@@ -4,6 +4,36 @@
...
+    EXPECT_EQ(Units("1.0 kg^0.5").unit_str(), "kg^0.5");
+    EXPECT_EQ(Units("kg / m^3").unit_str(), "kg / m^3");
+    EXPECT_EQ(Units("1 / s").unit_str(), "1 / s");
+}
+
+TEST(Units, copy_construct) {
+    EXPECT_EQ(Units(Units(1.)).str(), "1.0");
+    EXPECT_EQ(Units(Units(1., 1.)).str(), "1.0 kg");
+    EXPECT_EQ(Units(Units(1., 2.)).str(), "1.0 kg^2");
+    EXPECT_EQ(Units(Units(1., .5)).str(), "1.0 kg^0.5");
```
speth: For a class that doesn't have/require a manually-implemented copy constructor, I'm having trouble imagining how these tests could possibly fail, and I think testing the copy constructor is superfluous.

test/general/test_units.cpp
```diff
@@ -4,6 +4,36 @@
...
 using namespace Cantera;
 
+TEST(Units, from_string) {
+    EXPECT_EQ(Units("").str(), "1.0");
+    EXPECT_EQ(Units("1.").str(), "1.0");
+    EXPECT_EQ(Units("2").str(), "2.0");
+    EXPECT_EQ(Units("kg").str(), "1.0 kg");
+    EXPECT_EQ(Units("1.0 kg^0.5").str(), "1.0 kg^0.5");
+    EXPECT_EQ(Units("kg / m^3").str(), "1.0 kg / m^3");
+    EXPECT_EQ(Units("0.001 m^3").str(), "0.001 m^3");
```
speth: If we're not allowing leading numerical values as input for the `Units` class in Python, I don't think we should do it here either. I guess this comes back to the initial string representation of the `Units` class being a bit misleading and making it look more like a "quantity" that included a value and corresponding units.

src/base/Units.cpp
```diff
@@ -252,6 +324,47 @@ UnitSystem::UnitSystem(std::initializer_list<std::string> units)
...
+    std::map<std::string, std::string> out;
+    for (const auto& unit : units) {
+        out[unit.first] = unit.second.str();
+        for (const auto& known : knownUnits) {
+            if (unit.second == known.second) {
+                // Units are known: use abbreviation
+                out[unit.first] = known.first;
+            }
+        }
+    }
```
speth: The implementation of `UnitSystem::defaults()` seems more complex than necessary to me. Is there a reason not to just start with the internal unit system and update it with entries from `m_defaults`, I suppose with a special case for activation energies?

speth (changes requested):
Thanks for the updates, @ischoegl, I think this is fairly close.

interfaces/cython/cantera/units.pyx
ischoegl: Not sure - I am just quite hesitant to use an exact comparison between `double`. In any case, I moved that check to C++ by adding a flag to the `Units` constructor as:
```C++
    //! Create an object with the specified dimensions
    //! @param force_unity  ensure that conversion factor is equal to one
    explicit Units(const std::string& name, bool force_unity=false);
```
There, I am checking against `SmallNumber` ...

include/cantera/base/Units.h
ischoegl: :+1: ... we now have:
```C++
    //! Provide a string representation of these Units
    //! @param skip_unity  do not print '1' if conversion factor is equal to one
    std::string str(bool skip_unity=true) const;
```

src/base/Units.cpp
ischoegl: Removed - was left over from some intermediate testing.

test/general/test_units.cpp
ischoegl: Removed - I was indeed just following precedent but skipping unity makes sense.

src/base/Units.cpp
ischoegl: Ok. Let me walk through this as there may be some simplification I am overlooking. The internal unit system is the starting point in the `std::map<> units` variable that holds the various factors. From a generic view point, there are three scenarios here:

1. There is no conversion - ouput a regular unit string (line 345)
2. The unit can be converted to a known unit (line 349)
3. The unit has been overridden by `setDefaults`, which stores the original content in `m_defaults`.

Some thoughts:

* Scenario (3) will always take care of (2) for non-unity factors, as the current class always stores a value in `m_defaults` when a non-standard factor is assigned. 
* While (1) may be always overridden by either (2) or (3) this single line ensures correct behavior if any future (?) modification modifies a factor without changing `m_defaults`. 
* Some of the standard units used are not base units (e.g. `energy` and `pressure`) and have to be caught. 
 
Regarding the last comment, commenting out lines 346-351 would produce:
```python
{'activation-energy': 'kg * m^2 / s^2 / kmol',
 'current': 'A',
 'energy': 'kg * m^2 / s^2',
 'length': 'm',
 'mass': 'kg',
 'pressure': 'kg / m / s^2',
 'quantity': 'kmol',
 'temperature': 'K',
 'time': 's'}
```
so I believe the somewhat complex implementation is needed for `energy`, `pressure`, and - by effect - the case where activation energy is not specifically set. I added more comments in `UnitSystem::defaults` to clarify.

ischoegl:
@speth ... thank you for the additional comments, which should be taken care  of. Per suggestion, all the formatting flags are now implemented in C++ rather than just for Python. The unit tests were easily taken care of and the documentation is updated. Regarding `Unittest::default` the direction I am taking requires the steps, as I am starting from the factors and work my way from there.

ischoegl:
@speth ... let me know if there's anything else. From my side, I believe this is ready to be merged.

src/base/Units.cpp
speth: I went ahead and just implemented the alternative version of this that I had in mind in this commit: https://github.com/speth/cantera/commit/c92663253a179acc15a13341421fff6236bf49b1.

You're right that it does presume that `m_defaults` is consistent with the `m_whatever_factor` variables, but there are other features that rely on that correspondence -- namely, handling of nested `units` directives -- so this does get tested, at least implicitly.

Energy and pressure should be treated independent of the fact that they are composed of base units -- the behavior that Cantera implements is that just changing the units of mass, for example, does not automatically change the units of pressure or energy. However, the default for activation energy does depend on the energy and quantity units, and that does create a special case.

ischoegl: :+1: adopted. Thanks for claryfing!

ischoegl:
@speth ... thanks for clarifying! I simply cherry-picked your suggestion as it's indeed simpler (I was apparently too focused on the numeric factors ...).

speth (approved):
Thanks, @ischoegl, I think that wraps this all up.

PR #1077, opened 2021-07-25
Gas transport polynomial fits in Python

lavdwall:
Some modifications that allow accessing and modifying the gas transport properties polynomial fits, also from the python interface.

This is an addition to the earlier pull request #817 and more specifically, a reply to the suggestion of @ischoegl to make the access functions to the transport polynomials also available in the python interface (see https://github.com/Cantera/cantera/pull/817#issuecomment-817332657). 

Next to _access_ to the polynomial fits, I also added the possibility to _modify_ the polynomial coefficients. This in C++ as well as Python. Reason: there is literature that reports the polynomial fits to the transport properties explicitly (as function of log(T) or log(Tr)), rather than in terms of the classic L-J parameters. If this is important, the user now has the option to change the polynomial fits for some species or species pairs, after calculating them from the L-J parameters when first reading the *.cti or *.yaml input file. If the option to modify the polynomial fits is not OK for the public repo, I can remove those modification functions, and limit this pull request only to the access functions in the Python interface.

**Checklist**

### DIFF
--- a/include/cantera/cython/wrappers.h
+++ b/include/cantera/cython/wrappers.h
@@ -60,31 +60,20 @@ void CxxArray2D_set(Cantera::Array2D& array, size_t i, size_t j, double value)
     { object->FUNC_NAME(dim, data); }
 
 // Function which populates a 1D array, extra arguments
-#define ARRAY_FUNC1_1SPEC(PREFIX, CLASS_NAME, FUNC_NAME) \
+#define ARRAY_POLY(PREFIX, CLASS_NAME, FUNC_NAME) \
     void PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, size_t i, double* data) \
     { object->FUNC_NAME(i, data); }
     
-#define ARRAY_FUNC1_2SPEC(PREFIX, CLASS_NAME, FUNC_NAME) \
+#define ARRAY_POLY_BINARY(PREFIX, CLASS_NAME, FUNC_NAME) \
     void PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, size_t i, size_t j, double* data) \
     { object->FUNC_NAME(i, j, data); }
-    
-#define ARRAY_FUNC3_2SPEC(PREFIX, CLASS_NAME, FUNC_NAME) \
-    void PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, size_t i, size_t j, double* dataA, double* dataB, double* dataC) \
-    { object->FUNC_NAME(i, j, dataA, dataB, dataC); }
-    
-#define ARRAY_FUNC3_2SPEC_BOOL(PREFIX, CLASS_NAME, FUNC_NAME) \
-    void PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, size_t i, size_t j, double* dataA, double* dataB, double* dataC, bool flag) \
-    { object->FUNC_NAME(i, j, dataA, dataB, dataC, flag); }
-
 
 #define THERMO_1D(FUNC_NAME) ARRAY_FUNC(thermo, ThermoPhase, FUNC_NAME)
 #define KIN_1D(FUNC_NAME) ARRAY_FUNC(kin, Kinetics, FUNC_NAME)
 #define TRANSPORT_1D(FUNC_NAME) ARRAY_FUNC(tran, Transport, FUNC_NAME)
 #define TRANSPORT_2D(FUNC_NAME) ARRAY_FUNC2(tran, Transport, FUNC_NAME)
-#define TRANSPORT_POLY_1SPECIE(FUNC_NAME) ARRAY_FUNC1_1SPEC(tran, Transport, FUNC_NAME)
-#define TRANSPORT_POLY_2SPECIE(FUNC_NAME) ARRAY_FUNC1_2SPEC(tran, Transport, FUNC_NAME)
-#define TRANSPORT_3POLY_2SPECIE(FUNC_NAME) ARRAY_FUNC3_2SPEC(tran, Transport, FUNC_NAME)
-#define TRANSPORT_3POLY_2SPECIE_BOOL(FUNC_NAME) ARRAY_FUNC3_2SPEC_BOOL(tran, Transport, FUNC_NAME)
+#define TRANSPORT_POLY(FUNC_NAME) ARRAY_POLY(tran, Transport, FUNC_NAME)
+#define TRANSPORT_POLY_BINARY(FUNC_NAME) ARRAY_POLY_BINARY(tran, Transport, FUNC_NAME)
 
 THERMO_1D(getMassFractions)
 THERMO_1D(setMassFractions)
@@ -139,11 +128,9 @@ TRANSPORT_1D(getMobilities)
 TRANSPORT_2D(getMultiDiffCoeffs)
 TRANSPORT_2D(getBinaryDiffCoeffs)
 
-TRANSPORT_POLY_1SPECIE(getViscosityPolynomials)
-TRANSPORT_POLY_1SPECIE(setViscosityPolynomials)
-TRANSPORT_POLY_1SPECIE(getConductivityPolynomials)
-TRANSPORT_POLY_1SPECIE(setConductivityPolynomials)
-TRANSPORT_POLY_2SPECIE(getBinDiffusivityPolynomials)
-TRANSPORT_POLY_2SPECIE(setBinDiffusivityPolynomials)
-TRANSPORT_3POLY_2SPECIE(getCollisionIntegralPolynomials)
-TRANSPORT_3POLY_2SPECIE_BOOL(setCollisionIntegralPolynomials)
+TRANSPORT_POLY(getViscosityPolynomial)
+TRANSPORT_POLY(setViscosityPolynomial)
+TRANSPORT_POLY(getConductivityPolynomial)
+TRANSPORT_POLY(setConductivityPolynomial)
+TRANSPORT_POLY_BINARY(getBinDiffusivityPolynomial)
+TRANSPORT_POLY_BINARY(setBinDiffusivityPolynomial)
--- a/include/cantera/cython/wrappers.h
+++ b/include/cantera/cython/wrappers.h
@@ -59,11 +59,32 @@ void CxxArray2D_set(Cantera::Array2D& array, size_t i, size_t j, double value)
     void PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, size_t dim, double* data) \
     { object->FUNC_NAME(dim, data); }
 
+// Function which populates a 1D array, extra arguments
+#define ARRAY_FUNC1_1SPEC(PREFIX, CLASS_NAME, FUNC_NAME) \
+    void PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, size_t i, double* data) \
+    { object->FUNC_NAME(i, data); }
+    
+#define ARRAY_FUNC1_2SPEC(PREFIX, CLASS_NAME, FUNC_NAME) \
+    void PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, size_t i, size_t j, double* data) \
+    { object->FUNC_NAME(i, j, data); }
+    
+#define ARRAY_FUNC3_2SPEC(PREFIX, CLASS_NAME, FUNC_NAME) \
+    void PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, size_t i, size_t j, double* dataA, double* dataB, double* dataC) \
+    { object->FUNC_NAME(i, j, dataA, dataB, dataC); }
+    
+#define ARRAY_FUNC3_2SPEC_BOOL(PREFIX, CLASS_NAME, FUNC_NAME) \
+    void PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, size_t i, size_t j, double* dataA, double* dataB, double* dataC, bool flag) \
+    { object->FUNC_NAME(i, j, dataA, dataB, dataC, flag); }
+
 
 #define THERMO_1D(FUNC_NAME) ARRAY_FUNC(thermo, ThermoPhase, FUNC_NAME)
 #define KIN_1D(FUNC_NAME) ARRAY_FUNC(kin, Kinetics, FUNC_NAME)
 #define TRANSPORT_1D(FUNC_NAME) ARRAY_FUNC(tran, Transport, FUNC_NAME)
 #define TRANSPORT_2D(FUNC_NAME) ARRAY_FUNC2(tran, Transport, FUNC_NAME)
+#define TRANSPORT_POLY_1SPECIE(FUNC_NAME) ARRAY_FUNC1_1SPEC(tran, Transport, FUNC_NAME)
+#define TRANSPORT_POLY_2SPECIE(FUNC_NAME) ARRAY_FUNC1_2SPEC(tran, Transport, FUNC_NAME)
+#define TRANSPORT_3POLY_2SPECIE(FUNC_NAME) ARRAY_FUNC3_2SPEC(tran, Transport, FUNC_NAME)
+#define TRANSPORT_3POLY_2SPECIE_BOOL(FUNC_NAME) ARRAY_FUNC3_2SPEC_BOOL(tran, Transport, FUNC_NAME)
 
 THERMO_1D(getMassFractions)
 THERMO_1D(setMassFractions)
@@ -117,3 +138,12 @@ TRANSPORT_1D(getMobilities)
 
 TRANSPORT_2D(getMultiDiffCoeffs)
 TRANSPORT_2D(getBinaryDiffCoeffs)
+
+TRANSPORT_POLY_1SPECIE(getViscosityPolynomials)
+TRANSPORT_POLY_1SPECIE(setViscosityPolynomials)
+TRANSPORT_POLY_1SPECIE(getConductivityPolynomials)
+TRANSPORT_POLY_1SPECIE(setConductivityPolynomials)
+TRANSPORT_POLY_2SPECIE(getBinDiffusivityPolynomials)
+TRANSPORT_POLY_2SPECIE(setBinDiffusivityPolynomials)
+TRANSPORT_3POLY_2SPECIE(getCollisionIntegralPolynomials)
+TRANSPORT_3POLY_2SPECIE_BOOL(setCollisionIntegralPolynomials)
--- a/include/cantera/transport/GasTransport.h
+++ b/include/cantera/transport/GasTransport.h
@@ -117,36 +117,36 @@ public:
 
     //! Return the polynomial fits to the viscosity of species i
     //! @see fitProperties()
-    virtual void getViscosityPolynomials(size_t i, double* coeffs) const;
+    virtual void getViscosityPolynomial(size_t i, double* coeffs) const;
 
     //! Return the temperature fits of the heat conductivity of species i
     //! @see fitProperties()
-    virtual void getConductivityPolynomials(size_t i, double* coeffs) const;
+    virtual void getConductivityPolynomial(size_t i, double* coeffs) const;
 
     //! Return the polynomial fits to the binary diffusivity of species pair (i, j)
     //! @see fitDiffCoeffs()
-    virtual void getBinDiffusivityPolynomials(size_t i, size_t j, double* coeffs) const;
+    virtual void getBinDiffusivityPolynomial(size_t i, size_t j, double* coeffs) const;
 
     //! Return the polynomial fits to the collision integral of species pair (i, j)
     //! @see fitCollisionIntegrals()
-    virtual void getCollisionIntegralPolynomials(size_t i, size_t j, double* astar_coeffs, 
+    virtual void getCollisionIntegralPolynomial(size_t i, size_t j, double* astar_coeffs, 
                                                 double* bstar_coeffs, double* cstar_coeffs) const;
 
     //! Modify the polynomial fits to the viscosity of species i
     //! @see fitProperties()
-    virtual void setViscosityPolynomials(size_t i, double* coeffs);
+    virtual void setViscosityPolynomial(size_t i, double* coeffs);
 
     //! Modify the temperature fits of the heat conductivity of species i
     //! @see fitProperties()
-    virtual void setConductivityPolynomials(size_t i, double* coeffs);
+    virtual void setConductivityPolynomial(size_t i, double* coeffs);
 
     //! Modify the polynomial fits to the binary diffusivity of species pair (i, j)
     //! @see fitDiffCoeffs()
-    virtual void setBinDiffusivityPolynomials(size_t i, size_t j, double* coeffs);
+    virtual void setBinDiffusivityPolynomial(size_t i, size_t j, double* coeffs);
 
     //! Modify the polynomial fits to the collision integral of species pair (i, j)
     //! @see fitCollisionIntegrals()
-    virtual void setCollisionIntegralPolynomials(size_t i, size_t j, double* astar_coeffs, 
+    virtual void setCollisionIntegralPolynomial(size_t i, size_t j, double* astar_coeffs, 
                                                 double* bstar_coeffs, double* cstar_coeffs, bool actualT);
 
     virtual void init(ThermoPhase* thermo, int mode=0, int log_level=0);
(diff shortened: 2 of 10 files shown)


ischoegl:
@lavdwall ... thanks for posting! Personally, I believe that the ability to access & modify transport polynomials from Python is very useful.

Before I get into the details, would you mind adding some unit tests?

lavdwall:
Hi @ischoegl ,

Sorry for the late reply (holidays...).

Thanks for your positive response. I will have a look at adding some unit tests in the coming week.
Until now, I just checked the implementation by a series of 'get' and 'set' commands, making sure the calculated transport properties remain the same after 'setting' the coefficients acquired from a previous 'get' command. Is that OK for a testing approach? If so, I will try to implement it like that in the cython test folder.

What I implemented in the current pull request is really just a get and set of the polynomial coefficients of viscosity, thermal conductivity, binary diffusion coefficients and collision integrals, just as they are available in the C++ core of the code. This is regardless of the transport model implementation that is used, so the user has to be aware of possible differences between CK_Mode = True or  CK_Mode = False.  And the user needs to be aware of possible discrepancies between the coefficients stored in memory and how those are documented in the doxygen documentation. The documentation is in agreement with textbooks (Kee, and others) but internally in the code the definitions are a bit different it seems.

Anyway, I will add test functions to show that set/get works. But maybe on the longer term, the code documentation should be adjusted a bit as well so that everyone can correctly use the access/modify functions without knowledge of the underlying code.

ischoegl:
@lavdwall … yes, just checking whether coefficients are written correctly would be adequate. I’m not sure about the documentation; it may be worthwhile to raise an issue first so people with more intricate knowledge can chime in.

include/cantera/cython/wrappers.h
```diff
@@ -117,3 +138,12 @@ TRANSPORT_1D(getMobilities)
...
 TRANSPORT_2D(getBinaryDiffCoeffs)
+
+TRANSPORT_POLY_1SPECIE(getViscosityPolynomials)
+TRANSPORT_POLY_1SPECIE(setViscosityPolynomials)
+TRANSPORT_POLY_1SPECIE(getConductivityPolynomials)
+TRANSPORT_POLY_1SPECIE(setConductivityPolynomials)
+TRANSPORT_POLY_2SPECIE(getBinDiffusivityPolynomials)
+TRANSPORT_POLY_2SPECIE(setBinDiffusivityPolynomials)
+TRANSPORT_3POLY_2SPECIE(getCollisionIntegralPolynomials)
+TRANSPORT_3POLY_2SPECIE_BOOL(setCollisionIntegralPolynomials)
```
ischoegl: The `#define`'s for these functions are only used once each, so it may be easier to avoid inclusion in `wrappers.h`.

include/cantera/cython/wrappers.h
```diff
@@ -59,11 +59,32 @@ void CxxArray2D_set(Cantera::Array2D& array, size_t i, size_t j, double value)
...
+    void PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, size_t i, size_t j, double* data) \
+    { object->FUNC_NAME(i, j, data); }
+    
+#define ARRAY_FUNC3_2SPEC(PREFIX, CLASS_NAME, FUNC_NAME) \
+    void PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, size_t i, size_t j, double* dataA, double* dataB, double* dataC) \
+    { object->FUNC_NAME(i, j, dataA, dataB, dataC); }
+    
+#define ARRAY_FUNC3_2SPEC_BOOL(PREFIX, CLASS_NAME, FUNC_NAME) \
+    void PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, size_t i, size_t j, double* dataA, double* dataB, double* dataC, bool flag) \
+    { object->FUNC_NAME(i, j, dataA, dataB, dataC, flag); }
```
ischoegl: While I believe that using the C preprocessor here is generally a good idea, the versions taking three arrays / three arrays with flags may be avoidable. Generally, `#define`s are used for repetitive tasks to save code.

include/cantera/cython/wrappers.h
```diff
@@ -117,3 +138,12 @@ TRANSPORT_1D(getMobilities)
 
 TRANSPORT_2D(getMultiDiffCoeffs)
 TRANSPORT_2D(getBinaryDiffCoeffs)
+
+TRANSPORT_POLY_1SPECIE(getViscosityPolynomials)
+TRANSPORT_POLY_1SPECIE(setViscosityPolynomials)
+TRANSPORT_POLY_1SPECIE(getConductivityPolynomials)
+TRANSPORT_POLY_1SPECIE(setConductivityPolynomials)
+TRANSPORT_POLY_2SPECIE(getBinDiffusivityPolynomials)
+TRANSPORT_POLY_2SPECIE(setBinDiffusivityPolynomials)
```
ischoegl: One thing I overlooked previously is that each of those getters/setters are for a single polynomial, correct? Accordingly, it may be good to 'fix' the nomenclature to `getViscosityPolynomial` etc. (which would then also match the Python nomenclature that doesn't have the 's')

interfaces/cython/cantera/_cantera.pxd
```diff
@@ -1096,6 +1097,16 @@ cdef extern from "cantera/cython/wrappers.h":
     cdef void tran_getMultiDiffCoeffs(CxxTransport*, size_t, double*) except +translate_exception
     cdef void tran_getBinaryDiffCoeffs(CxxTransport*, size_t, double*) except +translate_exception
 
+    cdef void tran_getViscosityPolynomials(CxxTransport*, size_t, double*) except +translate_exception
+    cdef void tran_getConductivityPolynomials(CxxTransport*, size_t, double*) except +translate_exception
+    cdef void tran_getBinDiffusivityPolynomials(CxxTransport*, size_t, size_t, double*) except +translate_exception
+    cdef void tran_getCollisionIntegralPolynomials(CxxTransport*, size_t, size_t, double*, double*, double*) except +translate_exception
```
ischoegl: As mentioned, above, I believe this should be handled outside of `wrappesr.h`, so you would move `tran_getCollisionIntegralPolynomials` to where you defined `cbool CKMode()` around line 584. In this case, the `tran_` prefix is usually omitted.

interfaces/cython/cantera/_cantera.pxd
```diff
@@ -1096,6 +1097,16 @@ cdef extern from "cantera/cython/wrappers.h":
...
+
+    cdef void tran_setViscosityPolynomials(CxxTransport*, size_t, double*) except +translate_exception
+    cdef void tran_setConductivityPolynomials(CxxTransport*, size_t, double*) except +translate_exception
+    cdef void tran_setBinDiffusivityPolynomials(CxxTransport*, size_t, size_t, double*) except +translate_exception
+    cdef void tran_setCollisionIntegralPolynomials(CxxTransport*, size_t, size_t, double*, double*, double*, bool) except +translate_exception
```
ischoegl: Same as for `tran_getCollisionIntegralPolynomials`. Also, I believe `bool` should be `cbool` here.

interfaces/cython/cantera/transport.pyx
```diff
@@ -256,6 +260,79 @@ cdef class Transport(_SolutionBase):
         def __get__(self):
             return get_transport_1d(self, tran_getMobilities)
 
+    def viscosity_polynomial(self, i):
+        """Get the polynomial fit to the logarithm of temperature for 
+        the viscosity of species i."""
+        cdef np.ndarray[np.double_t, ndim=1] data = np.empty(4 if self.transport.CKMode() else 5)
+        tran_getViscosityPolynomials(self.transport, i, &data[0])
+        return data
```
ischoegl: You could simplify this procedure by introducing a new `cdef` function similar to `get_transport_1d` at the top, e.g.
```Python
cdef np.ndarray get_transport_polynomial(
        Transport tran, transportMethod1d method, int index, int n_coeffs):
    cdef np.ndarray[np.double_t, ndim=1] data = np.empty(n_coeffs)
    method(tran.transport, index, &data[0])
    return data
```
and declare it in `_cantera.pxd`.

This would allow you to simplify each of the  calls here and below to
```Python
    n_coeffs = 4 if self.transport.CKMode() else 5
    return get_transport_polynomial(self.transport, tran_getViscosityPolynomials, i, n_coeffs)
```
(or similar), making the implementation much easier to follow. Same goes for the two-index version.

include/cantera/cython/wrappers.h
```diff
@@ -59,11 +59,32 @@ void CxxArray2D_set(Cantera::Array2D& array, size_t i, size_t j, double value)
...
+
 
 #define THERMO_1D(FUNC_NAME) ARRAY_FUNC(thermo, ThermoPhase, FUNC_NAME)
 #define KIN_1D(FUNC_NAME) ARRAY_FUNC(kin, Kinetics, FUNC_NAME)
 #define TRANSPORT_1D(FUNC_NAME) ARRAY_FUNC(tran, Transport, FUNC_NAME)
 #define TRANSPORT_2D(FUNC_NAME) ARRAY_FUNC2(tran, Transport, FUNC_NAME)
+#define TRANSPORT_POLY_1SPECIE(FUNC_NAME) ARRAY_FUNC1_1SPEC(tran, Transport, FUNC_NAME)
+#define TRANSPORT_POLY_2SPECIE(FUNC_NAME) ARRAY_FUNC1_2SPEC(tran, Transport, FUNC_NAME)
+#define TRANSPORT_3POLY_2SPECIE(FUNC_NAME) ARRAY_FUNC3_2SPEC(tran, Transport, FUNC_NAME)
+#define TRANSPORT_3POLY_2SPECIE_BOOL(FUNC_NAME) ARRAY_FUNC3_2SPEC_BOOL(tran, Transport, FUNC_NAME)
```
ischoegl: A very minor point here is that you use `1SPEC` / `2SPEC` above and `1SPECIE`/`2SPECIE` here. You may be able to simplify/clarify this using names with no suffix (one index version), and having suffix `_BINARY` (2 indices)? Same goes for the proposed `cdef` access function in the Python interface.

src/transport/GasTransport.cpp
```diff
@@ -839,4 +844,102 @@ void GasTransport::getBinDiffusivityPolynomials(size_t i, size_t j, double* coef
...
+        m_visccoeffs[i][k] = coeffs[k];
+    }
+    
+    m_visc_ok = false;
+    m_spvisc_ok = false;
+    m_viscwt_ok = false;
+    m_bindiff_ok = false;
+    m_temp = -1;
+    
+    std::cout << "Modified viscosity polynomial for species "
```
ischoegl: None of those methods should generate output.

src/transport/MultiTransport.cpp
```diff
@@ -433,20 +434,20 @@ void MultiTransport::updateThermal_T()
     // evaluate polynomial fits for A*, B*, C*
     for (size_t i = 0; i < m_nsp; i++) {
         for (size_t j = i; j < m_nsp; j++) {
-            double z = m_logt - m_log_eps_k(i,j);
+            double z = m_star_poly_actualT[i][j] == 1? m_logt : m_logt - m_log_eps_k(i,j);
```
ischoegl: Could you comment on what this modification does?. (Also, there is a missing whitespace before the `?`)

Further, commenting out the `m_om22` lines here without removing it from the header files leaves the possibility that it is used elsewhere. Specifically, the missing `resize` call could result in a segfault.

interfaces/cython/cantera/transport.pyx
```diff
@@ -256,6 +260,79 @@ cdef class Transport(_SolutionBase):
...
+    def collision_integral_polynomials(self, i, j):
+        """Get the polynomial fit to the logarithm of temperature for 
+        the collision integral of species i and j."""
+        cdef np.ndarray[np.double_t, ndim=1] adata = np.empty(7 if self.transport.CKMode() else 9)
+        cdef np.ndarray[np.double_t, ndim=1] bdata = np.empty(7 if self.transport.CKMode() else 9)
+        cdef np.ndarray[np.double_t, ndim=1] cdata = np.empty(7 if self.transport.CKMode() else 9)
+        tran_getCollisionIntegralPolynomials(self.transport, i, j, &adata[0], &bdata[0], &cdata[0])
+        return adata, bdata, cdata
+
+    def modify_viscosity_polynomial(self, i, values):
```
ischoegl: For getters/setters, I would recommend the more commonly used prefixes `get_` (and potentially `set_`). ~More often than not, Cantera omits prefixes only when accessing `properties` (which don't use function arguments).~ Besides, nomenclature will also match underlying C++ methods.

include/cantera/transport/GasTransport.h
```diff
@@ -386,6 +412,10 @@ class GasTransport : public Transport
      * (i,j).
      */
     std::vector<vector_fp> m_omega22_poly;
+    
+    //! Flag to indicate for which (i,j) interaction pairs the 
+    //! actual temperature is used instead of the reduced temperature
+    std::vector<vector_int> m_star_poly_actualT;
```
ischoegl: This is only `true` or `false`, correct? While it may not be possible to use `std::vector<bool>` (which sometimes behaves funny), clarifying this as `m_star_poly_uses_actualT` may improve code readability.

ischoegl (changes requested):
Hi @lavdwall  ... thank you for the PR! While I have quite a few suggestions, this looks mostly good to me. Most of my comments affect details of the implementation.

One caveat here is that I'm not as familiar with some of the changes that affect theory (it's somewhat out of my wheelhouse), but I'm sure that others can fill in those gaps.

ischoegl:
Linking a recent request for this feature from the [User Group](https://groups.google.com/g/cantera-users/c/ozJebOvDokY/m/qB18byyUBAAJ).

ischoegl:
Hi @lavdwall ... just wanted to briefly check whether you had a chance to look over the review comments? Apart from the change to `Transport`, everything appears to be fairly close.

include/cantera/cython/wrappers.h
lavdwall: OK

include/cantera/cython/wrappers.h
lavdwall: OK

include/cantera/cython/wrappers.h
lavdwall: Fair point, I changed it

interfaces/cython/cantera/_cantera.pxd
lavdwall: Done

interfaces/cython/cantera/_cantera.pxd
lavdwall: Indeed

interfaces/cython/cantera/transport.pyx
lavdwall: Done

include/cantera/cython/wrappers.h
lavdwall: Done

src/transport/GasTransport.cpp
lavdwall: Those std::cout were for testing and I forgot to remove them. They are gone now.

interfaces/cython/cantera/transport.pyx
lavdwall: Done

src/transport/MultiTransport.cpp
lavdwall: If m_star_poly_actualT[i][j] is equal to 1 (or as you suggest below, if it is True), the polynomials are defined as a function of the logarithm of the actual temperature instead of the reduced temperature (T_ij = k_B*T/eps_ij).

lavdwall: It seems like m_om22 is not used anywhere, so I removed it from the header files as well.

include/cantera/transport/GasTransport.h
lavdwall: I renamed the variable as suggested.

lavdwall:
Hi @ischoegl 
Apologies for the delay. I believe most of your comments are addressed in the two commits I just pushed.

include/cantera/transport/GasTransport.h
```diff
@@ -117,17 +117,43 @@ class GasTransport : public Transport
...
+    virtual void setBinDiffusivityPolynomial(size_t i, size_t j, double* coeffs);
+
+    //! Modify the polynomial fits to the collision integral of species pair (i, j)
+    //! @see fitCollisionIntegrals()
+    virtual void setCollisionIntegralPolynomial(size_t i, size_t j, double* astar_coeffs, 
+                                                double* bstar_coeffs, double* cstar_coeffs, bool actualT);
 
     virtual void init(ThermoPhase* thermo, int mode=0, int log_level=0);
+    
+    bool CKMode() const {
```
ischoegl, line 160: needs a docstring.

include/cantera/transport/GasTransport.h
```diff
@@ -117,17 +117,43 @@ class GasTransport : public Transport
...
+    virtual void getConductivityPolynomial(size_t i, double* coeffs) const;
 
     //! Return the polynomial fits to the binary diffusivity of species pair (i, j)
     //! @see fitDiffCoeffs()
-    virtual void getBinDiffusivityPolynomials(size_t i, size_t j, double* coeffs) const;
+    virtual void getBinDiffusivityPolynomial(size_t i, size_t j, double* coeffs) const;
+
+    //! Return the polynomial fits to the collision integral of species pair (i, j)
+    //! @see fitCollisionIntegrals()
+    virtual void getCollisionIntegralPolynomial(size_t i, size_t j, double* astar_coeffs,
```
ischoegl: Now that all functions are first defined in `TransportBase.h` (with a specialization defined here), the docstrings should be moved to the base and can be omitted here. 

Same for all other newly introduced methods below.

include/cantera/transport/TransportBase.h
```diff
@@ -585,6 +585,41 @@ class Transport
         throw NotImplementedError("Transport::getMixDiffCoeffsMass");
     }
 
+    //! Get/set the polynomial fits to the transport properties
```
ischoegl: Replace with docstrings currently added to specializations in `GasTransport.h` (see other comment). Same for other functions below.

interfaces/cython/cantera/transport.pyx
```diff
@@ -256,6 +272,76 @@ cdef class Transport(_SolutionBase):
...
+        cdef np.ndarray[np.double_t, ndim=1] cdata = np.empty(7 if self.transport.CKMode() else 9)
+        self.transport.getCollisionIntegralPolynomial(i, j, &adata[0], &bdata[0], &cdata[0])
+        return adata, bdata, cdata
+
+    def set_viscosity_polynomial(self, i, values):
+        """Set the polynomial fit to the logarithm of temperature for 
+        the viscosity of species i."""
+        if len(values) != (4 if self.transport.CKMode() else 5):
+            msg = "Got {}. Expected {}".format(len(values), (4 if self.transport.CKMode() else 5))
+            raise ValueError('Array has incorrect length. ' + msg + '.')
```
ischoegl: it may be easier to read if you assigned the expected length of the list to a new variable, e.g.
```Python
n_values = 4 if self.transport.CKMode() else 5
if len(values) != n_values:
    raise ValueError(
        f"Array has incorrect length: expected {n_values} but received {len(values)}.")
```
There are also some additional minor changes above: f-strings make things more readable, and Cantera is now preferring double quotes (`"`) rather than single quotes (`'`) for the definition of strings.

The same comments apply for the remaining new functions below.

src/transport/GasTransport.cpp
```diff
@@ -839,4 +844,90 @@ void GasTransport::getBinDiffusivityPolynomials(size_t i, size_t j, double* coef
     }
 }
 
+void GasTransport::getCollisionIntegralPolynomial(size_t i, size_t j, double* astar_coeffs, 
+                                                double* bstar_coeffs, double* cstar_coeffs) const
+{   
+    for (size_t k = 0; k < (m_mode == CK_Mode ? 6 : COLL_INT_POLY_DEGREE)+1; k++) {
```
ischoegl: Missing white spaces ... should be `(m_mode == CK_Mode ? 6 : COLL_INT_POLY_DEGREE) + 1;`

ischoegl (changes requested):
Thanks for updating the PR, @lavdwall - things remaining from my side are mostly about formatting and should be easy to take care of.

include/cantera/transport/GasTransport.h
lavdwall: OK

include/cantera/transport/GasTransport.h
lavdwall: I added the docstrings in `TransportBase.h` but I will keep them in `GasTransport.h` as well, to have a link (@ see) to `fitProperties()`, `fitDiffCoeffs()` and `fitCollisionIntegrals()`, which are only defined in `GasTransport.h`

include/cantera/transport/TransportBase.h
lavdwall: OK

interfaces/cython/cantera/transport.pyx
lavdwall: OK

src/transport/GasTransport.cpp
lavdwall: OK

include/cantera/transport/TransportBase.h
```diff
@@ -667,6 +709,10 @@ class Transport
 
     //! Set root Solution holding all phase information
     virtual void setRoot(std::shared_ptr<Solution> root);
+    
+    virtual bool CKMode() const {
```
ischoegl, line 719: Missing docstring for the initial definition.

include/cantera/transport/GasTransport.h
```diff
@@ -117,17 +117,45 @@ class GasTransport : public Transport
...
+    virtual void setConductivityPolynomial(size_t i, double* coeffs);
+
+    //! Modify the polynomial fits to the binary diffusivity of species pair (i, j)
+    //! @see fitDiffCoeffs()
+    virtual void setBinDiffusivityPolynomial(size_t i, size_t j, double* coeffs);
+
+    //! Modify the polynomial fits to the collision integral of species pair (i, j)
+    //! @see fitCollisionIntegrals()
+    virtual void setCollisionIntegralPolynomial(size_t i, size_t j, double* astar_coeffs, 
+                                                double* bstar_coeffs, double* cstar_coeffs, bool actualT);
```
ischoegl: Please limit line length to 88 characters. (Same for `getCollisionIntegralPolynomial`, equivalent definitions in `TransportBase.h`, and lines in the corresponding `cpp` files).

interfaces/cython/cantera/transport.pyx
```diff
@@ -17,7 +17,18 @@ cdef np.ndarray get_transport_2d(Transport tran, transportMethod2d method):
...
-
+    
+cdef np.ndarray get_transport_polynomial(
+        Transport tran, transportPolyMethod1i method, int index, int n_coeffs):
+    cdef np.ndarray[np.double_t, ndim=1] data = np.empty(n_coeffs)
+    method(tran.transport, index, &data[0])
+    return data
+    
+cdef np.ndarray get_binary_transport_polynomial(
+        Transport tran, transportPolyMethod2i method, int indexi, int indexj, int n_coeffs):
```
ischoegl: Similar to other comment: line length exceeds 88 characters. Same is true for some of the setters/getters defined further down in this file.

ischoegl (changes requested):
@lavdwall ... thanks for taking care of the requests. Sorry for asking for one final edit: I hadn't noticed earlier that some of the lines exceed the 88 character limit we try to enforce (see `CONTRIBUTING.md` style guide). Beyond what I point out below, there's nothing else from my side.

ischoegl (approved):
@lavdwall ... thank you again for this PR and for taking care of the formatting right away. From my side, this looks ready to be merged. I will leave it open until later today, in case anyone else should have any requests.

Naikless:
Just a quick follow up on this: is there any documentation on the theory behind the current polynomial implementation of the viscosity?

ischoegl:
This PR (as well as the associated PR #817) are mostly just dealing with providing a better user interface. The code itself is documented in the C++ include files, which can be accessed by Cantera's doxygen documentation. Here is a link for [GasTransport](https://cantera.org/documentation/docs-2.6/doxygen/html/d8/d58/classCantera_1_1GasTransport.html#details) ... while it doesn't provide every detail in terms of theory, things should correspond to standard practice, see [reference](https://cantera.org/documentation/docs-2.6/doxygen/html/d8/d58/classCantera_1_1GasTransport.html#details).

PR #1079, opened 2021-07-29
Non-ideal capabilities for 1-D flame model

gkogekar:

Add non-ideal capabilities to the existing Flame models. Currently, the flame models utilize only ideal gas phase.  There has been some discussion in the user group and the feature request (https://github.com/Cantera/enhancements/issues/109) to include 1-D supercritical features in the Flame code. This pull request implements the real-gas flame models (R-K EoS, P-R EoS)  along with the modified transport. 

- Modify `StFlow` class to include real-gas thermophases (Redlich-Kwong, Peng-Robinson EoS)
- Modifications to the Transport / HighP transport models (if required)
- Modifications to Python interface
- Cantera tests to check real-gas flame capabilities

**References**
https://github.com/Cantera/enhancements/issues/109

**Theory Document**
[Non_ideal_Flame_equations.pdf](https://github.com/Cantera/cantera/files/6901849/Non_ideal_Flame_equations.pdf)


speth:
@gkogekar - If you can, please undo the merge commit and rebase this onto the current `main` branch instead.

bryanwweber:
Thanks @gkogekar, I just saw this was ready for review. It looks like there aren't any cases tested where you might expect to see some effect of the non-ideal EOS, can you please add something?

Similarly, the difference with the existing results is somewhat concerning to me. Things are changing in the third and fourth decimal place, (although I don't have a good sense for how much change would be expected). I guess I'm surprised there's so much difference by swapping the enthalpy for the temperature. Can you comment? Can you do a comparison of a simple calculation in the zero-D case to get a sense for how much difference to expect?

gkogekar:
Thanks, @bryanwweber. I had derived the energy equation for non-ideal EoS (the document is attached above). But we decided to move forward with the finite difference approach (I don't exactly remember why. ) In this implementation, the enthalpy flux is calculated based on a first order finite differencing, as opposed to exact calculation of Cp value in case of ideal gas EoS. So the difference in the results is bound to happen depending on the grid size. 

I am creating this PR as many people on the user group have asked for this functionality. But I haven't found any particular test case that will show the non-ideal effects. Since we are modeling flames (i.e. high temperatures), the gas-phase is always near ideal condition. Nonetheless, I will create a test case using a non-ideal EoS.

bryanwweber:
Thanks @gkogekar! I think the finite difference makes sense when we don't know anything about the underlying phase, but I wonder if there's room to allow the evaluation to specialize, analogous to `Reactor` vs. `IdealGasReactor`? One way I can think of is to add a flag to the constructor like `non_ideal_phase` (I don't love that name, but I hope it conveys the idea) to allow the user to specify they want to use the finite difference of the enthalpy, which could default to `false` so that the ideal gas solution is the default. Another way is to auto-detect the phase type; I'm not sure if it's possible to auto-detect the class type since it's a pointer, but that'd be the best case if possible.

As far as examples, I agree that under high-temperature (combustion) conditions, it's really unlikely to see a difference here. That'd be a reasonable test case anyways, if you compare to the ideal gas result. You can also consider something like high-pressure catalysis against a surface, a liquid flow, or something like that.

src/oneD/StFlow.cpp
```diff
@@ -451,22 +456,39 @@ void StFlow::evalResidual(double* x, double* rsd, int* diag,
...
+                    m_thermo->getCp_R(&cp_R[0]);
+                    m_thermo->getEnthalpy_RT_ref(&h_RT[0]);
+                    for (size_t k = 0; k < m_nsp; k++) {
+                        double flxk = 0.5*(m_flux(k,j-1) + m_flux(k,j));
+                        sum += wdot(k,j)*h_RT[k];
+                        sum2 += flxk*cp_R[k]/m_wt[k];
+                   }
+                   sum *= GasConstant * T(x,j);
+                   sum2 *= GasConstant * dtdzj;
+                } else if((m_thermo->type() == "RedlichKwong") || (m_thermo->type() == "PengRobinson"))
```
bryanwweber: I think this could just be `else`, you've already handled unsupported phase types earlier 😊

gkogekar:
@bryanwweber Thanks for your insights. Retrieving the `thermo` type in `StFlow `class is possible. So I have just modified the energy equation code, where it will use finite difference only for non-ideal phases. It will use the existing energy residual equation in the case of ideal gas EoS. This is the simplest workaround based on your suggestions. But I can always add an external flag if necessary. 

I will add the test for non-ideal EoS soon.

bryanwweber:
> But I can always add an external flag if necessary.

Thanks @gkogekar this is a really nice solution, I didn't think it was possible but I'm glad it is! Definitely not something to do here, but I wonder if it makes sense to eliminate some of the class hierarchy in reactors...

src/oneD/StFlow.cpp
```diff
@@ -35,9 +35,13 @@ StFlow::StFlow(ThermoPhase* ph, size_t nsp, size_t points) :
...
     if (ph->type() == "IdealGas") {
         m_thermo = static_cast<IdealGasPhase*>(ph);
+    } else if(ph->type() == "Redlich-Kwong" && ph->type() == "RedlichKwongMFTP") {
+        m_thermo = static_cast<RedlichKwongMFTP*>(ph);
+    } else if(ph->type() == "Peng-Robinson") {
+        m_thermo = static_cast<PengRobinson*>(ph);
     } else {
         throw CanteraError("StFlow::StFlow",
-                           "Unsupported phase type: need 'IdealGasPhase'");
+                           "Unsupported phase type");
```
speth: Changing `m_thermo` to be a `ThermoPhase*` rather than an `IdealGasPhase*` makes all of this `static_cast`ing and checking unnecessary.

include/cantera/oneD/StFlow.h
```diff
@@ -9,6 +9,8 @@
 #include "Domain1D.h"
 #include "cantera/base/Array.h"
 #include "cantera/thermo/IdealGasPhase.h"
+#include "cantera/thermo/RedlichKwongMFTP.h"
+#include "cantera/thermo/PengRobinson.h"
```
speth: It shouldn't be necessary to include these headers here. And actually, with the change of `m_thermo` to a simple `ThermoPhase` pointer, I think you could replace the include of `IdealGasPhase.h` with `ThermoPhase.h`.

include/cantera/oneD/StFlow.h
```diff
@@ -71,7 +73,7 @@ class StFlow : public Domain1D
      * Set the thermo manager. Note that the flow equations assume
      * the ideal gas equation.
      */
-    void setThermo(IdealGasPhase& th) {
+    void setThermo(ThermoPhase& th) {
```
speth: The comment above this method is no longer true.

src/oneD/StFlow.cpp
```diff
@@ -898,4 +920,41 @@ void StFlow::evalContinuity(size_t j, double* x, double* rsd, int* diag, double
...
+    m_thermo->getPartialMolarEnthalpies(&m_hk_left[0]);
+    
+    // Current Node j
+    setGas(x, j);
+    m_thermo->getPartialMolarEnthalpies(&m_hk_current[0]);
+
+    // Node (j+1)
+    setGas(x, j + 1);
+    m_thermo->getPartialMolarEnthalpies(&m_hk_right[0]);
+}
```
speth: To avoid calling `getPartialMolarEnthalpies` four times for each grid point, I think you could just create a single `Array2D` for `m_hk` and update it as part of the `updateThermo` method (and take care of resizing it in the same way that `m_wdot` is handled).

src/oneD/StFlow.cpp
```diff
@@ -898,4 +920,41 @@ void StFlow::evalContinuity(size_t j, double* x, double* rsd, int* diag, double
     }
 }
 
+vector_fp StFlow::grad_hk(const doublereal* x, size_t j) 
+{
+    vector_fp dhk_dz(m_nsp, 0.0);
```
speth: To avoid having to create/delete a lot of temporary arrays, I would suggest making an `Array2D` memberv variable `m_dhk_dz` to hold these values.

src/oneD/StFlow.cpp
```diff
@@ -451,22 +456,39 @@ void StFlow::evalResidual(double* x, double* rsd, int* diag,
...
                 double sum2 = 0.0;
-                for (size_t k = 0; k < m_nsp; k++) {
-                    double flxk = 0.5*(m_flux(k,j-1) + m_flux(k,j));
-                    sum += wdot(k,j)*h_RT[k];
-                    sum2 += flxk*cp_R[k]/m_wt[k];
+                
+                if(m_thermo->type() == "IdealGas")
+                {
+                    vector_fp cp_R(m_nsp, 0.0), h_RT(m_nsp, 0.0);
+                    m_thermo->getCp_R(&cp_R[0]);
```
speth: Here and elsewhere, please make sure your code follows the existing style that we fairly uniformly use, as below.
```suggestion
                if (m_thermo->type() == "IdealGas") {
                    vector_fp cp_R(m_nsp, 0.0), h_RT(m_nsp, 0.0);
                    m_thermo->getCp_R(&cp_R[0]);
```

src/oneD/StFlow.cpp
```diff
@@ -451,22 +456,39 @@ void StFlow::evalResidual(double* x, double* rsd, int* diag,
...
+                    m_thermo->getEnthalpy_RT_ref(&h_RT[0]);
+                    for (size_t k = 0; k < m_nsp; k++) {
+                        double flxk = 0.5*(m_flux(k,j-1) + m_flux(k,j));
+                        sum += wdot(k,j)*h_RT[k];
+                        sum2 += flxk*cp_R[k]/m_wt[k];
```
speth: I think the (somewhat suspect) logic behind the old implementation of this was to avoid copying the `h_RT` array that exists in the `m_thermo` object if it happens to be an `IdealGasPhase` object.

We already have an array for `m_cp`, so there's no reason to get `m_cp_R` as well, and I think it would be better to use the `m_hk` array that I've suggested introducing for the partial molar enthalpies, at which point you would also not need to fiddle around with redimensionalizing these arrays, or making a special case between the ideal and non-ideal phase below.

src/oneD/StFlow.cpp
```diff
@@ -898,4 +920,41 @@ void StFlow::evalContinuity(size_t j, double* x, double* rsd, int* diag, double
     }
 }
 
+vector_fp StFlow::grad_hk(const doublereal* x, size_t j) 
+{
```
speth: Remember that new code should not use the `doublereal` typedef.

speth (changes requested):
Thanks for introducing this PR, @gkogekar. Overall, I think it shows that there isn't all that much required to handle the non-ideal phases in the 1D flame model. Besides the in-line comments below, there are couple of things needed for this PR:
- Please eliminate the change to the `googletest` submodule
- I think we need at least a minimal test case for this. I guess that will require adding an input file that has both transport properties _and_ Redlich-Kwong or Peng-Robinson (or critical states) parameters, which I think we currently lack.
- A Python example would be nice as well. I don't know what conditions would actually show an effect from the nonideality.
- The "science" section of the Cantera website will need to be updated as well. I think for posterity, it would be nice to have a simpler derivation of the equations somewhere. The attached PDF is useful, but it has a lot of digressions and dead ends that aren't really necessary.

src/oneD/StFlow.cpp
gkogekar: Corrected.

src/oneD/StFlow.cpp
gkogekar: Corrected.

src/oneD/StFlow.cpp
gkogekar: Corrected.

src/oneD/StFlow.cpp
gkogekar: I have deleted this block of code.

include/cantera/oneD/StFlow.h
gkogekar: Updated.

include/cantera/oneD/StFlow.h
gkogekar: I have updated the comment.

ischoegl:
Hi @gkogekar ... thanks for updating this PR. Unfortunately, it seems like #1452 introduced a couple of minor merge conflicts since the latest rebase. Apart from this, #1448 introduced a couple of changes that should make the Python side quite a bit easier. Would you mind rebasing so this can be pushed over the finishing line? I'd be happy to assist in case you're facing any hurdles.

gkogekar:
@ischoegl Thanks for letting me know. I will rebase and resolve the merge conflicts.

src/oneD/StFlow.cpp
gkogekar: Updated.

src/oneD/StFlow.cpp
gkogekar: Updated.

src/extensions/pythonExtensions.h
ischoegl, line 1: I think this file was removed in `main` and shouldn't be reintroduced.

include/cantera/oneD/StFlow.h
```diff
@@ -320,6 +319,7 @@ class StFlow : public Domain1D
             m_rho[j] = m_thermo->density();
             m_wtm[j] = m_thermo->meanMolecularWeight();
             m_cp[j] = m_thermo->cp_mass();
+            m_thermo->getPartialMolarEnthalpies(&m_hk(0,j));
```
ischoegl:
```suggestion
            m_thermo->getPartialMolarEnthalpies(&m_hk(0, j));
```
whitespace after comma.

src/oneD/StFlow.cpp
```diff
@@ -60,6 +54,8 @@ StFlow::StFlow(ThermoPhase* ph, size_t nsp, size_t points) :
     m_multidiff.resize(m_nsp*m_nsp*m_points);
     m_flux.resize(m_nsp,m_points);
     m_wdot.resize(m_nsp,m_points, 0.0);
+    m_hk.resize(m_nsp,m_points, 0.0);
+    m_dhk_dz.resize(m_nsp,m_points-1, 0.0);
```
ischoegl:
```suggestion
    m_dhk_dz.resize(m_nsp, m_points - 1, 0.0);
```

src/oneD/StFlow.cpp
```diff
@@ -199,6 +195,7 @@ void StFlow::resize(size_t ncomponents, size_t points)
     }
     m_flux.resize(m_nsp,m_points);
     m_wdot.resize(m_nsp,m_points, 0.0);
+    m_hk.resize(m_nsp,m_points, 0.0);
```
ischoegl:
```suggestion
    m_hk.resize(m_nsp ,m_points, 0.0);
```
While you used formatting that matches pre-existing code, the style guide should be followed for new additions (here and elsewhere).

ischoegl (commented):
@gkogekar ... thank you for the rebase, and for taking care of most of the review comments!

Apart from some minor comments for code [style guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md#style-guidelines) I left below, there currently are some issues with the test suite. As they're consistent across platforms, you can likely reproduce them locally?

```
*********************************** Testing Summary ************************************

FAILED: One or more tests failed.
Tests passed: 3447
Up-to-date tests skipped: 0
Tests failed: 301
scons: *** [test_results] Explicit exit, status 1

Failed tests:
    - clib.passed ***no results for entire test suite***
    - oneD.passed ***no results for entire test suite***
    - python ***no results for entire test suite***
    - cxx-flamespeed

****************************************************************************************
```

src/extensions/pythonExtensions.h
speth: Specifically, this is a file automatically generated by Cython, leftover from a previous build. Due to some recent changes, the corresponding `pythonExtensions.pyx` no longer exists, and therefore this `.h` file was removed from the `.gitignore`. I think this may be a reminder of why to never use `git add .`.

src/oneD/StFlow.cpp
gkogekar: Corrected.

src/oneD/StFlow.cpp
gkogekar: Corrected.

include/cantera/oneD/StFlow.h
gkogekar: Corrected.

src/extensions/pythonExtensions.h
gkogekar: Thanks. I didn't realize it. I have removed this file now.

src/oneD/StFlow.cpp
gkogekar: I agree. I have removed the ideal gas code block. Now the energy equation uses the hk formulation for both ideal and non-ideal phases.

gkogekar:
Thanks, @ischoegl. I have taken care of most of the failures from the test suite. But one test (`Reaction.PythonExtensibleRate`) from Kinetics still fails on my Windows machine, and I am not able to figure out the reason for that. But I can see some other tests are now failing on Ubuntu and Mac platforms as well. Is there any way to fix that?

ischoegl:
> Thanks, @ischoegl. I have taken care of most of the failures from the test suite. But one test (`Reaction.PythonExtensibleRate`) from Kinetics still fails on my Windows machine, and I am not able to figure out the reason for that. But I can see some other tests are now failing on Ubuntu and Mac platforms as well. Is there any way to fix that?

Thanks for the prompt fixes! Funny that you get a local failure that CI doesn’t reproduce. For the other failures, I’d be happy to look into things as I’m familiar with that part.

gkogekar:
Thanks @ischoegl. Also let me know if I can do anything to help.

ischoegl:
> Thanks @ischoegl. Also let me know if I can do anything to help.

Fixed. Apparently your changes create slightly different results from the old implementation. So it is pretty much the equivalent to what you already did in the `*flamespeed_blessed.*` files. The context is that there now are (newly added) legacy HDF files that are kept around to ensure that new importers I recently added work with the old format. The tests I wrote had controls that caught the differences you introduced with this PR.

*PS: I cannot reproduce the failure for `Reaction.PythonExtensibleRate` on my machine - I'm on macOS)*

src/oneD/StFlow.cpp
```diff
@@ -20,15 +20,15 @@ StFlow::StFlow(ThermoPhase* ph, size_t nsp, size_t points) :
...
-    if (ph->type() == "ideal-gas") {
-        m_thermo = static_cast<IdealGasPhase*>(ph);
-    } else {
-        throw CanteraError("StFlow::StFlow",
-                           "Unsupported phase type: need 'IdealGasPhase'");
-    }
     m_type = cFlowType;
     m_points = points;
 
+    if (ph->type() == "ideal-gas" || ph->type() == "Redlich-Kwong" || ph->type() == "Peng-Robinson") {
```
ischoegl: Please stay within the 88 character limit; also, there's no need to move this check from the original position.

src/oneD/StFlow.cpp
```diff
@@ -20,15 +20,15 @@ StFlow::StFlow(ThermoPhase* ph, size_t nsp, size_t points) :
...
+        m_thermo = static_cast<ThermoPhase*>(ph);
```
ischoegl: The signature of the function indicates that this already is the correct type, so I'm not sure that the cast is necessary?

src/oneD/StFlow.cpp
```diff
@@ -20,15 +20,15 @@ StFlow::StFlow(ThermoPhase* ph, size_t nsp, size_t points) :
...
+    } else {
+        throw CanteraError("StFlow::StFlow",
+                           "Unsupported phase type: need 'IdealGasPhase' or 'Redlich-Kwong'or 'Peng-Robinson'");
+    }
     if (ph == 0) {
         return; // used to create a dummy object
```
ischoegl, line 27: This check (`ph == 0`) has to be run before calling `ph->type()` to prevent a segmentation fault in case the dummy object is actually used. It's a pre-existing condition, of course ...

ischoegl (commented):
Some additional comments. For the remainder, I defer to @speth as he had some additional comments. I believe he requested a minimal test case?

gkogekar:
> > Thanks @ischoegl. Also let me know if I can do anything to help.
> 
> Fixed. Apparently your changes create slightly different results from the old implementation. So it is pretty much the equivalent to what you already did in the `*flamespeed_blessed.*` files. The context is that there now are (newly added) legacy HDF files that are kept around to ensure that new importers I recently added work with the old format. The tests I wrote had controls that caught the differences you introduced with this PR.
> 
> _PS: I cannot reproduce the failure for `Reaction.PythonExtensibleRate` on my machine - I'm on macOS)_

Thanks, @ischoegl. I am surprised that these files did not give any error on Windows. Anyway, thanks for doing this. I will look at your other comments and update the PR accordingly.

ischoegl:
Hi @gkogekar ... thanks for rebasing. I believe you didn't retain the fix for the CI failure I had applied here when you force pushed. You can cherry-pick the last commit from [this branch](https://github.com/ischoegl/cantera/tree/fix-1079) on my repo, which should fix things again.

gkogekar:
> Hi @gkogekar ... thanks for rebasing. I believe you didn't retain the fix for the CI failure I had applied here when you force pushed. You can cherry-pick the last commit from [this branch](https://github.com/ischoegl/cantera/tree/fix-1079) on my repo, which should fix things again.

Thanks, @ischoegl. I completely missed out on that. I will apply that fix and push it again.

src/oneD/StFlow.cpp
gkogekar: Updated.

src/oneD/StFlow.cpp
gkogekar: Thanks. I have updated it now.

gkogekar:
@speth and @ischoegl I have added a simple python test in test_onedim.py. Basically it's the same as that of the corresponding ideal-gas case. I am still trying to find a better test, but it will need some digging into the literature.

data/h2o2.yaml
```diff
@@ -18,6 +18,15 @@ phases:
   transport: mixture-averaged
   state: {T: 300.0, P: 1 atm}
 
+# Note: The Redlich-Kwong thermodynamic data is obtained from nDodecane_Reitz.yaml
+- name: ohmech-RK
```
speth: Rather than this comment, can you add this to the `description` field at the top of the file, and mention the original source (as in the description in `nDodecane_Reitz.yaml`:
>   Redlich-Kwong coefficients are based on tabulated critical properties or
  estimated according to the method of Joback and Reid, "Estimation of pure-
  component properties from group-contributions," Chem. Eng. Comm. 57 (1987)
  233-243

data/h2o2.yaml
```diff
@@ -37,6 +46,10 @@ species:
     diameter: 2.92
     polarizability: 0.79
     rotational-relaxation: 280.0
+  equation-of-state:
+    model: Redlich-Kwong
+    a: [1.43319e+11, 0]
+    b: 18.42802577
```
speth: I know these are written this way in `nDodecane_Reitz.yaml`, but if the second term in the `a` coefficient is zero, you can just write this as:
```suggestion
    a: 1.43319e+11
    b: 18.42802577
```
I think this would be better, since the two-term model for "a" is kind of an oddity in Cantera's implementation of R-K.

include/cantera/oneD/StFlow.h
```diff
@@ -450,12 +453,16 @@ class StFlow : public Domain1D
     Array2D m_dthermal;
     Array2D m_flux;
 
+    // Arrays for saving molar enthalpies and enthalpy fluxes
+    Array2D m_hk;
+    Array2D m_dhk_dz;
```
speth: Can you describe the shape of these arrays? If you reference member variables with a preceding `#`, they will be linked by Doxygen, e.g. `#m_nsp by #m_points`. Also, these should be Doxygen-style comments (starting with `//!`) with a separate comment for each member variable.

include/cantera/oneD/StFlow.h
```diff
@@ -425,6 +425,9 @@ class StFlow : public Domain1D
     //! Update the diffusive mass fluxes.
     virtual void updateDiffFluxes(const doublereal* x, size_t j0, size_t j1);
 
+    //Get the gradient of speies specific molar enthalpies
```
speth:
```suggestion
    //! Get the gradient of species specific molar enthalpies 
```

src/oneD/StFlow.cpp
```diff
@@ -20,18 +20,19 @@ StFlow::StFlow(ThermoPhase* ph, size_t nsp, size_t points) :
...
     }
+    if (ph->type() == "ideal-gas" || ph->type() == "Redlich-Kwong" || 
+        ph->type() == "Peng-Robinson") {
+        m_thermo = static_cast<ThermoPhase*>(ph);
+    } else {
+        throw CanteraError("StFlow::StFlow", "Unsupported phase type:", 
+            "need 'IdealGasPhase' or 'Redlich-Kwong'or 'Peng-Robinson'");
+    }
```
speth: I'm not sure this check is needed at all. And the `static_cast` is unnecessary -- you can just assign `m_thermo = ph` since it's already at `ThermoPhase*`.

src/oneD/StFlow.cpp
```diff
@@ -561,22 +566,19 @@ void StFlow::evalResidual(double* x, double* rsd, int* diag,
...
+                    sum += wdot(k,j)*m_hk(k,j);
+                    sum2 += flxk * m_dhk_dz(k,j) / m_wt[k];
                 }
-                sum *= GasConstant * T(x,j);
-                double dtdzj = dTdz(x,j);
-                sum2 *= GasConstant * dtdzj;
-
+                
                 rsd[index(c_offset_T, j)] = - m_cp[j]*rho_u(x,j)*dtdzj
                                             - divHeatFlux(x,j) - sum - sum2;
```
speth: Since `sum` and `sum2` aren't being scaled by different values after the loop anymore, you could just combine them into a single value.

test/python/test_onedim.py
```diff
@@ -863,9 +863,9 @@ def check_save_restore(self, f):
         # pytest.approx is used as equality for floats cannot be guaranteed for loaded
         # HDF5 files if they were created on a different OS and/or architecture
         assert list(f.grid) == pytest.approx(list(self.sim.grid))
-        assert list(f.T) == pytest.approx(list(self.sim.T), rel=1e-5)
+        assert list(f.T) == pytest.approx(list(self.sim.T), rel=1e-4)
         k = self.gas.species_index('H2')
-        assert list(f.X[k, :]) == pytest.approx(list(self.sim.X[k, :]), rel=1e-5)
+        assert list(f.X[k, :]) == pytest.approx(list(self.sim.X[k, :]), rel=1e-4)
```
speth, line 882: Rather than relaxing the test tolerances, I think the "reference" files should be replaced with ones based on this PR.

ext/googletest
speth, line 1: This change to the `googletest` submodule needs to be eliminated from the commit history. If you need help doing that, let me know and I can force-push an update to this branch that takes care of it.

test/python/test_onedim.py
```diff
@@ -1301,6 +1301,116 @@ def test_restart(self):
...
+    # (3) Compare the reference files created in the current working directory with
+    #     the ones in test/data and replace them if needed.
+
+    def test_mixture_averaged(self, saveReference=False):
+        T_in = 373.0  # inlet temperature
+        comp = 'H2:1.6, O2:1, AR:7'  # premixed gas composition
+
+        gas = ct.Solution("h2o2.yaml", "ohmech-RK")
+        gas.TPX = T_in, 0.05 * ct.one_atm, comp
+        width = 0.2 # m
```
speth: I think it would be useful to have this regression test at a pressure high enough where the non-ideal model has some effect (if that's even possible for a H2 flame -- I'm not entirely sure).

speth (changes requested):
Thanks for updating this, @gkogekar -- I think it's looking pretty close. Please see below for a few suggestions.

test/python/test_onedim.py
ischoegl: The reference files were created with Cantera 2.6 using tools that have since been removed - these are some regression tests included for backward-compatibility of the legacy HDF format. From that perspective, adjusting the tolerances is all that can be done.

data/h2o2.yaml
gkogekar: I have updated the description.

data/h2o2.yaml
gkogekar: Updated.

include/cantera/oneD/StFlow.h
gkogekar: Corrected.

include/cantera/oneD/StFlow.h
gkogekar: Updated.

src/oneD/StFlow.cpp
gkogekar: If we delete the phase check, then the `test_badInstantiate` fails. So I have kept the check as it is and just changed it to `m_thermo = ph`.

src/oneD/StFlow.cpp
gkogekar: Thanks, I have updated it now.

test/python/test_onedim.py
gkogekar: I have modified the test to run at 10 bar, but truly speaking the temperatures are very high. So there is no difference between the idea gas and real gas models.

ext/googletest
gkogekar: I have no idea how to do that. I tried to search for the related commit, but failed to find it. Can you help me with that?

test/python/test_onedim.py
ischoegl: You could go farther away from stoichiometry? At the same time, this is `h2o2.yaml` …

test/python/test_onedim.py
```diff
@@ -1301,6 +1301,116 @@ def test_restart(self):
...
+            np.savetxt(referenceFile, data, '%11.6e', ', ')
+        else:
+            bad = utilities.compareProfiles(self.test_data_path / referenceFile, data,
+                                            rtol=1e-2, atol=1e-8, xtol=1e-2)
+            self.assertFalse(bad, bad)
+
+        filename = self.test_work_path / "CounterflowPremixedFlame-h2-mix-RK.csv"
+        # In Python >= 3.8, this can be replaced by the missing_ok argument
+        if filename.is_file():
+            filename.unlink()
```
ischoegl: We are requiring Python>=3.8, so 
```suggestion
        filename.unlink(missing_ok=True)
```

gkogekar: I have updated it now.

src/oneD/StFlow.cpp
speth: The phase check is not necessary. Without it, trying to instantiate `StFlow` based on the solid phase defined in `diamond.yaml` will fail because there is no valid transport model, which seems fine to me. You can just modify the test to match `transport model` instead.

ext/googletest
speth: I just force-pushed the branch with this change here. Please fetch and reset your local branch to match this one (head is commit `d655febfe`).

gkogekar: @speth Where is this commit `d655febfe`? The last commit I see is `d655feb `at https://github.com/gkogekar/cantera/tree/nonIdealFlame

speth: That's the same commit, just truncating the hash at a different point (the full hash is `d655febfeebff0c01e6b44effe4b5d64ce4dacba`)

gkogekar: Thanks, I just pushed those changes.

src/oneD/StFlow.cpp
gkogekar: I have updated the test.

src/oneD/StFlow.cpp
gkogekar: Yup. I have removed it now.

speth:
Please don't introduce the merge commit by using `git merge` or `git pull` -- this just brings back in the problematic changes.

With your branch checked out, and with **all uncommitted changes stashed or committed (or else they will be lost)**, you should:
- Run `git log` and note the commit hashes of your commits after `d655feb` (e.g. `cdf9e8c`)
- Run `git reset --hard d655feb`
- Run `git cherry-pick cdf9e8c` and repeat for any other commits you want to keep.

gkogekar:
> Please don't introduce the merge commit by using `git merge` or `git pull` -- this just brings back in the problematic changes.
> 
> With your branch checked out, and with **all uncommitted changes stashed or committed (or else they will be lost)**, you should:
> 
> * Run `git log` and note the commit hashes of your commits after `d655feb` (e.g. `cdf9e8c`)
> * Run `git reset --hard d655feb`
> * Run `git cherry-pick cdf9e8c` and repeat for any other commits you want to keep.

Apologies for that. I am still not getting used to Git advanced commands. Anyway, I have corrected it and also rebased the code with the latest Cantera build.

speth (approved):
Thanks, @gkogekar, this looks good to me. With this completed, could you also open a PR for the `cantera-website`  repository to document the updated 1D flame governing equations?

gkogekar:
> Thanks, @gkogekar, this looks good to me. With this completed, could you also open a PR for the `cantera-website` repository to document the updated 1D flame governing equations?

Thanks, @speth 
Sure. I will do that.

PR #1081, opened 2021-08-01
Refactor StoichManager for sparse Jacobian Calculations

ischoegl:

- Simplify `StoichManager` 
- Leverage `<Eigen/Sparse>` to handle sparse multiplications
- Implement calculation of sparse Jacobian terms

See #Cantera/enhancements#111 and #Cantera/enhancements#100

While calculations are handled by `<Eigen/Sparse>` internally, results are exposed to the Python API as:

```
>>> gas.net_rop_species_derivatives # similar for forward/reverse
>>> gas.net_rop_temperature_derivatives # similar for forward/reverse
>>> gas.net_production_rate_species_derivatives # similar for creation/destruction
>>> gas.net_production_rate_temperature_derivatives # similar for creation/destruction
```

All derivatives are checked against numerical implementations in unit tests. ~At the moment, third-body terms are not considered.~ (*Edit: now done.*) There may be some overlap with #1010, where Jacobians are evaluated for entire `ReactorNet`s (whereas this PR focuses on `GasKinetics`).

**Checklist**


bryanwweber:
I think to fix the failures on the Conda builds, you need to include `config.h` before you check whether `CT_USE_SYSTEM_EIGEN` is true. At least, that's what's done in `eigen_dense.h`, via `ct_defs.h`.

ischoegl:
> I think to fix the failures on the Conda builds, you need to include `config.h` before you check whether `CT_USE_SYSTEM_EIGEN` is true. At least, that's what's done in `eigen_dense.h`, via `ct_defs.h`.

:+1: - Ah - I was wondering about that. Thanks! I ended up moving the include of `StoichManager.h` to the `*.cpp` files, which appears to have solved the issue by a different route. Main reason was that it didn't need to be included everywhere ...

bryanwweber:
> I ended up moving the include of `StoichManager.h` to the `*.cpp` files, which appears to have solved the issue by a different route.

:+1: Nonetheless, I still think it's a good idea to make sure that `config.h` is included before checking `CT_USE_SYSTEM_EIGEN` in `StoichManager.h`, because it may get moved around again at a later date, or used in a different file, etc. and would lead to hard-to-debug errors.

ischoegl:
> Nonetheless, I still think it's a good idea to make sure that `config.h` is included before checking `CT_USE_SYSTEM_EIGEN` in `StoichManager.h`, because it may get moved around again at a later date, or used in a different file, etc. and would lead to hard-to-debug errors.

Fair enough. I changed `eigen_dense.h` to include sparse (and renamed to `eigen_defs.h`), and included that in `SolutionManager.h`. I believe that way everything is taken care of.

ischoegl:
Current benchmarks calculated for `gri30.yaml`:
```
In [4]: %timeit gas.creation_rates
The slowest run took 11.05 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 9.78 us per loop

In [5]: %timeit gas.rop_species_derivatives()
10000 loops, best of 3: 178 us per loop

In [6]: %timeit gas.production_rate_species_derivatives()
10000 loops, best of 3: 187 us per loop

In [7]: %timeit gas.reactant_stoich_coeffs()
10000 loops, best of 3: 86.7 us per loop
```
(note that this is underestimating speedups as gas states are not perturbed and `updateROP` exits early - a numeric version would have to perturb states and execute `updateROP` for each species; further, getting the sparse matrix to Python takes a significant portion of the time as illustrated by `In [7]` - which just exposes an existing sparse matrix).

*Edit: times improved somewhat after introducing `MappedSparseMatrix` (or `Map<SparseMatrix>`) ... 15-20 us less for derivative calculations.*

speth:
Following up on the discussion in Cantera/enhancements#111, I think the replacement of `incrementSpecies` with `dotReaction` introduces some unnecessary inefficiencies. Considering the following definitions:
```cpp
Eigen::VectorXd StoichManagerN::dotReaction(const double* R)
{
    std::fill(m_svec.begin(), m_svec.end(), 0.);
    _dotReaction(m_c1_list.begin(), m_c1_list.end(), R, m_svec);
    _dotReaction(m_c2_list.begin(), m_c2_list.end(), R, m_svec);
    _dotReaction(m_c3_list.begin(), m_c3_list.end(), R, m_svec);
    _dotReaction(m_cn_list.begin(), m_cn_list.end(), R, m_svec);

    return MappedVector(m_svec.data(), m_svec.size());
}

void Kinetics::getNetProductionRates(double* wdot)
{
    updateROP();
    MappedVector mapped(wdot, m_kk);

    // products are created for positive net rate of progress
    mapped = m_productStoich->dotReaction(m_ropnet.data());
    // reactants are destroyed for positive net rate of progress
    mapped -= m_reactantStoich->dotReaction(m_ropnet.data());
}
```
The operations which must be performed are:
1. Zero `StoichManagerN::m_svec`
2. Calculate elements of `StoichManagerN::m_svec` for products
3. Copy elements from `StoichManagerN::m_svec` to `wdot` **
4. Zero `StoichManagerN::m_svec` **
5. Calculate elements of `StoichManagerN::m_svec` for reactants
6. Subtract elements of `StoichManagerN::m_svec` from `wdot` **

Compare this to the set of operations for the existing implementation of `getNetProductionRates`:
```cpp
void Kinetics::getNetProductionRates(doublereal* wdot)
{
    updateROP();

    fill(wdot, wdot + m_kk, 0.0);
    // products are created for positive net rate of progress
    m_revProductStoich.incrementSpecies(m_ropnet.data(), wdot);
    m_irrevProductStoich.incrementSpecies(m_ropnet.data(), wdot);
    // reactants are destroyed for positive net rate of progress
    m_reactantStoich.decrementSpecies(m_ropnet.data(), wdot);
}
```
where the operations are:
1. Zero `wdot`
2. Add to elements of `wdot` for products of reversible reactions
3. Add to elements of `wdot` for products of irreversible reactions (probably a no-op)
4. Subtract from elements of `wdot` for reactants

So, the new implementation effectively adds three new array operations (steps 3, 4, and 6) for every call to `getNetProductionRates` without any simplification of the code. While I would agree that there are probably many places where Eigen could be used to simplify calculations in Cantera, I don't think this is one of them. A similar analysis applies to the introduction and use of the `speciesDot` method.

ischoegl:
@speth ... thanks for your analysis. Your points are well taken - I agree that there's no need to change the machinery for the aggregation in `StoichManagerN`. After thinking about your feedback yesterday, I ended up pursuing a different idea for getting individual components of ROP that no longer involves `Eigen` (I essentially pulled `updateROP` apart, see some of the most recent commits). I haven't had a chance to revert some of the other changes yet (it's not as high priority), so this is why it's still in. I will unwind it in a little bit. At the moment, I am still thinking about how to best set up everything that I can eventually take all of the required derivatives.

You ended up jumping in at a good time btw, so I appreciate your feedback.

ischoegl:
@speth ... I reverted all changes to the production rate calculations.

Here are current timings ...
```
In [2]: %timeit gas.creation_rates
The slowest run took 11.78 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 9.5 us per loop

In [3]: %timeit gas.rop_species_derivatives()
The slowest run took 6.51 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 164 us per loop

In [4]: %timeit gas.production_rate_species_derivatives()
10000 loops, best of 3: 166 us per loop

In [5]: %timeit gas.reactant_stoich_coeffs()
10000 loops, best of 3: 87.7 us per loop

In [6]: %timeit gas.rop_temperature_derivatives(reverse=False)
10000 loops, best of 3: 24.7 us per loop
```
As mentioned earlier, I believe there are some performance hits in the Python API (but those are mostly used for testing).

ischoegl:
As this PR experienced substantial 'feature drift', I am closing this, and continuing with #1089 (which also avoids a meandering commit history).

PR #1051, opened 2021-06-08
Proof-of-concept: Modify reaction rates in memory

ischoegl:
**Changes proposed in this pull request**

This PR builds on #995 ~(and will be rebased once #995 is merged, i.e. 80ish commits are not specific to this PR)~.

- Allow for modifications of rate parameters without having to 'modify' reactions
- Proof-of-concept implemented for `ArrheniusRate`

Addresses Cantera/enhancements#87.

**If applicable, provide an example illustrating new features this pull request is introducing**

```
In [1]: import cantera as ct
   ...: rxn = gas.reaction(0)
   ...: rxn
Out[1]: <ElementaryReaction: H2 + O <=> H + OH>

In [2]: gas.forward_rate_constants
Out[2]: 
array([5.19544737e+03, 1.34865000e+07, 2.55079922e+09, 4.25372812e+01,
       5.48180502e+06, 0.00000000e+00, 3.00385071e-09, 3.82748853e+09,
       5.81796100e+17, 1.93894637e+14, 9.91318424e+06])

In [3]: A = rxn.rate.pre_exponential_factor
   ...: A
Out[3]: 38.7

In [4]: rxn.rate.pre_exponential_factor = 2 * A
   ...: gas.reaction(0).rate.pre_exponential_factor
Out[4]: 77.4000015258789

In [5]: gas.forward_rate_constants
Out[5]: 
array([1.03908950e+04, 1.34865000e+07, 2.55079922e+09, 4.25372812e+01,
       5.48180502e+06, 0.00000000e+00, 3.00385071e-09, 3.82748853e+09,
       5.81796100e+17, 1.93894637e+14, 9.91318424e+06])
```

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

PR #1088, opened 2021-08-22
Small fixes and improvements / sparse coefficient matrices

ischoegl:

This PR contains several minor updates that fix small issues accumulated over several PR's (or address inconsistencies that were left over from the 2.5 code base). 

- Make `Plog` interface consistent: existing code mixed `std::vector<std::pair>` and `std::multimap` (deprecate `std::multimap` versions)
- Make `Chebyshev` interface consistent: existing code mixed `Array2D` and `vector_fp` (deprecate `vector_fp` versions)
- Update internal `Chebyshev` routines to consistently use column major arrays
- Fix warnings in compilation and test suite
- Clarify docstring for SCons `legacy_rate_constants`
- Allow for custom C++ `NotImplementedError` messages (allows for more granular user feedback)
- Renamed `eigen_dense.h` to `eigen_defs.h` (prepares for use of sparse matrices)
- Implement sparse stoichiometric coefficient matrices (`StoichManagerN`) via `Eigen::SparseMatrix`
- Expose sparse stoichiometric coefficient matrices to Python API via `scipy.sparse`

Most of these changes are straight-forward; some are meant to preempt questions that may arise in the context of #1051 and #1081. Specifically, the sparse interface is a preamble to sparse Jacobian output for the Python API in #1081.

Almost all updates are internal. The main API change allows for sparse matrix output (implemented for stoichiometric coefficients):

```
In [1]: import cantera as ct
   ...: gas = ct.Solution("test/data/kineticsfromscratch.yaml")

In [2]: gas.reactant_stoich_coefficients
Out[2]: 
array([[0., 0., 0., 0., 0., 0.],
       [1., 2., 0., 0., 0., 0.],
       [1., 0., 0., 1., 0., 0.],
       [0., 0., 0., 0., 0., 1.],
       [0., 0., 2., 0., 0., 0.],
       [0., 0., 0., 1., 0., 1.],
       [0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 1., 0.]])

In [3]: ct.use_sparse() # enable sparse output via scipy.sparse

In [4]: gas.reactant_stoich_coefficients
Out[4]: 
<9x6 sparse matrix of type '<class 'numpy.float64'>'
	with 9 stored elements in COOrdinate format>

In [5]: print(gas.reactant_stoich_coefficients)
  (1, 0)	1.0
  (2, 0)	1.0
  (1, 1)	2.0
  (4, 2)	2.0
  (2, 3)	1.0
  (5, 3)	1.0
  (8, 4)	1.0
  (3, 5)	1.0
  (5, 5)	1.0

In [6]: gas.reactant_stoich_coefficients.toarray()
Out[6]: 
array([[0., 0., 0., 0., 0., 0.],
       [1., 2., 0., 0., 0., 0.],
       [1., 0., 0., 1., 0., 0.],
       [0., 0., 0., 0., 0., 1.],
       [0., 0., 2., 0., 0., 0.],
       [0., 0., 0., 1., 0., 1.],
       [0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 1., 0.]])
```

**Checklist**

**Other thoughts**

I also considered using `Eigen::MatrixXd` for the external `Chebyshev` interface (2-D matrix of coefficients), but ended up staying closer to the previous implementation.

~For simplicity, the `Eigen::SparseMatrix` interface to Python is implemented using a C preprocessor; the alternative would be to expose some `Eigen` classes in `_cantera.pxd`.~

### DIFF
--- a/include/cantera/cython/wrappers.h
+++ b/include/cantera/cython/wrappers.h
@@ -2,6 +2,7 @@
 // at https://cantera.org/license.txt for license and copyright information.
 
 #include "cantera/base/logger.h"
+#include "cantera/numerics/eigen_sparse.h"
 #include "cantera/thermo/ThermoPhase.h"
 #include "cantera/transport/TransportBase.h"
 #include "cantera/kinetics/Kinetics.h"
@@ -49,6 +50,35 @@ void CxxArray2D_set(Cantera::Array2D& array, size_t i, size_t j, double value)
     array(i,j) = value;
 }
 
+// Service function to pass components describing sparse matrix
+size_t sparseComponents(const Eigen::SparseMatrix<double>& mat,
+    int* rows, int* cols, double* data, size_t length)
+{
+    size_t count = 0;
+    for (int i = 0; i < mat.outerSize(); i++) {
+        for (Eigen::SparseMatrix<double>::InnerIterator it(mat, i); it; ++it) {
+            if (count < length) {
+                rows[count] = it.row();
+                cols[count] = it.col();
+                data[count] = it.value();
+            }
+            count++;
+        }
+    }
+    if (count > length) {
+        throw Cantera::CanteraError("sparseComponents",
+            "Output arrays have insufficient length. Required size is {}, "
+            "while provided length is {}.", count, length);
+    }
+    return count;
+}
+
+// Function which passes sparse matrix components to a set of 1D arrays
+#define SPARSE_FUNC(PREFIX, CLASS_NAME, FUNC_NAME) \
+    size_t PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, \
+    int* rows, int* cols, double* data, size_t length) \
+    { return sparseComponents(object->FUNC_NAME(), rows, cols, data, length); }
+
 // Function which populates a 1D array
 #define ARRAY_FUNC(PREFIX, CLASS_NAME, FUNC_NAME) \
     void PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, double* data) \
@@ -62,6 +92,7 @@ void CxxArray2D_set(Cantera::Array2D& array, size_t i, size_t j, double value)
 
 #define THERMO_1D(FUNC_NAME) ARRAY_FUNC(thermo, ThermoPhase, FUNC_NAME)
 #define KIN_1D(FUNC_NAME) ARRAY_FUNC(kin, Kinetics, FUNC_NAME)
+#define KIN_SPARSE(FUNC_NAME) SPARSE_FUNC(kin, Kinetics, FUNC_NAME)
 #define TRANSPORT_1D(FUNC_NAME) ARRAY_FUNC(tran, Transport, FUNC_NAME)
 #define TRANSPORT_2D(FUNC_NAME) ARRAY_FUNC2(tran, Transport, FUNC_NAME)
 
@@ -88,6 +119,9 @@ THERMO_1D(getGibbs_RT)
 THERMO_1D(getCp_R)
 THERMO_1D(getActivities)
 THERMO_1D(getActivityCoefficients)
+
+KIN_SPARSE(reactantStoichCoeffs)
+KIN_SPARSE(productStoichCoeffs)
 
 KIN_1D(getFwdRatesOfProgress)
 KIN_1D(getRevRatesOfProgress)
--- a/include/cantera/cython/wrappers.h
+++ b/include/cantera/cython/wrappers.h
@@ -73,11 +73,10 @@ size_t sparseComponents(const Eigen::SparseMatrix<double>& mat,
     return count;
 }
 
-// Function which passes sparse matrix components to a set of 1D arrays
-#define SPARSE_FUNC(PREFIX, CLASS_NAME, FUNC_NAME) \
-    size_t PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object, \
-    int* rows, int* cols, double* data, size_t length) \
-    { return sparseComponents(object->FUNC_NAME(), rows, cols, data, length); }
+// Function which passes sparse matrix
+#define SPARSE_MATRIX(PREFIX, CLASS_NAME, FUNC_NAME) \
+    Eigen::SparseMatrix<double> PREFIX ## _ ## FUNC_NAME(Cantera::CLASS_NAME* object) \
+    { return object->FUNC_NAME(); }
 
 // Function which populates a 1D array
 #define ARRAY_FUNC(PREFIX, CLASS_NAME, FUNC_NAME) \
@@ -92,7 +91,7 @@ size_t sparseComponents(const Eigen::SparseMatrix<double>& mat,
 
 #define THERMO_1D(FUNC_NAME) ARRAY_FUNC(thermo, ThermoPhase, FUNC_NAME)
 #define KIN_1D(FUNC_NAME) ARRAY_FUNC(kin, Kinetics, FUNC_NAME)
-#define KIN_SPARSE(FUNC_NAME) SPARSE_FUNC(kin, Kinetics, FUNC_NAME)
+#define KIN_SPARSE_MATRIX(FUNC_NAME) SPARSE_MATRIX(kin, Kinetics, FUNC_NAME)
 #define TRANSPORT_1D(FUNC_NAME) ARRAY_FUNC(tran, Transport, FUNC_NAME)
 #define TRANSPORT_2D(FUNC_NAME) ARRAY_FUNC2(tran, Transport, FUNC_NAME)
 
@@ -120,8 +119,9 @@ THERMO_1D(getCp_R)
 THERMO_1D(getActivities)
 THERMO_1D(getActivityCoefficients)
 
-KIN_SPARSE(reactantStoichCoeffs)
-KIN_SPARSE(productStoichCoeffs)
+KIN_SPARSE_MATRIX(reactantStoichCoeffs)
+KIN_SPARSE_MATRIX(productStoichCoeffs)
+KIN_SPARSE_MATRIX(revProductStoichCoeffs)
 
 KIN_1D(getFwdRatesOfProgress)
 KIN_1D(getRevRatesOfProgress)
--- a/include/cantera/cython/wrappers.h
+++ b/include/cantera/cython/wrappers.h
@@ -50,8 +50,8 @@ void CxxArray2D_set(Cantera::Array2D& array, size_t i, size_t j, double value)
     array(i,j) = value;
 }
 
-// Service function to pass components describing sparse matrix
-size_t sparseComponents(const Eigen::SparseMatrix<double>& mat,
+// Service function to pass index/value triplets describing sparse matrix
+size_t sparseTriplets(const Eigen::SparseMatrix<double>& mat,
     int* rows, int* cols, double* data, size_t length)
 {
     size_t count = 0;
@@ -71,6 +71,32 @@ size_t sparseComponents(const Eigen::SparseMatrix<double>& mat,
             "while provided length is {}.", count, length);
     }
     return count;
+}
+
+// Service function to pass CSC data describing sparse matrix
+void sparseCscData(const Eigen::SparseMatrix<double>& mat,
+    double* value, int* inner, int* outer)
+{
+    if (!mat.isCompressed()) {
+        throw Cantera::CanteraError("sparseCscData",
+            "Invalid input: Eigen matrix is not compressed.");
+    }
+    if (mat.IsRowMajor) {
+        throw Cantera::CanteraError("sparseCscData",
+            "Invalid input: Eigen matrix is not in column major format.");
+    }
+
+    const double* valuePtr = mat.valuePtr();
+    const int* innerPtr = mat.innerIndexPtr();
+    for (size_t i = 0; i < mat.nonZeros(); ++i) {
+        value[i] = valuePtr[i];
+        inner[i] = innerPtr[i];
+    }
+
+    const int* outerPtr = mat.outerIndexPtr();
+    for (size_t i = 0; i < mat.outerSize() + 1; ++i) {
+        outer[i] = outerPtr[i];
+    }
 }
 
 // Function which passes sparse matrix
--- a/include/cantera/cython/wrappers.h
+++ b/include/cantera/cython/wrappers.h
@@ -81,10 +81,6 @@ void sparseCscData(const Eigen::SparseMatrix<double>& mat,
         throw Cantera::CanteraError("sparseCscData",
             "Invalid input: Eigen matrix is not compressed.");
     }
-    if (mat.IsRowMajor) {
-        throw Cantera::CanteraError("sparseCscData",
-            "Invalid input: Eigen matrix is not in column major format.");
-    }
 
     const double* valuePtr = mat.valuePtr();
     const int* innerPtr = mat.innerIndexPtr();
--- a/include/cantera/kinetics/Kinetics.h
+++ b/include/cantera/kinetics/Kinetics.h
@@ -558,6 +558,15 @@ public:
      * @param i   reaction index
      */
     virtual double reactantStoichCoeff(size_t k, size_t i) const;
+
+    /**
+     * Stoichiometric coefficient matrix for reactants.
+     */
+    Eigen::SparseMatrix<double> reactantStoichCoeffs() const
+    {
+        return m_reactantStoich.stoichCoeffs();
+    }
+
     /**
      * Stoichiometric coefficient of species k as a product in reaction i.
      *
@@ -565,6 +574,22 @@ public:
      * @param i   reaction index
      */
     virtual double productStoichCoeff(size_t k, size_t i) const;
+
+    /**
+     * Stoichiometric coefficient matrix for products.
+     */
+    Eigen::SparseMatrix<double> productStoichCoeffs() const
+    {
+        return m_productStoich.stoichCoeffs();
+    }
+
+    /**
+     * Stoichiometric coefficient matrix for products of reversible reactions.
+     */
+    Eigen::SparseMatrix<double> revProductStoichCoeffs() const
+    {
+        return m_revProductStoich.stoichCoeffs();
+    }
 
     //! Reactant order of species k in reaction i.
     /*!
--- a/include/cantera/kinetics/Kinetics.h
+++ b/include/cantera/kinetics/Kinetics.h
@@ -136,7 +136,7 @@ public:
     }
 
     //! Finalize Kinetics object and associated objects
-    virtual void finalizeSetup();
+    virtual void resizeReactions();
 
     //! Number of reactions in the reaction mechanism.
     size_t nReactions() const {
@@ -762,11 +762,11 @@ public:
      * Add a single reaction to the mechanism. Derived classes should call the
      * base class method in addition to handling their own specialized behavior.
      *
-     * @param r      Pointer to the Reaction object to be added.
-     * @param finalize  If `true`, finalize Kinetics object.
+     * @param r       Pointer to the Reaction object to be added.
+     * @param resize  If `true`, resizeReactions is called after reaction is added.
      * @return `true` if the reaction is added or `false` if it was skipped
      */
-    virtual bool addReaction(shared_ptr<Reaction> r, bool finalize=true);
+    virtual bool addReaction(shared_ptr<Reaction> r, bool resize=true);
 
     /**
      * Modify the rate expression associated with a reaction. The
@@ -925,8 +925,8 @@ protected:
     StoichManagerN m_revProductStoich;
     //@}
 
-    //! Boolean indicating whether Kinetics object setup is finalized
-    bool m_finalized;
+    //! Boolean indicating whether Kinetics object is fully configured
+    bool m_ready;
 
     //! The number of species in all of the phases
     //! that participate in this kinetics mechanism.
--- a/include/cantera/kinetics/Kinetics.h
+++ b/include/cantera/kinetics/Kinetics.h
@@ -896,11 +896,11 @@ protected:
     //! Stoichiometry manager for the reactants for each reaction
     StoichManagerN m_reactantStoich;
 
+    //! Stoichiometry manager for the products for each reaction
+    StoichManagerN m_productStoich;
+
     //! Stoichiometry manager for the products of reversible reactions
     StoichManagerN m_revProductStoich;
-
-    //! Stoichiometry manager for the products of irreversible reactions
-    StoichManagerN m_irrevProductStoich;
     //@}
 
     //! Boolean indicating whether Kinetics object setup is finalized
--- a/include/cantera/kinetics/Kinetics.h
+++ b/include/cantera/kinetics/Kinetics.h
@@ -562,8 +562,7 @@ public:
     /**
      * Stoichiometric coefficient matrix for reactants.
      */
-    Eigen::SparseMatrix<double> reactantStoichCoeffs() const
-    {
+    Eigen::SparseMatrix<double> reactantStoichCoeffs() const {
         return m_reactantStoich.stoichCoeffs();
     }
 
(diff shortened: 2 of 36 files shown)


ischoegl:
@speth ... as #1081 was starting to exceed 2k lines, I decided to push a portion that includes relevant changes to the API first. In addition, there is some streamlining of code that arose in the context of #1051.

include/cantera/cython/wrappers.h
```diff
@@ -49,6 +50,35 @@ void CxxArray2D_set(Cantera::Array2D& array, size_t i, size_t j, double value)
     array(i,j) = value;
 }
 
+// Service function to pass components describing sparse matrix
+size_t sparseComponents(const Eigen::SparseMatrix<double>& mat,
+    int* rows, int* cols, double* data, size_t dim)
```
speth: `dim` is kind of a confusing name for the length of the output arrays - maybe something along the lines of `length` or `out_len` would be more clear?

src/kinetics/RxnRates.cpp
```diff
@@ -275,18 +260,37 @@ std::vector<std::pair<double, Arrhenius> > Plog::rates() const
...
+            // starting a new group
+            pressures_[logp] = {j, j+1};
+        } else {
+            // another rate expression at the same pressure
+            pressures_[logp].second = j+1;
+        }
+
+        j++;
+        rates_.push_back(rate.second);
+    }
```
speth: I don't understand the motivation for changing the input data for `Plog` rates from `multimap<double, Arrhenius>` to `vector<pair<double, Arrhenius>>`. What is the benefit?

At a minimum, the method for adding rates here relies on them being sorted from lowest pressure to highest pressure, which occurs automatically for `multimap`, but not with `vector<pair>`. If rates in an input file are provided in a different order, the rates will now be calculated incorrectly.

src/kinetics/RxnRates.cpp
```diff
@@ -345,29 +348,52 @@ void Chebyshev::setup(double Tmin, double Tmax, double Pmin, double Pmax,
...
+const vector_fp& Chebyshev::coeffs() const
+{
+    warn_deprecated("Chebyshev::coeffs", "Behavior to change after Cantera 2.6; "
+        "for new behavior, use getCoeffs().");
+    return chebCoeffs_;
+}
+
+void Chebyshev::setCoeffs(const Array2D& coeffs)
+{
+    m_coeffs = Array2D(coeffs);
```
speth:
```suggestion
    m_coeffs = coeffs;
```

src/kinetics/RxnRates.cpp
```diff
@@ -345,29 +348,52 @@ void Chebyshev::setup(double Tmin, double Tmax, double Pmin, double Pmax,
...
+    nP_ = coeffs.nColumns();
+    nT_ = coeffs.nRows();
```
speth: At this point, I think the `nP_` and `nT_` variables could just be removed in favor of `m_coeffs.nColumns()` and `m_coeffs.nRows()`.

include/cantera/numerics/eigen_defs.h
```diff
@@ -0,0 +1,29 @@
...
+#define CT_EIGEN_DEFS_H
+
+#include "cantera/base/config.h"
+#if CT_USE_SYSTEM_EIGEN
+#include <Eigen/Dense>
+#include <Eigen/Sparse>
+#else
+#include "cantera/ext/Eigen/Dense"
+#include "cantera/ext/Eigen/Sparse"
+#endif
```
speth: If we continue to expand the use of Eigen through Cantera, I think it might be better to keep separate "wrapper" headers for `eigen_dense` and `eigen_sparse`, since there will probably be a lot more source files making use of the dense arrays than the sparse arrays, and including the sparse stuff unnecessarily will slow down compilation.

interfaces/cython/cantera/kinetics.pyx
```diff
@@ -243,28 +268,65 @@ cdef class Kinetics(_SolutionBase):
         The array of reactant stoichiometric coefficients. Element *[k,i]* of
         this array is the reactant stoichiometric coefficient of species *k* in
         reaction *i*.
+
+        .. deprecated:: 2.6
+
+            To be removed after Cantera 2.6; replaceable by property
+            `Kinetics.reactant_stoich_coefficients`.
+
```
speth: What's wrong with the existing name? We use the abbreviation `coeffs` an awful lot. And this still leaves the probably less common `stoich` abbreviation (not that I'd really suggest `reactant_stoichiometric_coefficients` as the name).

interfaces/cython/cantera/kinetics.pyx
```diff
@@ -18,6 +20,29 @@ cdef np.ndarray get_reaction_array(Kinetics kin, kineticsMethod1d method):
...
+    for i in xrange(size):
+        out[rows[i], cols[i]] = data[i]
+    return out
+
+cdef tuple get_sparse(Kinetics kin, kineticsMethodSparse method, size_t max_dim):
+    cdef np.ndarray[int, ndim=1, mode="c"] rows = np.empty(max_dim, dtype=c_int)
+    cdef np.ndarray[int, ndim=1, mode="c"] cols = np.empty(max_dim, dtype=c_int)
+    cdef np.ndarray[np.double_t, ndim=1] data = np.empty(max_dim)
+    size = method(kin.kinetics, &rows[0], &cols[0], &data[0], max_dim)
+    return data[:size], tuple([rows[:size], cols[:size]])
```
speth: As used, I think this is losing much of the benefit of using sparse matrices -- this is allocating enough space for a fully dense array (times two, if `c_int` is a 32-bit type).

I think your alternative suggestion of using a few methods of `Eigen::SparseMatrix` in Cython would allow you to avoid this. I suspect it probably ends up being simpler than this structure paralleling the interface that we use for functions that "return" arrays via their input arguments, since you can just write a Cython function that returns a `scipy.sparse` matrix given an `Eigen::SparseMatrix`. Then the implementation of `reactant_stoich_coeffs` is just `return get_sparse(self.reactantStoichCoeffs())` without having to fiddle around with any macros.

speth (changes requested):
I like the addition of the option to use sparse matrices for the stoichiometric coefficients, but I think there are a few tweaks needed to get it to make full use of the sparsity.

The one change here that I don't really understand is to the data structure used to initialize `Plog` objects.

ischoegl:
@speth ... thank you for the comments. I'll respond to some of your points right away, and should be able to get back to this before long.

interfaces/cython/cantera/kinetics.pyx
ischoegl: Fully agreed: I actually spent some time trying to expose some `Eigen` classes within `_cantera.pxd`, but gave up (I had the C preprocessors already working). I'll give it another shot, but may require some input (specifically, the `Eigen::SparseMatrix<double>::InnerIterator` gave me endless grief).

include/cantera/cython/wrappers.h
ischoegl: ~Would be moot if I can get the Cython route to work.~

src/kinetics/RxnRates.cpp
ischoegl: The reason here was that my goal was to create symmetric `set/getParameters` argument types. Specifically, the import via `AnyMap` uses `vector`. I wasn't aware of the sorting aspect, and will certainly address.

interfaces/cython/cantera/kinetics.pyx
ischoegl: I switched from a method to a property, which I believe is more consistent with the Python API. The intent here was to reduce the churn by just renaming. Alternatives would be to use  `product_stoichiometric_coeffs`, the usual two-step deprecation, or to leave the method alone.

src/kinetics/RxnRates.cpp
ischoegl: ~The alternative to `Array2D` here would be to go to `Eigen::MatrixXd` ... thoughts? The two matrix classes are somewhat redundant, but interfacing `Eigen` with various language API's may be somewhat thorny.~

Keeping the simpler `Array2D` interface.

src/kinetics/RxnRates.cpp
ischoegl: There was actually quite a bit of clutter left - I ended up making everything even more consistent - `Tmin`/`Tmax` and `Pmin`/`Pmax` are replaced by `temperatureRange` and `pressureRange`, respectively. And new constructors use `std::pair<double,double>` to better reflect the `AnyMap` construction route.

interfaces/cython/cantera/kinetics.pyx
ischoegl: I was able to resolve this by keeping `sparseComponents` as a C++ function and only exposing the `Eigen::SparseMatrix<double>` ... that way allocations can be done with exact size information.

interfaces/cython/cantera/kinetics.pyx
ischoegl: I ended up doing the usual - name the new property `Kinetics.product_stoich_coeffs3` for transitional behavior. After 2.6, the old name will be maintained with new behavior.

src/kinetics/RxnRates.cpp
ischoegl: The lack of symmetry was actually for `Plog::setup` and `Plog::rates`, where the former uses `multimap<>` and the latter uses `vector<pair<>>`. I had opted for `vector` as it was more similar to the `AnyMap` route, but didn't realize that there was a benefit to using `multimap`.

I reversed the previous change and instead implemented `Plog::setRates`/`Plog::getRates` that both use `multimap`, and deprecated the old `Plog::setup` and `Plog::rates` in favor of the more consistent setters/getters.

ischoegl:
@speth ... thanks again for the feedback (and I'm happy you liked the sparse coefficient interface, despite the naming squabbles).

I believe I took care of everything you mentioned thus far. The interface to `Eigen::SparseMatrix` no longer uses excessive memory and the `Plog`/`multimap` issue is taken care of. Regarding the Python coefficient getter, I opted for a transitional name, so the old names won't change long-term. 

One new significant change is that I streamlined `Chebyshev` to consistently use range pairs, e.g.
```
In [1]: import cantera as ct

In [2]:  rxn = ct.ChebyshevReaction(
   ...:             equation="HO2 <=> OH + O",
   ...:             rate={"temperature-range": [290.0, 3000.0],
   ...:                   "pressure-range": [1e3, 1e8],
   ...:                   "data": [[8.2883, -1.1397, -0.12059, 0.016034],
   ...:                            [1.9764, 1.0037, 7.2865e-03, -0.030432],
   ...:                            [0.3177, 0.26889, 0.094806, -7.6385e-03]]},
   ...:             kinetics=gas)

In [3]: rxn.rate.temperature_range # previously rxn.Tmin / rxn.Tmax
Out[3]: (290.0, 3000.0)

In [4]: rxn.rate.pressure_range # previously rxn.Pmin / rxn.Pmax
Out[4]: (1000.0, 100000000.0)
```
with corresponding changes in the C++ underpinnings (setters/getters/constructors).

*Edit: Writing this, I realized that `rxn.rate.coeffs` really should be `rxn.rate.data` to be consistent with YAML, which is now fixed:*
```
In [5]: rxn.rate.data  # previously rxn.coeffs
Out[5]: 
array([[ 8.2883e+00, -1.1397e+00, -1.2059e-01,  1.6034e-02],
       [ 1.9764e+00,  1.0037e+00,  7.2865e-03, -3.0432e-02],
       [ 3.1770e-01,  2.6889e-01,  9.4806e-02, -7.6385e-03]])
```

Finally, the transitional names just end with `3`, i.e.
```
In [6]: ct.use_sparse() # enable sparse output via scipy.sparse
   ...: gas.reactant_stoich_coeffs3
   ...: 
Out[6]: 
<10x29 sparse matrix of type '<class 'numpy.float64'>'
	with 50 stored elements in COOrdinate format>
```

ischoegl:
:tada: ... finally found something that will help exposing sparse matrices without copying ... it does look like [scipy.sparse.csc_matrix](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csc_matrix.html#scipy.sparse.csc_matrix) can be initialized from the data structure used by [_compressed_ sparse matrices in`Eigen`](https://eigen.tuxfamily.org/dox/group__TutorialSparse.html) 
> **csc_matrix((data, indices, indptr), [shape=(M, N)])**
>
>    is the standard CSC representation where the row indices for column i are stored in indices[indptr[i]:indptr[i+1]] and their corresponding values are stored in data[indptr[i]:indptr[i+1]]. If the shape parameter is not supplied, the matrix dimensions are inferred from the index arrays.

ischoegl:
Switching to CSC (from COO) format in `scipy.sparse` leverages the internal `Eigen` storage order, and has some impact on performance:
```
In [2]: gas = ct.Solution("gri30.yaml")
   ...: dense = gas.reactant_stoich_coeffs3

In [3]: %timeit gas.reactant_stoich_coeffs3
10000 loops, best of 3: 64.3 us per loop

In [4]: ct.use_sparse()

In [5]: sparse = gas.reactant_stoich_coeffs3
   ...: sparse
Out[5]: 
<53x325 sparse matrix of type '<class 'numpy.float64'>'
	with 626 stored elements in Compressed Sparse Column format>

In [6]: (sparse == dense).all()
Out[6]: True

In [7]: %timeit gas.reactant_stoich_coeffs3
The slowest run took 4.78 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 48 us per loop

In [8]: %timeit gas.reactant_stoich_coeffs3_prev # previous implementation outputting COO matrix
The slowest run took 9.96 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 61.6 us per loop
```
These tests show a 20% speed improvement with CSC (*not* a major saving); CSC is, however, better suited for matrix-vector products, so the switch makes sense overall.

include/cantera/kinetics/StoichManager.h
```diff
@@ -533,7 +459,24 @@ class StoichManagerN
...
+    //! finalize the sparse coefficient matrix setup
+    void finalizeSetup(size_t nSpc, size_t nRxn)
+    {
+        size_t nCoeffs = m_coeffList.size();
+
+        // Stoichiometric coefficient matrix
+        m_stoichCoeffs.setZero();
+        m_stoichCoeffs.resize(nSpc, nRxn);
+        m_stoichCoeffs.reserve(nCoeffs);
+        m_stoichCoeffs.setFromTriplets(m_coeffList.begin(), m_coeffList.end());
```
speth, line 475: While reading some of the Eigen documentation while trying to understand how this works, I noticed that `setFromTriplets` says the existing contents are destroyed, so I think calling `setZero` is unnecessary. It also says the result is a compressed sparse matrix, so I think the following call to `makeCompressed` is also unnecessary.

include/cantera/cython/wrappers.h
```diff
@@ -49,6 +50,60 @@ void CxxArray2D_set(Cantera::Array2D& array, size_t i, size_t j, double value)
...
+    double* value, int* inner, int* outer)
+{
+    if (!mat.isCompressed()) {
+        throw Cantera::CanteraError("sparseCscData",
+            "Invalid input: Eigen matrix is not compressed.");
+    }
+    if (mat.IsRowMajor) {
+        throw Cantera::CanteraError("sparseCscData",
+            "Invalid input: Eigen matrix is not in column major format.");
+    }
```
speth: I think checking `IsRowMajor` is unnecessary here -- for the matrix to be row major, the type would have to be `Eigen::SparseMatrix<double,RowMajor>`.

include/cantera/kinetics/StoichManager.h
```diff
@@ -650,11 +594,30 @@ class StoichManagerN
...
 private:
+    bool m_finalized; //!< Boolean flag indicating whether setup is finalized
+
     std::vector<C1> m_c1_list;
     std::vector<C2> m_c2_list;
     std::vector<C3> m_c3_list;
     std::vector<C_AnyN> m_cn_list;
+
+    //! Sparse matrices for stoichiometric coefficients
+    std::vector<Eigen::Triplet<double>> m_coeffList;
```
speth:
```suggestion
    SparseTriplets m_coeffList;
```
(or eliminate the typedef in `eigen_sparse.h` as it's otherwise unused)

include/cantera/kinetics/Kinetics.h
```diff
@@ -555,6 +558,15 @@ class Kinetics
...
      */
     virtual double reactantStoichCoeff(size_t k, size_t i) const;
+
+    /**
+     * Stoichiometric coefficient matrix for reactants.
+     */
+    Eigen::SparseMatrix<double> reactantStoichCoeffs() const
+    {
+        return m_reactantStoich.stoichCoeffs();
+    }
```
speth: For these one liners, I think the following style is fine (and consistent with the rest of this file):
```suggestion
    Eigen::SparseMatrix<double> reactantStoichCoeffs() const {
        return m_reactantStoich.stoichCoeffs();
    }
```

include/cantera/kinetics/RxnRates.h
```diff
@@ -534,52 +578,120 @@ class Chebyshev
...
+    /*!
+     * @deprecated  Deprecated in Cantera 2.6.
+     *              Accessible as number of columns of the coefficient matrix
+     */
+    size_t nPressure() const
+    {
+        warn_deprecated("Chebyshev::nPressure", "Deprecated in Cantera 2.6; "
+            "accessible as number of colums of the coefficient matrix.");
+        return m_coeffs.nColumns();
     }
```
speth, line 586: I think there is enough of a semantic difference between "number of points in the pressure dimension" and "number of columns in the coefficient matrix" to keep this function (and `nTemperature`) as well.

include/cantera/kinetics/RxnRates.h
```diff
@@ -534,52 +578,120 @@ class Chebyshev
...
-    const vector_fp& coeffs() const {
+    const vector_fp& coeffs() const
+    {
+        warn_deprecated("Chebyshev::coeffs", "Deprecated in Cantera 2.6 "
+            "and to be removed thereafter; replaceable by data().");
         return chebCoeffs_;
     }
 
+    //! Access Chebyshev coefficients as 2-dimensional array with temperature and
+    //! pressure dimensions corresponding to rows and columns, respecitively.
```
speth:
```suggestion
    //! pressure dimensions corresponding to rows and columns, respectively.
```

include/cantera/kinetics/RxnRates.h
```diff
@@ -412,8 +420,15 @@ class Plog
 
     //! Return the pressures and Arrhenius expressions which comprise this
     //! reaction.
+    /*!
+     * @deprecated   To be removed after Cantera 2.6. Replaceable by getRates.
+     */
```
speth, line 424: Is the intention to remove `rates`, or change it to match the updated return type of `getRates`? I think just `rates` should be the preferred name for this getter (since it actually _returns_ the property in question), but I understand the need for a transitional name.

include/cantera/kinetics/StoichManager.h
```diff
@@ -580,9 +523,9 @@ class StoichManagerN
         }
         bool frac = false;
         for (size_t n = 0; n < stoich.size(); n++) {
+            m_coeffList.push_back(Eigen::Triplet<double>(k[n], rxn, stoich[n]));
```
speth: This is a good place for `emplace_back`:
```suggestion
            m_coeffList.emplace_back(k[n], rxn, stoich[n]);
```

interfaces/cython/cantera/_cantera.pxd
```diff
@@ -43,6 +43,18 @@ cdef extern from "cantera/numerics/Func1.h":
...
+    cdef cppclass CxxSparseMatrix "Eigen::SparseMatrix<double>":
+        CxxSparseMatrix()
+        size_t nonZeros()
+        size_t rows()
+        size_t cols()
+        size_t innerSize()
+        size_t outerSize()
+        int* outerIndexPtr()
+        int* innerIndexPtr()
+        double* valuePtr()
```
speth: I don't think these are actually used within any of the `.pyx` files.

include/cantera/kinetics/Kinetics.h
```diff
@@ -135,6 +135,9 @@ class Kinetics
         return "Kinetics";
     }
 
+    //! Finalize Kinetics object and associated objects
+    virtual void finalizeSetup();
```
speth: I'm struggling a little with this function name. At least to me, it implies that further modification of the object is not possible, but that's not the case. I'd say that `initialize` is almost closer to the right term, and is one which we use in the reactor classes for somewhat similar purposes.

speth (changes requested):
Thanks for the updates, @ischoegl. I'm glad to see the reasonably efficient interface for exposing Eigen's sparse matrices to Python. I think that bodes well for some of the other things that I expect are on the horizon like Jacobians.

The change from `Plog::setup` to `Plog::setRates` makes sense, although I think the new function name implies that you could call it again to replace the existing rates and the current implementation doesn't support that as it leaves the existing contents of `rates_` and `pressures_` in tact.

I don't really understand the motivation for packing Chebyshev rate Tmin/Tmax and Pmin/Pmax into `std::pair`. All the extra use of `std::make_pair` and `foo.first`/`foo.second` that's now required doesn't really seem like an improvement over the old interface. In Python, on the other hand, implicit tuple packing/unpacking is clean and idiomatic, so I think that's a reasonable API change.

include/cantera/kinetics/Kinetics.h
ischoegl: :+1: ... I was under the impression that for methods the initial `{` should preferably be on its own line and hence started to apply this whenever touching code. I (personally) prefer what you proposed anyways and will adopt. 

So ... the initial `{` will only be put on its own line if the argument list doesn't fit on the first line?

include/cantera/kinetics/Kinetics.h
ischoegl: The reason here is that `Kinetics::init()` already exists - I picked `Kinetics::finalizeSetup()` to clearly differentiate. The two methods need to be run _before_ and _after_ adding reactions, respectively, and thus serve as bookends of the setup process.

include/cantera/kinetics/RxnRates.h
ischoegl: This is indeed the intent. I'll change the text of the deprecation warning.

ischoegl:
@speth ... thank you for your comments!

> The change from `Plog::setup` to `Plog::setRates` makes sense, although I think the new function name implies that you could call it again to replace the existing rates and the current implementation doesn't support that as it leaves the existing contents of `rates_` and `pressures_` in tact.

Good catch! Some of the renaming here pre-empts #1051 (which is currently on hold), so it will be possible to change that eventually even in memory.

> I don't really understand the motivation for packing `Chebyshev` rate Tmin/Tmax and Pmin/Pmax into `std::pair`. All the extra use of `std::make_pair` and `foo.first`/`foo.second` that's now required doesn't really seem like an improvement over the old interface. In Python, on the other hand, implicit tuple packing/unpacking is clean and idiomatic, so I think that's a reasonable API change.

I partially reverted those changes to C++. I guess I *wanted* to make things as similar as possible for YAML / Python and C++, but for the latter I agree that the resulting API change wasn't great. The updated Python API remains of course.

Regarding `Kinetics::finalizeSetup()` I see your point, but there is indeed a need to call it at a different time than `Kinetics::init()` ... let me know if you can think of a better name, but `initialize` would be imho too confusing.

include/cantera/kinetics/Kinetics.h
speth: Sorry, I forgot about `Kinetics::init`. That would be confusing. Maybe `completeSetup`?

ischoegl: I think complete still has the connotation of being completely done. How about `reset`?

speth: `reset` sounds like it's going to remove something that's already there. Maybe by analogy with `resizeSpecies`, we could have `resizeReactions`, although the use case is slightly different?

ischoegl: :+1: I think `resizeReactions` is an adequate description. Thanks for the suggestion!

ischoegl:
@speth ... after settling on `Kinetics::resizeReactions` (and `StoichManagerN::resizeCoeffs`), I believe everything is taken care of and this should be ready to be merged. Thanks for the review and the suggestions!

speth (approved):
Thanks, @ischoegl! This looks good to me.

ischoegl:
@speth ... thanks again for the review. Now that this is merged, I will post a new PR wrapping up `GasKinetics` (much of this is done, but I'll have to clean up some of the commit history). 

None of my other pending PR's are close to being merged, as most need to be rebased at this point.

PR #1089, opened 2021-08-25
Sparse Jacobians for GasKinetics

ischoegl:

- Add sparse Jacobians for `GasKinetics` 
- Implement mechanism to adjust behavior of Jacobian evaluation via `gas.jacobian_settings` (Python API)
- Allow for disabling of terms to obtain approximate results
- Cover changes with extensive test suite
- Sparse output to Python is supported via `ct.use_sparse()` (requires Scipy)

~The proposed enhancement is fully functional for Jacobians at constant pressure; some caveats remain:~

- ~As the transition from old to new `ReactionRate` is not complete (`Falloff` reactions remain to be ported), some band-aid implementations were necessary, which mostly affect temperature derivatives and a missing derivative term related to reaction rates depending on third-body concentrations.~ … *addressed*
- ~While the calculation of individual terms is optimized for speed, the addition of multiple sparse Jacobian terms is slow. A solution would involve the creation of internal buffers. As Jacobians are expected to be used elsewhere, I am leaving this up for discussion.~
- ~Derivatives with respect to pressure - while relatively straightforward - are currently not implemented.~ … *addressed*

~While I am leaving the PR in draft status,~ - (*edit: work is no longer draft*) - I am explicitly welcoming feedback at this moment (@speth, @kyleniemeyer, @bryanwweber , @anthony-walker, @DavidAkinpelu, among others). It may make sense to discuss some more generic issues in Cantera/enhancements#100.

Closes #Cantera/enhancements#19

Supersedes #1081; incorporates changes proposed in #1088. 

While calculations are handled by `<Eigen/Sparse>` internally, results are exposed to the Python API as:

```python
>>> gas.net_rates_of_progress_ddX # similar for forward/reverse
>>> gas.net_rates_of_progress_ddT # similar for forward/reverse
>>> gas.net_rates_of_progress_ddP # similar for forward/reverse
>>> gas.net_rates_of_progress_ddC # similar for forward/reverse
>>> gas.net_production_rate_ddX # similar for creation/destruction
>>> gas.net_production_rate_ddT # similar for creation/destruction
>>> gas.net_production_rate_ddP # similar for creation/destruction
>>> gas.net_production_rate_ddC # similar for creation/destruction
```

In order to avoid a lengthy interface, derivative evaluation depends on 'settings'.
```
>>> gas.derivative_settings
{'skip-third-bodies': False, 'skip-falloff': False, 'rtol-delta': 1e-08}
```

All derivatives are checked against numerical implementations in unit tests. 
**Checklist**

Additional items:

- [x] The PR is ready for testing
- [x] The PR is feature complete / ready to be merged


include/cantera/kinetics/Kinetics.h
```diff
@@ -568,6 +568,217 @@ class Kinetics
...
+     * Jacobian evaluation is set by keyword/value pairs using the methods
+     * @see getJacobianSettings and @see setJacobianSettings.
+     *
+     * For GasKinetics, the following keyword/value pairs are supported:
+     *  - 'constant-pressure' (boolean) ... if 'true' (default), constant pressure
+     *    is assumed, wheareas constant volume is assumed otherwise. For constant
+     *    pressure, species concentrations are a function of temperature.
+     *  - 'mole-fractions' (boolean) ... if 'true' (default), Jacobians are
+     *    calculated with respect to mole fractions; if 'false' they are calculated
+     *    with respect to species concentrations.
```
speth: Would this be better as `basis: mole-fraction` vs `basis: concentration`, to allow the possibility of other options (e.g. mass fractions) in the future?

ischoegl:
:tada: ... rebased and updated. Some metrics are given below:
```
In [1]: import cantera as ct

In [2]: gas = ct.Solution("gri30.yaml")

In [3]: gas.TPX = 500, ct.one_atm, "CH4:1, O2:2, N2:7"
   ...: gas.equilibrate("HP")

In [4]: ct.use_legacy_rate_constants(False)

In [5]: def test_rop():
   ...:     gas.TP = gas.T + 1e-4, gas.P
   ...:     gas.forward_rates_of_progress

In [6]: def test_rop_ddT():
   ...:     gas.TP = gas.T + 1e-4, gas.P
   ...:     gas.forward_rop_temperature_derivatives

In [7]: def test_rop_ddP():
   ...:     gas.TP = gas.T + 1e-4, gas.P
   ...:     gas.forward_rop_pressure_derivatives

In [8]: def test_rop_ddC():
   ...:     gas.TP = gas.T + 1e-4, gas.P
   ...:     gas.forward_rop_density_derivatives

In [9]: def test_rop_ddX():
   ...:     gas.TP = gas.T + 1e-4, gas.P
   ...:     gas.forward_rop_species_derivatives

In [10]: %timeit test_rop()
14.4 µs ± 22.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [11]: %timeit test_rop_ddT()
19.1 µs ± 558 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [12]: %timeit test_rop_ddP()
15.5 µs ± 45.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [13]: %timeit test_rop_ddC()
20.3 µs ± 156 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [14]: %timeit test_rop_ddX()
308 µs ± 1.92 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [15]: ct.use_sparse() # enable sparse matrix output

In [16]: %timeit test_rop_ddX()
113 µs ± 3.17 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [17]: gas.jacobian_settings
Out[17]: {'skip-third-bodies': False, 'skip-falloff': False, 'rtol-delta': 1e-08}

In [18]: gas.jacobian_settings = {"skip-falloff": True}

In [19]: %timeit test_rop_ddX()
109 µs ± 3.12 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [20]: gas.jacobian_settings = {"skip-falloff": True, "skip-third-bodies": True}

In [21]: %timeit test_rop_ddX()
70.9 µs ± 2.43 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
For `_ddX` runs, there is some overhead for matrix output to Python involved, so the actual computation is more efficient than the timing indicates.

ischoegl:
Adding a sample for speed tests (`samples/cxx/jacobian/jacobian_speed.cpp`):
```
$ ./jacobian_speed 
Benchmark tests for jacobian evaluations.

getFwdRatesOfProgress:  10.701 μs ± 0.133 μs per loop (7 runs, 10000 loops each)
fwdRatesOfProgress_ddT: 14.408 μs ± 0.119 μs per loop (7 runs, 10000 loops each)
fwdRatesOfProgress_ddP: 11.15 μs ± 0.047 μs per loop (7 runs, 10000 loops each)
fwdRatesOfProgress_ddC: 16.042 μs ± 1.2 μs per loop (7 runs, 10000 loops each)
fwdRatesOfProgress_ddX: 40.282 μs ± 0.882 μs per loop (7 runs, 1000 loops each)

getRevRatesOfProgress:  10.796 μs ± 0.0469 μs per loop (7 runs, 10000 loops each)
revRatesOfProgress_ddT: 16.937 μs ± 0.396 μs per loop (7 runs, 10000 loops each)
revRatesOfProgress_ddP: 11.613 μs ± 0.289 μs per loop (7 runs, 10000 loops each)
revRatesOfProgress_ddC: 15.889 μs ± 0.375 μs per loop (7 runs, 10000 loops each)
revRatesOfProgress_ddX: 40.764 μs ± 0.318 μs per loop (7 runs, 1000 loops each)

getNetRatesOfProgress:  10.977 μs ± 0.123 μs per loop (7 runs, 10000 loops each)
netRatesOfProgress_ddT: 17.276 μs ± 1.29 μs per loop (7 runs, 10000 loops each)
netRatesOfProgress_ddP: 11.36 μs ± 0.0522 μs per loop (7 runs, 10000 loops each)
netRatesOfProgress_ddC: 19.624 μs ± 0.24 μs per loop (7 runs, 10000 loops each)
netRatesOfProgress_ddX: 95.056 μs ± 1.22 μs per loop (7 runs, 1000 loops each)

getCreationRates:       12.363 μs ± 0.158 μs per loop (7 runs, 10000 loops each)
creationRates_ddT:      26.087 μs ± 1.78 μs per loop (7 runs, 10000 loops each)
creationRates_ddP:      16.865 μs ± 0.697 μs per loop (7 runs, 10000 loops each)
creationRates_ddC:      25.362 μs ± 0.451 μs per loop (7 runs, 10000 loops each)
creationRates_ddX:      172.49 μs ± 5.23 μs per loop (7 runs, 1000 loops each)

getDestructionRates:    14.38 μs ± 1.79 μs per loop (7 runs, 10000 loops each)
destructionRates_ddT:   27.015 μs ± 1.77 μs per loop (7 runs, 10000 loops each)
destructionRates_ddP:   15.656 μs ± 0.0515 μs per loop (7 runs, 10000 loops each)
destructionRates_ddC:   24.765 μs ± 1.55 μs per loop (7 runs, 10000 loops each)
destructionRates_ddX:   168.58 μs ± 5.99 μs per loop (7 runs, 1000 loops each)

getNetProductionRates:  13.051 μs ± 0.502 μs per loop (7 runs, 10000 loops each)
netProductionRates_ddT: 20.761 μs ± 0.269 μs per loop (7 runs, 10000 loops each)
netProductionRates_ddP: 15.04 μs ± 0.292 μs per loop (7 runs, 10000 loops each)
netProductionRates_ddC: 24.406 μs ± 3.1 μs per loop (7 runs, 10000 loops each)
netProductionRates_ddX: 151.87 μs ± 1.37 μs per loop (7 runs, 1000 loops each)
```

ischoegl:
I believe this is ready for review / testing. ~While some of the CI runners fail at the moment, none of the failures are due to cantera (there appear to be some issues with downloads of external repositories / submodules at the moment).~

include/cantera/kinetics/MultiRateBase.h
```diff
@@ -55,6 +55,36 @@ class MultiRateBase
...
+    //! Evaluate all rate constant temperature derivatives handled by the evaluator;
+    //! which are multiplied with the array of rate-of-progress variables.
+    //! Depending on the implementation of a rate object, either an exact derivative or
+    //! a numerical approximation may be used.
+    //! @param rop  array of rop, which is modified by the method
+    //! @param kf  array of forward rate constants (numerical derivative only)
+    //! @param deltaT  relative temperature perturbation (numerical derivative only)
+    virtual void processRateConstants_ddT(double* rop,
+                                          const double* kf,
+                                          double deltaT) = 0;
```
speth: I think the documentation here is a bit confusing - am I understanding the `rop` array correctly in that for the elements affected by a given `MultiRate` object, it contains the rate of progress on input and d(rop)/dT on output?

For cases like this where it's not obvious, I think it's good to use the extra annotations to `@param`, e.g. `@param[in,out]` or `@param[in]`.

include/cantera/kinetics/Kinetics.h
```diff
@@ -568,6 +568,273 @@ class Kinetics
...
+    {
+        throw NotImplementedError("Kinetics::setJacobianSettings",
+            "Not implemented for kinetics type '{}'.", kineticsType());
+    }
+
+    /**
+     * Calculate Jacobian for forward rate constants with respect to temperature
+     * at constant temperature, molar density and mole fractions
+     */
+    virtual Eigen::VectorXd fwdRateConstants_ddT()
```
speth: Would it be more accurate to describe this as calculating the _column_ of the Jacobian containing partial derivatives of the forward rate constants with respect to T at constant molar density and mole fractions. Also, these derivatives are at constant pressure as well, correct?

include/cantera/kinetics/ReactionData.h
```diff
@@ -128,8 +141,19 @@ struct FalloffData : public ReactionData
...
 
+    //! Perturb third-body concentration vector of data container
+    /**
+     * The method is used for the evaluation of numerical derivatives.
+     * @param deltaM  relative third-body perturbation
+     */
+    virtual bool perturbM(double deltaM);
+
+    //! Restore data container after a perturbation
+    virtual bool restore();
```
speth: Missing `override` specifier here, and for a couple other versions of this function.

src/kinetics/ReactionData.cpp
```diff
@@ -107,6 +131,32 @@ bool FalloffData::update(const ThermoPhase& bulk, const Kinetics& kin)
     return changed;
 }
 
+bool FalloffData::perturbM(double deltaM)
+{
+    // only perturb if there is no buffered value
+    if (m_perturbed) {
+        return false;
+    }
+    std::copy(conc_3b.begin(), conc_3b.end(), m_conc_3b_buf.begin());
```
speth: Simpler alternative:
```c++
m_conc_3b_buf = conc_3b;
```

include/cantera/kinetics/ReactionData.h
```diff
@@ -53,6 +53,16 @@ struct ReactionData
...
 
+    //! Perturb temperature of data container
+    /**
+     * The method is used for the evaluation of numerical derivatives.
+     * @param  deltaT  relative temperature perturbation
+     */
+    virtual bool perturbT(double deltaT);
+
+    //! Restore data container after a perturbation
+    virtual bool restore();
```
speth: What does the return value indicate?

src/kinetics/ReactionData.cpp
```diff
@@ -141,4 +214,27 @@ bool ChebyshevData::update(const ThermoPhase& bulk, const Kinetics& kin)
     return false;
 }
 
+bool ChebyshevData::perturbP(double deltaP)
+{
+    // only perturb if there is no buffered value
+    if (m_pressure_buf > 0.) {
+        return false;
```
speth: Would it make more sense to have calling `pertub` be an error for an already-perturbed `ReactionData` object (for any given perturbation)?

include/cantera/kinetics/ReactionData.h
```diff
@@ -144,6 +168,8 @@ struct FalloffData : public ReactionData
 
 protected:
     int m_state_mf_number; //!< integer that is incremented when composition changes
+    bool m_perturbed; //!< boolean indicating whether perturbed values are used
```
speth: Specifically, whether perturbed 3rd-body values, right?

include/cantera/kinetics/ReactionData.h
```diff
@@ -53,6 +53,16 @@ struct ReactionData
      */
     virtual bool update(const ThermoPhase& bulk, const Kinetics& kin) = 0;
 
+    //! Perturb temperature of data container
+    /**
+     * The method is used for the evaluation of numerical derivatives.
+     * @param  deltaT  relative temperature perturbation
+     */
+    virtual bool perturbT(double deltaT);
```
speth: I think calling this `perturbTemperature` be more consistent with our normal method-naming conventions. Likewise `perturbPressure` and `perturbThirdBodies` (or something along those lines).

include/cantera/kinetics/GasKinetics.h
```diff
@@ -88,11 +107,39 @@ class GasKinetics : public BulkKinetics
...
+    /*!
+     *  This (scaled) derivative is handled by a finite difference.
+     */
+    void processEquilibriumConstants_ddT(double* drkcn);
+
+    //! Apply temperature derivative
+    Eigen::VectorXd ddT(const vector_fp& in);
+
+    //! Apply pressure derivative
+    Eigen::VectorXd ddP(const vector_fp& in);
```
speth: The docstrings for these methods are a bit cryptic. What are the inputs and outputs? `apply_ddT` etc. might be a better naming convention as well.

include/cantera/kinetics/GasKinetics.h
```diff
@@ -88,11 +107,39 @@ class GasKinetics : public BulkKinetics
...
+    Eigen::VectorXd ddC(StoichManagerN& stoich,
+                        const vector_fp& in,
+                        bool mass_action=true);
+
+    //! Apply mole fraction derivative
+    Eigen::SparseMatrix<double> ddX(StoichManagerN& stoich,
+                                    const vector_fp& in);
+
+    //! Calculate rate coefficients
+    void checkLegacyRates(const std::string& name);
```
speth: Incorrect docstring?

include/cantera/kinetics/Kinetics.h
```diff
@@ -568,6 +568,273 @@ class Kinetics
      */
     virtual void getNetProductionRates(doublereal* wdot);
 
+    //! @}
+    //! @name Routines to Calculate Jacobians / Derivatives
+    /*!
+     * Jacobians are calculated with respect to temperature, pressure, molar density
+     * and species concentration for forward/reverse/net rates of progress as well as
+     * creation/destruction and net production of species.
```
speth: Given the abbreviated names used in the actual methods, it might help to describe the naming patterns here explicitly. 

If `ddC` means derivatives with respect to the molar density, would it help the mnemonic if we used the term "molar concentration" here?

And to clarify, the derivatives w.r.t. species concentration are derivatives w.r.t. mole fraction?

include/cantera/kinetics/Kinetics.h
```diff
@@ -568,6 +568,273 @@ class Kinetics
...
+     * Jacobian evaluation is set by keyword/value pairs using the methods
+     * @see getJacobianSettings and @see setJacobianSettings.
+     *
+     * For GasKinetics, the following keyword/value pairs are supported:
+     *
+     *  - 'skip-third-bodies' (boolean) ... if 'false' (default), third body
+     *    concentrations are considered for the evaluation of jacobians
+     *
+     *  - 'skip-falloff' (boolean) ... if 'true' (default), third-body effects
+     *    on reaction rates are not considered.
```
speth: Is this a typo? The name seems confusing if this description is correct.

include/cantera/kinetics/Kinetics.h
```diff
@@ -568,6 +568,273 @@ class Kinetics
...
+    {
+        throw NotImplementedError("Kinetics::revRatesOfProgress_ddC",
+            "Not implemented for kinetics type '{}'.", kineticsType());
+    }
+
+    /**
+     * Calculate Jacobian for reverse rates-of-progress with respect to species
+     * mole fractions at constant temperature, pressure and molar density.
+     */
+    virtual Eigen::SparseMatrix<double> revRatesOfProgress_ddX()
```
speth: Is it worth mentioning that the derivatives w.r.t. `X_i` are with all other `X_j` held constant (rather than satisfying `sum(X_i) = 1`)?

Also, it might help to document that the output is a `nTotalSpecies()` by `nReactions()` matrix.

include/cantera/kinetics/Kinetics.h
```diff
@@ -948,6 +1215,9 @@ class Kinetics
 
     //! Stoichiometry manager for the products of reversible reactions
     StoichManagerN m_revProductStoich;
+
+    //! Effective stoichiometry (products - reactants)
+    Eigen::SparseMatrix<double> m_stoichMatrix;
```
speth: "Net" stoichiometry, rather than "effective"?

include/cantera/kinetics/Kinetics.h
```diff
@@ -1009,6 +1279,9 @@ class Kinetics
     //! Forward rate constant for each reaction
     vector_fp m_rfn;
 
+    //! Delta G^0 times gas constant and temperature for all reactions
+    vector_fp m_delta_gibbs0_RT;
```
speth: The way this is used, I think this vector just contains `Delta G^0` itself.

include/cantera/kinetics/MultiRateBase.h
```diff
@@ -55,6 +55,36 @@ class MultiRateBase
...
+                                          const double* kf,
+                                          double deltaP) = 0;
+
+    //! Evaluate all rate constant third-body derivatives handled by the evaluator;
+    //! which are multiplied with the array of rate-of-progress variables.
+    //! @param rop  array of rop, which is modified by the method
+    //! @param kf  array of forward rate constants
+    //! @param deltaM  relative perturbation of third-body concentrations;
+    //!     if deltaM < 0, rop entries not affected by M are not set to zero
+    virtual void processRateConstants_ddM(double* rop,
```
speth, line 87: Would it make sense to name this `processRateConstants_ddC`, for consistency with the methods in `Kinetics`, since there isn't really any difference between the total concentration and the "third body concentration"?

include/cantera/kinetics/MultiRate.h
```diff
@@ -86,6 +116,100 @@ class MultiBulkRate final : public MultiRateBase
...
+                rop[rxn.first] *= dTinv * (k1 / kf[rxn.first] - 1.);
+            } // else not needed: derivative is already zero
+        }
+
+        // revert changes
+        m_shared.restore();
+    }
+
+    //! Helper function to process third-body derivatives for rate data that
+    //! implement the `perturbM` method.
```
speth: The option to implement `perturbM` and `perturbP` needs to be mentioned in the documentation for the `ReactionRate` class. These are even more important than `ddTScaledFromStruct`, since from the below, the assumption if they are not implemented is that the rate is independent of total concentration / pressure.

include/cantera/kinetics/MultiRate.h
```diff
@@ -86,6 +116,100 @@ class MultiBulkRate final : public MultiRateBase
...
+        m_shared.restore();
+    }
+
+    //! Helper function for rate data that do not implement `perturbM`
+    template <typename T=RateType, typename D=DataType,
+        typename std::enable_if<!has_ddM<D>::value, bool>::type = true>
+    void _process_ddM(double* rop, const double* kf, double deltaM) {
+        if (deltaM < 0.) {
+            return;
+        }
```
speth: When would this case come up? And why does it matter here but not for the case where `perturbM` is defined?

include/cantera/kinetics/StoichManager.h
```diff
@@ -148,11 +148,42 @@ class C1
         R[m_rxn] -= S[m_ic0];
     }
 
+    void updateSetup(const std::vector<std::pair<int, int>>& indices)
+    {
+        size_t count = 0;
```
speth: I'm not really sure what `updateSetup` means, based on the name. Some documentation for this and the following methods would be nice. While the other methods in this class don't have individual docstrings, they are described in the class docstring for `C1`.

include/cantera/kinetics/StoichManager.h
```diff
@@ -474,6 +691,31 @@ class StoichManagerN
...
+        }
+        m_values.resize(nCoeffs, 0.);
+
+        // Set up index pairs for jacobians
+        std::vector<std::pair<int, int>> indices;
+        for (int i = 0; i < tmp.outerSize(); i++) {
+            for (Eigen::SparseMatrix<double>::InnerIterator it(tmp, i); it; ++it) {
+                indices.emplace_back(it.row(), it.col());
+            }
+        }
```
speth: Assuming I've followed the logic correctly, I think the following would substantially simplify the implementation of the `updateSetup` methods:
```suggestion
        std::map<std::pair<int, int>, size_t> indices;
        size_t n = 0;
        for (int i = 0; i < tmp.outerSize(); i++) {
            for (Eigen::SparseMatrix<double>::InnerIterator it(tmp, i); it; ++it) {
                indices[{it.row(), it.col()}] = n++;
            }
        }
```

include/cantera/kinetics/ThirdBodyCalc.h
```diff
@@ -104,7 +106,35 @@ class ThirdBodyCalc3
...
+        Eigen::SparseMatrix<double> efficiencies;
+        efficiencies.setZero();
+        efficiencies.resize(nRxn, nSpc);
+        efficiencies.reserve(m_efficiencyList.size());
+        efficiencies.setFromTriplets(
+            m_efficiencyList.begin(), m_efficiencyList.end());
+
+        // Jacobian matrix multipliers
+        std::vector<Eigen::Triplet<double>> triplets;
+        triplets.reserve(nRxn * nSpc);
```
speth: Isn't this a huge overestimate of the space required?

include/cantera/kinetics/ThirdBodyCalc.h
```diff
@@ -142,6 +209,12 @@ class ThirdBodyCalc3
 
     //! The default efficiency for each reaction
     vector_fp m_default;
+
+    //! Sparse efficiency matrix (compensated for defaults)
+    std::vector<Eigen::Triplet<double>> m_efficiencyList;
```
speth, line 215: Can you make this documentation more clear? Each triplet is a (reaction index, species index, efficiency)?

interfaces/cython/cantera/kinetics.pyx
```diff
@@ -452,6 +456,281 @@ cdef class Kinetics(_SolutionBase):
...
+        For ``GasKinetics``, the following keyword/value pairs are supported:
+
+        -  ``skip-third-bodies`` (boolean) ... if `False` (default), third body
+           concentrations are considered for the evaluation of jacobians
+
+        -  ``skip-falloff`` (boolean) ... if `True` (default), third-body effects
+           on reaction rates are not considered.
+
+        -  ``rtol-delta`` (double) ... relative tolerance used to perturb properties
+           when calculating numerical derivatives.
```
speth: What's the default value?

interfaces/cython/cantera/kinetics.pyx
```diff
@@ -452,6 +456,281 @@ cdef class Kinetics(_SolutionBase):
...
+            return get_mapped(self, kin_fwdRatesOfProgress_ddC, self.n_reactions)
+
+    property forward_rop_species_derivatives:
+        """
+        Calculate Jacobian for forward rates-of-progress with respect to species
+        concentrations at constant temperature, pressure and molar density.
+        For sparse output, set ``ct.use_sparse(True)``.
+        """
+        def __get__(self):
+            if __use_sparse__:
```
speth: We should not invent new double-underscored names. I think this should just be `_use_sparse`.

interfaces/cython/cantera/kinetics.pyx
```diff
@@ -452,6 +456,281 @@ cdef class Kinetics(_SolutionBase):
...
+
+    property forward_rate_constants_density_derivatives:
+        """
+        Calculate Jacobian for forward rate constants with respect to molar density
+        at constant temperature, pressure and mole fractions.
+        """
+        def __get__(self):
+            return get_mapped(self, kin_fwdRateConstants_ddC, self.n_reactions)
+
+    property forward_rop_temperature_derivatives:
```
speth: I know thames are awfully long, but introducing the `rop` abbreviation here makes it a bit harder to discover that there are in fact derivatives available for `foward_rates_of_progress`, for example if you're using tab completion from IPython.

ischoegl:
Adding a minor update to the Jacobian benchmark sample to illustrate the impact of number of species on performance. Here are some results for doubling the number of species ('nDodecane_Reitz.yaml`):
```
nDodecane_Reitz.yaml: 100 species, 553 reactions.

getFwdRatesOfProgress:  12.242 μs ± 0.0352 μs per loop (7 runs, 10000 loops each)
fwdRatesOfProgress_ddT: 14.871 μs ± 0.0279 μs per loop (7 runs, 10000 loops each)
fwdRatesOfProgress_ddP: 12.999 μs ± 0.0848 μs per loop (7 runs, 10000 loops each)
fwdRatesOfProgress_ddC: 15.752 μs ± 0.0196 μs per loop (7 runs, 10000 loops each)
fwdRatesOfProgress_ddX: 55.234 μs ± 0.262 μs per loop (7 runs, 1000 loops each)

getRevRatesOfProgress:  12.272 μs ± 0.0395 μs per loop (7 runs, 10000 loops each)
revRatesOfProgress_ddT: 18.223 μs ± 0.0432 μs per loop (7 runs, 10000 loops each)
revRatesOfProgress_ddP: 12.962 μs ± 0.0295 μs per loop (7 runs, 10000 loops each)
revRatesOfProgress_ddC: 15.539 μs ± 0.0159 μs per loop (7 runs, 10000 loops each)
revRatesOfProgress_ddX: 51.379 μs ± 0.14 μs per loop (7 runs, 1000 loops each)

getNetRatesOfProgress:  12.238 μs ± 0.00841 μs per loop (7 runs, 10000 loops each)
netRatesOfProgress_ddT: 18.186 μs ± 0.0357 μs per loop (7 runs, 10000 loops each)
netRatesOfProgress_ddP: 12.968 μs ± 0.0167 μs per loop (7 runs, 10000 loops each)
netRatesOfProgress_ddC: 19.076 μs ± 0.03 μs per loop (7 runs, 10000 loops each)
netRatesOfProgress_ddX: 116.62 μs ± 0.261 μs per loop (7 runs, 1000 loops each)

getCreationRates:       13.907 μs ± 0.0381 μs per loop (7 runs, 10000 loops each)
creationRates_ddT:      26.293 μs ± 0.0615 μs per loop (7 runs, 10000 loops each)
creationRates_ddP:      18.969 μs ± 0.0303 μs per loop (7 runs, 10000 loops each)
creationRates_ddC:      24.373 μs ± 0.0342 μs per loop (7 runs, 10000 loops each)
creationRates_ddX:      212.18 μs ± 0.327 μs per loop (7 runs, 1000 loops each)

getDestructionRates:    13.362 μs ± 0.0515 μs per loop (7 runs, 10000 loops each)
destructionRates_ddT:   26.806 μs ± 0.0375 μs per loop (7 runs, 10000 loops each)
destructionRates_ddP:   18.765 μs ± 0.0255 μs per loop (7 runs, 10000 loops each)
destructionRates_ddC:   24.123 μs ± 0.0504 μs per loop (7 runs, 10000 loops each)
destructionRates_ddX:   201.09 μs ± 0.3 μs per loop (7 runs, 1000 loops each)

getNetProductionRates:  13.891 μs ± 0.0461 μs per loop (7 runs, 10000 loops each)
netProductionRates_ddT: 20.557 μs ± 0.0467 μs per loop (7 runs, 10000 loops each)
netProductionRates_ddP: 15.341 μs ± 0.0544 μs per loop (7 runs, 10000 loops each)
netProductionRates_ddC: 21.505 μs ± 0.0493 μs per loop (7 runs, 10000 loops each)
netProductionRates_ddX: 202.07 μs ± 0.276 μs per loop (7 runs, 1000 loops each)
```
Nothing else changed (I'm aware of some pending review comments).

ischoegl:
@speth / @anthony-walker  ... one additional comment about the philosophy used here. The Jacobian implemented in this PR is strictly for the source term; it does not consider constraints posed by the EoS, which can only be handled once additional assumptions inherent to a specific model are known (e.g. constant pressure / volume etc.). As such, results are different from a strictly numerical Jacobian, which includes additional terms due to the EoS closure. Note that these terms are implemented in the test suite where numerical / semi-analytical jacobians are compared.

I had included some of the EoS constraints in an earlier version of this PR, but decided to remove them as there were too many possible combinations to anticipate. It is imho a better approach to address these 'mixed' Jacobian terms (mainly an application of the chain rule for derivatives) in the context of a specific application (e.g. various reactor implementations or 1D problems).

src/kinetics/GasKinetics.cpp
```diff
@@ -260,6 +269,302 @@ void GasKinetics::getFwdRateConstants(double* kfwd)
...
+        factor *= rrt_dTinv;
+        factor += m_dn[irxn] * Tinv - m_delta_gibbs0_RT[irxn] * rrtt;
+        drkcn[irxn] *= factor;
+    }
+
+    for (size_t i = 0; i < m_irrev.size(); ++i) {
+        drkcn[m_irrev[i]] = 0.0;
+    }
+
+    thermo().setState_TP(T, P);
```
speth: It might be better to use `saveState` / `restoreState` to make sure that the state is exactly restored -- for any non-ideal phase, using `setState_TP` may vary by whatever the tolerance is on the iterative method for setting the pressure.

src/kinetics/GasKinetics.cpp
```diff
@@ -260,6 +269,302 @@ void GasKinetics::getFwdRateConstants(double* kfwd)
...
+}
+
+Eigen::VectorXd GasKinetics::ddT(const vector_fp& in)
+{
+    // apply temperature derivative
+    Eigen::VectorXd out(nReactions());
+    copy(in.begin(), in.end(), &out[0]);
```
speth: Prefer to use `out.data()` over `&out[0]`.

src/kinetics/GasKinetics.cpp
```diff
@@ -260,6 +269,302 @@ void GasKinetics::getFwdRateConstants(double* kfwd)
...
+    stoich.multiply(m_act_conc.data(), outV.data());
+
+    // derivatives due to reaction rates depending on third-body colliders
+    if (!m_jac_skip_falloff) {
+        for (auto& rates : m_bulk_rates) {
+            // processing step does not modify entries not dependent on M
+            rates->processRateConstants_ddM(
+                outV.data(), m_rfn.data(), -m_jac_rtol_delta);
+        }
+    }
```
speth, line 531: What's the logic behind the negative perturbation?

samples/cxx/jacobian/jacobian_speed.cpp
```diff
@@ -0,0 +1,249 @@
...
+}
+
+//! timer for Eigen::SparseMatrix<double> getters
+void timeit_matrix(Eigen::SparseMatrix<double> (Kinetics::*function)(),
+                   Kinetics* kin,
+                   ThermoPhase& gas,
+                   size_t loops=1000,
+                   size_t runs=7)
+{
+    Eigen::VectorXd out;
```
speth: It looks like this function assigning a `SparseMatrix<double>` to a `VectorXd`? What does that do?

interfaces/cython/cantera/test/test_jacobian.py
```diff
@@ -0,0 +1,708 @@
...
+        cls.gas.X = [0.1, 1e-4, 1e-5, 0.2, 2e-4, 0.3, 1e-6, 5e-5, 0.4]
+        cls.gas.TP = 800, 2 * ct.one_atm
+        super().setUpClass()
+
+
+class TestElementaryIrr(EdgeCases, utilities.CanteraTest):
+    # Irreversible elementary reaction with two reactants
+    # - equation: O + HO2 => OH + O2  # Reaction 1
+    #   rate-constant: {A: 2.0e+13, b: 0.0, Ea: 9.252008e+04}
+    rxn_idx = 0
```
speth, line 433: Rather than having to duplicate the reaction definitions here as a means of documentatation, would it make more sense to just define the parameters of the `Reaction` and use those to instantiate a phase with that one reaction?

interfaces/cython/cantera/test/test_jacobian.py
```diff
@@ -0,0 +1,708 @@
...
+    def test_reverse_rop_ddT(self):
+        pass
+
+    @utilities.unittest.skip("change of reaction enthalpy is not considered")
+    def test_net_rop_ddT(self):
+        pass
+
+
+class FullTests:
+    # Generic test class to check Jacobians evaluated for an entire reaction mechanims
```
speth: "mechanisms"

interfaces/cython/cantera/test/test_jacobian.py
```diff
@@ -0,0 +1,708 @@
...
+        cls.tpx = cls.gas.TPX
+
+    def setUp(self):
+        self.gas.TPX = self.tpx
+        self.gas.jacobian_settings = {} # reset
+
+    def rop_ddX(self, mode, rtol_deltac=1e-9, atol_deltac=1e-20):
+        # numerical derivative for rates-of-progress with respect to mole fractions
+        def calc(fwd):
+            if mode == "forward":
```
speth, line 542: The parameter `fwd` isn't used in this function.

speth (changes requested):
Thanks, @ischoegl, this is a very nice addition, and the core implementation seems very solid. Most of my comments here are aimed at improving the documentation of this capability. Beyond that, I just had a few broader questions.

* I'm wondering about the choice to have functions that return `Eigen::VectorXd`, rather than an array pointer argument as we usually do. The rationale for the latter approach is that (a) it can avoid unnecessary per-call memory allocation when these functions are called repeatedly and (b) it doesn't limit users to our choice of array container, whether that be `std::vector<double>`, `Eigen::ArrayXd` or anything else.

* I think the derivatives with respect to mole fraction at least, if not others, assume that the phase is ideal, i.e. that the activity coefficients are all unity. Maybe it's about time to add an `isIdeal()` method to `ThermoPhase` so this can be checked and the derivative calculations can throw some exception if it is not.

* As you explained, it makes sense to me that these derivatives are calculated with other quantities held constant as the equations are written, independent of whether that satisfies the equation of state or the sum(X) = 1 constraint. I think it would be very useful to provide a bit of guidance to users on this point, that is, noting that they can calculate other derivatives using the chain rule and whatever equation they need to satisfy, and maybe providing an example of how to calculate a Jacobian at constant volume or at constant pressure.

* Lastly, I thought that describing some of these terms as "Jacobians" ended up being a bit awkward for the case where what's being calculated is the derivative with respect to a single variable. Maybe it would be better to more broadly describe these just as "derivatives", though of course it can be noted that they can be combined to form a Jacobian.

samples/cxx/jacobian/jacobian_speed.cpp
ischoegl: Good catch! It's just used to receive the output value. Fixing it doesn't appear to change much for benchmarks, thankfully.

include/cantera/kinetics/Kinetics.h
ischoegl: This is already removed (i.e. comment referenced an earlier version) - I decided to default to mole fractions, and introduce a separate derivative for molar concentration (which resulted in a cleaner / more consistent interface).

include/cantera/kinetics/Kinetics.h
ischoegl: Indeed a typo. It was `true` while the feature was not implemented (and it strictly affects rate constants).

include/cantera/kinetics/MultiRateBase.h
ischoegl: I am hesitant doing this, as it really is a different context. Also, note that `GasKinetics` uses uppercase `M` to denote third-bodies elsewhere (also, this method is called by both `(process_)ddX` and `(process_)ddC`).

include/cantera/kinetics/GasKinetics.h
ischoegl: Using the prefix `process_`, as this is already in use within `MultiRate`.

include/cantera/kinetics/MultiRate.h
ischoegl: While `GasKinetics::(process_)ddC` replaces entries with zeros, `GasKinetics::(process_)ddX` does not (i.e. existing values are not overwritten). I am using the sign of `deltaM` to differentiate theses cases, as indicated in the docstring (see `MultiRateBase.h`). For the alternative case, the derivative is always non-zero.

Adding a dedicated parameter `overwrite` to clarify.

src/kinetics/GasKinetics.cpp
ischoegl: The flag indicates that existing values should not be overwritten. Implementing a dedicated parameters to clarify what is done.

ischoegl: The negative sign was used to indicate that existing values shouldn't be replaced by zeros. Changed this to use a separate parameter to clarify.

include/cantera/kinetics/ThirdBodyCalc.h
ischoegl: ~Yes. But I'm not sure how to estimate as in theory the matrix could be dense. There likely is a way to optimize - although per Perini et al (2012) the sparsity appears to be hard to predict; Table 1 gives examples for sparsity being anywhere between 46 and 98%.~

*Edit: Nevermind ... this is in `ThirdBodyCalc` ... `m_reaction_index.size() * nSpc` should be a much better estimate*

include/cantera/kinetics/Kinetics.h
ischoegl: I am fairly certain that this docstring is correct - there are some instances where pre-existing comments appear to erroneously refer to similar quantities as `Delta G^0`. If you look at line 126 in https://github.com/Cantera/cantera/blob/0706ab32c4f6577b4d12055a6daad8116bc2aab9/src/kinetics/GasKinetics.cpp#L115-L133 the quantity is divided by RT. I also had to track down a correction term to match numerical and semi-analytical results that arises from the multiplication. _PS: when I implemented this, I used Eq 5 in Perini (2012) as comparison as I didn't have Kee on hand at the time ... (will look this up again when I'm back in my office)_

Interestingly, this would mean that the docstring for `Kinetics::getDeltaGibbs` may be incorrect?

interfaces/cython/cantera/kinetics.pyx
ischoegl: 1e-8. (added)

interfaces/cython/cantera/kinetics.pyx
ischoegl: I'm pretty torn on this: `forward_rop_temperature_derivatives` (36 characters) would become `forward_rates_of_progress_temperature_derivatives` (50 characters). 

The alternative would be to use `_ddT` etc. suffixes as in C++, which are less descriptive and (potentially) not pythonic.

interfaces/cython/cantera/kinetics.pyx
ischoegl: This is a pre-existing condition (introduced in #1088). The idea was to use a global similar to `__git_commit__` etc. defined in `utils.pyx`.

*Edit: Replacing by `_USE_SPARSE`*

include/cantera/kinetics/Kinetics.h
ischoegl: Per another comment, I am changing this to *derivative*. There also is a typo as temperature cannot be constant here :joy:

include/cantera/kinetics/StoichManager.h
ischoegl: Could you elaborate? Wouldn't iterating over a  `map` be slower than a classic loop over a `vector`, where `n` is increased automatically? (see `Cx:updateSetup`, which I renamed to `Cx:resizeCoeffs` for consistency). 

As an aside, buffering these indices is key to being able to directly write into an `Eigen::SparseMatrix` while avoiding costly index look-ups. I believe @anthony-walker may have adopted this in #1010, as we discussed this a while back: these offsets can also be used for any composite Jacobian that is assembled from `Kinetics`.

include/cantera/kinetics/StoichManager.h
ischoegl: Renamed to `resizeCoeffs` for consistency. An explanation is added to the group docstring.

interfaces/cython/cantera/test/test_jacobian.py
ischoegl: The way these tests are implemented, all reactions except one are 'switched off', while keeping the original size of the stoichiometric coefficient matrix. That allows for checking that all indices are assigned correctly, which is imho a stricter / more reliable test than assigning a single reaction for a given phase.

While I see your point, the docstring is purely a reminder of what is done, and could in theory be shortened: all the information is easily accessible regardless. 

*Edit: decided to check that equation string and type are consistent, while removing most of the irrelevant information*

ischoegl:
Thank you @speth for the review! I addressed most of your comments, although a few will require clarifications.

Per your broader questions:

> I'm wondering about the choice to have functions that return `Eigen::VectorXd`, rather than an array pointer argument as we usually do. The rationale for the latter approach is that (a) it can avoid unnecessary per-call memory allocation when these functions are called repeatedly and (b) it doesn't limit users to our choice of array container, whether that be `std::vector<double>`, `Eigen::ArrayXd` or anything else.

I believe it makes sense to switch to an array pointer argument. I resorted to this 'easy' solution here as I'm not sure what the best internal C++ API would be in the context of #1010. I'd be interested to hear from @anthony-walker here. For the Python API (which currently handles the test suite and also serves as the interface for our own work), the internal C++ handling is secondary. 

> I think the derivatives with respect to mole fraction at least, if not others, assume that the phase is ideal, i.e. that the activity coefficients are all unity. Maybe it's about time to add an `isIdeal()` method to `ThermoPhase` so this can be checked and the derivative calculations can throw some exception if it is not.

This is correct. I developed this with `IdealGasPhase` in mind. I'll add this once I hear back about the other open questions. ... *Update: done.*

> As you explained, it makes sense to me that these derivatives are calculated with other quantities held constant as the equations are written, independent of whether that satisfies the equation of state or the sum(X) = 1 constraint. I think it would be very useful to provide a bit of guidance to users on this point, that is, noting that they can calculate other derivatives using the chain rule and whatever equation they need to satisfy, and maybe providing an example of how to calculate a Jacobian at constant volume or at constant pressure.

It would probably make sense to type up a short section on this topic on Cantera/cantera-website? There really isn't a great spot to add this information coherently in the source code, mainly due to the 20+ access functions (for *each* API). I tried to point out essential features 'holding XYZ constant' in a succinct way, and have updated the docstrings to further clarify based on your suggestions. ... *Update: see new issue Cantera/cantera-website#169*
 
> Lastly, I thought that describing some of these terms as "Jacobians" ended up being a bit awkward for the case where what's being calculated is the derivative with respect to a single variable. Maybe it would be better to more broadly describe these just as "derivatives", though of course it can be noted that they can be combined to form a Jacobian.

Fair point. I replaced most references to "Jacobian" with "derivatives" (which I had already adopted in the Python API). A few references remain (mostly internal to `StoichManager.h` and `ThirdBodyCalc.h`, plus the `jacobian_settings`). For `jacobian_settings`, most of what is implemented now are place-holders for what I believe will be needed in #1010, where the nomenclature should be consistent. As mentioned in one of my recent comments in #1010, there are some approaches in literature to avoid dense rows/columns for third-bodies; what I put in at the moment may help with estimating speed penalties, but is imho an overly coarse method (ignoring third bodies may save some calculations, but the error introduced may be substantial and thus completely outweigh those savings).

@anthony-walker ... as indicated above, there are some remaining questions about the API as it applies to #1010. These can be discussed now or in a follow-up PR. As mentioned, I'd be happy to assist, but will need some input on this.

include/cantera/kinetics/Kinetics.h
ischoegl: I replaced "molar density" by "molar concentration" throughout (which resulted in renaming of several methods in the Python API).

ischoegl:
@speth / @anthony-walker ... rebased after a merge conflict introduced by #1099; I also took care of a couple of additional items. At this point, GitHub unhelpfully collapses some of the critical comments ...

anthony-walker:
@ischoegl I just rebased with main but I am still having an issue with a couple windows OS's in the CI check but I am working to address them. I do believe however that I am ready to integrate #1089 into #1010. You mentioned that you have created some place holder functions to do this? I can start trying to merge the two branchs.
>I believe it makes sense to switch to an array pointer argument. I resorted to this 'easy' solution here as I'm not sure what the best internal C++ API would be in the context of #1010. I'd be interested to hear from @anthony-walker here. For the Python API (which currently handles the test suite and also serves as the interface for our own work), the internal C++ handling is secondary.
>
I have found the most success with passing and returning raw pointers that map to Eigen objects. That is currently what `ReactionDerivativeManager.h` does in #1010.

ischoegl:
@anthony-walker  ... thanks for your thoughts!

> You mentioned that you have created some place holder functions to do this? I can start trying to merge the two branchs.

Yes and no. For the C++API, the main placeholders are in `Kinetics.h`:
```C++
void getJacobianSettings(AnyMap& settings);
void setJacobianSettings(AnyMap& settings);
```
which presumably should govern how derivatives are calculated (various assumptions for relative perturbations and treatment of 3rd-bodies).

> I have found the most success with passing and returning raw pointers that map to `Eigen` objects.

That makes sense. For derivatives with respect to scalars, this is straight forward:
```C++
void getNetProductionRates(doublereal* wdot); // existing
void getNetProductionRates_ddT(double* dwdot); // in analogy ...
void getNetProductionRates_ddP(double* dwdot);
void getNetProductionRates_ddC(double* dwdot);
```
But for the derivative with respect to mole fractions (`void getNetProductionRates_ddX`) I'm not sure. I believe it would make sense to directly write into the global `Eigen::SparseMatrix`? For this, I believe this would take passing a `size_t` offset value (and this offset needs to be threaded through internally).

speth:
> But for the derivative with respect to mole fractions (`void getNetProductionRates_ddX`) I'm not sure. I believe it would make sense to directly write into the global `Eigen::SparseMatrix`? For this, I believe this would take passing a `size_t` offset value (and this offset needs to be threaded through internally).

I had been thinking about the need for an offset argument as well, so that may be a key part of the interface expansion. What I'm less sure about is whether it's generally useful to just update an existing `SparseMatrix` or whether it would be better to simply return a `SparseMatrix` with that offset applied. I think in a lot of cases, you're going to need to manipulate those entries anyway, for example to go from d/dX at constant P and molar volume to d/dn at constant P only. Also, I think the primary efficient way of building a `SparseMatrix` is from the list of triplets, and I don't know if there's a comparably efficient path for updating an existing matrix.

anthony-walker:
@ischoegl 


> > You mentioned that you have created some place holder functions to do this? I can start trying to merge the two branchs.
> 
> Yes and no. For the C++API, the main placeholders are in `Kinetics.h`:
> 
> ```c++
> void getJacobianSettings(AnyMap& settings);
> void setJacobianSettings(AnyMap& settings);
> ```

Are these methods to apply a custom mapping to the a sparse representation of the derivatives? If so do you have methods to form the sparse mapping? I use the current `ReactionDerivativeManager` in #1010  to do both to some extent.

> ```c++
> void getNetProductionRates(doublereal* wdot); // existing
> void getNetProductionRates_ddT(double* dwdot); // in analogy ...
> void getNetProductionRates_ddP(double* dwdot);
> void getNetProductionRates_ddC(double* dwdot);
> ```

I am also using derivatives with respect to moles in #1010, is that a possibility here?

ischoegl:
@anthony-walker  ...  the jacobian settings are mainly to pass some assumptions to C++, e.g. from Python
```
>>> gas.jacobian_settings
{'skip-third-bodies': False, 'skip-falloff': False, 'rtol-delta': 1e-08}
```
> I am also using derivatives with respect to moles in #1010, is that a possibility here?

... total or individual? Here, `getNetProductionRates_ddC` is the derivative with respect to (total) molar concentration.

anthony-walker:
@ischoegl  Ah, I would most likely use the C++ interface. I will need a way to remap the output which I believe I discussed a bit in #1010. I can work on that when I start merging the two of them.
> @anthony-walker ... the jacobian settings are mainly to pass some assumptions to C++, e.g. from Python

My approximate Jacobian based preconditioner is built from derivatives with respect to moles of individual species. 
> ... total or individual? Here, `getNetProductionRates_ddC` is the derivative with respect to (total) molar concentration.

ischoegl:
@speth / @anthony-walker  ... thanks for your comments thus far. Based on what was said, I believe I will proceed as follows:

* [x] Rename `get/setJacobianSettings` to `get/setDerivativeSettings`
* [x] Change the `*_ddT`, `*_ddP`, `*_ddC` methods to use raw pointers

but leave `*_ddX` as is.

ischoegl:
@anthony-walker 

> I would most likely use the C++ interface.

It is equivalent, and uses an `AnyMap` instead of a dictionary.

> I will need a way to remap the output which I believe I discussed a bit in #1010. I can work on that when I start merging the two of them.

I don't think that this is mutually exclusive: this PR should be using the same mapping as we discussed a while back; it's just that entries are shifted by an offset within the sparse element storage vector.

> My approximate Jacobian based preconditioner is built from derivatives with respect to moles of individual species

All it takes is to multiply `_ddX` by the molar concentration. It would be simple to pass an optional scaling factor.

**Note:** As mentioned somewhere above (in the collapsed portion) ... the derivatives here do not incorporate constraints posed by the EoS, which may be another difference.

include/cantera/kinetics/Kinetics.h
speth: Right, in the expression `exp(m_rkcn[irxn]*rrt - m_dn[irxn]*m_logStandConc)`, it's dividing `Delta G^0` by `RT` to get a nondimensional quantity, since the argument to `exp()` should be nondimensional.

include/cantera/kinetics/MultiRateBase.h
speth: I suppose this is a bit theoretical -- for all current rate parameterizations, if the rate constant depends on the total concentration, it is expressed as a "third body" where the additional notion of efficiencies come in, and I can see wanting to avoid conflating a general dependence on total concentration with concentrations of specific species that comes in though non-unity efficiencies. So, I'm fine with leaving this as is.

include/cantera/kinetics/ThirdBodyCalc.h
speth: Building a `std::vector` element-by-element is always just an O(n) operation, even if it invokes some internal resize operations. If there's an easy way to know how much space is needed ahead of time, it can save a small amount of time, but I think making a much larger than necessary allocation could actually backfire, and it would be better to just take the simpler route of not reserving anything.

interfaces/cython/cantera/kinetics.pyx
speth: Well, `forward_rate_constants_temperature_derivatives` (which already exists) is 46 characters, so 4 more isn't all that much.

But now that you mention it, abbreviating `T`, `P`, and `X` is actually consistent with the Python interface, where those abbreviations are already used. So perhaps either the `_ddT` suffix like in C++ would be reasonable, or something like `forward_rates_of_progress_T_derivatives` which is "only" 39 characters. I'm less sure about `C` as an abbreviation for the Python module.

include/cantera/kinetics/StoichManager.h
speth: With this map structure, though, you don't need to iterate and check each of the elements -- you just look up the element using the (reaction index, species index) tuple as the key. I went ahead and implemented this to demonstrate -- feel free to cherry-pick https://github.com/speth/cantera/commit/f1c1e686bcb09763c5efc8f2a19ddbee9bd350a9.

It's both a lot less code, and faster. Looking up an element in the map is O(log n), while the linear search is O(n); using an `unordered_map` would actually cut this down to O(1), but I didn't think it was worth writing a custom hash function for.

include/cantera/kinetics/Kinetics.h
ischoegl: Makes sense. I guess there is a discrepancy in nomenclature.

include/cantera/kinetics/StoichManager.h
ischoegl: Cherry-picked the suggested update.

interfaces/cython/cantera/kinetics.pyx
ischoegl: ok. Since `_ddT` etc are already used in the C++ API, and uppercase letters are valid in some instances, I am going with the shortest version after all.

include/cantera/kinetics/Kinetics.h
ischoegl: Fixed.

ischoegl:
@speth ... thanks for responding to my queries quickly! At this point, I believe everything that was pointed out is addressed; let me know if there's anything else.

Otherwise, this is ready for another round of reviews from my end.

ischoegl:
**PS:** While I'd suggest to leave tweaks of the internal C++ `_ddX` API so it fits with #1010 for a follow-up PR, I wanted to respond to one comment:

> I had been thinking about the need for an offset argument as well, so that may be a key part of the interface expansion. What I'm less sure about is whether it's generally useful to just update an existing `SparseMatrix` or whether it would be better to simply return a `SparseMatrix` with that offset applied. I think in a lot of cases, you're going to need to manipulate those entries anyway, for example to go from d/dX at constant P and molar volume to d/dn at constant P only. Also, I think the primary efficient way of building a `SparseMatrix` is from the list of triplets, and I don't know if there's a comparably efficient path for updating an existing matrix.

One alternative would be to pass a fully created `SparseMatrix` (initially zeroed out, but with properly referenced memory layout), and have the `_ddX` methods modify values in memory (i.e. equivalent to `+=` or `-=`); correct indices should be easily found based on the offset argument and internally stored indices. At least in this PR, I tried to avoid constructing new `SparseMatrix` variables from triplet lists at all cost: it is my understanding that anything that involves indexing/building is slow, which is why I resorted to accessing the underlying memory map directly and otherwise relied on copy constructors.

**PPS:** The best approach here may be to leave the current C++ getters that return `Eigen::SparseMatrix<double>` for Python (or any other higher-level API), but also have an alternative `void` accessor that passes a fully formed `SparseMatrix` for internal use within the C++ core.

include/cantera/kinetics/ReactionData.h
speth: But this return value is never used, is it? The only place I see this function called is by `MultiRate`, which doesn't use it.

ischoegl: :+1: It is not necessary. Removing.

speth (approved):
Thanks for the updates, @ischoegl. Looks like you also found my one last request for update (regarding `ReactionData::restore`) that GitHub was trying to hide.

ischoegl:
@speth ... thanks for your prompt (and thorough!) review! Beyond: @anthony-walker I'd be happy to assist with blending this with #1010. Let me know ...

PR #1103, opened 2021-09-20
Fix some 1D solver errors

### LINKED ISSUE #1067: Extracted submechanism causes failure with CounterflowDiffusionFlame
**Problem description**

A submechanism created using the extract_submechanism.py example causes a `CounterflowDiffusionFlame` simulation to fail, returning only a `floating point exception`, and no further debug info (even with `loglevel=5`).  The submechanism does not cause failure with `CounterflowPremixedFlame` or `IdealGasConstPressureReactor`.  I also see the same behavior  when using mechanisms other than gri30.yaml

**Steps to reproduce**

Run the attached modified version of extract_submechanism.py.  The simulation runs as normal with the full mechanism, but fails with the submechanism.  Uncomment the original `CounterflowPremixedFlame` function for a solution with both the full mechanism and submechanism.

**System information**

- Cantera: same results with 2.5.1 or dev version 2.6.0a2
- OS: Mac OS 11.4 and Windows 10
- Python 3.9

**Attachments**

[extract_submechanism_bug.zip](https://github.com/Cantera/cantera/files/6748698/extract_submechanism_bug.zip)

**Additional context**

Looking at the extracted submechanism after writing with write_yaml, I don't see any obvious errors.  I have also tried removing undeclared third-body species from the submechanism, but without any change in behavior.

speth:

* Fixed an error in the error handler for singular Jacobians where the wrong domain was being identified as containing the  problematic row, that ultimately resulted in an integer divide by zero error instead of a `CanteraError` identifying which domain, grid point, and solution component the row corresponded to. This fixes a regression introduced in e8194afeb.
* Fix some cases where the solution involves large values of lambda and the initial guess of lambda = 0 was leading to singular Jacobians. A reasonable initial guess for lambda can be obtained by noting the proportionality between lambda and rho*V**2 in the radial momentum equation.
* Fix use of user-specified solver tolerances after internal solver restarts, e.g. automatic grid expansion of `FreeFlame`

Resolves #1067

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/onedim.py
+++ b/interfaces/cython/cantera/onedim.py
@@ -1160,7 +1160,8 @@ class CounterflowDiffusionFlame(FlameBase):
         Yin_f = self.fuel_inlet.Y
         self.gas.TPY = self.fuel_inlet.T, self.P, Yin_f
         mdotf = self.fuel_inlet.mdot
-        u0f = mdotf / self.gas.density
+        rho0f = self.gas.density
+        u0f = mdotf / rho0f
         T0f = self.fuel_inlet.T
 
         sFuel = moles('O')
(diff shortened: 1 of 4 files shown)


ischoegl (approved):
Looks good to me - thanks for tracking down the issue!

PR #1110, opened 2021-10-02
Avoid C/C++ math constants specified as #define

### LINKED ISSUE #1108: Math constants not found for mingw
**Problem description**

Per [UG report](https://groups.google.com/g/cantera-users/c/7TsfXeWGIy8), compilation fails with the following error:
```
g++ -o build\src\thermo\PengRobinson.o -c -std=c++0x -O3 -Wno-inline -g -Wall -include src/pch/system.h -DNDEBUG -Iinclude -Iinclude\cantera\ext -Ibuild\src src\thermo\PengRobinson.cpp
src\thermo\PengRobinson.cpp: In member function 'virtual double Cantera::PengRobinson::cp_mole() const':
src\thermo\PengRobinson.cpp:110:28: error: 'M_SQRT2' was not declared in this scope
     double vpb = mv + (1 + M_SQRT2)*m_b;
                            ^~~~~~~
[...]
```

`M_SQRT2` is a math constant defined in ~`<cmath>`, which is included via `include/cantera/base/ct_defs.h`~ `<math.h>`, which is included at the top of `PengRobinson.cp`, including the required
```C++
#define _USE_MATH_DEFINES
```
(see [here](https://www.quantstart.com/articles/Mathematical-Constants-in-C/ )).
 
**Steps to reproduce**

See [UG report](https://groups.google.com/g/cantera-users/c/7TsfXeWGIy8). In other words, the error is thrown despite constants being explicitly included.

**System information**

- Cantera version: 2.6.0a3
- OS: Windows 10

**Attachments**

* [buildlog.txt](https://github.com/Cantera/cantera/files/7261819/buildlog.txt)

ischoegl:

Switch from math constants defined in C-library `math.h` to a regular `const double` definition - `M_SQRT2` was the only constant used.

Closes #1108

**Checklist**

**Other context**

Per [stackoverflow](https://stackoverflow.com/questions/10694255/cmath-vs-math-h-and-similar-c-prefixed-vs-h-extension-headers) comment there may be some quirks related to what version is included (i.e. `<cmath>` vs `<math.h>`):

> `<math.h>` and any `<xxx.h>` headers are not standard C++, despite being supported on every major implementation. However, since they are deprecated, there is no guarantee of what is inside those headers when you include them on your implementation. In fact, it has been observed on certain implementations that they provide functions that behave differently to the `<cxxx>` versions.

As there is only a single constant used (`M_SQRT2`), the proposed fix is likely more economical.

### DIFF
--- a/include/cantera/base/ct_defs.h
+++ b/include/cantera/base/ct_defs.h
@@ -1,12 +1,11 @@
 /**
  * @file ct_defs.h
- * This file contains definitions of terms that are used in internal
- * routines and are unlikely to need modifying (text for module physConstants (see \ref physConstants) is found here).
- * This file is included
- * in every file that is in the Cantera Namespace.
+ * This file contains definitions of constants, types and terms that are used
+ * in internal routines and are unlikely to need modifying.
  *
- * All physical constants are stored here.
- * The module physConstants is defined here.
+ * All physical constants are stored here (see module \ref physConstants).
+ *
+ * This file is included in every file within the Cantera namespace.
  */
 
 // This file is part of Cantera. See License.txt in the top-level directory or
--- a/include/cantera/base/ct_defs.h
+++ b/include/cantera/base/ct_defs.h
@@ -51,6 +51,9 @@ using std::isnan; // workaround for bug in libstdc++ 4.8
 //! Pi
 const double Pi = 3.14159265358979323846;
 
+//! Sqrt(2)
+const double Sqrt2 = 1.41421356237309504880;
+
 /*!
  * @name Defined Constants
  * These constants are defined by CODATA to have a particular value.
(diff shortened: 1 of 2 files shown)


ischoegl:
This PR should resolve the issue reported on the [UG](https://groups.google.com/g/cantera-users/c/7TsfXeWGIy8) as it completely sidesteps the import that is not working correctly for mingw.

speth (approved):
This looks fine to me.

Someday, when we're ready to require a C++20 compatible compiler as our minimum, we can replace both this and `Pi` with definitions from [`std::numbers`](https://en.cppreference.com/w/cpp/header/numbers), but I think that's going to be at least a couple of years from now.

PR #1111, opened 2021-10-04
clarifying coverage dependencies doc

### LINKED ISSUE #1068: Documentation for coverage-dependence E parameter
I suggest/request an improvement to the documentation for coverage-dependencies that shows up at 
https://cantera.org/documentation/dev/sphinx/html/yaml/reactions.html#interface

> coverage-dependencies
> A mapping of species names to coverage dependence parameters, where these parameters are contained in a mapping with the fields:
> 
> a
> Coefficient for exponential dependence on the coverage
> 
> m
> Power-law exponent of coverage dependence
> 
> E
> Activation energy dependence on coverage
> Example:
> 
> `equation: 2 H(s) => H2 + 2 Pt(s)`
> `rate-constant: {A: 3.7e21 cm^2/mol/s, b: 0, Ea: 67400 J/mol}`
> `coverage-dependencies: {H(s): {a: 0, m: 0, E: -6000 J/mol}}`

It's factually correct, but you need to cross-reference with the documentation at 
https://cantera.org/science/reactions.html#surface-reactions
and read the equations carefully and think to be sure that you get the sign right on $E_k$.

Some researchers publish using $\eplison_i$ which is the negative of $E_k$, and Detchem uses that convention.

There's no mistake in Cantera's documentation, and it's consistent with Chemkin's, but given the high potential for error here (speaking from experience), I think we could be more explicit and can afford a few extra words (especially as the chosen example has a negative value for E). 

I can open a PR if nobody beats me to it, but wanted to at least open an issue before I forget.

mazeau:
I'm just adding a bit of clarification to make the docs a bit more idiot-proof. This is in response to and should close #1068

### DIFF
--- a/doc/sphinx/yaml/reactions.rst
+++ b/doc/sphinx/yaml/reactions.rst
@@ -321,7 +321,8 @@ Includes the fields of an :ref:`sec-yaml-elementary` reaction plus:
         Power-law exponent of coverage dependence
 
     ``E``
-        Activation energy dependence on coverage
+        Activation energy dependence on coverage.
+        This is the negative of :math:`{\epsilon}_i` and the Detchem convention
 
 Example::
 


speth (changes requested):
Thanks for looking into this, as it provides a good starting point for discussion.

I'm not sure what the reference to "epsilon_i" is supposed to be (here or in @rwest's original issue report). The Kee et al _Chemically Reacting Flow_ book provides the following expression:
image
which uses the name "epsilon" but otherwise seems consistent with what Cantera does.

Rather than trying to explain this only in reference to sources that do not follow the same convention, I think it would be helpful to further describe what Cantera does, so something along the line of
> Here, _E_ has the same sign convention as the leading-order activation energy term. Note that this is the opposite sign convention convention of that used by Detchem and some literature sources.

bryanwweber:
@mazeau Are you interested in fixing up the comments from @speth on this PR?

ischoegl:
I am closing this due to inaction. @mazeau feel free to reopen. The PR remains linked to #1068, if anyone else wants to take another stab here.

PR #1112, opened 2021-10-06
Use YAML for saving and loading 1D flame simulations

speth:

In order to move ahead with the deprecation and removal of the XML format, we need a replacement for saving and loading 1D simulation results for all language interfaces. While the HDF5-based format available in the Python module is clearly the best choice for that interface, using YAML is the simplest option for the other interfaces.

- Save 1D flame simulations to YAML files
- Load 1D flame simulations from YAML files
- Fix a couple of errors in the Matlab toolbox related to the use of the now-removed component names `u` and `v` (replaced by `velocity` and `spread_rate` in Cantera 2.5.0).

**Checklist**

### DIFF
--- a/include/cantera/oneD/Domain1D.h
+++ b/include/cantera/oneD/Domain1D.h
@@ -518,7 +518,6 @@ protected:
 
     //! Identity tag for the domain
     std::string m_id;
-    std::string m_desc;
     std::unique_ptr<Refiner> m_refiner;
     std::vector<std::string> m_name;
     int m_bw;
--- a/include/cantera/oneD/Domain1D.h
+++ b/include/cantera/oneD/Domain1D.h
@@ -27,6 +27,7 @@ const int cPorousType = 109;
 class MultiJac;
 class OneDim;
 class Refiner;
+class AnyMap;
 class XML_Node;
 
 /**
@@ -339,6 +340,12 @@ public:
      *     Cantera 3.0.
      */
     virtual void restore(const XML_Node& dom, doublereal* soln, int loglevel);
+
+    //! Save the state of this domain as an AnyMap
+    /*!
+     * @param soln local solution vector for this domain
+     */
+    virtual AnyMap serialize(const double* soln) const;
 
     size_t size() const {
         return m_nv*m_points;
--- a/include/cantera/oneD/Domain1D.h
+++ b/include/cantera/oneD/Domain1D.h
@@ -347,6 +347,15 @@ public:
      */
     virtual AnyMap serialize(const double* soln) const;
 
+    //! Restore the solution for this domain from an AnyMap
+    /*!
+     * @param[in]  state AnyMap defining the state of this domain
+     * @param[out] soln Value of the solution vector, local to this domain
+     * @param[in]  loglevel 0 to suppress all output; 1 to show warnings; 2 for
+     *      verbose output
+     */
+    virtual void restore(const AnyMap& state, double* soln, int loglevel);
+
     size_t size() const {
         return m_nv*m_points;
     }
--- a/samples/matlab/catcomb.m
+++ b/samples/matlab/catcomb.m
@@ -177,7 +177,7 @@ solve(sim1D, loglevel, refine_grid);
 sim1D
 
 % save the solution
-saveSoln(sim1D,'catcomb.xml','energy',['solution with energy equation']);
+saveSoln(sim1D,'catcomb.yaml','energy',['solution with energy equation']);
 
 %%%%%%%%%% show statistics %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 writeStats(sim1D);
--- a/samples/matlab/catcomb.m
+++ b/samples/matlab/catcomb.m
@@ -125,10 +125,11 @@ setTemperature(surf,tsurf);
 sim1D = Stack([inlt, flow, surf]);
 
 % set the initial profiles.
-setProfile(sim1D, 2, {'u', 'V', 'T'}, [0.0            1.0       % z/zmax
-                                       0.06           0.0       % u
-                                       0.0            0.0       % V
-                                       tinlet         tsurf]);  % T
+setProfile(sim1D, 2, {'velocity', 'spread_rate', 'T'}, ...
+           [0.0            1.0       % z/zmax
+            0.06           0.0       % velocity (u)
+            0.0            0.0       % spread rate (V)
+            tinlet         tsurf]);  % T
 names = speciesNames(gas);
 for k = 1:nSpecies(gas)
   y = massFraction(inlt, k);
--- a/src/base/AnyMap.cpp
+++ b/src/base/AnyMap.cpp
@@ -213,6 +213,7 @@ Cantera::AnyValue Empty;
 namespace YAML { // YAML converters
 
 using namespace Cantera;
+static const int max_line_length = 87;
 
 template<>
 struct convert<Cantera::AnyMap> {
@@ -275,7 +276,9 @@ YAML::Emitter& operator<<(YAML::Emitter& out, const AnyMap& rhs)
             }
 
             if (foundType) {
-                if (width + name.size() + valueStr.size() + 4 > 87) {
+                // Check if this item will fit on the current line, including spaces
+                // for delimiters and whitespace. If not, wrap to the next line.
+                if (width + name.size() + valueStr.size() + 4 > max_line_length) {
                     out << YAML::Newline;
                     width = 15;
                 }
@@ -301,6 +304,10 @@ YAML::Emitter& operator<<(YAML::Emitter& out, const AnyMap& rhs)
     return out;
 }
 
+//! Write a vector in YAML "flow" style, wrapping lines to avoid exceeding the
+//! preferred maximum line length (set by `max_line_length`). Specialized for
+//! `vector<double>` to be able to use the custom `formatDouble` function with
+//! a given precision.
 void emitFlowVector(YAML::Emitter& out, const vector<double>& v, long int precision)
 {
     out << YAML::Flow;
@@ -308,16 +315,19 @@ void emitFlowVector(YAML::Emitter& out, const vector<double>& v, long int precis
     size_t width = 15; // wild guess, but no better value is available
     for (auto& x : v) {
         string xstr = formatDouble(x, precision);
-        if (width + xstr.size() > 87) {
+        // Wrap to the next line if this item would exceed the target line length
+        if (width + xstr.size() > max_line_length) {
             out << YAML::Newline;
             width = 15;
         }
         out << xstr;
-        width += xstr.size() + 2;
+        width += xstr.size() + 2; // Update width including comma and space
     }
     out << YAML::EndSeq;
 }
 
+//! Write a vector in YAML "flow" style, wrapping lines to avoid exceeding the
+//! preferred maximum line length (set by `max_line_length`).
 template <typename T>
 void emitFlowVector(YAML::Emitter& out, const vector<T>& v)
 {
@@ -326,7 +336,8 @@ void emitFlowVector(YAML::Emitter& out, const vector<T>& v)
     size_t width = 15; // wild guess, but no better value is available
     for (const auto& x : v) {
         string xstr = fmt::format("{}", x);
-        if (width + xstr.size() > 87) {
+        // Wrap to the next line if this item would exceed the target line length
+        if (width + xstr.size() > max_line_length) {
             out << YAML::Newline;
             width = 15;
         }
--- a/src/base/AnyMap.cpp
+++ b/src/base/AnyMap.cpp
@@ -1657,6 +1657,14 @@ bool AnyMap::addOrderingRules(const string& objectType,
     return true;
 }
 
+void AnyMap::clearCachedFile(const std::string& filename)
+{
+    std::string fullName = findInputFile(filename);
+    if (s_cache.count(fullName)) {
+        s_cache.erase(fullName);
+    }
+}
+
 AnyMap AnyMap::fromYamlString(const std::string& yaml) {
     AnyMap amap;
     try {
--- a/src/base/AnyMap.cpp
+++ b/src/base/AnyMap.cpp
@@ -275,7 +275,7 @@ YAML::Emitter& operator<<(YAML::Emitter& out, const AnyMap& rhs)
             }
 
             if (foundType) {
-                if (width + name.size() + valueStr.size() + 4 > 79) {
+                if (width + name.size() + valueStr.size() + 4 > 87) {
                     out << YAML::Newline;
                     width = 15;
                 }
@@ -308,7 +308,7 @@ void emitFlowVector(YAML::Emitter& out, const vector<double>& v, long int precis
     size_t width = 15; // wild guess, but no better value is available
     for (auto& x : v) {
         string xstr = formatDouble(x, precision);
-        if (width + xstr.size() > 79) {
+        if (width + xstr.size() > 87) {
             out << YAML::Newline;
             width = 15;
         }
@@ -326,7 +326,7 @@ void emitFlowVector(YAML::Emitter& out, const vector<T>& v)
     size_t width = 15; // wild guess, but no better value is available
     for (const auto& x : v) {
         string xstr = fmt::format("{}", x);
-        if (width + xstr.size() > 79) {
+        if (width + xstr.size() > 87) {
             out << YAML::Newline;
             width = 15;
         }
--- a/src/oneD/StFlow.cpp
+++ b/src/oneD/StFlow.cpp
@@ -850,15 +850,6 @@ AnyMap StFlow::serialize(const double* soln) const
     AnyMap state = Domain1D::serialize(soln);
     state["type"] = flowType();
     state["pressure"] = m_press;
-    state["grid"] = m_z;
-
-    vector_fp data(nPoints());
-    for (size_t i = 0; i < nComponents(); i++) {
-        for (size_t j = 0; j < nPoints(); j++) {
-            data[j] = soln[index(i,j)];
-        }
-        state[componentName(i)] = data;
-    }
 
     state["phase"]["name"] = m_thermo->name();
     AnyValue source = m_thermo->input().getMetadata("filename");
@@ -900,6 +891,15 @@ AnyMap StFlow::serialize(const double* soln) const
     if (m_zfixed != Undef) {
         state["fixed-point"]["location"] = m_zfixed;
         state["fixed-point"]["temperature"] = m_tfixed;
+    }
+
+    state["grid"] = m_z;
+    vector_fp data(nPoints());
+    for (size_t i = 0; i < nComponents(); i++) {
+        for (size_t j = 0; j < nPoints(); j++) {
+            data[j] = soln[index(i,j)];
+        }
+        state[componentName(i)] = data;
     }
 
     return state;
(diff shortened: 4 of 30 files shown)


ischoegl:
@speth ... thanks for taking this on. Without looking at details, I believe this is a very workable approach to moving away from XML. Do you anticipate creating an equivalent to `SolutionArray` (currently Python API only) at the Cantera C++ core eventually? (if yes, we may be able to move HDF to C++, and offer this as one of the input/output formats it for all interfaces). *PS: The HDF5 structure should be writable by a C++ library; and I am absolutely not advocating for tackling this without a C++ `SolutionArray` equivalent.*

speth:
While I'd love to have the ability to write HDF5 output files directly, my previous experience with using the HDF5 C++ library was that it was a beastly dependency to add to a project, especially when trying to build on Windows where there isn't (or at least wasn't) an easy way to install the library and headers. That was the main driver for choosing to base this on the already-available YAML input/output capabilities, even if HDF5 is better suited for this mostly numeric array-centric data.

If you want to start an enhancement discussion to talk about a C++ equivalent to the Python `SolutionArray` class, I'd be happy to share my thoughts on it.

ischoegl:
> If you want to start an enhancement discussion to talk about a C++ equivalent to the Python `SolutionArray` class, I'd be happy to share my thoughts on it.

Not sure that this is on my own short-term horizon, but I believe it may be a good new feature for 3.0. It's a substantial project.

ischoegl:
@speth ... I think it still makes sense to start the discussion now (Cantera/enhancements#119). 

To me, the ability to export to and import from YAML would be parallel to existing CSV and HDF5 (with the latter only supported for the Python API). In other words, it would be neat if we could import the YAML format from the Python API also (which presumably would be a simple pass-through to C++). You may already have that in the works ...

ischoegl:
Leaving a comment about HDF vs YAML before looking at details. 

The output produced by `save` for the `adiabatic_flame.py` example for the proposed YAML export is
[adiabatic_flame.txt](https://github.com/Cantera/cantera/files/7356233/adiabatic_flame.txt)

The same case output generated by `write_hdf` has the following structure
```
$ h5dump -n 1 adiabatic_flame_.h5 
HDF5 "adiabatic_flame_.h5" {
FILE_CONTENTS {
 group      /
 group      /group0
 attribute  /group0/Sim1D_type
 attribute  /group0/cantera_version
 attribute  /group0/curve
 attribute  /group0/date
 attribute  /group0/energy_enabled
 attribute  /group0/fixed_temperature
 attribute  /group0/git_commit
 attribute  /group0/max_grid_points
 attribute  /group0/max_time_step_count
 attribute  /group0/prune
 attribute  /group0/radiation_enabled
 attribute  /group0/ratio
 attribute  /group0/slope
 attribute  /group0/soret_enabled
 attribute  /group0/transport_model
 group      /group0/flame
 attribute  /group0/flame/Domain1D_type
 attribute  /group0/flame/emissivity_left
 attribute  /group0/flame/emissivity_right
 attribute  /group0/flame/name
 attribute  /group0/flame/steady_abstol
 attribute  /group0/flame/steady_reltol
 attribute  /group0/flame/transient_abstol
 attribute  /group0/flame/transient_reltol
 dataset    /group0/flame/T
 dataset    /group0/flame/X
 dataset    /group0/flame/density
 dataset    /group0/flame/grid
 group      /group0/flame/phase
 attribute  /group0/flame/phase/name
 attribute  /group0/flame/phase/source
 dataset    /group0/flame/velocity
 group      /group0/products
 attribute  /group0/products/Domain1D_type
 attribute  /group0/products/name
 dataset    /group0/products/T
 dataset    /group0/products/X
 dataset    /group0/products/density
 group      /group0/products/phase
 attribute  /group0/products/phase/name
 attribute  /group0/products/phase/source
 group      /group0/reactants
 attribute  /group0/reactants/Domain1D_type
 attribute  /group0/reactants/name
 dataset    /group0/reactants/T
 dataset    /group0/reactants/X
 dataset    /group0/reactants/density
 group      /group0/reactants/phase
 attribute  /group0/reactants/phase/name
 attribute  /group0/reactants/phase/source
 dataset    /group0/reactants/velocity
 }
}
```

samples/matlab/catcomb.m
```diff
@@ -125,10 +125,11 @@
...
 % set the initial profiles.
-setProfile(sim1D, 2, {'u', 'V', 'T'}, [0.0            1.0       % z/zmax
-                                       0.06           0.0       % u
-                                       0.0            0.0       % V
-                                       tinlet         tsurf]);  % T
+setProfile(sim1D, 2, {'velocity', 'spread_rate', 'T'}, ...
+           [0.0            1.0       % z/zmax
+            0.06           0.0       % u
+            0.0            0.0       % V
+            tinlet         tsurf]);  % T
```
ischoegl: There are some remaining 'u' and 'V' in the comments

samples/matlab/catcomb.m
```diff
@@ -177,7 +178,7 @@
 sim1D
 
 % save the solution
-saveSoln(sim1D,'catcomb.xml','energy',['solution with energy equation']);
+saveSoln(sim1D,'catcomb.yaml','energy',['solution with energy equation']);
```
ischoegl, line 181: As the line is changed, some after-comma white-spaces should be added (here and elsewhere)

ischoegl (commented):
@speth ... thank you for taking on the replacement of XML export. While this is not meant to be a full review, I commented on a couple of minor things I noticed. Beyond, there are a couple of Python examples left where XML is not replaced (and probably should be).

Regarding the PR itself, I don't have any concerns about the implementation per se, but I think it makes sense to discuss the long-term fate of storage structures for YAML. Cantera 2.5.1 introduced HDF output for Python, which is largely centered on `SolutionArray`, and extends this philosophy to various 1D objects (the way it was conceived centers on the user-facing Python API). The YAML structure here is - as far as I am aware - mostly a translation of the legacy XML structure that focuses on the implementation, and not the user interface. 

As mentioned earlier, I hope to see both `SolutionArray` and (hopefully) HDF support pushed into the C++ core eventually (perhaps for Cantera 3.0?). That way, we can offer users a more convenient 'save' operation for reactor simulations, and have full array support in all languages. Specifically for 1D simulations, I believe more often than not users are interested in post-processing of results when loading an old solution. From that viewpoint, a direct access as `SolutionArray` would be convenient. So designing the YAML structure with `SolutionArray` import/export in mind would make some sense.

While I don't think that we necessarily have to have exact equivalence, here is what I noticed when comparing the two structures: 

* the HDF format is a lot flatter, and uses some domain names directly rather than storing them as the `id` field. I believe that it would make sense to do the same for YAML: if the YAML file were to be imported directly using a generic YAML parser, you can access what belongs to an object by its name, rather than pick from a list after checking the `id` field. 
* I do not think that the `domains` level is necessary.
* `soret_enabled`, `radiation_enabled` and `emissivity_left/right` appear to be missing from YAML
* The location of some parameters is consistent with C++ for YAML, whereas the HDF implementation follows the Python interface. I do not have an opinion here as you can argue both ways (C++ implementation vs user interface, and a revision of HDF *could* be done)
* Storing by species name makes sense for YAML as it is a human-readable format (HDF uses `X` for efficiency, and isn't meant to be read) 
* One thing I tried to ensure for HDF is that the heritage of the `Solution` is stored (see `phase`) entry; the rationale here was that a solution can only be restored if you know what mechanism was used in the first place.

PS: out of curiosity ... what are the long-term plans to support import of `xml` results after Cantera 2.6? (Drop support or use converter script?)

samples/matlab/catcomb.m
bryanwweber: I believe the typical Matlab style (following their documentation) is not to have spaces after commas.

ischoegl: 🤷‍♂️

ischoegl: We do not have anything about that in our `CONTRIBUTING.md`, but the code style of surrounding code implies having them (it is also imho more readable). White space isn’t required for C++ either, where it is nevertheless enforced at the moment.

bryanwweber: Imo we should follow the idiomatic style of the language. In Python the spaces are likewise not required syntactically but are recommended style by PEP8. I'm not aware of a style guide for Matlab (I looked not that long ago), so I think the closest thing would be the docs.

I wouldn't put too much stock in what's not in `CONTRUBUTING.md`, we have not really discussed style things in the past, and when we have, we haven't been good about recording them.

ischoegl: I guess we should allow for single quotes in Python again, then :wink: … it’s a really minor thing that I won’t lose sleep over either way.

bryanwweber: ~Sometimes~ Most times I can't help [bikeshedding](https://en.m.wiktionary.org/wiki/bikeshedding) 🤣

ischoegl: Yup. 7 comments and counting. I’m not immune to that either!

speth: The rest of `catcomb.m` seems to be pretty irregular about spaces after commas. Given that the current Matlab toolbox's days are numbered (pending https://github.com/Cantera/enhancements/issues/32), I'm going to suggest just leaving this.

speth:
> @speth ... thank you for taking on the replacement of XML export. While this is not meant to be a full review, I commented on a couple of minor things I noticed. Beyond, there are a couple of Python examples left where XML is not replaced (and probably should be).

Good catch -- I had this on my to-do list for this PR, but it completely slipped my mind.
 
> * the HDF format is a lot flatter, and uses some domain names directly rather than storing them as the `id` field. I believe that it would make sense to do the same for YAML: if the YAML file were to be imported directly using a generic YAML parser, you can access what belongs to an object by its name, rather than pick from a list after checking the `id` field.

I agree that taking some cues from the Python HDF5 format for 1D flames is a good thought, though I think it's probably fine for these two formats to have some structural differences. I've updated the location of the domains so that they are stored in the top-level mapping, using the "id" as the key.

> * I do not think that the `domains` level is necessary.

Eliminated, per above.

> * `soret_enabled`, `radiation_enabled` and `emissivity_left/right` appear to be missing from YAML

Good catch. These were never included in the XML format, so I missed them while adapting the existing serialization code.

> * The location of some parameters is consistent with C++ for YAML, whereas the HDF implementation follows the Python interface. I do not have an opinion here as you can argue both ways (C++ implementation vs user interface, and a revision of HDF _could_ be done)

Can you be more specific? I think the grouping of the grid-related properties (`curve`, `prune`, etc.) into a subgroup makes sense, at least for the YAML format. Likewise for the solver tolerances.

One other difference is that YAML output puts the grid refinement criteria and the `energy_enabled` etc. fields with the "flow" domain, as they really belong to that domain. In theory you could have multiple flow domains, though we currently have no examples of this and the Python `FlameBase` class where `write_hdf` is implemented is built around the assumption of a single flow domain.

> * One thing I tried to ensure for HDF is that the heritage of the `Solution` is stored (see `phase`) entry; the rationale here was that a solution can only be restored if you know what mechanism was used in the first place.

Done.

> PS: out of curiosity ... what are the long-term plans to support import of `xml` results after Cantera 2.6? (Drop support or use converter script?)

The minimalist option is to say that Cantera 2.6 can be used convert existing XML solutions to YAML, as it has support for both. That will be viable as long as installing Cantera 2.6 in a conda environment works, which I think ought to be a fairly long time. Given that, I was leaning toward not creating a standalone converter script.

ischoegl:
Attaching the updated YAML output (generated by `adiabatic_flame.py`):
[adiabatic_flame.txt](https://github.com/Cantera/cantera/files/7376470/adiabatic_flame.txt)

ischoegl:
@speth ... thank you for taking the time to consider my observations and also to implement most of them. 

I really like the approach of letting `AnyMap` do all the data collection, and writing YAML from there. If we ever get around to moving HDF to the C++ core, I believe it would make sense to follow the same approach.

You are correct that locating refinement criteria with the flow domain makes sense from the C++ perspective, even if the Python API currently goes a different route. We could address this with a revision of the HDF standard _if/when_ we move it to the C++ core; same with the grouping of tolerances and refinement criteria (otherwise it's non-essential).

Regarding tolerances, I have a slight preference for a flatter structure: it's never clear to me whether to group by absolute/relative tolerance or transient/steady solver, so I'd prefer to see something that is close to the Python/HDF API, i.e.
```yaml
    tolerances:
      steady_abstol: 1.0e-09
      steady_reltol: 1.0e-04
      transient_abstol: 9.999999999999999e-12 # <-- I know this formatting issue is a current limitation
      transient_reltol: 1.0e-04
```

Using uppercase `Soret_enabled` looks like an outlier as other fields are all lowercase (there is a clash of two conventions).

Is there a way to write all settings before the numeric blocks? Currently, tolerances are written before, whereas refinement criteria are written after.

Finally, one observation is that currently some number blocks don't use horizontal space efficiently; this may be addressed by changing:
```yaml
    velocity: [0.7112641405849059, 0.7112641405822596, 0.7112641404451637,
    0.7112641377426332, 0.7112641104415781, 0.7112637339406135,
    0.7112617435453226, 0.7112560064702751, 0.7112322153842345,
    ...
    4.055246182368149, 4.055246182377726]
```
to 
```yaml
    velocity: [
      0.7112641405849059, 0.7112641405822596, 0.7112641404451637, 0.7112641377426332, 
      0.7112641104415781, 0.7112637339406135, 0.7112617435453226, 0.7112560064702751, 
      0.7112322153842345, 0.7111349509095231, 0.7109781680863129, 0.7106086071388092,
      ...
      4.055246182368149, 4.055246182377726]
```
(which in this case still fits 88 characters; I'd even consider going wider for this type of file). 

_PS: this is what 102 characters would allow for (current output only has 2 columns)_
```yaml
    OH: [
      -2.014801730705695e-17, -4.629141939774994e-19, -4.813801146169683e-20, 1.771106802152406e-20,
      2.545810916390064e-20, 1.623461193735384e-18, 1.841329907298661e-17, 1.31188752889368e-16,
      ...
      4.556290451629765e-04, 4.556290451629765e-04]
```

_PPS: the HDF writer suppresses `eField` and `lambda` where it is not applicable._

speth:
> Regarding tolerances, I have a slight preference for a flatter structure: it's never clear to me whether to group by absolute/relative tolerance or transient/steady solver, so I'd prefer to see something that is close to the Python/HDF API, i.e.
> 
> ```yaml
>     tolerances:
>       steady_abstol: 1.0e-09
>       steady_reltol: 1.0e-04
>       transient_abstol: 9.999999999999999e-12 # <-- I know this formatting issue is a current limitation
>       transient_reltol: 1.0e-04
> ```

Done, though I used names like `steady-abstol` to be consistent with the naming pattern established for Cantera's YAML input files.

> Using uppercase `Soret_enabled` looks like an outlier as other fields are all lowercase (there is a clash of two conventions).

For the YAML input file format, the convention is to capitalize proper nouns, but little less, so I'd prefer to leave this as `Soret-enabled`.

> Is there a way to write all settings before the numeric blocks? Currently, tolerances are written before, whereas refinement criteria are written after.

Done.

> Finally, one observation is that currently some number blocks don't use horizontal space efficiently; this may be addressed by changing:

You're right, getting only two values per line was a little silly, so I increased the nominal line length to 88 characters globally, though this is really just an estimate because the YAML emitter object doesn't provide enough information to know how many characters in you already are when you start serializing a particular sub-object. It also doesn't indent wrapped lines the way you or I might want, so the best we can do without a lot of extra effort is something like:
```yaml
    H2O: [9.397257458934713e-10, 4.507432204414695e-08, 2.0583678723563e-06,
    7.962288661066346e-05, 8.587434349611319e-04, 7.741273165985068e-03,
    0.0196275630231541, 0.03800532491016567, 0.05808037087942033,
    0.07665600042698147]
```

> _PPS: the HDF writer suppresses `eField` and `lambda` where it is not applicable._

Resolved (and likewise `spread_rate` for freely-propagating flames).

src/base/AnyMap.cpp
```diff
@@ -275,7 +275,7 @@ YAML::Emitter& operator<<(YAML::Emitter& out, const AnyMap& rhs)
             }
 
             if (foundType) {
-                if (width + name.size() + valueStr.size() + 4 > 79) {
+                if (width + name.size() + valueStr.size() + 4 > 87) {
```
ischoegl: Would it be possible to leave some comments on what this does?

src/base/AnyMap.cpp
```diff
@@ -308,7 +308,7 @@ void emitFlowVector(YAML::Emitter& out, const vector<double>& v, long int precis
     size_t width = 15; // wild guess, but no better value is available
     for (auto& x : v) {
         string xstr = formatDouble(x, precision);
-        if (width + xstr.size() > 79) {
+        if (width + xstr.size() > 87) {
```
ischoegl: Some comments would help understand this. The internal function `emitFlowVector` appears to be undocumented (which predates this change,  but some rudimentary comments could help).

src/base/AnyMap.cpp
```diff
@@ -326,7 +326,7 @@ void emitFlowVector(YAML::Emitter& out, const vector<T>& v)
     size_t width = 15; // wild guess, but no better value is available
     for (const auto& x : v) {
         string xstr = fmt::format("{}", x);
-        if (width + xstr.size() > 79) {
+        if (width + xstr.size() > 87) {
```
ischoegl: see above

src/oneD/StFlow.cpp
```diff
@@ -846,6 +859,142 @@ XML_Node& StFlow::save(XML_Node& o, const doublereal* const sol)
     return flow;
 }
 
+AnyMap StFlow::serialize(const double* soln) const
+{
+    AnyMap state = Domain1D::serialize(soln);
+    state["type"] = flowType();
+    state["pressure"] = m_press;
+    state["grid"] = m_z;
```
ischoegl: should this (`state["grid"] = m_z;`) be moved further down (as first of the numeric blocks?)

include/cantera/oneD/Domain1D.h
```diff
@@ -340,6 +341,21 @@ class Domain1D
      */
     virtual void restore(const XML_Node& dom, doublereal* soln, int loglevel);
 
+    //! Save the state of this domain as an AnyMap
+    /*!
+     * @param soln local solution vector for this domain
+     */
+    virtual AnyMap serialize(const double* soln) const;
```
ischoegl, line 348: I have a general comment/question about this approach. Returning `AnyMap` requires full information on the object, i.e. the inclusion of `#include "cantera/base/AnyMap.h"` in the header, whereas the alternative implementation
```C++
virtual void serialize(AnyMap& dom, const double* soln) const;
```
would allow to get away with a forward declaration (I believe - the docs build may still complain), and push the `#include` into the `cpp` file. What's the impact of those differing approaches on overall performance? (As an aside: the current implementation is more intuitive.)

ischoegl (changes requested):
Overall, this implementation looks good to me. Using the `AnyMap` emitters makes a lot of sense.

While I don't think that addressing some of the remaining YAML formatting issues needs to be taken on at this moment (these weren't introduced here after all), I hope to see them addressed before the release of 2.6 (I opened issue #1128).

include/cantera/oneD/Domain1D.h
speth: Declaring a function that returns `AnyMap` doesn't require the full definition of `AnyMap` to be available. I just `#include`d the header without really thinking about it. This has now been updated to use a forward declaration.

src/oneD/StFlow.cpp
speth: I've moved this so the grid is given just before the solution components.

ischoegl (approved):
@speth ... thank you for addressing the last comments!

PR #1116, opened 2021-10-09
scons build checks wrong location for eigen

### LINKED ISSUE #1115: scons build checks wrong location for system installations for sundials and eigen
**Problem description**

**Steps to reproduce**

Example: sundials
1. Install sundials `dnf install sundials-devel`
2. Build cantera `scons build`
3. Output `INFO: Using private installation of Sundials version 5.3.`

**Behavior**

Build should utilize locally-installed sundials and eigen headers, not download and use a private installation

**System information**

- Cantera version: commit a27eb1fff413beddf6d64d013a0ceb4dc08b4139
- OS: Fedora Linux 35
- Python/MATLAB/other software versions: Python 3.10

**Additional context**
I have identified that for `Eigen/Dense` the path ought to be `eigen3/Eigen/Dense`, but I am not entirely sure where/how the search process for sundials functions in `SConstruct`. (this is where I could use a little help to preserve the current functionality)
I have started a branch to address this: https://github.com/mefuller/cantera/tree/package_rpm - build completes but test fail due to the change I made to `sundials`
N.B. I want to chase down the dependencies to draw on system installations so that I can build and package Cantera for Fedora and other systems utilizing RPM packages

mefuller:

- Update `SConstruct` to find system installations of `eigen` when under `eigen3`
- Update `include/cantera/numerics/eigen_sparse.h` and `include/cantera/numerics/eigen_dense.h` to use system file location

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1000,37 +1000,44 @@ if env['googletest'] == 'none':
     print("""INFO: Not using Googletest -- unable to run complete test suite""")
 
 # Check for Eigen and checkout submodule if needed
-if env['system_eigen'] in ('y', 'default'):
-    if conf.CheckCXXHeader('Eigen/Dense', '<>'):
-        env['system_eigen'] = True
+if env["system_eigen"] in ("y", "default"):
+    if conf.CheckCXXHeader("eigen3/Eigen/Dense", "<>"):
+        env["system_eigen"] = True
+        env["system_eigen_prefixed"] = True
         print("""INFO: Using system installation of Eigen.""")
-    elif env['system_eigen'] == 'y':
-        config_error('Expected system installation of Eigen, but it '
-                     'could not be found.')
+        eigen_include = "<eigen3/Eigen/Core>"
+    elif conf.CheckCXXHeader("Eigen/Dense", "<>"):
+        env["system_eigen"] = True
+        env["system_eigen_prefixed"] = False
+        print("""INFO: Using system installation of Eigen.""")
+        eigen_include = "<Eigen/Core>"
+    elif env["system_eigen"] == "y":
+        config_error("Expected system installation of Eigen, but it "
+                     "could not be found.")
 
-if env['system_eigen'] in ('n', 'default'):
-    env['system_eigen'] = False
+if env["system_eigen"] in ("n", "default"):
+    env["system_eigen"] = False
     print("""INFO: Using private installation of Eigen.""")
-    if not os.path.exists('ext/eigen/Eigen/Dense'):
-        if not os.path.exists('.git'):
-            config_error('Eigen is missing. Install Eigen in ext/eigen.')
+    if not os.path.exists("ext/eigen/Eigen/Dense"):
+        if not os.path.exists(".git"):
+            config_error("Eigen is missing. Install Eigen in ext/eigen.")
 
         try:
-            code = subprocess.call(['git','submodule','update','--init',
-                                    '--recursive','ext/eigen'])
+            code = subprocess.call(["git","submodule","update","--init",
+                                    "--recursive","ext/eigen"])
         except Exception:
             code = -1
         if code:
-            config_error('Eigen not found and submodule checkout failed.\n'
-                         'Try manually checking out the submodule with:\n\n'
-                         '    git submodule update --init --recursive ext/eigen\n')
+            config_error("Eigen not found and submodule checkout failed.\n"
+                         "Try manually checking out the submodule with:\n\n"
+                         "    git submodule update --init --recursive ext/eigen\n")
+    eigen_include = "'../ext/eigen/Eigen/Core'"
 
-eigen_include = '<Eigen/Core>' if env['system_eigen'] else '"../ext/eigen/Eigen/Core"'
 eigen_versions = 'QUOTE(EIGEN_WORLD_VERSION) "." QUOTE(EIGEN_MAJOR_VERSION) "." QUOTE(EIGEN_MINOR_VERSION)'
 eigen_version_source = get_expression_value([eigen_include], eigen_versions)
-retcode, eigen_lib_version = conf.TryRun(eigen_version_source, '.cpp')
-env['EIGEN_LIB_VERSION'] = eigen_lib_version.strip()
-print('INFO: Found Eigen version {}'.format(env['EIGEN_LIB_VERSION']))
+retcode, eigen_lib_version = conf.TryRun(eigen_version_source, ".cpp")
+env["EIGEN_LIB_VERSION"] = eigen_lib_version.strip()
+print("INFO: Found Eigen version {}".format(env["EIGEN_LIB_VERSION"]))
 
 # Determine which standard library to link to when using Fortran to
 # compile code that links to Cantera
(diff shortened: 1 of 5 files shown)


site_scons/buildutils.py
```diff
@@ -11,7 +11,7 @@
 from pathlib import Path
 import logging
 from typing import TYPE_CHECKING
-import collections
+from collections.abc import Mapping
```
bryanwweber: Now that you've confirmed the fix works, I think `from collections.abc import Mapping as MappingABC` would be more clear. Also need to confirm that this doesn't generate any warnings down to Python 3.6.

speth (commented):
I think this solution works well for most cases where the Eigen headers have been installed by a system package manager, but I think it complicates things in cases where a user has downloaded the Eigen tarball and extracted it themselves to a different location, where the `Eigen` directory won't necessarily be a subdirectory of a directory named `eigen3`.

I'm also not sure whether placing the `eigen3` directory as a subdirectory of a system include directory (such as `/usr/include`) is universal or not. It seems to be the case for Ubuntu/Debian, Fedora/Redhat, and conda, but I don't know about other Linux distros or package managers.

What I would suggest is having SCons check for both `Eigen/Dense` and `eigen3/Eigen/Dense` and use that to set another variable in `config.h` so you can select between them.

mefuller:
thanks @speth - I'll work on implementing that check and variable in the coming week and hopefully have a ready PR soon

mefuller:
the implementation of the switch for the location of Eigen is not as clean as I would like, but I was having some difficulty getting the compiler directive formatted properly

mefuller:
@speth @bryanwweber  I'd like to solicit comments and suggestions on how I implemented my compiler switch. I'm not really happy with how it looks, but I believe it works.

SConstruct
```diff
@@ -1003,7 +1003,16 @@ if env['googletest'] == 'none':
 if env['system_eigen'] in ('y', 'default'):
     if conf.CheckCXXHeader('Eigen/Dense', '<>'):
         env['system_eigen'] = True
+        env['system_eigen0'] = True
+        env['system_eigen3'] = False
```
speth: I think you only need one new variable, since we're ultimately only going to include Eigen from one location. I might suggest slightly different names as well -- we're always using Eigen 3, so these are a bit confusing. Maybe just `env["system_eigen_prefixed"] = False` in this case, with `CT_SYSTEM_EIGEN_PREFIXED` as the corresponding preprocessor define.

Also, for efficiency, you might want to check for `eigen3/Eigen/Dense` before plain `Eigen/Dense`, since we seem to expect the former to be more common than the latter.

SConstruct
```diff
@@ -1024,8 +1033,9 @@ if env['system_eigen'] in ('n', 'default'):
             config_error('Eigen not found and submodule checkout failed.\n'
                          'Try manually checking out the submodule with:\n\n'
                          '    git submodule update --init --recursive ext/eigen\n')
+    eigen_include = '"../ext/eigen/Eigen/Core"'
 
-eigen_include = '<Eigen/Core>' if env['system_eigen'] else '"../ext/eigen/Eigen/Core"'
+# eigen_include = '<Eigen/Core>' if env['system_eigen'] else '"../ext/eigen/Eigen/Core"'
```
speth: Delete this line rather than commenting it out.

speth (changes requested):
Thanks, Mark. I think this does the job pretty well. I had one suggestion for simplifying it a bit, as mentioned below.

One small formatting request (channeling @bryanwweber) is that in all new or changed lines of Python code, you should use double quotes instead of single quotes.

SConstruct
mefuller: @speth so I'm clearly misunderstanding something because I can't understand why the compiler now looks in the wrong place during build
```
INFO: Generating build/src/config.h.build with the following settings:
    CANTERA_SHORT_VERSION               "2.6"
    CANTERA_VERSION                     "2.6.0a3"
    CT_LEGACY_RATE_CONSTANTS            1
    CT_SUNDIALS_USE_LAPACK              0
    CT_SUNDIALS_VERSION                 57
    CT_USE_DEMANGLE                     1
    CT_USE_SYSTEM_EIGEN                 1
    CT_USE_SYSTEM_EIGEN_PREFIXED        1
    CT_USE_SYSTEM_FMT                   1
    CT_USE_SYSTEM_YAMLCPP               1
    FTN_TRAILING_UNDERSCORE             1
    LAPACK_FTN_STRING_LEN_AT_END        1
    LAPACK_FTN_TRAILING_UNDERSCORE      1
    LAPACK_NAMES_LOWERCASE              1
    CT_USE_LAPACK                       *undefined*
    DARWIN                              *undefined*
    NEEDS_GENERIC_TEMPL_STATIC_DECL     *undefined*
    SOLARIS                             *undefined*
...
In file included from include/cantera/kinetics/StoichManager.h:11,
                 from include/cantera/kinetics/Kinetics.h:14,
                 from include/cantera/kinetics/KineticsFactory.h:11,
                 from src/fortran/fct.cpp:14:
include/cantera/numerics/eigen_sparse.h:12:10: fatal error: Eigen/Dense: No such file or directory
   12 | #include <Eigen/Dense>
      |          ^~~~~~~~~~~~~
compilation terminated.
scons: *** [build/src/fortran/fct.os] Error 1
```

include/cantera/numerics/eigen_sparse.h
```diff
@@ -5,12 +5,14 @@
...
+#if CT_USE_SYSTEM_EIGEN
+#if CT_USE_SYSTEM_EIGEN_PREFIXED
+#include <eigen3/Eigen/Dense>
 #else
-#include "cantera/ext/Eigen/Sparse"
+#include <Eigen/Dense>
+#endif
+#else
+#include "cantera/ext/Eigen/Dense"
 #endif
```
speth: Once you get these multiply-nested preprocessor statements, I like to start indenting each level, e.g. as done in `src/numerics/CVodesIntegrator.cpp`.

SConstruct
speth: You need to add a line to `include/cantera/base/config.h.in` for the new preprocessor variable.

mefuller:
@speth Thanks again for all your help.
This PR should now be all set

SConstruct
```diff
@@ -1001,9 +1001,16 @@ if env['googletest'] == 'none':
 
 # Check for Eigen and checkout submodule if needed
 if env['system_eigen'] in ('y', 'default'):
-    if conf.CheckCXXHeader('Eigen/Dense', '<>'):
+    if conf.CheckCXXHeader('eigen3/Eigen/Dense', '<>'):
```
bryanwweber: As @speth mentioned, please use double quotes (`"` instead of `'`) in all Python code. SCons scripts count too 😊

mefuller: Done

bryanwweber: Hi @mefuller, I'm sorry I wasn't clear 😕 I only meant for the lines you were already changing, not everywhere in the file. Changes across so many lines end up causing lots of merge conflicts for other folks with changes in this file on their branches.

mefuller: @bryanwweber ok, I made the quote mark changes now only to the"eigen" components

speth (approved):
Looks good to me. Thanks, @mefuller!

PR #1101, opened 2021-09-17
Wrap up GasKinetics

ischoegl:
**Changes proposed in this pull request**

- Introduce `BlowersMaselReaction3` (and associated `BlowersMaselRate`) to new framework
- Introduce `FalloffReaction3` and associated reaction rate handlers for pre-existing `Falloff`, `Troe`, `SRI`, and `Tsang` 
- Introduce a new `Arrhenius.h` header that collects reaction rates for `Arrhenius3`, `BlowersMasel3` and potential future additions (e.g. `ElectronTemperature` in #1099) in analogy to `Falloff.h`.
- Simplify handling of reaction rate units (new framework stores `Units` with rate objects)
- Make shared `ReactionData` more flexible by providing `Kinetics` references to `update` methods
- Clarify interaction with `ThirdBody` and `ThirdbodyCalc`
- `ChemicallyActivatedReaction` are treated as a special case of `FalloffReaction3`

New `FalloffReaction3` and associated rate objects bypass `FalloffFactory.h` and `FalloffMgr`, which at this point are only used by legacy implementations.

**If applicable, fill in the issue number this pull request is fixing**

Addresses Cantera/enhancements#87

**If applicable, provide an example illustrating new features this pull request is introducing**

A simple benchmark for YAML (new framework) vs XML (legacy objects) implies a slight overall speedup:
```
In [4]: %run custom_reactions.py
Average time of 100 simulation runs for 'gri30.yaml' (CH4)
- New framework (YAML): 157.51 ms (T_final=2736.60)
- One Python reaction: 179.26 ms (T_final=2736.60) ... +13.81%
- Old framework (XML): 161.23 ms (T_final=2736.60) ... +2.36%
```
(for comparison, the current `main` shows YAML tests about 1% faster)

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

PR #1120, opened 2021-10-17
Fix logic error in CanteraTest.assertNear

speth:

This fixes an issue I found while trying to explore the changes in #1101 where I realized that some problems weren't getting caught by the tests even though they were clearly covered by the test cases.

- Fix logic error in `assertNear` and `assertArrayNear` functions that caused `inf` and `nan` inputs to unexpectedly pass
- Modify tests that were failing as a result of this

Plus two more minor updates:

- Fix exception specification of `Transport::CKMode` in Cython module
- Fail after the first problem adding a reaction when compiling with optimizations disabled

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/test/utilities.py
+++ b/interfaces/cython/cantera/test/utilities.py
@@ -70,7 +70,7 @@ class CanteraTest(unittest.TestCase):
 
     def assertNear(self, a, b, rtol=1e-8, atol=1e-12, msg=None):
         cmp = 2 * abs(a - b)/(abs(a) + abs(b) + 2 * atol / rtol)
-        if cmp > rtol:
+        if not cmp < rtol:
             message = ('AssertNear: %.14g - %.14g = %.14g\n' % (a, b, a-b) +
                        'Relative error of %10e exceeds rtol = %10e' % (cmp, rtol))
             if msg:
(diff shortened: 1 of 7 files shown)


ischoegl (approved):
Looks good - I noticed the `NaN` passing as well and changed some of the tests in #1101 for that reason; thanks for tracking the root cause down! Also, noting the chained setter issue: this should be addressed once I get to #1051 at the latest.

interfaces/cython/cantera/test/utilities.py
```diff
@@ -70,7 +70,7 @@ def tearDownClass(cls):
 
     def assertNear(self, a, b, rtol=1e-8, atol=1e-12, msg=None):
         cmp = 2 * abs(a - b)/(abs(a) + abs(b) + 2 * atol / rtol)
-        if cmp > rtol:
+        if not cmp < rtol:
```
bryanwweber, line 73: I can never remember the priority order of `not` versus comparisons. I think it'd make sense to group the comparison for clarity, although it's not required. Also if you can add a comment here as to why the comparison is done this way, I think that'd help too. The previous version is more intuitive.

bryanwweber (commented):
Thanks @speth. Just a small comment for clarity.

interfaces/cython/cantera/test/utilities.py
ischoegl: Oops - should have waited longer before merging. I thought the logic made sense. I can add the comment in #1101.

PR #1127, opened 2021-10-22
cti2yaml: fix issues with composition strings and kinetics model of "None"

### LINKED ISSUE #1009: cti2yaml issues with composition strings and kinetics model of "None"
**Problem description**

While trying to convert a mechanism posted in the [Users' Group](https://groups.google.com/g/cantera-users/c/NIHKk7xd78k) from CTI to YAML, I ran into a couple of issues where `cti2yaml` has trouble with syntax that is handled by `ctml_writer`.

**Steps to reproduce**

Run `cti2yaml` on [Mayur-ISSP.cti.txt](https://github.com/Cantera/cantera/files/6281583/Mayur-ISSP.cti.txt). The attached file is modified from what is posted in the Users' Group to replace the use of the now-removed `incompressible_solid` phase type (Se the related issue #1008).

**Behavior**

The first error is related to setting the kinetics model of the `ideal_gas` phases in the input file, and gives the partial traceback:
```
  File ".../cantera/cti2yaml.py", line 1108, in to_yaml
    node.get_yaml(out)
  File ".../cantera/cti2yaml.py", line 1147, in get_yaml
    out['kinetics'] = _newNames[self.kinetics]
KeyError: 'None'
```

which is associated with the phase definition in the input file:
```python
ideal_gas(
	name = "gas_anode",
	elements = " O H C N",
	species = "N2 C3H4O3 C2H4 O2 CO2 H2O H2",
	initial_state = state(mole_fractions = "N2: 0.99999993, C3H4O3:1.0e-8, C2H4:1.0e-8, O2:1.0e-8, CO2:1.0e-8, H2O:1.0e-8, H2:1.0e-8"),
	reactions = "None",
	kinetics = "None",
	transport = "None"
)
```

Modifying these to specify `kinetics = None` instead works, but is different from what `ctml_writer` requires, which is the string rather than the built-in `None` object. After that, the next error is:
```
  File ".../cantera/cti2yaml.py", line 991, in to_yaml
    out['X'] = FlowMap(get_composition(node.X).items())
  File ".../cantera/cti2yaml.py", line 219, in get_composition
    d[b[0]] = float(b[1])
ValueError: could not convert string to float: ''
```

which appears to caused by not being able to handle the space after the first colon in the composition string `"N2: 0.99999993, C3H4O3:1.0e-8, C2H4:1.0e-8, O2:1.0e-8, CO2:1.0e-8, H2O:1.0e-8, H2:1.0e-8"`.

After fixing _that_, the mechanism finally converts, with several warnings:
```
WARNING: Reaction specification 'None' not supported
WARNING: Reaction specification 'None' not supported
WARNING: Reaction specification 'rxn-SEI-comp' not supported
```

The first two warnings are basically harmless, since they're for phases with no corresponding kinetics model, but the last one corresponds to an error in generated phase definition for `interface_SEIformation` that prevents it from being imported using the YAML input file. Instead of creating a reactions section containing just the reaction with ID `rxn-SEI-comp`, that reaction is just placed in the main `reactions` section.

**System information**

- Cantera version: 2.6.0a1 (b42989f36)
- OS: Ubuntu 20.04
- Python 3.8.3

**Attachments**

[Mayur-ISSP.cti.txt](https://github.com/Cantera/cantera/files/6281583/Mayur-ISSP.cti.txt)

**Additional context**

ischoegl:

- Handle `transport`/`kinetics` models specified as `None` consistently
- Allow for spaces in CTI composition definition (e.g. `N2: 0.9` rather than `N2:0.9`)
- Allow for specification of exact reaction id's

Closes #1009

see conversion of `Mayur-ISSP.cti` atttached to #1009. The result after applying this fix is [Mayur-ISSP.yaml.txt](https://github.com/Cantera/cantera/files/7401901/Mayur-ISSP.yaml.txt)

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/cti2yaml.py
+++ b/interfaces/cython/cantera/cti2yaml.py
@@ -1087,15 +1087,11 @@ class phase:
                 rnum = 'declared-species'
             if rnum != 'none':
                 self.reactions.append([datasrc, rnum])
-            if rnum in ('all', 'declared-species', 'none'):
+            if rnum.lower() in ("all", "declared-species", "none"):
                 continue
-            if '*' in rnum:
-                if datasrc != 'reactions':
-                    _printerr("WARNING: Reaction id-pattern matching from remote"
-                        " files not supported ({}: {})".format(datasrc, rnum))
-            else:
-                _printerr("WARNING: Reaction specification"
-                          " '{}' not supported".format(rnum))
+            if datasrc != "reactions":
+                _printerr("WARNING: Reaction id-pattern matching from remote"
+                    " files not supported ({}: {})".format(datasrc, rnum))
 
         self.initial_state = initial_state
 
@@ -1129,19 +1125,31 @@ class phase:
         # Convert reaction pattern matching to use of multiple reaction sections
         for i in range(len(self.reactions)):
             spec = self.reactions[i][1]
-            name = self.name + '-reactions'
-            if '*' in spec and name not in _reactions:
-                pattern = re.compile(spec.replace('*', '.*'))
-                misses = []
-                hits = []
-                for reaction in _reactions['reactions']:
-                    if pattern.match(reaction.id):
-                        hits.append(reaction)
-                    else:
-                        misses.append(reaction)
-                _reactions[name] = hits
-                _reactions['reactions'] = misses
-                self.reactions[i] = [name, 'all']
+            name = self.name + "-reactions"
+
+            if name in _reactions:
+                continue
+            if spec.lower() in ("all", "declared-species", "none"):
+                continue
+
+            pattern = None
+            if "*" in spec:
+                pattern = re.compile(spec.replace("*", ".*"))
+
+            misses = []
+            hits = []
+            for reaction in _reactions['reactions']:
+                if reaction.id == spec:
+                    # exact match (single reaction is specified)
+                    hits.append(reaction)
+                elif pattern and pattern.match(reaction.id):
+                    # regex match (wildcard is specified)
+                    hits.append(reaction)
+                else:
+                    misses.append(reaction)
+            _reactions[name] = hits
+            _reactions["reactions"] = misses
+            self.reactions[i] = [name, "all"]
 
         if self.kinetics and self.reactions:
             out['kinetics'] = _newNames[self.kinetics]
--- a/interfaces/cython/cantera/cti2yaml.py
+++ b/interfaces/cython/cantera/cti2yaml.py
@@ -208,7 +208,7 @@ def units(length = '', quantity = '', mass = '', time = '',
 
 def get_composition(atoms):
     if isinstance(atoms, dict): return atoms
-    a = atoms.replace(',',' ')
+    a = atoms.replace(",", " ").replace(": ", ":")
     toks = a.split()
     d = OrderedDict()
     for t in toks:
--- a/interfaces/cython/cantera/cti2yaml.py
+++ b/interfaces/cython/cantera/cti2yaml.py
@@ -305,7 +305,7 @@ class species:
         else:
             self.thermo = const_cp()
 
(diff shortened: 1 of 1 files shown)


ischoegl:
@speth / @bryanwweber ... I dropped the 'fix' for #1132 from here (per @speth's comment this will resolve itself).

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1087,15 +1087,11 @@ def __init__(self, name='', elements='', species='', note='',
...
-            if rnum in ('all', 'declared-species', 'none'):
+            if rnum.lower() in ("all", "declared-species", "none"):
                 continue
-            if '*' in rnum:
-                if datasrc != 'reactions':
-                    _printerr("WARNING: Reaction id-pattern matching from remote"
-                        " files not supported ({}: {})".format(datasrc, rnum))
-            else:
-                _printerr("WARNING: Reaction specification"
-                          " '{}' not supported".format(rnum))
```
speth, line 1098: Changing this set of allowed reaction specifications to be permissive means that certain cases that we know are not handled, like `n005 to n008` are passed through without raising a warning, and the resulting error when importing the mechanism is not easy to diagnose as it no longer mentions the unsupported user input.

speth (commented):
Thanks, @ischoegl. The first two changes look fine to me.

I'm not sure about the change to "allow for specification of exact reaction id's", though. If I'm reading this correctly, it only supports the case where a single reaction id is given (which I don't think is frequently used), correct? As currently implemented, this has the side-effect of obscuring the warning that was in place to let users know that `cti2yaml` does not support the range notation that CTI used (which I think was also rarely used).

interfaces/cython/cantera/cti2yaml.py
ischoegl: > If I'm reading this correctly, it only supports the case where a single reaction id is given

this is correct. I didn’t see a way around replacing the warnings as I had assumed that correct cti conversion was most important.

Could you be more specific about `n005 to n008`?

ischoegl: Actually, there's always the option to still produce a warning while leaving the same in place. ~I added this as a separate commit, but can squash if necessary. As is, the conversion works, but produces a warning, i.e.~
```
$ cti2yaml Mayur-ISSP.cti 
WARNING: Conversion of reaction specification 'rxn-SEI-comp' may not be supported
```
*Edit: I ended up changing things, so this is obsolete*

speth: Take a look at some of the examples given here: https://cantera.org/tutorials/cti/phases.html#declaring-the-reactions

Admittedly, I've never seen this construct used in the wild, which is why I didn't worry about it for `cti2yaml` in the first place. But I've also the reaction selection used to pick just a single reaction, either.

ischoegl: @speth ... thanks! I think it's possible to add the conversion of `n005 to n008` if it *ever* comes up. As it stands, the single reaction already came up, so I'd suggest to update the warning message and leave the import in place.

*~Edit: I added a separate error for anything containing ` to `, and tweaked the warning message.~*

*PS: found a better way by catching exceptions at a later point*

ischoegl:
@speth ... thank you for the comment. I believe I found a good solution by raising exceptions when things get parsed at a later point. As an example, the unimplemented `n005 to n008` input will produce the following warning:
```
$ cti2yaml test.cti 
WARNING: Unable to generate field 'test_not_implemented-reactions'
from reaction specification 'n005 to n008'
```
The specification of a single reaction works, and will not raise a confusing warning message (the `test_not_implemented` above points to the CTI phase definition that attempts to add those reactions).

speth (approved):

PR #1129, opened 2021-10-23
Read YAML using C++ newSolution in Python / access root level data

### LINKED ISSUE #1013: Serialization of root-level information in YAML

### LINKED ISSUE #1131: Improve handling of transport_model="None"
**Problem description**

In the current implementation, specifying a transport model of `"None"` (C++) or `None` (Python) prevents the instantiation of a transport manager object, with various internal pointers being set to `NULL`. At least in Python, there is a band-aid in place to make handling safe (using a `newDefaultTransportMgr`), but this sometimes fails - see example below. Also, it is not ideal to display `transport_model = ‘Transport’`, where the meaning is not clear to a user.

A safer implementation would use an _empty_ Transport object (similar to a `Kinetics` object with zero reactions added). A `Solution` object with no `Transport` created in C++ is unsafe as it contains `NULL` as the `Transport` reference. 

**Steps to reproduce**

The following should create equivalent objects (using `mock_ion.xml`, which does not include reactions):
```
In [5]: _ = ct.ThermoPhase("mock_ion.xml")

In [6]: _ = ct.Solution("mock_ion.xml", transport_model=None)
<ipython-input-6-5d233d73507d> in <module>()
----> 1 _ = ct.Solution("mock_ion.xml", transport_model="None")

/src/interfaces/cython/cantera/thermo.pyx in cantera._cantera.ThermoPhase.__init__()
    301     # The signature of this function causes warnings for Sphinx documentation
    302     def __init__(self, *args, **kwargs):
--> 303         super().__init__(*args, **kwargs)
    304         if 'source' not in kwargs:
    305             self.thermo_basis = mass_basis

/src/interfaces/cython/cantera/transport.pyx in cantera._cantera.Transport.__init__()
    175                 if not model:
    176                     model = 'None'
--> 177                 self.transport = newTransportMgr(stringify(model), self.thermo)
    178             self._transport.reset(self.transport)
    179 

CanteraError: 
CanteraError thrown by IonsFromNeutralVPSSTP::calcNeutralMoleculeMoleFractions:
neutral molecule calc error - anion
```

**Behavior**

While the former will prevent `Kinetics` and `Transport` methods from being called, the second command should result in an equivalent object with appropriate error handling *if a user attempts to access transport properties*.

**System information**

- Cantera version: 2.6
- OS: any
- Python/MATLAB/other software versions: Python 3.6.9

**Other thoughts**

* Prior to the implementation of `Solution.h`, there was no central object managing `Transport` (making this less of an issue)
* `KineticsFactory` does include an empty object, but there is no obvious way to reach it
* The issue surfaced in #1129: the same complication becomes apparent in the C++ interface

ischoegl:

- Use C++ `newSolution` to parse YAML strings and file input  
- Save root level YAML information and input source in C++
- Ensure that an empty transport manager is created in C++ `newSolution`
- Replaces previous Python implementation of YAML input
- Implementation expands existing serialization introduced with #984
- Allow for user-defined top-level YAML fields in `YamlWriter` (e.g. `description`)

Also:
- Deprecate `Species.fromYaml` in favor of `Species.from_yaml`
- Ensure that resetting transport in Python is registered at the C++ level
- ~Fix a minor glitch where LaTeX formatting in a figure title crashes example CI runs~

Closes #1013, closes #1131, supersedes #881, needed for #692

```python
In [1]: import cantera as ct
   ...: gas = ct.Solution("h2o2.yaml", transport_model=None)
   ...: gas.input_header
   ...:
Out[1]: 
{'cantera-version': '2.5.0',
 'date': 'Wed, 11 Dec 2019 16:59:04 -0500',
 'description': 'Hydrogen-Oxygen submechanism extracted from GRI-Mech 3.0.\nModified from the original to include N2.',
 'generator': 'ck2yaml',
 'input-files': ['h2o2.inp', 'gri30_tran.dat'],
 'units': {'activation-energy': 'cal/mol', 'length': 'cm', 'quantity': 'mol'}}
```
The `transport_model` is no longer `Transport` if not set, but yields
```
In [2]: gas.transport_model
Out[2]: 'None'
```
where appropriate safeguards are now also implemented in C++.

Further, it is now possible to specify top-level YAML fields for `YamlWriter`. For example in `extract_submechanism.py`
```python
gas2 = ct.Solution(name="gri30-CO-H2-submech",
                   thermo="ideal-gas", kinetics="gas", transport_model="Mix",
                   species=species, reactions=reactions)

# Save the resulting mechanism for later use
gas2.update_user_header({"description": "CO-H2 submechanism extracted from GRI-Mech 3.0"})
gas2.write_yaml("gri30-CO-H2-submech.yaml", header=True)
```
which produces
```yaml
description: CO-H2 submechanism extracted from GRI-Mech 3.0
generator: YamlWriter
cantera-version: 2.6.0a3
git-commit: ec800b863
date: Mon Oct 25 03:01:22 2021
...
```

**Checklist**


ischoegl:
@speth / @tsikes ... this PR replaces #881 (access to top-level YAML annotations), with a - hopefully - better fate. I improved the linkage of YAML input in Python and the underlying C++ so both use the C++ version of YAML input for `Solution`. I also replaced the `Transport` band-aid in Python with something that works universally. 

There is currently *one* failing test (Sundials 5.7 not getting correct `thermal_diff_coeffs` values), where I'm running out of ideas as I cannot replicate the issue on either of my two machines (may have to create a docker with the exact same environment).

_Edit: this [docker container](https://github.com/ischoegl/cantera-docker/tree/master/cantera-conda-fork) can reproduce the failure._

include/cantera/base/Solution.h
```diff
@@ -64,10 +64,22 @@ class Solution : public std::enable_shared_from_this<Solution>
...
 
+    //! Access input data associated with the species thermo definition
+    const AnyMap& input() const;
+    AnyMap& input();
+
+    //! Retrieve source used for object creation
+    const std::string source() const;
+
+    //! Overwrite source (only required if object is not created using newSolution)
+    void setSource(std::string source);
```
speth:
```suggestion
    void setSource(const std::string& source);
```

src/base/Solution.cpp
```diff
@@ -94,10 +146,99 @@ shared_ptr<Solution> newSolution(const std::string& infile,
...
+
+    return sol;
+}
+
+std::vector<std::string> getExcludes(const AnyValue& node)
+{
+    std::vector<std::string> exclude;
+    exclude.push_back("phases");
+    exclude.push_back("species");
+    exclude.push_back("reactions");
```
speth:
```suggestion
    std::vector<std::string> exclude{"phases", "species", "reactions"};
```

src/base/Solution.cpp
```diff
@@ -94,10 +146,99 @@ shared_ptr<Solution> newSolution(const std::string& infile,
...
+
+    // retrieve alternate names of reactions sections from phases entries
+    if (node.is<std::vector<AnyMap>>()) {
+
+        for (const auto& phaseNode : node.as<std::vector<AnyMap>>()) {
+
+            if (phaseNode.hasKey("reactions")) {
```
speth: This method for identifying "header" fields currently does not handle different "species" sections that may be present. I also think it would be a little strange to end up treating a section as being part of the header if it's actually just an unused reactions or species section. An alternative strategy you might consider, which I think will be both simpler and more true to the meaning of "header", would be to iterate over `rootNode.ordered()` and stop when you get to the `phases` node.

I also think the global `units` field should not be considered part of the header.

include/cantera/base/Solution.h
```diff
@@ -64,10 +64,22 @@ class Solution : public std::enable_shared_from_this<Solution>
 
     AnyMap parameters(bool withInput=false) const;
 
+    //! Access input data associated with the species thermo definition
+    const AnyMap& input() const;
+    AnyMap& input();
```
speth: This name is a bit misleading -- this `AnyMap` only contains the header information, right? And the docstring looks like copypasta.

src/base/Solution.cpp
```diff
@@ -94,10 +146,99 @@ shared_ptr<Solution> newSolution(const std::string& infile,
...
+        phases.push_back(adj->thermo().get());
+    }
+    sol->setKinetics(newKinetics(phases, phaseNode, rootNode));
+
+    // transport
+    if (transport == "") {
+        sol->setTransport(shared_ptr<Transport>(
+            newDefaultTransportMgr(sol->thermo().get())));
+    } else if (transport == "None") {
+        sol->setTransport(shared_ptr<Transport>(newTransportMgr("None")));
```
speth, line 186: Is this special case needed?

include/cantera/base/Solution.h
```diff
@@ -64,10 +64,22 @@ class Solution : public std::enable_shared_from_this<Solution>
...
+
+    //! Retrieve source used for object creation
+    const std::string source() const;
```
speth, line 72: Documentation here could be more clear, e.g. that this is typically the input file name.

include/cantera/base/YamlWriter.h
```diff
@@ -23,6 +24,9 @@ class YamlWriter
 public:
     YamlWriter();
 
+    //! Include top-level information used in YAML header block
+    void addHeader(const AnyMap& header);
```
speth: Given that this replaces any existing header, should this be `setHeader`?

src/base/YamlWriter.cpp
```diff
@@ -45,10 +49,32 @@ void YamlWriter::addPhase(shared_ptr<ThermoPhase> thermo,
...
+    output["date"].setLoc(-2, 0);
+
+    // Add remaining header information, ignoring obsolete information
+    std::vector<std::string> exclude = {
+        "description", "generator", "cantera-version", "git-commit", "date"};
+    for (const auto& item : m_header) {
+        std::string key = item.first;
+        if (find(exclude.begin(), exclude.end(), key) == exclude.end()) {
+            output[key] = item.second;
+        }
```
speth, line 77: If you use `std::set<std::string> exclude`, then the test can just be `if (!exclude.count(key))`

include/cantera/transport/TransportBase.h
```diff
@@ -751,6 +788,15 @@ class Transport
     std::weak_ptr<Solution> m_root;
 };
 
+
+//! Empty transport object used for exception handling
+class NoTransport : public Transport
+{
+    virtual std::string transportType() const {
+        return "None";
+    }
```
speth: Couldn't `Transport::transportType` just return `None`?

interfaces/cython/cantera/base.pyx
```diff
@@ -273,8 +319,14 @@ cdef class _SolutionBase:
             If `True`, user-defined fields which are not used by Cantera will
             be stripped from the output. These additional contents can also be
             controlled using the `update_user_data` and `clear_user_data` functions.
+        :param header:
+            If `True`, fields of the `input_header` will be added to the YAML header;
+            note that ``generator``, ``cantera-version``, ``git-commit`` and ``date``
+            will be replaced.
```
speth: It's a little unclear from this which versions of these fields will be replaced.

interfaces/cython/cantera/base.pyx
```diff
@@ -252,8 +282,24 @@ cdef class _SolutionBase:
...
+        """
+        Clear all saved header data, so that the data given by `input_header` or
+        `write_yaml` will only include values generated by Cantera based on the
+        current object state.
+        """
+        self.base.input().clear()
+
     def write_yaml(self, filename, phases=None, units=None, precision=None,
-                   skip_user_defined=None):
+                   skip_user_defined=None, header=False):
```
speth: What do you think about having this default to `header=True`?

interfaces/cython/cantera/base.pyx
```diff
@@ -95,7 +99,7 @@ cdef class _SolutionBase:
         The source of this object (such as a file name).
         """
         def __get__(self):
-            return self._source
+            return pystr(self.base.source())
```
speth, line 97: This means the `_source` member of the `_SolutionBase` class can be removed, right?

interfaces/cython/cantera/test/test_transport.py
```diff
@@ -99,7 +99,8 @@ def test_multiComponent(self):
         self.assertTrue(all(self.phase.thermal_diff_coeffs.flat != 0.0))
 
     def test_add_species_mix(self):
-        S = {s.name: s for s in ct.Species.listFromFile('gri30.xml')}
+        yaml = (self.cantera_data_path / "gri30.yaml").read_text()
+        S = {s.name: s for s in ct.Species.list_from_yaml(yaml, 'species')}
```
speth: Double quotes.

interfaces/cython/cantera/test/test_transport.py
```diff
@@ -122,7 +123,8 @@ def test_add_species_mix(self):
         self.assertArrayNear(gas1.mix_diff_coeffs, gas2.mix_diff_coeffs)
 
     def test_add_species_multi(self):
-        S = {s.name: s for s in ct.Species.listFromFile('gri30.xml')}
+        yaml = (self.cantera_data_path / "gri30.yaml").read_text()
+        S = {s.name: s for s in ct.Species.list_from_yaml(yaml, 'species')}
```
speth: Ditto.

interfaces/cython/cantera/base.pyx
```diff
@@ -112,44 +116,61 @@ cdef class _SolutionBase:
...
+            self._base = newSolution(phaseNode, root, cxx_transport, cxx_adjacent)
 
-        phaseNode = root[stringify("phases")].getMapWhere(stringify("name"),
-                                                          stringify(name))
+        self.base = self._base.get()
 
         # Thermo
         if isinstance(self, ThermoPhase):
-            self._thermo = newPhase(phaseNode, root)
+            self._thermo = self.base.thermo()
```
speth: If we have access to `shared_ptr<ThermoPhase>` from the `Solution` class, can we just remove `_SolutionBase._thermo` and access it as `self.base.thermo()` in the handful of places where that's necessary? And likewise for `_SolutionBase._kinetics` and `_SolutionBase._transport`?

test/transport/transportModels.cpp
```diff
@@ -50,3 +51,91 @@ TEST_F(WaterTransportTest, thermal_conductivity)
...
+    auto tr = soln_->transport();
+    ASSERT_EQ(tr->transportType(), "Mix");
+}
+
+TEST_F(DefaultTransportTest, check_scalar)
+{
+    // scalar quantities
+    auto tr = soln_->transport();
+    EXPECT_TRUE(tr->viscosity() >= 0);
+    EXPECT_TRUE(tr->thermalConductivity() >= 0);
```
speth: You can use `EXPECT_GE` here.

speth (changes requested):
Thanks, @ischoegl, this generally looks pretty good to me. I had a couple of suggestions for simplifications, as well as some other minor comments.

src/base/Solution.cpp
ischoegl: Yes. This is introduced to prevent segfaults that are currently caught with a band-aid in Python. (Or: as the Python solution predates the C++ handling … I am simply moving that exception handling to C++)

include/cantera/transport/TransportBase.h
ischoegl: In C++ it may still be best to stick with `”None”`? I may not be completely following ...

interfaces/cython/cantera/base.pyx
ischoegl: Yes. Which is probably for the better!

interfaces/cython/cantera/base.pyx
ischoegl: I think not having the shadow objects will avoid some curious behaviors.

interfaces/cython/cantera/base.pyx
ischoegl: Yes, they already are removed.

src/base/Solution.cpp
ischoegl: Makes sense.

ischoegl:
@speth ... I believe I took care of all the comments. Regarding the transport model, it's possible to replace the C++ `"None"` with a Python `None`, but the string version may be more intuitive/descriptive to users.

include/cantera/transport/TransportBase.h
speth: Sorry, GH trimmed off the relevant part of this code snippet. I wasn't questioning the string `"None"` versus the Python builtin symbol `None`. This comment is on the `NoTransport` class, and I was wondering if it would be simpler to just have `Transport::transportType` return this value rather than adding a special new class to represent the lack of a transport model.

ischoegl: :+1: ... makes sense. Adopted!

ischoegl:
@speth ... thanks for the clarification. I updated as suggested ...

include/cantera/thermo/ThermoFactory.h
```diff
@@ -107,6 +107,24 @@ inline ThermoPhase* newThermoPhase(const std::string& model)
     return ThermoFactory::factory()->create(model);
 }
 
+//! Create a new ThermoPhase instance.
+/*!
+ * @param model   String to look up the model against
+ * @returns a shared pointer to a new ThermoPhase instance matching the model string.
+ *   Throws an exception UnknownThermoPhaseModel if the string wasn't matched; or a
+ *   CanteraError if something else went wrong.
```
speth: This isn't actually true -- the `UnknownThermoPhaseModel` exception class is never used, and already deprecated. I guess this bit of erroneous documentation from `newThermoPhase` ought to be corrected.

include/cantera/thermo/ThermoFactory.h
```diff
@@ -107,6 +107,24 @@ inline ThermoPhase* newThermoPhase(const std::string& model)
...
+ */
+inline shared_ptr<ThermoPhase> newThermo(const std::string& model)
+{
+    ThermoPhase* tptr = ThermoFactory::factory()->create(model);
+    if (tptr == 0) {
+        CanteraError(
+            "newThermo", "Encountered exception while creating thermo "
+            "model '{}'.", model);
+    }
```
speth: I don't think this check is necessary, though I do see some documentation that incorrectly suggests otherwise. Any exception raised in the `create` method will just propagate through, and I don't think there's any case where it will return a null pointer.

include/cantera/transport/TransportFactory.h
```diff
@@ -96,6 +96,25 @@ class TransportFactory : public Factory<Transport>
...
+//!  Create a new Transport instance.
+/*!
+ *  @param model    name of transport model; if "default", the default
+ *                  transport model for the ThermoPhase object is created
+ *  @param thermo   the ThermoPhase object associated with the phase
+ *  @returns a Transport object for the phase
+ * @ingroup tranprops
+ */
+inline shared_ptr<Transport> newTransport(const std::string& model = "",
+                                          ThermoPhase* thermo = 0) {
```
speth: The introduction of this function might be a good time to flip the arguments here around to make more sense - the `thermo` model is required, so it should come before an optional parameter.

Also, wouldn't it make more sense for the default value for `model` to be `"default"`, rather than constructing the "no transport" model?

interfaces/cython/cantera/yamlwriter.pyx
```diff
@@ -9,6 +9,10 @@ cdef class YamlWriter:
         self._writer.reset(new CxxYamlWriter())
         self.writer = self._writer.get()
 
+    def add_header(self, _SolutionBase soln):
+        """ Include top-level information for the specified Solution object """
+        self.writer.setHeader(soln.base.header())
```
speth: I suppose the Python function name ought to be updated to match the C++ function name here.

speth (changes requested):
Thanks for the updates, @ischoegl, I think this is almost set. I just had a couple of further comments, mainly related to the new factory functions.

include/cantera/thermo/ThermoFactory.h
ischoegl: copypasta :joy:

include/cantera/thermo/ThermoFactory.h
ischoegl: tried to retain what I found elsewhere. But this makes sense.

include/cantera/transport/TransportFactory.h
ischoegl: This isn't necessarily how it is currently used, see `_cantera.pxd`:
```
cdef shared_ptr[CxxTransport] newTransport(string) except +translate_exception
cdef shared_ptr[CxxTransport] newTransport(string, CxxThermoPhase*) except +translate_exception
```
In essence, `newTransport("")` is how what was formerly `NoTransport` is instantiated. The class serves as a safe-guard against segfaults (i.e. there will always be a `Transport` model, albeit one that just handles exceptions and holds no data). This ensures that users see exceptions that point them in the right direction.

speth: Yes, being able to create an empty `Transport` object that just raises exceptions makes perfect sense. All I'm suggesting is reorganizing the input arguments, since this is a new function and it's used is still flexible. Currently, it is only ever called with both arguments, from within the Python module.

I think the signature `newTransport(ThermoPhase* thermo, const std::string& model="default")` would provide the simplest interface for the two most common use cases, e.g. creating either the transport model specified by the input file, or a value specified in the call:
```c++
newTransport(phase);
newTransport(phase, "mixture-averaged");
```
This still allows creating an "empty" `Transport` manager with
```c++
newTransport(phase, "None");
```
The rare case (currently unnecessary, as far as I am aware) where you want to create an empty `Transport` object with no attached phase is still possible, and is the only case where the syntax is slightly awkward:
```c++
newTransport(nullptr, "None");
```

ischoegl: Np. Was probably too stuck on the old
```C++
Transport* newTransportMgr(const std::string& transportModel = "",
                           ThermoPhase* thermo = 0, int loglevel = 0);
```
and the associated `TransportFactory::newTransport(const std::string& model, ThermoPhase* thermo, int log_level=0)`. 

I'm not sure about the long-term prospects of the factory service functions that return raw pointers (similar for various `ThermoPhase` and `Kinetics` service functions); once this becomes clear, it may make sense to update some of the function signatures. Not sure whether this should be part of 2.6, so those can be gone for 3.0. I certainly don't want to attach this to this PR, as it should be ready to be merged at this point.

ischoegl:
@speth ... I implemented the one remaining suggestion for `newTranport`. I believe this should be ready ...

speth (approved):

Regarding one of your last comments, I would agree that the various global factory functions for new Cantera thermo / kinetics / transport / etc. objects could use some housekeeping. At a minimum, there's both the inconsistency whether they return raw pointers, `unique_ptr` or `shared_ptr` as well as confusing variability in their names (do we need the `Mgr` suffix in `newKineticsMgr`?).

PR #1133, opened 2021-10-26
Improve YAML formatting

### LINKED ISSUE #1128: Improve YAML output formatting
**Problem description**

As observed in #1112 (and likely elsewhere), some of the human-readable output generated by YAML emitters could be improved. 

**Steps to reproduce**

The YAML output below can be produced using the new `save` method proposed for 1D objects (#1112), but the underlying implementation predates this PR (goes back to #984).

**Behavior**

There are three formatting issues documented thus far:

#### (1) Scalars / Rounding

As already pointed out in #984 (where serialization was initially introduced), there are some _"floating point rounding issues [that] result in kind of ugly values"_. For example:

```yaml
    tolerances:
      steady-abstol: 1.0e-09
      steady-reltol: 1.0e-04
      transient-abstol: 9.999999999999999e-12 # <-- should be 1.0e-11
      transient-reltol: 1.0e-04
```
I have not looked into details, but at least for scalar values there may be a way to catch these items. 

#### (2) Vectors / Efficient use of space

Number blocks don't use horizontal space efficiently; this may be addressed by changing:
```yaml
    velocity: [0.7112641405849059, 0.7112641405822596, 0.7112641404451637,
    0.7112641377426332, 0.7112641104415781, 0.7112637339406135,
    0.7112617435453226, 0.7112560064702751, 0.7112322153842345,
    ...
    4.055246182368149, 4.055246182377726]
```
to 
```yaml
    velocity: [
      0.7112641405849059, 0.7112641405822596, 0.7112641404451637, 0.7112641377426332, 
      0.7112641104415781, 0.7112637339406135, 0.7112617435453226, 0.7112560064702751, 
      0.7112322153842345, 0.7111349509095231, 0.7109781680863129, 0.7106086071388092,
      ...
      4.055246182368149, 4.055246182377726]
```
which uses space more efficiently and also improves readability from a human perspective. Beyond, there is likely an optimized number of characters per line that prevents ragged output in *most* cases.

**System information**

- Cantera version: 2.6
- OS: any
- Python/MATLAB/other software versions: any

**Additional context**

* PR #984
* PR #1112

ischoegl:
Leverage `fmt::format` to appropriately round values.

- Update rounding rules for `double` emitter
- Implement multi-line strings

Closes #1128 (together with improved formatting documentation introduced in #1112)

_(A) Rounding of floating point values_

Ensure that floating point values are properly rounded, e.g.
```
<       - {P: 10.0, A: 1.2866e+47, b: -9.0246, Ea: 2.00263761648836e+04}
<       - {P: 100.0, A: 5.9632e+56, b: -11.529, Ea: 2.64691461742217e+04}
---
>       - {P: 9.99999999999999, A: 1.2866e+47, b: -9.0246,
>       Ea: 2.00263761648836e+04}
>       - {P: 99.9999999999999, A: 5.9632e+56, b: -11.529,
>       Ea: 2.64691461742217e+04}
```

_(B) Serialization of lines spanning multiple lines (aka 'Literal' ~or folded strings~)_

Strings that contain `'\n'` are written as multiple lines, e.g. `"spam\nand\neggs"` becomes:
```yaml
literal-string: |
  spam
  and
  eggs
```
which leverages `YAML::Literal`. At the moment, there doesn't appear to be a way to skip the trailing `\n` in the equivalent string that is written (i.e. there is no `|-` rather than `|` in `yaml-cpp` for the first line as far as I can tell).

FWIW, here's a [demo](https://yaml-multiline.info/) illustrating different literal/folded multiline string versions in YAML. `yaml-cpp` appears to only support one version.

**Checklist**

### DIFF
--- a/src/base/AnyMap.cpp
+++ b/src/base/AnyMap.cpp
@@ -157,40 +157,56 @@ long int getPrecision(const Cantera::AnyValue& precisionSource)
 
 string formatDouble(double x, long int precision)
 {
-    int log10x = static_cast<int>(std::floor(std::log10(std::abs(x))));
     if (x == 0.0) {
         return "0.0";
-    } else if (log10x >= -2 && log10x <= 3) {
-        // Adjust precision to account for leading zeros or digits left of the
-        // decimal point
-        precision -= log10x;
-        string s = fmt::format("{:.{}f}", x, precision);
-        // Trim trailing zeros, keeping at least one digit to the right of
-        // the decimal point
-        size_t last = s.size() - 1;
-        for (; last != 0; last--) {
-            if (s[last] != '0' || s[last-1] == '.') {
-                break;
-            }
-        }
-        s.resize(last + 1);
-        return s;
-    } else {
-        string s = fmt::format("{:.{}e}", x, precision);
-        // Trim trailing zeros, keeping at least one digit to the right of
-        // the decimal point
-        size_t eloc = s.find('e');
-        size_t last = eloc - 1;
-        for (; last != 0; last--) {
-            if (s[last] != '0' || s[last-1] == '.') {
-                break;
-            }
-        }
-        if (last != eloc - 1) {
-            s = string(s, 0, last + 1) + string(s.begin() + eloc, s.end());
-        }
-        return s;
     }
+
+    // Build string with full precision
+    bool useExp = std::abs(x) < 1e-2 || std::abs(x) >= 1e4;
+    int log10x;
+    size_t last;
+    string s0;
+    if (useExp) {
+        s0 = fmt::format(fmt::format("{:.{}e}", x, precision));
+        last = s0.size() - 5; // last digit of significand
+    } else {
+        log10x = static_cast<int>(std::floor(std::log10(std::abs(x))));
+        s0 = fmt::format("{:.{}f}", x, precision - log10x);
+        last = s0.size() - 1; // last digit
+    }
+    if (s0[last - 2] == '0' && s0[last - 1] == '0' && s0[last] < '5') {
+        // Value ending in '00x' and should be rounded down
+    } else if (s0[last - 2] == '9' && s0[last - 1] == '9' && s0[last] > '4') {
+        // Value ending in '99y' and should be rounded up
+    } else {
+        // Value should not be rounded / do not round last digit
+        return s0;
+    }
+
+    // Remove trailing zeros
+    string s1;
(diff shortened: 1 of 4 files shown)


src/base/AnyMap.cpp
```diff
@@ -304,6 +319,40 @@ YAML::Emitter& operator<<(YAML::Emitter& out, const AnyMap& rhs)
     return out;
 }
 
+//! Write YAML strings spanning multiple lines if input includes endline '\n'
+void emitString(YAML::Emitter& out, const string str0) {
```
speth:
```suggestion
void emitString(YAML::Emitter& out, const string& str0) {
```

speth (changes requested):
Thanks, @ischoegl. I think these changes to the output formatting generally look very good. However, I did notice that there were a few cases where the output is now outputting trailing zeros in some circumstances, and I don't think this is expected based on the comments in the `formatDouble` method. A couple of examples can be seen in the test suite. In `test/work/generated-ptcombust.yaml`, I see:
```yaml
    site-density: 1.629771953878800e+13
```
and in `test/work/generated-h2o2-from-xml.yaml`, we have:
```yaml
    rate-constant: {A: 3.57e+04, b: 2.4, Ea: -1061.79321568240}
```
I'm not sure exactly how this is happening, but perhaps it's worth a few tests that this formatter is actually generating the desired output (recognizing that the lack of tests for this method is an oversight on my fault to begin with 😬).

ischoegl:
@speth ... thank you for the review!

Interesting find about the double trailing zeros. It is not entirely surprising as I was very careful to not round and inadvertently introduce error. I am now removing double '00' but also added the following explanation to `formatDouble`:
```C++
string formatDouble(double x, long int precision)
{
    // This function ensures that trailing zeros resulting from round-off error
    // are removed. Values are only rounded if at least three digits are removed,
    // or the displayed value has multiple trailing zeros.
```
This choice is partially aesthetic (e.g. long blocks of values), but also tries to stay on the safe side while catching compounded round-off error (not just trailing `001`'s and `999`'s).

I also added the requested unit test. It's somewhat tricky as the exact number representation may be impacted by the architecture, but I _believe_ what I implemented should have the desired output. One curious finding was that removing every trailing zero breaks the test suite (i.e. removing a single zero may change the actual value!). Removing double trailing zeros appears to be safe.

speth (approved):
Thank you, @ischoegl. This looks good to me.

PR #1136, opened 2021-10-30
[SCons] Implement script to extract options from SConstruct

ischoegl:

- Implement `scons2rst.py`, which extracts options from `SConstruct` via AST
- Refactor `SConstruct` in order to provide access to platform/compiler dependent defaults
- `SConstruct` options (including platform/compiler dependent defaults) are extracted and written to a ReST file

Addresses cantera/cantera-website#26

```
$ python scons2rst.py 
$ python scons2rst.py --dev SConstruct
```

An example for generated output for the development branch is [config-options.rst.txt](https://github.com/Cantera/cantera/files/7447402/config-options.rst.txt)

At the moment, platform-dependent options are described as `<platform dependent>`.

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -326,7 +326,12 @@ defaults.debug_link_flags = {"cl": "/DEBUG", "default": ""}
 defaults.no_debug_link_flags = ""
 defaults.warning_flags = {"cl": "/W3", "icc": "-Wcheck", "default": "-Wall"}
 defaults.build_pch = {"icc": False, "default": True}
-
+defaults.pch_flags = {
+    "cl": "/FIpch/system.h",
+    "gcc": "-include src/pch/system.h",
+    "clang": "-include-pch src/pch/system.h.gch",
+    "default": "",
+}
 defaults.thread_flags = {"Windows": "", "macOS": "", "default": "-pthread"}
 defaults.fs_layout = {"Windows": "compact", "default": "standard"}
 defaults.python_prefix = {"Windows": "", "default": "$prefix"}
@@ -335,7 +340,6 @@ defaults.env_vars = "PATH,LD_LIBRARY_PATH,PYTHONPATH"
 defaults.versioned_shared_library = {"mingw": False, "default": True}
 defaults.sphinx_options = "-W --keep-going"
 
-env['pch_flags'] = []
 env['openmp_flag'] = ['-fopenmp'] # used to generate sample build scripts
 
 env['using_apple_clang'] = False
@@ -350,11 +354,9 @@ if 'gcc' in env.subst('$CC') or 'gnu-cc' in env.subst('$CC'):
     if env['OS'] == 'Cygwin':
         defaults.select("Cygwin")
     defaults.select("gcc")
-    env['pch_flags'] = ['-include', 'src/pch/system.h']
 
 elif env['CC'] == 'cl': # Visual Studio
     defaults.select("cl")
-    env['pch_flags'] = ['/FIpch/system.h']
     env['openmp_flag'] = ['/openmp']
 
 elif 'icc' in env.subst('$CC'):
@@ -363,7 +365,6 @@ elif 'icc' in env.subst('$CC'):
 
 elif 'clang' in env.subst('$CC'):
     defaults.select("clang")
-    env['pch_flags'] = ['-include-pch', 'src/pch/system.h.gch']
 
 else:
     print("WARNING: Unrecognized C compiler '%s'" % env['CC'])
@@ -583,6 +584,10 @@ config_options = [
         'use_pch',
         """Use a precompiled-header to speed up compilation""",
         defaults.build_pch),
+    (
+        'pch_flags',
+        """Compiler flags when using precompiled-header.""",
+        defaults.pch_flags),
     (
         'cxx_flags',
         """Compiler flags passed to the C++ compiler only. Separate multiple
@@ -886,6 +891,7 @@ env['LINKFLAGS'] += listify(env['thread_flags'])
 env['CPPDEFINES'] = {}
 
 env['warning_flags'] = listify(env['warning_flags'])
+env["pch_flags"] = listify(env["pch_flags"])
 
 if env['optimize']:
     env['CCFLAGS'] += listify(env['optimize_flags'])
--- a/SConstruct
+++ b/SConstruct
@@ -195,8 +195,10 @@ if os.name == 'nt':
         (
             "msvc_version",
             """Version of Visual Studio to use. The default is the newest
-                installed version. Specify '12.0' for Visual Studio 2013 or '14.0'
-                for Visual Studio 2015. Windows MSVC only.""",
+               installed version. Specify '12.0' for Visual Studio 2013, '14.0' for
+               Visual Studio 2015, '14.1' ('14.1x') Visual Studio 2017, or '14.2'
+               ('14.2x') for Visual Studio 2019. For version numbers in parentheses,
+               'x' is a placeholder for a minor version number. Windows MSVC only.""",
             ""),
         EnumVariable(
             "target_arch",
@@ -216,7 +218,7 @@ if os.name == 'nt':
         EnumVariable(
             "toolchain",
             """The preferred compiler toolchain. If MSVC is not on the path but
-            'g++' is on he path, 'mingw' is used as a backup. Windows only.""",
+               'g++' is on the path, 'mingw' is used as a backup. Windows only.""",
             defaults.toolchain, ("msvc", "mingw", "intel")))
     opts.AddVariables(windows_compiler_options[-1])
     opts.Update(pickCompilerEnv)
@@ -444,7 +446,7 @@ config_options = [
         'matlab_path',
         """Path to the MATLAB install directory. This should be the directory
            containing the 'extern', 'bin', etc. subdirectories. Typical values
-           are: "C:/Program Files/MATLAB/R2011a" on Windows,
+           are: "C:\Program Files\MATLAB\R2021a" on Windows,
            "/Applications/MATLAB_R2011a.app" on macOS, or
            "/opt/MATLAB/R2011a" on Linux.""",
         '', PathVariable.PathAccept),
@@ -501,7 +503,7 @@ config_options = [
         'system_fmt',
         """Select whether to use the fmt library from a system installation
            ('y'), from a Git submodule ('n'), or to decide automatically
-           ('default').  If you do not want to use the Git submodule and fmt
+           ('default'). If you do not want to use the Git submodule and fmt
            is not installed directly into system include and library
            directories, then you will need to add those directories to
            'extra_inc_dirs' and 'extra_lib_dirs'. This installation of fmt
--- a/SConstruct
+++ b/SConstruct
@@ -64,6 +64,11 @@ if not COMMAND_LINE_TARGETS:
     logger.info(__doc__, print_level=False)
     sys.exit(0)
 
+if parse_version(SCons.__version__) < parse_version("3.0.0"):
+    logger.info("Cantera requires SCons with a minimum version of 3.0.0. Exiting.",
+                print_level=False)
+    sys.exit(0)
+
 valid_commands = ("build", "clean", "install", "uninstall",
                   "help", "msi", "samples", "sphinx", "doxygen", "dump")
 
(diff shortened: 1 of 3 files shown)


ischoegl:
Done. For output generated by the script, refer to Cantera/cantera-website#160. As this PR stands, the parsed options are uploaded together with documentation into a new `scons` folder.

ischoegl:
Noticed a couple of additional manual edits in the original `config-options-dev.rst` file, which are now also considered.

SConstruct
```diff
@@ -141,44 +141,76 @@ opts = Variables('cantera.conf')
 windows_compiler_options = []
 extraEnvArgs = {}
 
+
+class Defaults:
+    """Class enabling selection of options based on attribute dictionary entries"""
+
+    def apply(self, key="defaults"):
```
bryanwweber: I think `select` would be a better name for this function.

SConstruct
```diff
@@ -141,44 +141,76 @@ opts = Variables('cantera.conf')
...
 if os.name == 'nt':
+    defaults.apply("Windows")
+
     # On Windows, target the same architecture as the current copy of Python,
     # unless the user specified another option.
-    if '64 bit' in sys.version:
-        target_arch = 'amd64'
-    else:
-        target_arch = 'x86'
+    if '64 bit' not in sys.version:
```
```suggestion
    if "64 bit" not in sys.version:
```

SConstruct
```diff
@@ -141,44 +141,76 @@ opts = Variables('cantera.conf')
...
-        ('msvc_version',
-         """Version of Visual Studio to use. The default is the newest
-            installed version. Specify '12.0' for Visual Studio 2013 or '14.0'
-            for Visual Studio 2015.""",
-         ''),
+        (
+            "msvc_version",
+            """Version of Visual Studio to use. The default is the newest
+                installed version. Specify '12.0' for Visual Studio 2013 or '14.0'
+                for Visual Studio 2015. Windows MSVC only.""",
```
bryanwweber: Can you add the versions for 2017, 2019, and 2022 as well?

SConstruct
```diff
@@ -141,44 +141,76 @@ opts = Variables('cantera.conf')
...
 
-    windows_compiler_options.append(EnumVariable(
-        'toolchain',
-        """The preferred compiler toolchain.""",
-        defaultToolchain, ('msvc', 'mingw', 'intel')))
+    windows_compiler_options.append(
+        EnumVariable(
+            "toolchain",
+            """The preferred compiler toolchain. If MSVC is not on the path but
+            'g++' is on he path, 'mingw' is used as a backup. Windows only.""",
```
```suggestion
            'g++' is on the path, 'mingw' is used as a backup. Windows only.""",
```

Can you clarify what "backup" means in this context?

SConstruct
```diff
@@ -237,41 +269,61 @@ if 'FRAMEWORKS' not in env:
...
-defaults.optimizeCcFlags = '-O3'
-defaults.debugCcFlags = '-g'
-defaults.noDebugCcFlags = ''
-defaults.debugLinkFlags = ''
-defaults.noDebugLinkFlags = ''
-defaults.warningFlags = '-Wall'
-defaults.buildPch = False
-defaults.sphinx_options = '-W --keep-going'
+defaults.cxxFlags = {"cl": "/EHsc", "Cygwin": "-std=gnu++11", "default": "-std=c++11"}
+defaults.ccFlags = {
```
bryanwweber: If we're redoing this anyways, can you change them from camelCase to snake_case?

SConstruct
```diff
@@ -284,102 +336,80 @@ if env['OS'] == 'Darwin':
...
 # *** Read user-configurable options ***
 # **************************************
 
 config_options = [
     PathVariable(
-        'prefix',
-        'Set this to the directory where Cantera should be installed.',
+        "prefix",
+        """Set this to the directory where Cantera should be installed. On Windows
+           systems, '$ProgramFiles' typically refers to "C:\Program Files".""",
```
bryanwweber, line 401: Can you clarify what "typically" means here? If a user doesn't know what `'$ProgramFiles'` is or what it does, how should they interpret this help?

SConstruct
```diff
@@ -284,102 +336,80 @@ if env['OS'] == 'Darwin':
...
+           Some distributions (for example, Fedora/RHEL) use 'lib64' instead of 'lib'
+           on 64-bit systems or could use some other library directory name instead of
+           'lib' depends on architecture and profile (for example, Gentoo 'libx32' on
+           x32 profile). If the user didn't set the 'libdirname' configuration
+           variable, set it to the default value 'lib'""",
         'lib', PathVariable.PathAccept),
     EnumVariable(
         'python_package',
-        """If you plan to work in Python, then you need the ``full`` Cantera Python
+        """If you plan to work in Python, then you need the 'full' Cantera Python
```
bryanwweber, line 414: I haven't gotten to the option processing script yet (so I don't know if this is handled), but the \`\` will put the `full` into monospace in ReST. I think that should be preferred on the web, although it's obviously not optimal for the CLI.

SConstruct
```diff
@@ -506,13 +537,13 @@ config_options = [
...
            needed if the libraries are installed in a standard location, for example,
-           ``/usr/lib``.""",
+           '/usr/lib'.""",
         '', PathVariable.PathAccept),
     EnumVariable(
         'lapack_names',
         """Set depending on whether the procedure names in the specified
            libraries are lowercase or uppercase. If you don't know, run 'nm' on
-           the library file (for example, 'nm libblas.a').""",
+           the library file (for example, "nm libblas.a").""",
```
bryanwweber, line 564: This would be OK to leave as `'` since it's representing a program/CLI thingy, and that's our convention elsewhere.

site_scons/scons2rst.py
```diff
@@ -0,0 +1,350 @@
+#!/usr/bin/env python3
+# encoding: utf-8
+
+# This file is part of Cantera. See License.txt in the top-level directory or
+# at https://cantera.org/license.txt for license and copyright information.
+
+"""
+scons2rst.py: Extract SCons configuration options from SConstruct
+"""
```
bryanwweber, line 9: IMO, this would be better living the in website as a plugin for Nikola, so that it could run each time the website is generated and we don't have to remember to download/extract/upload the zip file from the artifacts. I'm happy to help turn this into a plugin over there.

bryanwweber (changes requested):
Thanks @ischoegl Reworking the `defaults` class is certainly welcome and a nice cleanup. I do have a small concern that everything is equivalent for the compilers that we don't run on GH Actions (Cygwin, mingw, Intel), since some of that stuff broke when I updated `buildutils.py`. Is there any way we can address that?

SConstruct
ischoegl: The conversion script handles all the back tick \`\` replacements. For the regular `scons help` output the regular `'` is imho more readable. At the moment, the `regex` parser expressions exclude whitespace for single quotes, but include whitespace for double quotes (which can be tweaked)

site_scons/scons2rst.py
ischoegl: I thought of that, but I believe that wouldn't cover changes in `SConstruct` on the main repo. The converter script relies on some conventions (e.g. specific variable names and the convention that the first assignment of a new class variable to `defaults` contains information on compiler/platform dependent defaults). This PR would cover the conversion as part of CI, where a failing test would automatically point towards the issue, which isn't possible if this would be moved to `cantera-website`.

ischoegl:
@bryanwweber ... thanks for the comments!

> I do have a small concern that everything is equivalent for the compilers that we don't run on GH Actions (Cygwin, mingw, Intel), since some of that stuff broke when I updated `buildutils.py`. Is there any way we can address that?

Not sure - what I did here is to map the *current* logic of `SConstruct`, where I explicitly didn't want introduce any changes (the refactoring was mainly done so I could parse for default options). For whatever is not on GH Actions, we appear to rely on user feedback, so it's hard to cover all of that ... we may want to state that we only test for certain compilers, although others are *theoretically* covered. Longterm, I hope that we can get Cantera/enhancements#117 et al. to take care of this problem.

*PS: It is correct that this assumes that `g++` options remain the same across toolchains, which I believe is* reasonable. *There is one Cygwin override but as far as I can tell the logic isn't complex enough to warrant nested option dictionaries.*

bryanwweber:
> Not sure - what I did here is to map the _current_ logic of `SConstruct`, where I explicitly didn't want introduce any changes

Yes, I know, but in any refactoring there's a chance of introducing bugs or incidentally changing behavior, I just hope that doesn't happen here. I guess the resolution is just Cantera/enhancements#117.

> we may want to state that we only test for certain compilers, although others are _theoretically_ covered.

That's a good idea, I wonder where that should go... probably in the compiling docs on the website would be a good spot.

---

I'm going to move the larger conversation about the home for this functionality out here into the main thread, so hopefully it doesn't get collapsed by GitHub quite so quickly... I never said, thank you @ischoegl for taking on the first implementation of this! Clearly that issue has been around for almost 3 years, so it's nice to finally have something concrete to discuss.

So, on looking through the `scons2rst.py` script a little more thoroughly, I think that @speth was correct in https://github.com/Cantera/cantera-website/issues/26#issuecomment-408452199, insofar as I think it'd be better to include an option to get this output directly into SConstruct. The reason for that is SCons already knows how to parse its options and get default values, see [`site_scons/buildutils.py:help`](https://github.com/Cantera/cantera/blob/main/site_scons/buildutils.py#L696-L780).

I think it would be considerably less fragile to have a new SCons option (like `scons build` etc.) that produces the appropriate output to stdout and this can be saved by anyone to a file. I'm worried that with a separate script and using the AST, that API changes made by SCons will require rewriting the conversion script, and possibly supporting multiple versions of SCons with such a script; or, that refactoring of SConstruct by us will cause the conversion script to fail. That just increases our already overloaded capacity for maintenance.

Like `scons help`, this new option could be something near the top of SConstruct that simply calls a function in `buildutils` and then exits. The only trick would be making sure that platform-specific options get included, regardless of what platform is running `scons get-configuration`. Maybe a separate `all_options` list that updates the environment if `get-configuration` is in `COMMAND_LINE_TARGETS`.

Once that new option is added to SConstruct here, we can add a plugin to the website repo to run `scons get-options` (or whatever the name is) and save that output to the `config-options-dev.rst` (or whatever that filename is).

ischoegl:
@bryanwweber ... thanks for the additional feedback. My initial reaction is that I am not sure that I agree with that assessment. The nice thing about AST is that you know what the original expressions are that are plugged into other expressions, while leaving things in plain code. 

The alternative would be to build a mechanism that detects options for `scons get-options` directly, meaning that the construction of parameter lists would happen in `SConstruct` (imho, a bad idea), or be pushed into `buildutils`, which hides what is done from the user. 

Either way, much of the work that is already done can be leveraged - the `Defaults` object will remain the same (although perhaps moved to `buildutils`), and the object handling the ReST formatting can be copied directly over. The difference is honestly not as big as it appears ... the main price for not using AST is that more is moved out of `SConstruct`, which may be a good thing after all.

bryanwweber:
> The alternative would be to build a mechanism that detects options

If I understand your concern correctly, this is already done by SCons when `opts.AddVariables` is called. The `Variables` object stores all the options and their defaults. This is how the `help()` function works. The only trick will be to make sure that all the variables get added, even if they're not available on the platform where the generator is being run (for example, running `scons get-options` on Linux, but it has to include options for Windows).

SConstruct
ischoegl: That's unfortunately a little tricky, as it's very hard to guess what single quotes enclose (for example, you'd have to catch exceptions for all kinds of contractions "isn't", "doesn't", "can't", "you'd", "that's"). My way around this was to limit single quote extents to whatever doesn't include whitespace. So using double quotes here is imho the easier solution (and it is still clear from a CLI perspective).

SConstruct
ischoegl: This should become clear from the context once the ReST is assembled ... the corresponding portion is
```rest
.. _prefix:

*  ``prefix``: [ ``path/to/prefix`` ]
   Set this to the directory where Cantera should be installed. On Windows
   systems, ``$ProgramFiles`` typically refers to ``"C:\Program Files"``.

   -  default: platform dependent

      -  Windows: ``'$ProgramFiles\Cantera'``
      -  Otherwise: ``'/usr/local'``
```
The original (hand-coded) ReST explicitly referred to `C:\Program Files\Cantera`, but that is not what `SConstruct` actually implements.

ischoegl:
>  this is already done by SCons when `opts.AddVariables` is called. The `Variables` object stores all the options and their defaults. This is how the `help()` function works. 

That's not exactly what happens for platform/compiler-dependent options, as the default option is not known a priori. The way around this that I can see is to implement wrappers, that can generate SCons variables, but also output ReST. (e.g. wrap `PathVariable` in a `PathOption` that is defined in `buildutils`). 

It's much easier to resolve for the `scons help` function as you already know what compiler/platform you are using :wink:

SConstruct
bryanwweber: OK, the ReST looks good, but does `scons help` return all that context too?

ischoegl: Point taken!

ischoegl:
Overall, I'm :+1: with adding `scons get-options`, but that will definitely mean more lines to review ...

ischoegl:
Superseded by #1137

PR #1137, opened 2021-11-02
Refactor SConstruct option handling / extract scons options v2

ischoegl:

- Refactor `SConstruct`
- Implement `scons help --restructured-text` to create ReST output
- Refactor `scons help` to avoid redundant parsers
- Implement help on individual option via `scons help --option=<name_of_option>`
- Replaces #1136

Addresses Cantera/cantera-website#26

```
$ scons help --restructured-text
[...]
$ scons help --restructured-text --dev --output=config-options-dev.rst
[...]
```
or
```
$ scons help --restructured-text --option=prefix
INFO: SCons is using the following Python interpreter: /usr/bin/python3

.. _prefix:

*  ``prefix``: [ ``path/to/prefix`` ]
   Set this to the directory where Cantera should be installed. On Windows
   systems, ``$ProgramFiles`` typically refers to ``"C:\Program Files"``.

   -  default: platform dependent
      -  Windows: ``'$ProgramFiles\Cantera'``
      -  Otherwise: ``'/usr/local'``
```
Further
```
$ scons help --option=CXX
INFO: SCons is using the following Python interpreter: /usr/bin/python3

* CXX: [ string ]
    The C++ compiler to use.
    - default: '${CXX}'
    - actual: 'g++'
```
and
```
$ scons help --defaults --option=cc_flags
INFO: SCons is using the following Python interpreter: /usr/bin/python3

* cc_flags: [ string ]
    Compiler flags passed to both the C and C++ compilers, regardless of
    optimization level.
    - default: compiler dependent
      - If using MSVC: '/MD /nologo /D_SCL_SECURE_NO_WARNINGS /D_CRT_SECURE_NO_WARNINGS'
      - If using ICC: '-vec-report0 -diag-disable 1478'
      - If using Clang: '-fcolor-diagnostics'
      - Otherwise: ''
```

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -326,7 +326,12 @@ defaults.debug_link_flags = {"cl": "/DEBUG", "default": ""}
 defaults.no_debug_link_flags = ""
 defaults.warning_flags = {"cl": "/W3", "icc": "-Wcheck", "default": "-Wall"}
 defaults.build_pch = {"icc": False, "default": True}
-
+defaults.pch_flags = {
+    "cl": "/FIpch/system.h",
+    "gcc": "-include src/pch/system.h",
+    "clang": "-include-pch src/pch/system.h.gch",
+    "default": "",
+}
 defaults.thread_flags = {"Windows": "", "macOS": "", "default": "-pthread"}
 defaults.fs_layout = {"Windows": "compact", "default": "standard"}
 defaults.python_prefix = {"Windows": "", "default": "$prefix"}
@@ -335,7 +340,6 @@ defaults.env_vars = "PATH,LD_LIBRARY_PATH,PYTHONPATH"
 defaults.versioned_shared_library = {"mingw": False, "default": True}
 defaults.sphinx_options = "-W --keep-going"
 
-env['pch_flags'] = []
 env['openmp_flag'] = ['-fopenmp'] # used to generate sample build scripts
 
 env['using_apple_clang'] = False
@@ -350,11 +354,9 @@ if 'gcc' in env.subst('$CC') or 'gnu-cc' in env.subst('$CC'):
     if env['OS'] == 'Cygwin':
         defaults.select("Cygwin")
     defaults.select("gcc")
-    env['pch_flags'] = ['-include', 'src/pch/system.h']
 
 elif env['CC'] == 'cl': # Visual Studio
     defaults.select("cl")
-    env['pch_flags'] = ['/FIpch/system.h']
     env['openmp_flag'] = ['/openmp']
 
 elif 'icc' in env.subst('$CC'):
@@ -363,7 +365,6 @@ elif 'icc' in env.subst('$CC'):
 
 elif 'clang' in env.subst('$CC'):
     defaults.select("clang")
-    env['pch_flags'] = ['-include-pch', 'src/pch/system.h.gch']
 
 else:
     print("WARNING: Unrecognized C compiler '%s'" % env['CC'])
@@ -583,6 +584,10 @@ config_options = [
         'use_pch',
         """Use a precompiled-header to speed up compilation""",
         defaults.build_pch),
+    (
+        'pch_flags',
+        """Compiler flags when using precompiled-header.""",
+        defaults.pch_flags),
     (
         'cxx_flags',
         """Compiler flags passed to the C++ compiler only. Separate multiple
@@ -886,6 +891,7 @@ env['LINKFLAGS'] += listify(env['thread_flags'])
 env['CPPDEFINES'] = {}
 
 env['warning_flags'] = listify(env['warning_flags'])
+env["pch_flags"] = listify(env["pch_flags"])
 
 if env['optimize']:
     env['CCFLAGS'] += listify(env['optimize_flags'])
--- a/SConstruct
+++ b/SConstruct
@@ -195,8 +195,10 @@ if os.name == 'nt':
         (
             "msvc_version",
             """Version of Visual Studio to use. The default is the newest
-                installed version. Specify '12.0' for Visual Studio 2013 or '14.0'
-                for Visual Studio 2015. Windows MSVC only.""",
+               installed version. Specify '12.0' for Visual Studio 2013, '14.0' for
+               Visual Studio 2015, '14.1' ('14.1x') Visual Studio 2017, or '14.2'
+               ('14.2x') for Visual Studio 2019. For version numbers in parentheses,
+               'x' is a placeholder for a minor version number. Windows MSVC only.""",
             ""),
         EnumVariable(
             "target_arch",
@@ -216,7 +218,7 @@ if os.name == 'nt':
         EnumVariable(
             "toolchain",
             """The preferred compiler toolchain. If MSVC is not on the path but
-            'g++' is on he path, 'mingw' is used as a backup. Windows only.""",
+               'g++' is on the path, 'mingw' is used as a backup. Windows only.""",
             defaults.toolchain, ("msvc", "mingw", "intel")))
     opts.AddVariables(windows_compiler_options[-1])
     opts.Update(pickCompilerEnv)
@@ -444,7 +446,7 @@ config_options = [
         'matlab_path',
         """Path to the MATLAB install directory. This should be the directory
            containing the 'extern', 'bin', etc. subdirectories. Typical values
-           are: "C:/Program Files/MATLAB/R2011a" on Windows,
+           are: "C:\Program Files\MATLAB\R2021a" on Windows,
            "/Applications/MATLAB_R2011a.app" on macOS, or
            "/opt/MATLAB/R2011a" on Linux.""",
         '', PathVariable.PathAccept),
@@ -501,7 +503,7 @@ config_options = [
         'system_fmt',
         """Select whether to use the fmt library from a system installation
            ('y'), from a Git submodule ('n'), or to decide automatically
-           ('default').  If you do not want to use the Git submodule and fmt
+           ('default'). If you do not want to use the Git submodule and fmt
            is not installed directly into system include and library
            directories, then you will need to add those directories to
            'extra_inc_dirs' and 'extra_lib_dirs'. This installation of fmt
--- a/SConstruct
+++ b/SConstruct
@@ -64,6 +64,11 @@ if not COMMAND_LINE_TARGETS:
     logger.info(__doc__, print_level=False)
     sys.exit(0)
 
+if parse_version(SCons.__version__) < parse_version("3.0.0"):
+    logger.info("Cantera requires SCons with a minimum version of 3.0.0. Exiting.",
+                print_level=False)
+    sys.exit(0)
+
 valid_commands = ("build", "clean", "install", "uninstall",
                   "help", "msi", "samples", "sphinx", "doxygen", "dump")
 
@@ -378,10 +383,8 @@ if env['OS'] == 'Windows':
 elif env["OS"] == "Darwin":
     defaults.select("macOS")
 
-# InstallVersionedLib only fully functional in SCons >= 2.4.0
 # SHLIBVERSION fails with MinGW: http://scons.tigris.org/issues/show_bug.cgi?id=3035
-if (env['toolchain'] == 'mingw'
-    or parse_version(SCons.__version__) < parse_version('2.4.0')):
+if (env["toolchain"] == "mingw"):
     defaults.select("mingw")
 
 defaults.select("default")
--- a/SConstruct
+++ b/SConstruct
@@ -143,9 +143,16 @@ extraEnvArgs = {}
 
 
 class Defaults:
-    """Class enabling selection of options based on attribute dictionary entries"""
+    """
+    Class enabling selection of options based on attribute dictionary entries that
+    allof for differentiation between platform/compiler dependent options.
 
-    def apply(self, key="defaults"):
+    Individual member variables are assigned explicitly after object instantiation, and
+    should contain strings specifying platform/compiler independent default options, or
+    dictionaries with platform/compiler dependent options.
+    """
+
+    def select(self, key="default"):
         """Select attribute dictionary entries corresponding to *key*"""
         for attr, val in self.__dict__.items():
 
@@ -173,11 +180,11 @@ defaults.toolchain = {"Windows": "msvc"}
 defaults.target_arch = {"Windows": "amd64"}
 
 if os.name == 'nt':
-    defaults.apply("Windows")
+    defaults.select("Windows")
 
     # On Windows, target the same architecture as the current copy of Python,
     # unless the user specified another option.
-    if '64 bit' not in sys.version:
+    if "64 bit" not in sys.version:
         defaults.target_arch = "x86"
 
     # Make an educated guess about the right default compiler
@@ -270,7 +277,7 @@ if 'FRAMEWORKS' not in env:
 add_RegressionTest(env)
 
 defaults.prefix = {"Windows": "$ProgramFiles\Cantera", "default": "/usr/local"}
-defaults.boostIncDir = ""
+defaults.boost_inc_dir = ""
 
 if os.name == 'posix':
     env['INSTALL_MANPAGES'] = True
@@ -281,47 +288,51 @@ else:
     print("Error: Unrecognized operating system '%s'" % os.name)
     sys.exit(1)
 
-defaults.CXX = "${CXX}"
-defaults.CC = "${CC}"
+defaults.cxx = "${CXX}"
+defaults.cc = "${CC}"
 
 compiler_options = [
     ('CXX',
      """The C++ compiler to use.""",
-     defaults.CXX),
+     defaults.cxx),
     ('CC',
      """The C compiler to use. This is only used to compile CVODE.""",
-     defaults.CC)]
+     defaults.cc)]
 
 compiler_options = substitute(compiler_options, env=env)
 opts.AddVariables(*compiler_options)
 opts.Update(env)
 
-defaults.cxxFlags = {"cl": "/EHsc", "Cygwin": "-std=gnu++11", "default": "-std=c++11"}
-defaults.ccFlags = {
+defaults.cxx_flags = {
+    "cl": "/EHsc",
+    "Cygwin": "-std=gnu++11", # See http://stackoverflow.com/questions/18784112
+    "default": "-std=c++11"
+}
+defaults.cc_flags = {
     "cl": "/MD /nologo /D_SCL_SECURE_NO_WARNINGS /D_CRT_SECURE_NO_WARNINGS",
     "icc": "-vec-report0 -diag-disable 1478",
     "clang": "-fcolor-diagnostics",
     "default": "",
 }
-defaults.noOptimizeCcFlags = {"cl": "/Od /Ob0", "default": "-O0"}
-defaults.optimizeCcFlags = {
+defaults.no_optimize_cc_flags = {"cl": "/Od /Ob0", "default": "-O0"}
+defaults.optimize_cc_flags = {
     "cl": "/O2",
     "gcc": "-O3 -Wno-inline",
     "default": "-O3",
 }
-defaults.debugCcFlags = {"cl": "/Zi /Fd${TARGET}.pdb", "default": "-g"}
-defaults.noDebugCcFlags = ""
-defaults.debugLinkFlags = {"cl": "/DEBUG", "default": ""}
-defaults.noDebugLinkFlags = ""
-defaults.warningFlags = {"cl": "/W3", "icc": "-Wcheck", "default": "-Wall"}
-defaults.buildPch = {"icc": False, "default": True}
+defaults.debug_cc_flags = {"cl": "/Zi /Fd${TARGET}.pdb", "default": "-g"}
+defaults.no_debug_cc_flags = ""
+defaults.debug_link_flags = {"cl": "/DEBUG", "default": ""}
+defaults.no_debug_link_flags = ""
+defaults.warning_flags = {"cl": "/W3", "icc": "-Wcheck", "default": "-Wall"}
+defaults.build_pch = {"icc": False, "default": True}
 
-defaults.threadFlags = {"Windows": "", "macOS": "", "default": "-pthread"}
-defaults.fsLayout = {"Windows": "compact", "default": "standard"}
+defaults.thread_flags = {"Windows": "", "macOS": "", "default": "-pthread"}
+defaults.fs_layout = {"Windows": "compact", "default": "standard"}
 defaults.python_prefix = {"Windows": "", "default": "$prefix"}
 defaults.python_cmd = "${PYTHON_CMD}"
 defaults.env_vars = "PATH,LD_LIBRARY_PATH,PYTHONPATH"
-defaults.versionedSharedLibrary = {"mingw": False, "default": True}
+defaults.versioned_shared_library = {"mingw": False, "default": True}
 defaults.sphinx_options = "-W --keep-going"
 
 env['pch_flags'] = []
@@ -336,40 +347,39 @@ if env['OS'] == 'Darwin':
         env['openmp_flag'].insert(0, '-Xpreprocessor')
 
 if 'gcc' in env.subst('$CC') or 'gnu-cc' in env.subst('$CC'):
-    defaults.apply("gcc")
     if env['OS'] == 'Cygwin':
-        # See http://stackoverflow.com/questions/18784112
-        defaults.apply("Cygwin")
+        defaults.select("Cygwin")
+    defaults.select("gcc")
     env['pch_flags'] = ['-include', 'src/pch/system.h']
 
 elif env['CC'] == 'cl': # Visual Studio
-    defaults.apply("cl")
+    defaults.select("cl")
     env['pch_flags'] = ['/FIpch/system.h']
     env['openmp_flag'] = ['/openmp']
 
 elif 'icc' in env.subst('$CC'):
-    defaults.apply("icc")
+    defaults.select("icc")
     env['openmp_flag'] = ['-openmp']
 
 elif 'clang' in env.subst('$CC'):
-    defaults.apply("clang")
+    defaults.select("clang")
     env['pch_flags'] = ['-include-pch', 'src/pch/system.h.gch']
 
 else:
     print("WARNING: Unrecognized C compiler '%s'" % env['CC'])
 
 if env['OS'] == 'Windows':
-    defaults.apply("Windows")
+    defaults.select("Windows")
 elif env["OS"] == "Darwin":
-    defaults.apply("macOS")
+    defaults.select("macOS")
 
 # InstallVersionedLib only fully functional in SCons >= 2.4.0
 # SHLIBVERSION fails with MinGW: http://scons.tigris.org/issues/show_bug.cgi?id=3035
 if (env['toolchain'] == 'mingw'
     or parse_version(SCons.__version__) < parse_version('2.4.0')):
-    defaults.apply("mingw")
+    defaults.select("mingw")
 
-defaults.apply("default")
+defaults.select("default")
 defaults.python_cmd = sys.executable
 
 # **************************************
@@ -572,21 +582,21 @@ config_options = [
     BoolVariable(
         'use_pch',
         """Use a precompiled-header to speed up compilation""",
-        defaults.buildPch),
+        defaults.build_pch),
     (
         'cxx_flags',
         """Compiler flags passed to the C++ compiler only. Separate multiple
            options with spaces, for example, "cxx_flags='-g -Wextra -O3 --std=c++11'"
            """,
-        defaults.cxxFlags),
+        defaults.cxx_flags),
     (
         'cc_flags',
         """Compiler flags passed to both the C and C++ compilers, regardless of optimization level.""",
-        defaults.ccFlags),
+        defaults.cc_flags),
     (
         'thread_flags',
         """Compiler and linker flags for POSIX multithreading support.""",
-        defaults.threadFlags),
+        defaults.thread_flags),
     BoolVariable(
         'optimize',
         """Enable extra compiler optimizations specified by the
@@ -596,11 +606,11 @@ config_options = [
     (
         'optimize_flags',
         """Additional compiler flags passed to the C/C++ compiler when 'optimize=yes'.""",
-        defaults.optimizeCcFlags),
+        defaults.optimize_cc_flags),
     (
         'no_optimize_flags',
         """Additional compiler flags passed to the C/C++ compiler when 'optimize=no'.""",
-        defaults.noOptimizeCcFlags),
+        defaults.no_optimize_cc_flags),
     BoolVariable(
         'debug',
         """Enable compiler debugging symbols.""",
@@ -608,25 +618,25 @@ config_options = [
     (
         'debug_flags',
         """Additional compiler flags passed to the C/C++ compiler when 'debug=yes'.""",
-        defaults.debugCcFlags),
+        defaults.debug_cc_flags),
     (
         'no_debug_flags',
         """Additional compiler flags passed to the C/C++ compiler when 'debug=no'.""",
-        defaults.noDebugCcFlags),
+        defaults.no_debug_cc_flags),
     (
         'debug_linker_flags',
         """Additional options passed to the linker when 'debug=yes'.""",
-        defaults.debugLinkFlags),
+        defaults.debug_link_flags),
     (
         'no_debug_linker_flags',
         """Additional options passed to the linker when 'debug=no'.""",
-        defaults.noDebugLinkFlags),
+        defaults.no_debug_link_flags),
     (
         'warning_flags',
         """Additional compiler flags passed to the C/C++ compiler to enable
            extra warnings. Used only when compiling source code that is part
            of Cantera (for example, excluding code in the 'ext' directory).""",
-        defaults.warningFlags),
+        defaults.warning_flags),
     (
         'extra_inc_dirs',
         """Additional directories to search for header files, with multiple
@@ -641,7 +651,7 @@ config_options = [
         'boost_inc_dir',
         """Location of the Boost header files. Not needed if the headers are
            installed in a standard location, for example, '/usr/include'.""",
-        defaults.boostIncDir, PathVariable.PathAccept),
+        defaults.boost_inc_dir, PathVariable.PathAccept),
     PathVariable(
         'stage_dir',
         """Directory relative to the Cantera source directory to be
@@ -675,7 +685,7 @@ config_options = [
            actual library and 'libcantera_shared.so' and 'libcantera_shared.so.2'
            as symlinks.
            """,
-        defaults.versionedSharedLibrary),
+        defaults.versioned_shared_library),
     BoolVariable(
         'use_rpath_linkage',
         """If enabled, link to all shared libraries using 'rpath', i.e., a fixed
@@ -690,7 +700,7 @@ config_options = [
            files in the subdirectory defined by 'prefix'. This layout is best
            with a prefix like '/opt/cantera'. 'debian' installs to the stage
            directory in a layout used for generating Debian packages.""",
-        defaults.fsLayout, ('standard','compact','debian')),
+        defaults.fs_layout, ('standard','compact','debian')),
     BoolVariable(
         "fast_fail_tests",
         """If enabled, tests will exit at the first failure.""",
--- a/SConstruct
+++ b/SConstruct
@@ -156,11 +156,12 @@ if os.name == 'nt':
         defaultToolchain = 'msvc'
 
     windows_compiler_options.extend([
-        ('msvc_version',
-         """Version of Visual Studio to use. The default is the newest
-            installed version. Specify '12.0' for Visual Studio 2013 or '14.0'
-            for Visual Studio 2015.""",
-         ''),
+        (
+            "msvc_version",
+            """Version of Visual Studio to use. The default is the newest
+                installed version. Specify '12.0' for Visual Studio 2013 or '14.0'
+                for Visual Studio 2015.""",
+            ""),
         EnumVariable(
             'target_arch',
             """Target architecture. The default is the same architecture as the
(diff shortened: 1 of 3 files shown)


ischoegl:
@bryanwweber ... I opened a new PR as the implementation did change a bit. ~I left the old `buildutils.py:help` in for the time being, but would replace it with a new version to avoid code redundancies.~

ischoegl:
Never liked the command `get-options` in this context. I decided to make it part of `scons help`, which uses the option `--restructured-text` for ReST output. Also implemented help for an individual option, which works for both ReST and CLI.

ischoegl:
@bryanwweber / @speth .... that's all I have at the moment. I believe this PR prepares for a resolution of Cantera/cantera-website#26 pretty thoroughly while also introducing some improvements of the user interface (various `scons help` options). Under the hood, I somewhat reorganized the first portion of `SConstruct` where newly introduced wrapper objects (`Option`+specializations and `Configuration`) allow for a much closer integration of options and various platform/compiler-dependent defaults. The AST is indeed fragile, so this PR proposes some new additions to `buildutils.py`.

SConstruct
```diff
@@ -64,6 +64,15 @@ if not COMMAND_LINE_TARGETS:
     logger.info(__doc__, print_level=False)
     sys.exit(0)
 
+if parse_version(SCons.__version__) < parse_version("3.0.0"):
+    logger.info("Cantera requires SCons with a minimum version of 3.0.0. Exiting.",
+                print_level=False)
+    sys.exit(0)
```
bryanwweber: This seems like an error, any reason to have it be `info` and `exit(0)`?

SConstruct
```diff
@@ -64,6 +64,15 @@ if not COMMAND_LINE_TARGETS:
...
+
+if os.name not in ["nt", "posix"]:
+    print(f"Error: Unrecognized operating system '{os.name}'")
+    sys.exit(1)
```
bryanwweber: Is there a specific reason to include this check? I suppose if we knew this didn't work on other platforms, it'd make sense to have this, but is there any chance that it does work on other platforms, even though we don't test it?

SConstruct
```diff
@@ -130,257 +139,100 @@ if "test-clean" in COMMAND_LINE_TARGETS:
...
 
 config_options = [
-    PathVariable(
-        'prefix',
-        'Set this to the directory where Cantera should be installed.',
-        defaults.prefix, PathVariable.PathAccept),
-    PathVariable(
+    Option(
+        "CXX",
+        """The C++ compiler to use.""",
```
```suggestion
        "The C++ compiler to use.",
```

SConstruct
```diff
@@ -130,257 +139,100 @@ if "test-clean" in COMMAND_LINE_TARGETS:
...
+        "${CXX}"),
+    Option(
+        'cxx_flags',
```
```suggestion
        "cxx_flags",
```

SConstruct
```diff
@@ -130,257 +139,100 @@ if "test-clean" in COMMAND_LINE_TARGETS:
...
+           options with spaces, for example, "cxx_flags='-g -Wextra -O3 --std=c++11'"
+           """,
+        {
+            "cl": "/EHsc",
+            "Cygwin": "-std=gnu++11", # See http://stackoverflow.com/questions/18784112
+            "default": "-std=c++11"
+        }),
+    Option(
+        "CC",
+        """The C compiler to use. This is only used to compile CVODE.""",
```
```suggestion
        "The C compiler to use. This is only used to compile CVODE.",
```

SConstruct
```diff
@@ -130,257 +139,100 @@ if "test-clean" in COMMAND_LINE_TARGETS:
...
+        "${CC}"),
+    Option(
+        "cc_flags",
+        """Compiler flags passed to both the C and C++ compilers, regardless of
+            optimization level.""",
```
bryanwweber: The leading extra space here looks strange to me. Is it needed because there's no space at the end of the previous line? What if `\n` is replaced by ` ` instead of being stripped?

SConstruct
```diff
@@ -130,257 +139,100 @@ if "test-clean" in COMMAND_LINE_TARGETS:
...
+    PathOption(
         'libdirname',
         """Set this to the directory where Cantera libraries should be installed.
-           Some distributions (for example, Fedora/RHEL) use 'lib64' instead of 'lib' on 64-bit systems
-           or could use some other library directory name instead of 'lib' depends
-           on architecture and profile (for example, Gentoo 'libx32' on x32 profile).
-           If user didn't set 'libdirname' configuration variable set it to default value 'lib'""",
+           Some distributions (for example, Fedora/RHEL) use 'lib64' instead of 'lib'
+           on 64-bit systems or could use some other library directory name instead of
+           'lib' depends on architecture and profile (for example, Gentoo 'libx32' on
```
```suggestion
           'lib', depending on architecture and profile (for example, Gentoo 'libx32' on
```

SConstruct
```diff
@@ -130,257 +139,100 @@ if "test-clean" in COMMAND_LINE_TARGETS:
...
            behavior is to build the full Python module for whichever version of
            Python is running SCons if the required prerequisites (NumPy and
-           Cython) are installed. Note: ``y`` is a synonym for ``full`` and ``n``
-           is a synonym for ``none``.""",
+           Cython) are installed. Note: 'y' is a synonym for 'full' and 'n'
+           is a synonym for 'none'.""",
         'default', ('full', 'minimal', 'none', 'n', 'y', 'default')),
-    PathVariable(
+    PathOption(
         'python_cmd',
```
```suggestion
        "python_cmd",
```

SConstruct
```diff
@@ -130,257 +139,100 @@ if "test-clean" in COMMAND_LINE_TARGETS:
...
         """Cantera needs to know where to find the Python interpreter. If
-           PYTHON_CMD is not set, then the configuration process will use the
+           'PYTHON_CMD' is not set, then the configuration process will use the
            same Python interpreter being used by SCons.""",
-        sys.executable, PathVariable.PathAccept),
-    PathVariable(
+        "${PYTHON_CMD}",
+        PathVariable.PathAccept),
+    PathOption(
         'python_prefix',
```
```suggestion
        "python_prefix",
```

SConstruct
```diff
@@ -389,62 +241,63 @@ config_options = [
...
            'matlab_path' is set.""",
         'default', ('y', 'n', 'default')),
-    PathVariable(
+    PathOption(
         'matlab_path',
         """Path to the MATLAB install directory. This should be the directory
            containing the 'extern', 'bin', etc. subdirectories. Typical values
-           are: "C:/Program Files/MATLAB/R2011a" on Windows,
-           "/Applications/MATLAB_R2011a.app" on OS X, or
+           are: "C:\Program Files\MATLAB\R2021a" on Windows,
```
bryanwweber: If you switch this to forward slashes, then they either need to be escaped or it needs to be a raw string.

SConstruct
```diff
@@ -690,7 +553,212 @@ config_options = [
...
+        action="store_true", help="Format defaults as ReST")
+    AddOption(
+        "--option", dest="option", nargs=1, type="string",
+        action="store", help="Output help for specific option")
+    AddOption(
+        "--defaults", dest="defaults",
+        action="store_true", help="All defaults (CLI only)")
+    AddOption(
+        "--dev", dest="dev",
+        action="store_true", help="Append -dev (ReST only)")
```
bryanwweber: FWIW, the docutils docs abbreviate reStructuredText as `reST`

SConstruct
```diff
@@ -690,7 +553,212 @@ config_options = [
...
+            extraEnvArgs["MSVC_VERSION"] = windows_compiler_env["msvc_version"]
+        msvc_version = (windows_compiler_env["msvc_version"] or
+                        windows_compiler_env["MSVC_VERSION"])
+        logger.info(f"Compiling with MSVC {msvc_version}", print_level=False)
+
+    elif windows_compiler_env["toolchain"] == "mingw":
+        toolchain = ["mingw", "f90"]
+        extraEnvArgs["F77"] = None
+        # Next line fixes http://scons.tigris.org/issues/show_bug.cgi?id=2683
+        extraEnvArgs["WINDOWS_INSERT_DEF"] = 1
```
bryanwweber, line 700: This appears to be fixed now, see: https://github.com/SCons/scons/issues/2856 At the least, the link should be updated to the copied GitHub issue: https://github.com/scons/scons/issues/2683

SConstruct
```diff
@@ -690,7 +553,212 @@ config_options = [
...
+
+    config["prefix"].default = pjoin(os.environ["ProgramFiles"], "Cantera")
+    config.select("Windows")
+
+    # On Windows, target the same architecture as the current copy of Python,
+    # unless the user specified another option.
+    if "64 bit" not in sys.version:
+        config["target_arch"].default = "x86"
+
+    opts.AddVariables(*config.to_scons(["msvc_version", "target_arch"]))
```
bryanwweber: Could this be a tuple? Although it probably doesn't make a difference, tuples are slightly faster, and semantically indicate the argument won't be change.
```suggestion
    opts.AddVariables(*config.to_scons(("msvc_version", "target_arch")))
```

SConstruct
```diff
@@ -705,19 +773,22 @@ for option in opts.keys():
...
 if 'sphinx' in COMMAND_LINE_TARGETS:
     env['sphinx_docs'] = True
 
-valid_arguments = (set(opt[0] for opt in windows_compiler_options) |
-                   set(opt[0] for opt in compiler_options) |
-                   set(opt[0] for opt in config_options))
 for arg in ARGUMENTS:
-    if arg not in valid_arguments:
+    if arg not in config:
         print('Encountered unexpected command line argument: %r' % arg)
```
bryanwweber: Could this be a call to the logger?

site_scons/buildutils.py
```diff
@@ -10,20 +10,22 @@
 import enum
 from pathlib import Path
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Union, List, Dict, Generator
```
bryanwweber: If you quote these when they're used, it functions as a forward declaration for typing purposes, but it isn't evaluated at runtime.

site_scons/buildutils.py
```diff
@@ -34,6 +36,373 @@
     TPathLike = TypeVar("TPathLike", Path, str)
 
 
+class Option:
+    """Object corresponding to SCons configuration option"""
+
+    def __init__(self,
+            name: str,
+            description: str,
+            default: Union[bool, str],
```
bryanwweber: To my comment above, this would become:
```suggestion
            default: "Union[bool, str]",
```

site_scons/buildutils.py
```diff
@@ -34,6 +36,373 @@
...
+        self.set_wrapper_indent(4)
+
+    @property
+    def parameters(self) -> tuple:
+        """Parameters of the SCons configuration option"""
+        if self.choices is None:
+            return self.name, self.description, self.default
+        return self.name, self.description, self.default, self.choices
+
+    def to_scons(self, env: "SCEnvironment"=None) -> "SCVariables":
```
bryanwweber: When type information is included, there's usually spaces around `=`, even in the function definition. This should also be optional:
```suggestion
    def to_scons(self, env: "Optional[SCEnvironment]" = None) -> "SCVariables":
```

site_scons/buildutils.py
```diff
@@ -34,6 +36,373 @@
...
+        self.name = name
+        self.description = Option._deblank(description)
+        self.default = default
+        self.choices = choices
+
+        self._wrapper = textwrap.TextWrapper(width=80)
+        self.set_wrapper_indent(4)
+
+    @property
+    def parameters(self) -> tuple:
```
bryanwweber: Ideally, this could indicate the types in the tuple. Something like `"Union[Tuple[str, str, str, str], Tuple[str, str, str]]"` (I'm not sure about all the types).

site_scons/buildutils.py
```diff
@@ -34,6 +36,373 @@
...
+        if isinstance(self.default, str):
+            out = list(self.parameters)
+            if env is not None and "$" in out[2]:
+                out[2] = env.subst(out[2])
+            return tuple(out)
+
+        raise TypeError(f"Invalid defaults option with type '{type(self.default)}'")
+
+    @property
+    def wrapper(self):
```
bryanwweber: This should have a return type

site_scons/buildutils.py
```diff
@@ -34,6 +36,373 @@
...
+            key = f"'{key}'" if tick else key
+            return f"``{key}``" if backticks else key
+
+        # format choices
+        if isinstance(self, PathOption):
+            choices = f"path/to/{self.name}"
+            choices = f"{decorate(choices, False)}"
+        elif isinstance(self.choices, (list, tuple)):
+            choices = list(self.choices)
+            for yes_no in ["n", "y"]:
```
bryanwweber:
```suggestion
            for yes_no in ("y", "n"):
```
For yes-no in "no", "yes" 😂

bryanwweber (changes requested):
Thanks for taking this on @ischoegl! I think the changes look good overall. The main question I had, besides minor typos, is about type hinting in the `Option` class. It seems like the type hinting isn't consistently applied. I think it's generally fine to leave out the types if you don't want to include them, but then it should be consistently applied for all the class's methods.

SConstruct
ischoegl: There isn't, but the check existed before as:
https://github.com/Cantera/cantera/blob/2d211dbca1e3ab6bbf6cf8feeea8d59089c74b1b/SConstruct#L242-L252

SConstruct
ischoegl: Removed - this was presumably due to hitting a tab.

bryanwweber: This happened in a few other places too, can you just check for it elsewhere?

site_scons/buildutils.py
ischoegl: This one I'm actually not sure about. While I added the spaces around `=` throughout, the `Optional[...]` appears to add clutter: from the context, it is clear that this is optional already, as default values are being specified?

ischoegl:
@bryanwweber ... thanks for the review! I ended up replacing all single quotes in the passage that contains the configuration options, as it was a mix and match of new and old (same for the triple `"""` you wanted to have changed). 

Regarding type hinting, I adopted your recommendations (with the notable exception of `Option[...]` which appears to unnecessarily add clutter). While I've been using PEP-484 decorations for a while, I wasn't necessarily enforcing things (it was more for documentation purposes). Your comments certainly clarified several things for me that I had seen (and used!), but were mainly based on copy-paste or StackOverflow ...

I can remove the check for operating system, but since it hasn't ever come up before, I opted to leave it.

I also added a documentation of the new help command options as I believe that those may be useful. I don't think that it makes sense to add the `--restructured-text --dev` there as they are mostly for internal use.

ischoegl:
@bryanwweber  ... done. I ended up improving the output formatting of one of the help command options I almost forgot about:
```
$ scons help --list-options
INFO: SCons is using the following Python interpreter: /usr/bin/python3
INFO: Available configuration options:
CC                   gtest_flags                      renamed_shared_libraries   
CXX                  lapack_ftn_string_len_at_end     show_long_tests            
FORTRAN              lapack_ftn_trailing_underscore   skip_slow_tests            
FORTRANFLAGS         lapack_names                     sphinx_cmd                 
VERBOSE              layout                           sphinx_docs                
blas_lapack_dir      legacy_rate_constants            sphinx_options             
blas_lapack_libs     libdirname                       stage_dir                  
boost_inc_dir        matlab_path                      sundials_include           
cc_flags             matlab_toolbox                   sundials_libdir            
coverage             msvc_version                     system_eigen               
cxx_flags            no_debug_flags                   system_fmt                 
debug                no_debug_linker_flags            system_sundials            
debug_flags          no_optimize_flags                system_yamlcpp             
debug_linker_flags   openmp_flag                      target_arch                
doxygen_docs         optimize                         thread_flags               
env_vars             optimize_flags                   toolchain                  
extra_inc_dirs       pch_flags                        use_pch                    
extra_lib_dirs       prefix                           use_rpath_linkage          
f90_interface        python_cmd                       verbose_tests              
fast_fail_tests      python_package                   versioned_shared_library   
googletest           python_prefix                    warning_flags      
```
and added it to the help output
```
$ scons

SCons build script for Cantera

Basic usage:
    'scons help' - print a description of user-specifiable options.

    'scons build' - Compile Cantera and the language interfaces using
                    default options.

    ...

    'scons doxygen' - Build the Doxygen documentation

Additional help command options:
    'scons help --list-options' - List all available configuration options.

    'scons help --option=<opt>' - Print the description of a specific option
                                  with name <opt>, for example
                                  'scons help --option=prefix'

    'scons help --defaults' - Print a description of default values. This can
                              be combined with --option, for example
                              'scons help --defaults --option=cc_flags'
```
_PS: as an observation, getting help by typing `scons` and a list of configuration options by typing `scons help` is not necessarily intuitive. Changing this would break with a long-standing tradition though!_

site_scons/buildutils.py
```diff
@@ -34,6 +36,396 @@
...
+        self.set_wrapper_indent(4)
+
+    @property
+    def parameters(self) -> "Union[Tuple[str, str, str, str], Tuple[str, str, str]]":
+        """Parameters of the SCons configuration option"""
+        if self.choices is None:
+            return self.name, self.description, self.default
+        return self.name, self.description, self.default, self.choices
+
+    def to_scons(self, env: "SCEnvironment" = None) -> "SCVariables":
```
bryanwweber: `"Optional[SCnvironment]" ` is equivalent to `"Union[SCEnvironment, None]"`, so it is required here, and everywhere `None` can be passed. I agree its not a great syntax, and starting with 3.10, we can use `|` instead of `Union`, so this would become `"SCEnvironment | None" `.

site_scons/buildutils.py
bryanwweber: See my other comment, sorry I didn't make that comment here

site_scons/buildutils.py
ischoegl: Hm. So anytime we pass `None` as a sentinel value, it is required to add this. I routinely use `None` as default, which I should probably stop then.

site_scons/buildutils.py
ischoegl: Alright. Replaced by `"Union[SCEnvironment, None]"`

site_scons/buildutils.py
ischoegl: I can live with `"Union[SCEnvironment, None]"` ... extrapolating the `Optional[...]` would apply to every single keyword argument!

ischoegl:
@bryanwweber ... fixed.

I also ended up changing the way the help system works (I can drop the commit):
```
$ scons              
ERROR: Missing command argument: type 'scons help' for information.
```
and 
```
$ scons help
INFO: SCons is using the following Python interpreter: /usr/bin/python3

SCons build script for Cantera

Basic usage:
    'scons help' - Show this help message.

    'scons build' - Compile Cantera and the language interfaces using
                    default options.

...

Additional help command options:
    'scons help --options' - Print a description of user-specifiable options.

    'scons help --list-options' - Print formatted list of available options.

    'scons help --option=<opt>' - Print the description of a specific option
                                  with name <opt>, for example
                                  'scons help --option=prefix'

    'scons help --defaults' - Print a description of default values. This can
                              be combined with --option, for example
                              'scons help --defaults --option=cc_flags'
```
as it's too easy to overlook that the basic help text even exists!

The old behavior is preserved as
```
$ scons help --options
INFO: SCons is using the following Python interpreter: /usr/bin/python3

        **************************************************
        *   Configuration options for building Cantera   *
        **************************************************

The following options can be passed to SCons to customize the Cantera
build process. They should be given in the form:

    scons build option1=value1 option2=value2

...
```

site_scons/buildutils.py
bryanwweber: I think the preference is generally for `Optional[value]` when `None` is used as a sentinel value (indicating that there's some default behavior), and the `Union` is for when `None` is a valid value. Obviously that's not enforced at a type-checking level, but IMO it communicates intent more clearly.

`Optional` is only required for cases where `None` is the default value (or a possible value). In cases where there is another type as the default, that needs to be in the `Union` instead, and you have to use `Union` (only if there are multiple possible types) 😊

ischoegl: I guess I learned something new here - thanks for explaining! Will change to `Optional[...]` then. I replaced all other occurrences of `None` as I really don't like the clutter it creates.

ischoegl:
@bryanwweber ... at this point, I adopted all comments (except for leaving the check for the OS in place). Let me know if you see anything else.

SConstruct
speth: I think the [only other possibility](https://docs.python.org/dev/library/os.html#os.name) here is `java`, which I guess corresponds to Jython and definitely wouldn't work for compiling Cantera.

ischoegl: @speth ... thanks for the comment! My personal preference would be to leave things as is (i.e. not support other OS's), and wait for an enhancement request, rather than potentially receiving an issue where no appropriate exception is raised.

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+    SCons configuration options, which are made available to both help text and
+    reST documentation. The class works in conjunction with the `Configuration` class
+    to select and convert system-dependent parameters to SCons configuration options.
+    """
+
+    def __init__(self,
+            name: str,
+            description: str,
+            default: "Union[bool, str]",
+            choices: list = []):
```
bryanwweber: Sorry if I missed this before... I saw it added in one of your new-ish changes too. Using mutable values as default arguments isn't usually a good idea. For example,
```
In [1]: def f(x = []):
   ...:     x.append(1)
   ...:     return x
   ...:

In [2]: f()
Out[2]: [1]

In [3]: f()
Out[3]: [1, 1]

In [4]: f()
Out[4]: [1, 1, 1]
```
An empty tuple should work here as a replacement.

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+        self.name = name
+        self.description = Option._deblank(description)
+        self.default = default
+        self.choices = choices
+
+        self._wrapper = textwrap.TextWrapper(width=80)
+        self.set_wrapper_indent(4)
+
+    @property
+    def parameters(self) -> "Union[Tuple[str, str, str, str], Tuple[str, str, str]]":
```
bryanwweber:
```suggestion
    def parameters(self) -> "Union[Tuple[str, str, Union[bool, str], Iterator[str]], Tuple[str, str, Union[bool, str]]]":
```
Based on the signature of `__init__()`.

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+        """Parameters of the SCons configuration option"""
+        if not self.choices:
+            return self.name, self.description, self.default
+        return self.name, self.description, self.default, self.choices
+
+    def to_scons(self, env: "Optional[SCEnvironment]" = None) -> "SCVariables":
```
bryanwweber: There's another option for a return value here, a `tuple` of `out`, which is possibly a tuple of `SCVariables`?

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+        if not len(out[-1]):
+            out = out[:-1]
+        out = "\n".join(out)
+        return out
+
+    def _build_title(self, backticks: bool = True, indent: int = 3) -> str:
+        """Build title describing option and defaults"""
+        # First line: "* option-name: [ 'choice1' | 'choice2' ]"
+
+        def decorate(key: str, tick: bool=True):
```
bryanwweber:
```suggestion
        def decorate(key: str, tick: bool=True) -> str:
```

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+    @staticmethod
+    def _build_defaults(
+            defaults: "Union[str, Dict[str, Union[str, bool]]]",
+            backticks: bool = True,
+            title: str = "default",
+            indent: int = 3,
+            hanging: int = 0) -> str:
+        """Assemble defaults from dictionary"""
+
+        def decorate(key: str, tick: bool=True):
```
bryanwweber:
```suggestion
        def decorate(key: str, tick: bool=True) -> str:
```

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+            default = "yes" if defaults else "no"
+            return f"{level1}{title}: {decorate(default)}\n"
+
+        if not isinstance(defaults, dict):
+            return f"{level1}{title}: {decorate(defaults)}\n"
+
+        # First line of default options
+        compilers = {"cl": "MSVC", "gcc": "GCC", "icc": "ICC", "clang": "Clang"}
+        toolchains = {"mingw", "intel", "msvc"}
+        if any([d in compilers for d in defaults]):
```
bryanwweber: You don't need the square brackets here and in the next `elif`, there's no need to actually generate the list, `any()` will do the right thing with just a generator expression.

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+            option2 = 'value2'
+        """
+        )
+    ]
+
+    def __init__(self):
+        self.options = {}
+        self.exported = []
+
+    def add(self, options: "Union[Option, List[Option]]" = []) -> None:
```
bryanwweber: Ah, yes. Here's the other place where a list is used as the default 😁 If it doesn't have to be a list specifically, then there's the general `Iterator` type.

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+            if not isinstance(item, Option):
+                raise TypeError(f"Invalid option with type '{type(item)}'")
+            self.options[item.name] = item
+
+    def list_options(self) -> str:
+        """Create formatted list of available configuration options"""
+        options = list(self.options.keys())
+
+        # create alphabetically sorted list
+        options.sort()
```
bryanwweber: This can be simplified a bit:
```
options = sorted(self.options.keys())
```
since `sorted()` returns a `list`.

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+
+    def __getitem__(self, key: str) -> "Union[str, bool, dict]":
+        """Make class subscriptable"""
+        return self.options[key]
+
+    def __len__(self) -> int:
+        """Length corresponds to number of keys"""
+        return len(self.options)
+
+    def __iter__(self) -> "Generator":
```
bryanwweber: I think this return type should be `Iterable[str]` (presuming all the keys are strings). According to [this SO answer](https://stackoverflow.com/a/42227485), a `Generator` should be annotated with a `send_type`, as well as the `yield_type` and `return_type`. I've never seen it before, but apparently, with a true generator, you can do `gen.send(value)` to have the generator run itself. New to me!

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+        for attr, item in self.options.items():
+
+            if isinstance(item.default, dict) and key in item.default:
+                self.options[attr].default = item.default[key]
+
+    def to_scons(
+            self,
+            keys: "Union[str, List[str]]" = "",
+            env: "Optional[SCEnvironment]" = None
+        ) -> "Union[SCVariables, List[SCVariables]]":
```
bryanwweber: This looks like it can only be the `List` version as the return value.

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+        message.append("")
+        for key in self:
+            message.append(self.options[key].to_rest(dev=dev))
+
+        return "\n".join(message)
+
+    def help(
+            self,
+            option: str = "",
+            env: "Optional[SCEnvironment]" = None) -> "List[str]":
```
bryanwweber: The last return statement in this function is a plain `str`, not a list.

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+        message.append(f"{'':^<{len(message[-1])}}")
+        message.append("")
+        for key in self:
+            message.append(self.options[key].to_rest(dev=dev))
+
+        return "\n".join(message)
+
+    def help(
+            self,
+            option: str = "",
```
bryanwweber: I think `None` is fine here as a sentinel value. I think it's clearer than an empty string. Using `None` here can simplify the `elif` structure below too.

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+            env: "Optional[SCEnvironment]" = None) -> "List[str]":
+        """Convert configuration help for command line interface (CLI) output"""
+        if option == "":
+            pass
+        elif option in self.options:
+            return "\n" + self.options[option].help(env=env)
+        else:
+            raise KeyError(f"Unknown option '{option}'")
```
bryanwweber: Assuming you have `None` as the sentinel, you can do:
```python
if option in self.options:
    return "\n" ... # what you've got here already
elif option is not None:
    raise ValueError(f"Unknown option: '{option}'")
```
That eliminates the `pass`. I suppose you could still use the empty string as a sentinel with this simplification... Up to you!

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+    def select(self, key: str = "default") -> None:
+        """Select attribute dictionary entries corresponding to *key*"""
+        for attr, item in self.options.items():
+
+            if isinstance(item.default, dict) and key in item.default:
+                self.options[attr].default = item.default[key]
+
+    def to_scons(
+            self,
+            keys: "Union[str, List[str]]" = "",
```
bryanwweber: Assuming tuples are also OK, you can do:

```suggestion
            keys: "Union[str, Iterator[str]]" = "",
```

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+
+            option1 = 'value1'
+            option2 = 'value2'
+        """
+        )
+    ]
+
+    def __init__(self):
+        self.options = {}
+        self.exported = []
```
bryanwweber:
```suggestion
        self.options: "Dict[str, Union[str, bool, dict]]" = {}
        self.exported: "List[str]" = []
```
Ideally that `dict` in the `Union` would also be typed 😁

site_scons/buildutils.py
bryanwweber: This also applies to `to_rest()`.

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+            return "\n" + self.options[option].help(env=env)
+        else:
+            raise KeyError(f"Unknown option '{option}'")
+
+        message = self.header
+        message.append(f"{'':->80}")
+        message.append("")
+
+        for key in self:
+            message.append(self.options[key].help(env=env))
```
bryanwweber: Do you think it makes sense to implement a `values()` method that returns an iterable of the `self.options` values? Or, why not just use
```python
for value in self.options.values():
    message.append(value.help(env=env))
```
Then there's no need to add the `__iter__` method.

site_scons/buildutils.py
bryanwweber: And I think the typing would be `choices: Iterator[str] = ()`. I'm only just now starting to grok the difference between `Iterator` and `Iterable`... For example, this will pass mypy, but is presumably not what we want:

```python
def f(n: Iterable[str]):
    for q in n:
        print(q)

f("asdf")  # this is not a type error
f(("asdf", "asdf"))  # neither is this
f((1, 2, 3))  # this is an error, which is correct
```

But switching to `Iterator` gives the behavior we want:
```python
def f(n: Iterator[str]):
    for q in n:
        print(q)

f("asdf")  # this is an error now
f(("asdf", "asdf"))  # this is not an error
f((1, 2, 3))  # this is an error
```

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+        """Returns itself as an iterator."""
+        for k in self.options.keys():
+            yield k
+
+    def select(self, key: str = "default") -> None:
+        """Select attribute dictionary entries corresponding to *key*"""
+        for attr, item in self.options.items():
+
+            if isinstance(item.default, dict) and key in item.default:
+                self.options[attr].default = item.default[key]
```
bryanwweber: Isn't `self.options[attr]` the same thing as `item`? So you could do `item.default = item.default[key]`? Then you can simplify the loop to be just over the `values()` and eliminate `attr`.

site_scons/buildutils.py
bryanwweber: Although it may be simpler to just return the empty structure in `self.choices`

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+    def to_rest(self, dev: bool = False, indent: int = 3) -> str:
+        """Convert description of option to restructured text (reST)"""
+        # assemble output
+        tag = self.name.replace("_", "-").lower()
+        if dev:
+            tag += "-dev"
+        out = f".. _{tag}:\n\n"
+        out += f"{self._build_title(backticks=True)}"
+        out += f"{self._build_description(indent=indent)}\n"
+        out += f"{self._build_defaults(self.default, backticks=True)}"
```
bryanwweber, line 250: The type of `self.default` (from the `__init__()`) doesn't match the type of the `defaults` argument to `_build_defaults`. In particular, `__init__()` is missing the `Dict` and `_build_defaults` is missing the `bool`.

bryanwweber (changes requested):
Thanks @ischoegl! Since we decided to keep the type hints, I did a more thorough review of those. Sorry for doubling you up with changes here 😕 I noted a one or two other minor things as well.

I'm on board with the OS check. And thanks for taking care of the `scons`/`scons help` paper cut!

site_scons/buildutils.py
ischoegl: Hm. Ended up turning on `"python.analysis.typeCheckingMode": "basic"` as this certainly goes way beyond what I've been doing elsewhere. For whatever reason `choices: Iterator[str] = ()` triggers:
```
Expression of type "tuple[()]" cannot be assigned to parameter of type "Iterator[str]"
  "tuple[()]" is incompatible with protocol "Iterator[str]"
    "__next__" is not presentPylancereportGeneralTypeIssues)
```
But `"Optional[Iterator[str]]" = None` works ...

bryanwweber: Maybe try `Sequence` instead?

site_scons/buildutils.py
bryanwweber: As discussed above, this  might need to be `Sequence` instead.

site_scons/buildutils.py
bryanwweber: Ugh, this is an open issue in the `typing` repo: https://github.com/python/typing/issues/256 I also suppose that `Sequence/Iterable[str] = ()` should fail, because that's an empty sequence, not a sequence of `str`s. Blergh, maybe the `None` sentinel is the best way to go. It avoids the mutable default argument problem, for sure :joy:

ischoegl: Well, on close inspection, the `__init__` declaration actually has a few more items:
```Python
    def __init__(self,
            name: str,
            description: str,
            default: "Union[str, bool, Dict[str, Union[str, bool]]]",
            choices: "Union[Optional[Iterator[str]], SCVariables]" = None):
```
(as you hinted at in your last comment)

site_scons/buildutils.py
ischoegl: I got rid of that helper function, as a 'correct' type-hinting string would have run 100+ characters :rofl:

site_scons/buildutils.py
ischoegl: The return values here were admittedly somewhat messy. I changed this function to avoid another unwieldy type hint.

site_scons/buildutils.py
ischoegl:
```Python
    def add(self, options: "Union[Option, Iterable[Option]]") -> None:
```

site_scons/buildutils.py
ischoegl: Yeah. I had this as `None` before ... tried to avoid another lengthy type hint, but it's probably the lesser of two evils

SConstruct
```diff
@@ -35,6 +35,19 @@ Basic usage:
...
+
+    'scons help --list-options' - Print formatted list of available options.
+
+    'scons help --option=<opt>' - Print the description of a specific option
+                                  with name <opt>, for example
+                                  'scons help --option=prefix'
+
+    'scons help --defaults' - Print a description of default values. This can
+                              be combined with --option, for example
+                              'scons help --defaults --option=cc_flags'
```
speth: I'm a little confused by the documentation for this option. It looks like the default values are printed even without this option, but with this option, the "actual" values are suppressed and the defaults for all platforms are shown.

site_scons/buildutils.py
ischoegl: `Sequence` indeed!

site_scons/buildutils.py
```diff
@@ -34,6 +36,409 @@
...
+        # assemble output
+        out = f"{self._build_title(backticks=False, indent=2)}"
+        out += self._build_description(backticks=False, indent=4)
+        out += self._build_defaults(self.default, backticks=False, indent=2, hanging=2)
+
+        if env is None:
+            return out
+
+        # Get the actual value in the current environment
+        actual = env.subst(f"${self.name!s}") if self.name in env else None
```
speth: Is this equivalent?
```suggestion
        actual = env[self.name] if self.name in env else None
```

site_scons/buildutils.py
ischoegl: No. This would break things.

speth (commented):
I like a lot of what this is accomplishes in terms of addressing Cantera/cantera-website#26. I had just a couple of small comments -- I think @bryanwweber's review of the implementation is detailed enough that I wouldn't have much useful to add.

SConstruct
ischoegl: Ok - I see what you're saying. The `--defaults` flag gives values for all build platforms, whereas `--options` gives current values (which may be the default or a user-specified override), 

I believe most users won't need `--defaults` (it was just something that naturally emerged when implementing the reST output). As it may confuse some, I'm not listing it (similar to the `--restructured-text` and `--dev` options). I'd still like to keep it, as it should be useful for debugging purposes.

ischoegl: To better document the hidden `scons help` features, I'm adding the following comment
```
# Note that 'scons help' supports additional command options that are intended for
# internal use (debugging or reST parsing of options) and thus are not listed above:
#  --defaults ... list default values for all supported platforms
#  --restructured-text ... format configuration as reST
#  --dev ... add '-dev' to reST output
#  --output=<fname> ... send output to file (reST only)
```

ischoegl:
@bryanwweber and @speth ... thanks for the reviews! I believe I took care of everything.

Also, thank you @bryanwweber for giving me a crash course on PEP 484. Although I've been using PEP 484 annotations for a while, I was apparently shockingly ignorant about some of the features. What is submitted now, passes Pylance's muster (no complaints on the added sections with `basic` type-hinting checks). Had to change a couple of things to avoid overly long type hints, but I think that the code quality benefited from those changes.

_PS: ugh ... fixing PEP 484 broke the actual build :rofl: (which for whatever reason still works on my machine ... may be due to a pretty old version of `scons` I am using)_

ischoegl:
Looks like I was overly aggressive when removing `Configuration.__iter__` :joy: ... it's useful when checking for the existence of keys as in `if key in config:` in `SConstruct` ... but things work now.

In addition, adding a couple of `scons help` commands to one of the CI runners (I picked `Build docs`) will prevent inadvertent breakages. I also added an upload of `Build docs` artifacts (commit 03b4243) just in case - those are kept for 2 days.

site_scons/buildutils.py
bryanwweber: How would it break things? This seems to work:

```ipython
In [1]: class Option:
   ...:     def __init__(self, default):
   ...:         self.default = default
   ...:
In [4]: class Config:
   ...:     def __init__(self, options):
   ...:         self.options = options
   ...:     def assign_default(self):
   ...:         for item in self.options.values():
   ...:             if isinstance(item.default, dict):
   ...:                 item.default = item.default["a"]
In [5]: c = Config({"asdf": Option("value"), "bdfg": Option({"a": "default"})})
In [6]: c.assign_default()
In [9]: c.options["bdfg"].default
Out[9]: 'default'
```

At least, that's what I would expect to happen. Is `self.options` something more complicated?

site_scons/buildutils.py
ischoegl: Thought I had this one, but didn't. Now fixed.

site_scons/buildutils.py
ischoegl: Odd. What you said made sense, but I could swear that last time I ended up with an error (not every time, but some times). Presumably the other changes made this possible? Anyhow, in this case I'm adopting (and force-pushing again) ...

ischoegl:
:tada: ... I think this (finally?) takes care of everything. Found another instance of an issue with type hints (which percolated through a couple of things), but that's now taken care of. @bryanwweber ... all PEP 484 updates are squashed in 2ccda5376147d532a3efd6c5ac20c2f85f99cebe.

As an aside, `SCons` and `buildutils` aren't resolved in `SConstruct` and raise a `reportMissingImports` flag in Pylance, causing *a lot* of 'Problems' to be reported. I don't think that I want to address this pre-existing condition in this PR :joy:

_PS: looks like I jinxed it - and had to force-push yet another time. At this point, I_ think _it is done for real._

site_scons/buildutils.py
ischoegl: yes

bryanwweber:
> As an aside, `SCons` and `buildutils` aren't resolved in `SConstruct` and raise a `reportMissingImports` flag in Pylance, causing _a lot_ of 'Problems' to be reported. I don't think that I want to address this pre-existing condition in this PR 😂

Yeah, this is because SCons has a lot of "magic" things that are automatically imported for you when it runs. It basically has to be addressed at the Pylance level, as far as I can tell, since those things don't get imported in SConstruct/SConscript.

ischoegl:
> Yeah, this is because SCons has a lot of "magic" things that are automatically imported […]

That’s what I figured but I was hoping that there may be a clever way to account for this without having to tinker with whatever delinter/etc. someone may be using.

Regardless, looks like I learned a bit about PEP 484 in this PR … it was definitely worthwhile. In hindsight knowing about some details I know now would have made me avoid a couple of pitfalls. Fanciful helper functions are certainly something to stay away from, but it certainly makes for cleaner code. Also, enabling Pylance’s type hinting capabilities did clarify a couple of things.

That said, I think I’m happy with this as it stands … I think that apart from the reST, the added options for `scons help` should be beneficial down the line.

bryanwweber (approved):
Looks great @ischoegl! Thanks for the quick efforts to fix this up. It was likewise educational for me to look all this stuff up.

PR #1141, opened 2021-11-21
Update handling of critical properties

speth:

- Use critical properties stored in the species `critical-parameters` YAML field to compute Redlich-Kwong and Peng-Robinson coefficients
- Use the acentric factor from the `critical-properties YAML field in the high-pressure gas transport model
- Replace `critProperties.xml` with `critical-properties.yaml`. Entries in the new file are consistent with YAML species definitions.

I believe this last change has the distinction of being the last place in the code where the XML format was being used without a YAML alternative. This therefore allows us to start making "noisy" deprecation warnings about CTI and XML.

This resolves a secondary point that was raised in Cantera/enhancements#20, and further discussed in Cantera/enhancements#107.

A species entry in a YAML input file:
```yaml
- name: H2
  composition: {H: 2}
  thermo:
    model: NASA7
    temperature-ranges: [200.0, 1000.0, 3500.0]
    data:
    - [2.34433112, 0.00798052075, -1.9478151e-05, 2.01572094e-08, -7.37611761e-12,
      -917.935173, 0.683010238]
    - [3.3372792, -4.94024731e-05, 4.99456778e-07, -1.79566394e-10, 2.00255376e-14,
      -950.158922, -3.20502331]
  critical-parameters:
    critical-temperature: 32.938
    critical-pressure: 1.2858 MPa
    acentric-factor: -0.22
```

An entry from `critical-properties.yaml`:
```yaml
- name: CS2
  descriptive-name: carbon-disulfide
  molecular-weight: 76.131
  critical-parameters:
    critical-temperature: 552
    critical-pressure: 7.9e+06
    critical-molar-volume: 0.160
    critical-compressibility: 0.276
    acentric-factor: 0.109
```

**Checklist**


src/thermo/PengRobinson.cpp
```diff
@@ -348,119 +348,82 @@ bool PengRobinson::addSpecies(shared_ptr<Species> spec)
...
-            spCoeff[0] = omega_a * (GasConstant * GasConstant) * (T_crit * T_crit) / P_crit; //coeff a
-            spCoeff[1] = omega_b * GasConstant * T_crit / P_crit; // coeff b
-            spCoeff[2] = w_ac; // acentric factor
-            break;
-        }
-    }
-    // If the species is not present in the database, throw an error
-    if(isnan(spCoeff[0]))
-    {
+    AnyMap data = AnyMap::fromYamlFile("critical-properties.yaml");
```
ischoegl: While this isn’t an approach that was introduced here, this means that the entire file is being parsed any time coefficients are needed? Is there a more efficient way to accomplish this (e.g. by buffering the file content the first time the file is read?)

src/thermo/RedlichKwongMFTP.cpp
```diff
@@ -572,70 +591,20 @@ void RedlichKwongMFTP::getSpeciesParameters(const std::string& name,
...
-                throw CanteraError("RedlichKwongMFTP::getCoeff",
-                                   "Critical Pressure not in database ");
-            }
-
-            //Assuming no temperature dependence
-            spCoeff[0] = omega_a * pow(GasConstant, 2) * pow(T_crit, 2.5) / P_crit; //coeff a
-            spCoeff[1] = omega_b * GasConstant * T_crit / P_crit; // coeff b
-            break;
-        }
+    AnyMap data = AnyMap::fromYamlFile("critical-properties.yaml");
```
ischoegl, line 619: Same here.

ischoegl (commented):
@speth … this looks overall very good. The only thing I could see is that the current (and previous!) implementation requires repeated parsing of a large file (6000+ lines for YAML). Would it make sense to read the file once and buffer either everything or only the subset of species that are used by a thermo model (which may, however, not be known at the time this is needed … ). A middle ground would be to introduce a `PengRobinson::getCoeffs(const std::vector<std::string>& iNames)` so everything can be done in one sweep if so required.

ischoegl (commented):
The only other improvement I can see is to use preferred IUPAC names (PIN’s), as the `descriptive-name` choices are somewhat arbitrary and probably do not consider various isomers (e.g. `octane` presumably refers to ‘isooctane’, which however is `2,2,4-trimethylpentane`). I am not necessarily suggesting to add this here, but just to prepare for this capability by using fields for `iupac-name` and another one for synonyms). Down the line, this may allow for disambiguation of alternative species names a mechanism may be using.

it looks like there are several packages that support lookup, e.g. [PubChemPy](https://github.com/mcs07/PubChemPy/blob/master/docs/source/index.rst) or [ChemSpiPy](https://github.com/mcs07/ChemSpiPy/blob/master/docs/index.rst) … there are likely others

speth:
> The only other improvement I can see is to use preferred IUPAC names (PIN’s), as the `descriptive-name` choices are somewhat arbitrary and probably do not consider various isomers (e.g. `octane` presumably refers to ‘isooctane’, which however is `2,2,4-trimethylpentane`). I am not necessarily suggesting to add this here, but just to prepare for this capability by using fields for `iupac-name` and another one for synonyms). Down the line, this may allow for disambiguation of alternative species names a mechanism may be using.

I chose the name `descriptive-name` for this information (copied directly from `critProperties.xml`) specifically because the information available doesn't uniformly follow the IUPAC convention. This way, if someone in the future wants to augment this dataset, they can add an `iupac-name` field or a `smiles` field. I'm open to a different suggestion, as long as it doesn't involve updating information for all ~700 species in this dataset.

ischoegl:
> This way, if someone in the future wants to augment this dataset, they can add an `iupac-name` field or a `smiles` field. 

Certainly understand, but as implemented, the somewhat arbitrary choices for `name` and `descriptive-name` are a limitation to this approach. I guess this is an enhancement request ...

src/thermo/PengRobinson.cpp
speth: This isn't quite as bad as it looks -- `AnyMap::fromYamlFile` caches the `AnyMap` generated for any given input file, and only re-reads it from disk if it's been modified. Changing this is a bit of scope creep for this PR, but I think solving it (for both P-R and R-K) won't be too complicated.

ischoegl: If the resulting `AnyMap` is indeed cached (i.e. `yaml-cpp` does not have to parse again), then this observation could be laid to rest by including a comment at the two instances where the `AnyMap::fromYamlFile` is called (i.e. no other action needed).

decaluwe:
FWIW, I think that is sort of the point of the `descriptive-name` tag - an acknowledgment that it is not quite an ideal solution (no pun intended),  but a placeholder until the longer-term project of adding `iupac` or `smiles` entries is implemented.  But I don't know it's fair to put that task on this PR (which I think we all agree with, from the comments above).

Certainly worth waiting to see if this is in much demand, before undertaking the significant amount of work it would require.

speth:
> > This way, if someone in the future wants to augment this dataset, they can add an `iupac-name` field or a `smiles` field.
> 
> Certainly understand, but as implemented, the somewhat arbitrary choices for `name` and `descriptive-name` are a limitation to this approach. I guess this is an enhancement request ...

I think the underlying issue is that most user input files come from Chemkin-format input files where there is no canonical species name, so matching the `name` is really the best we can do, until someone (RMG?) starts generating Cantera YAML files directly with additional species identifiers.

I'd say that the bigger feature of this PR (besides freeing us from XML) is defining a home for the critical state properties in user input files, so they can be used with any real gas model without having to compute the model-specific parameters manually. I'm not as sure about how useful this particular database is, but I think it's probably at least worth keeping around.

ischoegl (approved):
@speth ... thanks for updating! This looks good to me.

PR #1143, opened 2021-11-24
Modify third body efficiencies using modify_reaction

lavdwall:

Until now, when modifying a reaction using the `modifyReaction` function in C++ (or `modify_reaction` from Python), only the rates are modified, not the third-body efficiencies. Although the composition map that stores the efficiencies can be modified from Python (using get/set function `efficiencies`), these modifications are not taken into account when consequently calculating the reaction rates.

In this pull request, the functionality of `modifyReaction` is extended to also include modifications to third-body efficiencies. This is possible with only a few additions to `GasKinetics.cpp` and `ThirdBodyCalc.h`

For testing, I modified the `add_falloff_reaction` test in `kineticsFromScratch.cpp`: I added a reaction with wrong third-body-efficiencies first, verified that this gives wrong results, then used `modifyReaction` to correct the third-body efficiencies and verify the result is as expected.

**Checklist**

### DIFF
--- a/include/cantera/kinetics/ThirdBodyCalc.h
+++ b/include/cantera/kinetics/ThirdBodyCalc.h
@@ -32,6 +32,22 @@ public:
             m_eff.back().push_back(eff.second - dflt);
         }
     }
+    
+    void modifyEfficiencies(size_t rxnNumber, const std::map<size_t, double>& enhanced,
+                 double dflt=1.0) {
+        auto it = find(m_reaction_index.begin(), m_reaction_index.end(), rxnNumber);
+        if (it != m_reaction_index.end()) {
+            int index = it - m_reaction_index.begin();  
+            for (const auto& eff : enhanced) {
+                auto si = find(m_species[index].begin(), 
+                               m_species[index].end(), eff.first);   
+                if (it != m_reaction_index.end()) {
+                    int j = si - m_species[index].begin();
+                    m_eff[index][j] = (eff.second - m_default[index]);
+                }
+            }
+        }
+    }
 
     void update(const vector_fp& conc, double ctot, double* work) {
         for (size_t i = 0; i < m_species.size(); i++) {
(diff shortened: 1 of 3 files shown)


ischoegl:
Hi @lavdwall  ... thanks for the pull request! While I believe the objective of modifying the collision efficiencies is well conceived, there has been some ongoing work on the code base, which unfortunately affects your work. Cantera 2.6 will introduce significant changes in the way reactions and associated rates are handled (see Cantera/enhancements#87, a substantial portion was recently merged with #1101). Specifically, your PR targets `ThreeBodyReaction2`, together with `GasKinetics::modifyThreeBodyReaction` and member variables `GasKinetics::m_3b_concm` and `GasKinetics::m_falloff_concm`, all of which are de-facto deprecated and replaced by new handlers. At this point, anything loaded from YAML input bypasses these methods.

The way things are handled going forward involves a shared pointer `Reaction::thirdBody()` which is only initialized for `ThreeBodyReaction3` and `FalloffReaction3`. Similarly, testing should involve `kineticsFromScratch3.cpp`. As an aside, the internal extension `3` is used for the new framework; everything marked with `2` will be deprecated in 2.6, and removed for the major 3.0 release.

lavdwall:
Hm, I see.. Thanks for the clarification! I tested it two weeks ago on a mechanism loaded from YAML and back then everything still worked correctly, but that was before #1101 :)
I will look at the details of the new implementation. Is it worth it if I look at implementing this feature in the new framework, or is the possibility to modify third-body efficiencies there by default? If it is not worth it, I will just continue with my old version of code for now.

ischoegl:
Yes, the changes hit `main` very recently (at least for `Falloff`).

> Is it worth it if I look at implementing this feature in the new framework, or is the possibility to modify third-body efficiencies there by default? 

Good question: the setters all appear to be there, but that is somewhat misleading: similar to the handling of reaction rates, things won't get updated for `GasKinetics` internally once a reaction is added (which is why `modify_reaction` is important, which however lacks the ability to change efficiencies). So what you have is still relevant; note that there now is a `ThirdBodyCalc3` ...

ischoegl:
Fwiw, Cantera/enhancements#133 may implicitly address this issue.

PR #1047, opened 2021-05-28
Implementation of Peng-Robinson EoS (Revised)

gkogekar:
This is a more cleaned up version of the Peng-Robinson EoS implementation. 
(Original PR: https://github.com/Cantera/cantera/pull/641).

PR #1144, opened 2021-11-25
Add GH Actions runners for mingw and Intel compiler

### LINKED ISSUE #117: Add MinGW runner to GH Actions
**Abstract/Motivation**

While the official Cantera documentation lists [MinGW as one of the compilers](https://cantera.org/compiling/installation-reqs.html#windows-requirements) users can use on windows, compilation using MinGW is currently not tested using GitHub actions, where regressions, bugs and/or issues caused by compiler quirks are not caught in time.

**Possible Solutions**

There appear to be several GH actions that support `mingw` on the Actions marketplace.

**References**

* recent [UG discussion](https://groups.google.com/g/cantera-users/c/7TsfXeWGIy8)
* GH [Actions Marketplace](https://github.com/marketplace?type=actions&query=mingw+OR+msys2+)

### LINKED ISSUE #120: Add ICC Runner to GH Actions
**Abstract**

While the official Cantera documentation lists ICC (Intel compilers) as one of the compilers users can use, compilation using ICC is currently not tested using GitHub actions, where regressions, bugs and/or issues caused by compiler quirks are not caught in time.

**Possible Solutions**

Intel OneAPI compilers are freely (?) avaliable, see [Intel OneAPI plug](https://www.scivision.dev/intel-oneapi-github-actions/)
or [project using ICC runner](https://github.com/plumed/plumed2/blob/master/.github/workflows/ci.yml). There is likely more out there.

**References**

* [Intel OneAPI plug](https://www.scivision.dev/intel-oneapi-github-actions/)
* [project using ICC runner](https://github.com/plumed/plumed2/blob/master/.github/workflows/ci.yml)

### LINKED ISSUE #1150: Optimized Cantera build with the LLVM-based Intel compilers do not work correctly
I know the [compilation instructions recommend against the Intel compilers](https://cantera.org/compiling/special-cases.html), but the now-called classic Intel compilers `icc` and `icpc` have worked for us over the past year. However, the new LLVM-based compilers `icx` and `icpx` in the Intel oneAPI 2021 release are causing issues for us. 

When the `optimize=y` flag is passed to scons and the `icx`, `icpx`, and `ifort` compilers are used, all of the thermodynamic routines return nonsense results:

```
>>> gas2 = cantera.Solution('gri30.cti')
>>> gas2.partial_molar_cp
array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0.])
>>> gas2.partial_molar_entropies
array([-1.84618157e-12,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06])
```

Rebuilding with `optimize=n` gives logical results:

```
>>> gas.partial_molar_cp
array([28850.78499483, 20786.1565461 , 21900.310829  , 29388.07113248, 
       29877.9662108 , 33596.45144474, 34929.99364444, 42458.34282025, 
       20837.772867  , 29175.01279838, 35041.92327681, 33796.1755915 , 
       38476.66517763, 35760.53544168, 29143.05598224, 37217.74697861, 
       34620.17089036, 35443.46293798, 48473.38874297, 37990.95184739, 
       44154.86358601, 42047.04769266, 44134.49205447, 42846.9589907 , 
       43054.5662317 , 50667.04415886, 52719.37349824, 48539.96252103, 
       51898.37630042, 58510.53771334, 20786.15654538, 29192.46893339, 
       33869.64884343, 35699.25554991, 34695.31836395, 29858.13985786, 
       37234.10116356, 38710.67916751, 33901.48770831, 29158.80082924, 
       35925.14870757, 38337.31040521, 49223.39253868, 46913.04968495, 
       46367.88873358, 46574.66814261, 40180.03990305, 29075.48227818, 
       20786.15654538, 71585.98993627, 73950.0234428 , 55131.64313731, 
       55506.3227774 ])
>>> 
```

Building with `g++`, `gcc`, and `ifort` with `optimize=y` gives the right answers, as does `icpc`, `icc`, and `ifort`. 

We can deal with the workaround of using the GNU compilers for C and C++ or the classic Intel compilers, but I thought I'd bring it to your attention because the classic compilers will be retired and mixing the GNU compilers and Intel is messy on many of the HPC machines we use, so the workaround is not always trivial to implement.

ischoegl:

- Add GH runner for `mingw` toolchain (on Windows)
- Add GH runner for `icc` compiler (on Ubuntu)
- Add minor updates for compiler flags
- New intel compilers `icx`/`icpx` are recognized, ~but not tested (to be more exact: compilation works, but tests segfault)~ … compilers work as long as optimization is disabled.
- Add GH runner for `icx` compiler (on Ubuntu)

Closes Cantera/enhancements#117, closes Cantera/enhancements#120, closes #1150.

**Checklist**

**Other**

FWIW, 'out-of-the-box', the following tests are failing:

For `mingw`:
```
Failed tests:
    - thermo: cubicSolver_Test.solve_cubic
```

For `icc`:
```
Failed tests:
    - VCS-LiSi-verbose
    - cxx-kinetics1
```
 ... *while `icc` unit tests can be addressed, no action is taken as `icc` wil be replaced by a new compiler generation `icx` (which does not need modifications)*

Also, I created a docker container for testing, see [here](https://github.com/ischoegl/cantera-docker/blob/main/cantera-intel/Dockerfile)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -372,3 +372,51 @@ jobs:
         run: scons test show_long_tests=yes verbose_tests=yes --debug=time
         env:
           GITHUB_ACTIONS: "true"
+
+  windows-mingw:
+    name: mingw on Windows, Python 3.8
+    runs-on: windows-2019
+    env:
+      BOOST_ROOT: ${{github.workspace}}/3rdparty/boost
+      BOOST_URL: https://pilotfiber.dl.sourceforge.net/project/boost/boost/1.75.0/boost_1_75_0.7z
+    steps:
+    - uses: actions/checkout@v2
+      name: Checkout the repository
+      with:
+        submodules: recursive
+    - name: Set Up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+        architecture: x64
+    - name: Install Python dependencies
+      run: |
+        python -m pip install -U pip 'setuptools>=47.0.0,<48'
+        python -m pip install scons pypiwin32 numpy ruamel.yaml cython h5py pandas pytest pytest-github-actions-annotate-failures
+    - name: Restore Boost cache
+      uses: actions/cache@v2
+      id: cache-boost
+      with:
+        path: ${{env.BOOST_ROOT}}
+        key: boost
+    - name: Set up MinGW
+      uses: egor-tensin/setup-mingw@v2
+      with:
+        platform: x64
+    - name: Install Boost Headers
+      if: steps.cache-boost.outputs.cache-hit != 'true'
+      run: |
+        BOOST_ROOT=$(echo $BOOST_ROOT | sed 's/\\/\//g')
+        mkdir -p $BOOST_ROOT
+        curl --progress-bar --location --output $BOOST_ROOT/download.7z $BOOST_URL
+        7z -o$BOOST_ROOT x $BOOST_ROOT/download.7z -y -bd boost_1_75_0/boost
+        mv $BOOST_ROOT/boost_1_75_0/boost $BOOST_ROOT/boost
+        rm $BOOST_ROOT/download.7z
+      shell: bash
+    - name: Build Cantera
+      run: scons build -j2 boost_inc_dir=%BOOST_ROOT% debug=n VERBOSE=y
+        python_package=full env_vars=PYTHONPATH,GITHUB_ACTIONS
+        toolchain=mingw f90_interface=n --debug=time
+      shell: cmd
+    - name: Test Cantera
+      run: scons test show_long_tests=yes verbose_tests=yes --debug=time
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -418,6 +418,50 @@ jobs:
       run:
         python3 `which scons` test show_long_tests=yes verbose_tests=yes --debug=time
 
+  linux-intel-oneapi-classic:
+    name: intel-oneAPI classic on Ubuntu, Python 3.8
+    runs-on: ubuntu-latest
+    env:
+      INTEL_REPO: https://apt.repos.intel.com
+      INTEL_KEY: GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+    steps:
+    - name: Intel Apt repository
+      timeout-minutes: 1
+      run: |
+        wget ${INTEL_REPO}/intel-gpg-keys/${INTEL_KEY}
+        sudo apt-key add ${INTEL_KEY}
+        rm ${INTEL_KEY}
+        echo "deb ${INTEL_REPO}/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
+        sudo apt-get update
+    - name: Install Intel oneAPI
+      timeout-minutes: 5
+      run: |
+        sudo apt-get install intel-oneapi-compiler-fortran intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic \
+        intel-oneapi-mpi intel-oneapi-mpi-devel intel-oneapi-mkl ninja-build libboost-dev
+    - uses: actions/checkout@v2
+      name: Checkout the repository
+      with:
+        submodules: recursive
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+        architecture: x64
+    - name: Upgrade pip
+      run: python3 -m pip install -U pip setuptools wheel
+    - name: Install Python dependencies
+      run: python3 -m pip install ruamel.yaml scons numpy cython h5py pandas pytest
+        pytest-github-actions-annotate-failures
+    - name: Setup Intel oneAPI environment
+      run: |
+        source /opt/intel/oneapi/setvars.sh
+        printenv >> $GITHUB_ENV
+    - name: Build Cantera
+      run: python3 `which scons` build env_vars=all CC=icc CXX=icpc FORTRAN=ifort -j2 debug=n --debug=time
+    - name: Test Cantera
+      run:
+        python3 `which scons` test show_long_tests=yes verbose_tests=yes --debug=time
+
   windows-mingw:
     name: mingw on Windows, Python 3.8
     runs-on: windows-2019
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -373,6 +373,51 @@ jobs:
         env:
           GITHUB_ACTIONS: "true"
 
+  # Adapted from https://www.scivision.dev/intel-oneapi-github-actions/
+  linux-intel-oneapi:
+    name: intel-oneAPI on Ubuntu, Python 3.8
+    runs-on: ubuntu-latest
+    env:
+      INTEL_REPO: https://apt.repos.intel.com
+      INTEL_KEY: GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+    steps:
+    - name: Intel Apt repository
+      timeout-minutes: 1
+      run: |
+        wget ${INTEL_REPO}/intel-gpg-keys/${INTEL_KEY}
+        sudo apt-key add ${INTEL_KEY}
+        rm ${INTEL_KEY}
+        echo "deb ${INTEL_REPO}/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
+        sudo apt-get update
+    - name: Install Intel oneAPI
+      timeout-minutes: 5
+      run: |
+        sudo apt-get install intel-oneapi-compiler-fortran intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic \
+        intel-oneapi-mpi intel-oneapi-mpi-devel intel-oneapi-mkl ninja-build libboost-dev
+    - uses: actions/checkout@v2
+      name: Checkout the repository
+      with:
+        submodules: recursive
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+        architecture: x64
+    - name: Upgrade pip
+      run: python3 -m pip install -U pip setuptools wheel
+    - name: Install Python dependencies
+      run: python3 -m pip install ruamel.yaml scons numpy cython h5py pandas pytest
+        pytest-github-actions-annotate-failures
+    - name: Setup Intel oneAPI environment
+      run: |
+        source /opt/intel/oneapi/setvars.sh
+        printenv >> $GITHUB_ENV
+    - name: Build Cantera
+      run: python3 `which scons` build env_vars=all CC=icx CXX=icpx FORTRAN=ifx -j2 debug=n --debug=time
+    - name: Test Cantera
+      run:
+        python3 `which scons` test show_long_tests=yes verbose_tests=yes --debug=time
+
   windows-mingw:
     name: mingw on Windows, Python 3.8
     runs-on: windows-2019
--- a/SConstruct
+++ b/SConstruct
@@ -437,11 +437,16 @@ config_options = [
         """Enable extra compiler optimizations specified by the
            'optimize_flags' variable, instead of the flags specified by the
            'no_optimize_flags' variable.""",
-        {"icx": False, "default": True}),
+        True),
     Option(
         "optimize_flags",
         "Additional compiler flags passed to the C/C++ compiler when 'optimize=yes'.",
-        {"cl": "/O2", "gcc": "-O3 -Wno-inline", "default": "-O3"}),
+        {
+            "cl": "/O2",
+            "icx": "-O3 -fp-model precise", # cannot assume finite math
+            "gcc": "-O3 -Wno-inline",
+            "default": "-O3",
+        }),
     Option(
         "no_optimize_flags",
         "Additional compiler flags passed to the C/C++ compiler when 'optimize=no'.",
@@ -473,7 +478,6 @@ config_options = [
            of Cantera (for example, excluding code in the 'ext' directory).""",
         {
             "cl": "/W3",
-            "icx": "-Wall -Wno-tautological-constant-compare",
             "default": "-Wall",
         }),
     Option(
--- a/SConstruct
+++ b/SConstruct
@@ -212,7 +212,6 @@ config_options = [
            optimization level.""",
         {
             "cl": "/MD /nologo /D_SCL_SECURE_NO_WARNINGS /D_CRT_SECURE_NO_WARNINGS",
-            "icc": "-vec-report0 -diag-disable 1478",
             "clang": "-fcolor-diagnostics",
             "default": "",
         }),
@@ -443,6 +442,7 @@ config_options = [
         "Additional compiler flags passed to the C/C++ compiler when 'optimize=yes'.",
         {
             "cl": "/O2",
+            "icc": "-O3 -fp-model precise",
             "icx": "-O3 -fp-model precise", # cannot assume finite math
             "gcc": "-O3 -Wno-inline",
             "default": "-O3",
--- a/SConstruct
+++ b/SConstruct
@@ -227,8 +227,8 @@ config_options = [
         """Set this to the directory where Cantera libraries should be installed.
            Some distributions (for example, Fedora/RHEL) use 'lib64' instead of 'lib'
            on 64-bit systems or could use some other library directory name instead of
-           'lib', depending on architecture and profile (for example, Gentoo 'libx32' on
-           x32 profile). If the user didn't set the 'libdirname' configuration
+           'lib', depending on architecture and profile (for example, Gentoo 'libx32'
+           on x32 profile). If the user didn't set the 'libdirname' configuration
            variable, set it to the default value 'lib'""",
         "lib", PathVariable.PathAccept),
     EnumOption(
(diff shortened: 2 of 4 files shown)


ischoegl:
@bryanwweber / @speth ... I think this is as far as I can go with reasonable effort.

I had a closer look at the failing tests, and believe the only reason were minor numerical differences between Intel MKL (?) calculations and what is used elsewhere. I loosened the tolerances for `cxx-kinetics1`(same 'looser' tolerance that is used in other tests), and skipped the verbose test `VCS-LiSi-verbose` for lack of better options (verbose output is generated from a different execution branch when equilibrating, resulting in a different log output).

One thing that I briefly looked at, but decided to leave for a later point is `pch` (precompiled headers), which had already been disabled. I realize that I am not familiar enough with `scons` to find the correct flags; as far as I was able to find out, `scons` should actually recognize and use the correct options for `pch`. But at least for `icc` I am stuck with a `Warning #672: the command line options do not match those used when precompiled header file "src/pch/system.pchi" was created`. 

Beyond, I put in some rudimentary information related to Intel's new `icx`/`icpx`/`ifx` compilers. While cantera builds (even with `pch`!) things segfault during testing.

ischoegl:
I had a half-finished `mingw` GH runner laying around, and just gave it another shot - as it appears to work, I am adding it here. 

@gkogekar ... interestingly, one of the `solveCubic` tests failed: see band-aid in commit d8b3fc8. If I interpret this correctly, a different root is found for `mingw` which may or may not create issues for `PengRobinson`.

ischoegl:
Well ... I ended up digging a little more, and realized that `warning #2259` (and some related warnings I saw on `icpx`) boil down to the following minimally reproducible example:
```c++
// file answer.cpp: place in root cantera folder and compile via
// $ icpc -o answer.o -c -Iinclude/cantera/ext -std=c++11 -Wcheck answer.cpp
// $ icpc -o answer answer.o build/ext/fmt/src/format.os
#include <iostream>
#include <fmt/printf.h>

int main()
{
    std::cout << fmt::sprintf("The answer is %g", 42.) << std::endl;
    std::cout << fmt::sprintf("The answer is %2d", 42) << std::endl; // either line creates the issue
}
```

Resulting in _long_ warnings (repeated 5(!) times):
```
include/cantera/ext/fmt/printf.h(484): warning #2259: non-pointer conversion from "int" to "char" may lose significant bits
        c = it != end ? *it : 0;
          ^
          detected during:
            instantiation of "OutputIt fmt::v6::basic_printf_context<OutputIt, Char>::format() [with OutputIt=std::back_insert_iterator<fmt::v6::internal::buffer<char>>, Char=char, ArgFormatter=fmt::v6::printf_arg_formatter<fmt::v6::buffer_range<char>>]" at line 177
            instantiation of "void fmt::v6::internal::printf(fmt::v6::internal::buffer<Char> &, fmt::v6::basic_string_view<Char>, fmt::v6::basic_format_args<Context>) [with Char=char, Context=fmt::v6::printf_context]" at line 613
            instantiation of "std::__cxx11::basic_string<Char, std::char_traits<Char>, std::allocator<Char>> fmt::v6::vsprintf(const S &, fmt::v6::basic_format_args<fmt::v6::basic_printf_context_t<fmt::v6::type_identity_t<Char>>>) [with S=fmt::v6::basic_string_view<char>, Char=char]" at line 630
            instantiation of "std::__cxx11::basic_string<Char, std::char_traits<Char>, std::allocator<Char>> fmt::v6::sprintf(const S &, const Args &...) [with S=char [17], Args=<double>, Char=char]" at line 9 of "answer.cpp"
```
Offending lines are:
https://github.com/Cantera/cantera/blob/7f350156ae545465576a5c05bb0874520d1178e6/src/base/ctml.cpp#L49
https://github.com/Cantera/cantera/blob/7f350156ae545465576a5c05bb0874520d1178e6/src/base/xml.cpp#L439
https://github.com/Cantera/cantera/blob/7f350156ae545465576a5c05bb0874520d1178e6/src/thermo/HMWSoln.cpp#L4360-L4361
https://github.com/Cantera/cantera/blob/7f350156ae545465576a5c05bb0874520d1178e6/include/cantera/numerics/ResidEval.h#L137
(among others)

I figure it's worth fixing the root cause, rather than disabling the warning ...

_PS: found a better way of handling this (using `Wall` rather than `Wcheck`). The offending code probably should be replaced eventually, but it isn’t urgent. At least the warnings coming out of CTI/XML should be gone before long._

ischoegl:
Ad `icc`/`icpc`: changing the `Wcheck` (perform compile-time code checking for certain code) to `Wall` (enable warning and error diagnostics) option gets rid of most compiler warnings that are due to `fmt`. The same isn't true for `icx`/`icpx` (new Intel compilers), where both `fmt` 6.2.1 and 8.0.1 cause issues. (_Edit: fwiw, I reported this as fmtlib/fmt#2619, which however is already closed_)

ischoegl:
It looks like I figured out `icc`'s PCH flags after all: it takes some relatively intrusive updates in `*.cpp` files as the initial header sections need to be identical (I updated some files where I expected improvements due to the inclusion of `AnyMap`; for where this isn’t done, there are numerous `Warning #672 / #673` which are presumably harmless). However, I don't see any (significant) impact on compilation times (at least for `icc`). Leaving this as is in case anyone wants to play with this.

*Edit: I dropped the commit that tested PCH flags for `icc`, as it is probably not worth pursuing a fix for a toolchain that will be replaced by a new set of compilers.*

SConstruct
```diff
@@ -663,6 +671,19 @@ if "help" in COMMAND_LINE_TARGETS:
...
+    default="cantera.conf", action="store", help="Cantera configuration file")
+cantera_conf = GetOption("cantera-conf")
+if cantera_conf != "cantera.conf":
+    logger.info(f"Using alternate configuration file '{cantera_conf}'")
+    if Path("cantera.conf").is_file():
+        logger.info("Saving existing 'cantera.conf' as 'cantera.conf.bkp'")
+        sh_copy("cantera.conf", "cantera.conf.bkp")
+    # replace cantera.conf
+    sh_copy(cantera_conf, "cantera.conf")
+
```
bryanwweber: If you're allowing the option of using an alternate configuration file, why require having a `cantera.conf` file at all? Why not just read the file passed here?

ischoegl: Having multiple `cantera.conf` was what made me implement this. Funny enough, I barely ever used it, so I'm starting to seriously question whether this is worth keeping in this PR :smile:

bryanwweber: That's fine too 😄  I think if it's going to stay, `cantera_conf` should get passed into `Variables()` below, then you don't need to copy the user-specified file into a file called `"cantera.conf"`.

ischoegl: gone!

ischoegl:
@bryanwweber ... I think #1150 shed a few insights here. I dropped some of the commits that were not essential, and added 'experimental' support for `icx`: as long as optimization is disabled, things work. With that, I believe what is here is ready for review.

tpg2114:
Just a thought -- you might not need to relax the tolerance on `cxx-kinetics1` with the classic Intel compilers if the `fp-model strict` flag was added to that compiler set too. I don't know how important the tolerance and accuracy is for that case, but it might be worth checking (and deciding on importance).

ischoegl:
@bryanwweber and @tpg2114 ... as the compilation issues are apparently not compiler related (see new issue #1155), I believe that Intel OneAPI can be added to the list of fully supported compilers at this point (as long as the required optimization settings are applied). I consolidated all commits and replaced the GH runner for `icc` by one for `icx`, where the previous 'fixes' for tolerances are no longer needed. As compilation with `icx` works (including pch support), this PR will close #1150, with the `-ffast-math` becoming a separate issue (which may or may not be worth pursuing).

I left the `mingw` runner in, but can drop it (or make a separate PR) in case the one failing test (where a band-aid is applied) needs further investigations.

.github/workflows/main.yml
```diff
@@ -372,3 +372,96 @@ jobs:
...
+  linux-intel-oneapi:
+    name: intel-oneAPI on Ubuntu, Python 3.8
+    runs-on: ubuntu-latest
+    steps:
+    - name: Intel Apt repository
+      timeout-minutes: 1
+      run: |
+        wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+        sudo apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+        rm GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
```
bryanwweber: You can add an `env` key to the job to store the name of the `.PUB` file. Then it'll be easier to update to new versions of the API.

.github/workflows/main.yml
```diff
@@ -372,3 +372,96 @@ jobs:
...
+        echo "deb https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
+        sudo apt-get update
+    - name: Install Intel oneAPI
+      timeout-minutes: 5
+      run: |
+        sudo apt-get install intel-oneapi-compiler-fortran intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic \
+        intel-oneapi-mpi intel-oneapi-mpi-devel intel-oneapi-mkl ninja-build
```
bryanwweber:
```suggestion
        sudo apt-get install intel-oneapi-compiler-fortran intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic \
        intel-oneapi-mpi intel-oneapi-mpi-devel intel-oneapi-mkl ninja-build libboost-devel
```

Might as well save the time of a whole 'nother round of `apt-get`.

.github/workflows/main.yml
```diff
@@ -372,3 +372,96 @@ jobs:
...
+      run: python3 -m pip install -U pip setuptools wheel
+    - name: Install Python dependencies
+      run: python3 -m pip install ruamel.yaml scons numpy cython h5py pandas pytest
+        pytest-github-actions-annotate-failures
+    - name: Setup Intel oneAPI environment
+      run: |
+        source /opt/intel/oneapi/setvars.sh
+        printenv >> $GITHUB_ENV
+    - name: Build Cantera
+      run: python3 `which scons` build CC=icx CXX=icpx FORTRAN=ifx -j2 debug=n --debug=time
```
bryanwweber: I'm surprised you don't need to pass more environment variables through SCons here to get everything to work. As I recall, last I looked at this some 10 years ago, it was most convenient to set `env_vars='all'` to get the right shell variables through.

SConstruct
```diff
@@ -212,7 +212,6 @@ config_options = [
            optimization level.""",
         {
             "cl": "/MD /nologo /D_SCL_SECURE_NO_WARNINGS /D_CRT_SECURE_NO_WARNINGS",
-            "icc": "-vec-report0 -diag-disable 1478",
```
bryanwweber, line 215: I'd prefer if we didn't drop the `icc` stuff, especially since folks are likely to keep using it for some time to come, even though it's been deprecated.

SConstruct
```diff
@@ -778,6 +787,9 @@ elif env["CC"] == "cl": # Visual Studio
 elif "icc" in env.subst("$CC"):
     config.select("icc")
 
+elif "icx" in env.subst("$CC"):
+    config.select("icx")
+
```
bryanwweber, line 793: I know you didn't add this particular edge-case, but it occurs to me that if someone has the string `icx` somewhere in their path to the compiler, this case (or another one) may be incorrectly selected even if they didn't set `CC=icx`. I'm not sure how to best resolve this though, and it hasn't been a problem up to now, so maybe it doesn't need to be resolved.

interfaces/cython/SConscript
```diff
@@ -70,6 +70,9 @@ if (
 if localenv["HAS_CLANG"] and py_version_short == parse_version("3.8"):
     localenv.Append(CXXFLAGS='-Wno-deprecated-declarations')
 
+if localenv["CC"] == "icc":
+    localenv.Append(CPPDEFINES={"CYTHON_FALLTHROUGH":" __attribute__((fallthrough))"})
+
```
bryanwweber: Related to my previous comment, here the comparison is equality, so if someone sets `CC=/path/to/icc` this won't be picked up. This is why I'm not sure how to resolve the edge case 😄

test/thermo/cubicSolver_Test.cpp
```diff
@@ -99,7 +99,11 @@ TEST_F(cubicSolver_Test, solve_cubic)
     //Find cubic roots
     nSolnValues = peng_robinson_phase->solveCubic(Tcrit, pCrit, a_coeff, b_coeff, alpha * a_coeff, Vroot);
     EXPECT_NEAR(expected_result[2], Vroot[0], 1.e-6);
-    EXPECT_NEAR(nSolnValues, -2, 1.e-6);
+    if (nSolnValues < 0) {
+        EXPECT_NEAR(nSolnValues, -2, 1.e-6);
+    } else {
+        EXPECT_NEAR(nSolnValues, 2, 1.e-6);
```
bryanwweber: I assume this is the case that is added for `mingw`? It'd be nice to leave a comment here explaining what's going on, since it's not been covered by a test.

bryanwweber (changes requested):
Thanks @ischoegl! A few small comments here.

speth:
> I left the `mingw` runner in, but can drop it (or make a separate PR) in case the one failing test (where a band-aid is applied) needs further investigations.

I would suggest leaving the `mingw` compiler in, but I think the inconsistent return value arising from `MixtureFugacityTP` shouldn't be ignored by just making the test succeed if the result is either +2 or -2. The `PengRobinson` and `RedlichKwong` models both use this value and do not treat the positive and negative return values equally. I can create a new issue for this so it doesn't get lost.

For this PR, I would suggest skipping this test if the `__MINGW32__` preprocessor define is present, e.g. something like:
```c++
#ifdef __MINGW32__
TEST_F(cubicSolver_Test, solve_cubic_DISABLED)
#else
TEST_F(cubicSolver_Test, solve_cubic)
#endif
```

SConstruct
ischoegl: At this point, the `-vec-report0` triggers warnings, whereas `-diag-disable 1478` has no effect ... I mainly removed it as it appears to be lint left over from years ago. In other words, the `icc` compilation works just fine without those flags, as the old options were mainly for generating reports and disabling of warnings.

SConstruct
ischoegl: I agree, but I'd prefer to leave this as is, as this just follows what is done elsewhere.

interfaces/cython/SConscript
ischoegl: Me neither ... let's wait for an issue report (if it ever arises)

.github/workflows/main.yml
ischoegl: I'm just using the standard SCons options here. Let's see whether adding `env_vars='all'` changes anything ...

test/thermo/cubicSolver_Test.cpp
ischoegl: Adopting @speth's suggestion

interfaces/cython/SConscript
ischoegl: Changing to `if "icc" in localenv["CC"]:` presumably takes care of the path issue, but the bigger one remains ... if I understand correctly, this is something that applies to *any* test for any string in an environment variable ...

.github/workflows/main.yml
ischoegl: ... it doesn't. For the time being, the variables are added, but it doesn't appear to make a difference.

ischoegl:
@bryanwweber ... thanks for the review! 

@speth ... thanks for the suggestion!

> The `PengRobinson` and `RedlichKwong` models both use this value and do not treat the positive and negative return values equally. I can create a new issue for this so it doesn't get lost.

I excluded the test as suggested, but am not familiar enough to create an issue report. I tried tagging @gkogekar as this presumably is an edge case introduced by the `PengRobinson` PR #1047. FWIW, this is the output for the time being
```
[----------] 1 test from cubicSolver_Test
[ RUN      ] cubicSolver_Test.solve_cubic_DISABLED
[       OK ] cubicSolver_Test.solve_cubic_DISABLED (3 ms)
[----------] 1 test from cubicSolver_Test (3 ms total)
```

ischoegl:
@bryanwweber ... you had a point when you mentioned that `icc` is likely going to be used for a while, so I tried @tpg2114's suggestion to disable fast math even for that compiler. This indeed took care of tolerances, and I added the *OneAPI classic* runner back in.

SConstruct
ischoegl: I added the runner back in ... as you can see in https://github.com/Cantera/cantera/runs/4460844601?check_suite_focus=true there aren't any issues

bryanwweber (approved):
Thanks @ischoegl !

PR #1145, opened 2021-11-26
Add Python 3.10 to the CI runners

### LINKED ISSUE #1147: Attribute error during build
**Problem description**

Local build on Enterprise Linux fails with `Attribute Error`

**Steps to reproduce**

1. `scons-3 build system_sundials=y f90_interface=y renamed_shared_libraries=n python_package=full system_eigen=y extra_inc_dirs=/usr/include/eigen3 system_fmt=y`
2. `scons: *** [build/ext/LICENSE.txt] AttributeError : 'dict_values' object has no attribute 'get_ninfo'
scons: building terminated because of errors.`

**Behavior**

**System information**
    Cantera version: 0805854
    OS: EL 8 x86_64
    Python/MATLAB/other software versions: Python 3.6

**Extra Information**
   Error described in #1146 not present in this local build

### LINKED ISSUE #1149: scons install places some files in wrong location on latest Fedora Linux
**Problem description**
Building for latest/greatest Fedora Rawhide, scripts are installed to `/usr/local/bin` not `/usr/bin` as anticipated from building with `scons build prefix=/usr libdirname=lib64 system_sundials=y f90_interface=y renamed_shared_libraries=n python_package=full system_eigen=y extra_inc_dirs=/usr/include/eigen3 system_fmt=y`
```
Installing ck2cti script to /builddir/build/BUILDROOT/cantera-2.6.0-0.1.a3.fc36.x86_64/usr/local/bin
Installing ck2yaml script to /builddir/build/BUILDROOT/cantera-2.6.0-0.1.a3.fc36.x86_64/usr/local/bin
Installing cti2yaml script to /builddir/build/BUILDROOT/cantera-2.6.0-0.1.a3.fc36.x86_64/usr/local/bin
Installing ctml2yaml script to /builddir/build/BUILDROOT/cantera-2.6.0-0.1.a3.fc36.x86_64/usr/local/bin
Installing ctml_writer script to /builddir/build/BUILDROOT/cantera-2.6.0-0.1.a3.fc36.x86_64/usr/local/bin
```
```
    File not found: /builddir/build/BUILDROOT/cantera-2.6.0-0.1.a3.fc36.x86_64/usr/bin/ck2cti
    File not found: /builddir/build/BUILDROOT/cantera-2.6.0-0.1.a3.fc36.x86_64/usr/bin/ck2yaml
    File not found: /builddir/build/BUILDROOT/cantera-2.6.0-0.1.a3.fc36.x86_64/usr/bin/cti2yaml
    File not found: /builddir/build/BUILDROOT/cantera-2.6.0-0.1.a3.fc36.x86_64/usr/bin/ctml2yaml
    File not found: /builddir/build/BUILDROOT/cantera-2.6.0-0.1.a3.fc36.x86_64/usr/bin/ctml_writer
```
see buildlog at https://download.copr.fedorainfracloud.org/results/fuller/Cantera/fedora-rawhide-x86_64/03000194-cantera/builder-live.log.gz

Files install properly for earlier versions, e.g. current Fedora 35: https://download.copr.fedorainfracloud.org/results/fuller/Cantera/fedora-35-x86_64/03000194-cantera/builder-live.log.gz

**Steps to reproduce**

1. On current Fedora Rawhide with up-to-date system, `scons build prefix=/usr libdirname=lib64 system_sundials=y f90_interface=y renamed_shared_libraries=n python_package=full system_eigen=y extra_inc_dirs=/usr/include/eigen3 system_fmt=y`
2. `scons install`
3. Observe that scripts are placed in wrong directory

**Behavior**
Based on `scons build prefix=/usr`, expect `ck2cti` and other scripts mentioned above to be placed at `/usr/bin` not at `/usr/local/bin`

**System information**

- Cantera version: d5e075ac1e687df0133dc3854b2052419a1748cf
- OS: Fedora Linux - 36/rawhide
- Python/MATLAB/other software versions: python 3.10.0-4.fc36, python3-scons 4.3.0-1.fc36
(Successful build on Fedora 35 is scons 4.2, also python 3.10)

bryanwweber:

- [x] Add Python 3.10 to the CI runners
- [x] Clean up the workflow file
- [x] Use the new Codecov action instead of the Bash script. The Bash script is deprecated and possible insecure.
- [x] Add OpenMP to test the samples
- [ ] ~Add Fedora test runner and fix #1149~
- [x] Fix #1147 
- [X] Replace SUNDIALS 5.7 with 5.8
- [x] Add the clang compiler as a job? Clang-9 is installed on Ubuntu 18.04 runners

I decided not to look into #1149 at this time because I suspect that's related to `setuptools` and how it processes arguments. I think that needs some more discussion in the issue before we try to tackle a fix for it.

Otherwise, I think the changes above are fairly self-explanatory. The only fly in the ointment is the change in coverage, due to using `gcovr` to collect coverage from the runs, rather than letting CodeCov handle it magically. 🤷 The reason seems to be how branches are counted. Doesn't seem to be much to do, and we only use the coverage metric as a relative number anyways (as in, did it go up or down for a given PR), so the change here is unfortunate but not terrible.

**Checklist**

As discussed in [#1075](https://github.com/Cantera/cantera/pull/1075#issuecomment-980061291)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -262,7 +262,7 @@ jobs:
         shell: bash -l {0}
     strategy:
       matrix:
-        sundials-ver: [ 2, 3, 4, 5.7 ]
+        sundials-ver: [ 2, 3, 4, 5.8 ]
       fail-fast: false
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -112,20 +112,33 @@ jobs:
       run: python3 -m pip install -U pip setuptools wheel
     - name: Install Python dependencies
       run: python3 -m pip install ruamel.yaml scons numpy cython h5py pandas scipy pytest
-        pytest-github-actions-annotate-failures
+        pytest-github-actions-annotate-failures gcovr
     - name: Build Cantera
       run: |
         python3 `which scons` build blas_lapack_libs=lapack,blas coverage=y \
-        optimize=n skip_slow_tests=y no_optimize_flags=-DNDEBUG env_vars=all -j2 --debug=time
+        optimize=n skip_slow_tests=y no_optimize_flags='-DNDEBUG -O0' \
+        FORTRANFLAGS='-O0' env_vars=all -j2 --debug=time
     - name: Test Cantera
       run:
         python3 `which scons` test show_long_tests=yes verbose_tests=yes --debug=time
-      env:
-        GITHUB_ACTIONS: "true"
+    - name: Process coverage files
+      run: |
+        gcovr --root . --exclude-unreachable-branches --exclude-throw-branches \
+        --exclude-directories '\.sconf_temp' --exclude-directories 'build/ext$' \
+        --exclude '.*ext.*' --exclude '(.+/)?_cantera\.cpp$' --exclude '^test.*' \
+        --xml coverage.xml --html-details htmlcoverage.html --txt
+    - name: Archive code coverage results
+      uses: actions/upload-artifact@v2
+      with:
+        name: code-coverage-report
+        path: htmlcoverage*
+        retention-days: 5
     - name: Upload Coverage to Codecov
-      run: bash <(curl -s https://codecov.io/bash)
-      env:
-        CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+      uses: codecov/codecov-action@v2
+      with:
+        verbose: true
+        files: ./coverage.xml
+        fail_ci_if_error: true
 
   docs:
     name: Build docs
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [ '3.6', '3.7', '3.8', '3.9' ]
+        python-version: [ '3.6', '3.9', '3.10' ]
         os: ['ubuntu-18.04', 'ubuntu-20.04']
       fail-fast: false
     steps:
@@ -36,7 +36,7 @@ jobs:
     - name: Install Apt dependencies
       run: |
         sudo apt update
-        sudo apt install libboost-dev gfortran
+        sudo apt install libboost-dev gfortran libopenmpi-dev
     - name: Upgrade pip
       run: python3 -m pip install -U pip setuptools wheel
     - name: Install Python dependencies
@@ -47,15 +47,13 @@ jobs:
     - name: Test Cantera
       run:
         python3 `which scons` test show_long_tests=yes verbose_tests=yes --debug=time
-      env:
-        GITHUB_ACTIONS: "true"
 
   macos-multiple-pythons:
     name: macOS with Python ${{ matrix.python-version }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [ '3.7', '3.8', '3.9' ]
+        python-version: [ '3.7', '3.9', '3.10' ]
         os: ['macos-10.15']
       fail-fast: false
     steps:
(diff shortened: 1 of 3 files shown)


ischoegl (commented):
@bryanwweber ... thanks for your work on this! 

Most of the updates here look straight-forward, but the coverage is indeed puzzling. Some of the coverage percentages went down significantly (e.g. `Units.h`, presumably from 100% to 70%). As I'm not familiar with the details, one question I have is whether the new coverage tests include both C++ and Python test suites? 

One thing that's conceivable is to split the PR and make a separate PR for the coverage changes, as the rest looks good to me.

bryanwweber:
> As I'm not familiar with the details, one question I have is whether the new coverage tests include both C++ and Python test suites?

It should, or at least, that shouldn't be any different than before. My interpretation of the reason for the difference is that the old Bash Codecov script counted code branches one way (perhaps by ignoring them?) and `gcovr` counts each `if`/`for`/`while` statement as a separate branch, and not all of them are covered. This is why the "Partials" count in the coverage summary goes from zero to +2500.

> One thing that's conceivable is to split the PR and make a separate PR for the coverage changes, as the rest looks good to me.

Any reason for that suggestion? The coverage change will be the same, and this PR doesn't change any files that are counted in the coverage.

ischoegl:
> It should, or at least, that shouldn't be any different than before. My interpretation of the reason for the difference is that the old Bash Codecov script counted code branches one way (perhaps by ignoring them?) and `gcovr` counts each `if`/`for`/`while` statement as a separate branch, and not all of them are covered. This is why the "Partials" count in the coverage summary goes from zero to +2500.

Good. I'm still wondering what the huge difference is ... the partials only cover about a third of the lines that are now reported as uncovered (i.e. almost 4000 lines are not accounted for by this). One thing that worked in the past is that C++ code could be implicitly covered by Python tests, and it would be pretty bad if this ends up what's happening here. I don't understand enough about `gcovr` to have a good opinion; at the same time, it's certainly a good idea to move away from something that's cryptic and deprecated.

> Any reason for that suggestion? The coverage change will be the same, and this PR doesn't change any files that are counted in the coverage.

The switch to `gcovr` appears to be a major change of a very important CI job, and blending this together with routine maintenance may not be ideal. If the loss of coverage is 'real', it would be a pretty serious concern; I do understand that, technically, coverage shouldn't change, but some parts that are covered may not be recognized as such that were before.

*PS:* It appears that the number of lines 'found' by `gcovr` makes up for the difference (4000+ lines?!), which still doesn't make full sense to me.

speth:
> Good. I'm still wondering what the huge difference is ... the partials only cover about a third of the lines that are now reported as uncovered (i.e. almost 4000 lines are not accounted for by this).

I think these other 4000 lines are the C++ test suite itself, which was previously being counted in the coverage, but is now being excluded. @bryanwweber and I discussed this, and we thought it made sense to exclude the test code itself, because it was a huge source of misleading "partial" coverage, since the branch where each test fails is never run, and worrying about the coverage of the test suite itself seems a bit too meta.

> The switch to `gcovr` appears to be a major change of a very important CI job, and blending this together with routine maintenance may not be ideal. If the loss of coverage is 'real', it would be a pretty serious concern; I do understand that, technically, coverage shouldn't change, but some parts that are covered may not be recognized as such that were before.

I think this is pretty clearly just a difference in what's being reported -- we're running all the same tests as before, so the code that is actually getting run is the same. I don't really see the point of separating this into separate PRs -- the changes to other CI jobs obviously can't affect what happens in the "coverage" job, and I don't see why there would be any concern that the two other changes to the build scripts would affect coverage either.

ischoegl:
> I think these other 4000 lines are the C++ test suite itself, which was previously being counted in the coverage, but is now being excluded.

That makes a *lot* of sense. If that's all this is, then all's good. 

Regarding separating PR's, I saw a possible hangup about coverage, but now that this is clarified my questions are resolved!

ischoegl (commented):
Based on my understanding, this looks good from my side.

speth (approved):
Looks good to me as well.

PR #1151, opened 2021-12-04
Conditional rate updates for MultiRate reactions

speth:

- Use the `RateData::update` method (executed once per reaction type) to determine whether or not both `updateFromStruct` and `evalFromStruct` need to be called for all reactions of that type.
- Remove other conditions that could cause evaluation of `BulkKinetics::m_bulk_rates` to be skipped incorrectly

The timing results from `custom_reactions.py` show an improvement compared to the "legacy" framework, with the new framework now outperforming the "legacy" one.

- Current `main` branch:
  - New framework (YAML): 257.54 ms (T_final=2736.60)
  - One Python reaction: 273.18 ms (T_final=2736.60) ... +6.07%
  - Old framework (XML): 253.31 ms (T_final=2736.60) ... -1.64%

- This PR:
  - New framework (YAML): 248.20 ms (T_final=2736.60)
  - One Python reaction: 268.20 ms (T_final=2736.60) ... +8.06%
  - Old framework (XML): 254.35 ms (T_final=2736.60) ... +2.48%

**Checklist**

### DIFF
--- a/include/cantera/kinetics/MultiRate.h
+++ b/include/cantera/kinetics/MultiRate.h
@@ -35,6 +35,7 @@ public:
     virtual void add(const size_t rxn_index, ReactionRate& rate) override {
         m_indices[rxn_index] = m_rxn_rates.size();
         m_rxn_rates.emplace_back(rxn_index, dynamic_cast<RateType&>(rate));
+        m_shared.invalidateCache();
     }
 
     virtual bool replace(const size_t rxn_index, ReactionRate& rate) override {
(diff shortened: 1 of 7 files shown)


ischoegl (commented):
@speth … this looks like a really good solution to the update process. Thanks for looking into this, as this was one of the items that needed to be taken care of. Hadn’t thought of moving the update checks into `ReactionData` before seeing this, but I like this solution quite a bit, as it appears to resolve multiple issues once and for all.

ischoegl (approved):
Now that `ReactionData` has become a lot *smarter*, it may make sense to have a pure abstract parent class/struct for documentation purposes, where each of the current classes is a specialization. I don't want to include this in this PR, but I think it would make sense to implement this prior to 2.6.

PR #1154, opened 2021-12-06
FEAT: Added additional tools for Fortran calling codes

tpg2114:
This wraps a few additional functions into Fortran that we use in our calling code, and also adds a new feature -- an error logger that does not exit from within Cantera so a calling code can handle errors that arise. Mostly useful for Fortran calling codes that cannot catch exceptions.

### DIFF
--- a/include/cantera/base/NoExitLogger.h
+++ b/include/cantera/base/NoExitLogger.h
@@ -9,23 +9,18 @@
 #include <string>
 #include "cantera/base/logger.h"
 
-using namespace std;
-
-static std::string ss = "print \"\"\" ";
-
 namespace Cantera {
-
-  /// Logger that doesn't exit when an error is thrown.
-  /// @ingroup textlogs
-  class NoExitLogger : public Logger {
-  public:
+/// Logger that doesn't exit when an error is thrown.
+/// @ingroup textlogs
+class NoExitLogger : public Logger {
+public:
     NoExitLogger() {}
     virtual ~NoExitLogger() {}
 
     virtual void error(const std::string& msg)
     {
-      std::cerr << msg << std::endl;
+       std::cerr << msg << std::endl;
     }
-  };
+};
 }
 #endif
--- a/include/cantera/base/NoExitLogger.h
+++ b/include/cantera/base/NoExitLogger.h
@@ -1,0 +1,31 @@
+//! @file NoExitLogger.h
+
+// This file is part of Cantera. See License.txt in the top-level directory or
+// at https://cantera.org/license.txt for license and copyright information.
+
+#ifndef NOEXITLOGGER_H
+#define NOEXITLOGGER_H
+
+#include <string>
+#include "cantera/base/logger.h"
+
+using namespace std;
+
+static std::string ss = "print \"\"\" ";
+
+namespace Cantera {
+
+  /// Logger that doesn't exit when an error is thrown.
+  /// @ingroup textlogs
+  class NoExitLogger : public Logger {
+  public:
+    NoExitLogger() {}
+    virtual ~NoExitLogger() {}
+
+    virtual void error(const std::string& msg)
+    {
+      std::cerr << msg << std::endl;
+    }
+  };
+}
+#endif
--- a/src/fortran/cantera_funcs.f90
+++ b/src/fortran/cantera_funcs.f90
@@ -86,10 +86,9 @@ module cantera_funcs
       ierr = ctgetCanteraError(buf)
     end subroutine ctfunc_getCanteraError
 
-    subroutine ctfunc_turnOffExitOnError(self)
+    subroutine ctfunc_turnOffExitOnError()
       implicit none
-      type(phase_t), intent(inout) :: self
-      self%err = ctturnOffExitOnError(self%thermo_id)
+      call ctturnOffExitOnError()
     end subroutine ctfunc_turnOffExitOnError
 
     subroutine ctfunc_addCanteraDirectory(self, buf)
--- a/src/fortran/cantera_funcs.f90
+++ b/src/fortran/cantera_funcs.f90
@@ -86,6 +86,12 @@ module cantera_funcs
       ierr = ctgetCanteraError(buf)
     end subroutine ctfunc_getCanteraError
 
+    subroutine ctfunc_turnOffExitOnError(self)
+      implicit none
+      type(phase_t), intent(inout) :: self
+      self%err = ctturnOffExitOnError(self%thermo_id)
+    end subroutine ctfunc_turnOffExitOnError
+
     subroutine ctfunc_addCanteraDirectory(self, buf)
       implicit none
       type(phase_t), intent(inout) :: self
--- a/src/fortran/cantera_thermo.f90
+++ b/src/fortran/cantera_thermo.f90
@@ -217,13 +217,6 @@ contains
       self%err = phase_setmassfractionsbyname(self%thermo_id, y)
     end subroutine ctthermo_setmassfractionsbyname
 
-    subroutine ctthermo_setMassFractions_NoNorm(self, x)
-      implicit none
-      type(phase_t), intent(inout) :: self
-      double precision, intent(in) :: x(self%nsp)
-      self%err = phase_setMassFractions(self%thermo_id, x, 0)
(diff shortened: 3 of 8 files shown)


include/cantera/base/NoExitLogger.h
```diff
@@ -0,0 +1,26 @@
+#ifndef NOEXITLOGGER_H
+#define NOEXITLOGGER_H
+
+#include <string>
+#include "cantera/base/logger.h"
+
+using namespace std;
```
speth: `using namespace std` should not appear in a header file (and doesn't appear to be necessary here in any case).

include/cantera/base/NoExitLogger.h
```diff
@@ -0,0 +1,26 @@
...
+
+static std::string ss = "print \"\"\" ";
```
speth: This doesn't appear to be used.

include/cantera/base/NoExitLogger.h
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Cantera {
+
+  /// Logger that doesn't exit when an error is thrown.
+  /// @ingroup textlogs
+  class NoExitLogger : public Logger {
+  public:
+    NoExitLogger() {}
```
speth: Cantera's style is not to indent inside a `namespace` directive, and then to use 4 spaces for each level after that.

src/fortran/cantera_funcs.f90
```diff
@@ -86,6 +86,12 @@ subroutine ctfunc_getCanteraError(buf)
       ierr = ctgetCanteraError(buf)
     end subroutine ctfunc_getCanteraError
 
+    subroutine ctfunc_turnOffExitOnError(self)
+      implicit none
+      type(phase_t), intent(inout) :: self
+      self%err = ctturnOffExitOnError(self%thermo_id)
+    end subroutine ctfunc_turnOffExitOnError
```
speth, line 92: I'm a little confused by the interface here - the logger is a global, so why does there need to be an argument to either `ctfunc_turnOffExitOnError` or `ctturnOffExitOnError`? And the value passed to `ctturnOffExitOnError` doesn't seem to match it's argument name ("buflen").

src/fortran/cantera_thermo.f90
```diff
@@ -217,6 +217,13 @@ subroutine ctthermo_setMassFractionsByName(self, y)
       self%err = phase_setmassfractionsbyname(self%thermo_id, y)
     end subroutine ctthermo_setmassfractionsbyname
 
+    subroutine ctthermo_setMassFractions_NoNorm(self, x)
+      implicit none
+      type(phase_t), intent(inout) :: self
+      double precision, intent(in) :: x(self%nsp)
+      self%err = phase_setMassFractions(self%thermo_id, x, 0)
+    end subroutine ctthermo_setMassFractions_NoNorm
```
speth: Is there a reason to introduce this function instead of just calling `ctthermo_setMassFractions` with the third argument set to 0?

src/fortran/cantera_thermo.f90
```diff
@@ -545,6 +552,13 @@ subroutine ctthermo_getEnthalpies_RT(self, h_rt)
       self%err = th_getenthalpies_rt(self%thermo_id, h_rt)
     end subroutine ctthermo_getenthalpies_rt
 
+    subroutine ctthermo_getGibbs_RT(self, grt)
+      implicit none
+      type(phase_t), intent(inout) :: self
+      double precision, intent(out) :: grt(self%nsp)
+      self%err = th_getpartialmolarenthalpies(self%thermo_id, grt)
+    end subroutine ctthermo_getGibbs_RT
```
speth: Presumably this should call `th_getgibbs_rt` instead of `th_getpartialmolarenthalpies`.

src/fortran/fct.cpp
```diff
@@ -614,6 +621,17 @@ extern "C" {
         return 0;
     }
 
+    status_t th_getgibbs_rt_(const integer* n,
+                             doublereal* grt) {
+      try {
+        thermo_t* thrm = _fth(n);
+        thrm->getGibbs_RT(grt);
```
speth: The `thermo_t` typedef is deprecated. You can write this more compactly as just
```suggestion
        _fth(n)->getGibbs_RT(grt);
```

src/fortran/fct.cpp
```diff
@@ -614,6 +621,17 @@ extern "C" {
         return 0;
     }
 
+    status_t th_getgibbs_rt_(const integer* n,
+                             doublereal* grt) {
```
speth: The `doublereal` typedef is deprecated. This should be just
```suggestion
    status_t th_getgibbs_rt_(const integer* n, double* grt) {
```

src/fortran/fct_interface.f90
```diff
@@ -269,6 +274,16 @@ integer function th_getpartialmolarintenergies_r(n, ie)
         double precision, intent(out) :: ie(*)
     end function th_getpartialmolarintenergies_r
 
+    integer function th_getpartialmolarcp(n, cpbar)
+      integer, intent(in) :: n
+      double precision, intent(out) :: cpbar(*)
+    end function th_getpartialmolarcp
```
speth: Please match the indentation style of the surrounding code (4 spaces per level).

speth (changes requested):
Thanks for this contribution. While I eventually expect to be able to generate much of this interface in an automated fashion (Cantera/enhancements#39), that's still in it's early stages, so updates to this interface as it exists now are welcome.

I mostly had a few formatting suggestions. The other thing that I think would be valuable is calling a few of these new functions from the existing Fortran 90 test program (`samples/f90/demo.f90`).

src/fortran/fct.cpp
tpg2114: Should the more compact way of writing it be used even though the other routines that are similar do not do so, for example in `th_getenthalpies_rt` just prior it has:

```
ThermoPhase* therm = _fth(n);
therm->getEnthalpy_RT(h_rt);
```

I am happy to update existing routines to make them use the compact style if that is the preferred approach, or to use the 2-line style in this routine to be consistent with the others.

src/fortran/fct.cpp
tpg2114: I will also update existing routines to change `doublereal` to `double` and add those to the commit.

src/fortran/fct.cpp
tpg2114: I got rid of `thermo_t` but left it as the 2-line style to conform to the similar routines around it -- I was hesitant to update all the routines since there is not any testing (that I can find) for the Fortran interface, so I didn't want to unintentionally break something. If you'd like me to update to the compact style for this routine, and/or if you'd like them all updated, let me know.

speth (commented):
Thanks for the updates -- the changes look good to me. I hadn't intended for you to have to update the style of any surrounding code, only what was new in this PR, but thanks for taking on a few of those updates anyway.

I just pushed an update to your branch to add a couple comments to the `NoExitLogger.h` header so that it will appear in the documentation and so that it has the usual license statement carried by all Cantera source files.

Two small things, which you can update at your discretion, would be (1) to add yourself to the `AUTHORS` file and (2) to update update the email address that's set for Git on your computer to one that is associated with your GitHub account so those commits will be attributed correctly. There are some hopefully helpful instructions here: https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit.

tpg2114:
No problem on updates -- I ask the folks I develop with to always leave a file better than they left it, so bringing the whole file up to date is worth the couple of extra seconds it took to find-replace things. I think the tests that failed when I first pushed the latest updates were due to the Fortran test having additional output from the calls to the new functions, and I wasn't comfortable changing the blessed solution without somebody "official" checking it out first.

speth:
Right, I forgot about the changes to the blessed output file. Since the changes are pretty simple (just some additional output lines), I would suggest copying those extra lines from `test_problems/fortran/f90_demo_output.txt` to `test_problems/fortran/f90_demo_blessed.txt`, without editing any of the other lines (which may have changed in the last decimal places, below the test's threshold).

tpg2114:
Done and pushed! Hopefully the tests will all pass now.

speth (approved):
Thanks! This looks good to me.

PR #1156, opened 2021-12-08
Use normal sign convention for reactor heat transfer

speth:
Heat transfer _into_ a system is normally considered positive, and the internal variables used within `Reactor` should reflect this convention.

This does not change the sign convention from the perspective of a wall between two reactors is unchanged, where a positive value indicates heat transfer from the left to the right, which I think is consistent.

- Calculate `Reactor::m_Qdot` using the updated sign convention, and use this to set `Reactor::m_Q` following the existing convention
- Use `Reactor::m_Qdot` in calculations for all reactor models instead of `m_Q`
- Deprecate `Reactor::m_Q`

The `Reactor::m_Q` variable is a `protected` member of the `Reactor` class, with no accessor, so this can only affect user-generated classes derived from `Reactor`. However, fixing this is important for #1003, which does introduce a getter/setter pair for this net heat transfer, and using the normal sign convention would prevent some surprising behavior.

Closes #

**Checklist**

### DIFF
--- a/include/cantera/zeroD/Reactor.h
+++ b/include/cantera/zeroD/Reactor.h
@@ -204,7 +204,14 @@ protected:
     Kinetics* m_kin;
 
     doublereal m_vdot; //!< net rate of volume change from moving walls [m^3/s]
-    doublereal m_Q; //!< net heat transfer through walls [W]
+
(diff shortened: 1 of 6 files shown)


ischoegl (approved):
Looks good to me!

PR #1102, opened 2021-09-18
[SCons] Fix stage directory from appearing in setup_cantera

speth:
**Changes proposed in this pull request**

- Fix setting the SCons variable `python_module_loc` to remove the stage directory prefix so that doesn't show up in the `setup_cantera` script or the post-install message.
- Changes how the Python module is installed to a "staging directory" to avoid the staging directory name from ending up in the `.pyc` files.

On this second point, I'm hoping we're finally using the `--root` option to `setuptools` correctly. Previously (#296) we had tried using it in some cases and not been able to get it to do what we wanted.

**If applicable, fill in the issue number this pull request is fixing**

Fixes #1094

**If applicable, provide an example illustrating new features this pull request is introducing**

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

PR #1158, opened 2021-12-14
Add package workflow

### LINKED ISSUE #1192: Applications fail to install in `bin` on Windows
**Problem description**

While the post-install message claims to install **applications** in
```
[...]
postInstallMessage(["finish_install"], [])

File locations:

  applications                C:\Users\<user>\cantera\bin
  library files               C:\Users\<user>\cantera\lib
  C++ headers                 C:\Users\<user>\cantera\include
  samples                     C:\Users\<user>\cantera\samples
  data files                  C:\Users\<user>\cantera\data
  Python package (cantera)    C:\Users\<user>\miniconda3\envs\cantera-dev\Lib\site-packages
  Python samples              C:\Users\<user>\miniconda3\envs\cantera-dev\Lib\site-packages\cantera\examples
scons: done building targets.
```
nothing is installed in `cantera\bin` (the folder is actually not even created). While this may be a limitation of the OS, the post-install message is pointing at an issue.

**Steps to reproduce**

Compile & install Cantera from source.

**System information**

- Cantera version: 2.6.0a4
- OS: Windows 10
- Python/MATLAB/other software versions: Python 3.9 from miniconda (see [toolchain](https://gist.github.com/ischoegl/ab777f871a48d7f40ecc32c1c849faa3))

**Attachments**

Initial SCons output
```
INFO: SCons is using the following Python interpreter: C:\Users\<user>\miniconda3\envs\cantera-dev\python.exe
Compiling with MSVC 14.2
Compiling for architecture: amd64
Compiling using the following toolchain(s): ['default']
INFO: Building Cantera from git commit '63b7d27de'
Configuration variables read from 'cantera.conf' and command line:
    prefix = 'C:\\Users\\<user>\\cantera'
    boost_inc_dir = 'C:/Users/<user>/miniconda3/envs/cantera-dev/Library/include'
    verbose_tests = True
```

bryanwweber:

- Trigger workflows in the Conda, PyPI, Windows, and macOS packaging repositories. The workflows are triggered on commits to `main` in this repo and to tags pushed to this repo.
- Change the Python interfaces to use `pip` as the builder and installer

**Checklist**

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -1,0 +1,48 @@
+name: Packaging Tests
+on:
+  workflow_dispatch:  # allow manual triggering of this workflow
+    inputs:
+      outgoing_ref:
+        description: "The ref to be built. Can be a tag, commit hash, or branch name"
+        required: true
+        default: "main"
+      upload_to_pypi:
+        description: "Try to upload wheels and sdist to PyPI after building"
+        required: false
+        default: false
+  push:
+    # Run on tags that look like releases
+    tags:
+      - v*
+    # Run when main is pushed to
+    branches:
+      - main
+      - add-package-workflow
+
+jobs:
+  build-packages:
+    name: Trigger building packages from external repos
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Set up the job
+        run: |
+          if [ -n "${{ github.event.inputs.outgoing_ref }}" ]; then
+            echo "REF=${{ github.event.inputs.outgoing_ref }}" >> $GITHUB_ENV
+          else
+            echo "REF=${{ github.ref }}" >> $GITHUB_ENV
+          fi
+          if [ -n "${{ github.event.inputs.upload }}" ]; then
+            echo "UPLOAD=${{ github.event.inputs.upload}}" >> $GITHUB_ENV
+          elif [[ "${{ github.ref }}" == refs/tags* ]]; then
+            echo "UPLOAD=true" >> $GITHUB_ENV
+          else
+            echo "UPLOAD=false" >> $GITHUB_ENV
+          fi
+      - name: Trigger PyPI/Wheel builds
+        run: >
+          gh workflow run -R cantera/pypi-packages
+          python-package.yml
+          -f incoming_ref=${{ env.REF }}
+          -f upload=${{ env.UPLOAD }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.PYPI_PACKAGE_PAT }}
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -9,7 +9,11 @@ on:
       upload_to_pypi:
         description: "Try to upload wheels and sdist to PyPI after building"
         required: false
-        default: false
+        default: "false"
+      upload_to_anaconda:
+        description: "Try to upload package to Anaconda after building"
+        required: false
+        default: "false"
   push:
     # Run on tags that look like releases
     tags:
@@ -31,18 +35,33 @@ jobs:
           else
             echo "REF=${{ github.ref }}" >> $GITHUB_ENV
           fi
-          if [ -n "${{ github.event.inputs.upload }}" ]; then
-            echo "UPLOAD=${{ github.event.inputs.upload}}" >> $GITHUB_ENV
+          if [ -n "${{ github.event.inputs.upload_to_pypi }}" ]; then
+            echo "UPLOAD_TO_PYPI=${{ github.event.inputs.upload_to_pypi }}" >> $GITHUB_ENV
           elif [[ "${{ github.ref }}" == refs/tags* ]]; then
-            echo "UPLOAD=true" >> $GITHUB_ENV
+            echo "UPLOAD_TO_PYPI=true" >> $GITHUB_ENV
           else
-            echo "UPLOAD=false" >> $GITHUB_ENV
+            echo "UPLOAD_TO_PYPI=false" >> $GITHUB_ENV
+          fi
+          if [ -n "${{ github.event.inputs.upload_to_anaconda }}" ]; then
+            echo "UPLOAD_TO_ANACONDA=${{ github.event.inputs.upload_to_anaconda }}" >> $GITHUB_ENV
+          elif [[ "${{ github.ref }}" == refs/tags* ]]; then
+            echo "UPLOAD_TO_ANACONDA=true" >> $GITHUB_ENV
+          else
+            echo "UPLOAD_TO_ANACONDA=false" >> $GITHUB_ENV
           fi
       - name: Trigger PyPI/Wheel builds
         run: >
           gh workflow run -R cantera/pypi-packages
           python-package.yml
           -f incoming_ref=${{ env.REF }}
-          -f upload=${{ env.UPLOAD }}
+          -f upload=${{ env.UPLOAD_TO_PYPI }}
         env:
           GITHUB_TOKEN: ${{ secrets.PYPI_PACKAGE_PAT }}
+      - name: Trigger Conda builds
+        run: >
+          gh workflow run -R cantera/conda-recipes
+          main.yml
+          -f incoming_ref=${{ env.REF }}
+          -f upload=${{ env.UPLOAD_TO_ANACONDA }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.CONDA_PACKAGE_PAT }}
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -21,7 +21,6 @@ on:
     # Run when main is pushed to
     branches:
       - main
-      - add-package-workflow
 
 jobs:
   build-packages:
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -65,3 +65,10 @@ jobs:
           -f upload=${{ env.UPLOAD_TO_ANACONDA }}
         env:
           GITHUB_TOKEN: ${{ secrets.CONDA_PACKAGE_PAT }}
+      - name: Trigger Windows MSI Builds
+        run: >
+          gh workflow run -R cantera/cantera-windows-binaries
+          main.yaml
+          -f incoming_ref=${{ env.REF }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.WINDOWS_MSI_PAT }}
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -72,3 +72,10 @@ jobs:
           -f incoming_ref=${{ env.REF }}
         env:
           GITHUB_TOKEN: ${{ secrets.WINDOWS_MSI_PAT }}
+      - name: Trigger macOS Package Builds
+        run: >
+          gh workflow run -R cantera/cantera-macos-pkg
+          main.yml
+          -f incoming_ref=${{ env.REF }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.MACOS_PKG_PAT }}
--- a/SConstruct
+++ b/SConstruct
@@ -1600,24 +1600,6 @@ if env['python_package'] != 'none':
             print('INFO: Building the full Python package for Python {0}'.format(python_version))
             env['python_package'] = 'full'
 
-if env["python_package"] == "full" and env["OS"] == "Darwin":
-    # sysconfig.get_platform() looks something like: macosx-11.0-arm64
-    # MACOSX_DEPLOYMENT_TARGET should be the same as or lower than
-    # the one set when Python was compiled to ensure ABI compatibility.
-    # If it's been set in the user's environment and passed through env_vars
-    # configuration, use that value instead.
-    if not env["ENV"].get("MACOSX_DEPLOYMENT_TARGET", False):
-        info = get_command_output(
-            env["python_cmd"],
-            "-c",
-            "import sysconfig; print(sysconfig.get_platform())"
-        )
-        env["ENV"]["MACOSX_DEPLOYMENT_TARGET"] = info.split("-")[1]
-
-    env.Append(
-        CXXFLAGS=f"-mmacosx-version-min={env['ENV']['MACOSX_DEPLOYMENT_TARGET']}"
-    )
-
 # Matlab Toolbox settings
 if env['matlab_path'] != '' and env['matlab_toolbox'] == 'default':
     env['matlab_toolbox'] = 'y'
@@ -2123,9 +2105,9 @@ def removeDirectories(target, source, env):
 uninstall = env.Command("uninstall", None, Delete(allfiles))
 env.AddPostAction(uninstall, Action(removeDirectories))
 if env["python_package"] == "full":
-    env.AddPostAction(uninstall, Action("$python_cmd_esc -m pip uninstall Cantera"))
+    env.AddPostAction(uninstall, Action("$python_cmd_esc -m pip uninstall -y Cantera"))
 elif env["python_package"] == "minimal":
-    env.AddPostAction(uninstall, Action("$python_cmd_esc -m pip uninstall Cantera_minimal"))
+    env.AddPostAction(uninstall, Action("$python_cmd_esc -m pip uninstall -y Cantera_minimal"))
 
 ### Windows MSI Installer ###
 if 'msi' in COMMAND_LINE_TARGETS:
--- a/SConstruct
+++ b/SConstruct
@@ -268,7 +268,7 @@ config_options = [
            will be installed to the system default 'site-packages' directory.
            To install to the current user's 'site-packages' directory, use
            'python_prefix=USER'.""",
-        {"Windows": "", "default": "$prefix"},
+        {"default": ""},
         PathVariable.PathAccept),
     EnumOption(
         "matlab_toolbox",
--- a/SConstruct
+++ b/SConstruct
@@ -1801,7 +1801,7 @@ configh['DARWIN'] = 1 if env['OS'] == 'Darwin' else None
 
 if env['OS'] == 'Solaris' or env['HAS_CLANG']:
     configh['NEEDS_GENERIC_TEMPL_STATIC_DECL'] = 1
-    env["RPATHPREFIX"] = "-rpath,"
+    env["RPATHPREFIX"] = "-Wl,-rpath,"
 
 configh['CT_SUNDIALS_VERSION'] = env['sundials_version'].replace('.','')
 
--- a/SConstruct
+++ b/SConstruct
@@ -1649,11 +1649,6 @@ env["default_prefix"] = True
 if "prefix" in selected_options:
     env["default_prefix"] = False
 
-# Remove back slashes from paths. For example, C:\Users results in a Unicode error
-# because \U is the prefix for a Unicode sequence. This kind of thing can happen
-# on other platforms too, so this replacement isn't conditional.
-env["prefix"] = env["prefix"].replace("\\", "/")
-
 # Check whether Cantera should be installed into a conda environment
 if conda_prefix is not None and sys.executable.startswith(conda_prefix):
     # use conda layout unless any 'blocking' options were specified
@@ -1689,6 +1684,11 @@ else:
     env["ct_incroot"] = pjoin(env["prefix"], "include")
 
 env["ct_installroot"] = env["prefix"]
+
+# Remove back slashes from paths. For example, C:\Users results in a Unicode error
+# because \U is the prefix for a Unicode sequence. This kind of thing can happen
+# on other platforms too, so this replacement isn't conditional.
+env["prefix"] = env["prefix"].replace("\\", "/")
 
 if env['layout'] == 'compact':
     env['ct_datadir'] = pjoin(env['prefix'], 'data')
--- a/SConstruct
+++ b/SConstruct
@@ -1474,6 +1474,7 @@ elif env['python_package'] == 'n':
 
 env['install_python_action'] = ''
 env['python_module_loc'] = ''
+env["ct_pyscriptdir"] = "<not installed>"
 
 if env['python_package'] != 'none':
     # Test to see if we can import numpy and Cython
@@ -2064,11 +2065,11 @@ def postInstallMessage(target, source, env):
 
         File locations:
 
-          applications                {ct_bindir!s}
           library files               {ct_libdir!s}
           C++ headers                 {ct_incroot!s}
           samples                     {ct_sampledir!s}
-          data files                  {ct_datadir!s}""".format(**env_dict))
+          data files                  {ct_datadir!s}
+          input file converters       {ct_pyscriptdir!s}""".format(**env_dict))
 
     if env['sphinx_docs'] or env['doxygen_docs']:
         install_message += indent(textwrap.dedent("""
@@ -2077,7 +2078,7 @@ def postInstallMessage(target, source, env):
     if env['python_package'] == 'full':
         env['python_example_loc'] = pjoin(env['python_module_loc'], 'cantera', 'examples')
         install_message += indent(textwrap.dedent("""
-            Python package (cantera)    {python_module_loc!s}
+            Python package              {python_module_loc!s}
             Python samples              {python_example_loc!s}""".format(**env_dict)), '  ')
     elif env['python_package'] == 'minimal':
         install_message += indent(textwrap.dedent("""
--- a/SConstruct
+++ b/SConstruct
@@ -1994,7 +1994,8 @@ VariantDir('build/src', 'src', duplicate=0)
 SConscript('build/src/SConscript')
 
 if env['python_package'] == 'full':
-    SConscript('interfaces/cython/SConscript')
+    VariantDir("build/python", "interfaces/cython", duplicate=True)
+    SConscript('build/python/SConscript')
 elif env['python_package'] == 'minimal':
     SConscript('interfaces/python_minimal/SConscript')
 
--- a/SConstruct
+++ b/SConstruct
@@ -2095,23 +2095,6 @@ def postInstallMessage(target, source, env):
               {matlab_ctpath_loc!s}
         """.format(**env_dict))
 
-    if os.name != 'nt':
-        env['setup_cantera'] = pjoin(env['ct_bindir'], 'setup_cantera')
-        env['setup_cantera_csh'] = pjoin(env['ct_bindir'], 'setup_cantera.csh')
-        install_message += textwrap.dedent("""
-
-            Setup scripts to configure the environment for Cantera are at:
-
-              setup script (bash)         {setup_cantera!s}
-              setup script (csh/tcsh)     {setup_cantera_csh!s}
-
-            It is recommended that you run the script for your shell by typing:
-
-              source {setup_cantera!s}
-
-            before using Cantera, or else include its contents in your shell login script.
-        """.format(**env_dict))
-
     print(install_message)
 
 finish_install = env.Command('finish_install', [], postInstallMessage)
--- a/SConstruct
+++ b/SConstruct
@@ -876,7 +876,7 @@ env['cantera_short_version'] = re.match(r'(\d+\.\d+)', env['cantera_version']).g
 try:
     env["git_commit"] = get_command_output("git", "rev-parse", "--short", "HEAD")
     logger.info(f"Building Cantera from git commit '{env['git_commit']}'")
-except subprocess.CalledProcessError:
+except (subprocess.CalledProcessError, FileNotFoundError):
     env["git_commit"] = "unknown"
 
 # Print values of all build options:
@@ -2135,23 +2135,6 @@ def getParentDirs(path, top=True):
 # Files installed by SCons
 allfiles = FindInstalledFiles()
 
-# Files installed by the Python installer
-if os.path.exists('build/python-installed-files.txt'):
-    with open('build/python-installed-files.txt', 'r') as f:
-        file_list = f.readlines()
-
-    install_base = os.path.dirname(file_list[0].strip())
-    if os.path.exists(install_base):
-        not_listed_files = [s for s in os.listdir(install_base) if not any(s in j for j in file_list)]
-        for f in not_listed_files:
-            f = pjoin(install_base, f)
-            if not os.path.isdir(f) and os.path.exists(f):
-                allfiles.append(File(f))
-    for f in file_list:
-        f = f.strip()
-        if not os.path.isdir(f) and os.path.exists(f):
-            allfiles.append(File(f))
-
 # After removing files (which SCons keeps track of),
 # remove any empty directories (which SCons doesn't track)
 def removeDirectories(target, source, env):
@@ -2177,6 +2160,8 @@ def removeDirectories(target, source, env):
 
 uninstall = env.Command("uninstall", None, Delete(allfiles))
 env.AddPostAction(uninstall, Action(removeDirectories))
+if env["python_package"] == "full":
+    env.AddPostAction(uninstall, Action("$python_cmd_esc -m pip uninstall Cantera"))
 
 ### Windows MSI Installer ###
 if 'msi' in COMMAND_LINE_TARGETS:
--- a/SConstruct
+++ b/SConstruct
@@ -1621,6 +1621,24 @@ if env['python_package'] != 'none':
             print('INFO: Building the full Python package for Python {0}'.format(python_version))
             env['python_package'] = 'full'
 
+if env["python_package"] == "full" and env["OS"] == "Darwin":
+    # sysconfig.get_platform() looks something like: macosx-11.0-arm64
+    # MACOSX_DEPLOYMENT_TARGET should be the same as or lower than
+    # the one set when Python was compiled to ensure ABI compatibility.
+    # If it's been set in the user's environment and passed through env_vars
+    # configuration, use that value instead.
+    if not env["ENV"].get("MACOSX_DEPLOYMENT_TARGET", False):
+        info = get_command_output(
+            env["python_cmd"],
+            "-c",
+            "import sysconfig; print(sysconfig.get_platform())"
+        )
+        env["ENV"]["MACOSX_DEPLOYMENT_TARGET"] = info.split("-")[1]
+
+    env.Append(
+        CXXFLAGS=f"-mmacosx-version-min={env['ENV']['MACOSX_DEPLOYMENT_TARGET']}"
+    )
+
 # Matlab Toolbox settings
 if env['matlab_path'] != '' and env['matlab_toolbox'] == 'default':
     env['matlab_toolbox'] = 'y'
--- a/SConstruct
+++ b/SConstruct
@@ -196,13 +196,13 @@ config_options = [
         }),
     PathOption(
         "prefix",
-        """Set this to the directory where Cantera should be installed. If the Python
+        r"""Set this to the directory where Cantera should be installed. If the Python
            executable found during compilation is managed by 'conda', the installation
            'prefix' defaults to the corresponding environment and the 'conda' layout
            will be used for installation (specifying any of the options 'prefix',
-           'python_prefix', 'python_cmd' or 'layout' will override this default). On
+           'python_prefix', 'python_cmd', or 'layout' will override this default). On
            Windows systems, '$ProgramFiles' typically refers to "C:\Program Files".""",
-        {"Windows": "$ProgramFiles\Cantera", "default": "/usr/local"},
+        {"Windows": r"$ProgramFiles\Cantera", "default": "/usr/local"},
         PathVariable.PathAccept),
     PathOption(
         "libdirname",
@@ -1645,6 +1645,15 @@ env['debian'] = any(name.endswith('dist-packages') for name in sys.path)
 selected_options = set(line.split("=")[0].strip()
     for line in cantera_conf.splitlines())
 
+env["default_prefix"] = True
+if "prefix" in selected_options:
+    env["default_prefix"] = False
+
+# Remove back slashes from paths. For example, C:\Users results in a Unicode error
+# because \U is the prefix for a Unicode sequence. This kind of thing can happen
+# on other platforms too, so this replacement isn't conditional.
+env["prefix"] = env["prefix"].replace("\\", "/")
+
 # Check whether Cantera should be installed into a conda environment
 if conda_prefix is not None:
     if env["layout"] != "conda" and sys.executable.startswith(conda_prefix):
@@ -1675,10 +1684,6 @@ if env["layout"] == "conda" and os.name == "nt":
     env["ct_incroot"] = pjoin(env["prefix"], "Library", "include")
 else:
     env["prefix"] = os.path.normpath(env["prefix"])
-    if env["prefix"] != config["prefix"].default:
-        env["default_prefix"] = False
-    else:
-        env["default_prefix"] = True
     env["ct_libdir"] = pjoin(env["prefix"], env["libdirname"])
     env["ct_bindir"] = pjoin(env["prefix"], "bin")
     env["ct_python_bindir"] = pjoin(env["prefix"], "bin")
--- a/SConstruct
+++ b/SConstruct
@@ -1994,8 +1994,9 @@ if env['f90_interface'] == 'y':
 VariantDir('build/src', 'src', duplicate=0)
 SConscript('build/src/SConscript')
 
-if env['python_package'] == 'full':
+if env["python_package"] == "full":
     VariantDir("build/python", "interfaces/cython", duplicate=True)
+    SConscript("build/python/SConscript")
 elif env["python_package"] == "minimal":
     VariantDir("build/python_minimal", "interfaces/python_minimal", duplicate=True)
     SConscript("build/python_minimal/SConscript")
--- a/SConstruct
+++ b/SConstruct
@@ -1675,6 +1675,10 @@ if env["layout"] == "conda" and os.name == "nt":
     env["ct_incroot"] = pjoin(env["prefix"], "Library", "include")
 else:
     env["prefix"] = os.path.normpath(env["prefix"])
+    if env["prefix"] != config["prefix"].default:
+        env["default_prefix"] = False
+    else:
+        env["default_prefix"] = True
     env["ct_libdir"] = pjoin(env["prefix"], env["libdirname"])
     env["ct_bindir"] = pjoin(env["prefix"], "bin")
     env["ct_python_bindir"] = pjoin(env["prefix"], "bin")
--- a/SConstruct
+++ b/SConstruct
@@ -1745,7 +1745,6 @@ if 'msi' in COMMAND_LINE_TARGETS:
     COMMAND_LINE_TARGETS.append('install')
     env['stage_dir'] = 'stage'
     env['prefix'] = '.'
-    env['PYTHON_INSTALLER'] = 'binary'
 elif env['layout'] == 'debian':
     COMMAND_LINE_TARGETS.append('install')
     env['stage_dir'] = 'stage/cantera'
--- a/SConstruct
+++ b/SConstruct
@@ -105,14 +105,6 @@ if "clean" in COMMAND_LINE_TARGETS:
     remove_file("include/cantera/base/config.h")
     remove_file("src/pch/system.h.gch")
     remove_directory("include/cantera/ext")
-    remove_file("interfaces/cython/cantera/_cantera.cpp")
-    remove_file("interfaces/cython/cantera/_cantera.h")
-    remove_file("interfaces/cython/setup.cfg")
-    remove_file("interfaces/cython/LICENSE.txt")
-    remove_file("interfaces/cython/README.rst")
-    remove_file("interfaces/python_minimal/setup.cfg")
-    remove_file("interfaces/python_minimal/LICENSE.txt")
-    remove_file("interfaces/python_minimal/README.rst")
     remove_file("config.log")
     remove_directory("doc/sphinx/matlab/examples")
     remove_file("doc/sphinx/matlab/examples.rst")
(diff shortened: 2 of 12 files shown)


bryanwweber:
@Cantera/committers I'd appreciate some eyes on this PR, as I'm getting into rewriting the Python installer logic in the SConscript. I added it to this branch because I'm hoping to use the rewritten Python installer to fix the Conda packages, and then add a job here to run the Conda builds when a PR is merged to `main`. Note the change to `.github/workflow/main.yml` to disable it on PRs is temporary for debugging here, I'll remove that change. I didn't want to run a zillion builds of Cantera when I wasn't changing any files that are part of the build :smile:

See the details section 👇 for some justification/explanation of the changes:
<details><summary>Click for more</summary>

1. I elected to use a `VariantDir` with `duplicate=True` for the Cython interface. This copies all the dependent files into the `build/python` directory, so all the building commands can be run against that single directory, for example, `pip wheel build/python` works now to create a wheel.
2. The change to a `VariantDir` also means that `setup.py build` is no longer required, as far as I can tell. Since all the files are put into the `build/python` directory by SCons, the same trick of `PYTHONPATH=build/python ...` still works as it always has. There are also no longer any built files put into the `interfaces/cython` directory, which is a nice bonus for cleanup (although I haven't changed the `scons clean` code yet).
2. I opted to use pip as the build frontend to install the Python package. This is because as of Setuptools 60.0.0, using `setup.py install` is deprecated in favor of the explicit frontend/backend separation specified in PEP 517/518. In this sense, a frontend tells the backend what to build (sdist, wheel, egg, etc.) while the backend actually creates that resource. Our backend is setuptools (and has been for a long time); prior to this change, we were also using setuptools as the frontend via `setup.py install`. As this behavior is now deprecated, I opted for `pip` as the more likely candidate to be installed on someone's machine than `build` (which is used to generate the `sdist` via `scons sdist`). This means that `pip install build/python` should also work as expected.
</details>

FYI @speth, as we discussed in Cantera/enhancements#129, this branch has an example of triggering a workflow in another repo. You made a good point that it'd be nice if we can somehow drag the status of the triggered workflow back into the status of this repo. I'm sure it's possible, I'll see how to do it. Worst case, it will be a "badge" on the README with the workflow status in the other repositories.

ischoegl:
:eyes: ... (kidding)

I am aware of what you pointed out about (3) ... have been using the `pip` route for a while elsewhere, so this is absolutely needed! Beyond, I'm probably not super-familiar with the SCons specific changes (sorry), but the direction sounds great!

Triggering Conda builds upon PR merges sounds terrific! :tada:  ... this comes very close to Cantera/enhancements#24 (which is superseded by Cantera/enhancements#37) and should also eliminate issues like Cantera/conda-recipes#24 (haven't checked recently whether this has since been resolved).

bryanwweber:
@speth / @ischoegl / @mefuller / @yurivict / @band-a-prend (I believe this is the group that maintain distro installers for Cantera) another question I had on this pull request was whether it makes sense to set the default `python_prefix` to `--user` if the system Python is set as `python_cmd` on Linux distributions. Modern practice is that it's generally discouraged to install using `pip` with `sudo` (that is, into protected directories) because the version of `pip` running with `sudo` might be different from that when run as a regular user. Obviously, if the user has selected their own `python_prefix`, that would be respected.

speth:
> another question I had on this pull request was whether it makes sense to set the default `python_prefix` to `--user` if the system Python is set as `python_cmd` on Linux distributions.

I think that seems like a reasonable default for builds by normal users. But what are you supposed to do when you _do_ want to install into the default location for that copy of Python? Would it make sense to have a different special value for the `python_prefix` SCons option to omit the `--user` flag? Also, how do you decide whether the Python that is running is the "system" Python or not?

bryanwweber:
> But what are you supposed to do when you _do_ want to install into the default location for that copy of Python? Would it make sense to have a different special value for the `python_prefix` SCons option to omit the `--user` flag?

@speth Sure, perhaps `SYSTEM` to mirror the current `USER` option.

> Also, how do you decide whether the Python that is running is the "system" Python or not?

`sys.executable` starts with `/usr/bin` was my first idea.

band-a-prend:
@bryanwweber 

As for gentoo currently any python package (i.e. PyPi package) is always installed by package manager by handling PyPi package tarball. Pip is never used directly to install packages into system as fas as I know. The build/installation proccess is going within networkless sandbox. The Using `pip` directly is avoiding.

The gentoo itself usually has available several Python3 version to be installed with on is set up as default. Currently it's 3.9 with planning to switch to 3.10 as default but user could choose one of available for system. Python 3.8 is still available too. So some packages as numpy could be installed for several versions of Python. Cantera for gentoo is installed only for one system defined python version (or user defined by setup build preferencies for specific package). 

Early python implementation in gentoo x86_64 were placed under `/usr/lib64` but a couple of years it was changed to `/usr/lib` (while in common case `/usr/lib64` is used for x86_64 libraries).  Due to this changes I have to additionally patch `interfaces/cython/SConscript` and `/interfaces/python_minimal/SConscript` with
```patch
-            ' --prefix=${{python_prefix}}'
-            ' --install-lib=${{python_prefix}}/${{libdirname}}/python{}/site-packages'.format(py_version))
+            ' --prefix=${stage_dir}${prefix} --install-lib=${python_prefix}')
```
(https://github.com/gentoo/gentoo/blob/b25ddfa29f7369a77a54158e5bff0c3473c10ef5/sci-libs/cantera/files/cantera-2.5.1_env.patch#L44
and
https://github.com/gentoo/gentoo/blob/b25ddfa29f7369a77a54158e5bff0c3473c10ef5/sci-libs/cantera/files/cantera-2.5.1_env.patch#L60) to take into account this gentoo specific feature.
With this patch the Cantera then is installed with command
```
escons install stage_dir="${D}" libdirname="$(get_libdir)" python_prefix="$(python_get_sitedir)
```
where `escons` gentoo function call from helper script and `$(...)` gentoo build env variables provides system library directory and ditectory of Python site-packages directory for system setup python version. I didn't proposed this changes into Cantera upstream as assumed this is too distro specific.

If Cantera is planning to be installed from PyPi from own tarball then it will be easy for me split gentoo package into cantera (c++, fortran) and cantera-python.

mefuller:
@bryanwweber I think I'm missing something regarding `conda` and `pip` and how these all fit in, but these are my thoughts:
- `sudo pip` is like `GOTO`: there's just a visceral negative reaction. That said, I don't see `pip` in `SConstruct`, so maybe I don't understand how the install is being handled exactly.
- I would leave the default installation as system-level and perhaps just out an error message is `scons install` is run with insufficient privilege that mentions running the installation with `--user`. If someone is at the level of doing their own build, then appropriate documentation of the default installation directory and how to change it should be adequate. I also feel like system-level as default is expected behavior when running with `su`/`sudo` and this would be blocked if the install is triggered as a normal user.

bryanwweber:
@g3bk47 / @mefuller Thanks for replying! I should provide some context and clarification. I'm referring specifically to the behavior when a user runs `scons install` in the source directory.

Python (via a package called distribute, which has been replaced by setuptools) has traditionally installed components relative to the path of the Python executable using a `setup.py` script which is passed the `install` command line argument. In our SConscript for the Cython interface, we currently execute the equivalent of `python setup.py install` when the user writes `scons install`. For system Python installations, which put their files into protected directories, this has always required the use of `sudo scons install`.

As of Setuptools 60.0.0, the behavior of installing via `python setup.py install` is deprecated. The suggested replacement is to use `pip install <path-to-folder-with-setup.py>`. I'm proposing that we do precisely this in our SConscript file for the Cython interface when the user runs `scons install`. Ideally, this will achieve the same result as `setup.py install` but without relying on deprecated CLI arguments.

The change of behavior that I'm proposing is that when the system Python is being used to install the Cython interface (detected by a `/usr/bin` prefix on `sys.executable`), that our SConscript default to including the `--user` flag to pip, which installs files into the user-local directory (by default `~/.local`). Since this directory is in the user's home directory, it would not require `sudo` to install. My question is whether you foresee this change of behavior causing any problems for users on your platforms. Thanks!

mefuller:
@bryanwweber thanks for the clarification.
I think that this all sounds fine provided that there is an easy CLI option to attach to `sudo scons install` to duplicate the current placement of the Python files in protected directories.
Packages distributed in Fedora (and I am about to shift to having Cantera in the official repos) may not place files under `/home/`.

g3bk47:
On our systems, users don’t have sudo rights. So Cantera is always installed with `scons build prefix="xxx"`. If I understand correctly, the proposed change wouldn’t affect this.

But if installing the python files to a protected directory by default is a problem, shouldn’t the default installation directory for Cantera itself be changed from `/usr/local` to somewhere in `/home?`

>The suggested replacement is to use pip install <path-to-folder-with-setup.py>

Our systems don’t have a standalone `pip`, but `python -m pip`, as used in this PR, works.

bryanwweber:
> Packages distributed in Fedora (and I am about to shift to having Cantera in the official repos) may not place files under `/home/`.

Thanks @mefuller that's good to know. This would only be the case when the user is running `scons install` from a source checkout, not installing the system package, which is managed by the system package manager.

Likewise thanks @g3bk47 good to know about that policy. Indeed, I anticipate that this change won't impact you then.

> But if installing the python files to a protected directory by default is a problem, shouldn’t the default installation directory for Cantera itself be changed from `/usr/local` to somewhere in `/home?`

`/usr/local` is the traditional directory on Linux systems to install system-wide user-installed software. Those directories, at least on Debian-based distros, are in the default search path for the compiler and linker. So I think users on gentoo will have to stick with `scons install prefix=...` which seems like an easy enough workaround.

speth:
If you're installing with `layout=standard` (the default except for Windows), the default behavior is to install the Python package under `$prefix`. So if you have specified a `prefix` in a user-writable directory, then that's where the module will end up by default. These directories are probably not on the python search path, so this is the case where it's most convenient to use the `--user` flag. But I'm not sure that's actually a very good practice, as `~/.local/lib/python3.X/site-packages` will still be on your search path when using things like a conda environment, and it seems like there's a lot of potential for getting conflicting package versions.

The only case where the build system should be trying to write to `/usr/lib/python3.X/...` is if you've specified `prefix=/usr` (mostly applicable for the scripts that build the distro-specific packages). If `prefix` is set to `/usr/local`, I think the correct assumption would be that you're an administrator doing a system-wide installation, and the Python package should go into `/usr/local/lib/python3.X/`. So for these cases, I don't think setting `--user` as a default would make sense.

I'm not sure what the most robust way of distinguishing between these two scenarios is (`$USER == root` would work frequently, but I'm not sure if that's sufficient), assuming that we want to make such a distinction.

bryanwweber:
> If `prefix` is set to `/usr/local`, I think the correct assumption would be that you're an administrator doing a system-wide installation, and the Python package should go into `/usr/local/lib/python3.X/`. So for these cases, I don't think setting `--user` as a default would make sense.

Thanks for the thoughts @speth! I agree with your distinction between intentional and unintentional installation into write-protected directories. I think if `prefix` is explicitly set to `/usr/local` then the assumption you've got here is warranted. On the other hand, since `/usr/local` is also the default `prefix`, I don't think it's correct to assume that a system-wide install of the Python interface is intended in the general case. (Side note: I do think `/usr/local` is the correct place for the library and headers.)

Recent versions of `pip` actually display a warning when they're run with `sudo`:

```bash
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
```

So I think we need to change the default behavior of the Python installation via SCons when `python_cmd` is the system Python and/or if `sys.exec_prefix` is in a write-protected location. I definitely think that, for people who know what they're doing, _allowing_ system-wide installation should be possible, but I don't think it should be the _default_.

One option would be to test if the user has write permissions in the folder where they're trying to install and error out preemptively with a useful error message (which is not "just use `sudo`" 😄 ) if they don't have permission. This might be more effective than defaulting to `--user`.

speth:
At least on the Linux distros I've used, `/usr/local` is
1. Only writable by root (Is this different on macOS?).
2. Not used by the system package manager.

And the main purpose is to provide a location for installing software system-wide when you aren't using the system package manager. So if the Cantera headers and libraries are being installed in `/usr/local`, I don't see why we wouldn't also install the Python package there.

> Recent versions of `pip` actually display a warning when they're run with `sudo`:
> 
> ```shell
> WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
> ```

This seems bizarre. As root, `pip install` installs to `/usr/local`, and everything there should be owned by root, so I don't see how you can end up with "broken permissions". And while this does mix responsibility for managing Python package versions between the system package manager and pip and some issues with keeping dependencies satsified, the same thing is true of installation with `--user` (and actually worse if you're using Conda environments, as those still leave the package directory from `--user` on your `PYTHONPATH`). Further, in this case, what's the recommended way of installing system-wide Python packages that aren't available via the system package manager? While having every user maintain their own Python installations is an option, I don't see how it should be the _only_ one.

> I definitely think that, for people who know what they're doing, _allowing_ system-wide installation should be possible, but I don't think it should be the _default_.

I don't think there's a useful default for `prefix` within the user's own space, but I guess you could make an argument for `prefix` being a required argument to `scons install` so the installation location will never be a surprise.

bryanwweber:
> At least on the Linux distros I've used, `/usr/local` is
> 
> 1. Only writable by root (Is this different on macOS?).

macOS doesn't include this folder by default, IIRC. Homebrew on the M1 installs to `/opt` now, on Intel I believe it still installs to `/usr/local` for historical reasons.

> This seems bizarre. As root, `pip install` installs to `/usr/local`, and everything there should be owned by root, so I don't see how you can end up with "broken permissions".

I guess I don't find it particularly bizarre. Pip is meant to manage user-specific software, which probably shouldn't be going into write protected directories in general. There are certainly specific cases, but the most general case is a user who doesn't quite know what they're doing, and in that case I think it's a sensible policy. One place where the broken permissions refers to is the cache, because running `pip` as root will put files into the cache which are then owned by root and can't be deleted by the user (without another `sudo`). And since `sudo` also doesn't carry over the user's `PATH`, `pip` might be referring to a different version than expected (even `python -m pip`), which can also do unexpected things. Finally, there's the case of malicious packages, since `setup.py` can run arbitrary code...

As far as sys-admin for a multi-user machine, I think the presumption is that you know what you're doing at that point, and you know enough to ignore the warning when appropriate. I think that's why it's a warning and not an error.

The most common recommendation these days to avoid having to run `pip` as root is to use virtual environments, which I guess amounts to users maintaining their own Python installations, although I think that's not really what you were referring to. I agree that `--user` is not a great choice, especially in the world of conda/virtual environments, for the reasons you've pointed out.

It seems like we're converging on having some sort of extra action required by the user running `scons install` as the solution here. I don't think requiring `prefix` to run `scons install` is a great solution, since it will recompile some stuff when/if it changes, which could be confusing. And really, the only "problem" is the Python interface installer... Maybe requiring `python_prefix` instead of `prefix`? And only if the install is going into a write-protected location? I guess it remains to be determined how 'write-protected' can be reliably identified...

speth:
> macOS doesn't include this folder by default, IIRC. Homebrew on the M1 installs to `/opt` now, on Intel I believe it still installs to `/usr/local` for historical reasons.

OK, but I assume `/` is only writable by root, in which case any invented directory at that level has the same requirement, correct?

> One place where the broken permissions refers to is the cache, because running `pip` as root will put files into the cache which are then owned by root and can't be deleted by the user (without another `sudo`).

On Linux, running `sudo pip install...` will create cache files in `/root/.cache/pip`, not the user's home directory. So I really don't see how this can be an issue.

> Pip is meant to manage user-specific software, which probably shouldn't be going into write protected directories in general.

If that's the case, then what's the recommended way to install Python packages for all users, for packages where there aren't system packages (or where you want to install a different version)?

> There are certainly specific cases, but the most general case is a user who doesn't quite know what they're doing, and in that case I think it's a sensible policy.

Honestly, on a single user system, I don't think there's much of a meaningful difference in these locations. And if someone who doesn't know what they're doing is trying to manage a multi-user system, well...good luck. But, if you want to try steering users away from using `sudo`, then that must mean installing all of Cantera (not just the Python module) to a `prefix` other than `/usr/local`.

> Finally, there's the case of malicious packages, since `setup.py` can run arbitrary code...

Running a malicious `setup.py` as a non-privileged user can still delete all of my files, and leave nothing on my computer besides a bunch of system files that are trivially re-installable. The level of trust required of PyPI packages seems like a separate problem to me.

> The most common recommendation these days to avoid having to run `pip` as root is to use virtual environments, which I guess amounts to users maintaining their own Python installations, although I think that's not really what you were referring to. I agree that `--user` is not a great choice, especially in the world of conda/virtual environments, for the reasons you've pointed out.

No, that's exactly what I was referring to. But if you think that `--user` is not a great choice, then I don't think we should be making that our default for any value of `prefix`.

> It seems like we're converging on having some sort of extra action required by the user running `scons install` as the solution here. I don't think requiring `prefix` to run `scons install` is a great solution, since it will recompile some stuff when/if it changes, which could be confusing.

Well, you could require it as an argument to `scons build`, then. One downside to that is it's pretty different from the default expectation for other software packages, like anything where you would run `./configure && make && make install` where the default would always be to install to `/usr/local`. I think the behavior of having `python_prefix` default to following `prefix` still makes sense.

bryanwweber:
@speth I don't want to reply point-by-point because I think we're bikeshedding about `pip` and not about the actual problem we've both identified.

Sorry if it wasn't clear, I agree that a bare `--user` flag to pip is a bad default for our installer, no matter the value of `prefix`. I also don't think we need to require the `prefix` argument, which would be an unnecessary burden and change the default expectations of `./configure && make && make install`. That's also why I don't think changing the default install location of the library and headers is worth the headache, there's a lot of history and most people seem fine with that convention for C/C++ code. I think, however, that the Python community is moving in a different direction.

Rereading my thoughts and considering your points for some more time, I think what I'd like to do is try to avoid UG posts about the warning message from pip from using `sudo scons install`. I think there are 2 cases we need to try to address:

1. The user runs `scons install` and `python_cmd` is `/usr/bin/python3`, such that `python_cmd -m pip install` will try to put files into write-protected directories. This, of course, will result in an error without `sudo` and the most common advice online is "just yolo and sudo".
2. The user runs `sudo scons install` and unexpectedly hits the warning about not running `pip` as the root user.

I think the solution to 1. is to check for the ability to write into the system folders and provide a proactive error message indicating that the resolution is to set `python_cmd` to a Python installed in a virtual environment, conda environment, set `python_prefix` to an accessible location, or use `sudo`.

I think the solution to 2. is to require that users set `python_prefix = /usr/local` so that we're sure the user knows what they want. Of course, we'd provide a useful error message indicating what users should do if this condition isn't satisfied.

Of the solutions here, I'm very strongly in favor of 1., which I think should be fairly uncontroversial (providing improved error messages is usually a win 😉 ). I can be talked out of 2., but I think it'd be a really good idea.

ischoegl:
@bryanwweber ... I only want to comment on one minor aspect here

> That's also why I don't think changing the default install location of the library and headers is worth the headache, there's a lot of history and most people seem fine with that convention for C/C++ code. I think, however, that the Python community is moving in a different direction.

fwiw, I believe it's to be expected to install into write-protected system folders if the 'system' Python version is used to run `scons`. However, if a user uses conda or any other virtual environment when running the installation, the __default prefix should default to whatever makes the most sense for the virtual environment__, as presumably there is an implied desire to not 'pollute' system-wide folders (i.e. a departure from past practice would make sense in this case). This should take care of at least some cases where `sudo` would otherwise be required; specifying `prefix` for those who don't want this would be the expected alternative.

*PS: As an example, following some discussion over at Cantera/conda-recipes#20, the 'expected' install location for conda would be `'/home/<user>/.conda/envs/<some_env>/include/'` etc. (for Linux) and `'C:\\Users\\<user>\\anaconda3\\envs\\<some_env>\\Library\\include'` etc. (for Windows, i.e. wherever conda/miniconda was installed)*?

speth:
@bryanwweber 

> 1. The user runs `scons install` and `python_cmd` is `/usr/bin/python3`, such that `python_cmd -m pip install` will try to put files into write-protected directories. This, of course, will result in an error without `sudo` and the most common advice online is "just yolo and sudo".

But this will only attempt to install the Python package to `/usr/local` if `prefix` isn't set (or is set to another directory not writable for unprivileged users), in which case you will also hit other errors trying to install the other files to `/usr/local`. I guess which one terminates the build will depend on the order that SCons decides to execute the installation steps. I'd certainly be 👍 on providing an error message if the installation directories aren't writable that suggests setting `prefix` to a location that can be written by the user, but just setting `python_prefix` isn't going to be sufficient.

> I think the solution to 2. is to require that users set `python_prefix = /usr/local` so that we're sure the user knows what they want. Of course, we'd provide a useful error message indicating what users should do if this condition isn't satisfied.

Are you suggesting that an explicit setting for `python_prefix` is required in this case, while still allowing `prefix` to default to `/usr/local`? If not, what other default would make sense? A different system directory somewhere? It shouldn't be a directory in the user's own space, otherwise you will end up in the situation where there are now root-owned files in the user's directories. And in any case, this still won't get rid of the obnoxious warning from `pip` (which was kind of my earlier point, but I'm happy to try to let the discussion of "what's wrong with pip" lapse).

bryanwweber:
> the **default prefix should default to whatever makes the most sense for the virtual environment**,

@ischoegl Yep, definitely agree with this. Assuming that SCons is installed into the virtual/conda environment, then `pip` will install for that Python. Otherwise, `python_cmd` will need to be set to the Python in the environment. But that's consistent with current behavior.

> I'd certainly be 👍 on providing an error message if the installation directories aren't writable that suggests setting `prefix` to a location that can be written by the user, but just setting `python_prefix` isn't going to be sufficient.

@speth Yes, agreed.

> I guess which one terminates the build will depend on the order that SCons decides to execute the installation steps.

I _think_ it should be possible to get the directory permissions via the `os` module, which means that a check could be done in SConstruct pre-install, so that it won't matter what SCons chooses to do first.

> Are you suggesting that an explicit setting for `python_prefix` is required in this case, while still allowing `prefix` to default to `/usr/local`? ... And in any case, this still won't get rid of the obnoxious warning from `pip`

Yes, that's exactly what I'm suggesting. Essentially a way of enforcing users to opt-in to installing into directories that aren't recommended, which I hope would be useful. I think the resolution of the warning message will come once a [separate Python interface installer](https://github.com/Cantera/enhancements/issues/75) is available, and we can suggest setting `python_prefix` to a user-writable location and install interfaces separately.

ischoegl:
> ... . Assuming that SCons is installed into the virtual/conda environment, then `pip` will install for that Python. Otherwise, `python_cmd` will need to be set to the Python in the environment. But that's consistent with current behavior.

@bryanwweber ... actually, not quite. While this is the current behavior for Python, this is not true for headers, etc.. I have some (Linux) development toolchains where I am using a local (user) conda installation where there is no system-wide `scons`. Yet, I would have to provide a non-protected `prefix` if I wanted to install things to override defaults (which I don't as I just test with `PYTHONPATH`).

bryanwweber:
@ischoegl Ah, I see what you're saying. There's an enhancement issue open regarding this: https://github.com/Cantera/enhancements/issues/76

ischoegl:
> There's an enhancement issue open regarding this: Cantera/enhancements#76

@bryanwweber ... I guess I have to retract some of the comments I made over there :sob:. I since have run into exactly that issue, which I believe is worth addressing.

speth:
> > Are you suggesting that an explicit setting for `python_prefix` is required in this case, while still allowing `prefix` to default to `/usr/local`? ... And in any case, this still won't get rid of the obnoxious warning from `pip`
> 
> Yes, that's exactly what I'm suggesting. Essentially a way of enforcing users to opt-in to installing into directories that aren't recommended, which I hope would be useful.

But what other directory even makes sense to install to, if you're using `/usr/bin/python3` and installing the rest of Cantera to `/usr/local`? If you're following pip's recommendation to use virtual environments, then `python3` should be a location within a `venv` or `conda` environment. By this logic, should the build just fail if `python_cmd` isn't set explicitly but evaluates to `/usr/bin/python3`? Why wait for the `install` step?

bryanwweber:
> But what other directory even makes sense to install to, if you're using `/usr/bin/python3` and installing the rest of Cantera to `/usr/local`? If you're following pip's recommendation to use virtual environments, then `python3` should be a location within a `venv` or `conda` environment. By this logic, should the build just fail if `python_cmd` isn't set explicitly but evaluates to `/usr/bin/python3`? Why wait for the `install` step?

@speth / @ischoegl OK, after reflecting this week, I think enforcing installation directories is probably a step too far. The justification for it isn't quite as strong as I thought, and you've both raised some really good points. In the end, I think I'll defer improved error messages from this PR to a later one, and just focus this one on using pip as the Python installer so the Conda packages can work again. I'm going to mark a bunch of the comments in this thread as off topic to reduce the clutter as this PR moves towards review.

SConstruct
```diff
@@ -263,7 +263,7 @@ config_options = [
            will be installed to the system default 'site-packages' directory.
            To install to the current user's 'site-packages' directory, use
            'python_prefix=USER'.""",
-        {"Windows": "", "default": "$prefix"},
+        {"default": ""},
```
bryanwweber, line 249: I think this change will work out because `pip` does the right thing at install time, whereas pure Setuptools needed a little more hand-holding. But I didn't get a chance to test on Windows, and I'd appreciate at least some ad-hoc testing on your guys' end.

SConstruct
```diff
@@ -1578,6 +1579,24 @@ if env['python_package'] != 'none':
...
+            env["python_cmd"],
+            "-c",
+            "import sysconfig; print(sysconfig.get_platform())"
+        )
+        env["ENV"]["MACOSX_DEPLOYMENT_TARGET"] = info.split("-")[1]
+
+    env.Append(
+        CXXFLAGS=f"-mmacosx-version-min={env['ENV']['MACOSX_DEPLOYMENT_TARGET']}"
+    )
+
```
bryanwweber: This is a decently significant change of behavior for compiling on macOS. As shown here, I scoped it to only be done when the Python package is being built, since that's the main case where we want to be more careful about ABI compatibility I think. But if you think this should be a more general change, that's fine too.

SConstruct
```diff
@@ -2090,6 +2090,10 @@ def removeDirectories(target, source, env):
 
 uninstall = env.Command("uninstall", None, Delete(allfiles))
 env.AddPostAction(uninstall, Action(removeDirectories))
+if env["python_package"] == "full":
+    env.AddPostAction(uninstall, Action("$python_cmd_esc -m pip uninstall Cantera"))
+elif env["python_package"] == "minimal":
+    env.AddPostAction(uninstall, Action("$python_cmd_esc -m pip uninstall Cantera_minimal"))
```
bryanwweber: I'm not convinced it's worth trying to manage an uninstall from SCons, YMMV. I didn't try either of these commands.

site_scons/buildutils.py
```diff
@@ -1193,6 +1196,45 @@ def get_command_output(cmd: str, *args: str):
     return data.stdout.strip()
 
 
+def get_pip_install_location(
```
bryanwweber, line 1209: I hope getting the module location this way doesn't break the fixes in #1102.

interfaces/cython/SConscript
```diff
@@ -89,96 +89,72 @@ elif localenv['OS'] == 'Cygwin':
...
     # Get the absolute path to the stage directory. If the stage directory is a relative
-    # path, consider it to be relative to the root of the Cantera source directory,
-    # which is two directories up from the current working directory of this SConscript
-    # file.
-    stage_absolute = Path.cwd().parents[1].joinpath(env['stage_dir']).resolve()
-    extra += f" --root={stage_absolute}"
+    # path, consider it to be relative to the root of the Cantera source directory.
+    stage_dir = Path(env["stage_dir"])
+    if not stage_dir.is_absolute():
+        stage_dir = Path(Dir("#").abspath) / stage_dir
```
bryanwweber, line 144: I think this is more robust than `Path.cwd().parents[1]` if the stage directory is relative, and that I haven't broken the fixes from #1102, but I don't have a FreeBSD machine to test. @yurivict can you confirm this works for you?

bryanwweber (commented):
@ischoegl @speth Can you guys please take this branch for a spin with the new `pip` installation mode? I also left a few comments in the code for discussion. There are a few things here that are kind of unrelated to this code that I'd be happy to put into another PR if it makes sense to get it merged faster:

* The change of `MACOSX_DEPLOYMENT_TARGET`
* The bugfix for the test writing to the local directory
* Adding `FileNotFoundError` in the check for the git commit

ischoegl:
@bryanwweber ... while I may not have much to add in terms of review comments, I just gave this a spin on windows.

Without specifying `prefix`, installation requires administrator privileges (no surprises), although I assume that most users would be ok with installing in a non-admin environment.
```
[...]
scons: done reading SConscript files.
scons: Building targets ...
scons: *** [C:\Program Files\Cantera] C:\Program Files\Cantera: Access is denied
```

Using `pip 21.2.4`, I got this (colorful) warning during compilation (removal appears to be imminent, I didn't investigate though)
```
"C:\Users\<user>\miniconda3\envs\cantera-dev\python.exe" -m pip wheel --no-build-isolation --no-deps --wheel-dir=build/python/dist build/python
Processing c:\users\<user>\github\cantera\build\python
  DEPRECATION: A future pip version will change local packages to be built in-place without first copying to a temporary directory. We recommend you use --use-feature=in-tree-build to test your packages with this new behavior before it becomes the default.
   pip 21.3 will remove support for this functionality. You can find discussion regarding this at https://github.com/pypa/pip/issues/7555.
```

With `prefix`, things run without problems, with this post-install message.
```

File locations:

  library files               C:\Users\<user>\cantera\lib
  C++ headers                 C:\Users\<user>\cantera\include
  samples                     C:\Users\<user>\cantera\samples
  data files                  C:\Users\<user>\cantera\data
  input file converters       C:\Users\<user>\miniconda3\envs\cantera-dev\Scripts
  Python package              C:\Users\<user>\miniconda3\envs\cantera-dev\Lib\site-packages
  Python samples              C:\Users\<user>\miniconda3\envs\cantera-dev\Lib\site-packages\cantera\examples
scons: done building targets.
```
All tests run through, and spot-testing the installed cantera version didn't show any problems.

SConstruct
ischoegl: It does.

bryanwweber:
> Using pip 21.2.4, I got this (colorful) warning during compilation (removal appears to be imminent, I didn't investigate though)

Thanks for doing all this @ischoegl! Especially on Windows, which I don't have access to any more ☺️ pip versions are calendar versions, where the first two digit number is the year and the second is the quarter. 21.3 was released in June (?) last year, and 22.0 is the current version 😊 I've been using 22.0 with no problem, so I can add something to suppress that warning on older pips

> With prefix, things run without problems, with this post-install message.

Out of curiosity, what `prefix` did you set? You can see there that the `Scripts` directory is correctly set for the conversion scripts 🎉

ischoegl:
np. While I do most of my stuff on Fedora (or an Ubuntu docker), I keep a windows laptop around … so this was easy. As you see, this tool chain involves `conda` (see [details](https://gist.github.com/ischoegl/ab777f871a48d7f40ecc32c1c849faa3)) so it’s just one of several ways users can be installing.

> Out of curiosity, what `prefix` did you set?

Just my home directory. 

Great that this  takes care of #1192!

interfaces/cython/SConscript
ischoegl: Fwiw, this is the exact line that gave me grief in #1191.

bryanwweber: Were you working off of this branch? This line changed in this PR

ischoegl: No. But my troubles started at the line this PR is changing (I believe that a more robust implementation is a good thing here)

ischoegl:
As a minor comment ... per #1189, there was an update you suggested to adopt here, which is removing/updating the following:
```
Setup scripts to configure the environment for Cantera are at:

  setup script (bash)         /home/<user>/.conda/envs/cantera-dev/bin/setup_cantera
  setup script (csh/tcsh)     /home/<user>/.conda/envs/cantera-dev/bin/setup_cantera.csh

It is recommended that you run the script for your shell by typing:

  source /home/ischoegl/.conda/envs/cantera-dev/bin/setup_cantera

before using Cantera, or else include its contents in your shell login script.
```

bryanwweber:
@ischoegl @speth This is _finally_ complete, and triggers builds in all the packaging repos 🎉 It also gives a blueprint for triggering builds in the Jupyter Notebooks repo whenever we get around to setting up CI for those.

As it stands, the new workflow added here will only run on commits to `main` and for tags (well, and this branch, but that line should be removed before merging). I think this makes the most sense, since we don't package all that often, so failures in packaging shouldn't hold up a PR. This _should_ post statuses to commits after they've been merged to `main` so we'll be able to see from the Cantera/cantera repo homepage the status of the most recent commit, so we'll have a heads up when packages are failing and we can fix them.

Reviews welcome 😄

bryanwweber:
> 2\. (although I haven't changed the `scons clean` code yet). (https://github.com/Cantera/cantera/pull/1158#issuecomment-1015028238)

~I think I still need to do this. I also need to re-enable the standard testing workflow.~

OK, all of this is taken care of now 😄

SConstruct
speth: First, I think if you define the `MACOSX_DEPLOYMENT_TARGET` environment variable, you don't need to specify the compiler option. If you want to use the compiler option instead, though, it should also be set for `CCFLAGS` so it will affect pure C code like SUNDIALS.

Second, while I'll admit this is as clear as mud to me, I'm not sure it's necessary for libraries that get linked together to have the same OS target version. The conda-forge maintainer docs do have a bit to say about this (https://conda-forge.org/docs/maintainer/knowledge_base.html#requiring-newer-macos-sdks), which mostly boils down to saying that they compile for 10.9 to maintain compatibility with macOS 10.9, but suggests that requiring a higher version is something that *can* be done.

Certainly, when building packages that we distribute, we should be setting MACOSX_DEPLOYMENT_TARGET to an older value than whatever it defaults to, and I think that should be done regardless of whether you're compiling the Python module (but again, I don't think anything needs to be written here to achieve that).

SConstruct
```diff
@@ -1976,10 +1973,12 @@ if env['f90_interface'] == 'y':
 VariantDir('build/src', 'src', duplicate=0)
 SConscript('build/src/SConscript')
 
-if env['python_package'] == 'full':
-    SConscript('interfaces/cython/SConscript')
-elif env['python_package'] == 'minimal':
-    SConscript('interfaces/python_minimal/SConscript')
+if env["python_package"] == "full":
+    VariantDir("build/python", "interfaces/cython", duplicate=True)
+    SConscript("build/python/SConscript")
```
speth, line 1967: I was sort of hoping we could move toward doing away with any copying of the Python module parts, since that would make it easier to do rapid iteration and interactive debugging of the pure-Python components.

interfaces/python_minimal/pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools>=43.0.0", "wheel", "Cython>=0.29.12"]
```
speth: The minimal module doesn't need Cython for anything, does it?

.github/workflows/packaging.yml
```diff
@@ -0,0 +1,80 @@
...
+    inputs:
+      outgoing_ref:
+        description: "The ref to be built. Can be a tag, commit hash, or branch name"
+        required: true
+        default: "main"
+      upload_to_pypi:
+        description: "Try to upload wheels and sdist to PyPI after building"
+        required: false
+        default: "false"
+      upload_to_anaconda:
```
speth, line 13: What would set these to `true`, when we do want to upload a version to PyPI / conda? Or is the idea that we would trigger builds from the `pypi-packages` and `conda-recipes` repos when we want that to happen?

SConstruct
speth: I tested the former, and it works. The only thing is you may want to add the `-y` argument, since this prompts for confirmation.

speth (commented):
I'll admit that I understand approximately nothing about how Python packaging is supposed to work anymore, but I don't see any major problems with this, and it seems to work correctly for the couple of different configurations I tried.

.github/workflows/packaging.yml
bryanwweber: These can be set to true when the workflow is run manually. Otherwise, the appropriate parameters are sent to the packaging workflows on tags.

interfaces/python_minimal/pyproject.toml
bryanwweber: Right, thanks, this is copy-pasta!

SConstruct
bryanwweber: That's unfortunate... For me, it's so much easier to reason about actual files than abstract file nodes. I think it also streamlines the installation method (pip), which doesn't have to account for the built library being in a different location, and which also doesn't know anything about file nodes.

When possible, SCons hard links files rather than copying them. However, most editors (to my understanding) use a copy-then-move strategy to save files, which creates a new inode and breaks the hardlink, so that doesn't actually help us.

What I've done is use `pytest-watch`, which allows automatically running a command before running the test suite whenever a file is saved. This doesn't support the interactive feature development case so much, but it's better than nothing.

SConstruct
bryanwweber: This is likewise clear-as-mud to me. Since I don't think any of us understands this (and maybe no one outside compiler experts at Apple do), I'm happy to drop this change. I'm happy to admit that I was trying to copy some of the conda-forge magic from their recipes in here, but I don't understand it, so I'll just drop it.

SConstruct
```diff
@@ -2077,23 +2055,6 @@ def postInstallMessage(target, source, env):
...
-              setup script (bash)         {setup_cantera!s}
-              setup script (csh/tcsh)     {setup_cantera_csh!s}
-
-            It is recommended that you run the script for your shell by typing:
-
-              source {setup_cantera!s}
-
-            before using Cantera, or else include its contents in your shell login script.
-        """.format(**env_dict))
-
```
ischoegl, line 2096: It would be nice if there'd be a newline at the end of `install_message`. The current formatting (compiled within conda on fedora) looks like:
```shell
[...]
postInstallMessage(["finish_install"], [])

File locations:

  library files               /home/ischoegl/.conda/envs/cantera-dev/lib
  C++ headers                 /home/ischoegl/.conda/envs/cantera-dev/include
  samples                     /home/ischoegl/.conda/envs/cantera-dev/share/cantera/samples
  data files                  /home/ischoegl/.conda/envs/cantera-dev/share/cantera/data
  input file converters       /home/ischoegl/.conda/envs/cantera-dev/bin  
  Python package              /home/ischoegl/.conda/envs/cantera-dev/lib/python3.10/site-packages
  Python samples              /home/ischoegl/.conda/envs/cantera-dev/lib/python3.10/site-packages/cantera/examples
scons: done building targets.
```

ischoegl (dismissed):
I gave this a spin on both Linux (conda on Fedora) and Windows (likewise from within conda). Once I made sure to run `scons clean`, things worked as expected 😆 

For windows conda with prefix (i.e. `scons build -j4 prefix=C:\Users\ischo\cantera2.6.0a4`, and `scons install`), I get

```
[...]
postInstallMessage(["finish_install"], [])

File locations:

  library files               C:\Users\ischo\cantera2.6.0a4\lib
  C++ headers                 C:\Users\ischo\cantera2.6.0a4\include
  samples                     C:\Users\ischo\cantera2.6.0a4\samples
  data files                  C:\Users\ischo\cantera2.6.0a4\data
  input file converters       C:\Users\ischo\miniconda3\envs\cantera-dev\Scripts
  Python package              C:\Users\ischo\miniconda3\envs\cantera-dev\Lib\site-packages
  Python samples              C:\Users\ischo\miniconda3\envs\cantera-dev\Lib\site-packages\cantera\examples
scons: done building targets.
```

so the input file converter location is now displayed correctly. While it's not what I would have expected, it works and is user-friendly. As such, this closes #1192.

Uninstalling works in all cases (again, `scons clean` in between compilations is pretty important).

ischoegl (dismissed):
While I lack the background to really comment on the packaging aspect, I ran a couple of tests installing and everything works as expected. The install message newline and script location when `prefix` is specified is optional (presumably, the script location is dictated by `python_prefix`?).

Fwiw, this PR closes both #1192 and #1194.

SConstruct
bryanwweber: @speth Do you have any further thoughts here?

bryanwweber:
@ischoegl Thanks, I think I addressed your comments! It also helped me find the problem with the `prefix` and the new Python installer, so it was 100% worth doing 😄

ischoegl (commented):
Thanks! Tried on windows with `prefix` again, but something apparently broke:

```
(cantera-dev) PS C:\Users\ischo\GitHub\cantera> scons build -j4 prefix=C:\Users\ischo\cantera2.6.0a4
INFO: SCons is using the following Python interpreter: C:\Users\ischo\miniconda3\envs\cantera-dev\python.exe
Compiling with MSVC 14.3
Compiling for architecture: amd64
Compiling using the following toolchain(s): ['default']
INFO: Building Cantera from git commit '8e1079a21'
INFO: Configuration variables read from 'cantera.conf' and command line:
    prefix = 'C:\\Users\\ischo\\cantera2.6.0a4'

INFO: Adding conda include and library paths: C:\Users\ischo\miniconda3\envs\cantera-dev
Checking for C++ header file cmath... yes
Checking for C++ header file fmt/ostream.h... no
INFO: Using private installation of fmt library.
INFO: Found fmt version 6.2.1
Checking for YAML::Node().Mark()... yes
INFO: Using system installation of yaml-cpp library.
Checking for C++ header file gtest/gtest.h... no
Checking for C++ header file gmock/gmock.h... no
INFO: Using Googletest from Git submodule
Checking for C++ header file eigen3/Eigen/Dense... yes
INFO: Using system installation of Eigen.
INFO: Found Eigen version 3.4.0
Checking whether __GLIBCXX__ is declared... no
Checking whether _LIBCPP_VERSION is declared... no
Checking whether __clang__ is declared... no
Checking for C++ library iomp5... no
Checking for C++ library omp... no
Checking for C++ library gomp... no
INFO: Found Boost version 1.77
Checking whether boost::core::demangle is declared... yes
Checking for CVodeCreate(CV_BDF, CV_NEWTON) in C++ library sundials_cvodes... no
Checking for CVodeCreate(CV_BDF) in C++ library sundials_cvodes... no
Checking for SUNContext ctx; SUNContext_Create(0, &ctx) in C++ library sundials_cvodes... no
Checking for double x; log(x) in C library None... yes
INFO: Using private installation of Sundials version 5.3.
INFO: Skipping compilation of the Fortran 90 interface.
INFO: Using NumPy version 1.22.2.
INFO: Using Cython version 0.29.28.
INFO: Building the full Python package for Python 3.10
  File "<string>", line 10
    prefix="C:\Users\ischo\cantera2.6.0a4")
                                          ^
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape
CalledProcessError: Command '['C:\\Users\\ischo\\miniconda3\\envs\\cantera-dev\\python.exe', '-c', '\nfrom pip import __version__ as pip_version\nfrom pkg_resources import parse_version\nimport pip\nimport json\npip_version = parse_version(pip_version)\nif pip_version < parse_version("10.0.0"):\n    from pip.locations import distutils_scheme\n    scheme = distutils_scheme("Cantera", user=False, root=None,\n                              prefix="C:\\Users\\ischo\\cantera2.6.0a4")\nelse:\n    from pip._internal.locations import get_scheme\n    scheme = get_scheme("Cantera", user=False, root=None,\n                        prefix="C:\\Users\\ischo\\cantera2.6.0a4")\n\nif not isinstance(scheme, dict):\n    scheme = {k: getattr(scheme, k) for k in dir(scheme)\n               if not k.startswith("_")}\nprint(json.dumps(scheme))\n']' returned non-zero exit status 1.:
  File "C:\Users\ischo\GitHub\cantera\SConstruct", line 1964:
    SConscript("build/python/SConscript")
  File "C:\Users\ischo\miniconda3\envs\cantera-dev\lib\site-packages\SCons\Script\SConscript.py", line 660:
    return method(*args, **kw)
  File "C:\Users\ischo\miniconda3\envs\cantera-dev\lib\site-packages\SCons\Script\SConscript.py", line 597:
    return _SConscript(self.fs, *files, **subst_kw)
  File "C:\Users\ischo\miniconda3\envs\cantera-dev\lib\site-packages\SCons\Script\SConscript.py", line 285:
    exec(compile(scriptdata, scriptname, 'exec'), call_stack[-1].globals)
  File "C:\Users\ischo\GitHub\cantera\build\python\SConscript", line 154:
    install_locs = get_pip_install_location(localenv["python_cmd"], user_install,
  File "C:\Users\ischo\GitHub\cantera\site_scons\buildutils.py", line 1245:
    return json.loads(get_command_output(python_cmd, "-c", install_script))
  File "C:\Users\ischo\GitHub\cantera\site_scons\buildutils.py", line 1199:
    data = subprocess.run(
  File "C:\Users\ischo\miniconda3\envs\cantera-dev\lib\subprocess.py", line 524:
    raise CalledProcessError(retcode, process.args,
(cantera-dev) PS C:\Users\ischo\GitHub\cantera>
```
(I made sure to run `scons clean`)

bryanwweber:
> Thanks! Tried on windows with prefix again, but something apparently broke:

Yeah, this is the same failure as on the Windows CI. Won't be hard to fix 👍

bryanwweber:
> Thanks! Tried on windows with `prefix` again, but something apparently broke:

@ischoegl Ah, this is apparently a different error, actually, due to Windows insistence on using the wrong slash for a path separator. The paths should be normalized by that point, so I'm not sure what's going on, but I can reproduce locally and I'll fix it 👍

ischoegl:
post-install message for `scons build prefix=C:\Users\ischo\cantera2.6.0a4`
```
[...]
postInstallMessage(["finish_install"], [])

    File locations:
      library files               C:\Users\ischo\cantera2.6.0a4\lib
      C++ headers                 C:\Users\ischo\cantera2.6.0a4\include
      samples                     C:/Users/ischo/cantera2.6.0a4\samples
      data files                  C:/Users/ischo/cantera2.6.0a4\data
      input file converters       C:\Users\ischo\cantera2.6.0a4\Scripts
      Python package              C:\Users\ischo\cantera2.6.0a4\Lib\site-packages
      Python examples             C:\Users\ischo\cantera2.6.0a4\Lib\site-packages\cantera\examples

scons: done building targets.
```

ischoegl (approved):
While the forward/backward slashes aren't pretty, this shouldn't hold back the beta release.

PR #1159, opened 2021-12-19
Preparations for removal of CTI/XML

speth:

This PR takes care of changes that are needed to allow clean removal of the CTI and XML formats after the release of Cantera 2.6.

- Use of CTI/XML for input/output generates deprecation warnings
- CK to CTI and CTI to XML conversions generate deprecation warnings
- Read element standard entropies from a YAML database, eliminating the dependence on `elements.xml`
- Use YAML reference files for `ck2yaml`, `cti2yaml`, and `ctml2yaml` instead of CTI or XML files these tests can be maintained even after the removal of the CTI and XML functionality
- Fix 1D solver debug mode to save residual as YAML instead of XML
- Fix `cti2yaml` to create an empty Kinetics model when one is implied (for example, by the `ideal_gas` directive)
- Update instructions for creating "reference" files for some 1D solver regression tests
- Fix setting of `T-max` for species thermo in XML to YAML conversions
- Update standard concentration model names when converting from XML to YAML
- Fix unit handling for reactions with non-reactant orders in `ctml_writer`
- Transition all Python tests to use YAML input files
- Deprecate some additional XML helper functions and classes like `toSI`
- Deprecate unused `ThermoPhase::getParameters(int, double*)` and `ThermoPhase::setParameters(int, double*)`
- Update Python docstrings to use YAML examples instead of CTI/XML
- Remove XML-centric, model-specific instructions for instantiating phases in C++
- Remove test data files that are no longer used

**Checklist**

  - The decrease in coverage comes from a lot of XML-based initialization no longer being called, which I think is to be expected at this point.


data/element-standard-entropies.yaml
```diff
@@ -0,0 +1,282 @@
...
+      Americium and Technetium. OECD Nuclear Energy Agency.
+
+elements:
+- symbol: H
+  entropy298: 65.340e+03
+  source: 1/2 H2 gas from Chase (1998), p. 1310.
+- symbol: D
+  entropy298: 72.480e+03
+  source: 1/2 D2 gas from Chase (1998), p. 1040.
+- symbol: Tr
```
bryanwweber, line 31: Yttrium has a note that there is no data found, should that be added here?

interfaces/cython/cantera/ctml2yaml.py
```diff
@@ -1598,7 +1603,7 @@ def const_cp(self, thermo: etree.Element) -> Dict[str, Union[str, "QUANTITY"]]:
         if tmin is not None and tmin != '100.0':
             thermo_attribs['T-min'] = float(tmin)
         tmax = const_cp_node.get('Tmax')
-        if tmax is not None and tmin != '5000.0':
+        if tmax is not None and tmax != '5000.0':
```
bryanwweber, line 1606: Nice catch!

interfaces/cython/cantera/ctml_writer.py
```diff
@@ -2756,6 +2761,11 @@ def convert(filename=None, outName=None, text=None):
 def main():
     if len(sys.argv) not in (2,3):
         raise ValueError('Incorrect number of command line arguments.')
+    print("WARNING: The CTI and XML input file formats are deprecated and will be\n"
```
bryanwweber, line 2764: I think switching this to `DeprecationWarning` would be better.

interfaces/cython/cantera/ck2cti.py
```diff
@@ -2335,6 +2335,12 @@ def main(argv):
                                   surfaceFile, phaseName, outName,
                                   permissive=permissive)
 
+    logging.warning(
```
bryanwweber, line 2338: I think this can be turned into a deprecation warning

interfaces/cython/cantera/onedim.pyx
```diff
@@ -1408,10 +1408,13 @@ cdef class Sim1D:
...
 
-    def save(self, filename='soln.xml', name='solution', description='none',
+    def save(self, filename='soln.yaml', name='solution', description='none',
              loglevel=1):
         """
-        Save the solution in XML format.
+        Save the solution in YAML or XML format.
+
+        .. deprecated:: 2.6
+           XML-based output is deprecated and will be removed in Cantera 3.0.
```
bryanwweber, line 1417: I think it'd be good to check if the filename ends with `.xml` and log a deprecation warning.

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -244,7 +246,7 @@ def test_explicit_reverse_rate(self):
 
     def test_explicit_forward_order(self):
         output = self.convert('explicit-forward-order.inp', thermo='dummy-thermo.dat')
-        ref, gas = self.checkConversion('explicit-forward-order.xml', output)
+        ref, gas = self.checkConversion('explicit-forward-order.yaml', output)
```
bryanwweber:
```suggestion
        ref, gas = self.checkConversion("explicit-forward-order.yaml", output)
```

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -430,19 +432,26 @@ def test_surface_mech2(self):
         self.assertNear(covdeps['H_Pt'][2], -6e6)
 
     def test_third_body_plus_falloff_reactions(self):
-        self.convert('third_body_plus_falloff_reaction.inp')
-        gas = ct.Solution('third_body_plus_falloff_reaction' + self.ext)
+        output = self.convert('third_body_plus_falloff_reaction.inp')
```
bryanwweber:
```suggestion
        output = self.convert("third_body_plus_falloff_reaction.inp")
```

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -430,19 +432,26 @@ def test_surface_mech2(self):
...
+        gas = ct.Solution(output)
         self.assertEqual(gas.n_reactions, 2)
 
     def test_blank_line_in_header(self):
-        self.convert('blank_line_in_header.inp')
-        gas = ct.Solution('blank_line_in_header' + self.ext)
+        output = self.convert('blank_line_in_header.inp')
```
bryanwweber:
```suggestion
        output = self.convert("blank_line_in_header.inp")
```

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -467,11 +476,10 @@ def _convert(self, inputFile, *, thermo, transport, surface, output, extra,
 
     @utilities.slow_test
     def test_extra(self):
-        self.convert('gri30.inp', thermo='gri30_thermo.dat',
-                     transport='gri30_tran.dat', output='gri30_extra',
-                     extra='extra.yaml')
+        output = self.convert('gri30.inp', thermo='gri30_thermo.dat',
+                              transport='gri30_tran.dat', output='gri30_extra',
+                              extra='extra.yaml')
```
bryanwweber: Double quotes :smiley:

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -739,46 +748,29 @@ def test_sofc(self):
...
     def test_liquidvapor(self):
-        output_file = self.test_work_path / "liquidvapor.yaml"
-        cti2yaml.convert(
-            self.cantera_data_path / "liquidvapor.cti",
-            output_file,
-        )
-        for name in ['water', 'nitrogen', 'methane', 'hydrogen', 'oxygen',
-                     'hfc134a', 'carbondioxide', 'heptane']:
+        self.convert("liquidvapor", self.cantera_data_path)
+        for name in ['water', 'nitrogen', 'methane', 'hydrogen', 'oxygen', 'heptane']:
```
bryanwweber:
```suggestion
        for name in ["water", "nitrogen", "methane", "hydrogen", "oxygen", "heptane"]:
```

interfaces/cython/cantera/test/test_convert.py
```diff
@@ -983,45 +963,29 @@ def test_sofc(self):
...
 
     def test_liquidvapor(self):
-        output_file = self.test_work_path / "liquidvapor.yaml"
-        ctml2yaml.convert(
-            self.cantera_data_path / "liquidvapor.xml", output_file,
-        )
-        for name in ['water', 'nitrogen', 'methane', 'hydrogen', 'oxygen',
-                     'hfc134a', 'carbondioxide', 'heptane']:
+        self.convert("liquidvapor", self.cantera_data_path)
+        for name in ['water', 'nitrogen', 'methane', 'hydrogen', 'oxygen', 'heptane']:
```
bryanwweber: Double quotes :smiley:

bryanwweber (changes requested):
Thanks for taking this on @speth! Just a few small comments here.

include/cantera/thermo/LatticePhase.h
```diff
@@ -178,29 +178,6 @@ namespace Cantera
...
- * or by the following constructor:
- *
- * @code
- *    XML_Node *xc = get_XML_File("O_lattice_SiO2.xml");
- *    XML_Node * const xs = xc->findNameID("phase", "O_lattice_SiO2");
- *    LatticePhase *o_lattice = new LatticePhase(*xs);
- * @endcode
- *
- * The XML file used in this example is listed in the next section
- *
```
ischoegl, line 203: Should this be replaced by instructions for instantiation via `AnyMap` (or any other alternative to XML)? Also, there is a subsequent XML Example that should likely be removed?

include/cantera/thermo/SurfPhase.h
```diff
@@ -89,26 +89,6 @@ namespace Cantera
...
- * @endcode
- *
- * or by the following constructor:
- *
- * @code
- *    XML_Node *xc = get_XML_File("diamond.xml");
- *    XML_Node * const xs = xc->findNameID("phase", "diamond_100");
- *    SurfPhase *diamond100TP = new SurfPhase(*xs);
- * @endcode
- *
```
ischoegl, line 111: Similar here.

include/cantera/thermo/WaterSSTP.h
```diff
@@ -63,36 +63,6 @@ class WaterProps;
...
- * @endcode
- *
- * or by the following call to importPhase():
- *
- * @code
- *    XML_Node *xm = get_XML_NameID("phase", "waterSSTPphase.xml#water", 0);
- *    WaterSSTP water;
- *    importPhase(*xm, &water);
- * @endcode
- *
```
ischoegl, line 95: ditto

src/base/xml.cpp
```diff
@@ -772,6 +772,11 @@ void XML_Node::build(const std::string& filename)
 
 void XML_Node::build(std::istream& f, const std::string& filename)
 {
+    warn_deprecated("XML_Node::build",
+        "\nThe CTI and XML input file formats are deprecated and will be removed in\n"
+        "Cantera 3.0. Use `cti2yaml.py` or `ctml2yaml.py` to convert CTI or XML input\n"
+        "files to the YAML format. See https://cantera.org/tutorials/legacy2yaml.html\n"
+        "for more information.");
```
ischoegl: Should these be regular single quotes rather than back-ticks? (Here and elsewhere)

interfaces/cython/cantera/test/test_purefluid.py
```diff
@@ -512,7 +512,7 @@ class HFC134a(PureFluidTestCases, utilities.CanteraTest):
     def __init__(self, *args, **kwargs):
         refState = StateData('critical', 374.21, 4.05928,
                              rho=511.900, u=381.70937, s=1.5620991)
-        PureFluidTestCases.__init__(self, 'hfc134a', refState)
+        PureFluidTestCases.__init__(self, 'HFC-134a', refState)
```
ischoegl: some missed double quotes here and below :stuck_out_tongue_winking_eye:

ischoegl (changes requested):
The `@allow_deprecated` decorator is quite nice!

Beyond, I have a couple of comments.

src/base/xml.cpp
ischoegl: On a separate note, would it make sense to promote the `legacy2yaml` page to a more prominent position on the tutorials page (right now it's within the input-files sub-category)?

interfaces/cython/cantera/ctml_writer.py
speth: Using `raise DeprecationWarning(...)` has the effect of also printing out the stack trace, which most users are not interested in and I think has the effect of causing people to miss the actual important warning text, e.g.:
```
  File "/home/speth/src/cantera/interfaces/cython/cantera/ctml_writer.py", line 2772, in <module>
    main()
  File "/home/speth/src/cantera/interfaces/cython/cantera/ctml_writer.py", line 2764, in main
    raise DeprecationWarning("WARNING: The CTI and XML input file formats are deprecated and will be\n"
DeprecationWarning: WARNING: The CTI and XML input file formats are deprecated and will be
removed in Cantera 3.0. Use `cti2yaml.py` to convert CTI input files to
the YAML format. See https://cantera.org/tutorials/legacy2yaml.html for
more information.
```

interfaces/cython/cantera/onedim.pyx
speth: This already emits a deprecation warning deriving from `XML_Node::write`.

include/cantera/thermo/LatticePhase.h
speth: The reason I went ahead with removing these "instantiation" examples is because I think they are poor examples of how to instantiate phases in the first place, and because nothing about this is specific to a particular `ThermoPhase` model. Good ways to do this are better documented by the C++ examples, such as https://cantera.org/examples/cxx/demo.html.

I agree that the the XML input file examples also need to be on their way out, but I didn't want to lose those just yet given that (a) the file format is still supported for this Cantera version and (b) I think we do need a home somewhere for some complete input file examples, beyond just the snippets currently available in the YAML API documentation (https://cantera.org/documentation/docs-2.5/sphinx/html/yaml/phases.html#phase-thermodynamic-models).

src/base/xml.cpp
speth: I'm starting to wondering if a large chunk of the "Input" documentation needs to be promoted a level in the website -- Right now, it's all buried as a subsection under "tutorials", which has the effect of both putting the input file related tutorials and extra layer deep, and hiding the YAML format documentation inside the tutorials section, which isn't really correct.

include/cantera/thermo/LatticePhase.h
ischoegl: I believe it’s :+1: to leave the XML input examples as long as they are marked as ‘deprecated’

interfaces/cython/cantera/ctml_writer.py
bryanwweber: I don't think you need to `raise` the `DeprecatuonWarning`. I think importing `warnings` and using that interface would be better. https://docs.python.org/3/library/warnings.html#warnings.warn

Perhaps `FutureWarning` would be better, too: https://docs.python.org/3/library/exceptions.html#FutureWarning

speth: I see, I was misunderstanding how `DeprecationWarning` is supposed to be used, based on the fact that it is a type of `Exception`. But the output from `warnings.warn` still seems clunkier than I'd prefer:
```
/home/speth/src/cantera/build/python/cantera/ctml_writer.py:2765: FutureWarning: 
The CTI and XML input file formats are deprecated and will be
removed in Cantera 3.0. Use `cti2yaml.py` to convert CTI input files to
the YAML format. See https://cantera.org/tutorials/legacy2yaml.html for
more information.
  warnings.warn(
```
The line reference makes sense, but I don't really get why it prints the line of code that starts the call to `warnings.warn` (which, if the message were shorter, would just be repeating the message again).

data/element-standard-entropies.yaml
speth: Sure, why not?

src/base/xml.cpp
speth: Actually, I take that back - Besides being accessible from within the "Tutorials" section, the general YAML documentation is available from the main `cantera.org` page under Documentation -> YAML Input File Users' Guide, and the documentation for specific models is available as Documentation -> YAML Input File API Reference, which seems about right.

I'm not sure how much more prominent the `legacy2yaml` page needs to be. Within the tutorials section, I think it's in the right place, though I suppose it could come before the deprecated options of "creating a new CTI file" and "conversion from Chemkin to CTI". For users who don't think they need tutorials, an additional link on the "Documentation" page wouldn't hurt either.

ischoegl: I personally think that the input format description should be linked on the main page. It’s tangential here, and the url in the warnings is fine

ischoegl (approved):

interfaces/cython/cantera/ctml_writer.py
bryanwweber: Looks like the recommendation in the docs is to replace [`warnings.formatwarning`](https://docs.python.org/3/library/warnings.html#warnings.formatwarning) (via [SO](https://stackoverflow.com/a/2187390/2449192)). Something like

```python
def custom_formatwarnings(msg, category, filename, lineno, line=None):
    return f"{filename}:{lineno}:{msg}\n"
warnings.formatwarning = custom_formatwarnings
```

Since the message is longer though, it doesn't bother me too much.

speth: I don't understand what is "better" about using the `warnings` interface when:
- This message is triggered only when this file is called as a script (i.e., through the `main` method), so there's no benefit/opportunity for anything to interact with settings that would either suppress this warning or elevate it into an error
- All other warning output from this script is just handled by `print` anyway
- The entire script is deprecated and I'm planning deleting it as soon as we release Cantera 2.6.

bryanwweber: Sounds good to me!

bryanwweber (approved):
Thanks @speth!

PR #1161, opened 2021-12-22
Prevent errors due to unsafe math optimizations

### LINKED ISSUE #1155: Optimization with 'fast math' produces incorrect results and segfaults
**Problem description**

As reported in #1150, compilation with default optimization options for the Intel compiler suite `icx`/`icpx` results in incorrect results. The behavior can be reproduced for `gcc` with the `-ffinite-math-only` option (which is one of the `-ffast-math` flags), e.g. for a vanilla `gcc` toolchain on Ubuntu 20.04:

```
$ scons build optimize_flags="-O3 -ffinite-math-only"
```
The option breaks strict IEEE compliance, as it does "*Allow optimizations for floating-point arithmetic that assume that arguments and results are not NaNs or +-Infs.*"  There are some instances where Cantera uses `NaN` internally which presumably breaks 'fast math' optimizations.

**Steps to reproduce**

1. Compile as indicated above.
2. An attempt to run the test suite results in numerous segfaults. 
3. Output for `gcc` with fast math (same as for `icx` in #1150) ...

```
In [3]: gas2 = ct.Solution('gri30.yaml')

In [4]: gas2.partial_molar_cp
Out[4]: 
array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0.])

In [5]: gas2.partial_molar_entropies
Out[5]: 
array([-1.84618157e-12,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06,  5.74342730e+06,  5.74342730e+06,  5.74342730e+06,
        5.74342730e+06])
```

**Behavior**

**System information**

- Cantera version: 2.5.1, 2.6.0a3
- OS: Ubuntu 20.04 (and presumably others)
- Python/MATLAB/other software versions: `gcc` / `icx`

**Additional context**

* See initial issue report #1150; a new issue is opened as the behavior is not compiler-dependent.
* In case the issue is worth pursuing, some workarounds need to be found for `{fmt}` output, which currently produces numerous warnings related to `NaN` compliance.
* An alternative check for `NaN` can be implemented, see [SO](https://stackoverflow.com/questions/38278300/successfully-enabling-fno-finite-math-only-on-nan-removal-method)

speth:
This is my proposed resolution to #1155.

- Require a working implementation of `std::isnan`, and suggest appropriate compiler flags if this is not the case
- Adjust test tolerances so that running with `-ffast-math -fno-finite-math-only` passes the test suite
- Add these flags to one set of CI builds (I added this to the "multiple Sundials" runs somewhat arbitrarily)

Closes #1155

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -960,8 +960,6 @@ int main(int argc, char** argv) {
     context.Result(result)
     return result
 
-conf = Configure(env, custom_tests={'CheckStatement': CheckStatement})
-
 # Set up compiler options before running configuration tests
 env['CXXFLAGS'] = listify(env['cxx_flags'])
 env['CCFLAGS'] = listify(env['cc_flags']) + listify(env['thread_flags'])
@@ -1006,6 +1004,8 @@ def config_error(message):
     else:
         print("See 'config.log' for details.")
     sys.exit(1)
+
+conf = Configure(env, custom_tests={'CheckStatement': CheckStatement})
 
 # First, a sanity check:
 if not conf.CheckCXXHeader('cmath', '<>'):
--- a/SConstruct
+++ b/SConstruct
@@ -1027,6 +1027,16 @@ def get_expression_value(includes, expression, defines=()):
               '}\n'))
     return '\n'.join(s)
 
+# Check that NaN is treated correctly
+nan_check_source = get_expression_value(["<cmath>"], 'std::isnan(NAN + argc)')
+retcode, nan_works = conf.TryRun(nan_check_source, ".cpp")
+if nan_works.strip() != "1":
+    config_error(
+        "Cantera requires a working implementation of 'std::isnan'.\n"
+        "If you have specified '-ffast-math' or equivalent as an optimization option,\n"
+        "either remove this option or add the '-fno-finite-math-only option'."
+    )
+
 # Check for fmt library and checkout submodule if needed
 # Test for 'ostream.h' to ensure that version >= 3.0.0 is available
 if env['system_fmt'] in ('y', 'default'):
--- a/test/kinetics/kineticsFromYaml.cpp
+++ b/test/kinetics/kineticsFromYaml.cpp
@@ -494,10 +494,10 @@ public:
         kin->getRevRateConstants(kr.data());
         kin->getFwdRatesOfProgress(ropf.data());
         kin->getRevRatesOfProgress(ropr.data());
-        EXPECT_DOUBLE_EQ(kf[iOld], kf[iNew]);
-        EXPECT_DOUBLE_EQ(kr[iOld], kr[iNew]);
-        EXPECT_DOUBLE_EQ(ropf[iOld], ropf[iNew]);
-        EXPECT_DOUBLE_EQ(ropr[iOld], ropr[iNew]);
+        EXPECT_NEAR(kf[iOld], kf[iNew], 1e-13*kf[iOld]);
+        EXPECT_NEAR(kr[iOld], kr[iNew], 1e-13*kr[iOld]);
+        EXPECT_NEAR(ropf[iOld], ropf[iNew], 1e-13*ropf[iOld]);
+        EXPECT_NEAR(ropr[iOld], ropr[iNew], 1e-13*ropr[iOld]);
     }
 
     shared_ptr<Solution> soln;
--- a/test/kinetics/kineticsFromYaml.cpp
+++ b/test/kinetics/kineticsFromYaml.cpp
@@ -494,10 +494,10 @@ public:
         kin->getRevRateConstants(kr.data());
         kin->getFwdRatesOfProgress(ropf.data());
         kin->getRevRatesOfProgress(ropr.data());
-        EXPECT_NEAR(kf[iOld], kf[iNew], 1e-13*kf[iOld]);
-        EXPECT_NEAR(kr[iOld], kr[iNew], 1e-13*kr[iOld]);
-        EXPECT_NEAR(ropf[iOld], ropf[iNew], 1e-13*ropf[iOld]);
-        EXPECT_NEAR(ropr[iOld], ropr[iNew], 1e-13*ropr[iOld]);
+        EXPECT_NEAR(kf[iOld], kf[iNew], 1e-13 * kf[iOld]);
+        EXPECT_NEAR(kr[iOld], kr[iNew], 1e-13 * kr[iOld]);
+        EXPECT_NEAR(ropf[iOld], ropf[iNew], 1e-13 * ropf[iOld]);
+        EXPECT_NEAR(ropr[iOld], ropr[iNew], 1e-13 * ropr[iOld]);
     }
 
     shared_ptr<Solution> soln;
--- a/test/thermo/PengRobinson_Test.cpp
+++ b/test/thermo/PengRobinson_Test.cpp
@@ -186,13 +186,13 @@ TEST_F(PengRobinson_Test, setTP)
         const double temp = 294 + i*2;
         set_r(0.999);
         test_phase->setState_TP(temp, 5542027.5);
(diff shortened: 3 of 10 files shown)


SConstruct
```diff
@@ -1027,6 +1027,16 @@ def get_expression_value(includes, expression, defines=()):
...
     return '\n'.join(s)
 
+# Check that NaN is treated correctly
+nan_check_source = get_expression_value(["<cmath>"], 'std::isnan(NAN + argc)')
+retcode, nan_works = conf.TryRun(nan_check_source, ".cpp")
+if nan_works.strip() != "1":
+    config_error(
+        "Cantera requires a working implementation of 'std::isnan'.\n"
+        "If you have specified '-ffast-math' or equivalent as an optimization option\n"
+        "either remove this option or add the '-fno-finite-math-only option"
```
ischoegl, line 1037: Missing `'`. Also, `-fno-finite-math-only` does not exist on all compilers.

test/kinetics/kineticsFromYaml.cpp
```diff
@@ -494,10 +494,10 @@ class ReactionToYaml : public testing::Test
...
         kin->getFwdRatesOfProgress(ropf.data());
         kin->getRevRatesOfProgress(ropr.data());
-        EXPECT_DOUBLE_EQ(kf[iOld], kf[iNew]);
-        EXPECT_DOUBLE_EQ(kr[iOld], kr[iNew]);
-        EXPECT_DOUBLE_EQ(ropf[iOld], ropf[iNew]);
-        EXPECT_DOUBLE_EQ(ropr[iOld], ropr[iNew]);
+        EXPECT_NEAR(kf[iOld], kf[iNew], 1e-13*kf[iOld]);
+        EXPECT_NEAR(kr[iOld], kr[iNew], 1e-13*kr[iOld]);
+        EXPECT_NEAR(ropf[iOld], ropf[iNew], 1e-13*ropf[iOld]);
+        EXPECT_NEAR(ropr[iOld], ropr[iNew], 1e-13*ropr[iOld]);
```
ischoegl: White-space left and right of `*`?

SConstruct
ischoegl: Missing `'`. Also, `-fno-finite-math-only` does not exist on all compilers.

test/thermo/PengRobinson_Test.cpp
```diff
@@ -189,7 +189,7 @@ TEST_F(PengRobinson_Test, setTP)
         EXPECT_NEAR(test_phase->density(),rho1[i],1.e-8);
 
         test_phase->setState_TP(temp, 7388370.);
-        EXPECT_NEAR(test_phase->density(),rho2[i],1.e-8);
+        EXPECT_NEAR(test_phase->density(),rho2[i],1.e-7);
```
ischoegl: white-spaces. (Assuming that minor formatting issues are fixed whenever lines are touched)

ischoegl (changes requested):
Thank for addressing #1155! This looks mostly good to me, and should take care of most concerns from #1155 (I still think that `NaN` should be avoided long-term, as not all compilers offer granular control over fast-math options). As an error is in place, closing #1155 via this PR is :+1:.

Regarding `self.assertNear` vs. `self.assertEqual`: equivalent changes probably should be applied across the test suite whenever floats are compared? (which would be a *lot* of churn, so I'm not exactly advocating for this here.) At the same time, the issues probably only arise here due to the fact that activation energy input requires more processing than other items.

SConstruct
bryanwweber: Do you know which compilers so the comment can include that info?

ischoegl: For the intel compilers, you have to disable fast-math altogether (which was an outcome of #1150). The option `-fp-model fast` includes this optimization and does not work (hence default options for `icc`/`icx` in `SConstruct`) ... see [documentation](https://www.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/compiler-options/compiler-option-details/floating-point-options/fp-model-fp.html)

I believe @speth's intent of 'or equivalent' is clear the first time it's used, but an equivalent caveat should be added for the `'-fno-finite-math-only'` option.

speth: > Regarding `self.assertNear` vs. `self.assertEqual`: equivalent changes probably should be applied across the test suite whenever floats are compared? (which would be a _lot_ of churn, so I'm not exactly advocating for this here.)

While it's not obvious from the diff alone, the instances I changed are more than just the test cases that failed -- these are the other instances in these files where a calculated value was being compared (because we go back and forth between Ea/R and E).

speth:
> Regarding `self.assertNear` vs. `self.assertEqual`: equivalent changes probably should be applied across the test suite whenever floats are compared? (which would be a _lot_ of churn, so I'm not exactly advocating for this here.) At the same time, the issues probably only arise here due to the fact that activation energy input requires more processing than other items.

While it's not obvious from the diff alone, the instances I changed are more than just the test cases that failed -- these are the other instances in at least these Python tests where a _calculated_ value was being compared for exact equality (because we go back and forth between Ea/R and E). I didn't go through `kineticsFromYaml.cpp` as thoroughly.

SConstruct
speth: While `icx' doesn't list this flag in its man page, I tested this and it does work. I guess it's implemented for the sake of compatibility with GCC's options.

ischoegl:
>  I didn't go through `kineticsFromYaml.cpp` as thoroughly.

I think those discrepancies of exact bit-wise matches can occur anytime a unit conversion happens. I'm not advocating for large-scale changes (if *any* beyond what you already changed) in this PR, but it may be safer to avoid exact comparisons for floating point values in general?

speth:
> it may be safer to avoid exact comparisons for floating point values in general?

Of course, and I think in general, we do a good job of that. There are just a lot comparisons in the test suite, and it's easy to type `ASSERT_EQ` when you mean `ASSERT_NEAR` (and also have to decide how near you mean).

SConstruct
ischoegl: That's great to know. I guess google was not my friend here :rofl:

ischoegl:
>  it's easy to type `ASSERT_EQ` when you mean `ASSERT_NEAR`

... guilty as charged! Certainly has happened in some of the tests I wrote.

ischoegl (approved):
Thank you, this looks good to me!

PR #1162, opened 2021-12-23
Add support for Sundials 6.0

### LINKED ISSUE #124: Check compatibility and add support for Sundials 6.0.0
**Abstract**

Sundials 6.0.0 is out and with some serious-looking revisions:
https://github.com/LLNL/sundials/blob/master/CHANGELOG.md

**Motivation**
Sundials is a dependency and the latest version will need to be supported eventually

speth:

- Add support for Sundials 6.0
- Add a corresponding CI job

Closes Cantera/enhancements#124

**Checklist**

### DIFF
--- a/src/numerics/CVodesIntegrator.cpp
+++ b/src/numerics/CVodesIntegrator.cpp
@@ -46,6 +46,19 @@ typedef int sd_size_t;
 typedef long int sd_size_t;
 #endif
 
+namespace {
+
+N_Vector newNVector(size_t N, Cantera::SundialsContext& context)
+{
+#if CT_SUNDIALS_VERSION >= 60
+    return N_VNew_Serial(static_cast<sd_size_t>(N), context.get());
+#else
+    return N_VNew_Serial(static_cast<sd_size_t>(N));
+#endif
+}
+
+} // end anonymous namespace
+
 namespace Cantera
 {
 
@@ -130,9 +143,14 @@ CVodesIntegrator::~CVodesIntegrator()
         N_VDestroy_Serial(m_dky);
     }
     if (m_yS) {
-        N_VDestroyVectorArray_Serial(m_yS, static_cast<sd_size_t>(m_np));
+        #if CT_SUNDIALS_VERSION >= 60
+            N_VDestroyVectorArray(m_yS, static_cast<sd_size_t>(m_np));
+        #else
+            N_VDestroyVectorArray_Serial(m_yS, static_cast<sd_size_t>(m_np));
+        #endif
     }
 }
+
 
 double& CVodesIntegrator::solution(size_t k)
 {
@@ -152,7 +170,7 @@ void CVodesIntegrator::setTolerances(double reltol, size_t n, double* abstol)
         if (m_abstol) {
             N_VDestroy_Serial(m_abstol);
         }
-        m_abstol = N_VNew_Serial(static_cast<sd_size_t>(n));
+        m_abstol = newNVector(n, m_sundials_ctx);
     }
     for (size_t i=0; i<n; i++) {
         NV_Ith_S(m_abstol, i) = abstol[i];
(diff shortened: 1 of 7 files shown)


src/numerics/CVodesIntegrator.cpp
```diff
@@ -551,8 +620,8 @@ double CVodesIntegrator::sensitivity(size_t k, size_t p)
 
 string CVodesIntegrator::getErrorInfo(int N)
 {
-    N_Vector errs = N_VNew_Serial(static_cast<sd_size_t>(m_neq));
-    N_Vector errw = N_VNew_Serial(static_cast<sd_size_t>(m_neq));
+    N_Vector errs = newNVector(m_neq, m_sundials_ctx);;
+    N_Vector errw = newNVector(m_neq, m_sundials_ctx);;
```
ischoegl: double `;;`

src/numerics/IDA_Solver.cpp
```diff
@@ -446,23 +478,37 @@ void IDA_Solver::init(doublereal t0)
...
+            #if CT_SUNDIALS_VERSION >= 60
+                #if CT_SUNDIALS_USE_LAPACK
+                    m_linsol = SUNLinSol_LapackBand(m_y, (SUNMatrix) m_linsol_matrix,
+                                                    m_sundials_ctx.get());
+                #else
+                    m_linsol = SUNLinSol_Band(m_y, (SUNMatrix) m_linsol_matrix,
+                                              m_sundials_ctx.get());
+                #endif
+                    IDASetLinearSolver(m_ida_mem, (SUNLinearSolver) m_linsol,
+                                       (SUNMatrix) m_linsol_matrix);
```
ischoegl: Indent level?

ischoegl (commented):
Thanks, @speth! This looks great, although I found some minor formatting glitches.

One question that arises here is that with the inclusion of 6.0, there are 5 major Sundials version supported, with the last release of 2.7 being more than 5 years old (and version 2.4 dating back to 2009). Some of the preprocessor branches could be simplified if a minimum version of `CT_SUNDIALS_VERSION >= 30` would be required.

bryanwweber:
@ischoegl 

> One question that arises here is that with the inclusion of 6.0, there are 5 major Sundials version supported, with the last release of 2.7 being more than 5 years old (and version 2.4 dating back to 2009). Some of the preprocessor branches could be simplified if a minimum version of `CT_SUNDIALS_VERSION >= 30` would be required.

I've wondered the same thing... I've always held back because I'm not sure what versions various HPC services might be using. 10 years is not a _super_ long time for them... Certainly, the pace of SUNDIALS development has picked up in the last few years, so this is becoming a bigger concern though.

ischoegl:
@bryanwweber ... yes, I am aware of HPC and generally agree. At the same time, Cantera is a somewhat odd application for HPC (unless it is tied into user code), as there really isn't much of an advantage to put things on a cluster based on how Sundials is being used at the moment?

speth:
I did consider dropping support for some older Sundials versions as part of this. However, Ubuntu 18.04 has Sundials 2.7, and I'd like to continue supporting that fully at least until it enters the "extended maintenance" period in Fall 2022. So I think we can drop support for Sundials 2.x after the Cantera 2.6 release.

Looking further ahead, Ubuntu 20.04 has Sundials 3.1, so I guess I would expect to continue supporting that through Fall 2024.

src/numerics/IDA_Solver.cpp
speth: What do you mean? That seems consistent with one of the ways we normally indent multiline statements.

speth: Oh, sorry, I was missing the context lines around this. I see now.

ischoegl (approved):
> Looking further ahead, Ubuntu 20.04 has Sundials 3.1, so I guess I would expect to continue supporting that through Fall 2024.

Makes sense! Thanks for the context!

mefuller:
Checking [pkgs.org](https://pkgs.org/search/?q=sundials), CentOS 7 is also on Sundials 2.7, is set to receive maintenance updates until 2024-06-30 (https://wiki.centos.org/About/Product), and is what's presently being used by the HPCs that I use.
Good news is that once the jump is made from EL7 to EL8, sundials jumps to 5.8 (although with the CentOS situation, who knows when and to what exact distro those EL7 clusters will upgrade)

bryanwweber:
I think the next version of Cantera after 2.6 will be 3.0, so with the major version bump, I think dropping support for Sundials 2.7 would make sense. That will likely occur sometime in 2023 given recent release cadence, so it'd be a much smaller remaining support window for CentOS 7 than it looks right now :smile:

PR #1168, opened 2022-01-15
Fix MSVC-specific compiler warnings / tweak method names

ischoegl:

Over the past couple of merges, several windows (MSVC)-specific compiler warnings appeared that were not caught when developing with `g++`. Some of the warnings are quite long, despite being triggered by trivial causes. The following are affected:

- `StoichManager.h` / `ThirdBodyCalc.h` ... `Eigen` uses `int` to index, compared to most internal indices being `size_t`
- Some formatting / type-cast issues in `ReactorNet.h`
- Some other minor issues that resulted in warnings.
- Add missing 's' in derivative accessor names (e.g. `Kinetics.net_production_rates_ddX` instead of `Kinetics.net_production_rate_ddX`), as introduced in #1089

**Checklist**

### DIFF
--- a/include/cantera/kinetics/StoichManager.h
+++ b/include/cantera/kinetics/StoichManager.h
@@ -155,7 +155,7 @@ public:
         R[m_rxn] -= S[m_ic0];
     }
 
-    void resizeCoeffs(const std::map<std::pair<int, int>, size_t>& indices)
+    void resizeCoeffs(const std::map<std::pair<size_t, size_t>, size_t>& indices)
     {
         m_jc0 = indices.at({m_rxn, m_ic0});
     }
@@ -217,7 +217,7 @@ public:
         R[m_rxn] -= (S[m_ic0] + S[m_ic1]);
     }
 
-    void resizeCoeffs(const std::map<std::pair<int, int>, size_t>& indices)
+    void resizeCoeffs(const std::map<std::pair<size_t, size_t>, size_t>& indices)
     {
         m_jc0 = indices.at({m_rxn, m_ic0});
         m_jc1 = indices.at({m_rxn, m_ic1});
@@ -288,7 +288,7 @@ public:
         R[m_rxn] -= (S[m_ic0] + S[m_ic1] + S[m_ic2]);
     }
 
-    void resizeCoeffs(const std::map<std::pair<int, int>, size_t>& indices)
+    void resizeCoeffs(const std::map<std::pair<size_t, size_t>, size_t>& indices)
     {
         m_jc0 = indices.at({m_rxn, m_ic0});
         m_jc1 = indices.at({m_rxn, m_ic1});
@@ -395,7 +395,7 @@ public:
         }
     }
 
-    void resizeCoeffs(const std::map<std::pair<int, int>, size_t>& indices)
+    void resizeCoeffs(const std::map<std::pair<size_t, size_t>, size_t>& indices)
     {
         for (size_t i = 0; i < m_n; i++) {
             m_jc[i] = indices.at({m_rxn, m_ic[i]});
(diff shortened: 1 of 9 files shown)


ischoegl:
@speth ... looks like I overlooked an 's' in the accessor method names in #1089 (oops, had skipped a redundant ‘s’ on purpose when the names were longer, but with the updated `ddX` nomenclature this rationale does not apply). Also, it turns out that some types used by `Eigen::Triplet` created copious warnings in MSVC (that I didn't see as I routinely use `gcc` to develop). This PR should fix all of those issues (and some other minor warnings).

include/cantera/kinetics/StoichManager.h
```diff
@@ -635,11 +635,11 @@ class StoichManagerN
...
 
         // Set up index pairs for derivatives
-        std::map<std::pair<int, int>, size_t> indices;
-        size_t n = 0;
+        std::map<std::pair<size_t, size_t>, int> indices;
+        int n = 0;
         for (int i = 0; i < tmp.outerSize(); i++) {
             for (Eigen::SparseMatrix<double>::InnerIterator it(tmp, i); it; ++it) {
-                indices[{it.row(), it.col()}] = n++;
+                indices[{(size_t)it.row(), (size_t)it.col()}] = n++;
```
speth: This is quite a bit more verbose, but I've seen it strongly recommended to avoid the C-style cast in C++:
```suggestion
                indices[{static_cast<size_t>(it.row()), static_cast<size_t>(it.col())}] = n++;
```

include/cantera/kinetics/StoichManager.h
```diff
@@ -177,7 +177,7 @@ class C1
     //! Species number
     size_t m_ic0;
 
-    size_t m_jc0; //!< Index in derivative triplet vector
+    int m_jc0; //!< Index in derivative triplet vector
```
speth: This is used as an index into a `std::vector`, so I think `size_t` is still appropriate, right?

src/kinetics/ReactionFactory.cpp
```diff
@@ -253,7 +253,7 @@ bool isThreeBody(const Reaction& R)
        if (trunc(reac.second) != reac.second) {
            return false;
        }
-       nreac += reac.second;
+       nreac += (size_t)reac.second;
```
speth: In this case, it may be cleaner to just change the type of `nreac` to `int`.

src/zeroD/ReactorNet.cpp
```diff
@@ -269,16 +269,17 @@ void ReactorNet::eval(doublereal t, doublereal* y,
                     "eval(double t, double* ydot)");
                 m_have_deprecated_eval[n] = true;
             } catch (NotImplementedError& err) {
+                (void)err; // suppress compiler warning
```
speth: The better way to avoid this compiler warning is to just write the `catch` statement without giving the exception a name, e.g.:
```c++
} catch (NotImplementedError&) {
```

speth (changes requested):
Thanks for trying to clean these up. Compiler warnings that sneak in on other platforms are always kind of an unpleasant surprise when you switch to a new build environment.

src/kinetics/ReactionFactory.cpp
ischoegl: Actually, this is due to `reac.second` being `double`.

include/cantera/kinetics/StoichManager.h
ischoegl: True. Was overthinking this, changed all of these back.

ischoegl:
@speth ... thanks for the review suggestions! I believe it's all taken care of.

src/kinetics/ReactionFactory.cpp
ischoegl: Hm. But I guess this is another case for `static_cast`? (which I just realized)

ischoegl:
This should be it though ...

speth (approved):
Looks good to me.

ischoegl:
Thanks!!

PR #1169, opened 2022-01-15
Simplify creation of interface phases

### LINKED ISSUE #103: Simpler way to create objects defining an interface
**Abstract**

The currently-available methods for instantiating the set of `ThermoPhase` and `Kinetics` objects defining an interface is somewhat tedious, especially in C++. With appropriate information included in the input file, this could be simplified to a single function call to construct all of the objects and link them together.

**Motivation**

In C++, constructing a surface and its adjoining bulk phases requires something like the following:
```cxx
shared_ptr<Solution> gas = newSolution("diamond.yaml", "gas");
shared_ptr<Solution> bulk = newSolution("diamond.yaml", "diamond");
shared_ptr<Solution> surfSoln = newSolution("diamond.yaml", "diamond_100", "None", {gas, solid});
auto surf = std::dynamic_pointer_cast<SurfPhase>(soln->thermo());
auto surfKin = std::dynamic_pointer_cast<InterfaceKinetics>(soln->kinetics());
```
Where the two somewhat arcane `dynamic_pointer_cast` calls are required in the relatively common case where the user needs to access methods that are defined only on these derived classes, e.g. `setCoverages` and `advanceCoverages`. This is not to mention some of the more convoluted and fragile ways of initializing surface phase objects which litter the test suite.

**Possible Solutions**

The removal of the old `Interface` class in Cantera 2.5 paves the way for a new `Interface` class that follows the structure of the `Solution` class and is not tied to a specific phase model for the surface phase (e.g. providing flexibility to use a non-ideal surface phase model as planned in #59). For this class the `thermo()` and `kinetics()` would return `shared_ptr<SurfPhase>` and `shared_ptr<InterfaceKinetics>`, respectively, removing the need for the user to invoke `dynamic_pointer_cast`.

We could also go a step further and include information in the YAML definition of the surface phase which specifies where the adjacent phases are defined in order to instantiate all of the phases at once. (While this information was included in CTI files, it was used only to require that these adjacent bulk phases were present during construction, but not to actually help construct them.)
For example:
```yaml
phases:
- name: diamond_100
  adjacent-phases: [gas, diamond]
  ...
- name: gas
  ...
- name: diamond
  ...
```

Then, we could define the `newInterface` function so that the only required parameters were the input file name and the (optional) phase name. There are a couple of options for what to do with the adjacent phases. Perhaps the best would be to store those `Solution` objects as part of the `Interface` object. For example:
```c++
shared_ptr<Interface> surf = newInterface("diamond.yaml", "diamond_100");
shared_ptr<ThermoPhase> gas = surf->adjacent(0)->thermo();
shared_ptr<ThermoPhase> bulk = surf->adjacent("diamond")->thermo(); // accessible by name or index
```

In the case where the additional phase names were not in the input file (or where the user wanted to override those values), they could also be provided as an additional argument to the `newInterface` function.

A few other thoughts:
- It would probably make sense to extend this idea to an `Edge` class as well, since instantiating the set of phases involved is even more complicated, with up to 3 bulk phases and 3 interfaces involved.
- For the Python module, adding this should be pretty simple, as the generic `Interface` class already exists, and importing the adjacent phases could be done as part of the constructor for that class. One issue here would be how to transition, given that the `Interface` constructor doesn't currently import those phases.

**References**
* Originally brought up in the context of Cantera/cantera#1062
* #59

speth:

- Automatically import phases adjacent to an interface based on the `adjacent-phases` YAML field
- Make these phases available using the `Solution.adjacent(name)` or `adjacent(index)` methods of the C++ `Solution` class
- Make these phases available as the `Solution.adjacent` `dict` in Python (keyed by phase name)
- Write an `adjacent-phases` field when serializing interfaces to YAML
- Set the `adjacent-phases` field when converting mechansims from CK, CTI, or XML formats
- Deprecate the no-longer-used `Phase::setRoot` method
- Add the missing constructor `Edge(filename, phasename)`

Closes Cantera/enhancements#103

Python:
```python
tpb_a = ct.Interface("sofc.yaml", "tpb")
anode_surf, oxide_surf_a, anode_bulk = tpb_a.adjacent.values()
oxide_a, gas_a = oxide_surf_a.adjacent.values()
```

C++:
```cxx
shared_ptr<Interface> soln = newInterface("stoichSolidKinetics.yaml", "reaction_surface");
shared_ptr<ThermoPhase> gas = soln->adjacent("air")->thermo();
```

**Checklist**


interfaces/cython/cantera/test/test_composite.py
```diff
@@ -806,3 +814,51 @@ def test_species_transport(self):
...
+          adjacent-phases: [{ptcombust.yaml/phases: [gas]}]
+          species: [{ptcombust.yaml/species: all}]
+          kinetics: surface
+          reactions: [{ptcombust.yaml/reactions: all}]
+          site-density: 2.7063e-09
+        """
+
+        surf = ct.Interface(yaml=yaml)
+        self.assertEqual(surf.adjacent["gas"].n_species, 32)
+        self.assertEqual(surf.n_reactions, 24)
```
ischoegl: Missing carriage return

src/thermo/SurfPhase.cpp
```diff
@@ -381,6 +381,12 @@ EdgePhase::EdgePhase(doublereal n0)
     setNDim(1);
 }
 
+EdgePhase::EdgePhase(const std::string& infile, const std::string& id_)
+{
+    setNDim(1);
```
ischoegl, line 386: It may be good to add a unit test as this constructor is not covered.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -1618,10 +1618,20 @@ cdef class ThermoPhase(_SolutionBase):
 cdef class InterfacePhase(ThermoPhase):
     """ A class representing a surface or edge phase"""
     def __cinit__(self, *args, **kwargs):
+        if not kwargs.get('init', True):
```
ischoegl: Double-quotes

interfaces/cython/cantera/thermo.pyx
```diff
@@ -1618,10 +1618,20 @@ cdef class ThermoPhase(_SolutionBase):
...
+            return
         if pystr(self.thermo.type()) not in ("Surf", "Edge"):
             raise TypeError('Underlying ThermoPhase object is of the wrong type.')
         self.surf = <CxxSurfPhase*>(self.thermo)
 
+    property adjacent:
+        """
```
ischoegl, line 1628: ~Why ‘higher-dimensional’?~ Is the requirement that adjacent phases require a higher dimension that the current explained somewhere?

test_problems/stoichSolidKinetics/stoichSolidKinetics.cpp
```diff
@@ -40,42 +41,29 @@ void printRates(InterfaceKinetics& iKin)
...
 
-    mll[igco2] = 0.2;
-    mll[igo2] = 0.1;
-    mll[ign2] = 0.7;
+    shared_ptr<Interface> soln = newInterface("stoichSolidKinetics.yaml",
+                                              "reaction_surface");
+
+    auto gasTP = soln->adjacent("air")->thermo();
+    auto& iKin = *soln->kinetics();
+    string mech = "stoichSolidKinetics.yaml";
```
ischoegl: Should probably be moved above line 46 and used for `newInterface`

include/cantera/base/Solution.h
```diff
@@ -109,13 +153,19 @@ shared_ptr<Solution> newSolution(const std::string& infile,
...
  */
-shared_ptr<Solution> newSolution(AnyMap& phaseNode,
-                                 const AnyMap& rootNode=AnyMap(),
-                                 const std::string& transport="",
-                                 const std::vector<shared_ptr<Solution>>& adjacent={});
+shared_ptr<Solution> newSolution(const AnyMap& phaseNode,
+                                const AnyMap& rootNode=AnyMap(),
+                                const std::string& transport="",
+                                const std::vector<shared_ptr<Solution>>& adjacent={},
+                                const std::map<std::string, shared_ptr<Solution>>& related={});
```
ischoegl: Do we align the code with the end of the opening parenthesis? (although line 168 may prompt an alternative indenting style anyways)

doc/sphinx/yaml/phases.rst
```diff
@@ -56,6 +56,17 @@ and optionally reactions that can take place in that phase. The fields of a
...
+``adjacent-phases``
+    For interface phases, specification of adjacent phases that participate in reactions
+    on the interface. This can be:
+
+    - a list of phase names that appear in the ``phases`` section of the file.
+    - A list of single-key mappings of section names to a list of phase names. These
+      sections can be in the same file as the current phase definition, or from another
+      file if written as ``file-path/section-name``. If a relative path is specified,
+      the directory containing the current file is searched first, followed by the
+      Cantera data path.
```
ischoegl, line 68: Is this feature covered by a unit test?

include/cantera/base/Solution.h
```diff
@@ -109,13 +153,19 @@ shared_ptr<Solution> newSolution(const std::string& infile,
  * @param rootNode the root node of the tree containing the phase definition, which
  *     will be used as the default location from which to read species definitions.
  * @param transport name of the transport model.
- * @param adjacent vector containing adjacent solution objects.
+ * @param adjacent vector containing adjacent Solution objects. If empty, adjacent
+ *                 phases will be instantiated based on the phase definition.
+ * @param related  Vector of phases related to the same root Solution object. Used
```
ischoegl: 'Vector' should be lowercase?

interfaces/cython/cantera/ck2yaml.py
```diff
@@ -1952,6 +1952,7 @@ def write_yaml(self, name='gas', out_name='mech.yaml'):
                 phase = BlockMap()
                 phase['name'] = surf.name
                 phase['thermo'] = 'ideal-surface'
+                phase['adjacent-phases'] = FlowList([name])
```
ischoegl, line 1955: Is this covered by a unit test?

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1480,9 +1480,12 @@ def __init__(self, name='', elements='', species='', note='',
         self.kinetics = None if kinetics == "None" else kinetics
         self.transport = None if transport == "None" else transport
         self.site_density = site_density
+        self.adjacent_phases = phases.split()
 
     def get_yaml(self, out):
         super().get_yaml(out)
+        if self.adjacent_phases:
+            out['adjacent-phases'] = FlowList(self.adjacent_phases)
```
ischoegl, line 1488: Is this covered by a unit test?

interfaces/cython/cantera/examples/surface_chemistry/sofc.py
```diff
@@ -89,21 +89,14 @@ def NewtonSolver(f, xstart, C=0.0):
...
-# import the surfaces on the anode side
-anode_surf = ct.Interface('sofc.yaml', 'metal_surface', [gas_a])
-oxide_surf_a = ct.Interface('sofc.yaml', 'oxide_surface', [gas_a, oxide_a])
-
-# import the anode-side triple phase boundary
-tpb_a = ct.Interface('sofc.yaml', 'tpb', [anode_bulk, anode_surf, oxide_surf_a])
+# import the anode-side triple phase boundary and adjacent phases
+tpb_a = ct.Interface("sofc.yaml", "tpb")
+anode_surf, oxide_surf_a, anode_bulk = tpb_a.adjacent.values()
+oxide_a, gas_a = oxide_surf_a.adjacent.values()
```
ischoegl: I think getting entries from the dictionary by key is easier to follow. (here and elsewhere)

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -718,28 +718,36 @@ def test_yaml_outunits2(self):
...
-        surf2 = ct.Solution('ptcombust-generated.yaml', 'Pt_surf', [gas2])
+        surf2 = ct.Solution("ptcombust-generated.yaml", "Pt_surf")
         self.assertArrayNear(surf.concentrations, surf2.concentrations)
         self.assertArrayNear(surf.partial_molar_enthalpies,
                              surf2.partial_molar_enthalpies)
         self.assertArrayNear(surf.forward_rate_constants,
                              surf2.forward_rate_constants)
 
+    def test_yaml_surface_explicit(self):
+        gas = ct.Solution('ptcombust.yaml', 'gas')
```
ischoegl: double quotes ... (here and below)

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -718,28 +718,36 @@ def test_yaml_outunits2(self):
...
+        surf.coverages = np.ones(surf.n_species)
+        surf.write_yaml('ptcombust-generated.yaml')
+        self.check_ptcombust(gas, surf)
+
+    def test_yaml_surface_adjacent(self):
+        surf = ct.Interface('ptcombust.yaml', 'Pt_surf')
+        gas = surf.adjacent["gas"]
+        gas.TPY = 900, ct.one_atm, np.ones(gas.n_species)
+        surf.coverages = np.ones(surf.n_species)
+        surf.write_yaml("ptcombust-generated.yaml")
```
ischoegl, line 749: is there a `self.check_ptcombust(gas, surf)` missing?

src/base/Solution.cpp
```diff
@@ -40,23 +41,33 @@ void Solution::setName(const std::string& name) {
...
+            name(), adjacent->name());
+    }
+    if (m_thermo && adjacent->thermo()
+        && adjacent->thermo()->nDim() <= m_thermo->nDim())
+    {
+        throw CanteraError("Solution::addAdjacent",
+            "Adjacent phases should have higher dimensionality than the reacting ",
+            "phase.\n'{}' is {}-dimensional while '{}' is {}-dimensional",
+            adjacent->thermo()->name(), adjacent->thermo()->nDim(),
+            m_thermo->name(), m_thermo->nDim());
```
ischoegl, line 67: Ah - got the context for 'higher-dimensional' that was mentioned in a docstring!

interfaces/cython/cantera/base.pyx
```diff
@@ -4,10 +4,55 @@
...
+            name = pystr(adj_soln.get().name())
+            soln._adjacent[name] = _wrap_Solution(adj_soln)
+
+
+cdef object _wrap_Solution(shared_ptr[CxxSolution] cxx_soln):
+    """
+    Wrap an existing Solution object with a Python object of the correct
+    derived type.
+    """
+    import cantera as ct
```
ischoegl: I guess this is due to the inheritance structure (where objects aren't defined in Cython)? Would
```Python
from cantera import Interface, Solution
```
suffice?

src/base/Interface.cpp
```diff
@@ -0,0 +1,74 @@
...
+        auto rootNode = AnyMap::fromYamlFile(infile);
+        AnyMap& phaseNode = rootNode["phases"].getMapWhere("name", name);
+        throw InputFileError("newInterface", phaseNode,
+            "Phase definition does not define a surface phase");
+    }
+    return iface;
+}
+
+shared_ptr<Interface> newInterface(const std::string& infile,
+    const std::string& name, const std::vector<shared_ptr<Solution>>& adjacent)
```
ischoegl, line 55: Would be good to have a unit test (same for alternate `newInterface` below).

src/base/YamlWriter.cpp
```diff
@@ -27,14 +27,23 @@ void YamlWriter::setHeader(const AnyMap& header) {
...
 }
 
-void YamlWriter::addPhase(shared_ptr<Solution> soln) {
+void YamlWriter::addPhase(shared_ptr<Solution> soln, bool includeAdjacent) {
     for (auto& phase : m_phases) {
         if (phase->name() == soln->name()) {
-            throw CanteraError("YamlWriter::addPhase",
-                "Duplicate phase name '{}'", soln->name());
+            if (phase.get() == soln.get()) {
+                return;
```
ischoegl, line 37: Why does this return early?

ischoegl (changes requested):
Thanks for this PR - this is a great addition, which should greatly simplify the instantiation of `Interface`s + associated phases. Most of my comments are minor.

As a minor comment, the dictionary `values` aren't easy to parse in examples, e.g.
```python
tpb_a = ct.Interface("sofc.yaml", "tpb")
anode_surf, oxide_surf_a, anode_bulk = tpb_a.adjacent.values()
oxide_a, gas_a = oxide_surf_a.adjacent.values()
```
While it can be assumed that users are familiar with what is listed in the input files, this assumes knowledge of a specific order of adjacent phases. I overall believe that a dictionary is the right choice here, but would suggest to access entries by key (this should facilitate understanding in the case of a user just downloading an example file without ever looking into the YAML source).

interfaces/cython/cantera/test/test_composite.py
speth: Good catch.

test_problems/stoichSolidKinetics/stoichSolidKinetics.cpp
speth: Or just deleted.

include/cantera/base/Solution.h
speth: No, that was just bad formatting on my part.

doc/sphinx/yaml/phases.rst
speth: Yes. See `test_composite.py`, `TestSpeciesSerialization.test_remote_file`.

interfaces/cython/cantera/ck2yaml.py
speth: It is now...

interfaces/cython/cantera/cti2yaml.py
speth: This one and the route via `ctml2yaml` are covered now as well.

interfaces/cython/cantera/examples/surface_chemistry/sofc.py
speth: I was kind of on the fence here. Unpacking the `values()` like this was partly motivated by a suggestion you made in Cantera/enhancements#103. Going the more explicit route ends up being not that much less verbose than the old syntax in the cases where you actually want to assign variables for all the adjacent phases, but so it goes.

I've switched to the explicit style in all the examples, but some of the test code is still using the more compact syntax.

interfaces/cython/cantera/base.pyx
speth: Right, the `Solution` and `Interface` classes are defined in `.py` files, so those names can't be available directly inside the compiled `_cantera` module.

I guess the import could be written as suggested, but is there a particular advantage to doing that?

src/base/YamlWriter.cpp
speth: Added a comment documenting this case.

interfaces/cython/cantera/base.pyx
ischoegl: It's likely just semantics: to me seeing the suggested code means that we just need to import those classes, and I don't need to look over the code below to figure out why the import is needed.

interfaces/cython/cantera/examples/surface_chemistry/sofc.py
ischoegl: True, I did suggest to unpack from a tuple (which would imply a given order). I think your avenue of upgrading this to a dictionary makes a given order less obvious (although it is - at least for recent Python - built in), while also allowing for better 'self-documentation'. ~If you want to keep things short, there's always the option of using a generator (although they're harder to read).~

interfaces/cython/cantera/base.pyx
speth: I don't see that as adding much clarity, but I changed it anyway, and added a comment explaining the reason for needing an import here.

src/base/Solution.cpp
speth: Are you still looking for an explanation of this somewhere, or does this answer your question?

ischoegl: I think the docstring for `ThermoPhase.adjacent` should explain this in more detail. No action required here though.

interfaces/cython/cantera/thermo.pyx
ischoegl: I think this is fine after all.

ischoegl (approved):
Thank you @speth, this looks good to me. As said, I believe this will greatly simplify the creation of `Interface`s (at the least, it makes this much easier to understand for new users).

The only remaining thing will be to update the corresponding YAML tutorial on [Cantera/cantera-website](https://cantera.org/tutorials/yaml/defining-phases.html).

Issue #1170, opened 2022-01-15
Documentation typo: wrong variable printed in equivalenceRatio.py


blackrabbit17:
In the documentation here:
[documentation link](https://cantera.org/examples/python/thermo/equivalenceRatio.py.html)

The last line of the code is:
`
print("Z(burnt) = {:1.3f}".format(Z))
`

this should be:
`
print("Z(burnt) = {:1.3f}".format(Z_burnt))
`

bryanwweber:
Thank you for reporting this! Feel free to create a pull request with the fix, if you'd like to.

blackrabbit17:
OK cool, I have opened a pull request

PR #1173, opened 2022-01-17
Typo: equivalenceRatio.py is printing wrong results

blackrabbit17:
Fixes a typo in equivalenceRatio.py example

The last line of the code is:
`
print("Z(burnt) = {:1.3f}".format(Z))
`

this should be:
`
print("Z(burnt) = {:1.3f}".format(Z_burnt))
`


ischoegl (approved):
Thanks for the prompt fix! (Closes #1170)

Issue #1174, opened 2022-01-18
Builds/Tests now failing on aarch64, ppc64le architectures (Fedora 36=Rawhide)


mefuller:
**Problem description**

All builds on the ppc64le architecture with F36/Rawhide now fail. This was not the case four days ago (see https://copr.fedorainfracloud.org/coprs/fuller/Cantera/builds/).

Concurrently, the `kinetics: KineticsAddSpecies3.add_species_sequential` test now fails on "successful" builds on F34/35 for ppc64le architectures (and also for aarch64, i686 and s390x on all three Fedoras 34, 35, and 36/Rawhide, but not x86_64). This may not be a new problem as I only just added testing automation to the build automation for all architectures. (Yes, I know, I should have done that earlier)

**Steps to reproduce**

`scons build && scons test`

**Behavior**

Error message:
Build failure (ppc64le:F36):
```
In file included from /usr/include/eigen3/Eigen/Core:210,
                 from /usr/include/eigen3/Eigen/SparseCore:11,
                 from /usr/include/eigen3/Eigen/Sparse:26,
                 from include/cantera/numerics/eigen_sparse.h:10,
                 from include/cantera/kinetics/StoichManager.h:11,
                 from include/cantera/kinetics/Kinetics.h:14,
                 from src/base/Solution.cpp:12:
/usr/include/eigen3/Eigen/src/Core/arch/AltiVec/PacketMath.h:78:8: internal compiler error: Segmentation fault
   78 | static _EIGEN_DECLARE_CONST_FAST_Packet4i(ZERO, 0); //{ 0, 0, 0, 0,}
      |        ^
```
Test failure (all other affected systems):
```
[----------] 2 tests from KineticsAddSpecies3
[ RUN      ] KineticsAddSpecies3.add_species_sequential
test/kinetics/kineticsFromScratch3.cpp:371: Failure
Expected equality of these values:
  k_ref[i]
    Which is: 591054161.41004908
  k[i]
    Which is: 591054161.41004813
i = 0; N = 4
test/kinetics/kineticsFromScratch3.cpp:377: Failure
Expected equality of these values:
  k_ref[i]
    Which is: 150.5822178080069
  k[i]
    Which is: 150.58221780800667
i = 0; N = 4
test/kinetics/kineticsFromScratch3.cpp:384: Failure
Expected equality of these values:
  w_ref[iref]
    Which is: 150.58221780800866
  w[i]
    Which is: 150.58221780800844
sp = O; N = 4
[...]

PR #1175, opened 2022-01-18
Include explicit git commit hash in config.h

ischoegl:

During compilation, the git commit hash is currently passed through the commandline:
```
g++ -o build/src/base/global.os -c -std=c++11 -pthread -O3 -Wno-inline -Wall -include src/pch/system.h -fPIC -DNDEBUG -DGIT_COMMIT=\"fcff59292\" -Iinclude -Iinclude/cantera/ext -Ibuild/src src/base/global.cpp
```
which is, however, hard to locate. As an example, the git hash for the last successful build for #1174 would have been much easier to find.

The proposed change adds a `CANTERA_GIT_COMMIT` to the configuration summary, e.g.
```
ConfigBuilder(["build/src/config.h.build"], ["include/cantera/base/config.h.in"])
INFO: Generating build/src/config.h.build with the following settings:
    CANTERA_GIT_COMMIT                  "fcff59292"
    CANTERA_SHORT_VERSION               "2.6"
    CANTERA_VERSION                     "2.6.0a4"
[...]
```

**Checklist**

### DIFF
--- a/src/base/global.cpp
+++ b/src/base/global.cpp
@@ -129,6 +129,8 @@ void thread_complete()
 
 std::string gitCommit()
 {
+    warn_deprecated("gitCommit",
+        "Replaced by 'CANTERA_GIT_COMMIT'. To be removed after Cantera 2.6.");
     return CANTERA_GIT_COMMIT;
 }
 
--- a/src/base/global.cpp
+++ b/src/base/global.cpp
@@ -129,11 +129,7 @@ void thread_complete()
 
 std::string gitCommit()
 {
-#ifdef GIT_COMMIT
-    return GIT_COMMIT;
-#else
-    return "unknown";
-#endif
+    return CANTERA_GIT_COMMIT;
 }
 
 XML_Node* get_XML_File(const std::string& file, int debug)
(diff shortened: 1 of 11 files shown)


ischoegl:
Also eliminated `Cantera::gitCommit` as having the define in `config.h` makes things easier to follow. Should be good now (as long as CI passes).

src/base/global.cpp
```diff
@@ -127,15 +127,6 @@ void thread_complete()
...
 
-std::string gitCommit()
-{
-#ifdef GIT_COMMIT
-    return GIT_COMMIT;
-#else
-    return "unknown";
-#endif
-}
-
```
bryanwweber: Why not keep this function around? It can still use the macro from `config.h`, but keeping this would save a lot of the other churn here, I think.

ischoegl: The reason for eliminating this was to ensure that both `CANTERA_VERSION` and `CANTERA_GIT_COMMIT` are handled the same. It probably should be deprecated though ... (*Edit: done*)

speth:
The current approach is used for a reason -- this way, each time you make a local git commit, rebuilding Cantera only requires recompiling `global.cpp` and re-linking. If this is defined in `config.h`, then every single source file will need to be recompiled.

Printing out the discovered git commit somewhere near the start of the build process isn't a bad idea, though.

ischoegl:
> The current approach is used for a reason -- this way, each time you make a local git commit, rebuilding Cantera only requires recompiling global.cpp and re-linking. If this is defined in config.h, then every single source file will need to be recompiled.

Makes a lot of sense (in hindsight :joy:)! Closing this as much of the work here is rendered moot.

PR #1179, opened 2022-01-19
Resolve Yaml paper cuts

### LINKED ISSUE #1053: Paper cuts in YAML serialization API

ischoegl:

This PR aims to address remaining points raised in #1053

- Add `Species.from_dict` creation (in analogy to `Reaction.from_dict`)
- Add ability to pass YAML syntax to `anymap_from_dict`
- Address some YAML code formatting in Sphinx

Closes #1053

E.g. mixing YAML syntax when defining a species:
```python
data = {
    "name": "H2O",
    "composition": "{H: 2, O: 1}",
    "thermo": "{model: constant-cp, h0: 100}", # value is formatted as YAML code
}
species = ct.Species.from_dict(data)
```

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -1172,7 +1172,7 @@ class TestSpecies(utilities.CanteraTest):
         S = ct.Species.listFromCti((self.cantera_data_path / "h2o2.cti").read_text())
         self.assertEqual(S[3].name, self.gas.species_name(3))
 
-    def test_listFomYaml(self):
+    def test_list_from_yaml(self):
         yaml = '''
         - name: H2O
           composition: {H: 2, O: 1}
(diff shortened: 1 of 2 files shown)


ischoegl:
Pre-emptively dropping commits that will clash with #1183.

interfaces/cython/cantera/utils.pyx
```diff
@@ -284,7 +284,10 @@ cdef CxxAnyValue python_to_anyvalue(item, name=None) except *:
         else:
             v = list_to_anyvalue(item)
     elif isinstance(item, str):
-        v = stringify(item)
+        try:
+            v = AnyMapFromYamlString(stringify(item))
+        except BaseException:
+            v = stringify(item)
```
speth: How often would you really want to use the mixed syntax that this allows? Especially keeping in mind the existing support for specifying compositions as strings, which is probably the most common case. I don't think it's great that what I think must be the far more common case here ends up throwing an exception.

speth (changes requested):
I would prefer to drop the implicit YAML parsing that this introduces. Other than that, what's implemented here looks fine.

ischoegl:
> I would prefer to drop the implicit YAML parsing that this introduces. 

@speth ... no problem: dropped. Also made some of the unit test method names more pythonic (i.e. switched to snake_case), but nothing else changed.

interfaces/cython/cantera/utils.pyx
ischoegl: True. As this is not an important feature, it's better to just drop it.

speth (approved):
Looks good to me. Thanks, @ischoegl!

PR #1181, opened 2022-01-20
Refactor interface kinetics

### LINKED ISSUE #8: extend modify_reaction to work correctly for interfaceReactions
**This should be considered low priority.**

**Is your feature request related to a problem? Please describe**
modify_reaction "executes" for surface reactions, but it does not actually work. Though in many cases, it does crash with an error.
Note: I know this is not critical because one can just make a new cti file entirely, and also because I have checked and the set_multiplier function does execute correctly.

**Describe the desired solution**
It would be great if cantera could get modify_reaction to work with sticking and coverage dependent parameters. (Even better if the species of coverage dependence could be changes this way, too).

**Describe alternatives you have considered**
I have considered simply using the set_multiplier function. While this does work for many cases, it does not fix the issue of coverage dependence, since that is a dynamic quantity during simulatio

**Additional context**
I have attached a zipfile with a python file to run to see the issue.
[surf_pfr_modify_Reaction_Bug.zip](https://github.com/Cantera/cantera/files/3903552/surf_pfr_modify_Reaction_Bug.zip)

### LINKED ISSUE #87: Refactor Reaction Rate Evaluators
**Abstract**

Following #63 (and based on comments in #79), a new framework for reaction factories and reaction rate evaluators (`MultiRateBase`) is now introduced with Cantera/cantera#982. Once fully implemented, the new framework will allow for a unified handling that avoids the rigid `Rate1` templates, and eliminates the need for `FalloffFactory`, etc.. 

**Motivation**

Describe the need for the work being done:

* Simplify handling of reaction rates
* Replace hard-coded reaction rate implementation
* Simplify implementation of derivatives/Jacobians
* Improve API handling (take full advantage of `AnyMap` serialization)

**Description**

An intermediate `Reaction3` class allows for a clean deprecation of reaction types that use the old framework; until the transition is complete (Cantera 3.0?), both will remain fully functional. The refactoring is planned to involve multiple phases:

1. `BulkKinetics`: use existing `MultiBulkRates` rate evaluator to handle replacement classes, and deprecate current approaches

    a. PR to replace everything except `FalloffReaction`, i.e. create `ElementaryReaction3`, `ThreeBodyReaction3`, ... see Cantera/cantera#995
    b. PR’s to update infrastructure … see Cantera/cantera#1051 (modification of rate parameters in memory), Cantera/cantera#1061 (rate factories), Cantera/cantera#1084 (resolve role of `M`), Cantera/cantera#1088 (remove some inconsistencies of API's) 
    c. PR to replace `FalloffReactions` and associated `Falloff` and `FalloffFactory` classes, i.e. `LindemannFalloff`, `TroeFalloff`, `SRIFalloff` ... see Cantera/cantera#1101

2. `InterfaceKinetics`: a new `MultiInterfaceRates` rate evaluator will require some discussion prior to implementation

    a. PR to streamline ... see Cantera/cantera#1184
    b. Once everything is fully ported, deprecated classes (and associated handlers) can be removed and the `'3'` can be dropped from names of all replacement classes. ... see Cantera/cantera#1181

**References**

The following issues/PR's are tagged as there may be synergies and/or repercussions

* #8 - extend `modify_reaction` to `InterfaceKinetics`
* #19 - analytical Jacobians
* #38/#44 - electrochemical reactions
* #61 - dualism of `InterfaceKinetics` and reactor networks
* #63 - (closed) discussion of extendable  reactions
* #84 - make Python interface more intuitive
* #86 - Jacobians for Sundials solvers
* Cantera/cantera#597 - (closed) Langmuir-Hinshelwood kinetics
* Cantera/cantera#749 - (closed) deprecation of old electrochemical reaction types
* Cantera/cantera#982 - (closed) implementation of the initial framework 
* Cantera/cantera#984 - serialization
* Cantera/cantera#987 - Blowers Masel reactions
* Cantera/cantera#1010 - preconditioning for Sundials solvers (formerly Cantera/cantera#951)
* Cantera/cantera#1089 - analytical Jacobians (formerly Cantera/cantera#1081)
* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *

### LINKED ISSUE #1068: Documentation for coverage-dependence E parameter

### LINKED ISSUE #1210: Sticking species not detected correctly when ‘solid’ species are involved
**Problem description**

**Steps to reproduce**

Convert `haca2.cti` ([CTI file pulled verbatim from Cantera 2.5.1](https://github.com/Cantera/cantera/blob/2.5/test_problems/surfSolverTest/haca2.cti))

Open interface as:

```
In [1]: import cantera as ct
   ...: 
   ...: phase = ct.Interface("haca2.yaml", "soot_interface")
Input In [2], in <module>
      1 import cantera as ct
----> 3 phase = ct.Interface("haca2.yaml", "soot_interface")

File /work/GitHub/cantera/interfaces/cython/cantera/base.pyx:63, in cantera._cantera._SolutionBase.__cinit__()

File /work/GitHub/cantera/interfaces/cython/cantera/base.pyx:120, in cantera._cantera._SolutionBase._cinit()

File /work/GitHub/cantera/interfaces/cython/cantera/base.pyx:207, in cantera._cantera._SolutionBase._init_yaml()

CanteraError: 
CanteraError thrown by addReactions:

InputFileError thrown by InterfaceKinetics::buildSurfaceArrhenius:
Error on line 89 of ./haca2.yaml:
Multiple non-interface species ('CB-CB3' and 'OH')
found in sticking reaction: 'CB-CB3 + Csoot-H + OH => CO + Csoot-* + H2'.
Sticking species must be explicitly specified.
|  Line |
|    84 |   rate-constant: {A: 2.0e+13, b: 0.0, Ea: 0.0}
|    85 | - equation: Csoot-* + C2H2 => Csoot-H + H + 2 CB-CB3  # Reaction 6
|    86 |   rate-constant: {A: 8.0e+07, b: 1.56, Ea: 3.8}
|    87 | - equation: Csoot-* + O2 + 2 CB-CB3 => Csoot-* + 2 CO  # Reaction 7
|    88 |   rate-constant: {A: 2.2e+12, b: 0.0, Ea: 7.5}
>    89 > - equation: OH + Csoot-H + CB-CB3 => Csoot-* + CO + H2  # Reaction 8
            ^
|    90 |   sticking-coefficient: {A: 0.13, b: 0.0, Ea: 0.0}
```

**Behavior**

presumably the original CTI input file is valid (it has been part of the test suite), ~so it is not entirely clear what is going on~ … it looks like an issue with the logic in the validation of the sticking species that somehow has evaded detection: it probably should only consider gas phases?

**System information**

- Cantera version: 2.6.0a4
- OS: any
- Python/MATLAB/other software versions: Python 3.10 (conda)

**Additional context**

Issue was found when working on #1205

Resolution is contingent on #1181, ~and can be incorporated into #1205~ [*Edit: the unit test using `haca2` bypasses the (newly introduced) validation step, so this issue doesn't need to be resolved for #1205*]

As a reference, the sticking species is (presumably manually) added in the current ‘official’ YAML file https://github.com/Cantera/cantera/blob/main/test_problems/surfSolverTest/haca2.yaml

**Possible Solutions**

1. Add detection and specification of sticking species to `cti2yaml`
2. Only consider species that are in the gas phase in `InterfaceKinetics::buildSurfaceArrhenius`

I believe (2) is both more robust and easier to implement.

ischoegl:

Refactor `InterfaceKinetics` objects to reflect recent changes in `GasKinetics`.

- Replace `SurfaceArrhenius` by `InterfaceRate<>` and `StickRate<>` class templates
- Update `InterfaceKinetics` to use new framework
- Exception: `ElectrochemicalReaction` remains for the time being (the PR is getting long, and some issues pertaining to Cantera/enhancements#38 remain to be discussed)

Checklist:
- [x] Update C test suite
- [x] Issue warnings for deprecated `Reaction` specializations in Python
- [x] Use new framework by default when importing from YAML

Closes Cantera/enhancements#87, closes Cantera/enhancements#8, closes #1068, closes #1210

**Checklist**


ischoegl:
@speth ... while two tests are still failing (`test-diamondSurf-yaml` and `test-surfSolver2`), I believe this PR is ready for initial comments. This PR introduces significant changes which will likely result in some discussions.

As stated elsewhere, I believe it would make sense to decouple `ElectroChemicalReaction`s as this PR is getting long and the discussion may not have converged). Also tagging @decaluwe ...

include/cantera/kinetics/Arrhenius.h
```diff
@@ -71,26 +71,28 @@ class ArrheniusBase : public ReactionRate
 
     //! Return the pre-exponential factor *A* (in m, kmol, s to powers depending
     //! on the reaction order)
-    double preExponentialFactor() const {
+    //! class specializations may provide alternate definitions that describe
+    //! an effective activation energy that depends on the thermodynamic state.
+    virtual double preExponentialFactor() const {
```
speth, line 78: To get Doxygen to recognize this as the extended description separate from the short summary, either the summary on the line above should end in a period or there should be a blank line with just `//!` between the two.

include/cantera/kinetics/Arrhenius.h
```diff
@@ -175,6 +177,10 @@ class Arrhenius3 : public ArrheniusBase
         return m_logA + m_b * logT - m_Ea_R * recipT;
     }
 
+    //! Update information - used for InterfaceRate<> and BulkRate<>
+    void update(const BlowersMaselData& shared_data) {
+    }
+
```
speth: This seems very strange to have the base `Arrhenius3` class have a function that is only used for Blowers-Masel parameterizations. Also, why is this `update` rather than `updateFromStruct`?

include/cantera/kinetics/Arrhenius.h
```diff
@@ -366,19 +382,28 @@ class BlowersMasel : public ArrheniusBase
...
     }
 
-    //! Return the effective activation energy [J/kmol]
-    /*!
-     *  @param deltaH  Enthalpy change of reaction [J/kmol]
-     */
-    double effectiveActivationEnergy(double deltaH) const {
-        return effectiveActivationEnergy_R(deltaH / GasConstant) * GasConstant;
+    virtual double activationEnergy() const override {
+        return effectiveActivationEnergy_R(m_deltaH_R) * GasConstant;
```
speth, line 375: I'm wondering if it would be worth limiting the interface to _only_ the dimensional versions of these methods. It seems like we're ending up with a lot of boilerplate to just multiply (or not) values by the gas constant.

include/cantera/kinetics/Kinetics.h
```diff
@@ -173,6 +173,14 @@ class Kinetics
         return m_thermo.size();
     }
 
+    /**
+     * Return index specifying the beginning position for the species vector for
+     * the n-th phase in the kinetics class.
+     */
+    size_t startIndex(size_t n) const {
+        return m_start.at(n);
+    }
```
speth: This is already available as `kineticsSpeciesIndex(0, n)`.

include/cantera/kinetics/Coverage.h
```diff
@@ -0,0 +1,437 @@
...
+    void getStickParameters(AnyMap& node) const;
+
+    //! Get flag indicating whether sticking rate uses the correction factor developed
+    //! by Motz & Wise for reactions with high (near-unity) sticking coefficients.
+    //! Defaults to 'false'.
+    bool motzWiseCorrection() const {
+        return m_motzWise;
+    }
+
+    //! Set flag for by Motz & Wise correction factor
```
speth:
```suggestion
    //! Set flag for Motz & Wise correction factor
```

include/cantera/kinetics/Coverage.h
```diff
@@ -0,0 +1,437 @@
...
+    //! @param kin  Kinetics object associated with the sticking coefficient
+    //! Parameters can be accessed using the method getStickParameters
+    void buildStickParameters(const Reaction& rxn, const Kinetics& kin);
+
+    //! Return sticking coefficients
+    //! @param order  exponent applied to site density term
+    //! @param multiplier  multiplicative factor in rate expression
+    //! @param species  sticking species
+    void getStickParameters(double& order, double& multiplier,
+                            std::string& species) const {
```
speth: I think just two regular getters for `stickingOrder()` and `stickingMultiplier()` might be simpler (given there's already a `stickingSpecies()` getter) than having to use this method to return multiple values.

include/cantera/kinetics/Coverage.h
```diff
@@ -0,0 +1,437 @@
...
+    }
+};
+
+using ArrheniusInterfaceRate = InterfaceRate<Arrhenius3, CoverageData>;
+using BlowersMaselInterfaceRate = InterfaceRate<BlowersMasel, CoverageData>;
+
+
+//! A class template for interface sticking rate specifications
+template <class RateType, class DataType>
+class StickRate : public RateType, public StickCoverage
```
speth: Should this be `StickingRate`?

include/cantera/kinetics/Arrhenius.h
```diff
@@ -320,6 +326,17 @@ class BlowersMasel : public ArrheniusBase
...
+    //! Evaluate reaction rate
+    double evalRate(double logT, double recipT) const {
+        double Ea_R = effectiveActivationEnergy_R(m_deltaH_R);
+        return m_A * std::exp(m_b * logT - Ea_R * recipT);
+    }
+
+    //! Update information - used for InterfaceRate<> and BulkRate<>
+    void update(const BlowersMaselData& shared_data) {
+        updateFromStruct(shared_data);
+    }
```
speth: Do we really need `update(shared_data)` if we already ahve `updateFromStruct(shared_data)`?

interfaces/cython/cantera/_cantera.pxd
```diff
@@ -462,6 +462,9 @@ cdef extern from "cantera/kinetics/Reaction.h" namespace "Cantera":
         string type()
         double eval(double) except +translate_exception
         double eval(double, double) except +translate_exception
+        double eval(double, vector[double]&) except +translate_exception
+        double eval(double, double, vector[double]&) except +translate_exception
+        double eval(double, double, double, vector[double]&) except +translate_exception
```
speth: These latter two versions of `ReactionRate::eval` don't actually exist, do they?

interfaces/cython/cantera/reaction.pyx
```diff
@@ -679,6 +680,263 @@ cdef class CustomRate(ReactionRate):
         self.cxx_object().setRateFunction(self._rate_func._func)
 
 
+cdef class InterfaceTypeRate(ArrheniusTypeRate):
```
speth: Would naming these classes `InterfaceRateBase` and `ArrheniusRateBase` be more consistent with our other usage?

interfaces/cython/cantera/reaction.pyx
```diff
@@ -679,6 +680,263 @@ cdef class CustomRate(ReactionRate):
...
+    """
+    Base class collecting commonly used features of Arrhenius-type rate objects
+    that include coverage dependencies.
+    """
+
+    def __call__(self, double temperature, np.ndarray coverages):
```
speth, line 694: While I understand the rationale for having a simple method for evaluating simple rates (like the normal Arrhenius rate, or even P-log/Chebyshev rates), I don't think it really helps here. At least from the examples I've seen, most surface reactions do not specify any coverage dependencies, so having to specify the coverages here for such reactions makes this sort of tedious to use. I'd rather just have a method for evaluating individual rates from the corresponding `Solution` object's state, which is always general enough to work and can use the existing evaluation infrastructure more directly.

interfaces/cython/cantera/reaction.pyx
```diff
@@ -679,6 +680,263 @@ cdef class CustomRate(ReactionRate):
...
+            return anymap_to_dict(cxx_deps)
+        def __set__(self, dict deps):
+            cdef CxxAnyMap cxx_deps = dict_to_anymap(deps)
+
+            self.coverage.setCoverageDependencies(cxx_deps)
+
+    def set_species(self, species):
+        """
+        Set association with species
+        """
```
speth, line 728: This is not very clear - what is `species` expected to be?

interfaces/cython/cantera/reaction.pyx
```diff
@@ -679,6 +680,263 @@ cdef class CustomRate(ReactionRate):
...
+            cxxvector.push_back(stringify(s))
+        self.coverage.setSpecies(cxxvector)
+
+    property site_density:
+        """
+        Site density [kmol/m^2]
+
+        The site density is not an independent property. Accordingly, the setter
+        should be only used for testing purposes, as the value will be overwritten
+        by an update of the thermodynamic state.
```
speth: Wouldn't it be simpler to just acknowledge that evaluation of the rate in the absence of a `Kinetics` object doesn't make sense, and then the site density can be pulled from there?

include/cantera/kinetics/Coverage.h
```diff
@@ -0,0 +1,437 @@
...
+ *          \exp \left( \frac{- E_k \theta_k}{RT} \right)
+ *  \f]
+ * it is evident that this expression combines a regular modified Arrhenius rate
+ * expression \f$ A T^b \exp \left( - \frac{E_a}{RT} \right) \f$ with coverage-related
+ * terms, where the parameters \f$ (a_k, E_k, m_k) \f$ describe the dependency on the
+ * surface coverage of species \f$ k, \theta_k \f$. The Coverage class implements terms
+ * related to coverage only, which allows for combinations with arbitrary rate
+ * parameterizations (for example Arrhenius and BlowersMasel).
+ */
+class Coverage
```
speth: The name `Coverage` for this class seems a bit too generic for what this class represents. Maybe something like `CoverageDependency` or `CoverageRateBase` would be more descriptive? I guess the latter would avoid having to have a naming fight with the existing struct named `CoverageDependency`.

include/cantera/kinetics/Coverage.h
```diff
@@ -0,0 +1,437 @@
...
+    //! Perform object setup based on AnyMap node information
+    //! @param dependencies  Coverage dependencies
+    //! @param units  Unit system
+    void setCoverageDependencies(const AnyMap& dependencies,
+                                 const UnitSystem& units=UnitSystem());
+
+    //! Store parameters needed to reconstruct an identical object
+    //! @param dependencies  AnyMap receiving coverage information
+    //! @param asVector  Optional boolean flag to override map output
+    void getCoverageDependencies(AnyMap& dependencies, bool asVector=false) const;
```
speth: Is the `asVector` flag here really needed, or can we go without it? We don't provide such an option to for `Arrhenius::getRateParameters`.

include/cantera/kinetics/Coverage.h
```diff
@@ -0,0 +1,437 @@
...
+
+    //! Add a coverage dependency for species *sp*, with exponential dependence
+    //! *a*, power-law exponent *m*, and activation energy dependence *e*,
+    //! where *e* is in Kelvin, i.e. energy divided by the molar gas constant.
+    void addCoverageDependence(std::string sp, double a, double m, double e);
```
speth: Should be `const std::string&`

include/cantera/kinetics/Coverage.h
```diff
@@ -0,0 +1,437 @@
...
+/**
+ * Rate expressions defined for interfaces may include coverage dependent terms,
+ * where an example is given by [Kee, R. J., Coltrin, M. E., & Glarborg, P.(2005).
+ * Chemically reacting flow: theory and practice. John Wiley & Sons. Eq 11.113]:
+ *  \f[
+ *      k_f = A T^b \exp \left(
+ *          \ln 10 \sum a_k \theta_k
+ *          - \frac{1}{RT} \left( E_a + \sum E_k\theta_k \right)
+ *          + \sum m_k \ln \theta_k
+ *          \right)
```
speth: This version of the equation is not what's given in _Chemically Reacting Flow_ -- the book uses the form containing a product as below, while Cantera's implementation corresponds more closely to the version above to reduce evaluation to a single call to `exp`.

include/cantera/kinetics/Coverage.h
```diff
@@ -0,0 +1,437 @@
+/**
+ * @file Coverage.h
+ * Header for reaction rates that occur at interfaces.
+ */
```
speth: I would say that if anything, this file should be named `InterfaceRate.h`.

include/cantera/kinetics/Coverage.h
```diff
@@ -0,0 +1,437 @@
...
+    void getCoverageDependencies(AnyMap& dependencies, bool asVector=false) const;
+
+    //! Add a coverage dependency for species *sp*, with exponential dependence
+    //! *a*, power-law exponent *m*, and activation energy dependence *e*,
+    //! where *e* is in Kelvin, i.e. energy divided by the molar gas constant.
+    void addCoverageDependence(std::string sp, double a, double m, double e);
+
+    //! Set species indices within coverage array
+    void setSpecies(const Kinetics& kin);
+    void setSpecies(const std::vector<std::string>& species);
```
speth: The documentation for these methods is kind of vague. I don't know what the argument is in the latter case. When and why is this method called?

include/cantera/kinetics/Coverage.h
```diff
@@ -0,0 +1,437 @@
...
+                            std::string& species) const {
+        order = m_surfaceOrder;
+        multiplier = m_multiplier;
+        species = m_stickingSpecies;
+    }
+
+    //! Specify sticking coefficients, @see getStickParameters
+    //! @internal This method is used for testing purposes only
+    void setStickParameters(double order, double multiplier,
+                            const std::string& species, bool specified=false) {
```
speth: Should we standardize on always using `sticking` in the function/class names, rather than just `stick`? There currently seems to be a mix of the two, and I think I prefer the former.

include/cantera/kinetics/Coverage.h
```diff
@@ -0,0 +1,437 @@
...
+    void setStickingSpecies(const std::string& stickingSpecies) {
+        m_stickingSpecies = stickingSpecies;
+        m_explicitSpecies = true;
+    }
+
+    //! Build rate-specific parameters based on Reaction and Kinetics context
+    //! @param rxn  Reaction associated with the sticking coefficient
+    //! @param kin  Kinetics object associated with the sticking coefficient
+    //! Parameters can be accessed using the method getStickParameters
+    void buildStickParameters(const Reaction& rxn, const Kinetics& kin);
```
speth: This is effectively `setContext`, right? Would it be better to just use that name?

src/kinetics/Coverage.cpp
```diff
@@ -0,0 +1,203 @@
...
+    const auto& surf = dynamic_cast<const SurfPhase&>(phase);
+    m_siteDensity = surf.siteDensity();
+    if (!m_explicitMotzWise) {
+        m_motzWise = kin.thermo().input().getBool("Motz-Wise", false);
+    }
+
+    // Identify the interface phase
+    size_t iInterface = npos;
+    size_t min_dim = 4;
+    for (size_t n = 0; n < kin.nPhases(); n++) {
```
speth: This should be available directly as `kin.reactionPhaseIndex()` (I realize this is just copying from `InterfaceKinetics`, but might as well simplify what we can).

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -569,12 +572,23 @@ bool InterfaceKinetics::addReaction(shared_ptr<Reaction> r_base, bool resize)
...
 {
     Kinetics::modifyReaction(i, r_base);
-    if (r_base->reaction_type == BMINTERFACE_RXN) {
-        BlowersMaselInterfaceReaction& r = dynamic_cast<BlowersMaselInterfaceReaction&>(*r_base);
-        BMSurfaceArrhenius rate = buildBMSurfaceArrhenius(i, r, true);
-        m_blowers_masel_rates.replace(i, rate);
+    if (!(r_base->usesLegacy())) {
+        shared_ptr<ReactionRate> rate = r_base->rate();
+        // Ensure that MultiBulkRate evaluator is available
+        if (m_interface_types.find(rate->type()) == m_interface_types.end()) {
```
speth: Simpler as: `if (!m_interface_types.count(rate->type()))`

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -569,12 +572,23 @@ bool InterfaceKinetics::addReaction(shared_ptr<Reaction> r_base, bool resize)
...
+            throw CanteraError("InterfaceKinetics::modifyReaction",
+                 "Evaluator not available for type '{}'.", rate->type());
+        }
+
+        // Replace reaction rate to evaluator
```
speth: 'to' seems extraneous?

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -569,12 +572,23 @@ bool InterfaceKinetics::addReaction(shared_ptr<Reaction> r_base, bool resize)
...
+        size_t index = m_interface_types[rate->type()];
+        rate->setRateIndex(i);
+        rate->setContext(*r_base.get(), *this);
```
speth: The `.get()` is unnecessary.

src/kinetics/Reaction.cpp
```diff
@@ -59,7 +60,30 @@ Reaction::Reaction(const AnyMap& node, const Kinetics& kin)
...
+        size_t nDim = kin.thermo(kin.reactionPhaseIndex()).nDim();
+        if (nDim == 3) {
+            setRate(newReactionRate(node, calculateRateCoeffUnits3(kin)));
+        } else {
+            // Reaction type is not specified
+            AnyMap rateNode = node;
+            if (!rateNode.hasKey("type")) {
+                rateNode["type"] = "Arrhenius";
+            }
+            std::string type = rateNode["type"].asString();
```
speth, line 72: I feel like this all ought to be the job of `newReactionRate`, but I do see that the problem is that it doesn't currently get the `Kinetics` object as an argument to be able to see if `nDim < 3`.

test/data/kineticsfromscratch.yaml
```diff
@@ -86,3 +98,49 @@ reactions:
 - equation: O + OH => O + OH  # Reaction 13
   type: two-temperature-plasma
   rate-constant: {A: 17283, b: -3.1}
+
+surface:
+- equation: H(S) + O(S) <=> OH(S) + PT(S)
+  rate-constant: {A: 3.7e+21, b: 0, Ea: 11500}
+  units: {length: cm, quantity: mol, activation-energy: J/mol}
```
speth: Rather than putting the same `units` entry in every reaction, you can put a `units` entry alone as the first entry in the `surface` reaction section (and this would be a good test of a form that doesn't get used all that often).

interfaces/cython/cantera/reaction.pyx
```diff
@@ -1202,7 +1460,7 @@ cdef class Reaction:
 
         .. deprecated:: 2.6
              This property is for temporary backwards-compatibility with the deprecated
-             `PlogReaction` class. Replaced by ``Reaction.rate.rates`` for reactions
+             `PlogReaction` class. Replaced by `Reaction.rate.rates` for reactions
```
speth: I don't think this formatting change is correct (likewise, many instances that follow). I think the single backtick formatting should only be used for identifiers that Sphinx is able to resolve. For things that should just be formatted as code, we should use double backticks (which I admit, is hard to do when also frequently using Markdown).

interfaces/cython/cantera/reaction.pyx
```diff
@@ -2423,110 +2762,37 @@ cdef class InterfaceReaction(ElementaryReaction):
...
 
-        if reactants and products:
-            self.reactants = reactants
-            self.products = products
-        elif equation:
-            self.reaction.setEquation(stringify(equation))
+        if init and equation and kinetics:
+            # warnings.warn("Class 'InterfaceReaction' to be removed after Cantera 2.6.\n"
+            #     "These reactions can be constructed by passing a 'InterfaceTypeRate' object "
+            #     "as the 'rate' argument to the 'Reaction' class.")
```
speth: There's a bit of commented out code here and further down in this method that should presumably be removed.

interfaces/cython/cantera/test/test_reaction.py
```diff
@@ -144,45 +144,48 @@ def tearDownClass(cls):
...
     def setUp(self):
-        self.gas.X = "H2:0.1, H2O:0.2, O2:0.7, O:1e-4, OH:1e-5, H:2e-5, H2O2:1e-7"
-        self.gas.TP = 900, 2 * ct.one_atm
-        self.gas.Te = 2300
+        self.soln.X = "H2:0.1, H2O:0.2, O2:0.7, O:1e-4, OH:1e-5, H:2e-5, H2O2:1e-7"
+        self.soln.TP = 900, 2 * ct.one_atm
+        self.soln.Te = 2300
+
+    def finalize(self, rate):
+        return rate
```
speth, line 145: This could use a bit of documentation -- It's kind of hard to follow why it exists when the base version is a no-op.

include/cantera/kinetics/ReactionData.h
```diff
@@ -282,6 +297,42 @@ struct ChebyshevData : public ReactionData
...
+
+//! Data container holding shared data for reaction rate specification with interfaces
+/**
+ * The data container CoverageData holds precalculated data common to
+ * InterfaceRate and StickingRate objects.
+ *
+ * The data container inherits from BlowersMaselData, where density is used to
+ * hold the site density [kmol/m^2].
+ */
+struct CoverageData : public BlowersMaselData
```
speth, line 312: Why does `CoverageData` need to inherit from `BlowersMaselData`? Is this just to support the case of `SurfaceBlowersMasel` rates without introducing multiple inheritance of the data classes?

speth (changes requested):
Thanks for tackling this, @ischoegl. Overall, I think most of this looks very good. I'm glad to see the elimination of so much specialization around the interface `BlowersMasel` rate, in particular, as I think that bodes well or the level of simplification we'll have once we're able to eliminate all the "legacy" rate types.

I noticed that serializing surface reactions ends up with all of the reactions having their `type` explicitly given, e.g. `type: Arrhenius-interface` even though this isn't required. Can we skip emitting these `type` fields when they aren't required?

I'm really not convinced about the need to be able to evaluate coverage-dependent rates outside the context of a `Kinetics` object. Adding this requires a whole bunch of new specializations for `ReactionData::update`, `ReactionRate::eval`, and `MultiRateBase::update`, among other things. The general solution to evaluating such rates is to provide the thermo/kinetics object in the right state, and this doesn't really seem that much more complicated as a user interface. More generally, I think the desire to evaluate rates "independently" is creating a lot of bloat and weird interfaces like having the settable `delta_enthalpy` property attached to the Blowers-Masel rate or the `site_density` being attached to every interface reaction.

include/cantera/kinetics/Coverage.h
ischoegl: Makes sense. Also changed the multiplier to use the molecular weight directly (which adds a more rigorous test).

include/cantera/kinetics/Arrhenius.h
ischoegl: I removed all the getters for temperature units from the new interface.

include/cantera/kinetics/Coverage.h
ischoegl: Renamed to `CoverageBase`. It's not a rate expression, and in a way similar to `ArrheniusBase`.

interfaces/cython/cantera/reaction.pyx
ischoegl: Good catch. These changes are courtesy of a rebase. Fixed everything, but also included the missing blank line, as well the incorrect indent for Sphinx.

include/cantera/kinetics/ReactionData.h
ischoegl: Correct. Including the partial molar enthalpies for all reactions was ultimately a lot cleaner. There may be a small speed penalty, but as this is only evaluated once per reaction type I prefer shorter code here.

include/cantera/kinetics/Coverage.h
ischoegl: This is mostly implemented as I didn't want to drop/change a pre-existing test. I am unsure to resolve
```
InputFileError thrown by AnyValue::as:
Key 'E' contains a '(anonymous namespace)::Quantity',
not a 'double'
```
for the mapped version. I added a `todo` item as a reminder for future removal.

interfaces/cython/cantera/reaction.pyx
ischoegl: This is currently used for testing purposes. I will add `experimental` decorations for several of these instances.

interfaces/cython/cantera/reaction.pyx
ischoegl: While I agree that it does make sense to not allow for this approach eventually, I have been relying on this type of access for transitional testing. My preferred way of action would be to table the discussion, while marking some of the key methods as `experimental`

include/cantera/kinetics/Arrhenius.h
ischoegl: See other `update` comment.

include/cantera/kinetics/Coverage.h
ischoegl: Good catch. I mainly was copying from `SurfaceArrhenius`, with minor adjustments in wording.

include/cantera/kinetics/Arrhenius.h
ischoegl: Removed - required some of the generated method machinery from `MultiRate`, but it worked no problem.

include/cantera/kinetics/Arrhenius.h
ischoegl: Removed.

src/kinetics/Reaction.cpp
ischoegl: Correct - there isn't enough information to differentiate between surface and bulk versions of `Arrhenius` (type completely omitted) and `Blowers-Masel` (specialization `-stick` or `-interface` omitted). I still believe that the current location is the logical one - especially as the constructor augments the information that is stripped off in `Reaction::getParameters` (which I had forgotten about but is now implemented).

ischoegl:
@speth ... thank you for the review! I believe I took care of most of the suggestions. The main issue I wasn't able to resolve is about the mapped `CoverageBase::getCoverageDependencies` output, where I'm not sure how to get around a `Quantity`-related issue.

> I noticed that serializing surface reactions ends up with all of the reactions having their type explicitly given, e.g. `type: Arrhenius-interface` even though this isn't required. Can we skip emitting these type fields when they aren't required?

Good catch - while I had planned on this, it looks like this got left out in the end. It is implemented now.

> I'm really not convinced about the need to be able to evaluate coverage-dependent rates outside the context of a `Kinetics` object. [...] The general solution to evaluating such rates is to provide the thermo/kinetics object in the right state, and this doesn't really seem that much more complicated as a user interface. [...]

I generally agree with the statement, as it ultimately just requires the `Kinetics` object, which also contains a reference to the `Phase`. At the same time, I hope to eventually revisit the `CustomRate` specialization, where there will be the option to pass `T`/`P`/`X` back and forth (the interface would look similar to what is already there, although I don't see that before 3.0). I'd prefer to just mark some of the currently used methods as `experimental` for the time being, as they are imho very useful for testing purposes (I've been relying on those extensively). Once `legacy` is removed, the dust will hopefully settle ...

include/cantera/kinetics/InterfaceRate.h
```diff
@@ -0,0 +1,487 @@
...
+ * Using Cantera nomenclature, this expression can be rewritten as
+ *  \f[
+ *      k_f = A T^b \exp \left( - \frac{E_a}{RT} \right)
+ *          \prod_k 10^{a_k \theta_k} \theta_k^{m_k}
+ *          \exp \left( \frac{- E_k \theta_k}{RT} \right)
+ *  \f]
+ * It is evident that this expression combines a regular modified Arrhenius rate
+ * expression \f$ A T^b \exp \left( - \frac{E_a}{RT} \right) \f$ with coverage-related
+ * terms, where the parameters \f$ (a_k, E_k, m_k) \f$ describe the dependency on the
+ * surface coverage of species \f$ k, \theta_k \f$. The Coverage class implements terms
```
speth: Should be `CoverageBase`?

include/cantera/kinetics/InterfaceRate.h
```diff
@@ -0,0 +1,487 @@
...
+    }
+
+    unique_ptr<MultiRateBase> newMultiRate() const override {
+        return unique_ptr<MultiRateBase>(
+            new MultiRate<StickingRate<RateType, DataType>, DataType>);
+    }
+
+    //! Identifier of reaction rate type
+    virtual const std::string type() const override {
+        return RateType::type() + "-stick";
```
speth: For the sake of readability, what about `sticking-` as a prefix, so you'd have `sticking-Arrhenius`, for example?

interfaces/cython/cantera/reaction.pyx
```diff
@@ -267,8 +256,25 @@ cdef class BlowersMaselRate(ArrheniusTypeRate):
...
+        """
+        Enthalpy change of reaction ``deltaH`` [J/kmol]
+
+        The enthalpy change of reaction is a function of temperature and thus not
+        an independent property. Accordingly, the setter should be only used for
+        testing purposes, as any value will be overwritten by an update of the
+        thermodynamic state.
+
+        **Warning:** this property is an experimental part of the Cantera API and
+        may be changed or removed without notice.
```
speth, line 269: The RST `.. warning` directive is also an option for this (though you may have tried that and rejected it based on how it ends up being formatted in HTML)

src/kinetics/ReactionData.cpp
```diff
@@ -267,4 +262,74 @@ void ChebyshevData::restore()
...
+        changed = true;
+    }
+    if (T != temperature) {
+        ReactionData::update(T);
+        sqrtT = sqrt(T);
+        changed = true;
+    }
+    if (changed || mf != m_state_mf_number) {
+        phase.getActivityConcentrations(coverages.data());
+        surf.getCoverages(coverages.data());
```
speth, line 315: The call to `getActivityConcentrations` is unnecessary, isn't it?

src/kinetics/ReactionData.cpp
```diff
@@ -267,4 +262,74 @@ void ChebyshevData::restore()
...
+        for (size_t n = 0; n < coverages.size(); n++) {
+            logCoverages[n] = std::log(std::max(coverages[n], Tiny));
+        }
+        for (size_t n = 0; n < kin.nPhases(); n++) {
+            kin.thermo(n).getPartialMolarEnthalpies(
+                grt.data() + kin.kineticsSpeciesIndex(0, n));
```
speth: Sorry I hadn't noticed this before, but the variable `BlowersMaselData::grt` should probably be named something like `partial_molar_enthalpies` if that's what it holds. The name `grt` seems like it should only be a shorthand for nondimensional Gibbs free energy.

test/kinetics/kineticsFromScratch3.cpp
```diff
@@ -308,6 +308,87 @@ TEST_F(KineticsFromScratch3, invalid_nonreactant_order)
...
+        kin_ref->getRevRateConstants(&k_ref[0]);
+        EXPECT_DOUBLE_EQ(k_ref[iRef], k[0]);
+    }
+};
+
+TEST_F(InterfaceKineticsFromScratch3, add_surface_reaction)
+{
+    // Reaction 3 on the metal surface
+    // surface_reaction( "H(m) + O(m) <=> OH(m) + (m)",
+    //                   [5.00000E+22, 0, 100.0], id = 'metal-rxn4')
```
speth: The comments here should probably reference the YAML syntax for this reaction rather than the CTI format (and below for the `add_sticking_reaction` test).

speth (changes requested):
Thanks for the updates, @ischoegl.

Besides the relatively minor in-line suggestions, while trying this out I noticed that the behavior of `ArrheniusInterfaceRate.activation_energy` and friends is very non-intuitive:
```python
>>> iface = ct.Interface('ptcombust.yaml', 'Pt_surf')
>>> R = iface.reaction(1)
>>> print(R.rate.activation_energy)
67400000.0  # this is the base value with no coverage dependence
>>> R.rate(500, iface.coverages)
>>> print(R.rate.activation_energy)
64400000.0  # now this includes the coverage dependence
```

The fact that the initial value doesn't correspond to the state of the `Interface` object is definitely surprising, and the idea that you need to evaluate the rate to get the correct "effective" value is kind of non-intuitive. Should we at least be updating the rate when you first access the Reaction object? Is there any useful way to tie this to the `Kinetics` object? I can think of some complicated ones, but I'm not sure they're worth it.

Also, it seems that there is no longer a way to get the base value of this parameter, although I guess you can always get it by looking at the `input_data`, which might be fine.

Regarding the future of `CustomRate`, I'd like to build on the `Delegator` infrastructure to create a fully-featured `ExtensibleReaction` class, but this is definitely a post-2.6 task. I think there will be some work required to make this usable without slowing things down too much.

interfaces/cython/cantera/reaction.pyx
ischoegl: Yes, the rendering is somewhat disruptive. There's probably a way to update the CSS formatting, but that's something I don't want to tackle in this PR.

include/cantera/kinetics/InterfaceRate.h
ischoegl: NP. Did the same for `interface-Arrhenius` as well as all the associated object names (i.e. `StickingArrheniusRate` etc.).

ischoegl:
@speth … thank you for the comments! I addressed all except …

> The fact that the initial value doesn't correspond to the state of the `Interface` object is definitely surprising […]. Should we at least be updating the rate when you first access the Reaction object? Is there any useful way to tie this to the Kinetics object? I can think of some complicated ones, but I'm not sure they're worth it.

The behavior is definitely not what I expected either. I can think of something (updating when adding reactions), but my first tests were not successful. I need to make sure that the state of the phase is set before reactions are added … which may or may not be already the case (fwiw, I didn’t get the expected results). I‘ll try again later today.

> Also, it seems that there is no longer a way to get the base value of this parameter, although I guess you can always get it by looking at the `input_data`, which might be fine.

This is correct, and using serialization here was exactly my rationale for eliminating a couple of methods.

ischoegl:
> [...] trying this out I noticed that the behavior of `ArrheniusInterfaceRate.activation_energy` and friends is very non-intuitive:

```python
>>> iface = ct.Interface('ptcombust.yaml', 'Pt_surf')
>>> R = iface.reaction(1)
>>> print(R.rate.activation_energy)
67400000.0  # this is the base value with no coverage dependence
>>> R.rate(500, iface.coverages)
>>> print(R.rate.activation_energy)
64400000.0  # now this includes the coverage dependence
```

> [...]  Is there any useful way to tie this to the `Kinetics` object? I can think of some complicated ones, but I'm not sure they're worth it.

@speth ... you are probably aware of this, but the behavior is due to the fact that `R.rate` holds the original `ReactionRate` object that is created when the `Reaction` is first instantiated, which is separate from the copy that `Kinetics` uses to actually evaluate the rate. In my tests, I was partially successful in updating `ReactionData` right away (the best entry point is likely around the `Kinetics::resizeReactions` specializations), but at the moment there is no way of probing the content.

A resolution would require something akin to #1051, but I'm not sure that I want to include this in 2.6.

ischoegl:
> [...] trying this out I noticed that the behavior of `ArrheniusInterfaceRate.activation_energy` and friends is very non-intuitive:

```python
>>> iface = ct.Interface('ptcombust.yaml', 'Pt_surf')
>>> R = iface.reaction(1)
>>> print(R.rate.activation_energy)
67400000.0  # this is the base value with no coverage dependence
>>> R.rate(500, iface.coverages)
>>> print(R.rate.activation_energy)
64400000.0  # now this includes the coverage dependence
```

> [...]  Is there any useful way to tie this to the `Kinetics` object? I can think of some complicated ones, but I'm not sure they're worth it.

@speth ... you are probably aware of this, but the behavior is due to the fact that `R.rate` holds the original `ReactionRate` object that is created when the `Reaction` is first instantiated, which is separate from the copy that `Kinetics` uses to actually evaluate the rate. In my tests, I was partially successful in updating `ReactionData` right away (the best entry point is likely around the `Kinetics::resizeReactions` specializations), but at the moment there is no way of probing the content.

A resolution would require something akin to #1051, but I'm not sure that I want to include this in ~2.6~ *this PR*. There may be a way to copy the initial state over, but it would be imho more important to keep things synchronized at all times. *PS: I opened #1211 as this goes well beyond the changes proposed in this PR.*

speth (approved):
Thanks for the updates, @ischoegl. The update to the automatic sticking species detection looks good to me.

I agree that the synchronization issues is better handled outside of this PR, but is something that I think we need to deal with, so thanks for creating a new issue for that.

Issue #1165, opened 2022-01-04
ck2yaml crashes when parsing a third body reaction with PLOG


corykinney:
**Problem description**

ck2yaml crashes when parsing a third body reaction with PLOG. 
I've tried searching open/closed issues for this, but either I'm searching with the wrong keywords or this is a new issue. 

**Steps to reproduce**

1. Download AramcoMech 3.0 CHEMKIN files (zip file in [supplementary materials](https://www.sciencedirect.com/science/article/pii/S0010218018303675?via%3Dihub#sec0028))
2. Run `ck2yaml --input="AramcoMech 3.0.MECH" --thermo="AramcoMech 3.0.THERM" --transport="AramcoMech 3.0.TRAN"`
3. See error:

```
Error reading reaction starting on line 5544:
"""
C4H6+M<=>C2H3+C2H3+M                                         +1.3489600E+072 -1.6550000E+001 +1.4340000E+005
PLOG /                                       +3.9473700E-002 +2.3442300E+074 -1.7690000E+001 +1.4080000E+005 /
PLOG /                                       +7.8947400E-002 +2.2908700E+074 -1.7510000E+001 +1.4210000E+005 /
PLOG /                                       +1.5789500E-001 +4.3651600E+073 -1.7130000E+001 +1.4290000E+005 /
PLOG /                                       +3.1578900E-001 +1.3489600E+072 -1.6550000E+001 +1.4340000E+005 /
"""

ERROR: Unable to parse 'AramcoMech 3.0.MECH' near line 5544:
Reaction C4H6+M<=>C2H3+C2H3+M
 contains parameters for more than one reaction type.
```

**System information**

- Cantera version: 2.6.0a3
- OS: Windows 10
- Python version: 3.8.10

ischoegl:
Hi @corykinney … thanks for reporting. To date, third-body collision partners are only supported for `ThreeBodyReaction` and `FalloffReaction`, which is why `ck2yaml` complains about this input.

corykinney:
I was afraid that it might not be supported yet. Is there a recommended way to convert mechanisms with these reactions in them, or any plans to support them in the future, @ischoegl?

ischoegl:
Not sure about a recommended way at the moment. From my perspective, adding this capability to 2.6 would make sense (e.g. most likely by allowing third-body collision partners for *any* reaction type supported by `GasKinetics`). This will likely require some discussion over at Cantera/enhancements as this would be a departure from past practice.

[...]

PR #1183, opened 2022-01-24
Eliminate unnecessary specialized Reaction types

speth:
With the introduction of the `ReactionRate` class, most of the behavior that is specialized for particular kinds of reactions is contained in those classes, but we still have a (more or less) 1:1 correspondence with classes derived from `Reaction`, even though many of them don't have any unique behavior.

- Eliminate unnecessary specializations of C++ `Reaction` class.
  - So far, this covers `ElementaryReaction`, `PlogReaction`, and `BlowersMaselReaction`, and should be easily expandable to cover `CustomFunc1Reaction` and `TwoTempPlasmaReaction`.
  - The C++ `ThreeBodyReaction`, `FalloffReaction` and `ChebyshevReaction` do implement some specialized behaviors, mostly around parsing and representing the reaction equation, so I think they need to stay for now.

- In the Python module, my goal is to eliminate all the specializations of `Reaction`, at least for homogeneous phase reactions (I think `InterfaceReaction` will probably need to stay)
  - This would be more or less analogous to how we handle the `ThermoPhase` class in Python -- with a few exceptions, there aren't wrappers for different `ThermoPhase` types. Right now, I'm a bit stuck based on one of the routes that's been introduced for constructing `Reaction` objects.

Before this PR, a `PlogReaction` could be created in one step as:
```python
R = PlogReaction(
    equation="H2 + O2 <=> 2 OH",
    rate=[(1013.25, Arrhenius(1.2124e+16, -0.5779, 45491376.8)),
          (101325., Arrhenius(4.9108e+31, -4.8507, 103649395.2)),
          (1013250., Arrhenius(1.2866e+47, -9.0246, 166508556.0)),
          (10132500., Arrhenius(5.9632e+56, -11.529, 220076726.4))],
    kinetics=gas)
```
Or in a two steps:
```python
R = PlogReaction({"H2": 1, "O2": 1}, {"OH": 2})
R.rate = PlogRate(
    [(1013.25, Arrhenius(1.2124e+16, -0.5779, 45491376.8)),
     (101325., Arrhenius(4.9108e+31, -4.8507, 103649395.2)),
     (1013250., Arrhenius(1.2866e+47, -9.0246, 166508556.0)),
     (10132500., Arrhenius(5.9632e+56, -11.529, 220076726.4))]
```

With this PR, the first example becomes:
```python
R = Reaction(
    equation="H2 + O2 <=> 2 OH",
    rate=PlogRate([(1013.25, Arrhenius(1.2124e+16, -0.5779, 45491376.8)),
          (101325., Arrhenius(4.9108e+31, -4.8507, 103649395.2)),
          (1013250., Arrhenius(1.2866e+47, -9.0246, 166508556.0)),
          (10132500., Arrhenius(5.9632e+56, -11.529, 220076726.4))]),
    kinetics=gas)
```
which is pretty similar.

The second case becomes:
```python
R = ct.Reaction({"H2": 1, "O2": 1}, {"OH": 2})
R.rate = ct.PlogRate(
    [(1013.25, ct.Arrhenius(1.2124e+16, -0.5779, 45491376.8)),
     (101325., ct.Arrhenius(4.9108e+31, -4.8507, 103649395.2)),
     (1013250., ct.Arrhenius(1.2866e+47, -9.0246, 166508556.0)),
     (10132500., ct.Arrhenius(5.9632e+56, -11.529, 220076726.4))])
```
This is fine in the case of `PlogRate`, where the base C++ reaction type is just `Reaction` but presents a problem for the reaction types where we still need C++ specializations of the `Reaction` class, as there's no information about the rate type available in the initial constructor.

What I'm wondering is whether we need this second constructor, where no rate information is provided. Would it make sense require specifying at least an empty `ReactionRate` object of the correct type here, e.g.
```python
R = ct.Reaction({"H2": 1, "O2": 1}, {"OH": 2}, rate=ct.PlogRate())
```
I'm also wondering if there's a good way to clean up the split between specifying the reactants and products separately versus specifying the reaction equation in these constructors.

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/examples/kinetics/blowers_masel.py
+++ b/interfaces/cython/cantera/examples/kinetics/blowers_masel.py
@@ -23,7 +23,7 @@ import numpy as np
 import matplotlib.pyplot as plt
 
 #Create an elementary reaction O+H2<=>H+OH
-r1 = ct.ElementaryReaction({'O':1, 'H2':1}, {'H':1, 'OH':1})
+r1 = ct.Reaction({"O":1, "H2":1}, {"H":1, "OH":1})
 r1.rate = ct.ArrheniusRate(3.87e1, 2.7, 6260*1000*4.184)
 
 #Create a Blowers-Masel reaction O+H2<=>H+OH
--- a/interfaces/cython/cantera/examples/kinetics/blowers_masel.py
+++ b/interfaces/cython/cantera/examples/kinetics/blowers_masel.py
@@ -27,12 +27,12 @@ r1 = ct.Reaction({"O":1, "H2":1}, {"H":1, "OH":1})
 r1.rate = ct.ArrheniusRate(3.87e1, 2.7, 6260*1000*4.184)
 
 #Create a Blowers-Masel reaction O+H2<=>H+OH
-r2 = ct.BlowersMaselReaction({'O':1, 'H2':1}, {'H':1, 'OH':1})
+r2 = ct.Reaction({"O":1, "H2":1}, {"H":1, "OH":1})
 r2.rate = ct.BlowersMaselRate(3.87e1, 2.7, 6260*1000*4.184, 1e9)
 
 #Create a Blowers-Masel reaction with same parameters with r2
 #reaction equation is H+CH4<=>CH3+H2
-r3 = ct.BlowersMaselReaction({'H':1, 'CH4':1}, {'CH3':1, 'H2':1})
+r3 = ct.Reaction({"H":1, "CH4":1}, {"CH3":1, "H2":1})
 r3.rate = ct.BlowersMaselRate(3.87e1, 2.7, 6260*1000*4.184, 1e9)
 
 gas = ct.Solution(thermo='IdealGas', kinetics='GasKinetics',
--- a/interfaces/cython/cantera/examples/kinetics/blowers_masel.py
+++ b/interfaces/cython/cantera/examples/kinetics/blowers_masel.py
@@ -23,17 +23,17 @@ import numpy as np
 import matplotlib.pyplot as plt
 
 #Create an elementary reaction O+H2<=>H+OH
-r1 = ct.Reaction({"O":1, "H2":1}, {"H":1, "OH":1})
-r1.rate = ct.ArrheniusRate(3.87e1, 2.7, 6260*1000*4.184)
+r1 = ct.Reaction({"O":1, "H2":1}, {"H":1, "OH":1},
+                 ct.ArrheniusRate(3.87e1, 2.7, 6260*1000*4.184))
 
 #Create a Blowers-Masel reaction O+H2<=>H+OH
-r2 = ct.Reaction({"O":1, "H2":1}, {"H":1, "OH":1})
-r2.rate = ct.BlowersMaselRate(3.87e1, 2.7, 6260*1000*4.184, 1e9)
+r2 = ct.Reaction({"O":1, "H2":1}, {"H":1, "OH":1},
+                 ct.BlowersMaselRate(3.87e1, 2.7, 6260*1000*4.184, 1e9))
 
 #Create a Blowers-Masel reaction with same parameters with r2
 #reaction equation is H+CH4<=>CH3+H2
-r3 = ct.Reaction({"H":1, "CH4":1}, {"CH3":1, "H2":1})
-r3.rate = ct.BlowersMaselRate(3.87e1, 2.7, 6260*1000*4.184, 1e9)
+r3 = ct.Reaction({"H":1, "CH4":1}, {"CH3":1, "H2":1},
+                 ct.BlowersMaselRate(3.87e1, 2.7, 6260*1000*4.184, 1e9))
 
 gas = ct.Solution(thermo='IdealGas', kinetics='GasKinetics',
                    species=ct.Solution('gri30.yaml').species(), reactions=[r1, r2, r3])
--- a/interfaces/cython/cantera/reaction.pyx
+++ b/interfaces/cython/cantera/reaction.pyx
@@ -782,8 +782,8 @@ cdef class Reaction:
     _has_legacy = False
     _hybrid = False # indicate whether legacy implementations are separate or merged
 
-    def __cinit__(self, reactants="", products="", init=True, legacy=False, **kwargs):
-
+    def __cinit__(self, reactants="", products="", rate=None, *, legacy=False,
+                  init=True, **kwargs):
         if init:
             rxn_type = self._reaction_type
             if (not self._hybrid and self._has_legacy) or (self._hybrid and legacy):
@@ -795,16 +795,28 @@ cdef class Reaction:
             if products:
                 self.products = products
 
-    def __init__(self, reactants=None, products=None, *, equation=None, rate=None,
+    def __init__(self, reactants=None, products=None, rate=None, *, equation=None,
                  Kinetics kinetics=None, init=True, legacy=False, **kwargs):
 
-        if legacy:
+        if legacy or not init:
+            return
+
+        if not rate:
+            raise TypeError("No *rate* argument provided")
+
+        if isinstance(rate, ReactionRate):
+            if reactants and products:
+                self.reactants = reactants
+                self.products = products
+            else:
+                self.reaction.setEquation(stringify(equation))
+            self.reaction.setRate((<ReactionRate>rate)._rate)
             return
 
         if reactants and products and not equation:
             equation = self.equation
 
-        if init and equation and kinetics:
+        if equation and kinetics:
             rxn_type = self._reaction_type
             spec = {"equation": equation, "type": rxn_type}
             if isinstance(rate, dict):
@@ -818,10 +830,7 @@ cdef class Reaction:
                                             deref(kinetics.kinetics))
             self.reaction = self._reaction.get()
 
-            if isinstance(rate, ReactionRate):
-                self.rate = rate
-                rate = None
-            elif rate is not None:
+            if rate is not None:
                 try:
                     if isinstance(rate[0][1], Arrhenius):
                         self.rate = PlogRate(rate)
@@ -1606,8 +1615,11 @@ cdef class ElementaryReaction(Reaction):
             raise AttributeError("Incorrect accessor for legacy implementation")
         return <CxxElementaryReaction2*>self.reaction
 
-    def __init__(self, equation=None, rate=None, Kinetics kinetics=None,
-                 init=True, **kwargs):
+    def __init__(self, reactants=None, products=None, rate=None, *, equation=None,
+                 Kinetics kinetics=None, init=True, **kwargs):
+
+        if reactants and products and not equation:
+            equation = self.equation
 
         if init and equation and kinetics:
             rxn_type = self._reaction_type + "-legacy"
@@ -1625,8 +1637,8 @@ cdef class ElementaryReaction(Reaction):
                                             deref(kinetics.kinetics))
             self.reaction = self._reaction.get()
 
-            if isinstance(rate, Arrhenius):
-                self.rate = rate
+        if isinstance(rate, (Arrhenius, ArrheniusRate)):
+            self.rate = rate
 
     cdef _legacy_set_rate(self, Arrhenius rate):
         cdef CxxElementaryReaction2* r = self.cxx_object2()
@@ -1727,11 +1739,27 @@ cdef class ThreeBodyReaction(ElementaryReaction):
             return &(self.cxx_threebody2().third_body)
         return <CxxThirdBody*>(self.cxx_threebody().thirdBody().get())
 
-    def __init__(self, equation=None, rate=None, efficiencies=None,
-                 Kinetics kinetics=None, legacy=False, init=True, **kwargs):
+    def __init__(self, reactants="", products="", rate=None, *, equation=None,
+                 efficiencies=None, Kinetics kinetics=None, legacy=False, init=True,
+                 **kwargs):
+
+        if reactants and products and not equation:
+            equation = self.equation
+
+        if isinstance(rate, ArrheniusRate):
+            self._reaction.reset(new CxxThreeBodyReaction3())
+            self.reaction = self._reaction.get()
+            if reactants and products:
+                self.reactants = reactants
+                self.products = products
+            else:
+                self.reaction.setEquation(stringify(equation))
+            self.reaction.setRate((<ReactionRate>rate)._rate)
+            if efficiencies:
+                self.efficiencies = efficiencies
+            return
 
         if init and equation and kinetics:
-
             rxn_type = self._reaction_type
             if legacy:
                 rxn_type += "-legacy"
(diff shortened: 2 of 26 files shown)


ischoegl:
@speth ... I am overall in favor of this change, which moves along discussions we had during earlier PR's. My main question was about timing: should this be adopted prior to 2.6, or completed for 3.0? My assumption was to remove the old framework first, which would make a lot of specializations obsolete (this applies more to Python than to C++). In C++ we can move forward, and I do believe that some of the specializations can be further removed.

Regarding
```Python
R = Reaction(
    equation="H2 + O2 <=> 2 OH",
    rate=PlogRate([(1013.25, Arrhenius(1.2124e+16, -0.5779, 45491376.8)),
          (101325., Arrhenius(4.9108e+31, -4.8507, 103649395.2)),
          (1013250., Arrhenius(1.2866e+47, -9.0246, 166508556.0)),
          (10132500., Arrhenius(5.9632e+56, -11.529, 220076726.4))]),
    kinetics=gas)
```
it was my impression that this should work already? (I believe this should be covered by `test_reaction.py`).

Regarding the second constructor
```Python
R = ct.Reaction({"H2": 1, "O2": 1}, {"OH": 2}, rate=ct.PlogRate())
```
we can certainly enforce this, but for some of the legacy reactions, there is no dedicated `rate` object. So at least to me, the two-parameter version mostly exists for historical reasons.

ischoegl:
@speth … I ended up deleting my earlier comment as I’m curious what you’re coming up with. I believe that `Reaction` specializations can be removed completely in C++ relatively soon (and in Python after 2.6). From what I’m seeing over in #1181, this includes `InterfaceReactions` (Which means that `ReactionFactory` would be gone also). All we’d have to do is to track dimensionality, which is easy based on a `ReactionRate` when defined in the API (or phase in the context of YAML input).

speth:
When I started this, my hope was that it would lead to enabling removal of all the `Reaction` specializations in  C++ and along with that the `ReactionFactory` class, but I'm not so sure about that right now - there's a fair bit of specialization related to handling of third bodies and efficiencies for `ThreeBodyReaction` and `FalloffReaction`, and I don't know if we want to actually move that up to the base `Reaction` class, and I don't think it would make sense to make the `ReactionRate` classes know anything about the reactants / products. I'd be interested in any ideas you have that might make this possible, though.

ischoegl:
>  I don't think it would make sense to make the `ReactionRate` classes know anything about the reactants / products. 

Agreed. But for setup purposes it may be necessary to let them have a peek (without saving anything)

> I'd be interested in any ideas you have that might make this possible, though.

... sure! For interface reactions, I had to do a fair amount of parsing when it comes to `StickingRate` (I split up `SurfaceArrhenius` as keeping everything together made little sense to me). The hook I came up with is https://github.com/Cantera/cantera/blob/d211185afd6a310976cfe86d145bf9f5297eddfd/include/cantera/kinetics/ReactionRate.h#L114-L116
I am planning to use the same to eliminate `ElectrochemicalReaction` (and the double-parsing required for their detection).

FWIW, here's the (preliminary) implementation used for `StickingRate::setContext` ('lifted' from `InterfaceKinetics`)
https://github.com/Cantera/cantera/blob/d211185afd6a310976cfe86d145bf9f5297eddfd/src/kinetics/Coverage.cpp#L176-L245

This function is called when the reaction is attached to a `Kinetics` object.

**PS:** 

> there's a fair bit of specialization related to handling of third bodies and efficiencies for `ThreeBodyReaction` and `FalloffReaction`, and I don't know if we want to actually move that up to the base `Reaction` class.

The situation is not too dissimilar to coverage dependence for `InterfaceReaction`. For that case, I ended up handling these by `ReactionRate` specializations.

test/data/kineticsfromscratch.yaml
```diff
@@ -24,6 +24,7 @@ phases:
 reactions:
 - equation: O + H2 <=> H + OH  # Reaction 1
   rate-constant: {A: 38.7, b: 2.7, Ea: 6260.0}
+  duplicate: true  # duplicate of Reaction 7
```
ischoegl: Do we need to re-think the approach used for checking of duplicate reactions? Are reaction expressions distinct if there is a mismatch of reaction rate type?

src/kinetics/Reaction.cpp
```diff
@@ -200,6 +202,14 @@ void Reaction::setRate(shared_ptr<ReactionRate> rate)
...
         m_rate = rate;
     }
+
+    if (reactants.count("(+M)")) {
+        warn_deprecated("Chebyshev reaction equation", input, "Specifying '(+M)' "
+            "in the reaction equation for Chebyshev reactions is deprecated.");
+        // remove optional third body notation
+        reactants.erase("(+M)");
+        products.erase("(+M)");
+    }
```
ischoegl, line 225: Behavior should be specific to `ChebyshevRate3`; as an aside, I am wondering whether #1165 could be resolved by a similar 'fix' (both here and in the conversion scripts).

src/kinetics/ReactionFactory.cpp
```diff
@@ -22,11 +23,20 @@ std::mutex ReactionFactory::reaction_mutex;
...
-        return new ElementaryReaction3(node, kin);
+    reg("reaction", [](const AnyMap& node, const Kinetics& kin) {
+        unique_ptr<Reaction> R(new Reaction());
+        R->setParameters(node, kin);
+        if (kin.nPhases()) {
+            R->setRate(newReactionRate(node, R->calculateRateCoeffUnits3(kin)));
+        } else {
+            R->setRate(newReactionRate(node));
+        }
+        return R.release();
```
ischoegl: Could this behavior be absorbed by a 'smarter' constructor `Reaction(const AnyMap& node, const Kinetics& kin)`? I am cautiously hopeful that `ReactionFactory` will no longer be necessary once XML/CTI are removed.

interfaces/cython/cantera/examples/kinetics/blowers_masel.py
```diff
@@ -23,16 +23,16 @@
 import matplotlib.pyplot as plt
 
 #Create an elementary reaction O+H2<=>H+OH
-r1 = ct.ElementaryReaction({'O':1, 'H2':1}, {'H':1, 'OH':1})
+r1 = ct.Reaction({"O":1, "H2":1}, {"H":1, "OH":1})
```
ischoegl: I think that
```Python
r1 = ct.Reaction({"O":1, "H2":1}, {"H":1, "OH":1})
r1.rate = ct.ArrheniusRate(3.87e1, 2.7, 6260*1000*4.184)
```
could be combined to
```Python
r1 = ct.Reaction({"O":1, "H2":1}, {"H":1, "OH":1}, 
    ct.ArrheniusRate(3.87e1, 2.7, 6260*1000*4.184))
```
for clarity.

Same goes for instances below and similar occurrences in `test_kinetics.py`.

interfaces/cython/cantera/test/test_reaction.py
```diff
@@ -707,7 +711,11 @@ def from_rate(self, rate):
     def from_parts(self):
         # create reaction rate object from parts
         orig = self.gas.reaction(self._index)
-        rxn = self._cls(orig.reactants, orig.products, legacy=self._legacy)
+        if self._rate_cls:
+            rxn = self._cls(orig.reactants, orig.products, rate=self._rate_cls(),
+                            legacy=self._legacy)
```
ischoegl: Should use `rate=self._rate_obj` directly.

interfaces/cython/cantera/test/test_reaction.py
```diff
@@ -804,7 +815,7 @@ def test_no_rate(self):
         # check behavior for instantiation from keywords / no rate
         if self._rate_obj is None:
             return
-        rxn = self.from_rate(None)
+        rxn = self.from_rate(self._rate_cls() if self._rate_cls else None)
```
ischoegl: I assume this is necessary as an empty type is needed to check that uninitialized rate objects work correctly?

I actually think that this PR will make uninitialized `ReactionRate` objects unnecessary. I.e. if no rate is attached, evaluation should fail. Likewise, generation of partial YAML input (e.g. `equation` without rate information) can be removed as most of the YAML content is now attached to the rate object anyways.

interfaces/cython/cantera/reaction.pyx
```diff
@@ -795,6 +795,44 @@ cdef class Reaction:
             if products:
                 self.products = products
 
+    def __init__(self, reactants=None, products=None, *, equation=None, rate=None,
+                 Kinetics kinetics=None, init=True, legacy=False, **kwargs):
```
ischoegl: Would it make sense to write this as
```Python
    def __init__(self, reactants=None, products=None, rate=None, *, equation=None, 
                 Kinetics kinetics=None, init=True, legacy=False, **kwargs):
```
so the default sequence matches that of the C++ call?

Also see comment in `blowers_masel.py`.

interfaces/cython/cantera/reaction.pyx
```diff
@@ -795,6 +795,44 @@ cdef class Reaction:
...
+
+        if legacy:
+            return
+
+        if reactants and products and not equation:
+            equation = self.equation
```
ischoegl: This leaves the object in an interesting state. Should there be any tests? I.e. what does calling the `rate` property getter do?

interfaces/cython/cantera/reaction.pyx
```diff
@@ -795,6 +795,44 @@ cdef class Reaction:
...
+
+        if init and equation and kinetics:
+            rxn_type = self._reaction_type
+            spec = {"equation": equation, "type": rxn_type}
+            if isinstance(rate, dict):
+                # Arrhenius-like rates
+                spec["rate-constant"] = rate
+                rate = None
```
ischoegl: this should use `ReactionRate.from_dict` instead.

interfaces/cython/cantera/reaction.pyx
```diff
@@ -795,6 +795,44 @@ cdef class Reaction:
...
+            if isinstance(rate, ReactionRate):
+                self.rate = rate
+                rate = None
+            elif rate is not None:
+                try:
+                    if isinstance(rate[0][1], Arrhenius):
+                        self.rate = PlogRate(rate)
+                        rate = None
+                except (IndexError, TypeError, KeyError):
+                    pass
```
ischoegl: with the exception of a dictionary (see comment above) and the potential exception of a (YAML-compatible) string, I believe all other options should be removed. For `PlogRate`, it may make sense to convert the output from the nested list to a dictionary (an idea that I played with, but ultimately dropped for reasons of back-ward compatibility).

ischoegl (changes requested):
@speth ... this looks very good as it does cut a substantial amount of clutter that accumulated as Cantera/enhancements#87 progressed: you were even able to untangle the Python API, which is something I though would have to be deferred!

I have a couple of relatively small comments. The main question I have is whether the instantiation of a `ReactionRate` object shouldn't be called from within `Reaction.h` rather than `ReactionFactory`, as I have some hopes that we can eliminate the latter eventually.

Regarding merges, I believe this can be safely merged around the same time as #1166, and I will build on this to rework #1181 (which would otherwise introduce a slew of new objects that would be immediately removed). 

Beyond, I think it may make sense to have another look at `ThirdBodyCalc` as this is the only obstacle I see to removing some of the remaining reaction specializations. Of course, some changes will also be necessary for the science section on Cantera/cantera-website :cry:

**PS:** ugh, it looks like all of my multi-line comments were converted to regular comments (I reviewed from VSCode this time).

test/data/kineticsfromscratch.yaml
speth: Yes, but I think this is a bigger issue than this PR. I've created a new enhancement post detailing some of the thoughts that I've been entertaining on this topic (Cantera/enhancements#132).

src/kinetics/Reaction.cpp
speth: Oops. I had that and lost it in a rebase option, I guess.

interfaces/cython/cantera/examples/kinetics/blowers_masel.py
speth: Are you suggesting that we make `rate` a required argument to the `Reaction` constructor? I certainly think that would simplify matters.

test/data/kineticsfromscratch.yaml
ischoegl: I think the new enhancement request goes in the right direction, but for the time being, these reactions should not be labeled as a duplicate as the rate parameterizations differ (according to the 'traditional' approach)

interfaces/cython/cantera/examples/kinetics/blowers_masel.py
ischoegl: I wasn't necessarily suggesting this, but it would probably avoid undesirable edge cases that need to be caught otherwise.

interfaces/cython/cantera/test/test_reaction.py
speth: Without this, the tests of the `__call__` method fail for the rate types where the `__call__` method takes two parameters, but the default `ReactionRate` object is an `ArrheniusRate` that only takes one.

ischoegl: Ah, ok. I think we can leave the handling of empty rate parameterizations for another time (but a comment should be added) … *edit: actually, requiring rate as a parameter should resolve this?*

interfaces/cython/cantera/reaction.pyx
ischoegl: Per your comment above, requiring rate as a parameter would resolve this edge case

interfaces/cython/cantera/test/test_reaction.py
speth: ~~Well, it would resolve it in the sense that `None` would not be a viable argument to `from_rate`.~~
I put in a comment in `from_rate` that I hope explains this well enough. I'm looking forward to getting rid of the "legacy" rates so we can reduce the complexity of both the actual implementations and this testing infrastructure.

interfaces/cython/cantera/reaction.pyx
speth: Right now, this gives you a `nan`-initialized `ArrheniusRate` object. I think that ultimately stems from the fact that not specifying a reaction `type` implies an elementary/Arrhenius rate.

ischoegl:
@speth ... I thought about this some more over the weekend, and ended up opening Cantera/enhancements#133.

Beyond, to unwind the remaining tasks pertaining to Cantera/enhancements#87, I believe this PR should go first, then #1184, then #1181, and finally whatever we decide for Cantera/enhancements#133. The outcome would be complete removal of `Reaction` specializations.

interfaces/cython/cantera/examples/kinetics/blowers_masel.py
speth: Well, I went ahead with implementing that option, so we'll see if that ends up improving things.

interfaces/cython/cantera/reaction.pyx
speth: I've adopted this as the signature for all non-deprecated `Reaction` types, as part of making the `rate` argument required.

speth:
@ischoegl - I think I've addressed all of your review comments on this.

ischoegl (approved):
@speth ... thank you for addressing the comments! This looks good to go - I believe #1184 should be next. As an aside, I am likewise looking forward to removing the legacy bits from the testing framework - but it does give me some peace of mind until the transition is complete.

PR #1184, opened 2022-01-28
Preamble to InterfaceKinetics refactoring

### LINKED ISSUE #131: Optionally disable legacy reactions in Cantera 2.6
**Abstract**

At the moment, there already is a preprocessor flag `CT_NO_LEGACY_REACTIONS_26` introduced in `Reaction.h`, although it remains untested (at the moment, I am aware of at least one instance where toggling this flag will result in compilation errors). In order to ensure that the new framework introduced in #87 works without legacy support, the use of this preprocessor flag (or an additional `CT_NO_LEGACY_26`) could be extended to completely eliminate legacy reactions (as well as various CTI/XML tags). 

The flag could be adopted in `SCons`

**Motivation**

Describe the need for the work being done:

* What problem is it trying to solve? ... ensure that changes introduced in 2.6 are feature complete and do not rely on legacy objects.
* Who is affected by the change? ... early adopters and developers
* Why is this a good solution? ... the preprocessor will allow for CI tests without legacy support prior to the 2.6 release; deprecated code is automatically marked. As 2.6 is the last release to include legacy features, this will add a layer ensuring that everything is ready for the transition to 3.0.

**Alternatives**

Just make sure that toggling the flag to change defaults works (while still adding a CI runner that covers this configuration).

*Edit: Changes proposed in Cantera/cantera#1184 largely address this issue, as remaining 'hybrid' implementations no longer use legacy objects.*

ischoegl:

This PR seeks to address several issues that came up while working on #1181.

- Fix compilation errors if `NO_LEGACY_REACTIONS_26` is set
- Streamline nomenclature for `activationEnergy` in the context of `BlowersMasel` ... use `effectiveActivationEnergy` for version that includes dependency on thermodynamic state (same will be introduced for `ReactionRate` classes that handle interface reactions).
- Limit use of `Arrhenius` ... clarify as `Arrhenius2` in legacy framework and make it equivalent with `ArrheniusBase` in new framework (`ArrheniusBase` should be renamed to `Arrhenius` after Cantera 2.6).  *Note: Defining a reduced `Arrhenius(Base)` class will also allow for additional templated classes - see e.g. Cantera/enhancements#133.*
- Remove leftover '3' from `ChebyshevRate(3)`
- Calculate `deltaH` within `BlowersMasel` without requiring associated `Kinetics` object

* Preamble to #1181 
* Closes Cantera/enhancements#131

This PR will allow to disable legacy reactions (actually `typedef`s) from `SCons` as
```
$ scons build no_legacy_reactions=y
```

**Checklist**

### DIFF
--- a/include/cantera/kinetics/Arrhenius.h
+++ b/include/cantera/kinetics/Arrhenius.h
@@ -71,10 +71,6 @@ public:
     //! Return parameters
     void getRateParameters(AnyMap& node) const;
 
-    //! Return parameters - required for legacy framework
-    //! @todo: merge with single-parameter version after removal of old framework
-    virtual void getParameters(AnyMap& node, const Units& rate_units) const;
-
     //! Check rate expression
     void checkRate(const std::string& equation, const AnyMap& node);
 
--- a/include/cantera/kinetics/Arrhenius.h
+++ b/include/cantera/kinetics/Arrhenius.h
@@ -75,11 +75,17 @@ public:
     void checkRate(const std::string& equation, const AnyMap& node);
 
     //! Evaluate reaction rate
+    /*!
+     *  @internal  Non-virtual method that should not be overloaded
+     */
     double evalRate(double logT, double recipT) const {
         return m_A * std::exp(m_b * logT - m_Ea_R * recipT);
     }
 
     //! Evaluate natural logarithm of the rate constant.
+    /*!
+     *  @internal  Non-virtual method that should not be overloaded
+     */
     double evalLog(double logT, double recipT) const {
         return m_logA + m_b * logT - m_Ea_R * recipT;
     }
@@ -189,7 +195,7 @@ public:
 
     virtual void getParameters(AnyMap& node) const override;
 
-    void check(const std::string& equation, const AnyMap& node) {
+    void check(const std::string& equation, const AnyMap& node) override {
         checkRate(equation, node);
     }
 
@@ -208,11 +214,6 @@ public:
      */
     virtual double ddTScaledFromStruct(const ReactionData& shared_data) const {
         return (m_Ea_R * shared_data.recipT + m_b) * shared_data.recipT;
-    }
-
-    //! Return the activation energy *Ea* [J/kmol]
-    double activationEnergy() const {
-        return m_Ea_R * GasConstant;
     }
 
     //! Return the activation energy divided by the gas constant (i.e. the
@@ -282,7 +283,7 @@ public:
 
     virtual void getParameters(AnyMap& node) const override;
 
-    void check(const std::string& equation, const AnyMap& node) {
+    void check(const std::string& equation, const AnyMap& node) override {
         checkRate(equation, node);
     }
 
@@ -389,7 +390,7 @@ public:
 
     virtual void getParameters(AnyMap& node) const;
 
-    void check(const std::string& equation, const AnyMap& node) {
+    void check(const std::string& equation, const AnyMap& node) override {
         checkRate(equation, node);
     }
 
@@ -402,7 +403,7 @@ public:
     void updateFromStruct(const BlowersMaselData& shared_data) {
         if (shared_data.ready) {
             m_deltaH_R = 0.;
-            for (const auto& item : m_multipliers) {
+            for (const auto& item : m_stoich_coeffs) {
                 m_deltaH_R += shared_data.grt[item.first] * item.second;
             }
             m_deltaH_R /= GasConstant;
@@ -461,7 +462,7 @@ public:
 
 protected:
     //! Pairs of species indices and multiplers to calculate enthalpy change
-    std::vector<std::pair<size_t, double>> m_multipliers;
+    std::vector<std::pair<size_t, double>> m_stoich_coeffs;
 
     double m_deltaH_R; //!< enthalpy change of reaction (in temperature units)
 };
--- a/include/cantera/kinetics/Arrhenius.h
+++ b/include/cantera/kinetics/Arrhenius.h
@@ -173,7 +173,7 @@ public:
     }
 
     unique_ptr<MultiRateBase> newMultiRate() const override {
-        return unique_ptr<MultiRateBase>(new MultiBulkRate<ArrheniusRate, ArrheniusData>);
+        return unique_ptr<MultiRateBase>(new MultiRate<ArrheniusRate, ArrheniusData>);
     }
 
     //! Identifier of reaction rate type
@@ -257,7 +257,7 @@ public:
 
     unique_ptr<MultiRateBase> newMultiRate() const override {
         return unique_ptr<MultiRateBase>(
-            new MultiBulkRate<TwoTempPlasmaRate, TwoTempPlasmaData>);
+            new MultiRate<TwoTempPlasmaRate, TwoTempPlasmaData>);
     }
 
     //! Constructor based on AnyMap content
@@ -370,7 +370,7 @@ public:
 
     unique_ptr<MultiRateBase> newMultiRate() const {
         return unique_ptr<MultiRateBase>(
-            new MultiBulkRate<BlowersMaselRate, BlowersMaselData>);
+            new MultiRate<BlowersMaselRate, BlowersMaselData>);
     }
 
     //! Constructor based on AnyMap content
--- a/include/cantera/kinetics/Arrhenius.h
+++ b/include/cantera/kinetics/Arrhenius.h
@@ -26,6 +26,16 @@ class AnyMap;
  */
 
 //! Base class for Arrhenius-type Parameterizations
+/*!
+ * This base class provides a minimally functional interface that allows for parameter
+ * access from derived classes as well as classes that use Arrhenius-type expressions
+ * internally, for example FalloffRate and PlogRate.
+ *
+ * @todo supersedes Arrhenius2 and will replace Arrhenius(2) after Cantera 2.6,
+ *      The class should be renamed to Arrhenius after removal of Arrhenius2. The new
+ *      behavior can be forced in self-compiled Cantera installations by defining
+ *      CT_NO_LEGACY_REACTIONS_26 via the 'no_legacy_reactions' option in SCons.
+ */
 class ArrheniusBase
 {
 public:
@@ -138,16 +148,6 @@ protected:
     std::string m_b_str = "b"; //!< The string for temperature exponent
     std::string m_Ea_str = "Ea"; //!< The string for activation energy
     Units m_rate_units; //!< Reaction rate units
-};
-
-
-/*!
- * @todo supersedes Arrhenius2: replace existing instances with this class, rename,
- *      and deprecate Arrhenius3.
- */
-class Arrhenius3 final : public ArrheniusBase
-{
-    using ArrheniusBase::ArrheniusBase; // inherit constructors
 };
 
 
--- a/include/cantera/kinetics/Arrhenius.h
+++ b/include/cantera/kinetics/Arrhenius.h
@@ -99,10 +99,11 @@ public:
         return m_b;
     }
 
-    //! Return the intrinsic activation energy *Ea* [J/kmol]
+    //! Return the activation energy *Ea* [J/kmol]
     //! The value corresponds to the constant specified by input parameters;
-    //! in the simplest case, this is equal to the activation energy itself
-    double intrinsicActivationEnergy() const {
+    //! class specializations may provide alternate definitions that describe
+    //! an effective activation energy that depends on the thermodynamic state.
+    double activationEnergy() const {
         return m_Ea_R * GasConstant;
     }
 
@@ -423,7 +424,7 @@ public:
      */
     virtual double ddTScaledFromStruct(const BlowersMaselData& shared_data) const;
 
-    //! Return the actual activation energy (a function of the delta H of reaction)
+    //! Return the effective activation energy (a function of the delta H of reaction)
     //! divided by the gas constant (i.e. the activation temperature) [K]
     double activationEnergy_R(double deltaH_R) const {
         if (deltaH_R < -4 * m_Ea_R) {
@@ -438,17 +439,12 @@ public:
             (vp * vp - 4 * m_w_R * m_w_R + deltaH_R * deltaH_R); // in Kelvin
     }
 
-    //! Return the actual activation energy [J/kmol]
+    //! Return the effective activation energy [J/kmol]
     /*!
      *  @param deltaH  Enthalpy change of reaction [J/kmol]
      */
-    double activationEnergy(double deltaH) const {
+    double effectiveActivationEnergy(double deltaH) const {
         return activationEnergy_R(deltaH / GasConstant) * GasConstant;
-    }
-
-    //! Return the intrinsic activation energy [J/kmol]
-    double activationEnergy0() const {
-        return m_Ea_R * GasConstant;
     }
 
     //! Return the bond dissociation energy *w* [J/kmol]
--- a/include/cantera/kinetics/Arrhenius.h
+++ b/include/cantera/kinetics/Arrhenius.h
@@ -194,7 +194,9 @@ public:
     }
 
     //! Evaluate reaction rate
-    //! @param shared_data  data shared by all reactions of a given type
+    /*!
+     *  @param shared_data  data shared by all reactions of a given type
+     */
     double evalFromStruct(const ReactionData& shared_data) const {
         return m_A * std::exp(m_b * shared_data.logT - m_Ea_R * shared_data.recipT);
     }
@@ -284,6 +286,10 @@ public:
         checkRate(equation, node);
     }
 
+    //! Evaluate reaction rate
+    /*!
+     *  @param shared_data  data shared by all reactions of a given type
+     */
     double evalFromStruct(const TwoTempPlasmaData& shared_data) {
         // m_E4_R is the electron activation (in temperature units)
         return m_A * std::exp(m_b * shared_data.logTe -
@@ -387,14 +393,30 @@ public:
         checkRate(equation, node);
     }
 
-    double evalFromStruct(const BlowersMaselData& shared_data) {
-        double deltaH_R;
+    virtual void setContext(const Reaction& rxn, const Kinetics& kin) override;
+
+    //! Update information specific to reaction
+    /*!
+     *  @param shared_data  data shared by all reactions of a given type
+     */
+    void updateFromStruct(const BlowersMaselData& shared_data) {
         if (shared_data.ready) {
-            deltaH_R = shared_data.dH[m_rate_index] / GasConstant;
+            m_deltaH_R = 0.;
+            for (const auto& item : m_multipliers) {
+                m_deltaH_R += shared_data.grt[item.first] * item.second;
+            }
+            m_deltaH_R /= GasConstant;
         } else {
-            deltaH_R = shared_data.dH[0] / GasConstant;
+            m_deltaH_R = shared_data.dH_direct / GasConstant;
         }
-        double Ea_R = activationEnergy_R(deltaH_R);
+    }
+
+    //! Evaluate reaction rate
+    /*!
+     *  @param shared_data  data shared by all reactions of a given type
+     */
+    double evalFromStruct(const BlowersMaselData& shared_data) {
+        double Ea_R = activationEnergy_R(m_deltaH_R);
         return m_A * std::exp(m_b * shared_data.logT - Ea_R * shared_data.recipT);
     }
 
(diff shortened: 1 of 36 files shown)


ischoegl:
@speth ... this should take care of a lot of accumulated lint, and clearly mark legacy objects wherever used. Efficient handling of the `BlowersMasel` API will be important for #1181, where some preparatory work is done here. The new CI runner ensures that the `typedef Arrhenius2 Arrhenius` can be safely disabled.

There should be only minimal overlap with #1183 (if any). I added one instance of `setContext`, which I believe is a workable solution to the removal of remaining `Reaction` specializations that involve third bodies.

ischoegl:
@speth ... rebased now that #1183 is merged. Once this PR is resolved, I can get back to InterfaceKinetics (#1181).

include/cantera/kinetics/RxnRates.h
```diff
@@ -342,14 +220,17 @@ class PlogRate final : public ReactionRate
     PlogRate();
 
     //! Constructor from Arrhenius rate expressions at a set of pressures
-    explicit PlogRate(const std::multimap<double, Arrhenius>& rates);
+    explicit PlogRate(const std::multimap<double, ArrheniusBase>& rates);
+
+    //! Constructor using legacy Arrhenius2 framework
+    PlogRate(const std::multimap<double, Arrhenius2>& rates);
```
speth: Single-argument constructor should usually be marked `explicit`.

include/cantera/kinetics/RxnRates.h
```diff
@@ -469,18 +357,18 @@ class PlogRate final : public ReactionRate
      * @deprecated  Behavior to change after Cantera 2.6.
      *              @see getRates for new behavior.
      */
-    std::vector<std::pair<double, Arrhenius> > rates() const;
+    std::vector<std::pair<double, Arrhenius2> > rates() const;
```
speth: One minor formatting thing that can be cleaned up here would be to drop the space between the two closing `>` (a wart that was fixed by C++11).

include/cantera/kinetics/Falloff.h
```diff
@@ -104,8 +104,8 @@ class FalloffRate : public ReactionRate
         updateTemp(T, m_work.data());
         FalloffData data;
         data.update(T);
-        m_rc_low = m_lowRate.evalFromStruct(data);
-        m_rc_high = m_highRate.evalFromStruct(data);
+        m_rc_low = m_lowRate.evalRate(data.logT, data.recipT);
+        m_rc_high = m_highRate.evalRate(data.logT, data.recipT);
```
speth: At this point, we might as well just call `evalRate(std::log(T), 1/T)` and not bother with the `FalloffData` object for this function.

include/cantera/kinetics/FalloffMgr.h
```diff
@@ -1,5 +1,8 @@
 /**
  *  @file FalloffMgr.h
+ *
+ *  @deprecated  Deprecated in Cantera 2.6 and removed thereafter. Replaced by
+ *      FalloffRate objects managed by MultiRate evaluators.
```
speth, line 5: In addition to marking the header file as deprecated, I think it would be good to mark the `FalloffMgr` class deprecated as well.

include/cantera/kinetics/Arrhenius.h
```diff
@@ -54,17 +64,25 @@ class ArrheniusBase : public ReactionRate
...
+    //! Check rate expression
+    void checkRate(const std::string& equation, const AnyMap& node);
 
-    //! Return parameters - required for legacy framework
-    //! @todo: merge with single-parameter version after removal of old framework
-    virtual void getParameters(AnyMap& node, const Units& rate_units) const;
+    //! Evaluate reaction rate
+    double evalRate(double logT, double recipT) const {
+        return m_A * std::exp(m_b * logT - m_Ea_R * recipT);
+    }
```
speth, line 83: Having this as a non-virtual function in the base class used by `BlowersMaselRate` and `TwoTempPlasmaRate` seems like it will lead to some unfortunate surprising behavior. I'm starting to wonder if we're trying to hard to make those two classes fit into a heirarchy that doesn't quite fit.

include/cantera/kinetics/Arrhenius.h
```diff
@@ -400,32 +439,31 @@ class BlowersMaselRate final : public ArrheniusBase
...
     //! Return the bond dissociation energy *w* [J/kmol]
     double bondEnergy() const {
-        return m_w_R * GasConstant;
+        return m_E4_R * GasConstant;
     }
 
 protected:
-    double m_w_R; //!< Bond dissociation energy (in temperature units)
+    //! Pairs of species indices and multiplers to calculate enthalpy change
+    std::vector<std::pair<size_t, double>> m_multipliers;
```
speth: `m_stoich_coeffs` would be a more descriptive name of what these are rather than just how they're used.

include/cantera/kinetics/Arrhenius.h
```diff
@@ -170,8 +189,14 @@ class ArrheniusRate final : public ArrheniusBase
 
     virtual void getParameters(AnyMap& node) const override;
 
+    void check(const std::string& equation, const AnyMap& node) {
+        checkRate(equation, node);
+    }
```
speth, line 200: This needs to be marked `override`, and I suppose, `virtual` (likewise for the other classes derived from `ArrheniusBase`)

include/cantera/kinetics/RateCoeffMgr.h
```diff
@@ -1,5 +1,8 @@
 /**
  *  @file RateCoeffMgr.h
+ *
+ *  @deprecated  Deprecated in Cantera 2.6 and removed thereafter. Replaced by
+ *      ReactionRate objects managed by MultiRate evaluators.
```
speth, line 5: Likewise here, I think the `Rate1` class itself should be marked as deprecated.

include/cantera/kinetics/ReactionData.h
```diff
@@ -45,13 +45,15 @@ struct ReactionData
...
     virtual void update(double T, double extra);
 
-    //! Update data container based on *bulk* phase state
+    //! Update data container based on thermodynamic phase state
     /**
      * This update mechanism is used by Kinetics reaction rate evaluators.
      * @returns  A boolean element indicating whether the `evalFromStruct` method
      *      needs to be called (assuming previously-calculated values were cached)
+     *
+     * @todo  Remove Kinetics argument
```
speth, line 54: How do you anticipate removing the `Kinetics` argument? I had been expecting that since the `Kinetics` object actually knows about all of the phases participating in a reaction, that was the only way to guarantee that you had everything you needed available.

src/kinetics/BulkKinetics.cpp
```diff
@@ -143,6 +143,7 @@ bool BulkKinetics::addReaction(shared_ptr<Reaction> r, bool resize)
 
         // Set index of rate to number of reaction within kinetics
         rate->setRateIndex(nReactions() - 1);
+        rate->setContext(*r.get(), *this);
```
speth: You can write `*r.get()` as just `*r`. Likewise below in `modifyReaction`.

src/kinetics/FalloffFactory.cpp
```diff
@@ -1,5 +1,8 @@
 /**
  *  @file FalloffFactory.cpp
+ *
+ *  @deprecated  Deprecated in Cantera 2.6 and removed thereafter. Replaced by
+ *      FalloffRate objects managed by MultiRate evaluators.
```
speth, line 5: No point removing this, but I would say that for future reference there's never really a need to mark a `.cpp` file as deprecated, since there's way to reference it externally anyway.

src/kinetics/RxnRates.cpp
```diff
@@ -32,56 +30,37 @@ Arrhenius2::Arrhenius2(const AnyValue& rate,
...
-
-void BlowersMasel2::getParameters(AnyMap& rateNode, const Units& rate_units) const
+void Arrhenius2::getParameters(AnyMap& node, const Units& rate_units) const
 {
     if (rate_units.factor() != 0.0) {
-        rateNode["A"].setQuantity(preExponentialFactor(), rate_units);
+        node["A"].setQuantity(m_A, rate_units);
     } else {
-        rateNode["A"] = preExponentialFactor();
+        node["A"] = m_A;
```
speth: I know this isn't followed universally, but one of the ways to make sure that objects have a complete interface without having to interact with a specific input format is by using public methods in the `getParameters`/`setParameters`-like methods instead of just accessing all the member variables directly. I don't think we necessarily need to enforce this rigidly, but I think it's worth at least leaving cases that use that style as-is.

src/kinetics/RxnRates.cpp
```diff
@@ -240,17 +228,19 @@ void PlogRate::validate(const std::string& equation)
     }
 }
 
-std::vector<std::pair<double, Arrhenius> > PlogRate::rates() const
+std::vector<std::pair<double, Arrhenius2> > PlogRate::rates() const
 {
-    warn_deprecated("PlogRate::rates", "Behavior to change after Cantera 2.6; "
-        "see getRates() for new behavior.");
     auto rateMap = getRates();
```
speth, line 233: Why did the deprecation warning here go away?

interfaces/cython/cantera/reaction.pyx
```diff
@@ -120,72 +120,102 @@ cdef class ReactionRate:
...
-            self.set_cxx_object()
+    Base class collecting commonly used features of Arrhenius-type rate objects.
+    """
+    _reaction_rate_type = None
 
-    cdef CxxArrheniusRate* cxx_object(self):
-        return <CxxArrheniusRate*>self.rate
+    def _cinit(self, input_data, **kwargs):
+        """Helper function called by __cinit__"""
+        if self._reaction_rate_type is None:
```
speth, line 137: I think it would help to explain why this can't just be the implementation of `__cinit__` directly. If I'm understanding it correctly, this comes down to the arguments for the derived classes being slightly different, but not wanting to make these all be keyword-only?

interfaces/cython/cantera/reaction.pyx
```diff
@@ -120,72 +120,102 @@ cdef class ReactionRate:
...
+            self._from_dict(input_data)
+        elif all([kwargs[k] is not None for k in kwargs]):
+            self._from_parameters(*[kwargs[k] for k in kwargs])
+        elif all([kwargs[k] is None for k in kwargs]) and input_data is None:
+            self._from_dict({})
+        elif input_data:
+            raise TypeError("Invalid parameter 'input_data'")
+        else:
+            par_list = [f"'{k}'" for k in kwargs]
+            par_string = ", ".join([par_list[:-1]])
```
speth: I think this needs to be:
```suggestion
            par_string = ", ".join(par_list[:-1])
```

interfaces/cython/cantera/reaction.pyx
```diff
@@ -120,72 +120,102 @@ cdef class ReactionRate:
...
-    A reaction rate coefficient which depends on temperature only and follows
-    the modified Arrhenius form:
-
-    .. math::
-
-        k_f = A T^b \exp{-\tfrac{E}{RT}}
-
-    where ``A`` is the `pre_exponential_factor`, ``b`` is the `temperature_exponent`,
-    and ``Ea`` is the `activation_energy`.
+cdef class _ArrheniusTypeRate(ReactionRate):
```
speth: This class needs to be added to `kinetics.rst`

test/kinetics/kineticsFromScratch3.cpp
```diff
@@ -136,11 +136,11 @@ TEST_F(KineticsFromScratch3, add_plog_reaction)
     //                [(100.0, 'atm'), 5.963200e+56, -11.529, 52599.6])
     Composition reac = parseCompString("H2:1, O2:1");
     Composition prod = parseCompString("OH:2");
-    std::multimap<double, Arrhenius> rates {
-        { 0.01*101325, Arrhenius(1.212400e+16, -0.5779, 10872.7 / GasConst_cal_mol_K) },
-        { 1.0*101325, Arrhenius(4.910800e+31, -4.8507, 24772.8 / GasConst_cal_mol_K) },
-        { 10.0*101325, Arrhenius(1.286600e+47, -9.0246, 39796.5 / GasConst_cal_mol_K) },
-        { 100.0*101325, Arrhenius(5.963200e+56, -11.529, 52599.6 / GasConst_cal_mol_K) }
+    std::multimap<double, ArrheniusBase> rates {
+        { 0.01*101325, ArrheniusBase(1.212400e+16, -0.5779, 10872.7 * 4184.0) },
```
speth, line 140: You're anticipating that `ArrheniusBase` will become just `Arrhenius` eventually, right? I really hope this change from temperature units to energy units won't catch too many people by surprise.

include/cantera/kinetics/Arrhenius.h
```diff
@@ -77,10 +95,11 @@ class ArrheniusBase : public ReactionRate
...
     }
 
-    //! Return the intrinsic activation energy *Ea* [J/kmol]
+    //! Return the activation energy *Ea* [J/kmol]
     //! The value corresponds to the constant specified by input parameters;
-    //! in the simplest case, this is equal to the activation energy itself
-    double intrinsicActivationEnergy() const {
+    //! class specializations may provide alternate definitions that describe
+    //! an effective activation energy that depends on the thermodynamic state.
+    double activationEnergy() const {
```
speth, line 108: Big :+1: on this change in terminology -- I think it's much better.

speth (changes requested):
Thanks for splitting this out, since a lot of this is just reorganization and name changes. I think it will make reviewing #1181 a fair bit easier.

I like the addition of `setContext` -- it makes a lot of sense to have a function that can be called to set things like species indices that can be resolved at the point when a reaction is added to a Kinetics object, but not before.

include/cantera/kinetics/ReactionData.h
ischoegl: Whatever is needed can already be set/buffered by `setContext`. Once `InterfaceKinetics` is complete, I believe that the `Kinetics` argument can be safely removed.

src/kinetics/RxnRates.cpp
ischoegl: In order to properly split `Arrhenius` into legacy and new behavior, this route is used again whenever you instantiate from XML and/or access the legacy objects. I'm not sure that this is the best place to make noisy warnings.

In a nutshell, this object needs to be instantiated both from `PlogRate` as well as `PlogReaction2`, which use `ArrheniusBase/Arrhenius3` and `Arrhenius2`, respectively.

interfaces/cython/cantera/reaction.pyx
ischoegl: Correct. As far as I'm aware, Cython doesn't allow for `__cinit__` overloads.

interfaces/cython/cantera/reaction.pyx
ischoegl: Ok. While this is a hidden class, the documentation needs to be added, correct? Is there a way to specify inherited members (by level) in Sphinx?

test/kinetics/kineticsFromScratch3.cpp
ischoegl: ~This is correct (to clarify, I just renamed `ArrheniusBase` to `Arrhenius3`).~ *Edit: there's one option I want to test next where wouldn't be the case* ... I'm not sure that many people instantiate `Arrhenius` directly, and it would be odd to have a default unit that is different from the YAML specification.

One way would be to add an optional boolean. At the same time, it is really `ArrheniusRate` that is a successor to `Arrhenius2`; as this goes hand-in-hand with the introduction of YAML, I believe it is the right time to switch.

As an aside, the Python API already used energy units by default; as this is (presumably) the most commonly used API, I do not anticipate many problems.

include/cantera/kinetics/Arrhenius.h
ischoegl: I added an `@internal` and augmented the docstring mentioning that this method is not meant for overloading.

Overall, the `Arrhenius` class is purely non-virtual, as none of the methods are meant to be overloaded.

interfaces/cython/cantera/reaction.pyx
ischoegl: I ended up un-hiding the class in Python, and added it to `kinetics.rst` as requested.

ischoegl:
@speth ... thank you for your review! Beyond addressing review comments, I also renamed `ArrheniusBase` to `Arrhenius3` to clarify the anticipated renaming. I believe all concerns are addressed ...

*PS: removed a superfluous `ReactionRate::activationEnergy()` after requesting the review - had overlooked that.*

ischoegl:
@speth ... I addressed all comments. GH actions appears down for the time being; I'll try to re-trigger later on. Regarding the renaming of `ArrheniusBase`, I am tabling this as there is another tweak that I'd like to test first.

ischoegl:
GH actions are back up, and everything ran through …

include/cantera/kinetics/ReactionData.h
speth: You want to make the `Reaction` hold a pointer back to the `Kinetics` object if it needs that data, then?

src/kinetics/RxnRates.cpp
speth: I wonder if we need to have a way to do noisy warnings that are silenced if you're using CTI/XML input. Right now, we have a lot of functions and classes that are marked as deprecated but don't emit warnings for this reason.

test/kinetics/kineticsFromScratch3.cpp
speth: You're right that this is probably not the most common means of instantiating `Arrhenius` objects for end-users. One thing that would be simple to do, though not foolproof, would be to call out this difference in the Doxygen deprecation note that describes `ArrheniusBase` as the replacement. Though right now `Arrhenius2` is still missing such a marking.

ischoegl: I think that is appropriate. Once `InterfaceKinetics` is done, there won't be a problem raising noisy warnings, etc..

ischoegl: I believe that adding a `todo` in the docstring of the constructor is likely the best option for the moment (which I just added).

interfaces/cython/cantera/reaction.pyx
speth: `__cinit__` can be defined in both the base and derived classes, and functions more like a C++ constructor than the `__init__` method, in that it is always called for all classes in the hierarchy, starting with the base class. The one thing which I guess is relevant here is that default arguments defined in the child class's `__cinit__` won't be seen by the parent class's `__cinit__`.

In any case, the behavior in question is "overriding", not "overloading" which Python doesn't do at all.

include/cantera/kinetics/Arrhenius.h
```diff
@@ -370,14 +390,34 @@ class BlowersMaselRate final : public ArrheniusBase
 
     virtual void getParameters(AnyMap& node) const;
 
-    double evalFromStruct(const BlowersMaselData& shared_data) {
-        double deltaH_R;
+    void check(const std::string& equation, const AnyMap& node) override {
+        checkRate(equation, node);
+    }
```
speth, line 396: Sorry I missed this before, but there are several other functions in `BlowersMaselRate` that now need to be marked `override` (at least, to avoid getting warnings from Clang). Namely, `newMultiRate`, `type`, `setParameters`, and `getParameters`.

ischoegl: No problem - fixed. Should probably switch to clang!

src/kinetics/RxnRates.cpp
ischoegl: Maybe ... at the same time, it may be a good incentive to switch to YAML. I'd advocate for issuing warnings once `InterfaceKinetics` is done, and `Arrhenius2` is only used for XML/CTI.

include/cantera/kinetics/ReactionData.h
ischoegl: I think it's not strictly necessary, but we can discuss when the time comes.

interfaces/cython/cantera/reaction.pyx
ischoegl: > [...] functions more like a C++ constructor than the `__init__` method, in that it is always called for all classes in the hierarchy, starting with the base class The one thing which I guess is relevant here is that default arguments defined in the child class's `__cinit__` won't be seen by the parent class's `__cinit__`.

Thanks - yes, the call-order is relevant here, which is why the regular `__cinit__` route doesn't work.

speth (approved):
Thanks for the updates, @ischoegl. This looks good to me.

PR #1187, opened 2022-02-07
Introduce BulkRate class template

ischoegl:

The clear separation of `ArrheniusBase` and `ArrheniusRate` introduced in #1184 created an opening for the creation of class templates that can be further developed in #1181. Specifically, this PR introduces `BulkRate<RateType,DataType>`, along with stripped-down classes `Arrhenius3`,  `TwoTempPlasma` and `BlowersMasel`.

The pre-existing classes are now defined as `typedef`'s as:
```C++
typedef BulkRate<Arrhenius3, ArrheniusData> ArrheniusRate;
typedef BulkRate<BlowersMasel, BlowersMaselData> BlowersMaselRate;
typedef BulkRate<TwoTempPlasma, TwoTempPlasmaData> TwoTempPlasmaRate;
```

Beyond, additional streamlining targets the elimination of `TwoTempPlasmaReaction`, as well as some updates in `test_reaction.py`.

It is anticipated that the definition of  `Arrhenius3` and `BlowersMasel` will facilitate the introduction of class templates `InterfaceRate<>` and `StickingRate<>` in #1181. In addition, there is a potential use of `Arrhenius3` for a `ThreeBodyRate<>` template in a future resolution of Cantera/enhancements#133.

Also: eliminate multiple inheritance; while this allowed avoiding several `virtual` methods, multiple inheritance in combination with templates resulted in a small speed penalty, although the origin is inconclusive. While elimination of multiple inheritance shows a small improvement, tests on windows/MSVC indicate that #1184 did not introduce any noticeable regressions.

**Checklist**

### DIFF
--- a/include/cantera/kinetics/Arrhenius.h
+++ b/include/cantera/kinetics/Arrhenius.h
@@ -166,25 +166,29 @@ public:
     }
 
     //! Evaluate reaction rate
-    //! @internal  Non-virtual method that should not be overloaded
     double evalRate(double logT, double recipT) const {
         return m_A * std::exp(m_b * logT - m_Ea_R * recipT);
     }
 
     //! Evaluate natural logarithm of the rate constant.
-    //! @internal  Non-virtual method that should not be overloaded
     double evalLog(double logT, double recipT) const {
         return m_logA + m_b * logT - m_Ea_R * recipT;
     }
 
     //! Evaluate reaction rate
-    double evalRate(const ArrheniusData& shared_data) const {
+    /*!
+     *  @param shared_data  data shared by all reactions of a given type
+     */
+    double evalFromStruct(const ArrheniusData& shared_data) const {
         return m_A * std::exp(m_b * shared_data.logT - m_Ea_R * shared_data.recipT);
     }
 
     //! Evaluate derivative of reaction rate with respect to temperature
     //! divided by reaction rate
-    double ddTScaled(const ArrheniusData& shared_data) const {
+    /*!
+     *  @param shared_data  data shared by all reactions of a given type
+     */
+    double ddTScaledFromStruct(const ArrheniusData& shared_data) const {
         return (m_Ea_R * shared_data.recipT + m_b) * shared_data.recipT;
     }
 };
@@ -228,12 +232,13 @@ public:
         return "two-temperature-plasma";
     }
 
-    //! Context
     virtual void setContext(const Reaction& rxn, const Kinetics& kin) override;
 
     //! Evaluate reaction rate
-    //! @internal  Non-virtual method that should not be overloaded
-    double evalRate(const TwoTempPlasmaData& shared_data) const {
+    /*!
+     *  @param shared_data  data shared by all reactions of a given type
+     */
+    double evalFromStruct(const TwoTempPlasmaData& shared_data) const {
         // m_E4_R is the electron activation (in temperature units)
         return m_A * std::exp(m_b * shared_data.logTe -
                               m_Ea_R * shared_data.recipT +
@@ -246,9 +251,9 @@ public:
     /*!
      *  This method does not consider changes of electron temperature.
      *  A corresponding warning is raised.
-     *  @internal  Non-virtual method that should not be overloaded
+     *  @param shared_data  data shared by all reactions of a given type
      */
-    double ddTScaled(const TwoTempPlasmaData& shared_data) const;
+    double ddTScaledFromStruct(const TwoTempPlasmaData& shared_data) const;
 
     //! Return the electron activation energy *Ea* [J/kmol]
     double activationElectronEnergy() const {
@@ -313,7 +318,6 @@ public:
         return "Blowers-Masel";
     }
 
-    //! Set context
     virtual void setContext(const Reaction& rxn, const Kinetics& kin) override;
 
     //! Update information specific to reaction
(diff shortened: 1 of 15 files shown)


ischoegl:
This is ready for review. The four cancelled CI runs are due to a Windows-2016 brown-out. *Edit:* retriggered ...

include/cantera/kinetics/Arrhenius.h
```diff
@@ -468,7 +381,73 @@ class BlowersMaselRate final : public ArrheniusBase, public ReactionRate
...
+        }
+    }
+
+    //! Evaluate reaction rate
+    /*!
+     *  @param shared_data  data shared by all reactions of a given type
+     */
+    double evalFromStruct(const DataType& shared_data) const {
+        return RateType::evalRate(shared_data);
+    }
```
speth: If you rename `evalRate(DataType&)` to `evalFromStruct(DataType&)` in `ArrheniusBase` and friends, I think this method can just be deleted. Likewise for `ddTScaledFromStruct` below.

ischoegl: Correct, although they would have to be declared virtual, as there will be different implementations in other anticipated class templates. Due to the `final` declaration it should not make any difference. At the same time, the compiler inlines this and it does not appear to make any difference either way (I already tested this).

speth: *Sorry, I had clicked the wrong button -- meant to add this as part of a review.*

It's perfectly acceptable to override a non-virtual method in a derived class. In such a case, calls to this function will always be based on the visible type of the object, not a more derived type. This can be (and is) resolved at compile time, which is exactly what we want in this case, since the `MultiRate` class always has a pointer to the most-derived type, and no virtual function calls are needed (which furthermore means there's no need to mark the class as `final`).

speth (commented):
I like the simplification that this achieves, both in complexity and lines of code, by eliminating the use of multiple inheritance with the `ReactionRate` class. But I think we might be able to actually go a step further and skip the `BulkRate` template entirely with no real loss, since it's just delegating almost all of its behavior to `ArrheniusBase`.

I can still see this template structure potentially being useful in the case you mentioned of interface reactions, where we need to handle some common operations on top of either the base Arrhenius or Blowers-Masel rate types. And I can see retaining `BulkRate` if there's some set of behaviors that we need for bulk rates that shouldn't be part of the interface rates, but I don't see anything like that in the current implementation.

ischoegl:
@speth ... thanks for the comment! I removed the chained function calls, although I don't think there is any performance penalty. Based on your explanation, I don't see any benefit in keeping this around.

>  But I think we might be able to actually go a step further and skip the `BulkRate` template entirely with no real loss, since it's just delegating almost all of its behavior to `ArrheniusBase`.

There is the exception of `unique_ptr<MultiRateBase> newMultiRate()`, which cannot be pushed back to a lower level - we need to have this attached to a class definition that is `final`. So I believe `BulkRate<>` makes sense from this perspective. I am btw wondering whether we can construct a `BulkRate<Custom, CustomData>` ... .

> And I can see retaining `BulkRate` if there's some set of behaviors that we need for bulk rates that shouldn't be part of the interface rates, but I don't see anything like that in the current implementation.

Fwiw, `StickingRate<>` will need a different serialization, although this is minor.

So I *think* it may make sense to keep `BulkRate<>`, at least for the time being? As long as there's no performance penalty, it would imho be a self-documenting feature as well.

speth:
> There is the exception of `unique_ptr<MultiRateBase> newMultiRate()`, which cannot be pushed back to a lower level - we need to have this attached to a class definition that is `final`. So I believe `BulkRate<>` makes sense from this perspective.

You're right that `newMultiRate()` needs to be overridden in each class, but I don't think any of the classes actually need to be declared `final`, since the key performance sensitive methods (`xyzFromStruct`) are all non-virtual in the first place.

I guess eliminating `BulkRate` later is an option, if it turns out not to be useful in the `InterfaceKinetics` implementation. Although that would I think introduce more churn of renaming what is currently the `ArrheniusBase` class yet another time.

ischoegl:
> I guess eliminating `BulkRate` later is an option, if it turns out not to be useful in the `InterfaceKinetics` implementation. Although that would I think introduce more churn of renaming what is currently the `ArrheniusBase` class yet another time.

That would be my preference. I don’t see `ArrheniusBase` going away either way, and the `typedef`’s hide `BulkRate` from the API already.

ischoegl:
@speth ... fwiw, I believe that `BulkRate<>` can be easily leveraged to create
```
typedef BulkRate<CustomFunc1, ArrheniusData> CustomFunc1Rate;
```
by renaming `getParameters` to `getRateParameters` (mainly to raise an exception), and adding a corresponding `setRateParameters` (again to handle an exception). With that, I believe it would be simple to remove `CustomFunc1Reaction`. I'm not tackling this at the moment though as it is a tangent that I'd like to keep for later (mainly as a nomenclature has to be developed - `CustomFunc1` is clearly not ideal, - and there may be ways to introduce `Delegator`s). But again, the idea would be to have a `CustomFunc1` as a simple reusable unit for class templates `BulkRate<>`, `InterfaceRate<>`, `StickingRate<>`, etc..

At the moment, I'll wait for this PR to be merged before getting back to #1181.

include/cantera/kinetics/Arrhenius.h
```diff
@@ -255,55 +226,35 @@ class TwoTempPlasmaRate final : public ArrheniusBase, public ReactionRate
...
-     */
-    virtual void setParameters(const AnyMap& node, const UnitStack& rate_units) override;
-
-    virtual void getParameters(AnyMap& node) const override;
-
-    void check(const std::string& equation, const AnyMap& node) override {
-        checkRate(equation, node);
-    }
+    //! Context
+    virtual void setContext(const Reaction& rxn, const Kinetics& kin) override;
```
speth, line 235: I think it's better to just have no docstring for this method. That way, Doxygen will at least inherit the docstring from the parent class. Likewise for Blowers-Masel.

include/cantera/kinetics/Arrhenius.h
```diff
@@ -468,7 +388,56 @@ class BlowersMaselRate final : public ArrheniusBase, public ReactionRate
     double m_deltaH_R; //!< enthalpy change of reaction (in temperature units)
 };
 
-}
 
+//! A class template for bulk phase reaction rate specifications
+template <class RateType, class DataType>
+class BulkRate final : public RateType
```
speth: I don't think `final` is really necessary.

src/kinetics/Arrhenius.cpp
```diff
@@ -111,92 +111,62 @@ void ArrheniusBase::getRateParameters(AnyMap& node) const
...
 {
-    if (m_negativeA_ok) {
-        rateNode["negative-A"] = true;
-    }
-    AnyMap node;
-    ArrheniusBase::getRateParameters(node);
-    if (!node.empty()) {
-        // object is configured
-        rateNode["rate-constant"] = std::move(node);
+    // TwoTempPlasmaReaction is for a non-equilirium plasma, and the reverse rate
```
speth: "equilibrium" is misspelled.

speth (changes requested):
I had just a couple of very minor comments. Otherwise this looks good to me.

include/cantera/kinetics/Arrhenius.h
ischoegl: I believe it signals that nothing can inherit from this class, which I believe is ok. Still removing it ...

ischoegl:
@speth ... thank you for the review! I updated as requested.

speth (approved):

PR #1191, opened 2022-02-08
[SCons] Change prefix default if conda is detected

### LINKED ISSUE #76: Change the default prefix to $CONDA_PREFIX if it is set
**Abstract**

Conda environments are self-contained package installation locations, including C/C++, Python, and other languages. They support all the usual *nix directories (`share`, `lib`, `include`) etc., and have a similarly appropriate structure for Windows. Under certain conditions, it may make sense to set the `prefix` configuration variable to the `$CONDA_PREFIX` environment variable, which holds the path to the currently activated environment.

**Motivation**

In Cantera/cantera#963, the user is using Anaconda Python to build and install Cantera from source. They attempted to use `sudo scons install` to be able to install, but the `sudo` shell did not inherit the shell environment that Conda uses to activate itself, and the installation failed.

One proposed remedy for this case is to assume that the user would like to install all of the files into their Conda environment. Changing this as a default shouldn't cause any problems, since the Conda environment should already have all of the appropriate folders.

**Possible Solutions**

If this enhancement is approved, the default should be changed only when:

1) `prefix` is not set AND
2) `python_prefix` is not set AND
3) `python_cmd` is not set AND
4) `$CONDA_PREFIX` is set AND
5) the Python running SCons is located in `$CONDA_PREFIX`

**References**

https://github.com/Cantera/cantera/issues/963#issuecomment-774597675

ischoegl:

The default installation `prefix` is changed only when:

1. `prefix` is not set AND
2. `python_prefix` is not set AND
3. `python_cmd` is not set AND
4. `$CONDA_PREFIX` is set AND
5. the Python running SCons is located in `$CONDA_PREFIX`

**Also:** add `conda` include/library folders if conda is used (as long as `extra_inc_dirs` and `extra_lib_dirs` are not set.

Closes Cantera/enhancements#76

```
[...]
postInstallMessage(["finish_install"], [])

File locations:

  applications                C:\Users\<user>\miniconda3\envs\cantera-dev\Scripts
  library files               C:\Users\<user>\miniconda3\envs\cantera-dev\Library\lib
  C++ headers                 C:\Users\<user>\miniconda3\envs\cantera-dev\Library\include
  samples                     C:\Users\<user>\miniconda3\envs\cantera-dev\share\cantera\samples
  data files                  C:\Users\<user>\miniconda3\envs\cantera-dev\share\cantera\data
  Python package (cantera)    C:\Users\<user>\miniconda3\envs\cantera-dev\Lib\site-packages
  Python samples              C:\Users\<user>\miniconda3\envs\cantera-dev\Lib\site-packages\cantera\examples

scons: done building targets.
```

**Checklist**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -313,9 +313,8 @@ jobs:
           cython boost-cpp fmt eigen yaml-cpp h5py pandas libgomp openblas pytest
       - name: Build Cantera
         run: |
-          scons build extra_inc_dirs=$CONDA_PREFIX/include:$CONDA_PREFIX/include/eigen3 \
-          extra_lib_dirs=$CONDA_PREFIX/lib system_fmt=y system_eigen=y system_yamlcpp=y \
-          system_sundials=y blas_lapack_libs='lapack,blas' -j2 VERBOSE=True debug=n \
+          scons build system_fmt=y system_eigen=y system_yamlcpp=y system_sundials=y \
+          blas_lapack_libs='lapack,blas' -j2 VERBOSE=True debug=n \
           optimize_flags='-O3 -ffast-math -fno-finite-math-only'
       - name: Test Cantera
         run: scons test show_long_tests=yes verbose_tests=yes
--- a/SConstruct
+++ b/SConstruct
@@ -870,6 +870,7 @@ except subprocess.CalledProcessError:
     env["git_commit"] = "unknown"
 
 # Print values of all build options:
+# the (updated) "cantera.conf" combines all options that were specified by the user
 cantera_conf = Path("cantera.conf").read_text()
 logger.info("Configuration variables read from 'cantera.conf' and command line:")
 logger.info(textwrap.indent(cantera_conf, "    "), print_level=False)
@@ -1642,17 +1643,16 @@ if env['matlab_toolbox'] == 'y':
 # /usr/local because of dist-packages vs site-packages
 env['debian'] = any(name.endswith('dist-packages') for name in sys.path)
 
+# Identify options selected either on command line or in cantera.conf
+selected_options = set(line.split("=")[0].strip()
+    for line in cantera_conf.splitlines())
+
 # Check whether Cantera should be installed into a conda environment
 if conda_prefix is not None:
-    if env["layout"] != "conda":
-        # identify options selected either on command line or in cantera.conf
-        selected_options = set(line.split("=")[0].strip()
-            for line in cantera_conf.splitlines())
-        use_conda = (
-            not selected_options &
-            {"layout", "prefix", "python_prefix", "python_cmd"})
-        use_conda &= sys.executable.startswith(conda_prefix)
-        if use_conda:
+    if env["layout"] != "conda" and sys.executable.startswith(conda_prefix):
+        # use conda layout unless any 'blocking' options were specified
+        blocking_options = {"layout", "prefix", "python_prefix", "python_cmd"}
+        if not selected_options & blocking_options:
             env["layout"] = "conda"
             logger.info(
                 f"Using conda environment as default 'prefix': {conda_prefix}")
@@ -1663,7 +1663,11 @@ elif env["layout"] == "conda":
 # Directories where things will be after actually being installed. These
 # variables are the ones that are used to populate header files, scripts, etc.
 if env["layout"] == "conda":
-    env["prefix"] = os.path.normpath(conda_prefix)
+    if "stage_dir" in selected_options:
+        conda_prefix = Path(conda_prefix)
+        env["prefix"] = str(conda_prefix.relative_to(conda_prefix.parents[2]))
+    else:
+        env["prefix"] = os.path.normpath(conda_prefix)
 
 if env["layout"] == "conda" and os.name == "nt":
     env["ct_libdir"] = pjoin(env["prefix"], "Library", env["libdirname"])
@@ -1725,7 +1729,7 @@ if env["stage_dir"]:
     if stage_prefix.is_absolute():
         stage_prefix = Path(*stage_prefix.parts[1:])
 
-    instRoot = Path.cwd().joinpath(env["stage_dir"], stage_prefix)
+    instRoot = str(Path.cwd().joinpath(env["stage_dir"], stage_prefix))
 else:
     instRoot = env["prefix"]
 
@@ -1736,6 +1740,7 @@ if os.path.abspath(instRoot) == Dir('.').abspath:
 
 if env['layout'] == 'debian':
     base = pjoin(os.getcwd(), 'debian')
+    env["inst_root"] = base
 
     env['inst_libdir'] = pjoin(base, 'cantera-dev', 'usr', env['libdirname'])
     env['inst_incdir'] = pjoin(base, 'cantera-dev', 'usr', 'include', 'cantera')
@@ -1753,6 +1758,7 @@ if env['layout'] == 'debian':
     env['inst_python_bindir'] = pjoin(base, 'cantera-python', 'usr', 'bin')
     env['python_prefix'] = pjoin(base, 'cantera-python3')
 else:
+    env["inst_root"] = instRoot
     locations = ["libdir", "bindir", "python_bindir", "incdir", "incroot",
         "matlab_dir", "datadir", "sampledir", "docdir", "mandir"]
     for loc in locations:
@@ -2051,13 +2057,12 @@ def postInstallMessage(target, source, env):
 
               {matlab_ctpath_loc!s}
         """.format(**env_dict))
-    else:
-        install_message += "\n"
 
     if os.name != 'nt':
         env['setup_cantera'] = pjoin(env['ct_bindir'], 'setup_cantera')
         env['setup_cantera_csh'] = pjoin(env['ct_bindir'], 'setup_cantera.csh')
         install_message += textwrap.dedent("""
+
             Setup scripts to configure the environment for Cantera are at:
 
               setup script (bash)         {setup_cantera!s}
--- a/SConstruct
+++ b/SConstruct
@@ -217,8 +217,10 @@ config_options = [
         }),
     PathOption(
         "prefix",
-        """Set this to the directory where Cantera should be installed. On Windows
-           systems, '$ProgramFiles' typically refers to "C:\Program Files".""",
+        """Set this to the directory where Cantera should be installed. If the Python
+           executable found during compilation is managed by 'conda', the
+           installation 'prefix' will default to the corresponding environment. On
+           Windows systems, '$ProgramFiles' typically refers to "C:\Program Files".""",
         {"Windows": "$ProgramFiles\Cantera", "default": "/usr/local"},
         PathVariable.PathAccept),
     PathOption(
@@ -553,9 +555,11 @@ config_options = [
            conjunction with "prefix='/usr/local'". 'compact' puts all installed
            files in the subdirectory defined by 'prefix'. This layout is best
            with a prefix like '/opt/cantera'. 'debian' installs to the stage
-           directory in a layout used for generating Debian packages.""",
+           directory in a layout used for generating Debian packages. If the Python
+           executable found during compilation is managed by 'conda', the layout
+           will default to 'conda' irrespective of operating system.""",
         {"Windows": "compact", "default": "standard"},
-        ("standard", "compact", "debian")),
+        ("standard", "compact", "debian", "conda")),
     BoolOption(
         "fast_fail_tests",
         "If enabled, tests will exit at the first failure.",
@@ -819,6 +823,9 @@ config["python_cmd"].default = sys.executable
 opts.AddVariables(*config.to_scons())
 opts.Update(env)
 opts.Save('cantera.conf', env)
+cantera_conf = []
+for line in open('cantera.conf'):
+    cantera_conf.append(line.strip())
 
 # Expand ~/ and environment variables used in cantera.conf (variables used on
 # the command line will be expanded by the shell)
@@ -866,10 +873,10 @@ except subprocess.CalledProcessError:
     env["git_commit"] = "unknown"
 
 # Print values of all build options:
-print("Configuration variables read from 'cantera.conf' and command line:")
-for line in open('cantera.conf'):
-    print('   ', line.strip())
-print()
+msg = ["Configuration variables read from 'cantera.conf' and command line:"]
+for line in cantera_conf:
+    msg.append(f"    {line}")
+logger.info("\n".join(msg))
 
 # ********************************************
 # *** Configure system-specific properties ***
@@ -904,6 +911,19 @@ if os.pathsep == ';':
 
 env['extra_inc_dirs'] = [d for d in env['extra_inc_dirs'].split(os.pathsep) if d]
 env['extra_lib_dirs'] = [d for d in env['extra_lib_dirs'].split(os.pathsep) if d]
+
+# Add conda library/include paths (if applicable) to extra
+conda_prefix = os.environ.get("CONDA_PREFIX")
+if conda_prefix is not None:
+    if os.name == "nt":
+        conda_inc_dir = pjoin(conda_prefix, "Library", "include")
+        conda_lib_dir = pjoin(conda_prefix, "Library", env["libdirname"])
+    else:
+        conda_inc_dir = pjoin(conda_prefix, "include")
+        conda_lib_dir = pjoin(conda_prefix, env["libdirname"])
+    env["extra_inc_dirs"] = env["extra_inc_dirs"] + [conda_inc_dir]
+    env["extra_lib_dirs"] = env["extra_lib_dirs"] + [conda_lib_dir]
+    logger.info("Adding conda include and library paths.")
 
 env.Append(CPPPATH=env['extra_inc_dirs'],
            LIBPATH=env['extra_lib_dirs'])
@@ -1626,26 +1646,59 @@ if env['matlab_toolbox'] == 'y':
 # /usr/local because of dist-packages vs site-packages
 env['debian'] = any(name.endswith('dist-packages') for name in sys.path)
 
+# Check whether Cantera should be installed into a conda environment
+if conda_prefix is not None:
+    if env["layout"] != "conda":
+        # identify options selected either on command line or in cantera.conf
+        selected_options = set(line.split("=")[0].strip() for line in cantera_conf)
+        use_conda = not selected_options & \
+            {"layout", "prefix", "python_prefix", "python_cmd"}
+        use_conda &= sys.executable.startswith(conda_prefix)
+        if use_conda:
+            env["layout"] = "conda"
+            logger.info(
+                f"Using conda environment as default 'prefix': {conda_prefix}")
+elif env["layout"] == "conda":
+    logger.error("Layout option 'conda' requires a conda environment.")
+    sys.exit(1)
+
 # Directories where things will be after actually being installed. These
 # variables are the ones that are used to populate header files, scripts, etc.
-env['prefix'] = os.path.normpath(env['prefix'])
-env['ct_installroot'] = env['prefix']
-env['ct_libdir'] = pjoin(env['prefix'], env['libdirname'])
-env['ct_bindir'] = pjoin(env['prefix'], 'bin')
-env['ct_incdir'] = pjoin(env['prefix'], 'include', 'cantera')
-env['ct_incroot'] = pjoin(env['prefix'], 'include')
+if env["layout"] == "conda":
+    env["prefix"] = os.path.normpath(conda_prefix)
+
+if env["layout"] == "conda" and os.name == "nt":
+    env["ct_libdir"] = pjoin(env["prefix"], "Library", env["libdirname"])
+    env["ct_bindir"] = pjoin(env["prefix"], "Scripts")
+    env["ct_python_bindir"] = pjoin(env["prefix"], "Scripts")
+    env["ct_incdir"] = pjoin(env["prefix"], "Library", "include", "cantera")
+    env["ct_incroot"] = pjoin(env["prefix"], "Library", "include")
+else:
+    env["prefix"] = os.path.normpath(env["prefix"])
+    env["ct_libdir"] = pjoin(env["prefix"], env["libdirname"])
+    env["ct_bindir"] = pjoin(env["prefix"], "bin")
+    env["ct_python_bindir"] = pjoin(env["prefix"], "bin")
+    env["ct_incdir"] = pjoin(env["prefix"], "include", "cantera")
+    env["ct_incroot"] = pjoin(env["prefix"], "include")
+env["ct_installroot"] = env["prefix"]
 
 if env['layout'] == 'compact':
     env['ct_datadir'] = pjoin(env['prefix'], 'data')
     env['ct_sampledir'] = pjoin(env['prefix'], 'samples')
+    env["ct_docdir"] = pjoin(env["prefix"], "doc")
     env['ct_mandir'] = pjoin(env['prefix'], 'man1')
     env['ct_matlab_dir'] = pjoin(env['prefix'], 'matlab', 'toolbox')
 else:
     env['ct_datadir'] = pjoin(env['prefix'], 'share', 'cantera', 'data')
     env['ct_sampledir'] = pjoin(env['prefix'], 'share', 'cantera', 'samples')
+    env["ct_docdir"] = pjoin(env["prefix"], "share", "cantera", "doc")
     env['ct_mandir'] = pjoin(env['prefix'], 'share', 'man', 'man1')
-    env['ct_matlab_dir'] = pjoin(env['prefix'], env['libdirname'],
-                                 'cantera', 'matlab', 'toolbox')
+    if env["layout"] == "conda":
+        env["ct_matlab_dir"] = pjoin(
+            env["prefix"], "share", "cantera", "matlab", "toolbox")
+    else:
+        env["ct_matlab_dir"] = pjoin(
+            env["prefix"], env["libdirname"], "cantera", "matlab", "toolbox")
 
 # Always set the stage directory before building an MSI installer
 if 'msi' in COMMAND_LINE_TARGETS:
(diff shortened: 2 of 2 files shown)


SConstruct
```diff
@@ -1635,7 +1656,12 @@ env['ct_bindir'] = pjoin(env['prefix'], 'bin')
 env['ct_incdir'] = pjoin(env['prefix'], 'include', 'cantera')
 env['ct_incroot'] = pjoin(env['prefix'], 'include')
 
-if env['layout'] == 'compact':
+if use_conda:
+    env["ct_datadir"] = pjoin(env["prefix"], "cantera", "data")
+    env["ct_sampledir"] = pjoin(env["prefix"], "cantera", "samples")
+    env["ct_mandir"] = pjoin(env["prefix"], "cantera", "man1")
+    env["ct_matlab_dir"] = pjoin(env["prefix"], "cantera", "matlab", "toolbox")
+elif env["layout"] == "compact":
```
bryanwweber: At least on *nix, the folders in Conda environments follow the standard *nix convention in the `else` clause here. So this change may need to be scoped to Windows?

bryanwweber (commented):
Thanks for proposing something here @ischoegl! I have a high level question before reviewing the code in more detail. I wonder if it would be clearer to a user if we added `conda` as an allowed default value for some of the options? Or, if that's too intrusive, appended/overwrote some of the default options before the options are all saved to `cantera.conf.` This could be used to write things out into the `cantera.conf` file, which may prompt people to edit that file to restore old behavior if they want. I _think_ that could also simplify some of the `use_conda` logic and checking lines from `cantera.conf`. Does that make sense? I'd be happy to throw together a small prototype to clarify.

SConstruct
ischoegl: I think conda should take precedence even on Linux. I can run a test later on

bryanwweber: That's what I'm saying. On Linux/macOS, the folder structure is something like:

```
envs
|--- env_name/
|---|--- etc/
|---|--- share/
|---|--- lib/
...
```
So the default behavior here will already fit into a conda environment.

ischoegl:
@bryanwweber ... thank you for the feedback! I limited the conda-specific installation folders to Windows (`share` btw also exists there, which I am now using). I still need to test on Linux ...

SConstruct
```diff
@@ -553,9 +556,11 @@ config_options = [
            conjunction with "prefix='/usr/local'". 'compact' puts all installed
            files in the subdirectory defined by 'prefix'. This layout is best
            with a prefix like '/opt/cantera'. 'debian' installs to the stage
-           directory in a layout used for generating Debian packages.""",
+           directory in a layout used for generating Debian packages. If the Python
+           executable found during compilation is managed by 'conda', the layout
+           will default to 'conda' irrespective of operating system.""",
         {"Windows": "compact", "default": "standard"},
-        ("standard", "compact", "debian")),
+        ("standard", "compact", "debian", "conda")),
```
bryanwweber, line 571: This is a great use of the layout variable, nice!

SConstruct
```diff
@@ -819,6 +824,11 @@ config["python_cmd"].default = sys.executable
 opts.AddVariables(*config.to_scons())
 opts.Update(env)
 opts.Save('cantera.conf', env)
+cantera_conf = []
+for line in open('cantera.conf'):
+    cantera_conf.append(line.strip())
+selected_options = set(line.split("=")[0].strip() for line in cantera_conf)
+
```
bryanwweber: I don't think this set is needed, see below.

SConstruct
```diff
@@ -1046,6 +1056,17 @@ if nan_works.strip() != "1":
...
+conda_prefix = os.environ.get("CONDA_PREFIX")
+if conda_prefix is not None:
+    if not selected_options & {"extra_inc_dirs", "extra_lib_dirs"}:
+        if os.name == "nt":
+            env.Append(CPPPATH=pjoin(conda_prefix, "Library", "include"),
+                    LIBPATH=pjoin(conda_prefix, "Library", env["libdirname"]))
+        else:
+            env.Append(CPPPATH=pjoin(conda_prefix, "include"),
+                    LIBPATH=pjoin(conda_prefix, env["libdirname"]))
+
```
bryanwweber: Love this change, so that we don't have to put `boost_inc_dir` into the settings anymore 🎉

Since it's just appending to the paths, I don't think it needs to be conditional on not having user values for those two variables. Also, we may want to log that this is happening.

SConstruct
```diff
@@ -1626,26 +1647,58 @@ if env['matlab_toolbox'] == 'y':
 # /usr/local because of dist-packages vs site-packages
 env['debian'] = any(name.endswith('dist-packages') for name in sys.path)
 
+# Check whether Cantera should be installed into a conda environment
+if conda_prefix is not None:
+    if env["layout"] != "conda":
+        use_conda = not selected_options & \
+            {"layout", "prefix", "python_prefix", "python_cmd"}
```
bryanwweber: Here, rather than the set intersection, I think you can just compare these values to their defaults, right?

SConstruct
```diff
@@ -1626,26 +1647,58 @@ if env['matlab_toolbox'] == 'y':
...
+        use_conda &= sys.executable.startswith(conda_prefix)
+        if use_conda:
+            env["layout"] = "conda"
+            logger.info(
+                f"Using conda environment as default 'prefix': {env['prefix']}")
```
bryanwweber: I don't think the `prefix` will have been replaced at this point, right? That happens in the next block

SConstruct
```diff
@@ -1626,26 +1647,58 @@ if env['matlab_toolbox'] == 'y':
...
+else:
+    if env["layout"] == "conda":
```
bryanwweber: I think this can just be an `elif`

bryanwweber (changes requested):
This is really shaping up @ischoegl I'm 👍 on the direction in general here. A few comments/questions below

SConstruct
ischoegl: that's really tricky. I believe if a user specifies a default from the command line *that would be overwritten otherwise*, this would not take effect? I believe it's safer to just read from `cantera.conf`, which is already done anyways for some of the logging output.

SConstruct
ischoegl: Yup. It appears to work on the GH Sundials runner also ... ~hmm - almost.~ ... fixed!

SConstruct
bryanwweber: No, variables set on the command line are updated in `cantera.conf` on the `opts.Save` line, so they should be different from the default by this point. In terms of the logging, I think that can be simplified to one line if we don't need the data (something like `logger.info(indent(Path('cantera.conf').read_text()))`) plus it's one fewer data structure to mentally keep track of.

Not a big deal, it just feels weird to me to check the secondary output file rather than directly from the source of the configuration... in other words, `cantera.conf` is a representation of the current configuration, but it's not the actual values that are used in `SConstruct`.

ischoegl:
:tada: ... this is for Linux with the following `conda` environment:
```
$ conda create -n cantera-dev -c conda-forge sundials=6.0 scons numpy ruamel.yaml cython boost-cpp fmt eigen yaml-cpp libgomp openblas pytest ipython
```

This is the result of `scons build` & `scons install` ... which detects all 'system' dependencies (no additional options needed)

```
postInstallMessage(["finish_install"], [])

File locations:

  applications                /home/<user>/.conda/envs/cantera-dev/bin
  library files               /home/<user>/.conda/envs/cantera-dev/lib
  C++ headers                 /home/<user>/.conda/envs/cantera-dev/include
  samples                     /home/<user>/.conda/envs/cantera-dev/share/cantera/samples
  data files                  /home/<user>/.conda/envs/cantera-dev/share/cantera/data  
  Python package (cantera)    /home/<user>/.conda/envs/cantera-dev/lib/python3.10/site-packages
  Python samples              /home/<user>/.conda/envs/cantera-dev/lib/python3.10/site-packages/cantera/examples

Setup scripts to configure the environment for Cantera are at:

  setup script (bash)         /home/<user>/.conda/envs/cantera-dev/bin/setup_cantera
  setup script (csh/tcsh)     /home/<user>/.conda/envs/cantera-dev/bin/setup_cantera.csh

It is recommended that you run the script for your shell by typing:

  source /home/<user>/.conda/envs/cantera-dev/bin/setup_cantera

before using Cantera, or else include its contents in your shell login script.

scons: done building targets.
```

I am unfortunately not able to test on OSX, but as long as the file structure follows `*nix`, this should work.

SConstruct
ischoegl: > No, variables set on the command line are updated in `cantera.conf` on the `opts.Save` line, so they should be different from the default by this point. 

Correct. My implementation deliberately took advantage of this 'feature', as I believe *anything* specified either on the command line or in a pre-existing `cantera.conf` needs to take precedent. Unfortunately SCons doesn't expand variables in
```Python
opts = Variables("cantera.conf")
```
so I didn't see an obvious alternative solution. So overall, I stuck with this implementation as it reliably captures all options that are specified by the user.

bryanwweber: I can't get over that it somehow feels wrong or weird to be parsing the secondary level of configuration specification...

> Unfortunately SCons doesn't expand variables

I don't see how this is a problem? I assume you mean shell variables (like `$CONDA_PREFIX`). This is done by string operations somewhere around here (`os.path.expandvars` and `os.path.expanduser`).

If a variable is not specified in `cantera.conf` or on the command line (which are really equivalent), then they should still be equal to their default values as defined in the `Options` structure above. Checking whether the values for `layout`, `prefix`, `python_prefix`, and `python_cmd` are different from their default value (opts[value].default, I think...) should be equivalent and get directly to the data source that SCons is using to do its work.

For example, the default value of `python_prefix` is `"$prefix"` (as of this PR, #1158 changes that to the empty string, but that's not really important here).

ischoegl: PI’m not sure that we are disagreeing. The ‘secondary’ level succinctly contains what was specified? Even before this PR, this is exactly what was displayed for the summary.

The one case I am concerned about is when a user specifies the the default value on the command line (or cantera.conf). For example, if you specify `standard` as `layout` this would be equivalent to the default on some OSs, but would be overriden if conda is detected unless you query what was actually specified? You can probably come up with other scenarios where checking against the default could have unintended outcomes. Overall, I believe the current implementation does what it is supposed to?

ischoegl: > Unfortunately SCons doesn't expand variables

this refers to variables read from `cantera.conf`. It’s hard to get to those

SConstruct
```diff
@@ -819,6 +823,9 @@ config["python_cmd"].default = sys.executable
 opts.AddVariables(*config.to_scons())
 opts.Update(env)
 opts.Save('cantera.conf', env)
+cantera_conf = []
+for line in open('cantera.conf'):
+    cantera_conf.append(line.strip())
```
```suggestion
cantera_conf = [Path("cantera.conf").read_text().splitlines()]
```
Can add `l.strip() for l in ...` if you want

bryanwweber: Actually, I think this can be deleted here, see the next couple comments...

SConstruct
```diff
@@ -866,10 +873,10 @@ except subprocess.CalledProcessError:
...
 
 # Print values of all build options:
-print("Configuration variables read from 'cantera.conf' and command line:")
-for line in open('cantera.conf'):
-    print('   ', line.strip())
-print()
+msg = ["Configuration variables read from 'cantera.conf' and command line:"]
+for line in cantera_conf:
+    msg.append(f"    {line}")
+logger.info("\n".join(msg))
```
```suggestion
cantera_conf = Path("cantera.conf").read_text()
logger.info("Configuration variables read from 'cantera.conf' and command line:")
logger.info(cantera_conf)
```

SConstruct
```diff
@@ -1626,26 +1646,59 @@ if env['matlab_toolbox'] == 'y':
 # /usr/local because of dist-packages vs site-packages
 env['debian'] = any(name.endswith('dist-packages') for name in sys.path)
 
+# Check whether Cantera should be installed into a conda environment
+if conda_prefix is not None:
+    if env["layout"] != "conda":
+        # identify options selected either on command line or in cantera.conf
+        selected_options = set(line.split("=")[0].strip() for line in cantera_conf)
```
```suggestion
        selected_options = set(line.split("=")[0].strip() for line in cantera_conf.splitlines())
```

SConstruct
```diff
@@ -1626,26 +1646,59 @@ if env['matlab_toolbox'] == 'y':
...
+        use_conda = not selected_options & \
+            {"layout", "prefix", "python_prefix", "python_cmd"}
```
bryanwweber: Can you group this operation with parentheses instead of the backslash continuation? Otherwise, the second set could become its own variable for clarity.

SConstruct
```diff
@@ -905,6 +912,19 @@ if os.pathsep == ';':
...
+conda_prefix = os.environ.get("CONDA_PREFIX")
+if conda_prefix is not None:
+    if os.name == "nt":
+        conda_inc_dir = pjoin(conda_prefix, "Library", "include")
+        conda_lib_dir = pjoin(conda_prefix, "Library", env["libdirname"])
+    else:
+        conda_inc_dir = pjoin(conda_prefix, "include")
+        conda_lib_dir = pjoin(conda_prefix, env["libdirname"])
+    env["extra_inc_dirs"] = env["extra_inc_dirs"] + [conda_inc_dir]
+    env["extra_lib_dirs"] = env["extra_lib_dirs"] + [conda_lib_dir]
```
```suggestion
    env["extra_inc_dirs"] = env["extra_inc_dirs"].append(conda_inc_dir)
    env["extra_lib_dirs"] = env["extra_lib_dirs"].append(conda_lib_dir)
```
Creating a list so that it can be added seems excessive here to me 😄

SConstruct
```diff
@@ -905,6 +912,19 @@ if os.pathsep == ';':
...
+    logger.info("Adding conda include and library paths.")
```
bryanwweber: Can you add at least the common part of the paths to the info message?

SConstruct
```diff
@@ -1626,26 +1646,59 @@ if env['matlab_toolbox'] == 'y':
...
+    env["ct_incdir"] = pjoin(env["prefix"], "Library", "include", "cantera")
+    env["ct_incroot"] = pjoin(env["prefix"], "Library", "include")
+else:
+    env["prefix"] = os.path.normpath(env["prefix"])
+    env["ct_libdir"] = pjoin(env["prefix"], env["libdirname"])
+    env["ct_bindir"] = pjoin(env["prefix"], "bin")
+    env["ct_python_bindir"] = pjoin(env["prefix"], "bin")
+    env["ct_incdir"] = pjoin(env["prefix"], "include", "cantera")
+    env["ct_incroot"] = pjoin(env["prefix"], "include")
+env["ct_installroot"] = env["prefix"]
```
bryanwweber, line 1695: Can you add a newline before this line to more clearly indicate it doesn't belong in the `else` block?

SConstruct
```diff
@@ -1984,11 +2022,11 @@ def postInstallMessage(target, source, env):
...
-          applications                {ct_bindir!s}
-          library files               {ct_libdir!s}
-          C++ headers                 {ct_incroot!s}
-          samples                     {ct_sampledir!s}
-          data files                  {ct_datadir!s}""".format(**env_dict))
+          applications                {inst_bindir!s}
+          library files               {inst_libdir!s}
+          C++ headers                 {inst_incroot!s}
+          samples                     {inst_sampledir!s}
+          data files                  {inst_datadir!s}""".format(**env_dict))
```
bryanwweber: I suspect that these were the `ct`-prefixed variables because the `inst` variables include the stage, which hasn't been printed here in the past. I'm fine either way 🤷‍♂️

bryanwweber (changes requested):
OK, one last round of suggestions here, I hope. Now that we've sorted out the `set` issue 😄

SConstruct
bryanwweber: Could also do `textwrap.indent(cantera.conf, '    ')` to retain the previous behavior.

SConstruct
ischoegl: Correct. From what I understand, this should be `stage` (before this PR, changing some of the `ct_*dir` variables would not have an effect on the actual installation locations).

ischoegl:
@bryanwweber ... thanks for the suggestions! I adopted all. I further double-checked that things still work as expected on Windows, and the Linux variant is taken care of by the Sundials runners. Should be good to go!

speth:
I'd like a chance to review this before it gets merged, but won't be able to do so immediately. I saw few comments that make me think that this might not be doing the right thing for "staged" installations, and I'd like a chance to check that first.

bryanwweber:
@ischoegl I don't see a reason to block `conda` layouts for staged installs. I'm also planning to check that option.

ischoegl:
> this might not be doing the right thing for "staged" installations, [...]

@speth ... good catch! I added `stage_dir` to the 'blocking' options, as I cannot see a scenario where this would be doing the right thing for a `conda` layout. Also added a couple of comments - the logic should be easy to follow at this point.

_... fwiw ..._
```
$ scons help --option=stage_dir
INFO: SCons is using the following Python interpreter: /home/ischoegl/.conda/envs/cantera-dev/bin/python

* stage_dir: [ path/to/stage_dir ]
    Directory relative to the Cantera source directory to be used as a staging
    area for building for example, a Debian package. If specified, 'scons
    install' will install files to 'stage_dir/prefix/...'.
    - default: ''
```

bryanwweber:
@ischoegl Can you please remove the block on `stage_dir` for Conda environments? I use `stage_dir` a lot to test whether options are working without putting files into the real locations on my filesystem. All that should happen with `stage_dir` is that it is prepended to all the installation paths, so the conda layout can be contained within the `stage_dir`.

ischoegl:
> I use `stage_dir` a lot to test whether options are working without putting files into the real locations on my filesystem. All that should happen with `stage_dir` is that it is prepended to all the installation paths, so the `conda` layout can be contained within the `stage_dir`.

@bryanwweber ... could you explain further? The `conda_prefix` detected in this PR is an absolute file path, as stated in the logging output:
```
[...]
INFO: Using conda environment as default 'prefix': /home/ischoegl/.conda/envs/cantera-dev
[...]
```
so prepending a `stage_dir` doesn't apply?

Fwiw, one of the benefits of this PR would be to have an efficient way of keeping files outside of the ordinary file system and to instead install into a `conda` environment. So in a way, I see the `conda` layout as a feature that is orthogonal to (and mutually exclusive with) `stage_dir`.

bryanwweber:
> @bryanwweber ... could you explain further? The `conda_prefix` detected in this PR is an absolute file path, as stated in the logging output:
> 
> ```
> [...]
> INFO: Using conda environment as default 'prefix': /home/ischoegl/.conda/envs/cantera-dev
> [...]
> ```
> 
> so prepending a `stage_dir` doesn't apply?

Why not? If `stage_dir` is `stage` (relative to the source root), everything would go into `stage/home/ischoegl/.conda/envs/cantera-dev`. That way I don't have to actually have/create those folders and then remember to delete them.

> So in a way, I see the `conda` layout as a feature that is orthogonal to (and mutually exclusive with) `stage_dir`.

I agree they're orthogonal, which is why I don't think `stage_dir` and `conda` layout should be mutually exclusive 😄

ischoegl:
@bryanwweber ... I start to see what you're talking about, but in this case, the following doesn't make sense to me.
```
* stage_dir: [ path/to/stage_dir ]
    Directory relative to the Cantera source directory to be used as a staging
    area for building for example, a Debian package. If specified, 'scons
    install' will install files to 'stage_dir/prefix/...'.
    - default: ''
```
I assume that the staging feature is important for packaging purposes? (for other work, I think a conda environment is already equivalent to a 'staging' environment). Could you give me an example for what the output folder should look like? (*PS: both windows and `*nix` versions*)

bryanwweber:
@ischoegl Yes, that's perhaps the original intention of `stage_dir`. I also find it useful to test the behavior of `scons install` without putting files in their real homes in the filesystem, so that I don't accidentally corrupt anything and so I don't have to clean up folders out of a bunch of places.

Since the `$CONDA_PREFIX` is usually only set (I suppose I could do it manually) when an environment is activated, I don't want to install Cantera into that environment if I'm using it for dev, because I want to, for example, run the tests out of the `build` directory.

I don't think you need to handle anything. As the docs say, everything goes into `stage_dir/prefix`, so as long as you set `prefix` right, everything else should just work. But that's what I want to check.

ischoegl:
@bryanwweber ... I think in this case, the safest option is to not replicate `conda_prefix` in the installation path at all. I.e. if you specify `stage_dir=conda` *without* a prefix then things end up in `<cantera_src_dir>/conda/...`

Would this work? (just pushed a version that *should* accomplish that.)

bryanwweber:
> @bryanwweber ... I think in this case, the safest option is to not replicate `conda_prefix` in the installation path at all. I.e. if you specify `stage_dir=conda` _without_ a prefix then things end up in `<cantera_src_dir>/conda/...`
> 
> Would this work? (just pushed a version that _should_ accomplish that.)

What I want is to replicate the exact conda installation paths inside the `stage_dir` so that I can verify that it all looks correct. So your most recent change is also going to not quite accomplish the right thing, since the conda path structure is not set as the `prefix` when `stage_dir` is set.

ischoegl:
> What I want is to replicate the exact conda installation paths inside the `stage_dir` so that I can verify that it all looks correct.

That's easy on `*nix`, but will fail on windows. Regarding the other issue, that should be fixed. If you have a robust suggestion for a (truncated) prefix generated from `conda_prefix`, we can likely put this to rest. What about the last three 'parents', i.e.
* `stage/.conda/envs/cantera-dev` (`*nix`) or  
* `stage\miniconda3\envs\cantera-dev` (windows)?

ischoegl:
ok. ran a test for `scons build stage_dir=stage` on windows, with the following output
```
postInstallMessage(["finish_install"], [])

File locations:

  applications                C:\Users\<user>\GitHub\cantera\stage\miniconda3\envs\cantera-dev\Scripts
  library files               C:\Users\<user>\GitHub\cantera\stage\miniconda3\envs\cantera-dev\Library\lib
  C++ headers                 C:\Users\<user>\GitHub\cantera\stage\miniconda3\envs\cantera-dev\Library\include
  samples                     C:\Users\<user>\GitHub\cantera\stage\miniconda3\envs\cantera-dev\share\cantera\samples
  data files                  C:\Users\<user>\GitHub\cantera\stage\miniconda3\envs\cantera-dev\share\cantera\data
  Python package (cantera)    \Users\<user>\miniconda3\envs\cantera-dev\Lib\site-packages
  Python samples              \Users\<user>\miniconda3\envs\cantera-dev\Lib\site-packages\cantera\examples

scons: done building targets.
```
So it does look like the `python` packages aren't redirected as required.

*Edit: Python Package is not installed in correct location, but*
```
C:\Users\<user>\GitHub\cantera\stage\Users\<user>\miniconda3\envs\cantera-dev\Lib\site-packages\cantera
```

ischoegl:
@bryanwweber / @speth ... I ran some tests but am not completely following why the logic used for setting `'python_module_loc'` fails when `stage_dir` is set (on Windows). I'm not seeing what the conda layout does differently for things to break? ~I'm wondering whether this is OS specific?~

... *update* ... 
```
postInstallMessage(["finish_install"], [])

File locations:

  applications                /work/GitHub/cantera/stage/.conda/envs/cantera-dev/bin
  library files               /work/GitHub/cantera/stage/.conda/envs/cantera-dev/lib
  C++ headers                 /work/GitHub/cantera/stage/.conda/envs/cantera-dev/include
  samples                     /work/GitHub/cantera/stage/.conda/envs/cantera-dev/share/cantera/samples
  data files                  /work/GitHub/cantera/stage/.conda/envs/cantera-dev/share/cantera/data  
  Python package (cantera)    /.conda/envs/cantera-dev/lib/python3.10/site-packages
  Python samples              /.conda/envs/cantera-dev/lib/python3.10/site-packages/cantera/examples

Setup scripts to configure the environment for Cantera are at:

  setup script (bash)         /work/GitHub/cantera/stage/.conda/envs/cantera-dev/bin/setup_cantera
  setup script (csh/tcsh)     /work/GitHub/cantera/stage/.conda/envs/cantera-dev/bin/setup_cantera.csh

It is recommended that you run the script for your shell by typing:

  source /work/GitHub/cantera/stage/.conda/envs/cantera-dev/bin/setup_cantera

before using Cantera, or else include its contents in your shell login script.

scons: done building targets.
```
*Edit: Python is installed in the expected locations, although they are not displayed correctly*

ischoegl:
Alright ... mystery solved: on Linux, the `stage_dir` option works, but the wrong Python installation folders are displayed in the post-install message. On Windows, the Python module is installed into the stagedir, but not correctly. I believe this due to how `python_module_loc` is handled. At this point, this PR fixes most of the displayed installation locations, but the Python issue runs deeper.

On windows, the installation folder for the Python package is
```
 C:\Users\ischo\GitHub\cantera\stage\Users\ischo\miniconda3\envs\cantera-dev\Lib\site-packages\cantera
```

SConstruct
ischoegl: Prior to this PR, the paths were all incorrect if `stage_dir` is specified. On `*nix`, the installation appears to be successful though. As this is a pre-existing bug that also installs Python into an incorrect folder on Windows, I am not fixing this here.

ischoegl:
@bryanwweber / @speth ... I just opened #1194 as the issues with the 'stage' location opens a can of worms that goes well beyond the scope of what this PR intended to achieve. I rolled back all the 'fixes' for the post-build message, as I believe that they need to be addressed in another PR with a more narrow scope.

In a nutshell:
* Installation with the `conda` layout works without restrictions if no `stage_dir` is used
* If a `stage_dir` is used, installation works on `*nix` although installation locations are displayed incorrectly
* On Windows, there are issues with the Python package installation

There are two reasons why I'd like to defer a full resolution: (i) #1158 may change some of the behavior, and (ii) having the `conda` layout available will provide for a convenient means to resolve #1194.

That said, I believe the PR is ready with the caveats stated above.

.github/workflows/main.yml
```diff
@@ -313,9 +313,7 @@ jobs:
           cython boost-cpp fmt eigen yaml-cpp h5py pandas libgomp openblas pytest
       - name: Build Cantera
         run: |
-          scons build extra_inc_dirs=$CONDA_PREFIX/include:$CONDA_PREFIX/include/eigen3 \
-          extra_lib_dirs=$CONDA_PREFIX/lib system_fmt=y system_eigen=y system_yamlcpp=y \
-          system_sundials=y blas_lapack_libs='lapack,blas' -j2 VERBOSE=True debug=n \
+          scons build blas_lapack_libs='lapack,blas' -j2 VERBOSE=True debug=n \
```
speth: I think it was actually helpful to have the explicit `system_foo=y` options specified here to make sure that we were actually using a "system" installation of these libraries, rather than inadvertently falling back to the submodules.

SConstruct
```diff
@@ -905,6 +909,19 @@ if os.pathsep == ';':
...
+conda_prefix = os.environ.get("CONDA_PREFIX")
+if conda_prefix is not None:
+    if os.name == "nt":
+        conda_inc_dir = pjoin(conda_prefix, "Library", "include")
+        conda_lib_dir = pjoin(conda_prefix, "Library", env["libdirname"])
+    else:
+        conda_inc_dir = pjoin(conda_prefix, "include")
+        conda_lib_dir = pjoin(conda_prefix, env["libdirname"])
+    env["extra_inc_dirs"].append(conda_inc_dir)
+    env["extra_lib_dirs"].append(conda_lib_dir)
```
speth, line 931: This behavior should be described in the docstrings for the `extra_inc_dirs` and `extra_lib_dirs` options so users know they don't need to specify the Conda directories.

SConstruct
```diff
@@ -553,9 +555,11 @@ config_options = [
            conjunction with "prefix='/usr/local'". 'compact' puts all installed
            files in the subdirectory defined by 'prefix'. This layout is best
            with a prefix like '/opt/cantera'. 'debian' installs to the stage
-           directory in a layout used for generating Debian packages.""",
+           directory in a layout used for generating Debian packages. If the Python
+           executable found during compilation is managed by 'conda', the layout
+           will default to 'conda' irrespective of operating system.""",
```
speth: I think this also needs to provide some description of what the `conda` layout does, that is, that it installs Cantera into the active conda environment.

SConstruct
```diff
@@ -217,8 +217,10 @@ config_options = [
         }),
     PathOption(
         "prefix",
-        """Set this to the directory where Cantera should be installed. On Windows
-           systems, '$ProgramFiles' typically refers to "C:\Program Files".""",
+        """Set this to the directory where Cantera should be installed. If the Python
+           executable found during compilation is managed by 'conda', the
+           installation 'prefix' will default to the corresponding environment. On
+           Windows systems, '$ProgramFiles' typically refers to "C:\Program Files".""",
```
speth, line 225: I think this should also mention the other preconditions for chosing the conda environment as the `prefix`.

speth (changes requested):
Thanks for implementing this, @ischoegl. I think this will simplify from-source installation a lot in one of the most common use scenarios. My suggestions here are mostly around expanding the documentation of this feature.

ischoegl:
@speth ... thank you for the suggestions - these were all easily adopted. While the `Sundials` runner does detect all system installs correctly, I agree that it's safer to leave the explicit options in. 

@bryanwweber ... unless you have any other suggestions, I believe this is ready to merge!

speth (approved):

bryanwweber (approved):

PR #1201, opened 2022-02-19
Translate warnings to Python

### LINKED ISSUE #136: Channel C++ warnings to Python's warning system
While exceptions raised in the C++ core are already captured and thrown by the Python API, the same is not true for the warning system. I had originally added this as a discussion topic, but it may be more appropriate to discuss this as an issue (feature request).

### Discussed in https://github.com/Cantera/enhancements/discussions/112

<div type='discussions-op-text'>

<sup>Originally posted by **ischoegl** August 18, 2021</sup>
At the moment, the C++ code base issues warning messages via `warn_deprecated` and `warn_user` (the latter introduced in Cantera/cantera#741). In this context, @bryanwweber commented on a new deprecation warning proposed in Cantera/cantera#1084:

> One other comment, and I suppose this applies to all the `warn_deprecated()` calls in C++, I wonder if it's possible to catch those in Python and turn them into real Python `DeprecationWarnings` so they can be controlled by Python's warning interfaces? 

(see [here](https://github.com/Cantera/cantera/pull/1084#issuecomment-901377885); the same presumably applies to `warn_user`).

@speth's response was:

> I think this would require something a bit like how we interface with Python for writing messages to `stdout`, that is, registering a handler in the `Application` class and in that handler making whatever the correct calls are to the Python C API. 

**Additional context:**

* Cantera/cantera#796
* Comment about [Boost logger](https://github.com/Cantera/cantera/pull/741#pullrequestreview-314480685) in Cantera/cantera#741

**Other thoughts:**

As there may be a relatively straight-forward route to this after all, it may make sense to convert this discussion to an enhancement request (i.e. issue).</div>

ischoegl:

- Translate warnings raised in C++ to Python warnings: `warn_deprecated` is raised as `DeprecationWarning` and `warn_user` is raised as `UserWarning` (on C++ the output is changed from `CanteraWarning` to `UserWarning` for consistency)
- Update year in License

Closes Cantera/enhancements#136

```
>>> import cantera as ct
>>> gas = ct.Solution("h2o2.cti")
<stdin>:1: DeprecationWarning: XML_Node::build: 
The CTI and XML input file formats are deprecated and will be removed in
Cantera 3.0. Use 'cti2yaml.py' or 'ctml2yaml.py' to convert CTI or XML input
files to the YAML format. See https://cantera.org/tutorials/legacy2yaml.html
for more information.
```
and
```
>>> import warnings
>>> import cantera as ct
>>> with warnings.catch_warnings():
...     warnings.simplefilter("ignore")
...     gas = ct.Solution("gri30.cti")
... 
>>> gas
<cantera.composite.Solution object at 0x7f630d950dc0>
```

Another nice thing is that users will be pointed to the line in existing Python code where the warning is invoked, e.g.
```
In [1]: %run custom_reactions.py
/work/GitHub/cantera/interfaces/cython/cantera/examples/kinetics/custom_reactions.py:34: DeprecationWarning: XML_Node::build: 
The CTI and XML input file formats are deprecated and will be removed in
Cantera 3.0. Use 'cti2yaml.py' or 'ctml2yaml.py' to convert CTI or XML input
files to the YAML format. See https://cantera.org/tutorials/legacy2yaml.html
for more information.
```
(this is from an instance where a `DeprecationWarning` in an example was previously not caught by the test suite; this is also addressed in this PR)

**Checklist**

### DIFF
--- a/License.txt
+++ b/License.txt
@@ -6,7 +6,7 @@ Copyright (c) 2009 Sandia Corporation. Under the terms of
 Contract AC04-94AL85000 with Sandia Corporation, the U.S. Government
 retains certain rights in this software.
 
-Copyright (c) 2011-2021, Cantera Developers.
+Copyright (c) 2011-2022, Cantera Developers.
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/include/cantera/base/global.h
+++ b/include/cantera/base/global.h
@@ -195,6 +195,22 @@ void writeline(char repeat, size_t count,
 //! helper function passing deprecation warning to global handler
 void _warn_deprecated(const std::string& method, const std::string& extra="");
 
+//! Print a deprecation warning raised from *method*. @see Application::warn_deprecated
+/*!
+ * @param method  method name
+ * @param msg  Python-style format string with the following arguments
+ * @param args  arguments for the format string
+ */
+template <typename... Args>
+void warn_deprecated(const std::string& method, const std::string& msg,
+                     const Args&... args) {
+    if (sizeof...(args) == 0) {
+        _warn_deprecated(method, msg);
+    } else {
+        _warn_deprecated(method, fmt::format(msg, args...));
+    }
+}
+
 //! @copydoc Application::suppress_deprecation_warnings
 void suppress_deprecation_warnings();
 
@@ -232,22 +248,6 @@ void warn_user(const std::string& method, const std::string& msg,
         _warn("Cantera", method, msg);
     } else {
         _warn("Cantera", method, fmt::format(msg, args...));
-    }
-}
-
-//! Print a deprecation warning raised from *method*. @see Application::warn_deprecated
-/*!
- * @param method  method name
- * @param msg  Python-style format string with the following arguments
- * @param args  arguments for the format string
- */
-template <typename... Args>
-void warn_deprecated(const std::string& method, const std::string& msg,
-                     const Args&... args) {
-    if (sizeof...(args) == 0) {
-        _warn_deprecated(method, msg);
-    } else {
-        _warn_deprecated(method, fmt::format(msg, args...));
     }
 }
 
--- a/include/cantera/base/global.h
+++ b/include/cantera/base/global.h
@@ -219,7 +219,7 @@ void warn(const std::string& warning, const std::string& method,
     }
 }
 
-//! Print a user warning raised from *method*.
+//! Print a user warning raised from *method* as `CanteraWarning`.
 /*!
  * @param method  method name
  * @param msg  Python-style format string with the following arguments
@@ -229,9 +229,9 @@ template <typename... Args>
 void warn_user(const std::string& method, const std::string& msg,
                const Args&... args) {
     if (sizeof...(args) == 0) {
-        _warn("User", method, msg);
+        _warn("Cantera", method, msg);
     } else {
-        _warn("User", method, fmt::format(msg, args...));
+        _warn("Cantera", method, fmt::format(msg, args...));
     }
 }
 
--- a/include/cantera/base/global.h
+++ b/include/cantera/base/global.h
@@ -192,18 +192,35 @@ void writelogendl();
 void writeline(char repeat, size_t count,
                bool endl_after=true, bool endl_before=false);
 
-//! @copydoc Application::warn_deprecated
-void warn_deprecated(const std::string& method, const std::string& extra="");
+//! helper function passing deprecation warning to global handler
+void _warn_deprecated(const std::string& method, const std::string& extra="");
 
 //! @copydoc Application::suppress_deprecation_warnings
 void suppress_deprecation_warnings();
 
-//! helper function passing user warning to global handler
-void _warn_user(const std::string& method, const std::string& extra);
+//! helper function passing generic warning to global handler
+void _warn(const std::string& warning,
+           const std::string& method, const std::string& extra);
 
+//! Print a generic warning raised from *method*. @see Application::warn
 /*!
- * Print a user warning raised from *method*.
- *
+ * @param warning  type of warning; @see Logger::warn
+ * @param method  method name
+ * @param msg  Python-style format string with the following arguments
+ * @param args  arguments for the format string
+ */
+template <typename... Args>
+void warn(const std::string& warning, const std::string& method,
+          const std::string& msg, const Args&... args) {
+    if (sizeof...(args) == 0) {
+        _warn(warning, method, msg);
+    } else {
+        _warn(warning, method, fmt::format(msg, args...));
+    }
+}
+
+//! Print a user warning raised from *method*.
+/*!
  * @param method  method name
  * @param msg  Python-style format string with the following arguments
  * @param args  arguments for the format string
@@ -212,9 +229,25 @@ template <typename... Args>
 void warn_user(const std::string& method, const std::string& msg,
                const Args&... args) {
     if (sizeof...(args) == 0) {
-        _warn_user(method, msg);
+        _warn("User", method, msg);
     } else {
-        _warn_user(method, fmt::format(msg, args...));
+        _warn("User", method, fmt::format(msg, args...));
+    }
+}
+
+//! Print a deprecation warning raised from *method*. @see Application::warn_deprecated
+/*!
+ * @param method  method name
+ * @param msg  Python-style format string with the following arguments
+ * @param args  arguments for the format string
+ */
+template <typename... Args>
+void warn_deprecated(const std::string& method, const std::string& msg,
+                     const Args&... args) {
+    if (sizeof...(args) == 0) {
+        _warn_deprecated(method, msg);
+    } else {
+        _warn_deprecated(method, fmt::format(msg, args...));
     }
 }
 
--- a/include/cantera/base/logger.h
+++ b/include/cantera/base/logger.h
@@ -74,13 +74,7 @@ public:
      * @param msg      String message to be written to cout
      */
     virtual void warn(const std::string& warning, const std::string& msg) {
-
-        std::clog << warning << "Warning: " << msg;
-    }
-
-    //! Write an end of line character to the warning and flush output.
-    virtual void warnendl() {
-        std::clog << std::endl;
+        std::clog << warning << "Warning: " << msg << std::endl;
     }
 
     //! Write an error message and quit.
--- a/include/cantera/base/logger.h
+++ b/include/cantera/base/logger.h
@@ -67,6 +67,22 @@ public:
         std::cout << std::endl;
     }
 
+    //! Write a warning message.
+    /*!
+     * The default behavior is to write to the logging output.
+     * @param warning  String specifying type of warning; @see Logger::warn
+     * @param msg      String message to be written to cout
+     */
+    virtual void warn(const std::string& warning, const std::string& msg) {
+
+        std::clog << warning << "Warning: " << msg;
+    }
+
+    //! Write an end of line character to the warning and flush output.
+    virtual void warnendl() {
+        std::clog << std::endl;
+    }
+
     //! Write an error message and quit.
     /*!
      * The default behavior is to write to the standard error stream, and then
--- a/include/cantera/cython/wrappers.h
+++ b/include/cantera/cython/wrappers.h
@@ -9,6 +9,23 @@
 
 #include "Python.h"
 
+// Warning types supported by the Python C-API.
+// See https://docs.python.org/3/c-api/exceptions.html#issuing-warnings
+std::map<std::string, PyObject*> mapped_PyWarnings = {
+    {"", PyExc_Warning},
+    {"Bytes", PyExc_BytesWarning},
+    {"Cantera", PyExc_UserWarning}, // pre-existing warning
+    {"Deprecation", PyExc_DeprecationWarning},
+    {"Future", PyExc_FutureWarning},
+    {"Import", PyExc_ImportWarning},
+    {"PendingDeprecation", PyExc_PendingDeprecationWarning},
+    {"Resource", PyExc_ResourceWarning},
+    {"Runtime", PyExc_RuntimeWarning},
+    {"Syntax", PyExc_SyntaxWarning},
+    {"Unicode", PyExc_UnicodeWarning},
+    {"User", PyExc_UserWarning}
+};
+
 // Wrappers for preprocessor defines
 std::string get_cantera_version()
 {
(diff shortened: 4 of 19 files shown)


License.txt
```diff
@@ -6,7 +6,7 @@ Copyright (c) 2009 Sandia Corporation. Under the terms of
 Contract AC04-94AL85000 with Sandia Corporation, the U.S. Government
 retains certain rights in this software.
 
-Copyright (c) 2011-2021, Cantera Developers.
+Copyright (c) 2011-2022, Cantera Developers.
```
bryanwweber, line 9: There are lots of opinions on the web about this, most saying that leaving the second year off implies "to present". I think I'd like to do that to avoid having to update this every year.

bryanwweber: See
* https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated
* https://softwareengineering.stackexchange.com/q/130478
* https://softwareengineering.stackexchange.com/q/210472
* https://softwareengineering.stackexchange.com/q/227812

include/cantera/base/logger.h
```diff
@@ -67,6 +67,20 @@ class Logger
         std::cout << std::endl;
     }
 
+    //! Write a warning message.
+    /*!
+     * The default behavior is to write to the logging output.
+     * @param msg      String message to be written to cout
```
bryanwweber, line 74: Can you add the `warning` param here?

include/cantera/base/logger.h
```diff
@@ -67,6 +67,20 @@ class Logger
...
+     */
+    virtual void warn(const std::string& msg, const std::string& warning) {
```
bryanwweber: I noticed that you've got the message as the first argument and the type of warning second. This is the opposite of the `Py_*` functions, so it'd be nice to be consistent. Unless, of course, this is consistent with other Cantera functions, in which case, it should stay 😄

include/cantera/cython/wrappers.h
```diff
@@ -37,9 +37,20 @@ class PythonLogger : public Cantera::Logger
         std::cout.flush();
     }
 
+    virtual void warn(const std::string& msg, const std::string& warning) {
```
bryanwweber: Can you add explicit tests for this warning with `pytest.warns(..., match="...")`? These are showing up as not covered.

interfaces/cython/cantera/examples/kinetics/custom_reactions.py
```diff
@@ -31,7 +32,9 @@
 
 # old framework - use xml input
 
-gas2 = ct.Solution(mech.replace('.yaml', '.xml'))
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
+    gas2 = ct.Solution(mech.replace(".yaml", ".xml"))
```
bryanwweber: I actually think it'd be better to leave the warning message here, so anyone who copy-pastes this gets that warning and reads about it.

bryanwweber (changes requested):
Thanks @ischoegl! This looks like it will be a nice change, and it'll be nice to be able to manage warnings by built-in Python methods.

interfaces/cython/cantera/examples/kinetics/custom_reactions.py
ischoegl: This one is unfortunately breaking CI, as the warning is raised to an exception.

bryanwweber: Ah, that's annoying. I wonder if there's an easy way to fix that... With pytest, it must be possible to run these examples now, possibly just by `import`ing them... then we wouldn't have to do the hacky `find ... -exec ...`... hmm. That'd be a pretty significant scope expansion here though. What do you think?

ischoegl: I think I'll add a comment to the example, which is a lot simpler. The speed-test part comparing things to XML will have to be removed after 2.6 regardless - this part of the example was mainly intended as a coarse benchmark test (which btw was highly useful).

ischoegl: The `pytest` route is imho a separate topic / issue. As an unrelated note, I've used pytest to cover jupyter notebooks btw, see example [here](https://github.com/ischoegl/ctwrap/blob/main/tests/test_notebooks.py)

License.txt
ischoegl: NP. I skipped the second year in `License.txt`, but changed the verbiage to `2001-present` in Sphinx' `conf.py`.

include/cantera/base/logger.h
ischoegl: Good point. I switched the order around and created a more generic warning type (starting in `global.h`).

ischoegl:
@bryanwweber ... thank you for the review!

I ended up implementing a more comprehensive C++ backend for a more fine-grained warning system that should mesh well with various API's. I also included a test that ensures that warnings are raised correctly.

include/cantera/base/global.h
```diff
@@ -212,9 +229,25 @@ template <typename... Args>
...
+
+//! Print a deprecation warning raised from *method*. @see Application::warn_deprecated
+/*!
+ * @param method  method name
+ * @param msg  Python-style format string with the following arguments
+ * @param args  arguments for the format string
+ */
+template <typename... Args>
+void warn_deprecated(const std::string& method, const std::string& msg,
+               const Args&... args) {
```
bryanwweber: The indentation here looks strange in the diff view, can you double check it?

interfaces/cython/cantera/test/test_utils.py
```diff
@@ -6,6 +6,20 @@
...
+
+    def test_user(self):
+        yaml = """
+            type: falloff
+            low-P-rate-constant: {A: 2.3e+12, b: -0.9, Ea: -7112800.0}
+            high-P-rate-constant: {A: 7.4e+10, b: -0.37, Ea: 0.0}
+            Troe: {A: 0.7346, T3: 94.0, T1: 1756.0, T2: 0.}
+            """
+
+        with self.assertWarnsRegex(Warning, "Unexpected parameter value T2=0"):
```
bryanwweber: Might as well use the pytest version, `pytest.warns`. Should this be `UserWarning`?

bryanwweber (changes requested):
Thanks @ischoegl! A couple more small things here.

ischoegl:
@bryanwweber ... :sweat_smile: Testing the warning system ended up creating battles with some of the work-arounds that were put in since 2.5.1 for testing purposes, specifically 5f5a005 and 80c1e56. As those methods do not allow for fine-grained control and are de-facto redundant, I decided to take them out.

Regarding your comment about *pytest*: using warning filters would allow for further removing methods that @speth put in earlier, specifically `ct.make_deprecation_warnings_fatal()` and `ct.suppress_deprecation_warnings()`. Removing those would be a lot more intrusive and should probably be handled in a different PR (if at all) ...

interfaces/cython/cantera/test/test_utils.py
ischoegl: This is arguably a `SyntaxWarning` (which can be raised with this PR :smile:), but I'd prefer to leave all C++ warnings as `CanteraWarnings`, which are now converted to `UserWarnings` on the Python side. So the test (which is moved to `test_kinetics.py`) now checks for a `UserWarning`.

License.txt
```diff
@@ -6,7 +6,7 @@ Copyright (c) 2009 Sandia Corporation. Under the terms of
 Contract AC04-94AL85000 with Sandia Corporation, the U.S. Government
 retains certain rights in this software.
 
-Copyright (c) 2011-2021, Cantera Developers.
+Copyright (c) 2011, Cantera Developers.
```
speth: I disagree with the rationale for removing the range here. The range indicates years in which new copyrighted material was created. If anything, a single year implies that the work has not been modified since. Python itself lists every single year separately (see https://github.com/python/cpython/blob/main/LICENSE). A bit of Googling suggests that there is some argument that specifying a _range_ rather than multiple individual years is not a correct assertion of copyright.

include/cantera/base/global.h
```diff
@@ -212,30 +229,37 @@ template <typename... Args>
...
-void make_warnings_fatal();
-
 //! @copydoc Application::suppress_thermo_warnings
 void suppress_thermo_warnings(bool suppress=true);
 
 //! @copydoc Application::thermo_warnings_suppressed
 bool thermo_warnings_suppressed();
 
-//! @copydoc Application::suppress_user_warnings
-void suppress_warnings();
```
speth: I don't quite understand the rationale for removing `suppress_warnings` and `make_warnings_fatal`. Just because they aren't needed in the Python module anymore doesn't mean they aren't useful, for example for user C++ applications.

include/cantera/base/logger.h
```diff
@@ -67,6 +67,22 @@ class Logger
...
+     */
+    virtual void warn(const std::string& warning, const std::string& msg) {
+
+        std::clog << warning << "Warning: " << msg;
+    }
+
+    //! Write an end of line character to the warning and flush output.
+    virtual void warnendl() {
+        std::clog << std::endl;
+    }
```
speth, line 78: Given that the `warn` function is always meant to be a complete warning message, I think it would be simpler to just eliminate this and include a `std::endl` in the `warn` function above.

interfaces/cython/cantera/test/utilities.py
```diff
@@ -27,6 +27,15 @@ def wrapper(*args, **kwargs):
 
     return wrapper
 
+def suppress_temperature_derivative_warnings(test):
+    def wrapper(*args, **kwargs):
+        with warnings.catch_warnings():
+            # suppress warning raised for BlowersMasel and TwoTempPlasma derivatives
+            warnings.simplefilter("ignore")
+            test(*args, **kwargs)
```
speth: The name for this method seems a bit misleading -- how does this not just ignore all warnings?

src/base/application.h
```diff
@@ -137,6 +137,17 @@ class Application
         //! Write an end of line character to the screen and flush output
         void writelogendl();
 
+        //!  Write a warning message to the screen.
+        /*!
+         * @param warning  String specifying type of warning; @see Logger::warn
+         * @param msg  String to be written to the screen
+         * @ingroup textlogs
+         */
+        void warnlog(const std::string& warning, const std::string& msg);
```
speth, line 146: Any reason to not just name this `warn`, for symmetry with the global function?

include/cantera/base/global.h
```diff
@@ -192,18 +192,35 @@ void writelogendl();
 void writeline(char repeat, size_t count,
                bool endl_after=true, bool endl_before=false);
 
-//! @copydoc Application::warn_deprecated
-void warn_deprecated(const std::string& method, const std::string& extra="");
+//! helper function passing deprecation warning to global handler
+void _warn_deprecated(const std::string& method, const std::string& extra="");
```
speth, line 196: I think it would be more clear to put the implementation of `warn_deprecated` immediately after this.

speth (changes requested):
Thanks, @ischoegl. Overall, I think this is a good change, and most of the implementation makes sense to me. There were just a few particular issues that I wanted to address.

License.txt
bryanwweber: Would an unbounded `2011-` work for what you're thinking? If not, then continuing to bikeshed whether one commit a year is worth the effort is not worth the effort, and we can put back `2011-2022` 😄

interfaces/cython/cantera/test/utilities.py
bryanwweber: Also rather than suppressing the warnings, why not test for them? pytest has the `pytest.warns(..., match="")` context manager, which could probably be put into this decorator.

License.txt
ischoegl: One alternative option would be `2011-present` ... not sure whether this is permissible though.

src/base/application.h
ischoegl: I think there are naming conflicts, and I was aiming to mimic the `write`/`writelog` interface:
```
call of overloaded 'warn(const char [12], std::string)' is ambiguous
```
I would honestly like to keep the parallel implementations here ...

include/cantera/base/global.h
ischoegl: NP. I dropped the commits, but removed them from Python, as they interfere with the native warning system.

License.txt
ischoegl: Leaving it at `2011-2022`. we can re-address this down the line.

License.txt
speth: I'm unfamiliar with any precedent for "YYYY-present", and would rather stick to the example set by peers like NumPy ([docs](https://numpy.org/doc/stable/), [license](https://github.com/numpy/numpy/blob/main/LICENSE.txt)), Astropy ([docs](https://docs.astropy.org/en/stable/), [license](https://github.com/astropy/astropy/blob/main/LICENSE.rst)) or pandas ([docs](https://pandas.pydata.org/docs/), [license](https://github.com/pandas-dev/pandas/blob/main/LICENSE)), all of which use the year range notation in both locations. Or we can run this question by our friends at NumFOCUS who might be able to give an informed answer.

src/base/application.h
speth: I guess this doesn't affect the public API in any way, since `application.h` is purely internal.

License.txt
ischoegl: As this discussion is completely tangential to the main scope of this PR, I am reverting to the original approach, i.e. the full year is now given as the end date in both instances. This can be revisited as needed.

ischoegl:
@bryanwweber and @speth ... I believe both of your change requests are now fully addressed. Thanks for all suggestions!

speth (dismissed):

bryanwweber:
This passes for me if I run `pytest` directly, but not running `pytest` via `scons test-python`. Will revert those changes if I can't find a quick fix 😢

ischoegl:
@bryanwweber ... thanks for trying to simplify by introducing `pytest.fixture`s. While I agree that it's worthwhile testing this approach, it may be best to drop it from this PR. It's a tangent here and it can be easily tested in a dedicated PR?

bryanwweber:
@ischoegl I've got it working locally, I just need to commit & push

bryanwweber:
@ischoegl Presuming the checks pass this time, this should be good to :shipit:

bryanwweber (approved):

ischoegl:
@bryanwweber ... :tada: ... things passed. Looks like the fixture is pretty similar to what `unittest` did, which is good news! So thank you for looking into this! Beyond, I think all that's left is your approval :wink: ... :shipit:

bryanwweber:
@ischoegl Great minds!!! Merged 😄

PR #1204, opened 2022-02-21
Fix `SolutionArray` extra slice

### LINKED ISSUE #1203: Looping over a `SolutionArray` does not include `extra` items
**Problem description**

Accessing `extra` items from a `SolutionArray` in a loop gives an `AttributeError`.

**Steps to reproduce**

```python
import cantera as ct
import numpy as np

T = np.arange(900, 1000, 50)
estimated_ignition_delay_times = np.ones_like(T) * 0.005
ignition_delays = ct.SolutionArray(gas, shape=T.shape, extra={"tau": estimated_ignition_delay_times})
ignition_delays.TP = T, 40.0 * 101325.0
for state in ignition_delays:
    print(state._extra)
```

Output:

```
OrderedDict()
...
```

**Behavior**

Looping should include the extra information.

**System information**

- Cantera version: main, https://github.com/Cantera/cantera/commit/041559d56d25945d25f5c94b160538b7e091f2a2

ischoegl:

Closes #1203

```
In [1]: import cantera as ct
   ...: import numpy as np
   ...: 
   ...: gas = ct.Solution("gri30.yaml")
   ...: 
   ...: T = np.arange(900, 1000, 50)
   ...: estimated_ignition_delay_times = np.ones_like(T) * 0.005
   ...: ignition_delays = ct.SolutionArray(
   ...:     gas, shape=T.shape, extra={"tau": estimated_ignition_delay_times}
   ...: )
   ...: ignition_delays.TP = T, 40.0 * 101325.0
   ...: for state in ignition_delays:
   ...:     print(state.tau)
0.005
0.005
```
**Checklist**

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -617,14 +617,15 @@ class SolutionArray:
 
     def __getitem__(self, index):
         states = self._states[index]
+        extra = OrderedDict({key: val[index] for key, val in self._extra.items()})
         if(isinstance(states, list)):
             num_rows = len(states)
             if num_rows == 0:
                 states = None
-            return SolutionArray(self._phase, num_rows, states)
+            return SolutionArray(self._phase, num_rows, states, extra=extra)
         else:
             shape = states.shape[:-1]
-            return SolutionArray(self._phase, shape, states)
+            return SolutionArray(self._phase, shape, states, extra=extra)
 
     def __getattr__(self, name):
         if name in self._extra:
--- a/interfaces/cython/cantera/test/test_composite.py
+++ b/interfaces/cython/cantera/test/test_composite.py
@@ -224,6 +224,14 @@ class TestSolutionArrayIO(utilities.CanteraTest):
         self.assertIn('X', collected)
         self.assertEqual(collected['X'].shape, (0, self.gas.n_species))
 
+    def test_getitem(self):
+        states = ct.SolutionArray(self.gas, 10, extra={"index": range(10)})
+        for ix, state in enumerate(states):
+            assert state.index == ix
+
+        assert list(states[:2].index) == [0, 1]
+        assert list(states[100:102].index) == [] # outside of range
+
     def test_append_state(self):
         gas = ct.Solution("h2o2.yaml")
         gas.TPX = 300, ct.one_atm, 'H2:0.5, O2:0.4'


ischoegl:
@bryanwweber … this fix for the issue you reported is ready for review (and should be quick).

interfaces/cython/cantera/composite.py
```diff
@@ -617,14 +617,15 @@ def __init__(self, phase, shape=(0,), states=None, extra=None, meta=None):
 
     def __getitem__(self, index):
         states = self._states[index]
+        extra = OrderedDict({key: val[index] for key, val in self._extra.items()})
         if(isinstance(states, list)):
             num_rows = len(states)
             if num_rows == 0:
                 states = None
```
bryanwweber, line 624: Is there a test that exercises this branch? If so, I think it makes sense to see if `extra` needs to be handled in a similar way. On the other hand, I'm not sure how `num_rows` could end up being 0...

To clarify, my concern is that `extra[key]` would have some value associated with it, when it shouldn't.

bryanwweber (changes requested):
Thanks @ischoegl, one case that I think needs to be investigated here.

As a total aside, outside the scope of this PR, but just to write it down somewhere, it'd be nice if `SolutionArray` supported:
1. `len()`
2. `.shape`
3. Accessing `extra` keys that can't be accessed by attributed because they have an invalid character (`-`, ` `, etc.)

interfaces/cython/cantera/composite.py
ischoegl: Interesting. It can be empty if you grab multiple elements outside of the defined range. It turns out to do the right thing already, but I added this to the test regardless.

interfaces/cython/cantera/test/test_composite.py
```diff
@@ -224,6 +224,14 @@ def test_collect_data(self):
         self.assertIn('X', collected)
         self.assertEqual(collected['X'].shape, (0, self.gas.n_species))
 
+    def test_getitem(self):
+        states = ct.SolutionArray(self.gas, 10, extra={"index": range(10)})
+        for ix, state in enumerate(states):
+            assert state.index == ix
+
+        assert list(states[:2].index) == [0, 1]
+        assert list(states[100:102].index) == [] # outside of range
```
bryanwweber, line 233: It seems like this should raise an `IndexError`, but insofar as this maintains current behavior, it seems fine to me to keep it like this and defer adding the `IndexError`.

bryanwweber (approved):
Thanks! Once the tests pass this is good for me

PR #1205, opened 2022-02-21
Detect description in `cti2yaml`

ischoegl:

- [x] Automatically convert CTI header block to YAML description as default (?)
- [x] Provide option to override header conversion
- [x] Add 'validation', i.e. make sure that the resulting mechanism imports without errors (only checked when called from CLI - I.e. same approach as `ck2yaml`)

Further:

- Add some missing description blocks that were lost during conversion
- Move lone input file in `data/sample-data` to `sample/data`

Partially addresses Cantera/enhancements#22 and Cantera/enhancements#23

```
$ cti2yaml haca2.cti 
Wrote YAML mechanism file to 'haca2.yaml'.
Mechanism contains 3 species and 1 reactions.
Validating mechanism...
PASSED
```
(after also checking interface phases, this actually fails, see #1210). Or (using the same options already in place for `ck2yaml`)
```
$ cti2yaml haca2.cti --quiet --no-validate
```

The description is detected and available ... 
```
In [1]: import cantera as ct
   ...: gas = ct.Solution("haca2.yaml")
   ...: print(gas.input_header["description"])
HACA Mechanism - very rough prototype
          Reference:
 (1)  M. Frenklach, H. Wang,
      "Detailed Mechanism and Modeling of Soot Particle Formation"
      in Soot Formation in Combustion: Mechanisms and Models"
      J. Bockhorn Ed., Springer Verlag, Heidelberg (1994).
```

(note that the odd indent for 'Reference' is preserved from the [CTI file pulled verbatim from Cantera 2.5.1](https://github.com/Cantera/cantera/blob/2.5/test_problems/surfSolverTest/haca2.cti))

**Checklist**

**Other thoughts**

I spot-checked old XML input from 2.5 and there were only two instances where a comment was used, so I do not believe that this capability has to be added to XML (which is usually generated via `ctml_writer` anyways).


ischoegl:
@bryanwweber … I am marking this as ready for review despite some mysterious test failures. As far as I can tell, a Boolean flag is not being passed correctly from the test suite. The ‘interesting’ part is that failures are not repeatable (different runners either pass or fail), and the code itself works as intended. I am stumped …

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1553,7 +1554,7 @@ def get_yaml(self, out):
         pass
 
 
-def convert(filename=None, output_name=None, text=None):
+def convert(filename=None, output_name=None, text=None, header=True):
```
bryanwweber: I don't think that this needs to be configurable. If a user doesn't want the description in the output, they can remove it from the output file manually.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1585,16 +1586,18 @@ def convert(filename=None, output_name=None, text=None):
 
     try:
         if filename is not None:
-            with filename.open('r', encoding='latin-1') as f:
-                code = compile(f.read(), str(filename), 'exec')
+            with filename.open("r", encoding="utf-8") as f:
```
bryanwweber: I think this is deliberately `latin-1` to catch old Windows files. See here: https://stackoverflow.com/questions/13868140/showing-a-latin-1-character-in-a-utf-8-page for a description of the difference.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1632,9 +1631,23 @@ def convert(filename=None, output_name=None, text=None):
...
         sys.exit(4)
 
+    # get file description from header block
+    if header:
+        description = []
+        for line in text.splitlines():
+            # only consider comments in the initial header block
+            # comments start with '#'; there may be empty leading lines
+            if line.startswith("#") or not description:
+                description.append(line[1:].rstrip())
```
bryanwweber: This won't be able to capture descriptions where multiple paragraphs are separated by blank lines, like:

```python
# Comment

# Comment
```

I'm also not 100% sure this won't give an `IndexError` for files that start with blank lines (I don't remember if `splitlines()` retains the trailing `\n` or not). I _think_ something like this might be more clear:

```python
leading_lines = True
for line in text.splitlines():
    if not line.strip() and leading_lines:
        continue
    else:
        leading_lines = False
    if line.strip() and not line.startswith("#"):
        break
    description.append(line[1:].rstrip())
```

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1644,14 +1657,23 @@ def convert(filename=None, output_name=None, text=None):
...
         metadata = BlockMap([
             ('generator', 'cti2yaml'),
             ('cantera-version', '2.6.0a4'),
             ('date', formatdate(localtime=True)),
         ])
-        if filename is not None:
+        if filename != "<string>":
             metadata['input-files'] = FlowList([base])
+        if header:
+            metadata.yaml_set_comment_before_after_key("generator", before="\n")
```
bryanwweber:
```suggestion
        if description:
            metadata.yaml_set_comment_before_after_key("description", after="\n")
```

bryanwweber: That's what we actually want to happen, right?

bryanwweber (changes requested):
As I said below, I don't think adding the description to the output needs to be configurable. I find the option name confusing and anyways, it might resolve the test failures 😄

interfaces/cython/cantera/cti2yaml.py
ischoegl: The issue with `latin-1` is that having a regular `"` symbol in the header text causes crashes. The file `data/inputs/lithium_ion_battery.cti` is one example that causes failures.

interfaces/cython/cantera/cti2yaml.py
ischoegl: It does work with leading blank lines. Regarding concatenation of multiple comment blocks, not permitting this is a deliberate choice. `data/inputs/lithium_ion_battery.cti` is again a good example where you'd get lines that you may not want

interfaces/cython/cantera/cti2yaml.py
bryanwweber: I'd rather handle cases like that with the `error` kwarg to `open`, setting it to either `ignore` or `replace`.  I think `latin-1` files are _extremely_ common and changing this behavior can cause unexpected effects.

interfaces/cython/cantera/cti2yaml.py
bryanwweber: > It does work with leading blank lines.

Thanks! I wonder if the leading lines should be dropped though? I guess this will only collect one blank line if there are several of them?

> Regarding concatenation of multiple comment blocks, not permitting this is a deliberate choice.

OK, I think the other choice is the better one :smile: I'm not sure there's a good way to decide here... I guess I come down on the side of capturing more information since it doesn't affect the syntax of the file, rather than losing that information.

interfaces/cython/cantera/cti2yaml.py
ischoegl: what about having an `encryption` option for the interface?

bryanwweber: Sorry, I don't get what you mean?

ischoegl: Oh, I meant to type *encoding*. Just
```
cti2yaml haca2.cti --encoding=utf-8
```
etc. (I *believe* `argparse` can handle that)

bryanwweber: Sure, sounds good. Perhaps `--input-encoding` would be more clear?

ischoegl: Implemented `--input-encoding`.

interfaces/cython/cantera/cti2yaml.py
ischoegl: I ended up with the following:
```python
    # get file description from header block
    description = []
    for line in text.splitlines():
        # only consider comments in the initial contiguous comment block
        # comments start with '#'; there may be empty leading lines
        if line.startswith("#") or not description:
            description.append(line[1:].rstrip())
        else:
            break
    description = textwrap.dedent("\n".join(description).strip("\n"))
```
which I believe is short and does exactly what we want (`description` remains empty until the first `#` is encountered; after that, the first line not starting with `#` concludes the header block).

I ultimately think that only using the first comment block makes the most sense. Users can very easily update the CTI input to modify the description (presumably they are more familiar with the legacy CTI syntax than the new YAML format). I intentionally mentioned 'contiguous' in what I added to the `cti2yaml --help` text for documentation.

interfaces/cython/cantera/cti2yaml.py
ischoegl: Removed.

ischoegl:
@bryanwweber ... thanks for the suggestions! Omitting the configuration option indeed resolved the pipeline failures (although I still don't know *why* they happened in the first place).

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1553,7 +1554,7 @@ def get_yaml(self, out):
         pass
 
 
-def convert(filename=None, output_name=None, text=None):
+def convert(filename=None, output_name=None, text=None, encoding=None):
```
bryanwweber: IMO, the default here should be `latin-1`. I see that `latin-1` is the default from the CLI, but I think it should be the default for the API as well.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1585,16 +1586,18 @@ def convert(filename=None, output_name=None, text=None):
 
     try:
         if filename is not None:
-            with filename.open('r', encoding='latin-1') as f:
-                code = compile(f.read(), str(filename), 'exec')
+            with filename.open("r", encoding=encoding) as f:
+                text = f.read()
```
bryanwweber: I think you can get this to one line 😉 
```suggestion
            text = filename.read_text(encoding=encoding)
```

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1632,9 +1631,19 @@ def convert(filename=None, output_name=None, text=None):
...
 
+    # get file description from header block
+    description = []
+    for line in text.splitlines():
+        # only consider comments in the initial contiguous comment block
+        # comments start with '#'; there may be empty leading lines
+        if line.startswith("#") or not description:
+            description.append(line[1:].rstrip())
+        else:
+            break
```
bryanwweber: What if you have multiple leading blank lines? On line 0, you'd get `description` is empty, so you'd append an empty string to `description`. On line 1, you'd find it doesn't start with `#` and `description` isn't empty, so this would `break`.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1723,7 +1760,34 @@ def main():
...
+    # Do full validation by importing the resulting mechanism
+    if args.no_validate:
+        return
+
+    try:
+        import cantera as ct
+    except ImportError:
+        print("WARNING: Unable to import Cantera Python module. ")
+        print("Output mechanism has not been validated")
+        sys.exit(0)
```
bryanwweber, line 1778: This is perhaps mirroring some other code, but I suspect that `sys.exit(1)` here would be more appropriate? Also, printing to `sys.stderr`?

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1723,7 +1760,34 @@ def main():
...
 
-    convert(input_file, output_file)
+    n_spc, n_rxn, surfaces, output_name = convert(
+        input_file, output_file, encoding=args.input_encoding)
+
+    if not args.quiet:
+        print(f"Wrote YAML mechanism file to '{output_name}'.")
+        print(f"Mechanism contains {n_spc} species and {n_rxn} reactions.")
+
+    # Do full validation by importing the resulting mechanism
```
bryanwweber: This comment seems misplaced, since it's right above the `no_validate` line.

bryanwweber (changes requested):
Thanks @ischoegl A few more changes here. The main ones are the file encoding and multiple blank lines.

If we're deciding not to collect other comment blocks with this PR (which I think is fine) then this shouldn't close the enhancements issue. I think there is still value in attempting to extract those, even if only to try and insert them somewhere in the YAML as a comment, not an actual field. Likewise, I think it's worth trying to do something similar for CTML, which should actually be easier since lxml keeps comment tags.

interfaces/cython/cantera/cti2yaml.py
ischoegl: Other than not using `logger`, this is exactly equivalent to what `ck2yaml` does; as the script could be run without cantera installed on purpose, I think that may have been a deliberate choice. (Aside: `ck2yaml` prints a single line, which is now also adopted.)

interfaces/cython/cantera/cti2yaml.py
ischoegl: Fair enough - this would slip through. Fixed and covered by test.

ischoegl:
@bryanwweber ... thank you for the review! I believe everything is addressed. Regarding the `ImportError`, I just want to keep things consistent with `ck2yaml`.

> If we're deciding not to collect other comment blocks with this PR (which I think is fine) then this shouldn't close the enhancements issue. [...]

I have no issue keeping it open :smiley:

bryanwweber (approved):

PR #1206, opened 2022-02-22
Add dilution options to equivalence ratio functions

g3bk47:
Two changes are proposed in this PR:
1. add a new argument to `equivalence_ratio` to allow users to ignore certain species, e.g. as a result of dilution with an inert species
2. add two new arguments to `set_equivalence_ratio` to dilute a mixture after the mixture composition based on equivalence ratio has been determined.

All changes are backward compatible. Discussion on these changes can be found here:
https://github.com/Cantera/enhancements/issues/108 
https://github.com/Cantera/cantera/pull/521
https://www.google.com/search?q=cantera+users+group&rlz=1C1GCEB_enDE918DE918&oq=canter&aqs=chrome.0.69i59l3j69i57j69i60l3j69i65.6761j0j7&sourceid=chrome&ie=UTF-8

The first change is used like this:
`gas.equivalence_ratio(fuel="H2", oxidizer="O2:0.21,N2:0.79", include_species=["H2","O2"])`
This computes the equivalence ratio as before, but ignores all species except H2 and O2

The second change is used like this:
`gas.set_equivalence_ratio(2.0, "H2", "O2:0.21,N2:0.79", fraction="diluant:0.1", diluant="H2O")`
This first creates a mixture of H2 and air with equivalence ratio 2 and then dilutes it with H2O so that the mole fraction of the diluant in the final mixture is 0.1. Similar options exist to fix the fuel and oxygen fraction in the final mixture.

Comprehensive descriptions have been added to the examples in `interfaces/cython/cantera/examples/thermo/equivalenceRatio.py` to showcase these additions to Cantera.

### DIFF
--- a/interfaces/cython/cantera/examples/thermo/equivalenceRatio.py
+++ b/interfaces/cython/cantera/examples/thermo/equivalenceRatio.py
@@ -9,20 +9,23 @@ import cantera as ct
 
 gas = ct.Solution('gri30.yaml')
 
+# Define the oxidizer composition, here air with 21 mol-% O2 and 79 mol-% N2
+air = "O2:0.21,N2:0.79"
+
 # Set the mixture composition according to the stoichiometric mixture
 # (equivalence ratio phi = 1). The fuel composition in this example
-# is set to 100 mol-% CH4 and the oxidizer to 21 mol-% O2 and
-# 79 mol-% N2.
-gas.set_equivalence_ratio(phi=1.0, fuel="CH4:1", oxidizer="O2:0.21,N2:0.79")
+# is set to 100 mol-% CH4 and the oxidizer to 21 mol-% O2 and 79 mol-% N2.
+# This function changes the composition of the gas object and keeps temperature
+# and pressure constant
+gas.set_equivalence_ratio(phi=1.0, fuel="CH4:1", oxidizer=air)
 
-# The following is a short hand notation and equivalent to the call above
-gas.set_equivalence_ratio(1.0, "CH4", "O2:0.21,N2:0.79")
-
-# By default, the composition of fuel and oxidizer are interpreted as mole
+# if fuel or oxidizer consist of a single species, a short hand notation can be
+# used, for example fuel="CH4" is equivalent to fuel="CH4:1".
+# By default, the compositions of fuel and oxidizer are interpreted as mole
 # fractions. If the compositions are given in mass fractions, an
 # additional argument can be provided. Here, the fuel is 100 mass-% CH4
-# and the oxidizer is 23.3 mass-% O2 and 76.7 mass-% N2.
-gas.set_equivalence_ratio(1.0, "CH4", "O2:0.233,N2:0.767", basis='mass')
+# and the oxidizer is 23.3 mass-% O2 and 76.7 mass-% N2
+gas.set_equivalence_ratio(1.0, fuel="CH4:1", oxidizer="O2:0.233,N2:0.767", basis='mass')
 
 # This function can be used to compute the equivalence ratio for any mixture.
 # The first two arguments specify the compositions of the fuel and oxidizer.
@@ -30,7 +33,7 @@ gas.set_equivalence_ratio(1.0, "CH4", "O2:0.233,N2:0.767", basis='mass')
 # are provided in terms of mass or mole fractions. Default is mole fractions.
 # Note that for all functions shown here, the compositions are normalized
 # internally so the species fractions do not have to sum to unity
-phi = gas.equivalence_ratio(fuel="CH4", oxidizer="O2:233,N2:767", basis='mass')
+phi = gas.equivalence_ratio(fuel="CH4:1", oxidizer="O2:233,N2:767", basis='mass')
 print(f"phi = {phi:1.3f}")
 
 # If the compositions of fuel and oxidizer are unknown, the function can
@@ -45,13 +48,16 @@ print(f"phi = {phi:1.3f}")
 # The mixture fraction is always kg fuel / (kg fuel + kg oxidizer), independent
 # of the basis argument. For example, the mixture fraction Z can be computed as
 # follows. Again, the compositions by default are interpreted as mole fractions
-Z = gas.mixture_fraction(fuel="CH4", oxidizer="O2:0.21,N2:0.79")
+Z = gas.mixture_fraction(fuel="CH4:1", oxidizer=air)
 print(f"Z = {Z:1.3f}")
+
 # By default, the mixture fraction is the Bilger mixture fraction. Instead,
 # a mixture fraction based on a single element can be used. In this example,
 # the following two ways of computing Z are the same:
-Z = gas.mixture_fraction(fuel="CH4", oxidizer="O2:0.21,N2:0.79", element="Bilger")
-Z = gas.mixture_fraction(fuel="CH4", oxidizer="O2:0.21,N2:0.79", element="C")
+Z = gas.mixture_fraction(fuel="CH4:1", oxidizer=air, element="Bilger")
+print(f"Z(Bilger mixture fraction) = {Z:1.3f}")
+Z = gas.mixture_fraction(fuel="CH4:1", oxidizer=air, element="C")
+print(f"Z(mixture fraction based on C) = {Z:1.3f}")
 
 # Since the fuel in this example is pure methane and the oxidizer is air,
 # the mixture fraction is the same as the mass fraction of methane in the mixture
@@ -60,24 +66,24 @@ print(f"mass fraction of CH4 = {gas['CH4'].Y[0]:1.3f}")
 # To set a mixture according to the mixture fraction, the following function
 # can be used. In this example, the final fuel/oxidizer mixture
 # contains 5.5 mass-% CH4:
-Z = gas.set_mixture_fraction(0.055, fuel="CH4", oxidizer="O2:0.21,N2:0.79")
-print(f"mass fraction of CH4 = {gas['CH4'].Y[0]:1.3f}")
+gas.set_mixture_fraction(0.055, fuel="CH4:1", oxidizer=air)
+print(f"Z = {gas['CH4'].Y[0]:1.3f}")
 
 # Mixture fraction and equivalence ratio are invariant to the reaction progress.
 # For example, they stay constant if the mixture composition changes to the burnt
-# state or for any intermediate state
-fuel = "CH4:1"
-oxidizer = "O2:0.21,N2:0.79"
-gas.set_equivalence_ratio(1, fuel, oxidizer)
+# state or for any intermediate state. Fuel and oxidizer composition for all functions
+# shown in this example can be given as string, dictionary or numpy array
+fuel = {"CH4":1} # provide the fuel composition as dictionary instead of string
+gas.set_equivalence_ratio(1, fuel, air)
 gas.equilibrate('HP')
-phi_burnt = gas.equivalence_ratio(fuel, oxidizer)
-Z_burnt = gas.mixture_fraction(fuel, oxidizer)
+phi_burnt = gas.equivalence_ratio(fuel, air)
+Z_burnt = gas.mixture_fraction(fuel, air)
 print(f"phi(burnt) = {phi_burnt:1.3f}")
 print(f"Z(burnt) = {Z_burnt:1.3f}")
 
 # If fuel and oxidizer compositions are specified consistently, then
 # equivalence_ratio and set_equivalence_ratio are consistent as well, as
-# shown in the following example
+# shown in the following example with arbitrary fuel and oxidizer compositions:
 gas.set_equivalence_ratio(2.5, fuel="CH4:1,O2:0.01,CO:0.05,N2:0.1",
                           oxidizer="O2:0.2,N2:0.8,CO2:0.05,CH4:0.01")
 
@@ -98,20 +104,29 @@ print(f"phi = {phi:1.3f}")
 # hydrogen and oxygen and then dilute it with H2O. In this example, the final mixture
 # consists of 30 mol-% H2O and 70 mol-% of the H2/O2 mixture at phi=2
 gas.set_equivalence_ratio(2.0, "H2:1", "O2:1", diluent="H2O", fraction={"diluent":0.3})
-print(f"mole fraction of H2O = {gas['H2O'].X[0]:1.3f}") # final mixture contains 30 mol-% H2O
+print(f"mole fraction of H2O = {gas['H2O'].X[0]:1.3f}") # mixture contains 30 mol-% H2O
 print(f"ratio of H2/O2: {gas['H2'].X[0] / gas['O2'].X[0]:1.3f}") # according to phi=2
 
 # Another option is to specify the fuel or oxidizer fraction in the final mixture.
 # The following example creates a mixture with equivalence ratio 2 from pure
 # hydrogen and oxygen (same as above) and then dilutes it with a mixture of 50 mass-%
-# CO2 and 50 mass-% H2O so that the mass fraction of the fuel in the final mixture is 0.1
+# CO2 and 50 mass-% H2O so that the mass fraction of fuel in the final mixture is 0.1
 gas.set_equivalence_ratio(2.0, "H2", "O2", diluent="CO2:0.5,H2O:0.5",
                           fraction={"fuel":0.1}, basis="mass")
-print(f"mole fraction of H2O = {gas['H2'].Y[0]:1.3f}") # final mixture contains 10 mass-% fuel
+print(f"mole fraction of H2 = {gas['H2'].Y[0]:1.3f}") # mixture contains 10 mass-% fuel
 
 # To compute the equivalence ratio given a diluted mixture, a list of
 # species names can be provided which will be considered for computing phi.
 # In this example, the diluents H2O and CO2 are ignored and only H2 and O2 are
 # considered to get the equivalence ratio
 phi = gas.equivalence_ratio(fuel="H2", oxidizer="O2", include_species=["H2", "O2"])
+print(f"phi = {phi:1.3f}") # prints 2
+
+# If instead the diluent should be included in the computation of the equivalence ratio,
+# the mixture can be set in the following way. Assume the fuel is diluted with
+# 50 mol-% H2O:
+gas.set_equivalence_ratio(2.0, fuel="H2:0.5,H2O:0.5", oxidizer=air)
+
+# This creates a mixture with the specified equivalence ratio including the diluent:
+phi = gas.equivalence_ratio(fuel="H2:0.5,H2O:0.5", oxidizer=air)
 print(f"phi = {phi:1.3f}") # prints 2
--- a/interfaces/cython/cantera/examples/thermo/equivalenceRatio.py
+++ b/interfaces/cython/cantera/examples/thermo/equivalenceRatio.py
@@ -2,52 +2,115 @@
 This example demonstrates how to set a mixture according to equivalence ratio
 and mixture fraction.
 
-Requires: cantera >= 2.5.0
+Requires: cantera >= 2.6.0
 """
 
 import cantera as ct
 
 gas = ct.Solution('gri30.yaml')
 
-# fuel and oxidizer compositions
-fuel = "CH4"
-oxidizer = "O2:0.21,N2:0.79"
+# Set the mixture composition according to the stoichiometric mixture
+# (equivalence ratio phi = 1). The fuel composition in this example
+# is set to 100 mol-% CH4 and the oxidizer to 21 mol-% O2 and
+# 79 mol-% N2.
+gas.set_equivalence_ratio(phi=1.0, fuel="CH4:1", oxidizer="O2:0.21,N2:0.79")
 
-gas.TP = 300, ct.one_atm
+# The following is a short hand notation and equivalent to the call above
+gas.set_equivalence_ratio(1.0, "CH4", "O2:0.21,N2:0.79")
 
-# set the mixture composition according to the stoichiometric mixture
-# (equivalence ratio = 1)
-gas.set_equivalence_ratio(1, fuel, oxidizer)
+# By default, the composition of fuel and oxidizer are interpreted as mole
+# fractions. If the compositions are given in mass fractions, an
+# additional argument can be provided. Here, the fuel is 100 mass-% CH4
+# and the oxidizer is 23.3 mass-% O2 and 76.7 mass-% N2.
+gas.set_equivalence_ratio(1.0, "CH4", "O2:0.233,N2:0.767", basis='mass')
 
 # This function can be used to compute the equivalence ratio for any mixture.
-# An optional argument "basis" indicates if fuel and oxidizer compositions are
-# provided in terms of mass or mole fractions. Default is mole fractions.
-# If fuel and oxidizer are given in mass fractions, use basis='mass'
-phi = gas.equivalence_ratio(fuel, oxidizer)
+# The first two arguments specify the compositions of the fuel and oxidizer.
+# An optional third argument "basis" indicates if fuel and oxidizer compositions
+# are provided in terms of mass or mole fractions. Default is mole fractions.
+# Note that for all functions shown here, the compositions are normalized
+# internally so the species fractions do not have to sum to unity
+phi = gas.equivalence_ratio(fuel="CH4", oxidizer="O2:233,N2:767", basis='mass')
 print("phi = {:1.3f}".format(phi))
 
-# The equivalence ratio can also be computed from the elemental composition
-# assuming that there is no oxygen in the fuel and no C,H and S elements
-# in the oxidizer so that the composition of fuel and oxidizer can be omitted
+# If the compositions of fuel and oxidizer are unknown, the function can
+# be called without arguments. This assumes that all C, H and S atoms come from
+# the fuel and all O atoms from the oxidizer. In this example, the fuel was set
+# to be pre CH4 and the oxidizer O2:0.233,N2:0.767 so that the assumption is true
+# and the same equivalence ratio as above is computed
 phi = gas.equivalence_ratio()
-
-# In this example, the result is the same as above
 print("phi = {:1.3f}".format(phi))
 
-# the mixture fraction Z can be computed as follows:
-Z = gas.mixture_fraction(fuel, oxidizer)
+# Instead of working with equivalence ratio, mixture fraction can be used.
+# The mixture fraction is always kg fuel / (kg fuel + kg oxidizer), independent
+# of the basis argument. For example, the mixture fraction Z can be computed as
+# follows. Again, the compositions by default are interpreted as mole fractions
+Z = gas.mixture_fraction(fuel="CH4", oxidizer="O2:0.21,N2:0.79")
 print("Z = {:1.3f}".format(Z))
+# By default, the mixture fraction is the Bilger mixture fraction. Instead,
+# a mixture fraction based on a single element can be used. In this example,
+# the following two ways of computing Z are the same:
+Z = gas.mixture_fraction(fuel="CH4", oxidizer="O2:0.21,N2:0.79", element="Bilger")
+Z = gas.mixture_fraction(fuel="CH4", oxidizer="O2:0.21,N2:0.79", element="C")
 
-# The mixture fraction is kg fuel / (kg fuel + kg oxidizer). Since the fuel in
-# this example is pure methane and the oxidizer is air, the mixture fraction
-# is the same as the mass fraction of methane in the mixture
+# Since the fuel in this example is pure methane and the oxidizer is air,
+# the mixture fraction is the same as the mass fraction of methane in the mixture
+print("mass fraction of CH4 = {:1.3f}".format(gas["CH4"].Y[0]))
+
+# To set a mixture according to the mixture fraction, the following function
+# can be used. In this example, the final fuel/oxidizer mixture
+# contains 5.5 mass-% CH4:
+Z = gas.set_mixture_fraction(0.055, fuel="CH4", oxidizer="O2:0.21,N2:0.79")
 print("mass fraction of CH4 = {:1.3f}".format(gas["CH4"].Y[0]))
 
 # Mixture fraction and equivalence ratio are invariant to the reaction progress.
 # For example, they stay constant if the mixture composition changes to the burnt
-# state
+# state or for any intermediate state
+fuel = "CH4:1"
+oxidizer = "O2:0.21,N2:0.79"
+gas.set_equivalence_ratio(1, fuel, oxidizer)
 gas.equilibrate('HP')
 phi_burnt = gas.equivalence_ratio(fuel, oxidizer)
 Z_burnt = gas.mixture_fraction(fuel, oxidizer)
 print("phi(burnt) = {:1.3f}".format(phi_burnt))
 print("Z(burnt) = {:1.3f}".format(Z_burnt))
+
+# If fuel and oxidizer compositions are specified consistently, then
+# equivalence_ratio and set_equivalence_ratio are consistent as well, as
+# shown in the following example
+gas.set_equivalence_ratio(2.5,fuel="CH4:1,O2:0.01,CO:0.05,N2:0.1",
+        oxidizer="O2:0.2,N2:0.8,CO2:0.05,CH4:0.01")
+
+phi = gas.equivalence_ratio(fuel="CH4:1,O2:0.01,CO:0.05,N2:0.1",
+        oxidizer="O2:0.2,N2:0.8,CO2:0.05,CH4:0.01")
+print("phi = {:1.3f}".format(phi)) # prints 2.5
+
+# Without specifying the fuel and oxidizer compositions, it is assumed that
+# all C, H and S atoms come from the fuel and all O atoms from the oxidizer,
+# which is not true for this example. Therefore, the following call gives a
+# different equivalence ratio based on that assumption
+phi = gas.equivalence_ratio()
+print("phi = {:1.3f}".format(phi))
+
+# After computing the mixture composition for a certain equivalence ratio given
+# a fuel and mixture composition, the mixture can optionally be diluted. The
+# following function will first create a mixture with equivalence ratio 2 from pure
+# hydrogen and oxygen and then dilute it with H2O. In this example, the final mixture
+# consists of 30 mol-% H2O and 70 mol-% of the H2/O2 mixture at phi=2
+gas.set_equivalence_ratio(2.0, "H2:1", "O2:1", fraction={"diluent":0.3}, diluent="H2O")
+print("mole fraction of H2O = {:1.3f}".format(gas["H2O"].X[0])) # final mixture contains 30 mol-% H2O
+print("ratio of H2/O2: {:1.3f}".format(gas["H2"].X[0]/gas["O2"].X[0])) # according to phi=2
+
+# Another option is to specify the fuel or oxidizer fraction in the final mixture.
+# The following example creates a mixture with equivalence ratio 2 from pure
+# hydrogen and oxygen (same as above) and then dilutes it with a mixture of 50 mass-%
+# CO2 and 50 mass-% H2O so that the mass fraction of the fuel in the final mixture is 0.1
+gas.set_equivalence_ratio(2.0, "H2", "O2", fraction={"fuel":0.1}, diluent="CO2:0.5,H2O:0.5", basis="mass")
+print("mole fraction of H2O = {:1.3f}".format(gas["H2"].Y[0])) # final mixture contains 10 mass-% fuel
+
+# To compute the equivalence ratio given a diluted mixture, a list of
+# species names can be provided which will be considered for computing phi.
+# In this example, the diluents H2O and CO2 are ignored and only H2 and O2 are
+# considered to get the equivalence ratio
+phi = gas.equivalence_ratio(fuel="H2", oxidizer="O2", include_species=["H2","O2"])
+print("phi = {:1.3f}".format(phi)) # prints 2
--- a/interfaces/cython/cantera/examples/thermo/equivalenceRatio.py
+++ b/interfaces/cython/cantera/examples/thermo/equivalenceRatio.py
@@ -19,7 +19,7 @@ air = "O2:0.21,N2:0.79"
 # and pressure constant
 gas.set_equivalence_ratio(phi=1.0, fuel="CH4:1", oxidizer=air)
 
-# if fuel or oxidizer consist of a single species, a short hand notation can be
+# If fuel or oxidizer consist of a single species, a short hand notation can be
 # used, for example fuel="CH4" is equivalent to fuel="CH4:1".
 # By default, the compositions of fuel and oxidizer are interpreted as mole
 # fractions. If the compositions are given in mass fractions, an
@@ -67,11 +67,11 @@ print(f"mass fraction of CH4 = {gas['CH4'].Y[0]:1.3f}")
 # can be used. In this example, the final fuel/oxidizer mixture
 # contains 5.5 mass-% CH4:
 gas.set_mixture_fraction(0.055, fuel="CH4:1", oxidizer=air)
-print(f"Z = {gas['CH4'].Y[0]:1.3f}")
+print(f"mass fraction of CH4 = {gas['CH4'].Y[0]:1.3f}")
 
 # Mixture fraction and equivalence ratio are invariant to the reaction progress.
 # For example, they stay constant if the mixture composition changes to the burnt
-# state or for any intermediate state. Fuel and oxidizer composition for all functions
+# state or for any intermediate state. Fuel and oxidizer compositions for all functions
 # shown in this example can be given as string, dictionary or numpy array
 fuel = {"CH4":1} # provide the fuel composition as dictionary instead of string
 gas.set_equivalence_ratio(1, fuel, air)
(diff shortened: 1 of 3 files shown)


interfaces/cython/cantera/thermo.pyx
```diff
@@ -844,6 +860,85 @@ cdef class ThermoPhase(_SolutionBase):
...
+        if fraction is not None and diluent is not None:
+
+            # parse the fraction argument
+            if isinstance(fraction,str):
+                if ':' not in fraction:
+                    raise ValueError("The fraction argument requires a value in the"
+                        " form of e.g. fraction='fuel:0.7'")
+                colon_pos = fraction.rfind(":")
+                fraction_type  = fraction[:colon_pos]
+                fraction_value = float(fraction[colon_pos+1:])
```
ischoegl: You should be able to use Cantera's `parseCompString` (see `include/cantera/base/stringutils.h` which is C++ exposed to Cython, but not to to Python) in combination with `comp_map_to_dict` here (a helper function within Cython).

Also, there should be a ` ` before the comma in `isinstance(fraction, str)`

interfaces/cython/cantera/thermo.pyx
```diff
@@ -809,15 +809,19 @@ cdef class ThermoPhase(_SolutionBase):
         self.state = original_state
         return arr
 
-    def set_equivalence_ratio(self, phi, fuel, oxidizer, basis='mole'):
+    def set_equivalence_ratio(self, phi, fuel, oxidizer, basis='mole', fraction=None, diluent=None):
```
ischoegl: I believe `diluent` should be listed after `oxidizer` and `basis` should remain the last parameter. That way, the interface should be a little more intuitive).

As an aside: Cantera is transitioning to consistently using `"` to identify strings, where single ticks `'` are replaced whenever a line is changed. I.e. `basis='mole'` would be changed to `basis="mole"` in this line. This only affects lines that are changed.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -844,6 +860,85 @@ cdef class ThermoPhase(_SolutionBase):
...
+                    Z = 1.0-Z_fuel_mole
+                if fraction_value > Z:
+                    raise ValueError("fuel or oxidizer fraction after dilution cannot be higher"
+                            " than fuel fraction in the original mixture.")
+                self.X = diluent
+                factor = fraction_value / Z
+                self.X = factor*X_mix + (1.0-factor)*self.X
+
+        if (fraction is None and diluent is not None) or (fraction is not None and diluent is None):
+                raise ValueError("if dilution is used, both fraction and diluent parameters are required.")
```
ischoegl: Simpler as
```Python
if (fraction is None) != (diluent is None):
```
Also, error message should start with uppercase letter (here and elsewhere). Finally, for code readability, it would also make sense to move this check up above the previous `if` clause.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -844,6 +860,85 @@ cdef class ThermoPhase(_SolutionBase):
...
+                fraction_type  = fraction[:colon_pos]
+                fraction_value = float(fraction[colon_pos+1:])
+            elif isinstance(fraction,dict):
+                fraction_type  = list(fraction.keys())[0]
+                fraction_value = float(list(fraction.values())[0])
+            else:
+                raise ValueError("fraction argument must be given as string or dictionary.")
+
+            if isinstance(diluent, str) and ':' not in diluent:
+                diluent += ':1'
```
ischoegl: I believe you could use `ThermoPhase.__composition_to_array` here, which should simplify some code further down.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -844,6 +860,85 @@ cdef class ThermoPhase(_SolutionBase):
...
+            if isinstance(oxidizer, str) and ':' not in oxidizer:
+                oxidizer += ':1'
+
+            # if 'fraction' is specified for diluent, just scale the mass or mole fractions
+            # of the fuel/oxidizer mixture accordingly
+            if fraction_type == 'diluent':
+                if basis == 'mole':
+                    X_fuelox = self.X
+                    self.X = diluent
+                    self.X = (1.0-fraction_value)*X_fuelox + fraction_value*self.X
```
ischoegl: Cantera uses spaces left and right of most operators, as in
```python
self.X = (1.0 - fraction_value) * X_fuelox + fraction_value * self.X
```

interfaces/cython/cantera/thermo.pyx
```diff
@@ -844,6 +860,85 @@ cdef class ThermoPhase(_SolutionBase):
...
+                else:
+                    Y_fuelox = self.Y
+                    self.Y = diluent
+                    self.Y = (1.0-fraction_value)*Y_fuelox + fraction_value*self.Y
+                return
+
+            if not fraction_type == 'fuel' and not fraction_type == 'oxidizer':
+                raise ValueError(fraction_type+" is unknown. Fraction must specify"
+                        " 'diluent', 'fuel' or 'oxidizer'.")
```
ischoegl: Check should be moved to wherever `fraction_type` is initially determined. Using the optional input of `parseCompString` you can limit `names` to `["fuel", "oxidizer", "diluent"]` and then ensure that only one of them is non-zero.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -836,6 +840,18 @@ cdef class ThermoPhase(_SolutionBase):
...
             fractions (``basis='mole'``) or mass fractions (``basis='mass'``)
+        :param: fraction:
+            Dilute the mixture with equivalence ratio ``phi`` according to
+            ``fraction``. Fraction can refer to the fraction of diluent in the
+            final mixture (e.g. ``fraction="diluent:0.7``), the fraction of fuel
+            in the final mixture (e.g. ``fraction="fuel:0.7") or fraction of
+            oxidizer in the final mixture (e.g. ``fraction="oxidizer:0.7"). The
+            fraction itself is interpreted as mole or mass fraction based on ``basis``.
+            Default is no dilution or ``fraction=None``. May be given as string
+            or dictionary (e.g. ``fraction={"fuel":0.7})``
```
ischoegl: Should mention that this is only applicable (*and* required) when dilution is used. Also 'final' mixture should be 'reactant' mixture.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -844,6 +860,85 @@ cdef class ThermoPhase(_SolutionBase):
...
+
+            if not fraction_type == 'fuel' and not fraction_type == 'oxidizer':
+                raise ValueError(fraction_type+" is unknown. Fraction must specify"
+                        " 'diluent', 'fuel' or 'oxidizer'.")
+
+            # get the mixture fraction before scaling / diluent addition
+            Z_fuel = self.mixture_fraction(fuel,oxidizer,basis)
+
+            if (Z_fuel == 0.0 and fraction_type == "fuel") or (Z_fuel == 1.0 and fraction_type == "oxidzer"):
+                raise ValueError("Invalid fuel/oxidizer mixture")
```
ischoegl: Separating the error clauses would give more valuable feedback to the user.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -844,6 +860,85 @@ cdef class ThermoPhase(_SolutionBase):
...
+            if isinstance(diluent, str) and ':' not in diluent:
+                diluent += ':1'
+            if isinstance(fuel, str) and ':' not in fuel:
+                fuel += ':1'
+            if isinstance(oxidizer, str) and ':' not in oxidizer:
+                oxidizer += ':1'
+
+            # if 'fraction' is specified for diluent, just scale the mass or mole fractions
+            # of the fuel/oxidizer mixture accordingly
+            if fraction_type == 'diluent':
```
ischoegl: All single quotes `'` should be replaced by `"` in new code.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -887,10 +982,14 @@ cdef class ThermoPhase(_SolutionBase):
         equivalence ratio is computed from the available oxygen and the
         required oxygen for complete oxidation. The ``basis`` determines the
         composition of fuel and oxidizer: ``basis='mole'`` (default) means mole
-        fractions, ``basis='mass'`` means mass fractions::
-
-            >>> gas.set_equivalence_ratio(0.5, 'CH3:0.5, CH3OH:.5, N2:0.125', 'O2:0.21, N2:0.79, NO:0.01')
-            >>> gas.equivalence_ratio('CH3:0.5, CH3OH:.5, N2:0.125', 'O2:0.21, N2:0.79, NO:0.01')
+        fractions, ``basis='mass'`` means mass fractions. Additionally, a
+        list of species cnd be provided with ``include_species``. This means that
```
ischoegl: 'can'

interfaces/cython/cantera/thermo.pyx
```diff
@@ -900,17 +999,35 @@ cdef class ThermoPhase(_SolutionBase):
         :param basis:
             Determines if ``fuel`` and ``oxidizer`` are given in mole fractions
             (``basis='mole'``) or mass fractions (``basis='mass'``)
-        """
+        :param include_species:
+            List of species names. Only these species are considered for the
+            computation of the equivalence ratio. The default ``None`` means all
+            species are considered
```
ischoegl: "List of species names (optional)." and 
"By default, all species are considered." avoids a technical detail of the implementation.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -422,6 +422,134 @@ def test_equil_results(gas, fuel, ox, Y_Cf, Y_Of, Y_Co, Y_Oo, basis):
...
+        X_diluent = gas.X
+        gas.Y = diluent
+        Y_diluent = gas.Y
+
+        gas.set_equivalence_ratio(2, fuel, oxidizer)
+        X_Mix = gas.X
+        gas.set_equivalence_ratio(2, fuel,oxidizer, fraction="diluent:0.6", diluent=diluent)
+        X_expected = X_Mix*0.4 + 0.6*X_diluent
+        for k in range(gas.n_species):
+            self.assertNear(gas.X[k], X_expected[k])
```
ischoegl: You can use Cantera's custom
```
self.assertArrayNear(gas.X, X_Mix * 0.4 + 0.6 * X_diluent)
```
here (and elsewhere).

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -422,6 +422,134 @@ def test_equil_results(gas, fuel, ox, Y_Cf, Y_Of, Y_Co, Y_Oo, basis):
...
+        test_simple_dilution({"fuel":0.1}, "mole")
+        test_simple_dilution("fuel:0.1",   "mass")
+        test_simple_dilution({"fuel":0.1}, "mass")
+
+        test_simple_dilution("oxidizer:0.1",    "mole")
+        test_simple_dilution({"oxidizer":0.1}, "mole")
+        test_simple_dilution("oxidizer:0.1",   "mass")
+        test_simple_dilution({"oxidizer":0.1}, "mass")
+
+
```
ischoegl, line 505: Extraneous empty line (only one line should separate tests).

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -422,6 +422,134 @@ def test_equil_results(gas, fuel, ox, Y_Cf, Y_Of, Y_Co, Y_Oo, basis):
...
+        gas.X = fuel
+        M_fuel = gas.mean_molecular_weight
+        gas.X = oxidizer
+        M_oxidizer = gas.mean_molecular_weight
+
+        gas.set_equivalence_ratio(0.8, fuel, oxidizer)
+        AFR = M_fuel/M_oxidizer*gas.stoich_air_fuel_ratio(fuel,oxidizer)/0.8
+
+        gas.set_equivalence_ratio(0.8, fuel, oxidizer, fraction="fuel:0.05", diluent=diluent)
+        X_expected = 0.05*( (X_fuel + X_oxidizer*AFR)) + X_diluent*(1-(0.05+0.05*AFR))
```
ischoegl: Watch out for spacing and superfluous parentheses (here and elsewhere).

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -422,6 +422,134 @@ def test_equil_results(gas, fuel, ox, Y_Cf, Y_Of, Y_Co, Y_Oo, basis):
...
+            if isinstance(fraction,str):
+                fractiontype = fraction[:fraction.find(":")]
+            else:
+                fractiontype = list(fraction.keys())[0]
+
+            M_H2 = gas.molecular_weights[gas.species_index("H2")]
+            M_O2 = gas.molecular_weights[gas.species_index("O2")]
+
+            gas.set_equivalence_ratio(2, "H2","O2", fraction=fraction, diluent="CO2", basis=basis)
+            if basis=="mole" and fractiontype=="diluent":
```
ischoegl: Cantera uses spaces left and right of comparisons also, i.e.
```python
if basis == "mole" and fractiontype == "diluent":
```
You could also use the more pythonic `fraction_type`.

interfaces/cython/cantera/examples/thermo/equivalenceRatio.py
```diff
@@ -2,52 +2,115 @@
...
+print("phi = {:1.3f}".format(phi))
+
+# After computing the mixture composition for a certain equivalence ratio given
+# a fuel and mixture composition, the mixture can optionally be diluted. The
+# following function will first create a mixture with equivalence ratio 2 from pure
+# hydrogen and oxygen and then dilute it with H2O. In this example, the final mixture
+# consists of 30 mol-% H2O and 70 mol-% of the H2/O2 mixture at phi=2
+gas.set_equivalence_ratio(2.0, "H2:1", "O2:1", fraction={"diluent":0.3}, diluent="H2O")
+print("mole fraction of H2O = {:1.3f}".format(gas["H2O"].X[0])) # final mixture contains 30 mol-% H2O
+print("ratio of H2/O2: {:1.3f}".format(gas["H2"].X[0]/gas["O2"].X[0])) # according to phi=2
```
ischoegl: With our most recent Python requirements, you can start using f-strings (which are simpler), e.g.
```python
print(f"mole fraction of H2O = {gas['H2O'].X[0]:1.3f}")
```

interfaces/cython/cantera/thermo.pyx
```diff
@@ -809,15 +809,19 @@ cdef class ThermoPhase(_SolutionBase):
...
         specified equivalence ratio ``phi``, holding temperature and pressure
         constant. Considers the oxidation of C to CO2, H to H2O and S to SO2.
         Other elements are assumed not to participate in oxidation (that is,
         N ends up as N2). The ``basis`` determines the fuel and oxidizer
         compositions: ``basis='mole'`` means mole fractions (default),
-        ``basis='mass'`` means mass fractions::
+        ``basis='mass'`` means mass fractions. After a mixture given a
+        ``phi`` from ``fuel`` and ``oxidizer`` is  created, it can further
+        be diluted by a ``diluent`` based on a mixing ``fraction``. See also
+        interfaces/cython/cantera/examples/thermo/equivalenceRatio.py for more
```
ischoegl: I am not certain about the most appropriate link here, but believe that this should point directly to the [online example](https://cantera.org/examples/python/thermo/equivalenceRatio.py.html) using Sphinx notation (@bryanwweber may have some thoughts). I think it's ok to link to the permanent location as this hopefully will make it into the 2.6 release.

ischoegl (changes requested):
Hi @g3bk47 ... thank you for the PR! I believe this is a nice addition that addresses Cantera/enhancements#108 as well as some recent discussions on the user forum.

While there may be another round of suggestions, I mainly focused on the code for the time being. Cantera has a couple of built-in capabilities that are not obvious, but should help simplify some of what you implemented. Also, any new code addition should follow some style conventions more closely: e.g. empty spaces left and right of comparisons and most operators, double quotes for Python strings (whenever possible), etc.

I appreciate that you already added comprehensive unit tests, and updated the example. I honestly haven't checked the actual implementation yet, but it looks like you have most of the bases covered ...

interfaces/cython/cantera/thermo.pyx
bryanwweber: Changing the order of the arguments would possibly break existing code, so I don't think we should do that. I'd be in favor of making everything after `basis` into a keyword-only argument, I think if people are listing that many arguments they should be using keywords anyways.

ischoegl: Good point. I'd still advocate for reversing the order, i.e. `diluent` before `fraction`.

g3bk47: My reasoning for the current argument order was to mirror the existing structure (first how to mix and then what to mix). At the moment, it is first phi, i.e. a type of mixing ratio, and then the reactant compositions. This is now followed by the dilution fraction and then by the diluent composition. But I don't feel strongly either way.

ischoegl: I still believe it makes sense to have `fraction` last as it can reference fuel, oxidizer and diluent. Regarding @bryanwweber’s comment, I believe it makes sense to force keyword arguments. See explanation here: [vegardstikbakke.com/python-keyword-only/](https://www.vegardstikbakke.com/python-keyword-only/)

interfaces/cython/cantera/thermo.pyx
g3bk47: I think the second argument of parseCompString is not exposed to Cython https://github.com/Cantera/cantera/blob/36717bd1de73cd514904afe0db6699a9d350a0d1/interfaces/cython/cantera/_cantera.pxd#L166 So I wrote the check of valid input names by hand for now.

interfaces/cython/cantera/examples/thermo/equivalenceRatio.py
```diff
@@ -2,52 +2,116 @@
...
+phi = gas.equivalence_ratio()
+print(f"phi = {phi:1.3f}")
+
+# After computing the mixture composition for a certain equivalence ratio given
+# a fuel and mixture composition, the mixture can optionally be diluted. The
+# following function will first create a mixture with equivalence ratio 2 from pure
+# hydrogen and oxygen and then dilute it with H2O. In this example, the final mixture
+# consists of 30 mol-% H2O and 70 mol-% of the H2/O2 mixture at phi=2
+gas.set_equivalence_ratio(2.0, "H2:1", "O2:1", diluent="H2O", fraction={"diluent":0.3})
+print(f"mole fraction of H2O = {gas['H2O'].X[0]:1.3f}") # final mixture contains 30 mol-% H2O
```
ischoegl: Minor formatting: Cantera limits line lengths to 88 characters (similar below)

interfaces/cython/cantera/thermo.pyx
```diff
@@ -809,15 +809,19 @@ cdef class ThermoPhase(_SolutionBase):
         self.state = original_state
         return arr
 
-    def set_equivalence_ratio(self, phi, fuel, oxidizer, basis='mole'):
+    def set_equivalence_ratio(self, phi, fuel, oxidizer, basis="mole", *, diluent=None, fraction=None):
```
ischoegl: Limit line length to 88 characters.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -836,13 +840,107 @@ cdef class ThermoPhase(_SolutionBase):
...
+            can refer to the fraction of diluent in the reactant mixture
+            (e.g. ``fraction="diluent:0.7``), the fraction of fuel
+            in the reactant mixture (e.g. ``fraction="fuel:0.7") or fraction of
+            oxidizer in the reactant mixture (e.g. ``fraction="oxidizer:0.7"). The
+            fraction itself is interpreted as mole or mass fraction based on ``basis``.
+            Default is no dilution or ``fraction=None``. May be given as string
+            or dictionary (e.g. ``fraction={"fuel":0.7})``
+        :param: diluent:
+            Optional parameter. Required if dilution is used. Specifies the composition
+            of the diluent in mole/mass fractions as a string, array or dict
```
ischoegl, line 845: This should be moved above `fraction` (also 'Required if dilution is used.' can be skipped in this case).

interfaces/cython/cantera/thermo.pyx
```diff
@@ -809,15 +809,19 @@ cdef class ThermoPhase(_SolutionBase):
...
         Set the composition to a mixture of ``fuel`` and ``oxidizer`` at the
         specified equivalence ratio ``phi``, holding temperature and pressure
         constant. Considers the oxidation of C to CO2, H to H2O and S to SO2.
         Other elements are assumed not to participate in oxidation (that is,
         N ends up as N2). The ``basis`` determines the fuel and oxidizer
         compositions: ``basis='mole'`` means mole fractions (default),
-        ``basis='mass'`` means mass fractions::
+        ``basis='mass'`` means mass fractions. After a mixture given a
+        ``phi`` from ``fuel`` and ``oxidizer`` is  created, it can further
+        be diluted by a ``diluent`` based on a mixing ``fraction``. `See also
```
ischoegl: Isn't it that either fuel, oxidizer, or the mixture can be diluted?

interfaces/cython/cantera/thermo.pyx
```diff
@@ -809,15 +809,19 @@ cdef class ThermoPhase(_SolutionBase):
...
+        here for more examples
```
ischoegl: simpler as: "For more information, see `Python example <url>`_," Similar for the URL provided in `equivalence_ratio`.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -836,13 +840,107 @@ cdef class ThermoPhase(_SolutionBase):
...
-        cdef np.ndarray[np.double_t, ndim=1] o = \
+        cdef np.ndarray[np.double_t, ndim=1] ox_comp = \
                 np.ascontiguousarray(self.__composition_to_array(oxidizer, basis), dtype=np.double)
 
-        self.thermo.setEquivalenceRatio(phi, &f[0], &o[0], ThermoBasis.mass if basis == 'mass' else ThermoBasis.molar)
+        self.thermo.setEquivalenceRatio(phi, &fuel_comp[0], &ox_comp[0],
+                                        ThermoBasis.mass if basis == "mass" else ThermoBasis.molar)
+
+        if (fraction is None) != (diluent is None):
+            raise ValueError("If dilution is used, both 'fraction' and 'diluent' parameters are required.")
```
ischoegl: Use strategic line breaks to stay within 88 character limits (similar in passage above)

interfaces/cython/cantera/thermo.pyx
```diff
@@ -836,13 +840,107 @@ cdef class ThermoPhase(_SolutionBase):
...
+
+        if isinstance(fraction, str):
+            fraction_dict = comp_map_to_dict(parseCompString(stringify(fraction)))
+        elif isinstance(fraction, dict):
+            fraction_dict = fraction
+        else:
+            raise ValueError("fraction argument must be given as string or dictionary.")
+
+        if len(fraction_dict) != 1:
+            raise ValueError("Invalid format for the fraction. Must be provided e.g. as "
```
ischoegl: Avoid 'e.g.', as writing 'example' is easier to follow (Cantera has recently replaced most instances of 'i.e.' and 'e.g.' in the documentation)

interfaces/cython/cantera/thermo.pyx
```diff
@@ -836,13 +840,107 @@ cdef class ThermoPhase(_SolutionBase):
...
+                             "fraction='fuel:0.1'")
+
+        fraction_type  = list(fraction_dict.keys())[0]
+        fraction_value = float(list(fraction_dict.values())[0])
+
+        if fraction_type not in ["fuel", "oxidizer", "diluent"]:
+            raise ValueError("fraction must specify 'fuel', 'oxidizer' or 'diluent'")
```
ischoegl: Uppercase initial letter (there are some other instances); rephrase if you want to keep it consistent with the lower case argument `'fraction'`.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -836,13 +840,107 @@ cdef class ThermoPhase(_SolutionBase):
...
+        if Z_fuel == 1.0 and fraction_type == "oxidzer":
+            raise ValueError("No oxidizer in the fuel/oxidizer mixture")
+
+        if basis == "mass": # for mass basis, it is straight forward
+            if fraction_type == "fuel":
+                Z = Z_fuel
+            else:  # oxidizer
+                Z = 1.0 - Z_fuel
+            if fraction_value > Z:
+                raise ValueError(f"{fraction_type} fraction after dilution cannot "
```
ischoegl: Use `fraction_type.capitalize()` at beginning of sentence. Same further below

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -422,6 +422,125 @@ def test_equil_results(gas, fuel, ox, Y_Cf, Y_Of, Y_Co, Y_Oo, basis):
...
+        gas.set_equivalence_ratio(2, fuel, oxidizer, basis="mass")
+        Y_Mix = gas.Y
+        gas.set_equivalence_ratio(2, fuel, oxidizer, fraction="diluent:0.6", diluent=diluent, basis="mass")
+        Y_expected = Y_Mix * 0.4 + 0.6 * Y_diluent
+        self.assertArrayNear(gas.Y, Y_expected)
+
+        gas.set_equivalence_ratio(0.8, fuel, oxidizer, basis="mass")
+        AFR = gas.stoich_air_fuel_ratio(fuel, oxidizer, basis="mass") / 0.8
+        gas.set_equivalence_ratio(0.8, fuel, oxidizer, fraction="fuel:0.05", diluent=diluent, basis="mass")
+        Y_expected = 0.05 * (Y_fuel + Y_oxidizer*AFR) + Y_diluent * (1 - (0.05 + 0.05 * AFR))
```
ischoegl:
```suggestion
        Y_expected = 0.05 * (Y_fuel + AFR * Y_oxidizer) + (1 - 0.05 * (1 + AFR)) * Y_diluent
```
is a minor simplification (same for `X` version)

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -422,6 +422,125 @@ def test_equil_results(gas, fuel, ox, Y_Cf, Y_Of, Y_Co, Y_Oo, basis):
...
+        self.assertArrayNear(gas.Y, Y_expected)
+
+        gas.set_equivalence_ratio(0.8, fuel, oxidizer, fraction="oxidizer:0.05", diluent=diluent, basis="mass")
+        Y_expected = 0.05 / AFR * (Y_fuel + Y_oxidizer * AFR) + Y_diluent * (1 - 0.05 / AFR * (1 + AFR))
```
ischoegl:
```suggestion
        Y_expected = 0.05 * (Y_fuel / AFR + Y_oxidizer) + (1 - 0.05 * (1 + 1 / AFR)) * Y_diluent
```
is a minor simplification (same for `X` version)

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -422,6 +422,125 @@ def test_equil_results(gas, fuel, ox, Y_Cf, Y_Of, Y_Co, Y_Oo, basis):
...
+                self.assertNear(gas["H2"].Y[0] / gas["O2"].Y[0], 4 * M_H2 / M_O2)
+
+            state = gas.state
+            self.assertNear(gas.equivalence_ratio("H2", "O2", include_species=["H2", "O2"], basis=basis), 2)
+            self.assertNear(gas.equivalence_ratio(include_species=["H2", "O2"], basis=basis), 2)
+            self.assertArrayNear(state, gas.state)
+
+        # brute force all possible input combinations
+        test_simple_dilution("diluent:0.3",   "mole")
+        test_simple_dilution({"diluent":0.3}, "mole")
```
ischoegl:
```suggestion
        test_simple_dilution("diluent:0.3", "mole")
        test_simple_dilution({"diluent": 0.3}, "mole")
```
(here and elsewhere)

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -422,6 +422,125 @@ def test_equil_results(gas, fuel, ox, Y_Cf, Y_Of, Y_Co, Y_Oo, basis):
...
+    def test_equivalence_ratio_simple_dilution(self):
+        gas = ct.Solution("gri30.yaml")
+
+        gas.X = "H2:4,O2:1,CO:3,CO2:4,N2:5,CH4:6"
+        state = gas.state
+        self.assertNear(gas.equivalence_ratio(include_species=["H2", "O2"]), 2)
+        self.assertNear(gas.equivalence_ratio("H2", "O2", include_species=["H2", "O2"]), 2)
+        self.assertArrayNear(state, gas.state)
+
+        def test_simple_dilution(fraction,basis):
```
ischoegl:
```suggestion
        def test_simple_dilution(fraction, basis):
```

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -422,6 +422,125 @@ def test_equil_results(gas, fuel, ox, Y_Cf, Y_Of, Y_Co, Y_Oo, basis):
...
+                fraction_type = fraction[:fraction.find(":")]
+            else:
+                fraction_type = list(fraction.keys())[0]
+
+            M_H2 = gas.molecular_weights[gas.species_index("H2")]
+            M_O2 = gas.molecular_weights[gas.species_index("O2")]
+
+            gas.set_equivalence_ratio(2, "H2", "O2", fraction=fraction, diluent="CO2", basis=basis)
+            if basis == "mole" and fraction_type == "diluent":
+                self.assertNear(gas["H2"].X[0], (1 - 0.3) * 0.8)
```
ischoegl: I don't think that the value `0.3` should be hard-coded here. Instead, strip it off the fraction string.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -422,6 +422,125 @@ def test_equil_results(gas, fuel, ox, Y_Cf, Y_Of, Y_Co, Y_Oo, basis):
 
         test_equil_results(gas, fuel, ox, Y_Cf, Y_Of, Y_Co, Y_Oo, 'mass')
 
+    def test_equivalence_ratio_simple_dilution(self):
+        gas = ct.Solution("gri30.yaml")
+
+        gas.X = "H2:4,O2:1,CO:3,CO2:4,N2:5,CH4:6"
```
ischoegl: You could be more explicit about setting the equivalence ratio here. Also, avoid hard-coding '2' in all the calculations. Instead, define `phi = 2` and use `phi` elsewhere, as this will provide for more readable code.

interfaces/cython/cantera/examples/thermo/equivalenceRatio.py
```diff
@@ -2,52 +2,116 @@
...
-oxidizer = "O2:0.21,N2:0.79"
+# Set the mixture composition according to the stoichiometric mixture
+# (equivalence ratio phi = 1). The fuel composition in this example
+# is set to 100 mol-% CH4 and the oxidizer to 21 mol-% O2 and
+# 79 mol-% N2.
+gas.set_equivalence_ratio(phi=1.0, fuel="CH4:1", oxidizer="O2:0.21,N2:0.79")
 
-gas.TP = 300, ct.one_atm
+# The following is a short hand notation and equivalent to the call above
+gas.set_equivalence_ratio(1.0, "CH4", "O2:0.21,N2:0.79")
```
ischoegl: I believe just using the 'long' notation (i.e. keyword arguments) throughout this example will be easier to follow (as it is self-documenting). It still makes sense to document that fuel can be `CH4` or `CH4:1`

interfaces/cython/cantera/examples/thermo/equivalenceRatio.py
```diff
@@ -2,52 +2,116 @@
...
 
-# set the mixture composition according to the stoichiometric mixture
-# (equivalence ratio = 1)
-gas.set_equivalence_ratio(1, fuel, oxidizer)
+# By default, the composition of fuel and oxidizer are interpreted as mole
```
ischoegl:
```suggestion
# By default, the compositions of fuel and oxidizer are interpreted as mole
```

interfaces/cython/cantera/examples/thermo/equivalenceRatio.py
```diff
@@ -2,52 +2,116 @@
...
+print(f"phi = {phi:1.3f}")
 
-# In this example, the result is the same as above
-print("phi = {:1.3f}".format(phi))
+# Instead of working with equivalence ratio, mixture fraction can be used.
+# The mixture fraction is always kg fuel / (kg fuel + kg oxidizer), independent
+# of the basis argument. For example, the mixture fraction Z can be computed as
+# follows. Again, the compositions by default are interpreted as mole fractions
+Z = gas.mixture_fraction(fuel="CH4", oxidizer="O2:0.21,N2:0.79")
+print(f"Z = {Z:1.3f}")
```
ischoegl, line 52:
```suggestion
print(f"Z = {Z:1.3f}")

```
(missing line break)

interfaces/cython/cantera/examples/thermo/equivalenceRatio.py
```diff
@@ -2,52 +2,116 @@
...
+# By default, the mixture fraction is the Bilger mixture fraction. Instead,
+# a mixture fraction based on a single element can be used. In this example,
+# the following two ways of computing Z are the same:
+Z = gas.mixture_fraction(fuel="CH4", oxidizer="O2:0.21,N2:0.79", element="Bilger")
+Z = gas.mixture_fraction(fuel="CH4", oxidizer="O2:0.21,N2:0.79", element="C")
```
ischoegl: Print example output and add line break.

interfaces/cython/cantera/examples/thermo/equivalenceRatio.py
```diff
@@ -2,52 +2,116 @@
...
 gas = ct.Solution('gri30.yaml')
 
-# fuel and oxidizer compositions
-fuel = "CH4"
-oxidizer = "O2:0.21,N2:0.79"
+# Set the mixture composition according to the stoichiometric mixture
+# (equivalence ratio phi = 1). The fuel composition in this example
+# is set to 100 mol-% CH4 and the oxidizer to 21 mol-% O2 and
+# 79 mol-% N2.
+gas.set_equivalence_ratio(phi=1.0, fuel="CH4:1", oxidizer="O2:0.21,N2:0.79")
```
ischoegl: Since `oxidizer="O2:0.21,N2:0.79"` is used frequently, it could be defined as 
```
air = "O2:0.21,N2:0.79"
```
for clarity, and referenced as `oxidizer=air` thereafter.

ischoegl (changes requested):
@g3bk47 ... thank you for the fast turn-around! While I believe that the implementation is sound, I have some additional requests (which are mostly formatting).

interfaces/cython/cantera/thermo.pyx
g3bk47: I'm not sure if I understand correctly. So just to clarify: Calling `set_equivalence_ratio` with the new dilution option basically performs two mixing operations (this is why I used "final mixture" in the previous version of the description). The first mixing operation creates a mixture of fuel and oxidizer according to `phi`. This first step is always the same (with or without dilution and independent of the fraction argument). If dilution is used, a second mixing operation is performed. The fuel/oxidizer mixture at `phi` is then mixed with the diluent. This mixing is controlled by the `fraction` argument. If fraction specifies "diluent", then the final mixture consists of `(1-fraction)` parts fuel/oxidizer and fraction parts dilution. However. If `fraction` specifies the fuel, then the diluent is added to the mixture until the fraction of fuel in the final mixture is `fraction`. This means, in general, that the final mixture will have a different equivalence ratio than specified, but the ratio of fuel to oxidizer in the final mixture is alawys the same (according to `phi`) independent of the `fraction` argument. For example
set_equivalence_ratio(1.0, "H2", "O2", fraction="fuel:0.1", diluent="H2O")
results in a mixture with X_H2 = 0.1 and X_H2/X_O2 = 2. This is the requested behavior from https://github.com/Cantera/enhancements/issues/108.

A different interpretation of dilution would be to reverse the two mixing operations. For example, to first dilute only the fuel with 10% H2O, you could just easily write:
`set_equivalence_ratio(1, "H2:0.9,H2O:0.1", air)`
This behavior is of course different, since now the resulting mixture is guaranteed to be at phi=1, but X_H2/X_O2 is not necessarily 2 anymore.

Both interpretations are valid, but the second one is trivial for a user to write while the first is a bit tricky to implement, so in my opinion having a dedicated function for the first case is justified. Maybe I could include this second case as an example in equivalenceRatio.py? I have incorporated all other requested changes. I will commit them once this discussion is clarified.

ischoegl: Hmm. I’m not sure that I am following completely: as long as an inert mixture is used, the ratio of fuel to oxidizer should not be affected by a diluent (a user could certainly wreak havoc by specifying reactants as diluents, but that is not something that needs to be handled). To my understanding, `fraction` just specifies how the diluent ratio is specified: either as a fraction of “fuel, oxidizer, or the [reactant] mixture”. If I’m not mistaken, this is how your unit tests are already implemented?

PS: This argument should extend to the equivalence ratio, even if the AFR’s are changed by the diluent in the process.

g3bk47: Maybe we mean the same thing. But even if the diluent is just another fuel species, everything is well defined.
Lets make an example:
Assume I want a mixture of H2 and O2 at stoichiometric conditions. So I would write:
set_equivalence_ratio(1, "H2", "O2")
So I get a mixture with `X_H2 = 2 / 3` and` X_O2 = 1 / 3` or `X_H2 / X_O2 = 2`.
Now I want to dilute this mixture with "CH4" in a way that the fraction of fuel (i.e. H2) in the final mixture is 0.1.
For this, I would write:
set_equivalece_ratio(1, "H2", "O2", fraction={"fuel": 0.1}, diluent="CH4")
This gives a mixture with `X_H2 = 0.1, X_O2 = 0.05` and `X_CH4 = 0.85`.
Whatever this composition is, it is always guaranteed that X_H2 is 0.1 (i.e. according to `fraction={"fuel": 0.1}`) and X_H2/X_O2 is guaranteed to be 2 (according to. `phi=2`).
The goal of the function with dilution is not to get a final mixture with the specified `phi`, but to dilute a oxidizer/fuel mixture at a specified `phi `with an arbitrary diluent. At least this is what I get from https://github.com/Cantera/enhancements/issues/108

P.S:
I thought about having the interface of `equivalence_ratio` mirror the one of `set_equivalence_ratio`, but I am not sure how useful this is in practice. I could implement something like this:
```
#find the equivalence ratio of the following gas object:
phi = equivalence_ratio("H2","O2",fraction={"fuel":0.1},diluent="CH4") # prints "2" according to the example above
```
But this would not cover the case where a user just wants to ignore certain species (as in the users' group discussion), since you would always have to specify a concrete fraction value for the dilution.
To be consistent, this would also probably have to be added to "mixture_fraction", "set_mixture_fraction" and "stoich_fuel_air_ratio".  In theory, there could be a special case where only a diluent composition is given without fraction and all species in the diluent are ignored. This would mirror the new "include_species" option.

ischoegl: > Maybe we mean the same thing.

I think so. The way it works is that the equivalence ratio is defined by fuel and oxidizer, whereas the diluent is not considered. The fraction just specifies how the diluent is quantified. So overall, I think we are in agreement, although the docstring should be tweaked.

g3bk47: OK, nice 👍  Then I will prepare the next commit

interfaces/cython/cantera/thermo.pyx
ischoegl: Some of the reformatting could be improved by avoiding `\`, for example
```python
        cdef np.ndarray[np.double_t, ndim=1] fuel_comp = np.ascontiguousarray(
            self.__composition_to_array(fuel, basis), dtype=np.double)
```
instead of 
```python
        cdef np.ndarray[np.double_t, ndim=1] fuel_comp = \
                np.ascontiguousarray(self.__composition_to_array(fuel, basis),
                                     dtype=np.double)
```

interfaces/cython/cantera/test/test_thermo.py
ischoegl: I think `phi = 2` should be defined outside of `test_simple_dilution` (right around where you define the mixture); it will be accessible without passing it as argument. Also, define `afr = 0.5 / phi` (or the reciprocal fuel-to-air ratio) and use this to avoid the hard-coded factor '4'.

interfaces/cython/cantera/thermo.pyx
ischoegl: Well, I think I should have suggested "``Python example `equivalenceRatio.py <url>`_,``" (sorry)

interfaces/cython/cantera/thermo.pyx
ischoegl: I don't think that the docstring accurately describes the dilution option yet. Please address the section "specifying
the amount of diluent, fuel or oxidizer in the mixture." ... perhaps "[...]. The amount of diluent is quantified as a fraction of fuel, oxidizer or the fuel/oxidizer mixture." ?

ischoegl (approved):
@g3bk47 … thank you for taking care of all the requested changes! I’ll approve, but leave it open for another day before merging in case anyone else wanted to comment.

PR #1218, opened 2022-03-14
Fix Redlich-Kwong partial molar properties

### LINKED ISSUE #998: Redlich-Kwong implementations of getPartialMolarIntEnergies and getPartialMolarCp are incorrect
**Problem description**

The implementations of `getPartialMolarIntEnergies` and `getPartialMolarCp` in the `RedlichKwongMFTP` class incorrectly return the standard state internal energies and heat capacities, respectively. This can be seen pretty clearly from their implementations

https://github.com/Cantera/cantera/blob/908a2cceb50bf490eb2808905ee9ba5192d2e8ad/src/thermo/RedlichKwongMFTP.cpp#L385-L395

which are just scaling the values returned by functions which return the non-dimensional standard state values:

https://github.com/Cantera/cantera/blob/908a2cceb50bf490eb2808905ee9ba5192d2e8ad/include/cantera/thermo/ThermoPhase.h#L620-L639

In reality, the calculation of the partial molar heat capacities should be significantly more involved. The correct implementation of the partial molar internal energies can be done fairly easily by using the identity `h_k = u_k + P * v_k`, where the complexity is in the already-implemented methods for the partial molar enthalpies and partial molar volumes.

**Steps to reproduce**

This can also be seen in by computing these values as (a) the total quantity directly, (b) the dot product of the partial molar quantities with the mole fractions and (c) the dot product of the standard state quantities with the mole fractions. 

```python
>>> gas = ct.Solution('nDodecane_Reitz.yaml')
>>> gas.TPY = 300, 100 * ct.one_atm, 'CO2: 1.0, CH4: 1.0'

>>> gas.int_energy_mole
-164146896.2111432
>>> sum(gas.partial_molar_int_energies * gas.X)
-162451424.60398442
>>> sum(gas.X * gas.standard_int_energies_RT * ct.gas_constant * gas.T)
-162451424.6039844

>>> gas.cp_mole
56325.864657238315
>>> sum(gas.partial_molar_cp * gas.X)
35800.6557848247
>>> sum(gas.X * gas.standard_cp_R * ct.gas_constant)
35800.655784824696
```

**Behavior**

(a) and (b) should give identical values, but instead, (b) and (c) give identical values:

**System information**

- Cantera version: 2.5.1 / `main` branch at bbb752ac6

speth:

- Remove incorrect implementation of `RedlichKwongMFTP::getPartialMolarCp`
- Correct implementation of `RedlichKwongMFTP::getPartialMolarIntEnergies`

Closes #998

**Checklist**

### DIFF
--- a/src/thermo/RedlichKwongMFTP.cpp
+++ b/src/thermo/RedlichKwongMFTP.cpp
@@ -341,12 +341,6 @@ void RedlichKwongMFTP::getPartialMolarIntEnergies(doublereal* ubar) const
     scale(ubar, ubar+m_kk, ubar, RT());
 }
 
-void RedlichKwongMFTP::getPartialMolarCp(doublereal* cpbar) const
-{
-    getCp_R(cpbar);
-    scale(cpbar, cpbar+m_kk, cpbar, GasConstant);
-}
-
 void RedlichKwongMFTP::getPartialMolarVolumes(doublereal* vbar) const
 {
     for (size_t k = 0; k < m_kk; k++) {
--- a/src/thermo/RedlichKwongMFTP.cpp
+++ b/src/thermo/RedlichKwongMFTP.cpp
@@ -337,8 +337,13 @@ void RedlichKwongMFTP::getPartialMolarEntropies(doublereal* sbar) const
 
 void RedlichKwongMFTP::getPartialMolarIntEnergies(doublereal* ubar) const
 {
-    getIntEnergy_RT(ubar);
-    scale(ubar, ubar+m_kk, ubar, RT());
+    // u_k = h_k - P * v_k
+    getPartialMolarVolumes(m_partialMolarVolumes.data());
+    getPartialMolarEnthalpies(ubar);
+    double p = pressure();
+    for (size_t k = 0; k < nSpecies(); k++) {
+        ubar[k] -= p * m_partialMolarVolumes[k];
+    }
 }
 
 void RedlichKwongMFTP::getPartialMolarVolumes(doublereal* vbar) const
--- a/include/cantera/thermo/RedlichKwongMFTP.h
+++ b/include/cantera/thermo/RedlichKwongMFTP.h
@@ -117,7 +117,9 @@ public:
     virtual void getPartialMolarEnthalpies(doublereal* hbar) const;
     virtual void getPartialMolarEntropies(doublereal* sbar) const;
     virtual void getPartialMolarIntEnergies(doublereal* ubar) const;
-    virtual void getPartialMolarCp(doublereal* cpbar) const;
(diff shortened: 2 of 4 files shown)


src/thermo/RedlichKwongMFTP.cpp
```diff
@@ -337,8 +337,13 @@ void RedlichKwongMFTP::getPartialMolarEntropies(doublereal* sbar) const
...
 void RedlichKwongMFTP::getPartialMolarIntEnergies(doublereal* ubar) const
 {
-    getIntEnergy_RT(ubar);
-    scale(ubar, ubar+m_kk, ubar, RT());
+    // u_k = h_k - P * v_k
+    getPartialMolarVolumes(m_partialMolarVolumes.data());
+    getPartialMolarEnthalpies(ubar);
+    double p = pressure();
+    for (size_t k = 0; k < nSpecies(); k++) {
+        ubar[k] -= p * m_partialMolarVolumes[k];
```
decaluwe, line 345: Nice!  I was fearing some long derivation on which I'd have to go and run due diligence to check 😂

decaluwe (dismissed):
Thanks for tackling this one, @speth.  It looks good to me.

Is the idea that hopefully some day we will remove the `check_cp` flag from the `checkThermo` definition?  I think it's the right call, right now, but does seem odd that just that specific property is excepted (and I certainly wouldn't want it viewed as a precedent for anyone who just doesn't feel like implementing a full thermo feature set, but I think that's easy enough for us to head off).

speth:
Ah, that's an interesting point -- as I had it, it would have probably gone unnoticed to remove the `check_cp=False` settings from those tests if `getPartialMolarCp` was later implemented, so I changed it to test that an exception is raised when `check_cp=False`.

I'm not necessarily worried about thermo models not providing a "complete" interface. I'd be more interested in defining what the minimal interface that a model needs to provide is, say to be compatible with the reactor network or 1D flame models. As it is, I don't know when you actually need to know the partial molar specific heats -- they appear in the 1D flame for the ideal gas case, but end up being replaced by the partial molar enthalpies in the non-ideal formulation (according to #1079).

decaluwe:
Ah, good point - the partial molar enthalpies being the generally correct approach, I would be slightly curious to see if there is a speed difference, or if we could just replace the partial molar specific heats in the ideal gas reactor with partial molar enthalpies, and be no worse for it.

IIRC, one of J. Shepherd's students originally wrote asking for access to these partial molar entities, for use with non-ideal phases in `SD-ToolBox`.

ischoegl (approved):
@decaluwe ... is this ready to be merged? If not, just dismiss my approval.

PR #1222, opened 2022-03-20
Disable Python 3.6

ischoegl:

- Disable Python 3.6 (which will stop working after #1158)
- Reformat some build messages during Python checks

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -56,8 +56,9 @@ Additional help command options:
 # SConstruct is parsed by Python 2. This seems to be the most robust
 # and simplest option that will reliably trigger an error in Python 2
 # and provide actionable feedback for users.
+python_min_build_support = "3.7"
 f"""
-Cantera must be built using Python 3.6 or higher. You can invoke SCons by executing
+Cantera must be built using Python 3.7 or higher. You can invoke SCons by executing
     python3 `which scons`
 followed by any desired options.
 """
@@ -72,6 +73,15 @@ import textwrap
 from os.path import join as pjoin
 from pkg_resources import parse_version
 import SCons
+
+# ensure that Python version is sufficient for build process
+python_version = "{v.major}.{v.minor}".format(v=sys.version_info)
+if parse_version(python_version) < parse_version(python_min_build_support):
+    print(
+        f"ERROR: Cantera must be built using Python {python_min_build_support} or "
+        f"higher; Python {python_version} is not supported.", file=sys.stderr)
+    sys.exit(1)
+
 from buildutils import *
 
 if not COMMAND_LINE_TARGETS:
--- a/SConstruct
+++ b/SConstruct
@@ -1568,7 +1568,6 @@ if env['python_package'] != 'none':
                 f"Could not execute the Python interpreter '{env['python_cmd']!r}'")
             sys.exit(1)
     elif python_version < python_min_version:
-        msg = ""
         if env["python_package"] in ("minimal", "full", "default"):
             logger.error(
                 f"Python version is incompatible. Found {python_version} but "
@@ -1581,9 +1580,9 @@ if env['python_package'] != 'none':
     else:
 
         if len(info) > expected_output_lines:
-            msg = "\n| ".join(info[expected_output_lines:])
-            logger.warning(
-                f"Unexpected output while checking Python dependency versions:{msg}")
+            msg = ["Unexpected output while checking Python dependency versions:"]
+            msg.extend(info[expected_output_lines:])
+            logger.warning("\n| ".join(msg))
 
         warn_no_full_package = False
         if python_version >= parse_version("3.8"):
--- a/SConstruct
+++ b/SConstruct
@@ -1536,43 +1536,44 @@ if env['python_package'] != 'none':
         if check_for_ruamel_yaml:
             ruamel_yaml_version = parse_version(info[3])
             if ruamel_yaml_version == parse_version("0.0.0"):
-                print("ERROR: ruamel.yaml was not found. {} or newer is "
-                    "required".format(ruamel_min_version))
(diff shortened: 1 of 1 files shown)


SConstruct
```diff
@@ -1445,7 +1445,7 @@ if env['VERBOSE']:
 env['python_cmd_esc'] = quoted(env['python_cmd'])
 
 # Python Package Settings
-python_min_version = parse_version("3.6")
+python_min_version = parse_version("3.7")
 # The string is used to set python_requires in setup.py.in
```
```suggestion
# The string is used to set python_requires in setup.cfg.in
```

Anticipating #1158 being merged. Suggesting the change here to avoid having to push another commit to #1158 😄

bryanwweber (commented):
I think we should have a check at the top of `SConstruct` as well that the Python running SCons is >=3.7. But that can wait until after 2.6, I suppose, since it would preclude system Pythons from even running SCons and we're not using (nor will we add any, I bet) 3.7-specific features into the various SCons* files before 2.6 is released.

ischoegl:
Done!

ischoegl:
I wonder if it wouldn't be preferable to error out for Python 3.6 if `python_package = 'default'`, and indicate that it can still be installed for `python_package = 'none'`. Going through the entire build/installation process only to realize that Python was automatically skipped is not ideal ...

speth:
As a result of #1158, the requirement of Python 3.7 or newer applies to the Python interpreter running `scons`. If you try to build using Python 3.6, you will get the error:
```
  File "/src/cantera/site_scons/buildutils.py", line 1

    from __future__ import annotations
                                     ^

SyntaxError: future feature annotations is not defined
```
I don't think you'll ever get to the error handling added here except in the somewhat unlikely case where `scons` is running using Python 3.7+ but `python_cmd` has been explicitly pointed to an older Python installation.

The two options I see for providing a more helpful are either:
- Put an end of line comment after `from __future__ import annotations`.
- Check the Python version before importing `buildutils` in `SConstruct`

bryanwweber:
> * Check the Python version before importing `buildutils` in `SConstruct`

This would be my preference using `sys.version`

ischoegl:
> Check the Python version before importing `buildutils` in `SConstruct`

Np. What I had worked well prior to merging #1158. I added the check as requested, but left the version decision separate from the pre-existing checks.

SConstruct
```diff
@@ -72,6 +74,15 @@ import textwrap
 from os.path import join as pjoin
 from pkg_resources import parse_version
 import SCons
+
+# ensure that Python version is sufficient for build process
+python_version = "{v.major}.{v.minor}".format(v=sys.version_info)
+if parse_version(python_version) < parse_version(python_min_build_support):
+    print(
+        f"ERROR: Cantera must be built using Python {python_min_build_support} or "
```
bryanwweber, line 81:
```suggestion
        f"higher; Python {python_version} is not supported.", file=sys.stderr)
```
I'm on my phone, but I think that's the right syntax.

SConstruct
```diff
@@ -56,8 +56,10 @@ Additional help command options:
 # SConstruct is parsed by Python 2. This seems to be the most robust
 # and simplest option that will reliably trigger an error in Python 2
 # and provide actionable feedback for users.
+python_min_build_support = "3.7"
 f"""
-Cantera must be built using Python 3.6 or higher. You can invoke SCons by executing
+Cantera must be built using Python {python_min_build_support} or higher.
+You can invoke SCons by executing
```
speth: The problem with changing this to make the version a variable is that the version won't be printed in the error message returned to the user. For example, if this is invoked using Python 2.7, the error is:
```
  File "/src/cantera/SConstruct", line 65

    f"""
Cantera must be built using Python {python_min_build_support} or higher.
You can invoke SCons by executing
    python3 `which scons`
followed by any desired options.
"""
```

speth (changes requested):

SConstruct
ischoegl: :+1: Realized this right after pushing but had to leave my computer. Fixed!

SConstruct
```diff
@@ -1536,43 +1546,44 @@ if env['python_package'] != 'none':
...
+        logger.info(f"Building the minimal Python package for Python {python_version}")
     else:
 
         if len(info) > expected_output_lines:
-            print("WARNING: Unexpected output while checking Python "
-                  "dependency versions:")
-            print('| ' + '\n| '.join(info[expected_output_lines:]))
+            msg = "\n| ".join(info[expected_output_lines:])
+            logger.warning(
+                f"Unexpected output while checking Python dependency versions:{msg}")
```
speth: I think this is now going to print the first line of the unexpected output immediately after "...dependency versions", rather than on the following line with the `| ` prefix that is done for the following lines.

SConstruct
```diff
@@ -1536,43 +1546,44 @@ if env['python_package'] != 'none':
...
-                env['python_cmd']))
+            logger.error(
+                f"Could not execute the Python interpreter '{env['python_cmd']!r}'")
             sys.exit(1)
     elif python_version < python_min_version:
-        msg = ("{}: Python version is incompatible. Found {} but {} "
-               "or newer is required")
-        if env["python_package"] in ("minimal", "full"):
-            print(msg.format("ERROR", python_version, python_min_version))
+        msg = ""
```
speth: This definition of `msg` is unused, right?

speth (changes requested):

SConstruct
ischoegl: :+1: Oops.

speth (approved):
Thanks, @ischoegl. Feel free to merge once the CI passes.

bryanwweber (approved):

PR #1219, opened 2022-03-15
Proof-of-Concept: Eliminate reaction specializations

ischoegl:
**Changes proposed in this pull request**

Remove remaining specializations per Cantera/enhancements#142

- Create `ThreeBodyArrheniusRate` specialization for  `three-body-Arrhenius`
- Eliminate `ThreeBodyReaction` in C++ and Python
- :construction:  

Once fully implemented, this PR will allow for a deprecation of `ReactionFactory`.

**If applicable, fill in the issue number this pull request is fixing**

Closes Cantera/enhancements#142, closes Cantera/enhancements#133, interferes with #1143

**If applicable, provide an example illustrating new features this pull request is introducing**

**Checklist**

- [ ] The pull request includes a clear description of this code change
- [ ] Commit messages have short titles and reference relevant issues
- [ ] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [ ] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [ ] The pull request is ready for review

**Thoughts**

Due to the introduction of `BulkRate.h`, this PR has some minor conflicts with #1217

PR #1223, opened 2022-03-20
Reorganize reaction rate classes

speth:

- Move each implementation of a `ReactionRate` (e.g. `TwoTempPlasmaRate`) to a header file of its own, along with the implementation of the corresponding `ReactionData` class. By making rate implementations self-contained, I think this makes it easier for users to see how to add new ones of their own.
- Move the `ReactionFactory` declaration into an internal header file, so it can be removed after 2.6 without having to go through a deprecation cycle. (this class was introduced after Cantera 2.5, and so has never been a part of the released API)
- Deprecate the `update_C` and `updateRC` methods of `ChebyshevRate` and `PlogRate`, which are used directly only by the "legacy" rate methods.

**Checklist**


ischoegl:
@speth ... I have absolutely no problem with this proposition (it makes sense). Would it make sense to also introduce `BulkRate.h` as envisioned in #1219? It is the direct correspondence to `InterfaceRate` ....

PS: also marked this as relevant for the beta release ...

PPS: For #1219 this is just 96d42a05691d5b8d2d8a85ce0604d6939c09cf94 and bcf6ca7006e0b6e72a11bed23f2773564f0e9213 (and potentially ab47b50fb1a23ba77365f71e6fa2ed5c707dfe5f)

speth:
I like the changes in https://github.com/Cantera/cantera/commit/bcf6ca7006e0b6e72a11bed23f2773564f0e9213. I actually think they make the `BulkRate` class almost unnecessary. I'd like to see what I can get out of pursuing that in a separate PR, since this one already has so much cut-and-paste rearrangement that I think it would be harder to see the specific changes related to that refactoring.

ischoegl (approved):
As far as I am aware, this is just code reorganization …

PR #1227, opened 2022-03-22
Multicomponent diffusion coefficient documentation.

### LINKED ISSUE #143: [Cython] Use properties X and Y to get and set mole/mass fractions on Sim1D boundaries

### LINKED ISSUE #143: Add to documentation: multi_diff_coeffs

decaluwe:
- Adds content to multicomponent diffusion coefficient docstrings,
to explain the matrix indexing.
- Corrects an error in the multiDiffCoeffs.m docstring, regarding
the data returned by the function (array vs. matrix).

Closes https://github.com/Cantera/enhancements/issues/143

### DIFF
--- a/include/cantera/transport/TransportBase.h
+++ b/include/cantera/transport/TransportBase.h
@@ -576,7 +576,7 @@ public:
      * @param[in] ld  The dimension of the inner loop of d (usually equal to m_nsp)
      * @param[out] d  flat vector of diffusion coefficients, fortran ordering.
      *            d[ld*j+i] is the D_ij diffusion coefficient (the diffusion
-     *            coefficient for species i due to species j).
+     *            coefficient for species i due to concentration gradients in *            species j).
      */
     virtual void getMultiDiffCoeffs(const size_t ld, doublereal* const d) {
         throw NotImplementedError("Transport::getMultiDiffCoeffs",
--- a/include/cantera/transport/TransportBase.h
+++ b/include/cantera/transport/TransportBase.h
@@ -576,7 +576,8 @@ public:
      * @param[in] ld  The dimension of the inner loop of d (usually equal to m_nsp)
      * @param[out] d  flat vector of diffusion coefficients, fortran ordering.
      *            d[ld*j+i] is the D_ij diffusion coefficient (the diffusion
-     *            coefficient for species i due to concentration gradients in *            species j).
+     *            coefficient for species i due to concentration gradients in 
+     *            species j).
      */
     virtual void getMultiDiffCoeffs(const size_t ld, doublereal* const d) {
         throw NotImplementedError("Transport::getMultiDiffCoeffs",
--- a/include/cantera/transport/TransportBase.h
+++ b/include/cantera/transport/TransportBase.h
@@ -577,7 +577,7 @@ public:
      * @param[out] d  flat vector of diffusion coefficients, fortran ordering.
      *            d[ld*j+i] is the D_ij diffusion coefficient (the diffusion
      *            coefficient for species i due to concentration gradients in 
-     *            species j).
+     *            species j). Units: m^2/s
      */
     virtual void getMultiDiffCoeffs(const size_t ld, doublereal* const d) {
         throw NotImplementedError("Transport::getMultiDiffCoeffs",
--- a/interfaces/cython/cantera/transport.pyx
+++ b/interfaces/cython/cantera/transport.pyx
@@ -263,7 +263,9 @@ cdef class Transport(_SolutionBase):
     property multi_diff_coeffs:
         """Multicomponent diffusion coefficients, D_{i,j} [m^2/s].
         
-        The coefficient D_{row,column} is used for computing the contribution of concentration gradients in species [column] to the diffusive flux of species [row]."""
+        The coefficient D_{row,column} is used for computing the contribution 
+        of concentration gradients in species [column] to the diffusive flux of 
+        species [row]."""
         def __get__(self):
             return get_transport_2d(self, tran_getMultiDiffCoeffs)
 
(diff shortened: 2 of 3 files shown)


speth:
My thought on this is that the more critical thing to explain is that `D[i,j]` is the coefficient used for computing the contribution of the gradient in the concentration of species `j` to the diffusive flux of species `i`. The more generic statement that the first index is the row of the matrix and the second is the column doesn't even need to come into it.

Also, if you're updating the Python and Matlab docstrings, could you do the same for the C++ docstring as well?

decaluwe:
I'll try to craft an update that covers both possible uncertainties.

By the C++ docstrings, do you mean in the `include` header files?  Or is there another location?

ischoegl:
> Also, if you're updating the Python and Matlab docstrings, could you do the same for the C++ docstring as well?

C++ docstring would be here ...
https://github.com/Cantera/cantera/blob/5f35bd805e7430092aa63d803f375eb8afebef99/include/cantera/transport/TransportBase.h#L570-L580

interfaces/cython/cantera/transport.pyx
```diff
@@ -261,7 +261,9 @@ cdef class Transport(_SolutionBase):
             return get_transport_1d(self, tran_getThermalDiffCoeffs)
 
     property multi_diff_coeffs:
-        """Multicomponent diffusion coefficients [m^2/s]."""
+        """Multicomponent diffusion coefficients, D_{i,j} [m^2/s].
+        
+        The coefficient D_{row,column} is used for computing the contribution of concentration gradients in species [column] to the diffusive flux of species [row]."""
```
speth: I feel like it's less obvious that `row` and `column` are placeholders for particular species indices here than just continuing to use `i` and `j`.

ischoegl: Could you limit line length to 88 characters?

include/cantera/transport/TransportBase.h
```diff
@@ -576,7 +576,7 @@ class Transport
      * @param[in] ld  The dimension of the inner loop of d (usually equal to m_nsp)
      * @param[out] d  flat vector of diffusion coefficients, fortran ordering.
      *            d[ld*j+i] is the D_ij diffusion coefficient (the diffusion
-     *            coefficient for species i due to species j).
+     *            coefficient for species i due to concentration gradients in *            species j).
```
ischoegl: Missing line break 😄

ischoegl (commented):
Minor stuff …

interfaces/cython/cantera/transport.pyx
decaluwe: You are assuming, then, that all users understand that in `D_{i,j}` we mean that `i = row` and `j = column`? 

I don't really know that I'd agree.  But could cover this a different way.

decaluwe: How about:

> The coefficient D_{i,j} is used for computing the contribution 
        of concentration gradients in species ``j`` to the diffusive flux of 
        species ``i``, where ``i = row`` and ``j = column``.

decaluwe: Actually, @speth I may be coming around to your point - knowing `row` and `column` is almost irrelevant, since the indices are how you access the individual coefficients, anyway. So regardless of which is row and which is column, `D[i,j]` in Python or `D(i,j)` in Matlab is how you access the value (C++ is different, since it is a flattened matrix, but that is already spelled out in the docstring).

I would be fine either with what I suggest in the comment above, or just replacing `D_{i,j}` with `D[i,j]` or `D(i,j)`, as appropriate, in what I've currently got in the PR, dropping the `row` and `column` placeholders.

Thinking about it more, it is probably better to replace the `D_{i,j}`, no matter what.

decaluwe:
Okay, one last set of tweaks.  The C++ docstring was already pretty close to what we've been discussing, here (only missing the word "concentration" and the units, imho), so I just fixed that up, applied that in the Matlab and Python docstrings, and made each indexing expression match the format for that particular language.

I'll let y'all do with it what you will, at this point. :)

interfaces/cython/cantera/transport.pyx
```diff
@@ -261,7 +261,8 @@ cdef class Transport(_SolutionBase):
             return get_transport_1d(self, tran_getThermalDiffCoeffs)
 
     property multi_diff_coeffs:
-        """Multicomponent diffusion coefficients [m^2/s]."""
+        """Multicomponent diffusion coefficients, D[i,j], the diffusion coefficient for species i due to concentration gradients in
```
ischoegl: Looks good to me, except for line length 😉

ischoegl (changes requested):

ischoegl (approved):
🎉

bryanwweber (approved):
Thanks @decaluwe, this came out really well!

PR #1232, opened 2022-03-29
Fix building sdist and wheels and conda packages

bryanwweber:

- Reduce file size of wheels built on Linux by removing debug symbols
- Fix writing the sdist so that it is platform-independent
- Fix tests that write to the local directory
- Fix the name of the versioned Boost folder in CI
- Allow instances of `pathlib.Path` as the filename for `write_yaml`

---

### DIFF
--- a/interfaces/python_sdist/SConscript
+++ b/interfaces/python_sdist/SConscript
@@ -9,7 +9,8 @@ from build.env import IsolatedEnvBuilder
 
 from buildutils import logger
 
-Import("env")
+Import("env", "configh")
+configh = configh.copy()
 
 localenv = env.Clone()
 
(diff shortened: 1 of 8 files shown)


interfaces/python_sdist/SConscript
```diff
@@ -89,18 +90,79 @@ def replace_git_hash(target, source, env):
...
+    def __init__(self, data):
+        self.data = data
+
+    def __getitem__(self, key):
+        if key not in self.data:
+            return f"{{{key!s}!s}}"
+        elif self.data[key] is None:
+            return f"/* #undef {key!s} */"
+        else:
+            return f"#define {key!s} {self.data[key]!s}"
```
speth: Why doesn't the existing `buildutils.DefineDict` work here?

interfaces/python_sdist/setup.py
```diff
@@ -92,28 +93,95 @@ def cythonize(extensions):
...
-    sundials_configh = {
-        "SUNDIALS_USE_GENERIC_MATH": "#define SUNDIALS_USE_GENERIC_MATH 1",
-        "SUNDIALS_BLAS_LAPACK": "/* #undef SUNDIALS_BLAS_LAPACK */"
-    }
-    sundials_cflags = ["-w"]
-    sundials_macros = []
+
+def configure_build():
+    boost_version = ""
+    boost_locs = (os.environ.get("BOOST_INCLUDE", None), BOOST_INCLUDE, "/usr/include",
```
speth, line 99: Is all of what follows just to figure out whether we have Boost >= 1.56? That's pretty old (2014), and I'd say that just dropping it is definitely an option. Though I we might want to optionally use some newer boost feature in the future, in which case having this available would be useful.

speth (dismissed):
Looks good to me. I had a couple of questions, but they don't necessarily require any changes.

interfaces/python_sdist/SConscript
bryanwweber: It replaces keys that don't exist with the `undef` line, but I need some way to make sure that some keys are left in `config.h.in` for `setup.py` to fill in. I'll add a docstring 👍

interfaces/python_sdist/setup.py
bryanwweber: The Linux wheels are built (by default) in the `manylinux2014` container which is built on CentOS 7, which has maybe 1.54 in the repos (older than 1.56 anyways)? Santosh posted an issue about missing demangle not that long ago as well. Anyways, as you say, this isn't much code and it will eventually allow us to use newer Boost features.

bryanwweber:
@speth Since I pushed new commits, this needs another green check, then I think it's good to :shipit:

speth (approved):

PR #1234, opened 2022-04-02
Add keywords to examples

speth:

- Add a list of keywords to all of the examples (Python, Matlab, C++, and Fortran, which can (eventually) be used to provide additional sorting / searching options for the examples on the Cantera website, rather than categorizing only by implementation language. 
- To keep some regularity of the keywords used, add a script that parses the keywords and determines if any examples are missing keywords or introduce new keywords that are not in the list of previously-used keywords (which can be explicitly updated)
- Add a CI check that fails if the script above is not satisfied.

Partially addresses https://github.com/Cantera/enhancements/issues/88

If a PR contains examples with no keywords, or keywords that haven't been declared, you'll get a CI failure that looks like this:
image

**Checklist**

### DIFF
--- a/doc/example-keywords.py
+++ b/doc/example-keywords.py
@@ -66,7 +66,7 @@ def get_matlab_keywords(filename):
         logging.error(f"Couldn't parse docstring for {filename}")
         return False
     docstring = match.group(0) + "\n\n"
-    docstring = "\n".join(line.lstrip("% ") for line in docstring.split("\n"))
+    docstring = "\n".join(line.lstrip("% ") for line in docstring.splitlines())
 
     match = re.search(r"\s*Keywords:(.*?)\n\n", docstring, re.DOTALL | re.MULTILINE)
     if not match:
@@ -86,7 +86,7 @@ def get_cxx_keywords(filename):
         logging.error(f"Couldn't parse docstring for {filename}")
         return False
     docstring = match.group(1) + "\n\n"
-    docstring = "\n".join(line.lstrip("* ") for line in docstring.split("\n"))
+    docstring = "\n".join(line.lstrip("* ") for line in docstring.splitlines())
     match = re.search(r"\s*Keywords:(.*?)\n\n", docstring, re.DOTALL | re.MULTILINE)
     if not match:
         EXIT_CODE = 1
@@ -107,7 +107,7 @@ def get_fortran_keywords(filename, comment_char):
         return False
     docstring = match.group(0) + "\n\n"
     docstring = "\n".join(line.lstrip(f"{comment_char} ")
-                          for line in docstring.split("\n"))
+                          for line in docstring.splitlines())
 
     match = re.search(r"\s*Keywords:(.*?)\n\n", docstring, re.DOTALL | re.MULTILINE)
     if not match:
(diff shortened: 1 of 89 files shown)


doc/example-keywords.py
```diff
@@ -0,0 +1,211 @@
...
+    keywords = set(kw.strip() for kw in match[1].split(","))
+    return keywords
+
+
+def get_all_keywords():
+    """
+    Read keywords from all Cantera examples and print out a summary list
+    """
+    text = (Path(__file__).parent / "example-skip-keywords.txt").read_text()
+    skip = set(text.split("\n"))
```
bryanwweber:
```suggestion
    skip = set(text.splitlines())
```

This will handle `\r\n` as well

doc/example-keywords.py
```diff
@@ -0,0 +1,211 @@
...
+
+    all_keywords = Counter()
+    for d in Path("interfaces/cython/cantera/examples").glob("**"):
```
bryanwweber: This will cause an error if the cwd is the doc directory. Can we be more robust to that case?

doc/example-keywords.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def save_keywords():
+    """
+    Save an updated version of the known keywords list based on keywords appearing in
+    any of the examples.
+    """
+    found_kw = get_all_keywords()
+    with open(Path(__file__).parent / "example-keywords.txt", "w") as known:
+        known.writelines(kw + "\n" for kw in sorted(found_kw))
```
bryanwweber:
```suggestion
    (Path(__file__).parent / "...").write_text("\n".join(sorted(found_kw)))
```
I think this is simpler, but YMMV 😊 might take 2 lines anyways so there might not be an improvement.

doc/example-keywords.py
```diff
@@ -0,0 +1,211 @@
...
+        print(f"{kw} ({count})")
+
+
+if __name__ == "__main__":
+    if "compare" in sys.argv:
+        delta = compare()
+        if delta:
+            sys.exit(1)
+    elif "save" in sys.argv:
+        save_keywords()
```
bryanwweber, line 206: I think it is a little strange for the script to exit with `1` but still save the file to disk. Is that behavior deliberate? If so, I think it's probably fine, but just curious.

bryanwweber (changes requested):
Thanks @speth! A few small things here the list of keywords looks good, although I didn't check each example if they were labeled appropriately.

doc/example-keywords.py
speth: The file is being read in the "universal newlines" mode, so `\r\n` is going to be automatically translated into `\n` no matter what. But I can change this regardless.

doc/example-keywords.py
speth: Your example loses the newline at the end of the file, but I was able to fix that and fit it all on one line.

doc/example-keywords.py
speth: I hadn't really thought about it, but I think this behavior makes sense -- I'd rather be able to see what the script was able to generate than not, even if there are missing or incorrectly formatted keywords that still have to be fixed. The output file isn't that precious that we need to worry about overwriting it with garbage.

bryanwweber (approved):
Thanks, LGTM! ~Can't be merged while it's in draft though 🤪~ you took it out of draft in between my refreshes

PR #1236, opened 2022-04-04
Fix/deprecate 'from_pandas' interface

### LINKED ISSUE #1235: 1D Flame to/from pandas dataframe
**Problem description**

The to/from pandas dataframe methods, which use the to/from solution array methods, do not work as expected

**Steps to reproduce**

I created counterflow twin premixed diffusion flame and then wrote the solution vector as a dataframe. I then created a second flame and tried to initialize it with data from the data frame and got this error
image

**Behavior**
Screenshots below are from the "onedim.py" file in the cantera installation. Line 516 tries to pass a "restore_boundaries" argument to from_solution_array() but line 463 shows that from_solution_array() doesn't take this keyword argument

image

image

**System information**

- Cantera version: 2.5.1
- OS: Win10 
- Python

**Attachments**

**Additional context**

ischoegl:
*pandas* dataframes do not include information on boundaries and thus cannot
be used to recreate all information required to restore a simulation object.

- fix `from_pandas` keyword arguments
- deprecate `from_pandas` as it does not allow for a complete restoration of the object

Note that this API was used for a previous implementation of a pandas-based HDF export/import route, which has since been refactored.

Closes #1235

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/onedim.py
+++ b/interfaces/cython/cantera/onedim.py
@@ -472,7 +472,8 @@ class FlameBase(Sim1D):
 
     def from_pandas(self, df, restore_boundaries=True, settings=None):
         """
-        Restore the solution vector from a `pandas.DataFrame`.
+        Restore the solution vector from a `pandas.DataFrame`; currently disabled
+        (`save`/`restore` or `write_hdf`/`read_hdf` should be used as alternatives).
 
         :param df:
             `pandas.DataFrame` containing data to be restored
@@ -488,10 +489,9 @@ class FlameBase(Sim1D):
         requires a working *pandas* installation. The package ``pandas`` can be
         installed using pip or conda.
         """
-        arr = SolutionArray(self.gas, extra=self.other_components())
-        arr.from_pandas(df)
-        self.from_solution_array(arr, restore_boundaries=restore_boundaries,
-                                 settings=settings)
+        # @todo: Discuss implementation that allows for restoration of boundaries
+        raise NotImplementedError(
+            "Use 'save'/'restore' or 'write_hdf'/'read_hdf' as alternatives.")
 
     def write_hdf(self, filename, *args, group=None, species='X', mode='a',
                   description=None, compression=None, compression_opts=None,


interfaces/cython/cantera/onedim.py
```diff
@@ -487,11 +484,30 @@ def from_pandas(self, df, restore_boundaries=True, settings=None):
...
-        """
+
+        Note: As `to_pandas` does not save information on boundary conditions, those
+        values cannot be restored automatically and have to be specified by the user.
+
+        .. deprecated:: 2.6
+           *pandas* input will no longer be supported after Cantera 2.6 as it is not
+           capable of recreating all information required to restore the object. As
+           alternatives, `pandas` dataframes can be read into a `SolutionArray` and
+           passed to the flame domain using `from_solution_array`. For automatic
```
bryanwweber: As a naive person, I don't see how this is different from direct pandas input. What does `SolutionArray` do that pandas can't do directly? Can we modify `to_pandas` to include the boundary information?

ischoegl: I think that deprecating the method is preferable - it is essentially the same situation as for `write_csv` - we can write CSV files, but there is no direct way to import the data. 

All objects in the `onedim` category are composed of three domains: upstream/downstream plus the actual solution domain. Both `to_pandas` and `write_csv` only export the solution itself. From my perspective, if we decide to include boundary values here (which is doable) the interface will become a little more complex (we will have to save a row at the top/bottom that does not correspond to a location), with the actual implementation requiring handling of quite a few edge cases. The alternative is to just use what was specifically intended for saving/restoring full information on these objects (`save`/`restore` and `write_hdf`/`read_hdf`), which ultimately is a much safer route.

I'm not completely opposed to an implementation here, but I think what is done for pandas should also be done for CSV, and I am hesitant to roll out a new feature at the moment.

**PS:** Ultimately, I am afraid that users will keep asking why they get stuck importing from CSV / pandas, and not allowing this will avoid this situation.

bryanwweber: If it's a bug/broken, then it should be fixed or outright removed, IMO. If it's just not going to work at all, why bother warning? Just throw a clear error with the solution and be done with it.

~That said, I don't see how passing the output of `to_pandas` through `from_solution_array` fixes the problem of not having the boundary conditions.~ On rereading, I guess you're saying that the user should set up the `FreeFlame` and then pass the `SolutionArray` created from pandas to specifically set the state of the `flame` domain within the `FreeFlame`? I also see that this method is doing that, so why do we need to deprecate/remove this method? I think it would make more sense to ensure that the boundary conditions are set before loading, but still allow this behavior of loading from pandas DataFrames as long as that is satisfied.

Can you mention some of the edge cases you thought of in implementing `from_pandas`/`read_csv` with boundaries?

> All objects in the `onedim` category are composed of three domains: upstream/downstream plus the actual solution domain.

In principle, they can be composed of an arbitrary number of domains, I think; as a practical matter, the most convenient ones that we've implemented are composed of three domains 😄 

> I'm not completely opposed to an implementation here, but I think what is done for pandas should also be done for CSV, and I am hesitant to roll out a new feature at the moment.

I agree with this, but let's not shoot ourselves in the foot by removing things only to add them back later.

ischoegl:
@bryanwweber ... I'm responding outside of the feedback, as I believe there are some larger issues.

Fwiw, `from_pandas` is a vestigial method that is left over from the original HDF interface that built on pandas and was replaced by the current implementation between 2.4 and 2.5.1 (PR #840); the original HDF implementation was never released. Before #840, this function was an intermediate step of the HDF interface and was implicitly covered by the test suite. The only reason `from_pandas` still exists is that I overlooked it when removing the previous HDF incarnation. 

>  I guess you're saying that the user should set up the `FreeFlame` and then pass the `SolutionArray` created from pandas to specifically set the state of the flame domain within the `FreeFlame`? I also see that this method is doing that, so why do we need to deprecate/remove this method? 

While this is correct, I believe that having a dedicated function has the potential to create more confusion than actual benefit. Users will need to know some details about how these objects work, and it's not too much to ask to just write two extra lines in Python. Fwiw, CSV support is something that would be used in 95% of the cases, as I don't expect pandas to play a major role.

> let's not shoot ourselves in the foot by removing things only to add them back later.

I think that what is done here should ultimately be handled by `restore` ... even if this will require some heavy lifting on the C++ backend. At this point we need to start thinking about how to create `SolutionArray` equivalents that can be used from other API's, so I don't want to spend too much time creating yet another Python-centric feature. I am mostly referring to CSV support here, which `pandas` would piggy-back on.

**PS:** 

> In principle, they can be composed of an arbitrary number of domains, I think; as a practical matter, the most convenient ones that we've implemented are composed of three domains

This is of course correct, but it's not used. There are comments about this in [1D Flame Development Ideas](https://github.com/Cantera/cantera/projects/2) (or at least there were at some time); imho this is really just a tangential issue here.

interfaces/cython/cantera/onedim.py
ischoegl: > Can you mention some of the edge cases you thought of in implementing `from_pandas`/`read_csv` with boundaries?

Some of this was implemented prior to #840 (a working implementation can be seen in the deleted code segments of that PR, mostly in `FlameBase.to_solution_array` and `FlameBase.from_solution_array`) ... in order for this approach to work, compatible output needs to include rows that are not grid points (i.e. CSV and pandas output need to be prepped). I did not retain this as it won't work for `Surface` interactions (see `catalytic_combustion.py` example).

bryanwweber:
> The only reason `from_pandas` still exists is that I overlooked it when removing the previous HDF incarnation.

Then I think we should just raise a `NotImplementedError` for now and come back to this after the release.

decaluwe (changes requested):
I think this is the correct approach for now.  Only one suggestion I have: should the docstring lead with a notification that this feature is currently deprecated/disabled?  Seems like that would be most pertinent, before somebody decides to read/parse the rest of the docstring.

ischoegl:
@decaluwe ... fair point - docstring is updated per your suggestion.

decaluwe (approved):
Thanks, @ischoegl - LGTM.

PR #1238, opened 2022-04-05
Introduce cantera/core.h for C++ Examples

ischoegl:

- ~Add relevant headers to `Solution.h`~
- *Edit: Introduce `cantera/core.h`*
- Remove redundant headers from samples

Also:

-  Implement `Solution::setTransport(const std::string&)`

In many instances where `newSolution` handles the creation of a `Cantera::Solution` object, the main include needed in C++ would be
``` C++
#include "cantera/core.h"
```

~An alternative would be to introduce a new `core.h` header at the same level as `thermo.h` etc. (I.e. `#include "cantera/core.h"`)~ ... *Edit: this was adopted as an alternative to include common headers in `Solution.h`*

**Checklist**

### DIFF
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -44,8 +44,12 @@ public:
     //! Set the Kinetics object
     virtual void setKinetics(shared_ptr<Kinetics> kinetics);
 
-    //! Set the Transport object
+    //! Set the Transport object directly
     virtual void setTransport(shared_ptr<Transport> transport);
+
+    //! Set the Transport object by name
+    //! @param model  name of transport model
+    virtual void setTransport(const std::string& model);
 
     //! Accessor for the ThermoPhase pointer
     shared_ptr<ThermoPhase> thermo() {
--- a/include/cantera/core.h
+++ b/include/cantera/core.h
@@ -1,0 +1,18 @@
+/**
+ * @file core.h
+ *
+ * Support for Cantera core calculations from C++ application programs. This
+ * header file includes a minimal set of headers needed to create and use objects
+ * that evaluate thermo properties, chemical kinetics and transport properties.
+ */
+
+#ifndef CT_INCL_CORE_H
+#define CT_INCL_CORE_H
+
+// #include "cantera/base/global.h"
+#include "cantera/base/Solution.h"
+#include "cantera/thermo/ThermoPhase.h"
+#include "cantera/kinetics/Kinetics.h"
+#include "cantera/transport/TransportBase.h"
+
+#endif
--- a/include/cantera/core.h
+++ b/include/cantera/core.h
@@ -9,7 +9,6 @@
 #ifndef CT_INCL_CORE_H
 #define CT_INCL_CORE_H
 
-// #include "cantera/base/global.h"
 #include "cantera/base/Solution.h"
 #include "cantera/thermo/ThermoPhase.h"
 #include "cantera/kinetics/Kinetics.h"
--- a/samples/cxx/custom/custom.cpp
+++ b/samples/cxx/custom/custom.cpp
@@ -13,9 +13,7 @@
 // This file is part of Cantera. See License.txt in the top-level directory or
 // at https://cantera.org/license.txt for license and copyright information.
 
-#include "cantera/thermo.h"
-#include "cantera/kinetics.h"
-#include "cantera/base/Solution.h"
+#include "cantera/core.h"
 #include "cantera/numerics/Integrator.h"
 #include <fstream>
 
@@ -75,7 +73,7 @@ public:
         double *massFracs = &y[1];
         double *dTdt = &ydot[0];
         double *dYdt = &ydot[1];
-        
+
         /* ------------------------- UPDATE GAS STATE ------------------------- */
         // the state of the ThermoPhase is updated to reflect the current solution
         // vector, which was calculated by the integrator.
@@ -87,7 +85,7 @@ public:
         double cp = m_gas->cp_mass();
         m_gas->getPartialMolarEnthalpies(&m_hbar[0]);
         m_kinetics->getNetProductionRates(&m_wdot[0]);
-        
+
         /* -------------------------- ENERGY EQUATION ------------------------- */
         // the rate of change of the system temperature is found using the energy
         // equation for a closed-system constant pressure ideal gas:
@@ -110,7 +108,7 @@ public:
             dYdt[k] = m_wdot[k] * m_gas->molecularWeight(k) / rho;
         }
     }
-    
+
     /**
      * Number of equations in the ODE system.
      *   - overridden from FuncEval, called by the integrator during initialization.
@@ -153,7 +151,7 @@ int main() {
     // for each simulation time point.
     // create the csv file, overwriting any existing ones with the same name.
     std::ofstream outputFile("custom_cxx.csv");
-    
+
     // for convenience, a title for each of the state vector's components is written to
     // the first line of the csv file.
     outputFile << "time (s), temp (K)";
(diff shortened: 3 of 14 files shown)


samples/cxx/custom/custom.cpp
```diff
@@ -13,8 +13,6 @@
 // This file is part of Cantera. See License.txt in the top-level directory or
 // at https://cantera.org/license.txt for license and copyright information.
 
-#include "cantera/thermo.h"
-#include "cantera/kinetics.h"
```
bryanwweber, line 17: My understanding of the purpose of these headers is as a user-facing import point for the commonly used Cantera classes. This avoids users having to know about the folder structure inside `include/cantera`. I don't think we should remove the use of these.

include/cantera/base/Solution.h
```diff
@@ -8,14 +8,13 @@
 
 #include "cantera/base/ctexceptions.h"
 #include "cantera/base/AnyMap.h"
+#include "cantera/thermo/ThermoPhase.h"
+#include "cantera/kinetics/Kinetics.h"
+#include "cantera/transport/TransportBase.h"
```
bryanwweber: Adding these here means that this file will have to be recompiled any time one of these dependencies changes. It's a small change to compiling times probably, but I don't think it's necessary here.

bryanwweber (changes requested):
Per my comments below, IMO the preferred method to make `newSolution` more accessible to users would be to add a new header, say, `solution.h`.

include/cantera/base/Solution.h
ischoegl: Removed. Going with a new header file `cantera/core.h` instead.

ischoegl:
@bryanwweber / @speth ... thank you for the feedback. 

> Per my comments below, IMO the preferred method to make `newSolution` more accessible to users would be to add a new header, say, `solution.h`.

I had already eyed that idea (see comments on top). Ultimately, I believe 
```C++
#include "cantera/core.h"
```
is the most intuitive nomenclature - which is now implemented (it combines all commonly used features of what I consider the core - `Solution`, `ThermoPhase`, `Kinetics` and `TransportBase`).

decaluwe (commented):
Thanks, @ischoegl, this seems pretty non-controversial, and really cleans up the include headers in some of the examples!

Naive question: You have added `core.h` to the `zerodim` and `onedim` headers, but I noticed that there are no lines that get taken out (i.e. these files did not previously include any of the include files listed in `core.h`.  Am I understanding that correctly?  And if so, why is it necessary to include `core.h` now?  (note: this is not criticism, more just me trying to understand).

Edit: Never mind, I think I figured it out.  Including it here cleans up header statements for things like `Combustor.cpp` and `kinetics1.cpp`, which already include `zerodim`, et al.  Is that correct?

include/cantera/core.h
```diff
@@ -0,0 +1,23 @@
...
+#include "cantera/base/Solution.h"
+#include "cantera/thermo/ThermoPhase.h"
+#include "cantera/kinetics/Kinetics.h"
+
+#ifndef CT_SKIP_TRANSPORT
+// TransportFactory facilitates instantiation of alternative transport objects
+// @todo implement convenient way to switch transport model within Solution objects
+//      so dependency can be simplified to transport/TransportBase.h
+#include "cantera/transport/TransportFactory.h"
+#endif
```
bryanwweber: I don't think we should do this. This precludes transport calculations in the same file as a Reactor Net if zerodim.h is included. Does this significantly improve compilation time?

ischoegl: I don't think that it has any major effects. Will take it out ...

ischoegl:
@decaluwe ... glad you like it!

>  Including it here cleans up header statements for things like `Combustor.cpp` and `kinetics1.cpp`, which already include `zerodim`, et al. Is that correct?

Correct. I view `core.h` as something that is needed for most Cantera applications. Beyond, `zerodim` and `onedim` are expansions that build on top of this.

ischoegl:
Per discussion elsewhere, it may make sense to postpone this feature until after the 2.6 release. I will nevertheless keep the PR open as I think the feature is worthwhile.

ischoegl:
This should be one of the easier PR's left over from ongoing work prior to 2.6. Now that we're on our way to 3.0, this should be good for a review.

I ended up adding `Solution::setTransport(const std::string&)`, which I believe should be a helpful addition.

~**PS:** I'm a little torn about including `cantera/base/global.h` in `core.h`, as most examples don't use it. At the same time, it is useful for logging, warnings, etc. (and it will be a lot shorter after removal of CTI/XML)~ _Edit: removed._

ischoegl:
@speth / @bryanwweber … rebased and ready for review.

samples/cxx/gas_transport/gas_transport.cpp
```diff
@@ -13,11 +13,10 @@
 // This file is part of Cantera. See License.txt in the top-level directory or
 // at https://cantera.org/license.txt for license and copyright information.
 
-#include "cantera/thermo.h"
-#include "cantera/transport.h"
-#include "cantera/base/Solution.h"
+#include "cantera/core.h"
 #include "cantera/base/Array.h"
 
+#include <fstream>
```
speth: `fstream` is already included two lines below.

samples/cxx/jacobian/derivative_speed.cpp
```diff
@@ -17,9 +17,7 @@
 #include <iostream>
 #include <iomanip>
 #include <numeric>
-#include "cantera/base/Solution.h"
-#include "cantera/thermo/IdealGasPhase.h"
-#include "cantera/kinetics.h"
+#include "cantera/core.h"
```
speth, line 20: I think the include graph for `core.h` already gets `cantera/numerics/eigen_sparse.h` (included below).

src/base/Solution.cpp
```diff
@@ -51,6 +51,18 @@ void Solution::setTransport(shared_ptr<Transport> transport) {
     m_transport = transport;
 }
 
+void Solution::setTransport(const std::string& model) {
+    if (model == "") {
+        setTransport(shared_ptr<Transport>(
+            newDefaultTransportMgr(m_thermo.get())));
+    } else if (model == "None") {
+        setTransport(shared_ptr<Transport>(newTransportMgr("None")));
```
speth: Is this special case needed?

speth (changes requested):
Thanks, @ischoegl. Overall, I think this looks good. In a way, this is a better thought out version of the `Cantera.h` header that used to exist before around Cantera 2.2.

samples/cxx/gas_transport/gas_transport.cpp
ischoegl: whoops

src/base/Solution.cpp
ischoegl: I kept this special case as it was already included in the passage I moved over from `newSolution` (which now calls this new method). If I recall, the original point was to not have dangling pointers within the `Solution` object, so all exceptions can be thrown as needed.

*Edit: I am checking whether it can be deleted, as the regular call may already cover this.*

samples/cxx/jacobian/derivative_speed.cpp
ischoegl: indeed!

src/base/Solution.cpp
ischoegl: Alright. The short answer is: no, special case is not needed :tada:

ischoegl:
@speth ... thanks for the review! I think this is ready ...

speth (approved):

ischoegl:
@bryanwweber ... do you approve?

bryanwweber (approved):

PR #1242, opened 2022-04-14
Add `SolutionArray :set_mixture_fraction` method.

decaluwe:

- Adds the capability to set the mixture fraction of a solutionArray.

This largely follows the method already implemented for `solutionArray:set_equivalence_ratio`.

Addresses Cantera/enchancements#147

As noted in the enhancement request, an example implementation is:
```python
import cantera as ct

yfuel = {'C': 43.41, 'H': 5.82, 'O': 44.32, 'N': 0.25}
yair = {'O2': 23, 'N2': 76, 'Ar': 1}
mixfrac = np.linspace(0.1, 0.9, 20)

states = ct.SolutionArray(gas, shape=len(mixfrac))
states.TP = 900, ct.one_atm
states.set_mixture_fraction(mixfrac, yfuel, yair, basis='mass')
states.equilibrate('TP')
```

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -1001,6 +1001,25 @@ class SolutionArray:
             self._phase.set_equivalence_ratio(phi[index], *args, **kwargs)
             self._states[index][:] = self._phase.state
 
+    def set_mixture_fraction(self, mixture_fraction, *args, **kwargs):
+        """
+        See `ThermoPhase.set_mixture_fraction`
+
+        Note that ``mixture_fraction`` either needs to be a scalar value or 
+        dimensions have to be matched to the `SolutionArray`.
+        """
+
+        # broadcast argument shape
+        mixture_fraction, _ = np.broadcast_arrays(mixture_fraction, 
+            self._output_dummy)
+
+        # loop over values
+        for index in self._indices:
+            self._phase.state = self._states[index]
+            self._phase.set_mixture_fraction(mixture_fraction[index], *args, 
+                **kwargs)
+            self._states[index][:] = self._phase.state
+
     def collect_data(self, cols=None, tabular=False, threshold=0, species=None):
         """
         Returns the data specified by ``cols`` in an ordered dictionary, where
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -992,7 +992,9 @@ class SolutionArray:
         to be matched to the `SolutionArray`.
         """
 
-        # If ``phi`` is a scalar, broadcast it to the correct argument shape:
+        # If ``phi`` is lower-dimensional than the SolutionArray's  shape (for 
+        # example, a scalar), broadcast it to have the right number of 
+        # dimensions.
         phi, _ = np.broadcast_arrays(phi, self._output_dummy)
 
         for index in self._indices:
@@ -1008,8 +1010,9 @@ class SolutionArray:
         dimensions have to be matched to the `SolutionArray`.
         """
 
-        # If ``mixture_fraction`` is a scalar, broadcast it to the correct 
-        # argument shape:
+        # If ``mixture_fraction`` is lower-dimensional than the SolutionArray's 
+        # shape (for example, a scalar), broadcast it to have the right number 
+        # of dimensions.
         mixture_fraction, _ = np.broadcast_arrays(mixture_fraction, 
             self._output_dummy)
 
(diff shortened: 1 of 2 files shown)


decaluwe:
@ischoegl I want to add a test to cover this (and will cover `set_equivalence_ratio` as well), then it should be ready for review.

interfaces/cython/cantera/composite.py
```diff
@@ -1001,6 +1001,25 @@ def set_equivalence_ratio(self, phi, *args, **kwargs):
...
+
+        Note that ``mixture_fraction`` either needs to be a scalar value or 
+        dimensions have to be matched to the `SolutionArray`.
+        """
+
+        # broadcast argument shape
+        mixture_fraction, _ = np.broadcast_arrays(mixture_fraction, 
+            self._output_dummy)
+
+        # loop over values
```
bryanwweber: A nit, but this comment and the one above don't add anything to the explanation. I see they're copied from the other method, but I don't see a reason to copy them over.

decaluwe: Certainly for line 1016.  

For line 1013, it is not entirely clear to me what line 1012 does.  My best guess: if `mixture_fraction` is a scalar, we convert it to an array of the correct size, where are elements are set to the value `mixture_fraction`. Is that correct?

decaluwe: To put a finer point on it: @bryan, would you be supportive of me adding a comment that actually explains what line 1012 does (while deleting the comment on line 1016)?

bryanwweber: Yes, I think that would be perfect, thank you 😊

decaluwe: Okay, but so: is my understanding above of line 1012 correct?  😂

interfaces/cython/cantera/composite.py
```diff
@@ -992,15 +992,33 @@ def set_equivalence_ratio(self, phi, *args, **kwargs):
         to be matched to the `SolutionArray`.
         """
 
-        # broadcast argument shape
+        # If ``phi`` is a scalar, broadcast it to the correct argument shape:
```
bryanwweber: Thanks for taking a shot at improving this! I think `phi ` can be any shape that's compatible with `self._output_dummy`. I _think_ the key is that `phi` needs to have a lower dimensionality than the other array, but I'm not totally sure. I'd have to read the docs for numpy to see, I can never remember how broadcasting works.

The one-line version in my head is

> Broadcast `phi` to match the output shape

But I don't think that adds much either. It's basically restating the name of the function without any more explanation about _why_ this needs to be done.

I think I'll try to grok the Numpy docs and see if there's an alternative that comes to mind.

decaluwe:
Thanks—I read the docs earlier, even looked at some examples, and _still_ can’t figure out what/why it does 😂

interfaces/cython/cantera/composite.py
ischoegl: Fwiw, and if memory serves, I used this function upon suggestion when implementing the set equivalence ratio method. I believe @bryanwweber is describing the behavior correctly: it just makes sure that dimensions are matched. We mostly use 1d where things are simple. For 2d and above, broadcasting adds dimensions based on established rules.

speth: > But I don't think that adds much either. It's basically restating the name of the function without any more explanation about why this needs to be done.

You could write:
```
If phi is lower-dimensional than the SolutionArray's shape (for example, a scalar),
broadcast it to have the right number of dimensions.
```

ischoegl:
> I want to add a test to cover this (and will cover `set_equivalence_ratio` as well), then it should be ready for review.

@decaluwe … this looks really good! I think once the unit test is added, this is ready to go.

decaluwe:
Thanks, @ischoegl  - just added! Once it passes build 🤞, please have a look!

bryanwweber:
Thanks @decaluwe I made some changes to the `test_set_mixture_fraction()` methods. Please feel free to drop those changes if you don't agree (I figured this would be faster than going back-and-forth in a review).

ischoegl (approved):
@decaluwe thanks! ... also thanks to @bryanwweber for beating me to a review. I think this is ready ... :shipit:

bryanwweber:
All green, in it goes! Thanks @decaluwe!

decaluwe:
Thanks for the help, @bryanwweber. Certainly agree, re: just making the changes directly rather than a back and forth. I'm not particularly territorial about these things :)

PR #1239, opened 2022-04-06
Prevent aggressive pip uninstallation of existing package

speth:
**Changes proposed in this pull request**

- Determine whether the Python module is being re-installed to the same location. If not, add the `--ignore-installed` option to `pip install`

**If applicable, fill in the issue number this pull request is fixing**

Fixes #1230

**If applicable, provide an example illustrating new features this pull request is introducing**

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

Issue #1230, opened 2022-03-27
Install procedure tries to uninstall Cantera-python version installed by system pkg manager


band-a-prend:
**System information**

- Cantera version: 2.6.0b1
- OS: Gentoo linux
- Python: 3.9, 3.10

**Problem description**

New implemented Cantera-python installation procedure uses `pip install` with `--force-reinstall` option that tries to uninstall the Cantera version that was early placed by system package manager into system python site-packages directory. This attempt results in an error as system package manager blocks this attempt.

E.g. Gentoo package manager (portage) initially tries to install all files to specified `stage_dir` and then place it to system replacing early installed files if necessary. Thus the attempt to uninstall Cantera-python 2.5.1 files (installed by Gentoo package manager) by  `pip` is failed as Gentoo package manager blocks such try while installing Cantera-python 2.6.0b1 files into python site-packages under site_dir build directory.
Linux distributives that uses similar aproach (e.g. Debian, Ubuntu, Fedora etc) could be affected this issue too.

The quick workaround for distributive maintainers is to patch `interfaces/cython/SConscript` and/or `interfaces/python_minimal/SConscript` files replacing `--force-reinstall` substring with `--ignore-installed`.

Otherwise the conditional check is required within `interfaces/cython/SConscript` and `interfaces/python_minimal/SConscript`  to apply `--force-reinstall` if `stage_dir` option isn't used or `--ignore-installed` for `pip install` if `stage_dir` option is specified for Cantera scons build env.

bryanwweber:
Thanks for reporting this @band-a-prend! I'll look into it before `2.6.0b2`.

band-a-prend:
The initial description looks complicated even for me now. So I try to clarify.

1. I have installed Cantera 2.5.1 package from gentoo repository.
2. I tried to prepare Cantera 2.6.0b1 package for gentoo for testing before final 2.6.0 release.
3. On package manager installation phase of prepared package `scons install` is ran with specified `stage_dir` and `libdirname` options.
4. For installation of python Cantera implementation scons calls `pip` command.
5. On package manager installation phase the `pip` installs files into site-packages directory under sandboxed (isolated) `stage_dir` (it's OK) and then tries to uninstall Cantera 2.5.1 from system (and got access denied).
6. Gentoo package manager doesn't allow to directly change files in system by call of external tool like `pip`. It assumed that package manager merge files to system from stage_dir on it's own.

speth:
[...]

PR #1243, opened 2022-04-16
Fix RPATH related issues

speth:

- Set the Cantera shared library to use RPATH on macOS. Previously, the embedded path just pointed to the build directory, which would never work. Now, linking to this library will work if you specify the `-Wl,-rpath,/path/to/lib` option when linking an application against this library.
- Update `Cantera.mak` to add rpath flags when linking to Cantera or other libraries. This helps whenever these are shared libraries but not in standard system locations, for instance when linking against libraries in a conda installation. 
- Fix an unrelated issue reported in the Users' Group (see https://groups.google.com/g/cantera-users/c/3PsUAFqeEhw) related to the fix applied for #1230 in #1239. 

**Checklist**

### DIFF
--- a/interfaces/cython/SConscript
+++ b/interfaces/cython/SConscript
@@ -40,7 +40,7 @@ import site
 vars = get_config_vars()
 vars["plat"] = get_platform()
 vars["numpy_include"] = numpy.get_include()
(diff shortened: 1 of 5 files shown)


ischoegl (approved):
Looks good to me.

PR #1256, opened 2022-04-28
Fix bug with electrochemical-legacy

dschmider-HSOG:

- Change alias in ReactionFactory.cpp to enable correct legacy behavior for electrochemical reactions
- Fix problem described in [Cantera Users' Group](https://groups.google.com/g/cantera-users/c/KnHmndNoqu4)

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/test/test_kinetics.py
+++ b/interfaces/cython/cantera/test/test_kinetics.py
@@ -910,6 +910,12 @@ class TestSofcKinetics2(TestSofcKinetics):
     _mech = "sofc2.yaml"
 
 
+@pytest.mark.usefixtures("allow_deprecated")
+class TestSofcKinetics3(TestSofcKinetics):
+    """ Test using legacy framework; included to retain coverage """
+    _mech = "sofc.cti"
+
+
 class TestDuplicateReactions(utilities.CanteraTest):
     infile = 'duplicate-reactions.yaml'
 
--- a/src/kinetics/ReactionFactory.cpp
+++ b/src/kinetics/ReactionFactory.cpp
@@ -220,7 +220,7 @@ ReactionFactoryXML::ReactionFactoryXML()
         setupElectrochemicalReaction(*(ElectrochemicalReaction2*)R, node);
         return R;
     });
-    addAlias("interface-legacy", "electrochemical");
+    addAlias("electrochemical-legacy", "electrochemical");
 }
 
 }


speth:
I think this change makes sense, but I'm a little concerned about what it says about how we're testing this reaction type to verify that the "legacy" and new implementations corresponding to CTI and YAML input are actually the same. If the test suite didn't catch this error, I think we need to take another look at those tests.

ischoegl:
Thanks, @dschmider-HSOG for tracking that bug down! This was a clear oversight that happened to fall through the cracks of the test suite.

@speth Regarding unit test coverage, both `electrochemical` and `electrochemical-legacy` were already covered from the YAML side. I didn't realize that we still had a version for  `sofc.cti` in `data/inputs`, which I am now also checking against. The pre-existing test is thorough, so I didn't feel it to be necessary to add things to `test_reaction.py`.

To facilitate a resolution here, I pushed an update to include testing of CTI input to this PR.

interfaces/cython/cantera/test/test_kinetics.py
```diff
@@ -910,6 +910,12 @@ class TestSofcKinetics2(TestSofcKinetics):
     _mech = "sofc2.yaml"
 
 
+@pytest.mark.usefixtures("allow_deprecated")
+class TestSofcKinetics3(TestSofcKinetics):
+    """ Test using legacy framework; included to retain coverage """
```
bryanwweber, line 915: Thanks @ischoegl! Can you add a note about when this should be removed? (Unless it's obvious already, I don't have as much context about the legacy reaction changes)

ischoegl: Every single `-legacy` reaction is slated to be removed immediately after the release of 2.6, so I believe it's not necessary to add anything beyond.

speth:
Thanks for adding this test, @ischoegl. The thing that I'm still surprised to see is that if you revert the change that fixes the reaction type, this new test still passes.

One difference that I did notice between @dschmider-HSOG's example posted on the Users' Group and `sofc.cti` is the use of the `exchangecurrentdensity` option in the former case:
```py
surface_reaction("Li+[electrolyte] + V[NCA] + electron <=> Li[NCA]",
                 [2.62895E+10, 0.0, (61.0098120252686, 'kJ/mol')],
                 rate_coeff_type = "exchangecurrentdensity",
                 beta = 0.5,
                 id="NCA_reaction")
```
I'm thinking we don't actually have a test that covers this case for the "legacy" reaction framework / CTI input, though the feature is used in `data/inputs/lithium_ion_battery.cti`.

ischoegl:
> I'm thinking we don't actually have a test that covers this case for the "legacy" reaction framework / CTI input, though the feature is used in `data/inputs/lithium_ion_battery.cti`.

Thanks, @speth. When refactoring, I relied on existing tests. From the implementation side, the `exchangecurrentdensity` is a relatively small portion where I actually was able to avoid major refactoring, as most of what is required is to set `InterfaceRateBase::m_exchangeCurrentDensityFormulation` (none of the implementation changed even if the 'plumbing' was reshuffled).

If there wasn't a test covering this behavior from the get-go this is a different issue. It's outside of my wheel house and should probably be added by a separate issue? Tagging @decaluwe here ...

ischoegl:
@speth ... I opened #1259 which should resolve a larger issue. I don't think this should be added to this PR as this goes well beyond the scope of what @dschmider-HSOG signed up for.

ischoegl:
Based on #1263, I'd suggest to merge here (fwiw, the new `sofc.cti` test does add coverage, while not covering the actual issue).

speth (approved):
Thanks for identifying this fix and submitting the PR, @dschmider-HSOG.

PR #1257, opened 2022-04-29
Enforce third-body concentration as argument to falloff_function

### LINKED ISSUE #1254: `FalloffRate.third_body_concentration` is not defined
In `FalloffRate.falloff_function()`, the `conc3b` parameter can be passed as `None`, in which case `self.third_body_concentration` is used. However, `FalloffRate.third_body_concentration` is not defined in the pyx file, so this raises an attribute error:

```python
import cantera as ct
r = gas.reaction(11).rate
r.falloff_function(300)
```

I think we can improve the user experience in this case, either by eliminating the default argument or raising a more useful error message.

https://github.com/Cantera/cantera/blob/a3944235b8a0cdb246fbd308de7baad0535f667d/interfaces/cython/cantera/reaction.pyx#L421-L427

ischoegl:

- Make third-body-concentration mandatory

The method is new in 2.6, and simply breaks out C++ code to the Python API.

Closes #1254

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/test/test_reaction.py
+++ b/interfaces/cython/cantera/test/test_reaction.py
@@ -486,6 +486,11 @@ class TestLindemannRate(FalloffRateTests, utilities.CanteraTest):
         high-P-rate-constant: {A: 7.4e+10, b: -0.37, Ea: 0.0}
         """
 
+    def test_falloff_function(self):
+        rate = self.from_parts()
+        # Falloff-function for Lindemann is unity by definition
+        assert np.isclose(rate.falloff_function(300, 0.), 1.)
+
 
 class TestTroeRate(FalloffRateTests, utilities.CanteraTest):
     # test Troe rate expressions
--- a/interfaces/cython/cantera/reaction.pyx
+++ b/interfaces/cython/cantera/reaction.pyx
@@ -418,12 +418,11 @@ cdef class FalloffRate(ReactionRate):
         def __set__(self, cbool activated):
             self.falloff.setChemicallyActivated(activated)
 
-    def falloff_function(self, double temperature, conc3b=None):
+    def falloff_function(self, double temperature, double conc3b):
         """
-        Evaluate the falloff function
+        Evaluate the falloff function based on temperature and third-body
+        concentration.
         """
-        if conc3b is None:
-            conc3b = self.third_body_concentration
         return self.falloff.evalF(temperature, conc3b)
 
 


interfaces/cython/cantera/test/test_reaction.py
```diff
@@ -486,6 +486,10 @@ class TestLindemannRate(FalloffRateTests, utilities.CanteraTest):
         high-P-rate-constant: {A: 7.4e+10, b: -0.37, Ea: 0.0}
         """
 
+    def test_falloff_function(self):
+        rate = self.from_parts()
+        assert rate.falloff_function(300, 0.) == 1.
```
bryanwweber:
```suggestion
        assert np.allclose(rate.falloff_function(300, 0.0), 1.0)
```

I think floating point comparisons shouldn't use `==`, but maybe it's OK in this case.

ischoegl: In this case, it's ok, as that specific falloff function is 1.0 by definition. But I agree in general. Let me know if you'd like to change it.

bryanwweber: The only concern I have is that `rate.falloff_function` returns `1.000000000001` as it goes through any calculations. If this isn't exercising any calculation, why have the test?

ischoegl: I think `pytest` has something better than `numpy.allclose` here. How does
```python
assert rate.falloff_function(300, 0.) == pytest.approx(1, abs=1e-16)
```
look to you?

bryanwweber: Why is that better than `np.allclose`? There are three options here, `math.isclose(a, b)`, `pytest.approx(a, b)`, and `np.isclose/allclose(a, b)`. I think we should be consistent throughout the test suite which one we use, maybe this should use `self.assertNear` until (if) we switch to `pytest`-style asserts?

ischoegl: > If this isn't exercising any calculation, why have the test?

I mainly added the test to ensure that the Python API is covered and won't show surprising behavior (again). The optional argument was left over from an earlier implementation that had the necessary infrastructure included.

ischoegl: Ok. settling for `np.isclose`. I'd like to keep the test, and I'm actually not sure that we actually test for unity elsewhere.

bryanwweber (approved):
LGTM, thanks @ischoegl! Will merge after the tests finish. Feel free to merge if I don't get to it first.

PR #1185, opened 2022-02-01
Yaml2ck script

tsikes:
**Changes proposed in this pull request**

- Convert a Cantera YAML mechanism into a CHEMKIN-style (CK) mechanism 
- Write (nearly) any solution object into a CK mechanism
- Can be called from command-line or within Python

**If applicable, fill in the issue number this pull request is fixing**

Closes Cantera/enhancements#52

**If applicable, provide an example illustrating new features this pull request is introducing**

```
yaml2ck [--input=filename_path]
        [--output_path=[filename_path for mech, thermo, transport]]
        [--reorder_reactions=bool]
        [--skip_thermo=bool]
        [--separate_thermo_file=bool]
        [--skip_transport=bool]
        [--no-validate]
```
Can also be called by 
```
from yaml2ck import convert_mech as soln2ck
soln2ck(solution, output_path='', reorder_reactions=False,
        skip_thermo=False, separate_thermo_file=False, 
        skip_transport=False)
```

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [ ] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [ ] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

PR #1266, opened 2022-04-30
[Species] Add molecular weight attribute

bryanwweber:
The molecular weight is computed for each species when it is created.
The Phase private attribute m_mwt is not changed.

This will be useful in general, but specifically for #1185. This does _not_ need to go in for 2.6.

**Checklist**

### DIFF
--- a/include/cantera/thermo/Elements.h
+++ b/include/cantera/thermo/Elements.h
@@ -92,41 +92,15 @@ const std::vector<std::string>& elementSymbols();
 //! @since New in version 3.0
 const std::vector<std::string>& elementNames();
 
-//! Get a map with the element symbols as keys and weights as values.
+//! Get a map with the element and isotope symbols and names as keys and weights as
+//! values.
 /*!
  * This is a constant in the application so it is only generated once
  * when it is first needed.
  *
  * @since New in version 3.0
  */
-const std::map<std::string, double>& elementSymbolToWeight();
-
-//! Get a map with the element names as keys and weights as values.
-/*!
- * This is a constant in the application so it is only generated once
- * when it is first needed.
- *
- * @since New in version 3.0
- */
-const std::map<std::string, double>& elementNameToWeight();
-
-//! Get a map with the isotope symbols as keys and weights as values.
-/*!
- * This is a constant in the application so it is only generated once
- * when it is first needed.
- *
- * @since New in version 3.0
- */
-const std::map<std::string, double>& isotopeSymbolToWeight();
-
-//! Get a map with the isotope names as keys and weights as values.
-/*!
- * This is a constant in the application so it is only generated once
- * when it is first needed.
- *
- * @since New in version 3.0
- */
-const std::map<std::string, double>& isotopeNameToWeight();
+const std::map<std::string, double>& elementWeights();
 
 //! Get the atomic weight of an element.
 /*!
--- a/include/cantera/thermo/Elements.h
+++ b/include/cantera/thermo/Elements.h
@@ -84,6 +84,40 @@ namespace Cantera
 //! stable state at 298.15 K and 1 bar.
 #define ENTROPY298_UNKNOWN -123456789.
 
+//! Get a vector of the atomic symbols of the elements defined in Cantera.
+const std::vector<std::string>& elementSymbols();
+
+//! Get a vector of the names of the elements defined in Cantera.
+const std::vector<std::string>& elementNames();
+
+//! Get a map with the element symbols as keys and weights as values.
+/*!
+ * This is a constant in the application so it is only generated once
+ * when it is first needed.
+ */
+const std::map<std::string, double>& elementSymbolToWeight();
+
+//! Get a map with the element names as keys and weights as values.
+/*!
+ * This is a constant in the application so it is only generated once
+ * when it is first needed.
+ */
+const std::map<std::string, double>& elementNameToWeight();
+
+//! Get a map with the isotope symbols as keys and weights as values.
+/*!
+ * This is a constant in the application so it is only generated once
+ * when it is first needed.
+ */
+const std::map<std::string, double>& isotopeSymbolToWeight();
+
+//! Get a map with the isotope names as keys and weights as values.
+/*!
+ * This is a constant in the application so it is only generated once
+ * when it is first needed.
+ */
+const std::map<std::string, double>& isotopeNameToWeight();
+
 //! Get the atomic weight of an element.
 /*!
  * Get the atomic weight of an element defined in Cantera by its symbol
--- a/include/cantera/thermo/Elements.h
+++ b/include/cantera/thermo/Elements.h
@@ -175,11 +175,11 @@ int getAtomicNumber(const std::string& ename);
 
 //! Get the number of named elements defined in Cantera.
 //! This array excludes named isotopes
-int numElementsDefined();
+size_t numElementsDefined();
 
 //! Get the number of named isotopes defined in Cantera.
 //! This array excludes the named elements
-int numIsotopesDefined();
+size_t numIsotopesDefined();
 
 } // namespace
 
--- a/include/cantera/thermo/Elements.h
+++ b/include/cantera/thermo/Elements.h
@@ -85,11 +85,11 @@ namespace Cantera
 #define ENTROPY298_UNKNOWN -123456789.
 
 //! Get a vector of the atomic symbols of the elements defined in Cantera.
-//! @since 3.0
+//! @since New in version 3.0
 const std::vector<std::string>& elementSymbols();
 
 //! Get a vector of the names of the elements defined in Cantera.
-//! @since 3.0
+//! @since New in version 3.0
 const std::vector<std::string>& elementNames();
 
 //! Get a map with the element symbols as keys and weights as values.
@@ -97,7 +97,7 @@ const std::vector<std::string>& elementNames();
  * This is a constant in the application so it is only generated once
  * when it is first needed.
  *
- * @since 3.0
+ * @since New in version 3.0
  */
 const std::map<std::string, double>& elementSymbolToWeight();
 
@@ -106,7 +106,7 @@ const std::map<std::string, double>& elementSymbolToWeight();
  * This is a constant in the application so it is only generated once
  * when it is first needed.
  *
- * @since 3.0
+ * @since New in version 3.0
  */
 const std::map<std::string, double>& elementNameToWeight();
 
@@ -115,7 +115,7 @@ const std::map<std::string, double>& elementNameToWeight();
  * This is a constant in the application so it is only generated once
  * when it is first needed.
  *
- * @since 3.0
+ * @since New in version 3.0
  */
 const std::map<std::string, double>& isotopeSymbolToWeight();
 
@@ -124,7 +124,7 @@ const std::map<std::string, double>& isotopeSymbolToWeight();
  * This is a constant in the application so it is only generated once
  * when it is first needed.
  *
- * @since 3.0
+ * @since New in version 3.0
  */
 const std::map<std::string, double>& isotopeNameToWeight();
 
--- a/include/cantera/thermo/Elements.h
+++ b/include/cantera/thermo/Elements.h
@@ -85,15 +85,19 @@ namespace Cantera
 #define ENTROPY298_UNKNOWN -123456789.
 
 //! Get a vector of the atomic symbols of the elements defined in Cantera.
+//! @since 3.0
 const std::vector<std::string>& elementSymbols();
 
 //! Get a vector of the names of the elements defined in Cantera.
+//! @since 3.0
 const std::vector<std::string>& elementNames();
 
 //! Get a map with the element symbols as keys and weights as values.
 /*!
  * This is a constant in the application so it is only generated once
  * when it is first needed.
+ *
+ * @since 3.0
  */
 const std::map<std::string, double>& elementSymbolToWeight();
 
@@ -101,6 +105,8 @@ const std::map<std::string, double>& elementSymbolToWeight();
 /*!
  * This is a constant in the application so it is only generated once
  * when it is first needed.
+ *
+ * @since 3.0
  */
 const std::map<std::string, double>& elementNameToWeight();
 
@@ -108,6 +114,8 @@ const std::map<std::string, double>& elementNameToWeight();
 /*!
  * This is a constant in the application so it is only generated once
  * when it is first needed.
+ *
+ * @since 3.0
  */
 const std::map<std::string, double>& isotopeSymbolToWeight();
 
@@ -115,6 +123,8 @@ const std::map<std::string, double>& isotopeSymbolToWeight();
 /*!
  * This is a constant in the application so it is only generated once
  * when it is first needed.
+ *
+ * @since 3.0
  */
 const std::map<std::string, double>& isotopeNameToWeight();
 
@@ -209,10 +219,12 @@ int getAtomicNumber(const std::string& ename);
 
 //! Get the number of named elements defined in Cantera.
 //! This array excludes named isotopes
+//! @since Type is `size_t` in Cantera 3.0
 size_t numElementsDefined();
 
 //! Get the number of named isotopes defined in Cantera.
 //! This array excludes the named elements
+//! @since Type is `size_t` in Cantera 3.0
 size_t numIsotopesDefined();
 
 } // namespace
--- a/include/cantera/thermo/Species.h
+++ b/include/cantera/thermo/Species.h
@@ -76,7 +76,7 @@ public:
      *
      * @since New in version 3.0
      */
-    void setMolecularWeight(double weight, bool compute=false);
+    void setMolecularWeight(double weight);
 
     shared_ptr<TransportData> transport;
 
--- a/include/cantera/thermo/Species.h
+++ b/include/cantera/thermo/Species.h
@@ -52,6 +52,13 @@ public:
     double size;
 
     //! The molecular weight [amu] of the species.
+    /*!
+     * Calculates and sets the molecular weight from the elemental composition of the
+     * species and element definitions in Elements.cpp, if the molcular weight is
+     * Undef.
+     *
+     * @since 3.0
+     */
     const double molecularWeight();
 
     //! Set the molecular weight of the species.
@@ -61,9 +68,13 @@ public:
      * method the first time the molecularWeight method is called if the species has
      * not been added to a phase.
      *
+     * @since 3.0
+     *
      * @param weight: The weight of this species to assign, unless `compute` is `true`
      * @param compute: If `true`, the weight will be computed from standard element
      * weights from Elements.cpp, and the value of `weight` is ignored.
+     * @exception If the molecular weight has already been set and the updated value
+     * is not close to the existing value.
      */
     void setMolecularWeight(double weight, bool compute=false);
 
--- a/include/cantera/thermo/Species.h
+++ b/include/cantera/thermo/Species.h
@@ -54,27 +54,27 @@ public:
     //! The molecular weight [amu] of the species.
     /*!
      * Calculates and sets the molecular weight from the elemental composition of the
-     * species and element definitions in Elements.cpp, if the molcular weight is
+     * species and element definitions in Elements.cpp, if the molecular weight is
      * Undef.
      *
-     * @since 3.0
+     * @since New in version 3.0
      */
-    const double molecularWeight();
+    double molecularWeight();
 
     //! Set the molecular weight of the species.
     /*!
      * Since phases can have custom element weights, the phase will always call this
      * method when a species is added to that phase. The species may also call this
-     * method the first time the molecularWeight method is called if the species has
+     * method the first time the molecularWeight() method is called if the species has
      * not been added to a phase.
-     *
-     * @since 3.0
      *
      * @param weight: The weight of this species to assign, unless `compute` is `true`
      * @param compute: If `true`, the weight will be computed from standard element
      * weights from Elements.cpp, and the value of `weight` is ignored.
-     * @exception If the molecular weight has already been set and the updated value
+     * @exception CanteraError If the molecular weight has already been set and the updated value
      * is not close to the existing value.
+     *
+     * @since New in version 3.0
      */
     void setMolecularWeight(double weight, bool compute=false);
 
--- a/include/cantera/thermo/Species.h
+++ b/include/cantera/thermo/Species.h
@@ -51,6 +51,22 @@ public:
     //! species, where it represents the number of sites occupied.
     double size;
 
+    //! The molecular weight [amu] of the species.
+    const double molecularWeight();
+
+    //! Set the molecular weight of the species.
+    /*!
+     * Since phases can have custom element weights, the phase will always call this
+     * method when a species is added to that phase. The species may also call this
+     * method the first time the molecularWeight method is called if the species has
+     * not been added to a phase.
+     *
+     * @param weight: The weight of this species to assign, unless `compute` is `true`
+     * @param compute: If `true`, the weight will be computed from standard element
+     * weights from Elements.cpp, and the value of `weight` is ignored.
+     */
+    void setMolecularWeight(double weight, bool compute=false);
+
     shared_ptr<TransportData> transport;
 
     //! Thermodynamic data for the species
@@ -58,6 +74,12 @@ public:
 
     //! Input parameters used to define a species, for example from a YAML input file.
     AnyMap input;
+
+protected:
+
+    //! The molecular weight of the species, in atomic mass units. Includes
+    //! electron mass for charged species.
+    double m_molecularWeight = Undef;
 };
 
 //! Create a new Species object from an AnyMap specification
@@ -65,7 +87,6 @@ unique_ptr<Species> newSpecies(const AnyMap& node);
 
 //! Generate Species objects for each item (an AnyMap) in `items`.
 std::vector<shared_ptr<Species>> getSpecies(const AnyValue& items);
-
 }
 
 #endif
--- a/include/cantera/thermo/Species.h
+++ b/include/cantera/thermo/Species.h
@@ -68,11 +68,7 @@ public:
      * method the first time the molecularWeight() method is called if the species has
      * not been added to a phase.
      *
-     * @param weight: The weight of this species to assign, unless `compute` is `true`
-     * @param compute: If `true`, the weight will be computed from standard element
-     * weights from Elements.cpp, and the value of `weight` is ignored.
-     * @exception CanteraError If the molecular weight has already been set and the updated value
-     * is not close to the existing value.
+     * @param weight: The weight of this species to assign
      *
      * @since New in version 3.0
      */
@@ -98,6 +94,7 @@ unique_ptr<Species> newSpecies(const AnyMap& node);
 
 //! Generate Species objects for each item (an AnyMap) in `items`.
 std::vector<shared_ptr<Species>> getSpecies(const AnyValue& items);
+
 }
 
 #endif
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -1234,17 +1234,72 @@ class TestSpecies(utilities.CanteraTest):
         self.assertEqual(c['H'], 4)
         self.assertNear(s.molecular_weight, 16.043)
 
+    def test_molecular_weight_with_unstable_element(self):
+        s = ct.Species("CPo", {"C": 1, "Po": 1})
+        with pytest.raises(ct.CanteraError, match="has no stable isotopes"):
+            s.molecular_weight
+
     def test_molecular_weight_with_electron(self):
         yaml = """
             name: Li+[elyt]
             composition: {Li: 1, E: -1}
-            thermo:
-              model: constant-cp
-              h0: -278.49 kJ/mol
-              s0: 13.4 J/mol/K
         """
         s = ct.Species.from_yaml(yaml)
         assert np.isclose(s.molecular_weight, 6.939451420091127)
+
+    def test_custom_element_weight_is_set_on_species(self):
+        gas = ct.Solution("ideal-gas.yaml", "element-override")
+        # Check that the molecular weight stored in the phase definition is the same
+        # as the one on the element
+        self.assertNear(gas["AR"].molecular_weights[0], gas.species("AR").molecular_weight)
+        # Check that the custom value is actually used
+        self.assertNear(gas.species("AR").molecular_weight, 36.0)
+
+    def test_species_can_be_added_to_phase(self):
+        s = ct.Species.from_dict({
+            "name": "AR2",
+            "composition": {"Ar": 2},
+            "thermo": {"model": "constant-cp", "h0": 100}
+        })
+        # Access the molecular weight to make sure it's been computed by the Species
+        self.assertNear(s.molecular_weight, 39.95 * 2)
+        # This should not cause an error because the existing Ar definition is the
+        # default molecular weight
+        self.gas.add_species(s)
+        self.assertNear(
+            self.gas["AR2"].molecular_weights[0],
+            self.gas.species("AR2").molecular_weight
+        )
+
+    def test_species_cannot_be_added_to_phase_custom_element(self):
+        s = ct.Species.from_dict({
+            "name": "AR2",
+            "composition": {"Ar": 2},
+            "thermo": {"model": "constant-cp", "h0": 100}
+        })
+        # Access the molecular weight to make sure it's been computed by the Species
+        self.assertNear(s.molecular_weight, 39.95 * 2)
+        gas = ct.Solution("ideal-gas.yaml", "element-override")
+        # The error here is because the weight of the Argon element has been changed in
+        # the phase definition, but the molecular weight of the species has already been
+        # computed
+        with pytest.raises(ct.CanteraError, match="Molecular weight.*changed"):
+            gas.add_species(s)
+
+    def test_species_cant_be_added_to_phase_custom_element(self):
+        s = ct.Species.from_dict({
+            "name": "AR2",
+            "composition": {"Ar": 2},
+            "thermo": {"model": "constant-cp", "h0": 100}
+        })
+        # DO NOT access the molecular weight to make sure it's not been computed by the Species
+        gas = ct.Solution("ideal-gas.yaml", "element-override")
+        gas.add_species(s)
+        self.assertNear(
+            gas["AR2"].molecular_weights[0],
+            gas.species("AR2").molecular_weight
+        )
+        self.assertNear(s.molecular_weight, gas.species("AR2").molecular_weight)
 
     def test_defaults(self):
         s = ct.Species('H2')
@@ -1639,13 +1694,6 @@ class TestMisc(utilities.CanteraTest):
         with self.assertRaisesRegex(ct.CanteraError, 'Unknown species'):
             gas.Y = 'h2:1.0, o2:1.0'
 
-        gas_cti = """ideal_gas(
-            name="gas",
-            elements=" S C Cs ",
-            species=" nasa: all ",
-            options=["skip_undeclared_elements"],
-            initial_state=state(temperature=300, pressure=(1, "bar"))
-        )"""
         gas_yaml = """
             phases:
             - name: gas
@@ -1661,6 +1709,19 @@ class TestMisc(utilities.CanteraTest):
         gas.case_sensitive_species_names = True
         with self.assertRaises(ValueError):
             gas.species_index('cs')
+
+    def test_unstable_element_in_phase(self):
+        gas_yaml = """
+            phases:
+            - name: gas
+              thermo: ideal-gas
+              elements: [Po]
+            species:
+            - name: Po
+              composition: {Po: 1}
+        """
+        with pytest.raises(ct.CanteraError, match="has no stable isotopes"):
+            ct.Solution(yaml=gas_yaml)
 
 
 class TestElement(utilities.CanteraTest):
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -1251,7 +1251,9 @@ class TestSpecies(utilities.CanteraTest):
         gas = ct.Solution("ideal-gas.yaml", "element-override")
         # Check that the molecular weight stored in the phase definition is the same
         # as the one on the element
-        self.assertNear(gas["AR"].molecular_weights[0], gas.species("AR").molecular_weight)
+        self.assertNear(
+            gas["AR"].molecular_weights[0], gas.species("AR").molecular_weight
+        )
         # Check that the custom value is actually used
         self.assertNear(gas.species("AR").molecular_weight, 36.0)
 
@@ -1286,13 +1288,14 @@ class TestSpecies(utilities.CanteraTest):
         with pytest.raises(ct.CanteraError, match="Molecular weight.*changed"):
             gas.add_species(s)
 
-    def test_species_cant_be_added_to_phase_custom_element(self):
+    def test_species_can_be_added_to_phase_custom_element(self):
         s = ct.Species.from_dict({
             "name": "AR2",
             "composition": {"Ar": 2},
             "thermo": {"model": "constant-cp", "h0": 100}
         })
-        # DO NOT access the molecular weight to make sure it's not been computed by the Species
+        # DO NOT access the molecular weight to make sure it's not been computed by the
+        # Species
         gas = ct.Solution("ideal-gas.yaml", "element-override")
         gas.add_species(s)
         self.assertNear(
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -1232,6 +1232,19 @@ class TestSpecies(utilities.CanteraTest):
         self.assertEqual(len(c), 2)
         self.assertEqual(c['C'], 1)
         self.assertEqual(c['H'], 4)
+        self.assertNear(s.molecular_weight, 16.043)
+
+    def test_molecular_weight_with_electron(self):
+        yaml = """
+            name: Li+[elyt]
+            composition: {Li: 1, E: -1}
+            thermo:
+              model: constant-cp
+              h0: -278.49 kJ/mol
+              s0: 13.4 J/mol/K
+        """
+        s = ct.Species.from_yaml(yaml)
+        assert np.isclose(s.molecular_weight, 6.939451420091127)
 
     def test_defaults(self):
         s = ct.Species('H2')
--- a/interfaces/cython/cantera/test/test_thermo.py
+++ b/interfaces/cython/cantera/test/test_thermo.py
@@ -1250,7 +1250,7 @@ class TestSpecies(utilities.CanteraTest):
     def test_custom_element_weight_is_set_on_species(self):
         gas = ct.Solution("ideal-gas.yaml", "element-override")
         # Check that the molecular weight stored in the phase definition is the same
-        # as the one on the element
+        # as the one on the Species instance
         self.assertNear(
             gas["AR"].molecular_weights[0], gas.species("AR").molecular_weight
         )
@@ -1265,15 +1265,15 @@ class TestSpecies(utilities.CanteraTest):
         })
         # Access the molecular weight to make sure it's been computed by the Species
         self.assertNear(s.molecular_weight, 39.95 * 2)
-        # This should not cause an error because the existing Ar definition is the
-        # default molecular weight
+        # This should not cause a warning because the Ar element definition in
+        # self.gas is the same as the default
         self.gas.add_species(s)
         self.assertNear(
             self.gas["AR2"].molecular_weights[0],
             self.gas.species("AR2").molecular_weight
         )
 
-    def test_species_cannot_be_added_to_phase_custom_element(self):
+    def test_species_warns_when_changing_molecular_weight(self):
         s = ct.Species.from_dict({
             "name": "AR2",
             "composition": {"Ar": 2},
@@ -1282,10 +1282,11 @@ class TestSpecies(utilities.CanteraTest):
         # Access the molecular weight to make sure it's been computed by the Species
         self.assertNear(s.molecular_weight, 39.95 * 2)
         gas = ct.Solution("ideal-gas.yaml", "element-override")
-        # The error here is because the weight of the Argon element has been changed in
+        # The warning here is because the weight of the Argon element has been changed in
         # the phase definition, but the molecular weight of the species has already been
-        # computed
-        with pytest.raises(ct.CanteraError, match="Molecular weight.*changed"):
+        # computed, so loading the phase definition changes the molecular weight on the
+        # species.
+        with pytest.warns(UserWarning, match="Molecular weight.*changing"):
             gas.add_species(s)
 
     def test_species_can_be_added_to_phase_custom_element(self):
@@ -1294,8 +1295,8 @@ class TestSpecies(utilities.CanteraTest):
             "composition": {"Ar": 2},
             "thermo": {"model": "constant-cp", "h0": 100}
         })
-        # DO NOT access the molecular weight to make sure it's not been computed by the
-        # Species
+        # DO NOT access the molecular weight on the Species instance before adding it
+        # to the phase to make sure the weight has not been computed by the Species
         gas = ct.Solution("ideal-gas.yaml", "element-override")
         gas.add_species(s)
         self.assertNear(
@@ -1736,6 +1737,7 @@ class TestElement(utilities.CanteraTest):
         cls.ar_num = ct.Element(18)
 
     def test_element_multiple_possibilities(self):
+        # Carbon starts with Ca, the symbol for calcium.
         carbon = ct.Element('Carbon')
         self.assertEqual(carbon.name, 'carbon')
         self.assertEqual(carbon.symbol, 'C')
--- a/interfaces/cython/cantera/thermo.pyx
+++ b/interfaces/cython/cantera/thermo.pyx
@@ -179,7 +179,7 @@ cdef class Species:
     property molecular_weight:
         """The molecular weight [amu] of the species.
 
-        ..versionadded:: 3.0
+        .. versionadded:: 3.0
         """
         def __get__(self):
             return self.species.molecularWeight()
--- a/interfaces/cython/cantera/thermo.pyx
+++ b/interfaces/cython/cantera/thermo.pyx
@@ -176,6 +176,11 @@ cdef class Species:
         def __get__(self):
             return self.species.size
 
+    property molecular_weight:
+        """ The molecular weight [amu] of the species. """
+        def __get__(self):
+            return self.species.molecularWeight()
+
     property thermo:
         """
         Get/Set the species reference-state thermodynamic data, as an instance
@@ -2054,6 +2059,7 @@ def _element_symbols():
 def _element_names():
     names = elementNames()
     return tuple(pystr(n) for n in names)
+
 
 class Element:
     """
--- a/interfaces/cython/cantera/thermo.pyx
+++ b/interfaces/cython/cantera/thermo.pyx
@@ -2050,7 +2050,9 @@ cdef class PureFluid(ThermoPhase):
         def __get__(self):
             return self.s, self.v, self.Q
 
-
+# TODO: Remove these helper methods when support for Python 3.8 is dropped. Python 3.9
+# allows the classmethod and property decorators to be chained, so these can be
+# implemented as properties in the Element class.
 def _element_symbols():
     syms = elementSymbols()
     return tuple(pystr(s) for s in syms)
--- a/interfaces/cython/cantera/thermo.pyx
+++ b/interfaces/cython/cantera/thermo.pyx
@@ -2046,6 +2046,15 @@ cdef class PureFluid(ThermoPhase):
             return self.s, self.v, self.Q
 
 
+def _element_symbols():
+    syms = elementSymbols()
+    return tuple(pystr(s) for s in syms)
+
+
+def _element_names():
+    names = elementNames()
+    return tuple(pystr(n) for n in names)
+
 class Element:
     """
     An element or a named isotope defined in Cantera.
@@ -2092,13 +2101,11 @@ class Element:
 
     #: A list of the symbols of all the elements (not isotopes) defined
     #: in Cantera
-    element_symbols = [pystr(getElementSymbol(<int>(m+1)))
-                       for m in range(num_elements_defined)]
+    element_symbols = _element_symbols()
 
     #: A list of the names of all the elements (not isotopes) defined
     #: in Cantera
-    element_names = [pystr(getElementName(<int>m+1))
-                     for m in range(num_elements_defined)]
+    element_names = _element_names()
 
     def __init__(self, arg):
         if isinstance(arg, (str, bytes)):
--- a/interfaces/cython/cantera/thermo.pyx
+++ b/interfaces/cython/cantera/thermo.pyx
@@ -177,7 +177,10 @@ cdef class Species:
             return self.species.size
 
     property molecular_weight:
-        """ The molecular weight [amu] of the species. """
+        """The molecular weight [amu] of the species.
+
+        ..versionadded:: 3.0
+        """
         def __get__(self):
             return self.species.molecularWeight()
 
--- a/src/thermo/Elements.cpp
+++ b/src/thermo/Elements.cpp
@@ -58,15 +58,15 @@ struct isotopeWeightData {
 };
 
 /*!
- * @var static struct atomicWeightData atomicWeightTable[]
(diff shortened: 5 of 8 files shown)


bryanwweber:
Posted a question about the test failures on the UG: https://groups.google.com/g/cantera-users/c/mdNdCulK4gc/m/3G53KMDzAAAJ

ischoegl:
I think you need to implement a `Species::setComposition` method in C++ that also calculates the weight, similar to what is done in `newSpecies` (which won't be called as it is only used for the YAML interface).

I also replied on the [UG](https://groups.google.com/g/cantera-users/c/mdNdCulK4gc/m/flqOavv3AAAJ)

speth:
Can you explain more about why this is a necessary / useful feature? I don't quite understand what role it plays in implementing `yaml2ck`. One reason that the molecular weight isn't already a property of the `Species` object is because in the most general case (which includes custom element definitions) it can't be determined without information that's only held by the `ThermoPhase` object.

bryanwweber:
Thanks @speth! Good question. It will be useful to simplify sorting species in the CK output. I also think it's a logical place to keep the molecular weight... ideally, a species knows its own data. It's where I'd expect to find it if I had only a `Species` object.

That said, one of my test failures is because I'm not handling custom elements right now. I think it would make sense to raise an error if a custom element is present and the species is not part of a phase. Then the phase can set the molecular weight when the species is added. What do you think?

ischoegl:
@speth and @bryanwweber ... I actually agree that having `Species` know their molecular weights adds a feature of interest. I can think of at least one instance where not having this available for species forced me to use some kludgy work-arounds.

Per implementation, I see that custom elements add a wrinkle to what I said on the [UG](https://groups.google.com/g/cantera-users/c/mdNdCulK4gc/m/QqJ15QsGAQAJ). My preferred solution here would still be to strengthen the C++ constructor for `Species`, i.e.

> [...] addressing things at the root, by adding a `Species(const std::string&, const Composition&, double, double)` constructor in C++, exposing it to Cython, and calling it in `__cinit__` instead of the default constructor. This constructor would then call the existing `Species(const std::string&, const compositionMap&, double, double)`, which has a more restrictive interface. [...] the pre-existing constructor (and not `newSpecies`) should handle the calculation of molecular weights

I haven't looked into details about custom elements. As these are just computational entities/placeholders, they presumably don't need a weight, so there isn't an issue? Regarding species that contain (custom or regular) elements that aren't part of the phase, this is absolutely an error, and can be handled separately from the instantiation.

speth:
I'm unclear on what the extra constructor is supposed to do differently -- `Composition` and `compositionMap` are both just typedefs for `std::map<std::string, double>`. In any case, I agree that this calculation does not belong in the `newSpecies` function. The `Species` constructor is one option, but `ThermoPhase::addSpecies` is the only place where you're guaranteed to have enough information to do so.

> It will be useful to simplify sorting species in the CK output.

Don't you have access to a `ThermoPhase` object containing the `Species` when generating this output?

> I also think it's a logical place to keep the molecular weight... ideally, a species knows its own data.

> I haven't looked into details about custom elements. As these are just computational entities/placeholders, they presumably don't need a weight, so there isn't an issue?

That's one use case. Another is handling isotopes, where of course you do want to include the different value of the atomic weight.

> Regarding species that contain (custom or regular) elements that aren't part of the phase, this is absolutely an error, and can be handled separately from the instantiation.

This is already handled, of course. The default behavior for "regular" elements is to just add them to the phase, but this can be changed by calling one of `Phase::ignoreUndefinedElements()`, `Phase::addUndefinedElements()`, or `Phase::throwUndefinedElements()`.

> I think it would make sense to raise an error if a custom element is present and the species is not part of a phase.

So how do you create such a `Species` and add it to a phase?

ischoegl:
> I'm unclear on what the extra constructor is supposed to do differently -- `Composition` and `compositionMap` are both just `typedefs for std::map<std::string, double>`. 

Ugh. That is correct - didn't track down the definitions

https://github.com/Cantera/cantera/blob/9573e6b1e1097dca3bf86da37726333ab62663bb/include/cantera/base/ct_defs.h#L172-L180

In this case, I honestly think that there is one `typedef` too many.

So ... no new constructor needed. Good point about isotopes, and it frankly would have been surprising if there hadn't already been exception handling for undefined elements.

Either way, my personal preference here would be to strengthen the constructor for `Species` ...

bryanwweber:
> Don't you have access to a `ThermoPhase` object containing the `Species` when generating this output?

Yes, but it feels strange to write:

```python
solution = ct.Solution(...)
species = {s.name: s for s in solution.species()}
dict(sorted(species.items(), key=lambda s: solution[s[0]].molecular_weights))
# or
dict(sorted(species, key=lambda s: solution.molecular_weight(solution.species_index(s[0]))))
```
or something like that, when it could be:

```python
sorted(solution.species(), key=operator.attrgetter("molecular_weight"))
```

since, at least in ~this~ the `yaml2ck` script, we'll always have the complete phase.

> So how do you create such a `Species` and add it to a phase?

I meant throw if someone tries to read/get the molecular weight when the `Species` is not attached to a `Phase` (possibly also only if there are unknown elements involved). Presumably, once the species is attached to a `Phase`, the phase can tell it "This is your molecular weight".

speth:
> I meant throw if someone tries to read/get the molecular weight when the `Species` is not attached to a `Phase` (possibly also only if there are unknown elements involved). Presumably, once the species is attached to a `Phase`, the phase can tell it "This is your molecular weight".

If you want to just let the `Phase` object set the value stored by the `Species` when `Phase::addSpecies` is called, I think I'd withdraw my objection to adding the molecular weight to the `Species` object. That would avoid the possibility of introducing any inconsistency between the two values.

ischoegl:
@bryanwweber ... I had the exact same issue (minus resorting to `lambda`), i.e. 
```python
solution = ct.Solution(...)
species = {s.name: s for s in solution.species()}
dict(sorted(species.items(), key=lambda s: solution[s[0]].molecular_weights))
```
while wanting to implement a dictionary comprehension.

@speth ... I still don't fully understand what the hangup is about phases needing to set molecular weights (instead of species themselves). Can you direct me to a YAML input file where this shows up? (I mention YAML and not C++ here intentionally)

bryanwweber:
> That would avoid the possibility of introducing any inconsistency between the two values.

@speth I _think_ it would be useful to be able to get the molecular weight of an unattached species, provided it used only the elements/isotopes defined in `Elements.h`. However, I recognize that this introduces some further complexity to the situation that may not be worth the effort (that is, there's the risk of values becoming inconsistent).

I think this can be resolved by introducing a new/modified constructor as @ischoegl suggested, or perhaps a setter/getter method pair... The question to me is what the signature should look like to be most clear. Or, put another way, how can `addSpecies` tell the `Species` constructor "don't worry about molecular weight" and Python (via `ct.Species`, although this would extend to classes created directly in C++ as well) can say "please calculate and set the molecular weight for me if you know about all these elements and throw an error if I try to get the molecular weight and you found an element you don't know about".

Maybe there isn't a way to resolve this satisfactorily, but I hope there is.

speth:
Here is an existing example from the test suite that would lead to errors/inconsistencies with some of the implementations proposed here. Phase definition:

https://github.com/Cantera/cantera/blob/9573e6b1e1097dca3bf86da37726333ab62663bb/test/data/ideal-gas.yaml#L31-L38

where the referenced element definitions are:
https://github.com/Cantera/cantera/blob/9573e6b1e1097dca3bf86da37726333ab62663bb/test/data/test_subdir/species-elements.yaml#L3-L7

ischoegl:
@speth ... thank you, this sheds some light! This is an interesting case. But the larger question here is: what do you do if you have more than one isotope? ... One alternative would be to specify the isotope further down in the definition of the species, i.e.
https://github.com/Cantera/cantera/blob/9573e6b1e1097dca3bf86da37726333ab62663bb/test/data/ideal-gas.yaml#L139-L149
for example by adding a key for molecular weight.

(which imho would be a cleaner solution to this conundrum). I believe that would allow for the co-existence of multiple isotopes of the same element.

speth:
> I think this can be resolved by introducing a new/modified constructor as @ischoegl suggested, or perhaps a setter/getter method pair... The question to me is what the signature should look like to be most clear. Or, put another way, how can `addSpecies` tell the `Species` constructor "don't worry about molecular weight" and Python (via `ct.Species`, although this would extend to classes created directly in C++ as well) can say "please calculate and set the molecular weight for me if you know about all these elements and throw an error if I try to get the molecular weight and you found an element you don't know about".

If you take my suggestion of letting `Phase::addSpecies` be the place where molecular weight is calculated, I would suggest the following:
- Add `m_molecular_weight` (or whatever you want to call it) as a `protected` member of `Species`
- Initialize it to `nan` in the `Species` constructor
- Add a setter that will be called from `Phase::addSpecies`
- Add a getter that checks if the value is `nan` and throws if it is
- Do not implement the setter in Python

ischoegl:
While I do not like the `Phase::addSpecies` avenue at all (it just *feels* wrong), it avoids code refactoring. Isotopes have come up elsewhere recently, so this is probably a good candidate for a feature request, which could address this more thoroughly (perhaps adding a `// @todo: revisit this feature` comment when initializing the value may also be good).

That said, what @speth suggests is exactly how things are usually implemented based on our coding style recommendations (no surprises!). I only have one friendly amendment: initialize it to a negative value, as it is an easier sentinel value to handle.

speth:
@ischoegl - that test case admittedly isn't super exciting. It uses just `Ar` as the element name precisely because the goal is to test that the default atomic weight for that element symbol can be overridden.

The way to have multiple isotopes of an element would be to give them distinct names, like `O16` and `O18`. The set of species you might end up defining would start to multiply, e.g. you would have up to three species for what we normally just call O2, with compositions set to `{O16: 2}`, `{O18: 2}` and `{O16: 1, O18: 1}`, and at least in the general case would have slightly different thermo data.

I don't think you want to just store the molecular weight in the species definition (this seems like it's a long way from the original topic of this PR...), because we still want reactions and equilibrium calculations to preserve isotopic abundances.

ischoegl:
@speth ... I agree.

> The way to have multiple isotopes of an element would be to give them distinct names, like `O16` and `O18`. The set of species you might end up defining would start to multiply, e.g. you would have up to three species for what we normally just call `O2`, with compositions set to `{O16: 2}`, `{O18: 2}` and `{O16: 1, O18: 1}`, and at least in the general case would have slightly different thermo data.

That makes sense (of course), which leads to a compromise: calculate the molecular weight if it's listed in `Elements.h` or defer if custom elements are found. To be honest, I find the test example somewhat contrived (*"the goal is to test that the default atomic weight for that element symbol can be overridden"*), and it's about the only case where what I just said would fail.

speth:
> To be honest, I find the test example somewhat contrived (_"the goal is to test that the default atomic weight for that element symbol can be overridden"_), and it's about the only case where what I just said would fail.

The test may be a bit contrived, but being able to do this has been a feature in the CTI format since Cantera 1.6, so it's better to test it than not. I'd agree that overriding the default element weights is kind of a niche feature, but I'm not ready to assert that there's _no_ use case for it just so a `Species` object not attached to a `Phase` can still have a `molecular_weight` property.

ischoegl:
Ok. As mentioned above: 

> While I do not like the `Phase::addSpecies` avenue at all (it just _feels_ wrong), it avoids code refactoring. […] 
> 
> That said, what @speth suggests is exactly how things are usually implemented based on our coding style recommendations (no surprises!). I only have one friendly amendment: initialize it to a negative value, as it is an easier sentinel value to handle.

At the same time, with CTI gone we could revisit this historical feature (which I don’t think is good practice). I just don’t like it. If anything, switching out an element at the phase level should not be viewed as an ambiguity, but a deliberate choice by the user, where changing values are to be expected.

speth:
> At the same time, with CTI gone we could revisit this historical feature (which I don’t think is good practice). I just don’t like it. If anything, switching out an element at the phase level should not be viewed as an ambiguity, but a deliberate choice by the user, where changing values are to be expected.

What ambiguity is there in the existing implementation? You have to be quite explicit in the phase definition to use alternative element definitions, as shown in the example I linked to before:
```yaml
 - name: element-remote 
   thermo: ideal-gas 
   elements: 
   - default: [N, O] 
   - test_subdir/species-elements.yaml/isotopes: [Ar] 
   species: [O2, NO, N2, AR] 
   note: replace Argon with custom element from a different file 
   state: {T: 900.0, P: 5 atm, Y: {O2: 0.4, N2: 0.4, AR: 0.2}} 
```

ischoegl:
I agree that there is no ambiguity. What I mean here is that on principle, there shouldn't be an issue with a default value (as created by a `Species` constructor) being overwritten due to a differing specification in the phase YAML input.

**PS:** I am stopping here, as I'm ok with any path taken at this point - it's ultimately not a big enough issue to me

bryanwweber:
@speth / @ischoegl I updated the implementation to remove code from `newSpecies` and move it to the `Species` constructor and `Phase::addSpecies`. I think I hit a compromise of having a useful `molecular_weight` attribute when a species is created without a phase, while hopefully avoiding inconsistent data. This does need a few more tests to address code coverage and I need to fix anything that breaks running on CI, but I wanted to post some actual code to move the discussion forward, I think we got a little bit off track in the thread here.

FWIW, I think having `Phase::addSpecies` be able to set attributes on `Species` is fine. There are places where the abstraction of a "species" is leaky, with one case being isotopes or "custom" elements that the `Species` can't know about without information that's only available from outside `Species` (in this case, the elements specification).

bryanwweber:
@speth / @ischoegl I think I need to add some more tests here to address the coverage, but can you guys take a look at the C++ changes? I think I got things correct, but I'd be happy to have some advice on improvements, especially with respect to improving the memory usage of the new functions I've added.

bryanwweber:
I think I fixed the pre-Python 3.9 failures with `Element`. Python 3.9 adds the ability to chain the `classmethod` decorator with the `property` decorator, see https://docs.python.org/3/library/functions.html#classmethod. Once we drop Python 3.8 support, we can remove the `_element_names` and `_element_symbols` helper methods. I'll add a note to that effect.

**Edit:** The helper functions are necessary because using a list/tuple comprehension with the `elementSymbols()` C++ function directly results in a compiler error.

ischoegl (commented):
While this isn't a comment on code, I believe this feature should be held back until we bump to version `3.0.0a0`. If there's consensus, I'd also like to suggest tagging new features using Sphinx/doxygen directives (similar to what was illustrated in #1274).

bryanwweber:
> While this isn't a comment on code, I believe this feature should be held back until we bump to version `3.0.0a0`

I think there's a few ways to approach this... I think if we want to do 2.6.1, we can add commits to the `2.6` branch, which was created (or can be created) from the commit where 2.6 is tagged. Then we'd avoid releasing features like this that shouldn't go into a patch release, and development can keep moving forward on stuff like this in the meantime.

include/cantera/thermo/Species.h
```diff
@@ -52,13 +52,31 @@ class Species
     //! species, where it represents the number of sites occupied.
     double size;
 
+    double molecularWeight();
```
speth, line 62: The getter should be marked `const`.

include/cantera/thermo/Species.h
```diff
@@ -52,13 +52,31 @@ class Species
...
+
+    void setMolecularWeight(double weight);
+
+    std::vector<std::string> unknownElements() {
+        return m_unknownElements;
+    }
```
speth: This method is both `const` and could return a `const` reference, rather than a copy (though I'd rathe we didn't need this at all)

src/thermo/Species.cpp
```diff
@@ -29,12 +32,54 @@ Species::Species(const std::string& name_, const compositionMap& comp_,
     , charge(charge_)
     , size(size_)
 {
+    double wt = 0.;
+    const map<string, double>& elements = elementSymbolToWeight();
+    const map<string, double>& isotopes = isotopeSymbolToWeight();
```
speth: You can use `auto&` here

src/thermo/Elements.cpp
```diff
@@ -196,28 +196,96 @@ static struct atomicWeightData atomicWeightTable[] = {
...
+	vector<string> values;
+	values.reserve(atomicWeightTable.size());
+
+    if (symbol) {
+        transform(atomicWeightTable.begin(), atomicWeightTable.end(),
+            back_inserter(values),
+            [](const atomicWeightData& data) {
+                return data.symbol;
+            }
+        );
```
speth: A simple `for` loop here would be both shorter and quite a bit easier to read.

src/thermo/Elements.cpp
```diff
@@ -196,28 +196,96 @@ static struct atomicWeightData atomicWeightTable[] = {
...
+            symMap.emplace(atom.symbol, atom.atomicWeight);
+        } else if (name) {
+            symMap.emplace(atom.fullName, atom.atomicWeight);
+        }
+    }
+    return symMap;
+}
+
+const map<string, double>& elementSymbolToWeight() {
+    const static map<string, double> symMap = mapAtomicWeights(atomicWeightTable, true, false);
```
speth: Please stick to our max line length of 88 characters.

src/thermo/Elements.cpp
```diff
@@ -196,28 +196,96 @@ static struct atomicWeightData atomicWeightTable[] = {
...
+}
+
 double getElementWeight(const std::string& ename)
 {
-    int numElements = numElementsDefined();
-    int numIsotopes = numIsotopesDefined();
+    const map<string, double> elementSymbolMap = elementSymbolToWeight();
+    const map<string, double> elementNameMap = elementNameToWeight();
+    const map<string, double> isotopeSymbolMap = isotopeSymbolToWeight();
+    const map<string, double> isotopeNameMap = isotopeNameToWeight();
```
speth: This is creating copies of these maps each time `getElementWeight` is called. I'd suggest `auto&` instead.

include/cantera/thermo/Elements.h
```diff
@@ -84,6 +84,54 @@ namespace Cantera
...
+ * @return A vector of the element names
+ */
+const std::vector<std::string>& elementNames();
+
+//! Get a map with the element symbols as keys and weights as values.
+/*!
+ * This is a constant in the application so it is only generated once
+ * when it is first needed.
+ *
+ * @return A map of string symbols to atomic weights
```
speth: I think most of these `@return` annotations are unnecessary -- they just repeat what is stated in the summary docstring. But if you want this, I'd suggest:
```suggestion
 * @return A map of element symbols to atomic weights
```

include/cantera/thermo/Elements.h
```diff
@@ -84,6 +84,54 @@ namespace Cantera
...
+ */
+const std::map<std::string, double>& elementSymbolToWeight();
+
+//! Get a map with the element names as keys and weights as values.
+/*!
+ * This is a constant in the application so it is only generated once
+ * when it is first needed.
+ *
+ * @return A map of string names to atomic weights
```
speth:
```suggestion
 * @return A map of element names to atomic weights
```

src/thermo/Elements.cpp
```diff
@@ -196,28 +196,96 @@ static struct atomicWeightData atomicWeightTable[] = {
...
     {"Tr", "tritium", 3.0160492820, 1},
     {"E", "electron", ElectronMass * Avogadro, 0},
 };
 
+vector<string> elementVectors(bool symbol, bool names) {
+	vector<string> values;
+	values.reserve(atomicWeightTable.size());
+
+    if (symbol) {
+        transform(atomicWeightTable.begin(), atomicWeightTable.end(),
```
speth: I think this should probably just be two separate methods -- the boolean flag to just implement a different function here is kind of messy.

src/thermo/Species.cpp
```diff
@@ -29,12 +32,54 @@ Species::Species(const std::string& name_, const compositionMap& comp_,
...
+    }
 }
 
 Species::~Species()
 {
 }
 
+double Species::molecularWeight() {
+    if (!m_unknownElements.empty()) {
+        throw CanteraError(
```
speth: Can we make the calculation of the molecular weight lazy, and do it in this method rather than the constructor? I think that would do a couple of things for us:
1. Eliminate the need for the `m_uknownElements` member variable (which is just wasting space most of the time) and the `unknownElements` member function.
2. Still allow a phase to have custom element that overrides a built-in one, by having `Phase::addSpecies` always call `setMolecularWeight`.

src/thermo/Species.cpp
```diff
@@ -29,12 +32,54 @@ Species::Species(const std::string& name_, const compositionMap& comp_,
...
+            "Species::molecularWeight",
+            "Species includes unknown element(s) or isotope(s): '{}'. Add it to a phase to calculate the molecular weight"
+            // m_unknownElements
```
speth: This exception message isn't fully working.

include/cantera/thermo/Species.h
```diff
@@ -52,13 +52,31 @@ class Species
...
     shared_ptr<SpeciesThermoInterpType> thermo;
 
     //! Input parameters used to define a species, for example from a YAML input file.
     AnyMap input;
+
+protected:
+
+    //! The molecular weight of the species, in atomic mass units. Includes
+    //! electron mass for charged species.
+    double m_molecularWeight = -999.;
```
speth: I think it would be better to use the constant `Undef` (from `ct_defs.h`) here, if you have to pick a value. My personal preference would be `nan`, but I know I'll get complaints elsewhere if we use that.

include/cantera/thermo/Species.h
```diff
@@ -52,13 +52,31 @@ class Species
...
+
+    //! If the species contains an element or isotope not defined in Elements.h,
+    //! this string will contain its name. Mostly useful for error messages.
+    std::vector<std::string> m_unknownElements = {};
```
speth: The initializer here is unnecessary.

include/cantera/thermo/Species.h
```diff
@@ -85,8 +103,7 @@ unique_ptr<Species> newSpecies(const AnyMap& node);
 std::vector<shared_ptr<Species> > getSpecies(const XML_Node& node);
 
 //! Generate Species objects for each item (an AnyMap) in `items`.
-std::vector<shared_ptr<Species>> getSpecies(const AnyValue& items);
-
+std::vector<shared_ptr<Species> > getSpecies(const AnyValue& items);
```
speth: The space here is not required (and not recommended) when using C++11.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -66,22 +66,23 @@ cdef class Species:
...
+            if composition is not None:
+                composition = comp_map(composition)
+            else:
+                composition = comp_map("")
+            self._species.reset(new CxxSpecies(name, composition))
             self.species = self._species.get()
 
-    def __init__(self, name=None, composition=None, charge=None, size=None,
-                 *args, init=True, **kwargs):
+    def __init__(self, name=None, composition=None, charge=None, size=None, *args, init=True, **kwargs):
```
speth: I don't undesrtand why this is being changed.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -66,22 +66,23 @@ cdef class Species:
...
-            self._species.reset(new CxxSpecies())
+            if name is not None:
+                name = stringify(name)
+            else:
+                name = stringify("")
+            if composition is not None:
+                composition = comp_map(composition)
+            else:
+                composition = comp_map("")
+            self._species.reset(new CxxSpecies(name, composition))
```
speth: If you switch to a lazy calculation of the molecular weight, I think this can all be left as it was.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1293,6 +1293,19 @@ def test_standalone(self):
...
+        self.assertNear(s.molecular_weight, 16.043)
+
+    def test_molecular_weight_with_electron(self):
+        yaml = """
+            name: Li+[elyt]
+            composition: {Li: 1, E: -1}
+            thermo:
+              model: constant-cp
+              h0: -278.49 kJ/mol
+              s0: 13.4 J/mol/K
```
speth: The `thermo` field here is unnecessary for this test.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -2182,6 +2188,15 @@ cdef class PureFluid(ThermoPhase):
...
 
+def _element_symbols():
+    syms = elementSymbols()
+    return tuple(pystr(s) for s in syms)
+
+
+def _element_names():
+    names = elementNames()
+    return tuple(pystr(n) for n in names)
+
```
speth, line 2067: Two blank lines after this function definition, right?

speth (changes requested):
I think this is moving in the right direction. I have some suggestions for improving the implementation. In particular, the current version does suffer from the inconsistency problem that I had been concerned about:
```py
>>> import cantera as ct
>>> gas = ct.Solution('test/data/ideal-gas.yaml', 'element-remote')
>>> gas['AR'].molecular_weights
array([38.])  # correct, based on custom element definition
>>> gas.species('AR').molecular_weight
39.95  # incorrect
```

include/cantera/thermo/Species.h
bryanwweber: I think `Undef` would be appropriate here. `nan` (to me) indicates that some calculation was done and returned something that's not a number. On the other hand, at this point, this value is undefined.

interfaces/cython/cantera/thermo.pyx
bryanwweber: A mistake! 😄 I had moved `name` and `composition` out of this initializer, which allowed everything to fit on one line, but then I added them back and didn't fix the line length 🤦‍♂️

include/cantera/thermo/Species.h
speth: I'd agree if `Undef` wasn't a made-up constant used only by Cantera that takes on a somewhat unlikely but still perfectly valid floating point value (-999.1234). IEE754 doesn't have any specific way of indicating an undefined value.

src/thermo/Species.cpp
bryanwweber: > Can we make the calculation of the molecular weight lazy, and do it in this method rather than the constructor?

Sure, I think this makes a lot of sense. One minor concern that is probably already present, what happens in this case:

```python
s = ct.Species(name='ARstandard', {"AR": 1})
s.molecular_weight
39.95
gas = ct.Solution('test/data/ideal-gas.yaml', 'element-remote')
gas.add_species(s)  # this is made up, I don't remember the API to do this off-hand, but I think this is possible
s.molecular_weight
????
```
The only reason to check if the value is `Undef` in the setter is to save some unnecessary computation... on the other hand, maybe this computation isn't that intensive, and it's only done once, so it should always be done when a species is added to a phase. What do you think? FWIW, I'm leaning towards always calculate the molecular weight when a species is added to a phase for simplicity.

speth: I suppose there are two options here, both of which start with `Phase::addSpecies` always setting the molecular weight of the `Species` object.

1. `Species::setMolecularWeight` just accepts the updated value, so it now returns 38.
2. `Species::setMolecularWeight` throws if the stored value is not either `Undef` or very near the input value, which would imply that the weight had already been calculated based standard element definitions and accessed elsewhere, leading to potential inconsistency.

I think the second option is better, and closer to what you already have.

bryanwweber:
Thanks @speth, I think I addressed all your comments now!

ischoegl:
> If there's consensus, I'd also like to suggest tagging new features using Sphinx/doxygen directives (similar to what was illustrated in #1274).

... just quoting my own comment. As things stand, the directives are still up for discussion.

bryanwweber:
Whoops, thanks for the reminder @ischoegl! I'll add them tonight

interfaces/cython/cantera/thermo.pyx
```diff
@@ -176,6 +176,14 @@ cdef class Species:
         def __get__(self):
             return self.species.size
 
+    property molecular_weight:
+        """The molecular weight [amu] of the species.
+
+        ..versionadded:: 3.0
```
speth:
```suggestion
        .. versionadded:: 3.0
```

include/cantera/thermo/Elements.h
```diff
@@ -84,6 +84,50 @@ namespace Cantera
 //! stable state at 298.15 K and 1 bar.
 #define ENTROPY298_UNKNOWN -123456789.
 
+//! Get a vector of the atomic symbols of the elements defined in Cantera.
+//! @since 3.0
```
speth: In the HTML docs, this just renders as:
> image

I guess this is a bit like the `@deprecated` notation, where Doxygen doesn't massage the provided text at all, so maybe we should write `@since New in version 3.0.` or `@since New in Cantera 3.0.`. The former is what Sphinx generates when the `.. versionadded:: 3.0` notation is used.

include/cantera/thermo/Species.h
```diff
@@ -51,21 +51,53 @@ class Species
     //! species, where it represents the number of sites occupied.
     double size;
 
+    //! The molecular weight [amu] of the species.
+    /*!
+     * Calculates and sets the molecular weight from the elemental composition of the
+     * species and element definitions in Elements.cpp, if the molcular weight is
+     * Undef.
```
speth:
```suggestion
     * species and element definitions in Elements.cpp, if the molecular weight is
     * Undef.
```

include/cantera/thermo/Species.h
```diff
@@ -51,21 +51,53 @@ class Species
...
+     *
+     * @since 3.0
+     */
+    const double molecularWeight();
+
+    //! Set the molecular weight of the species.
+    /*!
+     * Since phases can have custom element weights, the phase will always call this
+     * method when a species is added to that phase. The species may also call this
+     * method the first time the molecularWeight method is called if the species has
```
speth:
```suggestion
     * method the first time the molecularWeight() method is called if the species has
```

include/cantera/thermo/Species.h
```diff
@@ -51,21 +51,53 @@ class Species
...
+     * not been added to a phase.
+     *
+     * @since 3.0
```
speth: I'd suggest putting the `@since` annotation as the last item in the docstring.

include/cantera/thermo/Species.h
```diff
@@ -51,21 +51,53 @@ class Species
...
+     *
+     * @param weight: The weight of this species to assign, unless `compute` is `true`
+     * @param compute: If `true`, the weight will be computed from standard element
+     * weights from Elements.cpp, and the value of `weight` is ignored.
+     * @exception If the molecular weight has already been set and the updated value
+     * is not close to the existing value.
```
speth: The `@exception` notation expects an exception class before the description
```suggestion
     * @exception CanteraError If the molecular weight has already been set and the updated value
     * is not close to the existing value.
```

src/thermo/Species.cpp
```diff
@@ -34,6 +37,47 @@ Species::~Species()
 {
 }
 
+const double Species::molecularWeight() {
+    if (m_molecularWeight == Undef) {
+        setMolecularWeight(-1.0, true);
+    }
```
speth: Since the `compute=true` branch is only ever called from here, can we just put that code here, and eliminate this argument to `setMolecularWeight`?

include/cantera/thermo/Species.h
```diff
@@ -51,21 +51,53 @@ class Species
...
 
+    //! The molecular weight [amu] of the species.
+    /*!
+     * Calculates and sets the molecular weight from the elemental composition of the
+     * species and element definitions in Elements.cpp, if the molcular weight is
+     * Undef.
+     *
+     * @since 3.0
+     */
+    const double molecularWeight();
```
speth: Things that are returned by value should generally not be marked as `const`.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1232,6 +1232,74 @@ def test_standalone(self):
...
+            composition: {Li: 1, E: -1}
+        """
+        s = ct.Species.from_yaml(yaml)
+        assert np.isclose(s.molecular_weight, 6.939451420091127)
+
+    def test_custom_element_weight_is_set_on_species(self):
+        gas = ct.Solution("ideal-gas.yaml", "element-override")
+        # Check that the molecular weight stored in the phase definition is the same
+        # as the one on the element
+        self.assertNear(gas["AR"].molecular_weights[0], gas.species("AR").molecular_weight)
```
speth: I think this needs to be line-wrapped.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1232,6 +1232,74 @@ def test_standalone(self):
...
+        with pytest.raises(ct.CanteraError, match="Molecular weight.*changed"):
+            gas.add_species(s)
+
+    def test_species_cant_be_added_to_phase_custom_element(self):
+        s = ct.Species.from_dict({
+            "name": "AR2",
+            "composition": {"Ar": 2},
+            "thermo": {"model": "constant-cp", "h0": 100}
+        })
+        # DO NOT access the molecular weight to make sure it's not been computed by the Species
```
speth: This comment should be line-wrapped as well.

speth: Regarding the test name, this is a test that the species _can_ be added, right?

interfaces/cython/cantera/thermo.pyx
```diff
@@ -2045,6 +2053,18 @@ cdef class PureFluid(ThermoPhase):
         def __get__(self):
             return self.s, self.v, self.Q
 
+# TODO: Remove these helper methods when support for Python 3.8 is dropped. Python 3.9
+# allows the classmethod and property decorators to be chained, so these can be
+# implemented as properties in the Element class.
```
speth, line 2058: Wouldn't making these into class properties mean that this tuple was constructed again from the C++ `elementSymbols` method every time it was accessed, rather than just a single time as a class attribute?

src/thermo/Phase.cpp
```diff
@@ -846,6 +846,7 @@ bool Phase::addSpecies(shared_ptr<Species> spec) {
     wt = std::max(wt, Tiny);
     m_molwts.push_back(wt);
     m_rmolwts.push_back(1.0/wt);
+    spec->setMolecularWeight(wt);
```
speth: Given that this can throw, I think we should move the calculation of `wt` and this call to before any of the arrays belonging to the phase have values appended. I think this starts where `m_speciesNames` is updated. Otherwise, this will result in the `Phase` object ending up in an inconsistent state.

src/thermo/Species.cpp
```diff
@@ -34,6 +37,47 @@ Species::~Species()
...
+            } else {
+                search = isotopes.find(comp.first);
+                if (search != isotopes.end() && search->second > 0) {
+                    weight += search->second * comp.second;
+                }
+            }
+        }
+    }
+    if (m_molecularWeight != Undef) {
+        double weight_cmp = fabs(weight - m_molecularWeight) / max(weight, m_molecularWeight);
```
speth: This line is too long.

speth (changes requested):
Thanks for the updates, @bryanwweber. I had a few more suggested updates.

include/cantera/thermo/Species.h
bryanwweber: I think I misinterpreted your previous comment then,

> The getter should be marked `const`.

Can you clarify? Thanks :smile:

interfaces/cython/cantera/thermo.pyx
bryanwweber: I don't think so. In Python 3.10, if I put:

```python
class Element:
...
    @classmethod
    @property
    def element_symbols():
        syms = elementSymbols()
        return tuple(pystr(s) for s in syms)
...
```

and then do

```python
import cantera as ct
ct.Element.element_symbols is ct.Element.element_symbols
```
it returns `True`. So I think marking it as a `@classmethod` means that it's only computed once, when the class is defined.

interfaces/cython/cantera/test/test_thermo.py
bryanwweber: Yes, I think I deleted the `no` from `cannot` and missed the `t` :smile:

include/cantera/thermo/Species.h
speth: I was suggesting `double molecularWeight() const`, which means that the method doesn't modify the `Species` object. However, that's not actually true at this point, since this triggers the calculation of `m_molecularWeight`, so I think this is a moot point.

interfaces/cython/cantera/thermo.pyx
speth: For me, trying the above snippet in Python 3.10 gives the error: `TypeError: element_symbols() takes no arguments (1 given)`. Is there something else I'm missing?

bryanwweber: Oh, sorry, it should be `def element_symbols(cls):`

speth: Right, I should have known that. On my system, `ct.Element.element_symbols is ct.Element.element_symbols` returns `False` (executing from within IPython). And putting a `print` into the method shows that it is being executed every time.

bryanwweber: Ack, this is what I get for explaining from memory. I also added a class variable `_element_symbols = None`, so it looks like:

```python
class Element:
...
    _element_symbols = None
...
    @classmethod
    @property
    def element_symbols(cls):
        if cls._element_symbols is None:
            syms = elementSymbols()
            cls._element_symbols = tuple(pystr(s) for s in syms)
        return cls._element_symbols
```

and likewise for the `element_names`.

By the way, I'd prefer to do `element_symbols = tuple(pystr(s) for s in elementSymbols())` but that gives me a compiler error, something about `lvalue` or `rvalue` not being able to be used as a reference. By experimenting, I interpreted that to mean that since `elementSymbols()` is used in the comprehension, it gets assigned to a reference by Cython in `_cantera.cpp`, and if I assign it to a variable, things are better.

speth: Ah, I see. Does updating `cantera.pxd` to indicate that `elementSymbols` returns a `vector[string]&` get Cython to generate something that actually compiles?

bryanwweber: No, then I get:

```
clang++ -o build/temp-py/_cantera.os -c -std=c++11 -fcolor-diagnostics -pthread -O0 -g -fPIC -Iinclude -I/home/bryanw/mambaforge/envs/cantera-dev/include/python3.10 -I/home/bryanw/mambaforge/envs/cantera-dev/lib/python3.10/site-packages/numpy/core/include -I/home/bryanw/mambaforge/envs/cantera-dev/include build/python/cantera/_cantera.cpp
...
build/python/cantera/_cantera.cpp:63552:15: error: assigning to 'std::vector<std::string> *' (aka 'vector<basic_string<char>> *') from 'const std::vector<std::string> *' (aka 'const vector<basic_string<char>> *') discards qualifiers
  __pyx_t_2 = &Cantera::elementSymbols();
              ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

Well, checking now, I get the same error whether it's `vector[string]&` in `cantera.pxd` or not. The error message about `lvalue` must have been with an old implementation on the C++ side.

speth: Ah, too bad. I tried a couple of other things but couldn't come up with one that made the one-liner version of this actually work.

bryanwweber: Yeah, I tried looking around the internet and couldn't find a good answer either. Darn.

include/cantera/thermo/Elements.h
```diff
@@ -84,6 +84,50 @@ namespace Cantera
...
+const std::vector<std::string>& elementNames();
+
+//! Get a map with the element symbols as keys and weights as values.
+/*!
+ * This is a constant in the application so it is only generated once
+ * when it is first needed.
+ *
+ * @since New in version 3.0
+ */
+const std::map<std::string, double>& elementSymbolToWeight();
```
ischoegl: Can this be shortened to `elementWeight(s)`? I find the `To` not to be intuitive.

include/cantera/thermo/Elements.h
```diff
@@ -84,6 +84,50 @@ namespace Cantera
...
+
+//! Get a map with the element names as keys and weights as values.
+/*!
+ * This is a constant in the application so it is only generated once
+ * when it is first needed.
+ *
+ * @since New in version 3.0
+ */
+const std::map<std::string, double>& elementNameToWeight();
```
ischoegl: I see that you're making the distinction between symbols and names, but both can be held by the same map.

include/cantera/thermo/Elements.h
```diff
@@ -84,6 +84,50 @@ namespace Cantera
...
+
+//! Get a map with the isotope symbols as keys and weights as values.
+/*!
+ * This is a constant in the application so it is only generated once
+ * when it is first needed.
+ *
+ * @since New in version 3.0
+ */
+const std::map<std::string, double>& isotopeSymbolToWeight();
```
ischoegl: Similar to above, this can be shortened to `isotopeWeight`, with names being added into the same map.

include/cantera/thermo/Species.h
```diff
@@ -51,21 +51,53 @@ class Species
...
+    //! electron mass for charged species.
+    double m_molecularWeight = Undef;
 };
 
 //! Create a new Species object from an AnyMap specification
 unique_ptr<Species> newSpecies(const AnyMap& node);
 
 //! Generate Species objects for each item (an AnyMap) in `items`.
 std::vector<shared_ptr<Species>> getSpecies(const AnyValue& items);
-
```
ischoegl: I think those line breaks are commonly used and shouldn't be deleted. The closing `}` is from a different context.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1232,6 +1232,77 @@ def test_standalone(self):
...
+        self.assertNear(
+            gas["AR"].molecular_weights[0], gas.species("AR").molecular_weight
+        )
+        # Check that the custom value is actually used
+        self.assertNear(gas.species("AR").molecular_weight, 36.0)
+
+    def test_species_can_be_added_to_phase(self):
+        s = ct.Species.from_dict({
+            "name": "AR2",
+            "composition": {"Ar": 2},
```
ischoegl, line 1263: Interesting choice :joy:

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1232,6 +1232,77 @@ def test_standalone(self):
...
+            "thermo": {"model": "constant-cp", "h0": 100}
+        })
+        # Access the molecular weight to make sure it's been computed by the Species
+        self.assertNear(s.molecular_weight, 39.95 * 2)
+        gas = ct.Solution("ideal-gas.yaml", "element-override")
+        # The error here is because the weight of the Argon element has been changed in
+        # the phase definition, but the molecular weight of the species has already been
+        # computed
+        with pytest.raises(ct.CanteraError, match="Molecular weight.*changed"):
+            gas.add_species(s)
```
ischoegl: I'm not sure that I am following the rationale used here - the species definition does not include any information about molecular weight, so imho `gas.add_species` should be agnostic.

interfaces/cython/cantera/test/test_thermo.py
```diff
@@ -1232,6 +1232,77 @@ def test_standalone(self):
...
+    def test_species_can_be_added_to_phase_custom_element(self):
+        s = ct.Species.from_dict({
+            "name": "AR2",
+            "composition": {"Ar": 2},
+            "thermo": {"model": "constant-cp", "h0": 100}
+        })
+        # DO NOT access the molecular weight to make sure it's not been computed by the
+        # Species
+        gas = ct.Solution("ideal-gas.yaml", "element-override")
+        gas.add_species(s)
```
ischoegl, line 1301: It took me a while to understand why this works in this case. I think that generating a warning in the case where the molecular weight is overwritten would be preferable.

src/thermo/Elements.cpp
```diff
@@ -188,36 +188,106 @@ static struct atomicWeightData atomicWeightTable[] = {
...
-        } else if (name == atomicWeightTable[i].fullName) {
-            elementWeight = atomicWeightTable[i].atomicWeight;
-            break;
+    auto search = elementSymbolMap.find(symbol);
+    if (search != elementSymbolMap.end()) {
+        elementWeight = search->second;
+    } else {
+        search = elementNameMap.find(name);
+        if (search != elementNameMap.end()) {
+            elementWeight = search->second;
```
ischoegl: This will become much simpler if the two maps are merged. Same for the isotope version below.

include/cantera/thermo/Elements.h
```diff
@@ -84,6 +84,50 @@ namespace Cantera
 //! stable state at 298.15 K and 1 bar.
 #define ENTROPY298_UNKNOWN -123456789.
 
+//! Get a vector of the atomic symbols of the elements defined in Cantera.
+//! @since New in version 3.0
+const std::vector<std::string>& elementSymbols();
+
+//! Get a vector of the names of the elements defined in Cantera.
+//! @since New in version 3.0
+const std::vector<std::string>& elementNames();
```
ischoegl, line 93: I think that creating a 
```C++
const std::map<std::string, std::string>& elementNames();
```
with symbols as keys and names as values is more flexible, as it will simplify lookup.

For symmetry, a corresponding
```C++
const std::map<std::string, std::string>& isotopeNames();
```
seems intuitive?

src/thermo/Elements.cpp
```diff
@@ -252,25 +327,23 @@ double getElementWeight(int atomicNumber)
...
-    int numElements = numElementsDefined();
-    int numIsotopes = numIsotopesDefined();
     string name = toLowerCopy(trimCopy(ename));
-    for (int i = 0; i < numElements; i++) {
-        if (name == atomicWeightTable[i].fullName) {
-            return atomicWeightTable[i].symbol;
+    for (const auto& atom : atomicWeightTable) {
+        if (name == atom.fullName) {
+            return atom.symbol;
         }
```
ischoegl, line 300: If the suggested `elementNames` map is adopted, it could be used here (same for a potential `isotopeNames` map below).

src/thermo/Elements.cpp
```diff
@@ -280,25 +353,23 @@ string getElementSymbol(int atomicNumber)
...
-    int numElements = numElementsDefined();
-    int numIsotopes = numIsotopesDefined();
     string symbol = trimCopy(ename);
-    for (int i = 0; i < numElements; i++) {
-        if (symbol == atomicWeightTable[i].symbol) {
-            return atomicWeightTable[i].fullName;
+    for (const auto& atom : atomicWeightTable) {
+        if (symbol == atom.symbol) {
+            return atom.fullName;
         }
```
ischoegl, line 326: If the suggested `elementNames` map is adopted, this would simplify (same for a potential `isotopeNames` map below).

src/thermo/Species.cpp
```diff
@@ -34,6 +37,46 @@ Species::~Species()
...
+
+void Species::setMolecularWeight(double weight) {
+    if (m_molecularWeight != Undef) {
+        double maxWeight = max(weight, m_molecularWeight);
+        double weight_cmp = fabs(weight - m_molecularWeight) / maxWeight;
+        if (weight_cmp > 1.0e-9) {
+            throw CanteraError(
+                "Species::setMolecularWeight",
+                "Molecular weight of a species cannot be changed."
+            );
```
ischoegl: I believe this should raise a warning instead.

ischoegl (changes requested):
HI @bryanwweber ... thanks for working on this, as I believe that this is a nice addition. I have a few suggestions that should simplify a couple of things.

One thing I believe could be improved in this context (although I do not suggest to add to this PR) is the handling of modified element or isotope weights - while this can be done (see the `element-override` example), it appears to exist completely outside of the element handler.

include/cantera/thermo/Elements.h
bryanwweber: How would I do that, and ensure that `map::find` works properly? In Python, it would be very awkward, you'd have to do something like a tuple as the key, I think... Or are you thinking that weights would be duplicated in the map?

include/cantera/thermo/Elements.h
bryanwweber: As you note below, the `To` is there to distinguish from names. To me, it also indicates a mapping.

interfaces/cython/cantera/test/test_thermo.py
bryanwweber: Just making sure it's a species that almost certainly doesn't exist in the mechanism 😄

interfaces/cython/cantera/test/test_thermo.py
bryanwweber: The point is that the species definition _does_ include information about molecular weight now, since the `molecular_weight` method was called. Maybe I'm not understanding the question?

interfaces/cython/cantera/test/test_thermo.py
bryanwweber: See below... Also, the molecular weight is not being overridden in this case. I think your suggestion would result in a warning every time you add a species with a custom element weight?

include/cantera/thermo/Elements.h
bryanwweber: ~What advantage does the map provide? I don't think it'd simplify that much, since you still don't know whether you're looking up a symbol or name without trying it first.~

The other concern that I have is, what's the key of the map? The atomic number? The weight? The symbol? I don't know unless I look at the docstring. Then would we want maps for everything else? I'm not inclined to implement this at the moment... 😕

src/thermo/Elements.cpp
bryanwweber: I can see how this would be simpler, but I think the complexity would just move the the creation of the map, I don't think it would actually remove complexity overall...

src/thermo/Species.cpp
bryanwweber: Can you talk more about the advantages you see of a warning? I think allowing user code to overwrite the molecular weight would potentially be very confusing. The trade off that I see is people might find it confusing that sometimes the species they create can be added to a phase and sometimes they get this error.

include/cantera/thermo/Elements.h
ischoegl: > weights would be duplicated in the map?

Correct. In essence, I am just suggesting to merge the two maps - the naming becomes more intuitive, and you don't have to search two maps.

interfaces/cython/cantera/test/test_thermo.py
ischoegl: I was referring to the YAML definition - the user doesn't know how things are implemented under the hood.

interfaces/cython/cantera/test/test_thermo.py
ischoegl: No. If I understand your implementation correctly, the error is only thrown if the molecular weight is calculated prior to adding a species, *and* is overridden by a custom element weight.

interfaces/cython/cantera/test/test_thermo.py
bryanwweber: Sorry, I'm still not understanding what you mean. The YAML definition of a species implicitly contains information about the molecular weight, since it includes the elemental composition. This is testing for the case where a species has been created in Python, then added to a phase.

include/cantera/thermo/Elements.h
ischoegl: No worries - my suggestion would be useful if element handling is made consistent for the overrides - the Element classes  are currently not aware of these, which makes things quite hard to follow. While this is not being done, the internal `atomicWeightTable` and `isotopeWeightTable` objects (which are not accessible otherwise) are sufficient to handle everything.

interfaces/cython/cantera/test/test_thermo.py
bryanwweber: Yes, that's true. I forgot that it was also checking whether the molecular weight had been computed.

src/thermo/Species.cpp
bryanwweber: Another concern if this is a warning, what should the molecular weight be?

```
s = ct.Species(...)
s.molecular_weight
39.95
gas.add_species(s)  # with custom molecular weight
s.molecular_weight
????
```
Should it be the original value or the modified value? I can see cases where you'd want both behaviors...

interfaces/cython/cantera/test/test_thermo.py
ischoegl: I am mostly thinking about user intent here - at the end of the day, we're just discussing whether this should raise a warning or throw an error.

bryanwweber: :+1: Then let's move the discussion to that comment below 😄

include/cantera/thermo/Elements.h
bryanwweber: One option is a [`boost::bimap`](https://www.boost.org/doc/libs/1_79_0/libs/bimap/doc/html/boost_bimap/reference.html#boost_bimap.reference.headers) which is header only, I think!

src/thermo/Species.cpp
ischoegl: I agree that there is a tradeoff, where the reasoning you gave is ultimately more important from my perspective:

> The trade off that I see is people might find it confusing that sometimes the species they create can be added to a phase and sometimes they get this error.

include/cantera/thermo/Elements.h
ischoegl: That's an interesting option. ~How would you use it?~

Nevermind: https://www.boost.org/doc/libs/1_75_0/libs/bimap/doc/html/boost_bimap/one_minute_tutorial.html

src/thermo/Species.cpp
ischoegl: In your example, if the first `s.molecular_weight` is not called, then there is no error. Presumably, the user intent will be to use elements that are consistent with definitions of the phase the species is added to. I believe the most useful implementation here is to raise a warning and change the weight.

include/cantera/thermo/Elements.h
speth: That's an interesting idea. This would definitely simplify the locations where you have to search for the weights in multiple locations.

include/cantera/thermo/Elements.h
speth: Or you could go even a step further and include the isotope weights (by name and symbol) in the `elementWeights` map, since there are no overlaps, which would reduce the number of API functions that need to exist and simplify searching when you want to allow either elements or isotopes (which I think covers all the cases we have).

src/thermo/Species.cpp
speth: I think raising an exception was a suggestion that came from me, but I'd agree that a warning if the weight changes would be sufficient for letting the user know that they're doing something that might lead to an unexpected result.

src/thermo/Elements.cpp
```diff
@@ -188,36 +188,84 @@ static struct atomicWeightData atomicWeightTable[] = {
...
     {"D",  "deuterium", 2.0141017781, 1},
     {"Tr", "tritium", 3.0160492820, 1},
     {"E", "electron", ElectronMass * Avogadro, 0},
 };
 
+// This is implemented as a separate function from elementSymbols() because this pattern
+// allows elementSymbols() to return a const reference to the data.
+vector<string> elementVectorsFromSymbols() {
+	vector<string> values;
+    for (const auto& atom: atomicWeightTable) {
```
ischoegl:
```suggestion
    for (const auto& atom : atomicWeightTable) {
```

src/thermo/Elements.cpp
```diff
@@ -188,36 +188,84 @@ static struct atomicWeightData atomicWeightTable[] = {
...
+const vector<string>& elementSymbols() {
+    const static vector<string> values = elementVectorsFromSymbols();
+    return values;
+}
+
+// This is implemented as a separate function from elementNames() because this pattern
+// allows elementNames() to return a const reference to the data.
+vector<string> elementVectorsFromNames() {
+    vector<string> values;
+    for (const auto& atom: atomicWeightTable) {
```
ischoegl:
```suggestion
    for (const auto& atom : atomicWeightTable) {
```

src/thermo/Elements.cpp
```diff
@@ -188,36 +188,84 @@ static struct atomicWeightData atomicWeightTable[] = {
...
+}
+
+map<string, double> mapAtomicWeights() {
+    map<string, double> symMap;
+
+    for (auto const& atom : atomicWeightTable) {
+        symMap.emplace(atom.symbol, atom.atomicWeight);
+        symMap.emplace(atom.fullName, atom.atomicWeight);
+    }
+    for (auto const& isotope: isotopeWeightTable) {
```
ischoegl:
```suggestion
    for (auto const& isotope : isotopeWeightTable) {
```

ischoegl (dismissed):
Looks good to me. I did notice some really minor formatting issues that are trivially easy.

ischoegl (approved):
LGTM

speth (approved):

PR #1284, opened 2022-05-18
[SolutionArray] Add array size methods

bryanwweber:

- Add a few methods to `SolutionArray` to have it better match the NumPy API.

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -654,6 +654,33 @@ class SolutionArray:
         return SolutionArray(self._phase[species], states=self._states,
                              extra=self._extra)
 
+    def __len__(self) -> int:
+        return self._shape[0]
+
+    @property
+    def ndim(self) -> int:
+        """The number of dimensions in the SolutionArray.
+
+        :return: `int`
+        """
+        return len(self.shape)
+
+    @property
+    def shape(self) -> tuple[int, ...]:
+        """The shape of the SolutionArray.
+
+        :return: A tuple of integers with the number of elements in each dimension.
+        """
+        return self._shape
+
+    @property
+    def size(self) -> int:
+        """The number of elements in the SolutionArray.
+
+        :return: `int`
+        """
+        return np.prod(self.shape)
+
     def append(self, state=None, normalize=True, **kwargs):
         """
         Append an element to the array with the specified state. Elements can
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -1,5 +1,6 @@
 # This file is part of Cantera. See License.txt in the top-level directory or
 # at https://cantera.org/license.txt for license and copyright information.
+from __future__ import annotations
 
 from ._cantera import *
 import numpy as np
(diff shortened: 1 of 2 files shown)


interfaces/cython/cantera/composite.py
```diff
@@ -655,6 +656,33 @@ def __call__(self, *species):
...
                              extra=self._extra)
 
+    def __len__(self) -> int:
+        return self._shape[0]
+
+    @property
+    def ndim(self) -> int:
+        """The number of dimensions in the SolutionArray.
+
+        :return: `int`
```
ischoegl: Is the `:return:` needed if PEP 484 is used? It was my impression that sphinx is smart enough to deduce?

ischoegl (commented):
Nice addition! But I do have a question about the documenting style (other than the apparent typing issue, which should be easily resolved)

interfaces/cython/cantera/composite.py
bryanwweber: I dunno. I'll figure it out 😄  I guess I can just drop the `:return:` annotation since the docstring is clear enough.

ischoegl: I think the `:return:` documentation can be dropped. Other than potentially labeling this as new in 3.0 this looks pretty good - I have no objections against selectively adding PEP 484.

interfaces/cython/cantera/composite.py
```diff
@@ -655,6 +656,33 @@ def __call__(self, *species):
...
+        """
+        return len(self.shape)
+
+    @property
+    def shape(self) -> tuple[int, ...]:
```
ischoegl, line 670: Should likely be `Tuple` (imported from `typing`)

bryanwweber: I _think_ I did the `from __future__ import annotations` incantation at the top... if not, that's my preference, since we can eventually drop the import.

ischoegl: I’d be ok with that, as long as unit tests pass … will be a while before we drop support for Python 3.8 😄 …

interfaces/cython/cantera/composite.py
```diff
@@ -654,6 +656,35 @@ def __call__(self, *species):
...
+        """
+        return self._shape
+
+    @property
+    def size(self) -> int:
+        """The number of elements in the SolutionArray.
+
+        .. versionadded: 3.0
+        """
+        return math.prod(self.shape)
```
bryanwweber: @speth @ischoegl The reason the tests are failing for Python 3.7 is that `math.prod` was added in 3.8, see: https://docs.python.org/3/library/math.html#math.prod

I suppose I'll switch it to NumPy. The reason I chose `math.prod` was that the [NumPy docs](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.size.html) for `ndarray.size` specify that a standard arbitrary size Python `int` is returned from `ndarray.size`, similar to the result from `math.prod`, as opposed to `np.prod` that returns an `np.int_` type. It _seems_ unlikely that this will be a problem, but when trying to match an API, I think it's useful to match it as closely as possible.

~The real question here is, is this worth dropping 3.7 support over? Thoughts are welcome :smile:~

Edit: Although, the `math` docs say that the functions return a float, unless otherwise specified, so this may not be as meaningful as I thought. I'll just switch to NumPy, never mind!

ischoegl (approved):
This all looks good to me.

PR #1274, opened 2022-05-05
[Sphinx/doxygen] Add directives to new/experimental features

ischoegl:
**Changes proposed in this pull request**

- Add Sphinx `versionadded` directive to experimental features introduced in Cantera 2.6
- Add `versionadded` without additional information to `Kinetics` derivatives
- Add `@since` directive in corresponding C++ member functions
- ~Address some formatting issues per #1275 (limited to `include/cantera/kinetics` folder)~

See discussion in Cantera/enhancements#150 ... the `versionadded` is selectively applied here to investigate additional Sphinx directives for the generation of API documentation.

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

PR #1285, opened 2022-05-18
[Units] Add Cython interface for some functions

bryanwweber:
These functions are chosen for their potential utility. Setting units is
not enabled by this interface because it is meant to be handled
internally. However, these methods can help to determine the unit
factors for reaction rates.

### DIFF
--- a/interfaces/cython/cantera/units.pyx
+++ b/interfaces/cython/cantera/units.pyx
@@ -1,5 +1,6 @@
 # This file is part of Cantera. See License.txt in the top-level directory or
 # at https://cantera.org/license.txt for license and copyright information.
+from typing import Dict
 
 cdef class Units:
     """
@@ -20,6 +21,29 @@ cdef class Units:
 
     def __repr__(self):
         return f"<Units({pystr(self.units.str())}) at {id(self):0x}>"
+
+    def __str__(self):
+        return pystr(self.units.str())
+
+    def dimension(self, primary):
+        return self.units.dimension(stringify(primary))
+
+    @property
+    def dimensions(self) -> Dict[str, str]:
+        """A dictionary of the primary unit components to their dimensions.
+
+        .. versionadded: 3.0
+        """
+        dimensions = ("mass", "length", "time", "temperature", "current", "quantity")
+        return {d: self.dimension(d) for d in dimensions}
+
+    @property
+    def factor(self) -> float:
+        """The factor required to convert from this unit to Cantera's base units.
+
+        .. versionadded: 3.0
+        """
+        return self.units.factor()
 
     @staticmethod
     cdef copy(CxxUnits other):
--- a/interfaces/cython/cantera/units.pyx
+++ b/interfaces/cython/cantera/units.pyx
@@ -19,20 +19,29 @@ cdef class Units:
         if name:
             self.units = CxxUnits(stringify(name), True)
 
-    def __repr__(self):
-        return f"<Units({pystr(self.units.str())}) at {id(self):0x}>"
+    def __repr__(self) -> str:
+        return f"<Units({str(self)}) at {id(self):0x}>"
 
-    def __str__(self):
+    def __str__(self) -> str:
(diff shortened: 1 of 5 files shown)


interfaces/cython/cantera/units.pyx
```diff
@@ -21,6 +21,21 @@ cdef class Units:
     def __repr__(self):
         return f"<Units({pystr(self.units.str())}) at {id(self):0x}>"
 
+    def __str__(self):
```
ischoegl: This should probably be used in the f-string for `__repr__` above.

ischoegl (commented):
This dove-tails nicely with #1076. I have one minor suggestion ...

interfaces/cython/cantera/units.pyx
```diff
@@ -21,6 +21,21 @@ cdef class Units:
...
+        return pystr(self.units.str())
+
+    def dimension(self, primary):
+        return self.units.dimension(stringify(primary))
```
ischoegl, line 38: One more comment: some of these methods should probably get a docstring, in which case some points raised in #1274 could be considered (i.e. tagging new features with the Sphinx directive `versionadded`)

interfaces/cython/cantera/units.pyx
```diff
@@ -1,5 +1,6 @@
 # This file is part of Cantera. See License.txt in the top-level directory or
 # at https://cantera.org/license.txt for license and copyright information.
+from __future__ import annotations
```
ischoegl: This may have to go into `_cantera.pyx`?

bryanwweber: It probably needs to above the comment. I thought it would ignore comments, but I guess not.

bryanwweber: Well, nothing I tried with this worked. In any case, as of Cython 3.0, this import will not be needed for `*.pyx` files, see: https://github.com/cython/cython/issues/2950

interfaces/cython/cantera/units.pyx
```diff
@@ -18,8 +19,38 @@ cdef class Units:
...
+
+    def __str__(self) -> str:
+        return pystr(self.units.str())
+
+    def dimension(self, primary: str) -> float:
+        """The dimension of the given unit component.
+
+        .. versionadded:: 3.0
+
+        :param: primary, a string with the desired unit component. One of ``"mass"``,
```
ischoegl:
```suggestion
        :param primary:
            A string with the desired unit component. One of ``"mass"``,
```
This is mostly a question: is this `:param: primary ...`, or `:param primary: ...`? Most other instances use the latter formatting ... (there are only two instances of the former, which should probably be taken care of)

bryanwweber: Good point, the latter method is how the [sphinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#field-lists) suggest formatting this.

ischoegl: Here are the only two instances of the former in all of Cantera: https://github.com/Cantera/cantera/blob/790b507113d79c4901e319ed8837e5e7cbaaca1f/interfaces/cython/cantera/thermo.pyx#L735-L740
(and it is likely incorrect usage) ... it may be worth taking care of.

interfaces/cython/cantera/units.pyx
```diff
@@ -18,8 +19,40 @@ cdef class Units:
...
+    def __str__(self) -> str:
+        return pystr(self.units.str())
+
+    def dimension(self, primary: str) -> float:
+        """The dimension of the given unit component.
+
+        .. versionadded:: 3.0
+
+        :param primary:
+            a string with the desired unit component. One of ``"mass"``,
```
ischoegl: While I believe that this is resolved, I noticed that many comparable docstring entries start with uppercase?

ischoegl (approved):
Thank you - LGTM!

PR #1287, opened 2022-05-18
Bump python on CI

bryanwweber:

- Bump the minimum Python version required to 3.8. 3.7 is out of support now, plus we need to add 3.11, so I think it makes sense to drop 3.7.
- Bump the Cantera version number to 2.6.1a0. I chose 2.6.1a0 in case we end up having to make a 2.6.1 release, rather than going all the way to 3.0 now and having to backtrack.

**Checklist**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -365,35 +365,6 @@ jobs:
           cti2yaml test/data/air-no-reactions.cti air-no-reaction-test.yaml
           test -f air-no-reaction-test.yaml
 
-  cython-latest:
-    name: Test pre-release version of Cython
-    runs-on: ubuntu-latest
-    timeout-minutes: 60
-    steps:
-    - uses: actions/checkout@v2
-      name: Checkout the repository
-      with:
-        submodules: recursive
-    - name: Setup python
-      uses: actions/setup-python@v3
-      with:
-        python-version: '3.10'
-        architecture: x64
-    - name: Install Apt dependencies
-      run: |
-        sudo apt update
-        sudo apt install libboost-dev liblapack-dev libblas-dev
-    - name: Upgrade pip
-      run: python3 -m pip install -U pip setuptools wheel
-    - name: Install Python dependencies
-      run: |
-        python3 -m pip install ruamel.yaml scons numpy h5py pandas pytest pytest-github-actions-annotate-failures
-        python3 -m pip install https://github.com/cython/cython/archive/master.zip --install-option='--no-cython-compile';
-    - name: Build Cantera
-      run: python3 `which scons` build blas_lapack_libs=lapack,blas python_package='full' -j2 debug=n
-    - name: Test Cantera
-      run: python3 `which scons` test show_long_tests=yes verbose_tests=yes
-
   check-deprecations:
     name: Run test suite without legacy typedefs
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,7 +25,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        python-version: [ '3.7', '3.9', '3.10' ]
+        python-version: [ '3.8', '3.10', '3.11.0-alpha - 3.11.0' ]
         os: ['ubuntu-18.04', 'ubuntu-20.04']
       fail-fast: false
     steps:
@@ -34,7 +34,7 @@ jobs:
       with:
         submodules: recursive
     - name: Setup Python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v3
       with:
         python-version: ${{ matrix.python-version }}
         architecture: x64
@@ -63,7 +63,7 @@ jobs:
       with:
         submodules: recursive
     - name: Setup Python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v3
       with:
         python-version: 3.8
         architecture: x64
@@ -90,7 +90,7 @@ jobs:
     timeout-minutes: 90
     strategy:
       matrix:
-        python-version: [ '3.7', '3.9', '3.10' ]
+        python-version: [ '3.8', '3.10', '3.11.0-alpha - 3.11.0' ]
         os: ['macos-10.15']
       fail-fast: false
     steps:
@@ -107,7 +107,7 @@ jobs:
       with:
         submodules: recursive
     - name: Setup Python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v3
       with:
         python-version: ${{ matrix.python-version }}
         architecture: x64
@@ -138,7 +138,7 @@ jobs:
       with:
         submodules: recursive
     - name: Setup python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v3
       with:
         python-version: '3.8'
         architecture: x64
@@ -186,7 +186,7 @@ jobs:
       - uses: actions/checkout@v2
         name: Checkout the repository
       - name: Setup Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: '3.10'
       - name: Check keywords
@@ -204,7 +204,7 @@ jobs:
         with:
           submodules: recursive
       - name: Setup Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: '3.8'
           architecture: x64
@@ -267,7 +267,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        python-version: ['3.7', '3.9', '3.10']
+        python-version: ['3.8', '3.10', '3.11.0-alpha - 3.11.0']
       fail-fast: false
     steps:
       - uses: actions/checkout@v2
@@ -275,7 +275,7 @@ jobs:
         with:
           submodules: recursive
       - name: Setup Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python-version }}
           architecture: x64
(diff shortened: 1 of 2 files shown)


ischoegl (commented):
I think we should likely make a decision about whether or not to go to 2.6.1 rather sooner than later. My personal preference would be to go directly to 3.0.0a0, so we can start with the purge of legacy code. 

If you want to go ahead with the updated Python while punting the version bump, the review will be much easier.

bryanwweber:
@ischoegl Done! Hopefully I fixed the build setup failures too.

ischoegl (dismissed):
Thanks! This looks good to go as long as the tests pass.

On a separate note, I agree on your commit comment about Cython:

> This hasn't caught any problems yet, it's causing failures due to
changes in setuptools, and it more properly belongs in a separate
workflow because it doesn't need to run on every PR, only on commits to
main.

It may be a good idea to add a tracking feature request for the 'merge workflow' you propose and to re-address the build failures there.

bryanwweber:
> It may be a good idea to add a tracking feature request for the 'merge workflow' you propose and to re-address the build failures there.

See https://github.com/Cantera/enhancements/issues/37 ☺️

speth:
Given that the failures with Python 3.11.0-alpha have nothing at all to do with Cantera, isn't it a little premature to be adding this to our core CI suite? Maybe it would make more sense to have this as part of a workflow that runs on commits to `main`, as is being suggested for pre-release versions of Cython.

Also, where is the idea that Python 3.7 is out of support coming from? Its [official "end of life" date ](https://devguide.python.org/#branchstatus) is 2023-06-27, which is more than a year from now.

bryanwweber:
> Given that the failures with Python 3.11.0-alpha have nothing at all to do with Cantera, isn't it a little premature to be adding this to our core CI suite?

No, the first beta of 3.11 is already released (the version spec pulls the most recent in the series from alpha up to final). I don't think it's a problem to install the apt dependency for Ubuntu to build h5py, and then it should be possible to cache the wheel to avoid long build times.

> Also, where is the idea that Python 3.7 is out of support coming from? Its official "end of life" date is 2023-06-27, which is more than a year from now.

Whoops, that's my fault. I'm fine to revert that change. That said...

At the moment, I don't think there are any killer 3.8+ features that we'd want to use. However, Ubuntu 18.04 will also go out of support next year and 20.04 defaults to 3.8. Given our recent release cadence, it seems reasonable to target 3.8 as the minimum version, _especially_ if we identify anything from 3.8+ that'd be really useful.

Also, I tend to run the most recent version of Python while developing and it's quite frustrating to run into problems that are fixed by upgrading Python. It's not a great example because it requires 3.9, but it'd be really useful to chain decorators in the species molecular weight PR...

speth:
> No, the first beta of 3.11 is already released (the version spec pulls the most recent in the series from alpha up to final). I don't think it's a problem to install the apt dependency for Ubuntu to build h5py, and then it should be possible to cache the wheel to avoid long build times.

I'm expecting that this will be just the first of many issues before the stable version of Python 3.11 is released 5 months from now, and I'd rather we didn't have to deal with those in random, unrelated PRs. I feel like the logic for moving the "cython-latest" test out of the main test suite is pretty applicable here as well.


> At the moment, I don't think there are any killer 3.8+ features that we'd want to use. However, Ubuntu 18.04 will also go out of support next year and 20.04 defaults to 3.8. Given our recent release cadence, it seems reasonable to target 3.8 as the minimum version, _especially_ if we identify anything from 3.8+ that'd be really useful.

I'm a little disappointed that I didn't catch the fact that Ubuntu 18.04, which is still in its support period, runs Python 3.6 and that we lost support for that in Cantera 2.6 and weren't able to provide those Ubuntu packages. I'd rather not set the precedent of having Cantera drop support for Python versions that are still supported.

Also, dropping support for Python 3.7 would lose us the feature that we _just_ gained of being able to run Cantera in Google Colab.

bryanwweber:
> I'm expecting that this will be just the first of many issues before the stable version of Python 3.11 is released 5 months from now, and I'd rather we didn't have to deal with those in random, unrelated PRs. I feel like the logic for moving the "cython-latest" test out of the main test suite is pretty applicable here as well.

Sure, I'll drop the 3.11 stuff for now.

> I'm a little disappointed that I didn't catch the fact that Ubuntu 18.04, which is still in its support period, runs Python 3.6 and that we lost support for that in Cantera 2.6 and weren't able to provide those Ubuntu packages.

Can you talk about why you feel this? 18.04 will go out of support next year, so we're not missing a big chunk of the window. Do we know much about the users who download from the PPA? I'd be really interested to know why they prefer that as opposed to conda/pip. I think conda/pip are good enough from an end user perspective now, I really don't see the point of installing user-land applications that rely on frequently updated dependencies with a rarely updated system package manager.

I think supporting these really old versions of Python is a lot of weight that we carry, even if only in CI time or cognitive overhead of making sure we don't break old versions. I'd be very interested in pushing this forward a bit faster.

As one example [which I'm sure you're aware of, but just to do out the math], [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) (which is followed by most of the PyData community, including IPython, Matplotlib, pandas, SciPy, and more) suggests supporting the minor versions of Python released in the 42 months prior to our current release. As Python 3.7 was [released in June 2018](https://endoflife.date/python), NEP 29 suggests that we should have dropped 3.7 support in Cantera 2.6! Looking forward to Cantera 3.0, and assuming we release in roughly a year, NEP 29 suggests we drop 3.8 (released Oct. 2019, so dropped after April 2023 per NEP 29).

I don't think we need to adopt NEP 29 explicitly. However, to the extent that we're part of the PyData community and rely on many of those packages as dependencies, I think that we drop old versions of Python too slowly, especially now that CPython has moved to an annual release cadence.

> I'd rather not set the precedent of having Cantera drop support for Python versions that are still supported.

Assuming that Cantera 3.0 will be released in ~1 year, there will be a month of support left for Python 3.7. That hardly seems worth it... Also, supporting 5 (3.7-3.11) or 6 (3.7-3.12) minor versions of Python (depending on the release timing) seems a little ridiculous. Already for 2.6, we're going to be supporting the five minor versions, 3.7-3.11 assuming that we can build the conda and/or pip packages with the 2.6 branch when 3.11 is released.

> Also, dropping support for Python 3.7 would lose us the feature that we _just_ gained of being able to run Cantera in Google Colab.

Ugggggggghhhhhhhh 😢 You'd think one of the richest companies on Earth would be able to spend the time updating to a more recent version of Python, but noooooo...

speth:
> Can you talk about why you feel this? 18.04 will go out of support next year, so we're not missing a big chunk of the window.

To be clear, I'm talking about the current release -- 18.04 is today still supported by Canonical, and the current Cantera release can't be built using its system Python installation. While there are a lot of other distributions / OSes, I think Ubuntu is a pretty useful reference for how fast to move things forward. There are certainly other distros that are still supporting even older versions of Python (for instance, Debian `buster` is still fully supported until 8/2021 with an LTS period extending until 6/2024, and that's running Python 3.7).

> Do we know much about the users who download from the PPA? I'd be really interested to know why they prefer that as opposed to conda/pip.

Python isn't the only language we support. I'd at least like for the PPA to be a useful option for people who want to compile an link other codes against it, and I just don't think we gained that much by making Python 3.7 a _build_ requirement.

> As one example [which I'm sure you're aware of, but just to do out the math], [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) (which is followed by most of the PyData community, including IPython, Matplotlib, pandas, SciPy, and more) suggests supporting the minor versions of Python released in the 42 months prior to our current release.

You neglected two key words in NEP 29 (emphasis mine):
> When a project releases a new major or minor version, we recommend that they support **at least** all minor versions of Python introduced and released in the prior 42 months

So even if we were strictly following NEP 29, we could absolutely opt to extend that interval a small amount.

> Assuming that Cantera 3.0 will be released in ~1 year, there will be a month of support left for Python 3.7.

Maybe it's excessive optimism on my part to think that we might get 3.0 out in less than a year. So if our plan is to wait at least until next summer, then I think we'd be okay with 3.8 being the minimum supported version, based on the status of both official python.org support and supported Ubuntu versions. 

But certainly any 2.6.x version should keep 3.7 as the minimum version.

bryanwweber:
> Python isn't the only language we support. I'd at least like for the PPA to be a useful option for people who want to compile an link other codes against it, and I just don't think we gained that much by making Python 3.7 a _build_ requirement.

OK, this makes sense to me. I forget that the PPA has to be built with all the system packages, even if we don't support older Pythons for the Python interface. Sorry about that, I think this would be worth writing in a comment above the build-time check. I'll add it.

I _think_ the only need for Python 3.7 in building is for the `__annotations__` in the `buildutils.py` file. That should be easy enough to revert on the 2.6 branch, if we end up releasing 2.6.1. That should at least let us build the C/C++ interfaces on 18.04.

> Maybe it's excessive optimism on my part to think that we might get 3.0 out in less than a year.

Hey, if we can do this, I'm all for it! I guess it depends on how many breaking changes we want to try and squeeze in, aside from the input format.

> But certainly any 2.6.x version should keep 3.7 as the minimum version.

Agreed, I'm assuming that any further development for 2.6 will happen on the branch, not on main (except insofar as commits get cherry-picked back to the 2.6 branch).

> NEP 29...

We've been going back and forth for a long time about the pace that we drop Python versions 😃 I personally don't read much in to the **at least** in the text, obviously, since I left it out of my point. My main concern is supporting 5+ versions simultaneously, which either increases CI time or the cognitive overhead of feature development, or both. You're absolutely right to point out that there's a difference between the Python interface and the build system; however, I think a growing divergence between those versions will also be suboptimal. I don't know how to manage that, other than using a different build system, but I'm not taking that on!

I think the changes in this PR aren't needed at this time then, so I'll close it. @ischoegl, if you want to cherry-pick or just recreate the commit dropping the Cython test onto your version bump branch, please feel free!

ischoegl:
@speth and @bryanwweber … I likewise hope that the release cycle for 3.0 will be shorter. Regarding cython CI, I cherry-picked as suggested. Beyond, I agree that keeping old Python versions around until they are officially out of support is a good idea, even if it blocks some new features.

PR #1291, opened 2022-05-24
Remove deprecated CTI & XML formats

speth:
**Changes proposed in this pull request**

- Remove all functionality related to the deprecated XML and CTI formats
- Remove other code deprecated in Cantera 2.6, with the exception of items related to Kinetics / Reaction refactoring
- Fix the Fortran interface so interface phases can be imported using YAML input files

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

PR #1292, opened 2022-05-25
Remove legacy kinetics

### LINKED ISSUE #149: Remove legacy Kinetics infrastructure for Cantera 3.0
**Abstract**

Cantera 2.6 introduces a major redesign of how reaction rates are handled internally to improve extensibility and enable new capabilities. After release of 2.6, the "legacy" kinetics infrastructure is slated for removal. At the same time, the CTI/XML input format will be removed. 

With Cantera 2.6 being the last version that fully supports CTI/XML, this means that substantial amounts of legacy code will be removed for Cantera 3.0. 

**Possible Solutions**

The following roadmap is proposed:

1. Remove CTI/XML support first, while retaining `legacy-` reaction implementations (which can still be instantiated from YAML) … see Cantera/cantera#1291
2. Remove legacy `Kinetics` handlers second … see Cantera/cantera#1292
3. Resume work towards Cantera 3.0

**References**

New in Cantera 2.6:

- #63 
- #87 
- #139

Work deferred to Cantera 3.0:

- #142
- Cantera/cantera#1211
- Cantera/cantera#1217

ischoegl:

- Remove infrastructure used by legacy Kinetics evaluators
- Legacy code was used by objects imported from CTI/XML input (removed in #1291); new infrastructure is used by default when importing from YAML since Cantera 2.6.0

Closes Cantera/enhancements#149

**Checklist**


SConstruct
```diff
@@ -581,19 +581,11 @@ config_options = [
...
-           Flow', Wiley Interscience, 2003).""",
-        True),
-    BoolOption(
-        "no_legacy_reactions",
-        """If disabled ('no'/default), legacy 'Reaction' and associated rate objects
-           that are deprecated in Cantera 2.6 are used. If enabled ('yes'), internal
-           objects will use new objects introduced in Cantera 2.6. The flag is used
-           for testing purposes only and has no effect on results.""",
+           for three-body reactions, which corresponds to a legacy implementation used
+           prior to Cantera 3.0. The option is disabled by default, where rate
```
```suggestion
           prior to Cantera 3.0. The option is disabled by default, such that rate
```

include/cantera/kinetics/ChebyshevRate.h
```diff
@@ -138,7 +138,22 @@ class ChebyshevRate final : public ReactionRate
...
+                dotProd_[i] = m_coeffs(i, 0);
+            }
+            for (size_t j = 1; j < m_coeffs.nColumns(); j++) {
+                Cnp1 = 2 * Pr * Cn - Cnm1;
+                for (size_t i = 0; i < m_coeffs.nRows(); i++) {
+                    dotProd_[i] += Cnp1 * m_coeffs(i, j);
+                }
+                Cnm1 = Cn;
+                Cn = Cnp1;
+            }
```
bryanwweber, line 156: Is there a way to do this calculation without looping, by some array/vector operations?

include/cantera/kinetics/ChebyshevRate.h
```diff
@@ -242,19 +215,6 @@ class ChebyshevRate final : public ReactionRate
         return m_coeffs.nRows();
     }
 
-    //! Access the ChebyshevRate coefficients.
-    /*!
-     *  \f$ \alpha_{t,p} = \mathrm{coeffs}[N_P*t + p] \f$ where
-     *  \f$ 0 <= t < N_T \f$ and \f$ 0 <= p < N_P \f$.
-     *
```
bryanwweber, line 249: I think it'd be nice if these equations could be moved down to the docstring for `data()`.

include/cantera/kinetics/Kinetics.h
```diff
@@ -1000,19 +1000,18 @@ class Kinetics
...
 
     /**
-     * Flag specifying the type of reaction. The legal values and their meaning
-     * are specific to the particular kinetics manager.
+     * String specifying the type of reaction.
      *
      * @param i   reaction index
-     *
-     * @deprecated To be changed after Cantera 2.6.
+     * @version  Method returned magic number prior to Cantera 3.0.
```
bryanwweber: I think this should be `@since`, see https://www.doxygen.nl/manual/commands.html#cmdversion

include/cantera/kinetics/PlogRate.h
```diff
@@ -115,88 +110,63 @@ class PlogRate final : public ReactionRate
...
-     *  @param shared_data  data shared by all reactions of a given type
-     */
-    double evalFromStruct(const PlogData& shared_data) {
-        return updateRC(shared_data.logT, shared_data.recipT);
-    }
+            auto iter = pressures_.upper_bound(logP_);
+            AssertThrowMsg(iter != pressures_.end(), "PlogRate::updateFromStruct",
+                        "Pressure out of range: {}", logP_);
+            AssertThrowMsg(iter != pressures_.begin(), "PlogRate::updateFromStruct",
+                        "Pressure out of range: {}", logP_);
```
bryanwweber: The indentation here doesn't look quite right, the second line might need a few extra spaces? GitHub can be weird about this sometimes though.

interfaces/cython/cantera/composite.py
```diff
@@ -991,8 +990,8 @@ def set_equivalence_ratio(self, phi, *args, **kwargs):
         to be matched to the `SolutionArray`.
         """
 
-        # If ``phi`` is lower-dimensional than the SolutionArray's  shape (for 
-        # example, a scalar), broadcast it to have the right number of 
+        # If ``phi`` is lower-dimensional than the SolutionArray's  shape (for
```
```suggestion
        # If ``phi`` is lower-dimensional than the SolutionArray's shape (for
```

interfaces/cython/cantera/reaction.pyx
```diff
@@ -1041,15 +1041,11 @@ cdef class Reaction:
                 gas)
     """
     _reaction_type = ""
-    _has_legacy = False
-    _hybrid = False # indicate whether legacy implementations are separate or merged
 
-    def __cinit__(self, reactants=None, products=None, rate=None, *, legacy=False,
+    def __cinit__(self, reactants=None, products=None, rate=None, *,
                   init=True, **kwargs):
```
bryanwweber, line 1046: I wonder if this can collapse to one line now?

interfaces/cython/cantera/test/test_reaction.py
```diff
@@ -1123,30 +1089,16 @@ def test_no_rate(self):
...
-                               reactions=[rxn], adjacent=self.adj)
-            sol2.TPX = self.soln.TPX
-            self.assertNear(sol2.forward_rate_constants[0], 0.)
-            self.assertNear(sol2.net_rates_of_progress[0], 0.)
-        else:
-            with self.assertRaisesRegex(ct.CanteraError, "validate"):
-                ct.Solution(thermo=self.soln.thermo_model,
-                            kinetics=self.soln.kinetics_model,
-                            species=self.species, reactions=[rxn], adjacent=self.adj)
+        with self.assertRaisesRegex(ct.CanteraError, "validate"):
```
bryanwweber: I'm not sure where you're at with `pytest` features, but you could do `with pytest.raises(ct.CanteraError, match="validate"):`

interfaces/cython/cantera/test/test_reaction.py
```diff
@@ -1540,81 +1368,22 @@ def eval_rate(self, rate):
...
-            with self.assertWarnsRegex(DeprecationWarning, "Setter is replaceable"):
-                rxn.rates = rates
-            with self.assertWarnsRegex(DeprecationWarning, "property is moved"):
-                self.check_rates(rxn.rates, TestPlog2._rate)
-
-
-class TestPlog(TestPlog2):
-    # test updated version of Plog reaction
+class TestPlog(ReactionTests, utilities.CanteraTest):
+    # test Plog reaction
```
bryanwweber, line 1372: These comments seem redundant, you can probably remove them to improve the +/- lines ratio 😂

interfaces/cython/cantera/test/test_reaction.py
```diff
@@ -1695,6 +1439,10 @@ class TestChebyshev(TestChebyshev2):
         - [0.3177, 0.26889, 0.094806, -7.6385e-03]
         """
 
+    @classmethod
+    def setUpClass(cls):
+        ReactionTests.setUpClass()
```
bryanwweber: I _think_ if you do this, `CanteraTest.setUpClass()` won't be called, which may not be intentional. If it is, it'd be good a add comment about why. I'm not sure if `ReactionTests.setUpClass()` will call `CanteraTest.setUpClass()` though.

interfaces/cython/cantera/test/test_reaction.py
```diff
@@ -1954,17 +1633,14 @@ class StickReactionTests(InterfaceReactionTests):
 
     def finalize(self, rxn):
         rxn = super().finalize(rxn)
-        if not self._legacy:
-            weight = self.gas.molecular_weights[self.gas.species_index(self._sticking_species)]
-            rxn.rate.sticking_species = self._sticking_species
-            rxn.rate.sticking_order = self._sticking_order
-            rxn.rate.sticking_weight = weight
+        weight = self.gas.molecular_weights[self.gas.species_index(self._sticking_species)]
```
bryanwweber, line 1632: Hopefully, this could be replaced by `self.gas.species(self._sticking_species).molecular_weight` 😉

interfaces/cython/cantera/test/test_reaction.py
```diff
@@ -1973,18 +1649,11 @@ def from_parts(self):
...
-        if self._legacy:
-            pytest.skip("Legacy: interface rate does not include sticking terms")
-        super().test_rate()
-
-    @pytest.mark.skip("Legacy: construction from parts is not tested")
-    def test_from_parts(self):
-        super().test_from_parts()
+    # @pytest.mark.skip("Legacy: construction from parts is not tested")
+    # def test_from_parts(self):
+    #     super().test_from_parts()
```
bryanwweber: Should this be deleted?

interfaces/cython/cantera/utils.pyx
```diff
@@ -84,21 +84,13 @@ def use_legacy_rate_constants(pybool legacy):
...
-        three-body reactions are multiplied with third-body concentrations
-        (no change to legacy behavior). After Cantera 2.6, results will no longer
-        include third-body concentrations and be consistent with conventional
-        definitions (see Eq. 9.75 in Kee, Coltrin and Glarborg, 'Chemically
-        Reacting Flow', Wiley Interscience, 2003).
+    If set to `False` (default value), rate constants of three-body reactions are
+    consistent with conventional definitions (for example Eq. 9.75 in Kee, Coltrin
+    and Glarborg, 'Chemically Reacting Flow', Wiley Interscience, 2003). If set to
+    `True`, output for rate constants of three-body reactions is multiplied by
+    third-body concentrations (legacy behavior prior to Cantera 3.0).
```
bryanwweber:
```suggestion
    third-body concentrations, consistent with Cantera's behavior prior to version 3.0.
```

interfaces/cython/cantera/utils.pyx
```diff
@@ -84,21 +84,13 @@ def use_legacy_rate_constants(pybool legacy):
...
+    For user-compiled Cantera, the default behavior can be changed by the SCons flag
+    ``legacy_rate_constants``.
```
bryanwweber: I think this can be deleted. If you're reading the Python docstrings, you're already well past the compiling stage...

src/kinetics/Kinetics.cpp
```diff
@@ -391,16 +351,14 @@ double Kinetics::productStoichCoeff(size_t kSpec, size_t irxn) const
...
-        "Kinetics::reactionTypeStr during transition.");
-    return m_reactions[i]->reaction_type;
+std::string Kinetics::reactionType(size_t i) const {
+    return m_reactions[i]->type();
 }
 
 std::string Kinetics::reactionTypeStr(size_t i) const {
-    return m_reactions[i]->type();
+    warn_deprecated("Kinetics::reactionTypeStr",
+        "To be removed after Cantera 3.0.");
```
bryanwweber: This can probably fit on one line.

test/kinetics/kineticsFromScratch.cpp
```diff
@@ -51,11 +52,12 @@ class KineticsFromScratch : public testing::Test
...
 {
     // reaction 0:
-    // reaction('O + H2 <=> H + OH', [3.870000e+01, 2.7, 6260.0])
+    //     equation: O + H2 <=> H + OH  # Reaction 1
+    //     rate-constant: {A: 38.7, b: 2.7, Ea: 6260.0}
     Composition reac = parseCompString("O:1 H2:1");
     Composition prod = parseCompString("H:1 OH:1");
-    Arrhenius2 rate(3.87e1, 2.7, 6260.0 / GasConst_cal_mol_K);
-    auto R = make_shared<ElementaryReaction2>(reac, prod, rate);
+    auto rate = make_shared<ArrheniusRate>(3.87e1, 2.7, 2.619184e+07);
```
bryanwweber, line 59: Any reason to change from `6260.0 / GasConst_cal_mol_K`? The nice thing about that is the visible float number is the same as in the comment.

test/kinetics/kineticsFromScratch.cpp
```diff
@@ -93,56 +97,59 @@ TEST_F(KineticsFromScratch, skip_undefined_third_body)
...
-    std::multimap<double, Arrhenius2> rates {
-        { 0.01*101325, Arrhenius2(1.212400e+16, -0.5779, 10872.7 / GasConst_cal_mol_K) },
-        { 1.0*101325, Arrhenius2(4.910800e+31, -4.8507, 24772.8 / GasConst_cal_mol_K) },
-        { 10.0*101325, Arrhenius2(1.286600e+47, -9.0246, 39796.5 / GasConst_cal_mol_K) },
-        { 100.0*101325, Arrhenius2(5.963200e+56, -11.529, 52599.6 / GasConst_cal_mol_K) }
+    std::multimap<double, ArrheniusRate> rates {
+        { 0.01*101325, ArrheniusRate(1.212400e+16, -0.5779, 10872.7 * 4184.0) },
+        { 1.0*101325, ArrheniusRate(4.910800e+31, -4.8507, 24772.8 * 4184.0) },
+        { 10.0*101325, ArrheniusRate(1.286600e+47, -9.0246, 39796.5 * 4184.0) },
+        { 100.0*101325, ArrheniusRate(5.963200e+56, -11.529, 52599.6 * 4184.0) }
```
bryanwweber, line 149: I think if these are going to be replaced, they should be consistent with how other E_a are calculated.

test/kinetics/kineticsFromScratch.cpp
```diff
@@ -445,25 +458,25 @@ class KineticsAddSpecies : public testing::Test
         kin.getRevRateConstants(k.data());
         kin_ref->getRevRateConstants(k_ref.data());
         for (size_t i = 0; i < kin.nReactions(); i++) {
-            EXPECT_DOUBLE_EQ(k_ref[i], k[i]) << "i = " << i << "; N = " << N;
+            EXPECT_NEAR(k_ref[i], k[i], k_ref[i]*1e-12) << "i = " << i << "; N = " << N;
```
bryanwweber, line 461: Does this mean there's a relative tolerance of 1e-12? Why change from `EXPECT_DOUBLE_EQ`?

test/kinetics/kineticsFromScratch.cpp
```diff
@@ -384,19 +393,20 @@ TEST_F(InterfaceKineticsFromScratch, unbalanced_sites)
...
-    Arrhenius2 rate(5e21, 0, 100.0e6 / GasConstant);
-
-    auto R = make_shared<InterfaceReaction2>(reac, prod, rate);
+    auto rate = make_shared<InterfaceArrheniusRate>(5e21, 0, 100.0e6);
+    auto R = make_shared<Reaction>(reac, prod, rate);
     ASSERT_THROW(kin.addReaction(R), CanteraError);
 }
 
-class KineticsAddSpecies : public testing::Test
+class KineticsAddSpecies3 : public testing::Test
```
bryanwweber: I see this is in the test suite; nonetheless, it's the opposite of all the changes you're making elsewhere. I don't see any tests named `KineticsAddSpecies` (I probably missed it though...), so why this change?

bryanwweber (commented):
Thanks for this @ischoegl! I love the ratio of +/- lines in these last few PRs 😃 I've gotta work on that myself! Just a few minor comments about formatting here. I'll let @speth review the more substantive changes to the C++ code, since I'm not all that familiar with the appropriate data types, etc.

bryanwweber:
One other question I wanted to ask, is it necessary to keep the option to change the default status of the legacy rate constants at compile time? I think it makes sense to keep the ability to change that on the fly, but I don't think it's super helpful to keep the compile-time option. I'm worried that it will lead some research group to install a modified copy on their server, and we'll get questions on the UG that we can't figure out because the end-user doesn't know that option was changed.

include/cantera/kinetics/ChebyshevRate.h
speth: No, it's a second-order recurrence relation, where C<sub>n+1</sub> depends on C<sub>n</sub> and C<sub>n-1</sub>, so the outer loop must be done sequentially. Also, while the inner loop could theoretically be vectorized, these variables are using types (`std::vector` and `Cantera::Array2D`) that don't implement vectorized operations.

bryanwweber: Thanks for explaining @speth!

test/kinetics/kineticsFromScratch.cpp
ischoegl: There wasn't supposed to be a change ... `git` wasn't smart enough to track a renamed file that contained the new code.

test/kinetics/kineticsFromScratch.cpp
ischoegl: I believe ... this change was carried over from a renamed file, so I actually didn't introduce this. ~Trying to not do everything in one commit so hopefully this clears up.~ … *git refuses to cooperate*

test/kinetics/kineticsFromScratch.cpp
ischoegl: There was no change - this is an artifact introduced by git.

test/kinetics/kineticsFromScratch.cpp
ischoegl: There was no change - this is an artifact introduced by git.

interfaces/cython/cantera/test/test_reaction.py
ischoegl: Good catch - this was presumably left over from an earlier version that added other tasks. Should no longer be necessary ...

interfaces/cython/cantera/test/test_reaction.py
ischoegl: yup, this is one of the annoying workarounds that are still necessary

interfaces/cython/cantera/test/test_reaction.py
ischoegl: I added a comment for each of the classes (and most methods unless they are overloaded), so I'd like to keep this

interfaces/cython/cantera/reaction.pyx
ischoegl: Not yet - 91 characters.

interfaces/cython/cantera/composite.py
ischoegl: That's what you get for deblanking :joy:

interfaces/cython/cantera/test/test_reaction.py
bryanwweber: Why? It doesn't add anything that isn't already available from the test class name and inheritance. It's not really hurting anything, I suppose, but I don't see a reason to keep it either...

include/cantera/kinetics/Kinetics.h
ischoegl: This was an attempt to differentiate in analogy to Sphinx' `@versionchanged` - but I changed it ...

include/cantera/kinetics/ChebyshevRate.h
ischoegl: The new version uses a 2D array rather than the flattened coefficient vector, so it's not an exact correspondence. I believe the docstring of `data()` is adequate as it is.

include/cantera/kinetics/Kinetics.h
bryanwweber: I don't think Doxygen supports that idea... the `\version` command is supposed to take just a version, or at least, that's how it's shown in the example.

interfaces/cython/cantera/test/test_reaction.py
ischoegl: I think the main reason for keeping this is consistency - I'd prefer to keep this as it's not hurting anything.

ischoegl:
@bryanwweber ... thanks for the prompt review! And I was able to shave a couple more lines ... removing this was definitely easier than coding up what replaces the legacy code :joy:

> One other question I wanted to ask, is it necessary to keep the option to change the default status of the legacy rate constants at compile time? I think it makes sense to keep the ability to change that on the fly, but I don't think it's super helpful to keep the compile-time option. I'm worried that it will lead some research group to install a modified copy on their server, and we'll get questions on the UG that we can't figure out because the end-user doesn't know that option was changed.

Fair point - I removed this in the last commit.

src/base/application.h
```diff
@@ -361,21 +361,11 @@ class Application
...
-     *              concentrations (no change to legacy behavior). After Cantera 2.6,
-     *              results will no longer include third-body concentrations and be
-     *              consistent with conventional definitions (see Eq. 9.75 in
-     *              Kee, Coltrin and Glarborg, 'Chemically Reacting Flow', Wiley
-     *              Interscience, 2003).
+     * If set to 'false' (default value), rate constants of three-body reactions are
+     * consistent with conventional definitions (for example Eq. 9.75 in Kee, Coltrin
+     * and Glarborg, 'Chemically Reacting Flow', Wiley Interscience, 2003). If set to
+     * 'true', output for rate constants of three-body reactions is multiplied by
+     * third-body concentrations (legacy behavior prior to Cantera 3.0).
```
bryanwweber:
```suggestion
     * third-body concentrations, consistent with Cantera's behavior prior to version 3.0.
```

test/kinetics/kineticsFromScratch.cpp
bryanwweber: Ah, I see, it's a result of copying `kineticsFromScratch3.cpp` to this file. Hmm, I still think it makes sense to use the number that's listed in the comment, but I can see the other side too.

include/cantera/kinetics/Arrhenius.h
```diff
@@ -229,6 +225,7 @@ class ArrheniusRate : public ArrheniusBase
     }
 };
 
+typedef ArrheniusRate Arrhenius;
```
speth: Do we want this typedef, or do we just want to consistently use `ArrheniusRate`? I only find one case where this is used (in the test suite). The same thing applies to the typedef for `Plog`.

include/cantera/kinetics/Kinetics.h
```diff
@@ -1067,24 +1059,18 @@ class Kinetics
...
      *
-     * The computed values include all temperature-dependent, pressure-dependent,
-     * and third body contributions. Length is the number of reactions. Units are
-     * a combination of kmol, m^3 and s, that depend on the rate expression for
-     * the reaction. Note, this routine will return rate constants for
-     * irreversible reactions if the default for `doIrreversible` is overridden.
+     * The computed values include all temperature-dependent and pressure-dependent
+     * contributions. By default, third-body concentrations are only considered if
+     * they are part of the reaction rate definition; for a legacy implementation that
+     * includes third-body concentrations @see Cantera::use_legacy_rate_constants().
```
speth: This use of `@see` doesn't do what you want -- it breaks the sentence and adds a "See also" subheading. Take a look at the generated HTML. Same for `getFwdRateConstants` above.

src/fortran/cantera_kinetics.f90
```diff
@@ -114,11 +114,12 @@ double precision function ctkin_productStoichCoeff(self, k, i)
...
     end function ctkin_productstoichcoeff
 
-    integer function ctkin_reactionType(self, i)
+    integer function ctkin_reactionType(self, i, nm)
       implicit none
       type(phase_t), intent(in) :: self
       integer, intent(in) :: i
-      ctkin_reactiontype = kin_reactiontype(self%kin_id, i)
+      character*(*), intent(out) :: nm
+      ctkin_reactiontype = kin_reactiontype(self%kin_id, i, nm)
```
speth: I think the pattern this needs to follow is the one of `ctkin_getReactionString` (and I guess, the new function name ought to be `ctkin_getReactionType`).

It's a little unfortunate to have an abrupt API change like this, but I guess this is a rarely-used function and it is more allowable given we're jumping to a new major version.

src/clib/ct.cpp
```diff
@@ -988,12 +988,12 @@ extern "C" {
         }
     }
 
-    int kin_reactionType(int n, int i)
+    int kin_reactionType(int n, int i, size_t lennm, char* nm)
```
speth: I think this should likewise be renamed `kin_getReactionType`.

src/kinetics/Kinetics.cpp
```diff
@@ -755,14 +712,6 @@ void Kinetics::modifyReaction(size_t i, shared_ptr<Reaction> rNew)
             rOld->type(), rNew->type());
     }
 
-    if (!(rNew->usesLegacy())) {
-        if (rNew->rate()->type() != rOld->rate()->type()) {
-            throw CanteraError("Kinetics::modifyReaction",
-                "ReactionRate types are different: {} != {}.",
-                rOld->rate()->type(), rNew->rate()->type());
-        }
-    }
```
speth, line 764: Shouldn't this be retained, but always take place now? I didn't think switching the rate parameterization worked.

speth (changes requested):
Thanks, @ischoegl. I likewise enjoy the net decrease in lines of code here.

Besides a couple of specific comments below, I think there are several cases where we need to keep the `Whatever3` names as typedefs or passthroughs, at least for the next version. This is important for enabling code that is built to work with Cantera 2.6 without using any deprecated features to still work with the development version of Cantera 3.0. For example:
  - `ThirdBodyCalc3`
  - `Reaction::calculateRateCoeffUnits3`
  - `ThreeBodyReaction3`
  - `FalloffReaction3`
  - `Kinetics.reactant_stoich_coeffs3` and `product_stoich_coeffs3`

src/fortran/cantera_kinetics.f90
ischoegl: :+1: ... I agree that this is likely seldom used and ok with a major jump. But I *think* a deprecation warning has been triggered in 2.6, although the suggested workaround wasn't available for clib and f90.

include/cantera/kinetics/Kinetics.h
ischoegl: Makes sense. I used it here as some of the current documentation uses `@see` at the end of a sentence, which led me to believe this would work.

include/cantera/kinetics/Arrhenius.h
ischoegl: I think we can drop these.

src/kinetics/Kinetics.cpp
ischoegl: Good catch!

ischoegl:
@speth ... thanks for the review! I added all suggestions (except for `Kinetics.reactant_stoich_coeffs3` and `product_stoich_coeffs3` pass-throughs, which were already in place).

I *think* this is ready to go ...

bryanwweber (changes requested):
One more small doc change for consistency 🙂

test/kinetics/kineticsFromScratch.cpp
ischoegl: I’d prefer to leave this, as no actual code changes were introduced in this PR.

src/fortran/cantera_kinetics.f90
speth: Good point, users would have gotten a warning from the C++ layer, so the only difficulty is not having a soft transition path.

include/cantera/kinetics/Kinetics.h
```diff
@@ -581,7 +581,7 @@ class Kinetics
     //!  - `_ddX`: derivative with respect to species mole fractions (a matrix)
     //!
     //! Settings for derivative evaluation are set by keyword/value pairs using
-    //! the methods @see getDerivativeSettings and @see setDerivativeSettings.
+    //! the methods @see getDerivativeSettings and setDerivativeSettings.
```
speth: I think all you need here is 
```suggestion
    //! the methods getDerivativeSettings() and setDerivativeSettings().
```

src/fortran/cantera_kinetics.f90
ischoegl: True. It’s an interesting situation that apparently is only infrequent as the clib and f90 interfaces are incomplete.

ischoegl: On a separate note, we dropped a comparable interface from Python as it received a much more detailed reaction interface. I still think that the current change is the best short-time path.

speth (commented):
Thanks, @ischoegl. Just one minor suggestion for the docs.

include/cantera/kinetics/Kinetics.h
ischoegl: Done. I figure that the parentheses are required to detect methods in all cases? It may make sense to go through the documentation in a separate PR as they are often left off when used in combination with `@see`? (although doxygen may be smart enough to infer things in that case)

speth: Yeah, that's correct. The `@see` notation is just for creating a `see also` block within the documentation.

I think there are probably a lot of cases where the parens are missing when referencing functions, both with and without `@see`.

ischoegl:
@speth / @bryanwweber … let me know in case there’s anything else …

speth (approved):

bryanwweber (approved):
Thanks @ischoegl !

PR #1307, opened 2022-06-03
Use logging consistently in SCons

ischoegl:

- Replace `print` by appropriate `logging` level
- Deprecate SCons `VERBOSE` option in favor of `logging` option
- Reduce verbosity of SCons output for commands other than `scons build`

Closes N/A

```
$ scons build logging=debug
```
and
```
$ scons help --option=logging

* logging: [ 'debug' | 'info' | 'warning' | 'error' | 'default' ]
    Select logging level for SCons output. By default, logging messages use the
    'info' level for 'scons build' and the 'warning' level for all other
    commands. In case the SCons option '--silent' is passed, all messages below
    the 'error' level are suppressed.
    - default: 'default'
```

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -101,10 +101,20 @@ valid_commands = ("build", "clean", "install", "uninstall",
                   "help", "msi", "samples", "sphinx", "doxygen", "dump",
                   "sdist")
 
+# set default logging level
+if GetOption("silent"):
+    logger.logger.setLevel("ERROR")
+else:
+    logger.logger.setLevel("WARNING")
+
 for command in COMMAND_LINE_TARGETS:
     if command not in valid_commands and not command.startswith('test'):
         logger.error("Unrecognized command line target: {!r}", command)
         sys.exit(1)
+
+    # update default logging level
+    if command in ["build"] and not GetOption("silent"):
+        logger.logger.setLevel("INFO")
 
 if "clean" in COMMAND_LINE_TARGETS:
     remove_directory("build")
@@ -512,7 +522,7 @@ config_options = [
     EnumOption(
         "logging",
         """Select logging level for SCons output. """,
-        "info", ("debug", "info", "warning", "error")),
+        "default", ("debug", "info", "warning", "error", "default")),
     Option(
         "gtest_flags",
         """Additional options passed to each GTest test suite, for example,
@@ -885,7 +895,8 @@ logger.info(textwrap.indent(cantera_conf, "    "), print_level=False)
 # ********************************************
 
 loglevel = env["logging"]
-logger.logger.setLevel(loglevel.upper())
+if loglevel != "default":
+    logger.logger.setLevel(loglevel.upper())
 
 if env["VERBOSE"]:
     # @todo: Remove after Cantera 3.0
--- a/SConstruct
+++ b/SConstruct
@@ -2024,16 +2024,20 @@ if env['f90_interface'] == 'y':
 build_samples = Alias('samples', sampleTargets)
 
 def postBuildMessage(target, source, env):
-    print("*******************************************************")
-    print("Compilation completed successfully.\n")
-    print("- To run the test suite, type 'scons test'.")
-    print("- To list available tests, type 'scons test-help'.")
-    if env['googletest'] == 'none':
-        print("  WARNING: You set the 'googletest' to 'none' and all its tests will be skipped.")
-    print("- To install, type 'scons install'.")
+    build_message = [
+        f"\n{' Compilation completed successfully ':*^88}\n",
+        "- To run the test suite, type 'scons test'.",
+        "- To list available tests, type 'scons test-help'.",
+    ]
+    if env["googletest"] == "none":
+        build_message.append("  WARNING: You set the 'googletest' to 'none' "
+            "and all its tests will be skipped.")
+    build_message.append("- To install, type 'scons install'.")
     if os.name == 'nt':
-        print("- To create a Windows MSI installer, type 'scons msi'.")
-    print("*******************************************************")
+        build_message.append("- To create a Windows MSI installer, type 'scons msi'.")
+    build_message.append(f"\n{'*' * 88}\n")
+
+    logger.info("\n".join(build_message), print_level=False)
 
 finish_build = env.Command('finish_build', [], postBuildMessage)
 env.Depends(finish_build, buildTargets)
--- a/SConstruct
+++ b/SConstruct
@@ -909,15 +909,6 @@ elif env['env_vars']:
             logger.warning(f"failed to propagate environment variable '{repr(name)}'\n"
                 "Edit cantera.conf or the build command line to fix this.")
 
-# @todo: Remove this Warning after Cantera 2.5
-if os.pathsep == ';':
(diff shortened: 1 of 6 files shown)


ischoegl:
Ok. Added a couple of minor tweaks - stopping here. At this point this is ready for review.

SConstruct
```diff
@@ -789,7 +808,8 @@ elif "clang" in env.subst("$CC"):
     config.select("clang")
 
 else:
-    print(f"WARNING: Unrecognized C compiler '{env['CC']}'")
+    logger.error(f"Unrecognized C compiler {env['CC']!r}")
+    exit(1)
```
```suggestion
    sys.exit(1)
```

SConstruct
```diff
@@ -623,15 +642,15 @@ if "help" in COMMAND_LINE_TARGETS:
 
         if list:
             # show formatted list of options
-            logger.info("\nConfiguration options for building Cantera:", print_level=False)
-            logger.info(config.list_options(), print_level=False)
+            logger.status("\nConfiguration options for building Cantera:\n", print_level=False)
+            logger.status(config.list_options() + "\n", print_level=False)
```
bryanwweber, line 646: What do you see as the difference between `status` and `info`? Why did you want to add this distinction?

SConstruct
```diff
@@ -878,6 +898,15 @@ logger.info(textwrap.indent(cantera_conf, "    "), print_level=False)
 # *** Configure system-specific properties ***
 # ********************************************
 
+loglevel = env["logging"]
+if loglevel != "default":
+    logger.logger.setLevel(loglevel.upper())
```
bryanwweber, line 903: It looks like there can be a conflict between `scons build --silent logging=info` (or anything other than `default`, I suppose). Does `--silent` also silence some of the output from SCons itself? It'd be nice to unify that.

SConstruct
```diff
@@ -890,20 +919,10 @@ elif env['env_vars']:
...
-# @todo: Remove this Warning after Cantera 2.5
-if os.pathsep == ';':
-    for dirs in (env['extra_inc_dirs'], env['extra_lib_dirs']):
-        if re.search(r':\w:', dirs):
-            print('ERROR: Multiple entries in "extra_inc_dirs" and "extra_lib_dirs" '
-                  'should be separated by semicolons (;) on Windows. Use of OS-specific '
-                  'path separator introduced in Cantera 2.5.')
-            sys.exit(1)
+            logger.warning(f"failed to propagate environment variable {name!r}\n"
+                "Edit cantera.conf or the build command line to fix this.")
```
```suggestion
            logger.warning(f"Failed to propagate environment variable {name!r}\n"
                           "Edit cantera.conf or the build command line to fix this.")
```

SConstruct
```diff
@@ -1313,12 +1337,12 @@ if env['system_sundials'] == 'y':
     env['sundials_version'] = '.'.join(sundials_version.split('.')[:2])
     sundials_ver = parse_version(env['sundials_version'])
     if sundials_ver < parse_version("2.4") or sundials_ver >= parse_version("7.0"):
-        print("""ERROR: Sundials version %r is not supported.""" % env['sundials_version'])
+        logger.error(f"Sundials version {env['sundials_version']} is not supported.")
```
```suggestion
        logger.error(f"Sundials version {env['sundials_version']!r} is not supported.")
```

and likewise below, to match the previous output.

SConstruct
```diff
@@ -1351,10 +1375,10 @@ if env['system_sundials'] == 'y':
     # In the case where a user is trying to link Cantera to an external BLAS/LAPACK
     # library, but Sundials was configured without this support, print a Warning.
     if not env['has_sundials_lapack'] and env['use_lapack']:
-        print('WARNING: External BLAS/LAPACK has been specified for Cantera '
-              'but Sundials was built without this support.')
+        logger.warning("External BLAS/LAPACK has been specified for Cantera "
+              "but Sundials was built without this support.")
```
```suggestion
        logger.warning("External BLAS/LAPACK has been specified for Cantera "
                       "but Sundials was built without this support.")
```

SConstruct
```diff
@@ -1376,11 +1400,11 @@ end program main
         if success and 'Hello, world!' in output:
             return True
         else:
-            print("WARNING: Unable to use '%s' to compile the Fortran "
-                  "interface. See config.log for details." % compiler)
+            logger.warning(f"Unable to use {compiler!r} to compile the Fortran "
+                  "interface. See config.log for details.")
```
```suggestion
            logger.warning(f"Unable to use {compiler!r} to compile the Fortran "
                           "interface. See config.log for details.")
```

bryanwweber (changes requested):
Thanks @ischoegl! A few small suggestions and one question below. One further question, is it possible to capture the output from `env.Command` and send that through the logger too?

SConstruct
ischoegl: I think that most `INFO` level messages make most sense for `scons build`, but clutter output otherwise. I created status so we can trigger output at the `WARNING` level with a more appropriate nomenclature.

SConstruct
ischoegl: I tried to keep SCons settings and logger settings separate as much as possible. Fwiw, `--silent` is a native SCons command, so it does act on much of the SCons output. I still opted for tying reduced logging output to this as I felt this would make intuitive sense.
```
scons build --silent logging=info
```
will suppress SCons ouptut, but still create logging output (where the `logging=info` acts as an override). SCons also supports the `-Q` option (which is more aggressive than `--silent`), but I have not found a way to capture whether it is toggled within `SConstruct`.

ischoegl:
@bryanwweber ... thanks for the review! I think things should be addressed.

ischoegl:
@bryanwweber … could you elaborate on

> One further question, is it possible to capture the output from env.Command and send that through the logger too?

I have not tried redirecting SCons output to the logger. It’s an interesting idea, but I don’t want to add this capability yet.

*PS: something along [those lines](https://stackoverflow.com/questions/19425736/how-to-redirect-stdout-and-stderr-to-logger-in-python) may be possible, but I think this goes beyond the intent of this PR*

bryanwweber:
> @bryanwweber … could you elaborate on
> 
> > One further question, is it possible to capture the output from env.Command and send that through the logger too?
> 
> I have not tried redirecting SCons output to the logger. It’s an interesting idea, but I don’t want to add this capability yet.
> 
> _PS: something along [those lines](https://stackoverflow.com/questions/19425736/how-to-redirect-stdout-and-stderr-to-logger-in-python) may be possible, but I think this goes beyond the intent of this PR_

I'm not sure how `env.Command` works internally in SCons... assuming it's something like `subprocess.run`, you can capture the output and use `logger.info` etc, to display it. But if it isn't a relatively small change, then it may be better to address separately.

ischoegl:
> But if it isn't a relatively small change, then it may be better to address separately.

I don’t think that we have direct access to that output. Capturing and redirecting what SCons does internally is something where I honestly don’t know how much effort it takes. I think it’s better to leave that for another PR.

ischoegl:
@bryanwweber ... any additional comments?

bryanwweber (approved):
Thanks @ischoegl LGTM!

PR #1316, opened 2022-06-05
Add auto-detection of Intel MKL and OpenBLAS

### LINKED ISSUE #144: Use BLAS/LAPACK by default if installed
**Abstract**

While many *'system'*-installed packages are detected and used by default (e.g. `Eigen`, `fmt`, `sundials`, `yaml-cpp`), the same is not true for BLAS and LAPACK. Instead, it does require the `blas_lapack_libs` flag (as expected, there currently is no `system_blas_lapack` option).

While there are some issues related to incompatible versions packaged with MATLAB, it would make sense to use BLAS and LAPACK by default otherwise (instead of Eigen's linear algebra support). Based on communication with @speth there are couple of caveats to consider:

>  [...] there are a number of special cases that you would want handle. One is that one of the best BLAS/LAPACK implementations to link to would be the MKL if it’s present. The other is that if you’re trying to autodetect, you probably want to try figuring out if it’s a version that uses an 8-byte index type or not (which is the Matlab issue).

ischoegl:

The following priority is used:

1. Use Intel MKL if installed 
2. Use OpenBLAS if installed
3. Use `lapack,blas` if installed
4. Use prior behavior

Closes Cantera/enhancements#144

```
$ scons build locate_lapack=auto
[...]
INFO: Using private installation of Sundials version 5.3.
Checking for C library mkl_rt... (cached) yes
INFO: Using MKL 2022.0.1 for Intel(R) 64 architecture
INFO: Skipping compilation of the Fortran 90 interface.
[...]
```

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -371,28 +371,27 @@ config_options = [
            for example, '/usr/lib'.""",
         "", PathVariable.PathAccept),
     EnumOption(
-        "locate_lapack",
-        """Controls whether Cantera automatically checks for installations of BLAS and
-           LAPACK libraries if the configuration is not specified by 'blas_lapack_libs'.
-           By default, 'auto' updates 'blas_lapack_libs' based on build system and
-           detected libraries. On macOS, 'auto' defaults to the Accelerate framework,
-           whereas on other operating systems the preferred options depend on the CPU
-           manufacturer. In general, OpenBLAS ('openblas') is prioritized over
-           'standard' libraries ('lapack,blas'). On Intel CPU's, MKL (Windows:
-           'mkl_rt' / Linux: 'mkl_rt,dl') has highest priority, followed by the
-           other options. Note that 'locate_lapack' is disabled if the MATLAB toolbox
-           is installed.""",
-        "auto", ("auto", "mkl", "openblas", "standard", "off")),
+        "system_blas_lapack",
+        """Select whether to use BLAS/LAPACK from a system installation ('y'), use
+           Eigen linear algebra support ('n'), or to decide automatically based on
+           libraries detected on the system ('default'). Specifying 'blas_lapack_libs'
+           or 'blas_lapack_dir' changes the default to 'y', whereas installing the
+           Matlab toolbox changes the default to 'n'. On macOS, the 'default' option
+           uses the Accelerate framework, whereas on other operating systems the
+           preferred option depends on the CPU manufacturer. In general, OpenBLAS
+           ('openblas') is prioritized over standard libraries ('lapack,blas'), with
+           Eigen being used if no suitable BLAS/LAPACK libraries are detected. On Intel
+           CPU's, MKL (Windows: 'mkl_rt' / Linux: 'mkl_rt,dl') has the highest priority,
+           followed by the other options. Note that Eigen is required whether or not
+           BLAS/LAPACK libraries are used.""",
+        "default", ("default", "y", "n")),
     Option(
         "blas_lapack_libs",
         """Cantera can use BLAS and LAPACK libraries installed on your system if you
-           have optimized versions available. Cantera automatically checks for
-           installations of BLAS and LAPACK if the 'locate_lapack' option is enabled.
-           Otherwise, Cantera will use Eigen for linear algebra support. To use specific
-           versions of BLAS and LAPACK, set 'blas_lapack_libs' to the the list of
-           libraries that should be passed to the linker, separated by commas, for
-           example, "lapack,blas" or "lapack,f77blas,cblas,atlas". Eigen is required
-           whether or not BLAS/LAPACK are used.""",
+           have optimized versions available (see option 'system_blas_lapack'). To use
+           specific versions of BLAS and LAPACK, set 'blas_lapack_libs' to the the list
+           of libraries that should be passed to the linker, separated by commas, for
+           example, "lapack,blas" or "lapack,f77blas,cblas,atlas".""",
         ""),
     PathOption(
         "blas_lapack_dir",
@@ -1024,10 +1023,6 @@ if env['system_sundials'] in ('y','default'):
 if env['blas_lapack_libs'] != '':
     env['blas_lapack_libs'] = env['blas_lapack_libs'].split(',')
     env['use_lapack'] = True
-elif env['OS'] == 'Darwin':
-    env['blas_lapack_libs'] = []
-    env['use_lapack'] = True
-    env.Append(FRAMEWORKS=['Accelerate'])
 else:
     env['blas_lapack_libs'] = []
     env['use_lapack'] = False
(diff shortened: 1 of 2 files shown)


bryanwweber:
Thanks @ischoegl I think this is an interesting proposal. I'm concerned that turning these things on by default will cause more issues like https://github.com/Cantera/conda-recipes/issues/29, especially since we also default to adding the conda directories to the path. What do you think?

ischoegl:
@bryanwweber … packaging is an interesting aspect here, as local installations were my primary concern. For the latter, I don’t see any potential issues, whereas for the former things are certainly more interesting. What do you mean by ‘especially since we also default to adding the conda directories to the path’? Having conda paths included should be desirable, no?

bryanwweber:
The problem I can foresee is that you most likely don't want to `scons install` into the same environment with all your build dependencies, or at least, I wouldn't want to, for instance by setting `python_cmd`. In that case, MKL might not be installed (or a different version is installed) in the environment where `python_cmd` is located, and you'd get errors loading the module.

Similarly, if someone has the conda directories added into their linker path, but installs into the system directories, MKL might not be found. This _should_ be prevented by the fact that setting `prefix` would disable automatic conda paths, but then they might add the conda paths manually to pick up one of our other dependencies.

ischoegl:
> The problem I can foresee is that you most likely don't want to `scons install` into the same environment with all your build dependencies, or at least, I wouldn't want to, for instance by setting `python_cmd`. In that case, MKL might not be installed (or a different version is installed) in the environment where `python_cmd` is located, and you'd get errors loading the module.

I believe most users would want to install into the same environment, see [environment.yaml](https://github.com/comocheng/cantera-build) discussion. I believe there are a couple of edge cases, which need to be identified and where auto detection may need to be disabled.

> Similarly, if someone has the conda directories added into their linker path, but installs into the system directories, MKL might not be found. This _should_ be prevented by the fact that setting `prefix` would disable automatic conda paths, but then they might add the conda paths manually to pick up one of our other dependencies.

I think that if the conda layout is used, MKL is certainly desirable, but the same is true for other layouts? Linking against dependencies that are not available in a different context is imho user error, where the same arguments apply to various ‘system’ libraries, e.g. yaml-cpp, etc.? We likewise check for existence and provide fallback options only if nothing is found?

ischoegl:
@bryanwweber ... I ended up adding an option that allows for control of the detection of 'optimized' BLAS/LAPACK.

In addition, I ran some tests with some odd results for MKL when running speed tests for ignition
```
Without MKL (private Sundials 5.3)

In [2]: %run custom_reactions.py
- New framework (YAML): 92.27 ms (T_final=2736.60)
- One Python reaction: 99.00 ms (T_final=2736.60) ... +7.30%

With MKL (private Sundials 5.3)

In [2]: %run custom_reactions.py
- New framework (YAML): 102.01 ms (T_final=2736.60)
- One Python reaction: 99.66 ms (T_final=2736.60) ... -2.31%

With OpenBLAS (private Sundials 5.3)

In [2]: %run custom_reactions.py
- New framework (YAML): 100.79 ms (T_final=2736.60)
- One Python reaction: 103.94 ms (T_final=2736.60) ... +3.12%

Without MKL (system Sundials 6.0)

In [2]: %run custom_reactions.py
- New framework (YAML): 92.19 ms (T_final=2736.60)
- One Python reaction: 99.34 ms (T_final=2736.60) ... +7.76%

With MKL (system Sundials 6.0)

In [3]: %run custom_reactions.py
- New framework (YAML): 104.20 ms (T_final=2736.60)
- One Python reaction: 102.95 ms (T_final=2736.60) ... -1.20%
```
These results indicate that the 'optimized' libraries slow things down. Further, it looks like the inclusion of Python calls (which should slow things down) is advantageous for MKL. 

*PS: There's likely overhead involved that may require fine-tuning beyond just enabling/disabling some libraries.*

ischoegl:
@speth ... in light of the [recent UG posting](https://groups.google.com/g/cantera-users/c/pFQV0rsk5As/m/TPL3GVteAgAJ) about speed differences, I am curious about your opinion on the slow-downs I observed when testing changes this PR is proposing. While the mechanism is small, I'd appreciate any clues?

SConstruct
```diff
@@ -1246,6 +1254,55 @@ else:
...
+    # MATLAB provides the mwlapack and mwblas libraries in matlabroot/extern/lib.
+    logger.info("Skip auto-detection of Intel MKL / OpenBLAS as they are not "
+                "compatible with the MATLAB toolbox.")
+
+elif env["locate_lapack"] != "off":
+    # auto-detect versions
+    if env["locate_lapack"] in ["mkl", "auto"]:
+        blas_lapack_order = ["mkl_rt", "openblas"]
+    else:
+        blas_lapack_order = ["openblas", "mkl_rt"]
```
speth: The logic here isn't what I expected from the docstring for the `locate_lapack` option. I would have only expected it to try both OpenBLAS and MKL when using the `auto` option, not for that to be just the first preference with the fallback of the other. If you want to keep the behavior as it is here, then I guess the `auto` option is redundant.

SConstruct
```diff
@@ -1246,6 +1254,55 @@ else:
 env['has_demangle'] = conf.CheckDeclaration("boost::core::demangle",
                                 '#include <boost/core/demangle.hpp>', 'C++')
 
+# check BLAS/LAPACK installations
+if env["blas_lapack_libs"]:
+    for lib in env["blas_lapack_libs"]:
+        if not conf.CheckLib(lib):
+            config_error(f"Library {lib!r} could not be found.")
+
+elif env["matlab_path"] != "" and env["matlab_toolbox"] == "default":
```
speth: Shouldn't this apply for either `default` or `y` values of the `matlab_toolbox` option?

speth (commented):
I think as implemented now, with the default being to just use Eigen, this is a reasonable addition.

Regarding the performance differences you noted, I'm not really sure what's behind them. For systems as small as GRI 3.0, the linear algebra really should not be taking that much of the compute time. The place where MKL and OpenBLAS can really beat Eigen is more likely to be in larger systems where multithreading is actually helpful.

SConstruct
ischoegl: looking over this, I believe I never finished the logic here when I toggled to draft mode. After rebasing to the current `main`, I am sticking with the description of the SCons option.

ischoegl:
Thanks for your input on this, @speth!

I updated a couple of things, most notably a new `locate_lapack=standard` option that looks for `lapack,blas`, which is also part of the `auto` sequence.

Outside of packaging, I personally don’t see major issues where auto configuration could go awry, but I’m happy to leave things ‘off’.

speth:
A couple of reasons I see for leaving autodetection off by default are
1. MKL is known to have deliberately sub-optimal performance on non-Intel (i.e. AMD) processors. 
2. On macOS, the Accelerate framework that we use by default offers a well-optimized BLAS/LAPACK implementation, and using OpenBLAS instead (or worse, a copy of the reference BLAS/LAPACK libraries installed somewhere) would likely be worse in most cases.

speth (dismissed):
Thanks, @ischoegl, I think this is good. Sorry for the small merge conflict introduced by merging #1382.

ischoegl:
Thanks for the context here.

> On macOS, the Accelerate framework that we use by default offers a well-optimized BLAS/LAPACK implementation, and using OpenBLAS instead (or worse, a copy of the reference BLAS/LAPACK libraries installed somewhere) would likely be worse in most cases.

That is interesting, as `conda-recipes` does install OpenBLAS (and specifies it as `blas_lapack_libs`). ~I had assumed that this just indicates that the Accelerate framework is supposed to be used. Is there something that is falling through the cracks somewhere?~ ... *Edit: It looks like `Accelerate` is indeed not used for `conda-recipes`.*

*PS: apart from this I rebased, which removed the approval.*

ischoegl:
@speth ... I ended up adding exceptions to the `locate_lapack=auto` settings so it won't override `Accelerate` on macOS and also avoids `mkl_rt` on non-intel CPU's.

This may address most of the objections against enabling this by default?

speth:
Here are some timing results from running a set of ignition delay calculations on a M1 Pro. Times within 5% of the fastest are highlighted in bold.

| vars | steps | OpenBlas N | OpenBLAS 1 | Accelerate N | Accelerate 1 |
|------|-------|-----------:|-----------:|-------------:|-------------:|
|   11 |  705  |    **0.002**   |    **0.002**   |      **0.002**   |      **0.002**   |
|   54 | 1786  |    **0.038**   |    **0.038**   |      **0.040**   |      **0.040**   |
|  121 | 2887  |    0.474   |    **0.281**   |      0.303   |      0.302   |
|  202 | 2491  |    1.234   |    **0.521**   |      **0.527**   |      **0.527**   |
|  231 | 2502  |    1.272   |    **0.584**   |      0.648   |      0.648   |
|  494 | 2778  |    4.811   |    2.872   |      **2.330**   |      **2.327**   |
|  655 | 2319  |    5.722   |    3.841   |      **3.489**   |      3.805   |
|  875 | 3080  |    12.336  |    9.779   |      **7.033**   |      8.391   |
| 1379 | 2200  |    18.274  |    22.034  |      **13.375**  |      15.313  |
| 2116 | 2311  |    42.599  |    76.971  |      **33.105**  |      41.256  |
| 2650 | 3047  |    77.850  |    153.194 |      **66.371**  |      86.913  |
| 3788 | 2674  |    150.124 |    379.541 |      **120.374** |      179.087 |

- For small mechanisms (GRI 3.0 or smaller), the linear algebra library basically doesn't matter.
- For medium sized mechanisms (100 - ~400 species), using OpenBLAS on a single thread seems to be the fastest. The Accelerate framework is slightly slower, but seems to be smart enough to not try any parallelization that would just slow things down. Using OpenBLAS with its default behavior of trying to use all available cores is the slowest option of all, and burns a lot of extra CPU time.
- For large mechanisms, The Accelerate framework is the fastest, and is able to get a modest speedup from parallelization for mechanisms with > ~800 species.
- In the cases where parallelization leads to an overall speedup, the Accelerate framework is much more efficient than OpenBLAS. Average CPU usage for the largest mechanisms is ~600% for OpenBLAS but only ~200% for Accelerate.

So at least for our usage, I don't think we should go out of our way to use OpenBLAS. The best performance without having to fiddle with environment variables on a problem-size dependent basis will be found using the Accelerate framework. I guess one caveat here is that this applies only to the dense factorization algorithms used in the reactor network solver. The performance of the banded matrix algorithms used in the flame solver could be different.

ischoegl:
@speth … nice!

Great to see that what you mentioned earlier is backed up by this comparison. While you were running those tests, I added some code detecting the CPU, and adapted the `auto` setting accordingly. Let me know what you think?

SConstruct
```diff
@@ -370,14 +370,23 @@ config_options = [
            Not needed if the libraries are installed in a standard location,
            for example, '/usr/lib'.""",
         "", PathVariable.PathAccept),
+    EnumOption(
+        "locate_lapack",
+        """Controls whether Cantera automatically checks for installations of BLAS and
+           LAPACK libraries. The option is disabled by default ('off'). If automatic
+           detection is enabled, libraries are added to the 'blas_lapack_libs' path.
+           For the 'auto' setting, MKL is prioritized over OpenBLAS and the 'standard'
+           option corresponding to 'lapack,blas'.""",
```
speth: I think this needs to be updated to cover the different behavior for Intel vs other processors, and for macOS.

speth (changes requested):
I think this just needs a small fix to the documentation. Otherwise, this looks good to me.

ischoegl:
@speth … thanks! Before I fix the docstring, I *believe* that at this point the `auto` option does the correct thing in all cases I can think of. So it may be safe to make it the default after all?

speth:
🤷 I'm willing to give it a try.

ischoegl:
While updating the docstrings to use 'automatic' configuration of BLAS/LAPACK by default, I realized that `system_blas_lapack` would be a better option name, as it is consistent with what is done for other external dependencies (especially sundials). I came up with the following:
```
 % scons help --option=system_blas_lapack

* system_blas_lapack: [ 'n' | 'y' | 'default' ]
    Select whether to use BLAS/LAPACK from a system installation ('y'), use
    Eigen linear algebra support ('n'), or to decide automatically based on
    libraries detected on the system ('default'). Specifying 'blas_lapack_libs'
    or 'blas_lapack_dir' changes the default to 'y', whereas installing the
    Matlab toolbox changes the default to 'n'. On macOS, the 'default' option
    uses the Accelerate framework, whereas on other operating systems the
    preferred option depends on the CPU manufacturer. In general, OpenBLAS
    ('openblas') is prioritized over standard libraries ('lapack,blas'), with
    Eigen being used if no suitable BLAS/LAPACK libraries are detected. On Intel
    CPU's, MKL (Windows: 'mkl_rt' / Linux: 'mkl_rt,dl') has the highest
    priority, followed by the other options. Note that Eigen is required whether
    or not BLAS/LAPACK libraries are used.
    - default: 'default'
```

speth:
The new docstring looks good to me. What's behind the change to disable the Fortran interface on the OneAPI builders? Can you update the commit message to explain why that was needed?

ischoegl:
> The new docstring looks good to me. What's behind the change to disable the Fortran interface on the OneAPI builders? Can you update the commit message to explain why that was needed?

Done - I created issue #1393 to document the reason for this change: the `f77-demo` test fails for optimized LAPACK libraries (MKL is now automatically detected on the OneAPI tool chain).

ischoegl:
@speth … let me know if there’s anything else to do here. I don’t want to address #1393 here as it is imho a pre-existing condition.

speth (approved):
Sure, looks good to me.

PR #1327, opened 2022-06-08
Some Transport updates

ischoegl:

The header `TransportBase.h` does not match the naming convention used for other fundamental headers, while `Transport.h` does (see, e.g. includes in `core.h` introduced in #1238)

https://github.com/Cantera/cantera/blob/dceacdfba4519ee85d4e5f60e404fe25b8e5193b/include/cantera/core.h#L12-L15

Also, streamline construction of `Transport` objects in Python.

**Checklist**


ischoegl:
@speth / @bryanwweber ... this is a trivial PR - not sure how to avoid the apparent 'churn' (which really isn't there) due to renaming + recreating a rudimentary file with the old name for deprecation purposes. I tried this in separate commits before, but it doesn't appear to make any difference ...

src/transport/Transport.cpp
```diff
@@ -6,7 +6,7 @@
 // This file is part of Cantera. See License.txt in the top-level directory or
 // at https://cantera.org/license.txt for license and copyright information.
 
-#include "cantera/transport/TransportBase.h"
+#include "cantera/transport/Transport.h"
```
speth, line 9: Above this, the Doxygen docstring needs to have `TransportBase.cpp` replaced with `Transport.cpp` (never mind that the rest of that docstring is erroneous copypasta).

include/cantera/transport/Transport.h
```diff
@@ -0,0 +1,791 @@
...
+    //! @name Transport Properties
+    //! @{
+
+    /*!
+     * The viscosity in Pa-s.
+     */
+    virtual doublereal viscosity() {
+        throw NotImplementedError("Transport::viscosity",
+            "Not implemented for transport model '{}'.", transportType());
+    }
```
speth: If Git is going to see this as a "new" file, this would be a good opportunity to just replace all instances of `doublereal` in this file with `double`, since there's no way anyone could rebase a commit touching the old `TransportBase.h` header across this change without having to make manual updates anyway.

interfaces/cython/cantera/base.pyx
```diff
@@ -274,6 +277,13 @@ cdef class _SolutionBase:
                 self.kinetics.addReaction(reaction._reaction, False)
             self.kinetics.resizeReactions()
 
+        if not transport:
+            transport = "none"
+
```
speth: I think this is unnecessary -- leaving `transport` as the empty string also results in creating a base `Transport` object here.

speth (changes requested):
Thanks for fixing this annoying inconsistency, @ischoegl. I had just a few minor suggestions.

ischoegl:
@speth ... thanks for the review! Changes are taken care of.

speth (approved):

PR #1332, opened 2022-07-05
Fix CustomRate segfaults and remove CustomReaction

### LINKED ISSUE #1282: segfault when using CustomRate
**Problem description**

Creating a reaction in Python using the base `Reaction` class and a `CustomRate` object for the rate function will segfault unless a reference to the `CustomRate` object is held elsewhere.

**Steps to reproduce**

The following will segfault:
```python
import cantera as ct
R = ct.Reaction(equation='H + OH <=> H2O', rate=ct.CustomRate(lambda T: 1.234))
gas.add_reaction(R)
```

This problem can be worked around by either doing:
```python
R = ct.CustomReaction(equation='H + OH <=> H2O', rate=ct.CustomRate(lambda T: 1.234))
```
or 
```python
ratefunc = ct.CustomRate(lambda T: 1.234)
R = ct.Reaction(equation='H + OH <=> H2O', rate=ratefunc)
```
but neither of these should be required.

**Additional Context**

I ran into this in a case where I wanted to set a `CustomRate` on an existing `Reaction` object, created from an input file. The only way I could figure out how to do this was to create an elementary reaction with some dummy rate parameters, which gives an instance of the base `Reaction` class.

**System information**

- Cantera version: 2.6.0
- OS: macOS 12.2.1
- Python/MATLAB/other software versions: Python 3.10

ischoegl:

- Remove `CustomReaction` specialization
- Fix segfaults caused by temporary Python objects

Closes #1282, partially addresses Cantera/enhancements#142

**Checklist**

### DIFF
--- a/include/cantera/kinetics/Custom.h
+++ b/include/cantera/kinetics/Custom.h
@@ -68,7 +68,7 @@ public:
     void setRateFunction(shared_ptr<Func1> f);
 
 protected:
-    shared_ptr<Func1> m_ratefunc;
+    shared_ptr<Func1> m_ratefunc = 0;
 };
 
 
--- a/include/cantera/kinetics/Custom.h
+++ b/include/cantera/kinetics/Custom.h
@@ -68,7 +68,7 @@ public:
     void setRateFunction(shared_ptr<Func1> f);
 
 protected:
-    shared_ptr<Func1> m_ratefunc = 0;
+    shared_ptr<Func1> m_ratefunc;
 };
 
 
--- a/include/cantera/kinetics/Reaction.h
+++ b/include/cantera/kinetics/Reaction.h
@@ -31,7 +31,7 @@ class ThirdBody;
 class Reaction
 {
 public:
-    Reaction();
+    Reaction() {}
     Reaction(const Composition& reactants, const Composition& products,
              shared_ptr<ReactionRate> rate={});
 
@@ -147,17 +147,17 @@ public:
     std::string id;
 
     //! True if the current reaction is reversible. False otherwise
-    bool reversible;
+    bool reversible = true;
 
     //! True if the current reaction is marked as duplicate
-    bool duplicate;
+    bool duplicate = false;
 
     //! True if reaction orders can be specified for non-reactant species.
     //! Default is `false`.
-    bool allow_nonreactant_orders;
+    bool allow_nonreactant_orders = false;
 
     //! True if negative reaction orders are allowed. Default is `false`.
-    bool allow_negative_orders;
+    bool allow_negative_orders = false;
 
     //! Input data used for specific models
     AnyMap input;
@@ -165,7 +165,7 @@ public:
     //! The units of the rate constant. These are determined by the units of the
     //! standard concentration of the reactant species' phases of the phase
     //! where the reaction occurs.
-    Units rate_units;
+    Units rate_units{0.0};
 
     //! Get reaction rate pointer
     shared_ptr<ReactionRate> rate() {
@@ -187,7 +187,7 @@ protected:
     virtual void getParameters(AnyMap& reactionNode) const;
 
     //! Flag indicating whether reaction is set up correctly
-    bool m_valid;
+    bool m_valid = true;
 
     //! Helper function returning vector of undeclared third body species
     //! and a boolean expression indicating whether the third body is specified.
--- a/include/cantera/kinetics/Reaction.h
+++ b/include/cantera/kinetics/Reaction.h
@@ -287,27 +287,6 @@ public:
 };
 
 
-//! A reaction which follows mass-action kinetics with a custom reaction rate
-//! defined in Python.
-/**
- * @warning This class is an experimental part of the %Cantera API and
- *    may be changed or removed without notice.
- */
-class CustomFunc1Reaction : public Reaction
-{
-public:
-    CustomFunc1Reaction();
-    CustomFunc1Reaction(const Composition& reactants, const Composition& products,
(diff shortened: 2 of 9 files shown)


interfaces/cython/cantera/func1.pyx
```diff
@@ -3,6 +3,13 @@
 
 import sys
 
+
+# list holding references to Python objects used for Func1 calls;
+# prevents garbage collection of temporary Python objects, which would otherwise
+# lead to segfaults when C++ Func1 objects are evaluated
+cdef list _python_object_registry = []
```
bryanwweber: Is this going to leak memory when `Func1` objects go out of scope and _should_ be garbage collected? I don't understand quite enough about the problem to suggest an alternate approach, but just a thought I had looking through this

ischoegl: This just ensures that references to `Func1` (Python objects) will be kept around. I couldn’t think of a better way to guarantee that they will be available for callbacks from C++ …

bryanwweber: Yep, I got the purpose. I'm wondering what happens if someone creates lots of objects which aren't going to be used any more. Those should be garbage collected but this will retain a reference to them indefinitely, and not release that memory back. Of course when the Python process ends, the memory will be released so it's not a true memory leak.

However, I'm not sure how to tell when a Func1 object won't be used anymore... Or rather, I'm not sure why Python is deciding to GC the C++ object (that's what's happening, right?). So I'm not sure of a better solution either at the moment.

interfaces/cython/cantera/test/test_reaction.py
```diff
@@ -1490,6 +1491,22 @@ def test_custom_lambda(self):
...
+        for i in range(5):
+            R = ct.Reaction(equation='H + OH <=> H2O', rate=ct.CustomRate(lambda T: T))
+            gas.add_reaction(R)
+            R = ct.Reaction(equation='H + OH <=> H2O', rate=lambda T: T)
+            gas.add_reaction(R)
+            def func(T):
+                return T
+            R = ct.Reaction(equation='H + OH <=> H2O', rate=func)
+            gas.add_reaction(R)
+        assert (gas.forward_rate_constants == gas.T).all()
```
ischoegl, line 1508: @bryanwweber ... this is the crux of the issue ... `R`'s are temporary - see also @speth's example in the #1282

interfaces/cython/cantera/func1.pyx
ischoegl: No - it garbage collects the Python object that C++ still wants to call - any ideas would be appreciated. As an aside, I dropped some of the newer commits as this is probably enough to review by itself.

ischoegl: One more thing: `CustomRate` actually holds a reference to the Python `Func1` in Cython, but apparently Python isn't aware of that and still thinks that it's unused, see https://github.com/Cantera/cantera/blob/0925fc08634515458bd2feb018ca29888f8aecdd/interfaces/cython/cantera/_cantera.pxd#L1360-L1362

ischoegl: Hah! Nevermind - found a way around that. GC does work as intended, but not if the Python `Reaction` objects are not kept around.

interfaces/cython/cantera/kinetics.pyx
```diff
@@ -63,6 +63,9 @@ cdef class Kinetics(_SolutionBase):
     of progress, species production rates, and other quantities pertaining to
     a reaction mechanism.
     """
+    #: List holding references to CustomRate objects (prevents garbage collection)
+    _custom_rates = []
```
bryanwweber, line 74: This makes sense, does Cython have a destructor method for classes? It might make sense to clear this list when the Kinetics instance is GCed if you can do it in the destructor

Or will that happen automatically? I'm not sure of the interaction here with C++/Cython/Python objects

ischoegl: I believe this happens automatically.

ischoegl:
@bryanwweber / @speth ... rebased and ready for a review. The commits are also included in #1333, so it will be easier to review this first.

include/cantera/kinetics/Custom.h
```diff
@@ -68,7 +68,7 @@ class CustomFunc1Rate final : public ReactionRate
     void setRateFunction(shared_ptr<Func1> f);
 
 protected:
-    shared_ptr<Func1> m_ratefunc;
+    shared_ptr<Func1> m_ratefunc = 0;
```
speth: `shared_ptr` is initialized to `nullptr`/0 by default.

include/cantera/kinetics/Reaction.h
```diff
@@ -147,25 +147,25 @@ class Reaction
...
+    bool allow_negative_orders = false;
 
     //! Input data used for specific models
     AnyMap input;
 
     //! The units of the rate constant. These are determined by the units of the
     //! standard concentration of the reactant species' phases of the phase
     //! where the reaction occurs.
-    Units rate_units;
+    Units rate_units {0.0};
```
speth: I think the preferred style for the brace-initialization is to omit the space, analogous to the constructor initialization e.g.
```suggestion
    Units rate_units{0.0};
```

speth (changes requested):
Thanks, @ischoegl, this looks good to me. I had just two very minor suggested modifications.

ischoegl:
@speth ... thanks for the review! I updated both instances per recommendation.

speth (approved):

PR #1333, opened 2022-07-06
Remove reaction specializations

### LINKED ISSUE #1220: Two-temperature-plasma incompatible with third-body notation
**Problem description**

Per #967/#1015/#1027 the following reaction in `test/data/ET_test.yaml`

```yaml
- equation: E + O2 + O2 => O2^- + O2  # Kossyi (1992) Eq.45
  type: two-temperature-plasma
  rate-constant: {A: 1.523e+27, b: -1.0, Ea_T: -100, Ea_Te: 700}
```

is a three-body reaction, but is currently not handled as such.

**Steps to reproduce**

Found in #1219

**Behavior**

Reaction was found using a generic search for implicitly defined third-body colliders

**System information**

- Cantera version: [for example, 2.5.0 or the git commit hash] 2.6.0a4
- OS: [for example, Windows 10]
- Python/MATLAB/other software versions:

ischoegl:

- [x] Deprecate `ThreeBodyReaction` in C++ and Python
- [x] Deprecate `FalloffReaction` in C++ and Python
- [x] Remove `ReactionFactory` ~and deprecate `newReaction` constructors~
- [x] Update Python `Reaction.__repr__` output
- [x] Implement `ThirdBody` object in Python and deprecate `efficiencies` argument in constructor

~Incorporates #1332~

Closes #1220; partially addresses Cantera/enhancements#142

```
In [1]: import cantera as ct

In [2]: ct.Reaction(equation="2 O + O2 <=> 2 O2", rate=ct.ArrheniusRate())
Out[2]: 2 O + O2 <=> O2 + O2    <three-body-Arrhenius>

In [3]: ct.Reaction(equation="2 O (+M) <=> O2 (+M)", rate=ct.TroeRate())
Out[3]: 2 O (+M) <=> O2 (+M)    <falloff-Troe>

In [4]: ct.Reaction("O:2", "O2:1", rate=ct.ArrheniusRate(), third_body=ct.ThirdBody())
Out[4]: 2 O + M <=> O2 + M    <three-body-Arrhenius>

In [5]: ct.Reaction("O:2", "O2:1", rate=ct.ArrheniusRate(), third_body=ct.ThirdBody("O2"))
Out[5]: 2 O + O2 <=> O2 + O2    <three-body-Arrhenius>

In [6]: ct.Reaction(equation="2 O + M <=> O2 + M", rate=ct.BlowersMaselRate())
Out[6]: 2 O + M <=> O2 + M    <three-body-Blowers-Masel>
```

**Comments**

- `three-body` type entry is now optional in YAML input
- Third body is supported for `Blowers-Masel` rates

**Checklist**


ischoegl:
@speth … this is ready for a review

ischoegl:
Rebased and updated to incorporate changes to `yaml2ck`. This should be good for a review now (with the caveat that the first couple of commits are repeated from #1332)

ischoegl:
@speth … thanks for the quick review on #1332! After the merge, I rebased this again so now it’s just what goes beyond `CustomReaction`

include/cantera/kinetics/Reaction.h
```diff
@@ -16,6 +16,7 @@
 #include "Falloff.h"
 #include "InterfaceRate.h"
 #include "PlogRate.h"
+#include "TwoTempPlasmaRate.h"
```
speth: What necessitates this include? If anything, I'd expect this to allow the removal of most of the headers for various `ReactionRate` implementations.

interfaces/cython/cantera/reaction.pyx
```diff
@@ -992,6 +998,87 @@ cdef class StickingBlowersMaselRate(StickRateBase):
             self.cxx_object().setDeltaH(delta_H)
 
 
+cdef class ThirdBody:
+    r"""
+    Class representing third-body collision partners in three-body or falloff reactions.
+
+    :param collider:
+        Name of the third-body collider. If ``M`` (default), the `default_efficiency`
+        is set to 1 and the ollider is assumed to participate in a three-body reaction.
```
speth: "collider"

interfaces/cython/cantera/reaction.pyx
```diff
@@ -992,6 +998,87 @@ cdef class StickingBlowersMaselRate(StickRateBase):
...
+        If the collider includes parentheses, - for example ``(+M)``, - a falloff form
+        is assumed, where the collider is not considered for the law of mass action.
+        For species other than ``M``, the third-body collider represents a named species
+        with collision efficiency 1, and the `default_efficiency` is set to zero.
+    :param efficiencies:
+        Non-default third-body efficiencies
+    :param default_efficinecy:
```
speth: "efficiency"

interfaces/cython/cantera/reaction.pyx
```diff
@@ -1049,73 +1136,108 @@ cdef class Reaction:
...
+            # default to Arrhenius expression
+            raise ValueError("Missing reaction rate information.")
+        elif isinstance(rate, dict):
+            if {"A", "b"} - set(rate) == set():
+                # Allow simple syntax for Arrhenius-type rates
+                args = {"rate-constant": rate}
+                if "Ea0" in rate:
+                    args.update({"type": "Blowers-Masel"})
+                elif "Ea_gas" in rate:
+                    args.update({"type": "two-temperature-plasma"})
```
speth, line 1171: It's kind of strange for the `Reaction` class to know about these particular reaction rate parameterizations. Given that we require the `type` field in the YAML definition for these kinds of reactions, I think it would be preferable to also require it for this constructor style.

interfaces/cython/cantera/reaction.pyx
```diff
@@ -1368,6 +1479,85 @@ cdef class Reaction:
...
+        """
+        Returns an `ThirdBody` object if `Reaction` uses a third body collider, and
+        ``None`` otherwise.
+
+        .. versionadded:: 3.0
+        """
+        def __get__(self):
+            if not self.reaction.usesThirdBody():
+                return
+            return ThirdBody.wrap(self.reaction.thirdBody())
```
speth: Slightly simpler alternative:
```py
if self.reaction.usesThirdBody():
    return ThirdBody.wrap(self.reaction.thirdBody())
```

src/kinetics/Reaction.cpp
```diff
@@ -186,28 +205,55 @@ void Reaction::setParameters(const AnyMap& node, const Kinetics& kin)
...
 
-    if (reactants.count("M") && std::dynamic_pointer_cast<PlogRate>(m_rate)) {
-        throw InputFileError("Reaction::setRate", input, "Found superfluous 'M' in "
-            "pressure-dependent-Arrhenius reaction.");
+    std::string rate_type = input.getString("type", "");
+    if (m_third_body) {
+        if (std::dynamic_pointer_cast<FalloffRate>(m_rate)) {
+            m_third_body->mass_action = false;
+        } else if (std::dynamic_pointer_cast<TwoTempPlasmaRate>(m_rate)) {
+            // two-temperature-plasma rates do not support third-body colliders
```
speth: I don't understand why `TwoTempPlasmaRate` is singled out here. What about this rate type is incompatible with the more general handling of third bodies being introduced here? Wouldn't allowing the third-body object to exist provide a simple resolution to #1220?

src/kinetics/Reaction.cpp
```diff
@@ -186,28 +205,55 @@ void Reaction::setParameters(const AnyMap& node, const Kinetics& kin)
...
+            for (const auto& spc : m_third_body->efficiencies) {
+                reactants[spc.first] += 1.;
+                products[spc.first] += 1.;
+            }
+            m_third_body.reset();
+        } else if (m_third_body->name() == "<multiple>") {
+            m_third_body.reset();
+            throw InputFileError("Reaction::setRate", input,
+                "Reactants for reaction '{}'\n"
+                "contain multiple third body colliders.", equation());
```
speth: This seems like something that's a problem only with reactions that we've gone out of our way to identify for special handling as "three-body reactions", but I think such a reaction should still be allowable as a (somewhat unusual) elementary reaction.

include/cantera/kinetics/Reaction.h
```diff
@@ -33,29 +34,31 @@ class Reaction
 public:
     Reaction() {}
     Reaction(const Composition& reactants, const Composition& products,
-             shared_ptr<ReactionRate> rate={});
+             shared_ptr<ReactionRate> rate, shared_ptr<ThirdBody> tbody=0);
+    Reaction(const std::string& equation,
+             shared_ptr<ReactionRate> rate, shared_ptr<ThirdBody> tbody=0);
```
speth: If you don't want to use `{}` to indicate the default constructor for `shared_ptr<ReactionRate>`, I'd suggest `nullptr` rather than `0`.

interfaces/cython/cantera/test/test_kinetics.py
```diff
@@ -1152,15 +1151,15 @@ def test_negative_A_falloff(self):
             rate.low_rate = low_rate
 
         rate.low_rate = ct.Arrhenius(-2.16e13, -0.23, 0)
-        r.rate = rate
+        rxn = ct.Reaction("NH:1, NO:1", "N2O:1, H:1", rate, third_body=ct.ThirdBody())
```
speth: The need to explicitly supply the `third_body` argument here, when the `rate` object is already derived from `FalloffRate` surprises me a bit.

interfaces/cython/cantera/test/test_reaction.py
```diff
@@ -973,29 +973,19 @@ def from_dict(self):
...
-                        **self._kwargs)
+        # create reaction object from dictionary
+        rxn = ct.Reaction(equation=self._equation, rate=rate, third_body=self._3rd_body)
         return self.finalize(rxn)
 
     def from_parts(self):
         # create reaction rate object from parts
         orig = self.soln.reaction(self._index)
-        rxn = self._cls(orig.reactants, orig.products, rate=self._rate_obj)
+        rxn = ct.Reaction(orig.reactants, orig.products, rate=self._rate_obj, third_body=self._3rd_body)
```
speth: This should probably be line-wrapped. Ditto the line above in `from_empty`

src/kinetics/Reaction.cpp
```diff
@@ -186,28 +205,55 @@ void Reaction::setParameters(const AnyMap& node, const Kinetics& kin)
...
+            warn_deprecated("Chebyshev reaction equation", input, "Specifying '(+M)' "
+                "in the reaction equation for Chebyshev reactions is deprecated.");
+            m_third_body.reset();
+        } else if (std::dynamic_pointer_cast<PlogRate>(m_rate)) {
+            throw InputFileError("Reaction::setRate", input,
+                "Found superfluous '{}' in pressure-dependent-Arrhenius reaction.",
+                m_third_body->name());
+        }
+    } else {
+        if (std::dynamic_pointer_cast<FalloffRate>(m_rate)) {
```
speth: All of this `dynamic_pointer_cast`ing is a bit of an abstraction violation -- I think the the specialization of what to do about the third body based on the type of the `m_rate` object should be handled by a virtual method of the `ReactionRate` class that gets passed the `Reaction` object and can decide what to do.

src/kinetics/Reaction.cpp
```diff
@@ -253,11 +305,122 @@ std::string Reaction::equation() const
...
+        throw CanteraError("Reaction::type", "Empty Reaction does not have a type");
+    }
+
+    std::string rate_type = m_rate->type();
+    auto falloff = std::dynamic_pointer_cast<FalloffRate>(m_rate);
+    if (falloff) {
+        if (falloff->chemicallyActivated()) {
+            return "chemically-activated-" + rate_type;
+        }
+        return "falloff-" + rate_type;
```
speth: This seems like something that ought to be handled by `FalloffRate::type()`.

src/kinetics/Reaction.cpp
```diff
@@ -488,365 +623,198 @@ bool Reaction::usesElectrochemistry(const Kinetics& kin) const
...
 {
-    m_third_body.reset(new ThirdBody);
-    if (!node.empty()) {
-        setParameters(node, kin);
-        setRate(newReactionRate(node, calculateRateCoeffUnits(kin)));
-    } else {
-        setRate(newReactionRate(type()));
+    if (m_name != "M") {
+        return;
+    }
```
speth, line 688: Why does this silently return without changing anything?

include/cantera/kinetics/Reaction.h
```diff
@@ -210,34 +214,70 @@ class Reaction
...
+    void getParameters(AnyMap& node) const;
+
     //! Get the third-body efficiency for species *k*
     double efficiency(const std::string& k) const;
 
+    //! Name or placeholder of third body collider, for example `+ M`
+    //! @since  New in Cantera 3.0
+    std::string collider() const;
+
+    //! Verify that all species involved in collision efficiences are defined in the
```
speth: "efficiencies"

include/cantera/kinetics/Reaction.h
```diff
@@ -210,34 +214,70 @@ class Reaction
...
+    //! Get third-body efficiencies from AnyMap *node*
+    //! @since  New in Cantera 3.0
+    void getParameters(AnyMap& node) const;
+
     //! Get the third-body efficiency for species *k*
     double efficiency(const std::string& k) const;
 
+    //! Name or placeholder of third body collider, for example `+ M`
+    //! @since  New in Cantera 3.0
+    std::string collider() const;
```
speth, line 251: This is more than just the _name_ of the collider, given the inclusion of the ` + ` or ` (+` -- it's the suffix for the reactants / products expressions in the reaction equation.

test/kinetics/kineticsFromScratch.cpp
```diff
@@ -72,23 +85,77 @@ TEST_F(KineticsFromScratch, add_three_body_reaction)
...
+    auto rate = make_shared<ArrheniusRate>(1.2e11, -1.0, 0.0);
+    auto R = make_shared<Reaction>(equation, rate);
+    EXPECT_TRUE(R->usesThirdBody());
+}
+
+TEST_F(KineticsFromScratch, multiple_third_bodies1)
+{
+    std::string equation = "2 H + 2 O2 <=> H2 + 2 O2";
+    auto rate = make_shared<ArrheniusRate>(1.2e11, -1.0, 0.0);
+    ASSERT_THROW(Reaction(equation, rate), CanteraError);
```
speth: I don't understand why this should inherently be an error.

test/kinetics/kineticsFromScratch.cpp
```diff
@@ -72,23 +85,77 @@ TEST_F(KineticsFromScratch, add_three_body_reaction)
...
+}
+
+TEST_F(KineticsFromScratch, add_two_temperature_plasma)
+{
+    std::string equation = "O + H => O + H";
+    auto rate = make_shared<TwoTempPlasmaRate>(17283, -3.1, -5820000, 1081000);
+    auto R = make_shared<Reaction>(equation, rate);
+    EXPECT_FALSE(R->usesThirdBody());
```
speth, line 157: This should never be considered a three-body reaction by the rule that either the reactants or products side must involve 3 species, independent of the use of the `TwoTempPlasmaRate` parameterization.

speth (changes requested):
Thanks, @ischoegl, I think this is an exciting step forward in simplifying the handling of different reaction rate parameterizations.

One thing that you only barely touched on is that by handling third bodies in the common `Reaction` class, these can in principle be used for any rate parameterization. However, there are a couple of quirks here that I think are worth ironing out. One is that there are a few places that single out the `TwoTempPlasmaRate` class as not being able to handle third bodies, but I don't understand why that would be the case now. Shouldn't any reaction type be able to handle third bodies now, with the exception that we explicitly remove them from Plog and Chebyshev reactions in acknowledgment that the third-body pressure dependence is already embedded in those rate constants?

The second is that there are a few situations where the automatic detection of specific third body colliders prevents instantiation of reactions as non-third-body reactions, even though I don't think these should be seen as problematic. For instance, the following reaction definitions raise exceptions, but were (I think correctly) permitted in Cantera 2.6:

```py
R1 = ct.Reaction(
    equation="HO2 + O2 <=> OH + O + O2",
    rate={"type": "pressure-dependent-Arrhenius",
          "rate-constants": [
               {"P": 1013.25, "A": 1.2124e+16, "b": -0.5779, "Ea": 45491376.8},
               {"P": 101325., "A": 4.9108e+31, "b": -4.8507, "Ea": 103649395.2}
          ],
    }
)

R2 = ct.Reaction(
    equation="H + OH + 2 O2 = H2O + 2 O2",
    rate={"A": 1.234e5, "b": 0, "Ea": 4321.0}
)
```

test/kinetics/kineticsFromScratch.cpp
ischoegl: I don’t think we can handle multiple third-bodies at the moment.

test/kinetics/kineticsFromScratch.cpp
ischoegl: :tada: Ah. So far, I was not checking for the condition when multiple colliders are present. This will simplify things!

test/kinetics/kineticsFromScratch.cpp
speth: My suggestion is that this should fall back to being treated as an "elementary" reaction with no third-body defined.

include/cantera/kinetics/Reaction.h
ischoegl: I believe that at the moment, the headers for individual reaction rate specializations are included via `Reaction.h`. From that perspective, `TwoTempPlasmaRate.h` was an outlier. Other than that, reaction rate headers are only included in `ReactionRateFactory.cpp`.

src/kinetics/Reaction.cpp
ischoegl: Many of those can involve multiple third bodies, where I had blocked things before. Being more permissive resolves this issue.

include/cantera/kinetics/Reaction.h
speth: With the exception of the top-level headers (`cantera/thermo.h` etc.), the rest of the Cantera header files are not meant to be shortcuts to including many other headers. The reason so many `ReactionRate` headers were included here previously was only because that was necessary, since the derived `Reaction` types included members of those types.

If the abstraction that we've created in the `ReactionRate` class is robust, it should never be necessary to include the specific implementation headers anywhere except for `ReactionRateFactory.cpp` and classes that derive directly from them. You can see the parallel with the `ThermoPhase` types, which are by and large included only `ThermoFactory.cpp`.

ischoegl: This makes sense. In this case, I believe all specialization headers should be removed.

src/kinetics/Reaction.cpp
ischoegl: This should raise an error ...

src/kinetics/Reaction.cpp
ischoegl: This is a bit of a paradigm shift, but does make sense - I introduced `ReactionRate::subType` to differentiate between `SRI`/`Troe`/etc., which has the advantage of making `ReactionRate::type` consistent with the YAML nomenclature.

src/kinetics/Reaction.cpp
ischoegl: ... with the updated `ReactionRate::type` definitions for `FalloffRate`, the cast for `FalloffRate` is no longer necessary, as `FalloffRate::type` is either `falloff` or `chemically-activated`, which are easy to query.

interfaces/cython/cantera/test/test_kinetics.py
ischoegl: There is a bit of logic involved as the correct `(+M)` nomenclature should be checked if the reaction equation is provided. With that in place, it's simple to omit.

interfaces/cython/cantera/reaction.pyx
ischoegl: This is really just a shortcut. We do allow it for `ArrheniusRate`; the two others are very similar so I thought to include them.

ischoegl:
@speth ... thanks for the feedback. I ended up being a lot more permissive/selective with the detection of third bodies, which I believe does make sense.

Regarding the two reactions you mentioned, they now produce the following:
```
In [2]: ct.Reaction(
   ...:     equation="HO2 + O2 <=> OH + O + O2",
   ...:     rate={
   ...:         "type": "pressure-dependent-Arrhenius",
   ...:         "rate-constants": [
   ...:             {"P": 1013.25, "A": 1.2124e16, "b": -0.5779, "Ea": 45491376.8},
   ...:             {"P": 101325.0, "A": 4.9108e31, "b": -4.8507, "Ea": 103649395.2},
   ...:         ],
   ...:     },
   ...: )
Out[2]: HO2 + O2 <=> O + OH + O2    <three-body-pressure-dependent-Arrhenius>

In [3]: ct.Reaction(
   ...:     equation="H + OH + 2 O2 = H2O + 2 O2", rate={"A": 1.234e5, "b": 0, "Ea": 4321.0}
   ...: )
Out[3]: H + 2 O2 + OH <=> H2O + 2 O2    <Arrhenius>
```

src/kinetics/Reaction.cpp
ischoegl: Removed

include/cantera/kinetics/Falloff.h
```diff
@@ -302,7 +305,7 @@ class LindemannRate final : public FalloffRate
             new MultiRate<LindemannRate, FalloffData>);
     }
 
-    virtual const std::string type() const {
+    virtual const std::string subType() const override {
```
speth, line 309: All the other overridden methods now need the `override` qualifier to avoid copious compiler warnings (at least from Clang).

interfaces/cython/cantera/reaction.pyx
```diff
@@ -1368,6 +1486,84 @@ cdef class Reaction:
             cdef CxxUnits rate_units = self.reaction.rate_units
             return Units.copy(rate_units)
 
+    property third_body:
+        """
+        Returns an `ThirdBody` object if `Reaction` uses a third body collider, and
```
speth:
```suggestion
        Returns a `ThirdBody` object if `Reaction` uses a third body collider, and
```

src/kinetics/Reaction.cpp
```diff
@@ -186,28 +208,55 @@ void Reaction::setParameters(const AnyMap& node, const Kinetics& kin)
...
+                throw InputFileError("Reaction::setRate", input,
+                    "Found superfluous '{}' in pressure-dependent-Arrhenius reaction.",
+                    m_third_body->name());
+            }
+        }
+    } else {
+        if (rate_type == "falloff" || rate_type == "chemically-activated") {
+            if (!m_from_composition) {
+                throw InputFileError("Reaction::setRate", input,
+                    "Reaction eqution for falloff reaction '{}'\n does not "
```
speth:
```suggestion
                    "Reaction equation for falloff reaction '{}'\n does not "
```

src/kinetics/Reaction.cpp
```diff
@@ -253,11 +308,129 @@ std::string Reaction::equation() const
...
+    if (rate_type == "three-body") {
+        // state type when serializing
+        m_explicit_rate = true;
+    } else if (rate_type == "elementary") {
+        // user override
+        m_explicit_rate = true;
+        return;
+    } else if (rate_type == "two-temperature-plasma") {
+        // two-temperature-plasma reactions do not use third bodies
+        return;
```
speth: I'm confused by this comment -- this now seems to be permissible:
```py
>>> R = ct.Reaction(equation='OH + e + O2 => OH- + O2',
                    rate={'A': 1.0, 'b':0.0, 'Ea_gas': 3000, 'Ea_electron': 5000})
>>> print(repr(R))
OH + e + O2 => OH- + O2    <three-body-two-temperature-plasma>
```

src/kinetics/Reaction.cpp
```diff
@@ -253,11 +308,129 @@ std::string Reaction::equation() const
...
+    }
+
+    std::string third_body;
+    size_t count = 0;
+    size_t countM = 0;
+    for (const auto& reac : reactants) {
+        // detect explicitly specified collision partner
+        if (products.count(reac.first)) {
+            third_body = reac.first;
+            size_t generic = third_body == "M" || ba::ends_with(third_body, "M)");
```
speth: Isn't just looking for names ending with `M` likely to generate some false positives?

speth (changes requested):
Thanks @ischoegl, I had a few more comments / questions.

Given the degree to which this now separates the handling of third bodies from reaction rate parameterizations, I think there is a need for some updates to the corresponding "science" documentation that's shown at https://cantera.org/science/kinetics.html -- I think this will need to be done before we can fully close https://github.com/Cantera/enhancements/issues/142.

src/kinetics/Reaction.cpp
ischoegl: whoops. forgot to delete this one.

ischoegl:
@speth ... I agree that there should be some updates to the Science section, but I think the rest is done! Thank you for the suggestions ... this should complete the work left over on refactoring of `Reaction` and `ReactionRate` outside of the website.

ischoegl:
@speth ... I opened Cantera/cantera-website#216 to track the remaining documentation work.

speth (approved):
Thanks, @ischoegl - I think this is good to go.

PR #1337, opened 2022-07-15
Bump version to 3.0.0a2

ischoegl:
**Changes proposed in this pull request**

After the merge of #1333 (i.e. CTI/XML are completely removed and remaining `Reaction` specializations are deprecated, plus `yaml2ck` is merged), it may be time for updating the alpha version.

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

PR #1338, opened 2022-07-18
Update three body logic

ischoegl:

- Fix cases where three-body detection logic introduced in #1333 fails

```
In [2]: ct.Reaction(
   ...:     equation="CH2OCH + M <=> CH2CHO + M", rate={"A": 5.0e09, "b": 0.0, "Ea": 0.0}
   ...: )
Out[2]: CH2OCH + M <=> CH2CHO + M    <three-body-Arrhenius>
```

plus several edge cases where three-body reactions are explicitly or implicitly defined, e.g.

```
In [3]: gas = ct.Solution("h2o2.yaml")
   ...: _yaml = """
   ...:         equation: H2 + O2 <=> H2 + O2
   ...:         type: three-body-Blowers-Masel # or type: Blowers-Masel
   ...:         rate-constant: {A: 38700, b: 2.7, Ea0: 2.619184e4 cal/mol, w: 4.184e9 cal/mol}
   ...:         efficiencies: {O2: 1}
   ...:         default-efficiency: 0.
   ...:         """
   ...: ct.Reaction.from_yaml(_yaml, gas)
Out[3]: H2 + O2 <=> H2 + O2    <three-body-Blowers-Masel>
```
or
```
In [4]: rxn = ct.Reaction(
   ...:     equation="H2 + O2 <=> H2 + O2",
   ...:     rate={"A": 5.0e09, "b": 0.0, "Ea": 0.0},
   ...:     third_body=ct.ThirdBody("O2"),
   ...: )
   ...: rxn
Out[4]: H2 + O2 <=> H2 + O2    <three-body-Arrhenius>

In [5]: rxn.input_data
Out[5]:
{'equation': 'H2 + O2 <=> H2 + O2',
 'rate-constant': {'A': 5000000000.0, 'b': 0.0, 'Ea': 0.0},
 'efficiencies': {'O2': 1.0},
 'default-efficiency': 0.0}
```

**Checklist**

### DIFF
--- a/include/cantera/kinetics/Reaction.h
+++ b/include/cantera/kinetics/Reaction.h
@@ -195,7 +195,10 @@ protected:
     bool m_valid = true;
 
     //! Flag indicating that serialization uses explicit type
-    bool m_explicit_rate = false;
+    bool m_explicit_type = false;
+
+    //! Flag indicating that third body collider is ambiguous
+    bool m_explicit_3rd = false;
 
     //! Flag indicating that object was instantiated from reactant/product compositions
     bool m_from_composition = false;
@@ -239,8 +242,10 @@ public:
     void setParameters(const AnyMap& node);
 
     //! Get third-body efficiencies from AnyMap *node*
+    //! @param node  AnyMap receiving serialized parameters
+    //! @param explicit_3rd  Flag triggering explicit third body efficiency output
     //! @since  New in Cantera 3.0
-    void getParameters(AnyMap& node) const;
+    void getParameters(AnyMap& node, bool explicit_3rd=false) const;
 
     //! Get the third-body efficiency for species *k*
     double efficiency(const std::string& k) const;
--- a/include/cantera/kinetics/Reaction.h
+++ b/include/cantera/kinetics/Reaction.h
@@ -197,7 +197,7 @@ protected:
     //! Flag indicating that serialization uses explicit type
     bool m_explicit_type = false;
 
-    //! Flag indicating that third body collider is ambiguous
+    //! Flag indicating that third body requires explicit serialization
     bool m_explicit_3rd = false;
 
     //! Flag indicating that object was instantiated from reactant/product compositions
@@ -245,7 +245,7 @@ public:
     //! @param node  AnyMap receiving serialized parameters
     //! @param explicit_3rd  Flag triggering explicit third body efficiency output
     //! @since  New in Cantera 3.0
-    void getParameters(AnyMap& node, bool explicit_3rd=false) const;
+    void getParameters(AnyMap& node, bool explicit_3rd) const;
 
     //! Get the third-body efficiency for species *k*
     double efficiency(const std::string& k) const;
--- a/include/cantera/kinetics/Reaction.h
+++ b/include/cantera/kinetics/Reaction.h
@@ -197,9 +197,6 @@ protected:
     //! Flag indicating that serialization uses explicit type
     bool m_explicit_type = false;
 
-    //! Flag indicating that third body requires explicit serialization
-    bool m_explicit_3rd = false;
-
     //! Flag indicating that object was instantiated from reactant/product compositions
     bool m_from_composition = false;
 
@@ -243,9 +240,8 @@ public:
 
     //! Get third-body efficiencies from AnyMap *node*
     //! @param node  AnyMap receiving serialized parameters
-    //! @param explicit_3rd  Flag triggering explicit third body efficiency output
     //! @since  New in Cantera 3.0
-    void getParameters(AnyMap& node, bool explicit_3rd) const;
+    void getParameters(AnyMap& node) const;
 
     //! Get the third-body efficiency for species *k*
     double efficiency(const std::string& k) const;
@@ -273,6 +269,9 @@ public:
     //! Third body is used by law of mass action
     //! (`true` for three-body reactions, `false` for falloff reactions)
     bool mass_action = true;
+
+    //! Flag indicating whether third body requires explicit serialization
+    bool explicit_3rd = false;
 
 protected:
     //! Name of the third body collider
--- a/src/kinetics/Reaction.cpp
+++ b/src/kinetics/Reaction.cpp
@@ -62,6 +62,9 @@ Reaction::Reaction(const std::string& equation,
 {
     setEquation(equation);
     setRate(rate_);
+    if (tbody_ && tbody_->name() != "M") {
+        m_explicit_3rd = true;
+    }
 }
 
 Reaction::Reaction(const AnyMap& node, const Kinetics& kin)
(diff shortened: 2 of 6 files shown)


ischoegl:
I believe this is ready for a review.

src/kinetics/Reaction.cpp
```diff
@@ -37,6 +37,21 @@ Reaction::Reaction(const Composition& reactants_,
...
 {
     setRate(rate_);
+
+    // ensure safe serialization
+    size_t count = 0;
+    for (const auto& reac : reactants) {
+        if (products.count(reac.first)) {
+            count++;
+        }
+    }
```
bryanwweber: Seems like this could be a little simpler as:

```suggestion
    bool count = false;
    for (const auto& reac : reactants) {
        if (products.count(reac.first)) {
            count = true;
            break;
        }
    }
```

bryanwweber (commented):
I have one very minor suggestion, which may not be an improvement. I also wonder if you can clarify, in the docstring and the error message, what "ambiguous" means in this context. What should the user do to correct this problem?

ischoegl:
@bryanwweber ... thanks! I updated the comments and error messages. Let me know if this clarifies things sufficiently.

ischoegl:
PS: the only question I have is about another edge case. At the moment, the logic requires either three reactants or three products to detect a three-body reaction and thus misses the following
```
In [6]: ct.Reaction(
   ...:     equation="CH2OCH + O2 <=> CH2CHO + O2", rate={"A": 5.0e09, "b": 0.0, "Ea": 0.0}
   ...: )
Out[6]: CH2OCH + O2 <=> CH2CHO + O2    <Arrhenius>
```
... it is certainly perceivable that this should be a `three-body-Arrhenius` instead? Fwiw, the case should be equivalent to
```
In [7]: rxn = ct.Reaction(
   ...:     equation="CH2OCH + M <=> CH2CHO + M",
   ...:     rate={"A": 5.0e09, "b": 0.0, "Ea": 0.0},
   ...:     third_body=ct.ThirdBody("O2"),
   ...: )
   ...: rxn
Out[7]: CH2OCH + M <=> CH2CHO + M    <three-body-Arrhenius>

In [8]: rxn.input_data
Out[8]:
{'equation': 'CH2OCH + M <=> CH2CHO + M',
 'rate-constant': {'A': 5000000000.0, 'b': 0.0, 'Ea': 0.0},
 'efficiencies': {'O2': 1.0},
 'default-efficiency': 0.0}
```

PPS: regardless, serialization works and I'd rather not introduce another set of cases where duplicates are created where there weren't any before.

ischoegl:
I think this is finally ready. At this point I believe that unit tests cover everything I can think of - once this is merged, we should be good to bump the alpha version with #1337

include/cantera/kinetics/Reaction.h
```diff
@@ -195,7 +195,10 @@ class Reaction
     bool m_valid = true;
 
     //! Flag indicating that serialization uses explicit type
-    bool m_explicit_rate = false;
+    bool m_explicit_type = false;
+
+    //! Flag indicating that third body collider is ambiguous
+    bool m_explicit_3rd = false;
```
speth: The comment here is incorrect, isn't it?

src/kinetics/Reaction.cpp
```diff
@@ -37,6 +37,22 @@ Reaction::Reaction(const Composition& reactants_,
     , m_third_body(tbody_)
 {
     setRate(rate_);
+
+    // ensure safe serialization
```
speth: This comment is kind of confusing -- what does "safe" serialization mean?

include/cantera/kinetics/Reaction.h
```diff
@@ -239,8 +242,10 @@ class ThirdBody
     void setParameters(const AnyMap& node);
 
     //! Get third-body efficiencies from AnyMap *node*
+    //! @param node  AnyMap receiving serialized parameters
+    //! @param explicit_3rd  Flag triggering explicit third body efficiency output
     //! @since  New in Cantera 3.0
-    void getParameters(AnyMap& node) const;
+    void getParameters(AnyMap& node, bool explicit_3rd=false) const;
```
speth: I think the second argument can be made mandatory -- this method is only called from one place, in `Reaction::getParameters`. Alternatively, would it make more sense to make `Reaction::m_explicit_3rd` a member of the `ThirdBody` class instead, eliminating the need to pass this in here?

test/kinetics/kineticsFromYaml.cpp
```diff
@@ -91,6 +91,119 @@ TEST(Reaction, ThreeBodyFromYaml2)
...
+
+    auto R = newReaction(rxn, *(sol->kinetics()));
+    EXPECT_EQ(R->type(), "three-body-Arrhenius");
+    EXPECT_TRUE(R->usesThirdBody());
+    EXPECT_EQ(R->thirdBody()->name(), "O2");
+
+    const auto& rate = std::dynamic_pointer_cast<ArrheniusRate>(R->rate());
+    EXPECT_DOUBLE_EQ(rate->preExponentialFactor(), 5.0e+9);
+
+    AnyMap input = R->parameters();
```
speth: I think these tests should be run with `false` passed to the `parameters()` method to suppress the keys that came directly from the input, and test that the method is actually recreating the structure.

speth (changes requested):
Thanks for working on these refinements, @ischoegl.

Beside the suggestions below, I was thinking that we should probably limit the use of the `efficiencies` and `default-efficiency` fields to the case where third body is `M`, to avoid allowing the rather confusing example below:

```yaml
equation: H + O2 + N2 = HO2 + N2
rate-constant: {A: 1, b: 1, Ea: 200}
efficiencies: {N2: 10.0}
default-efficiency: 1
```

which is currently allowed and seems to be equivalent to writing the same reaction with `M` as the third body.

include/cantera/kinetics/Reaction.h
ischoegl: It's not necessarily incorrect, but also not clear. The flag indicates that a third body needs to be explicitly included for serialization. ... comment is clarified

src/kinetics/Reaction.cpp
ischoegl: clarified - it's really "correct" output.

ischoegl:
@speth ... thanks for the review! I adopted recommendations and tweaked the logic so that your example throws the following error:
```
InputFileError thrown by ThirdBody::setParameters:
Error on line 5 of input string:
Invalid default efficiency for explicit collider N2;
value is optional and/or needs to be zero
|  Line |
|     1 |
|     2 | equation: H + O2 + N2 = HO2 + N2
|     3 | rate-constant: {A: 1, b: 1, Ea: 200}
|     4 | efficiencies: {N2: 10.0}
>     5 > default-efficiency: 1
                              ^
```
(it does make sense that the default efficiency has to be zero in this case).

On the other hand, I'd like to keep the ability to change the efficiency for `N2`. Among others, specifying the efficiency does allow for the disambiguation of the following:
```yaml
equation: CH2 + O2 <=> CH2 + O2
rate-constant: {A: 5.0e+9, b: 0.0, Ea: 0.0}
efficiencies: {O2: 1.}
```
I.e. this is handled as a three-body reaction, where the user specifies which of the two participating species is treated as a third body. (*As an aside: I realize that the reaction makes little sense; I had originally blocked those but the feedback was that these reactions should be allowed.*)

speth (approved):

PR #1345, opened 2022-07-29
Integrate Solution in oneD objects

ischoegl:

- Create `StFlow` objects from `Solution`
- Make nomenclature for `Transport::transportModel` consistent

The PR excludes `clib/ctonedim`, where `Solution` is currently completely absent.

Preliminary work for Cantera/enhancements#137

**Checklist**

### DIFF
--- a/include/cantera/oneD/Domain1D.h
+++ b/include/cantera/oneD/Domain1D.h
@@ -28,6 +28,7 @@ class MultiJac;
 class OneDim;
 class Refiner;
 class AnyMap;
+class Solution;
 
 /**
  * Base class for one-dimensional domains.
@@ -323,6 +324,12 @@ public:
      */
     virtual void restore(const AnyMap& state, double* soln, int loglevel);
 
+    //! Return reference to thermo/kinetics/transport manager
+    //! @since  New in Cantera 3.0.
+    shared_ptr<Solution> sol() const {
+        return m_sol;
+    }
+
     size_t size() const {
         return m_nv*m_points;
     }
@@ -498,6 +505,9 @@ protected:
     std::vector<std::string> m_name;
     int m_bw;
     bool m_force_full_update;
+
+    //! Composite thermo/kinetics/transport handler
+    std::shared_ptr<Solution> m_sol;
 };
 }
 
--- a/include/cantera/oneD/Domain1D.h
+++ b/include/cantera/oneD/Domain1D.h
@@ -324,10 +324,10 @@ public:
      */
     virtual void restore(const AnyMap& state, double* soln, int loglevel);
 
-    //! Return reference to thermo/kinetics/transport manager
+    //! Return thermo/kinetics/transport manager used in the domain
     //! @since  New in Cantera 3.0.
-    shared_ptr<Solution> sol() const {
-        return m_sol;
+    shared_ptr<Solution> solution() const {
+        return m_solution;
     }
 
     size_t size() const {
@@ -507,7 +507,7 @@ protected:
     bool m_force_full_update;
 
     //! Composite thermo/kinetics/transport handler
-    std::shared_ptr<Solution> m_sol;
+    std::shared_ptr<Solution> m_solution;
 };
 }
 
--- a/include/cantera/oneD/IonFlow.h
+++ b/include/cantera/oneD/IonFlow.h
@@ -31,7 +31,10 @@ namespace Cantera
 class IonFlow : public StFlow
 {
 public:
-    IonFlow(IdealGasPhase* ph = 0, size_t nsp = 1, size_t points = 1);
+    IonFlow(ThermoPhase* ph = 0, size_t nsp = 1, size_t points = 1);
+
+    IonFlow(shared_ptr<Solution> sol, size_t nsp = 1, size_t points = 1);
+
     //! set the solving stage
     virtual void setSolvingStage(const size_t phase);
 
--- a/include/cantera/oneD/StFlow.h
+++ b/include/cantera/oneD/StFlow.h
@@ -59,10 +59,10 @@ public:
 
     virtual void resetBadValues(double* xg);
 
-
     ThermoPhase& phase() {
         return *m_thermo;
     }
+
     Kinetics& kinetics() {
         return *m_kin;
     }
@@ -70,10 +70,10 @@ public:
     /**
      * Set the thermo manager. Note that the flow equations assume
      * the ideal gas equation.
+     *
+     * @deprecated  To be removed after Cantera 3.0 (unused)
      */
-    void setThermo(IdealGasPhase& th) {
-        m_thermo = &th;
-    }
+    void setThermo(IdealGasPhase& th);
 
     //! Set the kinetics manager. The kinetics manager must
     void setKinetics(Kinetics& kin) {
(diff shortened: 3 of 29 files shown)


include/cantera/oneD/Domain1D.h
```diff
@@ -323,6 +324,12 @@ class Domain1D
      */
     virtual void restore(const AnyMap& state, double* soln, int loglevel);
 
+    //! Return reference to thermo/kinetics/transport manager
+    //! @since  New in Cantera 3.0.
+    shared_ptr<Solution> sol() const {
+        return m_sol;
+    }
```
bryanwweber: Since this is new, can we use the full `m_solution` and `solution()`? I think it'll probably be inconsistent with other interfaces but we're not working with 80-character punch cards anymore...

include/cantera/oneD/StFlow.h
```diff
@@ -52,37 +53,48 @@ class StFlow : public Domain1D
...
-    }
+    void setThermo(IdealGasPhase& th);
 
     //! Set the kinetics manager. The kinetics manager must
     void setKinetics(Kinetics& kin) {
         m_kin = &kin;
     }
 
-    //! set the transport manager
+    //! Set the transport manager directly
```
bryanwweber:
```suggestion
    //! Set a reference to an existing transport manager instance
```

include/cantera/transport/Transport.h
```diff
@@ -157,10 +158,20 @@ class Transport
 
     //! Identifies the Transport object type. Each derived class should override
```
bryanwweber:
```suggestion
    //! Identifies the model represented by this Transport object. Each derived class should override
```

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -468,10 +460,29 @@ cdef class _FlowBase(Domain1D):
...
     def set_transport(self, _SolutionBase phase):
         """
         Set the `Solution` object used for calculating transport properties.
+
+        .. deprecated:: 3.0
+
+            Method to be removed after Cantera 3.0. Replaceable by `transport_model`
         """
+        warnings.warn("Method to be removed after Cantera 3.0; use property "
+                    "'transport_model' instead.", DeprecationWarning)
```
bryanwweber:
```suggestion
                      "'transport_model' instead.", DeprecationWarning)
```

bryanwweber (changes requested):
Thanks @ischoegl I have a couple minor suggestions here. I'm not sure of the implementation details, so I think @speth will need to review that part.

include/cantera/oneD/Domain1D.h
ischoegl: no objections to `m_solution`, but there's an unfortunate clash for `solution()`: https://github.com/Cantera/cantera/blob/177356c6687f135f2b95cf566d803f948a032a10/include/cantera/oneD/Sim1D.h#L109-L111

... it's a slightly different context, but I'd like to avoid a clash of nomenclatures. Suggestions?

bryanwweber: That's disappointing. Even with the 3.0 break, I still think changing the functionality of the existing `solution()` so drastically as to return a `Solution` would need to go through a deprecation cycle, plus whatever we use for the `m_solution`-getter would immediately be deprecated... I don't have any other good suggestions 😦 

As another nitpick on these lines, should the comment say `Return a _pointer_ to thermo/kinetics/transport manager`?

speth: Maybe `getSolution()`? It's still a little confusing, but I'm not sure that can be entirely avoided.

Regarding the comment, I'd just say "Get the thermo/kinetics/transport managers used in the domain" -- I don't think it's necessary or helpful to repeat the information about the type that's more precisely given in the function signature.

ischoegl: As `Sim1D::solution` is unused at this point, I decided to deprecate it (together with `Sim1D::setSolution` and use `Domain1D::solution` after all as it is consistent with the nomenclature we use for getters in general.

ischoegl:
Rebased to latest `main`.

src/oneD/IonFlow.cpp
```diff
@@ -57,6 +58,23 @@ IonFlow::IonFlow(IdealGasPhase* ph, size_t nsp, size_t points) :
     m_do_electric_field.resize(m_points,false);
 }
 
+IonFlow::IonFlow(shared_ptr<Solution> sol, size_t nsp, size_t points) :
+    IonFlow(sol->thermo().get(), nsp, points)
+{
+    m_solution = sol;
+    m_kin = m_solution->kinetics().get();
+    m_trans_shared = m_solution->transport();
+    m_trans = m_trans_shared.get();
```
speth, line 67: I don't understand the need for the `m_trans_shared` object. It's always just a copy of the `m_solution->transport()` shared pointer, isn't it?

include/cantera/oneD/IonFlow.h
```diff
@@ -31,7 +31,11 @@ namespace Cantera
 class IonFlow : public StFlow
 {
 public:
-    IonFlow(IdealGasPhase* ph = 0, size_t nsp = 1, size_t points = 1);
+    IonFlow(ThermoPhase* ph = 0, size_t nsp = 1, size_t points = 1);
+
+    //! Delegating constructor
+    IonFlow(shared_ptr<Solution> sol, size_t nsp = 1, size_t points = 1);
```
speth: The "delegating" aspect is an implementation detail that I don't think needs to be documented here. But if you wanted to add an actual docstring for this constructor, I wouldn't object.

speth (changes requested):
This looks pretty good to me, @ischoegl. My main question is about the necessity of the `StFlow::m_trans_shared` object.

src/oneD/IonFlow.cpp
ischoegl: If I remember correctly, it is needed to prevent hard crashes. There are instances where the transport object is instantiated elsewhere and would get destroyed if the shared pointer isn’t copied here.

speth: Oh, I see it now -- if you call `setTransportModel` on the `Solution` object, then that would cause the current object to be deleted unless the `StFlow` object is also holding a copy of the `shared_ptr`. Arguably, the same thing is true for the `Kinetics` object and the use of the `Solution::setKinetics` method, though I don't think there's any case where we currently use the ability to replace the `Kinetics` object that way.

ischoegl: Your interpretation is spot-on. There’s no good way around this without replacing raw pointers, which is well beyond of what I was trying to do in this PR.

include/cantera/oneD/IonFlow.h
ischoegl: :+1: removed.

ischoegl:
@speth ... thank you for the review!

src/oneD/IonFlow.cpp
ischoegl: I updated the docstring of `StFlow::m_trans_shared` to clarify.

speth (approved):

bryanwweber (approved):

PR #1334, opened 2022-07-06
Allow Cython module components to be compiled independently

speth:
**Changes proposed in this pull request**

- Separate includes (both C++ and Cython) needed by each `.pyx` file so they can be compiled independently
- Combine the separately compiled submodules into a single shared library
- Use the solution from [this StackOverflow post](https://stackoverflow.com/a/52714500) to import the resulting submodules from the same shared library

**If applicable, provide an example illustrating new features this pull request is introducing**

This change substantially speeds up compilation of the Cython module when compiling with multiple cores, or when making changes that only affect a subset of the `.pyx` files.

*Timings for the current `main` branch:*
```shell
# Compile everything
$ time scons build -j4 &>/dev/null

real    2m35.281s
user    5m58.835s
sys     0m18.177s

# Re-cythonize and compile just the Python module
$ rm -rf build/temp-py build/python
$ time scons build -j4 &>/dev/null

real    1m14.313s
user    1m10.681s
sys     0m5.905s
```

*Timings for this PR:*
```sh
# Compile everything
$ rm -rf build/temp-py build/python
$ time scons build -j4

real    2m5.328s
user    6m42.189s
sys     0m20.347s

# Re-cythonize and compile just the Python module
$ time scons build -j4

real    0m42.615s
user    1m54.346s
sys     0m7.753s

# With a lot of cores
$ time scons build -j16

real    0m26.209s
user    1m55.752s
sys     0m8.489s

# With just 2 cores
$ time scons build -j2
real    1m6.462s
user    1m54.324s
sys     0m7.826s
```

Note that the total CPU time increases by a fair amount, due to the duplicated work done in compiling some common code for each `.pyx` file. However, even with 2 cores, the wall-clock time decreases.

In addition, this provides some of the structure envisioned by cantera/enhancements#153, in that the classes and methods are accessible using names qualified with the corresponding submodule, such as:
```python
from cantera.reactor import IdealGasReactor
from cantera.onedim import FreeFlame
```
although they are also available directly in the `cantera` module as well. While I don't want to pursue this as part of the current PR, I think this restructuring also provides most of the necessary machinery to support building only a subset of the C++ source and the corresponding portions of the Python module, such as compiling everything except the reactor/flame models.

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
[...]

PR #1346, opened 2022-07-30
Fix compilation warnings

ischoegl:

Fix compilation warnings caused by comparisons of `int` and `size_t`.

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1604,6 +1604,10 @@ if env['python_package'] != 'none':
                 f"Cython is an incompatible version: Found {cython_version} but "
                 f"{cython_min_version} or newer is required.")
             warn_no_full_package = True
+        elif cython_version < parse_version("3.0.0"):
+            logger.info(
+                f"Using Cython version {cython_version} (uses legacy NumPy API)")
+            env["numpy_1_7_API"] = True
         else:
             logger.info(f"Using Cython version {cython_version}")
 
--- a/include/cantera/numerics/PreconditionerBase.h
+++ b/include/cantera/numerics/PreconditionerBase.h
@@ -32,6 +32,8 @@ class PreconditionerBase
 public:
     PreconditionerBase() {}
 
+    virtual ~PreconditionerBase() {}
+
     //! Set a value at the specified row and column of the jacobian triplet vector
     //! @param row row in the jacobian matrix
     //! @param col column in the jacobian matrix
--- a/include/cantera/zeroD/IdealGasConstPressureMoleReactor.h
+++ b/include/cantera/zeroD/IdealGasConstPressureMoleReactor.h
@@ -43,7 +43,7 @@ public:
 protected:
     vector_fp m_hk; //!< Species molar enthalpies
 
-    const int m_sidx = 1;
+    const size_t m_sidx = 1;
 };
 
 }
--- a/include/cantera/zeroD/MoleReactor.h
+++ b/include/cantera/zeroD/MoleReactor.h
@@ -47,7 +47,7 @@ protected:
     virtual void getSurfaceInitialConditions(double* y);
 
     //! const value for the species start index
-    const int m_sidx = 2;
+    const size_t m_sidx = 2;
 };
 
 }
--- a/interfaces/cython/SConscript
+++ b/interfaces/cython/SConscript
@@ -78,6 +78,9 @@ elif localenv['OS'] == 'Cygwin':
     # extract 'pythonX.Y' from 'libpythonX.Y.dll.a'
     localenv.Append(LIBS=pylib[3:-6])
 
+if "numpy_1_7_API" in localenv:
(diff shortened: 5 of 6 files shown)


include/cantera/thermo/Elements.h
```diff
@@ -128,7 +128,7 @@ double getElementWeight(const std::string& ename);
  * @exception CanteraError if the element has no stable isotopes, and
  * therefore no standard atomic weight
  */
-double getElementWeight(int atomicNumber);
+double getElementWeight(size_t atomicNumber);
```
bryanwweber: I had intentionally left these as `int`, as semantically the atomic number is not a size (of an array). I guess that's a pretty pedantic point, but is there another way to resolve this warning that maintains the semantics of the types?

ischoegl: I see your intent, but from a coding perspective, `atomicNumber` is used to pick elements out of the array `atomicWeightTable` :smile: ... I believe this fix is easier than type-casting things internally.

bryanwweber: I'm not sure how much to push back, I'm not sure how much the semantics here really matter... I don't think either of us feels very strongly about this change but it just feels a little tiny bit... odd... to make this change. The impact to users is effectively nil, so this is me being purely academic and pedantic because Cantera is the project where I can do that 😜 

That said, Boost has a function to safely convert unsigned to signed integer, [here](https://www.boost.org/doc/libs/1_55_0/libs/numeric/conversion/doc/html/boost_numericconversion/improved_numeric_cast__.html#boost_numericconversion.improved_numeric_cast__.numeric_cast) (I didn't check if it was header-only). So casting `size_t` to `int` is at least feasible 🤷‍♂️

Another option, since atomic numbers can't physically be negative, is to do a local `typedef` for an unsigned integer (it could even be an 8 bit integer, since there are only a few hundred elements). I suspect that kind of automatic type casting won't result in warnings from the compiler?

On the other hand, an unsigned integer doesn't actually enforce that the number is positive, either at compile time or runtime, so in that sense it's also not what we want this type to really represent either.

Clearly, I've thought about this _way_ more than it needs to be. In the end, I'm fine with having said everything that's above so that it's out of my head and other people can judge how crazy it is, and I'm fine with whatever decision we think is best for Cantera maintenance. Hence, this is a comment instead of requested change ☺️

speth: The [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Integer_Types), which we reference but don't strictly follow, recommends against using the unsigned types in most cases, and offers some rationale as to why.

The fairly consistent use of `size_t` was actually something I'd introduced before the release of Cantera 2.0, but I doubt I'd make the same decision now. While the standard library uses `size_t` extensively, some libraries that we work with like SUNDIALS and Eigen use signed integers for sizes.

I'm not suggesting transitioning back to signed integers broadly, but in this particular instance, I think the additional rationale for wanting to use `int` makes it a reasonable choice.

ischoegl: Interesting. It is correct that I don’t feel strongly either way - the switch to `size_t` was just the most efficient way to get rid of the warnings. Ultimately, this PR is trivial and not high priority.

bryanwweber: > While the standard library uses size_t extensively

FWIW, this is why I changed the return types of the `num*` functions when I updated the overall API. I still think these warnings are worth resolving, though... Especially because this PR addresses more than just the elements case

ischoegl:
Alright - I fixed the warnings per discussion (together with two more I located, notably the mildly annoying `NPY_1_7_API_VERSION` warning generated by Cython). 

Otherwise, I believe this is good to go from my side.

.gitignore
```diff
@@ -54,6 +54,9 @@ environment.y*
 *.gch
 coverage/
 coverage.info
+*.h5
+*.csv
+/*.yaml
```
speth: I don't think we should ignoring `.csv` files by default, given that there CSV files that are properly part of the repository.

I'm a little ambivalent about excluding `.h5` and `.yaml` files in the root directory. I guess you end up with these the same way I do, by running some of the Python examples directly from the repository root. But I'd rather this be visible as clutter that I should delete rather than just teaching git to ignore it (you can always add these to your local `.git/info/exclude`).

speth (changes requested):
Thanks, @ischoegl. It's great to see that there is finally a fix for the Numpy/Cython deprecation warning, especially after #1334 made several more instances of it appear.

I just had one minor concern related to the changes to `.gitignore`.

.gitignore
ischoegl: I don't have a problem dropping the `.gitignore` portion.

And you're right - I end up with these files all the time while testing.

ischoegl:
@speth ... thanks for the review! I dropped the `.gitignore` bit, so this should be good to go.

speth (approved):

PR #1350, opened 2022-08-02
Fix issues with "volume" and compressibility of SurfPhase

### LINKED ISSUE #1312: Confusion on meaning of "volume" for Surface / Edge phases
**Problem description**

For `Surface` phases, the definition of molar volume and how it should (or shouldn't be) used in certain calculations needs to be resolved. Currently, the `molarVolume` method for the surface phase returns the inverse of the site density, that is, a quantity in kmol/m^2 (which is not documented). This molar volume is then used in the calculation of the value returned by `density()` which then gives a value in kg/m^2. However, this mass density is not particular meaningful as it depends on how the surface species are defined, for example whether the substrate species is included in the composition of a surface "site" or not.

This definition of the molar volume is tricky, because it cannot then be used in the identity _h = u + P*v<sub>m</sub>_. The implementation of `intEnergy_mole` for the phase currently returns `enthalpy_mole()` to get around this. This would be consistent with the idea that the molar volume should just be zero (and that the density would be infinity).

```python
>>> import cantera as ct
>>> s = ct.Solution('test/data/surface-phases.yaml', 'Pt-surf')

>>> s.site_density
2.7063000000000002e-08

>>> 1/s.volume_mole
2.7063000000000005e-08

>>> s.density
5.3337681873e-06

>>> s.site_density * s.mean_molecular_weight
5.3337681873e-06
```

Also, there seems to be some inconsistency with whether the phase is compressible and what therefore the equation of state is:

```python
>>> s.is_compressible
True

>>> s.TD = None, s.density * 1.5
>>> s.density
8.00065228095e-06

>>> s.site_density * s.mean_molecular_weight
5.3337681873e-06  # no longer corresponds to density (site density has not changed)
```

**System information**

- Cantera version: 2.6.0 or `main` at 1ab81ace5

**Additional context**

This behavior was identified as part of https://github.com/Cantera/cantera/pull/1299, which implements https://github.com/Cantera/enhancements/issues/114.

speth:

- Define the molar volume of a `SurfPhase` to be zero.
- Define concentration and density-like quantities of `SurfPhase` to be per unit area (and per unit length for `EdgePhase`.
- Make `SurfPhase` incompressible, which disables setting the density to a value inconsistent with the site density.
- Modify conversion between mole fractions and coverages to avoid needing to use the density
- Fix handling of variable species "sizes" / multi-site species, which was incorrect and not subject to any tests
- Fix setting surface state using mole fractions, which did not actually preserve that sum(concentrations) == coverages. This change is responsible for the changes to the blessed output of the F90 demo.
- Fix a couple of issues with problems generating detailed error messages
- Fix a couple of tangentially-related issues with the F90 "demo" program

Closes #1312 

**Checklist**


src/thermo/SurfPhase.cpp
```diff
@@ -26,6 +26,7 @@ SurfPhase::SurfPhase(const std::string& infile, const std::string& id_) :
     m_press(OneAtm)
 {
     setNDim(2);
+    assignDensity(INFINITY);
```
ischoegl: While I acknowledge that the PR is still in draft mode, I have a question about the introduction of C++11 `INFINITY` (as defined [here](https://en.cppreference.com/w/cpp/numeric/math/INFINITY)). 

Are there additional ramifications about this choice when it comes to Cantera/enhancements#125? If I understand correctly, there won't be an issue as long as we don't check for equality, but I'm not sure what the compiler warnings for `INFINITY` correspond to in actual usage. As I was hoping for a resolution of the `--fast-math` issue for 3.0, I wanted to get a better understanding.

bryanwweber: Great question!

ischoegl: Linking https://stackoverflow.com/questions/22931147/stdisinf-does-not-work-with-ffast-math-how-to-check-for-infinity

speth: I'm not interested in pursuing support of `--ffast-math`, and have already explained why I think we shouldn't hamstring ourselves with such a limitation in the issues linked in https://github.com/Cantera/enhancements/issues/125.

ischoegl: Would it make any difference if volumetric density were (arbitrarily) defined as zero instead in this case? While I see the point of not necessarily wanting to enable a controversial optimization approach, I’d still be in favor of not introducing roadblocks when not necessary.

speth: I think that would be a little awkward -- the phase would no longer satisfy molar volume = 1 / molar density, which is a pretty fundamental identity. It would also, at a minimum, require specifically overloading `Phase::molarVolume` (currently not a virtual method).

ischoegl: It is true that zero wouldn't be consistent with the original intent stated at the top, i.e.

> * Define the "volume" of a `SurfPhase` to be zero. This implies that the density must be `inf`.

However: as the concept of a volume evaluated for a 2D geometry is fundamentally flawed I am starting to wonder whether this shouldn't be better handled as an exception?

bryanwweber: > However: as the concept of a volume evaluated for a 2D geometry is fundamentally flawed I am starting to wonder whether this shouldn't be better handled as an exception?

To avoid losing any resulting discussion, can I propose that this question and any answers move to the main thread, out of this line comment?

ischoegl:
Per @bryanwweber's request, I am putting a line comment from https://github.com/Cantera/cantera/pull/1350#discussion_r937070427 into the main thread:

It is true that [setting density] zero wouldn't be consistent with the original intent stated at the top, i.e.

> * Define the "volume" of a `SurfPhase` to be zero. This implies that the density must be `inf`.

However: as the concept of a volume evaluated for a 2D geometry is fundamentally flawed I am starting to wonder whether this shouldn't be better handled as an exception?

speth:
> However: as the concept of a volume evaluated for a 2D geometry is fundamentally flawed I am starting to wonder whether this shouldn't be better handled as an exception?

By "as an exception", you mean by having `density()`, `molarDensity()` and `molarVolume()` all return zero? I guess that would save the trouble of having to handle `inf` in YAML input/output, at least.

ischoegl:
> By "as an exception", you mean by having `density()`, `molarDensity()` and `molarVolume()` all return zero? I guess that would save the trouble of having to handle `inf` in YAML input/output, at least.

I was thinking about returning zero while issuing a warning (potentially to be turned into an error after Cantera 3.0).

speth:
> I was thinking about returning zero while issuing a warning (potentially to be turned into an error after Cantera 3.0).

I think that issuing warnings or errors about accessing these properties would be more of a headache than it's worth. I'm pretty sure there are a *lot* of places that assume that density is a useful part of a phase's state.

ischoegl:
>  I'm pretty sure there are a *lot* of places that assume that density is a useful part of a phase's state.

I would certainly agree that it's a useful quantity for 3D phases, but for 2D it becomes quite murky, hence my suggestion for a warning / exception.

include/cantera/thermo/SurfPhase.h
```diff
@@ -189,8 +210,8 @@ class SurfPhase : public ThermoPhase
      *
      * @param k Optional parameter indicating the species. The default
      *          is to assume this refers to species 0.
-     * @return
-     *   Returns the standard Concentration in units of m3 kmol-1.
+     * @return the standard concentration in units of m^2/kmol for surface phases or
```
decaluwe: I don't understand the units in these docstrings.  Aren't they inverted from those of concentrations?

decaluwe:
Been following this convo and pondering.  To me, the concept of density still applies, but its units just change with the dimensionality of the phase.  So density would have units of `kg/m^3` for a 3-D phase, and `kg/m^2` for a 2-D phase.  I don't really know how useful this quantity would be for a surface phase, but that would match what we do with concentrations, no?

Would this be a tenable approach?  Any variable whose name specifically references volume or v would use `volume = 0`, whereas anything that does not would adjust the implicit dimensionality?  So concentrations and densities would be per m^2 (or per m for edge phases)?  I looked through the files and noticed that many calls related to concentration and setConcentration were removed--are we trying to move away from/remove those functions?

Also, would `v = 0` imply that `h = u` for a surface phase?  I guess I should go and look at what our current approach is, here.

ischoegl:
@decaluwe … thanks for your input on this. I think `density` being redefined based on 2D units would make sense. The only (friendly) amendment I have is that an attempt to access any variable whose name specifically references volume or v should also raise a warning (or error after Cantera 3.0) … that way we can hopefully avoid the introduction of  `inf`.

include/cantera/thermo/SurfPhase.h
speth: Good catch. I was just correcting the fact that it's m^2 for surfaces, but this is clearly inverted as well.

speth:
> I looked through the files and noticed that many calls related to concentration and setConcentration were removed--are we trying to move away from/remove those functions?

This is related to enforcing the idea that this is an "incompressible" phase, and that the site density is fixed (unless explicitly changed). Setting the vector of concentrations is overconstrained, since the sum of the concentrations must equal the site density.

decaluwe:
Thanks, @speth --that makes sense.  One related question - can the user still change the site density via the python interface (i.e. "on the fly")? I could see that remaining a useful feature, in terms of having that be a "fittable" parameter.  But I would be in favor of requiring any altering of that parameter to require a certain amount of "intentionality," i.e. they should not be able to do it unwittingly.

(ps. sorry about the accidental close + reopen.  Was trying to close a draft comment, and clicked "close with comment" before I thought it all the way through...) 🤦

speth:
> ne related question - can the user still change the site density via the python interface (i.e. "on the fly")? I could see that remaining a useful feature, in terms of having that be a "fittable" parameter.

Yes, setting the `site_density` property in Python (and the equivalent in C++) is still possible.



> Would this be a tenable approach? Any variable whose name specifically references volume or v would use `volume = 0`, whereas anything that does not would adjust the implicit dimensionality? So concentrations and densities would be per m^2 (or per m for edge phases)?

Having thought about this a bit more, I think I agree that this works. Basically, it's saying that "volume = 1/density" is only meaningful for bulk (3D) phases, which sort of makes sense. This sidesteps the need to have anything return "inf", so I don't think it's necessary for any methods to raise warnings or exceptions.

ischoegl:
> Having thought about this a bit more, I think I agree that this works. Basically, it's saying that "volume = 1/density" is only meaningful for bulk (3D) phases, which sort of makes sense. This sidesteps the need to have anything return "inf" […]

Perfect - I’m happy with that outcome!

bryanwweber:
> Having thought about this a bit more, I think I agree that this works.

@speth As we've reached a consensus, can you document/summarize this outcome in code comments/docstrings, since it may not be obvious how or why we ended up here? And also link to this PR somewhere in the code?

speth:
> > Having thought about this a bit more, I think I agree that this works.
> 
> @speth As we've reached a consensus, can you document/summarize this outcome in code comments/docstrings, since it may not be obvious how or why we ended up here? And also link to this PR somewhere in the code?

If you read the changes, I believe that I *have* documented the new(ish) behavior, with additions to the docstring for the `Phase` class and in the newly overloaded members of the `SurfPhase` class. The former is just documenting what was already the case, and I think the docstrings for the latter are brief but clear. I don't think it's necessary to link to this PR anywhere -- as far as I can tell, that's never something we've done in the past.

P.S. I would have thought I might get a _little_ bit of space to work on this without everyone piling on, given both that it is  marked "Draft" and I did not check the box for "The pull request is ready for review".

ischoegl (approved):
Thanks, @speth - this looks good to me!

bryanwweber:
> I don't think it's necessary to link to this PR anywhere -- as far as I can tell, that's never something we've done in the past.

To my naïve eye, this looks to have generated a fair amount of in-the-weeds thermo/physical chemistry discussion. As the reasoning for why we're making this choice of behavior may not be clear in 3 months or 5 years or..., especially to new developers, it seems useful to me to provide a link to the nitty gritty of the discussion  rather than adding all of that into a comment or something, or eliding it and assuming the result is obvious. YMMV 🤷‍♂️ but I won't hold up the PR over this.

ischoegl:
Personally, I *believe* that updated docstrings, together with links in commits e6f3707 and 656dfe2, are adequate.

PR #1286, opened 2022-05-18
Yaml2ck script

bryanwweber:
**Changes proposed in this pull request**

- Add a `yaml2ck` script to convert YAML input files/Solution instances to Chemkin file(s)
- ~TODO: Add a method to `Solution` to import and call `yaml2ck.convert`~
- ~TODO: Fix up the docstring~

**If applicable, fill in the issue number this pull request is fixing**

Supersedes #1185 
I tried to push an update to that branch and it closed the PR 🤷‍♂️ 

This also relies on the work in #1266 for some of the sorting, so it should be merged after that PR.

**If applicable, provide an example illustrating new features this pull request is introducing**

```python
from cantera import yaml2ck
yaml2ck.convert("input_file.yaml")
```

**Checklist**

- [X] The pull request includes a clear description of this code change
- [X] Commit messages have short titles and reference relevant issues
- [X] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [X] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [X] The pull request is ready for review

PR #1354, opened 2022-08-05
Move Python test suite

### LINKED ISSUE #1258: Changes to Python test suite requires building of wheels
**Problem description**

The new wheels-based Python package requires rebuilding of wheels for every change to the test suite.

**Steps to reproduce**

1. Open any file in `interfaces/cython/cantera/tests/test_<any>.py` and introduce a change
3. Run `scons test-python-<any>`
4. As the test suite is part of the package, the wheel needs to be rebuilt, leading to inefficient testing.

**Behavior**

**System information**

- Cantera version: 2.6.0b2
- OS: any
- Python/MATLAB/other software versions: any

**Additional context**

One potential option would be to no longer package the Python tests in the wheel ... resolution should not prevent the release of 2.6 though.

ischoegl:

* Move Python test suite out of Python module.
* Redirect output of `test_convert.py::yaml2ckTest` to work folder

Closes #1258

After the proposed changes
```
$ pytest test/python
[...]
$ pytest test/python/test_onedim.py::TestDiffusionFlame::test_mixture_averaged
[...]
```
just works.

**Checklist**


speth (changes requested):
Thanks for making these changes, @ischoegl. I was pleased to see that this change fixes an issue I had been experiencing, where pytest's fancy introspection didn't work and wasn't able to show the values of the arguments of `assert X == Y` when the assertion failed.

I have two suggestions. First, if you add a `test/python/pytest.ini` file with the contents
```ini
[pytest]
pythonpath = ../../build/python
```
you will be able to run `pytest test/python` and variants directly from the root of the repository, without needing to set the `PYTHONPATH` (or accidentally running the tests from a previously-installed copy of Cantera). I also found that this let me set up integration with VS Code, which I think will be really nice to use.

Second, I'd like to suggest squashing some of the commits here that are not logically separate. For example, moving the test suite and fixing the integration with SCons. Changes like could lead to annoying errors while using `git bisect` in the future, where trying to run the tests with this intermediate commit just doesn't work. I think the same concern may be applicable to some of the commits in #1352.

ischoegl:
@speth ... thank you for the review. Especially your comment about `pytest.ini` is valuable, as I did run my tests for `pytest` after checking that `README.txt` installs correctly (which presumably shadowed some issues that would arise without running `scons install` first). I am also happy that this PR appears to resolve some issues that you have been experiencing. At this point, I confirmed that running 

```
$ pytest test/python
```

works after removing previous installs via `scons uninstall`.

Per suggestion, I also squashed commits - thus far, when committing, I was usually more concerned about keeping steps separate so I can trace back easily, but your point about `git bisect` is well taken.

speth (changes requested):
Checking out just the first commit here does not build, with the error: 
> ```scons: *** [build/python/dist/Cantera-3.0.0a2-cp310-cp310-macosx_11_0_arm64.whl] Explicit dependency `build/python/cantera/test/README.txt' not found, needed by target `build/python/dist/Cantera-3.0.0a2-cp310-cp310-macosx_11_0_arm64.whl'.```

And as expected, running `scons test` with just the first two commits results in a bunch of pollution in the `test/data` directory.

So I'd like to suggest squashing at least the first of those commits, and possibly the second as well.

ischoegl:
@speth ... I reordered the commits, so they now all build individually. The 'pollution' issue actually goes back to #1286, but was not apparent as it was polluting the `build` folder. I applied that fix - which I believe is completely unrelated to the main point of this PR - first.

speth (approved):
Thanks, @ischoegl. I think this is good to go.

PR #1357, opened 2022-08-07
Update/simplify gitignore

ischoegl:

- remove `interfaces/cython/.gitignore` and `interfaces/python_minimal/.gitignore` (no longer needed)
- simplify logic in `test_problems/.gitignore` and remove obsolete code
- ~add `samples/.gitignore`~
- ~handle `ImportError` in `vapordome.py` example~

**Checklist**


samples/python/thermo/vapordome.py
```diff
@@ -7,10 +7,14 @@
...
 
-import cantera as ct
-import pandas as pd
 import numpy as np
-from matplotlib import pyplot as plt
+try:
+    import pandas as pd
+except ImportError:
+    print("Example requires pandas installation")
+    exit(0)
```
bryanwweber: I'm not a big fan of this pattern. The error message and resolution is already obvious from the import error, and this makes the happy path example code a little harder to follow for new Python people, especially if they're coming from Matlab. For optional dependencies, it _may_ make sense but even then, I'm skeptical. YMMV ☺️

ischoegl: Do you have a better suggestion? These patterns are widely used elsewhere in the Python sample folders. The only thing I did differently here was to add the `exit(0)` so I don't have to indent everything.

Fwiw, this came up when running your nifty:
```
find samples/python -type f -iname "*.py" \
```
locally, which crashes out for this example when `pandas` is not installed.

bryanwweber: That's kinda the point, isn't it? For it to fail when dependencies aren't installed, so we know one of the samples isn't being run/tested? I know this is used elsewhere in the examples but IMO it's an anti-pattern for example code. I think it's probably a holdover from when it was difficult to install dependencies, especially on Windows, but that's been more-or-less solved for a few years now. I don't think it's a huge deal, but I don't think it actually helps users, and it may be harmful insofar as it allows the CI to pass when it really shouldn't. I'd say if you're running locally, you can add an exception to the `find` not to match this example or install pandas 🤷‍♂️ ☺️ I don't think we need to immediately rip it out of the other examples but I'd prefer not to add it in new places. In fact, removing this pattern from old examples would probably be a good "first issue" for a new contributor 🤔

ischoegl: :+1: ... I see your point, and dropped the commit.

ischoegl:
I ended up not adding any new excludes, so this just eliminates entries that are no longer needed while consolidating others.

At this point, the build process no longer pollutes the `interfaces` Python folders, so several `.gitignore` files are obsolete. I verified that no pollution occurs on the following OS's:
* [x] Linux (Fedora)
* [x] Windows 10 (including MATLAB)
* [x] macOS

bryanwweber (approved):
Thanks!

PR #1363, opened 2022-08-11
Implementation of MoleReactor and ConstantPressureMoleReactor

anthony-walker:

- Implement MoleReactor class
- Implement ConstPressureMoleReactor class
- Add corresponding python interfaces and inherit existing python tests
- Add test to compare to integration of mass fraction reactors with surface chemistry
- Supports https://github.com/Cantera/enhancements/issues/47
-  Bug fix in `test_reactor.py::test_component_names`

**Checklist**

### DIFF
--- a/include/cantera/zeroD/ConstPressureMoleReactor.h
+++ b/include/cantera/zeroD/ConstPressureMoleReactor.h
@@ -1,0 +1,45 @@
+//! @file ConstPressureMoleReactor.h
+
+// This file is part of Cantera. See License.txt in the top-level directory or
+// at https://cantera.org/license.txt for license and copyright information.
+
+#ifndef CT_CONSTPRESSMOLE_REACTOR_H
+#define CT_CONSTPRESSMOLE_REACTOR_H
+
+#include "cantera/zeroD/MoleReactor.h"
+
+namespace Cantera
+{
+
+/*!
+ * ConstPressureMoleReactor is a class for constant-pressure reactors
+ * which use a state of moles.
+ */
+class ConstPressureMoleReactor : public MoleReactor
+{
+public:
+    ConstPressureMoleReactor() {}
+
+    virtual std::string type() const {
+        return "ConstPressureMoleReactor";
+    };
+
+    virtual size_t componentIndex(const std::string& nm) const;
+
+    virtual std::string componentName(size_t k);
+
+    virtual void getState(double* y);
+
+    virtual void initialize(double t0 = 0.0);
+
+    virtual void eval(double t, double* LHS, double* RHS);
+
+    virtual void updateState(double* y);
+
+protected:
+    const size_t m_sidx = 1;
+};
+
+}
+
+#endif
--- a/include/cantera/zeroD/ConstPressureMoleReactor.h
+++ b/include/cantera/zeroD/ConstPressureMoleReactor.h
@@ -14,6 +14,7 @@ namespace Cantera
 /*!
  * ConstPressureMoleReactor is a class for constant-pressure reactors
  * which use a state of moles.
+ * @SInCE New in Cantera 3.0
  */
 class ConstPressureMoleReactor : public MoleReactor
 {
--- a/include/cantera/zeroD/ConstPressureMoleReactor.h
+++ b/include/cantera/zeroD/ConstPressureMoleReactor.h
@@ -14,7 +14,7 @@ namespace Cantera
 /*!
  * ConstPressureMoleReactor is a class for constant-pressure reactors
  * which use a state of moles.
- * @SInCE New in Cantera 3.0
+ * @since New in Cantera 3.0
  */
 class ConstPressureMoleReactor : public MoleReactor
 {
--- a/interfaces/cython/cantera/reactor.pyx
+++ b/interfaces/cython/cantera/reactor.pyx
@@ -375,6 +375,8 @@ cdef class MoleReactor(Reactor):
     they are closed (no inlets or outlets), have fixed volume, and have adiabatic,
     chemically-inert walls. These properties may all be changed by adding
     appropriate components such as `Wall`, `MassFlowController` and `Valve`.
+    
+    .. versionadded:: 3.0    
     """
     reactor_type = "MoleReactor"
 
--- a/interfaces/cython/cantera/reactor.pyx
+++ b/interfaces/cython/cantera/reactor.pyx
@@ -369,6 +369,14 @@ cdef class Reactor(ReactorBase):
             limit = -1.
         self.reactor.setAdvanceLimit(stringify(name), limit)
 
+cdef class MoleReactor(Reactor):
+    """
+    A homogeneous zero-dimensional reactor with a mole based state vector. By default,
+    they are closed (no inlets or outlets), have fixed volume, and have adiabatic,
+    chemically-inert walls. These properties may all be changed by adding
+    appropriate components such as `Wall`, `MassFlowController` and `Valve`.
+    """
+    reactor_type = "MoleReactor"
 
 cdef class Reservoir(ReactorBase):
     """
@@ -385,6 +393,13 @@ cdef class ConstPressureReactor(Reactor):
     pressure constant.
     """
     reactor_type = "ConstPressureReactor"
+
+cdef class ConstPressureMoleReactor(Reactor):
+    """A homogeneous, constant pressure, zero-dimensional reactor with a mole based
+    state vector. The volume of the reactor changes as a function of time in order to
+    keep the pressure constant.
+    """
+    reactor_type = "ConstPressureMoleReactor"
 
 
 cdef class IdealGasReactor(Reactor):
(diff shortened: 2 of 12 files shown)


src/zeroD/ConstPressureMoleReactor.cpp
```diff
@@ -0,0 +1,146 @@
...
+{
+    if (m_thermo == 0) {
+        throw CanteraError("ConstPressureMoleReactor::getState",
+                           "Error: reactor is empty.");
+    }
+    m_thermo->restoreState(m_state);
+    // get mass for calculations
+    m_mass = m_thermo->density() * m_vol;
+    // set the first component to the temperature
+    y[0] = m_thermo->enthalpy_mass() * m_thermo->density() * m_vol;
```
speth: The comment here is incorrect and misleading.

src/zeroD/ConstPressureMoleReactor.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+void ConstPressureMoleReactor::getState(double* y)
+{
+    if (m_thermo == 0) {
+        throw CanteraError("ConstPressureMoleReactor::getState",
+                           "Error: reactor is empty.");
+    }
+    m_thermo->restoreState(m_state);
+    // get mass for calculations
+    m_mass = m_thermo->density() * m_vol;
```
speth: The value of `m_mass` is not used in this case, so I don't think there's a need to set it in this method.

test/python/test_reactor.py
```diff
@@ -220,10 +221,9 @@ def integrate(atol, rtol):
 
         n_baseline = integrate(1e-10, 1e-20)
         n_rtol = integrate(5e-7, 1e-20)
-        n_atol = integrate(1e-10, 1e-6)
-
-        self.assertTrue(n_baseline > n_rtol)
-        self.assertTrue(n_baseline > n_atol)
+        n_atol = integrate(1e-10, 1e-5)
+        self.assertTrue(n_baseline > n_rtol, f"{n_baseline} is less than {n_rtol}")
+        self.assertTrue(n_baseline > n_atol, f"{n_baseline} is less than {n_atol}")
```
speth: This is even simpler if you make use of the fact that we're using `pytest` now (which will use introspection magic to provide nice test failure messages):
```suggestion
        assert n_baseline > n_rtol
        assert n_baseline > n_atol
```

test/python/test_reactor.py
```diff
@@ -785,6 +785,55 @@ def test_bad_kwarg(self):
...
+        surf2.coverages = {"PT(S)":1}
+        rsurf2 = ct.ReactorSurface(surf2, r2, A=surf_area)
+        # reactor network setup
+        net1 = ct.ReactorNet([r1,])
+        net2 = ct.ReactorNet([r2,])
+        net1.rtol = net2.rtol = 1e-9
+        net1.atol = net2.atol = 1e-18
+        for i in np.arange(0.1, 1.1, 0.1):
+            net1.advance(i)
+            net2.advance(i)
```
speth: You should give some consideration to the time step and interval that's being tested here. For the problem at hand, the system has basically reached steady-state long before the first output time.

speth (changes requested):
Thanks, @anthony-walker, it's good to see the set of reactor types more fleshed out.

Besides the few comments below, one thing I think you should do is add the `Extensible...Reactor` wrappers for all of the mole-based reactor classes so they can be used as the basis for user-defined reactors as well.

src/zeroD/ConstPressureMoleReactor.cpp
anthony-walker: `m_mass` is used in `getMoles` so it has to be set here to properly get moles. I will make a comment of this in the code.

anthony-walker:
@speth Thank you for the review. I have made all of the requested changes. Aside from the following because `m_mass` is used in the `getMoles` function. I have added a comment to reflect this as well and avoid confusion.
> The value of `m_mass` is not used in this case, so I don't think there's a need to set it in this method.

I was considering potentially adding some tests for the `Extensible...MoleReactor` classes as well if you think necessary. Thoughts?

src/zeroD/ConstPressureMoleReactor.cpp
speth: But the purpose of the `getState` function is to determine the vector `y` that corresponds to the current state of the reactor, so `m_mass` should already be current. The only functions that should set `m_mass` are `updateState` (to set the state from a portion of the `ReactorNet` state vector) or `syncState` (to set the state from a `ThermoPhase` object).

include/cantera/zeroD/ConstPressureMoleReactor.h
```diff
@@ -0,0 +1,45 @@
...
+#define CT_CONSTPRESSMOLE_REACTOR_H
+
+#include "cantera/zeroD/MoleReactor.h"
+
+namespace Cantera
+{
+
+/*!
+ * ConstPressureMoleReactor is a class for constant-pressure reactors
+ * which use a state of moles.
```
ischoegl, line 16:
```suggestion
 * ConstPressureMoleReactor is a class for constant-pressure reactors
 * which use a state of moles.
 * @since New in Cantera 3.0
```
Newly introduced objects should be marked (same applies for new methods) ... (*Edit: looks like there is a user handle that prevents proper rendering in the comment; to make sure, it should be `@since`.*)

interfaces/cython/cantera/reactor.pyx
```diff
@@ -369,6 +369,14 @@ cdef class Reactor(ReactorBase):
             limit = -1.
         self.reactor.setAdvanceLimit(stringify(name), limit)
 
+cdef class MoleReactor(Reactor):
+    """
+    A homogeneous zero-dimensional reactor with a mole based state vector. By default,
+    they are closed (no inlets or outlets), have fixed volume, and have adiabatic,
+    chemically-inert walls. These properties may all be changed by adding
+    appropriate components such as `Wall`, `MassFlowController` and `Valve`.
+    """
```
ischoegl:
```suggestion
    """
    A homogeneous zero-dimensional reactor with a mole based state vector. By default,
    they are closed (no inlets or outlets), have fixed volume, and have adiabatic,
    chemically-inert walls. These properties may all be changed by adding
    appropriate components such as `Wall`, `MassFlowController` and `Valve`.
    
    .. versionadded:: 3.0    
    """
```
As in C++, new code should be marked.

ischoegl (commented):
Just as a minor comment, as of 3.0 we are trying to keep track of newly introduced features with doxygen and Sphinx tags (should have pointed that out in #1003 and #1377).

src/zeroD/ConstPressureMoleReactor.cpp
anthony-walker: @speth this is also done in `IdealGasReactor` and the `Reactor` classes. I can attempt to not repeat this in the mole reactor classes though. However, it seems to be causing issues with the test. I will try to track down the reasoning for this.

speth: I did not realize that this was the case in the standard `Reactor` and `IdealGasReactor` classes. In that case, I think it's OK. I still think there's something kind of funky about it, but I guess there is something in the existing evaluation logic that relies on this, and I don't think you need to unravel that here.

anthony-walker: In that case, I believe this is ready for further review or merging.
I will rebase with main and push of course first.

include/cantera/zeroD/ConstPressureMoleReactor.h
```diff
@@ -0,0 +1,46 @@
...
+ * @SInCE New in Cantera 3.0
```
speth: GitHub is really making a mess out of this. The annotation command should just be lowercase `since` preceded by an `@`, like every other Doxygen command. The funky capitalization came from GitHub deciding that this was _clearly_ a mention of a user who has their name capitalized that way.

interfaces/cython/cantera/reactor.pyx
```diff
@@ -386,6 +396,13 @@ cdef class ConstPressureReactor(Reactor):
     """
     reactor_type = "ConstPressureReactor"
 
+cdef class ConstPressureMoleReactor(Reactor):
+    """A homogeneous, constant pressure, zero-dimensional reactor with a mole based
+    state vector. The volume of the reactor changes as a function of time in order to
+    keep the pressure constant.
+    """
```
speth, line 405: I suppose this class should also get the `.. versionadded:: 3.0` annotation. Likewise all of the new `Extensible...Reactor` classes added in this PR.

speth (changes requested):
Thanks for the updates, @anthony-walker. I agree with you on leaving the use of `m_mass` in `getState` alone for now.

I think the only remaining issues are two small documentation updates.

include/cantera/zeroD/ConstPressureMoleReactor.h
anthony-walker: Ah, I assumed it was just rendered that way and not committed that way. Making these changes now.

anthony-walker:
@speth and @ischoegl, thank you for the review. Assuming all checks pass, I have addressed all of your comments and concerns.

speth (approved):
Thanks, @anthony-walker. This looks good to me. I'll merge once the tests all pass.

PR #1384, opened 2022-09-05
Add Solution to 1D boundaries

ischoegl:

This PR is a follow-up to #1345 and adds `Solution` objects to all `Boundary1D` specializations instantiated from the Python API.

- `ReactingSurf1D` receives interface (instead of gas)
- Other `Boundary1D` objects receive gas phase adjacent to boundary (no change)
- ~Deprecate support for non-keyword `phase` arguments in `Domain1D` interface~

Preliminary work for Cantera/enhancements#137

**Checklist**


bryanwweber:
> Deprecate support for non-keyword phase arguments in Domain1D interface

I'm curious about the motivation for this change. I didn't see it discussed elsewhere, although I may have missed it 🤪

ischoegl:
> > Deprecate support for non-keyword phase arguments in Domain1D interface
> 
> I'm curious about the motivation for this change. I didn't see it discussed elsewhere, although I may have missed it 🤪

... you didn't miss anything, but the old implementation was imho error-prone.

The Python 1D API uses `*args, **kwargs` extensively, effectively allowing to pass `phase` as either. One of the `super` constructors does eventually require `phase` to be a positional argument, which means that there are some lurking issues: `Boundary1D.gas` (which is rarely used) may not hold the expected object. I believe having a well-defined interface is ultimately the better option here.

src/oneD/Boundary1D.cpp
```diff
@@ -610,6 +610,30 @@ ReactingSurf1D::ReactingSurf1D()
     m_type = cSurfType;
 }
 
+ReactingSurf1D::ReactingSurf1D(shared_ptr<Solution> sol)
```
bryanwweber:
```suggestion
ReactingSurf1D::ReactingSurf1D(shared_ptr<Solution> solution)
```

include/cantera/oneD/Boundary1D.h
```diff
@@ -106,6 +106,10 @@ class Inlet1D : public Boundary1D
 public:
     Inlet1D();
 
+    Inlet1D(shared_ptr<Solution> sol) : Inlet1D(){
```
bryanwweber:
```suggestion
    Inlet1D(shared_ptr<Solution> solution) : Inlet1D(){
```
And likewise elsewhere

bryanwweber (commented):
> `Boundary1D.gas` (which is rarely used) may not hold the expected object

If this is the problem to fix, then enforcing `phase` to always be a kwarg is a pretty intrusive change, IMO. As it's always required it has the effect of increasing verbosity and restricting how the function can be called for the default use case of most of the API (although not the main user facing classes, `FreeFlame` etc. fortunately). It may as well be positional-only, to have the same effect of enforcing a standard API, no? That's not necessarily a serious suggestion, just to point out that I think it's not a clear win to make this kwarg-only. But this is deliberately a comment not a request-for-change review because in the end it probably doesn't make much of a difference whatever is chosen here.

speth:
> One of the `super` constructors does eventually require `phase` to be a positional argument, which means that there are some lurking issues: `Boundary1D.gas` (which is rarely used) may not hold the expected object.

Can you give an example of this? It isn't obvious to me. The signature of `Domain1D.__init__` does require a `phase` object, but it's not required to be a positional argument:
```py
cdef class Domain1D:
    ...
    def __init__(self, _SolutionBase phase, *args, name=None, **kwargs):
```

ischoegl:
@speth ... I just removed the code portions in question (while making it impossible to pass `phase` as a non-positional argument), so the issue is moot. (*I may have gotten lost with all the unspecific `*args, **kwargs`*)

ischoegl:
@bryanwweber ... based on your feedback, I reverted the changes. Instead, I made the `Boundary1D` interface much stricter - which is consistent with the existing test suite, while throwing more descriptive errors (which now also has a unit test).

speth (commented):
I get the impression that this is working toward something a bit larger -- can I ask what that might be, and how this contributes?

Adding `Solution` objects to the `StFlow` class in #1345  made sense, but I hadn't paid enough attention to the idea that it was also adding a `Solution` object to every `Domain1D` object. In particular, I think this ends up being a bit messy since for `Boundary1D` objects, there are in general two phases that matter: one associated with the bulk/flow domain to the left, and one to the right. Hence the `m_phase_left` and `m_phase_right` member variables of the `Boundary1D` class, which are automatically set when the containing `OneDim` object is initialized. I realize, however, that we don't currently provide an implementation of a `Boundary1D` that would actually sit between two flow domains, and this idea isn't supported at all in the Python interface.

The only case where a `Boundary1D` object has a phase object of its own that isn't already part of a neighboring domain is the `ReactingSurf1D` class. Other than that, the newly added `Solution` object seems to go unused, as far as I can tell.

ischoegl:
> I get the impression that this is working toward something a bit larger -- can I ask what that might be, and how this contributes?

See top of page: _Preliminary work for https://github.com/Cantera/enhancements/issues/137_

I have a reasonable draft for an import of C++ solution arrays from YAML and HDF formats for Sim1D that are already part of 2.6 (which I can post). `Solution` objects are necessary to ensure that the states at the boundaries can be set properly. This is of particular importance for `ReactingSurf1D`. 

*Edit: I posted #1385 to provide context*

ischoegl:
To answer a little more in detail: 

> for `Boundary1D` objects, there are in general two phases that matter: one associated with the bulk/flow domain to the left, and one to the right. Hence the `m_phase_left` and `m_phase_right` member variables of the `Boundary1D` class, which are automatically set when the containing `OneDim` object is initialized. I realize, however, that we don't currently provide an implementation of a `Boundary1D` that would actually sit between two flow domains, and this idea isn't supported at all in the Python interface.

This is of course all true. At the same time, Python does provide a `gas` object for boundaries objects without specifying left or right (and requires a phase object when the boundary is created). All of them are external boundaries, so there is no ambiguity and doing the same for those objects in C++ would also make sense. As you probably saw I did make the surface phase the primary phase being added for `ReactingSurf1D` which I believe is the intuitive approach. 

Beyond, I see boundary left/right in C++ as part of a more generic internal interface that was never fully implemented, as a dedicated internal boundary object is missing (as you noted). This PR is not trying to change any of that. At the moment, it is also impossible to replace the boundary left/right route as `clib` does not know about `Solution`.

include/cantera/oneD/Boundary1D.h
```diff
@@ -106,6 +106,10 @@ class Inlet1D : public Boundary1D
 public:
     Inlet1D();
 
+    Inlet1D(shared_ptr<Solution> solution) : Inlet1D(){
```
speth: Minor nitpick: there should be a space before the `{`.

ischoegl:
@speth … any further comments?

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -342,8 +345,8 @@ cdef class Inlet1D(Boundary1D):
     domain - it must be either the leftmost or rightmost domain in a
     stack.
     """
-    def __cinit__(self, *args, **kwargs):
-        self.inlet = new CxxInlet1D()
+    def __cinit__(self, _SolutionBase phase, **kwargs):
```
speth: The `__cinit__` functions should retain `*args` and `**kwargs` in addition to any specific arguments. Otherwise, it is impossible to introduce derived types that take additional arguments since `__cinit__` is always called with the original argument list.

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -10,15 +10,18 @@ from ._utils cimport stringify, pystr
...
+
 # Need a pure-python class to store weakrefs to
 class _WeakrefProxy:
     pass
 
 cdef class Domain1D:
-    def __cinit__(self, *args, **kwargs):
+    def __cinit__(self, _SolutionBase phase, **kwargs):
+        if not isinstance(phase, _SolutionBase):
+            raise TypeError(f"Received phase with invalid type '{type(phase)}'.")
```
speth: This should be sufficient to raise a `TypeError` when needed:
```suggestion
    def __cinit__(self, _SolutionBase phase not None, **kwargs):
```

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -396,24 +399,54 @@ cdef class SymmetryPlane1D(Boundary1D):
...
+        The (surface) phase corresponding to the boundary
+
+    .. versionchanged:: 3.0
+
+        Starting in Cantera 3.0, parameter `phase` should reference surface instead of
+        gas phase.
+    """
+    def __cinit__(self, _SolutionBase phase, **kwargs):
+        if isinstance(phase, _SolutionBase) and phase.phase_of_matter != "gas":
+            self.surf = new CxxReactingSurf1D(phase._base)
```
speth: I think the `isinstance` check is unnecessary -- the base class `__cinit__` should already catch this, right?

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -396,24 +399,54 @@ cdef class SymmetryPlane1D(Boundary1D):
...
+                gas = None
+                for val in sol._adjacent.values():
+                    if val.phase_of_matter == "gas":
+                        gas = val
+                        break
+                if gas is None:
+                    raise CanteraError("ReactingSurface1D needs an adjacent gas phase")
+                super().__init__(gas, name=name)
+        else:
+            super().__init__(phase, name=name)
```
speth: Is it actually possible to end up in this `else` clause?

src/oneD/Boundary1D.cpp
```diff
@@ -610,6 +610,30 @@ ReactingSurf1D::ReactingSurf1D()
...
+    }
+    if (!std::dynamic_pointer_cast<InterfaceKinetics>(solution->kinetics())) {
+        throw CanteraError("ReactingSurf1D::ReactingSurf1D",
+            "Detected incompatible kinetics type '{}'",
+            solution->kinetics()->kineticsType());
+    }
+    m_solution = solution;
+    m_kin = (InterfaceKinetics*)solution->kinetics().get();
+    m_surfindex = m_kin->surfacePhaseIndex();
+    m_sphase = (SurfPhase*)&m_kin->thermo(m_surfindex);
```
speth: To get rid of the C-style casts (which I would like to avoid), you could assign `m_kin` and `m_sphase` using `dynamic_pointer_cast` and immediately after each do the check that they are non-zero.

speth (changes requested):
I think this is generally fine. Just had a few small questions / suggestions.

interfaces/cython/cantera/_onedim.pyx
ischoegl: nice - didn't know this one!

interfaces/cython/cantera/_onedim.pyx
ischoegl: Nice catch - this was left over from an earlier version.

src/oneD/Boundary1D.cpp
ischoegl: Done. I also further simplified the logic, as some of these checks were redundant.

ischoegl:
@speth ... thanks for the review! I believe I addressed everything.

@bryanwweber ... I hope you don't mind me cherry-picking your fix for the matplotlib warning from the `units` PR.

speth (approved):
Thanks, looks good to me!

PR #1385, opened 2022-09-10
Add native SolutionArray / HDF support to C++ OneDim

ischoegl:

Implement native support for HDF data in C++ `OneDim` using [HighFive](https://github.com/BlueBrain/HighFive). Similar to the Python `h5py` implementation, a `SolutionArray` (also in C++) acts as an intermediate container for saving/restoring of `Domain1D` data.

- [x] Implement import of HDF data to C++ `SolutionArray` (using existing HDF format; since Cantera 2.5)
- [x] Implement import of YAML data to C++ `SolutionArray` (using existing YAML format; since Cantera 2.6)
- [x] Restore `Domain1D` from `SolutionArray`
- [x] Save `Domain1D` to `SolutionArray`
- [x] Save C++ `SolutionArray` to HDF (make HDF format consistent with YAML structure; implemented in C++)
- [x] Save C++ `SolutionArray` to YAML
- Link C++ `SolutionArray` data to Python API (~likely~ *definitely* as separate PR)

Extensions to `zeroD` reactor networks are anticipated.

Once `Solution` is supported by `clib`, HDF can be made available to any user-facing API.

Partially addresses https://github.com/Cantera/enhancements/issues/137 and https://github.com/Cantera/enhancements/issues/158

```
>>> sim = ct.FreeFlame(gas)
>>> sim.restore("adiabatic_flame.h5", name="mix")
```

**Checklist**


ischoegl:
@speth ... ~while this is not fully implemented,~ feel free to have a look (feedback is explicitly welcome). At this point, restoring of `Sim1D` from existing YAML and HDF file formats via a newly implemented C++ `SolutionArray::restore` works. The approach mostly follows the YAML serialization, but inserts the C++ `SolutionArray` so various file IO formats can be supported in one central place. HDF uses [BlueBrain/HighFive](https://github.com/BlueBrain/HighFive) - thus far, I haven't experienced any major difficulties.

Remaining work for this PR involves implementing `SolutionArray::save` for ~HDF and~ YAML. It's not a major effort, but it may take me a couple of days to get back to this.

ischoegl:
@speth ... at this point, round-trip HDF save/restore of `Sim1D` for the updated format works, and the old HDF format can be read without `h5py` also. There isn't really a pressing need to implement the YAML bits that are missing - they're not needed for `Sim1D`, but will become helpful once this is tied to the Python `SolutionArray` (another PR).

Some of the main points of clarification are:
- Go/no-go for HighFive
- At the moment, HighFive is installed from conda and is optional - I added CI tests for the runners that are based on conda; as our CI coverage jobs use apt-get, coverage isn't straight-forward.

Once I have some feedback, I'll go ahead and track down some of the missing pieces (not all of the Domain1D specializations are supported/tested yet)

speth:
I think an _optional_ dependency on libhdf5 and HighFive is completely fine.

You might consider adding a submodule for HighFive, like we do for Eigen, since it's a header-only dependency that isn't widely packaged (outside Conda, I guess).

ischoegl:
@speth ... thanks for the comments! I worked a little more on this and the structure is in good shape (one exception: one small bug that I need to track down). I can likely clean the code up further by wrapping HighFive in a dedicated class; I agree that it would make sense to create a submodule. 

I will address these remaining items together with review comments.

ischoegl:
@speth ... I *think* the submodule is starting to work - I have it working on Fedora (and it should run on macOS also). However, ubuntu appears to put `libhdf5.so` into a subfolder - potentially `hdf5/serial/libhdf5.so` (see [discussion here](hdf5/serial/libhdf5.so))? I don't have an ubuntu machine to check. Any recommendations?

speth:
Oh, that's kind of annoying, though I guess I get the reasoning for being able to support multiple HDF5 libraries compiled against different MPI libraries.

On most Linux systems, I think you could append the output of `pkg-config --libs hdf5` to the link line to get the right directories. For the sake of CI, that would be append to either the `debug_linker_flags` or `no_debug_linker_flags` SCons option. I'm less sure whether we'd want to incorporate that into the SCons config itself.

ischoegl:
> Oh, that's kind of annoying, though I guess I get the reasoning for being able to support multiple HDF5 libraries compiled against different MPI libraries.
> 
> On most Linux systems, I think you could append the output of `pkg-config --libs hdf5` to the link line to get the right directories. For the sake of CI, that would be append to either the `debug_linker_flags` or `no_debug_linker_flags` SCons option. I'm less sure whether we'd want to incorporate that into the SCons config itself.

@speth - thanks! Unfortunately, I haven't had much luck with `pkg-config` on my fedora (it doesn't find anything) ... the only way that worked was to use `whereis libhdf5.so` (which on Fedora yields `/lib64/libhdf5.so`). The way I detect `hdf5` in SCons is to use `conf.CheckLib` ...so presumably I would have to provide the  location of `libhdf5.so` using the option `'extra_lib_dirs'`. Do you happen to have an Ubuntu installation somewhere, so we know where `libhdf5-dev` (or `pip install h5py` for that matter) puts the libraries?

ischoegl:
Alright ... partial success. For Ubuntu, I guessed the apt `libhdf5-dev` install locations correctly, with:
```
        extra_lib_dirs=/usr/lib/x86_64-linux-gnu/hdf5/serial
        extra_inc_dirs=/usr/include/hdf5/serial
```

This works for Python 3.11, but apparently not for earlier Python versions, where there are failures due to the following:
```
Warning! ***HDF5 library version mismatched error***
[9266](https://github.com/Cantera/cantera/actions/runs/3605901930/jobs/6076720020#step:9:9267)
The HDF5 header files used to compile this application do not match
[9267](https://github.com/Cantera/cantera/actions/runs/3605901930/jobs/6076720020#step:9:9268)
the version used by the HDF5 library to which this application is linked.
```

I.e. `libhdf5-dev` appears to clash with `h5py`. I guess that the correct library depends on whether `h5py` is configured or not.

ischoegl:
:tada: ... CI finally runs through, with `h5py` and `HighFive` using the same HDF libraries even on Ubuntu; this apparently also resolves the restriction for `h5py` on Python 3.11.

I marked the new additions as 'experimental' as I am aware of the potential for a couple of improvements:

* ~HighFive is actually not installed (not sufficiently familiar with SCons); once it is installed, grabbing version information from HDF will be possible (a minor addition for SCons output).~ *Edit: this would involve a dependency to CMake; as there are only minor portions that require configuration, benefits of installing are overall questionable*
* Eigen can be used to avoid copying of data (HighFive does support it)
* ~Chunksize and compression level for HDF are currently left at default~ *Edit: done*

There currently is no connection between C++ and Python versions of `SolutionArray`; for the time being, the C++ version mostly handles file input and output (with both HDF and YAML being supported; the addition of CSV is straightforward but doesn’t make sense for `Sim1D`). As this PR just hit 2K lines, I will defer other enhancements to a follow-up PR. One motivation for this work is to extend this to handle reactor networks.

*Edit - some additional thoughts for things that should be taken care of prior to a merge:*

- ~`libhdf5` needs to be linked statically??~
- [x] Some remaining references to `HighFive::File` outside of `Storage.h` should be removed
- [x] Compression and chunk size should be addressed
- [x] An option to disable HDF support needs to be included in SCons
- [x] `SolutionArray.write_hdf` should be deprecated.

These items require only minimal code changes, and are independent of any other review comments.

ischoegl:
@speth ... from my side, the code portion of this PR is converged. 

The one remaining aspect is that this PR uncovered potential library version conflicts in calls from C++ to Python (e.g. `ExtensibleRate`) if a Python instance isn't already active - in this specific case, `h5py` ships with its own set of binaries, and checks whether versions match; while many environments use the same version, this isn't always the case (Ubuntu is one example). The version mismatch itself isn't a problem when used from Python (`h5py` links to one `libhdf5` version, while `cantera` links to another), but a call from C++ to Python creates issues, with a check run by `h5py` essentially breaking those calls (albeit only when called from the C++ core without launching Python, as is the case in the C++ unit tests?). For the case of `h5py`, installing with `pip install --no-binary=h5py h5py` is one workaround. At the moment, this PR disables the google test (only if native HDF support is present), which illustrates that `pytest` checks for `ExtensibleRate` still work.

include/cantera/base/stringUtils.h
```diff
@@ -102,6 +102,22 @@ doublereal fpValueCheck(const std::string& val);
 void tokenizeString(const std::string& oval,
                     std::vector<std::string>& v);
 
+//! This function separates a string up into tokens according to the location of
+//! path separators.
+/*!
+ * White space includes the new line character. tokens are stripped of leading
+ * and trailing white space.
```
speth: I'm guessing this business about "white space" is copypasta from `tokenizeString`.

src/base/stringUtils.cpp
```diff
@@ -195,6 +195,13 @@ void tokenizeString(const std::string& in_val, std::vector<std::string>& v)
     ba::split(v, val, ba::is_space(), ba::token_compress_on);
 }
 
+void tokenizePath(const std::string& in_val, std::vector<std::string>& v)
+{
+    std::string val = ba::trim_copy(in_val);
+    v.clear();
+    ba::split(v, val, ba::is_any_of("/\\:"), ba::token_compress_on);
```
speth: I understand splitting on either `/` or `\`, but why `:`?

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,680 @@
...
+        throw CanteraError("SolutionArray::SolutionArray",
+            "Unable to create SolutionArray from invalid Solution object.");
+    }
+}
+
+void SolutionArray::initialize(const std::vector<std::string>& extra)
+{
+    m_stride = m_sol->thermo()->stateSize();
+    m_work.reset(new vector_fp(m_size * m_stride, 0.));
+    m_data = m_work->data();
```
speth: This data structure seems more complex than necessary. Is there a reason not to just have `m_data` be a `vector_fp` of the correct size?

include/cantera/base/SolutionArray.h
```diff
@@ -0,0 +1,194 @@
...
+    static shared_ptr<SolutionArray> create(const shared_ptr<Solution>& sol,
+                                            size_t size=0,
+                                            const AnyMap& meta={})
+    {
+        return shared_ptr<SolutionArray>(
+            new SolutionArray(sol, size, meta));
+    }
+
+    /*!
+     *  Initialize SolutionArray with independent memory management
```
speth: What does "independent memory management" mean here?

include/cantera/base/SolutionArray.h
```diff
@@ -0,0 +1,194 @@
...
+    SolutionArray(const shared_ptr<Solution>& sol,
+                  size_t size,
+                  const AnyMap& meta);
+
+public:
+    virtual ~SolutionArray() {}
+
+    static shared_ptr<SolutionArray> create(const shared_ptr<Solution>& sol,
+                                            size_t size=0,
+                                            const AnyMap& meta={})
```
speth, line 47: This factory function needs a docstring.

src/oneD/Domain1D.cpp
```diff
@@ -173,6 +183,11 @@ void Domain1D::restore(const AnyMap& state, double* soln, int loglevel)
     }
 }
 
+void Domain1D::restore(SolutionArray& arr, double* soln, int loglevel)
+{
+    throw CanteraError("Domain1D::restore", "Needs to be overloaded.");
```
speth: Should throw `NotImplementedError`. I'd also tend to put these non-implementations in the header file, but I guess we're not that consistent about that.

include/cantera/base/SolutionArray.h
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /*!
+     *  Retrieve associated ThermoPhase object
+     */
+    std::shared_ptr<ThermoPhase> thermo();
+
+    /*!
+     *  Check whether SolutionArray contains a component.
+     */
```
speth, line 68: Can you define what a "component" is?

include/cantera/base/SolutionArray.h
```diff
@@ -0,0 +1,194 @@
...
+    /*!
+     *  Update the buffered index used to access entries.
+     */
+    void setIndex(size_t index, bool restore=true);
+
+    /*!
+     *  Retrieve the state vector for a single entry. If index is valid, it is updated;
+     *  otherwise, the last previously used index is referenced.
+     */
+    vector_fp getState(size_t index=npos);
```
speth: This interface of having a "current" index that is used by default seems a bit surprising, and it doesn't really seem to be significantly used anywhere. Would it not be simpler to just always specify the index? (in which case, I'd suggest that the index argument comes first in `setState`)

test/python/test_onedim.py
```diff
@@ -517,7 +517,20 @@ def test_prune(self):
         # TODO: check that the solution is actually correct (that is, that the
         # residual satisfies the error tolerances) on the new grid.
 
-    def test_save_restore_yaml(self):
+    def test_save_restore_yaml_legacy(self):
+        self.run_save_restore("legacy")
+
+    def test_save_restore_yaml_transition(self):
+        self.run_save_restore("transition")
```
speth: Can you add a few comments to define what these different modes mean?

samples/python/onedim/adiabatic_flame.py
```diff
@@ -2,13 +2,21 @@
...
 
+from pathlib import Path
 import cantera as ct
 
+
+if "native" in ct.hdf_support():
+    output = Path() / "adiabatic_flame.h5"
+else:
+    output = Path() / "adiabatic_flame.yaml"
+output.unlink(missing_ok=True)
```
speth: It might make sense to put this logic around picking the output file name just before the first place that `output` is used, rather than before any of the input parameters.

samples/python/onedim/flame_initial_guess.py
```diff
@@ -53,25 +53,20 @@ def describe(flame):
...
-    f.write_hdf(
-        hdf_filepath,
-        group="freeflame",
-        mode="w",
-        quiet=False,
-        description=("Initial methane flame"),
-    )
+    hdf_filepath = output_path / "flame.h5"
+    hdf_filepath.unlink(missing_ok=True)
+    f.save(hdf_filepath, name="freeflame", description=("Initial methane flame"))
```
speth: Parentheses around `"Initial methane flame"` are unnecessary.

include/cantera/oneD/Domain1D.h
```diff
@@ -315,6 +316,14 @@ class Domain1D
      */
     virtual AnyMap serialize(const double* soln) const;
 
+    //! Save the state of this domain as a SolutionArray
+    /*!
+     * @param soln local solution vector for this domain
+     *
+     * @since  New in Cantera 3.0.
+     */
+    virtual std::shared_ptr<SolutionArray> asArray(const double* soln) const;
```
speth: The `std::` prefix is unnecessary for `shared_ptr` since we have `using std::shared_ptr` in `ct_defs.h`.

include/cantera/oneD/Sim1D.h
```diff
@@ -117,9 +117,17 @@ class Sim1D : public OneDim
...
      */
     void save(const std::string& fname, const std::string& id,
-              const std::string& desc, int loglevel=1);
+              const std::string& desc, int loglevel=1, int compression=0);
+
+    /**
+     * Save the current solution to YAML (legacy implementation). @see save
+     * @since  New in Cantera 3.0.
+     */
+    void write_yaml(const std::string& fname, const std::string& id,
```
speth: If `save` is already able to generate a (nearly?) identical YAML output file, is there a reason to keep this implementation under a different name? Ditto on `read_yaml`.

SConstruct
```diff
@@ -356,6 +356,34 @@ config_options = [
            must include the shared version of the library, for example,
            'libfmt.so'.""",
         "default", ("default", "y", "n")),
+    EnumOption(
+        "hdf_support",
+        """Select whether to support HDF5 container files natively ('y'), disable HDF5
+           support ('n'), or to decide automatically based on the system configuration
+           ('default'). Native HDF5 support uses the headers-only HDF5 wrapper HighFive
+           (see option 'system_highfive'). Specifying 'hdf_include' or 'hdf_libdir'
```
speth: I think we should say that it uses the HDF5 library as well as the header-only HighFive C++ wrapper.

include/cantera/base/Storage.h
```diff
@@ -0,0 +1,519 @@
...
+{
+    m_file->flush();
+}
+
+bool Storage::checkGroupRead(const std::string& id) const
+{
+    std::vector<std::string> tokens;
+    tokenizePath(id, tokens);
+    std::string grp = tokens[0];
+    if (!m_file->exist(grp) || m_file->getObjectType(grp) != h5::ObjectType::Group) {
```
speth: This is a lot of code to put in a header file -- I think this all belongs in `Storage.cpp`.

Doing so would also let you move all the HighFive includes into the `.cpp` file, eliminating the need for users to have HighFive available when compiling C++ applications that use Cantera.

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,680 @@
...
+    for (auto& key : extra) {
+        m_other.emplace(key, std::make_shared<vector_fp>(m_size));
```
speth: Would it be more clear for this to just be `m_extra` for consistency?

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,680 @@
...
+{
+    if (!hasComponent(name)) {
+        throw CanteraError("SolutionArray::getComponent", "no component named " + name);
+    }
+
+    vector_fp out(m_size);
+    if (m_other.count(name)) {
+        // auxiliary data
+        auto other = m_other.at(name);
+        std::copy(other->begin(), other->end(), out.begin());
```
speth: I think you can do just `out = *m_other.at(name)`.

include/cantera/base/Storage.h
```diff
@@ -0,0 +1,519 @@
...
+        throw CanteraError("Storage::setCompressionLevel",
+                           "Invalid compression level '{}' (needs to be 0..9).", level);
+    }
+    m_compressionLevel = level;
+}
+
+void Storage::flush()
+{
+    m_file->flush();
+}
```
speth: Based on the way the `Storage` class is used, I'd suggest putting this into the destructor rather than requiring the user to call it. I'm kind of surprised that it isn't part of the destructor for the `HighFive::File` class.

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,680 @@
...
+    for (auto& field : tokens) {
+        path += "/" + field;
+        AnyMap& sub = *ptr;
+        if (sub.hasKey(field) && !sub[field].is<AnyMap>()) {
+            throw CanteraError("openField",
+                "Encountered invalid existing field '{}'", path);
+        } else if (!sub.hasKey(field)) {
+            sub[field] = AnyMap();
+        }
+        ptr = &sub[field].as<AnyMap>(); // AnyMap lacks 'operator=' for const AnyMap
```
speth: I don't understand the comment here -- this is just getting a reference, not assigning anything to an `AnyMap`. Plus, `AnyMap` definitely does support assignment, using the compiler-generated version of that operator.

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,680 @@
...
+    if (m_size == 1) {
+        setIndex(0);
+        data["temperature"] = phase->temperature();
+        data["pressure"] = phase->pressure();
+        auto surf = std::dynamic_pointer_cast<SurfPhase>(phase);
+        auto nSpecies = phase->nSpecies();
+        vector_fp values(nSpecies);
+        if (surf) {
+            surf->invalidateCache();
+            surf->getCoverages(&values[0]);
```
speth, line 298: Did you find a case where invalidating the cache is necessary? If so, I'm a bit worried and we should dig into this elsewhere.

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,680 @@
...
+            } else {
+                data[name] = getComponent(name);
+            }
+        }
+    }
+
+    // If this is not replacing an existing solution, put it at the end
+    if (!preexisting) {
+        data.setLoc(INT_MAX, 0);
+    }
```
speth, line 339: The default behavior if this is a new entry is already to put this after any existing entries, right?

src/oneD/Sim1D.cpp
```diff
@@ -151,8 +194,137 @@ void Sim1D::saveResidual(const std::string& fname, const std::string& id,
...
+        }
+        resize();
+        m_xlast_ts.clear();
+        for (auto dom : m_dom) {
+            dom->restore(*arrs[dom->id()], m_x.data() + dom->loc(), loglevel);
+        }
+        finalize();
+    } else {
+        throw CanteraError("Sim1D::restore",
+                           "Unknown file extension '{}'", extension);
```
speth: Since this method now accepts specific extensions only, we should probably list the known/valid ones here.

src/oneD/Sim1D.cpp
```diff
@@ -151,8 +194,137 @@ void Sim1D::saveResidual(const std::string& fname, const std::string& id,
     std::swap(res, m_x);
 }
 
+AnyMap legacyH5(shared_ptr<SolutionArray> arr, const AnyMap& header={})
+{
+    // convert data format used by Python h5py export (Cantera < 3.0)
```
speth: Putting this comment before the function signature, like a Doxygen-style docstring, will allow it to be parsed by various tools / editors.

src/oneD/Boundary1D.cpp
```diff
@@ -234,6 +235,24 @@ AnyMap Inlet1D::serialize(const double* soln) const
...
+std::shared_ptr<SolutionArray> Inlet1D::asArray(const double* soln) const
+{
+    AnyMap meta = Boundary1D::getMeta();
+    meta["type"] = "inlet";
+    meta["mass-flux"] = m_mdot;
+
+    // set gas state (using pressure from adjacent domain)
+    double pressure = m_flow->phase().pressure();
+    auto phase = m_solution->thermo();
+    phase->setState_TPY(m_temp, pressure, &m_yin[0]);
```
speth: Prefer `m_yin.data()` to `&m_yin[0]`.

include/cantera/base/SolutionArray.h
```diff
@@ -0,0 +1,194 @@
...
+     *  Initialize SolutionArray with independent memory management
+     *
+     *  @param extra Names of auxiliary data
+     */
+    void initialize(const std::vector<std::string>& extra={});
+
+    /*!
+     *  Size of SolutionArray (number of entries)
+     */
+    int size() const {
```
speth, line 53: If there's nothing else to write, I'd suggest the "short" form of the Doxygen comment, e.g.
```
    //! Size of SolutionArray (number of entries)
```

include/cantera/base/SolutionArray.h
```diff
@@ -0,0 +1,194 @@
...
+    /*!
+     *  Write header data to container file.
+     *
+     *  @param fname  Name of HDF container file
+     *  @param id  Identifier of SolutionArray root within the container file
+     *  @param desc  Description
+     */
+    static void writeHeader(const std::string& fname, const std::string& id,
+                            const std::string& desc);
+    static void writeHeader(AnyMap& root, const std::string& id,
```
speth, line 114: This second signature is sufficiently different (i.e, what is "root"?) that I think it needs its own docstring. Likewise for `writeEntry` and others.

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,680 @@
...
+    if (extension == "yaml" || extension == "yml") {
+        const AnyMap& root = AnyMap::fromYamlFile(fname);
+        readEntry(root, id);
+        return readHeader(root, id);
+    }
+    throw CanteraError("SolutionArray::restore",
+                        "Unknown file extension '{}'", extension);
+}
+
+std::string SolutionArray::detectMode(std::set<std::string> names, bool native)
```
speth: Can `names` be a const ref?

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,680 @@
...
+            break;
+        }
+    }
+    if (usesNativeState && native) {
+        return "native";
+    }
+    return mode;
+}
+
+std::set<std::string> SolutionArray::stateProperties(std::string mode, bool alias)
```
speth: `mode` should be a const ref?

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,680 @@
...
+{
+    if (!m_sol) {
+        throw CanteraError("SolutionArray::SolutionArray",
+            "Unable to create SolutionArray from invalid Solution object.");
+    }
+}
+
+void SolutionArray::initialize(const std::vector<std::string>& extra)
+{
+    m_stride = m_sol->thermo()->stateSize();
```
speth, line 51: Despite the name of the class, the `Solution` object here is only ever used to access the `ThermoPhase` object. Would it be simpler to just store the `shared_ptr<ThermoPhase>` instead, or do you anticipate extending this class such that access to kinetics and/or transport data would be of interest?

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,680 @@
...
+{
+    Storage file(fname, false);
+    file.checkGroup(id);
+    return file.readAttributes(id, false);
+}
+
+AnyMap SolutionArray::readHeader(const AnyMap& root, const std::string& id)
+{
+    throw CanteraError("SolutionArray::readHeader", "Not implemented.");
+}
```
speth, line 407: Are you anticipating subclassess of `SolutionArray`, or implementing this later? Otherwise, is this method just not needed?

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,680 @@
...
+        return readHeader(root, id);
+    }
+    throw CanteraError("SolutionArray::restore",
+                        "Unknown file extension '{}'", extension);
+}
+
+std::string SolutionArray::detectMode(std::set<std::string> names, bool native)
+{
+    // identify storage mode of state data
+    std::string mode = "";
```
speth, line 429: What is the "storage mode"? The algorithm below is a bit complex, so it's not that clear what the possibilities are and how this is determined.

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,680 @@
...
+        const auto& nativeState = m_sol->thermo()->nativeState();
+        for (const auto& item : sub) {
+            const std::string& name = item.first;
+            const AnyValue& value = item.second;
+            if (value.is<std::vector<double>>()) {
+                const vector_fp& data = value.as<std::vector<double>>();
+                if (data.size() == m_size) {
+                    setComponent(name, data, true);
+                    exclude.insert(item.first);
+                }
```
speth, line 642: This just ignores keys that specify a different number of data points? If the item needs to exist with a specific size, it might be useful to use `value.asVector<double>(m_size)` to check that requirement.

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,680 @@
...
+            const size_t offset_D = nativeState.find("D")->second;
+            const size_t offset_Y = nativeState.find("Y")->second;
+            for (size_t i = 0; i < m_size; i++) {
+                double T = m_data[offset_T + i * m_stride];
+                m_sol->thermo()->setState_TPY(T, P, &m_data[offset_Y + i * m_stride]);
+                m_data[offset_D + i * m_stride] = m_sol->thermo()->density();
+            }
+        } else if (missingProps.size()) {
+            throw CanteraError("SolutionArray::restore",
+                "Incomplete state information.");
```
speth: It would be nice to list the missing properties in the error message.

src/oneD/Boundary1D.cpp
```diff
@@ -257,6 +276,24 @@ void Inlet1D::restore(const AnyMap& state, double* soln, int loglevel)
...
+    m_temp = phase->temperature();
+    if (meta.hasKey("mass-flux")) {
+        m_mdot = meta.at("mass-flux").asDouble();
+    } else {
+        // convert data format used by Python h5py export (Cantera < 3.0)
+        auto aux = arr.getAuxiliary(0);
+        m_mdot = phase->density() * aux["velocity"];
+    }
+    auto Y = phase->massFractions();
+    std::copy(Y, Y + m_nsp, &m_yin[0]);
```
speth: The `massFractions` method is a weird wart that's best avoided in most cases. I'd recommend `phase->getMassFractions(m_yin.data())`.

src/oneD/Boundary1D.cpp
```diff
@@ -407,6 +460,14 @@ AnyMap Outlet1D::serialize(const double* soln) const
     return state;
 }
 
+std::shared_ptr<SolutionArray> Outlet1D::asArray(const double* soln) const
+{
+    AnyMap meta = Boundary1D::getMeta();
+    meta["type"] = "outlet";
+    auto arr = SolutionArray::create(m_solution, 0, meta);
+    return arr;
```
speth: You can just `return SolutionArray::create(...)` directly, here and in a few other places.

ext/SConscript
```diff
@@ -125,6 +125,14 @@ if not env['system_eigen']:
     copyenv.Depends(copyenv['config_h_target'], h)
     ext_copies.extend(h)
 
+if not env["system_highfive"]:
+    localenv = prep_default(env)
+    license_files["HighFive"] = File("#ext/HighFive/LICENSE")
+    h = build(copyenv.Command('#include/cantera/ext/HighFive', '#ext/HighFive/include/highfive',
+                              Copy('$TARGET', '$SOURCE')))
```
speth, line 132: As mentioned elsewhere, I think we may be able to avoid needing to install the HighFive headers.

SConstruct
```diff
@@ -1497,6 +1525,78 @@ else: # env['system_sundials'] == 'n'
     env['sundials_version'] = '5.3'
     env['has_sundials_lapack'] = int(env['use_lapack'])
 
+if env["hdf_include"]:
+    env["hdf_include"] = Path(env["hdf_include"]).as_posix()
+    env.Append(CPPPATH=[env["hdf_include"]])
+    env["hdf_support"] = "y"
+    env["extra_inc_dirs"].append(env["hdf_include"])
```
speth: I think there are two mutually exclusive options here. Either we introduce an `hdf_include` option and handle it independent of `extra_inc_dirs`, or we mention that it needs to be added to `extra_inc_dirs`. Otherwise, this would be redundant except for the fact that it's occuring after `extra_inc_dirs` has already been added to the SCons `CPPPATH`.

The same goes for `HDF_LIBDIR` and `extra_lib_dirs`.

.github/workflows/main.yml
```diff
@@ -79,16 +89,17 @@ jobs:
...
-        sudo apt install libboost-dev gfortran libomp-dev libomp5 libopenblas-dev
+        sudo apt install libboost-dev gfortran libomp-dev libomp5 libopenblas-dev libhdf5-dev
     - name: Upgrade pip
       run: python3 -m pip install -U pip setuptools wheel
     - name: Install Python dependencies
-      run: python3 -m pip install ruamel.yaml scons numpy cython h5py pandas pytest
-        pytest-github-actions-annotate-failures
+      run: |
+        python3 -m pip install ruamel.yaml scons numpy cython pandas pytest pytest-github-actions-annotate-failures
+        python3 -m pip install --no-binary=h5py h5py
```
speth: For the moment, shouldn't disabling the extensible reaction test make building h5py from source unnecessary?

In any case, I have a possible fix in mind for this interaction in mind -- would you mind if I push it to your branch to see if it works with the CI?

.github/workflows/main.yml
```diff
@@ -340,17 +360,26 @@ jobs:
...
         # (GCC 7.5.0 is both default and oldest supported version)
         run: python3 `which scons` build -j2 debug=n CC=gcc-9 CXX=g++-9
+        if: matrix.python-version != '3.10'
+      - name: Build Cantera (Python 3.10 with HDF)
+        # compile with GCC 9.4.0 on ubuntu-20.04 as an alternative to the default
+        # (GCC 7.5.0 is both default and oldest supported version)
+        run: |
+          python3 `which scons` build -j2 debug=n CC=gcc-9 CXX=g++-9 \
+          hdf_libdir=$HDF5_LIBDIR hdf_include=$HDF5_INCLUDEDIR
+        if: matrix.python-version == '3.10'
```
speth, line 381: Is the reason for this special case just to have one case that has HDF5 and one that doesn't? Either way, can you add a comment so it will be more clear what to do when updating this in the future?

speth (changes requested):
Thanks, @ischoegl. Overall, I think this is nice new feature. Certainly, it will be nice to have HDF5 output available even when not using the Python interface. I hope my comments below are useful in refining the interface defined by the new classes here.

include/cantera/oneD/Sim1D.h
ischoegl: This uses the old serialization that was implemented for YAML (the `save` route creates an intermediate `SolutionArray`), and is included to ensure that both options work and to maintain code coverage in the Python test suite. It is conceivable to remove the serialization approach completely and just store a pre-Cantera 3.0 YAML file to ensure backward compatibility. It is correct that formats are *nearly* identical.

include/cantera/base/SolutionArray.h
ischoegl: This is anticipating a connection to the Python API.  *Edit: adopting suggestion to make index mandatory anyways*

.github/workflows/main.yml
ischoegl: > In any case, I have a possible fix in mind for this interaction in mind -- would you mind if I push it to your branch to see if it works with the CI?

Not a problem whatsoever

src/base/SolutionArray.cpp
ischoegl: I wasn’t able to get this to compile. I may be overlooking something but attributed this to a lack of this operator for a `constant AnyMap`.

src/base/SolutionArray.cpp
ischoegl: As this can be tied to `SolutionArray` of other API’s, I’d like to keep this as is. *Edit: `SolutionArray::initialize` is gone*

src/base/SolutionArray.cpp
ischoegl: My thinking here is to return the header information when loading a `SolutionArray` once this is linked to Python. This API isn’t fleshed out yet, and will be added later.

*Edit: now implemented; `Sim1D::restore` now returns header information*

include/cantera/oneD/Sim1D.h
speth: I would say keeping a YAML file from Cantera 2.6 in `test/data` would be preferable to keeping this otherwise redundant code around.

src/base/SolutionArray.cpp
speth: But what's the alternative that you would want to do here? Since you want have `ptr` point to different `AnyMap` objects in different iterations of the loop, it needs to be a an `AnyMap*` rather than an `AnyMap&`. And since `sub[field].as<AnyMap>()` returns an `AnyMap&`, you need to take the address with the leading `&`. The assignment operator (`operator=`) would only be invoked if you wanted to copy the `AnyMap` object, which isn't the goal here.

ischoegl: Ok. I definitely wanted to avoid copying here; I looked through the header declarations and probably looked in the wrong spot.

include/cantera/base/SolutionArray.h
ischoegl: This refers to an earlier version of the code; shared memory access has since been removed.

src/base/SolutionArray.cpp
ischoegl: An early version of the implementation accessed memory of the `Sim1D` solution vector directly; this has since been removed. *Update: data structure is simplified*

src/base/SolutionArray.cpp
ischoegl: No. I believe this was copy-paste.

src/base/SolutionArray.cpp
ischoegl: This is cut and paste from the earlier implementation - my interpretation is the same as yours.

src/base/SolutionArray.cpp
ischoegl: A vector of different length is automatically added to meta data, so it is not ignored. ~There shouldn't be a check for this requirement, so I'll leave this as is.~ *Edit: I added a new field `components` that ensures that meta data and components can be differentiated; a check for size as suggested is added.*

.github/workflows/main.yml
ischoegl: Resolved by proposed fix.

src/base/SolutionArray.cpp
ischoegl: Alright - I tried again and the issue is actually a segfault. I'm leaving this for the time being.

SConstruct
ischoegl: This is actually not needed - it looks like I forgot to remove this after I introduced `hdf_include` and `hdf_libdir`.

ext/SConscript
ischoegl: I'm not sure that I am following: headers are required if Cantera is compiled from source (albeit only for HDF5 support in `Storage.cpp`)

include/cantera/oneD/Sim1D.h
ischoegl: OK. I removed redundant code.

src/base/SolutionArray.cpp
ischoegl: OK. On further inspection, here is the issue: defining `const AnyMap& sub`, there would need for a `sub = sub[field].as<AnyMap>()`, which causes the following error:
```
src/base/SolutionArray.cpp:394:13: error: no viable overloaded '='
        sub = sub[field].as<AnyMap>();
        ~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~
include/cantera/base/AnyMap.h:399:7: note: candidate function (the implicit copy assignment operator) not viable: 'this' argument has type 'const Cantera::AnyMap', but method is not marked const
class AnyMap : public AnyBase
      ^
include/cantera/base/AnyMap.h:399:7: note: candidate function (the implicit move assignment operator) not viable: 'this' argument has type 'const Cantera::AnyMap', but method is not marked const
```
The raw pointers side-step this issue.

ischoegl:
@speth - thank you for the review. I believe I addressed almost everything (there are some minor points where it would be good if you could clarify). Also, thanks for pushing the fix for the `libhdf5` clashes!

src/base/SolutionArray.cpp
speth: Right, the error message here is kind of confusing, but the thing to remember is that assigning to a reference type is asking to change the contents of the object being referred to. Once you create a reference, there is no way to have it refer to a different object, hence the need for pointers here.

ext/SConscript
speth: If we have `#include`s of HighFive only from `.cpp` files (as is currently the case), then those header files won't be needed to compile code that `#include`s Cantera's public header files. But I guess installing these with Cantera is mostly harmless.

include/cantera/base/SolutionArray.h
```diff
@@ -0,0 +1,209 @@
...
+     *  @param name  Name of component (property defining state or auxiliary variable)
+     *  @param data  Component data
+     *  @param force  If true, add new component to SolutionArray
+     */
+    void setComponent(const std::string& name, const vector_fp& data, bool force=false);
+
+    /*!
+     *  Update the buffered index used to access entries.
+     */
+    void setIndex(size_t index, bool restore=true);
```
speth, line 86: Since the index is now provided when calling `getState` and `getAuxillary`, I think the `setIndex` method could be made `protected`. The only remaining public use is before calls to `thermo()`, where that method could be modified to have the index as an optional argument (for cases where which state the thermo object is set to don't matter).

samples/python/onedim/adiabatic_flame.py
speth: Sorry if this wasn't clear, but I meant this comment to apply across most of the examples.

src/base/SolutionArray.cpp
speth: This bit of code is still present.

include/cantera/base/Storage.h
```diff
@@ -0,0 +1,96 @@
...
+     *  which involves considerable overhead for metadata).
+     */
+    void setCompressionLevel(int level);
+
+    //! Check whether path go location exists
+    //! If the file has write access, create location if necessary
+    bool checkGroup(const std::string& id);
+
+    //! Retrieve contents of file from a specified location
+    std::pair<size_t, std::set<std::string>> contents(const std::string& id) const;
```
speth: The structure returned by this method needs to be documented.

include/cantera/oneD/StFlow.h
```diff
@@ -56,7 +56,8 @@ class StFlow : public Domain1D
         StFlow(th.get(), nsp, points) {
     }
 
-    StFlow(shared_ptr<Solution> sol, size_t nsp = 1, size_t points = 1);
+    StFlow(shared_ptr<Solution> sol, const std::string& id="",
+           size_t nsp=1, size_t points=1);
```
speth: This constructor should have a docstring. Likewise for `IonFlow`.

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -1495,8 +1497,10 @@ cdef class Sim1D:
 
         >>> s.restore(filename='save.yaml', name='energy_off')
         """
-        self.sim.restore(stringify(str(filename)), stringify(name), loglevel)
+        cdef CxxAnyMap header
+        header = self.sim.restore(stringify(str(filename)), stringify(name), loglevel)
         self._initialized = True
+        return anymap_to_dict(header)
```
speth, line 1508: The docstring should mention this new return value (and note that this is "since Cantera 3.0")

samples/cxx/flamespeed/flamespeed.cpp
```diff
@@ -7,6 +7,7 @@
  *
  * Usage: flamespeed [equivalence_ratio] [refine_grid] [loglevel]
  *
+ * Requires: cantera >= 3.0
  * Keywords: combustion, 1D flow, premixed flame, flame speed
```
speth: You might want to add the "saving output" keyword here.

samples/cxx/flamespeed/flamespeed.cpp
```diff
@@ -113,6 +115,21 @@ int flamespeed(double phi, bool refine_grid, int loglevel)
...
 
+        // Save initial guess to container file
+
+        // Solution is saved in HDF5 or YAML file format
+        std::string fileName;
+        if (usesHDF5()) {
+            // Cantera is compiled with native HDF5 support
+            fileName = "flamespeed.h5";
+        } else {
+            fileName = "flamespeed.yml";
```
speth: We tend to use the full `.yaml` extension pretty much everywhere else.

samples/cxx/flamespeed/flamespeed.cpp
```diff
@@ -16,6 +17,7 @@
 #include "cantera/onedim.h"
 #include "cantera/base/stringUtils.h"
 #include <fstream>
+#include <cstdio>
```
speth: I don't think there's anything here that requires the `cstdio` header.

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,687 @@
...
+#include <set>
+#include <fstream>
+
+
+namespace ba = boost::algorithm;
+
+namespace Cantera
+{
+
+const std::map<std::string, std::string> aliasMap = {
```
speth, line 22: Variables/functions that are intended to have file scope should be put in a anonymous (unnamed) namespace.

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,687 @@
...
+        writeHeader(data, id, desc);
+
+        // Write the output file and remove the now-outdated cached file
+        std::ofstream out(fname);
+        out << data.toYamlString();
+        AnyMap::clearCachedFile(fname);
+        return;
+    }
+    throw CanteraError("SolutionArray::writeHeader",
+                       "Unknown file extension '{}'", extension);
```
speth, line 368: This error is being reported from `SolutionArray::save`.

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,687 @@
...
+    std::vector<std::string> tokens;
+    tokenizePath(id, tokens);
+    const AnyMap* ptr = &root; // use raw pointer to avoid copying
+    std::string path = "";
+    for (auto& field : tokens) {
+        path += "/" + field;
+        const AnyMap& sub = *ptr;
+        if (!sub.hasKey(field) || !sub[field].is<AnyMap>()) {
+            throw CanteraError("SolutionArray::restore",
+                "No field or solution with id '{}'", path);
```
speth, line 390: Method name should be `locateField`.

src/base/SolutionArray.cpp
```diff
@@ -0,0 +1,687 @@
...
+        auto Y = file.readMatrix(id, "Y", m_size, nSpecies);
+        for (size_t i = 0; i < m_size; i++) {
+            m_sol->thermo()->setState_TPY(T[i], P[i], Y[i].data());
+            m_sol->thermo()->saveState(nState, &m_data[i * m_stride]);
+        }
+    } else if (mode == "") {
+        throw CanteraError("SolutionArray::restore",
+            "Data are not consistent with full state modes.");
+    } else {
+        throw NotImplementedError("SolutionArray::restore",
```
speth: These error messages are arising from `SolutionArray::readEntry`. Similarly, there are several more of these in the implementation of `SolutionArray::readEntry(const AnyMap& root, const std::string& id)`.

src/base/Storage.cpp
```diff
@@ -0,0 +1,466 @@
...
+}
+
+std::vector<vector_fp> Storage::readMatrix(const std::string& id,
+                                           const std::string& name,
+                                           size_t rows, size_t cols) const
+{
+    h5::Group sub = m_file->getGroup(id);
+    if (!sub.exist(name)) {
+        throw CanteraError("Storage::readVector",
+            "DataSet '{}' not found in group '{}'.", name, id);
```
speth, line 341: Should be `Storage::readMatrix`

speth (changes requested):
Thanks, @ischoegl, I think these updates look good. I found a few more minor issues on a second pass. I somehow messed up using the GH interface, so there are a few comments that are in the preceding review.

One additional item: I think what ends up in `cantera.pc` needs to be double-checked. I tried this on Ubuntu using the paths from the GitHub Actions config, and I think both the library and include directories are missing.

include/cantera/base/SolutionArray.h
ischoegl: I agree that within the currently implemented interface there isn't much need for a `public` `setIndex` - the part where it is most useful - picking states within restored `SolutionArray`s from the Python API so that they can be further analyzed - is not implemented at the moment. As such, I'd like to leave the interface as is.

src/base/SolutionArray.cpp
ischoegl: Based on the review comment, I believe the ‘pointer’ implementation is good. I removed the notes that this needs to be improved.

ext/SConscript
ischoegl: Based on the review comment, I believe it's ok to leave things as they are.

ischoegl:
> Thanks, @ischoegl, I think these updates look good. I found a few more minor issues on a second pass. I somehow messed up using the GH interface, so there are a few comments that are in the preceding review.

@speth … thanks for the additional checks - I believe everything is addressed now.

> One additional item: I think what ends up in `cantera.pc` needs to be double-checked. I tried this on Ubuntu using the paths from the GitHub Actions config, and I think both the library and include directories are missing.

Until this came up I was actually unaware of `cantera.pc` - ~I’m not sure how to address this as I’m not familiar with any of the details (sorry).~ *Edit: Not too hard to figure out after all.*

As an aside, on my system (macOS), the `pkg-config` route does not work, with `cantera.pc` being:
```
prefix=/opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev
exec_prefix=${prefix}/bin
libdir=${prefix}/lib
includedir=${prefix}/include

Name: Cantera
Description: Cantera library
URL: https://cantera.org
Version: 3.0.0a3

Libs: -L${libdir} -L/opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/lib -L/opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/lib/python3.10/config-3.10-darwin -L/opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/lib -lcantera -lsundials_cvodes -lsundials_ida -lsundials_nvecserial -lsundials_sunlinsollapackdense -lsundials_sunlinsollapackband -lhdf5 -lfmt -lyaml-cpp -lpython3.10 -ldl
Cflags: -std=c++14 -framework Accelerate -I${includedir} -I/opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/include -I
```

An attempt to compile the [very simple C++ program](https://cantera.org/tutorials/cxx-guide/simple-example.html) fails with:
```
% export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/lib
% export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/lib/pkgconfig
% g++ simple.cpp -o myProgram $(pkg-config --cflags --libs cantera)
% ./myProgram
dyld[92219]: Library not loaded: @rpath/libsundials_cvodes.6.dylib
  Referenced from: <E62AB15F-607D-3047-996E-A59885DCC71C> /Volumes/Data/work/GitHub/cantera/myProgram
  Reason: tried: '/System/Volumes/Preboot/Cryptexes/OS@rpath/libsundials_cvodes.6.dylib' (no such file), '/usr/local/lib/libsundials_cvodes.6.dylib' (no such file), '/usr/lib/libsundials_cvodes.6.dylib' (no such file, not in dyld cache)
zsh: abort      ./myProgram
```

As I get the same failure for the `main` branch, I cannot further investigate and/or verify that my updates are correct (I switched to macOS relatively recently, and may not be aware of all differences yet - this may just not work; - it may also be an issue with compiling under a conda environment).

src/base/SolutionArray.cpp
ischoegl: This bit of code is the same as what you had originally implemented. Do you want to change this?

platform/posix/SConscript
```diff
@@ -58,6 +58,14 @@ if localenv["boost_inc_dir"] and not localenv["package_build"]:
 else:
     localenv['mak_boost_include'] = ''
 
+if localenv["use_hdf5"] and not localenv["package_build"]:
+    localenv["mak_hdf_include"] = f"-I{localenv['hdf_include']}"
+    pc_incdirs.append(localenv["hdf_include"])
+    localenv["mak_hdf_libs"] = f"-L{localenv['hdf_libdir']} -lhdf5"
+else:
+    localenv["mak_hdf_include"] = ""
+    localenv["mak_hdf_libs"] = "lhdf5"
```
speth: This should be `-lhdf5`, right?

speth (changes requested):
Thanks, @ischoegl, this all looks good to me. There is just one very minor issue with the `.mak` file generation, I think.

As far as the `cantera.pc`, I guess that requires getting the right `rpath` flags in when linking to shared libraries that live in non-system locations, such as a Conda install. I'd suggest we just ignore that here, and you can create a new issue to remind us that needs to be fixed at some point.

platform/posix/SConscript
ischoegl: Correct!

ischoegl:
> Thanks, @ischoegl, this all looks good to me. There is just one very minor issue with the `.mak` file generation, I think.

Thanks for catching - fixed!

> As far as the `cantera.pc`, I guess that requires getting the right `rpath` flags in when linking to shared libraries that live in non-system locations, such as a Conda install. I'd suggest we just ignore that here, and you can create a new issue to remind us that needs to be fixed at some point.

 See #1422.

speth (approved):

PR #1396, opened 2022-09-26
Accept floating point representations of valid integer geometry flags in ck2yaml

### LINKED ISSUE #1395: Floating point representations of valid integer geometry flags crash ck2yaml
**Problem description**

`c2yaml` crashes when the transport geometry flag is a float representation of a valid integer flag (e.g. `0.000`, `1.000`, `2.000`). 

**Expected Behavior**

This has been discussed previously in #446, but rather than reviving that issue I decided to create a new one. The resolution in #454 was that the user is responsible for fixing floats instead of silently fixing the issue. Since the point of `ck2yaml` is to convert existing CHEMKIN mechanisms to a Cantera format, I would argue that float representations of valid integer flags should be automatically accepted as they are valid CHEMKIN inputs. Manually fixing the geometry flags for mechanisms such as AramcoMech 3.0 (as distributed by [NUIGalway](https://c3.universityofgalway.ie/combustionchemistrycentre/mechanismdownloads/)) with hundreds of species is unnecessarily tedious.

**Version**

Cantera 2.6.0a3

corykinney:

This pull request automatically converts floating point representations of valid integer geometry flags to integers by casting to a `float` then to an `int` and checking equality between the two results. 

Closes #1395 

**Checklist**


bryanwweber:
Thanks for this @cory-kinney! To be honest, I feel quite ambivalent about this change... on the one hand, this change seems sensible and some simple testing on my end shows that it should be pretty robust. It would certainly improve the user experience.

On the other hand, the CHEMKIN spec (as much as can be found publicly) states that this value is an index, which I interpret to mean an integer. I think that's the most reasonable interpretation. In Python, indices to access elements in sequences need to be ints, for example. As CHEMKIN is closed source, we can only go by the published materials that are available when implementing this. Specifically, I feel quite reluctant to base Cantera's behavior on CHEMKIN's current (inconsistent in my opinion) behavior, as they can change behavior without warning or recourse from our side. In this specific case, it seems like they'd be unlikely to cause such a backwards-compatibility-breaking change for a fairly minor inconsistency, but the principle remains...

In the end, I think this change is probably worth it. I'll just ask you to fix up the failing test which explicitly checks for this error, and change it to a test that ensures this behavior is robust. The case I'd be specifically worried about is automatic writers that format something that should be 2.0 as just slightly less than 2.0 (likewise for 1.0)... Python's default behavior with `int` is to floor the input, that is:

```
In [45]: int(float("2.0") - sys.float_info.epsilon)
Out[45]: 1
```

That is, if some software writes out a non-linear molecule with a geometry of 1.999999999999999999, it will be turned into a linear molecule with this change. I'd rather error in that case, than produce incorrect results in Cantera.

---

On the less productive side, I also feel like this is another chance for Cantera to point out where CHEMKIN is either incorrect or (in this case) inconsistent with its own documentation. So I'm also inclined to keep this as an error, if only to point out to users that CHEMKIN should _not_ be the de-facto baseline. But, as I said, that's not super productive from a user standpoint 😄

ischoegl:
> I also feel like this is another chance for Cantera to point out where CHEMKIN is either incorrect or (in this case) inconsistent with its own documentation. So I'm also inclined to keep this as an error, if only to point out to users that CHEMKIN should not be the de-facto baseline. 

I believe the best way forward would be to issue warnings instead of errors, and convert for the user for cases like the one that is reported. Silently accepting incorrect syntax is not ideal.

interfaces/cython/cantera/ck2yaml.py
```diff
@@ -691,13 +691,14 @@ def __init__(self, label, geometry, well_depth, collision_diameter,
...
-                "Bad geometry flag '{}' for species '{}', is the flag a float "
-                "or character? It should be an integer.", geometry, label)
-        if geometry not in (0, 1, 2):
-            raise InputError("Bad geometry flag '{}' for species '{}'.", geometry, label)
+            if float(geometry) == int(float(geometry)):
+                geometry = int(float(geometry))
+            else:
+                raise ValueError("Flag should be an integer.")
+            if geometry not in (0, 1, 2):
+                raise ValueError("Flag value should be 0, 1, or 2.")
```
ischoegl: The old `InputError` should remain unchanged (outside of the `try` .. `except` block), as it will show the section where the error is raised.

interfaces/cython/cantera/ck2yaml.py
```diff
@@ -691,13 +691,14 @@ def __init__(self, label, geometry, well_depth, collision_diameter,
...
         try:
-            geometry = int(geometry)
-        except ValueError:
-            raise InputError(
-                "Bad geometry flag '{}' for species '{}', is the flag a float "
-                "or character? It should be an integer.", geometry, label)
-        if geometry not in (0, 1, 2):
-            raise InputError("Bad geometry flag '{}' for species '{}'.", geometry, label)
+            if float(geometry) == int(float(geometry)):
+                geometry = int(float(geometry))
```
ischoegl: The original `int` conversion should remain in the `try` branch, with the current check being moved to the `except` branch. A warning should be issued so that the automatic conversion/fix does not remain silent.

ischoegl (dismissed):
@cory-kinney ... thanks for the PR! As mentioned, I am not opposed to being more permissive when it comes to CHEMKIN input, but would request some changes to your approach.

interfaces/cython/cantera/ck2yaml.py
bryanwweber: Just to make sure I'm thinking the same thing, this would look like:

```python
try:
    geometry = int(geometry)
except ValueError:
    try:
        geometry = float(geometry)
    except ValueError:
        raise InputError(...) from None
    if geometry == int(geometry):
        warn(...)
        geometry = int(geometry)
if geometry not in (0, 1, 2):
     raise InputError(...)
```

The nested `try/except` is needed in case the original value of `geometry` is a nonsense string because of a mis-formatted file... `float(geometry)` is not guaranteed to succeed.

corykinney: I thought about doing the nested `try`/`except` block and keeping the `int` conversion in the original `try` branch, but I supposed that anything that failed `int(geometry)` would also fail `float(geometry)` (in the case of a random string) since the former would be stricter anyway. I also thought it would be interesting to jsut `raise InputError(...) from` whatever `ValueError` occurred, whether the two that were explicitly raised or the one that might occur from `float(geometry)`. 

But I'm more than happy to change it into form suggested! I don't know how pull requests normally work, so do I just make the changes on my branch, make a new commit, and then include them in the pull request somehow?

And I certainly agree on the warning, rather than silently fixing it. Is there an existing custom warning type that should be used for input syntax issues, or is the standard `warnings.warn` preferred?

bryanwweber: > I don't know how pull requests normally work, so do I just make the changes on my branch, make a new commit, and then include them in the pull request somehow?

Yes, make a new commit then `git push` to the same branch. That will automatically update this pull request with the new commits.

> And I certainly agree on the warning, rather than silently fixing it. Is there an existing custom warning type that should be used for input syntax issues, or is the standard warnings.warn preferred?

I think the best place to look would be to do whatever is done for duplicate thermo entries. At the least, this warning would be behind the `--permissive` flag, and an error if `--permissive` is not given.

ischoegl: @cory-kinney … please let us know in case you need any pointers.

interfaces/cython/cantera/ck2yaml.py
```diff
@@ -687,17 +687,26 @@ def reduce(self, output):
...
+                geometry = float(geometry)
+            except ValueError:
+                raise InputError(
+                    "Bad geometry flag '{}' for species '{}'. "
+                    "Flag should be an integer.", geometry, label) from None
+            if geometry == int(geometry):
+                geometry = int(geometry)
+                parser.warn("Incorrect syntax for geometry flag for species {0}. "
+                            "If --permissive was given, the flag was automatically "
+                            "converted to an integer.".format(label))
```
bryanwweber: There needs to be an `else` here that throws an `InputError` if `float != int` of the geometry string.

test/data/h2o2-float-geometry-tran.dat
```diff
@@ -1,7 +1,7 @@
 AR                 0   136.500     3.330     0.000     0.000     0.000
-H                  0.00   145.000     2.050     0.000     0.000     0.000
-H2                 1.00    38.000     2.920     0.000     0.790   280.000
-H2O                2.00   572.400     2.605     1.844     0.000     4.000
+H                  0   145.000     2.050     0.000     0.000     0.000
+H2                 0.999999999999999999    38.000     2.920     0.000     0.790   280.000
+H2O                1.999999999999999999   572.400     2.605     1.844     0.000     4.000
```
bryanwweber: I see you adjusted these so that we're testing the new behavior, but I think the tests are still going to fail because you need to adjust the code over here: https://github.com/Cantera/cantera/blob/332bb9b9d970c6ead5160f428ff49976fb2947fa/test/python/test_convert.py#L340

bryanwweber (changes requested):
Thanks @cory-kinney! A few more changes based on the updates

interfaces/cython/cantera/ck2yaml.py
corykinney: @ischoegl ... sorry, was a bit delayed by the hurricane but I believe I figured everything out 😅

@bryanwweber I changed the test case for floats (since that is now accepted behavior) to verify that it fails for floating point rounding errors instead (e.g. `1.999999999999999999`)

Let me know if there's anything else that needs to be changed or added.

bryanwweber: @cory-kinney see below :point_down: I think you need to change the test code as well to make sure that things are failing and succeeding when they're supposed to.

interfaces/cython/cantera/ck2yaml.py
corykinney: Ahh yes, I'm not sure how I skipped over that.

test/data/h2o2-float-geometry-tran.dat
corykinney: @bryanwweber what changes would need to be made? Wouldn't the existing test case fail since it would raise an `InputError` with the phrase "geometry flag" in it since `--permissive` won't apply?

bryanwweber: We also need a test where --permissive is passed and the conversion is successful. There should also be a case testing passing --permissive and the check fails because of the rounding issue.

The existing test is only checking one of the code paths, as you note the one without --permissive

ischoegl: @cory-kinney ... I think the only remaining task preventing this from being approved/merged is the test @bryanwweber mentioned?

corykinney: @ischoegl That's correct, I've just been a bit busy so I haven't gotten around to it! I'll try to do that sometime this next week.

corykinney:
It would seem that the additional tests implemented originally failed because, for the original value chosen to test handling of floating point arithmetic error, `float(0.999999999999999999)` evaluates to `1.0` so the error was not being raised; therefore, the value with the fewest number of digits after the decimal place that did not evaluate to `1.0` was chosen (`0.9999999999999999`) so that the error would be raised and the test would pass. Does that satisfy the concern about accidental floors @bryanwweber?

Let me know if there are any other changes that should be made. Should I go ahead and squash my commits and write a proper commit message now?

corykinney:
I believe it's ready for review now @ischoegl !

interfaces/cython/cantera/ck2yaml.py
```diff
@@ -687,17 +687,30 @@ def reduce(self, output):
...
-            raise InputError(
-                "Bad geometry flag '{}' for species '{}', is the flag a float "
-                "or character? It should be an integer.", geometry, label)
+            try:
+                geometry = float(geometry)
+            except ValueError:
+                raise InputError(
+                    "Bad geometry flag '{}' for species '{}'. "
+                    "Flag should be an integer.", geometry, label) from None
+            if geometry == int(geometry):
```
bryanwweber: It looks like there isn't a test of this case, which could be triggered by putting something that's not a number into the transport field. I'm basing this on the coverage comment here, although I do see what looks like it should be the right test below, `test_transport_bad_geometry`.

I'd suggest updating the error messages to be slightly different so that we can actually check for them appropriately and make sure that we have tests that cover all the cases:

1. Invalid input (character instead of number)
2. Float which rounds to 0, 1, or 2 and is only an error if `permissive=False`, and succeeds otherwise
3. Integer 0, 1, or 2 (this should already be well covered)
4. Integer or float >=3 which fails

test/python/test_convert.py
```diff
@@ -343,6 +343,15 @@ def test_transport_float_geometry(self):
                 transport='h2o2-float-geometry-tran.dat',
                 output='h2o2_transport_float_geometry')
 
+        self.convert('h2o2.inp',
+            transport='h2o2-float-geometry-tran.dat',
+            output='h2o2_transport_float_geometry', permissive=True)
```
bryanwweber: This test should also check that the converted mechanism has the appropriate type for the geometry. Otherwise we're not sure what the value is actually is.

bryanwweber (commented):
Thanks @cory-kinney, I have a few more suggested updates here to make sure that the tests provide appropriate coverage.

By the way, the failing tests here seem to be failing on the `main` branch too, so you don't need to worry about that.

ischoegl:
> I believe it's ready for review now @ischoegl !

I believe that my own initial review comments have been addressed a while ago. I know that @bryanwweber has things under control here, so I am happy with where this is going.

corykinney:
@bryanwweber I finally got around to implementing the changes you requested (I haven't had to convert a mechanism with this issue in a while, so it was on the back burner), let me know if the latest commit properly addresses them.

corykinney:
I'm having some trouble verifying the tests work properly now because of a build error that has started occurring recently for unknown reasons. 

```
In file included from C:\Users\Cory\AppData\Local\Programs\Python\Python311\Include/Python.h:94,
                 from src\extensions\pythonExtensions.h:6,
                 from src\extensions\PythonExtensionManager.cpp:10:
C:\Users\Cory\AppData\Local\Programs\Python\Python311\Include/pylifecycle.h:37:38: note: declared here
   37 | Py_DEPRECATED(3.11) PyAPI_FUNC(void) Py_SetProgramName(const wchar_t *);
      |                                      ^~~~~~~~~~~~~~~~~
src\extensions\PythonExtensionManager.cpp:127:22: error: '_PyNamespace_New' was not declared in this scope
  127 |     PyObject *spec = _PyNamespace_New(attrs);
      |                      ^~~~~~~~~~~~~~~~
scons: *** [build\src\extensions\PythonExtensionManager.o] Error 1
```

bryanwweber:
@speth @ischoegl Can you guys help out with Corey's error above?

interfaces/cython/cantera/ck2yaml.py
```diff
@@ -687,17 +687,30 @@ def reduce(self, output):
...
             geometry = int(geometry)
         except ValueError:
-            raise InputError(
-                "Bad geometry flag '{}' for species '{}', is the flag a float "
-                "or character? It should be an integer.", geometry, label)
+            try:
+                geometry = float(geometry)
+            except ValueError:
+                raise InputError(
+                    "Invalid geometry flag entry for species '{}'. "
```
bryanwweber:
```suggestion
                    "Invalid geometry flag '{}' for species '{}'. "
```

bryanwweber: Without this change, the wrong thing will be printed in the other placeholder.

bryanwweber (commented):
One teeny tiny change 😄 Great work @corykinney I'd approve but it looks like you're having some trouble compiling, plus there's the failing .NET job which I'm not sure about.

corykinney:
@bryanwweber I was able to build it just fine on my laptop with Python 3.10, so I can now confirm that the test cases pass. I just had to update the check for the error you fixed, as it was looking for `Invalid geometry flag entry`. To distinguish from the other error messages without further modification, I just switched it to `Invalid geometry flag '` to catch the first single quote which isn't in the other three possible error messages.

corykinney:
> @speth @ischoegl Can you guys help out with Corey's error above?

Look like this was already fixed in f04d48222351785328fbc75542ebd73d80de1d46 in November, which was slightly after my latest rebase of this pull request.

ischoegl:
@corykinney ... the error you posted looks like it's related to #1382, but apparently went away after the fix you indicated. Given that CI passes (coverage failed for a different reason), this should be ready to merge. I'll leave the approval to @bryanwweber ... :grin:

bryanwweber:
@corykinney Sorry, it looks like you committed the `.idea` folder, can you remove that? If you want to avoid this in the future, you can add `.idea` to the `.gitignore`, I think PyCharm is a common enough editor that we can ignore the config for it globally.

bryanwweber (approved):
Thanks for all the work here. In it goes!

corykinney:
@bryanwweber Good catch! For whatever reason I've avoided modifying the `.gitignore` like I normally do for fear of accidentally committing that, but then I ended up committing the `.idea` folder anyway haha

bryanwweber:
If you want to ignore files just for yourself, you can add it to `.git/info/exclude` in the repo. That file has the same syntax as `.gitignore` but it only works for the computer you're on

corykinney:
> If you want to ignore files just for yourself, you can add it to `.git/info/exclude` in the repo. That file has the same syntax as `.gitignore` but it only works for the computer you're on

That's a super helpful tip!

And thank you both @bryanwweber @ischoegl for your patience and help in figuring out my first pull request! 🎉
Hopefully next is #1421!

PR #1398, opened 2022-09-27
interface_current method and accompanying test added

c-randall:

- Here, a new method is added for the interface kinetics class. The method `interface_current(phase)` is useful when charge transfer reactions occur at an interface. The method returns the net positive charge moving into phase `phase` via these reactions (Units: A/m^2).
- To test the new method, a new `test_interface_current` test has been added. Prior to the pull request, the package has been built and tested locally. All tests were passed successfully.
- At present, the new method is implemented in the cython/python folders. If there is interest in implementing this functionality via the full C++ code, please reach out to let me know.  

Closes #

A short python script is included below to demonstrate how to use this new method. It references the existing `lithium_ion_battery.yaml` file already available in the Cantera package. 

```
"""
Interface current: net positve current into a specified phase (A/m^2)
"""

import cantera as ct

file = 'lithium_ion_battery.yaml'

anode = ct.Solution(file,"anode")
elect = ct.Solution(file,"electron")
elyte = ct.Solution(file,"electrolyte")
anode_interface = ct.Interface(file,"edge_anode_electrolyte",[anode,elect,elyte])

anode.X = [0.9, 0.1]
elyte.X = [0.4, 0.3, 0.15, 0.15]

anode.electric_potential = 0.
elyte.electric_potential = 3.

int_current = anode_interface.interface_current(elyte)

print(int_current)
```

**Checklist**

### DIFF
--- a/include/cantera/kinetics/InterfaceKinetics.h
+++ b/include/cantera/kinetics/InterfaceKinetics.h
@@ -287,7 +287,7 @@ public:
      */
     int phaseStability(const size_t iphase) const;
 
-    //! Gets the interface current for the ith phaseExistence
+    //! Gets the interface current for the ith phase
     /*!
     * @param iphase Phase Id
     * @return The double specifying the interface current. The interface Current
@@ -296,7 +296,7 @@ public:
     *         specified by the Phase Id. (Units: A/m^2 for a surface reaction,
     *         A/m for an edge reaction).
     */
-    double InterfaceCurrent(const size_t iphase);
+    double interfaceCurrent(const size_t iphase);
 
 protected:
     //! Temporary work vector of length m_kk
--- a/include/cantera/kinetics/InterfaceKinetics.h
+++ b/include/cantera/kinetics/InterfaceKinetics.h
@@ -287,7 +287,16 @@ public:
      */
     int phaseStability(const size_t iphase) const;
 
-    double InterfaceCurrent(int iPhase);
+    //! Gets the interface current for the ith phaseExistence
+    /*!
+    * @param iphase Phase Id
+    * @return The double specifying the interface current. The interface Current
+    *         is useful when charge transfer reactions occur at an interface. It
+    *         is defined here as the net positive charge entering the phase
+    *         specified by the Phase Id. (Units: A/m^2 for a surface reaction,
+    *         A/m for an edge reaction).
+    */
+    double InterfaceCurrent(const size_t iphase);
 
 protected:
     //! Temporary work vector of length m_kk
--- a/include/cantera/kinetics/InterfaceKinetics.h
+++ b/include/cantera/kinetics/InterfaceKinetics.h
@@ -287,6 +287,8 @@ public:
      */
     int phaseStability(const size_t iphase) const;
 
+    double InterfaceCurrent(int iPhase);
+
 protected:
     //! Temporary work vector of length m_kk
     vector_fp m_grt;
--- a/interfaces/cython/cantera/kinetics.pxd
+++ b/interfaces/cython/cantera/kinetics.pxd
@@ -69,7 +69,7 @@ cdef extern from "cantera/kinetics/InterfaceKinetics.h":
     cdef cppclass CxxInterfaceKinetics "Cantera::InterfaceKinetics":
         void advanceCoverages(double, double, double, double, size_t, size_t) except +translate_exception
         void solvePseudoSteadyStateProblem() except +translate_exception
-        double InterfaceCurrent(int) except +translate_exception
+        double InterfaceCurrent(size_t) except +translate_exception
 
 
 cdef extern from "cantera/cython/kinetics_utils.h":
--- a/interfaces/cython/cantera/kinetics.pxd
+++ b/interfaces/cython/cantera/kinetics.pxd
@@ -69,6 +69,7 @@ cdef extern from "cantera/kinetics/InterfaceKinetics.h":
     cdef cppclass CxxInterfaceKinetics "Cantera::InterfaceKinetics":
         void advanceCoverages(double, double, double, double, size_t, size_t) except +translate_exception
         void solvePseudoSteadyStateProblem() except +translate_exception
+        double InterfaceCurrent(int) except +translate_exception
 
 
 cdef extern from "cantera/cython/kinetics_utils.h":
--- a/interfaces/cython/cantera/kinetics.pxd
+++ b/interfaces/cython/cantera/kinetics.pxd
@@ -69,7 +69,7 @@ cdef extern from "cantera/kinetics/InterfaceKinetics.h":
     cdef cppclass CxxInterfaceKinetics "Cantera::InterfaceKinetics":
         void advanceCoverages(double, double, double, double, size_t, size_t) except +translate_exception
         void solvePseudoSteadyStateProblem() except +translate_exception
-        double InterfaceCurrent(size_t) except +translate_exception
+        double interfaceCurrent(size_t) except +translate_exception
 
 
 cdef extern from "cantera/cython/kinetics_utils.h":
--- a/interfaces/cython/cantera/kinetics.pyx
+++ b/interfaces/cython/cantera/kinetics.pyx
@@ -6,6 +6,7 @@ import warnings
 cimport numpy as np
 import numpy as np
 
+from .constants import *
 from .reaction cimport *
 from ._utils cimport *
 from . import _utils
@@ -879,6 +880,14 @@ cdef class InterfaceKinetics(Kinetics):
         species in all phases.
         """
         return self.net_production_rates[self._phase_slice(phase)]
+
+    def interface_current(self, phase):
+        """
+        The interface current is useful when charge transfer reactions occur at
+        an interface. It is defined here as the net positive charge entering the
+        phase ``phase`` (Units: A/m^2).
+        """
+        return sum(self.get_net_production_rates(phase)*phase.charges)*faraday
 
     def write_yaml(self, filename, phases=None, units=None, precision=None,
                    skip_user_defined=None):
--- a/interfaces/cython/cantera/kinetics.pyx
+++ b/interfaces/cython/cantera/kinetics.pyx
@@ -886,8 +886,8 @@ cdef class InterfaceKinetics(Kinetics):
         an interface. It is defined here as the net positive charge entering the
         phase ``phase`` (Units: A/m^2 for a surface, A/m for an edge reaction).
         """
-        iPhase = self.phase_index(phase)
-        return (<CxxInterfaceKinetics*>self.kinetics).InterfaceCurrent(iPhase)
+        i_phase = self.phase_index(phase)
+        return (<CxxInterfaceKinetics*>self.kinetics).InterfaceCurrent(i_phase)
 
     def write_yaml(self, filename, phases=None, units=None, precision=None,
                    skip_user_defined=None):
--- a/interfaces/cython/cantera/kinetics.pyx
+++ b/interfaces/cython/cantera/kinetics.pyx
@@ -885,9 +885,10 @@ cdef class InterfaceKinetics(Kinetics):
         """
         The interface current is useful when charge transfer reactions occur at
         an interface. It is defined here as the net positive charge entering the
-        phase ``phase`` (Units: A/m^2).
+        phase ``phase`` (Units: A/m^2 for a surface, A/m for an edge reaction).
         """
-        return sum(self.get_net_production_rates(phase)*phase.charges)*faraday
+        iPhase = self.phase_index(phase)
+        return (<CxxInterfaceKinetics*>self.kinetics).InterfaceCurrent(iPhase)
 
     def write_yaml(self, filename, phases=None, units=None, precision=None,
                    skip_user_defined=None):
--- a/interfaces/cython/cantera/kinetics.pyx
+++ b/interfaces/cython/cantera/kinetics.pyx
@@ -6,7 +6,6 @@ import warnings
 cimport numpy as np
 import numpy as np
 
-from .constants import *
 from .reaction cimport *
 from ._utils cimport *
 from . import _utils
--- a/interfaces/cython/cantera/kinetics.pyx
+++ b/interfaces/cython/cantera/kinetics.pyx
@@ -887,7 +887,7 @@ cdef class InterfaceKinetics(Kinetics):
         phase ``phase`` (Units: A/m^2 for a surface, A/m for an edge reaction).
         """
         i_phase = self.phase_index(phase)
-        return (<CxxInterfaceKinetics*>self.kinetics).InterfaceCurrent(i_phase)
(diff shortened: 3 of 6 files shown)


speth:
Thanks for this contribution, @c-randall.

As a procedural note, please don't introduce merge commits as part of your PR. We will rebase this on top of the current `main` branch when accepting the PR.

decaluwe:
@speth, @bryanwweber @ischoegl  etal - in addition to his comment on the C++ code base, @c-randall and I were wondering if there is any desire to also implement in the Matlab interface, at this point, or just wait for the toolbox redesign to be complete?

ischoegl:
> @speth, @bryanwweber @ischoegl etal - in addition to his comment on the C++ code base, @c-randall and I were wondering if there is any desire to also implement in the Matlab interface, at this point, or just wait for the toolbox redesign to be complete?

@decaluwe / @c-randall ... If you'd like to add it to MATLAB, I'd suggest to create the functionality in C++ (as suggested at the top), so it can be called from any API (either directly from C++ or from `clib`). I.e. the API would just add a wrapper for core functionality.

decaluwe:
Thanks @ischoegl --yes, that's what we would do, but the question is more whether we are holding off on Matlab updates for the time being until the toolbox redesign is complete.

(also, if/when we do put it in C++, would the correct method be something like `void InterfaceKinetics::getInterfaceCurrent(double* i_int)`)? TIA.

ischoegl:
> Thanks @ischoegl --yes, that's what we would do, but the question is more whether we are holding off on Matlab updates for the time being until the toolbox redesign is complete.
> 
> (also, if/when we do put it in C++, would the correct method be something like `void InterfaceKinetics::getInterfaceCurrent(double* i_int)`)? TIA.

Since there are plans to include this in other API's, I believe it should be implemented in C++ regardless of whether it's added to the current MATLAB toolbox or not. The signature should be
```C++
double InterfaceKinetics::interfaceCurrent() const;
```
where the `get` prefix can be omitted (while it is not accessing a member variable, it's still a simple calculation). If you add it to `clib` as well, adding the method to either of the two MATLAB toolboxes (or Fortran, or the .NET interface) will be very simple, i.e. it's ready to go.

include/cantera/kinetics/InterfaceKinetics.h
```diff
@@ -287,6 +287,8 @@ class InterfaceKinetics : public Kinetics
      */
     int phaseStability(const size_t iphase) const;
 
+    double InterfaceCurrent(int iPhase);
```
decaluwe: Given the other declarations in this file, I wonder if this might be preferred:

```suggestion
    double InterfaceCurrent(const size_t iphase);
```

interfaces/cython/cantera/kinetics.pxd
```diff
@@ -69,6 +69,7 @@ cdef extern from "cantera/kinetics/InterfaceKinetics.h":
     cdef cppclass CxxInterfaceKinetics "Cantera::InterfaceKinetics":
         void advanceCoverages(double, double, double, double, size_t, size_t) except +translate_exception
         void solvePseudoSteadyStateProblem() except +translate_exception
+        double InterfaceCurrent(int) except +translate_exception
```
decaluwe: If the declaration in `InterfaceKinetics.h` is changed, then change accordingly, here.

interfaces/cython/cantera/kinetics.pyx
```diff
@@ -880,6 +880,15 @@ cdef class InterfaceKinetics(Kinetics):
         """
         return self.net_production_rates[self._phase_slice(phase)]
 
+    def interface_current(self, phase):
+        """
+        The interface current is useful when charge transfer reactions occur at
+        an interface. It is defined here as the net positive charge entering the
+        phase ``phase`` (Units: A/m^2 for a surface, A/m for an edge reaction).
+        """
+        iPhase = self.phase_index(phase)
```
decaluwe: `i_phase` would be more python-ic.

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -595,4 +595,30 @@ void InterfaceKinetics::setPhaseStability(const size_t iphase, const int isStabl
     }
 }
 
+double InterfaceKinetics::InterfaceCurrent(int iPhase)
```
decaluwe: Again, change accordingly if the declaration changes.

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -595,4 +595,30 @@ void InterfaceKinetics::setPhaseStability(const size_t iphase, const int isStabl
...
+{
+    int sp = thermo(iPhase).nSpecies();
```
decaluwe: This is nitpicky, but I'd recommend something like `nSp` or `n_sp`.  `sp` in this class typically refers to an actual species object.

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -595,4 +595,30 @@ void InterfaceKinetics::setPhaseStability(const size_t iphase, const int isStabl
...
+    double charge_k[sp];
+    double net_k[sp];
```
decaluwe: Again, nitpicky, but something like `sdot_k` would be more consistent with the nomenclature in the literature, which makes it a little easier for an interested reader to comprehend.

test/python/test_kinetics.py
```diff
@@ -1065,6 +1065,32 @@ def cathode_current(phi_s, phi_l, X_Li_cathode):
...
+        anode.X = [0.9, 0.1]
+        elyte.X = [0.4, 0.3, 0.15, 0.15]
+
+        anode.electric_potential = 0.
+        elyte.electric_potential = 3.
+
+        phases = [anode_int, anode, elect, elyte]
+
+        for p in phases:
+            productions = anode_int.get_net_production_rates(p)
```
decaluwe: Again, a name like `sdot_k` or even just `net_prod_rates` would probably be more instructive.

decaluwe (commented):
Hi @c-randall, excellent work moving this into the C++.  I have left a few comments below, most of them are nitpicky and related to naming conventions and nomenclature, where I think consistency with other places in the code or with the literature at large might make it (slightly) easier for a reader to follow.

The only one which might matter is for the function declaration in the header file. TBQH, I've never really understood what the `size_t` data type is or what it's for, but have tried to mimic how it gets used, elsewhere in the code.  Given that, you might change the data type of your input phase index.  

I also don't know if the declaration needs an `const;` appended to the end. Nearby functions that return variables do have it, but my C++ is not such that I know why or when this is done.  @bryanwweber @speth @ischoegl , etc., can one of you weigh in?

Again, very nice job, and thanks for including a test of these capabilities.

include/cantera/kinetics/InterfaceKinetics.h
c-randall: I read about the const qualifier and size_t type and agree with this. I have updated the code here and in other areas to account for this.

interfaces/cython/cantera/kinetics.pyx
c-randall: Agreed, this change has been made.

src/kinetics/InterfaceKinetics.cpp
c-randall: Agreed, I changed to nSp.

src/kinetics/InterfaceKinetics.cpp
c-randall: Agreed, this has been changed.

test/python/test_kinetics.py
c-randall: Agreed, variable names are now updated. I went with `net_prod_rates` here.

include/cantera/kinetics/InterfaceKinetics.h
```diff
@@ -287,6 +287,17 @@ class InterfaceKinetics : public Kinetics
...
+    //! Gets the interface current for the ith phaseExistence
+    /*!
+    * @param iphase Phase Id
+    * @return The double specifying the interface current. The interface Current
+    *         is useful when charge transfer reactions occur at an interface. It
+    *         is defined here as the net positive charge entering the phase
+    *         specified by the Phase Id. (Units: A/m^2 for a surface reaction,
+    *         A/m for an edge reaction).
+    */
+    double InterfaceCurrent(const size_t iphase);
```
ischoegl:
```suggestion
    double interfaceCurrent(const size_t iphase);
```
methods should be lower case.

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -595,4 +595,30 @@ void InterfaceKinetics::setPhaseStability(const size_t iphase, const int isStabl
     }
 }
 
+double InterfaceKinetics::InterfaceCurrent(const size_t iphase)
+{
+    int nSp = thermo(iphase).nSpecies();
+    double charge_k[nSp];
+    double sdot_k[nSp];
```
ischoegl: These should be `vector_fp`'s, as sizes aren't known at compile time.

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -595,4 +595,30 @@ void InterfaceKinetics::setPhaseStability(const size_t iphase, const int isStabl
...
+    doublereal netProdRates[m_kk];
```
ischoegl: no `doublereal` in new code. Also, should be `vector_fp`.

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -595,4 +595,30 @@ void InterfaceKinetics::setPhaseStability(const size_t iphase, const int isStabl
...
+    {
+        sdot_k[k] = netProdRates[m_start[iphase]+k];
+    }
+
+    double dotProduct = 0.0;
+
+    for(int k=0; k<nSp; k++)
+    {
+        dotProduct += charge_k[k]*sdot_k[k];
+    }
```
ischoegl: those two loops can be combined, which eliminates the need for `s_dot_k`. As a minor formatting quibble, there should be spaces before and after operators.

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -595,4 +595,30 @@ void InterfaceKinetics::setPhaseStability(const size_t iphase, const int isStabl
...
+
+    return dotProduct*Faraday;
```
ischoegl:
```suggestion
    return dotProduct * Faraday;
```

ischoegl (changes requested):
Hi @c-randall ... thanks for the updates!

I have some minor comments that should ensure that the code complies with our C++ coding style (see [CONTRIBUTING.md](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md#c)) and avoids unnecessary loops.

include/cantera/kinetics/InterfaceKinetics.h
c-randall: Thank you for all the comments. I updated the method to be lower case everywhere that it is referenced.

src/kinetics/InterfaceKinetics.cpp
c-randall: Changed both of these to `vector_fp` type.

src/kinetics/InterfaceKinetics.cpp
c-randall: Removed the `s_dot_k` variable and integrated both loops into one single loop.

src/kinetics/InterfaceKinetics.cpp
c-randall: Added spaces around all operations.

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -595,4 +595,21 @@ void InterfaceKinetics::setPhaseStability(const size_t iphase, const int isStabl
...
+double InterfaceKinetics::interfaceCurrent(const size_t iphase)
+{
+    vector_fp charges(m_kk, 0.0);
+    vector_fp netProdRates(m_kk, 0.0);
+    double dotProduct = 0.0;
+
+    thermo(iphase).getCharges(charges.data());
+    getNetProductionRates(netProdRates.data());
+
+    for(size_t k=0; k < thermo(iphase).nSpecies(); k++)
```
ischoegl:
```suggestion
    for (size_t k = 0; k < thermo(iphase).nSpecies(); k++)
```

test/python/test_kinetics.py
```diff
@@ -1065,6 +1065,32 @@ def cathode_current(phi_s, phi_l, X_Li_cathode):
...
         assert np.allclose(data, ref, rtol=1e-7)
 
+    def test_interface_current(self):
+        file = "lithium_ion_battery.yaml"
+
+        # The 'elde' electrode phase is needed as a source/sink for electrons:
+        anode = ct.Solution(file,"anode")
+        elect = ct.Solution(file,"electron")
+        elyte = ct.Solution(file,"electrolyte")
+        anode_int = ct.Interface(file,"edge_anode_electrolyte",[anode,elect,elyte])
```
ischoegl:
```suggestion
        anode = ct.Solution(file, "anode")
        elect = ct.Solution(file, "electron")
        elyte = ct.Solution(file, "electrolyte")
        anode_int = ct.Interface(file, "edge_anode_electrolyte", [anode, elect, elyte])
```

test/python/test_kinetics.py
```diff
@@ -1065,6 +1065,32 @@ def cathode_current(phi_s, phi_l, X_Li_cathode):
...
+        phases = [anode_int, anode, elect, elyte]
+
+        for p in phases:
+            net_prod_rates = anode_int.get_net_production_rates(p)
+            charges = p.charges
+
+            method = anode_int.interface_current(p)
+            manual = sum(net_prod_rates*charges)*ct.faraday
+
+            self.assertEqual(method,manual)
```
ischoegl:
```suggestion
            manual = sum(net_prod_rates * charges) * ct.faraday

            self.assertEqual(method, manual)
```

ischoegl (changes requested):
@c-randall ... thanks for the quick turn-around. I located some more missing white spaces, but otherwise this looks great!

ischoegl (approved):
LGTM - thanks!

decaluwe:
@ischoegl - many thanks for looking this over, and for the improvements.

One question on one of your comments, for my own knowledge. I nearly commented on this, but since `Interface::netProdRates` returns a `doublereeal`, wasn't sure: does the function calling it not need to accept it's return type as such?  Thanks again.

ischoegl:
Sure. `doublereal` and `double` can be used interchangeably as they really are the same type - there are some historical Fortran-related definitions in `config.h`, see

```C++
// define types doublereal, integer, and ftnlen to match the
// corresponding Fortran data types on your system. The defaults
// are OK for most systems

typedef double doublereal;   // Fortran double precision
typedef int integer;      // Fortran integer
typedef int ftnlen;       // Fortran hidden string length type
```
As far as I am aware, there isn't a reason to keep `doublereal` around any longer. At the same time, replacing them all - while easy - would create unnecessary churn that should be avoided.

PR #1416, opened 2022-12-19
WIP - Update electrochemical reaction type input for BV and Marcus kinetics

korffdm:

- Change input method from YAML to choose Butler-Volmer kinetics for electrochemical reactions by string instead of bool
- Adds input option for electrochemical reactions to use Marcus theory kinetics

The current WIP updates the input field of the YAML for reactions to choose specific electrochemical kinetics models. This new field is called "echem-kinetics-form" and accepts a string input for either "Butler-Volmer" or "Marcus". If either of these strings is used then the appropriate term to modify the reaction rate evaluation will be added to the correction factor in voltageCorrection of InterfaceRate.h

Related to Cantera/enhancements#38 

**Checklist**

### DIFF
--- a/include/cantera/kinetics/InterfaceRate.h
+++ b/include/cantera/kinetics/InterfaceRate.h
@@ -123,6 +123,10 @@ public:
         return m_exchangeCurrentDensityFormulation;
     }
 
+    std::string eChemForm() {
+        return m_eChemForm;
+    }
+
     //! Build rate-specific parameters based on Reaction and Kinetics context
     //! @param rxn  Reaction associated with rate parameterization
     //! @param kin  Kinetics object associated with rate parameterization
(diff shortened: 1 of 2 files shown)


decaluwe:
Hi everyone, this begins our long-awaited move to overhaul and expand our electrochemistry.  @korffdm and I decided to stop and do a WIP PR to hash things out before moving too much farther and making test, documentation, etc.

We decide to add the Marcus kinetics routine, for now, as a parallel option to the current Butler-Volmer for (mostly accomplished via `m_exchangeCurrentDensityFormulation` in `InterfaceRate.h` header file), rather than radically changing how both are done.  But  honestly I am not in love with the current approach. One of the attractive aspects of the B-V type reaction rate form (I would call the Marcus rate a "flavor" of the B-V form) is its simplicity:
$$\dot{q}_{\rm rop} =\frac{i_o}{nF}\left[\exp\left( -\frac{\beta nF\eta}{RT}\right)  - \exp\left( \frac{(1-\beta) nF\eta}{RT}\right)  \right]$$
where the exchange current density $i_o$ is a function of the forward rate constant and the activity concentrations, and the overpotential $\eta$ is the net electrochemical driving force on the reaction (calculated in Cantera via the `echemPotentials` property in the C++ code.

Anyway, long story short: if you look at our earlier implementation of these kinetics ([here](https://github.com/CSM-Offenburg/cantera/blob/a91ed333a0f6448f0ab99fa76c9be0f1368ce03a/src/kinetics/InterfaceKinetics.cpp#L349), for example), it was much easier to trace the theory than in our current implementation.  This was all before the creation of the `KineticsFactory` approach. I certainly get why the new approach is preferable over a series of flags, bools, and switch statements. But even as someone who knew what he was looking for, it was very difficult to even verify that the current Butler-Volmer routine does what I think it should do.

So, even shorter: is there any way to implement this kinetics in the new framework that looks more similar to the earlier implementation (again, [here](https://github.com/CSM-Offenburg/cantera/blob/a91ed333a0f6448f0ab99fa76c9be0f1368ce03a/src/kinetics/InterfaceKinetics.cpp#L349))? If not, we have the current approach working and can stick with it. But it really will make it difficult (imho) for new developers to follow behind and verify, alter, or expand the approach, unless they are extremely motivated.

speth:
I think it would make sense to set a direction for this in the context of some previous discussions we've had on the topic of electrochemical reactions. Namely, https://github.com/Cantera/enhancements/issues/38, which includes what I think is the most detailed description of the [theory](https://github.com/Cantera/cantera/files/3883908/Charge.Transfer.in.Cantera.pdf) that we have as written up by @decaluwe, and https://github.com/Cantera/enhancements/issues/44.

Quite a bit has changed since those topics were opened, from the complete transition to the YAML input format to the major refactoring and generalization of how reaction rates are handled in the C++ layer, so I think there should be an opportunity to implement this in a much cleaner way that might have been possible in the past.

I'd suggest that either of those Enhancement discussions would be a better location for hashing out the details of these features, rather than in a PR review, where GitHub has the unfortunate habit of starting to hide posts in once there are detailed code reviews mixed in with the higher-level discussion.

PR #1413, opened 2022-12-10
Removed redundant error checks to fix issue #1412

ssun30:
**Changes proposed in this pull request**

- Removes a redundant error check in `clib/ct.cpp` so external language interfaces can correctly handle errors returned by the method `thermo_setDensity` and `thermo_setMolarDensity`. 

**If applicable, fill in the issue number this pull request is fixing**

Closes #1412 

**Checklist**

- [ ] The pull request includes a clear description of this code change
- [ ] Commit messages have short titles and reference relevant issues
- [ ] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [ ] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [ ] The pull request is ready for review

PR #1417, opened 2022-12-21
Customizable data objects for ExtensibleRate

speth:

This PR introduces `ExtensibleRateData` class that is used for collecting state data and being passed as the argument to the `eval` method of `ExtensibleRate` objects, analogous to the C++ implementation of reaction rates.

This is the first follow-up to https://github.com/Cantera/cantera/pull/1382, and contributes to the implementation of https://github.com/Cantera/enhancements/issues/79.

The implementation of an `ExtensibleRate` now looks like the following (from `custom_reactions.py`):
```py
class ExtensibleArrheniusData(ct.ExtensibleRateData):
    __slots__ = ("T",)
    def __init__(self):
        self.T = None

    def update(self, gas):
        T = gas.T
        if self.T != T:
            self.T = T
            return True
        else:
            return False

@ct.extension(name="extensible-Arrhenius", data=ExtensibleArrheniusData)
class ExtensibleArrhenius(ct.ExtensibleRate):
    __slots__ = ("A", "b", "Ea_R")
    def set_parameters(self, params, units):
        self.A = params["A"]
        self.b = params["b"]
        self.Ea_R = params["Ea_R"]

    def eval(self, data):
        return self.A * data.T**self.b * exp(-self.Ea_R/data.T)
```

**Checklist**


speth:
I'm not sure what's going on with code coverage here. The deltas shown in Codecov seem to be all over the place for several recent PRs (for example, #1413) that should have had almost no changes, and in this case, I see lines/functions being showed as uncovered (such as `ExtensionManager::wrapSolution`) that are definitely called during the test suite on the coverage runner. These lines are shown as uncovered in the gcov report that's uploaded as an artifact, so it's not (just) a Codecov issue.

ischoegl:
@speth ... thank you for your continued work on extensible objects. Before going through a review, I wanted to ask about the long game - for lack of a better term, extensible objects rely on 'plugins' written in a high level language (e.g. code in `test/python`). In the long run, will these be auto-generated from YAML code or will a user be expected to write those plugins themselves? Depending on the answer, it may or may not make sense to start adding examples for each `extensible` object in `samples/python`.

samples/python/kinetics/custom_reactions.py
```diff
@@ -25,41 +25,67 @@
...
+    def __init__(self):
+        self.T = None
+
+    def update(self, gas):
+        T = gas.T
+        if self.T != T:
+            self.T = T
+            return True
+        else:
+            return False
```
ischoegl, line 49: While this may go beyond the scope of this PR (it is based on choices made in the C++ core), it is likely confusing to users to return a boolean. For C++, things are well hidden so the added complexity needed for avoiding unnecessary evaluations doesn't matter. For `ExtensibleRateData`, it may be easier to ignore things, meaning:
```Python
def update(self, gas):
    self.T = gas.T
```

interfaces/cython/cantera/delegator.pyx
```diff
@@ -362,16 +372,20 @@ def extension(*, name):
 
         import cantera as ct
 
-        @ct.extension(name="cool-rate")
+        class CoolRateData(ct.ExtensibleRateData):
+            def update(self, soln):
+                ...
```
ischoegl, line 379: Docstring should mention that a boolean needs to be returned.

samples/python/kinetics/custom_reactions.py
ischoegl, line 49: If a user does not implement a return value (as in the following, where I simply disabled the if/else block), the following error is thrown:
```
CanteraError:
CanteraError thrown by CVodesIntegrator::integrate:
CVodes error encountered. Error code: -9
The right-hand side routine failed at the first call.

Exceptions caught during RHS evaluation:

CanteraError thrown by Lambda generated by Delegator::makeDelegate:
Delegate for function of type 'std::__1::function<double (void*)>'
did not return a value

Components with largest weighted error estimates:
38: -5.51375772662609
18: -0.3289589063852491
17: 0.20937118185164436
5: 0.11976563250530743
7: 0.10814966603409117
8: 0.035912234521081284
3: -0.011193624655501031
39: -0.007491081474199196
9: 0.0017188199018761563
4: 0.0007281059930890118
```
which will likely create some traffic on the user group.

It would be good if the error message could identify what delegated methods is responsible (in this case `ExtensibleArrheniusData.update`), while eliminating the misleading `CVodesIntegrator::integrate` reference.

test/python/test_reaction.py
```diff
@@ -1499,21 +1500,44 @@ def func(T):
         assert (gas.forward_rate_constants == gas.T).all()
 
 
-@ct.extension(name="user-rate-1")
+class UserRate1Data(ct.ExtensibleRateData):
+    def __init__(self):
+        super().__init__()
```
ischoegl: I don't think that the call to `super` is necessary?

ischoegl (changes requested):
@speth ... thank you for your continued work on *extensible* objects! (and sorry about false starts for this review.) The implementation itself looks good to me, but I have some (minor?) comments, which will hopefully improve the user experience. It is to be expected that users will create buggy extensions, where improved exception handling will give valuable feedback.

On a separate note, it turns out that the simplistic `CustomRate` is about twice as fast as `ExtensibleRate`:
```
%run custom_reactions.py
- New framework (YAML): 18.44 μs/step (T_final=2780.21)
- Two Custom reactions: 18.86 μs/step (T_final=2780.21) ...+2.29%
- Two Extensible reactions: 19.43 μs/step (T_final=2780.21) ...+5.37%
```
There is likely little that can be done about this (`ExtensibleRate/ExtensibleRateData` is *extremely* flexible and can be serialized); at the same time, I was hoping that `CustomRate` could be deprecated in favor of `ExtensibleRate`. What's your opinion here?

interfaces/cython/cantera/delegator.pyx
speth: This is a sample, showing how to use the `extension` decorator. The documentation for the `update` method is in the `ExtensibleRateData` class, and does document the requirement for the return value.

samples/python/kinetics/custom_reactions.py
speth: I've updated this so more information is available at the point where the exception is thrown (so the class and method name can be identified). The output is now something like:
```
CanteraError thrown by CVodesIntegrator::integrate:
CVodes error encountered. Error code: -9
The right-hand side routine failed at the first call.

Exceptions caught during RHS evaluation:

CanteraError thrown by Lambda generated by Delegator::makeDelegate:
Method 'update' of class 'ExtensibleArrheniusData' did not return a value of type 'double'.

Components with largest weighted error estimates:
0: -1.3642641433719054e+86
38: -10.117372889372218
7: 0.5206771669918789
8: -0.33262364359287805
18: -0.27521731705348845
5: 0.18190017240387882
17: 0.17516792601392808
39: -0.014443245050882536
3: 0.003992263897614869
9: 0.0023360039408578046
```
I think the reference to `CVodesIntegrator::integrate` should be expected because `CVodesIntegrator` does some of its own error handling and not every exception raised during the integration necessarily has to propagate back to the user.

speth:
> There is likely little that can be done about this (`ExtensibleRate/ExtensibleRateData` is _extremely_ flexible and can be serialized); at the same time, I was hoping that `CustomRate` could be deprecated in favor of `ExtensibleRate`. What's your opinion here?

In terms of performance, the sample problem doesn't tell the whole story. Both `CustomRate` and `ExtensibleRate` will call the Python rate evaluation function the same number of times, and this call count will scale with the number of custom/extensible reactions. However, `CustomRate` still has its function for updating the `ReactionData` object implemented in C++. This function is called only once per reaction *type*, and so this overhead won't increase any further if more rates are implemented in Python, assuming there are only a few distinct reaction types. It just so happens that for this particular case, because of the conditions where the rate data needs to (potentially) be updated, we end up with approximately twice as many calls to Python-wrapped functions as for the `CustomRate` route.

I would say that the decision on whether to keep `CustomRate` should come down to usability. For the cases where it can be used now, it is certainly less code to write than an equivalent version of `ExtensibleRate`. Perhaps the right compromise would be a built-in `ExtensibleRate` implementation for rates that are functions of temperature only.

I have a few ideas on how to improve the performance of these Python rates, but those are definitely for a later PR.

ischoegl (dismissed):
@speth ... thanks for taking care of the improved exception handling, which I believe will go a long way.

*PS: I also appreciate your thoughts on performance.*

speth:
@ischoegl, can you re-approve? My fiddling around to clear up some unrelated CI issues cleared your previous approval.

ischoegl (approved):

PR #1299, opened 2022-05-31
Perform a comprehensive set of thermodynamic consistency tests for all thermo models

speth:
**Changes proposed in this pull request**

- Introduce a uniform set of thermodynamic consistency tests that can be performed for all thermo models with a minimal amount of code duplication, totaling nearly 2000 distinct unit tests.
- Provide options for skipping certain tests for specific thermo models when there are known failures
- Specific tests are skipped automatically if a method raises a `NotImplementedError`
- Bump the version of Googletest used, which also bumps the minimum version of Visual Studio to 2017 (though the oldest version we can test on is 2019, due to a SCons bug).
- Remove a number of now-redundant ad hoc tests of thermodynamic consistency

I've already uncovered several bugs using these tests, and have created issues for the ones that aren't being resolved as part of this PR (all of which tag this PR).

**If applicable, fill in the issue number this pull request is fixing**

Resolves Cantera/enhancements#114

**If applicable, provide an example illustrating new features this pull request is introducing**

All of the configuration for a set of consistency tests is specified in a YAML file, `consistency-tests.yaml`:
```yaml
binary-solution-tabulated:
  setup:
    file: thermo-models.yaml
    phase: graphite-anode
    known-failures:
      g_eq_h_minus_Ts: Inconsistent results when P != 1 atm (states 2 and 3)
      g_eq_sum_gk_Xk: Inconsistent result when P != 1 atm (states 2 and 3)
  states:
  - {T: 300, P: 1 atm, X: {"Li[anode]": 0.3, "V[anode]": 0.7}}
  - {T: 320, P: 1 atm, X: {"Li[anode]": 0.3, "V[anode]": 0.7}}
  - {T: 300, P: 10 atm, X: {"Li[anode]": 0.6, "V[anode]": 0.4}}
  - {T: 300, P: 10 atm, X: {"Li[anode]": 0.0, "V[anode]": 1.0}}
```

And there is just a short block of code to create the parameterized test suite:
```c++
INSTANTIATE_TEST_SUITE_P(BinarySolutionTabulated, TestConsistency,
    testing::Combine(
        testing::Values(getSetup("binary-solution-tabulated")),
        testing::ValuesIn(getStates("binary-solution-tabulated")))
);
```

Which runs a set of different tests for each of the specified state vectors. Failing tests show the information about the particular case that is failing, e.g.:
```
[ RUN      ] BinarySolutionTabulated/TestConsistency.h_eq_sum_hk_Xk/0
[...]

PR #1421, opened 2023-01-03
Add isothermal_compressibility and thermal_expansion_coefficient for Redlich-Kwong and Peng-Robinson EOS

### LINKED ISSUE #122: Add `isothermal_compressibility` and `thermal_expansion_coefficient` for Redlich-Kwong and Peng-Robinson EOS
**Abstract**

Implementing the required functions to calculate the isothermal compressibility and thermal expansion coefficient for the Peng-Robinson EOS added in https://github.com/Cantera/cantera/pull/1047. 

**Motivation**

I'm currently developing a Python package that solvers for shock conditions (in a shock tube) using a real gas EOS. Currently, the implementation of Peng-Robinson does not define the `isothermal_compressibility` and `thermal_expansion_coefficient` properties that the algorithm requires. 

**Description**

I found [this paper](https://journals.flvc.org/cee/article/view/122983) that includes the derivation of the derivative of molar mass with respect to temperature at constant pressure, which can be used for the thermal expansion coefficient. I couldn't find a similar resource relevant to the isothermal compressibility factor. I have attached [my attempt at a derivation](https://github.com/Cantera/enhancements/files/7688575/Peng_Robinson_Derivatives.pdf) of the isothermal compressibility factor in case there are any errors in my understanding of the thermodynamic principles or the derivatives.

I have not yet started on any code, as I have limited C++ experience and expect to have some trouble implementing it. I will document my progress here, and if anyone is willing to help, that would be great!

corykinney:

Add `isothermalCompressibility` and `thermalExpansionCoeff` definitions to the `RedlichKwongMFTP` and `PengRobinson` models.
Add corresponding thermo consistency tests `betaT_eq_minus_dmv_dP_const_T_div_mv` and `alphaV_eq_dmv_dT_const_P_div_mv` based on finite differences.

Closes https://github.com/Cantera/enhancements/issues/122

**Checklist**

### DIFF
--- a/include/cantera/thermo/PengRobinson.h
+++ b/include/cantera/thermo/PengRobinson.h
@@ -205,6 +205,9 @@ protected:
 
 public:
 
+    virtual double isothermalCompressibility() const;
+    virtual double thermalExpansionCoeff() const;
+
     //! Calculate \f$dp/dV\f$ and \f$dp/dT\f$ at the current conditions
     /*!
      *  These are stored internally.
@@ -248,7 +251,7 @@ protected:
      */
     double m_a;
 
-    //! Value of \f$\alpha\f$ in the equation of state
+    //! Value of \f$a \alpha\f$ in the equation of state
     /*!
      *  `m_aAlpha_mix` is a function of the temperature and the mole fractions.
      */
--- a/test/thermo/consistency.cpp
+++ b/test/thermo/consistency.cpp
@@ -74,6 +74,8 @@ public:
         atol = setup.getDouble("atol", 1e-5);
         rtol_fd = setup.getDouble("rtol_fd", 1e-6);
         atol_v = setup.getDouble("atol_v", 1e-11);
+        atol_c = setup.getDouble("atol_c", 1e-14);
+        atol_e = setup.getDouble("atol_e", 1e-18);
 
         phase = cache[key];
         phase->setState(state);
@@ -104,6 +106,8 @@ public:
     double T, p, RT;
     double atol; // absolute tolerance for molar energy comparisons
     double atol_v; // absolute tolerance for molar volume comparisons
+    double atol_c; // absolute tolerance for compressibility comparison
+    double atol_e; // absolute tolerance for expansion comparison
     double rtol_fd; // relative tolerance for finite difference comparisons
 };
 
@@ -343,6 +347,58 @@ TEST_P(TestConsistency, dSdv_const_T_eq_dPdT_const_V) {
     } else {
         GTEST_SKIP() << "Undefined for incompressible phase";
     }
+}
+
+TEST_P(TestConsistency, betaT_eq_minus_dmv_dP_const_T_div_mv)
+{
+    double betaT1;
+    try {
+        betaT1 = phase->isothermalCompressibility();
+    } catch (NotImplementedError& err) {
+        GTEST_SKIP() << err.getMethod() << " threw NotImplementedError";
+    }
+
+    double T = phase->temperature();
+    double P1 = phase->pressure();
+    double mv1 = phase->molarVolume();
+    
+    double P2 = P1 * (1 + 1e-6);
+    phase->setState_TP(T, P2);
+    double betaT2 = phase->isothermalCompressibility();
+    double mv2 = phase->molarVolume();
+
+    double betaT_mid = 0.5 * (betaT1 + betaT2);
+    double mv_mid = 0.5 * (mv1 + mv2);
+    double betaT_fd = -1 / mv_mid * (mv2 - mv1) / (P2 - P1);
+
+    EXPECT_NEAR(betaT_fd, betaT_mid, 
+                max({rtol_fd * betaT_mid, rtol_fd * betaT_fd, atol_c}));
+}
+
+TEST_P(TestConsistency, alphaV_eq_dmv_dT_const_P_div_mv)
+{
+    double alphaV1;
+    try {
+        alphaV1 = phase->thermalExpansionCoeff();
+    } catch (NotImplementedError& err) {
+        GTEST_SKIP() << err.getMethod() << " threw NotImplementedError";
+    }
+
+    double P = phase->pressure();
+    double T1 = phase->temperature();
+    double mv1 = phase->molarVolume();
+
+    double T2 = T1 * (1 + 1e-6);
+    phase->setState_TP(T2, P);
+    double alphaV2 = phase->thermalExpansionCoeff();
+    double mv2 = phase->molarVolume();
+
+    double alphaV_mid = 0.5 * (alphaV1 + alphaV2);
+    double mv_mid = 0.5 * (mv1 + mv2);
+    double alphaV_fd = 1 / mv_mid * (mv2 - mv1) / (T2 - T1);
+
+    EXPECT_NEAR(alphaV_fd, alphaV_mid, 
+                max({rtol_fd * alphaV_mid, rtol_fd * alphaV_fd, atol_e}));
 }
 
 // ---------- Tests for consistency of standard state properties ---------------
--- a/AUTHORS
+++ b/AUTHORS
@@ -29,6 +29,7 @@ John Hewson (@jchewson), Sandia National Laboratory
 Trevor Hickey (@luxe)
 Yuanjie Jiang
 Benjamin Kee (@lionkey)
+Cory Kinney (@corykinney)
 Gandhali Kogekar (@gkogekar)
 Daniel Korff (@korffdm), Colorado School of Mines
 Corey R. Randall (@c-randall), Colorado School of Mines
--- a/include/cantera/thermo/RedlichKwongMFTP.h
+++ b/include/cantera/thermo/RedlichKwongMFTP.h
@@ -173,6 +173,9 @@ public:
     virtual doublereal densSpinodalGas() const;
     virtual doublereal dpdVCalc(doublereal TKelvin, doublereal molarVol, doublereal& presCalc) const;
 
+    virtual double isothermalCompressibility() const;
+    virtual double thermalExpansionCoeff() const;
+
     //! Calculate dpdV and dpdT at the current conditions
     /*!
      *  These are stored internally.
--- a/src/thermo/PengRobinson.cpp
+++ b/src/thermo/PengRobinson.cpp
@@ -630,6 +630,18 @@ double PengRobinson::dpdVCalc(double T, double molarVol, double& presCalc) const
     return -GasConstant * T / (vmb * vmb) + 2 * m_aAlpha_mix * vpb / (denom*denom);
 }
 
+double PengRobinson::isothermalCompressibility() const
+{
+    calculatePressureDerivatives();
+    return -1 / (molarVolume() * m_dpdV);
+}
+
+double PengRobinson::thermalExpansionCoeff() const
+{
+    calculatePressureDerivatives();
+    return -m_dpdT / (molarVolume() * m_dpdV);
+}
+
 void PengRobinson::calculatePressureDerivatives() const
 {
     double T = temperature();
--- a/src/thermo/RedlichKwongMFTP.cpp
+++ b/src/thermo/RedlichKwongMFTP.cpp
@@ -707,6 +707,18 @@ doublereal RedlichKwongMFTP::dpdVCalc(doublereal TKelvin, doublereal molarVol, d
     return dpdv;
 }
 
+double RedlichKwongMFTP::isothermalCompressibility() const
+{
+    pressureDerivatives();
+    return -1 / (molarVolume() * dpdV_);
+}
+
+double RedlichKwongMFTP::thermalExpansionCoeff() const
+{
+    pressureDerivatives();
+    return -dpdT_ / (molarVolume() * dpdV_);
+}
+
 void RedlichKwongMFTP::pressureDerivatives() const
 {
     doublereal TKelvin = temperature();


corykinney:
For reference, the following finite difference script was used to calculate the values in the test and set reasonable tolerances:

```python
import cantera as ct
import numpy as np


def isothermal_compressibility(state, dP):
    T, P = state.TP
    state.TP = T, P - dP
    rho1 = state.density_mass
    state.TP = T, P + dP
    rho2 = state.density_mass
    state.TP = T, P
    return - state.density_mass * (1 / rho2 - 1 / rho1) / (2 * dP)


def thermal_expansion_coeff(state, dT):
    T, P = state.TP
    state.TP = T - dT, P
    rho1 = state.density_mass
    state.TP = T + dT, P
    rho2 = state.density_mass
    state.TP = T, P
    return state.density_mass * (1 / rho2 - 1 / rho1) / (2 * dT)


gas = ct.Solution("thermo-models.yaml", "CO2-PR")
gas.X = "CO2: 1"

betaT_ct = []
betaT_fd = []
alphaV_ct = []
alphaV_fd = []

P = 5e5
for i in range(10):
    T = 300 + 25 * i
    gas.TP = T, P
    betaT_ct.append(gas.isothermal_compressibility)
    alphaV_ct.append(gas.thermal_expansion_coeff)
    betaT_fd.append(isothermal_compressibility(gas, P * 1e-5))
    alphaV_fd.append(thermal_expansion_coeff(gas, T * 1e-5))

betaT_error = np.array(betaT_fd) - np.array(betaT_ct)
alphaV_error = np.array(alphaV_fd) - np.array(alphaV_ct)

print(f"Isothermal compressibility\nCantera: {betaT_ct}")
print(f"Finite difference error: ≤ {max(abs(betaT_error)):.5e}\n")
print(f"Thermal expansion coefficient\nCantera: {alphaV_ct}")
print(f"Finite difference error: ≤ {max(abs(alphaV_error)):.5e}")
```

which gives the following output for `PengRobinson`

```
Isothermal compressibility
Cantera: [2.0565735288179827e-06, 2.043354114849639e-06, 2.0336631623657743e-06, 2.0263828696824072e-06, 2.0208047779848385e-06, 2.0164611841648566e-06, 2.0130329255806156e-06, 2.010296131734592e-06, 2.0080900822010764e-06, 2.0062970726140575e-06]
Finite difference error: ≤ 2.20314e-16

Thermal expansion coefficient
Cantera: [0.0036236953242238055, 0.0032907286131624216, 0.003018150217266326, 0.0027901653673486353, 0.0025961927676142714, 0.0024288429329921953, 0.0022827789874782295, 0.0021540380684750378, 0.0020396082523816846, 0.0019371545393275069]
Finite difference error: ≤ 6.56762e-14
```

Note: For the `PengRobinson` functions, in addition to the finite difference check, the values from the functions were also compared to CoolProp output (CoolProp doesn't have RedlichKwong). 

It seems that there might be a slight difference in critical parameters used for carbon dioxide, as the latest version of CoolProp gives slightly different values for density than were used in the original `PengRobinson` test for density. Thus, the derivatives evaluated by CoolProp were also very close but not exact.

corykinney:
Note: For the `PengRobinson` functions, in addition to the finite difference check, the values from the functions were also compared to CoolProp output (CoolProp doesn't have RedlichKwong). 

It seems that there might be a slight difference in critical parameters used for carbon dioxide, as the latest version of CoolProp gives slightly different values for density than were used in the original `PengRobinson` test for density. Thus, the derivatives evaluated by CoolProp were also very close but not exact.

corykinney:
I marked it as ready for review a bit prematurely, but everything passes `scons test` now so it should be ready for review @ischoegl. Unless the checks need to be implemented using the framework from #1299

corykinney:
@ischoegl Sorry to tag you again, but the finite difference tests are now implemented as thermodynamic consistency tests as you suggested using the framework by @speth, which is great because now I don't have to do any extra work for #1423 😄. Everything passes `scons test` and is ready for review. Let me know if I need to clean up the commits, as I added the tests in the individual classes before removing and switching them.

include/cantera/thermo/PengRobinson.h
```diff
@@ -248,7 +254,7 @@ class PengRobinson : public MixtureFugacityTP
      */
     double m_a;
 
-    //! Value of \f$\alpha\f$ in the equation of state
+    //! Value of \f$\a alpha\f$ in the equation of state
```
ischoegl:
```suggestion
    //! Value of \f$a \alpha\f$ in the equation of state
```

ischoegl (commented):
@corykinney ... thanks - while I'm not familiar with the theory (and don't have time to catch up at the moment), I looked over the code; the only thing I found was a LaTeX docstring typo. @decaluwe - the actual changes look like they are well aligned with recent work by your group - could you review and approve as appropriate?

Regarding your question about squashing commits: as you introduce tests that are subsequently removed it makes sense to squash corresponding commits.

include/cantera/thermo/PengRobinson.h
corykinney: Ahh I thought I fixed that typo before committing! I'm not as familiar with the review process, do I go ahead and commit the suggestion and squash it later?

ischoegl: You can commit the suggestion and squash it / do a fix up or just update it in one of your own commits (and still squash / fix up)

corykinney: Perfect. I'll wait to squash it until the other reviews are done so that I can do it all in one. Thanks!

decaluwe:
Sorry for the delay - taking a look at the derivation, just to get another set of eyes on it... Stay tuned :)

corykinney:
@decaluwe I appreciate you taking the time to review it! The derivations in the original enhancement proposal were a bit outdated, as I didn't originally realize that $d(a\alpha)/dT$ was already implemented, so I compiled the [final version](https://github.com/Cantera/cantera/files/10501914/Compressibility_derivatives__PR__RK_.pdf) used for the implementation which should be easier to review.

corykinney:
@speth For the consistency tests with finite differences implemented here, I followed the tolerance pattern used for checking the final result of many of the other properties, in this case it looked like:

```
EXPECT_NEAR(betaT_fd, betaT_mid, max({rtol_fd * betaT_mid, rtol_fd * betaT_fd, atol}));
```

I would think `min` would be more appropriate in this case where `atol` might be relatively large compared to the magnitude of the value being checked, in which case using the product of `rtol_fd` and the expected value would be desired. But I wanted to check with you since every other test uses `max`.

speth:
> I would think `min` would be more appropriate in this case where `atol` might be relatively large compared to the magnitude of the value being checked, in which case using the product of `rtol_fd` and the expected value would be desired. But I wanted to check with you since every other test uses `max`.



> @speth For the consistency tests with finite differences implemented here, I followed the tolerance pattern used for checking the final result of many of the other properties, in this case it looked like:
> 
> ```
> EXPECT_NEAR(betaT_fd, betaT_mid, max({rtol_fd * betaT_mid, rtol_fd * betaT_fd, atol}));
> ```
> 
> I would think `min` would be more appropriate in this case where `atol` might be relatively large compared to the magnitude of the value being checked, in which case using the product of `rtol_fd` and the expected value would be desired. But I wanted to check with you since every other test uses `max`.

`max` is correct. The point of an absolute tolerance is to specify a limit below which we don't care about differences because both values are effectively zero. You need make sure to use an absolute tolerance that corresponds to "effectively zero" for these specific tests, though. The default value of `atol` used in other tests is 1e-5, where it is used to compare values that are in J/kg, and 1e-5 J/kg is indeed negligible for most purposes. But that's not really the right scale for the compressibility or thermal expansion coefficient.

corykinney:
@speth Thank you for the explanation. For some reason my impression was the opposite, which is that `atol` might be a required level of precision that would be used when `rtol` times the actual quantity is larger, but what you said makes a lot of sense!

I tried to select some reasonable numbers for the two quantities with units $1/K$ and $1/Pa$, for isothermal compressibility and thermal expansion coefficient, respectively. Based on the order of magnitudes of the ideal values `1/T` and `1/P`, considering that the real gas deviation will be one or two orders of magnitude smaller, I selected `1e-8` and `1e-12`. However, when setting these values the following tests failed:

```
Failed tests:
    - thermo: NitrogenPureFluid/TestConsistency.betaT_eq_minus_dmv_dP_const_T_div_mv/1
    - thermo: NitrogenPureFluid/TestConsistency.betaT_eq_minus_dmv_dP_const_T_div_mv/2
    - thermo: LiquidWaterIapws95/TestConsistency.betaT_eq_minus_dmv_dP_const_T_div_mv/0
```

@speth Do you think these tolerances are reasonable, or are they perhaps too restrictive?

corykinney:
Actually I realized the values were swapped, so now the tests pass and the only question is whether the `atol` values should be more restrictive or if they are appropriate.

corykinney:
@decaluwe Have you had a chance to look at the [final version](https://github.com/Cantera/cantera/files/10501914/Compressibility_derivatives__PR__RK_.pdf) (also linked above) of the derivations and the implementation in the last couple weeks?

corykinney:
@ischoegl Any chance this might get reviewed soon?

ischoegl:
@corykinney … I was under the impression that @decaluwe would review the theory.

include/cantera/thermo/PengRobinson.h
```diff
@@ -205,6 +205,12 @@ class PengRobinson : public MixtureFugacityTP
 
 public:
 
+    //! Returns the isothermal compressibility. Units: 1/Pa.
+    virtual double isothermalCompressibility() const;
+
+    //! Return the volumetric thermal expansion coefficient. Units: 1/K.
+    virtual double thermalExpansionCoeff() const;
```
speth: I think this would be a good place to document the formulas for these properties. Likewise in `RedlichKwongMFTP.h`

test/thermo/consistency.cpp
```diff
@@ -345,6 +349,56 @@ TEST_P(TestConsistency, dSdv_const_T_eq_dPdT_const_V) {
...
+        betaT1 = phase->isothermalCompressibility();
+    } catch (NotImplementedError& err) {
+        GTEST_SKIP() << err.getMethod() << " threw NotImplementedError";
+    }
+
+    double T = phase->temperature();
+    double P1 = phase->pressure();
+    double mv1 = phase->molarVolume();
+    
+    double P2 = P1 * (1 + 1e-7);
```
speth: If you use the same relative perturbation of 1e-6 that is used on other finite difference tests, I think you can set much tighter absolute tolerances for these tests. At least locally, with this larger perturbation, I was able to pass the test suite with `atol_c = 1e-14` and `atol_e = 1e-18`.

test/thermo/consistency.cpp
```diff
@@ -345,6 +349,56 @@ TEST_P(TestConsistency, dSdv_const_T_eq_dPdT_const_V) {
...
+    double T2 = T1 * (1 + 1e-7);
+    phase->setState_TP(T2, P);
+    double alphaV2 = phase->thermalExpansionCoeff();
+    double mv2 = phase->molarVolume();
+
+    double alphaV_mid = 0.5 * (alphaV1 + alphaV2);
+    double mv_mid = 0.5 * (mv1 + mv2);
+    double alphaV_fd = 1 / mv_mid * (mv2 - mv1) / (T2 - T1);
+
+    EXPECT_NEAR(alphaV_fd, alphaV_mid, max({rtol_fd * alphaV_mid, rtol_fd * alphaV_fd, atol_e}));
```
speth: Please wrap long lines (over 88 characters).

speth (changes requested):
Thank you for this PR, @corykinney. I think both the implementation of these properties for the P-R and R-K models and the additional consistency tests for these properties across all phase models are useful contributions.

I had some comments on the derivation of these formulas that I shared in https://github.com/Cantera/enhancements/issues/122 that I think can be used to simplify the calculations here. Other than that, I have just a few minor comments shown below.

test/thermo/consistency.cpp
corykinney: Wrapped both lines in a9b594528dbfc1c13cdf63c24abf16e5b4f38254

test/thermo/consistency.cpp
corykinney: Adjust tolerances in a9b594528dbfc1c13cdf63c24abf16e5b4f38254  as mentioned

include/cantera/thermo/PengRobinson.h
corykinney: Added the formulas in a9b594528dbfc1c13cdf63c24abf16e5b4f38254
Unfortunately I'm not certain how to build the documentation to verify, but it should be formatted correctly

corykinney:
@speth Currently working on testing the simplified equations mentiond in [your other comment](https://github.com/Cantera/enhancements/issues/122#issuecomment-1474973680). Regarding the applicability of the inverse function theorem, the Jacobian just needs to be invertible at whichever point we're evaluating the derivatives at in order to use these simplifications, correct?

corykinney:
@speth Switching to the derivative identity approach allowed for leveraging existing pressure derivative functions, so now the implementation is extremely simple. Let me know if everything looks good, and I can squash the changes.

include/cantera/thermo/PengRobinson.h
```diff
@@ -205,6 +205,22 @@ class PengRobinson : public MixtureFugacityTP
 
 public:
 
+    //! Returns the isothermal compressibility (\f$\beta_T\f$). Units: 1/Pa.
+    /*!
+     *  \f[ 
+     *    \beta_T = -\frac{1}{v} \left(\frac{\partial v}{\partial p}\right)_T
+     *  \f]
+     */
+    virtual double isothermalCompressibility() const;
```
speth: There's no need to repeat documentation for overrides unless there's some specific differences in the documentation, which isn't the case here.

speth (changes requested):
Thanks for the update to this PR, @corykinney. The revised implementation looks good to me -- I wasn't even expecting it to be that simple!

Besides the minor doc update, please add yourself to the `AUTHORS` file, and then I think this will be good to go.

include/cantera/thermo/PengRobinson.h
corykinney: I didn't realize the equations were already documented in `ThermoPhase`! The unnecessary comments are now removed.

corykinney:
@speth Let me know if there are any other changes :)

speth (approved):

Issue #1372, opened 2022-08-23
read_csv does not work for species name with a comma


BangShiuh:
**Problem description**

read_csv does not properly handle a comma in a species name.
Ex.  CH(1,3).

```
gas = ct.Solution("gri30_mod.yaml")
states = ct.SolutionArray(gas)
gas.TPX = 300, ct.one_atm, "O2:1.0"
states.append(gas.state)
states.write_csv("output.csv")
states2 = ct.SolutionArray(gas)
states2.read_csv("output.csv")
```

Line #2 (got "x" columns instead of "x+1")
x is the number of variables.

**System information**

- Cantera version: 3.0.0

**Attachments**

**Additional context**

ischoegl:
@BangShiuh ... I think that csv being a 'comma-separated-list' by definition/default, this behavior is expected. So I do not believe that this is improper handling of output, but a problematic (i.e. difficult-to-parse) choice of species name. 

I'd recommend using a different output format (although even YAML may face similar issues; HDF may not), or replace the comma in the species name with a different character.

bryanwweber:
I'm surprised this isn't handled by the `write_csv` method. You can quote values with commas in CSV files so that this doesn't happen. Doesn't `write_csv` use NumPy or pandas underneath?

ischoegl:
`write_csv` uses the `csv` package, while `read_csv` uses `numpy.genfromtxt` (with `delimiter` set to `','`). I believe there are other internal writers in use beyond what is defined in `composite.py`, so there isn't a uniform approach to CSV. I.e. I don't think that there can be any guarantees for a species name that contains a comma at the moment.

bryanwweber:
I think this is worth fixing... This is a common enough convention for species naming to indicate a particular shape of the molecule, so I don't see "don't use those species" as a long term solution. As I said this is a known/solved problem if you have to embed the delimiter in the value for any *sv type file, so it may just take a little more configuration to resolve on the write side.

ischoegl:
> I think this is worth fixing

I don’t have an issue with this, but would like to see solutions that are not API-centric.

corykinney:
> > I think this is worth fixing
> 
> I don’t have an issue with this, but would like to see solutions that are not API-centric.

Could you clarify what you mean when you say API-centric? 

[...]

PR #1426, opened 2023-01-20
Link C++ and Python SolutionArray implementations

### LINKED ISSUE #137: C++ Equivalent of SolutionArray
**Abstract**

In Python, the `SolutionArray` class offers a convenient way of handling multiple states. Equivalent objects are not available for other API's, where in particular a means for efficient file I/O would be important. While there now is the ability to save to YAML as ASCII text (see Cantera/cantera#1112), this will create unnecessarily large files for large mechanisms.

### Discussed in https://github.com/Cantera/enhancements/discussions/119

<div type='discussions-op-text'>

<sup>Originally posted by **ischoegl** October  6, 2021</sup>
In the Python API, `SolutionArray` objects hold multiple states, which allows for efficient manipulation of time series, spatial profiles etc. The implementation of this feature uses Python (Cython) exclusively, where an equivalent capability does not exist in the C++ core. Beyond manipulation of data sets, `SolutionArray` objects also offers a convenient vehicle to save and load data (e.g. CSV, HDF5), where interoperability between different interfaces would be beneficial.

This discussion is meant to collect some ideas within this context.</div>

**Other Thoughts**

For a potential underlying storage structure, I am also wondering whether the discussion in #91 should be given some thought.

ischoegl:

Follow-up to #1385.

- [x] Expose C++ core `SolutionArray` to Python (as `SolutionArrayBase`)
- [x] Load YAML and HDF `SolutionArray` formats from Python
- [x] Add support for additional `extra` types in C++ `SolutionArray` that are already available in Python, e.g. entries with types `string`, `long int`, `vector<double>` and `vector<long int>` (internal handling leverages `AnyValue`)
- [x] Link C++ and Python versions of `SolutionArray`
- [x] Deprecate Python `h5py` support and associated dedicated functions; continue support for pre-existing HDF format with core C++ HDF routines; disable creation of new HDF files with legacy format.

With this change, all data available for `SolutionArray` are stored by the C++ core object, with the Python `SolutionArray` serving as an external API. At the same time, all property calculations are still handled by the Python interface, although states are being updated in the C++ core. 

Due to the size of the PR, some additional features are postponed for a follow-up:
* Save/restore from CSV data (see #1372)
* Backport selected `SolutionArray.write_hdf` capabilities for `SolutionArray.save`: e.g. default group names
* Implement `SolutionArray::show()` with ascii pandas-style formatting.
* Make Python `SolutionArray` picklable/copyable (via YAML)
* Initial `ReactorNet.save` implementation for single-reactor networks
* Allow for replacing groups in HDF via `H5Ldelete` [see SO post](https://stackoverflow.com/questions/447854/delete-or-update-a-dataset-in-hdf5), which is exposed in HighFive as `unlink` (see [declaration](https://github.com/BlueBrain/HighFive/blob/20be06fe1882cdd2423bc72cf7adbec32c529c24/include/highfive/bits/H5Node_traits_misc.hpp#L273-L280))

Resolves Cantera/enhancements#137

While the way data are handled internally is completely refactored, most changes are completely _under the hood_, with `SolutionArray.save` and `SolutionArray.restore` being an exception. 
```
In [1]: import cantera as ct
   ...: arr = ct.SolutionArray(gas, (5), extra={"spam": [1, 2, 3, 4, 5]})
   ...: arr.save("test.yaml", "entry", description="SolutionArray saved as YAML")
   ...: arr.save("test.h5", "entry", description="SolutionArray saved as HDF")

In [2]: a = ct.SolutionArray(gas)
   ...: a.restore("test.yaml", "entry")
Out[2]:
{'description': 'SolutionArray saved as YAML',
 'generator': 'Cantera SolutionArray',
 'cantera-version': '3.0.0a3',
 'git-commit': "'6b3079996'",
 'date': 'Wed Feb  1 07:13:13 2023'}

In [3]: b = ct.SolutionArray(gas)
   ...: b.restore("test.h5", "entry")
Out[3]:
{'cantera-version': '3.0.0a3',
 'date': 'Wed Feb  1 07:13:13 2023',
 'description': 'SolutionArray saved as HDF',
 'generator': 'Cantera SolutionArray',
 'git-commit': "'6b3079996'"}
```

**Checklist**


ischoegl:
@speth ... at this point this is ready for a review. I spent some time working on a comprehensive test suite (which makes up for a good bit of the line count). It's still a large PR, but I didn't feel like there was a good intermediate stopping point.

ischoegl:
@speth ... merge conflicts are resolved - the PR is ready for a review.

include/cantera/base/SolutionArray.h
```diff
@@ -49,159 +52,277 @@ class SolutionArray
...
+    void resize(size_t size);
+
+    //! SolutionArray shape information used by high-level API's.
+    vector<long int> apiShape() const {
+        return m_apiShape;
+    }
+
+    //! Set SolutionArray shape information used by high-level API's.
+    //! The size of the SolutionArray is adjusted automatically.
+    void setApiShape(const vector<long int>& shape);
```
speth, line 84: This idea of an array shape that doesn't mean anything to `SolutionArray` but is important to the external consumer / API seems like something that could be stored as "metadata" instead.

include/cantera/base/SolutionArray.h
```diff
@@ -49,159 +52,277 @@ class SolutionArray
...
+        return m_apiShape.size();
+    }
+
     //! SolutionArray meta data.
     AnyMap& meta() {
         return m_meta;
     }
 
+    //! Set SolutionArray meta data.
+    void setMeta(const AnyMap& meta) {
```
speth, line 97: Given the `meta()` method already returns a mutable reference to `m_meta`, I think you can always set the metadata by just calling `solnarray.meta() = new_metadata;` without using this method.

include/cantera/base/SolutionArray.h
```diff
@@ -49,159 +52,277 @@ class SolutionArray
...
     //! Retrieve associated ThermoPhase object
     shared_ptr<ThermoPhase> thermo();
 
-    /*!
-     *  Check whether SolutionArray contains a component (property defining state or
-     *  auxiliary variable)
-     */
-    bool hasComponent(const std::string& name) const;
+    //! Retrieve list of component names
+    vector<string> components() const;
```
speth: I'd suggest `componentNames` for this, to better contrast with the other methods like `getComponent` that operate on the actual data.

src/base/SolutionArray.cpp
```diff
@@ -36,30 +35,231 @@ const std::map<std::string, std::string> aliasMap = {
...
+        }
+    }
+    set<int> unique(selected.begin(), selected.end());
+    if (unique.size() < selected.size()) {
+        throw CanteraError("SolutionArray::SolutionArray", "Indices must be unique.");
+    }
+}
+
+template<class T>
+void __resetSingle(AnyValue& extra, const vector<int>& slice);
```
speth: Identifiers containing a double underscore are reserved in C++ and should not be used.

The better way to define "local" functions is within an unnamed namespace (which can be nested within the `Cantera` namespace here or not, it doesn't really matter), e.g.
```c++
namespace {

void resetSingle(...)

} // end unnamed namespace
```

include/cantera/base/SolutionArray.h
```diff
@@ -49,159 +52,277 @@ class SolutionArray
...
+
+    //! Check whether SolutionArray contains a component (property defining state or
+    //! auxiliary variable)
+    bool hasComponent(const string& name) const;
 
     //! Retrieve a component of the SolutionArray by name
-    vector_fp getComponent(const std::string& name) const;
+    AnyValue getComponent(const string& name) const;
```
speth, line 122: Given the opaqueness of the return type, it would be worth documenting that this `AnyValue` contains an array with length `size()` of some type that is particular to the component (and for the "simple" components this is just `double`).

Likewise, for `setComponent`, a bit more explanation of what the input value should contain would make this more clear.

ischoegl:
Rebased and updated to C++17.

ischoegl:
Ugh - tried to fix a minor issue and accidentally pushed an obsolete branch. Will fix once I'm back in my office tomorrow.

ischoegl:
Fixed.

include/cantera/base/AnyMap.h
```diff
@@ -120,10 +120,36 @@ class AnyValue : public AnyBase
     template<class T>
     bool is() const;
 
+    //! Returns `true` if the held value is a vector of the specified type, such as
+    //! `vector<double>`.
+    template<class T>
+    bool isVector() const;
```
speth: Needs `@since ...`

include/cantera/base/AnyMap.h
```diff
@@ -120,10 +120,36 @@ class AnyValue : public AnyBase
...
+    //! and `vector<bool`.
+    //! @since  New in Cantera 3.0.
+    size_t vectorSize() const;
+
+    //! Returns columns of a matrix or zero for a vector; if number of columns is
+    //! not consistent, or the type is not supported, a npos pair is returned.
+    //! Types considered include `vector<vector<double>>`, `vector<vector<long int>>`,
+    //! `vector<vector<string>>` and `vector<vector<bool>>`.
+    //! @since  New in Cantera 3.0.
+    pair<size_t, size_t> matrixShape() const;
```
speth: I found this docstring a bit confusing -- it doesn't clearly specify the "normal" case of returning (rows, cols) for a rectangular matrix. Also what does "... or zero for a vector" mean?

include/cantera/base/AnyMap.inl.h
```diff
@@ -19,10 +21,22 @@ const T &AnyValue::as() const {
             // Implicit conversion of long int to double
             const_cast<AnyValue*>(this)->m_value = static_cast<double>(as<long int>());
             m_equals = eq_comparer<double>;
-        } else if (typeid(T) == typeid(std::vector<double>)
-                   && m_value.type() == typeid(std::vector<AnyValue>)) {
+        } else if (typeid(T) == typeid(string) && m_value.type() == typeid(double)) {
+            // Implicit conversion of double to string
+            std::ostringstream s;
+            s << as<double>();
+            const_cast<AnyValue*>(this)->m_value = s.str();
```
speth:
```suggestion
            // Implicit conversion of double to string
            const_cast<AnyValue*>(this)->m_value = fmt::format("{}", as<double>());
```

include/cantera/base/SolutionArray.h
```diff
@@ -49,159 +52,277 @@ class SolutionArray
...
+     *  @param selected  List of locations for shared entries
+     */
+    static shared_ptr<SolutionArray> share(const shared_ptr<SolutionArray>& other,
+                                           const vector<int>& selected)
+    {
+        return shared_ptr<SolutionArray>(new SolutionArray(other, selected));
+    }
+
+    //! Reset SolutionArray to current Solution state
+    void reset();
```
speth: This docstring is a little unclear -- does it mean all entries of the array get set to the same state?

include/cantera/base/SolutionArray.h
```diff
@@ -49,159 +52,277 @@ class SolutionArray
...
-    static AnyMap readHeader(const AnyMap& root, const std::string& id);
+    static AnyMap readHeader(const AnyMap& root, const string& id);
 
     /*!
      *  Restore SolutionArray entry from a container file.
      *
      *  @param fname  Name of HDF container file
-     *  @param id  Identifier of SolutionArray within the file structure
+     *  @param id  Identifier of root location within the container file
+     *  @param sub  Name identifier for the subgroup holding actual data
```
speth: "Name identifier" seems a bit redundant (for both versions of `readEntry`).

include/cantera/base/Storage.h
```diff
@@ -37,79 +38,81 @@ namespace Cantera
...
+    //!
+    //! Compression is only applied to matrix-type data; note that compression may
+    //! increase file size for small data sets (compression requires setting of chunk
+    //! sizes, which involves considerable overhead for metadata).
     void setCompressionLevel(int level);
 
-    //! Check whether path go location exists
-    //! If the file has write access, create location if necessary
+    //! Check whether location id represents a group
+    bool hasGroup(const string& id) const;
```
speth:
```suggestion
    //! Check whether location `id` represents a group
    bool hasGroup(const string& id) const;
```

include/cantera/base/SolutionArray.h
```diff
@@ -49,159 +52,277 @@ class SolutionArray
...
+     */
+    void _resizeExtra(const string& name, const AnyValue& value=AnyValue());
+
+    /*!
+     *  Set extra SolutionArray component
+     *
+     *  @param name  Name of component (property defining auxiliary variable)
+     *  @param value  Value to be set
+     */
+    void _setExtra(const string& name, const AnyValue& data=AnyValue());
```
speth: Check agreement between name in docstring (`value`) and name in signature (`data`). This (and a few other discrepancies in this file) generate warnings from Doxygen.

include/cantera/base/Storage.h
```diff
@@ -37,79 +38,81 @@ namespace Cantera
...
+
+    //! Check whether path location exists
+    //! If the file has write access, create location; otherwise exceptions are thrown.
     //! @param id  storage location within file
-    bool checkGroup(const std::string& id);
+    //! @param permissive  if true, do not raise exceptions
+    bool checkGroup(const string& id, bool permissive=false);
```
speth: The first line, containing the Doxygen "summary", needs to end with a period or be followed by a blank line. Otherwise, both lines get merged as the summary.

include/cantera/base/Storage.h
```diff
@@ -37,79 +38,81 @@ namespace Cantera
...
 
     //! Retrieve contents of file from a specified location
     //! @param id  storage location within file
     //! @returns  pair containing size and list of entry names of stored data set
-    std::pair<size_t, std::set<std::string>> contents(const std::string& id) const;
+    std::pair<size_t, set<string>> contents(const string& id) const;
```
speth: `std::pair` is also brought into `namespace Cantera` in `ct_defs.h`.

include/cantera/base/Storage.h
```diff
@@ -37,79 +38,81 @@ namespace Cantera
...
-    //! Read matrix from a specified location
+    //! @param name  name of vector/matrix entry
+    //! @param rows  number of vector length or matrix rows
+    //! @param cols  number of matrix columns, if applicable; if 0, a vector is
+    //!     expected, if npos, the size is detected automatically; otherwise, an exact
+    //!     number of columns needs to be matched.
+    //! @returns  matrix or vector containing data; implemented for types
+    //!     `vector<double>`, `vector<long int>`, `vector<string>`,
+    //!     `vector<vector<double>>`, `vector<vector<long int>>` and
+    //!     `vector<vector<lstring>>`
```
speth:
```suggestion
    //!     `vector<vector<string>>`
```

include/cantera/base/Storage.h
```diff
@@ -37,79 +38,81 @@ namespace Cantera
...
-    //! @param id  storage location within file
-    //! @param name  name of matrix entry
-    //! @param data  matrix containing data (vector of vectors)
-    void writeMatrix(const std::string& id,
-                     const std::string& name, const std::vector<vector_fp>& data);
+    //! @param data  vector or matrix containing data; implemented for types
+    //!     `vector<double>`, `vector<long int>`, `vector<string>`
+    //!     `vector<vector<double>>`, `vector<vector<long int>>` and
+    //!     `vector<vector<string>>`
+    void writeData(const string& id, const string& name, const AnyValue& data);
```
speth: The phrase "for data type double" seems to contradict the set of allowed types that includes vectors/matrices of strings.

interfaces/cython/cantera/solutionbase.pyx
```diff
@@ -485,3 +485,166 @@ cdef object _wrap_Solution(shared_ptr[CxxSolution] cxx_soln):
...
+        dest.base = dest._base.get()
+        return dest
+
+    @property
+    def size(self):
+        """ The number of elements in the `SolutionArrayBase`. """
+        return self.base.size()
+
+    def _api_shape(self):
+        """ Retrieve shape information avalailable in C++ core. """
```
speth: "available"

interfaces/cython/cantera/composite.py
```diff
@@ -448,20 +448,20 @@ class SolutionArray:
         >>> states.read_csv('somefile.csv')
 
     As an alternative to comma separated export and import, data extracted from
-    `SolutionArray` objects can also be saved to and restored from a HDF
-    container file using the `write_hdf`::
+    `SolutionArray` objects can also be saved to and restored from YAML and HDF
+    container file using the `save`::
```
```suggestion
    `SolutionArray` objects can also be saved to and restored from YAML and HDF
    container files using the `save` function::
```

interfaces/cython/cantera/composite.py
```diff
@@ -699,15 +694,68 @@ def shape(self) -> tuple[int, ...]:
...
 
-        .. versionadded:: 3.0
+        Handle transition from member variable to property.
+
+        .. deprecated:: 3.0
+
+            Property to be removed after Cantera 3.0; use `shape` instead.
+        """
+        warnings.warn("Property '_shape' to be removed after Cantera 3.0; use property "
+            "'shape' instead.", DeprecationWarning)
```
speth: I don't think a deprecation warning is necessary before removing Python members that start with underscores. (for any of `_shape`, `_states`, or `_extra`) -- these should be regarded as implementation details.

interfaces/cython/cantera/composite.py
```diff
@@ -1253,6 +1272,52 @@ def from_pandas(self, df, normalize=True):
...
+        :param name:
+            Identifier of root location within the container file; the root location
+            contains header data and a subgroup holding the actual `SolutionArray`.
+        :param key:
+            Name identifier for the subgroup holding the `SolutionArray` data and
+            metadata objects. If `None`, the subgroup name default to ``data``.
+        :param description:
+            Custom comment describing the dataset to be stored.
+        :param compression:
+            Compression level; optional (default=0; HDF only)
```
speth: Maybe worth mentioning that the range is 0-9.

interfaces/cython/cantera/composite.py
```diff
@@ -1320,63 +1385,20 @@ def write_hdf(self, filename, *args, cols=None, group=None, subgroup=None,
...
-                dgroup.attrs[key] = val
-            for header, value in data.items():
-                if value.dtype.type == np.str_:
-                    dgroup.create_dataset(header, data=value.astype('S'),
-                                          **hdf_kwargs)
-                else:
-                    dgroup.create_dataset(header, data=value, **hdf_kwargs)
+            Method to be removed after Cantera 3.0; use `save` instead. Note that
+            the call is redirected to `save` in order to prevent the creation of a file
+            with deprecated HDF format.
```
speth: The wording here "deprecated HDF format" might be misconstrued to mean that _all_ HDF output is deprecated.

interfaces/cython/cantera/onedim.py
```diff
@@ -127,13 +127,18 @@ def set_initial_guess(self, *args, data=None, group=None, **kwargs):
...
-            if data.endswith('.hdf5') or data.endswith('.h5'):
+            arr = SolutionArray(self.gas, extra=self.other_components())
+            if any(data.endswith(suffix) for suffix in [".hdf5", ".h5", ".hdf"]):
+                # data source identifies a HDF file
+                if "native" in hdf_support():
+                    arr.restore(data, name=group, key=self.domains[1].name)
+                else:
+                    arr.read_hdf(data, group=group, subgroup=self.domains[1].name)
+            elif data.endswith(".yaml") or data.endswith(".yml"):
                 # data source identifies a HDF file
```
speth: Comment 
```suggestion
            elif data.endswith(".yaml") or data.endswith(".yml"):
                # data source identifies a YAML file
```

interfaces/cython/cantera/onedim.py
```diff
@@ -127,13 +127,18 @@ def set_initial_guess(self, *args, data=None, group=None, **kwargs):
...
-                arr = SolutionArray(self.gas, extra=self.other_components())
-                arr.read_hdf(data, group=group, subgroup=self.domains[1].name)
+                arr.restore(data, name=group, key=self.domains[1].name)
             elif data.endswith('.csv'):
                 # data source identifies a CSV file
-                arr = SolutionArray(self.gas, extra=self.other_components())
                 arr.read_csv(data)
             else:
                 raise ValueError(f"'{data}' does not identify CSV or HDF file.")
```
speth: Error message should be updated to acknowledge the possibility of YAML input.

interfaces/cython/cantera/onedim.py
```diff
@@ -161,10 +166,10 @@ def set_initial_guess(self, *args, data=None, group=None, **kwargs):
...
 
             self.gas.TPY = left.T, self.P, left.Y
-            u[:i] = u[:i] * left.mdot / self.gas.density / u[0]
+            arr[:i].velocity = u[:i] * left.mdot / self.gas.density / u[0]
 
             self.gas.TPY = right.T, self.P, right.Y
-            u[i:] = - u[i:] * right.mdot / self.gas.density / u[-1]
+            arr[i:].velocity = - u[i:] * right.mdot / self.gas.density / u[-1]
 
             arr.velocity = u
```
speth: With this change, setting `arr.velocity = u` should be unnecessary, right?

interfaces/cython/cantera/solutionbase.pyx
```diff
@@ -485,3 +485,166 @@ cdef object _wrap_Solution(shared_ptr[CxxSolution] cxx_soln):
...
+
+    **Warning:** this class is an experimental part of the Cantera API and
+    may be changed or removed without notice.
+
+    .. versionadded:: 3.0
+    """
+    _phase = None
+
+    def __cinit__(self, _SolutionBase phase, shape=(0,),
+                  states=None, extra=None, meta={}, init=True):
```
speth: I'm pretty sure it's not actually being mis-used here, but the default initialization to a mutable value (`meta={}`) always sets off alarm bells.

interfaces/cython/cantera/composite.py
```diff
@@ -1253,6 +1272,52 @@ def from_pandas(self, df, normalize=True):
...
+        :param description:
+            Custom comment describing the dataset to be stored.
+        :param compression:
+            Compression level; optional (default=0; HDF only)
+        :return:
+            Group identifier used for storing HDF data.
+
+        .. versionadded:: 3.0
+        """
+        return self._cxx_save(fname, name, key, description, compression)
```
speth, line 1245: If this is all this method does, and this is the only place `_cxx_save` is called, would it make sense to move this method up to `SolutionArrayBase` and avoid the need for `_cxx_save`. Likewise for `_cxx_restore`.

interfaces/cython/cantera/solutionbase.pyx
```diff
@@ -485,3 +485,166 @@ cdef object _wrap_Solution(shared_ptr[CxxSolution] cxx_soln):
...
+            cxx_data.push_back(item)
+        self.base.setState(loc, cxx_data)
+
+    def _list_extra(self, all=True):
+        """ Retrieve list of `SolutionArrayBase` extra components """
+        cdef vector[string] cxx_name = self.base.listExtra(all)
+        out = []
+        for item in cxx_name:
+            out.append(pystr(item))
+        return out
```
speth: Can't this just be the implementation of the `extra` property?

interfaces/cython/cantera/solutionbase.pyx
```diff
@@ -485,3 +485,166 @@ cdef object _wrap_Solution(shared_ptr[CxxSolution] cxx_soln):
...
+        """ Add component to `SolutionArrayBase` and initialize to default value """
+        self.base.addExtra(stringify(name), back)
+
+    def _get_auxiliary(self, loc):
+        """ Retrieve auxiliary data for a `SolutionArrayBase` location """
+        return anymap_to_dict(self.base.getAuxiliary(loc))
+
+    def _set_auxiliary(self, loc, data):
+        """ Set auxiliary data for a `SolutionArrayBase` location """
+        self.base.setAuxiliary(loc, dict_to_anymap(data))
```
speth: These two methods don't seem to end up getting used.

src/base/AnyMap.cpp
```diff
@@ -641,6 +641,66 @@ bool AnyValue::isScalar() const {
...
+    }
+    return npos;
+}
+
+pair<size_t, size_t> AnyValue::matrixShape() const {
+    if (isVector<vector<double>>()) {
+        auto& mat = as<vector<vector<double>>>();
+        if (isMatrix<double>()) {
+            if (mat.size()) {
+                return pair<size_t, size_t>(mat.size(), mat[0].size());
```
speth: I think here and in all following instances, you can just write:
```suggestion
                return {mat.size(), mat[0].size()};
```

src/base/SolutionArray.cpp
```diff
@@ -36,30 +36,229 @@ const std::map<std::string, std::string> aliasMap = {
...
 {
     if (!m_sol) {
         throw CanteraError("SolutionArray::SolutionArray",
             "Unable to create SolutionArray from invalid Solution object.");
     }
     m_stride = m_sol->thermo()->stateSize();
-    m_data.resize(m_size * m_stride, 0.);
+    m_data.reset(new vector<double>(m_dataSize * m_stride, 0.));
+    m_extra.reset(new map<string, AnyValue>());
+    m_order.reset(new map<int, string>());
```
speth: In line with some style preferences introduced in #1452:
```suggestion
    m_data = make_shared<vector<double>>(m_dataSize * m_stride, 0.);
    m_extra = make_shared<map<string, AnyValue>>();
    m_order = make_shared<map<int, string>>();
```

src/base/SolutionArray.cpp
```diff
@@ -36,30 +36,229 @@ const std::map<std::string, std::string> aliasMap = {
...
+template<class T>
+void __resetMulti(AnyValue& extra, const vector<int>& slice);
+
+void SolutionArray::reset()
+{
+    size_t nState = m_sol->thermo()->stateSize();
+    vector<double> state(nState);
+    m_sol->thermo()->saveState(state); // thermo contains current state
+    for (int k = 0; k < m_size; ++k) {
+        std::copy(state.begin(), state.end(), &(*m_data)[m_active[k] * m_stride]);
```
speth: An alternative that might be slightly less clunky (here and in several other locations):
```suggestion
        std::copy(state.begin(), state.end(), m_data->data() + m_active[k] * m_stride);
```

src/base/SolutionArray.cpp
```diff
@@ -36,30 +36,229 @@ const std::map<std::string, std::string> aliasMap = {
...
+                key, extra.type_str());
+        }
+    }
+}
+
+void SolutionArray::resize(size_t size)
+{
+    if (apiNdim() > 1) {
+        throw CanteraError("SolutionArray::resize",
+            "Resize is ambiguous for multi-dimensional arrays; use reshape instead.");
```
speth: "reshape" doesn't exist, does it?

src/base/SolutionArray.cpp
```diff
@@ -404,37 +815,200 @@ AnyMap SolutionArray::readHeader(const AnyMap& root, const std::string& id)
...
+            "'h5'/'hdf'/'hdf5' and 'yml'/'yaml'.", extension);
+    }
+    return header;
+}
+
+void SolutionArray::_initExtra(const string& name, const AnyValue& value)
+{
+    if (!m_extra->count(name)) {
+        throw CanteraError("SolutionArray::_initExtra",
+            "Component '{}' does not exists.", name);
```
speth: This appears a few times:
```suggestion
            "Component '{}' does not exist.", name);
```

src/base/SolutionArray.cpp
```diff
@@ -553,83 +1219,132 @@ void SolutionArray::readEntry(const std::string& fname, const std::string& id)
...
+            } else {
+                addExtra(name, back);
+                AnyValue data;
+                data = file.readData(path, name, m_dataSize);
+                setComponent(name, data);
+            }
+        }
+        m_meta.erase("components");
+    } else {
+        // data format used by Python h5py export (Cantera 2.5)
```
speth, line 1242: Should this raise a warning? Or do we want to support this format indefinitely?

src/base/SolutionArray.cpp
```diff
@@ -638,8 +1353,8 @@ void SolutionArray::readEntry(const AnyMap& root, const std::string& id)
 
         // check that state data are complete
         const auto& nativeState = m_sol->thermo()->nativeState();
-        std::set<std::string> props = {};
-        std::set<std::string> missingProps = {};
+        set<string> props = {};
+        set<string> missingProps = {};
```
speth: The `{}` initializers are unnecessary.

src/base/Storage.cpp
```diff
@@ -284,115 +345,215 @@ void writeH5Attributes(h5::Group sub, const AnyMap& meta)
...
     try {
+        checkGroupWrite(id, false);
+        h5::Group sub = m_file->getGroup(id);
         writeH5Attributes(sub, meta);
     } catch (const Cantera::NotImplementedError& err) {
         throw NotImplementedError("Storage::writeAttribute",
             "{} in group '{}'.", err.getMessage(), id);
+    } catch (const CanteraError& err) {
+        // rethrow with public method attribution
+        throw CanteraError("Storage::writeData", "{}", err.getMessage());
```
speth: I think I understand the comment, but shouldn't this reference "writeAttributes"?

src/oneD/Sim1D.cpp
```diff
@@ -216,13 +216,14 @@ AnyMap legacyH5(shared_ptr<SolutionArray> arr, const AnyMap& header={})
 
     if (header.hasKey("fixed_temperature")) {
         double temp = header.getDouble("fixed_temperature", -1.);
-        auto profile = arr->getComponent("T");
+        auto profile = arr->getComponent("T").as<std::vector<double>>();
```
speth: Here and below
```suggestion
        auto profile = arr->getComponent("T").asVector<double>();
```

test/general/test_composite.cpp
```diff
@@ -27,3 +28,436 @@ TEST(Interface, from_AnyMap_and_Solution)
...
+    AnyValue any;
+    any = arr->getComponent("spam");
+    ASSERT_TRUE(any.is<void>());
+
+    any = 1;
+    // arr->setComponent("spam", any);
+    ASSERT_THROW(arr->setComponent("spam", any), CanteraError);
+    any = vector<double>({1});
+    // arr->setComponent("spam", any);
+    ASSERT_THROW(arr->setComponent("spam", any), CanteraError);
```
speth: The commented out lines can be deleted.

test/python/test_composite.py
```diff
@@ -216,6 +219,51 @@ def test_empty_transport(self):
             ct.Transport()
 
 
+class TestSolutionArray(utilities.CanteraTest):
+    """ Test SolutionArray file IO """
```
speth: This is actually the tests that _don't_ do any file IO

test/python/test_composite.py
```diff
@@ -342,85 +388,307 @@ def test_to_pandas(self):
...
+        outfile.unlink(missing_ok=True)
 
         extra = {'foo': range(7), 'bar': range(7)}
         meta = {'spam': 'eggs', 'hello': 'world'}
         states = ct.SolutionArray(self.gas, 7, extra=extra, meta=meta)
         states.TPX = np.linspace(300, 1000, 7), 2e5, 'H2:0.5, O2:0.4'
         states.equilibrate('HP')
 
-        states.write_hdf(outfile, attrs={'foobar': 'spam and eggs'})
+        states.save(outfile, "group0") # , attrs={'foobar': 'spam and eggs'}
```
speth: I assume this comment should be deleted? Also the block of commented-out code below.

test/python/test_composite.py
```diff
@@ -342,85 +388,307 @@ def test_to_pandas(self):
...
+        assert b.shape == states.shape
+
+class TestLegacyHDF(utilities.CanteraTest):
+    # Test SolutionArray legacy HDF file input
+    #
+    # All input files were created using the Cantera 2.6 Python test suite:
+    # - solutionarray_fancy_legacy.h5 -> test_composite.py::TestSolutionArrayIO::test_write_hdf
+    # - solutionarray_str_legacy.h5 -> test_composite.py::TestSolutionArrayIO::test_write_hdf_str_column
+    # - solutionarray_multi_legacy.h5 -> test_composite.py::TestSolutionArrayIO::test_write_hdf_multi_column
+    # - solutionarray_no_norm_legay.h5 -> test_composite.py::TestSolutionArrayIO::test_import_no_norm_data
```
speth:
```suggestion
    # - solutionarray_no_norm_legacy.h5 -> test_composite.py::TestSolutionArrayIO::test_import_no_norm_data
```

test/python/test_composite.py
```diff
@@ -342,85 +388,307 @@ def test_to_pandas(self):
...
+            arr.read_hdf(infile, group='group1')
+        with pytest.raises(IOError, match='does not contain data'):
+            arr.read_hdf(infile, subgroup='foo')
+
+    @pytest.mark.skipif("native" not in ct.hdf_support(), reason="Cantera compiled without HDF support")
+    def test_legacy_hdf_exceptions(self):
+        infile = self.test_data_path / f"solutionarray_fancy_legacy.h5"
+        # infile contains 'group0' with h2o2 data, 'group1' with gri30 data, and
+        # an empty group 'spam'
+        # TOOD: resolve remaining edge
```
speth: What is covered by this "TODO"? All the commented out cases? Or is this out of date?

test/python/test_onedim.py
```diff
@@ -1327,100 +1402,72 @@ def test_reacting_surface_case2(self):
...
-        # save and restore legacy h5py format
-        self.run_impingingjet_write("legacy")
+    def test_restore_legacy_hdf_h5py(self):
+        filename = self.test_data_path / f"impingingjet_legacy.h5"
+        jet = ct.ImpingingJet(gas=self.gas, surface=self.surf_phase)
+        with pytest.raises(IOError, match="Unable to load surface phase"):
+            # legacy HDF format uses TDX state information, which is incomplete for surfaces
+            jet.read_hdf(filename)
+
+    @pytest.mark.skipif("native" not in ct.hdf_support(), reason="Cantera compiled without HDF support")
```
speth: Please keep to the 88-character line length in this file.

speth (changes requested):
Thanks, @ischoegl, for putting together this substantial bit of work. I think having a unified `SolutionArray` implementation starting in C++ will simplify bringing this feature to other language interfaces. I was pleased to see that the net effect for the Python version was a modest speedup -- about 20% for some simple tests I ran.

Overall, I think this implementation looks very good. I think all of my comments below are regarding fairly minor issues.

interfaces/cython/cantera/composite.py
ischoegl: True. I mainly implemented those to find instances where it was used. While it may break user code, it can probably be removed without deprecation?

src/base/SolutionArray.cpp
ischoegl: Adding a `user_warning`. I think that we should support the legacy format indefinitely - users may have stored HDF data in legacy format; not being able to read it after Cantera 3.0 would be problematic.

As an aside, renaming the transport model from `Mix` to `mixture-averaged` raises a separate deprecation warning. Some of the deprecated default names should probably be kept around (e.g. by using a `addLegacyAlias` rather than a `addDeprecatedAlias`).

test/python/test_composite.py
ischoegl: After removing `h5py` to write data, this is no longer needed.

interfaces/cython/cantera/composite.py
ischoegl: ~Good point. Originally, `SolutionArrayBase` was only intended to be a thin interface; that said, there's no reason not to move this.~

On second thought, `restore` needs to update a couple of things within `SolutionArray` (`self._indices` and `self._output_dummy`), so it's not possible to move `restore`. For reasons of symmetry, I'm leaving things as they are.

include/cantera/base/SolutionArray.h
ischoegl: I see the point, but at the same time I view 'metadata' as something that is left to the user completely (within the group key / outside of the HDF header). Shape is the only 'standardized' information we keep within the groups (although technically, there's also the `phase` group), so I'd prefer to leave this as is.

include/cantera/base/SolutionArray.h
ischoegl: Cython balks at trying to do this: `Cannot assign to or delete this` (I declared this as `CxxAnyMap& meta()` in the `solutionbase.pxd`, and used the syntax proposed above for the assignment)

src/base/SolutionArray.cpp
ischoegl: Thanks for pointing this out. To keep things simple, I'm just using a single underscore.

interfaces/cython/cantera/solutionbase.pyx
ischoegl: promoting them to `get_auxiliary` and `set_auxiliary` (without underscore), plus adding a test.

interfaces/cython/cantera/composite.py
ischoegl: I removed these hidden properties in a stand-alone commit.

ischoegl:
@speth - thank you for the review! I believe that everything is addressed; I also rebased to #1455.

include/cantera/base/SolutionArray.h
speth: Oh, interesting. The first part suggests declaring the assignment operator for `AnyMap` in the `.pxd` header, but I'm not sure whether that would resolve the "or delete" aspect of the error. Either way, not a major issue.

src/base/SolutionArray.cpp
speth: That's a reasonable distinction -- user code is something that needs to be continuously updated to keep pace with new Cantera versions, but it would be a different thing to say that you need to update your data files (without a specific migration path as in the case of CTI/XML to YAML).

I wasn't necessarily planning on removing the deprecated model names anytime soon, and the warning doesn't specify that these will be removed after a particular Cantera version, unlike our warnings for deprecated functions/classes.

include/cantera/base/SolutionArray.h
```diff
@@ -49,159 +52,285 @@ class SolutionArray
...
+     *  Share locations from an existing SolutionArray and return new reference.
+     *
+     *  Both SolutionArray object share common data. The method is used for slicing
+     *  of SolutionArrays from high-level API's. Note that meta data are not inherited.
+     *  @param other  SolutionArray object containing shared data
+     *  @param selected  List of locations for shared entries
+     */
+    static shared_ptr<SolutionArray> share(const shared_ptr<SolutionArray>& other,
+                                           const vector<int>& selected)
+    {
```
speth: Would this make more sense as a member function?

src/base/SolutionArray.cpp
speth: If you want to prefix with a single underscore, that's fine (though not exactly a pattern we follow elsewhere for local functions in C++). However, these functions should still be defined inside an unnamed namespace to restrict their scope to this translation unit (that is, when compiling `SolutionArray.cpp`).

speth (changes requested):
Thanks for the updates, @ischoegl. I have just a couple of remaining comments.

include/cantera/base/SolutionArray.h
ischoegl: ~I guess it depends on whether `enable_shared_from_this` is preferable or change the constructor that is being called to accept references ... my own preference is the `static` function here. It probably won’t be used by itself (it’s a helper function for slicing) so I’d prefer to leave this as is.~

Upon further consideration, I'm adopting the suggestion.

src/base/SolutionArray.cpp
ischoegl: No need to keep a single underscore. I finally figured out why the unnamed namespace didn't compile last time I tried. This time it worked :tada: ...

ischoegl:
@speth ... I adopted both suggestions/requests.

speth (approved):
Thanks, @ischoegl! I think this is good to go. I'm going to ignore the transient failure of the "coverage" builder, as we already have the coverage testing results.

PR #1429, opened 2023-01-24
Link Python module to Cantera shared library

### LINKED ISSUE #140: Export C++ symbols in dynamic Windows libraries
**Abstract**

At the moment, symbols in dynamic libraries are not exported in Windows, i.e.
```shell
> dumpbin /exports cantera_shared.dll
Microsoft (R) COFF/PE Dumper Version 14.31.31104.0
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file cantera_shared.dll

File Type: DLL

  Summary

        1000 .00cfg
       19000 .data
        5000 .idata
       40000 .pdata
       F9000 .rdata
        B000 .reloc
      5E3000 .text
        1000 .tls
```
does not list any symbols (whereas running `dumpbin /LINKERMEMBER cantera_shared.lib` for the static library lists an extensive list of symbols - at least for the C API?)

**Motivation**

* What problem is it trying to solve? ... allow access to C++ methods in shared dynamic libraries
* Why is this a good solution? ... make shared dynamic libraries usable on Windows platforms

The ability of accessing Cantera C++ classes and functions at run-time will open additional pathways for the definition of alternative API's and uses in 3rd party applications.

**Possible Solutions**

Per [SO answer](https://stackoverflow.com/questions/45258925/c-declspec-static-libraries-and-dlls), a platform-independent preprocessor macro
```C++
// If we are Microsoft C/C++ Compiler
#if defined(_MSC_VER)
    #if defined(DLL_EXPORT)
        #define API __declspec(dllexport)
    #else
        #define API __declspec(dllimport)
    #endif

// If we are non Windows (Export by default) or compiling to a StaticLibrary
#else
    #define API 
#endif
```
will properly export symbols for classes / functions that include the `API` decoration. *Note: this is just copy-paste from SO, and thus is untested / may need tweaks. There are likely parts of the* SCons *build that will have to be adjusted.*

**References**

* #97 
* #102

speth:

The main purpose of this PR is to modify the implementation of `ExtensibleRate` so that the core Cantera library (`libcantera`) does not need to depend on a specific Python installation at compilation / installation time. Achieving this required some fairly far-reaching changes, though I think they will all have long-term benefits.

- The `PythonExtensionManager` class and associated Cython functions have been moved out of `libcantera` into a separate shim, `libcantera_python` that is loaded on demand. This new library is the one that links to `libpython` and loads the Cantera Python module.
- Because a user C++ program and the Cantera Python module (loaded via `libcantera_python`) both depend on `libcantera`, `libcantera` needs to be loaded as a shared library to avoid having two copies of the library with separate state (which would interfere instantiating user defined types that are only instance with one copy of `ReactionRateFactory`, for example). Multiple versions of Python are supported (not at the same time!) by having multiple versions of `libcantera_python`, e.g. `libcantera_python3_11.so` and `libcantera_python3_10.so`.
- On Windows, linking C++ applications and the Cantera Python module to `libcantera_shared` requires "exporting" the complete C++ interface. Due the the sheer number of methods that need to be exported, rather than the standard approach investigated in #1298, this PR uses the static library to generate a list of symbols to be exported from the DLL.
- The functions to load the `libcantera_python` library on-demand are OS-specific. Rather than use these low-level functions, the Boost.DLL library is used for its higher-level, cross-platform capabilities. While Boost.DLL is header-only, by default it depends on Boost.Filesystem, which has a compiled component. However, since C++17 introduced `std::filesystem` (derived from Boost.Filesystem), Boost.DLL provides an option to use `std::filesystem`. Therefore, in order to avoid introducing a dependence on the compiled components of Boost, **this PR bumps Cantera's compiler requirement to the C++17 standard**.

There are also a few more minor changes that ended up coming along for the ride:
- Add `using` directives within the `Cantera` namespace for some of the most-used standard library types: `string`, `vector`, `map`, `function`, `pair`, `unique_ptr`. I figure `map<string, string>` is more readable than `std::map<std::string, std::string>`.
- Bump vendored versions of `fmt` and `Eigen`.
- Move some function definitions out of the header files, especially in the case of the Factory classes.

- Continues the work on https://github.com/Cantera/enhancements/issues/79
- Provides a path to resolving https://github.com/Cantera/conda-recipes/issues/40
- Resolves https://github.com/Cantera/enhancements/issues/140
- The underlying functionality here should enable implementation of https://github.com/Cantera/enhancements/issues/126

**Checklist**

### DIFF
--- a/include/cantera/base/ct_defs.h
+++ b/include/cantera/base/ct_defs.h
@@ -22,9 +22,11 @@
 #include <cstdlib>
 #include <vector>
 #include <map>
+#include <set>
 #include <string>
 #include <algorithm>
 #include <memory>
+#include <functional>
 
 /**
  * Namespace for the Cantera kernel.
@@ -36,6 +38,12 @@ using std::shared_ptr;
 using std::make_shared;
 using std::unique_ptr;
 using std::isnan; // workaround for bug in libstdc++ 4.8
+using std::string;
+using std::vector;
+using std::map;
+using std::set;
+using std::function;
+using std::pair;
 
 /*!
  * All physical constants are stored here.
@@ -159,20 +167,20 @@ const double Tiny = 1.e-20;
 /*!
  * This is used mostly to assign concentrations and mole fractions to species.
  */
-typedef std::map<std::string, double> compositionMap;
+typedef map<string, double> compositionMap;
 
 //! Map from string names to doubles. Used for defining species mole/mass
 //! fractions, elemental compositions, and reaction stoichiometries.
-typedef std::map<std::string, double> Composition;
+typedef map<string, double> Composition;
 
 //! Turn on the use of stl vectors for the basic array type within cantera
 //! Vector of doubles.
-typedef std::vector<double> vector_fp;
+typedef vector<double> vector_fp;
 //! Vector of ints
-typedef std::vector<int> vector_int;
+typedef vector<int> vector_int;
 
 //! A grouplist is a vector of groups of species
-typedef std::vector<std::vector<size_t> > grouplist_t;
+typedef vector<vector<size_t>> grouplist_t;
 
 //! index returned by functions to indicate "no position"
 const size_t npos = static_cast<size_t>(-1);
--- a/include/cantera/base/global.h
+++ b/include/cantera/base/global.h
@@ -87,6 +87,11 @@ void loadExtension(const std::string& extType, const std::string& name);
 //! @since New in Cantera 3.0
 void loadExtensions(const AnyMap& node);
 
+//! Returns `true` if Cantera was loaded as a shared library in the current
+//! application. Returns `false` if it was statically linked.
+//! @since New in Cantera 3.0
+bool usingSharedLibrary();
+
 //! Delete and free all memory associated with the application
 /*!
  * Delete all global data.  It should be called at the end of the
--- a/include/cantera/base/global.h
+++ b/include/cantera/base/global.h
@@ -87,6 +87,9 @@ void loadExtension(const std::string& extType, const std::string& name);
 //! @since New in Cantera 3.0
 void loadExtensions(const AnyMap& node);
 
+//! @copydoc Application::searchPythonVersions
+void searchPythonVersions(const string& versions);
+
 //! Returns `true` if Cantera was loaded as a shared library in the current
 //! application. Returns `false` if it was statically linked.
 //! @since New in Cantera 3.0
--- a/include/cantera/base/global.h
+++ b/include/cantera/base/global.h
@@ -102,6 +102,11 @@ void appdelete();
 //! @copydoc Application::thread_complete
 void thread_complete();
 
+//! Returns the Cantera version. This function when needing to access the version from a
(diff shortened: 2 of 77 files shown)


include/cantera/base/ct_defs.h
```diff
@@ -36,6 +37,12 @@ using std::shared_ptr;
 using std::make_shared;
 using std::unique_ptr;
 using std::isnan; // workaround for bug in libstdc++ 4.8
+using std::string;
+using std::vector;
+using std::map;
+using std::function;
+using std::pair;
+using std::unique_ptr;
```
ischoegl: `unique_ptr` is already added above 😄 

Could we add
```C++
using std::set;
```
also?

include/cantera/base/global.h
```diff
@@ -87,6 +87,13 @@ void loadExtension(const std::string& extType, const std::string& name);
 //! @since New in Cantera 3.0
 void loadExtensions(const AnyMap& node);
 
+//! @copydoc Application::setPythonSearchVersions
+void searchPythonVersions(const string& versions);
+
+//! Returns `true` if Cantera was loaded as a shared library in the current
+//! application. Returns `false` if it was statically linked.
+bool usingSharedLibrary();
```
ischoegl, line 96: Should receive a `@since` tag; same for `version` below. There are likely a couple of other instances.

src/pch/system.h
```diff
@@ -12,9 +12,8 @@
 #include <string>
 #include <algorithm>
 #include <memory>
+#include <functional>
```
ischoegl, line 16: Should there be a couple of other standard includes? (those who were previously included in `AnyMap.h`, e.g.
```C++
#include <unordered_map>
#include <set>
#include <boost/any.hpp>
```
(for `any`, there are obviously alternatives in C++17, but that's well beyond this PR)

src/base/application.h
```diff
@@ -432,6 +437,9 @@ class Application
     //! Current vector of input directories to search for input files
     std::vector<std::string> inputDirs;
 
+    //! Versions of Python to consider when attempting to load user extensions
+    vector<string> m_pythonSearchVersions = {"3.11", "3.10", "3.9", "3.8"};
```
ischoegl, line 442: what about making this `set<string>` instead? When keeping order doesn't matter, sets are much more convenient.

ischoegl (changes requested):
@speth ... thanks for spending time on this gnarly (and elusive) feature! Changes look good to me; I only have a couple of minor suggestions.

src/base/application.h
speth: The order here isn't irrelevant, though. I want to start with the most recent supported Python version and only look for older versions if the first one doesn't work.

src/pch/system.h
speth: I added `<set>` since it's now in `ct_defs.h` for the sake of being able to add the Cantera-wide `using`. However, I didn't see any compilation speedup for adding the other two headers here.

I expect to give `std::any` a shot in a future PR also making use of some other C++17 features. Perhaps it would be worth checking again at that point whether including `<any>` in the PCH helps at all.

ischoegl (approved):
🎉

PR #1182, opened 2022-01-21
Matlab toolbox revamp

ssun30:
**Changes proposed in this pull request**

- Change the Matlab interface to the modern Matlab structure and syntaxes for OOP. 
- Replace MEX interface with directly calling from Cantera Clib. 
- Re-adapt examples to the new interface.

**If applicable, fill in the issue number this pull request is fixing**

Closes Cantera/enhancements#32

**Checklist**

- [x] The pull request includes a clear description of this code change
- [ ] Commit messages have short titles and reference relevant issues
- [ ] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

PR #1433, opened 2023-02-09
Deprecate 'R' in C++ setter nomenclature in favor of 'D'

ischoegl:

While Cantera's Python nomenclature uses `TD`/`TDX`/`TDY` and `DP`/`DPX`/`DPY`, other API's use the older (?) nomenclature `TR`/`TRX`/`TRY` and `RP`/`RPX`/`RPY`. This PR seeks to implement a uniform nomenclature.

In addition, a new 
```C++
    CANTERA_CAPI int thermo_set_TD(int n, double* vals);
```
is added to `ct.h`, which is needed for work in #1182.

**Checklist**

### DIFF
--- a/interfaces/matlab/toolbox/@ThermoPhase/setState_DP.m
+++ b/interfaces/matlab/toolbox/@ThermoPhase/setState_DP.m
@@ -1,6 +1,6 @@
-function setState_RP(tp, rp)
-% SETSTATE_RP  Set the density and pressure.
-% setState_RP(tp, [density,p])
+function setState_DP(tp, dp)
+% SETSTATE_DP  Set the density and pressure.
+% setState_DP(tp, [density,p])
 % The density is set first, then the pressure is set by
 % changing the temperature holding the density and
 % chemical composition fixed.
@@ -8,9 +8,9 @@ function setState_RP(tp, rp)
 % :param tp:
 %     Instance of class :mat:func:`ThermoPhase` (or another
 %     class derived from ThermoPhase)
-% :param rp:
+% :param dp:
 %     Vector of length 2 containing the desired values for the density (kg/m^3)
 %     and pressure (Pa)
 %
-warning('To be removed after Cantera 3.0. Renamed to setState_DP.')
-thermo_set(tp.tp_id, 26, rp);
+
+thermo_set(tp.tp_id, 26, dp);
--- a/src/thermo/PDSS.cpp
+++ b/src/thermo/PDSS.cpp
@@ -8,6 +8,7 @@
 // This file is part of Cantera. See License.txt in the top-level directory or
 // at https://cantera.org/license.txt for license and copyright information.
 
+#include "cantera/base/global.h"
 #include "cantera/thermo/PDSS.h"
 #include "cantera/thermo/VPStandardStateTP.h"
 
@@ -184,7 +185,14 @@ void PDSS::setState_TP(doublereal temp, doublereal pres)
 
 void PDSS::setState_TR(doublereal temp, doublereal rho)
 {
-    throw NotImplementedError("PDSS::setState_TR");
+    warn_deprecated("PDSS::setState_TR",
+        "To be removed after Cantera 3.0. Renamed to setState_TD.");
+    setState_TD(temp, rho);
+}
+
+void PDSS::setState_TD(double temp, double rho)
+{
+    throw NotImplementedError("PDSS::setState_TD");
 }
 
 doublereal PDSS::satPressure(doublereal t)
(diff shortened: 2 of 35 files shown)


speth:
I appreciate the attempt to improve the consistency of the different interfaces.

I'm wondering about the usefulness of this whole family of functions, though -- basically, `setState_DP`, `setState_DPX`, and `setState_DPY`, with the latter two having three variants each depending on how the mass/mole fractions are specified. I note the following:
- Besides `setState_TPX` and `setState_TPY`, these are the only "full-state" setters that we provide in C++.
- The only thermo model that implements `setState_DP` is `IdealGasPhase`.
- The `setState_PV` method is essentially equivalent, but this method is implemented only for `PureFluidPhase`.
- I'm having trouble coming up with the physical situation where this is the property pair that you know and that the thing that needs to be found is the temperature. It seems like a pretty unusual situation.
- This was all added by @bryanwweber some time ago, so maybe he remembers what these functions were for?

Unless there's some interesting use case, I think I'd support deprecating all of the "full-state" versions at least and combining `setState_DP` and `setState_PV` so only one of them needs to be implemented in any specific thermo model.

bryanwweber:
I'm not totally sure which functions you're referring to @speth. I will say that, with respect to the PV property pair, these are extremely common in thermodynamic systems from intro textbooks. Or, possibly, in a cycle where the volume or pressure are held constant through a process, so you don't know the temperature at the other end (for example, "the mixture is heated at constant pressure until the volume has tripled").

> Unless there's some interesting use case, I think I'd support deprecating all of the "full-state" versions at least and combining setState_DP and setState_PV so only one of them needs to be implemented in any specific thermo model.

I think removing the full-state versions would be fine. I think as long as both DP and PV are available (even if they're dependent/interrelated) it would also be fine.

ischoegl:
Looks like I opened a can of worms here ... this was supposed to be simple! :rofl:

My main incentive was to make the nomenclature consistent, but I do see that there's a lot of boilerplate code. At the same time, the `setState_TPX`/`setState_TPY` are used extensively in examples and the test suite, so I'm not necessarily in support of removing those just for the sake of avoiding a handful of service functions (which can likely be inlined/header-only). 

I tend to agree that keeping both `setState_PV` and `setState_DP` make sense; @bryanwweber mentions an example that is indeed often used in thermo (another one is an IC engine where you know pressures and specific volume from experimental data). Regarding `DP` vs `PV`, this probably depends on a user perspective. Specific volumes are certainly common when viewed from a thermo perspective, but in other contexts, I'd say that density is probably more natural (compressible flows?). As an aside, the same argument goes for `TD` and `TV` ... in both cases, context is key.

bryanwweber:
> so I'm not necessarily in support of removing those just for the sake of avoiding a handful of service functions

I think @speth was suggesting to remove PDY and PDX, as those aren't as likely to be used.

interfaces/matlab/toolbox/@ThermoPhase/setState_DP.m
```diff
@@ -0,0 +1,16 @@
+function setState_DP(tp, rp)
```
bryanwweber: The latter parameter should be dp?

src/thermo/PDSS.cpp
```diff
@@ -184,7 +185,14 @@ void PDSS::setState_TP(doublereal temp, doublereal pres)
 
 void PDSS::setState_TR(doublereal temp, doublereal rho)
 {
-    throw NotImplementedError("PDSS::setState_TR");
+    warn_deprecated("PDSS::setState_TR",
```
bryanwweber, line 188: If this is not implemented anyways, could it just be replaced without warning?

src/thermo/PDSS_ConstVol.cpp
```diff
@@ -78,12 +78,12 @@ void PDSS_ConstVol::setState_TP(doublereal temp, doublereal pres)
     setPressure(pres);
 }
 
-void PDSS_ConstVol::setState_TR(doublereal temp, doublereal rho)
+void PDSS_ConstVol::setState_TD(double temp, double rho)
```
bryanwweber, line 81: Should we consider remaining these parameters at the same time? Density instead of rho?

bryanwweber (commented):
Just a couple comments for food for thought

src/thermo/PDSS_ConstVol.cpp
ischoegl: I don't think this to be necessary

src/thermo/PDSS.cpp
ischoegl: `setState_TD` is not implemented by the base class, but is overloaded by specializations.

ischoegl:
@speth and @bryanwweber ... I deprecated all full state setters using `R` (i.e. `TRX`/`TRY` and `RPX`/`RPY`) but retained the `TD` and `DP` versions.

speth (approved):
Looks good to me!

I wasn't originally advocating for the removal of the `TPX`/`TPY` setters or the `TRX`/`TRY` setters, but I'm quite happy to see the latter set go.

PR #1434, opened 2023-02-10
[Scons] Remove "install" target from "check_for_pytest"

band-a-prend:

The `check_for_pytest` depends on `check_for_ruamel_yaml` the pytest package is required on `install` phase and it results in the Cantera installation fail if pytest absent even if tests were not required. 

This patch swap `check_for_pytest` and `check_for_ruamel_yaml` dependency order.

The explicit `test-python-convert`, `test-python` targets are dropped from `check_for_ruamel_yaml` because the `check_for_pytest` includes `target.startswith("test-python")`.

See commit https://github.com/Cantera/cantera/commit/72734b1edd562d4ff4050136821b5003c633f9ab#r97347952 discussion

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1697,17 +1697,17 @@ ruamel_min_version = parse_version('0.15.34')
 # Minimum pytest version assumed based on Ubuntu 20.04
 pytest_min_version = parse_version("4.6.9")
 
+# Pytest is required only to test the Python module
+check_for_pytest = any(
+    target.startswith("test-python") for target in COMMAND_LINE_TARGETS
+)
+
 # Check for the minimum ruamel.yaml version, 0.15.34, at install and test
 # time. The check happens at install and test time because ruamel.yaml is
 # only required to run the Python interface, not to build it.
-check_for_ruamel_yaml = any(
+check_for_ruamel_yaml = check_for_pytest or any(
     target in COMMAND_LINE_TARGETS
-    for target in ["install", "test", "test-python-convert", "test-python"]
-)
-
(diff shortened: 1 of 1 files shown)


ischoegl (approved):
Thanks, @band-a-prend. Changes look good to me - some tests appear to be failing for unrelated upstream reasons - I’ll retrigger them, expecting them to pass eventually.

SConstruct
```diff
@@ -1697,17 +1697,17 @@ ruamel_min_version = parse_version('0.15.34')
 # Minimum pytest version assumed based on Ubuntu 20.04
 pytest_min_version = parse_version("4.6.9")
 
+# Pytest is required only to test the Python module
+check_for_pytest = any(
+    target.startswith("test-python") for target in COMMAND_LINE_TARGETS
```
speth, line 1702: the command line target `test` should also  trigger this check.

ischoegl: this is correct - sorry for jumping the gun on this

band-a-prend: @speth 
If cantera built without "python_full" or "python_minimal" should then "test" trigger the check_for_pytest? Is pytest required just for C++ library and Fortran wrapper testing?

speth: No, of course not. But `test` causes the Python test suite to be executed when the Python module is installed, so this check does need to happen in that case.

It's not clear from this line alone, but if you look further down where this test is actually handled, you'll see that it happens in a block that checks that `env['python_package'] != 'none'`.

PR #1438, opened 2023-02-17
Cleaning up the redundant line inside the for loop in evalSurfaces()

gkogekar:
The LHS array in `evalSurfaces()` is set to 1 by default. The line `LHS[loc] = 1` inside the for loop is removed, as it is redundant and confusing.

### DIFF
--- a/src/zeroD/Reactor.cpp
+++ b/src/zeroD/Reactor.cpp
@@ -299,7 +299,6 @@ void Reactor::evalSurfaces(double* LHS, double* RHS, double* sdot)
             RHS[loc + k] = m_work[surfloc + k] * rs0 * surf->size(k);
             sum -= RHS[loc + k];
         }
-        LHS[loc] = 1.0;
         RHS[loc] = sum;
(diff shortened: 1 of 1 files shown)


ischoegl (commented):
LGTM, although only after checking https://github.com/Cantera/cantera/blob/3d434130bd99e6fbde9236d65c498fbd77d61d8f/src/zeroD/ReactorNet.cpp#L276
 ... I assume that there never was an intent to change LHS values in this loop? I'm wondering whether setting `LHS[loc] = 1.0` on line 302/303 is required at all.

ischoegl (dismissed):
Approving as the change has no effect on behavior and removes a puzzling indication that there should be a change where there is none.

speth:
Yes, I agree, the statement `LHS[loc] = 1.0;` on line 302/303 (right after the loop) could be removed as well.

gkogekar:
I just removed the other line as well.

ischoegl (approved):

PR #1444, opened 2023-02-27
Add callbacks for synchronizing with C++ Solution members

### LINKED ISSUE #1409: Segmentation fault after Sim1D.restore in IPython
**Problem description**

After restoring a `ct.FreeFlame` object, hitting tab to get a list of methods results in a segmentation fault.

**Steps to reproduce**

Restoring a `FreeFlame` object from the output file `adiabatic_flame.yaml` obtained after running the Python example `samples/python/onedim/adiabatic_flame.py` causes hard crashes when attempting tab completion in IPython.

```
 % ipython
Python 3.10.8 | packaged by conda-forge | (main, Nov 22 2022, 08:25:29) [Clang 14.0.6 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.6.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import cantera as ct

In [2]: gas = ct.Solution("h2o2.yaml")

In [3]: sim = ct.FreeFlame(gas)
   ...: sim.restore("adiabatic_flame.yaml", name="mix")

In [4]: sim.zsh: segmentation fault  ipython
```

**Behavior**

The segmentation fault occurs in `[4]` after typing `sim.` and hitting tab. Prior to calling the restore method, autocompletion works. Curiously, the crash does not occur on the regular `python` command prompt.

The expected behavior would be to see
```
In [4]: sim.
  boundary_emissivities                    concentrations                           creation_rates
  chemical_potentials                      cp                                       cv
  clear_stats()                            cp_mass                                  cv_mass                                  >
  collect_data()                           cp_mole                                  cv_mole
```
(which was obtained from Cantera 2.6 *after* restoring).

**System information**

- Cantera version: newest main 7d86e2611
- OS: macOS (Apple Silicon) ... *not* reproduced on Intel Mac
- Python/MATLAB/other software versions: Python 3.10.8

speth:

- Add a system for registering callback functions with the C++ `Solution` class that are invoked whenever one of the members of the `Solution` is replaced, e.g. by calling `setTransportModel`.
- Register callbacks from `StFlow` and `IonFlow`
- Register a callback from the Python `Solution` class

Fixes #1409

```py
>>> import cantera as ct
>>> gas = ct.Solution('h2o2.yaml')
>>> flame = ct.FreeFlame(gas, width=0.1)
>>> gas.transport_model = 'UnityLewis'
>>> print(gas.transport_model, flame.transport_model)
UnityLewis UnityLewis
>>> flame.transport_model = 'Multi'
>>> print(gas.transport_model, flame.transport_model)
Multi Multi
```

**Checklist**

### DIFF
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -108,6 +108,22 @@ public:
     //! Returns a null pointer if the requested handle does not exist.
     shared_ptr<ExternalHandle> getExternalHandle(const std::string& name) const;
 
+    //! Register a function to be called if any of the Solution's thermo, kinetics,
+    //! or transport objects is replaced.
+    //! @param id  A unique ID corresponding to the object affected by the callback.
+    //!   Typically, this is a pointer to an object that also holds a reference to the
+    //!   Solution object.
+    //! @param callback  The callback function to be called after any component of the
+    //!   Solution is replaced.
+    //! When the callback becomes invalid (for example, the corresponding object is
+    //! being deleted, the removeChangedCallback() method must be invoked.
+    //! @since New in Cantera 3.0
+    void registerChangedCallback(void* id, const function<void()>& callback);
+
+    //! Remove the callback function associated with the specified object.
+    //! @since New in Cantera 3.0
+    void removeChangedCallback(void* id);
+
 protected:
     shared_ptr<ThermoPhase> m_thermo;  //!< ThermoPhase manager
     shared_ptr<Kinetics> m_kinetics;  //!< Kinetics manager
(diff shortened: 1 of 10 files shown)


ischoegl:
@speth ... thank for working on this (would have taken a stab myself before long). One thing I wanted to briefly discuss are options (1) vs (2) you mentioned in #1409. 

My own thinking here was to pursue (1) and give each `Domain1D` a unique copy of the base C++ `Solution`, rather than sharing a single object. This doesn't negate sharing `ThermoPhase`, `Kinetics` and `Transport` pointers, but will keep things separate where it matters. The copied C++ `Solution`s would be wrapped in Python as new objects, so there wouldn't be the issue you mentioned in #1409.

One scenario where I could see this being important is a hypothetical composite `Sim1D` with multiple domains constructed from the same `Solution`, where transport models differ (we currently don't have standard examples for this, but the core implementation does allow for more than three domains).

ischoegl (approved):
Thanks! Per discussion in #1409 this appears to be the best way forward given the current state of the code basis, despite leaving some open questions about synchronization when a single object is used as a basis for multiple reactors or other objects.

PR #1445, opened 2023-02-28
OneD domain factories and cleanup

ischoegl:

- Implement `DomainFactory` (see https://github.com/Cantera/cantera/projects/2)
- Initial simplifications in Python API
- Deprecate magic numbers in `Domain1D::domainType`
- Replace raw pointers by shared pointers in assembly of `Sim1D`.
- Remove `solution` from `Domain1D::show` and `Sim1D::show` to avoid erroneous conflation with `Solution` objects
- Introduce `clib` function returning domain type as well as `new_domain` factory loader (to facilitate #1182)

While other tweaks are possible, they will create merge conflicts with #1079

Implements one 'idea' in https://github.com/Cantera/cantera/projects/2

```c++
shared_ptr<Domain1D> inlet = newDomain("inlet", sol, "reactants");
shared_ptr<Domain1D> flow = newDomain("gas-flow", sol, "flow");
```

for `clib`, the following items are added:
```c
int domain_type3(int i, size_t lennm, char* nm); /* returns char* instead of magic number */
int sim1D_show(int i, const char* fname); /* replaces sim1D_showSolution */
int bdry_setSpreadRate(int i, double v); /* replaces inlet_setSpreadRate */
double bdry_spreadRate(int i); /* add missing getter */

int soln_setTransportModel(int n, const char* model); /* addendum to #1448 */
int trans_transportModel(int n, int lennm, char* nm); /* returns char* specifying transport type */

int domain_new(const char* type, int i, const char* id); /* factory constructor */
```

For the Python API, the following items are (re-)introduced:
```Python
flow = ct.FreeFlow(gas)
flow = ct.AxisymmetricFlow(gas)
```

with `IdealGasFlow` being deprecated.

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/_onedim.pyx
+++ b/interfaces/cython/cantera/_onedim.pyx
@@ -16,7 +16,7 @@ class _WeakrefProxy:
     pass
 
 cdef class Domain1D:
-    _domain_type = "None"
+    _domain_type = "none"
     def __cinit__(self, _SolutionBase phase not None, *args, **kwargs):
         self.domain = NULL
 
@@ -673,6 +673,49 @@ cdef class _FlowBase(Domain1D):
             return pystr(self.flow.flowType())
 
 
+cdef class FreeFlow(_FlowBase):
+    r"""A free flow domain. The equations solved are standard equations for adiabatic
+    one-dimensional flow. The solution variables are:
+
+    *velocity*
+        axial velocity
+    *T*
+        temperature
+    *Y_k*
+        species mass fractions
+    """
+    _domain_type = "free-flow"
+
+
+cdef class AxisymmetricFlow(_FlowBase):
+    r"""
+    An axisymmetric flow domain. The equations solved are the similarity equations for
+    the flow in a finite-height gap of infinite radial extent. The solution variables
+    are:
+
+    *velocity*
+        axial velocity
+    *spread_rate*
+        radial velocity divided by radius
+    *T*
+        temperature
+    *lambda*
+        :math:`(1/r)(dP/dr)`
+    *Y_k*
+        species mass fractions
+
+    It may be shown that if the boundary conditions on these variables are independent
+    of radius, then a similarity solution to the exact governing equations exists in
+    which these variables are all independent of radius. This solution holds only in
+    the low-Mach-number limit, in which case :math:`(dP/dz) = 0`, and :math:`lambda` is
+    a constant. (Lambda is treated as a spatially-varying solution variable for
+    numerical reasons, but in the final solution it is always independent of :math:`z`.)
+    As implemented here, the governing equations assume an ideal gas mixture. Arbitrary
+    chemistry is allowed, as well as arbitrary variation of the transport properties.
+    """
+    _domain_type = "axisymmetric-flow"
+
+
 cdef class IdealGasFlow(_FlowBase):
     """
     An ideal gas flow domain. Functions `set_free_flow` and
@@ -702,8 +745,18 @@ cdef class IdealGasFlow(_FlowBase):
     it is always independent of z.) As implemented here, the governing
     equations assume an ideal gas mixture.  Arbitrary chemistry is allowed, as
     well as arbitrary variation of the transport properties.
+
+    .. deprecated:: 3.0
+
+        Class to be removed after Cantera 3.0; replaced by `FreeFlow` and
+        s`AxisymmetricFlow`.
     """
     _domain_type = "gas-flow"
+
+    def __init__(self, *args, **kwargs):
+        warnings.warn("Class to be removed after Cantera 3.0; use 'FreeFlow' "
+                      "or AxisymmetricFlow' instead.", DeprecationWarning)
+        super().__init__(*args, **kwargs)
 
 
 cdef class IonFlow(_FlowBase):
--- a/interfaces/cython/cantera/_onedim.pyx
+++ b/interfaces/cython/cantera/_onedim.pyx
@@ -16,6 +16,7 @@ class _WeakrefProxy:
     pass
 
 cdef class Domain1D:
+    _domain_type = "None"
     def __cinit__(self, _SolutionBase phase not None, *args, **kwargs):
         self.domain = NULL
 
@@ -42,6 +43,13 @@ cdef class Domain1D:
         """
         def __get__(self):
             return self.domain.domainIndex()
+
+    property domain_type:
+        """
+        String indicating the domain implemented.
+        """
+        def __get__(self):
+            return pystr(self.domain.domainType())
 
     property n_components:
         """Number of solution components at each grid point."""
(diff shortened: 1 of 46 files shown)


ischoegl:
@speth ... after running into #1447 (which likely involves tackling Cantera/enhancements#158) I am omitting factory constructors in `clib` for the moment.

ischoegl:
Converting back to draft, as this should be merged *after* #1448.

ischoegl:
Rebased after the merge of #1448; some experiments for the introduction of `SharedCabinet` and/or `newDomain` factory constructors indicate that updates to `clib/ctonedim` will have to be somewhat more extensive than anticipated, ~and are postponed for another time.~ *Edit: changes are now incorporated.*

ischoegl:
@speth .... At this point, I verified that the traditional MATLAB toolbox runs correctly. There will be some trivial merge conflicts with #1455 (tests for model names), but this is otherwise ready for a review (*Edit: this includes the force push*)

@ssun30 ... this PR should bring additional simplifications that are relevant to #1182.

ischoegl:
Rebased and updated after merging #1455 / ready for a review.

ischoegl:
Resolved a minor merge conflict introduced by #1426.

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -16,6 +16,7 @@ class _WeakrefProxy:
     pass
 
 cdef class Domain1D:
+    _domain_type = "None"
```
speth: For consistency with #1455, I suppose this should be `"none"`.

interfaces/cython/cantera/onedim.py
```diff
@@ -56,7 +56,7 @@ def other_components(self, domain=None):
         if isinstance(dom, Inlet1D):
             return tuple([e for e in self._other
                           if e not in {'grid', 'lambda', 'eField'}])
-        elif isinstance(dom, (IdealGasFlow, IonFlow)):
+        elif isinstance(dom, (FreeFlow, AxisymmetricFlow, IdealGasFlow, IonFlow)):
```
speth, line 59: Can this be just
```suggestion
        elif isinstance(dom, _FlowBase):
```

interfaces/cython/cantera/onedim.py
```diff
@@ -177,7 +177,7 @@ def set_initial_guess(self, *args, data=None, group=None, **kwargs):
             arr.TP = T + left.T - T[0], self.P
 
             # adjust velocities
-            if self.flame.flow_type != "Free Flame":
+            if self.flame.domain_type.startswith("axisymmetric"):
```
speth, line 180: Would it be better to rely on `isFree` for this check?

src/clib/Cabinet.h
```diff
@@ -344,6 +344,18 @@ class SharedCabinet
         return data[n];
     }
 
+    /**
+     * Return object n, cast to the specified type.
+     */
+    template <class T>
+    static shared_ptr<T> as(size_t n) {
```
speth, line 351: For all the cases where this is used, the existing `get` method could be used instead, right?

src/clib/ct.cpp
```diff
@@ -189,6 +189,19 @@ extern "C" {
         }
     }
 
+    int soln_setTransportModel(int n, const char* model)
+    {
+        try {
+            auto soln = SolutionCabinet::at(n);
+            TransportCabinet::del(
+                TransportCabinet::index(*(soln->transport())));
```
speth, line 197: The fact that this deletes the old `Transport` object is potentially a little surprising, and should probably be noted with the declaration of this function.

src/clib/ctmultiphase.cpp
```diff
@@ -298,7 +298,7 @@ extern "C" {
         }
     }
 
-    doublereal mix_equilibrate(int i, const char* XY, doublereal rtol,
+    double mix_equilibrate(int i, const char* XY, double rtol,
                                int maxsteps, int maxiter, int loglevel)
```
speth: The indentation on the following line is now ragged.

src/oneD/StFlow.cpp
```diff
@@ -288,7 +295,7 @@ void StFlow::setGasAtMidpoint(const doublereal* x, size_t j)
 }
 
 bool StFlow::fixed_mdot() {
-    return (domainType() != cFreeFlow);
+    return !m_isFree;
 }
```
speth, line 299: This suggests that `fixed_mdot` could be deprecated in favor of `!isFree`.

test/clib/test_clib.cpp
```diff
@@ -76,6 +76,12 @@ TEST(ct, soln_objects)
 
     int trans = soln_kinetics(ref2);
     ASSERT_EQ(trans, 1);
+    int buflen = trans_transportModel(trans, 0, 0);
+    char* buf = new char[buflen];
+    trans_transportModel(trans, buflen, buf);
+    string trName = buf;
+    ASSERT_EQ(trName, "mixture-averaged");
+    delete[] buf;
```
speth: If you want to avoid needing to do a manual new/delete, you can write
```suggestion
    vector<char> buf(buflen);
    trans_transportModel(trans, buflen, buf.data());
    string trName(buf.data());
    ASSERT_EQ(trName, "mixture-averaged");
```

speth (changes requested):
Thanks, @ischoegl. This looks very good, and I think having the 1D code start to use the factory infrastructure that we have for most other object types will simplify some future additions (say, `ExtensibleDomain`). I had only a handful of minor suggestions.

src/clib/Cabinet.h
ischoegl: This is correct. I do, however, believe that passing shared pointers is in the end more flexible than what can be done with references. I view the `get` method as useful for drop-in replacement of `SharedCabinet` for `Cabinet`, while the `as` method may be preferable for future work.

src/oneD/StFlow.cpp
ischoegl: Potentially, yes. ~At the same time, I am not 100% sure that `m_isFree` will be needed at all once all raw pointers are replaced, as checking `dynamic_pointer_cast` does the job by itself.~ *Edit: rather than introducing `StFlow::isFree`, all instances are replaced by `StFlow::fixed_mdot`).*

interfaces/cython/cantera/onedim.py
ischoegl: I believe that it makes more sense to refactor `IonFlow` to be part of `AxisymmetricFlow`, and then test for the object type directly (I don't want to introduce a new Python method `is_free` just to deprecate/remove it one release later). At the moment, the `startswith` does the job. Some of this is - to some extent - dependent on how far we take #1079 in terms of allowing for alternative thermo models. *Edit: I also no longer believe that `StFlow::isFree` should be introduced at all.*

interfaces/cython/cantera/onedim.py
ischoegl: After trying this, I remembered that Cython refuses to export class names starting with an underscore. There may be a trick to force export, but I'm not aware of it.

ischoegl:
@speth ... thanks for the review! I adopted some suggestions, and responded to others. *Edit: I removed `StFlow::isFree` as I don't think that a new methods should be introduced.*

speth (approved):
Thanks for the updates, @ischoegl. This looks good to me.

PR #1448, opened 2023-03-01
Smart clib cabinets / add Solution to clib API

### LINKED ISSUE #158: Expose C++ Solution to clib
**Abstract**

Since the initial introduction of the C++ `Solution` class, it has become central to the instantiation of `ThermoPhase`/`Kinetics`/`Transport` objects from YAML input. It is, however, still missing from the `clib` interface. 

**Motivation**

* What problem is it trying to solve? ... add a missing interface to `clib`
* Who is affected by the change? ... developers - this is largely 'under the hood'
* Why is this a good solution? ... reduce number of instances required to keep track of objects

**Possible Solutions**

Create a `template<> SolutionCabinet* SolutionCabinet` in analogy to existing "cabinets" for `Thermo` etc.

### LINKED ISSUE #1447: clib sim1D_save/restore broken
**Problem description**

Recently merged work in #1385 broke `clib` methods used for saving and restoring of `oneD` data. The new methods are routed through `SolutionArray`; as there is no underlying `Solution` available at the moment (see Cantera/enhancements#158), the methods no longer work. 

**Steps to reproduce**

1. Open '...' matlab
2. Run '....' flame1 sample
3. See error '....'  below

**Behavior**

```
[...]
..............................................................................
no new points needed in flow
Error using ctmethods

CanteraError thrown by SolutionArray::SolutionArray:
Unable to create SolutionArray from invalid Solution object.


Error in stack_methods (line 15)
    v = ctmethods(90, n, job, a, b, c);

Error in Stack/saveSoln (line 28)
stack_methods(s.stack_id, 107, fname, id, desc);

Error in flame1 (line 95)
saveSoln(fl,'h2flame.yaml','energy',['solution with energy equation']);
```

**System information**

- Cantera version: 3.0.0a4 (main)
- OS: Windows 10
- Python/MATLAB/other software versions: Matlab R2022a

ischoegl:

This PR required keeping track of shared pointers to core Cantera objects (`ThermoPhase`, `Kinetics` and `Transport`) after their construction. Several factory constructors and internal methods that previously used raw pointers or references are replaced by methods that use shared pointers, with the original methods being deprecated.

Modifications allowed for consistent tracking of objects behind the `clib` API, where a new `SharedCabinet` based on smart pointers is introduced. The new object is adopted for `ThermoCabinet`, `KineticsCabinet` and `TransportCabinet`. 

Other details:

- Fix bugs in `test_problems/clib_test/clib_test.c`
- Implement `SolutionCabinet` based on `SharedCabinet`
- Implement `clib` tests based on `googletests` (run via `scons test-clib-gtest`), which now includes tests for `ctonedim`

Fixes #1447
Resolves Cantera/enhancements#158

Solution types can now be loaded directly, i.e.

```c
int soln_newSolution(const char* infile, const char* name, const char* transport);
int soln_newInterface(const char* infile, const char* name, int na, const int* adjacent);
```

where parts are obtained as:
```c
int soln_thermo(int n);
int soln_kinetics(int n);
int soln_transport(int n);
```

**Checklist**


ischoegl:
After writing new google-test based `scons test-clib-gtest` routines for `ct.h`, `ctreactor.h` and `ctonedim.h` as an (expanded) alternative to `scons test-clib` (where bugs needed to be fixed), and successfully running `test_examples.m` on MATLAB (Windows 10), this is finally ready for a review ...

ischoegl:
@ssun30 ... together with #1445, there should be a couple of useful bits for #1182 in here (or: at least lay the foundation for several significant simplifications).

include/cantera/kinetics/Kinetics.h
```diff
@@ -1320,6 +1329,10 @@ class Kinetics
      */
     std::vector<ThermoPhase*> m_thermo;
 
+    //! vector of shared pointers, @see m_thermo
+    //! @todo replace m_thermo with shared version after Cantera 3.0
+    std::vector<shared_ptr<ThermoPhase>> m_sharedThermo;
```
speth: `std::` is no longer needed for `vector`.

src/clib/Cabinet.h
```diff
@@ -205,4 +206,174 @@ class Cabinet
...
+     */
+    static int index(const M& obj) {
+        dataRef data = getData();
+        int count = 0;
+        for (const auto& item : data) {
+            if (item.get() == &obj) {
+                return count;
+            }
+            count++;
+        }
```
speth: This linear search for the index could get slow if the data vector starts to get long, which could happen if the user creates and clears a lot of `Solution` objects. It might be better to create a reverse lookup table, i.e. `unordered_map<M*, int>` for this purpose. I realize this function existed before in a similar form, but I think now it starts to become a bit more important.

src/clib/Cabinet.h
```diff
@@ -205,4 +206,174 @@ class Cabinet
...
+    }
+
+private:
+    /**
+     * Static function that returns a pointer to the data member of
+     * the singleton SharedCabinet<M> instance. All member functions should
+     * access the data through this function.
+     */
+    static dataRef getData() {
+        if (s_storage == 0) {
```
speth: I don't always remember to do this, but I think it's preferable to write:
```suggestion
        if (s_storage == nullptr) {
```

interfaces/cython/cantera/solutionbase.pyx
```diff
@@ -440,8 +442,10 @@ cdef _assign_Solution(_SolutionBase soln, shared_ptr[CxxSolution] cxx_soln,
     soln.base = cxx_soln.get()
 
     def assign_pointers():
-        soln.thermo = soln.base.thermo().get()
-        soln.kinetics = soln.base.kinetics().get()
+        soln._thermo = soln.base.thermo()
+        soln.thermo = soln._thermo.get()
+        soln._kinetics = soln.base.kinetics()
```
speth: Given the very limited ways that the `shared_ptr<Kinetics>` and `shared_ptr<ThermoPhase>` members of `_SolutionBase` are directly used, can we avoid introducing them? I think in the small number of places they appear, we can just write `self.base.thermo()` or similar, right?

src/clib/Cabinet.h
```diff
@@ -205,4 +206,174 @@ class Cabinet
...
+    static int reset() {
+        getData().clear();
+        return 0;
+    }
+
+    /**
+     * Delete the nth object. After the object is deleted, the pointer to it in the list
+     * is replaced by a pointer to the first element in the list.
+     */
+    static void del(size_t n) {
```
speth: This docstring does not correspond to the behavior of the method.

src/clib/Cabinet.h
```diff
@@ -9,7 +9,8 @@
 #define CT_CABINET_H
 
 #include "cantera/base/ctexceptions.h"
-#include "clib_utils.h"
+
+using Cantera::CanteraError;
```
speth: Alternatively, you could just put this file's contents inside `namespace Cantera` like everything else.

src/clib/ct.cpp
```diff
@@ -49,6 +53,149 @@ extern "C" {
...
+    int soln_del(int n)
+    {
+        try {
+            if (n >= 0 && n < SolutionCabinet::size()) {
+                // remove all associated objects
+                auto sol = SolutionCabinet::at(n);
+                int index = ThermoCabinet::index(*(sol->thermo()));
+                if (index >= 0) {
+                    ThermoCabinet::del(index);
+                }
```
speth: I know `clib` has zero documentation, but this behavior of the `Solution` deleting its members probably should be made explicit -- none of the other `del` methods in `clib` delete things other than the specified object.

src/clib/ct.cpp
```diff
@@ -49,6 +53,149 @@ extern "C" {
...
+        } catch (...) {
+            return handleAllExceptions(-1, ERR);
+        }
+    }
+
+    int soln_thermo(int n)
+    {
+        try {
+            auto sol = SolutionCabinet::at(n);
+            return ThermoCabinet::index(*(sol->thermo()));
```
speth:
```suggestion
            return ThermoCabinet::index(*sol->thermo());
```

src/clib/ct.cpp
```diff
@@ -49,6 +53,149 @@ extern "C" {
...
+    {
+        try {
+            auto sol = SolutionCabinet::at(n);
+            if (!sol->kinetics()) {
+                return -1;
+            }
+            return KineticsCabinet::index(*(sol->kinetics()));
+        } catch (...) {
+            return handleAllExceptions(-1, ERR);
+        }
```
speth: I think the error handling here needs to change slightly -- if -1 is the return value for "there is no `Kinetics` object" and that isn't considered an error, then -1 can't also be the return value if a `CanteraError` was caught (first argument to `handleAllExceptions`).

src/clib/ct.cpp
```diff
@@ -1518,6 +1663,33 @@ extern "C" {
...
     }
 
+    int thermo_size()
+    {
+        try {
+            return ThermoCabinet::size();
+        } catch (...) {
+            return handleAllExceptions(-1, ERR);
+        }
+    }
```
speth: What is the use case for knowing the size of the `Cabinet`? The user should only ever be accessing it via known handles.

src/clib/ct.cpp
```diff
@@ -1547,4 +1719,34 @@ extern "C" {
...
+
+    int thermo_reset()
+    {
+        try {
+            ThermoCabinet::reset();
+            return 0;
+        } catch (...) {
+            return handleAllExceptions(-1, ERR);
+        }
+    }
```
speth: Are these functions needed, separate from the full `ct_clearStorage`?

src/fortran/fct.cpp
```diff
@@ -51,6 +51,11 @@ ThermoPhase* _fth(const integer* n)
     return &ThermoCabinet::item(*n);
 }
 
+shared_ptr<ThermoPhase> _fth3(const integer* n)
```
speth: Why the suffix `3`? This isn't transitional (right?), and there's no reason to keep the overly cryptic naming convention, given this is a purely internal method within an unnamed namespace.

include/cantera/kinetics/Kinetics.h
```diff
@@ -1099,6 +1105,9 @@ class Kinetics
      *
      * @param thermo    Reference to the ThermoPhase to be added.
      */
+    virtual void addPhase(shared_ptr<ThermoPhase> thermo);
+
+    //! @see Kinetics::addPhase(shared_ptr<ThermoPhase>)
     virtual void addPhase(ThermoPhase& thermo);
```
speth: I think this old version of `addPhase` needs a `@deprecated` label, and the shared pointer version needs something like `@since Changed ...`.

src/oneD/Boundary1D.cpp
```diff
@@ -644,8 +644,23 @@ ReactingSurf1D::ReactingSurf1D(shared_ptr<Solution> solution, const std::string&
     m_enabled = true;
 }
 
+void ReactingSurf1D::setKinetics(shared_ptr<Kinetics> kin)
+{
+    m_solution = Solution::create();
+    m_solution->setThermo(kin->reactionPhase());
+    m_solution->setKinetics(kin);
+    m_solution->setTransportModel(); // default transport
```
speth: I don't think the surface needs a transport model, does it? Creating transport models can be fairly expensive, so should be avoided if possible.

src/oneD/Boundary1D.cpp
```diff
@@ -644,8 +644,23 @@ ReactingSurf1D::ReactingSurf1D(shared_ptr<Solution> solution, const std::string&
...
+    m_kin = dynamic_pointer_cast<InterfaceKinetics>(kin).get();
+    m_surfindex = kin->reactionPhaseIndex();
+    m_sphase = (SurfPhase*)&kin->thermo(m_surfindex);
```
speth:
```suggestion
    m_sphase = dynamic_pointer_cast<SurfPhase>(kin->reactionPhase()).get();
```

ischoegl:
@speth … adding `soln_newInterface` (which was supposed to be a trivial addition) caused some unexpected complications - see #1450 - but none of the substance has changed. Will wait for a review at this point.

src/oneD/StFlow.cpp
```diff
@@ -141,6 +148,48 @@ void StFlow::setThermo(IdealGasPhase& th) {
...
+    //     "To be removed after Cantera 3.0. Replaced by Domain1D::setKinetics.");
+    m_kin = &kin;
+}
+
+void StFlow::setTransport(shared_ptr<Transport> trans)
+{
+    if (!m_solution) {
+        throw CanteraError("StFlow::setTransport",
+            "Unable to update object that was not constructed from smart pointers.");
+    }
```
speth: Is the intention to remove the ability to create `StFlow` from raw pointers? In which case, this needs the `@todo remove after Cantera 3.0` on these checks, right?

src/thermo/ThermoFactory.cpp
```diff
@@ -150,24 +169,33 @@ unique_ptr<ThermoPhase> newPhase(const AnyMap& phaseNode, const AnyMap& rootNode
...
         extension = toLowerCopy(infile.substr(dot+1));
     }
+    string id_ = id;
     if (id == "-") {
-        id = "";
+        id_ = "";
     }
     if (extension == "cti" || extension == "xml") {
-        throw CanteraError("newPhase",
+        throw CanteraError("newThermo",
```
speth, line 174:
```suggestion
        throw CanteraError("newThermoPhase",
```

test/clib/clib_interface.cpp
```diff
@@ -0,0 +1,227 @@
...
+    ASSERT_EQ(thermo_size(), 2);
+
+    thermo_reset();
+    ASSERT_EQ(thermo_size(), 0);
+}
+
+TEST(ct, cabinet_exceptions)
+{
+    soln_newSolution("h2o2.yaml", "ohmech", "");
+    soln_name(999, 0, 0);
```
speth: You might also want to test that `soln_name` returns -1 in this case.

test/SConscript
```diff
@@ -281,6 +281,7 @@ def addMatlabTest(script, testName, dependencies=None, env_vars=()):
     return run_program
 
 # Instantiate tests
+addTestProgram('clib', 'clib-gtest')
```
speth: For the sake of consistency, I think it would be better for this test to be what gets run when using `scons test clib` and the existing `clib` test can have some other name.

test/clib/clib_interface.cpp
```diff
@@ -0,0 +1,227 @@
...
+    soln_reset();
+    thermo_reset();
+    kin_reset();
+
+    int sol = soln_newSolution("ptcombust.yaml", "gas", "");
+    ASSERT_EQ(sol, 0);
+
+    std::vector<int> adj{sol};
+    int surf = soln_newInterface("ptcombust.yaml", "Pt_surf", 1, adj.data());
+    ASSERT_EQ(surf, 1);
```
speth: As part of this PR, did you want to consider how to handle access to "adjacent" phases of a Solution, and the case where an `Interface` is instantiated with the adjacent phases automatically handled based on the input file?

test/clib/clib_interface.cpp
```diff
@@ -0,0 +1,227 @@
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+#include <fstream>
+
+#include "cantera/core.h"
+
+extern "C" {
+    #include "cantera/clib/ct.h"
+}
```
speth: `ct.h` already has a guarded version of `extern "C"`, so this wrapper shouldn't be necessary.

test/clib/clib_interface.cpp
```diff
@@ -0,0 +1,227 @@
...
+    int ret = trans_getMixDiffCoeffs(tran, 53, c_dkm.data());
+    ASSERT_EQ(ret, 0);
+
+    vector<double> cpp_dkm(nsp);
+    auto sol = newSolution("gri30.yaml", "gri30");
+    auto transport = sol->transport();
+    transport->getMixDiffCoeffs(cpp_dkm.data());
+
+    for (size_t n = 0; n < nsp; n++) {
+        ASSERT_NEAR(cpp_dkm[n], c_dkm[n], 1e-6);
```
speth: 1e-6 is a pretty loose tolerance for quantities that are on the order of 1e-5.

test/clib/test_ctonedim.cpp
```diff
@@ -0,0 +1,305 @@
...
+    ASSERT_GE(dom, 0);
+}
+
+vector<double> T_ctonedim = {
+    300.0000, 300.0000, 300.0000, 300.0001, 300.0010, 300.0141, 300.0842,
+    300.6108, 302.1595, 308.7286, 319.8533, 331.8250, 352.4362, 367.7950,
+    387.6687, 441.9631, 513.3673, 596.8101, 717.1222, 838.6055, 956.3989,
+    1067.1676, 1259.9339, 1408.2242, 1517.2894, 1598.5709, 1661.5194, 1712.0943,
+    1786.7072, 1843.4863, 1920.5714, 2009.4328, 2066.5316, 2135.8423, 2174.3517,
+    2207.6584, 2228.0248, 2228.0248};
```
speth: Directly comparing to this output makes these tests a bit brittle, and subject to failures due to either minor changes in the implementation of the 1D classes, or even floating point differences on different platforms that cause a step change, say due to some point crossing the threshold for grid refinement.

To avoid spurious test failures, my suggestion would be to just check that the solution, with grid refinement disabled, follows simple checks like the temperature being monotonically increasing. We have a much more robust set of tests for the 1D solver itself elsewhere.

test/clib/test_ctonedim.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+TEST(ctonedim, freeflame_cpp)
+{
+    auto sol = newSolution("h2o2.yaml", "ohmech", "Mix");
+    auto gas = sol->thermo();
+    size_t nsp = gas->nSpecies();
+
```
speth: Do we need this test? It doesn't actually touch anything in clib, does it?

test_problems/clib_test/clib_test.c
```diff
@@ -34,7 +34,7 @@ int main(int argc, char** argv)
     ret = thermo_print(thermo, 1, 0);
     assert(ret == 0);
 
-    int kin = kin_newFromFile("gri30.yaml", "gri30", thermo, 0, 0, 0, 0);
+    int kin = kin_newFromFile("gri30.yaml", "gri30", thermo, -1, -1, -1, -1);
```
speth, line 37: This is quite the breaking change, if anyone is making direct use of `clib`.

speth (changes requested):
Thanks for taking this on, @ischoegl. Generally, I think this looks good, and most of my comments below are minor.

I appreciate the effort put into cleaning up all the `newX` constructors, that were previously returning a mishmash of raw pointers, `unique_ptr` and `shared_ptr`. There are a couple of things that I'd like to consider for further refinement.

The first is the need to make the "phase id" argument to `newThermoPhase` mandatory. I realize this is driven by wanting to not break any of the existing methods.

Second is the more frequent need for users to call `dynamic_pointer_cast` to get objects of the underlying type, when special methods are only available there.

My suggested solution to both of these issues is to template `newThermo` and have it do the `dynamic_pointer_cast` internally, with a signature something like:
```c++
template <class T=ThermoPhase>
shared_ptr<T> newThermo(const string& infile, const string& id="");
```
Then, users could write:
```c++
auto surf = newThermo<SurfPhase>(infile);
surf->setCoveragesByName(...);
```

This would necessitate removing the existing method `shared_ptr<ThermoPhase> newThermo(const string& model)`, but we could maintain backwards compatibility (with a deprecation warning) by checking the first argument for a "." and warning if it doesn't contain one. I think the existing behavior is not that useful to end-users, since without access to the derived type, it's not that easy to set up the phase. I also kind of like `newThermo` as a shorter function name.

test_problems/clib_test/clib_test.c
ischoegl: This is actually a bug in the code that took me about half a day to find! In all other instances (fortran and matlab), -1 is used when adjacent phases aren’t used. For the old `Cabinet`, 0 had a specific meaning where the output ended up being corrupted (!), see necessary changes in `output_blessed.txt`. After the fix, the `clib` output matches the C++ result.

*PS: What was most problematic in the old clib example is that results didn’t even remotely match what it was supposed to be. Mole fractions were set after a change of pressure, so it wasn’t anywhere close to the 5atm it  was supposed to be run at. It somehow used a default object instead, which was the only reason results made some sense (but still weren’t accurate). See fix in commit f695ea64dae35e687b8d22ef03fad638a8c1d47e*

test/SConscript
ischoegl: I believe the ideal situation would be to move the old test over to `samples` (as `clib-demo.c`) eventually, so I’ll use that name.

test/clib/test_ctonedim.cpp
ischoegl: I used this as a template when porting the clib version; as it's already implemented, I am using it to start a new gtest runner for `scons test-oneD`

src/clib/ct.cpp
ischoegl: Using 
```c
return handleAllExceptions(npos, ncos);
```
as in other pre-existing instances.

test/clib/clib_interface.cpp
ischoegl: Sure, why not ...

src/clib/ct.cpp
ischoegl: The `reset` version is a little different from `clear` (it actually resets the cabinet). I can merge things to `ct_resetStorage`.

src/clib/ct.cpp
ischoegl: It was useful for initial testing; not that things are working, they are probably not needed.

src/clib/Cabinet.h
ischoegl: That's a good point. One thing I'd like to ask before implementing this is whether there's a way to avoid having to reference a `s_lookup` in all of the files where the `SharedCabinet<M>` is used (similar to what is currently already present for `s_storage`?

src/oneD/StFlow.cpp
ischoegl: Correct.

test/clib/test_ctonedim.cpp
ischoegl: Fair point. The original intent was to ensure that both `cpp` and `clib` versions yield the same result. Whenever `SolutionArray` becomes available in `clib` (I have no plans to do this for 3.0), there's the option to have a reference solution instead. I guess it could be done with `sim1D_restore`, but I don't think it's necessary at this point.

Using the monotonic criterion with grid refinement disabled as suggested for the time being.

ischoegl:
Thanks for the prompt review, @speth! I took care of everything except the lookup table in `Cabinet.h`, where I have a question (see above)

> The first is the need to make the "phase id" argument to `newThermoPhase` mandatory. I realize this is driven by wanting to not break any of the existing methods.
> 
> Second is the more frequent need for users to call `dynamic_pointer_cast` to get objects of the underlying type, when special methods are only available there.
> 
> [...] but we could maintain backwards compatibility (with a deprecation warning) by checking the first argument for a "." and warning if it doesn't contain one. I think the existing behavior is not that useful to end-users, since without access to the derived type, it's not that easy to set up the phase. I also kind of like `newThermo` as a shorter function name.

I agree that the mandatory id argument wasn't ideal, and adopted your suggestion about checking for a ".". It's noteworthy that the Python API uses the existing behavior, so I renamed the existing route `newThermoModel`.

However, I didn't find many instances for where users may need to do a `dynamic_pointer_cast` (it’s mainly something that is helpful for the test suite; `Cantera/core.h` doesn’t provide access to factories, so I’m not sure that we should add this to the default API).

I believe that creating a template for `Solution::thermo<T>()` would be ultimately much more useful. I.e.
```C++
template <class T=ThermoPhase>
shared_ptr<T> thermo();
```
as most users will presumably use `newSolution` and `newInterface` for instantiations. I.e. 
```C++
auto soln = newInterface(infile);
auto surf = soln->thermo<SurfPhase>();
surf->setCoveragesByName(...);
```
The same logic applies for `Solution::kinetics<T>()` and `Solution::transport<T>()`. ~As this isn't really touched by this PR, I'd suggest to leave this for another time.~ *Edit: I just implemented those getters in #1451*

src/oneD/Boundary1D.cpp
ischoegl: changed to `m_solution->setTransportModel("none")`

ischoegl:
@speth ... I ended up renaming `Kinetics::addPhase` to `Kinetics::addThermo`, as I believe this to be more consistent with the updated nomenclature.

src/clib/Cabinet.h
ischoegl: Nevermind. It can be `m_lookup`, just like `m_table`.

ischoegl:
:tada: ... I think this PR is ready / all suggestions are taken care of.

src/clib/ct.cpp
speth: Sorry for being unclear here. Since the method returns an `int`, I think it's fine for `handleAllExceptions` to use int values rather than `npos` which is of type `size_t` (I'm sure we have other places where this is mixed up, but that's not that big of a problem).

The issue is the choice of sentinel value. The logic in the caller's code should be that if either of the values specified by `handleAllExceptions` is returned, an error occurred and the message for that error can be obtained by calling `ct_getCanteraError`. But if this method also returns `-1` (same thing as `npos` when it gets converted to an `int`) when the issue is just that the `Solution` doesn't have a `Kinetics` object (which is not considered an error), then the caller will incorrectly assume that there is an error message to be retrieved.

test_problems/clib_test/clib_test.c
speth: Sorry, I guess those bugs in `clib_test.c` were probably all my doing. 😳

include/cantera/thermo/ThermoFactory.h
```diff
@@ -12,6 +12,7 @@
 
 #include "ThermoPhase.h"
 #include "cantera/base/FactoryBase.h"
+#include "cantera/base/stringUtils.h"
```
speth: I think this `include` can be moved to the `.cpp` file, to avoid dragging it in everywhere `ThermoFactory.h` is included.

ischoegl: Absolutely. I had this included when I tested the templated version.

src/kinetics/InterfaceKinetics.cpp
```diff
@@ -16,19 +16,24 @@ using namespace std;
...
-    if (thermo != 0) {
-        addPhase(*thermo);
-    }
+}
+
+InterfaceKinetics::InterfaceKinetics(ThermoPhase* thermo)
+    : InterfaceKinetics()
+{
+    warn_deprecated("InterfaceKinetics::InterfaceKinetics",
+        "To be removed after Cantera 3.0. Use default constructor instead.");
```
speth: For clarity, I guess this should be:
```suggestion
    warn_deprecated("InterfaceKinetics::InterfaceKinetics(ThermoPhase*)",
        "To be removed after Cantera 3.0. Use default constructor instead.");
```

src/kinetics/KineticsFactory.cpp
```diff
@@ -65,10 +67,42 @@ shared_ptr<Kinetics> newKinetics(const string& model)
...
+    kin->init();
+    addReactions(*kin, phaseNode, rootNode);
+    return kin;
+}
+
 unique_ptr<Kinetics> newKinetics(const vector<ThermoPhase*>& phases,
                                  const AnyMap& phaseNode,
                                  const AnyMap& rootNode)
 {
+    warn_deprecated("newKinetics",
```
speth:
```suggestion
    warn_deprecated("newKinetics(vector<ThermoPhase*>&, AnyMap&, AnyMap&)",
```

src/oneD/StFlow.cpp
```diff
@@ -141,6 +148,50 @@ void StFlow::setThermo(IdealGasPhase& th) {
...
+    m_kin = kin.get();
+    m_solution->setKinetics(kin);
+}
+
+void StFlow::setKinetics(Kinetics& kin)
+{
+    // @todo: resolve crash for updated m_solution->registerChangedCallback
+    // warn_deprecated("StFlow::setKinetics",
+    //     "To be removed after Cantera 3.0. Replaced by Domain1D::setKinetics.");
+    m_kin = &kin;
```
speth, line 166: I forgot to ask before, but do you have an example of this crash?

Also, should the commented-out warning be un-commented, or removed?

speth (changes requested):
Thanks for the updates, @ischoegl. Besides the minor suggestions below, I think the one thing that still needs a little bit of work is the error handling for `soln_kinetics`, `soln_transport`, and `soln_adjacent`. I commented on that here: https://github.com/Cantera/cantera/pull/1448#discussion_r1126526672, but GitHub is probably trying to hide that from you.

src/oneD/StFlow.cpp
ischoegl: Ah - yes, this is something that I wanted to come back to, but didn't. The crash I experienced happened when I tried to update https://github.com/Cantera/cantera/blob/7f04d839f3f0bebbc3474ecb972007d896362af7/src/oneD/StFlow.cpp#L126-L129
to use the `shared_ptr` version. It may have masked another issue and I haven't tried since. Will certainly look at it again. The deprecation warning is commented out as it is de facto still actively used by the callbacks in `StFlow` and `IonFlow`.

ischoegl: @speth ... I just confirmed that this still crashes: simply remove the `*`'s (either of the two lines), recompile and run `scons test-python-onedim`. Any insights would be appreciated ...

src/clib/ct.cpp
ischoegl: Ok. Thanks for the clarification! (`clib`'s exception handling is something that I wasn't too familiar with.) So I guess, simply switching to
```C++
```
would do the trick? (with `ERR` being -999);

ischoegl:
@speth ... thanks for the re-review! You probably noticed that I updated the reverse lookup table to prevent memory leaks (see c5bf0db5c30be49b25683436aa871d525a023790). Otherwise, I *believe* the clib exception handling should now be addressed. I am stuck on the `m_solution->registerChangedCallback` crash (see https://github.com/Cantera/cantera/pull/1448#discussion_r1126865635), as I'm not sure why this is causing issues.

speth (approved):
Thanks @ischoegl, this looks good to me. The last commit I pushed took care of the callback-related errors, so I'll leave the final merge to you.

ischoegl:
Thank you, @speth!

PR #1451, opened 2023-03-05
Templates for solution getters

ischoegl:
**Changes proposed in this pull request**

Add templated getters for `Solution`, as motivated by review feedback in #1448.

**If applicable, provide an example illustrating new features this pull request is introducing**

```C++
auto soln = newInterface(infile);
auto surf = soln->thermo<SurfPhase>();
surf->setCoveragesByName(...);
auto kin = soln->kinetics<InterfaceKinetics>();
kin->advanceCoverages(...);
```

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

PR #1452, opened 2023-03-07
C++ code cleanup and modernization

speth:
This is some wide-ranging cleanup of the code to make use of some features from the modern C++ to make the code simpler and more readable.

Given the large number of files this touches, I know there will be a few merge conflicts, so I'm happy to try getting a couple of existing PRs merged ahead of this one and deal with those conflicts here.

- Use `if constexpr` (C++17) to greatly simplify the implementation of `MultiRate` methods that depend on whether or not the rate class implements specific methods like `updateFromStruct`
- Use `std::filesystem` (C++17) to avoid OS-specific implementations of `get_modified_time`
- Use "structured bindings" C++17 to unpack `std::pair`/`std::tuple` and improve readability (for example, `for (auto& [a, b] : some_map)`)
- Make use of C++11 "non-static data member initialization" to initialize member variables in the class definition rather than in the constructor, whenever possible.
- Make use of `make_unique` (C++14) and `make_shared` (C++11) to reduce usage of bare `new`.
- Use `ClassName() = default;` to implement trivial constructors and similar for destructors (C++11)
- Remove unnecessary standard library includes and `using` statements
- Use `std::any` (C++17) instead of `boost::any` to implement `AnyValue`. Since we no longer need to hide the `boost` type behind a `unique_ptr`, this simplifies the implementation of `AnyValue` slightly and eliminates an extra memory allocation for every `AnyValue` object.

**Checklist**

### DIFF
--- a/test_problems/clib_test/clib_test.c
+++ b/test_problems/clib_test/clib_test.c
@@ -15,7 +15,7 @@ int main(int argc, char** argv)
 {
     int ret;
     int thermo = thermo_newFromFile("gri30.yaml", "gri30");
-    assert(thermo > 0);
+    assert(thermo >= 0);
     size_t nsp = thermo_nSpecies(thermo);
     assert(nsp == 53);
 
@@ -35,7 +35,7 @@ int main(int argc, char** argv)
     assert(ret == 0);
 
     int kin = kin_newFromFile("gri30.yaml", "gri30", thermo, -1, -1, -1, -1);
-    assert(kin > 0);
+    assert(kin >= 0);
 
     size_t nr = kin_nReactions(kin);
     assert(nr == 325 );
@@ -55,6 +55,7 @@ int main(int argc, char** argv)
 
     printf("\n  Species    Mix diff coeff\n");
     int tran = trans_newDefault(thermo, 0);
+    assert(tran >= 0);
     double dkm[53];
     trans_getMixDiffCoeffs(tran, 53, dkm);
     int k; // declare this here for C89 compatibility
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -97,6 +97,8 @@
 * For any new code, do *not* use the `doublereal` and `integer` typedefs for the
   basic types `double` and `int`, but also do not go out of your way to change
   uses of these in otherwise unmodified code.
+* Initialize member variables with their declarations, when possible, rather than using
+  constructor-based initialization.
 
 ## Python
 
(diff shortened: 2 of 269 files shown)


ischoegl (commented):
Thanks for this extensive cleanup, @speth! The changes are straight-forward and all look good to me. I definitely like the unpacking of pairs, which definitely makes code easier to follow!

I will approve once the remaining Intel compiler failure is resolved.

ischoegl (commented):
Btw, feel free to cherry-pick 0973ae2159dfa7f18b47b872ce9d2cddbdf273d9 from #1451 ... it's a better fit here.

ischoegl:
@speth ... on a related note: what are differences between boost's `any` and the new `std::any` introduced in C++17? Is this a drop-in replacement? *Edit: here's a link to a [related SO question](https://stackoverflow.com/questions/49278691/differences-between-boostany-and-stdany)*

speth:
Yes, `std::any` is basically the adoption of `boost::any` into the standard library. I had started testing a transition, but it doesn't really have much of a benefit at the moment -- no impact on compilation time, and we still depend on Boost either way. Still, it might save some headache down the line.

ischoegl:
> Yes, `std::any` is basically the adoption of `boost::any` into the standard library. I had started testing a transition, but it doesn't really have much of a benefit at the moment -- no impact on compilation time, and we still depend on Boost either way. Still, it might save some headache down the line.

If I'm not mistaken, the `std::any` implementation may have some optimizations for small objects, so it still may pay off. Not sure whether this has any measurable impact on performance though ...

test_problems/clib_test/clib_test.c
```diff
@@ -15,7 +15,7 @@ int main(int argc, char** argv)
 {
     int ret;
     int thermo = thermo_newFromFile("gri30.yaml", "gri30");
```
ischoegl, line 17: Thanks for catching this: any idea why this was not raising exceptions? I had thought of removing those checks as they are superseded by the new test suite (where all checks are replicated), but held off in #1448.

speth: I was confused by this for a while, too. The answer is that `assert` statements all get optimized out when compiling with the `NDEBUG` macro enabled, which we generally do unless you specify `optimize=n` to SCons. I happened to catch this because I was debugging a different crash and needed to disable optimizations to be able to use the debugger effectively.

ischoegl (approved):
Thank you for the updates!

PR #1456, opened 2023-03-18
Simplify Python extension loading and other Python improvements

speth:

- Simplify loading of Python extension support by letting Cython handle the complexity of loading the extension module after Python has already been initialized. In the case of a non-Python main program, all we have to do is initialize Python and import the Cantera Python module.
- Add support for Cython 3.0.0b1
- Convert C++ `NotImplementedError` to Python native `NotImplementedError`
- Improve documentation of `_SolutionBase._references`

**Checklist**


interfaces/cython/cantera/thermo.pyx
```diff
@@ -268,7 +268,11 @@ cdef class ThermoPhase(_SolutionBase):
             self.thermo_basis = mass_basis
         # In composite objects, the ThermoPhase constructor needs to be called first
         # to prevent instantiation of stand-alone 'Kinetics' or 'Transport' objects.
-        # The following is used as a sentinel.
+        # The following is used as a sentinel. After initialization, the _references
```
ischoegl, line 271: I know this is still in draft mode, but aren’t sentinels for these cases relevant to other API’s (not just Python) as well? In other words, could these be set in C++?

speth: Yes, this feature probably ought to be implemented at the C++ level, but doing so is somewhat complicated. I think some of the things you've done recently to more generally use `shared_ptr` may make this easier after Cantera 3.0.

ischoegl: I'm :+1: with keeping things Python-specific for the moment, but would request a note/comment ...

ischoegl (changes requested):
@speth ... this looks great overall. The only item I'd request would be a `todo` comment to move sentinels to C++ after Cantera 3.0.

speth:
I created an issue (#1457) to track the needed feature. I think that's preferable to a to-do item buried somewhere in the code.

ischoegl (changes requested):
> I created an issue (#1457) to track the needed feature. I think that's preferable to a to-do item buried somewhere in the code.

Thanks for creating the issue! I agree that it is preferable to a `todo`.

It did, however, occur to me that `SolutionArray` needs a similar sentinel, somewhere around here: https://github.com/Cantera/cantera/blob/fbfd93f48f18b5072719a8c17f97549ce6ecfec0/interfaces/cython/cantera/composite.py#L558-L559 
Even prior to the introduction of C++ `SolutionArray`, adding species would have created inconsistencies in stored data.

PS: Another item I noticed is that we may want to test with Cython 3.0 again?

speth:
I've added the Python `weakref`-style checks to both `SolutionArray` and `Quantity`.

I've also added a new CI job that runs only after pushes to `main`, and tests whatever the current pre-release version of Cython is. This creates a home for some of the tests that we've discussed in https://github.com/Cantera/enhancements/issues/37, where running them on every push to a PR is more trouble than it's worth. I tested it by pushing this branch directly to `Cantera/cantera`. You can see the status (green checkmark) for this job reported here: https://github.com/Cantera/cantera/tree/python-ext-init. I'll delete that branch after we merge this PR.

ischoegl (approved):
Thanks, @speth … everything looks good to me!

PR #1458, opened 2023-03-20
Tweak SolutionArray IO

ischoegl:

This PR takes introduces minor improvements to the API after the merge of #1426

- Protect written data / enable overwriting of existing data
- Deprecate `loglevel` in `Sim1D::save/restore`
- Remove unnecessary output for `SolutionArray::save`

**Checklist**

### DIFF
--- a/include/cantera/base/Storage.h
+++ b/include/cantera/base/Storage.h
@@ -53,10 +53,17 @@ public:
     bool hasGroup(const string& id) const;
 
     //! Check whether path location exists.
-    //! If the file has write access, create location; otherwise exceptions are thrown.
+    //! If the location does not exist, an exception is thrown unless the *permissive*
+    //! flag is set; in this case, the method attempts to create a new location if the
+    //! file is accessed in write mode.
     //! @param id  storage location within file
-    //! @param permissive  if true, do not raise exceptions
+    //! @param permissive  if true, do not raise exceptions (default=false)
+    //! @returns  boolean indicating whether id is pre-existing
     bool checkGroup(const string& id, bool permissive=false);
+
+    //! Delete group
+    //! @param id  storage location within file
+    void deleteGroup(const string& id);
 
     //! Retrieve contents of file from a specified location
     //! @param id  storage location within file
--- a/interfaces/cython/cantera/_onedim.pyx
+++ b/interfaces/cython/cantera/_onedim.pyx
@@ -1523,7 +1523,7 @@ cdef class Sim1D:
             return self.sim.fixedTemperatureLocation()
 
     def save(self, filename='soln.yaml', name='solution', description=None,
-             loglevel=1, compression=0):
+             loglevel=None, overwrite=False, compression=0):
         """
         Save the solution in YAML or HDF format.
 
@@ -1533,17 +1533,24 @@ cdef class Sim1D:
             solution name within the file
         :param description:
             custom description text
+        :param overwrite:
+            Force overwrite if name exists; optional (default=`False`)
         :param compression:
             compression level 0..9; optional (HDF only)
 
         >>> s.save(filename='save.yaml', name='energy_off',
         ...        description='solution with energy eqn. disabled')
 
+        .. versionchanged:: 3.0
+            Argument loglevel is no longer supported
         """
+        if loglevel is not None:
+            warnings.warn("Argument 'loglevel' is deprecated and will be ignored.",
+                DeprecationWarning)
         self.sim.save(stringify(str(filename)), stringify(name),
-                      stringify(description), loglevel, compression)
+                      stringify(description), overwrite, compression)
 
-    def restore(self, filename='soln.yaml', name='solution', loglevel=2):
+    def restore(self, filename='soln.yaml', name='solution', loglevel=None):
         """Set the solution vector to a previously-saved solution.
 
         :param filename:
@@ -1559,10 +1566,13 @@ cdef class Sim1D:
         >>> s.restore(filename='save.yaml', name='energy_off')
 
         .. versionchanged:: 3.0
-            Implemented return value for meta data
+            Implemented return value for meta data; loglevel is no longer supported
         """
+        if loglevel is not None:
+            warnings.warn("Argument 'loglevel' is deprecated and will be ignored.",
+                DeprecationWarning)
         cdef CxxAnyMap header
-        header = self.sim.restore(stringify(str(filename)), stringify(name), loglevel)
+        header = self.sim.restore(stringify(str(filename)), stringify(name))
         self._initialized = True
         return anymap_to_dict(header)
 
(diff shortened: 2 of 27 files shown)


include/cantera/base/Storage.h
```diff
@@ -58,6 +58,10 @@ class Storage
     //! @param permissive  if true, do not raise exceptions
     bool checkGroup(const string& id, bool permissive=false);
 
+    //! Unlink group
+    //! @param id  storage location within file
+    void unlinkGroup(const string& id);
```
speth: I'm guessing that "unlink" is jargon for "delete"? I think the latter would be a better name -- even the base HDF5 API seems to be moving in this direction (see [here](https://docs.hdfgroup.org/hdf5/v1_12/group___h5_g.html#gacb843cbd5bbb816cfa9c855463d1e51c)).

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -1523,7 +1523,7 @@ cdef class Sim1D:
             return self.sim.fixedTemperatureLocation()
 
     def save(self, filename='soln.yaml', name='solution', description=None,
-             loglevel=1, compression=0):
+             loglevel=None, overwrite=False, compression=0):
```
bryanwweber, line 1526: I think you can avoid breaking code for anyone using positional arguments here by moving `overwrite` to the end. It'd be unfortunate if someone misses the warnings and accidentally overwrites a file due to positive integers being truthy. Minor concern, but I hope it's not a big lift to change 🤓

test/general/test_serialization.cpp
```diff
@@ -417,12 +417,12 @@ TEST(Storage, groups)
     // create group implicitly if permissive flag is set
     EXPECT_FALSE(file->hasGroup("one"));
     EXPECT_THROW(file->checkGroup("one"), CanteraError);
-    EXPECT_TRUE(file->checkGroup("one", true));
+    EXPECT_FALSE(file->checkGroup("one", true));
     EXPECT_TRUE(file->hasGroup("one"));
```
speth, line 421: I think I'm missing something here -- if `checkGroup("one", [false])` raises an exception, then it can't create the group with the `permissive=true` flag either, right? But how does `hasGroup("one")` return `true` afterwards?

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -1559,10 +1566,13 @@ cdef class Sim1D:
         >>> s.restore(filename='save.yaml', name='energy_off')
 
         .. versionchanged:: 3.0
-            Implemented return value for meta data
+            Implemented return value for meta data; loglevel is no longer supported
```
bryanwweber, line 1569: Does it make sense to add a second version changed directive here?

samples/cxx/flamespeed/flamespeed.cpp
```diff
@@ -141,18 +137,16 @@ int flamespeed(double phi, bool refine_grid, int loglevel)
         flame.solve(loglevel,refine_grid);
         double flameSpeed_mix = flame.value(flowdomain,
                                             flow->componentIndex("velocity"),0);
-        print("Flame speed with mixture-averaged transport: {} m/s\n",
-              flameSpeed_mix);
-        flame.save(fileName, "mix", "Solution with mixture-averaged transport", 0);
+        print("Flame speed with mixture-averaged transport: {} m/s\n", flameSpeed_mix);
```
bryanwweber, line 140: Does this actually fill in the variable to the string? I see it was this way before, but is this a bug to fix?

speth (changes requested):
Thanks, @ischoegl. I'm definitely pleased to see the `loglevel` flag go away in these instances. I just had one naming suggestion, and a point of clarification on `checkGroup`, which may suggest some additional documentation for that method.

interfaces/cython/cantera/_onedim.pyx
ischoegl: Compression (which only applies to HDF) is a very new feature for `save` that was never released (it first appeared when #1385 was merged on Jan 12); as we're still in alpha, I believe this is ok. I was more concerned about loglevel being caught.

interfaces/cython/cantera/_onedim.pyx
ischoegl: Since both were changed in the same version, it should be ok to combine?

include/cantera/base/Storage.h
ischoegl: Python uses `unlink` also, but I’m not opposed to changing this whatsoever.

test/general/test_serialization.cpp
ischoegl: The docstring was somewhat ambiguous / implementation was not clear, as the method captures both read and write modes. The docstring now reads:
```C++
    //! Check whether path location exists.
    //! If the location does not exist, an exception is thrown unless the *permissive*
    //! flag is set; in this case, the method attempts to create a new location if the
    //! file is accessed in write mode.
    //! @param id  storage location within file
    //! @param permissive  if true, do not raise exceptions (default=false)
    //! @returns  boolean indicating whether id is pre-existing
    bool checkGroup(const string& id, bool permissive=false);
```
Based on this,  `false` can be returned while creating a group as long as `permissive` is `true`. While this has been the intent before, the previous implementation was not consistent; this is corrected now.

samples/cxx/flamespeed/flamespeed.cpp
ischoegl: Output relies on `fmt::print` and works as expected. For `phi=1.`, the output is
```
Flame speed with mixture-averaged transport: 0.3800463682237259 m/s
```

ischoegl:
@speth and @bryanwweber ... thanks for the reviews! I believe all concerns are addressed.

samples/cxx/flamespeed/flamespeed.cpp
bryanwweber: Oh, thanks. I didn't see this was cpp code! 🤪

interfaces/cython/cantera/_onedim.pyx
bryanwweber: Oh, ok. This is what I get for commenting when I haven't been paying attention to the changes! You're right, this probably makes sense then the way it is.

interfaces/cython/cantera/_onedim.pyx
bryanwweber: I was thinking the second directive would make the existence of both changes more prominent, as opposed to putting the fairly unrelated things in the same HTML box. Just a thought though ☺️

ischoegl: The way it renders on sphinx, I believe having a single directive per version is preferable? (I realize that this is subjective). From that perspective, I prefer it the way it is at the moment.

bryanwweber: Alright, thanks for checking!

speth (approved):

PR #1463, opened 2023-03-22
Counterflow solution strategy

g3bk47:
Currently, an "extinct flame" solution is treated as successful solution for counterflow flames. Therefore, trying to find a solution by first disabling the energy equation is never tried. However, this can significantly increase the chance of finding the non-extinct solution. This PR modifies the solution strategy for counterflow flames so that after an extinct flame solution is found, the normal workflow of trying another solution with disabled energy equation is applied. Previous discussion can be found in https://github.com/Cantera/enhancements/issues/155.

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/_onedim.pyx
+++ b/interfaces/cython/cantera/_onedim.pyx
@@ -1336,7 +1336,10 @@ cdef class Sim1D:
             # If initial solve using energy equation fails, fall back on the
             # traditional fixed temperature solve followed by solving the energy
             # equation
-            if not solved:
+            if not solved or self.extinct():
+                if self.extinct():
+                    self.set_initial_guess(*self._initial_guess_args,
+                                           **self._initial_guess_kwargs)
                 log('Initial solve failed; Retrying with energy equation disabled')
                 self.energy_enabled = False
                 try:


ischoegl (commented):
@g3bk47 … thank you for the PR! Apart from squashing the ‘fixing line length’ commit, could you comment on the odd dip in temperature for U=7 m/s in your original post https://github.com/Cantera/enhancements/issues/155#issue-1306713572? I’m a little concerned as is looks like an artifact that we’d like to avoid. I realize that you pointed this out yourself …

g3bk47:
Hi @ischoegl,
I merged the two commits and took a closer look at the example from https://github.com/Cantera/enhancements/issues/155:
The cause for the odd dip (black dashed line below) was that I set the tolerances to very high values in my test script so that some of the flame "solutions" contained only a hand full of points. After changing
`f.set_refine_criteria(ratio=4, slope=0.2, curve=0.3, prune=0.04)`
to
`f.set_refine_criteria(ratio=2.5, slope=0.05, curve=0.05, prune=0.005)`
I get pretty much perfect agreement between the direct solution using the new code (blue solid line) and the continuation method (red dotted line) in terms of maximum temperature:
T

ischoegl (approved):
Thanks, @g3bk47 ... this looks good to me, and the explanation makes sense. Regarding the CI failures, they are unrelated - see #1471.

PR #1462, opened 2023-03-22
Solutionarray summary

ischoegl:
**Changes proposed in this pull request**

This PR adds a utility function `SolutionArray::info`, which provides a concise summary of the content. The styling is inspired by *pandas*' `DataFrame.info`, although it is implemented in C++ and thus available for all API's. Under the hood, the C++ library `fmt` is used to format content.

~In addition, this PR fixes a glitch in the order of `SolutionArray::componentNames`, where the native storage mode `TDY` was alphabetically sorted as `DTY` (same fix is also added in #1464).~

**If applicable, provide an example illustrating new features this pull request is introducing**

```
In [1]: import cantera as ct
   ...: gas = ct.Solution("h2o2.yaml")
   ...: arr = ct.SolutionArray(gas, 20, extra={"spam": "eggs"})

In [2]: arr
Out[2]:
      T          D   H2    H    O   O2   OH  H2O  HO2  H2O2   AR   N2  spam
0   300  0.0818939    1    0    0    0    0    0    0     0    0    0  eggs
1   300  0.0818939    1    0    0    0    0    0    0     0    0    0  eggs
2   300  0.0818939    1    0    0    0    0    0    0     0    0    0  eggs
3   300  0.0818939    1    0    0    0    0    0    0     0    0    0  eggs
4   300  0.0818939    1    0    0    0    0    0    0     0    0    0  eggs
..  ...        ...  ...  ...  ...  ...  ...  ...  ...   ...  ...  ...   ...
15  300  0.0818939    1    0    0    0    0    0    0     0    0    0  eggs
16  300  0.0818939    1    0    0    0    0    0    0     0    0    0  eggs
17  300  0.0818939    1    0    0    0    0    0    0     0    0    0  eggs
18  300  0.0818939    1    0    0    0    0    0    0     0    0    0  eggs
19  300  0.0818939    1    0    0    0    0    0    0     0    0    0  eggs

[20 rows x 13 components; state='TDY']
```

and
```
In [3]: import numpy as np
   ...: arr.set_equivalence_ratio(np.linspace(.5, 1.5, 20), "H2", "O2:1,N2:10")
   ...: arr.equilibrate("HP")
   ...: arr.spam = np.linspace(-1e-6, 1e-6, 20)

In [4]: print(arr.info(rows=7, width=100))
           T         D           H2            H  ...         H2O2   AR        N2         spam
[...]

PR #1464, opened 2023-03-25
Remove vestigial onedim Python functions and methods

### LINKED ISSUE #166: Link FlameBase.to_solution_array to C++ SolutionArray
**Abstract**

At the moment, the `FlameBase.to_solution_array` method assembles a `SolutionArray` in Python instead of accessing the `Domain1D::asArray` serialization method of the C++ object. One consequence is that the output order of components is incorrect.

**Motivation**

* What problem is it trying to solve? ... remove/deprecate redundant code
* Who is affected by the change? ... users of Cantera 3.0
* Why is this a good solution? ... the conversion should use the same tools as the C++ save/restore methods

**Possible Solutions**

There are two options:
1. Implement `Sim1D::asArray(const string& name)` and link to `FlameBase.to_solution_array`; one complication is that the C++ route does not normalize, which is the current default behavior in Python.
2. Implement `Domain1D::asArray()` and link to a new `Domain1D.to_solution_array`; here, the complication is that `Domain1D` knows its container, but is is accessed as `OneDim`, which does not hold the solution vector; as in other instances, the code uses raw pointers internally, which ideally would have to be converted to smart pointers.

**References**

* Cantera/cantera#1385
* Cantera/cantera#1462

ischoegl:

After moving `SolutionArray` support to C++ in #1426 and updating HDF support in #1385, remaining Python versions are replaced, vestigial functions are deprecated, and code is removed where appropriate. Some minor glitches are fixed.

- Make `Sim1D` data vector accessible to `Domain1D`
- Implement `SolutionArray::normalize` to normalize mass/mole fractions
- Implement `ThermoPhase::nativeMode` to produce acronym describing storage mode
- Fix minor bugs in new C++ `SolutionArray` interface
- Rename some Python / C++ methods for consistency across interfaces
- Use C++ `SolutionArray` methods for the interface between Python `Sim1D` and `SolutionArray`
- Retain ability to restore legacy HDF using `h5py` for Cantera 3.0, but deprecate all associated methods
- Remove `OrderedDict`, which is no longer necessary
- Convert Python `FutureWarning`s to `DeprecationWarning`s
- Replace some hardcoded indices
- _**Deprecate `h5py` support**_

Resolves Cantera/enhancements#166

**Checklist**

### DIFF
--- a/include/cantera/oneD/Domain1D.h
+++ b/include/cantera/oneD/Domain1D.h
@@ -540,12 +540,19 @@ public:
         m_force_full_update = update;
     }
 
+    //! Set shared data pointer
+    void setData(shared_ptr<vector<double>>& data) {
+        m_data = data;
+    }
+
 protected:
     //! Retrieve meta data
     virtual AnyMap getMeta() const;
 
     //! Retrieve meta data
     virtual void setMeta(const AnyMap& meta);
+
+    shared_ptr<vector<double>> m_data; //!< data pointer shared from OneDim
 
     double m_rdt = 0.0;
     size_t m_nv = 0;
--- a/include/cantera/oneD/Domain1D.h
+++ b/include/cantera/oneD/Domain1D.h
@@ -562,7 +562,7 @@ public:
 
     //! Set shared data pointer
     void setData(shared_ptr<vector<double>>& data) {
-        m_data = data;
+        m_state = data;
     }
 
 protected:
@@ -572,7 +572,7 @@ protected:
     //! Retrieve meta data
     virtual void setMeta(const AnyMap& meta);
 
-    shared_ptr<vector<double>> m_data; //!< data pointer shared from OneDim
+    shared_ptr<vector<double>> m_state; //!< data pointer shared from OneDim
 
     double m_rdt = 0.0;
     size_t m_nv = 0;
--- a/include/cantera/oneD/Domain1D.h
+++ b/include/cantera/oneD/Domain1D.h
@@ -344,15 +344,27 @@ public:
      */
     AnyMap serialize(const double* soln) const;
 
-    //! Save the state of this domain as a SolutionArray
+    //! Save the state of this domain as a SolutionArray.
     /*!
      * @param soln local solution vector for this domain
+     * @todo  Despite the method's name, data are copied; the intent is to access data
+     *      directly in future revisions, where a non-const version will be implemented.
      *
      * @since  New in Cantera 3.0.
      */
     virtual shared_ptr<SolutionArray> asArray(const double* soln) const {
         throw NotImplementedError("Domain1D::asArray", "Needs to be overloaded.");
     }
+
+    //! Save the state of this domain to a SolutionArray.
+    /*!
+     * This method serves as an external interface for high-level API's; it does not
+     * provide direct access to memory.
+     * @param normalize If true, normalize concentrations (default=false)
+     *
+     * @since  New in Cantera 3.0.
+     */
+    shared_ptr<SolutionArray> toArray(bool normalize=false) const;
 
     //! Restore the solution for this domain from an AnyMap
     /*!
@@ -372,9 +384,17 @@ public:
      *
      * @since  New in Cantera 3.0.
      */
-    virtual void restore(SolutionArray& arr, double* soln) {
-        throw NotImplementedError("Domain1D::restore", "Needs to be overloaded.");
+    virtual void fromArray(SolutionArray& arr, double* soln) {
+        throw NotImplementedError("Domain1D::fromArray", "Needs to be overloaded.");
     }
+
+    //! Restore the solution for this domain from a SolutionArray.
+    /*!
+     * This method serves as an external interface for high-level API's.
+     * @param  arr SolutionArray defining the state of this domain
+     * @since  New in Cantera 3.0.
+     */
+    void fromArray(const shared_ptr<SolutionArray>& arr);
 
     //! Return thermo/kinetics/transport manager used in the domain
     //! @since  New in Cantera 3.0.
--- a/include/cantera/thermo/Phase.h
+++ b/include/cantera/thermo/Phase.h
@@ -291,6 +291,11 @@ public:
     //! and restoreState().
     virtual std::map<std::string, size_t> nativeState() const;
 
+    //! Return string acronym representing native state
+    //! @see nativeState
+    //! @since  New in Cantera 3.0
+    string nativeMode() const;
+
     //! Return a vector containing full states defining a phase.
     //! Full states list combinations of properties that allow for the
     //! specification of a thermodynamic state based on user input.
--- a/include/cantera/thermo/Phase.h
+++ b/include/cantera/thermo/Phase.h
@@ -291,7 +291,8 @@ public:
     //! and restoreState().
     virtual std::map<std::string, size_t> nativeState() const;
 
-    //! Return string acronym representing native state
+    //! Return string acronym representing the native state of a Phase.
+    //! Examples: "TP", "TDY", "TPY".
     //! @see nativeState
     //! @since  New in Cantera 3.0
     string nativeMode() const;
--- a/interfaces/cython/cantera/_onedim.pyx
+++ b/interfaces/cython/cantera/_onedim.pyx
@@ -291,15 +291,18 @@ cdef class Domain1D:
 
     property settings:
         """
-        Return comprehensive dictionary describing type, name, and simulation
-        settings that are specific to domain types.
+        Return comprehensive dictionary describing type, name, and simulation settings
+        that are specific to domain types.
+
+        .. versionchanged:: 3.0
+
+            The dictionary layout in Cantera 3.0 combines features of HDF and YAML
+            metadata used in Cantera 2.6 to form a uniform interface.
         """
         def __get__(self):
-            out = {
-                'Domain1D_type': type(self).__name__,
-                'name': self.name}
-
-            return out
+            cdef shared_ptr[CxxSolutionArray] arr
+            arr = self.domain.toArray(False)
+            return anymap_to_dict(arr.get().meta())
 
 
 cdef class Boundary1D(Domain1D):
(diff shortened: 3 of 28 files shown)


ischoegl:
@speth ... while #1462 is mostly about cosmetics, this PR addresses some hanging chads left after #1385 and #1426. Changes are ready for a review.

ischoegl:
I added commits that replace hardcoded indices in `StFlow`. With these changes, tests show that changing the storage order now works for axisymmetric cases, but not for free flames. As locating the remaining hard-coded bits (presumably someplace around the fixed point) is beyond the scope of this PR, I am stopping here.

While the `h5py` deprecation warning needs improvement (there is a replacement!), I don’t want to add another force-push 😜 .

include/cantera/oneD/Domain1D.h
```diff
@@ -540,13 +560,20 @@ class Domain1D
...
+
 protected:
     //! Retrieve meta data
     virtual AnyMap getMeta() const;
 
     //! Retrieve meta data
     virtual void setMeta(const AnyMap& meta);
 
+    shared_ptr<vector<double>> m_data; //!< data pointer shared from OneDim
+
```
speth: I'd like to find a more descriptive name for this array than just "data" (and likewise the `setData` method). Perhaps `m_state`, given that it's replacing `m_x` at the `Sim1D` level?

Also, we normally use the `vector_fp` typedef instead of writing `vector<double>`. This shows up several times in this PR. Unless you're suggesting preferring this name more broadly?

include/cantera/thermo/Phase.h
```diff
@@ -291,6 +291,11 @@ class Phase
     //! and restoreState().
     virtual std::map<std::string, size_t> nativeState() const;
 
+    //! Return string acronym representing native state
+    //! @see nativeState
+    //! @since  New in Cantera 3.0
+    string nativeMode() const;
```
speth: An example or two of what this string looks like would be helpful.

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -270,15 +289,18 @@ cdef class Domain1D:
...
         """
-        Return comprehensive dictionary describing type, name, and simulation
-        settings that are specific to domain types.
+        Return comprehensive dictionary describing type, name, and simulation settings
+        that are specific to domain types.
+
+        .. versionchanged:: 3.0
+
+            The dictionary layout in Cantera 3.0 combines features of HDF and YAML
+            metadata used in Cantera 2.6 to form a uniform interface.
```
speth: Assuming that we'll keep the `versionchanged` documentation indefinitely, this description is pretty opaque. Can you elaborate on what fields are now in the dict?

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -1055,51 +1101,16 @@ cdef class Sim1D:
...
-        else:
-            msg = ("Export of '{}' is not implemented")
-            raise NotImplementedError(msg.format(type(self).__name__))
+            Method to be removed after Cantera 3.0. Unused after moving conversion to
+            `SolutionArray` to the C++ core. As `FlameBase.to_solution_array` (which
+            itself is deprecated) bypasses this function, this service function has no
+            use and is now disabled.
+        """
+        raise NotImplementedError(
+            "This method is no longer supported and will be removed after Cantera 3.0.")
```
speth: The claim that this is a "deprecated" method, but then having it just be "not implemented" is kind of strange. I think we need to choose between the two options:

1. Assume the existing method is an implementation detail that no one is using directly, and just remove it (or rely on the fact that a major version bump allows breaking changes)
2. Leave an implementation of this method, with a deprecation warning.

interfaces/cython/cantera/composite.py
```diff
@@ -20,6 +24,8 @@ def _import_h5py():
     except importlib.metadata.PackageNotFoundError:
         raise ImportError('Method requires a working h5py installation.')
     else:
+        warnings.warn(
+            "Support for 'h5py' to be removed after Cantera 3.0.", DeprecationWarning)
```
speth: I think this is just excess noise for the user -- `_import_h5py` is an internal method, and `SolutionArray.read_hdf` (the one place where this is called) already raises a `DeprecationWarning`.

interfaces/cython/cantera/onedim.py
```diff
@@ -396,11 +410,32 @@ def write_csv(self, filename, species='X', quiet=True, normalize=True):
...
 
+    def to_array(self, domain=None, normalize=False):
+        """
+        Retrieve domain data as a `SolutionArray` object.
+
+        :param domain:
+            Domain to be converted; by default, the method retrieves the flow domain
+        :param normalize:
+            Boolean flag indicating whether mass/mole fractions should be normalized
+            (default=`False`)
```
speth: `` `False` `` doesn't get formatted nicely by Sphinx here. Since the default is also shown in the method signature, we could just not mention it again here.

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -562,13 +584,35 @@ cdef class _FlowBase(Domain1D):
...
+
     property settings:
         """
-        Get a dictionary describing type, name, and simulation specific to this domain
-        type.
+        .. deprecated:: 3.0
+
+            Specialization to be removed after Cantera 3.0, as it is specific to the
+            legacy HDF structure for metadata. For new behavior of the getter, use the
+            temporary method 'get_settings3'; the setter will be removed, but is
```
speth:
```suggestion
            temporary method `get_settings3`; the setter will be removed, but is
```

interfaces/cython/cantera/onedim.py
```diff
@@ -622,7 +671,18 @@ def read_hdf(self, filename, group=None, restore_boundaries=True, normalize=True
...
     def settings(self):
-        """ Return a dictionary listing simulation settings """
+        """
+        Return a dictionary listing simulation settings
+
+        .. deprecated:: 3.0
+
+            To be removed after Cantera 3.0. The getter is replaceable by
+            `Domain1D.settings`; for the setter, use setters for individual settings.
+            Also see deprecation note in `_FlowBase.settings`.
```
speth: Redirecting users to the documentation for `_FlowBase.settings` is sort of unhelpful, since that method does not show up in the Sphinx docs -- only the derived classes like `IdealGasFlow` are there (but these do include the documentation for the `settings` property)

src/oneD/Domain1D.cpp
```diff
@@ -145,6 +145,31 @@ AnyMap Domain1D::serialize(const double* soln) const
...
+    if (!m_data) {
+        throw CanteraError("Domain1D::toArray",
+            "Domain needs to be installed in a container before calling asArray.");
+    }
+    if (!normalize) {
+        return asArray(m_data->data() + m_iloc);
+    }
+    auto ret = asArray(m_data->data() + m_iloc);
+    ret->normalize();
+    return ret;
```
speth: Very minor simplification:
```suggestion
    auto ret = asArray(m_data->data() + m_iloc);
    if (normalize) {
        ret->normalize();
    }
    return ret;
```

test/general/test_composite.cpp
```diff
@@ -79,6 +79,52 @@ TEST(SolutionArray, simple)
...
+    for (size_t loc = 0; loc < arr->size(); loc++) {
+        state = arr->getState(loc);
+        for (size_t i = 2; i < state.size() - 1; i++) {
+            ASSERT_NEAR(state[i], state[2], 1e-12);
+            ASSERT_NE(state[i], newState[i]);
+            ASSERT_LE(state[i], 1.);
+            ASSERT_GE(state[i], 0.);
+        }
+        ASSERT_EQ(state.back(), 0.);
+    }
```
speth: A couple comments here would help with understanding what behavior is being tested. It takes a moment to recognize that comparing `state[2]` to all species but the last one is a sensible thing to do.

test/python/test_composite.py
```diff
@@ -742,15 +740,14 @@ def check(a, b, atol=None):
...
-                                        if i]))
+            b.restore_data(dict([tup for i, tup in enumerate(data.items()) if i]))
 
         # add extra column
         t = np.arange(10, dtype=float)
 
         # auto-detection of extra
         b = ct.SolutionArray(self.gas)
-        data_mod = OrderedDict(data)
+        data_mod = dict(data) # create a copy
```
speth: PEP-8 suggests 2 spaces before the `#`.

speth (changes requested):
Thanks, @ischoegl, I think this is generally fine. I just had some minor suggestions/comments.

interfaces/cython/cantera/_onedim.pyx
ischoegl: It is indeed an implementation detail, which is not reachable any longer except by calling it directly. From that perspective, I prefer the breaking change.

interfaces/cython/cantera/onedim.py
ischoegl: Removing redirection to additional deprecation note.

interfaces/cython/cantera/_onedim.pyx
ischoegl: The main thing changing is that some domain-specific settings were added. I believe the only thing a user needs to know that there was a change in output, where HDF/YAML origins are not really needed. Details are quite hard, as they are specific to a specialization and this is a catch-all service function.

include/cantera/oneD/Domain1D.h
ischoegl: I renamed `m_data` to `m_state` as suggested.

Regarding `vector_fp` vs `vector<double>`, I do indeed have a preference for the latter, as it is self-documenting while the former is somewhat hidden in `ct_defs.h`. I believe the extra 5 characters are worthwhile, especially now that we no longer require `std::`.

ischoegl:
@speth ... thanks for the review! I believe I addressed most of the comments, with a remaining comment about `vector_fp`.

speth (approved):
Looks good to me!

PR #1465, opened 2023-03-26
Rebuild for post1

bryanwweber:
Updates to the build configuration for the 2.6 branch to explicitly indicate that Python 3.11 is not supported. Once this is merged, I will kick off a manual build in the pypi-packages repo, also on the 2.6 branch, since that's working now: https://github.com/Cantera/pypi-packages/actions/runs/4516536636

### DIFF
--- a/interfaces/cython/pyproject.toml
+++ b/interfaces/cython/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools>=43.0.0", "wheel", "Cython>=0.29.12"]
+requires = ["setuptools>=43.0.0,<61", "wheel", "Cython>=0.29.12"]
 build-backend = "setuptools.build_meta"
--- a/interfaces/cython/setup.cfg.in
+++ b/interfaces/cython/setup.cfg.in
@@ -42,7 +42,7 @@ include_package_data = True
 install_requires =
     numpy >= 1.12.0
     ruamel.yaml >= 0.15.34
-python_requires = >=@py_min_ver_str@
+python_requires = >=@py_min_ver_str@,<3.11
 packages =
     cantera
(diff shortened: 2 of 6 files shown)


ischoegl (commented):
:+1: Thanks! The contraction in the last commit message looks odd, but otherwise the PR looks great!

bryanwweber:
> 👍 Thanks! The contraction in the last commit message looks odd, but otherwise the PR looks great!

Yeah, wrote the commit message on the CLI and I always forget the quoting rules, so I just omit quotes in sentences 😄

ischoegl (approved):

ischoegl:
As an aside: should we place similar safeguards into `main`? At some point, Python 3.12 will start to create similar issues ...

bryanwweber:
> As an aside: should we place similar safeguards into `main`? At some point, Python 3.12 will start to create similar issues ...

Probably depends on when 3.0.0 will come out. 3.12 is slated for October, so I'd say if Cantera is going to come out in late Summer, we should probably support 3.12 and set the version to `<3.13`. YMMV 😄

ischoegl:
> > As an aside: should we place similar safeguards into `main`? At some point, Python 3.12 will start to create similar issues ...
> 
> Probably depends on when 3.0.0 will come out. 3.12 is slated for October, so I'd say if Cantera is going to come out in late Summer, we should probably support 3.12 and set the version to `<3.13`. YMMV 😄

My preference would be to put the safeguards in for 3.12 while we're aware of the need, and update version numbers if we decide to include 3.12 for Cantera 3.0.

ischoegl:
@bryanwweber ... see #1467; as I started with a cherry-pick, I guess you'll be the author :wink:.

PR #1467, opened 2023-03-26
Add safeguards against Python 3.12

ischoegl:

Implement safeguards against Python 3.12 in `main`; the PR replicates similar changes implemented in #1465 for updates of Cantera 2.6. *Edit: added a check in SCons to abort attempts that will fail before compiling.*

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -56,6 +56,7 @@ Additional command options:
 #  --output=<fname> ... send output to file (reST only)
 #
 # Other features not listed above:
+# 'scons sdist' - Build PyPI packages.
 # 'scons <command> dump' - Dump the state of the SCons environment to the
 #                          screen instead of doing <command>, for example
 #                          'scons build dump'. For debugging purposes.
@@ -1681,10 +1682,13 @@ env['python_cmd_esc'] = quoted(env['python_cmd'])
 
 # Python Package Settings
 python_min_version = parse_version("3.8")
+# Newest Python version not supported/tested by Cantera
+python_max_p1_version = parse_version("3.12")
 # The string is used to set python_requires in setup.cfg.in
-env['py_min_ver_str'] = str(python_min_version)
-# Note: cython_min_version is redefined below if the Python version is 3.8 or higher
-cython_min_version = parse_version('0.23')
+env["py_requires_ver_str"] = f">={python_min_version}"
+if env["python_sdist"] or env["package_build"]:
+    env["py_requires_ver_str"] += f",<{python_max_p1_version}"
+cython_min_version = parse_version("0.29.12")
 numpy_min_version = parse_version('1.12.0')
 
 # We choose ruamel.yaml 0.15.34 as the minimum version
@@ -1814,6 +1818,20 @@ if env['python_package'] != 'none':
                 f"{python_min_version} or newer is required. In order to install "
                 "Cantera without Python support, specify 'python_package=none'.")
             sys.exit(1)
+    elif python_version >= python_max_p1_version:
+        if env["python_package"] in ("minimal", "full", "default"):
+            if env["python_sdist"] or env["package_build"]:
+                # An error is triggered as the pip wheel cannot be built for
+                # 'scons build' due to safeguards in setup.cfg(.in) files.
+                logger.error(
+                    f"Python version is incompatible. Found {python_version}, which "
+                    "is not yet supported for package builds.")
+                sys.exit(1)
+            logger.warning(
+                f"Python {python_version} is not supported for Cantera "
+                f"{env['cantera_version']}. Python versions {python_max_p1_version} and "
+                "newer are untested and may result in unexpected behavior. Proceed "
+                "with caution.")
     elif env["python_package"] == "minimal":
         # If the minimal package was specified, no further checking needs to be done
         logger.info(f"Building the minimal Python package for Python {python_version}")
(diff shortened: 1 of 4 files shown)


SConstruct
```diff
@@ -1814,6 +1816,13 @@ if env['python_package'] != 'none':
                 f"{python_min_version} or newer is required. In order to install "
                 "Cantera without Python support, specify 'python_package=none'.")
             sys.exit(1)
+    elif python_version >= python_next_version:
+        if env["python_package"] in ("minimal", "full", "default"):
+            logger.error(
+                f"Python version is incompatible. Found {python_version}, which is "
+                f"not yet supported. In order to install Cantera without Python "
+                "support, specify 'python_package=none'.")
+            sys.exit(1)
```
bryanwweber: I think it makes more sense for this to be a warning. For the scons builder, it's more likely to work by "accident" with new Python versions it seems to me. Plus, if you're building from source, it's implied you have a little more knowledge. The concern to me is more with the pip install crowd getting a bad error message. 

The minimal interface also has a different set of requirements, so I don't think it needs to be included here.

SConstruct
```diff
@@ -1681,10 +1681,12 @@ env['python_cmd_esc'] = quoted(env['python_cmd'])
 
 # Python Package Settings
 python_min_version = parse_version("3.8")
+ # next (unsupported) version
+python_next_version = parse_version("3.12")
```
bryanwweber: Would `python_max_version` be better for symmetry and to indicate the purpose of this number?

bryanwweber (changes requested):
Thanks Ingmar! Couple of comments

SConstruct
ischoegl: :+1: ... I was vacillating between warning and error, and ended up with the latter as it was my assumption that the wheel build may fail when it comes to the Python installation? Still switched to warning as suggested.

ischoegl: I guess the CI failure on 3.11.2 when forcing a 'max' version <=3.11 confirms my suspicion that wheel builds will fail:
```
ERROR: Package 'cantera' requires a different Python: 3.11.2 not in '<=3.11,>=3.8'
``` 
I am reverting to what I had before, where the error will prevent failed attempts; those who really want to test can always edit `SConstruct`

SConstruct
ischoegl: Unfortunately, testing for versions <=3.11 fails, as it excludes patch numbers. Checking against the upcoming release is simpler.

SConstruct
ischoegl: I updated the error message to indicate that users will have to edit a specific variable in `SConstruct` in case they want to proceed.

bryanwweber: I still strongly prefer the warning. I think if you do `<=3.11.*` it'll work

ischoegl: Sorry, but the build *will* fail, so there is absolutely no point in issuing a warning. Regarding the version string, it’s still easier to stick with 3.12.

bryanwweber: I still strongly prefer the warning. I think if you do `<=3.11.*` it'll work

ischoegl: This unfortunately doesn’t address an attempt to compile with 3.12, which *will* fail due to the setup configuration for the pip wheel. For 3.11.*, what is suggested obviously works?

ischoegl: Regarding `3.11.*`, it is brittle:
```
>>> from pkg_resources import parse_version
>>> parse_version("3.11.*")
/opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/lib/python3.10/site-packages/pkg_resources/__init__.py:123: PkgResourcesDeprecationWarning: 3.11.* is an invalid version and will not be supported in a future release
  warnings.warn(
<LegacyVersion('3.11.*')>
>>> parse_version("3.12")
<Version('3.12')>
```
and
```
>>> from packaging.version import parse
>>> parse('3.11.*')
<LegacyVersion('3.11.*')>
>>> parse('3.12')
<Version('3.12')>
```
In a nutshell, both routes used in SCons produce a `LegacyVersion` if you use `3.11.*`. I am strongly opposed to going that route.

ischoegl:
I added the following clarification to the `SConstruct` safeguard:
```python
# An error is triggered as the pip wheel cannot be built for 'scons build'
# due to safeguards in setup.cfg(.in) files. Untested Python versions can
# introduce breaking changes; tests should be limited to experienced users.
```
As recent experience has shown, building wheels with `pip` is relatively brittle (or at least a lot more than other things we do). From my perspective, blocking attempts is - in this instance - preferable over bug reports for things that mostly occur while compiling with what I consider the bleeding edge.

ischoegl:
@bryanwweber ... thanks for your comments. Based on feedback, I updated the approach and am now conditioning behavior on the `python_sdist` and `package_build` flags: if `False`, there is only a warning as suggested; if it is `True`, SCons exits with an error. Also, the safeguards against newer Python versions in `setup.cfg` are only applied for packaged builds (*~Edit: there is a pending push, but GH currently has partial outages.~*).

SConstruct
bryanwweber: I still think that `python_next_version` isn't quite right... if we release 3.0 and it doesn't support 3.12, but October passes and 3.12 is released, this name implies to me it should be bumped to 3.13. I also get that it's not quite the max version, it's the max version + 1 🚲 🏠

bryanwweber: I still think that `python_next_version` isn't quite right... if we release 3.0 and it doesn't support 3.12, but October passes and 3.12 is released, this name implies to me it should be bumped to 3.13. I also get that it's not quite the max version, it's the max version + 1 🚲 🏠

bryanwweber:
Thanks for the change @ischoegl! I had one small nitpick from my previous review.

SConstruct
ischoegl: `python_max_p1_version` it is then (also updated the associated comment). The intent is that we bump the version whenever CI is updated to include Python 3.12. I believe the 🚲🏡 looks nice at this point?

ischoegl:
@bryanwweber … I believe that all comments are addressed.

SConstruct
```diff
@@ -1681,10 +1682,13 @@ env['python_cmd_esc'] = quoted(env['python_cmd'])
 
 # Python Package Settings
 python_min_version = parse_version("3.8")
+# Newest Python version not supported/tested by Cantera
+python_max_p1_version = parse_version("3.12")
```
bryanwweber, line 1686: I'm curious what this variable stands for.

bryanwweber (approved):
Thanks @ischoegl LGTM. One question for my own edification :smile:

SConstruct
ischoegl: Max supported version plus 1 … tried to heed to your suggestion 😃

PR #1471, opened 2023-04-01
Fix clash of names with upstream function

ischoegl:
HighFive 2.7.0 introduces `h5py`-style support for booleans, which is close to identical with what was already implemented in `Storage.cpp`. HighFive uses the exact same name for a new `create_enum_boolean` function, which clashes with an existing local function in `Storage.cpp`, causing CI failures when conda packages `highfive` with versions 2.7.0 and newer are used. 

The resolution is to use the full name for older versions of HighFive (2.7.0 and older), while using HighFive's implementation for boolean support for HighFive 2.7.1 and newer (this avoids a change of their API from 2.7.0 to 2.7.1). For the time being, the git submodule remains on 2.6.2, while conda will grab the newest package available. 

Other updates:
- use system installation of HighFive by default and other SCons logic fixes
- detect and display HighFive version

See https://github.com/BlueBrain/HighFive/issues/713 and https://github.com/BlueBrain/HighFive/issues/724

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -931,7 +931,7 @@ if "sdist" in COMMAND_LINE_TARGETS:
     elif env["python_package"] in ("full", "y"):
         logger.error("'sdist' target was specified. Cannot also build Python package.")
         sys.exit(1)
-    for ext_dependency in ("sundials", "fmt", "yamlcpp", "eigen"):
+    for ext_dependency in ("sundials", "fmt", "yamlcpp", "eigen", "highfive"):
         if env[f"system_{ext_dependency}"] == "y":
             logger.error(f"'sdist' target was specified. Cannot use 'system_{ext_dependency}'.")
             sys.exit(1)
@@ -1542,9 +1542,32 @@ else:
     if not env["use_hdf5"] and env["hdf_support"] == "y":
         config_error("HDF5 support has been specified but libraries were not found.")
 
+if env["use_hdf5"] and env["system_highfive"] in ("y", "default"):
+
+    if conf.CheckLibWithHeader(
+            "hdf5", "highfive/H5File.hpp", language="C++", autoadd=False):
+        env["system_highfive"] = True
+
+        highfive_include = "<highfive/H5Version.hpp>"
+        h5_version_source = get_expression_value(
+            [highfive_include], "QUOTE(HIGHFIVE_VERSION)")
+        retcode, h5_lib_version = conf.TryRun(h5_version_source, ".cpp")
+        if retcode:
+            env["HIGHFIVE_VERSION"] = h5_lib_version.strip()
+        else:
+            config_error("Detected invalid HighFive configuration.")
+
+        highfive_include = "<highfive/H5DataType.hpp>"
+        logger.info("Using system installation of HighFive library.")
+
+    elif env["system_highfive"] == "y":
+        config_error(
+            "Expected system installation of HighFive library, but it is either "
+            "corrupted or could not be found.")
+
 if env["use_hdf5"] and env["system_highfive"] in ("n", "default"):
     env["system_highfive"] = False
-    if not os.path.exists("ext/eigen/HighFive/include"):
+    if not Path("ext/HighFive/include").is_dir():
         if not os.path.exists(".git"):
             config_error("HighFive is missing. Install HighFive in ext/HighFive.")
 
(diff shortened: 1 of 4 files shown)


ischoegl:
While this is ready for a review, I am leaving this in draft mode until HighFive 2.7.1 is released. ~The current CI failures on windows-2022 are caused by a known issue with conda's `highfive` package 2.7.0 for MSVC 14.1 that are presumably unrelated to Cantera.~ ~*Edit: pinned conda’s `highfive` to older versions.*~

*Update: while HighFive 2.7.1 is released (and is expected to fully resolve all issues), the conda packages aren’t generated/uploaded yet.*

ischoegl:
@speth ... as expected, HighFive 2.7.1 conda packages resolve the CI issues (other than #1473, which is separate).

SConstruct
```diff
@@ -1564,28 +1577,55 @@ if env["use_hdf5"] and env["system_highfive"] in ("n", "default"):
...
+        lines = lines.replace("@PROJECT_VERSION_MINOR@", h5_version[1])
+        lines = lines.replace("@PROJECT_VERSION_PATCH@", h5_version[2])
+
+        # write file - file name stem eliminates '.in'
+        fname = fname.parent / fname.stem
+        with open(fname, "w") as fid:
+            fid.write(lines)
+
+    cmake_replace("ext/HighFive/CMakeLists.txt")
+    highfive_include = '"../ext/HighFive/include/highfive/H5Version.hpp"'
```
speth: This should be done as part of a SCons action -- right now, it's being repeated every time `scons` is called.

Also, I think the whole procedure for generating `H5Version.hpp` should be moved to `ext/SConscript`, which is where similar things like generating `sundials_config.h` are handled.

SConstruct
```diff
@@ -1564,28 +1577,55 @@ if env["use_hdf5"] and env["system_highfive"] in ("n", "default"):
...
+        #include {highfive_include}
+        // Decode corrupted version string defined prior to HighFive 2.7.1
+        #define STRINGIFY_VALUE(s) STRINGIFY_NAME(s)
+        #define STRINGIFY_NAME(s) #s
+        int main(int argc, char** argv) {{
+            std::cout << STRINGIFY_VALUE(HIGHFIVE_VERSION);
+            return 0;
+        }}
+    """)
+    retcode, h5_lib_version = conf.TryRun(h5_lib_version, ".cpp")
```
speth: I think this is essentially the same thing as the existing `get_expression_value` function, and could be replaced by just calling:
```py
h5_version_source = get_expression_value([highfive_include], "QUOTE(HIGHFIVE_VERSION)")
retcode, h5_lib_version = conf.TryRun(h5_version_source, ".cpp")
```

SConstruct
```diff
@@ -2147,6 +2190,7 @@ cdefine('LAPACK_NAMES_LOWERCASE', 'lapack_names', 'lower')
 cdefine('CT_USE_LAPACK', 'use_lapack')
 cdefine("CT_USE_HDF5", "use_hdf5")
 cdefine("CT_USE_SYSTEM_HIGHFIVE", "system_highfive")
+cdefine("HIGHFIVE_BOOLEAN", "highfive_boolean")
```
speth: To avoid namespace pollution (especially if `HIGHFIVE_` is already the prefix used by HIGHFIVE), we should use a different name like `CT_USE_HIGHFIVE_BOOLEAN`.

src/base/Storage.cpp
```diff
@@ -29,17 +29,23 @@
 
 namespace h5 = HighFive;
 
+#if HIGHFIVE_BOOLEAN
```
speth: Shouldn't this be `#ifdef`?

speth (changes requested):
Thanks for finding the solution to this, @ischoegl -- the CI error messages definitely did not make it clear what was going wrong. I had some suggestions that I think will simplify the configuration checks a bit.

SConstruct
ischoegl: @speth ... while it's clear why you are requesting this, the details get a little gnarly. The situation is that `ext/SConscript` is processed after the configuration step, but I need to create `H5Version.hpp` during configuration, as I read the version number from that file (I found out the hard way that it's good to know what version is being compiled when debugging for this PR).

As an alternative that does not rewrite `H5Version.hpp` every time `scons` is called, I am deleting the file during `scons clean`, and write it only if the file does not exist. I also left a comment explaining why replacement needs to happen during the configuration step.

ischoegl:
> Thanks for finding the solution to this, @ischoegl -- the CI error messages definitely did not make it clear what was going wrong. 

I guess I need to give a shoutout to folks over at the HighFive project, who were extremely helpful and responsive in the debugging effort.

Other than that, I rebased and addressed three of the four suggestions. For the last one, I no longer rewrite the file, but am not sure how to move things out of `SConstruct` without major overhauls of the `scons` workflow.

SConstruct
speth: Ah, thanks for the explanation. I think I have a way around this complexity. Basically, we can simplify the logic to:
- if `system_highfive` is set to `y` or `default`, try to get the version from `H5Version.hpp` using `get_expression_value`.
- If that fails, parse the version from `CMakeLists.txt`.
- Then, use the version to determine the value of `CT_USE_HIGHFIVE_BOOLEAN`.

This way, there's no need to generate `H5Version.hpp` (which is never included elsewhere), and this also avoids leaving the `ext/HighFive` directory in a "dirty" state.

ischoegl: Ok - I guess I was too focused on getting everything done at once! I adopted your suggestion.

ischoegl:
@speth ... thanks for the suggestion. I believe that everything is taken care of at this point. 

I submitted a minor PR #1477 that could be adopted before the version bump, but it's not a major issue (it's mostly about updating annoying formatting for something that I've been working on).

SConstruct
```diff
@@ -145,6 +145,8 @@ if "clean" in COMMAND_LINE_TARGETS:
     remove_file("interfaces/matlab/ctpath.m")
     for name in Path("interfaces/matlab/toolbox").glob("ctmethods.*"):
         remove_file(name)
+    for name in Path("ext/HighFive/include/highfive").glob("*.in"):
+        remove_file(name.parent / name.stem)
```
speth: This is no longer necessary if we aren't generating `H5Version.hpp`, right?

speth (dismissed):
Thanks, @ischoegl, this looks good to me. Just one small simplification left, I think.

SConstruct
ischoegl: Good catch! I had deleted it, but apparently undid some of my changes before committing.

ischoegl:
@speth ... could you re-approve? (fixing the last item dismissed your approval)

speth (approved):

PR #1472, opened 2023-04-05
Remove unused configuration variables

ischoegl:

Remove (mostly) unused environment variables in `config.h`

- `SOLARIS` is unused
- `NEED_ALT_MAIN` is used by a Fortran example but nowhere defined; there is no evidence of this being used outside of Cantera
- `LAPACK_NAMES_LOWERCASE` has a SCons option, but is nowhere mentioned in C++ or Fortran; there is no evidence of this being used outside of Cantera
- `CT_USE_DEMANGLE` is only required for Boost <1.56. The newly required DLL support was introduced in 1.61
- `DARWIN` is only used to set an obscure data location `/Applications/Cantera/data` - it *may* be used for `cantera-macos-pkg`, but I am not sure how much use the installer sees; its main use appears to be for an installation with Matlab, but I *believe* that the toolbox is broken for M1?

**Checklist**

### DIFF
--- a/src/base/application.cpp
+++ b/src/base/application.cpp
@@ -266,11 +266,6 @@ void Application::setDefaultDirectories()
 
 #endif
 
-#ifdef DARWIN
-    // add a default data location for Mac OS X
-    inputDirs.push_back("/Applications/Cantera/data");
-#endif
-
     // CANTERA_DATA is defined in file config.h. This file is written during the
     // build process (unix), and points to the directory specified by the
     // 'prefix' option to 'configure', or else to /usr/local/cantera.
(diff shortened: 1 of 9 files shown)


src/base/application.cpp
```diff
@@ -266,11 +266,6 @@ void Application::setDefaultDirectories()
 
 #endif
 
-#ifdef DARWIN
-    // add a default data location for Mac OS X
-    inputDirs.push_back("/Applications/Cantera/data");
-#endif
-
```
speth, line 273: I'm okay with eliminating this as a standard search location, but in that case there are a few docstrings that need to be updated, too (in `global.h` and `application.h`).

speth (changes requested):
Thanks for the cleanup work on this, @ischoegl. Looks good to me besides the one comment below.

ischoegl:
Thanks! I removed the references as suggested. *PS: ended up also removing references to Win XP and Win 2000 ...*

speth (approved):

PR #1478, opened 2023-04-17
Extensible rate: unit handling, serialization, and validation

speth:

- Expand handling of `AnyMap` units in Python
- Implement `ExtensibleRate.get_parameters` to enable serialization of user-defined rate types
- Implement `ExtensibleRate.validate` to enable validation checks for user-defined types
- Fix setting of `Reaction.rate_units`
- Fix incorrect rate coefficient unit calculations for falloff / chemically-activated reactions
- Fix confusion where rate coefficient units were being conflated with units of the parameter appearing in the rate expression (specifically, this applies to sticking coefficients)
- Add tests for rate coefficient units

Contributes to implementing https://github.com/Cantera/enhancements/issues/79

From the `custom_reactions.py` example:
```py
@ct.extension(name="extensible-Arrhenius", data=ExtensibleArrheniusData)
class ExtensibleArrhenius(ct.ExtensibleRate):
    __slots__ = ("A", "b", "Ea_R")
    def set_parameters(self, params, units):
        self.A = params.convert_rate_coeff("A", units)
        self.b = params["b"]
        self.Ea_R = params.convert_activation_energy("Ea", "K")

    def get_parameters(self, params):
        params.set_quantity("A", self.A, self.conversion_units)
        params["b"] = self.b
        params.set_activation_energy("Ea", self.Ea_R, "K")

    def validate(self, equation, soln):
        if self.A < 0:
            raise ValueError(f"Found negative 'A' for reaction {equation}")

    def eval(self, data):
        return self.A * data.T**self.b * exp(-self.Ea_R/data.T)
```

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/_utils.pyx
+++ b/interfaces/cython/cantera/_utils.pyx
@@ -184,6 +184,14 @@ cdef class AnyMap(dict):
         """
         return self.unitsystem.convert_activation_energy_to(self[key], dest)
 
+    def convert_rate_coeff(self, str key, dest):
+        """
+        Convert the value corresponding to the specified *key* to the units defined by
+        *dest*, with special handling for `UnitStack` input and potentially-undefined
+        rate coefficient units.
+        """
+        return self.unitsystem.convert_rate_coeff_to(self[key], dest)
+
 
 cdef anyvalue_to_python(string name, CxxAnyValue& v):
     cdef CxxAnyMap a
--- a/interfaces/cython/cantera/_utils.pyx
+++ b/interfaces/cython/cantera/_utils.pyx
@@ -7,7 +7,9 @@ import warnings
 from cpython.ref cimport PyObject
 import numbers
 import importlib.metadata
+from collections import namedtuple
 import numpy as np
+from .units cimport Units
 
 
 _scipy_sparse = None
@@ -151,6 +153,10 @@ cdef comp_map_to_dict(Composition m):
 class CanteraError(RuntimeError):
     pass
 
+_DimensionalValue = namedtuple('_DimensionalValue',
+                              ('value', 'units', 'activation_energy'),
+                              defaults=[False])
+
 cdef public PyObject* pyCanteraError = <PyObject*>CanteraError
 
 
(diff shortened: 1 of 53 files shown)


interfaces/cython/cantera/_utils.pyx
```diff
@@ -151,8 +153,75 @@ cdef comp_map_to_dict(Composition m):
 class CanteraError(RuntimeError):
     pass
 
+DimensionalValue = namedtuple('DimensionalValue',
```
ischoegl: Looks like there is starting to be overlap with what `pint` can do. I am personally on the fence about where this can/should go, so this is mostly an observation at this point.

interfaces/cython/cantera/_utils.pyx
```diff
@@ -151,8 +153,75 @@ cdef comp_map_to_dict(Composition m):
...
+                              ('value', 'units', 'activation_energy'),
+                              defaults=[False])
+
 cdef public PyObject* pyCanteraError = <PyObject*>CanteraError
 
+
+cdef class AnyMap(dict):
+    """
+    A key-value store representing objects defined in Cantera's YAML input format.
```
ischoegl, line 165: See above. We potentially have the option to use `pint` here, instead of grafting it on at a later point?

ischoegl (commented):
@speth … changes look good to me; while I did spend a fair amount of time figuring out units when I refactored the rate calculators, your simplifications make sense to me. Good to see the fixes also: for falloff reactions, I was mainly concerned about getting the final units right and making sure that everything matched the CTI implementation; regarding sticking coefficients, they weren’t meant to have units, but I may have adopted/duplicated infrastructure that brought them in the back door.

My main observation in this PR is about the broadening units interface in Python. It is apparent that extensible objects make exposing C++ units a requirement (some bare bones units functionality I had put in place before was merely a narrow interface that allowed for testing). At this point some potential overlaps with `pint` capabilities start to appear. While the motivations for these unit system interfaces are certainly different, I am wondering about where this would lead eventually? I don’t have a satisfactory answer to this myself at the moment …

*PS: from a technical viewpoint, it makes little to no sense to run things through `pint` for what is proposed here. So my main question is about how the unit handling approaches are differentiated from the user-facing perspective.*

ischoegl (dismissed):
Apart from the questions I had in my comment, the changes themselves look good to me.

speth:
You're right that there is some significant overlap in what pint can be configured to do and what the C++ `UnitSystem` class does, but I think there's something to be said for the consistency of always using our internal methods rather than introducing the possibility that we would see differing behavior between Python and C++. I don't anticipate the need for much further expansion of the Python unit handling capability -- this pretty much mirrors what we have in C++, and that has been sufficient for all of our needs on that side.

I think that even if we eventually expand the use of units as in #991 and that becomes a popular way to use Cantera, I'd still expect to retain a lower-level interface that worked in plain mks+kmol values, and you'd probably always want to use that interface for things like `ExtensibleRate` where having units around would just add a further speed penalty.

I updated the PR to (a) hide the `DimensonalValue` namedtuple, as this is just an implementation detail (the corresponding C++ `Quantity` struct is likewise internal) and (b) provide a little more guidance on how the `UnitSystem` class is meant to be used.

ischoegl (approved):

PR #1483, opened 2023-04-20
Merge GasKinetics into BulkKinetics and simplify rate updates

speth:

With all specializations for different reaction rate parameterizations now handled by `ReactionRate` classes, there is very little left in the `GasKinetics` class that doesn't apply to kinetics in any bulk phase. From this point, I think there should be no need to specialize `BulkKinetics` for the purpose of representing different thermodynamic or kinetic models -- these should be handled by the specializations of `ThermoPhase` and `ReactionRate`.

- Move methods of `GasKinetics` to `BulkKinetics` and deprecate class `GasKinetics`
- Consolidate the separate (and incorrectly named) methods used for updating rate coefficients and rates of progress into `updateROP`
- Deprecate creating `InterfaceKinetics` objects with the reacting phase not as the first phase
- Fix some Doxygen warnings
- Delete some unused member variables

**Checklist**


include/cantera/kinetics/KineticsFactory.h
```diff
@@ -82,10 +82,12 @@ unique_ptr<Kinetics> newKinetics(const std::vector<ThermoPhase*>& phases,
  *     the reactions occur. Searches the Cantera data for this file.
  * @param phase_name  The name of the reacting phase in the input file (that is, the
  *     name of the first phase in the `phases` vector)
+ * @deprecated The 'phase_name' argument is deprecated and will be removed after
+ *     Cantera 3.1.
```
ischoegl: Could you add a comment about the new behavior?

src/kinetics/KineticsFactory.cpp
```diff
@@ -6,7 +6,11 @@
 // at https://cantera.org/license.txt for license and copyright information.
 
 #include "cantera/kinetics/KineticsFactory.h"
-#include "cantera/kinetics/GasKinetics.h"
+#include "cantera/kinetics/BulkKinetics.h"
+
+#define CT_SKIP_DEPRECATION_WARNINGS
```
ischoegl, line 11: Is there a reason to still include `GasKinetics.h`?

src/kinetics/KineticsFactory.cpp
```diff
@@ -137,8 +143,14 @@ shared_ptr<Kinetics> newKinetics(const vector<shared_ptr<ThermoPhase>>& phases,
                                  const string& filename,
                                  const string& phase_name)
 {
+    if (phase_name != "" && phase_name != phases.at(0)->name()) {
+        warn_deprecated("newKinetics", "The reacting phase must be the first phase in "
+            "the 'phases' vector. Remove the 'phase_name' argument to use the first "
+            " phase by default. This warning will be an error after Cantera 3.0. ");
```
ischoegl: Isn’t the parameter removed after Cantera 3.0?

ischoegl (changes requested):
@speth … thanks for implementing these simplifications (I recall you hinting at wanting to do this for a while). I have only minor comments.

src/kinetics/KineticsFactory.cpp
speth: The `phase_name` parameter is gone, but having the reacting phase not in first position will raise an exception from `Kinetics::addThermo`.

src/kinetics/KineticsFactory.cpp
speth: The "gentle" deprecation approach that I've tried to implement here leaves setting `kinetics: gas` in an input file creating a `GasKinetics` object in Cantera 3.0, then switching that to `BulkKinetics` in Cantera 3.1. The reasons for this are:
1. I don't want to make users edit their input files -- `gas` can remain a synonym for `bulk` indefinitely
2. If a user explicitly asks for a `gas` object, in Cantera 3.0, they should still get one -- they could (at least in theory) `dynamic_cast` that pointer to `GasKinetics`, and that should only raise a deprecation warning for now.

The way I've achieved this (I think -- please feel free to check this logic) is to have the `gas` model continue to create a `GasKinetics` object without any warning. However, if you `#include "GasKinetics.h`, which is a prerequisite for directly creating or casting to `GasKinetics`, you will get a compiler warning from that input file via the `#pragma message` and runtime warnings from the methods that are only defined for the `GasKinetics` class.

ischoegl: Ok, this makes sense. My own, less "gentle", intuition would have been to switch this over directly, as we are bumping the major version. I don't have a problem with the gentle approach though.

src/kinetics/KineticsFactory.cpp
ischoegl: I guess I was puzzled by "this warning will be an error", as the clause would have to go somewhere else?

speth: OK, I think I was making this part too complicated -- the `phase_name` parameter can be deprecated immediately. Specifying the first phase's name to `newKinetics` can issue a warning, and having the reacting phase in a different order will issue a warning from `Kinetics::addThermo`.

ischoegl (approved):
Thanks - looks good to me!

PR #1319, opened 2022-06-06
Non homogenous flow reactor

gkogekar:
This is an updated version of the old PR - Non homogenous flow reactor rewrite  (https://github.com/Cantera/cantera/pull/575) created by @arghdos.  The current version is up-to-date with the latest Cantera build.

PR #575, opened 2018-11-17
Non homogenous flow reactor rewrite

skyreflectedinmirrors:
TODO:

- consolidate IDA / CVODEs into a single abstract base class
- make IDA solver use a FuncEval (rather than the un-used ResidualJacFunc)
- Implement a virtual eval w/ residuals in FuncEval

PR #1490, opened 2023-05-12
FlowReactor with heterogeneous chemistry

speth:
This is a new attempt to integrate the work originally done in #575 and updated in #1319, which enables solving the DAE system that arises when modeling a plug flow reactor with surface chemistry.

- Substantial rewrite of the `DAE_Solver` class (now `IdasIntegrator`)
- Implementation of heterogeneous PFR equations in the `FlowReactor` class
- New Python example based on the Jupyter notebook [`1D_pfr_surfchem`](https://github.com/Cantera/cantera-jupyter/blob/main/reactors/1D_pfr_surfchem.ipynb)
- Update `surf_pfr.py` to use the `FlowReactor` class instead of a chain of `IdealGasReactor`s integrated to steady-state (the previous version is still available as `surf_pfr_chain.py`

See the two new examples referenced above.

**Checklist**


data/SiF4_NH3_mec.yaml
```diff
@@ -0,0 +1,523 @@
+description: |-
+  nits(length='cm', time='s', quantity='mol', act_energy='cal/mol')
```
ischoegl: Minor issue: description needs an update (while I am aware that this PR is still in 'draft' mode, this is something that may get overlooked). I'll reserve other comments for later.

SConstruct
```diff
@@ -1474,7 +1474,8 @@ env['NEED_LIBM'] = not conf.CheckLibWithHeader(None, 'math.h', 'C',
 env['LIBM'] = ['m'] if env['NEED_LIBM'] else []
 
 if env['system_sundials'] == 'y':
-    for subdir in ('sundials', 'nvector', 'cvodes', 'ida', 'sunlinsol', 'sunmatrix'):
+    for subdir in ('sundials', 'nvector', 'cvodes', 'idas', 'sunlinsol', 'sunmatrix',
```
bryanwweber, line 1477: Can you make sure the same change is made to the `python_sdist` SConstruct? https://github.com/Cantera/cantera/blob/0d449cabaef0a5a2317d81b725e87a47c7c96505/interfaces/python_sdist/SConscript#L58

Thanks 😄

data/SiF4_NH3_mec.yaml
speth: Don't worry, I wasn't going to leave this mechanism without a description.

include/cantera/numerics/Integrator.h
```diff
@@ -288,23 +288,34 @@ class Integrator
         return stats;
     }
 
+    virtual void setMaxNonlinIterations(int n) {
+        warn("setMaxNonlinIterations");
```
ischoegl, line 301: What is this warning about? ... Same for similar warnings below.

interfaces/cython/cantera/reactor.pyx
```diff
@@ -446,20 +446,58 @@ cdef class FlowReactor(Reactor):
     reactor_type = "FlowReactor"
 
     property mass_flow_rate:
-        """ Mass flow rate per unit area [kg/m^2*s] """
+        """ Mass flow rate [kg/s] """
         def __set__(self, double value):
             (<CxxFlowReactor*>self.reactor).setMassFlowRate(value)
 
+    property area:
```
ischoegl: As far as I am aware of, Cython now recommends standard Python decorators
```
@property
def area(self):
    [...]

@area.setter
def area(self, value):
    [...]
```
(the previous syntax is deprecated, see [here](https://cython.readthedocs.io/en/stable/src/userguide/extension_types.html#properties), even in 0.29). Should we use the official style for new code?

src/zeroD/ReactorNet.cpp
```diff
@@ -9,24 +9,25 @@
...
 namespace Cantera
 {
 
 ReactorNet::ReactorNet() :
-    m_integ(newIntegrator("CVODE"))
+    m_integ(),
+    m_time(0.0), m_init(false), m_integrator_init(false),
+    m_nv(0), m_rtol(1.0e-9), m_rtolsens(1.0e-4),
+    m_atols(1.0e-15), m_atolsens(1.0e-6),
+    m_verbose(false)
```
ischoegl: Should default values be added where the class variables are defined?

samples/python/reactors/surf_pfr.py
```diff
@@ -30,91 +32,46 @@
...
-sim.atol = 1.0e-21
 
 output_data = []
+n = 0
+print('    distance       X_CH4        X_H2        X_CO')
+print('  {:10f}  {:10f}  {:10f}  {:10f}'.format(
+      0, *r.thermo['CH4', 'H2', 'CO'].X))
+
+while sim.time < length:
+    dist = sim.time * 1e3
```
ischoegl: I know that distance is proportional to time, but it's still a little confusing to write it like this.

include/cantera/zeroD/FlowReactor.h
```diff
@@ -6,55 +6,131 @@
...
+    void setSurfaceAreaToVolumeRatio(double sa_to_vol) {
+        m_sa_to_vol = sa_to_vol;
+    }
+
+    //! Set the steady state tolerances used to determine the initial state for
+    //! surface coverages
+    void setInletSurfaceAtol(double atol) {
+        m_ss_atol = atol;
     }
-    double distance() const {
```
ischoegl, line 39: Would you reconsider the removal of `distance`? In some contexts (examples!), using time as a placeholder for distance is somewhat confusing, and also not universally applicable.

interfaces/cython/cantera/reactor.pyx
```diff
@@ -446,20 +446,58 @@ cdef class FlowReactor(Reactor):
...
+        def __set__(self, sa_to_vol):
+            (<CxxFlowReactor*>self.reactor).setSurfaceAreaToVolumeRatio(sa_to_vol)
+
     property speed:
         """ Speed [m/s] of the flow in the reactor at the current position """
         def __get__(self):
             return (<CxxFlowReactor*>self.reactor).speed()
 
-    property distance:
-        """ The distance of the fluid element from the inlet of the reactor."""
```
ischoegl, line 459: See other comment.

ischoegl (changes requested):
@speth ... thank you for tackling this PR, which concludes a major effort started some time ago. I mainly have a couple of minor comments, plus the concern that removing `distance` may be confusing for some.

include/cantera/numerics/Integrator.h
speth: This is consistent with how other methods that aren't implemented in the `Integrator` base class are handled -- the `warn` method is a private member of `Integrator` used only for this purpose. It might make more sense for all of these to throw `NotImplementedError`, but I don't know that this PR is the time to take care of that cleanup.

interfaces/cython/cantera/reactor.pyx
speth: I updated the newly-added properties, which also required adding getters for several properties that only had setters implemented. I will say that this bit of refactoring is very tedious, and I'm not looking forward to doing it more broadly through the Cython module. I'm hoping the removal of this syntax from the Cython language is a long way off.

samples/python/reactors/surf_pfr.py
speth: I made some changes so that `ReactorNet` now has both `time` and `distance` properties, where the one that is not applicable (depending on the reactor types involved) raises an exception. I updated some of the documentation as well to reflect the fact that the integrator's independent variable could be either time or distance, but stopped short of creating alternative names for `set_initial_time()` and the `max_time_step` property.

include/cantera/zeroD/FlowReactor.h
speth: I brought this back, but in a deprecated form -- I think the value of the independent variable should be accessed via the `ReactorNet` object, the same as for other `Reactor` types.

ischoegl (approved):
Thanks, @speth!

PR #1491, opened 2023-05-17
Add speed of sound property for ThermoPhase

corykinney:

- Add `ThermoPhase::soundVelocity` virtual function that returns the equilibrium speed of sound in m/s.
- Add finite difference consistency test `c_eq_sqrt_dP_drho_const_s`
- Implement definitions for `IdealGasPhase`, `RedlichKwongMFTP`, and `PengRobinson`

Marking this as a draft to see if this would be useful or not. I'm also not sure whether `soundVelocity` is the best name, I also considered `soundSpeed` and `equilibriumSoundSpeed`, or something else could be better. I'm also not certain if `ThermoPhase` is the best place for it, and I'm not positive I added it to the cython interface properly.

**Checklist**


bryanwweber:
I have a preference for `equilibriumSoundSpeed` and `equilibrium_sound_speed` for C++ and Python, respectively.

speth:
Thanks, @corykinney. I think that adding this property is useful, especially when accompanied with implementations for phases other than the ideal gas model, where it's a non-trivial calculation.

I agree that "speed" is preferable to "velocity", but I think calling this the "equilibrium" sound speed would be a misnomer, since the speed is not being computed while holding the mixture at chemical equilibrium. We even have an example illustrating this difference ([sound_speed.py](https://cantera.org/examples/python/thermo/sound_speed.py.html)).

corykinney:
> Thanks, @corykinney. I think that adding this property is useful, especially when accompanied with implementations for phases other than the ideal gas model, where it's a non-trivial calculation.
> 
> I agree that "speed" is preferable to "velocity", but I think calling this the "equilibrium" sound speed would be a misnomer, since the speed is not being computed while holding the mixture at chemical equilibrium. We even have an example illustrating this difference ([sound_speed.py](https://cantera.org/examples/python/thermo/sound_speed.py.html)).

Yes, frozen was the term I was looking for, not equilibrium. I can rename them to `soundSpeed` and `sound_speed` for C++ and Python, respectively.

ischoegl (commented):
@corykinney ... thanks for the PR. Before a more in-depth review, could you please look into the failing tests? Most likely, you need to (i) add `sound_speed` to the list of `_scalar` definitions within the `SolutionArray` class in `composite.py` (around line 585), and (ii) add `sound_speed` to the scalar property `_attr` list in `onedim.py` (around line 770).

corykinney:
> @corykinney ... thanks for the PR. Before a more in-depth review, could you please look into the failing tests? Most likely, you need to (i) add `sound_speed` to the list of `_scalar` definitions within the `SolutionArray` class in `composite.py` (around line 585), and (ii) add `sound_speed` to the scalar property `_attr` list in `onedim.py` (around line 770).

Looks like that fixed the failing checks!

ischoegl (changes requested):
Hi @corykinney ... thanks for fixing CI - I'm happy that this did the trick! Changes look mostly good to me.

The only other thing I'd ask you to update is the Python example [sound_speed.py](https://github.com/Cantera/cantera/blob/main/samples/python/thermo/sound_speed.py) so it uses the new built-in function for the check with `afrozen2`. Also, please make sure to update the `requires` in the header.

corykinney:
@ischoegl I updated the example as requested (assuming this change will make it into 3.0.0). Also, I saw that there was a [sound_speed_units.py](https://github.com/Cantera/cantera/blob/main/samples/python/thermo/sound_speed_units.py) example as well, so I believe I implemented the units for `sound_speed` as required, but I'd like to get a second set of eyes on it.

ischoegl (dismissed):
Thanks, @corykinney ... this looks good from my side. I'll leave it open for a couple of days in case anyone else has anything to add.

ischoegl:
@corykinney ... it looks like there now is a merge conflict due to recent upstream changes (#1501 moved consistency tests to a separate test). Could you rebase?

corykinney:
> @corykinney ... it looks like there now is a merge conflict due to recent upstream changes (#1501 moved consistency tests to a separate test). Could you rebase?

@ischoegl Just force pushed with the rebased commits

include/cantera/thermo/IdealGasPhase.h
```diff
@@ -401,6 +401,10 @@ class IdealGasPhase: public ThermoPhase
         return 1.0 / temperature();
     }
 
+    virtual double soundSpeed() const {
+        return sqrt(cp_mole() / cv_mole() * GasConstant / meanMolecularWeight() * temperature());
+    }
```
speth: A small nitpick here -- can you please format this so it fits within our line length limit of 88 characters? It might be better to move the implementation into the `.cpp` file as well.

speth (changes requested):
Thanks, @corykinney, and my apologies for introducing the merge conflict with this. I had just one small suggestion before this is merged.

include/cantera/thermo/IdealGasPhase.h
corykinney: @speth Should be good now. I also squashed some commits.

speth:
Thanks for the quick update, @corykinney. However, I think your commits are now out of order, such that building from some of the intermediate ones will fail. I think it should be:
1. Add virtual method to `ThermoPhase` and implementation for `IdealGasPhase`
2. Add the implementations for P-R and R-K
3. Add the consistency test
4. Update the example

corykinney:
> Thanks for the quick update, @corykinney. However, I think your commits are now out of order, such that building from some of the intermediate ones will fail. I think it should be:
> 
>     1. Add virtual method to `ThermoPhase` and implementation for `IdealGasPhase`
> 
>     2. Add the implementations for P-R and R-K
> 
>     3. Add the consistency test
> 
>     4. Update the example

@speth The changes should be grouped and ordered properly now. Thanks for your patience! Let me know if there is anything else.

speth (approved):
Thanks for taking care of these details, @corykinney!

PR #1495, opened 2023-06-05
Polynomial Ea coverage dependency

jongyoonbae:

- This PR will enable polynomial coverage dependency on activation energy for `InterfaceRate` on top of the current linear coverage dependency.
- The current methods for linear coverage dependency remain unchanged when a surface reaction's `E` in `coverage-dependencies` is given as a scalar value.
- When `E` is given as a list of four values, the 4th-order polynomial dependency is applied.
- Accordingly, source code changes are made in `InterfaceRate.cpp` and `InterfaceRate.h` as well as additional documentations in `InterfaceRate.h` and `reactions.rst` are appended.
- Lastly, an additional test routine is added to `kineticsFromYaml.cpp` with additional dummy coverage dependency parameters in `surface-phases.yaml`.

Closes #

**Checklist**

### DIFF
--- a/include/cantera/kinetics/InterfaceRate.h
+++ b/include/cantera/kinetics/InterfaceRate.h
@@ -141,7 +141,7 @@ public:
 
     //! Set association with an ordered list of all species associated with a given
     //! `Kinetics` object.
-    void setSpecies(const std::vector<std::string>& species);
+    void setSpecies(const vector<string>& species);
 
     //! Update reaction rate parameters
     //! @param shared_data  data shared by all reactions of a given type
@@ -257,16 +257,16 @@ protected:
     //! Vector holding coverage-specific activation energy dependence as a
     //! 5-membered array of polynomial coeffcients starting from 0th-order to
     //! 4th-order coefficients
-    std::vector<vector_fp> m_ec;
-    std::vector<bool> m_lindep; //!< Vector holding boolean for linear dependence
+    vector<vector_fp> m_ec;
+    vector<bool> m_lindep; //!< Vector holding boolean for linear dependence
     vector_fp m_mc; //!< Vector holding coverage-specific power-law exponents
 
 private:
     //! Pairs of species index and multipliers to calculate enthalpy change
-    std::vector<std::pair<size_t, double>> m_stoichCoeffs;
+    vector<pair<size_t, double>> m_stoichCoeffs;
 
     //! Pairs of phase index and net electric charges (same order as m_stoichCoeffs)
-    std::vector<std::pair<size_t, double>> m_netCharges;
+    vector<pair<size_t, double>> m_netCharges;
 };
 
 
@@ -301,7 +301,7 @@ public:
     }
 
     //! Get sticking species.
-    std::string stickingSpecies() const {
+    string stickingSpecies() const {
         return m_stickingSpecies;
     }
 
@@ -311,7 +311,7 @@ public:
      *  to be explicitly identified. Note that species have to be specified prior
      *  to adding a reaction to a Kinetics object.
      */
-    void setStickingSpecies(const std::string& stickingSpecies) {
+    void setStickingSpecies(const string& stickingSpecies) {
         m_stickingSpecies = stickingSpecies;
         m_explicitSpecies = true;
     }
@@ -360,7 +360,7 @@ public:
 protected:
     bool m_motzWise; //!< boolean indicating whether Motz & Wise correction is used
     bool m_explicitMotzWise; //!< Correction cannot be overriden by default
-    std::string m_stickingSpecies; //!< string identifying sticking species
+    string m_stickingSpecies; //!< string identifying sticking species
     bool m_explicitSpecies; //!< Boolean flag
     double m_surfaceOrder; //!< exponent applied to site density term
     double m_multiplier; //!< multiplicative factor in rate expression
@@ -391,7 +391,7 @@ public:
     }
 
     //! Identifier of reaction rate type
-    virtual const std::string type() const override {
+    virtual const string type() const override {
         return "interface-" + RateType::type();
     }
 
@@ -482,7 +482,7 @@ public:
     }
 
     //! Identifier of reaction rate type
-    virtual const std::string type() const override {
+    virtual const string type() const override {
         return "sticking-" + RateType::type();
     }
 
@@ -527,7 +527,7 @@ public:
         StickingCoverage::setContext(rxn, kin);
     }
 
-    virtual void validate(const std::string &equation, const Kinetics& kin) override {
+    virtual void validate(const string &equation, const Kinetics& kin) override {
         RateType::validate(equation, kin);
         fmt::memory_buffer err_reactions;
         double T[] = {200.0, 500.0, 1000.0, 2000.0, 5000.0, 10000.0};
--- a/include/cantera/kinetics/InterfaceRate.h
+++ b/include/cantera/kinetics/InterfaceRate.h
@@ -84,9 +84,20 @@ struct InterfaceData : public BlowersMaselData
  * It is evident that this expression combines a regular modified Arrhenius rate
  * expression \f$ A T^b \exp \left( - \frac{E_a}{RT} \right) \f$ with coverage-related
  * terms, where the parameters \f$ (a_k, E_k, m_k) \f$ describe the dependency on the
- * surface coverage of species \f$ k, \theta_k \f$. The InterfaceRateBase class implements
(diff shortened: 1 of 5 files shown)


include/cantera/kinetics/InterfaceRate.h
```diff
@@ -84,9 +84,20 @@ struct InterfaceData : public BlowersMaselData
...
  * expression \f$ A T^b \exp \left( - \frac{E_a}{RT} \right) \f$ with coverage-related
  * terms, where the parameters \f$ (a_k, E_k, m_k) \f$ describe the dependency on the
- * surface coverage of species \f$ k, \theta_k \f$. The InterfaceRateBase class implements
- * terms related to coverage only, which allows for combinations with arbitrary rate
- * parameterizations (for example Arrhenius and BlowersMaselRate).
+ * surface coverage of species \f$ k, \theta_k \f$. In addition to the linear coverage
+ * dependence on the activation energy modifier \f$ E_k \f$, polynomial coverage
+ * dependence is also available. When the dependence parameter \f$ E_k \f$ is given as
+ * a scalar value, the linear dependency is applied whereas if a list of four values
+ * are given as \f$ [E^{(1)}_k-E^{(4)}_k] \f$, a polynomial dependency is applied as
```
speth:
```suggestion
 * are given as \f$ [E^{(1)}_k, ..., E^{(4)}_k] \f$, a polynomial dependency is applied as
```

include/cantera/kinetics/InterfaceRate.h
```diff
@@ -116,7 +127,7 @@ class InterfaceRateBase
     //! Add a coverage dependency for species *sp*, with exponential dependence
     //! *a*, power-law exponent *m*, and activation energy dependence *e*,
     //! where *e* is in Kelvin, that is, energy divided by the molar gas constant.
-    virtual void addCoverageDependence(const string& sp, double a, double m, double e);
+    virtual void addCoverageDependence(const string& sp, double a, double m, vector_fp e);
```
speth:
```suggestion
    virtual void addCoverageDependence(const string& sp, double a, double m, const vector_fp& e);
```

include/cantera/kinetics/InterfaceRate.h
```diff
@@ -243,7 +254,11 @@ class InterfaceRateBase
     map<size_t, size_t> m_indices;
     vector<string> m_cov; //!< Vector holding names of coverage species
     vector_fp m_ac; //!< Vector holding coverage-specific exponential dependence
-    vector_fp m_ec; //!< Vector holding coverage-specific activation energy dependence
+    //! Vector holding coverage-specific activation energy dependence as a
+    //! 5-membered array of polynomial coeffcients starting from 0th-order to
+    //! 4th-order coefficients
+    std::vector<vector_fp> m_ec;
+    std::vector<bool> m_lindep; //!< Vector holding boolean for linear dependence
```
speth: We now have a namespace-wide `using std::vector` declaration, so the `std::` prefix can always be removed from `std::vector` and a number of other types listed in `ct_defs.h`.

src/kinetics/InterfaceRate.cpp
```diff
@@ -140,18 +141,38 @@ void InterfaceRateBase::setCoverageDependencies(const AnyMap& dependencies,
...
-            E = units.convertActivationEnergy(cov_map["E"], "K");
+            if (cov_map["E"].isScalar()) {
+                m_lindep.push_back(true);
+                E[1] = units.convertActivationEnergy(cov_map["E"], "K");
+            } else {
+                m_lindep.push_back(false);
+                auto& E_temp = cov_map["E"].asVector<AnyValue>(4);
+                for (size_t i = 0; i < E_temp.size(); i++) {
+                    E[i+1] = units.convertActivationEnergy(E_temp[i], "K");
+                }
```
speth: You could make this input slightly more flexible by allowing the user to omit any trailing zeros in the coefficient array. I think all that is required is to get the values by writing
```c++
auto& E_temp = cov_map["E"].asVector<AnyValue>(1, 4);
```
to allow an array of length 1-4. Applies again in the case below.

src/kinetics/InterfaceRate.cpp
```diff
@@ -226,7 +264,7 @@ void InterfaceRateBase::updateFromStruct(const InterfaceData& shared_data) {
     m_mcov = 0.0;
     for (auto& [iCov, iKin] : m_indices) {
         m_acov += m_ac[iCov] * shared_data.coverages[iKin];
-        m_ecov += m_ec[iCov] * shared_data.coverages[iKin];
+        m_ecov += poly4(shared_data.coverages[iKin], m_ec[iCov].data());
```
speth: I was kind of expecting this to be conditioned on the value of `m_lindep[iCov]`, though I'm not sure whether the extra "if" outweighs the benefit of not evaluating the trivial polynomial terms in the linear case.

test/kinetics/kineticsFromYaml.cpp
```diff
@@ -605,8 +605,23 @@ TEST(Kinetics, InterfaceKineticsFromYaml)
...
+
+    auto soln2 = newInterface("surface-phases.yaml", "Pt-multi-sites");
+    auto kin2 = soln2->kinetics();
+    auto R4 = kin2->reaction(3);
+    vector_fp coeffs{1.0e6, 3.0e6, -7.0e7, 5.0e6};
+    const auto rate4 = std::dynamic_pointer_cast<InterfaceArrheniusRate>(R4->rate());
+    Cantera::AnyMap coverage_deps2;
+    rate4->getCoverageDependencies(coverage_deps2);
+    coverage_deps2.applyUnits();
+    auto& cov_map2 = coverage_deps2["O(s)"].as<Cantera::AnyMap>();
```
speth: The `Cantera::` prefix is unnecessary here since we have a `using namespace Cantera` directive at the top of the file.

speth (changes requested):
Thanks for this contribution, @jongyoonbae. I think it generally looks good, and I just had a couple of minor comments on the implementation.

src/kinetics/InterfaceRate.cpp
jongyoonbae: I was debating with myself about that exact aspect. I chose the slightly-simpler code block (polynomial evaluation for all) for my initial commit. But reconsidering this part made me think using the `m_lindep` flag probably makes the code more consistent throughout. I will go ahead and modify accordingly.

include/cantera/kinetics/InterfaceRate.h
jongyoonbae: That's great to know. While I am at it, shall I remove `std::` for the other parts of the `InterfaceRate` where applicable? Or change only for what I added?

speth: The only necessary changes are for lines of code that you added/modified. If you want, you can update nearby lines in the same function, but please do not expand the scope beyond that, since it makes for large diffs that obscure the non-trivial code changes.

speth: Looking at the commit you made to make some of these changes: this is fine -- there aren't that many of them in this particular file, so I think it was OK to update them.

src/kinetics/InterfaceRate.cpp
```diff
@@ -166,22 +187,40 @@ void InterfaceRateBase::getCoverageDependencies(AnyMap& dependencies,
...
+            if (m_lindep[k]) {
+                dep[2] = m_ec[k][1];
+            } else {
+                vector<AnyValue> dep(3);
+                vector_fp E_temp(4);
+                for (size_t i = 0; i < m_ec[k].size() - 1; i++) {
+                    E_temp[i] = m_ec[k][i+1];
+                }
+                dep[2] = E_temp;
+            }
```
speth: I don't think this case works as you expect -- the second declaration of `dep` as a `vector<AnyValue>` shadows the first one, but this vector is then deleted at the end of this scope and the outer `vector_fp` is used without any value set for `dep[2]`. Since this option to the function is deprecated anyway, I think you can rely on the idea that a new feature doesn't need to be compatible with some other deprecated option and have this second case throw an exception (I'd suggest `NotImplementedError`).

speth (changes requested):
Thanks for the updates, @jongyoonbae. I found just one spot where I think an update is needed.

src/kinetics/InterfaceRate.cpp
jongyoonbae: Great to know, thanks! I changed it as advised.

speth (approved):
Thanks, @jongyoonbae! This looks good to me.

PR #1496, opened 2023-06-06
Matlab interface (new) test suite

ssun30:

- Add unit tests for the new Matlab interface.
- Add those tests into CI workflow. 

Partially address cantera/enhancements#177

This PR will introduce a test suite using Mathwork's testing frameworks for the new Matlab interface. It will utilize a class-based structure for running tests on each module within the interface. 

It will be much more extensive than the legacy interface's test suite, which only included tests on a limited set of methods within the ThermoPhase class only. The aim is to replicate all tests that are present in the Python interface's test suite as long as the class and method exists. 

**Checklist**


test/matlab_experimental/ctRunTests.m
```diff
@@ -0,0 +1,29 @@
...
+disp('Cantera is loaded for test');
+
+% Run all tests
+results1 = runtests('ctTestThermo')
+
+% Unload Cantera and remove temporary library file
+unloadlibrary('libcantera_shared');
+delete([rootDir, ctName]);
+
+disp('Cantera has been unloaded');
```
rwest:
```suggestion
disp('Cantera has been unloaded');

```

test/matlab_experimental/ctTestPath.m
```diff
@@ -0,0 +1,44 @@
+% ctTestPath.m
+% Set up environment for testing the Cantera Matlab interface
+% from within the Cantera source tree. Run this file from the
+% root of the Cantera source tree, for example:
+%
+%    cd ~/src/cantera
+%    run interfaces/matlab_experimental/testpath.m
```
rwest, line 7:
```suggestion
%    run interfaces/matlab_experimental/ctTestPath.m
```

test/matlab_experimental/ctRunTests.m
```diff
@@ -0,0 +1,29 @@
+clear all
+
```
rwest: Docstrings or "how to use" advice in comments here?

test/matlab_experimental/ctTestThermo.m
```diff
@@ -0,0 +1,456 @@
...
+        function testInvalidProperty(self)
+            self.verifyError(@() self.getInvalidProperty(),...
+                                'MATLAB:noSuchMethodOrField');
+            self.verifyError(@() self.setInvalidProperty(300),...
+                                'MATLAB:noPublicFieldForClass');            
+        end
+
+    end
+
+end
```
rwest:
```suggestion
end

```

.github/workflows/main.yml
```diff
@@ -19,6 +20,45 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  matlab-test:
+    name: Run MATLAB Test Script
+    runs-on: ubuntu-latest
```
rwest, line 24:
```suggestion
    runs-on: ubuntu-latest
    env:
      LD_PRELOAD: /usr/lib/x86_64-linux-gnu/libstdc++.so.6
```

rwest (commented):
Some suggestions

speth:
Thanks for sharing the work-in-progress here, @ssun30. I'd like to suggest a pause on adding new tests so we can discuss structure and functionality of this test suite and limit the scope of this PR at this time. I think the first priority should be getting the tests to actually run and indicate success/failure as the status for the corresponding GitHub Action. The previously "successful" action ([here](https://github.com/Cantera/cantera/actions/runs/5249877081/jobs/9483188920)) did not actually run any of the tests that you wrote.

I also wanted to comment on a statement made in the initial post:
> The aim is to replicate all tests that are present in the Python interface's test suite as long as the class and method exists.

I don't think this is quite the correct goal for the Matlab test suite. I would suggest that the aim should be to have unit tests covering everything that is unique to the Matlab toolbox, but not to duplicate tests that are mainly about testing underlying functionality of the C++ library, which is true of quite a bit of the Python test suite.

ssun30:
> Thanks for sharing the work-in-progress here, @ssun30. I'd like to suggest a pause on adding new tests so we can discuss structure and functionality of this test suite and limit the scope of this PR at this time. I think the first priority should be getting the tests to actually run and indicate success/failure as the status for the corresponding GitHub Action. The previously "successful" action ([here](https://github.com/Cantera/cantera/actions/runs/5249877081/jobs/9483188920)) did not actually run any of the tests that you wrote.
> 
Hi @speth , the Matlab Github Action was looking for the wrong folder for the Matlab tests in CI#4006. This was fixed in the subsequent commit and the test suite was running correctly (and failed as expected) in [CI#4007](https://github.com/Cantera/cantera/actions/runs/5250181476). In a later commit, I changed the [qualification methods](https://www.mathworks.com/help/matlab/matlab_prog/types-of-qualifications.html) from verification to assumption so the CI workflow would label tests that failed due to verification (no errors but returning wrong results) as filtered. This is only temporary to see whether the CI would run to completion, which it did as in [CI#4011](https://github.com/Cantera/cantera/actions/runs/5259339240). 
> I also wanted to comment on a statement made in the initial post:
> 
> > The aim is to replicate all tests that are present in the Python interface's test suite as long as the class and method exists.
> 
> I don't think this is quite the correct goal for the Matlab test suite. I would suggest that the aim should be to have unit tests covering everything that is unique to the Matlab toolbox, but not to duplicate tests that are mainly about testing underlying functionality of the C++ library, which is true of quite a bit of the Python test suite.

Sorry the statement I made here was not my intent. I was not trying to replicate the entire Python test suite, but rather test all the features that are present in the Matlab interface, but using the same exact parameters (for example temperature, pressure, expected values, and edge cases) for consistency.

ssun30:
@rwest and I have discussed about how the test suite would be merged. As it stands, the complete test for each class could be over 500 lines long. So to not make the PR overly big, we are leaning towards introducing tests incrementally with several PRs. And this PR will include two test classes and the main objective is to make sure the Github Actions run them properly. 

Also, as I'm making the tests, I spotted some bugs that result in verification failures (wrong values returned). Some of these are due to inconsistent basis (molar vs. mass), some due to incorrect indexing (in Matlab integer indices start from 1 instead of 0), some due to incorrect error handling on the Clib side. For now I temporarily filtered these failed tests out so they only count as incomplete so the CI could run to the end. But ideally I would like to merge those patches so all tests could pass before merging the test itself. So, as we incrementally introduce more tests, we also merge a separate PR that address the test failures. The exception would be for fatal errors that I would patch immediately upon discovery, such as #1498 . 

Would this be a viable way forward?

speth:
I think it's worth distinguishing between this PR, which establishes the structure and patterns for testing the experimental Matlab toolbox, and subsequent PRs that make the test suite more comprehensive.

For this PR, I would suggest:
- Do not add any test cases beyond what you've already implemented here.
- If there are test cases that would not pass (or crash outright), filter them and leave a short comment noting the failure. I believe the idiomatic way of doing this in Matlab's `unittest` package is by calling `self.assumeFail()` at the start of the test.

For subsequent PRs, I think each one should either:
1. Add new test cases that pass (possibly requiring some minor code changes) and/or add disabled test cases that would not currently pass (if you don't immediately have a fix or the fix itself is complex enough to warrant its own PR).
2. Fix some errors and enable previously-disabled test cases that now succeed.

I would specifically like to avoid the pattern of #1498, where we have a bug fix that is not verified by any automated testing. I think this approach is error prone and introduces needless complexity on your end managing separate branches that are logically related.

.github/workflows/main.yml
ischoegl, line 1: Is it possible to piggy-back on prior builds? E.g. similar to what is done for .NET or the Python examples? That way it could be easily expanded to multiple platforms (obviously as long as the MATLAB actions exist)

ssun30: I believe that's possible.

speth: The Matlab action is currently only available for Linux. Given some other discussions that @ischoegl and I have had, I don't think it would be worth the effort of changing the build configuration to make this job depend on the `cantera_shared.so` built by one of the other jobs.

.github/workflows/main.yml
```diff
@@ -19,6 +20,54 @@ concurrency:
...
+        # h5py is optional; some versions don't have binaries (yet)
+        run: |
+          python3 -m pip install ruamel.yaml scons==3.1.2 numpy cython pandas pytest \
+          pytest-github-actions-annotate-failures pint
+          python3 -m pip install h5py
+      - name: Build Cantera
+        run: |
+          python3 `which scons` build env_vars=all -j2 debug=n --debug=time \
+          hdf_libdir=$HDF5_LIBDIR hdf_include=$HDF5_INCLUDEDIR \
+          cc_flags=-D_GLIBCXX_ASSERTIONS
```
speth: I think you can save a bit of build and setup time by building with the SCons option `python_package=n` and skipping installation of all the Python-related stuff except for SCons itself.

.github/workflows/main.yml
```diff
@@ -19,6 +20,54 @@ concurrency:
...
+      - name: Set up MATLAB
+        uses: matlab-actions/setup-matlab@v1
+      - name: Print environment variables
+        uses: matlab-actions/run-command@v1
+        with:
+          command: disp(getenv('CANTERA_ROOT')); disp(getenv('CANTERA_DATA')); disp(getenv('CANTERA_TEST'));
```
speth: I don't think `CANTERA_TEST` is ever defined or used (besides this).

.github/workflows/main.yml
```diff
@@ -19,6 +20,54 @@ concurrency:
...
+      - name: Run tests
+        uses: matlab-actions/run-tests@v1
+        with:
+          select-by-folder: /home/runner/work/cantera/cantera/test/matlab_experimental
+
```
speth, line 60: The documentation for the `run-tests` action says this should be a path "relative to the project root folder", that is just `test/matlab_experimental`. Did you try that and have problems? I'm wondering if this is behind the warnings in the logs related to files that are outside this path and shouldn't be examined at all as part of this test suite, for example:
```
[Warning: "/home/runner/work/cantera/cantera/test/matlab/TestImport.m" was
  excluded.
  Caused by:
      The specified superclass 'TestCase' contains a parse error, cannot be found
      on MATLAB's search path, or is shadowed by another file with the same name.]
```
and
```
[Warning:
  "/home/runner/work/cantera/cantera/ext/HighFive/deps/catch2/projects/XCode/OCTest/OCTest/TestObj.m"
  was excluded.
  Caused by:
      File:
      /home/runner/work/cantera/cantera/ext/HighFive/deps/catch2/projects/XCode/OCTest/OCTest/TestObj.m
      Line: 1 Column: 1
      Invalid use of operator.] 
  [> In matlab.unittest.internal/InvalidTestFactory/createSuiteImplicitly (line 49)
```

.github/workflows/main.yml
```diff
@@ -9,6 +9,7 @@ on:
     branches:
       - main
       - testing
+      - matlab_tests
```
speth: Please remove this -- the CI is automatically on PRs, which is the main way we handle doing this for code in development.

test/matlab_experimental/ctTestKinetics.m
```diff
@@ -0,0 +1,162 @@
...
+    methods (Test)
+
+        function testCounts(self)
+            self.assumeEqual(self.phase.nReactions, 29);
+            self.assumeEqual(self.phase.nTotalSpecies, 10);
+            self.assumeEqual(self.phase.nPhases, 1);
+
+            % Missing method
+            % self.assumeEqual(self.phase.reactionPhaseIndex, 0);
+        end
```
speth: This test passes, right? I think in general, you should be using `verifyXyz` rather than `assumeXyz`. The `assume` prefix (according to [the documentation](https://www.mathworks.com/help/matlab/matlab_prog/types-of-qualifications.html) is meant for filtering tests where the environment does not meet some condition, for example, a test that should only run on certain operating systems, and you do not want the test to be considered as failing if the `assumeXyz` statement is not satisfied.

test/matlab_experimental/ctTestKinetics.m
```diff
@@ -0,0 +1,162 @@
...
+
+        function testIsReversible(self)
+            % Fails because of incorrect indexing. 
+            self.assumeFail();
+
+            for i = 1:self.phase.nReactions
+                self.assumeTrue(self.phase.isReversible(i));
+            end
```
speth: Please add at least a to-do comment to try this on a mechanism where not all reactions are reversible.

test/matlab_experimental/ctTestKinetics.m
```diff
@@ -0,0 +1,162 @@
...
+
+        function testStoichCoeffs(self)
+            % Fails because StoichCoeffs methods does not convert species name to 
+            % species index. 
+            self.assumeFail();
+
+            nu_r = self.phase.reactantStoichCoeffs;
+            nu_p = self.phase.productStoichCoeffs;
+
+            function checkReactnat(s, i, val)
```
speth:
```suggestion
            function checkReactant(s, i, val)
```

test/matlab_experimental/ctTestPath.m
```diff
@@ -0,0 +1,44 @@
...
+
+% get the list of directories on the Matlab path
+dirs = regexp(path, ['([^' pathsep ']*)'], 'match');
```
speth: What is i
```suggestion
dirs = split(path, pathsep);
```

test/matlab_experimental/ctTestPath.m
```diff
@@ -0,0 +1,44 @@
...
+
+% if 'cantera' is already in the path, remove it
+for i = 1:length(dirs)
+    if strfind(dirs{i}, 'Cantera')
+        rmpath(dirs{i});
+        continue;
+    end
+    if strfind(dirs{i}, 'cantera')
+        rmpath(dirs{i});
+    end
```
speth: You can simplify this by using `contains` with the `'IgnoreCase', true` arguments.

test/matlab_experimental/ctTestPath.m
```diff
@@ -0,0 +1,44 @@
...
+         fullfile(cantera_root, 'test', 'matlab_experimental'));
+
+% Add the Cantera toolbox to the Matlab path
+addpath(genpath([cantera_root, '/interfaces/matlab_experimental']));
+addpath(genpath([cantera_root, '/test/matlab_experimental']));
+
+% Set path to Python module
+if strcmp(getenv('PYTHONPATH'), '')
+    setenv('PYTHONPATH', fullfile(cantera_root, 'build', 'python'))
+end
```
speth: What is Python needed for here?

test/matlab_experimental/ctTestSetUp.m
```diff
@@ -0,0 +1,20 @@
...
+% Load Cantera
+if ~libisloaded('libcantera_shared')
+    [~, warnings] = loadlibrary([rootDir, ctName], ...
+                                [rootDir, '/include/cantera/clib/ctmatlab.h'], ...
+                                'includepath', [rootDir, '/include'], ...
+                                'addheader', 'ct', 'addheader', 'ctfunc', ...
+                                'addheader', 'ctmultiphase', 'addheader', ...
+                                'ctonedim', 'addheader', 'ctreactor', ...
+                                'addheader', 'ctrpath', 'addheader', 'ctsurf');
+end
```
speth: Can't you use the existing `ctLoad` function here?

test/matlab_experimental/ctTestTearDown.m
```diff
@@ -0,0 +1,10 @@
+clear all
+
+cantera_root = getenv('CANTERA_ROOT');
+ctName = '/test/matlab_experimental/libcantera_shared.so';
+
+% Unload Cantera and remove temporary library file
+unloadlibrary('libcantera_shared');
+delete([cantera_root, ctName]);
```
speth: Do you need to worry about deleting the library here? If so, you should use the OS-specific name as determined in `ctTestPath.m`.

test/matlab_experimental/ctTestThermo.m
```diff
@@ -0,0 +1,482 @@
...
+
+    methods
+
+        % Generic function to set invalid values to attribute and assume errors
+        function setInvalidValue(self, attr, val, errMessage)
+            try 
+                self.phase.(attr) = val;
+            catch ME
+                self.assumeSubstring(ME.message, errMessage);
+            end
```
speth: You need a `verifyFail` statement in the `try` block, after the setter that should fail. Otherwise, this test will pass if the command unexpectedly succeeds.

test/matlab_experimental/ctTestThermo.m
```diff
@@ -0,0 +1,482 @@
...
+            self.setInvalidValue('UVX', {-4e5, 4.4, 'H2:1.0', -1}, ...
+                                     'incorrect number');           
+        end
+
+        function testInvalidProperty(self)
+            self.assumeError(@() self.getInvalidProperty(),...
+                                'MATLAB:noSuchMethodOrField');
+            self.assumeError(@() self.setInvalidProperty(300),...
+                                'MATLAB:noPublicFieldForClass');            
+        end
```
speth: Can you define `getInvalidProperty` and `setInvalidProperty` as local functions here, and then just call
```matlab
self.assumeError(@getInvalidProperty(),...
                    'MATLAB:noSuchMethodOrField');
```

test/matlab_experimental/ctTestThermo.m
```diff
@@ -0,0 +1,482 @@
...
+            self.setInvalidValue('X', 'H2:1.0,CO2:1.5', 'Unknown species');
+            self.setInvalidValue('X', 'H2:1.0,O2:asdf', 'Trouble processing');
+            self.setInvalidValue('X', 'H2:1.e-x4', 'Trouble processing');
+            self.setInvalidValue('X', 'H2:1e-1.4', 'decimal point in exponent');
+            self.setInvalidValue('X', 'H2:0.5,O2:1.0,H2:0.1', 'Duplicate key');
+            self.setInvalidValue('X', '', 'cannot be empty');
+        end
+
+        function testSetStateMole(self)
+            self.assumeFail();
```
speth: Please provide a brief explanation of why this test is currently failing.

test/matlab_experimental/ctTestSetUp.m
```diff
@@ -0,0 +1,20 @@
+clear all
+
+% Copy library to test folder
+ctTestPath;
+
+% Load Cantera
+rootDir = getenv('CANTERA_ROOT');
+ctName = '/test/matlab_experimental/libcantera_shared.so';
```
ischoegl: This appears specific to a Linux environment. Is it possible to implement this in a way that this can be run from any platform? (same in `ctTestTearDown.m`). Also, `/test/matlab_experimental` is installation-specific?

You do have some switches implemented in `ctTestPath`, which would also apply here:

```Matlab
if ispc
    [...]
elseif ismac
    [...]
elseif isunix
    [...]
end
```

test/matlab_experimental/ctTestThermo.m
```diff
@@ -0,0 +1,482 @@
...
+        % Test methods
+
+        function testBaseAttributes(self)
+            self.assumeInstanceOf(self.phase.solnName, ...
+                                      'char');
+            
+            % Missing method
+            % self.assumeInstanceOf(self.phase.phaseName, ...
+            %                           'char');
+            %
```
speth: Rather than commented-out code (which can go unnoticed for a very long time), please use the `self.assumeFail()` method, which will always leave evidence behind in the test logs.

test/matlab_experimental/ctTestThermo.m
```diff
@@ -0,0 +1,482 @@
...
+                
+                idx = chargePhase.speciesIndex(species);
+                self.assumeEqual(charges(idx), charge);
+            end
+            clear chargePhase
+        end
+
+        function testReport(self)
+            % Fails because report method is missing. 
+            self.assumeFail();
```
speth: Rather than providing the reason for the failure as a comment, it appears that you can also pass it as the argument to `assumeFail`, which will make it appear in the test log. I would recommend this, since it is analogous to what we do in the other test suites.
```suggestion
            self.assumeFail('Fails because report method is missing.');
```

test/matlab_experimental/ctTestThermo.m
```diff
@@ -0,0 +1,482 @@
...
+        end
+
+        function testSingleGetters(self)
+            self.assumeFail();
+
+            val = self.phase.T;
+            exp = 300;
+            self.assumeEqual(val, exp, 'RelTol', self.rtol);
+            self.assumeGreaterThan(self.phase.maxTemp, 0);
+            self.assumeGreaterThan(self.phase.minTemp, 0);
```
speth: These seem less useful than the tests of `maxTemp` and `minTemp` above in `testRefInfo`.

test/matlab_experimental/ctTestThermo.m
```diff
@@ -0,0 +1,482 @@
...
+            self.assumeSubstring(str, 'temperature');
+
+            for i = 1:self.phase.nSpecies
+                name = self.phase.speciesName(i);
+                self.assumeSubstring(str, name{:});
+            end
+        end
+
+        function testRefInfo(self)
+            self.assumeFail();
```
speth: Please document why this test fails.

test/matlab_experimental/ctTestThermo.m
```diff
@@ -0,0 +1,482 @@
...
+        end
+
+        function testSetCompositionStringBad(self)
+            self.setInvalidValue('X', 'H2:1.0,CO2:1.5', 'Unknown species');
+            self.setInvalidValue('X', 'H2:1.0,O2:asdf', 'Trouble processing');
+            self.setInvalidValue('X', 'H2:1.e-x4', 'Trouble processing');
+            self.setInvalidValue('X', 'H2:1e-1.4', 'decimal point in exponent');
+            self.setInvalidValue('X', 'H2:0.5,O2:1.0,H2:0.1', 'Duplicate key');
+            self.setInvalidValue('X', '', 'cannot be empty');
+        end
```
speth: This is an example of what I meant that for the Matlab toolbox, you do not need to test all the behavior of the underlying library. Here, I think all you need to to is test the different places an error can come from once. That is, the "cannot be empty" error, which is raised from the Matlab setter, and just one of the others, all of which are raised by the C++ `setMassFractionsByName` or `parseCompString` methods. The rest of these cases are adequately handled by the Python and C++ test suites.

speth (changes requested):
Thanks for working on this, @ssun30, and keeping this PR to a manageable size. I think most of the comments here should be relatively easy to address.

Besides the specific comments below, I have one general recommendation for all of the test methods: I think most if not all of your test checks should be using the `verifyXyz` form rather than the `assumeXyz` form, with the exception of using `assumeFail` to indicate known failures.

test/matlab_experimental/ctTestSetUp.m
ssun30: `/test/matlab_experimental` is not installation-specific. `ctTestSetUp` always temporarily copies the Cantera shared library from the build directory into the test directory and deletes it after `ctTestTearDown`. This is how the legacy Matlab test suite sets up the test. I don't think this intermediate step is necessary.

test/matlab_experimental/ctTestSetUp.m
ssun30: I did not use `ctLoad` since it reads the value of `ctRoot` which (for now) is set by the user. For the Github actions, the test set up uses environment variable `CANTERA_ROOT` instead.

speth: I think there should be a way to avoid needing to do this, but we can certainly work on that as a separate improvement.

speth (approved):
Thanks for taking care of the suggested changes, @ssun30. I think this is a solid foundation for continuing to build out the new Matlab interface and test suite.

test/matlab_experimental/ctTestSetUp.m
ischoegl: ~But isn't it true that `libcantera_shared.so` refers to a Linux platform? Presumably, some users may want to run the test suite on Windows?~

At the same time, I see that the code is updated. So I'm ok with this as the review has been addressed. Thanks!

Issue #47, opened 2015-03-26
Error compiling Cantera Demos


GoogleCodeExporter:
```
When i compile Cantera 1.8 demos with Microsoft Visual Studio Professional 2010 
the software return me a following error:

LINK : fatal error LNK1104: cannot open file 'NVEC_SER_d.lib'

I have searched this .lib in my Pc and online but i cannot find this library. 
I hope that somebody could help me.
Best Regards
Luigi

N.B. I have windows 7 64 bit. I have the same error with Cantera 32bit and 
Python 2.6 32bit and with Cantera 64bit and Python 2.6 64bit 
```

Original issue reported on code.google.com by `comunica...@gmail.com` on 12 May 2011 at 9:38

PR #1497, opened 2023-06-07
Implementation of PhotolysisRate 

anthony-walker:
This commit implements PhotolysisRate which is a type of atmospheric process where light exposure causes the reaction. Other slight changes to files are included such as a zenith angle to the ThermoPhase object which is the angle of light exposure to calculate rates from the applied parameterization. The parameterization implemented is from the Master Chemical Mechanism Protocol.

Saunders, S. M., Jenkin, M. E., Derwent, R. G., & Pilling, M. J. (2003). Protocol for the development of the Master Chemical Mechanism, MCM v3 (Part A): tropospheric degradation of non-aromatic volatile organic compounds. Atmospheric Chemistry and Physics, 3(1), 161-180.

- Implementation of PhotolysisRate to model the atmospheric process.
- Implementation of associated tests.
- Python interface for all implementations
- Minor implementations to thermo phase to facilitate Photolysis rate calculation

The implementation can be used simply for different photolysis rate calculations or other examples such as modeling photolysis reactions in a gas phase. For example

```
gas = ct.Solution("ozone-photolysis.yaml", "atmos")
gas.zenith_angle = 0.4
```

Presently changing the zenith angle does not force the rates to be updated. I thought perhaps it should but I think the zenith angle should live in the phase as my next step will be implementation of extended phases for heterogeneous chemistry which will require the zenith angle. I am open to suggestions in this regard.

Atmospheric chemistry is an interest in [#47](https://github.com/Cantera/enhancements/issues/47)

**Checklist**

### DIFF
--- a/data/ozone-photolysis.yaml
+++ b/data/ozone-photolysis.yaml
@@ -1,0 +1,230 @@
+description: |-
+  A simple photolysis mechanism for the breakdown of ozone and formation of NOx. This
+  model was constructed manually from a subset of MCM. Saunders, S. M., Jenkin, M. E.,
+  Derwent, R. G., & Pilling, M. J. (2003). Protocol for the development of the Master
+  Chemical Mechanism, MCM v3 (Part A): tropospheric degradation of non-aromatic
+  volatile organic compounds. Atmospheric Chemistry and Physics, 3(1), 161-180.
+cantera-version: 3.0
+
+units: {length: cm, time: s, quantity: molec}
+
+phases:
+- name: atmos
+  thermo: ideal-gas
+  elements: [O, N, H]
+  species: [O, O2, O1D, O3, NO, NO2, NO3, OH, H2O2, HNO3, HONO]
+  kinetics: gas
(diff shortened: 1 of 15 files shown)


ischoegl (commented):
Thanks, @anthony-walker, for this contribution! It's good to see an example outside of the ordinary. One thing that is missing from my perspective is an example illustrating how the zenith angle works in the context of the "Master Chemical Mechanism" protocol, which appears to be rather specialized and - in its present form - lacks context. 

Another thing that comes to mind is that this would be a terrific application for the new `ExtensibleRate` framework. We are getting quite a few inquiries about custom rate definitions on the user group, and any additional examples would be very welcome. This is mainly a comment, so please take it with a grain of salt.

anthony-walker:
@ischoegl I have had some discussions with @speth and I think this particular application may just be possible with the `ExtensibleRate`. I am going to see about some other more general parameterizations of photolysis rates which would make this more generic but for now I will close it out and save the work such that it may be more generalized in the future.

PR #1507, opened 2023-06-19
Replace Cabinet in clib / update zeroD factory constructors

ischoegl:

- Make `zeroD` factory constructors return `shared_ptr<ReactorBase>`, etc.
- Replace `Cabinet` by `SharedCabinet` in `clib` interface
- Add some unit tests for `ctfunc`
- Fix deprecation warnings in `sofc.py` and `lithium_ion_battery.py` samples

*Caveat:* `Cabinet<Func1>` remains as required updates go beyond the scope of this PR. While some functions can be replaced by classes, internal class methods `Func1::derivative` etc. would need extensive refactoring.

Addresses Cantera/enhancements#164

**Checklist**

### DIFF
--- a/test/clib/test_ctfunc.cpp
+++ b/test/clib/test_ctfunc.cpp
@@ -1,0 +1,104 @@
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+#include <fstream>
+
+#include "cantera/core.h"
+#include "cantera/clib/ctfunc.h"
+#include "cantera/numerics/Func1.h"
+
+using namespace Cantera;
+
+TEST(ctfunc, sin)
+{
+    double omega = 2.;
+    int fcn = func_new(SinFuncType, 0, 1, &omega);
+    ASSERT_GE(fcn, 0);
+    ASSERT_EQ(func_value(fcn, 0.), 0.);
+    ASSERT_EQ(func_value(fcn, 0.5), sin(omega * 0.5));
+}
+
+TEST(ctfunc, cos)
+{
+    double omega = 2.;
+    int fcn = func_new(CosFuncType, 0, 1, &omega);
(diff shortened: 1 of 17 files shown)


test/clib/test_ctfunc.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+using namespace Cantera;
+
+TEST(ctfunc, sin)
+{
+    double omega = 2.;
+    int fcn = func_new(SinFuncType, 0, 1, &omega);
+    ASSERT_GE(fcn, 0);
+    ASSERT_EQ(func_value(fcn, 0.), 0.);
+    ASSERT_EQ(func_value(fcn, 0.5), sin(omega * 0.5));
```
speth: I think the equality comparisons should be done using `EXPECT_DOUBLE_EQ`. Though I'll admit the difference here is fairly minor, it's more consistent with what we should be trying to maintain elsewhere. `ASSERT_` should only be used when we need to stop the test case immediately on a failure. And truly exact floating point results are always iffy -- this relaxes the tolerance to a few units in the last place.

speth (changes requested):
Thanks, @ischoegl. I'm glad to see something that adds a bit of testing for `Func1`, even if that was somewhat incidental to what you were setting out to do. I had just one suggested change.

ischoegl:
> Thanks, @ischoegl. I'm glad to see something that adds a bit of testing for `Func1`, even if that was somewhat incidental to what you were setting out to do. I had just one suggested change.

It was indeed incidental. I did some testing for replacement options for the remaining `Cabinet<Func1>`, where the tests were the quickest way to see whether things worked as expected. Alas, `Func1` does need some refactoring that I don't want to tackle at the moment.

speth (approved):

Issue #752, opened 2019-11-15
'composite' Func in Matlab


ane-sh:
**System information**

- Cantera version: 2.4.0
- OS: Windows 10
- MATLAB version:

Following the post in the google group forum titled 'composite' Functor Matlab, the problem is with composing Functors. One member advised that the code was incorrectly written/implemented.

**To Reproduce**
Taking the example of a function (1- (sin(x)^2))^(1/2), the following code was used

```matlab
fs = Func('fourier',1,[0,0,1,1])  % sin(x)
fs2 = Func('prod',fs,fs)  % (sin(x)^2) 
fs2_1 = Func('diff',Func('polynomial',0,1),fs2)  % 1-(sin(x))^2 
sqrtF = Func('arrhenius',1,[1,0.5,0])  % sqrt function
% now is the challenging part
Func('composite', sqrtF, fs2_1)
```

**Actual behavior**
Either always results in 0 or NaN.
I also tried composing other (simpler) functions but to no avail.

Nonetheless, Cantera great software and indispensable in my toolbox.
Thanks
Aneesh

bryanwweber:
Thank you for posting this issue! For reference, the Google Group discussion is at https://groups.google.com/forum/#!topic/cantera-users/I12aznWPpAk

ischoegl:
Implementing the reported composite function after updates in #1513 in `clib` does not reproduce the error. 

At least when using `clib`, the correct correspondence - $\left|\cos(x)\right|$ - is reproduced.

ischoegl:
Tested this after merging #1513, which included a unit test for the exact example, see: https://github.com/Cantera/cantera/blob/25fee187c474210c14ff45e8647e20507843b075/test/clib/test_ctfunc.cpp#L256-L293
While the unit test passes, the MATLAB result for the composite function defined at the top is still incorrect, although it now yields different results:
```
>> fun = Func('composite', sqrtF, fs2_1)
 
fun = 
 
   0
 
>> fun(.5)

ans =

    0.6924

>> cos(.5)

ans =

    0.8776
```

ischoegl:
As the issue is now isolated to the legacy MATLAB toolbox, which itself was removed via #1670, this will not be fixed.

Issue #567, opened 2018-10-17
Missing functors in matlab toolbox


ghost:
### Cantera version
2.4.0

### Operating System
Win10

### Python/MATLAB version
Matlab2016b

### Expected Behavior
func = Func('pow', base, exponent);
-> func = base^exponent

### Actual Behavior
Pow-Functor and other functors are missing in Matlab Toolbox (https://groups.google.com/forum/#!searchin/cantera-users/matlab$20functor$20in$20matlab%7Csort:date/cantera-users/jBWBBjilFxg/ilkOCfwoBgAJ). 

### Steps to reproduce
omega = 1500/30*pi;
offset   = 263/180*pi;
ct.x       = Func('polynomial', 1, [omega -offset]);
ct.func = Func('pow', ct.x, 0.5);

### Also not working to arbitrarily combine functors
ct.sin = Func('fourier', ct.x, 1, [0 0 omega 1]);
-> ct.sin = sin(Omega*x-Offset)

arjavvashi:
I am working on this issue.

ischoegl:
PR #1513 implements factory constructors for `Func1` objects that avoids hard-coding of functor names (or corresponding magic numbers).

ischoegl:
After removal of the legacy MATLAB toolbox via #1670, all functors will be supported in the experimental toolbox after #1741.

PR #1511, opened 2023-06-22
Fix Conda packaging issues

speth:
**Changes proposed in this pull request**

This is a companion to Cantera/conda-recipes#44 for updates that need to be made in the main Cantera repository.

*Note*: I've also pushed this branch to the `Cantera/cantera` repository to allow me to run the Conda packaging CI against this branch. This will be deleted later.

- Eliminate `CT_HAS_PYTHON` preprocessor variable to enable more efficient incremental builds
- Install `cantera_shared.dll` to the `Library\bin` directory, where all the other conda package DLLs go
- Fix propagation of compiler flags for MSVC in generated `SConstruct` / `CMakeLists.txt` files
- Fix include/lib dirs in the generated build scripts (updates from #1509 were a bit too aggressive in trimming these)
- Don't use SO-versioning for the shim library used to load the Python module from C++
- Fix some compiler warnings on Windows related to integer sizes
- Fix loading the Python "shim" library when installed as a Conda package

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

PR #1513, opened 2023-06-23
Refactor Func1

### LINKED ISSUE #164: Replace Cabinet by SharedCabinet in clib
**Abstract**

Cantera/cantera#1448 introduced `SharedCabinet` to replace raw pointers by smart pointers on the clib backend. While it is (mostly) a drop-in replacement, there are still several instances where the older `Cabinet` is in use. Some of this may be relevant to Cantera/cantera#1182, as shared pointers need significantly less effort for manual cleanup.

**Motivation**

* What problem is it trying to solve? ... eliminate raw pointers
* Who is affected by the change? ... developers - raw pointers are harder to handle
* Why is this a good solution? ... smart pointers have inherent advantages

**Possible Solutions**

Implementation is (in most cases) straight-forward, as it only takes replacing several lines at the top of the implementation files (i.e. replace `Cabinet` by `SharedCabinet`). Some testing will be necessary, as well as the resolution of one instance of `DiagramCabinet::newCopy` in `ctrpath.cpp`, where a suitable solution needs to be implemented by `SharedCabinet`.

**References**

* Cantera/cantera#1448
* Cantera/cantera#1182

ischoegl:

Functors are presumably the easiest pathway to create custom functions using API's that are based on `clib`. This PR seeks to update the framework to facilitate future usage.

- Replace `Func1&` references by `shared_ptr<Func1>` (avoids raw pointers and requirement to copy objects for various operations)
- Implement factory constructors
- Improve unit test coverage
- Eliminate `Cabinet`
- Create new `clib` functions that do not rely on hardcoded magic numbers; magic numbers are, however, not deprecated as they are still used by the stable MATLAB toolbox.
- Eliminate one loophole potentially causing segfaults in `SharedCabinet`

Closes Cantera/enhancements#164

Addresses #752, #567

Directly applicable to Cantera/enhancements#177

**Checklist**


ischoegl:
@speth ... at this point, this is ready for a review. While the line count is a bit larger than expected (a lot of this is due to deprecations, new unit tests and refactored code that otherwise doesn't introduce anything new), this PR should remove a lot of lint from the `Func1` interface, which will be a lot simpler after the deprecations flushed out. As an example, I created a new `Log1` functor, which only required a handful of lines.

The central enhancement are new factory functions (while eliminating the need for magic numbers), where I stuck to MATLAB nomenclature for the time being. Every single `Func1` class is accessible from `clib` at this point, which I hope will  be propagated into the new experimental MATLAB interface. Rather than using a single `clib` (or C++) function signature, I grouped as appropriate, which should make things simpler to use going forward.

include/cantera/clib/ctfunc.h
```diff
@@ -15,7 +15,12 @@ extern "C" {
 #endif
 
     CANTERA_CAPI int func_new(int type, size_t n, size_t lenp, const double* p);
+    CANTERA_CAPI int func_new_basic(const char* type, double c);
+    CANTERA_CAPI int func_new_advanced(const char* type, size_t n, size_t lenp, const double* p);
+    CANTERA_CAPI int func_new_math(const char* type, size_t a, size_t b);
```
speth: Calling one group of these "math" functions is a little confusing, since the rest are all mathematical functions as well. What about "compound" instead?

`func_new_misc` is also a little uninformative.

include/cantera/numerics/Func1Factory.h
```diff
@@ -0,0 +1,145 @@
...
+shared_ptr<Func1> newFunc1(const string& func1Type, size_t n=0,
+                           const vector<double>& params={});
+
+//! Create a new Func1 functor object based on mathematical operations.
+//! @param mathType  String identifying operation.
+//! @param f1  First Func1 object.
+//! @param f2  Second Func1 object.
+//! @since New in Cantera 3.0
+shared_ptr<Func1> newMath1(const string& mathType, const shared_ptr<Func1> f1,
+                           const shared_ptr<Func1> f2);
```
speth: Can we make both versions of `newMath1` just be additional overloads for `newFunc1`? I don't think the name "math" adds anything here.

include/cantera/numerics/Func1Factory.h
```diff
@@ -0,0 +1,145 @@
...
+    //! default constructor, which is defined as private
+    Math1FactoryB();
+
+    //! Mutex for use when calling the factory
+    static std::mutex domain_mutex;
+};
+
+
+//! Create a new Func1 functor object.
+//! @param func1Type  string identifying function type.
```
speth: In the absence of any broader documentation, I think this would be the best place to list the valid options for `func1Type`, for each version of `newFunc1`?

include/cantera/numerics/Func1Factory.h
```diff
@@ -0,0 +1,145 @@
...
+
+private:
+    //! Pointer to the single instance of the factory
+    static Math1FactoryA* s_factory;
+
+    //! default constructor, which is defined as private
+    Math1FactoryA();
+
+    //! Mutex for use when calling the factory
+    static std::mutex domain_mutex;
```
speth: `domain_mutex` is a bit of copypasta (for all three factory classes defined here). Arguably, this could just be `s_mutex` to avoid needing a unique name for every factory class.

include/cantera/numerics/Func1Factory.h
```diff
@@ -0,0 +1,145 @@
...
+//! Create a new Func1 functor object.
+//! @param func1Type  string identifying function type.
+//! @param coeff  Coefficient; definition depends on the function type.
+//! @since New in Cantera 3.0
+shared_ptr<Func1> newFunc1(const string& func1Type, double coeff=1.);
+
+//! Create a new Func1 functor object.
+//! @param func1Type  string identifying function type.
+//! @param n  Integer; definition depends on the function type.
+//! @param params  Parameter vector; definition depends on the function type.
```
speth: If I'm reading the `Func1` classes correctly, `n` is only ever used as a check on the length of `params`. In which case, I'd suggest just dropping it to simplify both this function and the clib `func_new_advanced` method.

include/cantera/numerics/Func1.h
```diff
@@ -285,22 +394,28 @@ class Tabulated1 : public Func1
     Tabulated1(size_t n, const double* tvals, const double* fvals,
                const string& method="linear");
 
+    //! Constructor uses 2*n parameters
+    //! [t0, t1, .. tn-1, f0, f1, .. fn-1]
+    Tabulated1(size_t n, const vector<double>& params);
+
+    //! Set the interpolation method
+    void setMethod(const string& mode);
```
speth: What are the possible options?

include/cantera/numerics/Func1.h
```diff
@@ -285,22 +394,28 @@ class Tabulated1 : public Func1
...
+    virtual string type() const {
         if (m_isLinear) {
-            return *(new Tabulated1(m_tvec.size(), &m_tvec[0], &m_fvec[0],
-                                    "linear"));
-        } else {
-            return *(new Tabulated1(m_tvec.size(), &m_tvec[0], &m_fvec[0],
-                                    "previous"));
+            return "tabulated-linear";
         }
+        return "tabulated-previous";
```
speth, line 515: what does "previous" mean in this context?

src/numerics/Func1.cpp
```diff
@@ -404,60 +860,93 @@ doublereal Func1::isProportional(Func1& other)
...
+    if (f.type() == "constant" && f.c() == 0.0) {
         return true;
     } else {
         return false;
     }
 }
 
+static bool isZero(shared_ptr<Func1> f)
+{
+    return f->type() == "constnat" && f->c() == 0.0;
```
speth:
```suggestion
    return f->type() == "constant" && f->c() == 0.0;
```

src/numerics/Func1.cpp
```diff
@@ -404,60 +860,93 @@ doublereal Func1::isProportional(Func1& other)
...
     } else {
         return false;
     }
 }
 
+static bool isConstant(shared_ptr<Func1> f)
+{
+    return f->type() == "constant";
+}
+
```
speth: Better than having these as `static` functions would be putting them into an unnamed namespace. Also, if you want a very small performance optimization, these could take the argument as `const shared_ptr<Func1>&` to avoid having to increment and decrement the reference count.

src/numerics/Func1.cpp
```diff
@@ -580,8 +1119,64 @@ Func1& newProdFunction(Func1& f1, Func1& f2)
...
+shared_ptr<Func1> newProdFunction(shared_ptr<Func1> f1, shared_ptr<Func1> f2)
+{
+    if (isOne(f1)) {
+        return f2;
+    }
+    if (isOne(f2)) {
+        return f1;
+    }
+    if (isZero(f1) || isZero(f2)) {
+        return shared_ptr<Func1>(new Const1(0.));
```
speth: Throughout here, I think this can be written like:
```suggestion
        return make_shared<Const1>(0.);
```

src/clib/ctfunc.cpp
```diff
@@ -28,78 +26,98 @@ extern "C" {
...
+                r = newFunc1("constant", params[0]);
             } else if (type == FourierFuncType) {
-                if (lenp < 2*n + 2) {
-                    throw CanteraError("func_new",
-                                       "not enough Fourier coefficients");
-                }
-                r = new Fourier1(n, params[n+1], params[0], params + 1,
-                                 params + n + 2);
+                vector<double> par(lenp);
+                std::copy(params, params + lenp, par.data());
```
speth: Any time you're initializing a `vector` from a `double*`:
```suggestion
                vector<double> par(params, params + lenp);
```

include/cantera/numerics/Func1.h
```diff
@@ -704,22 +859,22 @@ class Ratio1 : public Func1
...
-        Func1& a1 = newProdFunction(m_f1->derivative(), m_f2->duplicate());
-        Func1& a2 = newProdFunction(m_f1->duplicate(), m_f2->derivative());
-        Func1& s = newDiffFunction(a1, a2);
-        Func1& p = newProdFunction(m_f2->duplicate(), m_f2->duplicate());
+    virtual shared_ptr<Func1> derivative3() const {
+        auto a1 = newProdFunction(m_f1_shared->derivative3(), m_f2_shared);
+        auto a2 = newProdFunction(m_f1_shared, m_f2_shared->derivative3());
+        auto s = newDiffFunction(a1, a2);
+        auto p = newProdFunction(m_f2_shared, m_f2_shared);
         return newRatioFunction(s, p);
```
speth: I'd suggest moving functions with this length / complexity to the `.cpp` file.

include/cantera/numerics/Func1.h
```diff
@@ -993,11 +1191,12 @@ class Arrhenius1 : public Func1
         return *this;
     }
 
-    virtual Func1& duplicate() const {
-        Arrhenius1* np = new Arrhenius1(*this);
-        return *((Func1*)np);
+    virtual string type() const {
+        return "arrhenius";
```
speth: If we want to stick to the naming convention that we've used elsewhere (YAML), then we should capitalize proper names like Arrhenius, Gaussian, and Fourier.

test/clib/test_ctfunc.cpp
```diff
@@ -4,101 +4,540 @@
...
+
+    auto functor = newFunc1("sin", omega);
+    EXPECT_DOUBLE_EQ(functor->eval(0.5), func_value(fcn, 0.5));
+
+    int dfcn = func_derivative(fcn);
+    EXPECT_DOUBLE_EQ(func_value(dfcn, 0.), omega);
+    EXPECT_DOUBLE_EQ(func_value(dfcn, 0.5), omega * cos(omega * 0.5));
+
+    auto dfunctor = functor->derivative3();
+    EXPECT_DOUBLE_EQ(dfunctor->eval(0.5), func_value(dfcn, 0.5));
```
speth: What does testing the `Func1` object directly this way do that isn't covered by testing via the clib calls?

speth (changes requested):
Thanks for taking the time to clean up this interface. It was past due for an overhaul. Besides moving clib away from magic numbers, I appreciate especially the elimination of the manual memory management with backwards semantics (like returning references to objects that the caller was then responsible for deleting) and the expansion of the test suite.

include/cantera/numerics/Func1.h
ischoegl: Was basing nomenclature on Matlab, but I'm :+1: with this change.

include/cantera/numerics/Func1Factory.h
ischoegl: Not quite. `n` can be the order of a polynomial / Fourier series and does not indicate array size. I switched the order of arguments to make it optional.

include/cantera/numerics/Func1.h
ischoegl: Updated docstring.

include/cantera/numerics/Func1.h
ischoegl: See docstring for `setMethod`. There was some discussion on the naming convention back when this was introduced, with 'previous' being the consensus resolution.

include/cantera/clib/ctfunc.h
ischoegl: I'm :+1: with `func_new_compound`. Regarding `func_new_misc`, this is a catch all for very dissimilar functions (`TimesConstant1`, `PlusConstant1` and `Periodic1`); in other words, the 'uninformative' name is deliberate.

test/clib/test_ctfunc.cpp
ischoegl: Indeed there isn't much new information here. ~However, in absence of a pure C++ test suite covering `Func1`, I'd rather have a little redundancy. I did, however, change the order to test the C++ call before the `clib` call so it will fail first.~

include/cantera/numerics/Func1Factory.h
ischoegl: Documentation is tricky as these lists are unlikely to remain static. I'd much prefer to reorder content of `Func1.h` and use doxygen groupings; ~I refrained from this in this PR though as the diff will be impossible to review. I'd be happy to do this in a separate PR either before or after Cantera 3.0 is released.~

ischoegl: Nevermind - adding the doxygen groups didn't require any reordering. I added `\ref` tags to each of the `newFunc1` versions, and grouped all of the `Func1` classes accordingly.

test/clib/test_ctfunc.cpp
ischoegl: I ended up creating a separate C++ unit test, which is a little clearer.

ischoegl:
@speth ... thanks for the review! I definitely appreciate your suggestions, and implemented ~almost~ everything. ~There is only one minor item left: I can't really think of a good name for the miscellaneous category (other than 'other', which is not an improvement). Everything else should be taken care of.~

ischoegl:
Rebased due to merge conflict with #1511

include/cantera/clib/ctfunc.h
ischoegl: Suggesting 'func_new_modified' ... the classes in question share that they modify a functor.

ischoegl:
Resolved the final comment. 

*Edit: also renamed `func_new_basic` to `func_new_essential` in `clib`.*

include/cantera/clib/ctfunc.h
```diff
@@ -15,7 +15,13 @@ extern "C" {
 #endif
 
     CANTERA_CAPI int func_new(int type, size_t n, size_t lenp, const double* p);
+    CANTERA_CAPI int func_new_essential(const char* type, double c);
+    CANTERA_CAPI int func_new_advanced(
+        const char* type, size_t lenp, const double* p, size_t n);
+    CANTERA_CAPI int func_new_compound(const char* type, size_t a, size_t b);
+    CANTERA_CAPI int func_new_modified(const char* type, size_t a, double c);
```
speth: Sorry for not catching this before, but since the handles returned by clib are `int`s, then the handle arguments here should be `int`s as well.

include/cantera/clib/ctfunc.h
```diff
@@ -15,7 +15,13 @@ extern "C" {
 #endif
 
     CANTERA_CAPI int func_new(int type, size_t n, size_t lenp, const double* p);
+    CANTERA_CAPI int func_new_essential(const char* type, double c);
```
speth: Personally, I kind of preferred `basic` to `essential` -- I don't know what makes $\cos(\omega t)$ more essential than $t^2$. You could also consider `simple` as another option for these functions of a single parameter. I don't think this matters too much though, so your call.

include/cantera/numerics/Func1.h
```diff
@@ -285,22 +458,32 @@ class Tabulated1 : public Func1
     Tabulated1(size_t n, const double* tvals, const double* fvals,
                const string& method="linear");
 
+    //! Constructor uses 2*n parameters
+    //! [t0, t1, .. tn-1, f0, f1, .. fn-1]
```
speth: Take a look at how this is formatted by Doxygen -- it's not pretty. Doxygen takes everything up to the first period as the "summary" and the rest as "details".

include/cantera/numerics/Func1Factory.h
```diff
@@ -0,0 +1,145 @@
...
+    //! Mutex for use when calling the factory
+    static std::mutex s_mutex;
+};
+
+
+//! Create a new essential functor object (see \ref func1essential).
+//! @param func1Type  string identifying function type.
+//! @param coeff  Coefficient; definition depends on the function type.
+//! @since New in Cantera 3.0
+shared_ptr<Func1> newFunc1(const string& func1Type, double coeff=1.);
```
speth: I think it would help to have these functions be `@ingroup Func1`, and/or to have each overload in its corresponding sub-group.

interfaces/matlab_experimental/Func/Func.m
```diff
@@ -92,11 +92,11 @@
...
                 itype = 2;
-            elseif strcmp(typ, 'fourier')
+            elseif strcmp(typ, 'Fourier')
                 itype = 1;
-            elseif strcmp(typ, 'arrhenius')
+            elseif strcmp(typ, 'Arrhenius')
                 itype = 3;
-            elseif strcmp(typ, 'gaussian')
+            elseif strcmp(typ, 'Gaussian')
                 itype = 4;
```
speth, line 100: I'd suggest not changing the API of the legacy Matlab toolbox. If you want, you could make these comparisons case-insensitive to make it a little easier to translate between Matlab and elsewhere.

speth (changes requested):
Thanks for the updates, @ischoegl.

I like the new organization for the related Doxygen documentation. However, this still doesn't inform the user what name to pass to `newFunc1` or `func_new_xyz` to get a function of a particular type, so I think we need to find a home for these lists in the documentation, even if they risk being out of date if new `Func1` classes are added. I'd say that this is somewhat analogous to the list of thermo models that we maintain [here](https://cantera.org/documentation/docs-2.6/sphinx/html/yaml/phases.html#phase-thermodynamic-models).

include/cantera/clib/ctfunc.h
ischoegl: Going with `simple`

include/cantera/numerics/Func1.h
ischoegl: Replaced by math expression.

interfaces/matlab_experimental/Func/Func.m
ischoegl: This is the new experimental toolbox. (this will have to change regardless as it is still using the magic numbers, but I'm not sure when this will happen)

ischoegl:
@speth ... thanks! I believe I took care of everything.

> I like the new organization for the related Doxygen documentation. However, this still doesn't inform the user what name to pass to `newFunc1` or `func_new_xyz` to get a function of a particular type, so I think we need to find a home for these lists in the documentation, even if they risk being out of date if new `Func1` classes are added. I'd say that this is somewhat analogous to the list of thermo models that we maintain [here](https://cantera.org/documentation/docs-2.6/sphinx/html/yaml/phases.html#phase-thermodynamic-models).

I agree in general. At the same time, the generated doxygen documentation actually does a really good job summarizing everything, although it's probably not easy to find. As the code is largely absent from Python, this would have to be done manually on the website ...

image

speth:
Right, I think this documentation page is clear that `newFunc1("???", c)` is what should be called to create a new `Const1` function, but how is the user supposed to know that `???` should be `constant` instead of `const` or `Const1` or something else? This is _almost_ obvious for a lot of the function types, but not all of them.

If you don't want a list in a single place, then an example in each derived class's docstring would be an alternative option.

ischoegl:
@speth ... I added lists of available types to each of the four doxygen groupings, and mentioned the type of each class in (improved) class docstrings. I _hope_ this addresses the remaining comment :smile:

speth (approved):
Thanks, @ischoegl. This looks good to me. Sorry for all the nitpicking on the documentation.

PR #1514, opened 2023-06-24
Make 1D flow domains agnostic of transport model

### LINKED ISSUE #165: Make 1D flow domains agnostic of thermo/transport model
**Abstract**

While it does make sense to have separate implementations for `StFlow` and `IonFlow` in C++, there is no such requirement for high-level API's or factory loaders, as thermo and/or transport models can be used to infer what is needed.

For the Python API, this change would mean the elimination of classes `IdealGasFlow` and `IonFlow`, which would be absorbed by   `FreeFlow` and `AxisymmetricFlow`. As an aside, the nomenclature for `IdealGasFlow` is also affected by a pending PR (Cantera/cantera#1079), which proposes the addition of non-ideal capabilities. A similar rationale applies to other interfaces, as it avoids the creation of thermo/transport-model specific domain classes.

The same idea can be extended to composite objects: `IonFreeFlame` and `IonBurnerFlame` could be absorbed by standard objects, by simply assembling the domains that are correctly detected based on the associated `Solution`.

**Motivation**

* What problem is it trying to solve? ... simplify high-level interfaces
* Who is affected by the change? ... Cantera users
* Why is this a good solution? ... avoids proliferation of user-facing API classes with trivial differences

**Possible Solutions**

After implementing `DomainFactory` in Cantera/cantera#1445, domains can be instantiated using named types, where logic differentiates C++ implementations. This is already done here for `free-flow` and `axisymmetric-flow`: 
https://github.com/Cantera/cantera/blob/8b5a0e9fb02b1bc4994b06f3bbbf23a07dd1259e/src/oneD/DomainFactory.cpp#L46-L55
The main difference would be additional logic to either instantiate `StFlow` or `IonFlow`. In addition, merging of functions in highl-level API's will require some additional coding.

**References**

* Cantera/cantera#1445
* Cantera/cantera#1079

ischoegl:

Switch between `StFlow` and `IonFlow` in C++ factory constructors. This allows for the following:

- Absorb `IonFlow` behavior in conventional `FreeFlow`/`UnstrainedFlow`/`AxisymmetricFlow`
- Absorb `IonFreeFlame` in `FreeFlame`
- Absorb `IonBurnerFlame` in `BurnerFlame`
- Deprecate all "ion" specializations in the user-facing API

Closes Cantera/enhancements#165

```Python
# Instantiate Solution with 'ionized-gas' transport model
gas = ct.Solution('gri30_ion.yaml')
gas.TPX = Tin, p, reactants

# Set up flame object - automatically sets up correct flow domain model
f = ct.FreeFlame(gas, width=width)
```

**Checklist**

### DIFF
--- a/include/cantera/oneD/StFlow.h
+++ b/include/cantera/oneD/StFlow.h
@@ -205,6 +205,26 @@ public:
 
     void solveEnergyEqn(size_t j=npos);
 
+    //! Get the solving stage (used by IonFlow specialization)
+    //! @since New in Cantera 3.0
+    virtual size_t getSolvingStage() const;
+
+    //! Solving stage mode for handling ionized species (used by IonFlow specialization)
+    //! - \c stage=1: the fluxes of charged species are set to zero
+    //! - \c stage=2: the electric field equation is solved, and the drift flux for
+    //!     ionized species is evaluated
+    virtual void setSolvingStage(const size_t stage);
+
+    //! Set to solve electric field in a point (used by IonFlow specialization)
+    virtual void solveElectricField(size_t j=npos);
+
+    //! Set to fix voltage in a point (used by IonFlow specialization)
+    virtual void fixElectricField(size_t j=npos);
+
+    //! Retrieve flag indicating whether electric field is solved or not (used by
+    //! IonFlow specialization)
+    virtual bool doElectricField(size_t j) const;
+
     //! Turn radiation on / off.
     /*!
      *  The simple radiation model used was established by Y. Liu and B. Rogg
@@ -425,9 +445,9 @@ protected:
     //! Update the diffusive mass fluxes.
     virtual void updateDiffFluxes(const doublereal* x, size_t j0, size_t j1);
 
-    //! Get the gradient of species specific molar enthalpies 
+    //! Get the gradient of species specific molar enthalpies
     virtual void grad_hk(const double* x, size_t j);
-    
+
     //---------------------------------------------------------
     //             member data
     //---------------------------------------------------------
--- a/interfaces/cython/cantera/_onedim.pxd
+++ b/interfaces/cython/cantera/_onedim.pxd
@@ -77,11 +77,11 @@ cdef extern from "cantera/oneD/StFlow.h":
         double radiativeHeatLoss(size_t)
         double pressure()
         void setFixedTempProfile(vector[double]&, vector[double]&)
-        size_t getSolvingStage()
-        void setSolvingStage(size_t)
-        void solveElectricField()
-        void fixElectricField()
-        cbool doElectricField(size_t)
+        size_t getSolvingStage() except +translate_exception
+        void setSolvingStage(size_t) except +translate_exception
+        void solveElectricField() except +translate_exception
+        void fixElectricField() except +translate_exception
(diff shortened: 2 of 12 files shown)


interfaces/cython/cantera/_onedim.pxd
```diff
@@ -69,13 +69,19 @@ cdef extern from "cantera/oneD/StFlow.h":
...
         void enableRadiation(cbool)
         cbool radiationEnabled()
         double radiativeHeatLoss(size_t)
         double pressure()
         void setFixedTempProfile(vector[double]&, vector[double]&)
+        size_t getSolvingStage()
+        void setSolvingStage(size_t)
+        void solveElectricField()
+        void fixElectricField()
+        cbool doElectricField(size_t)
```
speth: Most of these need to be marked with `except +translate_exception`, since `StFlow` throws `NotImplementedError`.

src/oneD/DomainFactory.cpp
```diff
@@ -44,17 +45,32 @@ DomainFactory::DomainFactory()
         return new IonFlow(solution, id);
     });
     reg("free-flow", [](shared_ptr<Solution> solution, const string& id) {
-        StFlow* ret = new StFlow(solution, id);
+        StFlow* ret;
+        if (solution->transport()->transportModel() == "ionized-gas") {
+            ret = new IonFlow(solution, id);
+        } else {
+            ret = new StFlow(solution, id);
+        }
```
speth, line 53: This reliance on the specific model name is a little fragile/limiting, in that it doesn't provide a way for any other transport model to be used with `IonFlow`. I believe the only requirement the `IonFlow` class puts on the transport model is that it must implement the `getMobilities` method. Perhaps we can do something like adding a boolean `providesGetMobilities()` method to `Transport` that can be reimplemented by any relevant transport model.

include/cantera/oneD/StFlow.h
```diff
@@ -205,6 +205,21 @@ class StFlow : public Domain1D
 
     void solveEnergyEqn(size_t j=npos);
 
+    //! Get the solving stage
+    virtual size_t getSolvingStage() const;
+
+    //! Set the solving stage
+    virtual void setSolvingStage(const size_t stage);
```
speth: Some extra description to highlight that these are used by `IonFlow` (like in the Python docstrings) would be useful.

speth (changes requested):
Thanks, @ischoegl, I think this makes sense and simplifies flame instantiation a bit. I had just a few suggested changes.

src/oneD/DomainFactory.cpp
ischoegl: Valid comment. However, `MixTransport` defines the function also, which complicates this a bit. As long as "ionized-gas" is the only model that implements this, I believe it's safe as long as it's tested. I added two specific tests for all flame types to `test_oneD.cpp`, which will fail if any of this is touched.

ischoegl:
@speth ... thanks for catching the `translate_exception` - these are nasty crashes.

Regarding your point about the brittle implementation of the `IonFlow` factory constructor switch, I added a `googletest` that should be adequate until the point when someone adds a new transport model that is suitable for 1D simulations.

speth (approved):

PR #1515, opened 2023-06-25
Update reactor API

### LINKED ISSUE #160: Make handling of Reactor Wall functions consistent
**Abstract**

Reactor walls define `wall_Q` as well as `wall_vdot` in `clib`, which have an explicit time dependency although the time is known internally (see https://github.com/Cantera/cantera/pull/1182#discussion_r1044943917). The Python API provides non-pythonic setters, and omits getters. An improved solution should be implemented that is consistent across all interfaces.

**Motivation**

* What problem is it trying to solve? ... inconsistent/incomplete interface
* Who is affected by the change? ... new Matlab Interface Cantera/cantera#1182
* Why is this a good solution? ... consistent interface

**Possible Solutions**

Update underlying C++ functions as needed and adapt Python/Matlab API's.

**References**

Cantera/cantera#1182

### LINKED ISSUE #1460: Duplicate initial gas property in python example ic_engine.py
Gas properties are defined 2 times in [ic_engine.py](https://cantera.org/examples/python/reactors/ic_engine.py.html)

This is just a documentation issue.

Line `gas.TPX = T_inlet, p_inlet, comp_inlet` is duplicated:

```
...
#####################################################################
# Set up Reactor Network
#####################################################################

# load reaction mechanism
gas = ct.Solution(reaction_mechanism, phase_name)

# define initial state and set up reactor
gas.TPX = T_inlet, p_inlet, comp_inlet
cyl = ct.IdealGasReactor(gas)
cyl.volume = V_oT

# define inlet state
gas.TPX = T_inlet, p_inlet, comp_inlet
inlet = ct.Reservoir(gas)
...
```

Suggest to remove duplicate.

ischoegl:

Make Python API more pythonic

- use properties instead of of `set_XYZ` methods in Python
- evaluate functors at reactor network time
- replace `set_master` by `primary` property

API becomes consistent with existing `mfc` behavior
```
>>> mfc.mass_flow_rate = 0.3
>>> mfc.mass_flow_rate = lambda t: 2.5 * exp(-10 * (t - 0.5)**2)
```

Closes Cantera/enhancements#160
Closes #1460

```Python
net.initial_time = 0.                   # instead of net.set_initial_time(0.)
wall.velocity = lambda t: sin(t)        # instead of wall.set_velocity(lambda t: sin(t))
wall.heat_flux = lambda t: cos(t)       # instead of wall.set_heat_flux(lambda t: cos(t))
vdot = wall.expansion_rate              # instead of wall.vdot(net.time)
qdot = wall.heat_rate                   # instead of wall.qdot(net.time)
valve.time_function = lambda t: sin(t)  # instead of valve.set_time_function(lambda t: sin(t))
mfc.pressure_function = lambda p: p**2  # instead of mfc.set_pressure_function(lambda p: p**2)
pc.primary = mfc                        # instead of pc.set_master(mfc)
```

~In `clib`, the new functions `wall_vdot2` and `wall_qdot2` are introduced to preserve current behavior for the old MATLAB toolbox (i.e. no backport) while freeing up `wall_vdot` and `wall_qdot` for all revised interfaces (i.e. reactor network time is automatically used).~ *Edit: no change for traditional MATLAB, new functions `wall_expansionRate` and `wall_heatRate` for revised interfaces.*

**Checklist**

### DIFF
--- a/include/cantera/zeroD/FlowDevice.h
+++ b/include/cantera/zeroD/FlowDevice.h
@@ -85,7 +85,7 @@ public:
      * The calculation of mass flow rate depends to the flow device.
      * @since New in Cantera 3.0.
      */
-    double getPressureFunction();
+    double evalPressureFunction();
 
     //! Set a function of pressure that is used in determining the
     //! mass flow rate through the device. The evaluation of mass flow
@@ -99,7 +99,7 @@ public:
      * The calculation of mass flow rate depends on the flow device.
      * @since New in Cantera 3.0.
      */
-    double getTimeFunction();
+    double evalTimeFunction();
 
     //! Set a function of time that is used in determining
     //! the mass flow rate through the device. The evaluation of mass flow
--- a/include/cantera/zeroD/FlowDevice.h
+++ b/include/cantera/zeroD/FlowDevice.h
@@ -77,15 +77,42 @@ public:
         return *m_out;
     }
 
+    //! Return current value of the pressure function.
+    /*!
+     * The mass flow rate [kg/s] is calculated given the pressure drop [Pa] and a
+     * coefficient set by a flow device specific function; unless a user-defined
+     * pressure function is set, this is the pressure difference across the device.
+     * The calculation of mass flow rate depends to the flow device.
+     * @since New in Cantera 3.0.
+     */
+    double getPressureFunction();
+
     //! Set a function of pressure that is used in determining the
     //! mass flow rate through the device. The evaluation of mass flow
     //! depends on the derived flow device class.
     virtual void setPressureFunction(Func1* f);
 
+    //! Return current value of the time function.
+    /*!
+     * The mass flow rate [kg/s] is calculated for a Flow device, and multiplied by a
+     * function of time, which returns 1.0 unless a user-defined function is provided.
+     * The calculation of mass flow rate depends on the flow device.
+     * @since New in Cantera 3.0.
+     */
+    double getTimeFunction();
+
     //! Set a function of time that is used in determining
     //! the mass flow rate through the device. The evaluation of mass flow
     //! depends on the derived flow device class.
     virtual void setTimeFunction(Func1* g);
+
+    //! Set current reactor network time
+    /*!
+     * @since New in Cantera 3.0.
+     */
+    void setSimTime(double time) {
+        m_time = time;
+    }
 
 protected:
     double m_mdot = Undef;
@@ -98,6 +125,9 @@ protected:
 
     //! Coefficient set by derived classes; used by updateMassFlowRate
     double m_coeff = 1.0;
+
+    //! Current reactor network time
+    double m_time = 0.;
 
 private:
     size_t m_nspin = 0;
--- a/include/cantera/zeroD/Wall.h
+++ b/include/cantera/zeroD/Wall.h
@@ -45,6 +45,12 @@ public:
         return 0.0;
     }
 
+    //! Rate of volume change (m^3/s) for the adjacent reactors at current reactor
+    //! network time.
+    double vdot() {
+        return vdot(m_time);
+    }
+
     //! Heat flow rate through the wall (W).
     /*!
      * This method is called by Reactor::evalWalls(). Base class method
@@ -52,6 +58,11 @@ public:
      */
     virtual double Q(double t) {
         return 0.0;
+    }
+
+    //! Heat flow rate through the wall (W) at current reactor network time.
+    double qdot() {
+        return Q(m_time);
     }
 
     //! Area in (m^2).
@@ -83,11 +94,19 @@ public:
         return *m_right;
     }
 
+    //! Set current reactor network time
+    void setSimTime(double time) {
+        m_time = time;
+    }
+
 protected:
     ReactorBase* m_left = nullptr;
     ReactorBase* m_right = nullptr;
 
     std::vector<ReactorSurface> m_surf;
+
+    //! current reactor network time
+    double m_time = 0.0;
 
     double m_area = 1.0;
 };
@@ -108,6 +127,9 @@ public:
         return "Wall";
     }
 
+    //! Wall velocity \f$ v(t) \f$ at current reactor network time.
+    double velocity() const;
+
     //! Set the wall velocity to a specified function of time, \f$ v(t) \f$.
     void setVelocity(Func1* f=0) {
         if (f) {
@@ -127,6 +149,9 @@ public:
      * decreases in the volume of the reactor on the right.
      */
     virtual double vdot(double t);
+
+    //! Heat flux function \f$ q_0(t) \f$ evaluated at current reactor network time.
+    double heatFlux() const;
 
     //! Specify the heat flux function \f$ q_0(t) \f$.
     void setHeatFlux(Func1* q) {
--- a/include/cantera/zeroD/Wall.h
+++ b/include/cantera/zeroD/Wall.h
@@ -47,6 +47,9 @@ public:
 
     //! Rate of volume change (m^3/s) for the adjacent reactors at current reactor
     //! network time.
+    /*!
+     * @since New in Cantera 3.0.
+     */
     double vdot() {
         return vdot(m_time);
     }
@@ -67,13 +70,16 @@ public:
     /*!
      * This method is called by Reactor::evalWalls(). Base class method
      * does nothing (that is, an adiabatic wall), but may be overloaded.
-     * @since Cantera 3.0
+     * @since New in Cantera 3.0
      */
     virtual double qdot(double t) {
         return 0.0;
     }
 
     //! Heat flow rate through the wall (W) at current reactor network time.
+    /*!
+     * @since New in Cantera 3.0.
+     */
     double qdot() {
         return qdot(m_time);
     }
@@ -108,6 +114,9 @@ public:
     }
 
     //! Set current reactor network time
+    /*!
+     * @since New in Cantera 3.0.
+     */
     void setSimTime(double time) {
         m_time = time;
     }
--- a/include/cantera/zeroD/Wall.h
+++ b/include/cantera/zeroD/Wall.h
@@ -55,14 +55,27 @@ public:
     /*!
      * This method is called by Reactor::evalWalls(). Base class method
      * does nothing (that is, an adiabatic wall), but may be overloaded.
+     * @deprecated To be removed after Cantera 2.6; replaced by qdot(double).
      */
-    virtual double Q(double t) {
+    double Q(double t) {
+        warn_deprecated("WallBase::Q",
+            "To be removed after Cantera 3.0; replaced by qdot(double).");
+        return qdot(t);
+    }
+
+    //! Heat flow rate through the wall (W).
+    /*!
+     * This method is called by Reactor::evalWalls(). Base class method
+     * does nothing (that is, an adiabatic wall), but may be overloaded.
+     * @since Cantera 3.0
+     */
+    virtual double qdot(double t) {
         return 0.0;
     }
 
     //! Heat flow rate through the wall (W) at current reactor network time.
     double qdot() {
-        return Q(m_time);
+        return qdot(m_time);
     }
 
     //! Area in (m^2).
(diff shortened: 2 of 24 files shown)


interfaces/cython/cantera/reactor.pyx
```diff
@@ -911,21 +911,57 @@ cdef class WallBase:
         def __set__(self, double value):
             self.wall.setArea(value)
 
+    @property
```
bryanwweber, line 960: Can the time value here just take a default value that's checked at runtime to avoid defining this method then immediately deleting it? Especially because the warning indicates that people need to change but then change back what their code is doing.

interfaces/cython/cantera/reactor.pyx
```diff
@@ -1227,33 +1350,38 @@ cdef class Valve(FlowDevice):
...
-    .. math:: \dot m = \dot m_{\rm master} + K_v*f(P_1 - P_2)
+    .. math:: \dot m = \dot m_{\rm primary} + K_v*f(P_1 - P_2)
 
     where :math:`f` is the arbitrary function of a single argument.
     """
     flowdevice_type = "PressureController"
 
-    def __init__(self, upstream, downstream, *, name=None, master=None, K=1.):
+    def __init__(self, upstream, downstream, *,
+            name=None, primary=None, K=1., **kwargs):
```
bryanwweber, line 1422: Thanks for making this change!

interfaces/cython/cantera/reactor.pyx
```diff
@@ -1348,13 +1493,31 @@ cdef class ReactorNet:
         """
         return self.net.distance()
 
-    def set_initial_time(self, double t):
+    @property
+    def initial_time(self):
+        raise NotImplementedError("ReactorNet.initial_time")
```
bryanwweber: This seems like it'd be useful to be able to get after a simulation is started.

bryanwweber (commented):
Thanks Ingmar! The master thing has been a wart for a while now 😊

interfaces/cython/cantera/reactor.pyx
ischoegl: This is a valid concern. However, I'd rather leave `vdot(3)` and `qdot(3)` deprecations as proposed for the following three reasons:

1. Evaluating the functions at any time other than the current reactor time yields nonsensical results as they depend on current reactor pressures and temperatures (the `time` argument was only needed to evaluate velocity and/or heat_flux functions).
2. `ExtensibleReactor` recently introduced these as a property, so having differently defined properties and methods of the same name will be confusing, see: https://github.com/Cantera/cantera/blob/d717e8bc6f20afd5dba97b248016e9a9e12f365e/interfaces/cython/cantera/reactor.pyx#L654-L670
3. It's not really a commonly used part of the API (there isn't a single example that accesses either of the methods in question). So I don't expect too much impact on the community.

The deprecation process I am suggesting follows the standard protocol we have been using elsewhere.

ischoegl:
@bryanwweber ... Thanks for the review! I addressed one of your comments (and also made sure that the remaining new getters are consistent); I am hoping to convince you about the other suggested change.

*Edit: Also added a (trivial) fix for #1460 / caught a reference to 'master' in the new Matlab toolbox.*

include/cantera/zeroD/FlowDevice.h
```diff
@@ -77,16 +77,43 @@ class FlowDevice
...
 
+    //! Return current value of the pressure function.
+    /*!
+     * The mass flow rate [kg/s] is calculated given the pressure drop [Pa] and a
+     * coefficient set by a flow device specific function; unless a user-defined
+     * pressure function is set, this is the pressure difference across the device.
+     * The calculation of mass flow rate depends to the flow device.
+     * @since New in Cantera 3.0.
+     */
+    double getPressureFunction();
```
speth: Would `evalPressureFunction()` be a better name for this? `getPressureFunction` gives me the sense that this should return the `Func1` object. Likewise for the "time" function.

interfaces/cython/cantera/reactor.pxd
```diff
@@ -94,8 +94,10 @@ cdef extern from "cantera/zerodim.h" namespace "Cantera":
         void setCoverages(int, double*)
         void setCoverages(int, Composition&) except +translate_exception
         void syncCoverages(int)
+        double vdot()
         double vdot(double)
-        double Q(double)
+        double qdot()
+        double qdot(double)
```
speth: Any of these functions that call a `Func1` probably need to be marked with `except +translate_exception`, since `Func1` can throw in the case where it is implemented as a Python function.

include/cantera/zeroD/Wall.h
```diff
@@ -45,15 +45,45 @@ class WallBase
...
     }
 
+    //! Rate of volume change (m^3/s) for the adjacent reactors at current reactor
+    //! network time.
+    /*!
+     * @since New in Cantera 3.0.
+     */
+    double vdot() {
+        return vdot(m_time);
+    }
```
speth: I think we should avoid having both `vdot(t)` and `vdot()` functions -- I'd suggest deprecating the former and making the latter into the `virtual` method that derived classes (all one of them) should implement. Similar logic applies to the heat flux functions.

include/cantera/zeroD/Wall.h
```diff
@@ -45,15 +45,45 @@ class WallBase
...
+    /*!
+     * This method is called by Reactor::evalWalls(). Base class method
+     * does nothing (that is, an adiabatic wall), but may be overloaded.
+     * @deprecated To be removed after Cantera 2.6; replaced by qdot(double).
+     */
+    double Q(double t) {
+        warn_deprecated("WallBase::Q",
+            "To be removed after Cantera 3.0; replaced by qdot(double).");
+        return qdot(t);
+    }
```
speth: Why is `qdot` preferable to `Q`? Is this for alignment with `clib` and `ReactorDelegator`? This might be the opportunity to standardize on a more descriptive name like `heat_flow` across the board.

test/python/test_reactor.py
```diff
@@ -656,17 +655,26 @@ def test_valve_timing(self):
...
-        valve.set_time_function(lambda t: t>.01)
+        valve.time_function = lambda t: t > .01
 
+        delta_p = lambda: self.r1.thermo.P - self.r2.thermo.P
         mdot = lambda: valve.valve_coeff * (self.r1.thermo.P - self.r2.thermo.P)
         self.net.initialize()
-        self.assertEqual(valve.mass_flow_rate, 0.0)
+        assert valve.time_function == 0.0
+        assert valve.pressure_function == pytest.approx(delta_p())
+        assert valve.mass_flow_rate == 0.0
```
speth: Since we use it so often, this file already has `from pytest import approx` at the top.

interfaces/cython/cantera/reactor.pyx
```diff
@@ -911,21 +911,57 @@ cdef class WallBase:
...
+            To be changed after Cantera 3.0; for new behavior, see property ``vdot3``.
         """
+        warnings.warn(
+            "To be changed after Cantera 3.0; for new behavior, see property 'vdot3'.",
+            DeprecationWarning)
         return self.wall.vdot(t)
 
+    @property
+    def qdot3(self):
+        """
```
speth: Naming the new property `heat_flow` would also avoid needing to introduce this temporary/immediately deprecated property. I'm less sure what to do about `vdot3`; maybe `expansion_rate`?

speth (changes requested):
Thanks for these updates, @ischoegl. I think a lot of this makes sense, and I appreciate the work towards making the different language interfaces more consistent.

I am somewhat ambivalent about the increased level of magical behavior for these properties where the setter takes a function but the getter just returns a value, but as you note we've already made a change that direction with `MassFlowController.mass_flow_rate`, so at least it's increasing consistency.

interfaces/cython/cantera/reactor.pyx
ischoegl: Based on @speths's comments, this is now renamed to `expansion_rate`, where only a single replacement is necessary.

include/cantera/zeroD/Wall.h
ischoegl: I deprecated `vdot(t)` and `Q(t)`, and introduced new virtual functions `expansionRate()` and `heatRate()`.

include/cantera/zeroD/Wall.h
ischoegl: Correct ... there were multiple nomenclatures. Based on comments here and elsewhere, I replaced things by `heatRate` and `expansionRate`.

interfaces/cython/cantera/reactor.pyx
ischoegl: I like the suggestion. For the sake of symmetric naming, I settled on `expansion_rate` and `heat_rate`.

ischoegl:
@speth ... I believe I took care of everything, where I opted for `heat_rate` and `expansion_rate` to replace `Q/qdot` and `vdot`. Note that I applied the suggested updates to `ExtensibleReactor` also. In addition, I noticed the updates you had for `DeprecationWarning`s in #1519, which I applied preemptively here. 

Overall, the deprecations got a lot simpler, where an intermittent "3" suffix is not needed.

> I am somewhat ambivalent about the increased level of magical behavior for these properties where the setter takes a function but the getter just returns a value, but as you note we've already made a change that direction with `MassFlowController.mass_flow_rate`, so at least it's increasing consistency.

Agreed, but I think it is ultimately useful and intuitive. We have other properties that return values without having a setter, so it is in a way consistent from that perspective as well.

interfaces/cython/cantera/reactor.pyx
bryanwweber: Hooray @speth for having the solution 😄

bryanwweber:
@ischoegl @speth My only remaining suggestion here is to use `volumetric_expansion_rate` instead of `expansion_rate` to be explicit. I don't feel strongly though, as it is very wordy.

interfaces/cython/cantera/reactor.pyx
```diff
@@ -651,23 +651,23 @@ cdef class ExtensibleReactor(Reactor):
         def __set__(self, n):
             self.accessor.setNEq(n)
 
-    property vdot:
+    property expansion_rate:
```
speth: I think we need to keep the existing names in deprecated form, since `ExtensibleReactor` was part of the 2.6.0 release and these properties are meant for use by end users. (Only for the Python interface -- the C++ `ReactorDelegator` is effectively internal)

interfaces/cython/cantera/reactor.pyx
```diff
@@ -911,19 +911,49 @@ cdef class WallBase:
...
+            To be removed after Cantera 3.0; replaceable by ``expansion_rate``.
         """
         return self.wall.vdot(t)
 
+    @property
+    def heat_rate(self):
+        """
+        Get the total heat flux [W] through the wall  at the current reactor network
+        time. A positive value corresponds to the left-hand reactor volume increasing,
+        and the right-hand reactor volume decreasing.
```
speth: Bit of documentation copypasta here.

src/clib/ctreactor.cpp
```diff
@@ -503,6 +523,15 @@ extern "C" {
         }
     }
 
+    double walL_heatRate(int i)
```
speth:
```suggestion
    double wall_heatRate(int i)
```

speth (changes requested):
Thanks, @ischoegl. Just a few minor issues to resolve.

Regarding, @bryanwweber's suggestion, I think I lean slightly toward just `expansion_rate`, given the somewhat excessive length of `volumetric_expansion_rate`.

interfaces/cython/cantera/reactor.pyx
ischoegl: Ah, sorry. For some reason I thought it was a new feature. Will use the usual deprecation sequence, and also update the new property to the updated syntax.

ischoegl:
Thanks for catching those items, @speth. I corrected them all.
 
> Regarding, @bryanwweber's suggestion, I think I lean slightly toward just `expansion_rate`, given the somewhat excessive length of `volumetric_expansion_rate`.

I tend to agree.

speth (dismissed):
Looks good to me!

ischoegl:
> Looks good to me!

Thanks! will rebase as #1521 was just merged, so the one new instance of `TabulatedFunction` won't cause trouble.

ischoegl:
@speth ... could you reapprove?

speth (approved):

PR #1521, opened 2023-06-30
Enhance Python Func1 Interface

ischoegl:

- Provide access to all C++ `Func1` objects from Python with dynamically created classes.
- Remove unnecessary argument from advanced functor constructors (oversight in #1513)
- Replace Python callbacks returning constants with C++ functors
- Rename `TabulatedFunction` to `Tabulated1` to ensure names are consistent

While the `Func1` API in Python should mainly serve as an 'advanced feature', it can serve as a template for other API's (example: new MATLAB interface Cantera/enhancements#177).

```
In [1]: import cantera as ct

In [2]: f1 = ct.Func1.cxx_functor("sin")
   ...: f2 = ct.Func1.cxx_functor("Gaussian", [0, 1, 2])
   ...: func = ct.Func1.cxx_functor("sum", f1, f2)

In [3]: func
Out[3]: <cantera.func1.Sum1 at 0x10457cd00>

In [4]: print(func.write())
\sin(t) + \mathrm{Gaussian}(t)
```

**Checklist**

### DIFF
--- a/doc/sphinx/cython/zerodim.rst
+++ b/doc/sphinx/cython/zerodim.rst
@@ -13,7 +13,7 @@ Defining Functions
 
 .. autoclass:: Func1
 
-.. autoclass:: TabulatedFunction
+.. autoclass:: Tabulated1
 
 Base Classes
 ------------
--- a/include/cantera/numerics/Func1.h
+++ b/include/cantera/numerics/Func1.h
@@ -134,7 +134,7 @@ public:
 
     //! Returns a string describing the type of the function
     //! @since  New in Cantera 3.0.
-    string type_name() const;
+    string typeName() const;
 
     //! Calls method eval to evaluate the function
     doublereal operator()(doublereal t) const;
--- a/include/cantera/numerics/Func1.h
+++ b/include/cantera/numerics/Func1.h
@@ -132,6 +132,10 @@ public:
         return "functor";
     }
 
+    //! Returns a string describing the type of the function
+    //! @since  New in Cantera 3.0.
+    string type_name() const;
+
     //! Calls method eval to evaluate the function
     doublereal operator()(doublereal t) const;
 
--- a/include/cantera/numerics/Func1.h
+++ b/include/cantera/numerics/Func1.h
@@ -274,7 +274,7 @@ public:
     }
 
     //! Constructor uses single parameter (frequency)
-    Sin1(const vector<double>& params, size_t n=npos);
+    Sin1(const vector<double>& params);
 
     Sin1(const Sin1& b) :
         Func1(b) {
@@ -323,7 +323,7 @@ public:
     }
 
     //! Constructor uses single parameter (frequency)
-    Cos1(const vector<double>& params, size_t n=npos);
+    Cos1(const vector<double>& params);
 
     Cos1(const Cos1& b) :
         Func1(b) {
@@ -368,7 +368,7 @@ public:
     }
 
     //! Constructor uses single parameter (exponent factor)
-    Exp1(const vector<double>& params, size_t n=npos);
+    Exp1(const vector<double>& params);
 
     Exp1(const Exp1& b) :
         Func1(b) {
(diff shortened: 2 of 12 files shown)


speth:
I'm not really clear on the use case for constructing `Func1` objects in Python, except for `Func1Py` which allows calling arbitrary Python functions. The latter makes it easy to define both simple functions as lambdas as well as arbitrarily complex functions including capturing outside variables, which the piecewise `Func1` construction can never replicate.

ischoegl:
@speth ... this is ready for a review. For the Python `Func1` API, I kept the access outside of the main constructor. On of my initial reasons is that it is possible to recreate the C++ type names dynamically (see [SO example](https://stackoverflow.com/questions/52748764/python-3-set-namespace-for-dynamic-classes)), but I'm not sure that it's worth the effort. Another reason is that I want to keep this an "advanced feature" with the current Python callback behavior remaining the go-to solution.

ischoegl:
> I'm not really clear on the use case for constructing `Func1` objects in Python, except for `Func1Py` which allows calling arbitrary Python functions. The latter makes it easy to define both simple functions as lambdas as well as arbitrarily complex functions including capturing outside variables, which the piecewise `Func1` construction can never replicate.

I am aware that the `Func1Py` approach is a *lot* more flexible. This PR (where most lines are actually a clean-up for #1513) mainly illustrates that it is extremely simple to add access. Python can serve as a test-bed for other API's that lack the generic callbacks (e.g. MATLAB, or prototyping for C++ implementations). In theory, the C++ functors should also be faster than Python callbacks, although I'm not sure how much this matters. Regardless, I replaced the constant lambda functions with the C++ equivalent.

ischoegl:
@speth ... I renamed the factory class method to `cxx_functor`; the factory recreates the underlying C++ classes dynamically, where type names are preserved.

Ad use case: the Python `Func1` API will allow for assigning dedicated C++ functors from the Python interface. Beyond this PR, there is no additional Cython code necessary (all classes are created dynamically).

doc/sphinx/cython/zerodim.rst
```diff
@@ -13,7 +13,7 @@ Defining Functions
 
 .. autoclass:: Func1
 
-.. autoclass:: TabulatedFunction
```
speth, line 16: I thought `TabulatedFunction` was better than `Tabulated1` -- the use of `1` as a suffix to indicate this family of classes on the C++ side is pretty obscure.

include/cantera/numerics/Func1.h
```diff
@@ -132,6 +132,10 @@ class Func1
         return "functor";
     }
 
+    //! Returns a string describing the type of the function
+    //! @since  New in Cantera 3.0.
+    string type_name() const;
```
speth:
```suggestion
    string typeName() const;
```

speth (changes requested):
Thanks, @ischoegl. I'm glad to see that this incorporates the simplification I had been trying to get at in #1513 to eliminate the unnecessary integer argument to the `Func1` constructors. Besides that, just a couple of very minor comments.

doc/sphinx/cython/zerodim.rst
ischoegl: Yes, this was a last-minute change. The reason for this change was that:

```Python
>>> f1 = ct.TabulatedFunction([0, 1, 2], [0, 1, 0])
>>> f2 = ct.Func1.cxx_functor("tabulated-linear", [0, 1, 2, 0, 1, 0])
``` 
would end up having different types, despite being the exact same thing (the `TabulatedFunction` only adds a more user-friendly constructor).

ischoegl:
Thanks, @speth for the prompt review. I can retain the old name for `TabulatedFunction`, but explained why I opted to update. Let me know ... 

*PS: there is one newly introduced `ct.TabulatedFunction` in a test in #1515, which I'll address based on what is merged first.*

doc/sphinx/cython/zerodim.rst
speth: Understood. I think what you've go here makes sense in that case.

speth (approved):
Thanks, @ischoegl, I think this is good to go.

PR #1555, opened 2023-07-24
Simplify BurnerFlame simulations

### LINKED ISSUE #1533: Outlet1D conditions
**Problem description**

In `Outlet1D::eval`, the following conditions are used:

https://github.com/Cantera/cantera/blob/f89efe84ef8a027aaa037e681a421f696469cadf/src/oneD/Boundary1D.cpp#L372

and 

https://github.com/Cantera/cantera/blob/f89efe84ef8a027aaa037e681a421f696469cadf/src/oneD/Boundary1D.cpp#L387-L390

In order to make `lambda` zero, it should presumably use the same condition as for `Inlet1D::eval`, i.e.

https://github.com/Cantera/cantera/blob/f89efe84ef8a027aaa037e681a421f696469cadf/src/oneD/Boundary1D.cpp#L196-L199

While it is currently not used, there is a comment in `OutletRes1D::eval`:

https://github.com/Cantera/cantera/blob/f89efe84ef8a027aaa037e681a421f696469cadf/src/oneD/Boundary1D.cpp#L470-L472

Setting the residual for `c_offset_U` based on the current value of `c_offset_L` to drive the value to zero appears non-intuitive, especially as `Inlet1D::eval` uses the more intuitive `rb[c_offset_L] = xb[c_offset_L];` condition. It makes sense that things converge for axisymmetric/unstrained flows, but it nevertheless appears to be incorrect.

A brute-force replacement with the presumably correct version does lead to convergence issues/failures in `BurnerFlame` unit tests. The likely resolution is to implement the correct condition for `rb[c_offset_U]`.

**System information**

- Cantera version: 3.0.0b1 (and going back many Cantera versions; lines were last updated in #441, but the underlying issue predates that change)
- OS: Any

ischoegl:

Unstrained flames have historically been implemented as axisymmetric flames where 'lambda' was used to specify the mass flow rate implicitly (lambda was propagated left-to-right and velocity propagated right-to-left), which leads to confusing boundary conditions (see #1533). The calculation also involves 'V' (which is zero) for the calculation of continuity, which is certainly not intuitive. Beyond, 'V' was also integrated for free flames, although it is by definition zero.

This PR removes the dependency on 'lambda' and 'V', and instead propagates information on mass flow rate left-to-right directly for the 'U' solution component, meaning that 'lambda'/'V' entries are no longer used for unconstrained flames (i.e. `BurnerFlame` and `FreeFlame` variants).

In addition: 
* simplify logic used for the evaluation of boundary conditions
* only evaluate 'V' for strained flames
* deprecate 'left' outlets, as they are untested and unlikely to work, i.e. only support 'right-to-left' flow for strained flames

Closes #1533

**Checklist**

### DIFF
--- a/include/cantera/oneD/StFlow.h
+++ b/include/cantera/oneD/StFlow.h
@@ -292,7 +292,18 @@ public:
         return m_rho[j];
     }
 
+    //! @deprecated To be removed after %Cantera 3.0. Superseded by isFree()
     virtual bool fixed_mdot();
+
+    /**
+     * Retrieve flag indicating whether flow is freely propagating.
+     * The flow is unstrained and the axial mass flow rate is not specified.
+     * For free flame propagation, the axial velocity is determined by the solver.
+     * @since New in %Cantera 3.0
+     */
+    bool isFree() const {
+        return m_isFree;
+    }
 
     void setViscosityFlag(bool dovisc) {
         m_dovisc = dovisc;
--- a/include/cantera/oneD/StFlow.h
+++ b/include/cantera/oneD/StFlow.h
@@ -305,6 +305,17 @@ public:
         return m_isFree;
     }
 
+    /**
+     * Retrieve flag indicating whether flow uses radial momentum.
+     * If `true`, radial momentum equation for @f$ V @f$ as well as
+     * @f$ d\Lambda/dz = 0 @f$ are solved; if `false`, @f$ \Lambda(z) = 0 @f$ and
+     * @f$ V(z) = 0 @f$ by definition.
+     * @since New in %Cantera 3.0
+     */
+    bool usesLambda() const {
+        return m_usesLambda;
+    }
+
     void setViscosityFlag(bool dovisc) {
         m_dovisc = dovisc;
     }
--- a/include/cantera/oneD/StFlow.h
+++ b/include/cantera/oneD/StFlow.h
@@ -312,7 +312,7 @@ public:
      * @f$ V(z) = 0 @f$ by definition.
      * @since New in %Cantera 3.0
      */
-    bool usesLambda() const {
+    bool isStrained() const {
         return m_usesLambda;
     }
 
--- a/src/oneD/Boundary1D.cpp
+++ b/src/oneD/Boundary1D.cpp
@@ -142,14 +142,19 @@ void Inlet1D::init()
     // if a flow domain is present on the left, then this must be a right inlet.
     // Note that an inlet object can only be a terminal object - it cannot have
     // flows on both the left and right
-    if (m_flow_left) {
+    if (m_flow_left && !m_flow_right) {
+        if (!m_flow_left->usesLambda()) {
+            throw CanteraError("Inlet1D::init",
+                "Right inlets with right-to-left flow are only supported for "
+                "strained flow configurations.");
+        }
         m_ilr = RightInlet;
         m_flow = m_flow_left;
     } else if (m_flow_right) {
         m_ilr = LeftInlet;
         m_flow = m_flow_right;
     } else {
-        throw CanteraError("Inlet1D::init","no flow!");
+        throw CanteraError("Inlet1D::init", "Inlet1D is not properly connected.");
     }
 
     // components = u, V, T, lambda, + mass fractions
@@ -353,10 +358,13 @@ void Outlet1D::init()
     _init(0);
 
     if (m_flow_right) {
-        m_flow_right->setViscosityFlag(false);
+        throw CanteraError("Outlet1D::init",
+            "Left outlets with right-to-left flow are not supported.");
     }
     if (m_flow_left) {
         m_flow_left->setViscosityFlag(false);
+    } else {
+        throw CanteraError("Outlet1D::init", "Outlet1D is not connected.");
     }
 }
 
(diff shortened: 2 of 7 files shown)


ischoegl:
Rebased after merge conflict after #1568; i.e. this is ready for a review.

src/oneD/Boundary1D.cpp
```diff
@@ -366,31 +383,32 @@ void Outlet1D::eval(size_t jg, double* xg, double* rg, integer* diagg,
     integer* diag = diagg + loc();
 
     if (m_flow_right) {
+        throw CanteraError("hi", "hi");
```
speth: Hi. Hi?

src/oneD/StFlow.cpp
```diff
@@ -1014,17 +1025,14 @@ void StFlow::evalRightBoundary(double* x, double* rsd, int* diag, double rdt)
...
-            rsd[index(c_offset_T, j)] = T(x,j) - T_fixed(j);
-        }
+    if (m_usesLambda) {
+        rsd[index(c_offset_U, j)] = rho_u(x, j);
     } else {
-        rsd[index(c_offset_U,j)] = rho_u(x,j) - rho_u(x,j-1);
-        rsd[index(c_offset_T,j)] = T(x,j) - T(x,j-1);
+        rsd[index(c_offset_U, j)] = rho_u(x, j) - rho_u(x, j-1);
     }
+    rsd[index(c_offset_L, j)] = lambda(x, j);
```
speth: In the counterflow case at least, I think this right boundary condition should be `lambda(x, j) - lambda(x, j-1)`. A right `Inlet1D` object does not modify this value, and this results in $\Lambda = 0$ at the right, when it should have the same constant value as at all other grid points.

include/cantera/oneD/StFlow.h
```diff
@@ -292,8 +292,21 @@ class StFlow : public Domain1D
...
 
+    //! Retrieve flag indicating whether flow is free
+    //! @since New in %Cantera 3.0
+    bool isFree() const {
+        return m_isFree;
+    }
+
+    //! Retrieve flag indicating whether flow uses Lambda
+    //! @since New in %Cantera 3.0
+    bool usesLambda() const {
```
speth: I think this would be a good place to add a bit of documentation about what it means for the flow to "use lambda". That is, `usesLambda == true` means that we solve the radial momentum equation and the equation $d\Lambda/dz = 0$, and `usesLambda == false` means we solve the trivial equations $\Lambda(z) = 0$ and $V(z) = 0$.

include/cantera/oneD/StFlow.h
```diff
@@ -292,8 +292,21 @@ class StFlow : public Domain1D
...
     }
 
+    //! @deprecated To be removed after %Cantera 3.0. Superseded by isFree()
     virtual bool fixed_mdot();
 
+    //! Retrieve flag indicating whether flow is free
+    //! @since New in %Cantera 3.0
+    bool isFree() const {
+        return m_isFree;
+    }
```
speth: I think it would help to elaborate on what it means for the flow to be "free", namely, that the axial velocity at the inlet is unspecified and determined by the solver.

To be honest, the `fixed_mdot` nomenclature is more intuitive to me -- while we call the unstrained flame "freely propagating", we solve it by pinning the flame in place.

src/oneD/StFlow.cpp
```diff
@@ -1036,14 +1044,14 @@ void StFlow::evalContinuity(size_t j, double* x, double* rsd, int* diag, double
...
-    if (!m_isFree) {
+    if (m_usesLambda) {
         // Note that this propagates the mass flow rate information to the left
         // (j+1 -> j) from the value specified at the right boundary. The
         // lambda information propagates in the opposite direction.
         rsd[index(c_offset_U,j)] =
             -(rho_u(x,j+1) - rho_u(x,j))/m_dz[j]
             -(density(j+1)*V(x,j+1) + density(j)*V(x,j));
-    } else {
+    } else if (m_isFree) {
```
speth, line 1056: I think there is some further simplification possible here, given that in the `m_isFree` case $V=0$.

src/oneD/StFlow.cpp
```diff
@@ -1060,6 +1068,9 @@ void StFlow::evalContinuity(size_t j, double* x, double* rsd, int* diag, double
                 - (rho_u(x,j+1) - rho_u(x,j))/m_dz[j]
                 - (density(j+1)*V(x,j+1) + density(j)*V(x,j));
         }
+    } else {
+        // unstrained
+        rsd[index(c_offset_U, j)] = rho_u(x, j) - rho_u(x, j - 1);
```
speth: This is specifically the unstrained case with fixed mass flow rate. The "free flame" case above is also unstrained.

speth (changes requested):
Thanks for looking into this, @ischoegl. I think this generally makes sense, with a few possible adjustments:

- I like the change to how the energy equation BCs are handled, with the  `Boundary1D` object always responsible for subtracting a term, whether that's the   boundary temperature or the value from the fixed temperature profile.

- I agree that handling right-to-left flow for the unstrained configurations isn't really necessary. Further, there seem to be a number of errors in the BCs in such a case; I tried creating a version of `FreeFlame` that went in the other direction and couldn't get it to converge in even the simplest of cases. So, if this is non-functional, I think we could just disable it now, without going through the deprecation cycle.
- Reducing the equations for $\Lambda$ and $V$ to just $\Lambda_j = 0$ and $V_j = 0$ in the free flame and burner flame cases makes sense, but I think it needs to be documented a bit better that this is what's happening.

src/oneD/StFlow.cpp
ischoegl: Good catch.

include/cantera/oneD/StFlow.h
ischoegl: Added documentation. Also renamed to `isStrained` to have better symmetry with `isFree`.

include/cantera/oneD/StFlow.h
ischoegl: Added requested documentation.

Regarding nomenclature, `fixed_mdot` is unfortunately ambiguous: it encompasses both strained counter-flow and unstrained burner-stabilized flames. This was the motivation for splitting into `isFree` and `isStrained` (formerly `usesLambda`). Burner flames are neither free nor strained.

ischoegl:
Thanks for the suggestions, @speth! I addressed all comments, and removed support for left outlets. I suspect that there may be some other issues for right-to-left stagnation flow, but didn't test as this goes beyond the original intent of this PR (which was to untangle boundary conditions for unstrained flow).

One upside is that with this change, it becomes possible - for unstrained cases that do not involve electric fields - to use up to three solution array entries to implement custom equations in overloaded `StFlow` classes. For example in #1510, a `c_offset_Uo` could use an existing *unused* component (although I'm not advocating for this). After 3.0 is out, it would also be possible to not allocate memory for unused components altogether, although this would require significant changes.

speth:
> After 3.0 is out, it would also be possible to not allocate memory for unused components altogether, although this would require significant changes.

Yeah, it would be nice to be more flexible about the state vector components, and have that be determined based on the configuration / domain types. One reason I've hesitated about doing this is that there may be a performance hit in having `c_offset_Y` and company not being compile-time constants. These variables get used _a lot_ in the evaluation of the governing equations, so I think it's possible this matters, but it also may not.

speth (approved):

PR #1563, opened 2023-07-30
Add derivative documentation

### LINKED ISSUE #169: Add information on derivative calculation to science section
The PR Cantera/cantera#1089 introduces the capability to calculate derivatives of reaction rate expressions. As noted there:

> The Jacobian implemented in this PR is strictly for the source term; it does not consider constraints posed by the EoS, which can only be handled once additional assumptions inherent to a specific model are known (e.g. constant pressure / volume etc.). As such, results are different from a strictly numerical Jacobian, which includes additional terms due to the EoS closure. Note that these terms are implemented in the test suite where numerical / semi-analytical jacobians are compared.

Per @speth's review comment:

> [...] it makes sense to me that these derivatives are calculated with other quantities held constant as the equations are written, independent of whether that satisfies the equation of state or the sum(X) = 1 constraint. I think it would be very useful to provide a bit of guidance to users on this point, that is, noting that they can calculate other derivatives using the chain rule and whatever equation they need to satisfy, and maybe providing an example of how to calculate a Jacobian at constant volume or at constant pressure.

Due to the number of access function introduced in various API's, it would make sense to collect corresponding information in the Science section of the website.

ischoegl:

Add documentation of derivative calculation. Documentation is added to pre-existing Doxygen section.

Closes Cantera/cantera-website#169

**Checklist**

### DIFF
--- a/include/cantera/kinetics/Kinetics.h
+++ b/include/cantera/kinetics/Kinetics.h
@@ -597,42 +597,119 @@ public:
     //! @addtogroup derivGroup
     //! @{
 
-    //! @name Routines to Calculate Kinetics Derivatives (Jacobians)
-    //!
-    //! Kinetics derivatives are calculated with respect to temperature, pressure,
-    //! molar concentrations and species mole fractions for forward/reverse/net rates
-    //! of progress as well as creation/destruction and net production of species.
-    //!
-    //! The following suffixes are used to indicate derivatives:
-    //!  - `_ddT`: derivative with respect to temperature (a vector)
-    //!  - `_ddP`: derivative with respect to pressure (a vector)
-    //!  - `_ddC`: derivative with respect to molar concentration (a vector)
-    //!  - `_ddX`: derivative with respect to species mole fractions (a matrix)
-    //!  - `_ddCi`: derivative with respect to species concentrations (a matrix)
-    //!
-    //! Settings for derivative evaluation are set by keyword/value pairs using
-    //! the methods getDerivativeSettings() and setDerivativeSettings().
-    //!
-    //! For BulkKinetics, the following keyword/value pairs are supported:
-    //!  - `skip-third-bodies` (boolean) ... if `false` (default), third body
-    //!    concentrations are considered for the evaluation of jacobians
-    //!  - `skip-falloff` (boolean) ... if `false` (default), third-body effects
-    //!    on rate constants are considered for the evaluation of derivatives.
-    //!  - `rtol-delta` (double) ... relative tolerance used to perturb properties
-    //!    when calculating numerical derivatives. The default value is 1e-8.
-    //!
-    //! For InterfaceKinetics, the following keyword/value pairs are supported:
-    //!  - `skip-coverage-dependence` (boolean) ... if `false` (default), rate constant
-    //!    coverage dependence is not considered when evaluating derivatives.
-    //!  - `skip-electrochemistry` (boolean) ... if `false` (default), electrical charge
-    //!    is not considered in evaluating the derivatives and these reactions are
-    //!    treated as normal surface reactions.
-    //!  - `rtol-delta` (double) ... relative tolerance used to perturb properties
-    //!    when calculating numerical derivatives. The default value is 1e-8.
-    //!
-    //! @warning  The calculation of derivatives is an experimental part of the
-    //!      %Cantera API and may be changed or removed without notice.
-    //! @{
+    /**
+     * @name Routines to Calculate Kinetics Derivatives (Jacobians)
+     *
+     * Kinetics derivatives are calculated with respect to temperature, pressure,
+     * molar concentrations and species mole fractions for forward/reverse/net rates
+     * of progress as well as creation/destruction and net production of species.
+     *
+     * The following suffixes are used to indicate derivatives:
+     *  - `_ddT`: derivative with respect to temperature (a vector)
+     *  - `_ddP`: derivative with respect to pressure (a vector)
+     *  - `_ddC`: derivative with respect to molar concentration (a vector)
+     *  - `_ddX`: derivative with respect to species mole fractions (a matrix)
+     *  - `_ddCi`: derivative with respect to species concentrations (a matrix)
+     *
+     * @since New in Cantera 2.6
+     *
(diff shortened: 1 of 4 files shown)


ischoegl:
Rebased and ready for review. BibTeX sort moved to #1567.

include/cantera/kinetics/Kinetics.h
```diff
@@ -597,42 +597,121 @@ class Kinetics
...
-    //!
-    //! @warning  The calculation of derivatives is an experimental part of the
-    //!      %Cantera API and may be changed or removed without notice.
-    //! @{
+    /**
+     * @anchor kinDerivs
+     * @par Routines to Calculate Kinetics Derivatives (Jacobians)
+     * @name
+     *
+     * Kinetics derivatives are calculated with respect to temperature, pressure,
```
speth, line 605: Putting all of the following documentation as part of a `@name` tag results in a giant wall of italicized text, which seems less than ideal. I wonder if this makes more sense as a section of the docstring for the `Kinetics` class, with just a short heading for this member group.

include/cantera/kinetics/Kinetics.h
```diff
@@ -597,42 +597,121 @@ class Kinetics
...
+     * Unless the reaction involves third-body colliders, @f$ \nu_{M,i} = 0 @f$.
+     * For three-body reactions, effective ThirdBody collider concentrations @f$ C_M @f$
+     * are considered with @f$ \nu_{M,i} = 1 @f$. For more detailed information on
+     * relevant theory, see, for example, Perini, et al. @cite perini2012 or Niemeyer,
+     * et al. @cite niemeyer2017, although specifics of %Cantera's implementation may
+     * differ.
+     *
+     * Partial derivatives are obtained from the product rule, where resulting terms
+     * consider reaction rate derivatives, derivatives of the concentration product
+     * term, and, - if applicable, - third-body term derivatives. ReactionRate
```
speth:
```suggestion
     * term, and, if applicable, third-body term derivatives. ReactionRate
```

include/cantera/kinetics/Kinetics.h
```diff
@@ -597,42 +597,121 @@ class Kinetics
...
+     * approximate derivatives that neglect terms to increase computational speed
+     * and/or improve Jacobian sparsity (example: AdaptivePreconditioner).
+     * Derivative evaluations settings are accessible by keyword/value pairs
+     * using the methods getDerivativeSettings() and setDerivativeSettings().
+     *
+     * For BulkKinetics, the following keyword/value pairs are supported:
+     *  - `skip-third-bodies` (boolean) ... if `false` (default), third body
+     *    concentrations are considered for the evaluation of Jacobians
+     *  - `skip-falloff` (boolean) ... if `false` (default), third-body effects
+     *    on rate constants are considered for the evaluation of derivatives.
```
speth:
```suggestion
     *  - `skip-third-bodies` (boolean): if `false` (default), third body
     *    concentrations are considered for the evaluation of Jacobians
     *  - `skip-falloff` (boolean): if `false` (default), third-body effects
     *    on rate constants are considered for the evaluation of derivatives.
```
and so on below.

test/python/test_jacobian.py
```diff
@@ -368,6 +368,43 @@ def test_net_rop_ddP(self):
...
+    def test_net_rate_ddT(self):
+        # check derivatives of net creation rates with respect to temperature
+
+        # constant pressure - need to account for density change
+        dcdt = - self.gas.density_mole / self.gas.T
+        drate = self.gas.net_production_rates_ddT
+        drate += self.gas.net_production_rates_ddC * dcdt
+        drate_num = self.rate_ddT(mode="net", const_p=True)
+        for spc_ix in self.rix + self.pix:
+            self.assertNear(drate[spc_ix], drate_num[spc_ix], self.rtol)
```
speth: Can you use `assert` with `pytest.approx` here instead?

test/python/test_jacobian.py
```diff
@@ -368,6 +368,43 @@ def test_net_rop_ddP(self):
...
+        dt = self.tpx[0] * rtol
+        dp = 0 if const_p else self.tpx[1] * rtol
+        self.gas.TP = self.tpx[0] + dt, self.tpx[1] + dp
+        rate1 = calc()
+        self.gas.TP = self.tpx[:2]
+        rate0 = calc()
+        return (rate1 - rate0) / dt
+
+    def test_net_rate_ddT(self):
+        # check derivatives of net creation rates with respect to temperature
```
speth: I think it would help to explicitly write the identity that's being checked here.

speth (changes requested):

include/cantera/kinetics/Kinetics.h
ischoegl: Interesting, what Doxygen version were you using? It may have something to do with adding this section to both "kinetics" and "derivatives" groups (which is an argument for keeping this under a `@name`), although CSS settings are more likely. Myself, I don't see the 'wall': With Doxygen=1.9.1, I get no italics (under derivatives):

image

and (under Kinetics), normal text, although list items are italicized.

image

This is also the same behavior as you currently see under merged development documentation: https://cantera.org/documentation/dev/doxygen/html/d9/d08/group__derivGroup.html#func-members and 
https://cantera.org/documentation/dev/doxygen/html/d4/dc4/classCantera_1_1Kinetics.html#pub-methods (need to scroll down a bit)

I thought that it may have something to do with adding the 'anchor', but it makes little difference. I *believe* that this may be a CSS issue?

ischoegl:
Thanks for the review comments. I took care of everything except for what I believe is a CSS styling issue?

include/cantera/kinetics/Kinetics.h
speth: I'm seeing this with Doxygen 1.9.6, on the page for the `Kinetics` class:
image

I had not noticed that this content was also being duplicated into the `modules` section. Looking now, I see it is appearing on that page correctly, without the unnecessary italicization.

speth: In any case, I think the actual content here is a good addition, and I'm not that concerned about the formatting details.

speth (approved):

PR #1586, opened 2023-08-11
Fix problems in ThermoPhase and Kinetics classes for the experimental Matlab toolbox

ssun30:

- Fix known problems in ThermoPhase and Kinetics classes discovered by the new test suite.
- Remove filtered tests when problems are fixed.  

Partially address Cantera/enhancements#177

PR #1496 introduced a test suite to the experimental Matlab interfaces. Currently, 12 out of 26 tests implemented in the test suite for the experimental Matlab interface are marked as filtered since they could not pass. This PR will address these problems so they will pass. Some changes to the Clib may be required as some of these problems are not related to the Matlab side of the interface. 

**Checklist**


ssun30:
Sorry I've been busy with class and proposal defense in the past few weeks. I just got some time to go over the PR again. I mentioned to Ray earlier that some methods could not return correct error messages. It turns out these are methods that return certain indices such as `thermo_speciesIndex` and `thermo_elementIndex`. 

They do not throw exceptions when the returned values are out of bounds (npos). Instead, there are dedicated methods to check whether these indices are out of bounds, for example `Phase::checkSpeciesIndex`. From my understanding, it was done this way so that three-body reactions (with undefined species 'M') do not cause errors when they are imported. Therefore I've moved the error handling into the `Clib` layer. 

The other remaining issue with the `ThermoPhase` class in the Matlab interface is that multi-property setters (such as `ThermoPhase.TDY` may set incorrect values. I couldn't pinpoint the cause just yet and will test it on Thursday.

src/clib/ct.cpp
```diff
@@ -312,7 +316,11 @@ extern "C" {
     size_t thermo_speciesIndex(int n, const char* nm)
     {
         try {
-            return ThermoCabinet::item(n).speciesIndex(nm);
+             size_t k = ThermoCabinet::item(n).speciesIndex(nm);
+             if (k == npos) {
+                ThermoCabinet::item(n).checkSpeciesIndex(k);
+             }
+             return k;
```
speth: I think this is getting two things a bit mixed together. The C++ method `speciesIndex` returns `npos` if the named species does not exist. This is a pattern that's used outside of Cantera, for example the Python `str.find()` method returns -1 if the specified substring is not found.

I think it does make sense to change this function to indicate an error if the species is not found. However, this is not the purpose of `checkSpeciesIndex` -- that function is to check if an *input* species index is out of bounds. Instead, you should just throw a `CanteraError` here with the relevant message (which will then immediately be caught by the `catch (...)` block).

interfaces/matlab_experimental/Base/Kinetics.m
```diff
@@ -210,6 +214,10 @@ function delete(kin)
             n = ctFunc('kin_phaseIndex', kin.kinID, phase) + 1;
         end
 
+        function n = get.reactionPhaseIndex(kin)
+            n = ctFunc('kin_reactionPhaseIndex', kin.kinID) + 1;
+        end
+
```
speth: Starting with Cantera 3.0, the reaction phase is required to always be the first phase associated with the `Kinetics` object. Therefore, this method is not necessary.

interfaces/matlab_experimental/Base/Kinetics.m
```diff
@@ -256,7 +264,11 @@ function delete(kin)
                 krange = 1:nsp;
                 irange = 1:nr;
             elseif nargin == 3
-                krange = species;
+                if ischar(species)
+                    krange = {species};
+                else
+                    krange = species;
+                end
```
speth, line 268: Allowing `species` to be a species name makes sense, but please update the docstring accordingly, and be clear if this works for a list of species names or just a single species name.

interfaces/matlab_experimental/Base/Kinetics.m
```diff
@@ -265,17 +277,31 @@ function delete(kin)
...
             end
 
-            n = temp;
+            if nnz(temp) == 1
+                n = nonzeros(temp);
+            elseif nnz(temp) == 0
+                n = 0;
+            else
+                n = temp;
+            end
```
speth: This is changing the type of the return value, from a sparse matrix to just a single value. This is fine, in the case where the input was for a single species and reaction, but probably not desired if a range of species / reactions was specified and only happens to contain one non-zero element. I think you can simplify the logic in this method and decide whether to return a scalar or a matrix based only on the inputs. That is, the scalar option occurs if you have 3 arguments, and `length(krange)` and `length(irange)` are both 1; otherwise, you should return the matrix.

Also, don't forget to update the documentation to indicate the different return types (you can fix the references to the non-existent `stoich_p` method while you're at it).

And of course, this also applies to `productStoichCoeffs`.

interfaces/matlab_experimental/Base/ThermoPhase.m
```diff
@@ -449,19 +459,7 @@ function delete(tp)
...
-            ptr = libpointer('cstring', aa);
-            [iok, bb] = calllib(ctLib, 'thermo_report', tp.tpID, buflen, ptr, 1);
-
-            if iok < 0
-                error(ctGetErr);
-            else
-                disp(bb);
-            end
-
+            ctFunc('thermo_print', tp.tpID, 1, 1);
```
speth: I'd suggest implementing this as `disp(tp.report)`

interfaces/matlab_experimental/Base/ThermoPhase.m
```diff
@@ -34,6 +34,8 @@
         %     String. Can be 'mole'/'molar'/'Molar'/'Mole' or 'mass'/'Mass'.
         basis
 
+        phaseName % Name of the phase.
+
```
speth: Can this be just `name`, for consistency with the Python interface?

interfaces/matlab_experimental/Base/ThermoPhase.m
```diff
@@ -1464,6 +1499,9 @@ function display(tp)
         function set.SV(tp, input)
             s = input{1};
             v = input{2};
+            if strcmp(tp.basis, 'molar')
+                s = s/tp.meanMolecularWeight;
+            end
```
speth: `basis` also affects the dimensions of the "volume" term, doesn't it? For this and all other setters involving `V`? Likewise, the meaning of `D` changes from mass to molar density, if you want to be consistent with the Python interface.

speth (changes requested):
Thanks for these updates, @ssun30. I think this is important progress for the new Matlab toolbox, and I'm glad to see it moving forward. I had just a few suggestions related to the current updates.

interfaces/matlab_experimental/Base/ThermoPhase.m
ssun30: The problem I see with that approach is `thermo_report` currently doesn't have a display threshold for mass and mole fractions. If there are many species in the input file the display could be inconveniently long. Should we modify it so it could accept that as an argument? Then we could maybe deprecate `thermo_print`. 

I made a mistake here as the threshold should be set to a lower value like 1e-4. Currently the threshold is set to 1.

speth (approved):
Thanks, @ssun30! This looks good to me.

PR #1588, opened 2023-08-12
Fix issues for undeclared third-bodies and implicit third-body reactions

### LINKED ISSUE #1403: Unable to load serialized mechanism where reactions contain undeclared third bodies
**Problem description**

When adding reactions dynamically to a `Solution` object in Python, Cantera by default will ignore reactions with undeclared third bodies. However, the corresponding flag is not automatically set (if needed) when saving this `Solution` as a YAML file and the resulting file cannot be loaded without modification.

Specifically, the phase entry in the generated YAML file should contain the option:
```yaml
    skip-undeclared-third-bodies: true
```

**Steps to reproduce**

```python
import cantera as ct

### Reducing mechanism
gri30 = ct.Solution('gri30.yaml')
h2o2 = ct.Solution('h2o2.yaml')

reactions = []
dest_species = set(h2o2.species_names)
for R in gri30.reactions():
    if all(S in dest_species for S in (R.reactants | R.products).keys()):
        reactions.append(R)

gas = ct.Solution(thermo='ideal-gas', kinetics='gas', species=h2o2.species(),
                  reactions=reactions)

gas.write_yaml("reduced.yaml")
restored = ct.Solution("reduced.yaml")
```

**Behavior**

The following traceback is generated
```pytb
  File "/home/speth/src/cantera-related/undeclared_species.py", line 20, in <module>
    restored = ct.Solution("reduced.yaml")
  File "build/python/cantera/base.pyx", line 71, in cantera._cantera._SolutionBase.__cinit__
  File "build/python/cantera/base.pyx", line 128, in cantera._cantera._SolutionBase._cinit
  File "build/python/cantera/base.pyx", line 215, in cantera._cantera._SolutionBase._init_yaml
CanteraError thrown by addReactions:

InputFileError thrown by Reaction::checkSpecies:
Error on line 189 of ./reduced.yaml:
Reaction '2 O + M <=> O2 + M'
defines third-body efficiencies for undeclared species: 'C2H6', 'CH4', 'CO', 'CO2'
|  Line |
|   184 |       rotational-relaxation: 4.0
|   185 | reactions:
|   186 |   - equation: 2 O + M <=> O2 + M
|   187 |     type: three-body
|   188 |     rate-constant: {A: 1.2e+11, b: -1.0, Ea: 0.0}
>   189 >     efficiencies: {AR: 0.83, C2H6: 3.0, CH4: 2.0, CO: 1.75, CO2: 3.6, H2: 2.4, H2O: 15.4}
                            ^
|   190 |   - equation: H + O + M <=> OH + M
|   191 |     type: three-body
|   192 |     rate-constant: {A: 5.0e+11, b: -1.0, Ea: 0.0}
```
(plus similar errors for 6 other reactions).

**Workaround**

This problem can be worked around by setting the necessary flag as "user data". Adding the following line before calling `write_yaml` generates a mechanism file that can be loaded:

```py
gas.update_user_data({'skip-undeclared-third-bodies': True})
```

**System information**

- Cantera version: 2.6.0, or `main` branch at 7684653bb
- OS: Ubuntu 20.04
- Python 3.10 (conda-forge)

**Attachments**

**Additional context**

This is derived from a problem posted on the Users' Group: https://groups.google.com/g/cantera-users/c/ujcAgMT5Vgc

### LINKED ISSUE #1587: Segfault with undefined/explicit third-body collider
**Problem description**

Reactions with explicit third body colliders that are undefined are not caught.

**Steps to reproduce**

Load minimal YAML 

```yaml
generator: YamlWriter
cantera-version: 3.0.0b1
git-commit: 4b30fa74b
date: Sat Aug 12 07:34:23 2023
phases:
  - kinetics: gas
    name: ""
    skip-undeclared-third-bodies: true
    thermo: ideal-gas
    elements: [H, O, Ar, N]
    species:
    - gri30.yaml/species: [H2, H, O, O2, OH, H2O, HO2, H2O2, AR, N2]
    state:
      T: 1.0e-03
      density: 1.0e-03
      Y: {H2: 1.0}
reactions:
  - equation: 2 H + CO2 <=> H2 + CO2
    rate-constant: {A: 5.5e+14, b: -2.0, Ea: 0.0}
```

**Behavior**

Reaction should be skipped (`CO2` is not among species, and `skip-undeclared-third-bodies` is set), but causes segfault instead.

**System information**

- Cantera version: main
- OS: macOS

**Additional context**

Input file is based on output produced for an attempted fix for #1403 (the 'workaround' therein produces the segfault as well)

ischoegl:

Fix a series of bugs and edge cases that involve undeclared third-body colliders and/or implicitly defined third-body reactions.

- Add `skip-undeclared-third-bodies` to serialization (see #1403)
- Prevent segfault if an implicitly defined third-body reaction contains an invalid collider (see #1587)
- Prevent silent skipping of invalid reactions in Python API for creation of `Solution` from scratch
- Fix `samples/python/kinetics/extract_submechanism.py` sample, where reaction list included invalid reactions
- Address several edge cases where `Reaction` objects are created from reactant/product `Composition`, where several invalid combinations were not caught (also, implicit third-bodies were not implemented correctly).
- Deprecate `Reaction.reactants/products` setter in Python, as they allow for undefined behavior

Closes #1403 
Closes #1587

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/reaction.pyx
+++ b/interfaces/cython/cantera/reaction.pyx
@@ -1502,10 +1502,19 @@ cdef class Reaction:
         species names and the values, are the stoichiometric coefficients, for example
         ``{'CH4':1, 'OH':1}``, or as a composition string, for example
         ``'CH4:1, OH:1'``.
+
+        .. deprecated:: 3.0
+
+            Setter for reactants is deprecated and will be removed after Cantera 3.0.
+            Use constructor instead.
         """
         def __get__(self):
             return comp_map_to_dict(self.reaction.reactants)
         def __set__(self, reactants):
+            warnings.warn(
+                "'Reaction.reactants' setter is deprecated and will be removed after "
+                "Cantera 3.0.\nInstantiate using constructor instead.",
+                DeprecationWarning)
             self.reaction.reactants = comp_map(reactants)
 
     property products:
@@ -1514,10 +1523,19 @@ cdef class Reaction:
         species names and the values, are the stoichiometric coefficients, for example
         ``{'CH3':1, 'H2O':1}``, or as a composition string, for example
         ``'CH3:1, H2O:1'``.
+
+        .. deprecated:: 3.0
+
+            Setter for products is deprecated and will be removed after Cantera 3.0.
+            Use constructor instead.
         """
         def __get__(self):
             return comp_map_to_dict(self.reaction.products)
         def __set__(self, products):
+            warnings.warn(
+                "'Reaction.products' setter is deprecated and will be removed after "
+                "Cantera 3.0.\nInstantiate using constructor instead.",
+                DeprecationWarning)
             self.reaction.products = comp_map(products)
 
     def __contains__(self, species):
--- a/interfaces/cython/cantera/reaction.pyx
+++ b/interfaces/cython/cantera/reaction.pyx
@@ -1673,6 +1673,18 @@ cdef class Reaction:
             if self.reaction.usesThirdBody():
                 return ThirdBody.wrap(self.reaction.thirdBody())
 
+    @property
+    def third_body_name(self):
+        """
+        Returns name of `ThirdBody` collider if `Reaction` uses a third body collider,
+        and ``None`` otherwise.
+
+        .. versionadded:: 3.0
+        """
+        if self.reaction.usesThirdBody():
+            return ThirdBody.wrap(self.reaction.thirdBody()).name
+        return None
+
     property efficiencies:
         """
         Get/Set a `dict` defining non-default third-body efficiencies for this reaction,
--- a/src/kinetics/Reaction.cpp
+++ b/src/kinetics/Reaction.cpp
@@ -822,7 +822,9 @@ bool ThirdBody::checkSpecies(const Reaction& rxn, const Kinetics& kin) const
             throw InputFileError("ThirdBody::checkSpecies", rxn.input, "Reaction '{}'\n"
                 "is a three-body reaction with undeclared species: '{}'",
                 rxn.equation(), ba::join(undeclared, "', '"));
-        } else if (kin.skipUndeclaredSpecies() && m_name != "M") {
+        } else if (kin.skipUndeclaredThirdBodies() && m_name != "M") {
+            // Prevent addition of reaction silently as "skip-undeclared-third-bodies"
+            // is set to true
             return false;
         }
     }
--- a/src/kinetics/Reaction.cpp
+++ b/src/kinetics/Reaction.cpp
@@ -35,21 +35,35 @@ Reaction::Reaction(const Composition& reactants_,
     , m_from_composition(true)
     , m_third_body(tbody_)
 {
+    if (reactants.count("M") || products.count("M")) {
+        throw CanteraError("Reaction::Reaction",
+            "Third body 'M' must not be included in either reactant or product maps.");
+    }
     setRate(rate_);
 
     // set flags ensuring correct serialization output
-    bool count = 0;
+    Composition third;
     for (const auto& [name, stoich] : reactants) {
         if (products.count(name)) {
-            count = true;
-            break;
+            third[name] = products.at(name) - stoich;
         }
     }
-    if (count) {
-        if (tbody_ && tbody_->name() != "M") {
+    if (tbody_) {
+        string name = tbody_->name();
+        if (reactants.count(name) && products.count(name)) {
+            throw CanteraError("Reaction::Reaction",
+                "'{}' not acting as third body collider must not be included in both "
+                "reactant and product maps.", name);
+        }
+        if (name != "M") {
             m_third_body->explicit_3rd = true;
-        } else if (!tbody_) {
-            m_explicit_type = true;
+        }
+    } else if (!tbody_ && third.size() == 1) {
+        // implicit third body
+        string name = third.begin()->first;
+        m_third_body = make_shared<ThirdBody>(name);
+        if (name != "M") {
+            m_third_body->explicit_3rd = true;
         }
     }
 }
@@ -61,7 +75,7 @@ Reaction::Reaction(const string& equation,
 {
     setEquation(equation);
     setRate(rate_);
-    if (tbody_ && tbody_->name() != "M") {
+    if (m_third_body && m_third_body->name() != "M") {
         m_third_body->explicit_3rd = true;
     }
 }
--- a/test/python/test_kinetics.py
+++ b/test/python/test_kinetics.py
@@ -304,7 +304,7 @@ class KineticsFromReactions(utilities.CanteraTest):
         surf.net_rates_of_progress_ddCi
 
     def test_electrochemistry_flags(self):
-             # Phases
+        # Phases
         mech = "lithium_ion_battery.yaml"
         anode, cathode, metal, electrolyte = ct.import_phases(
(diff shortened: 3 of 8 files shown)


interfaces/cython/cantera/reaction.pyx
```diff
@@ -1514,10 +1523,19 @@ cdef class Reaction:
...
+
+            Setter for products is deprecated and will be removed after Cantera 3.0.
+            Use constructor instead.
         """
         def __get__(self):
             return comp_map_to_dict(self.reaction.products)
         def __set__(self, products):
+            warnings.warn(
+                "'Reaction.products' setter is deprecated and will be removed after"
+                "Cantera 3.0.\nInstantiate using constructor instead.",
```
speth: Missing a space in "afterCantera 3.0".

interfaces/cython/cantera/reaction.pyx
```diff
@@ -1655,6 +1673,18 @@ cdef class Reaction:
             if self.reaction.usesThirdBody():
                 return ThirdBody.wrap(self.reaction.thirdBody())
 
+    @property
+    def third_body_collider(self):
+        """
+        Returns name of `ThirdBody` collider if `Reaction` uses a third body collider,
+        and ``None`` otherwise.
```
speth, line 1680: Would this be better as `third_body_name`? I think that would make the distinction more clear between this and the `third_body` property.

interfaces/cython/cantera/solutionbase.pyx
```diff
@@ -12,6 +12,7 @@ from .kinetics cimport *
 from .transport cimport *
 from .reaction cimport *
 from ._utils cimport *
+from ._utils import CanteraError
```
speth: It doesn't look like this import is needed/used.

src/kinetics/Reaction.cpp
```diff
@@ -38,19 +38,34 @@ Reaction::Reaction(const Composition& reactants_,
...
+    if (tbody_) {
+        string name = tbody_->name();
+        if (reactants.count("M") || products.count("M")) {
+            throw CanteraError("Reaction::Reaction",
+                "Third body 'M' must not be included in reactant or product lists.");
+        } else if (reactants.count(name)
+                   && reactants.count(name) == products.count(name)) {
+            throw CanteraError("Reaction::Reaction",
+                "Third body '{}' must not have same stoichiometry in reactant and "
+                "product lists.", name);
```
speth: This is not checking the stoichiometry of `name` -- only that it exists in both the `reactants` and `products`.

src/kinetics/Reaction.cpp
```diff
@@ -38,19 +38,34 @@ Reaction::Reaction(const Composition& reactants_,
...
+            third[name] = products.at(name) - stoich;
         }
     }
-    if (count) {
-        if (tbody_ && tbody_->name() != "M") {
+    if (tbody_) {
+        string name = tbody_->name();
+        if (reactants.count("M") || products.count("M")) {
+            throw CanteraError("Reaction::Reaction",
+                "Third body 'M' must not be included in reactant or product lists.");
```
speth: This check is relevant even if `tbody_` is not defined, isn't it?

test/python/test_kinetics.py
```diff
@@ -315,6 +315,33 @@ def test_electrochemistry_flags(self):
         anode_int.derivative_settings = {"skip-electrochemistry": True}
         anode_int.net_rates_of_progress_ddCi
 
+    def test_submechanism(self):
+        # Simplified samples/python/kinetics/extract_submechanism.py
+        gri30 = ct.Solution('gri30.yaml')
+        h2o2 = ct.Solution('h2o2.yaml')
```
speth:
```suggestion
        gri30 = ct.Solution('gri30.yaml', transport_model=None)
        h2o2 = ct.Solution('h2o2.yaml', transport_model=None)
```

samples/python/kinetics/extract_submechanism.py
```diff
@@ -50,6 +51,9 @@
     if not all(product in species_names for product in R.products):
         continue
 
+    if R.third_body_collider not in colliders:
+        continue
+
```
speth: I would like it if we could avoid pushing the responsibility for this filtering step onto the user.

speth (changes requested):
Thanks, @ischoegl. I think this mostly makes sense, with one caveat.

Namely, I think that attempting to add a reaction like `2 H + CO2 <=> H2 + CO2` to a phase that doesn't contain `CO2` using the `Solution` "from parts" constructor should not be treated as an error. The traditional behavior of the `Solution` constructor has been to automatically skip unknown third bodies, and it shouldn't matter whether or not that the unknown species is the only valid third body for a reaction.

In part, this behavior is because we don't currently provide a way for the user to specify that unknown 3rd bodies should be skipped when using this constructor. But more importantly, it's because when constructing a mechanism directly from an input file, it's much more likely that an unknown species is typo that should be detected as an error, whereas when working with `Reaction` objects, I think it's more likely that an unknown species is intentionally not part of the phase definition, and it's easier to have Cantera skip it than make the user do the additional filtering.

src/kinetics/Reaction.cpp
ischoegl: This is actually the intent. The logic can still be simplified: `reactants.count(name) && products.count(name)` - I updated the error message to be less confusing: `"'{<name>}' not acting as third body collider must not be included in both reactant and product maps."`

ischoegl:
Thanks for the comments, @speth!

> [...] I think this mostly makes sense, with one caveat.
> 
> Namely, I think that attempting to add a reaction like `2 H + CO2 <=> H2 + CO2` to a phase that doesn't contain `CO2` using the `Solution` "from parts" constructor should not be treated as an error. The traditional behavior of the `Solution` constructor has been to automatically skip unknown third bodies, and it shouldn't matter whether or not that the unknown species is the only valid third body for a reaction.

The traditional treatment of this was to not detect explicit 3rd body colliders, which means that these species show up in product/reactant maps and thus these reactions were not included.

> In part, this behavior is because we don't currently provide a way for the user to specify that unknown 3rd bodies should be skipped when using this constructor. But more importantly, it's because when constructing a mechanism directly from an input file, it's much more likely that an unknown species is typo that should be detected as an error, whereas when working with `Reaction` objects, I think it's more likely that an unknown species is intentionally not part of the phase definition, and it's easier to have Cantera skip it than make the user do the additional filtering.

~I converted the errors to warnings. For example, the `extract_submechanism.py` now gives a warning, but runs through without modifications:~ ... *Edit: the reaction is now skipped silently*

```
path/to/extract_submechanism.py:58: UserWarning: Skipping three-body reaction(s) with undeclared species:
- H + O2 + AR <=> HO2 + AR    <three-body-Arrhenius>
```

interfaces/cython/cantera/solutionbase.pyx
ischoegl: Used it to trigger a warning in an earlier revision of the PR. Now that the warning is again the behavior, it is needed (again).

ischoegl:
Nevermind. I removed the `Skipping three-body reaction(s) with undeclared species` warning as it addresses a very minor issue that nevertheless opened a can of worms in CI testing (the `run-examples` runner fails if any warnings are issued).

speth (approved):
Thanks, @ischoegl. These updates look good to me. Feel free to merge once the CI passes.

Issue #1310, opened 2022-06-03
Thermo inconsistencies in LatticeSolidPhase


speth:
**Problem description**

The phase does not satisfy basic thermodynamic identities. Some of these are due to #1309, but I think the discrepancies here may be a separate bug.

I am also suspicious of the implementation of activities and activity coefficients for this phase, which has some unusual behavior that is not documented in any way.

**Steps to reproduce**

_Thermo identities:_
```python
>>> import cantera as ct
>>> p = ct.Solution('test/data/thermo-models.yaml', 'Li7Si3_and_interstitials')
>>> p.TP = 725, 50 * ct.one_atm
>>> p.h, p.u + p.P * p.v
-2066297.4664435755 -2077226.57539485  # should be equal
>>> p.chemical_potentials
[-4.62330156e+08 -4.56142503e+07  5.03718894e+05]
>>> p.partial_molar_enthalpies - p.T * p.partial_molar_entropies
[-4.62330156e+08 -4.66072353e+07  2.92167222e+04]
```
The discrepancy between h and u + Pv seems to grow with pressure.

_Activities:_
```python
>>> p.X = [0.2, 0.5, 0.3]
>>> p.activity_coefficients
array([1., 1., 1.])
>>> p.activities
array([1.   , 0.625, 0.375])  # should equal X * activity_coefficients
```

**System information**

- Cantera version: 2.6.0 or `main` at 1ab81ace5

**Additional context**

This was discovered as part of https://github.com/Cantera/cantera/pull/1299, which implements https://github.com/Cantera/enhancements/issues/114.

speth:
The docstring of the `LatticeSolidPhase` notes a rather peculiar modeling choice:
```
 * The mole fraction vector is redefined within the LatticeSolidPhase object.
 * Each of the mole fractions sum to one on each of the sublattices.  The
 * routine getMoleFraction() and setMoleFraction() have been redefined to use
 * this convention.
```
However, while the phase _attempts_ to override `getMoleFractions()`, the base class method `Phase::getMoleFractions` _is not virtual_. This leads to some surprising behavior, where calls to `getMoleFractions` from within the `LatticeSolidPhase` class will use `LatticeSolidPhase::getMoleFractions`, any external calls from a base class pointer will use `Phase::getMoleFractions`. I believe that this is the reason for the problem with computing consistent activities and activity coefficients.

[...]

PR #1589, opened 2023-08-15
Make use of 'override' specifier

speth:

Make use of the [override specifier](https://en.cppreference.com/w/cpp/language/override) for overrides of virtual functions. That is, in derived classes, write
```c++
int foo(double bar) override;
```
instead of
```c++
virtual foo(double bar);
```
where the `virtual` keyword in the derived class actually has no effect. This helps distinguish between virtual functions introduced in a class and overrides of base class methods. More importantly, it helps detect some subtle errors, where an override is desired but does not actually correspond to the base class method, for example by having a different argument type or a difference in `const`ness, where having the `override` specifier will generate an error.

Since the default C++11 behavior is that the `override` specifier is not _required_ (for backwards compatibility), this PR also updates the Clang CI runner to issue a warning in cases where it _could_ be used, and is further configured with `-Werror` to treat such warnings as errors.

**Checklist**

### DIFF
--- a/include/cantera/thermo/GibbsExcessVPSSTP.h
+++ b/include/cantera/thermo/GibbsExcessVPSSTP.h
@@ -94,28 +94,7 @@ public:
     //! @{
 
 protected:
-    /**
-     * Calculate the density of the mixture using the partial molar volumes and
-     * mole fractions as input
-     *
-     * The formula for this is
-     *
-     * @f[
-     * \rho = \frac{\sum_k{X_k W_k}}{\sum_k{X_k V_k}}
-     * @f]
-     *
-     * where @f$ X_k @f$ are the mole fractions, @f$ W_k @f$ are the molecular
-     * weights, and @f$ V_k @f$ are the pure species molar volumes.
-     *
-     * Note, the basis behind this formula is that in an ideal solution the
-     * partial molar volumes are equal to the pure species molar volumes. We
-     * have additionally specified in this class that the pure species molar
-     * volumes are independent of temperature and pressure.
-     *
-     * NOTE: This is a non-virtual function, which is not a member of the
-     *       ThermoPhase base class.
-     */
-    void calcDensity();
+    void calcDensity() override;
 
 public:
     //! @}
@@ -127,8 +106,8 @@ public:
     //! which depends only on temperature and pressure.
     //! @{
 
-    virtual Units standardConcentrationUnits() const;
-    virtual void getActivityConcentrations(double* c) const;
+    Units standardConcentrationUnits() const override;
+    void getActivityConcentrations(double* c) const override;
 
     /**
      * The standard concentration @f$ C^0_k @f$ used to normalize the
@@ -145,8 +124,8 @@ public:
      *
      * @param k species index. Defaults to zero.
      */
-    virtual double standardConcentration(size_t k=0) const;
-    virtual double logStandardConc(size_t k=0) const;
+    double standardConcentration(size_t k=0) const override;
+    double logStandardConc(size_t k=0) const override;
 
     //! Get the array of non-dimensional activities (molality based for this
     //! class and classes that derive from it) at the current solution
@@ -160,13 +139,12 @@ public:
      *
      * @param ac     Output vector of molality-based activities. Length: m_kk.
      */
-    virtual void getActivities(double* ac) const;
+    void getActivities(double* ac) const override;
 
-    virtual void getActivityCoefficients(double* ac) const;
+    void getActivityCoefficients(double* ac) const override;
 
     //! Get the array of temperature derivatives of the log activity coefficients
     /*!
-     * This function is virtual, and first appears in GibbsExcessVPSSTP.
      *
      * units = 1/Kelvin
      *
@@ -177,7 +155,7 @@ public:
         throw NotImplementedError("GibbsExcessVPSSTP::getdlnActCoeffdT");
     }
 
-    virtual void getdlnActCoeffdlnN(const size_t ld, double* const dlnActCoeffdlnN) {
+    void getdlnActCoeffdlnN(const size_t ld, double* const dlnActCoeffdlnN) override {
         throw NotImplementedError("GibbsExcessVPSSTP::getdlnActCoeffdlnN: "
                                   "nonzero and nonimplemented");
     }
@@ -216,15 +194,15 @@ public:
      *  @param vbar   Output vector of species partial molar volumes.
      *                Length = m_kk. units are m^3/kmol.
      */
-    virtual void getPartialMolarVolumes(double* vbar) const;
+    void getPartialMolarVolumes(double* vbar) const override;
     //! @deprecated Unused. To be removed after %Cantera 3.0.
     virtual const vector<double>& getPartialMolarVolumesVector() const;
     //! @}
 
-    virtual bool addSpecies(shared_ptr<Species> spec);
+    bool addSpecies(shared_ptr<Species> spec) override;
 
 protected:
-    virtual void compositionChanged();
+    void compositionChanged() override;
 
     //! utility routine to check mole fraction sum
     /*!
--- a/include/cantera/thermo/HMWSoln.h
+++ b/include/cantera/thermo/HMWSoln.h
@@ -796,7 +796,7 @@ public:
     //! @name  Utilities
     //! @{
 
-    virtual string type() const {
+    string type() const override {
         return "HMW-electrolyte";
     }
 
@@ -804,12 +804,7 @@ public:
     //! @name  Molar Thermodynamic Properties of the Solution
     //! @{
 
-    //! Molar enthalpy. Units: J/kmol.
-    /**
-     * Molar enthalpy of the solution. Units: J/kmol.
-     *      (HKM -> Bump up to Parent object)
-     */
-    virtual double enthalpy_mole() const;
+    double enthalpy_mole() const override;
 
     /**
      * Excess molar enthalpy of the solution from
@@ -845,21 +840,13 @@ public:
      *
      *      (HKM -> Bump up to Parent object)
      */
-    virtual double entropy_mole() const;
+    double entropy_mole() const override;
 
-    //! Molar Gibbs function. Units: J/kmol.
-    /*!
-     *      (HKM -> Bump up to Parent object)
-     */
-    virtual double gibbs_mole() const;
+    double gibbs_mole() const override;
 
-    virtual double cp_mole() const;
+    double cp_mole() const override;
 
-    //! Molar heat capacity at constant volume. Units: J/kmol/K.
-    /*!
-     *      (HKM -> Bump up to Parent object)
-     */
-    virtual double cv_mole() const;
+    double cv_mole() const override;
 
     //! @}
     //! @name Mechanical Equation of State Properties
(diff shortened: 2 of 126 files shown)


src/thermo/LatticeSolidPhase.cpp
```diff
@@ -32,10 +32,9 @@ double LatticeSolidPhase::minTemp(size_t k) const
             }
         }
     }
-    double mm = 1.0E300;
-    for (size_t n = 0; n < m_lattice.size(); n++) {
-        double ml = m_lattice[n]->minTemp();
-        mm = std::min(mm, ml);
+    double mm = 0;
+    for (auto& lattice : m_lattice) {
+        mm = std::max(mm, lattice->minTemp());
```
ischoegl, line 37: I'm not sure that this is equivalent. One appears to get the smallest `minTemp`, and the other one the largest? This may be intentional, but is not part of the `override` replacements.

src/thermo/LatticeSolidPhase.cpp
```diff
@@ -49,10 +48,9 @@ double LatticeSolidPhase::maxTemp(size_t k) const
             }
         }
     }
-    double mm = -1.0E300;
-    for (size_t n = 0; n < m_lattice.size(); n++) {
-        double ml = m_lattice[n]->maxTemp();
-        mm = std::max(mm, ml);
+    double mm = BigNumber;
+    for (auto& lattice : m_lattice) {
+        mm = std::min(mm, lattice->maxTemp());
```
ischoegl, line 53: Similar here. Changes *do* make sense, but the behavior changes. Given that the class is untested, it's hard to tell.

include/cantera/thermo/LatticePhase.h
```diff
@@ -498,17 +498,9 @@ class LatticePhase : public ThermoPhase
...
+    void getCp_R(double* cpr) const override;
 
-    //! Get the molar volumes of the species standard states at the current
-    //! *T* and *P* of the solution.
-    /*!
-     * units = m^3 / kmol
-     *
-     * @param vol     Output vector containing the standard state volumes.
-     *                Length: m_kk.
-     */
```
ischoegl, line 510: While some of the comments slated for removal may be redundant, not all of them *appear* to be interchangeable. Doxygen simply replaces docstrings that are redefined, so I'd prefer to leave those changes until Cantera/enhancements#6 is resolved (which hopefully will also address the lack of citations). 

I will mark all instances I located.

include/cantera/thermo/IdealSolnGasVPSS.h
```diff
@@ -98,27 +82,20 @@ class IdealSolnGasVPSS : public VPStandardStateTP
...
-    virtual double standardConcentration(size_t k=0) const;
+    double standardConcentration(size_t k=0) const override;
 
-    //! Get the array of non-dimensional activity coefficients at the current
-    //! solution temperature, pressure, and solution concentration.
-    /*!
-     *  For ideal gases, the activity coefficients are all equal to one.
-     *
-     * @param ac Output vector of activity coefficients. Length: m_kk.
-     */
```
ischoegl, line 109: Comment may be worth keeping?

(I can locate this twice in the documentation, but the documentation is not inherited)

include/cantera/thermo/IdealSolnGasVPSS.h
```diff
@@ -51,40 +51,24 @@ class IdealSolnGasVPSS : public VPStandardStateTP
...
-     * @f]
-     *
-     * where @f$ X_k @f$ are the mole fractions, @f$ W_k @f$ are the molecular
-     * weights, and @f$ V_k @f$ are the pure species molar volumes.
-     *
-     * Note, the basis behind this formula is that in an ideal solution the
-     * partial molar volumes are equal to the species standard state molar
-     * volumes. The species molar volumes may be functions of temperature and
-     * pressure.
-     */
```
ischoegl, line 81: Comment may be worth keeping?

include/cantera/thermo/IdealMolalSoln.h
```diff
@@ -166,25 +166,7 @@ class IdealMolalSoln : public MolalityVPSSTP
...
-     * @f]
-     *
-     * where @f$ X_k @f$ are the mole fractions, @f$ W_k @f$ are the molecular
-     * weights, and @f$ V_k @f$ are the pure species molar volumes.
-     *
-     * Note, the basis behind this formula is that in an ideal solution the
-     * partial molar volumes are equal to the pure species molar volumes. We
-     * have additionally specified in this class that the pure species molar
-     * volumes are independent of temperature and pressure.
-     */
```
ischoegl, line 186: Comment may be worth keeping?

include/cantera/thermo/HMWSoln.h
```diff
@@ -872,28 +859,7 @@ class HMWSoln : public MolalityVPSSTP
...
-     * weights, and @f$ V_k @f$ are the pure species molar volumes.
-     *
-     * Note, the basis behind this formula is that in an ideal solution the
-     * partial molar volumes are equal to the pure species molar volumes. We
-     * have additionally specified in this class that the pure species molar
-     * volumes are independent of temperature and pressure.
-     *
-     * NOTE: This is a non-virtual function, which is not a member of the
-     *       ThermoPhase base class.
-     */
```
ischoegl, line 895: Comment may be worth keeping?

include/cantera/thermo/GibbsExcessVPSSTP.h
```diff
@@ -94,28 +94,7 @@ class GibbsExcessVPSSTP : public VPStandardStateTP
...
-     * weights, and @f$ V_k @f$ are the pure species molar volumes.
-     *
-     * Note, the basis behind this formula is that in an ideal solution the
-     * partial molar volumes are equal to the pure species molar volumes. We
-     * have additionally specified in this class that the pure species molar
-     * volumes are independent of temperature and pressure.
-     *
-     * NOTE: This is a non-virtual function, which is not a member of the
-     *       ThermoPhase base class.
-     */
```
ischoegl, line 117: Comment may be worth keeping?

ischoegl (changes requested):
Thanks, @speth for another significant cleanup: I believe it is definitely worth the effort, as I noticed several instances where override behavior may have deviated from the original intent.

I did, however, notice that you removed several comments. While they may be indeed redundant, I'd prefer to keep them a little longer as there is little harm, and we're not really making inroads on Cantera/enhancements#6 yet. *Edit: in case the comments are indeed exact replicas, then I'm ok with removal.*

I also noticed a change of behavior in one implementation, which *does make sense*, but should probably be documented a little better.

include/cantera/thermo/LatticeSolidPhase.h
```diff
@@ -284,20 +284,13 @@ class LatticeSolidPhase : public ThermoPhase
...
-        throw NotImplementedError("LatticeSolidPhase::moleFraction");
-    }
-
-    virtual void getMassFractions(double* const y) const {
-        throw NotImplementedError("LatticeSolidPhase::getMassFractions");
-    }
-
-    virtual double massFraction(const int k) const {
-        throw NotImplementedError("LatticeSolidPhase::massFraction");
-    }
```
ischoegl, line 300: Removing these may allow things that are not meant to happen by the original author(s)? ... all of these are valid in the base class, so I believe the overrides are intentional?

src/thermo/LatticeSolidPhase.cpp
speth: Sorry, this is a bit unrelated to the rest of the changes here. I went down a bit of a rabbit hole while seeing if fixing any of the override-related problems in `LatticeSolidPhase` helped with #1310 before giving up and just leaving a comment there. This was the one change that I thought was worth keeping: The minimum temperature for a set of phases should be the minimum temperature that is valid for _all_ of the phases, not _any_ of them, which is what this now implements. That is the main point of this change; the formatting change is secondary. The logic for `maxTemp` is symmetric.

include/cantera/thermo/LatticePhase.h
speth: I am aware of how Doxygen treats docstrings on overridden methods. This method overrides `ThermoPhase::getStandardVolumes`, whose docstring is:

```c++
    //!  Get the molar volumes of the species standard states at the current
    //!  *T* and *P* of the solution.
    /*!
     * units = m^3 / kmol
     *
     * @param vol     Output vector containing the standard state volumes.
     *                Length: m_kk.
     */
```
which is character-for-character identical.

include/cantera/thermo/IdealSolnGasVPSS.h
speth: The only difference between this and the `ThermoPhase::getActivityCoefficients` documentation is the statement
> For ideal gases, the activity coefficients are all equal to one.

However, as you might recall, the ability for this class to represent ideal gases was removed in Cantera 2.6, so this note is irrelevant.

src/thermo/LatticeSolidPhase.cpp
ischoegl: I figured as much. ~As this rationale doesn't show up in the commit (I believe), could you add a comment?~

include/cantera/thermo/IdealSolnGasVPSS.h
speth: This is nearly identical to the documentation of the method that is being overridden ([`VPStandardStateTP::calcDensity`](https://github.com/speth/cantera/blob/e7862d20b4f03eff084c0491725240e2ea043045/include/cantera/thermo/VPStandardStateTP.h#L165-L185))

include/cantera/thermo/IdealMolalSoln.h
speth: Same as above.

include/cantera/thermo/HMWSoln.h
speth: And again.

include/cantera/thermo/LatticeSolidPhase.h
speth: The methods of the base class are not virtual and cannot be overridden. Removing these has no effect on the behavior when the model is used through a base class (i.e. `ThermoPhase`) pointer. I'm sure that things that were not meant to happen by the original author are happening regardless, but I think this is covered by my comment in #1310.

ischoegl: Hah! I clearly overlooked the part of the bases not being `virtual` ...

speth:
> I did, however, notice that you removed several comments. While they may be indeed redundant, I'd prefer to keep them a little longer as there is little harm, and we're not really making inroads on [Cantera/enhancements#6](https://github.com/Cantera/enhancements/issues/6) yet. _Edit: in case the comments are indeed exact replicas, then I'm ok with removal._

I think redundant documentation does do harm: it gives the illusion that there is any meaningful documentation for what these methods do, reducing the impetus to make improvements. While doing a broad scan for this is pretty tedious, I thought it was worth cleaning up a few instances while I was looking at these methods and their base class analogues.

ischoegl (approved):
> I think redundant documentation does do harm: it gives the illusion that there is any meaningful documentation for what these methods do, reducing the impetus to make improvements. While doing a broad scan for this is pretty tedious, I thought it was worth cleaning up a few instances while I was looking at these methods and their base class analogues.

I don't disagree. Not being too familiar with the `ThermoPhase` side of things, I still appreciate confirmations here, as there were no comments in the commit history. Sorry for trying to double-check, but once a comment is removed, it's hard to go back.

include/cantera/thermo/GibbsExcessVPSSTP.h
speth: This is identical to the documentation of the method that is being overridden ([VPStandardStateTP::calcDensity](https://github.com/speth/cantera/blob/e7862d20b4f03eff084c0491725240e2ea043045/include/cantera/thermo/VPStandardStateTP.h#L165-L185))

speth:
> I don't disagree. Not being too familiar with the `ThermoPhase` side of things, I still appreciate confirmations here, as there were no comments in the commit history. Sorry for trying to double-check, but once a comment is removed, it's hard to go back.

No worries. I appreciate the attention to detail. Sorry for not breaking these out in the commit history to make the intention and rationale more clear; I just didn't want to get into a merge conflict fight within my own PR, which all of these would have triggered since they're on adjacent lines to the `override` changes.

PR #1612, opened 2023-09-02
Add Python 3.12 post-merge-test

ischoegl:

This is a minor update to #1604.

- fix several raw strings that lead to syntax warnings in Python 3.12 
- add Python 3.12 post-merge test as recommended by @bryanwweber (note that the runner is currently still failing due to upstream limitations, see https://github.com/Cantera/cantera/actions/runs/6061035003/job/16445794850)
- force system libraries in ubuntu docker post-merge-test
- add hdf5 to fedora docker post-merge-test

At this point, fedora `rawhide` is still the only runner that uses Python 3.12 successfully.

**Checklist**

### DIFF
--- a/.github/workflows/post-merge-tests.yml
+++ b/.github/workflows/post-merge-tests.yml
@@ -124,10 +124,12 @@ jobs:
         python3-wheel sundials-devel yaml-cpp-devel hdf5-devel highfive-devel
     - name: Build Cantera
       run: |
-        scons build -j2 debug=n --debug=time \
-        extra_inc_dirs=/usr/include/eigen3 f90_interface=y \
-        libdirname=/usr/lib64 python_package=full system_eigen=y system_fmt=y \
-        system_sundials=y system_yamlcpp=y system_blas_lapack=y hdf_support=y
+        scons build -j2 debug=n --debug=time python_package=full f90_interface=y \
+        extra_inc_dirs=/usr/include/eigen3 libdirname=/usr/lib64 \
+        system_eigen=y system_fmt=y system_blas_lapack=y system_sundials=y \
+        system_yamlcpp=y system_blas_lapack=y hdf_support=y
+      # note: 'system_highfive=y' is omitted as the current packaged version is too old;
+      # once newer version is available in 'latest', this should be tested as well
     - name: Test Cantera
       run:
         scons test verbose_tests=yes --debug=time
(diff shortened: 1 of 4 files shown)


.github/workflows/post-merge-tests.yml
```diff
@@ -119,13 +121,57 @@ jobs:
...
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+        architecture: x64
+        allow-prereleases: true
+    - name: Install Apt dependencies
+      run: |
+        sudo apt update
+        sudo apt install libboost-dev gfortran libopenmpi-dev libpython3-dev \
+        libblas-dev liblapack-dev libhdf5-dev libfmt-dev
```
speth: I don't think you want/need to include `libpython3-dev` here, if we're trying to test the pre-release Python installed by the `setup-python` action.

.github/workflows/post-merge-tests.yml
```diff
@@ -119,13 +121,57 @@ jobs:
...
         python3-pytest python3-ruamel-yaml python3-scipy python3-scons \
-        python3-wheel sundials-devel yaml-cpp-devel
+        python3-wheel sundials-devel yaml-cpp-devel hdf5-devel highfive-devel
     - name: Build Cantera
       run: |
         scons build -j2 debug=n --debug=time \
         extra_inc_dirs=/usr/include/eigen3 f90_interface=y \
         libdirname=/usr/lib64 python_package=full system_eigen=y system_fmt=y \
-        system_sundials=y system_yamlcpp=y system_blas_lapack=y
+        system_sundials=y system_yamlcpp=y system_blas_lapack=y hdf_support=y
```
speth: Also `system_highfive=y`?

.github/workflows/post-merge-tests.yml
```diff
@@ -119,13 +121,57 @@ jobs:
...
+      run: python3 -m pip install -U pip setuptools wheel
+    - name: Install Python dependencies
+      run: |
+        python3 -m pip install ruamel.yaml scons==3.1.2 numpy cython pandas pytest \
+        pytest-github-actions-annotate-failures pint
+    - name: Build Cantera
+      run: |
+        python3 `which scons` build env_vars=all -j2 debug=n --debug=time \
+        system_fmt=y hdf_libdir=$HDF5_LIBDIR hdf_include=$HDF5_INCLUDEDIR \
+        cc_flags=-D_GLIBCXX_ASSERTIONS
```
speth: Do you also want to force `system_blas_lapack=y`, since it's being installed?

.github/workflows/post-merge-tests.yml
```diff
@@ -119,13 +121,57 @@ jobs:
...
+      run: |
+        sudo apt update
+        sudo apt install libboost-dev gfortran libopenmpi-dev libpython3-dev \
+        libblas-dev liblapack-dev libhdf5-dev libfmt-dev
+        gcc --version
+    - name: Upgrade pip
+      run: python3 -m pip install -U pip setuptools wheel
+    - name: Install Python dependencies
+      run: |
+        python3 -m pip install ruamel.yaml scons==3.1.2 numpy cython pandas pytest \
```
speth: I think we can probably stop pinning `scons==3.1.2` on all but one or two builds.

speth (changes requested):
Thanks, @ischoegl. This makes sense to me. I had just a couple of minor suggestions.

.github/workflows/post-merge-tests.yml
ischoegl: I included the fedora package, but it is currently not used - the version is too old and the installer uses the private installation as a fallback instead (there is a pending fix upstream, - see https://github.com/Cantera/cantera/pull/1604#issuecomment-1703996219, - so hopefully with the next fc release this will be moot). Based on this feedback, I added a separate test at the end of the runner that should illustrate this. The logic currently still uses the bypass first as this is the only way to get to test Python 3.12 at the moment.

speth: Right, I forgot about that discussion. I think a comment explaining the reason for not specifying `system_highfive=y` despite installing the corresponding package would be sufficient. I would rather not introduce new tests that just fail out of the box.

ischoegl: Sounds good.

ischoegl:
@speth - thank you for the suggestions. I believe everything is addressed.

speth (approved):
Thanks, @ischoegl - this looks good to me.

PR #1621, opened 2023-09-23
Render examples using Sphinx-gallery

### LINKED ISSUE #88: Recategorize Website Examples
**Abstract**

The examples on the [website](https://cantera.org/examples) are currently categorized by relevant programming language. This organization is fine as far as it goes, but isn't particularly useful if one wants to see examples related to thermodynamics, kinetics, etc. For these cases, it'd be useful to have multiple ways of sorting or categorizing the examples.

**Motivation**

In general, the examples are not all that easy to find, if you know the problem domain you're trying to solve but not where to find it.

**Possible Solutions**

This could be done by implementing a tag system, possibly leveraging the existing tagging functionality built in to Nikola for blog posts.

speth:

- Use `sphinx-gallery` to render the examples.
- Use `sphinx-tags` to parse the example keywords and create index pages for each tag
- Reformat example headers to use reST format so their content can be parsed by `sphinx-gallery` and `sphinx-tags`

**Notes:**
- There is a lot of room to use more reST formatting in the examples. Beyond the headers, I only did a little bit of that here mostly as a test. I think doing more of that can be handled in future PRs that address a smaller number of examples and include improvements beyond just formatting updates.
- Currently, about half the Python examples have plot output that is used to generate an "interesting" icon for the gallery. Since `sphinx-gallery` can't run the other language examples (at least for the time being), those galleries are full of the default image.
- To enhance the visual appeal, I was considering finding or making some icons to use for at least some of the examples.
- The automatic linking from the Python examples to the API docs provided by `sphinx-gallery` is pretty awesome.
- This PR relies on a bleeding-edge version of `sphinx-gallery` that includes a pending PR to introduce the ability to parse examples in languages other than Python.
- This PR relies on a bleeding-edge version of `sphinx-tags` that includes a few fixes I introduced. This extension is very small (a single file) so we could just vendor it if we're going to introduce significant customizations.
- Running all the Python samples massively increases the time it takes to build the docs. To make local testing of updates to the API docs easier, it will probably be worth adding an option to SCons to exclude all of the example-related content.
- For the example reformatting, I only did the new/experimental Matlab toolbox examples, rather than the "legacy" ones. I really didn't want to have to do all of those examples twice, and I'm hoping we can make the transition to the new toolbox as part of the next Cantera release.
- Migrating the Jupyter notebook examples into this will be a separate PR.

- One step toward addressing Cantera/enhancements#178
- Partially addresses Cantera/cantera-website#209
- Resolves Cantera/enhancements#88

The page for a Python example:

Screenshot 2023-09-23 at 5 04 55 PM

Part of an index page:

Screenshot 2023-09-23 at 5 20 57 PM

**Checklist**


doc/sphinx/_static/custom.css
ischoegl, line 1: Missing newline at end of file.

doc/sphinx/conf.py
```diff
@@ -42,8 +42,118 @@
...
+}
+
+# Override sphinx-gallery's method for determining which examples should be executed.
+# There's really no way to achieve this with the `filename_pattern` option, and
+# `ignore_pattern` excludes the example entirely.
+skip_run = {
+    # multiprocessing can't see functions defined in __main__ when run by
+    # sphinx-gallery, at least on macOS.
+    "multiprocessing_viscosity.py",
+    # __file__ deliberately not available when run by sphinx-gallery
```
ischoegl: Why? (Line 83)

samples/clib/demo.c
```diff
@@ -1,11 +1,13 @@
-/**
+/*
```
ischoegl, line 1: While it’s interesting to see ReST in a docstring, we really had not configured doxygen to collect examples. I’m :+1: on this. *Edit: fwiw, this is just an alternative to doxygen syntax, which itself isn’t specific to C++.*

samples/matlab_experimental/plug_flow_reactor.m
```diff
@@ -8,19 +8,19 @@
 %
 % The reactor assumes that the flow follows the Ideal Gas Law.
 %
-% The governing equations used in this code can be referenced at:
+% The governing equations used in this code (see :ref:`PFR_solver.m
+% <sphx_glr_examples_matlab_experimental_PFR_solver.m>`) can be referenced at:
```
ischoegl: The generated links are somewhat convoluted.

samples/matlab_experimental/plug_flow_reactor.m
```diff
@@ -8,19 +8,19 @@
...
 % McGraw Hill Education, India, 2012, 206-210.*
 %
 % The current example is written for methane combustion, but can be readily
 % adapted for other chemistries.
 %
-% Developed by Ashwin Kumar/Dr.Joseph Meadows (mgak@vt.edu/jwm84@vt.edu) on 3-June-2020
-% Research Assistant/Assistant Professor
-% Advanced Propulsion and Power Laboratory
-% Virginia Tech
+% Example originally developed by Ashwin Kumar (Research Assistant, mgak@vt.edu) and
```
ischoegl: This (pre-existing) acknowledgment is out of the norm, as we usually add contributors elsewhere.

samples/python/onedim/diffusion_flame_batch.py
ischoegl, line 1: Looks like we’re not consistent with the license string for examples at the top (others skip it).

samples/python/reactors/NonIdealShockTube.py
ischoegl, line 1: Do we need the first line?

samples/python/reactors/PorousMediaBurner.py
```diff
@@ -52,8 +48,9 @@
 radiation. For more realistic and quantitative predictions, running 1D simulations with
 full radiation transport is required.
 
-More details about the governing equations, submodels and physical setup and
-corresponding experiments can be found in
+Initial code by Thorsten Zirwes and Guillaume Vignat, Stanford University & Karlsruhe
```
ischoegl: Another example for inconsistent author acknowledgements. In this case, the citation should suffice.

samples/python/reactors/preconditioned_integration.py
ischoegl, line 1: Another `coding` directive.

samples/python/surface_chemistry/sofc.py
```diff
@@ -34,7 +39,7 @@
 # gas compositions. Change as desired.
 anode_gas_X = 'H2:0.97, H2O:0.03'
 cathode_gas_X = 'O2:1.0, H2O:0.001'
-
+0
```
ischoegl: ?

ischoegl (changes requested):
@speth … this is terrific. While I believe that your PR over at `sphinx-gallery` is making good progress, I have a couple of minor nitpicks (most are lines that touch or are adjacent to pre-existing conditions).

It may make sense to wait what happens with the contribution to the upstream library?

doc/sphinx/conf.py
speth: https://sphinx-gallery.github.io/stable/faq.html#why-is-file-not-defined-what-can-i-use

ischoegl: Ah, ok. To me, this wasn’t clear. I guess we could avoid this by using `os.getcwd()` as suggested?

samples/matlab_experimental/plug_flow_reactor.m
speth: They're certainly _long_. I realized that there is an alternative, which is to link to the page rather than the heading, which can be done like ``` :doc:`PFR_solver.m <PFR_solver>` ``` where the bracketed text needs to specify a relative path, e.g. for some of the Python links, we'd have something like ``` :doc:`reactor1.py <../reactors/reactor1>` ```.

samples/python/onedim/diffusion_flame_batch.py
speth: I'm not sure what the right way to think of the examples is, in terms of licensing. I think they are frequently used in a way that is not quite consistent with what even the permissive BSD license requires, that is, with users adapting the examples to their own purposes, but not retaining the license file. At a certain level, this use case would be better accommodated by saying that the examples are public domain, but that's definitely out of scope for this PR.

speth:
> It may make sense to wait what happens with the contribution to the upstream library?

Yes, we can wait to merge this in case there are any API changes requested upstream. I don't think I want to wait for the next release of `sphinx-gallery`, though. I may start building another PR on top of this one with more documentation updates.

doc/sphinx/conf.py
speth: Maybe, but this isn't exactly the same. For example, when this is run as part of the `run-examples` workflow, the working directory is the root of the repository. The output for this example isn't that interesting anyway -- it looks pretty much just like the other simple flame examples, so I don't think we're missing much.

ischoegl: True. At the same time, I hate to see an example that some users may find interesting left behind. I’d suggest to add the contents of `tdata.dat` as a multi-line string within the Python script, instead of omitting the example from the documentation.

samples/python/onedim/diffusion_flame_batch.py
ischoegl: Ok. I would have just removed it, but the question about licensing is a fair point.

doc/sphinx/conf.py
speth: Ah, to be clear: the example still appears in the documentation. We just don't run it and capture output (so, same as what's happening for the non-Python examples). However, I realize we do need to provide a link to the `tdata.dat` though, for anyone who downloads the example.

samples/python/surface_chemistry/sofc.py
speth: Whoops.

.github/workflows/main.yml
```diff
@@ -317,16 +317,22 @@ jobs:
...
-          sphinxcontrib-matlabdomain sphinxcontrib-doxylink pint \
-          pydata-sphinx-theme==0.13.3 sphinx-argparse
-      - name: Build Cantera with documentation
-        run: python3 `which scons` build -j2 doxygen_docs=y sphinx_docs=y debug=n optimize=n use_pch=n
+          python3 -m pip install ruamel.yaml scons numpy cython sphinx==6.2.1 \
+            sphinxcontrib-matlabdomain sphinxcontrib-doxylink pint \
+            pydata-sphinx-theme==0.13.3 sphinx-argparse sphinx_design matplotlib \
+            pandas scipy
+          python3 -m pip install "git+https://github.com/speth/sphinx-gallery.git@multi-src-lang"
+          python3 -m pip install "git+https://github.com/speth/sphinx-tags.git@tags-with-spaces"
```
ischoegl: Following up on the earlier conversation. I agree that it's definitely not necessary to wait for new upstream releases, but I'd suggest to wait until we can use official repos here.

ischoegl (commented):
As I didn't want to pip install from GitHub repos, I downloaded the artifacts from GH Actions: things look overall great. 

The only thing I noticed are mislabeled download buttons (Fortran is not Python and definitely doesn't know Jupyter 😃 ; C++/C/Matlab have the same issue ... not sure whether this needs to be fixed here or upstream):
image

I also noticed some "FortranFixed" reference that I am not sure about
image

doc/sphinx/conf.py
ischoegl: 👍 makes sense. At the same time, we recently had a confused user asking on the UG about the format to be used for this exact example. I believe that including data in the file itself would resolve all issues.

speth:
The "download all" buttons and corresponding zip files are something I still need to fix upstream.

For the individual examples, the name is the one used by Pygments (first column [here](https://pygments.org/languages/)), which uses "Fortran" for F90-style "free format" and "FortranFixed" for the older fixed-width format. I think it's a pretty minor wart.

bryanwweber:
Thanks for doing this Ray! I'm happy we'll be able to link to API docs from the examples, and that you've got all the examples building. Once this goes in, I'll rework the ongoing sphinx migration to use this instead of my bespoke solution.

ischoegl:
> The "download all" buttons and corresponding zip files are something I still need to fix upstream.

Thanks for the upstream fix - I retriggered GH actions, and things are taken care of :tada:

> For the individual examples, the name is the one used by Pygments (first column [here](https://pygments.org/languages/)), which uses "Fortran" for F90-style "free format" and "FortranFixed" for the older fixed-width format. I think it's a pretty minor wart.

Thanks for the explanation. It indeed is a minor wart.

ischoegl:
🥳  ... `sphinx-gallery` PR is merged.

.github/workflows/main.yml
speth: Per our discussion on Cantera/enhancements#178, these are now updated to use `sphinx-gallery/sphinx-gallery` and `Cantera/sphinx-tags`.

ischoegl (approved):
Thanks, @speth for this substantial addition!

PR #1626, opened 2023-09-24
[1D] use midpoint properties for diffusion fluxes

g3bk47:

Evaluates density and mean molecular weight at the grid midpoint location for the expression of diffusion fluxes for the mixture-averaged and unity Lewis number model.

Further discussion in https://github.com/Cantera/enhancements/issues/187.

**Checklist**


test/python/test_onedim.py
```diff
@@ -756,7 +756,7 @@ def test_restore_legacy_hdf(self):
         f = ct.FreeFlame(self.gas)
         meta = f.restore(filename, "group0")
         assert meta['description'] == desc
-        assert meta['cantera_version'] == "2.6.0"
+        #assert meta['cantera_version'] == "2.6.0"
```
g3bk47: I don't quite understand the logic behind these tests. So for now I disabled this version check. Let me know if there is a nicer way to do this.

ischoegl: These are just tests for metadata of the legacy HDF file format written with a known earlier Cantera version. This should not throw unless you changed the reference file (in which case the entire test is pointless). In essence, we continue support for a file format that Cantera can no longer write: the HDF format has been updated since.

ischoegl: To be clear: data that is loaded by this test should not be changed.

test/data/freeflame_legacy.h5
ischoegl, line 1: Please revert change.

test/data/impingingjet_legacy.h5
ischoegl, line 1: Please revert change.

ischoegl (commented):
The two files are explicitly included to ensure that newer Cantera versions still support an earlier version of HDF (Cantera switched from h5py to native HDF support in Cantera 3.0).

test/python/test_onedim.py
ischoegl: I guess the check is working 😅

g3bk47: Yes, I changed the reference file (I replaced it with a newly written one during the test routine). Ok, I will revert this change.

ischoegl: You will likely have to modify the test. But as the file names indicate, these are legacy and should not be  changed. I do understand that the flame speeds will no longer be the same.

g3bk47: I reverted the change for now. I will take a closer look at these two tests in the next few days.

ischoegl (commented):
Thanks! @g3bk47 . But rather than adding two commits that have zero effect, could you just drop the commits altogether, so no unnecessary binary data have to be written/changed in the repo?

g3bk47:
Without touching the legacy reference files, I think relaxing the tolerances for the two legacy tests might be the easiest solution.

ischoegl:
> Without touching the legacy reference files, I think relaxing the tolerances for the two legacy tests might be the easiest solution.

I'm not opposed to this fix to the legacy test, although the tolerances are rather large (it’s still not an issue, as the main concern for these tests is that it remains possible to import legacy files).

Regarding the review itself, I'd like to defer to @speth, as the proposed changes introduce differences of results. The enthalpy plot in https://github.com/Cantera/enhancements/issues/187#issuecomment-1732407678 does let me believe that we should adopt the updates. I thought I recalled seeing some more larger discrepancies reported earlier (I think there were some edits since), but changes are now reasonable (<1%, per current reports in Cantera/enhancements#187). 

Fwiw, it would be interesting to see how changes compare to results from other software, e.g. Chemkin / FlameMaster / zero-RK (I don't have any of these installed myself).

test/python/test_onedim.py
```diff
@@ -759,7 +759,7 @@ def test_restore_legacy_hdf(self):
         assert meta['cantera_version'] == "2.6.0"
 
         # check with reduced tolerance to account for machine dependent differences
-        self.check_save_restore(f)
+        self.check_save_restore(f, tol_T=1e-3, tol_X=1e-1)
```
speth: I think we should update the comment to say that the reduced tolerances are mainly to account for the change in the finite difference discretization between Cantera 2.6 and Cantera 3.1.

test/python/test_onedim.py
```diff
@@ -1573,13 +1573,13 @@ def run_save_restore(self, mode):
 
         self.check_save_restore(jet)
 
-    def check_save_restore(self, jet):
+    def check_save_restore(self, jet, tol_T=1e-3, tol_X=1e-4):
         # pytest.approx is used as equality for floats cannot be guaranteed for loaded
```
speth, line 1581: For both versions of this function, I think you can set the default values of `tol_T` and `tol_X` to `None`, which will use the default values in `pytest.approx`. The only tests that should need the relaxed tolerances are the comparisons with the "legacy" HDF5 format.

speth (changes requested):
I think the main code changes all make sense. I just had a couple of suggestions for the updates to the tests.

On the tests against the legacy HDF5 output files, the relatively large discrepancies aren't that surprising -- these are solutions on a very coarse initial grid, so the effect of the change to the discretization is exaggerated compared to what it would be with any reasonable level of grid refinement.

g3bk47:
> I'm not opposed to this fix to the legacy test, although the tolerances are rather large (it’s still not an issue, as the main concern for these tests is that it remains possible to import legacy files).

I think the very large tolerances (1e-1) are only due to some species mole fractions being very close to zero (e.g. comparing X = 1e-12 to 1e-13). I could play around with both the relative and absolute tolerances, but I wanted to keep the introduced changes a minimal as possible.

> I thought I recalled seeing some more larger discrepancies reported earlier (I think there were some edits since)

Yes, my initial commit only modified the diffusion coefficient of the 0ths species because I forgot to include the loop over all species (see the oldest version of my comment at https://github.com/Cantera/enhancements/issues/187#issuecomment-1732394253). But since this was just a mistake in the code, I edited the comment to include the correct values.

> Fwiw, it would be interesting to see how changes compare to results from other software, e.g. Chemkin / FlameMaster / zero-RK (I don't have any of these installed myself).

I will ask my colleagues to see if anyone has access to Chemkin.

> I think you can set the default values of tol_T and tol_X to None, which will use the default values in pytest.approx

I chose the default values to not change the original behavior of the tests (as these values were hard-coded before). I now changed them to `None` (at least locally for me, all tests still pass).

> I think we should update the comment to say that the reduced tolerances are mainly to account for the change in the finite difference discretization between Cantera 2.6 and Cantera 3.1.

Comments are now included.

Once the enthalpy issue is addressed, I think we should also include a new test that explicitly checks the conservation properties of a unity Lewis, freely propagating flame, i.e.:

```
# species
for e in gas.element_names:
    assertEqual(np.max(f.elemental_mass_fraction(e)), np.min(f.elemental_mass_fraction(e)))

# continuity
assertEqual(np.max(f.density*f.velocity), np.min(f.density*f.velocity))

# energy
assertEqual(np.max(f.enthalpy_mass), np.min(f.enthalpy_mass))
```

Same could be done for a non-unity Lewis number flame by comparing the corresponding first(inlet) and last(outlet) values.

speth (approved):
Thanks, this looks good to me. At least for this PR, I don't think a comparison with other solvers is necessary. To the extent that a comparison would be worthwhile more generally, I think Zero-RK is much more interesting than Chemkin, since in that case the source code is available and there's a chance of understanding any differences.

ischoegl (commented):
Thanks! - :shipit:

PR #1630, opened 2023-10-02
Fix some failing CI jobs

speth:

- Replace Clang 12 with Clang 14, based on updates to versions available on GitHub Actions runners
- Pin the version of MinGW installed by the `setup-mingw` action. This unfortunately slows down these builds a lot. There's already an issue created upstream (https://github.com/egor-tensin/setup-mingw/issues/14). Hopefully we can unpin once this is resolved.

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -866,8 +866,10 @@ elif "clang" in env.subst("$CC"):
     config.select("clang")
 
 else:
-    logger.error(f"Unrecognized C compiler {env['CC']!r}")
-    sys.exit(1)
+    # Assume a GCC compatible compiler if nothing else
+    logger.warning(f"Unrecognized C compiler {env['CC']!r}")
+    config.select("gcc")
+
 
 if env["OS"] == "Windows":
     config.select("Windows")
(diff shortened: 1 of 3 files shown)


SConstruct
```diff
@@ -866,8 +866,9 @@ elif "clang" in env.subst("$CC"):
     config.select("clang")
 
 else:
-    logger.error(f"Unrecognized C compiler {env['CC']!r}")
-    sys.exit(1)
+    # Assume a GCC compatible compiler if nothing else
+    config.select("gcc")
```
ischoegl: Will this make things harder to debug?

speth: Perhaps in some corner case, but I think it's somewhat unlikely. With the exception of MSVC, it seems that most compilers accept the same flags as GCC for the options we want to set. I changed this so there will at least be a warning that GCC variables were assumed.

There's only so we can do with our ad hoc, name-based identification of the compiler (and I'm not advocating making this more complicated), and I'd rather have a likely fallback in place than force the user to edit `SConstruct`. If this is wrong, at least the user can specify all the right flags through other SCons options.

ischoegl: The logger warning is definitely appreciated. Thanks!

ischoegl (approved):

PR #1634, opened 2023-10-17
Extensible jacobian and network jacobian interfaces

anthony-walker:

In this pull request I have setup basic structure for adding contributions to the jacobian from flow devices and walls.
I have also added an extensible jacobian interface to the `ExtensibleReactor` system and modified the existing jacobian system to pass a vector of triplets as is needed in lieu of creating a large number of separate sparse matrices. 

@speth @ischoegl some additional tests are needed for the wall and flow contributions but I wanted to get the content officially into a PR.

- Jacobian contributions from walls and flow devices
- Extensible jacobian interface
- Enhancements to existing jacobian backend

**Checklist**

### DIFF
--- a/include/cantera/zeroD/FlowDevice.h
+++ b/include/cantera/zeroD/FlowDevice.h
@@ -141,40 +141,47 @@ public:
         m_time = time;
     }
 
-    /*! Build the Jacobian terms specific to the flow device for the given connected reactor.
-     * @param r a pointer to the calling reactor
-     * @param jacVector a vector of triplets to be added to the jacobian for the reactor
-     * @warning This function is an experimental part of the %Cantera API and may be changed
-     * or removed without notice.
-     * @since New in %Cantera 3.0.
-     */
+    //! Build the Jacobian terms specific to the flow device for the given connected
+    //! reactor.
+    //! @param r a pointer to the calling reactor
+    //! @param jacVector a vector of triplets to be added to the jacobian for the
+    //! reactor
+    //! @warning This function is an experimental part of the %Cantera API and may be
+    //! changed
+    //! or removed without notice.
+    //! @since New in %Cantera 3.0.
+    //!
     virtual void buildReactorJacobian(ReactorBase* r, vector<Eigen::Triplet<double>>& jacVector) {
         throw NotImplementedError(type() + "::buildReactorJacobian");
     }
 
-    /*! Build the Jacobian terms specific to the flow device for the network. These terms
-     * will be adjusted to the networks indexing system outside of the reactor.
-     * @param jacVector a vector of triplets to be added to the jacobian for the reactor
-     * @warning This function is an experimental part of the %Cantera API and may be changed
-     * or removed without notice.
-     * @since New in %Cantera 3.0.
-     */
+    //! Build the Jacobian terms specific to the flow device for the network. These
+    //! terms
+    //! will be adjusted to the networks indexing system outside of the reactor.
+    //! @param jacVector a vector of triplets to be added to the jacobian for the
+    //! reactor
+    //! @warning This function is an experimental part of the %Cantera API and may be
+    //! changed
+    //! or removed without notice.
+    //! @since New in %Cantera 3.0.
+    //!
     virtual void buildNetworkJacobian(vector<Eigen::Triplet<double>>& jacVector) {
         throw NotImplementedError(type() + "::buildNetworkJacobian");
     }
 
-    /*! Specify the jacobian terms have been calculated and should not be recalculated.
-     * @warning This function is an experimental part of the %Cantera API and may be changed
-     * or removed without notice.
-     * @since New in %Cantera 3.0.
-     */
+    //! Specify the jacobian terms have been calculated and should not be recalculated.
+    //! @warning This function is an experimental part of the %Cantera API and may be
+    //! changed
+    //! or removed without notice.
+    //! @since New in %Cantera 3.0.
+    //!
     void jacobianCalculated() { m_jac_calculated = true; };
 
-    /*! Specify that jacobian terms have not been calculated and should be recalculated.
-     * @warning This function is an experimental part of the %Cantera API and may be changed
-     * or removed without notice.
-     * @since New in %Cantera 3.0.
-     */
+    //! Specify that jacobian terms have not been calculated and should be recalculated.
+    //! @warning This function is an experimental part of the %Cantera API and may be changed
+    //! or removed without notice.
+    //! @since New in %Cantera 3.0.
+    //!
     void jacobianNotCalculated() { m_jac_calculated = false; };
 
 protected:
--- a/include/cantera/zeroD/FlowDevice.h
+++ b/include/cantera/zeroD/FlowDevice.h
@@ -149,9 +149,10 @@ public:
     //! @warning This function is an experimental part of the %Cantera API and may be
     //! changed
     //! or removed without notice.
-    //! @since New in %Cantera 3.0.
+    //! @since New in %Cantera 3.1.
     //!
-    virtual void buildReactorJacobian(ReactorBase* r, vector<Eigen::Triplet<double>>& jacVector) {
+    virtual void buildReactorJacobian(ReactorBase* r,
+        vector<Eigen::Triplet<double>>& jacVector) {
         throw NotImplementedError(type() + "::buildReactorJacobian");
     }
 
@@ -163,28 +164,11 @@ public:
     //! @warning This function is an experimental part of the %Cantera API and may be
     //! changed
     //! or removed without notice.
-    //! @since New in %Cantera 3.0.
+    //! @since New in %Cantera 3.1.
     //!
     virtual void buildNetworkJacobian(vector<Eigen::Triplet<double>>& jacVector) {
-        if (!m_jac_calculated) {
-            throw NotImplementedError(type() + "::buildNetworkJacobian");
-        }
+        throw NotImplementedError(type() + "::buildNetworkJacobian");
     }
-
-    //! Specify the jacobian terms have been calculated and should not be recalculated.
-    //! @warning This function is an experimental part of the %Cantera API and may be
-    //! changed
-    //! or removed without notice.
-    //! @since New in %Cantera 3.0.
-    //!
-    void jacobianCalculated() { m_jac_calculated = true; };
-
-    //! Specify that jacobian terms have not been calculated and should be recalculated.
-    //! @warning This function is an experimental part of the %Cantera API and may be changed
-    //! or removed without notice.
-    //! @since New in %Cantera 3.0.
-    //!
-    void jacobianNotCalculated() { m_jac_calculated = false; };
 
 protected:
     string m_name;  //!< Flow device name.
--- a/include/cantera/zeroD/FlowDevice.h
+++ b/include/cantera/zeroD/FlowDevice.h
@@ -10,6 +10,7 @@
 #include "cantera/base/global.h"
 #include "cantera/base/ctexceptions.h"
 #include "ConnectorNode.h"
+#include "cantera/numerics/eigen_sparse.h"
 
 namespace Cantera
 {
@@ -140,7 +141,49 @@ public:
         m_time = time;
     }
 
+    /*! Build the Jacobian terms specific to the flow device for the given connected reactor.
+     * @param r a pointer to the calling reactor
+     * @param jacVector a vector of triplets to be added to the jacobian for the reactor
+     * @warning This function is an experimental part of the %Cantera API and may be changed
+     * or removed without notice.
+     * @since New in %Cantera 3.0.
+     */
+    virtual void buildReactorJacobian(ReactorBase* r, vector<Eigen::Triplet<double>>& jacVector) {
+        throw NotImplementedError(type() + "::buildReactorJacobian");
+    }
+
+    /*! Build the Jacobian terms specific to the flow device for the network. These terms
+     * will be adjusted to the networks indexing system outside of the reactor.
+     * @param jacVector a vector of triplets to be added to the jacobian for the reactor
+     * @warning This function is an experimental part of the %Cantera API and may be changed
+     * or removed without notice.
+     * @since New in %Cantera 3.0.
+     */
+    virtual void buildNetworkJacobian(vector<Eigen::Triplet<double>>& jacVector) {
+        throw NotImplementedError(type() + "::buildNetworkJacobian");
+    }
+
+    /*! Specify the jacobian terms have been calculated and should not be recalculated.
+     * @warning This function is an experimental part of the %Cantera API and may be changed
+     * or removed without notice.
+     * @since New in %Cantera 3.0.
+     */
+    void jacobianCalculated() { m_jac_calculated = true; };
+
+    /*! Specify that jacobian terms have not been calculated and should be recalculated.
+     * @warning This function is an experimental part of the %Cantera API and may be changed
+     * or removed without notice.
+     * @since New in %Cantera 3.0.
+     */
+    void jacobianNotCalculated() { m_jac_calculated = false; };
+
 protected:
+    string m_name;  //!< Flow device name.
+    bool m_defaultNameSet = false;  //!< `true` if default name has been previously set.
+    //! a variable to switch on and off so calculations are not doubled by the calling
+    //! reactor or network
+    bool m_jac_calculated = false;
+
     double m_mdot = Undef;
 
     //! Function set by setPressureFunction; used by updateMassFlowRate
--- a/include/cantera/zeroD/FlowDevice.h
+++ b/include/cantera/zeroD/FlowDevice.h
@@ -166,7 +166,9 @@ public:
     //! @since New in %Cantera 3.0.
     //!
     virtual void buildNetworkJacobian(vector<Eigen::Triplet<double>>& jacVector) {
-        throw NotImplementedError(type() + "::buildNetworkJacobian");
+        if (!m_jac_calculated) {
+            throw NotImplementedError(type() + "::buildNetworkJacobian");
+        }
     }
 
     //! Specify the jacobian terms have been calculated and should not be recalculated.
--- a/include/cantera/zeroD/Reactor.h
+++ b/include/cantera/zeroD/Reactor.h
@@ -231,7 +231,27 @@ public:
         throw NotImplementedError(type() + "::buildJacobian");
     }
 
-    // virtual void jacobian()
+    //! Calculate the Jacobian of a Reactor specialization for wall contributions.
+    //! @param jacVector vector where jacobian triplets are added
+    //! @warning Depending on the particular implementation, this may return an
+    //! approximate Jacobian intended only for use in forming a preconditioner for
+    //! iterative solvers.
+    //! @ingroup derivGroup
+    //!
+    //! @warning  This method is an experimental part of the %Cantera
+    //! API and may be changed or removed without notice.
+    virtual void buildWallJacobian(vector<Eigen::Triplet<double>>& jacVector);
+
+    //! Calculate flow contributions to the Jacobian of a Reactor specialization.
+    //! @param jac_vector vector where jacobian triplets are added
+    //! @warning Depending on the particular implementation, this may return an
+    //! approximate Jacobian intended only for use in forming a preconditioner for
+    //! iterative solvers.
+    //! @ingroup derivGroup
+    //!
+    //! @warning  This method is an experimental part of the %Cantera
+    //! API and may be changed or removed without notice.
+    virtual void buildFlowJacobian(vector<Eigen::Triplet<double>>& jacVector);
 
     //! Calculate the reactor-specific Jacobian using a finite difference method.
     //!
@@ -320,6 +340,10 @@ protected:
 
     //! Vector of triplets representing the jacobian
     vector<Eigen::Triplet<double>> m_jac_trips;
+    //! Boolean to skip walls in jacobian
+    bool m_jac_skip_walls = false;
+    //! Boolean to skip flow devices in jacobian
+    bool m_jac_skip_flow_devices = false;
 };
 }
 
--- a/include/cantera/zeroD/Reactor.h
+++ b/include/cantera/zeroD/Reactor.h
@@ -218,8 +218,7 @@ public:
     }
 
     //! Calculate the Jacobian of a specific Reactor specialization.
-    //! @param jac_vector vector where jacobian triplets are added
-    //! @param offset offset added to the row and col indices of the elements
+    //! @param jacVector vector where jacobian triplets are added
     //! @warning Depending on the particular implementation, this may return an
     //! approximate Jacobian intended only for use in forming a preconditioner for
     //! iterative solvers.
@@ -243,7 +242,7 @@ public:
     virtual void buildWallJacobian(vector<Eigen::Triplet<double>>& jacVector);
 
     //! Calculate flow contributions to the Jacobian of a Reactor specialization.
-    //! @param jac_vector vector where jacobian triplets are added
+    //! @param jacVector vector where jacobian triplets are added
     //! @warning Depending on the particular implementation, this may return an
     //! approximate Jacobian intended only for use in forming a preconditioner for
     //! iterative solvers.
--- a/include/cantera/zeroD/Reactor.h
+++ b/include/cantera/zeroD/Reactor.h
@@ -199,7 +199,7 @@ public:
     //! @param limit value for step size limit
     void setAdvanceLimit(const string& nm, const double limit);
 
-    //! Calculate the Jacobian of a specific Reactor specialization.
+    //! A wrapper for the Jacobian function to return the Eigen::SparseMatrix<double>
     //! @warning Depending on the particular implementation, this may return an
     //! approximate Jacobian intended only for use in forming a preconditioner for
     //! iterative solvers.
@@ -208,8 +208,30 @@ public:
     //! @warning  This method is an experimental part of the %Cantera
     //! API and may be changed or removed without notice.
     virtual Eigen::SparseMatrix<double> jacobian() {
-        throw NotImplementedError("Reactor::jacobian");
+        m_jac_trips.clear();
+        // Add before, during, after evals
+        buildJacobian(m_jac_trips);
+        // construct jacobian from vector
+        Eigen::SparseMatrix<double> jac(m_nv, m_nv);
+        jac.setFromTriplets(m_jac_trips.begin(), m_jac_trips.end());
+        return jac;
     }
+
+    //! Calculate the Jacobian of a specific Reactor specialization.
+    //! @param jac_vector vector where jacobian triplets are added
+    //! @param offset offset added to the row and col indices of the elements
+    //! @warning Depending on the particular implementation, this may return an
+    //! approximate Jacobian intended only for use in forming a preconditioner for
+    //! iterative solvers.
+    //! @ingroup derivGroup
+    //!
+    //! @warning  This method is an experimental part of the %Cantera
+    //! API and may be changed or removed without notice.
+    virtual void buildJacobian(vector<Eigen::Triplet<double>>& jacVector) {
+        throw NotImplementedError(type() + "::buildJacobian");
+    }
+
+    // virtual void jacobian()
 
     //! Calculate the reactor-specific Jacobian using a finite difference method.
     //!
--- a/include/cantera/zeroD/Reactor.h
+++ b/include/cantera/zeroD/Reactor.h
@@ -207,15 +207,7 @@ public:
     //!
     //! @warning  This method is an experimental part of the %Cantera
     //! API and may be changed or removed without notice.
-    virtual Eigen::SparseMatrix<double> jacobian() {
-        m_jac_trips.clear();
-        // Add before, during, after evals
-        buildJacobian(m_jac_trips);
-        // construct jacobian from vector
-        Eigen::SparseMatrix<double> jac(m_nv, m_nv);
-        jac.setFromTriplets(m_jac_trips.begin(), m_jac_trips.end());
-        return jac;
-    }
+    virtual Eigen::SparseMatrix<double> jacobian();
 
     //! Calculate the Jacobian of a specific Reactor specialization.
     //! @param jacVector vector where jacobian triplets are added
--- a/include/cantera/zeroD/ReactorBase.h
+++ b/include/cantera/zeroD/ReactorBase.h
@@ -302,11 +302,69 @@ public:
         return m_sensParams.size();
     }
 
+    /*! Calculate the derivative of temperature with respect to the temperature in the
+     * heat transfer equation based on the reactor specific equation of state.
+     * This function should also transform the state of the derivative to that
+     * appropriate for the jacobian's state/
+     * @warning This function is an experimental part of the %Cantera API and may be changed
+     * or removed without notice.
+     * @since New in %Cantera 3.0.
+     */
+    virtual double temperatureDerivative() {
+        throw NotImplementedError("Reactor::temperatureDerivative");
+    }
+
+    /*! Calculate the derivative of temperature with respect to the temperature in the
+     * heat transfer radiation equation based on the reactor specific equation of state.
+     * This function should also transform the state of the derivative to that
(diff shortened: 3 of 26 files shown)


anthony-walker:
@speth @ischoegl I am not entirely sure why the `.NET` tests are failing but it doesn't seem to be with this PR, is this a known issue?

Otherwise I believe this code is ready for review.

ischoegl (commented):
While I don't have enough experience in extensible things for a complete review, I noticed that a lot of test are failing with:
```
NotImplementedError thrown by MassFlowController::buildReactorJacobian:
Not implemented.
```
which doesn't appear to be .NET related.

anthony-walker:
@ischoegl these failures are from my latest bug fixes there is a separate commit which addresses the `.NET` failure issue, [0a34092](https://github.com/Cantera/cantera/pull/1634/commits/0a34092e04f6ad9f90294a4e2830e4d62860e950).

include/cantera/zeroD/FlowDevice.h
```diff
@@ -114,7 +115,56 @@ class FlowDevice
...
 
+    //! Build the Jacobian terms specific to the flow device for the given connected
+    //! reactor.
+    //! @param r a pointer to the calling reactor
+    //! @param jacVector a vector of triplets to be added to the jacobian for the
+    //! reactor
+    //! @warning This function is an experimental part of the %Cantera API and may be
+    //! changed
+    //! or removed without notice.
+    //! @since New in %Cantera 3.0.
```
speth: Here and elsewhere:
```suggestion
    //! @since New in %Cantera 3.1.
```

speth:
Before you get any further, I just wanted to suggest undoing the merge commit and rebasing onto the current `main` branch instead.

include/cantera/zeroD/ReactorBase.h
```diff
@@ -283,10 +283,44 @@ class ReactorBase
...
+    //! Calculate the derivative of T with respect to the ith species in the heat
+    //! transfer equation based on the reactor specific equation of state.
+    //! @param index index of the species the derivative is with respect too
+    //! @warning This function is an experimental part of the %Cantera API and may be changed
+    //! or removed without notice.
+    //! @since New in %Cantera 3.0.
+    //!
+    virtual double moleDerivative(size_t index) {
+        throw NotImplementedError("Reactor::moleDerivative");
+    }
```
speth: I find the name for this function fairly confusing -- Nothing in the name indicates that this is the derivative of the temperature. For consistency with how we named the derivative methods in `Kinetics`, I'd recommend `temperature_ddni`.

Regarding the docstring, nothing in this method is specific to the "heat transfer equation".

include/cantera/zeroD/ReactorBase.h
```diff
@@ -283,10 +283,44 @@ class ReactorBase
...
+    //! Calculate the derivative of T with respect to the ith species in the heat
+    //! transfer radiation equation based on the reactor specific equation of state.
+    //! @param index index of the species the derivative is with respect too
+    //! @warning This function is an experimental part of the %Cantera API and may be changed
+    //! or removed without notice.
+    //! @since New in %Cantera 3.0.
+    //!
+    virtual double moleRadiationDerivative(size_t index) {
+        throw NotImplementedError("Reactor::moleRadiationDerivative");
+    }
```
speth: I don't think this method should be necessary. You can compute $d(T^4)/dn_i$ using the chain rule as $4T^3 dT/dn_i$.

src/zeroD/IdealGasMoleReactor.cpp
```diff
@@ -242,14 +241,35 @@ Eigen::SparseMatrix<double> IdealGasMoleReactor::jacobian()
...
+    buildFlowJacobian(jacVector);
+}
+
+double IdealGasMoleReactor::moleDerivative(size_t index)
+{
+    // derivative of temperature transformed by ideal gas law
+    vector<double> moles(m_nsp);
+    getMoles(moles.data());
+    double dTdni = pressure() * m_vol / GasConstant / std::accumulate(moles.begin(), moles.end(), 0.0);
+    return dTdni;
```
speth: This is a lot of repeated memory allocations and repeated calculations when all you really want is the total number of moles. The simpler way to calculate this is just
`n_total = m_mass / m_thermo->meanMolecularWeight()`.

src/zeroD/IdealGasMoleReactor.cpp
```diff
@@ -242,14 +241,35 @@ Eigen::SparseMatrix<double> IdealGasMoleReactor::jacobian()
...
+        // build wall jacobian
+        buildWallJacobian(jacVector);
     }
-    // convert triplets to sparse matrix
-    Eigen::SparseMatrix<double> jac(m_nv, m_nv);
-    jac.setFromTriplets(m_jac_trips.begin(), m_jac_trips.end());
-    return jac;
+
+    // build flow jacobian
+    buildFlowJacobian(jacVector);
```
speth: The comments here add nothing

src/zeroD/IdealGasConstPressureMoleReactor.cpp
```diff
@@ -271,4 +273,23 @@ string IdealGasConstPressureMoleReactor::componentName(size_t k) {
                        "Index is out of bounds.");
 }
 
+double IdealGasConstPressureMoleReactor::moleDerivative(size_t index)
+{
+    // derivative of temperature transformed by ideal gas law
+    vector<double> moles(m_nsp);
+    getMoles(moles.data());
+    double dTdni = pressure() * m_vol / GasConstant / std::accumulate(moles.begin(), moles.end(), 0.0);
+    return dTdni;
```
speth: Likewise here, the calculation of the total number of moles can be done more efficiently.

test/python/conftest.py
```diff
@@ -1,3 +1,26 @@
...
+
+def pytest_collection_modifyitems(config, items):
+    # run only diagnostic tests
+    if config.getoption("--diagnose"):
+        reason = "running only diagnostic tests."
+        skip_diagnose = pytest.mark.skip(reason=reason)
+        for item in items:
+            if "diagnose" not in item.keywords:
+                item.add_marker(skip_diagnose)
+        return
```
speth, line 86: What does this do? How is this used? If this is adding new ways of selectively running tests, I think it should be documented in the developer docs (a page under `doc/sphinx/develop`)

src/zeroD/Reactor.cpp
```diff
@@ -37,6 +37,17 @@ void Reactor::setDerivativeSettings(AnyMap& settings)
...
+    // set reactor settings
+    bool force = settings.empty();
+    if (force || settings.hasKey("skip-walls")) {
+        m_jac_skip_walls = settings.getBool("skip-walls",
+            false);
+    }
+    if (force || settings.hasKey("skip-flow-devices")) {
+        m_jac_skip_flow_devices = settings.getBool("skip-flow-devices",
+            false);
+    }
```
speth, line 55: Minor, but the line wrapping here is unnecessary.

As a broader question, though, why would you want to skip these Jacobian elements? Is there a case-dependent performance implication?

src/zeroD/Reactor.cpp
```diff
@@ -592,4 +603,26 @@ void Reactor::setAdvanceLimit(const string& nm, const double limit)
     }
 }
 
+void Reactor:: buildWallJacobian(vector<Eigen::Triplet<double>>& jacVector)
```
speth: Minor formatting error:
```suggestion
void Reactor::buildWallJacobian(vector<Eigen::Triplet<double>>& jacVector)
```

src/zeroD/ReactorNet.cpp
```diff
@@ -580,4 +587,117 @@ void ReactorNet::checkPreconditionerSupported() const {
...
+            for (size_t i = 0; i < r->nWalls(); i++) {
+                r->wall(i).jacobianNotCalculated();
+            }
+        }
+        // flow devices
+        if (!m_jac_skip_flow_devices) {
+            // outlets
+            for (size_t i = 0; i < r->nOutlets(); i++) {
+                r->outlet(i).jacobianNotCalculated();
+            }
```
speth: Rather than having to toggle this flag on every wall and flow device each time the Jacobian is evaluated, I think it would make more sense to have the `ReactorNet` store a set of `Wall*` and `FlowDevice*` as part of `initialize()`, where the `set` will automatically take care of deduplication.

src/zeroD/ReactorNet.cpp
```diff
@@ -539,19 +549,16 @@ void ReactorNet::preconditionerSetup(double t, double* y, double gamma)
...
-                    it.value());
-            }
-        }
+    // Create jacobian triplet vector
+    vector<Eigen::Triplet<double>> jacVector;
+    buildJacobian(jacVector);
+    // Add to preconditioner with offset
+    for (auto it : jacVector) {
+        precon->setValue(it.row(), it.col(), it.value());
     }
```
speth, line 821: The inefficiency of having to copy the Jacobian elements here is a bit unfortunate. The best alternative I can think of is that the preconditioner should to provide access to its triplet vector, so you can pass that into `buildJacobian` directly rather than needing the temporary `jacVector` here.

src/zeroD/ReactorNet.cpp
```diff
@@ -580,4 +587,117 @@ void ReactorNet::checkPreconditionerSupported() const {
...
 }
 
+void ReactorNet::buildJacobian(vector<Eigen::Triplet<double>>& jacVector)
+{
+    // network must be initialized for the jacobian
+    if (! m_init) {
```
speth:
```suggestion
    if (!m_init) {
```

src/zeroD/ReactorNet.cpp
```diff
@@ -580,4 +587,117 @@ void ReactorNet::checkPreconditionerSupported() const {
...
+
+Eigen::SparseMatrix<double> ReactorNet::finiteDifferenceJacobian()
+{
+    // network must be initialized for the jacobian
+    if (! m_init) {
+        initialize();
+    }
+
+    // clear former jacobian elements
+    vector<Eigen::Triplet<double>> jac_trips;
```
speth: Comment seems incorrect.

src/zeroD/Wall.cpp
```diff
@@ -111,4 +113,66 @@ double Wall::heatRate()
...
+
+void Wall::buildReactorJacobian(ReactorBase* r, vector<Eigen::Triplet<double>>& jacVector)
+{
+    // get derivative of heat transfer for both reactors
+    vector<Eigen::Triplet<double>> network;
+    size_t nsp = r->contents().nSpecies();
+    size_t sidx = r->speciesOffset();
+    size_t eidx = r->energyIndex();
+    // define a scalar for direction based on left and right
+    double scalar = (r == m_left) ? 1.0 : -1.0;
```
speth: I think `direction` would be a better name -- this is more like a unit vector than a scalar.

src/zeroD/Wall.cpp
```diff
@@ -111,4 +113,66 @@ double Wall::heatRate()
...
+    // elements within the current reactor
+    // find dQdni for the current reactor w.r.t current reactor
+    for (size_t i = sidx; i < nsp + sidx; i++) {
+        double dQdni = m_rrth * m_area * scalar * r->moleDerivative(i);
+        dQdni += m_emiss * m_area * scalar * r->moleRadiationDerivative(i);
+        jacVector.emplace_back(eidx, i, dQdni);
```
speth: The Jacobian element is equal to $dQ/dn_i$ only if $U$ or $H$ is the state variable for the energy equation. However, for both `IdealGasMoleReactor` and `IdealGasConstPressureMoleReactor`, the energy equation is in terms of temperature.

src/zeroD/Wall.cpp
```diff
@@ -111,4 +113,66 @@ double Wall::heatRate()
...
+    // return if the jacobian has been calculated
+    if (m_jac_calculated) {
+        return;
+    }
+    // get derivatives for inter-dependent reactor terms
+    //variables for the right side
+    vector<Eigen::Triplet<double>> network;
+    size_t r_nsp = m_right->contents().nSpecies();
+    size_t r_sidx = m_right->speciesOffset();
+    size_t r_net = m_right->network().globalStartIndex((Reactor* ) m_right);
```
speth: To avoid the need for the cast here, I think we should modify `ReactorNet::globaStartIndex` to take a `ReactorBase*`. It will raise an exception if a `Reservoir` is provided as the input here, which is fine.

src/zeroD/Wall.cpp
```diff
@@ -111,4 +113,66 @@ double Wall::heatRate()
...
+    size_t r_eidx = m_right->energyIndex();
+
+    // variables for the left side
+    size_t l_nsp = m_left->contents().nSpecies();
+    size_t l_sidx = m_left->speciesOffset();
+    size_t l_net = m_left->network().globalStartIndex((Reactor* ) m_left);
+    size_t l_eidx = m_left->energyIndex();
+
+    if (((Reactor* ) m_right)->energyEnabled()) {
```
speth: To avoid the need for this cast, let's move `energyEnabled` up to the `ReactorBase` class as a virtual method.

src/zeroD/Wall.cpp
```diff
@@ -111,4 +113,66 @@ double Wall::heatRate()
...
+        // find dQdni for the right reactor w.r.t left reactor
+        for (size_t i = l_sidx; i < l_sidx + l_nsp; i++) {
+            double dQdni = m_rrth * m_area * m_left->moleDerivative(i);
+            dQdni += m_emiss * m_area * m_left->moleRadiationDerivative(i);
+            jacVector.emplace_back(r_eidx + r_net, i + l_net, dQdni);
```
speth: Again, this derivative is not correct for the reactors that use T as the state variable.

src/zeroD/Wall.cpp
```diff
@@ -111,4 +113,66 @@ double Wall::heatRate()
...
+        }
+    }
+
+    if (((Reactor* ) m_left)->energyEnabled()) {
+        // find dQdni for the left reactor w.r.t right reactor
+        for (size_t i = r_sidx; i < r_sidx + r_nsp; i++) {
+            double dQdni = - m_rrth * m_area * m_right->moleDerivative(i);
+            dQdni -= m_emiss * m_area * m_right->moleRadiationDerivative(i);
+            jacVector.emplace_back(l_eidx + l_net, i + r_net, dQdni);
```
speth: Besides the energy equation, `Wall` objects can also affect the volume equation for the reactor, if an expansion coefficient has been set, and those terms also need to be added to the Jacobian.

include/cantera/zeroD/Wall.h
```diff
@@ -115,6 +116,50 @@ class WallBase
         m_time = time;
     }
 
+    //! Build the Jacobian terms specific to the flow device for the given connected
+    //! reactor.
+    //! @param r a pointer to the calling reactor
+    //! @param jacVector a vector of triplets to be added to the jacobian for the
```
speth: "Jacobian" should always be capitalized.

include/cantera/zeroD/Wall.h
```diff
@@ -115,6 +116,50 @@ class WallBase
...
+    //! reactor
+    //! @warning This function is an experimental part of the %Cantera API and may be
+    //! changed
+    //! or removed without notice.
```
speth: The line wrapping here is kind of odd. Please fix it here and in the other instances of the same.

include/cantera/zeroD/Wall.h
```diff
@@ -115,6 +116,50 @@ class WallBase
...
+    //! @since New in %Cantera 3.0.
+    //!
+    virtual void buildReactorJacobian(ReactorBase* r, vector<Eigen::Triplet<double>>& jacVector) {
+        throw NotImplementedError("WallBase::buildReactorJacobian");
+    }
```
speth: Please wrap lines longer than 88 characters, here and elsewhere.

include/cantera/zeroD/Wall.h
```diff
@@ -115,6 +116,50 @@ class WallBase
...
+
+    //! Build the Jacobian terms specific to the flow device for the network. These
+    //! terms
+    //! will be adjusted to the networks indexing system outside of the reactor.
```
speth: This is incorrect, isn't it? The determination of the global component indices is done in this method. Only `buildReactorJacobian` uses local component indices. Also, I think it needs to be more clear that this method is only computing the cross-reactor terms.

include/cantera/zeroD/Wall.h
```diff
@@ -115,6 +116,50 @@ class WallBase
...
+    //! @param jacVector a vector of triplets to be added to the jacobian for the
+    //! reactor
```
speth: What reactor?

include/cantera/zeroD/Wall.h
```diff
@@ -255,6 +304,10 @@ class Wall : public WallBase
         return m_k;
     }
 
+    virtual void buildReactorJacobian(ReactorBase* r, vector<Eigen::Triplet<double>>& jacVector) override;
+
+    virtual void buildNetworkJacobian(vector<Eigen::Triplet<double>>& jacVector) override;
+
```
speth: You don't need to specify `virtual` on methods that are marked as `override`.

include/cantera/zeroD/Reactor.h
```diff
@@ -212,9 +212,50 @@ class Reactor : public ReactorBase
...
     virtual Eigen::SparseMatrix<double> jacobian() {
-        throw NotImplementedError("Reactor::jacobian");
+        m_jac_trips.clear();
+        // Add before, during, after evals
+        buildJacobian(m_jac_trips);
+        // construct jacobian from vector
+        Eigen::SparseMatrix<double> jac(m_nv, m_nv);
+        jac.setFromTriplets(m_jac_trips.begin(), m_jac_trips.end());
+        return jac;
+    }
```
speth: The body of this method is too much for a header file, and should be moved to the `.cpp` file.

include/cantera/zeroD/Reactor.h
```diff
@@ -212,9 +212,50 @@ class Reactor : public ReactorBase
...
+    //! Calculate flow contributions to the Jacobian of a Reactor specialization.
+    //! @param jacVector vector where jacobian triplets are added
+    //! @warning Depending on the particular implementation, this may return an
+    //! approximate Jacobian intended only for use in forming a preconditioner for
+    //! iterative solvers.
+    //! @ingroup derivGroup
+    //!
+    //! @warning  This method is an experimental part of the %Cantera
+    //! API and may be changed or removed without notice.
+    virtual void buildFlowJacobian(vector<Eigen::Triplet<double>>& jacVector);
```
speth, line 245: Given the need for `Wall::buildNetworkJacobian` and `FlowDevice::buildNetworkJacobian` to handle cross-reactor terms, I'm wondering if these methods have much value. At the reactor network level, you already have to (effectively) loop over each `Wall` and `FlowDevice` to get their contributions, so you might as well compute the "local" terms there as well.

include/cantera/zeroD/FlowDevice.h
```diff
@@ -114,7 +115,56 @@ class FlowDevice
...
+    //!
+    virtual void buildNetworkJacobian(vector<Eigen::Triplet<double>>& jacVector) {
+        if (!m_jac_calculated) {
+            throw NotImplementedError(type() + "::buildNetworkJacobian");
+        }
+    }
```
speth, line 171: Is there currently no implementation of this method for any flow device?

include/cantera/zeroD/ReactorNet.h
```diff
@@ -236,6 +236,20 @@ class ReactorNet : public FuncEval
...
+    //! reactor network.
+    size_t globalStartIndex(Reactor* curr_reactor) {
+        for (size_t i = 0; i < m_reactors.size(); i++) {
+            if (curr_reactor == m_reactors[i]) {
+                return m_start[i];
+            }
+        }
+        throw CanteraError("ReactorNet::globalStartIndex: ",
+                curr_reactor->name(), " not found in network.");
+        return npos;
```
speth: The `return` statement after `throw` is unnecessary.

This method is too long to for the header file -- basically anything that isn't a one-liner should be in the `.cpp` file.

interfaces/cython/cantera/_utils.pyx
```diff
@@ -519,6 +520,10 @@ cdef vector[vector[string]] list2_string_to_anyvalue(data):
             v[i][j] = stringify(jtem)
     return v
 
+cdef CxxEigenTriplet get_triplet(row, col, val):
+    cdef CxxEigenTriplet* trip_ptr = new CxxEigenTriplet(row, col, val)
+    return dereference(trip_ptr)
```
speth, line 533: This looks like a serious memory leak -- every call allocates a new `trip_ptr` that will never be deleted.

interfaces/cython/cantera/delegator.pyx
```diff
@@ -305,7 +319,6 @@ cdef int assign_delegates(obj, CxxDelegator* delegator) except -1:
 
         if when is None:
             continue
-
         cxx_name = stringify(options[0])
```
speth, line 323: Blank line here shouldn't have been deleted.

src/zeroD/IdealGasMoleReactor.cpp
```diff
@@ -242,14 +241,35 @@ Eigen::SparseMatrix<double> IdealGasMoleReactor::jacobian()
...
+}
+
+double IdealGasMoleReactor::moleDerivative(size_t index)
+{
+    // derivative of temperature transformed by ideal gas law
+    vector<double> moles(m_nsp);
+    getMoles(moles.data());
+    double dTdni = pressure() * m_vol / GasConstant / std::accumulate(moles.begin(), moles.end(), 0.0);
```
speth: I think there's a significant conceptual problem here: This is calculating the derivative of $T$ with respect to $n_i$ assuming that $p$ is held constant. However, for this reactor type, $T$ is one of the independent variables -- a change in $n_i$ will not change $T$, but will instead result in a change in $p$.

src/zeroD/IdealGasConstPressureMoleReactor.cpp
```diff
@@ -271,4 +273,23 @@ string IdealGasConstPressureMoleReactor::componentName(size_t k) {
                        "Index is out of bounds.");
 }
 
+double IdealGasConstPressureMoleReactor::moleDerivative(size_t index)
+{
+    // derivative of temperature transformed by ideal gas law
+    vector<double> moles(m_nsp);
+    getMoles(moles.data());
+    double dTdni = pressure() * m_vol / GasConstant / std::accumulate(moles.begin(), moles.end(), 0.0);
```
speth: Like in `IdealGasMoleReactor`, $T$ is an independent variable. Here, a change in $n_i$ will change the volume, not the temperature.

test/python/test_reactor.py
```diff
@@ -1378,19 +1433,88 @@ def create_reactors(self, **kwargs):
...
+        net = ct.ReactorNet([r1, r2])
+        jac = net.jacobian
+
+        # check for values
+        for i in range(gas1.n_species):
+            assert np.isclose(jac[0, i + 1 + r1.n_vars], - U * A * gas2.T)
+            assert np.isclose(jac[r1.n_vars, i + 1], U * A * gas1.T)
+            if (i + 1 != 2):
+                assert np.isclose(jac[0, i + 1], U * A * gas1.T, rtol=1e-2)
+                assert np.isclose(jac[r1.n_vars, i + 1 + r1.n_vars], - U * A * gas2.T, rtol=1e-2)
```
speth: I think you should check against the finite difference Jacobian. This test is just checking against the same incorrect formulas in Python as what's in the C++ implementation.

test/python/test_reactor.py
```diff
@@ -1378,19 +1433,88 @@ def create_reactors(self, **kwargs):
...
+        gas1 = ct.Solution("h2o2.yaml", "ohmech")
+        gas1.TPX = 600, ct.one_atm, "O2:1.0"
+        r1 = self.reactorClass(gas1)
+        r1.chemistry_enabled = False
+
+        # create second reactor
+        gas2 = ct.Solution("h2o2.yaml", "ohmech")
+        gas2.TPX = 300, ct.one_atm, "O2:1.0"
+        r2 = self.reactorClass(gas2)
+        r2.chemistry_enabled = False
```
speth: The analytical Jacobian doesn't currently respect the `chemistry_enabled` flag, so there are a lot of non-zero elements here that should be zero based on the setting of this flag.

test/zeroD/test_zeroD.cpp
```diff
@@ -212,6 +212,124 @@ TEST(AdaptivePreconditionerTests, test_precon_solver_stats)
...
+    // manually get wall jacobian elements
+    vector<Eigen::Triplet<double>>  wallJac;
+    // build jac for reactor 1 wall only
+    w.buildReactorJacobian(&reactor1, wallJac);
+    wallJacMat.setFromTriplets(wallJac.begin(), wallJac.end());
+    // check appropriate values
+    // double tol = 1e-8;
+    for (int k = 0; k < wallJacMat.outerSize(); k++) {
+        for (Eigen::SparseMatrix<double>::InnerIterator it(wallJacMat, k); it; ++it) {
+            EXPECT_DOUBLE_EQ(it.value(), 6000.0);
```
speth: What are "appropriate" values? Why should these elements equal 6000?

test/zeroD/test_zeroD.cpp
```diff
@@ -212,6 +212,124 @@ TEST(AdaptivePreconditionerTests, test_precon_solver_stats)
...
+                EXPECT_DOUBLE_EQ(it.value(), 6000.0);
+                EXPECT_EQ(it.row(), reactor1.neq()); // check that it is the first row
+                EXPECT_GE(it.col(), reactor2.speciesOffset());
+                EXPECT_LT(it.col(), reactor1.neq());
+            }
+        }
+    }
+}
+
+TEST(JacobianTests, test_flow_jacobian_build)
```
speth: I think this needs to be more clear that the flow device Jacobian elements are currently not implemented.

speth (changes requested):
Thanks for putting this together, @anthony-walker. I think it helps in laying out the necessary pieces for handling the cross-reactor terms in the Jacobian -- in particular, the mechanics surrounding `Wall::buildNetworkJacobian` seem like a good basis to keep building on.

I think there is some room for simplification of this structure, which I've left specific comments on. However, I think there is a significant issue with sorting out exactly which derivatives need to be computed and what the right formulas are for the different cases. The challenge I see us needing to resolve to generalize this across the different reactor types is that what derivative is needed for a single interaction (for example, wall heat transfer) depends on *both* reactor types and what variables they use to represent each term, and we need a way of selecting the correct formula for each Jacobian element based on that pair of reactor types.

speth:
For an example of where the current derivatives are clearly not correct, consider the following example, adapted from one of your new tests:
```py
import cantera as ct
import numpy as np

# create first reactor
gas1 = ct.Solution("h2o2.yaml", "ohmech")
gas1.TPX = 600, ct.one_atm, "O2:1.0"
r1 = ct.IdealGasMoleReactor(gas1)
r1.name = 'R1'
gas1.set_multiplier(0.0)

# create second reactor
gas2 = ct.Solution("h2o2.yaml", "ohmech")
gas2.TPX = 300, ct.one_atm, "O2:1.0"
gas2.set_multiplier(0.0)
r2 = ct.IdealGasMoleReactor(gas2)
r2.name = 'R2'

# create wall
U = 500.0
A = 3.0
w = ct.Wall(r1, r2, U=U, A=A)
net = ct.ReactorNet([r1, r2])

def print_jac(J):
    for i,j in zip(*np.nonzero(J)):
        name_i = net.component_name(i).replace('temperature', 'T')
        name_j = net.component_name(j).replace('temperature', 'T')
        net.component_name(i).replace('temperature', 'T')
        #print(f"J[{i: 3d}, {j: 3d}] = {J[i,j]:.3e}")
        print(f"J[{name_i:8s}, {name_j:8s}] = {J[i,j]:.3e}")

print('Semi-analytical Jacobian:')
print_jac(net.jacobian)
print('\nFinite difference Jacobian:')
print_jac(net.finite_difference_jacobian)
```
which outputs:
```
Semi-analytical Jacobian:
J[R1: T   , R1: T   ] = -2.727e+00
J[R1: T   , R1: H2  ] = 9.000e+05
J[R1: T   , R1: H   ] = 9.000e+05
J[R1: T   , R1: O   ] = 9.000e+05
J[R1: T   , R1: O2  ] = 9.000e+05
J[R1: T   , R1: OH  ] = 9.000e+05
J[R1: T   , R1: H2O ] = 9.000e+05
J[R1: T   , R1: HO2 ] = 9.000e+05
J[R1: T   , R1: H2O2] = 9.000e+05
J[R1: T   , R1: AR  ] = 9.000e+05
J[R1: T   , R1: N2  ] = 9.000e+05
J[R1: T   , R2: H2  ] = -4.500e+05
J[R1: T   , R2: H   ] = -4.500e+05
J[R1: T   , R2: O   ] = -4.500e+05
J[R1: T   , R2: O2  ] = -4.500e+05
J[R1: T   , R2: OH  ] = -4.500e+05
J[R1: T   , R2: H2O ] = -4.500e+05
J[R1: T   , R2: HO2 ] = -4.500e+05
J[R1: T   , R2: H2O2] = -4.500e+05
J[R1: T   , R2: AR  ] = -4.500e+05
J[R1: T   , R2: N2  ] = -4.500e+05
J[R2: T   , R1: H2  ] = 9.000e+05
J[R2: T   , R1: H   ] = 9.000e+05
J[R2: T   , R1: O   ] = 9.000e+05
J[R2: T   , R1: O2  ] = 9.000e+05
J[R2: T   , R1: OH  ] = 9.000e+05
J[R2: T   , R1: H2O ] = 9.000e+05
J[R2: T   , R1: HO2 ] = 9.000e+05
J[R2: T   , R1: H2O2] = 9.000e+05
J[R2: T   , R1: AR  ] = 9.000e+05
J[R2: T   , R1: N2  ] = 9.000e+05
J[R2: T   , R2: T   ] = -1.887e+00
J[R2: T   , R2: H2  ] = -4.500e+05
J[R2: T   , R2: H   ] = -4.500e+05
J[R2: T   , R2: O   ] = -4.500e+05
J[R2: T   , R2: O2  ] = -4.500e+05
J[R2: T   , R2: OH  ] = -4.500e+05
J[R2: T   , R2: H2O ] = -4.500e+05
J[R2: T   , R2: HO2 ] = -4.500e+05
J[R2: T   , R2: H2O2] = -4.500e+05
J[R2: T   , R2: AR  ] = -4.500e+05
J[R2: T   , R2: N2  ] = -4.500e+05

Finite difference Jacobian:
J[R1: T   , R1: T   ] = -2.727e+00
J[R1: T   , R1: O2  ] = 4.589e+04
J[R1: T   , R2: T   ] = 3.107e+00
J[R2: T   , R1: T   ] = 1.752e+00
J[R2: T   , R2: T   ] = -1.887e+00
J[R2: T   , R2: O2  ] = -1.294e+04
```

test/python/conftest.py
anthony-walker: This is to isolate python tests. I will add to docs.

src/zeroD/Wall.cpp
anthony-walker: @speth I think it would make sense to move chemistryEnabled in the same way for consistency.

anthony-walker:
@speth apologies for the delay, I have made some updates but still need to do some of the larger changes. It is still on my radar.

src/zeroD/Reactor.cpp
anthony-walker: That is a good question, I believe depending on the flow device there could be performance implications.

include/cantera/zeroD/Reactor.h
anthony-walker: My thought was when an individual wants to access a reactor only Jacobian.

include/cantera/zeroD/FlowDevice.h
anthony-walker: There is not one currently.

anthony-walker:
@speth pushed up some more changes, I will have to put a bit of thought into some of the other changes.

ischoegl:
Superseded by #2113.

PR #1647, opened 2023-11-20
Migrate "Science" docs

speth:
**Changes proposed in this pull request**

- Migrate content from the `cantera-website` "Science" section into the versioned Sphinx documentation
- Significantly revise "species thermo" documentation
  - document _all_ `SpeciesThermoInterpType` models (e.g. NASA polynomials), models derived from `PDSS` (e.g. temperature-dependent density), and models where the parameters are processed directly by the `ThermoPhase` model (e.g. Peng-Robinson).
  - Try to bridge the gap between the typical definition of "standard state" and Cantera's mixed usage of "reference state" and "standard state".
- Reorganize reactor network docs to split out a tutorial (now in the "User Guide" section) and a description of the CVODES inner workings (now in the "Develop" section).

**If applicable, fill in the issue number this pull request is fixing**

Continues work on Cantera/enhancements#178
Closes Cantera/cantera-website#249

**If applicable, provide an example illustrating new features this pull request is introducing**

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

PR #1648, opened 2023-12-04
[scons] Bump python_max_p1_version to 3.13

matthiasdiener:
Followup of #1604, #1612. See also https://github.com/conda-forge/cantera-feedstock/pull/35 for context.

- Allows building the package with Python 3.12

**Checklist**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -189,7 +189,7 @@ jobs:
         brew link --force --overwrite python@${{ matrix.python-version }}
         brew install --display-times scons
     - name: Upgrade pip
-      run: $PYTHON_CMD -m pip install -U pip 'setuptools>=47.0.0,<48' wheel
+      run: $PYTHON_CMD -m pip install -U pip setuptools wheel
     - name: Install Python dependencies
       run: |
         $PYTHON_CMD -m pip install ruamel.yaml numpy cython pandas pytest pytest-github-actions-annotate-failures pint
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,7 +65,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        python-version: ['3.8', '3.10', '3.11']
+        python-version: ['3.8', '3.10', '3.11', '3.12']
         os: ['ubuntu-20.04', 'ubuntu-22.04']
       fail-fast: false
     env:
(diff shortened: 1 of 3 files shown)


bryanwweber:
I think we also need to bump the CI runner configuration to make sure we test on 3.12.

matthiasdiener:
> I think we also need to bump the CI runner configuration to make sure we test on 3.12.

Thanks, I added 3.12 to the CI configs.

bryanwweber:
It looks like we might need to bump Cython to fix that build, but I'm not sure why it's only failing on macOS. Very strange.

matthiasdiener:
> It looks like we might need to bump Cython to fix that build, but I'm not sure why it's only failing on macOS. Very strange.

As far as I can see, the cython version is the same across all builds (3.0.6). I believe it is just using an old setuptools version for the macos build, which I hopefully resolved in [f0f7d28](https://github.com/Cantera/cantera/pull/1648/commits/f0f7d280e0cf27449480f50e5e29bba72a875b0f).

matthiasdiener:
That seems to have fixed it. The remaining CI errors look the same as in `main`, so I think this is ready to go.

.github/workflows/main.yml
```diff
@@ -423,7 +423,7 @@ jobs:
           -exec sh -c 'for n; do echo "$n" | tee -a results.txt && python3 "$n" >> results.txt || exit 1; done' sh {} +
         env:
           # The ignore setting here is due to a new warning introduced in Matplotlib==3.6.0
-          PYTHONWARNINGS: "error,ignore:warn_name_set_on_empty_Forward::pyparsing"
+          PYTHONWARNINGS: "error,ignore:warn_name_set_on_empty_Forward::pyparsing, ignore::DeprecationWarning"
```
speth: Can you make this more specific, rather than ignoring all deprecation warnings? Ideally, we would want to see deprecations from calls made directly by Cantera, just not ones that occur indirectly through another package. While that may not be possible, we can at least limit the scope of this to be whatever package is currently issuing such warnings.

speth (changes requested):
If you would like to also fix the other CI errors, you can cherry pick 5128a891145887c00eabe2b874f70dcba0b91da2 from #1647.

.github/workflows/main.yml
matthiasdiener: Thanks @speth! I was able to make the ignore more specific in [880de60](https://github.com/Cantera/cantera/pull/1648/commits/880de60b167977485684c2c5ed32ce9aa086926d), and cherry-picked your other CI fix.

speth (approved):
Thanks, @matthiasdiener, this looks good to me.

PR #1622, opened 2023-09-23
Counterflow Flame control

wandadars:
This pull request supersedes #1510 and is based on the newer coding in the StFlow class. This feature aims to provide a capability for users to generate 1D flame solutions that include solutions along the unstable burning branch of the typical s-curve that is used for flamelet progress variable methods.

- Introduce two-point flame control method that adjusts the Lambda equation and adds a new equation, Uo as a dummy equation.
- C++ adjustments to the StFlow class for axisymmetric stagnation flow configurations (only valid case for the two-point control currently)
- Python interface to allow users to control the two-point control.

**Checklist**

- [ ] The pull request includes a clear description of this code change
- [ ] Commit messages have short titles and reference relevant issues
- [ ] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [ ] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [ ] The pull request is ready for review

PR #1660, opened 2024-01-17
Minor adjustment to debugging output in Sim1d.cpp for log levels >7

wandadars:

I added the "true" overwrite flag to the debugging save() and saveResidual() calls in the Sim1d.cpp file to prevent code-stopping errors. I cam across this when trying to debug my work in #1622 . @ischoegl  suggested putting the changes in a separate pull request.

To reproduce the error, run a 1D flow case that experiences failures in the steady newton solve, after taking a few successful newton solves. The error consists of the data group at the highest level already existing in the debug file, causing the code to error out.

**Checklist**

### DIFF
--- a/src/oneD/Sim1D.cpp
+++ b/src/oneD/Sim1D.cpp
@@ -447,7 +447,7 @@ void Sim1D::solve(int loglevel, bool refine_grid)
                 }
 
                 if (loglevel > 6) {
-                    save("debug_sim1d.yaml", "debug",
+                    save("debug_sim1d.yaml", "solution",
                          "After successful Newton solve", true);
                 }
                 if (loglevel > 7) {
@@ -458,7 +458,7 @@ void Sim1D::solve(int loglevel, bool refine_grid)
             } else {
                 debuglog("    failure. \n", loglevel);
                 if (loglevel > 6) {
-                    save("debug_sim1d.yaml", "debug",
+                    save("debug_sim1d.yaml", "solution",
                          "After unsuccessful Newton solve", true);
                 }
                 if (loglevel > 7) {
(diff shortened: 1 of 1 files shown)


src/oneD/Sim1D.cpp
```diff
@@ -448,34 +448,34 @@ void Sim1D::solve(int loglevel, bool refine_grid)
...
+                                 "After unsuccessful Newton solve", true);
                 }
                 if (loglevel > 0) {
                     writelog("Take {} timesteps   ", nsteps);
                 }
                 dt = timeStep(nsteps, dt, m_state->data(), m_xnew.data(), loglevel-1);
                 m_xlast_ts = *m_state;
                 if (loglevel > 6) {
-                    save("debug_sim1d.yaml", "debug", "After timestepping");
+                    save("debug_sim1d.yaml", "debug", "After timestepping", true);
```
speth: A minor formatting nitpick: please wrap this line to keep the maximum line length no more than 88 characters.

speth (changes requested):
Thanks, @wandadars, the changes here makes sense to me. Besides one minor formatting quibble, I'd like to suggest rewriting the commit message to something along the lines of:

```
[1D] Set overwrite flag when saving debug output files

This prevents errors that could occur if the steady-state Newton solver
experiences failures after previous successful Newton solves. In this situation,
the data group at the highest level already exists in the debug file, and must
explicitly be overwritten.
```

src/oneD/Sim1D.cpp
wandadars: That line is 82 characters, is that just too close to the 88 limit?

ischoegl (approved):

src/oneD/Sim1D.cpp
speth: Sorry, my mistake. This is fine. I think I had something set incorrectly in my editor while looking at this.

speth (approved):

PR #1661, opened 2024-01-20
Prevent segfaults for partial 0-D objects

### LINKED ISSUE #1623: Access violation dealing with Wall between empty Reactors
**Problem description**
Trying to access `heat_rate` (and probably other stuff too) from a `Wall` connecting two empty Reactors (i.e. with no defined phase) crashes the interpreter.

Clearly, there should be no proper result without defined phases in the reactors, but an access violation seems a bit drastic to me. 

**Steps to reproduce**
```python
import cantera as ct

r1 = ct.Reactor()
r2 = ct.Reactor()

w = ct.Wall(r1,r2)

w.heat_rate
```

**Behavior**
The result for me is
```
Windows fatal exception: access violation
```

I suppose there is just a check missing on the C++ side of the object, because the conceptually similar `MassFlowController` handles the same situation by throwing
```python
CanteraError: 
CanteraError thrown by ReactorBase::contents:
Reactor contents not defined.
```

I suppose copying this behavior to the `Wall` object would be enough.

**System information**

- Cantera version: 3.1.0a1
- OS: Windows 10
- Python/MATLAB/other software versions: Python 3.11.5

ischoegl:

- Prevent segfaults in reactors with no contents
- Prevent segfaults for unconnected walls
- Add google tests

Currently existing guards are largely specific to the implementation of the Python API (example: retrieving temperature for an empty reactor fails when attempting to load the current reactor state into a thermo object). As underlying C++ code is difficult to reach from Python (but may be exposed elsewhere), unit tests are added for C++ directly.

Fixes #1623

Example taken from #1623:
```
In [1]: import cantera as ct
   ...:
   ...: r1 = ct.Reactor()
   ...: r2 = ct.Reactor()
   ...:
   ...: w = ct.Wall(r1,r2)

In [2]: w.heat_rate
Cell In[2], line 1
----> 1 w.heat_rate

File /Volumes/Data/work/GitHub/cantera/build/python/cantera/reactor.pyx:1066, in cantera.reactor.WallBase.heat_rate.__get__()
   1064 .. versionadded:: 3.0
   1065 """
-> 1066 return self.wall.heatRate()
   1067
   1068

CanteraError:
CanteraError thrown by ReactorBase::temperature:
Reactor state empty and/or contents not defined.
```

**Other thoughts**

An alternative to the added guards is to prevent the creation of empty (or unconnected) 0-D objects altogether (which should likely be done when moving from raw pointers to shared pointers for various held objects - `m_thermo`, etc.). As this is more involved and will require a deprecation cycle, the guards proposed in this PR are still appropriate.

**Checklist**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -412,7 +412,7 @@ jobs:
         run: python3 -m pip install -U pip setuptools wheel
       - name: Install Python dependencies
         run: |
-            python3 -m pip install numpy ruamel.yaml pandas matplotlib scipy pint graphviz
+            python3 -m pip install numpy ruamel.yaml pandas pyarrow matplotlib scipy pint graphviz
             python3 -m pip install --pre --no-index --find-links dist cantera
       - name: Run the examples
         # See https://unix.stackexchange.com/a/392973 for an explanation of the -exec part
--- a/include/cantera/zeroD/ReactorBase.h
+++ b/include/cantera/zeroD/ReactorBase.h
@@ -202,11 +202,19 @@ public:
 
     //! Returns the current density (kg/m^3) of the reactor's contents.
     double density() const {
+        if (m_state.empty()) {
+            throw CanteraError("ReactorBase::density",
+                               "Reactor state empty and/or contents not defined.");
+        }
         return m_state[1];
     }
 
     //! Returns the current temperature (K) of the reactor's contents.
     double temperature() const {
+        if (m_state.empty()) {
+            throw CanteraError("ReactorBase::temperature",
+                               "Reactor state empty and/or contents not defined.");
+        }
         return m_state[0];
     }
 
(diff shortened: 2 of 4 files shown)


speth (changes requested):
Thanks, @ischoegl, all the changes here look good to me, and I appreciate the thoughts on longer-term options to eliminate the need for these checks.

My only request here is to document what upstream deprecation warning is resolved by installing `pyarrow` as part of the commit message. Do we need to update the dependencies of some of the examples?

ischoegl:
> My only request here is to document what upstream deprecation warning is resolved by installing `pyarrow` as part of the commit message. 

:+1: done

> Do we need to update the dependencies of some of the examples?

I don't think so - it's just pandas complaining that it will require pyarrow down the line.

speth (approved):

Issue #1457, opened 2023-03-18
addSpecies should be disabled on ThermoPhase objects in use by reactors and flames


speth:
**Problem description**

The `ThermoPhase::addSpecies` method can be used to add species to a phase dynamically. However, this capability is not compatible with objects such as `Reactor` or `Domain1D`, which set internal array sizes based on the number of species in the phase at the time they are created.

In Python, this is protected against by having `Solution` objects hold a weak reference any `Reactor`, `Domain1D`, or `Mixture` objects that use that `Solution`. No equivalent exists at the C++ layer, though, and adding species after creating such objects can lead to crashes.

Managing this in a similar way in C++ would have been difficult until recently, but I think the transition to having objects like `Domain1D` hold the `Solution` as a `shared_ptr` (see #1385, for example) should make this more feasible after some of the old routes have been removed following Cantera 3.0. I think one prerequisite will be using `shared_ptr<Solution>` to set up `Reactor` objects, which have not had this change yet.

**Steps to reproduce**

Replace the integration loop in `samples/cxx/combustor/combustor.cpp` with the following:
```c++
    bool added = false;
    while (tnow < tfinal) {
        tnow += 0.005;
        sim.advance(tnow);
        tres = combustor.mass()/v.massFlowRate();
        f << tnow << ", "
          << combustor.temperature() << ", "
          << tres << ", ";
        ThermoPhase& c = combustor.contents();
        for (size_t i = 0; i < k_out.size(); i++) {
            f << c.moleFraction(k_out[i]) << ", ";
        }
        f << std::endl;
        if (!added) {
            auto c6h6 = make_shared<Species>("C6H6", Composition({{"C", 6}, {"H", 6}}));
            auto spthermo = c6h6->thermo->parameters();
            spthermo.applyUnits();
            c6h6->thermo = newSpeciesThermo(spthermo);
            gas->addSpecies(c6h6);
            added = true;
        }
    }
```
(also requires adding includes for `cantera/thermo/Species.h` and `cantera/thermo/SpeciesThermoFactory.h`)

**Behavior**

The above segfaults after the first timestep.

**Additional context**

Initially discussed in PR review for #1456.

ischoegl:
#1663 and #1670/#1685 ensure that individual `ThermoPhase` and `Kinetics` objects are always tied to a `Solution` object. With this, the `_WeakrefProxy` guards existing in Python can be moved to C++, which should allow for a clean resolution?

[...]

PR #1663, opened 2024-01-21
Use Solution for zeroD objects

ischoegl:

- Assign `Solution` to zeroD objects where possible
- Implement instantiation of non-empty `Reactor` objects, and trigger deprecation warnings when attempting to create reactors without content whenever feasible
- Some legacy routes for empty `Reactor` objects remain (example: legacy MATLAB toolbox)

Prepares a future solution for #1457

**Checklist**

### DIFF
--- a/include/cantera/clib/ctreactor.h
+++ b/include/cantera/clib/ctreactor.h
@@ -14,12 +14,12 @@
 extern "C" {
 #endif
 
-    CANTERA_CAPI int reactor_new(const char* type);
+    CANTERA_CAPI int reactor_new(const char* type);  //! @deprecated: remove after 3.1
+    CANTERA_CAPI int reactor_new3(const char* type, int n, const char* name);
     CANTERA_CAPI int reactor_del(int i);
     CANTERA_CAPI int reactor_setInitialVolume(int i, double v);
     CANTERA_CAPI int reactor_setChemistry(int i, int cflag);
     CANTERA_CAPI int reactor_setEnergy(int i, int eflag);
-    CANTERA_CAPI int reactor_setSolution(int i, int n);
     CANTERA_CAPI int reactor_setThermoMgr(int i, int n);
     CANTERA_CAPI int reactor_setKineticsMgr(int i, int n);
     CANTERA_CAPI int reactor_insert(int i, int n);
--- a/include/cantera/clib/ctreactor.h
+++ b/include/cantera/clib/ctreactor.h
@@ -19,6 +19,7 @@ extern "C" {
     CANTERA_CAPI int reactor_setInitialVolume(int i, double v);
     CANTERA_CAPI int reactor_setChemistry(int i, int cflag);
     CANTERA_CAPI int reactor_setEnergy(int i, int eflag);
+    CANTERA_CAPI int reactor_setSolution(int i, int n);
     CANTERA_CAPI int reactor_setThermoMgr(int i, int n);
     CANTERA_CAPI int reactor_setKineticsMgr(int i, int n);
     CANTERA_CAPI int reactor_insert(int i, int n);
--- a/include/cantera/zeroD/ReactorBase.h
+++ b/include/cantera/zeroD/ReactorBase.h
@@ -26,6 +26,7 @@ class ReactorNet;
 class ReactorSurface;
 class Kinetics;
 class ThermoPhase;
+class Solution;
 
 enum class SensParameterType {
     reaction,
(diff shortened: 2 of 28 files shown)


include/cantera/zeroD/Reservoir.h
```diff
@@ -18,21 +18,27 @@ namespace Cantera
 class Reservoir : public ReactorBase
 {
 public:
-    Reservoir() {}
+    Reservoir(shared_ptr<Solution> sol, const string& name = "(none)") {
+        m_solution = sol;
+        m_name = name;
+        setThermoMgr(*sol->thermo());
+    }
```
speth: Given the implementation of `ReactorBase::setSolution`, is this specialization of the constructor necessary?

include/cantera/zeroD/ReactorBase.h
```diff
@@ -78,12 +89,17 @@ class ReactorBase
         m_vol = vol;
     }
 
+    //! @deprecated To be removed after %Cantera 3.1. Superseded by setSolution.
+    void insert(shared_ptr<Solution> sol);
+
     //! Specify the mixture contained in the reactor. Note that a pointer to
     //! this substance is stored, and as the integration proceeds, the state of
     //! the substance is modified.
+    //! @todo: make protected (should only be used internally)
```
speth: Instead of labeling this as a "todo", would it make sense to write something like `@since After %Cantera 3.1, this method will be become a 'protected' method for internal use only.`?

interfaces/cython/cantera/reactor.pyx
```diff
@@ -27,16 +27,20 @@ cdef class ReactorBase:
...
         self._outlets = []
         self._walls = []
         self._surfaces = []
-        if isinstance(contents, ThermoPhase):
+        if isinstance(contents, _SolutionBase):
             self.insert(contents)
+        else:
+            warnings.warn(
+                "Starting in Cantera 3.1, the reactor contents must not be empty.",
+                DeprecationWarning)
```
speth, line 43: I would suggest also updating the docstring for the `contents` argument to `Reactor.__init__` below.

include/cantera/clib/ctreactor.h
```diff
@@ -19,6 +19,7 @@ extern "C" {
     CANTERA_CAPI int reactor_setInitialVolume(int i, double v);
     CANTERA_CAPI int reactor_setChemistry(int i, int cflag);
     CANTERA_CAPI int reactor_setEnergy(int i, int eflag);
+    CANTERA_CAPI int reactor_setSolution(int i, int n);
```
speth: I'm wondering about how adding this method fits with the goal of eventually eliminating instantiation of empty `Reactor` objects. Don't we need an alternative to `reactor_new` that takes both the reactor type and the handle for the `Solution` object?

test/python/test_reactor.py
```diff
@@ -67,7 +68,8 @@ def test_insert(self):
         self.assertEqual(len(R.kinetics.net_production_rates), g.n_species)
 
     def test_volume(self):
-        R = self.reactorClass(volume=11)
+        with pytest.warns(DeprecationWarning, match="must not be empty"):
+            R = self.reactorClass(volume=11)
```
speth: I would suggest modifing this to specify the contents rather than handling the `DeprecationWarning`. That would eliminate the need to revisit this as part of processing all the deprecated items after the 3.1 release.

test/python/test_reactor.py
```diff
@@ -853,7 +855,8 @@ def test_invalid_property(self):
                 x.foobar
 
     def test_bad_kwarg(self):
-        self.reactorClass(name='ok')
+        with pytest.warns(DeprecationWarning, match="must not be empty"):
+            self.reactorClass(name='ok')
```
speth: Likewise, I would suggest specifying `contents` here.

test/zeroD/test_zeroD.cpp
```diff
@@ -19,16 +19,17 @@ TEST(zerodim, simple)
 
     auto sol = newSolution("gri30.yaml", "gri30", "none");
     sol->thermo()->setState_TPX(T, P, X);
-    IdealGasReactor cppReactor;
-    cppReactor.insert(sol);
-    cppReactor.initialize();
+    auto cppReactor = newReactor("IdealGasReactor", sol, "simple");
+    // IdealGasReactor cppReactor(sol, "simple");
```
speth: The commented-out line here can be removed.

speth (changes requested):
Thanks, @ischoegl, making this transition makes sense to me. I had just a few minor suggestions and questions regarding the implementation.

include/cantera/zeroD/ReactorBase.h
ischoegl: I'm a little hesitant, as `@since` implies that it is already done; also, the nice thing about `@todo` is that it can be easily searched for and is recognized by doxygen. I'm combining what you suggest with what I had already.

include/cantera/zeroD/Reservoir.h
ischoegl: good catch!

include/cantera/clib/ctreactor.h
ischoegl: Excellent point. This does require a little more work in the experimental MATLAB interface, but it's definitely worth the effort.

ischoegl:
@speth ... thanks for the review! As there aren't unit tests for that portion available yet, I tested the modifications of the "experimental" MATLAB toolbox locally (with the side-effect that I now have MATLAB for Apple Silicon installed). So I guess this should be ready ...

speth (approved):

PR #1664, opened 2024-01-21
Expose Peng-Robinson EoS parameters to Python interface

wandadars:
This pull request seeks to expose the Peng-Robinson equation of state parameters, such as aAlpha, b, and the first and second derivatives of aAlpha with respect to temperature so that these values are available through the Python interface of the ThermoPhase object. These parameters are often used in the context of compressible flamelet table tabulations.

See: https://arxiv.org/abs/1704.02639

I'm attempting to use the AnyValue object for the return type to generalize a bit. The overall idea I had was to have a catch-all function that can dispatch to method that return values from the eos objects.

```
def add_additional_data(flame, gas):
    for i in range(flame.flame.n_points):
        # Set the state of the gas to match the local state
        T = flame.T[i]
        P = flame.P
        X = flame.X[:,i]
        gas.TPX = T, P, X

        # Compute additional data
        aAlpha_mix = gas.compute_extra_method('aAlpha_mix')
        b_mix = gas.compute_extra_method('b_mix')
        daAlpha_dT = gas.compute_extra_method('daAlpha_dT')
        d2aAlpha_dT2 = gas.compute_extra_method('d2aAlpha_dT2')
        print(f'aAlpha_mix: {aAlpha_mix}, b_mix: {b_mix}, daAlpha_dT: {daAlpha_dT}, d2aAlpha_dT2: {d2aAlpha_dT2}')

```

**Checklist**

### DIFF
--- a/include/cantera/thermo/PengRobinson.h
+++ b/include/cantera/thermo/PengRobinson.h
@@ -177,10 +177,10 @@ public:
     /*!
      * @return
      *   Returns an AnyMap containing the following key names and values:
-     *   - "aAlpha_mix": The value of m_aAlpha_mix
-     *   - "b_mix": The value of the m_b
-     *   - "daAlpha_dT": The value returned by the daAlpha_dT() method
-     *   - "d2aAlpha_dT2": The value returned by the d2aAlpha_dT2() method
+     *   - `aAlpha_mix`: The value of #m_aAlpha_mix
+     *   - `b_mix`: The value of the #m_b
+     *   - `daAlpha_dT`: The value returned by the daAlpha_dT() method
+     *   - `d2aAlpha_dT2`: The value returned by the d2aAlpha_dT2() method
      */
     AnyMap getAuxiliaryData() override;
 
--- a/include/cantera/thermo/PengRobinson.h
+++ b/include/cantera/thermo/PengRobinson.h
@@ -173,6 +173,14 @@ public:
     void setBinaryCoeffs(const string& species_i, const string& species_j, double a);
     //! @}
 
+    //! Compute additional method based on name
+    /*!
+     * @param name The name of the quantity to compute
+     * @return Computed value wrapped in AnyValue type for Peng-Robinson phase
+     *         quantities
+     */
+    AnyValue compute_extra_method(const std::string& name) override;
+
 protected:
     // Special functions inherited from MixtureFugacityTP
     double sresid() const override;
--- a/include/cantera/thermo/PengRobinson.h
+++ b/include/cantera/thermo/PengRobinson.h
@@ -173,13 +173,8 @@ public:
     void setBinaryCoeffs(const string& species_i, const string& species_j, double a);
     //! @}
 
-    //! Compute additional method based on name
-    /*!
-     * @param name The name of the quantity to compute
-     * @return Computed value wrapped in AnyValue type for Peng-Robinson phase
-     *         quantities
-     */
-    AnyValue compute_extra_method(const std::string& name) override;
+    //! Return common parameters used by the Peng-Robinson equation of state
+    AnyMap getEosParameters() override;
 
 protected:
     // Special functions inherited from MixtureFugacityTP
--- a/include/cantera/thermo/PengRobinson.h
+++ b/include/cantera/thermo/PengRobinson.h
@@ -173,8 +173,16 @@ public:
     void setBinaryCoeffs(const string& species_i, const string& species_j, double a);
     //! @}
 
-    //! Return common parameters used by the Peng-Robinson equation of state
-    AnyMap getEosParameters() override;
+    //! Return parameters used by the Peng-Robinson equation of state.
+    /*!
+     * @return
+     *   Returns an AnyMap containing the following key names and values:
+     *   - "aAlpha_mix": The value of m_aAlpha_mix
+     *   - "b_mix": The value of the m_b
+     *   - "daAlpha_dT": The value returned by the daAlpha_dT() method
+     *   - "d2aAlpha_dT2": The value returned by the d2aAlpha_dT2() method
+     */
+    AnyMap getAuxiliaryData() override;
 
 protected:
     // Special functions inherited from MixtureFugacityTP
--- a/include/cantera/thermo/ThermoPhase.h
+++ b/include/cantera/thermo/ThermoPhase.h
@@ -1556,6 +1556,16 @@ public:
             const Composition& oxComp, ThermoBasis basis=ThermoBasis::molar) const;
     //! @}
 
+    //! Compute additional method based on name
+    /*!
+     * @param name The name of the quantity to compute
+     * @return Computed value wrapped in AnyValue type
+     */
+    virtual AnyValue compute_extra_method(const std::string& name)
+    {
+        throw NotImplementedError("ThermoPhase::compute_extra_method");
+    }
+
 private:
 
     //! Carry out work in HP and UV calculations.
--- a/include/cantera/thermo/ThermoPhase.h
+++ b/include/cantera/thermo/ThermoPhase.h
@@ -1556,14 +1556,11 @@ public:
             const Composition& oxComp, ThermoBasis basis=ThermoBasis::molar) const;
     //! @}
 
-    //! Compute additional method based on name
-    /*!
-     * @param name The name of the quantity to compute
-     * @return Computed value wrapped in AnyValue type
-     */
-    virtual AnyValue compute_extra_method(const std::string& name)
+    //! Return common parameters used by an equation of state that a user may
+    //! want to access.
+    virtual AnyMap getEosParameters()
     {
-        throw NotImplementedError("ThermoPhase::compute_extra_method");
+        throw NotImplementedError("ThermoPhase::getEosParameters");
     }
 
 private:
--- a/include/cantera/thermo/ThermoPhase.h
+++ b/include/cantera/thermo/ThermoPhase.h
@@ -1556,11 +1556,12 @@ public:
             const Composition& oxComp, ThermoBasis basis=ThermoBasis::molar) const;
     //! @}
 
-    //! Return common parameters used by an equation of state that a user may
-    //! want to access.
-    virtual AnyMap getEosParameters()
+    //! Return intermediate or model-specific parameters used by particular
+    //! derived classes. Specific parameters are described in overidden
+    //! methods of classes that derive from the base class.
+    virtual AnyMap getAuxiliaryData()
     {
-        throw NotImplementedError("ThermoPhase::getEosParameters");
+        return AnyMap();
     }
 
 private:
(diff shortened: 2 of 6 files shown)


include/cantera/thermo/PengRobinson.h
```diff
@@ -173,6 +173,14 @@ class PengRobinson : public MixtureFugacityTP
     void setBinaryCoeffs(const string& species_i, const string& species_j, double a);
     //! @}
 
+    //! Compute additional method based on name
+    /*!
+     * @param name The name of the quantity to compute
+     * @return Computed value wrapped in AnyValue type for Peng-Robinson phase
+     *         quantities
+     */
+    AnyValue compute_extra_method(const std::string& name) override;
```
ischoegl:
```suggestion
    AnyValue computeExtraMethod(const std::string& name) override;
```
In C++, Cantera uses mostly camelCase methods rather than snake_case.

src/thermo/PengRobinson.cpp
```diff
@@ -780,4 +780,18 @@ int PengRobinson::solveCubic(double T, double pres, double a, double b, double a
                                          an, bn, cn, dn, tc, vc);
 }
 
+AnyValue PengRobinson::compute_extra_method(const std::string& name)
+{
+    if(name == "aAlpha_mix") {
```
ischoegl:
```suggestion
    if (name == "aAlpha_mix") {
```
Missing whitespace after `if`

ischoegl (commented):
I am not sure whether I am fully behind hard-coded names as parameters, as users need to know what those are (an alternative would be to return an `AnyMap` populated with all applicable entries, which can then be mapped to a Python `dict`). I am also unsure whether the name `compute_extra_method` is intuitive.

PS: for the failing examples, there's a pending fix in 62300d345a26226b4c9e85de69371fe33113f75a

speth:
I recall us having discussed this idea at some point before. Can you please update the initial post to describe the use case for making these parameters accessible?

wandadars:
> I recall us having discussed this idea at some point before. Can you please update the initial post to describe the use case for making these parameters accessible?

I have updated with a reference to a paper showing the compressible flamelet methodology that utilizes tabulations of real-gas parameters to create better flamelet tables.

wandadars:
Now that I've worked on the implementation, essentially all of the additional methods are just calling already defined functions in the PREOS implementation, so those could be exposed instead of adding this new virtual method. I don't know if that is better or worse.

ischoegl:
> Now that I've worked on the implementation, essentially all of the additional methods are just calling already defined functions in the PREOS implementation, so those could be exposed instead of adding this new virtual method. I don't know if that is better or worse.

From the perspective of the Python API implementation (or any other language interface), it is easier (preferable?) to have a single function signature that is common to all `ThermoPhase` models (i.e. the virtual method), which returns implementation-specific values. As mentioned above, it would likewise be good to not force the user to guess or look up what keywords are being used internally, so I'd recommend using the `AnyMap` route pointed out before. Of course, these are only my 2 cents ...

wandadars:
> I am not sure whether I am fully behind hard-coded names as parameters, as users need to know what those are (an alternative would be to return an `AnyMap` populated with all applicable entries, which can then be mapped to a Python `dict`). I am also unsure whether the name `compute_extra_method` is intuitive.
> 
> PS: for the failing examples, there's a pending fix in [62300d3](https://github.com/Cantera/cantera/commit/62300d345a26226b4c9e85de69371fe33113f75a)

So for an AnyMap return type, you're saying that the function should just collect all parameters and return them all to user instead of letting them choose?

ischoegl:
> So for an AnyMap return type, you're saying that the function should just collect all parameters and return them all to user instead of letting them choose?

Correct.

wandadars:
Ok, I switched to the AnyMap, and that is a more user-friendly option for sure.

wandadars:
I have an example here:

```
import cantera as ct

gas = ct.Solution('co2_PR_example.yaml', 'CO2-PR')
gas.TPX = 300, 101325, 'H2:1.0'
params = gas.get_eos_parameters()

print(params)
```
This just uses the Cantera co2_PR_example.yaml and the data in critical-properties.yaml

The CO2 file is at: /test/data/  and the critical file is at /build/data

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1556,6 +1556,13 @@ class ThermoPhase : public Phase
             const Composition& oxComp, ThermoBasis basis=ThermoBasis::molar) const;
     //! @}
 
+    //! Return common parameters used by an equation of state that a user may
+    //! want to access.
+    virtual AnyMap getEosParameters()
```
speth: I'd like to suggest the name `getAuxiliary` or `getAuxiliaryData`, as I don't think this needs to be specific to retrieving equation of state parameters (and avoids having to stick the abbreviation "eos" into the method name).

I don't think I'd describe the parameters returned by this method as "common" -- at least for the current and any other use I can think of, it's more that they are intermediate, model-specific values in the equation of state calculation. Also, it would be helpful to mention where the user can find what parameters are retrieved for a given thermo model.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -1556,6 +1556,13 @@ class ThermoPhase : public Phase
...
+    {
+        throw NotImplementedError("ThermoPhase::getEosParameters");
+    }
```
speth, line 1565: I think the base class can just return an empty `AnyMap` rather than throwing an exception. Then, derived classes can call the base class method first to get an `AnyMap` with these parameters from the base class and add any specific to the derived class.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -1172,6 +1172,13 @@ cdef class ThermoPhase(_SolutionBase):
         X = self.thermo.getMoleFractionsByName(threshold)
         return {pystr(item.first):item.second for item in X}
 
+    def get_eos_parameters(self):
```
speth: You could make this a `property` with the name `auxiliary_data`.

include/cantera/thermo/PengRobinson.h
```diff
@@ -173,6 +173,9 @@ class PengRobinson : public MixtureFugacityTP
     void setBinaryCoeffs(const string& species_i, const string& species_j, double a);
     //! @}
 
+    //! Return common parameters used by the Peng-Robinson equation of state
+    AnyMap getEosParameters() override;
```
speth: For the derived class, the docstring should what parameters are returned, and what the key names are in the `AnyMap`.

src/thermo/PengRobinson.cpp
```diff
@@ -780,4 +780,17 @@ int PengRobinson::solveCubic(double T, double pres, double a, double b, double a
                                          an, bn, cn, dn, tc, vc);
 }
 
+AnyMap PengRobinson::getEosParameters()
+{
+    AnyMap parameters;
+
+    // Add each parameter to the map.
```
speth: This comment is too literal to be useful.

src/thermo/PengRobinson.cpp
```diff
@@ -780,4 +780,17 @@ int PengRobinson::solveCubic(double T, double pres, double a, double b, double a
...
+    parameters["aAlpha_mix"] = AnyValue(m_aAlpha_mix);
+    parameters["b_mix"] = AnyValue(m_b);
+    parameters["daAlpha_dT"] = AnyValue(daAlpha_dT());
+    parameters["d2aAlpha_dT2"] = AnyValue(d2aAlpha_dT2());
```
speth: I don't think it should be necessary to explicitly cast these as `AnyValue`.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -1172,6 +1172,13 @@ cdef class ThermoPhase(_SolutionBase):
...
+        """
+        Return a dictionary giving common eos parameters.
+        """
```
speth: See the comments on the docstring for `ThermoPhase::getEosParameters`.

speth (changes requested):
Thanks for implementing this, @wandadars. I think it will be a useful feature, especially in some debugging applications. Besides the specific comments below, this also needs a test, ideally for the Python interface.

include/cantera/thermo/ThermoPhase.h
wandadars: That's a great name. I've updated to the getAuxiliaryData name, and changed the docstring.

wandadars:
Ok. I made some final changes that I had missed when we renamed the method. I added a Python test to the ThermoPhase Python testing file.  I'm not well versed in handling of YAML files that have multiple definitions in them (like the co2_PR one), but I think it does the right thing.  One thing I'm not 100% sure on with the Peng Robinson specifications is how we can set the Peng Robinson equation of state to be used, but to give it a species that is ideal-like. That way we could potentially test the values of the parameters that are returned, as I believe for a single species, those values would all be zero.  I just haven't quite figured out how to formalize such a configuration using Cantera yet.

include/cantera/thermo/PengRobinson.h
```diff
@@ -173,6 +173,17 @@ class PengRobinson : public MixtureFugacityTP
...
     //! @}
 
+    //! Return parameters used by the Peng-Robinson equation of state.
+    /*!
+     * @return
+     *   Returns an AnyMap containing the following key names and values:
+     *   - "aAlpha_mix": The value of m_aAlpha_mix
+     *   - "b_mix": The value of the m_b
+     *   - "daAlpha_dT": The value returned by the daAlpha_dT() method
+     *   - "d2aAlpha_dT2": The value returned by the d2aAlpha_dT2() method
```
speth: For better formatting and autolinking by Doxygen, I'd suggest:
```suggestion
     *   - `aAlpha_mix`: The value of #m_aAlpha_mix
     *   - `b_mix`: The value of the #m_b
     *   - `daAlpha_dT`: The value returned by the daAlpha_dT() method
     *   - `d2aAlpha_dT2`: The value returned by the d2aAlpha_dT2() method
```

test/python/test_thermo.py
```diff
@@ -970,6 +970,16 @@ def test_add_species_duplicate(self):
...
 
+    def test_auxiliary_data(self):
+        """
+        Should get back a dictionary with 4 keys present when calling the auxiliary_data
+        property for the Peng-Robinson thermo phase.
+        """
+        gas = ct.Solution('co2_PR_example.yaml', 'CO2-PR')
+        gas.TPX = 300, 101325, 'H2:1.0'
+        params = gas.auxiliary_data
+        self.assertEqual(len(params), 4)
```
speth: We've been transitioning to the `pytest` style of test assertions, so you can just write
```suggestion
        assert len(params) == 4
```

speth (changes requested):
Thanks for the updates, @wandadars. I only have a couple of minor formatting/style requests. 

For the test, I think you should check that the keys returned are what you expect, rather than just the number of keys is correct. I don't think you necessarily need to worry about whether the numerical values are correct in this test. But if you want to check something, you could use the returned values of `aAlpha_mix` and `b_mix` along with the current temperature and molar volume to compute the pressure of the mixture using the PR equation of state and see if it matches.

wandadars:
I have made those formatting updates and added that much better pressure testing unit test.

speth (approved):
Thanks, @wandadars. This looks good to me!

PR #1665, opened 2024-01-23
Complete Unit Test Suite for the Matlab Toolbox - Pt.1

ssun30:

Added unit tests for all methods and properties currently implemented in the following classes:
- ThermoPhase
- Kinetics
- Transport
- Func1

Fixed bugs that prevent these tests from passing. 

Updates to Cantera/enhancements#177

Introduced a temporary fix to https://github.com/Cantera/cantera/issues/1852 . 

Added tests for rates of progress, rate constants, species rates, and thermodynamic values. 

**Checklist**


ischoegl:
@ssun30 and @rwest ... what are the current plans on this PR and https://github.com/Cantera/enhancements/issues/177? With the old MATLAB toolbox removed, the next Cantera version will only provide for the experimental toolbox.

ssun30:
> @ssun30 and @rwest ... what are the current plans on this PR and [Cantera/enhancements#177](https://github.com/Cantera/enhancements/issues/177)? With the old MATLAB toolbox removed, the next Cantera version will only provide for the experimental toolbox.

Hi Ingmar, finishing the test suite is back on our to-do list and we were aiming to complete it by the end of the month. There is a slight delay so maybe the first week of July we shall have all the unit tests ready.

ssun30:
After a year of hiatus this PR is finally ready for review. Thanks to support from Mathworks the plan to make the Matlab toolbox official is back on track.

ssun30:
Checklist for all unit tests (that can be implemented in MATLAB). Now complete after 0D and 1D tests are separated. 

- [x] ThermoPhase
- [x] Kinetics
- [x] Transport
- [x] Functors
- [x] Equilibrium
- [x] Mixture

ssun30:
@ischoegl @speth It seems that I don't have write access to the repo and thus could not request reviews for this PR. Could you check whether that's true and grant me access?

ischoegl (commented):
Hi @ssun30 ... thanks for getting back to this! From my side, this looks mostly good to go (despite a minor nit that some newlines at file endings are missing). I did notice that the MATLAB CI run wasn't completed successfully, though.

Overall, it's great to have some of this covered, especially as the plan is to move towards replacing the current CLib with an autogenerated CLib (see Cantera/enhancements#220). Once this is in place, it will be relatively easy to expose parts that are currently missing from the MATLAB API (e.g. `Reaction`, `Species`, `ReactionRate`, `SolutionArray`, etc.). One thing that may also be necessary is a MATLAB equivalent to `AnyMap` to access the serialization interface - I believe that there's a new `dictionary` that was introduced in R2022b.

interfaces/matlab_experimental/Base/ThermoPhase.m
```diff
@@ -1393,6 +1393,51 @@ function display(tp)
 
         end
 
+        function set.T(tp, t)
```
ischoegl: I don't think that we should allow for single setter methods, and would in general avoid requiring `input` statements.

ischoegl (changes requested):

ssun30:
For a while I've kept the MATLAB tests as a separate job in the workflow to get quicker feedback from the CI runs. Now that the tests are stable, should I move these tests into the matrix jobs generated on different OS and Python versions?

ischoegl:
> For a while I've kept the MATLAB tests as a separate job in the workflow to get quicker feedback from the CI runs. Now that the tests are stable, should I move these tests into the matrix jobs generated on different OS and Python versions?

I'd suggest to address this in a separate PR (likely the one where installation is integrated into `scons`)

_PS: I think MATLAB tests should be run in a similar fashion to the ones for .NET (eventually)._

ssun30:
In commit https://github.com/Cantera/cantera/pull/1665/commits/4e921fd50354de765579e4299948ff54c161fc72 I added a getter method for heatTransferCoeff along with the corresponding function in clib. This is to temporarily pass `testHeatTransfer2`. But since there are lots of missing methods in the clib that need to be added later, I think a better approach is to filter this test for this PR and enable it later. 

I wonder what's the timeline for the implementation of code-generated clib? Mathworks expects us to release the MATLAB toolbox in an as complete state as possible and so far we have a few properties that lack getter methods (but do have setters, such as `heatTransferCoeff`) and will return an empty value when the user calls them. I suggest having a separate PR to expand the clib such that all properties that are already implemented in the MATLAB toolbox are functional.

ischoegl:
> I wonder what's the timeline for the implementation of code-generated clib? Mathworks expects us to release the MATLAB toolbox in an as complete state as possible and so far we have a few properties that lack getter methods (but do have setters, such as heatTransferCoeff) and will return an empty value when the user calls them. I suggest having a separate PR to expand the clib such that all properties that are already implemented in the MATLAB toolbox are functional.

The timeline for the generated `clib` is currently undefined. Still, I did comment on the roadmap in Cantera/enhancements#220 (the objective was to have this working for 3.2, but this all depends on the speed of reviews and my own varying bandwidth). Please note that I ignored the specific needs of the MATLAB toolbox (I'm not part of this effort), but adding getters is trivial as long as C++ provides for these. 

One area where I see a huge amount of potential improvement for the MATLAB API is code generation (this goes back to Cantera/enhancements#39). Adding in all those getters/setters (and associated docstrings) by hand is imho hard to maintain. I believe adding the test suite first is still the right approach.

interfaces/matlab_experimental/Base/Mixture.m
```diff
@@ -190,21 +194,24 @@
...
 
         function display(m)
             % Display the state of the mixture on the terminal.
 
-            calllib(ct, 'mix_updatePhases', m.mixID);
+            ctFunc('mix_updatePhases', m.mixID);
             [np, nc] = size(m.phases);
 
             for n = 1:np
                 s = [sprintf('\n*******************    Phase %d', n) ...
```
ischoegl, line 212: While this line (212) isn't changed in this PR, I believe that print commands like these should originate from the C++ core so things are consistent across all interfaces.

interfaces/matlab_experimental/Base/Solution.m
```diff
@@ -86,8 +91,32 @@
...
+        %% Solution Class Getter Methods
+
+        function str = get.transportModel(soln)
+            str = ctString('trans_transportModel', soln.trID);
+        end
+
+        %% Solution Class Setter Methods
+
+        function set.transportModel(soln, str)
+            if strcmp(str, soln.transportModel)
```
ischoegl: These checks are already implemented in C++, so this is redundant.

interfaces/matlab_experimental/Base/Water.m
```diff
@@ -17,6 +17,11 @@
     %
     % :return:
     %     Instance of class :mat:class:`Solution`.
-
-    w = Solution('liquidvapor.yaml', 'water');
+    if nargin == 0 | strcmp(backend, 'Reynolds')
+        w = Solution('liquidvapor.yaml', 'water', 'water');
+    elseif strcmp(backend, 'IAPWS95')
+        w = Solution('liquidvapor.yaml', 'liquid-water-IAPWS95', 'water');
```
ischoegl: Please don't introduce MATLAB-specific nomenclature. Strings referencing backends should be passed directly to C++, which should also handle exceptions.

interfaces/matlab_experimental/Reactor/Reactor.m
```diff
@@ -88,10 +88,10 @@
     methods
         %% Reactor Class Constructor
 
-        function r = Reactor(content, typ, name)
+        function r = Reactor(content, typ, name, V0)
```
ischoegl: Python uses class specializations rather than the `typ` argument.

interfaces/matlab_experimental/Reactor/Reactor.m
```diff
@@ -111,35 +111,53 @@
...
+            elseif nargin > 4
                 error('too many arguments');
             end
 
             if ~isa(content, 'Solution')
                 error('Reactor contents must be an object of type "Solution"');
             end
 
+            if isempty(typ)
+                typ = 'Reactor';
```
ischoegl: This exception handling should be left to C++?

interfaces/matlab_experimental/Reactor/Reactor.m
```diff
@@ -111,35 +111,53 @@
...
+            end
+
+            if isempty(name)
+                name = '(none)';
```
ischoegl: Some of this logic should be streamlined - see lines 126 and 128. I believe that C++ will automatically assign `(none)` for an empty name.

interfaces/matlab_experimental/Reactor/ReactorNet.m
```diff
@@ -54,46 +54,57 @@
...
+            if nargin == 0
+                r = {};
             end
 
-            if isa(reactors, 'Reactor')
+            if isa(r, 'Reactor')
                 % Allow simpler syntax for creating a network with one
                 % reactor.
-                reactors = {reactor};
+                r = {r};
```
ischoegl: I believe that single-letter variable names should be avoided except for simple cases (iterators).

interfaces/matlab_experimental/Base/Interface.m
ischoegl, line 1: This file should inherit from `Solution.m`, see https://github.com/Cantera/cantera/blob/c7910524da1388898f81d33b25ba81a1876144a1/interfaces/matlab_experimental/Base/Interface.m

ischoegl (commented):
Thanks, @ssun30. I have some comments on changes to the MATLAB API.

interfaces/matlab_experimental/Base/Mixture.m
ssun30: True. But in the case of `Mixture` class, the cython interface also adopted a similar approach where it calls the `report` method for each individual phase in the mixture. Have we implemented a `report` method for the `Mixture` class in C++?

interfaces/matlab_experimental/Base/Water.m
ssun30: This is also borrowed directly from the cython interface: https://github.com/Cantera/cantera/blob/main/interfaces/cython/cantera/liquidvapor.py

interfaces/matlab_experimental/Reactor/Reactor.m
ssun30: I agree this could definitely be streamlined. Going forward, I will be using MATLAB's [inputParser](https://www.mathworks.com/help/matlab/ref/inputparser.html) to handle all `varargin` cases. I will separate this part of the work to a different PR. 

As for setting the default name, when I pass an empty `cstring` type pointer from MATLAB to clib, `reactor_new` treats rhs3 as a non-empty input, and thus the default value for the C++ function `newReactor` is not used. The name is set to an empty string instead.

interfaces/matlab_experimental/Base/Interface.m
ssun30: The current inheritance structure is borrowed from the cython interface. Inheriting from `Solution` will cause the `Solution` class constructor to create a `ThermoPhase` and `Kinetics` object using the `solnID` of the `Solution` class instead of the `interfaceID` of the `Interface` class. As a result, the `Interface` object will be pointing to the wrong `ThermoPhase` and `Kinetics` objects and return incorrect results. The workaround is setting the `tpID` and `kinID` of the `Interface` object later, but for memory safety I've set all object IDs except the `Transport` class to immutable.

ischoegl: Both `Interface` and `Solution` are stored in the same `SolutionCabinet`, so the `solnID` should reference the correct object. You should not require `interfaceID`. I'm not sure how an object that should be correctly loaded in C++ ends up scrambled in clib? If there's an issue, it needs to be fixed in clib rather than having a workaround being created in MATLAB. Fwiw, `Interface` inherits from `Solution` in C++ and the same should be true in MATLAB (I believe there are technical reasons why it's implemented differently in Cython).

interfaces/matlab_experimental/Base/Water.m
ischoegl: I believe that all interfaces should be consistent, even if this means modifying cython as well.

interfaces/matlab_experimental/Base/Mixture.m
ischoegl: If there's not one implemented in C++, it should be. That way, we avoid redundant implementations.

interfaces/matlab_experimental/Reactor/Reactor.m
ischoegl: > I agree this could definitely be streamlined. Going forward, I will be using MATLAB's [inputParser](https://www.mathworks.com/help/matlab/ref/inputparser.html) to handle all `varargin` cases. I will separate this part of the work to a different PR.

Going away from the `nargin` checks could be a step in the right direction. Unfortunately, the syntax appears to be extremely clunky.

> As for setting the default name, when I pass an empty `cstring` type pointer from MATLAB to clib, `reactor_new` treats rhs3 as a non-empty input, and thus the default value for the C++ function `newReactor` is not used. The name is set to an empty string instead.

This may be due to passing an incorrect length? If you pass `" "` rather than `""` to C++ these are different. We have introduced `ReactorBase::setDefaultName` not too long ago, which should replace `(none)`. If `(none)` persists, there is a bug in the interfce.

ssun30: > > I agree this could definitely be streamlined. Going forward, I will be using MATLAB's [inputParser](https://www.mathworks.com/help/matlab/ref/inputparser.html) to handle all `varargin` cases. I will separate this part of the work to a different PR.
> 
> Going away from the `nargin` checks could be a step in the right direction. Unfortunately, the syntax appears to be extremely clunky.

This seems like a much [cleaner approach](https://www.mathworks.com/help/matlab/ref/arguments.html) to handle variable arguments and easier for code generation. We can essentially make all input arguments optional and let C++ set all the default values and handle error messages without any interference from the MATLAB side.

ssun30:
For several tests I added some new data files. The reason these is that the MATLAB toolbox only accepts yaml files as valid inputs but not yaml strings. I wonder how that's implemented for Python? 

Technically MATLAB could `fprint` yaml strings as a temporary file first and use that as the input. But that seems a bit clunky. There are some yaml parsers for MATLAB on the [File Exchange](https://www.mathworks.com/matlabcentral/fileexchange/106765-yaml/) but I don't think relying on a community toolbox that may not be maintained in the future is a good strategy. Do we want to have an in-house yaml writer for MATLAB for future releases? 

For now I would say we stay with yaml files only and keep those data files. Should I move them into the `test\matlab_experimental` folder instead of `test\data`, just to make sure we remove those in the future after yaml strings are supported?

ischoegl:
Hi @ssun30 … thanks for your continued work on this. I do, however, have one request: the PR is quickly approaching 5000 lines which is not practical for a review. Could you please wrap up one portion, and start a second one so this becomes manageable?

ssun30:
> Hi @ssun30 … thanks for your continued work on this. I do, however, have one request: the PR is quickly approaching 5000 lines which is not practical for a review. Could you please wrap up one portion, and start a second one so this becomes manageable?

Sure! I will separate the tests and fixes for Reactor, FlowDevice, and ReactorNet into a second PR, since these are the longest ones. 

Is there a reason why all our recent CI runs fail at the save wheel step for the `ubuntu-multiple-pythons` job?

ischoegl:
> Is there a reason why all our recent CI runs fail at the save wheel step for the `ubuntu-multiple-pythons` job?

I unfortunately don't know. @bryanwweber ... do you have any insights on this?

bryanwweber:
I think Ray just fixed this, you can rebase and run the checks again.

interfaces/matlab_experimental/Base/Solution.m
ischoegl: This remains to be addressed.

interfaces/matlab_experimental/Base/Mixture.m
ischoegl: I won't insist on this pre-existing condition.

interfaces/matlab_experimental/Utility/ImportPhases.m
ischoegl, line 1: I would recommend avoiding utilities that are not available at other APIs. If you want to use it for your test suite, it should be moved there.

interfaces/matlab_experimental/Utility/ctString.m
```diff
@@ -4,7 +4,7 @@
 
     err1 = -1;
 
-    buflen = calllib(ctLib, funcName, varargin{:}, 0, '');
+    buflen = calllib(ctLib, funcName, varargin{:}, 0, '') + 1;
```
ischoegl, line 8: This fix will no longer be necessary in the experimental clib; please add a comment that this is temporary.

test/data/invalid-inputs.yaml
ischoegl, line 1: Ideally, all APIs should use similar testing data; I would not be surprised if the existing Python test suite already has very similar entries. 

As you may have noticed, there are already a large number of input files. In order to avoid further proliferation, add to one of the existing files rather than adding new ones. I am asking for this as we need to maintain these input data long term.

test/data/undeclared-tests.yaml
ischoegl, line 1: See comment for `invalid-inputs.yaml`

test/matlab_experimental/ctTestDuplicateReactions.m
```diff
@@ -0,0 +1,101 @@
...
+        end
+
+        function testNonreactingSpecies(self)
+            self.check('K');
+            self.verifyEqual(self.phase.nReactions, 3);
+        end
+
+    end
+
+end
```
ischoegl: Missing empty line at file end (there are others as well). You can likely avoid this by updating your editor settings so this is done automatically when saving.

interfaces/matlab_experimental/readme.md
ischoegl, line 1: While I am specifically not asking to add this to the PR, it is probably time to move this file to the documentation, so it becomes visible outside of this folder.

My suggestion for the next PR would be to start thinking about how to integrate things into the `scons` installation process. You may glean some ideas from what recently added for the `clib_experimental` flag in [2a43239](https://github.com/Cantera/cantera/commit/2a432397f4766cbccde31e12dc169fdc87aea350)

.github/workflows/main.yml
```diff
@@ -21,11 +21,10 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  matlab-test:
-    name: Run MATLAB Test Script
+  matlab-test-ubuntu:
```
ischoegl, line 24: As mentioned elsewhere, I believe the Matlab CI runner should use a similar stategy as the one used by the .NET runner (use artifacts from a previous compilation so things can be checked on multiple platforms). 

I am not asking for this to be added here, but it should be very high on the list of priorities for the next PR.

ischoegl (dismissed):
@ssun30 ... thank you for tackling this. Most of my comments are quite minor.

The main feedback I have is the following:
1. Please avoid introducing new input files for `test/data`. I am fairly certain that the Python test suite covers very similar situations. Having to maintain redundant input makes little sense to me.
1. Does MATLAB provide for any hooks that we can use to check code coverage? For Python/C++ we can see what is covered and what is not. One important thing here would be to make sure that this is added to the CI/Coverage GitHub runner as soon as possible.
1. I believe that updating the testing infrastructure (integration with `scons`, CI runs and initial documentation) is now one of the highest priorities.

From where I stand, the single-most important thing about unit tests is that code is covered. I probably won't go through details of individual tests - I'd prefer to simply look at the code coverage percentage, which is a metric on how much could potentially break without us noticing.

PS: one thing I believe I recall from the workshop is that the tutorials can now be written as regular `.m` files (rather than livescripts). ~My preference would be to add these to the website rather than the samples folders, i.e., in the `doc` folder.~ _Edit: Reconsidering - as long as we can avoid binary files, adding tutorials to the sample folder may become an option again. For Python, we have CI runs that test all examples; you may want to introduce the same for MATLAB?_

interfaces/matlab_experimental/readme.md
```diff
@@ -12,25 +12,22 @@ function calling from Cantera CLib.
...
+   `matlab -softwareopengl`.
+5. In the Matlab command window, run
    `addpath(genpath([pwd, '/interfaces/matlab_experimental']))` to add search path for
    the experimental toolbox.
-5. In the Matlab command window, run
+6. In the Matlab command window, run
    `cd([pwd, '/interfaces/matlab_experimental/Utility'])` to navigate to the Utility
    folder.
-6. Open the file named 'ctRoot.m', in the second line, edit `output=` to
+7. Open the file named 'ctRoot.m', in the second line, edit `output=` to
```
ischoegl, line 28: Could you please add a comment on how to run the test suite to the usage guide below?

ssun30:
> 2. Does MATLAB provide for any hooks that we can use to check code coverage? For Python/C++ we can see what is covered and what is not. One important thing here would be to make sure that this is added to the CI/Coverage GitHub runner as soon as possible.

@ischoegl Coverage is now added to CI. As for adding new input files, I mentioned this earlier:

> For several tests I added some new data files. The reason these is that the MATLAB toolbox only accepts yaml files as valid inputs but not yaml strings. I wonder how that's implemented for Python?

> Technically MATLAB could fprint yaml strings as a temporary file first and use that as the input. But that seems a bit clunky. There are some yaml parsers for MATLAB on the [File Exchange](https://www.mathworks.com/matlabcentral/fileexchange/106765-yaml/) but I don't think relying on a community toolbox that may not be maintained in the future is a good strategy. Do we want to have an in-house yaml writer for MATLAB for future releases?

For now I would say we temporarily keep those files, but isolate them in the test\matlab_experimental folder instead of test\data, just to make sure we remove those in the future after yaml strings are supported.

> PS: one thing I believe I recall from the workshop is that the tutorials can now be written as regular `.m` files (rather than livescripts). ~My preference would be to add these to the website rather than the samples folders, i.e., in the `doc` folder.~ _Edit: Reconsidering - as long as we can avoid binary files, adding tutorials to the sample folder may become an option again. For Python, we have CI runs that test all examples; you may want to introduce the same for MATLAB?_

MATLAB natively supports direct conversion from `.m` scripts to Live Scripts. All we need to do is dividing the code sections with the double percent sign similar to Jupytext notebooks. This should be a trivial task and addressed in another PR. As for running the examples in a CI, this would come after the fix to the ILP64/LP64 incompatibility issue on Linux since some examples are broken by it. This would be my highest priority apart from completing the test suite.

test/data/invalid-inputs.yaml
ischoegl: @ssun30 ... thanks for acting on the suggestion, but I believe that moving things out of `test/data` is actually worse than adding another input file here. I.e., please leave things here. Likely, the proliferation of files should be addressed (eventually) by some other means.

I believe there was a question regarding text input instead of files. On Python, this involves `AnyMap` objects, which, however, are not available in CLib yet.

ischoegl:
@ssun30 ... thank you for adding this to the coverage run!

> For several tests I added some new data files. The reason these is that the MATLAB toolbox only accepts yaml files as valid inputs but not yaml strings. I wonder how that's implemented for Python?

The Python API uses `AnyMap` for this, which isn't available for CLib (I am not sure when it would be, as handling these objects would require a different paradigm).

> Technically MATLAB could fprint yaml strings as a temporary file first and use that as the input. But that seems a bit clunky. There are some yaml parsers for MATLAB on the [File Exchange](https://www.mathworks.com/matlabcentral/fileexchange/106765-yaml/) but I don't think relying on a community toolbox that may not be maintained in the future is a good strategy. Do we want to have an in-house yaml writer for MATLAB for future releases?

I am very much :-1: on this idea. It's redundant and would add unnecessary maintenance. Outside of `ck2yaml` (and other conversion scripts) all YAML parsing is done in C++. If you want to parse YAML strings in C++ you'd have to create new functions/constructors for `Solution` that add this capability to the core, and then expose this at the CLib interface.

> For now I would say we temporarily keep those files, but isolate them in the test\matlab_experimental folder instead of test\data, just to make sure we remove those in the future after yaml strings are supported.

I think that moving things out of `test/data` is worse than adding new files there.

Beyond, I am holding off on other comments until open questions in Cantera/enhancements#226 are addressed. It is unclear to me at the moment whether the plan would be to move MATLAB out of this repo (in which case any further review work here would be moot).

ssun30:
> The Python API uses `AnyMap` for this, which isn't available for CLib (I am not sure when it would be, as handling these objects would require a different paradigm).
> I think that moving things out of `test/data` is worse than adding new files there.

Would you suggest keeping those data for now, or skip those tests until Clib gets its equivalent of `AnyMap`?

> Beyond, I am holding off on other comments until open questions in [Cantera/enhancements#226](https://github.com/Cantera/enhancements/issues/226) are addressed. It is unclear to me at the moment whether the plan would be to move MATLAB out of this repo (in which case any further review work here would be moot).

Sorry for the confusion in that post. The intention was never to move MATLAB out of the main repo, but to have a separate package that users can download from File Exchange. MCT suggested having a separate repo so that the File Exchange package is always in sync with GitHub, but it's not a requirement.

ischoegl:
> > The Python API uses `AnyMap` for this, which isn't available for CLib (I am not sure when it would be, as handling these objects would require a different paradigm).
> > I think that moving things out of `test/data` is worse than adding new files there.
> 
> Would you suggest keeping those data for now, or skip those tests until Clib gets its equivalent of `AnyMap`?

Yes, please keep those data in `test/data` (sorry about the back-and-forth). Beyond, there's no guarantee for `AnyMap` coming to CLib anytime soon (after stumbling upon a related issue in #1863 I decided that this will be beyond the scope of Cantera/enhancements#220). For YAML strings that specify `Solution` objects, there is a much simpler solution, i.e., add new constructors that don't require the existence of a file.

ischoegl (commented):
Thanks for putting the test data back into the original location. I don't have any further requests but would like to have some second opinions before this is merged.

ischoegl (approved):
I am approving from my side. I will still wait a couple of days before the merge in case anyone else would like to comment.

PR #1677, opened 2024-03-14
add macOS arm runner

bryanwweber:
- Add a runner for `macos-14` which supports Apple Silicon (ARM) architecture. In this build, Python is installed from the GitHub `setup-python` action
- Remove Python 3.10 test for macOS
- Shift the dotnet test to Python 3.11 so that downloaded libs are consistent, due to the previous point
- Add a post-merge runner for `macos-14` that installs Python 3.12 from Homebrew. This represents what I see as the most likely case for people to be using Python from Homebrew to compile Cantera. It's possible to use older Python versions to compile Cantera or use older Python versions to build the Python interface, but it's trickier and I don't think we should encourage it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,810 +19,860 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  matlab-test:
-    name: Run MATLAB Test Script
-    runs-on: ubuntu-latest
-    env:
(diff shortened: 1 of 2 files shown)


bryanwweber:
The post-merge tests have passed: https://github.com/Cantera/cantera/actions/runs/8300167316/job/22717482677

ischoegl (approved):
Thank you - this all looks good to me!

PR #1670, opened 2024-02-21
Remove legacy MATLAB toolbox

ischoegl:
**Changes proposed in this pull request**

This PR seeks to map out the removal of the legacy MATLAB toolbox

- [x] Remove `interfaces/matlab`
- [x] Remove documentation of legacy toolbox
- [x] Remove tests and samples
- [x] Remove MATLAB installation from build system
- [x] Remove references to MATLAB toolbox from new sphinx web theme

Items to be deferred:

- Remove `clib` functions that were only used by legacy MATLAB toolbox
- Bump development version number as this is a significant change

**If applicable, fill in the issue number this pull request is fixing**

Closes Cantera/enhancements#196

**If applicable, provide an example illustrating new features this pull request is introducing**

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

PR #1685, opened 2024-03-30
Remove/deprecate legacy Solution handling

ischoegl:

Removal of legacy MATLAB toolbox #1670 allows for some simplifications / deprecations:

- Remove remaining legacy wall methods from CAPI and C++
- Deprecate `setThermoMgr`/`setKineticsMgr` for 0D
- Move deprecation warning for empty reactor objects from Python to C++

Partially addresses #1457

_A clean resolution of #1457 will require pushing the Python `_WeakrefProxy` approach to the C++ layer?_

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/reactor.pyx
+++ b/interfaces/cython/cantera/reactor.pyx
@@ -23,8 +23,20 @@ cdef class ReactorBase:
     Common base class for reactors and reservoirs.
     """
     reactor_type = "none"
-    def __cinit__(self, *args, **kwargs):
-        self._reactor = newReactor(stringify(self.reactor_type))
+    def __cinit__(self, _SolutionBase contents=None, name=None, *, **kwargs):
+        def reactor_name(name):
+            if name is not None:
+                return name
+            _reactor_counts[self.reactor_type] += 1
+            return f"{self.reactor_type}_{_reactor_counts[self.reactor_type]}"
+
+        if isinstance(contents, _SolutionBase):
+            self._reactor = newReactor(stringify(self.reactor_type),
+                                       contents._base, stringify(reactor_name(name)))
+        else:
+            # deprecated: will raise warnings in C++ layer
+            self._reactor = newReactor(stringify(self.reactor_type))
+            self._reactor.get().setName(stringify(reactor_name(name)))
         self.rbase = self._reactor.get()
 
     def __init__(self, _SolutionBase contents=None, name=None, *, volume=None,
@@ -36,18 +48,7 @@ cdef class ReactorBase:
         self._walls = []
         self._surfaces = []
         if isinstance(contents, _SolutionBase):
-            self.insert(contents)
-        else:
-            warnings.warn(
-                "Starting in Cantera 3.1, the reactor contents must not be empty.",
-                DeprecationWarning)
-
-        if name is not None:
-            self.name = name
-        else:
-            _reactor_counts[self.reactor_type] += 1
-            n = _reactor_counts[self.reactor_type]
-            self.name = '{0}_{1}'.format(self.reactor_type, n)
+            self.insert(contents)  # leave insert for the time being
 
         if volume is not None:
             self.volume = volume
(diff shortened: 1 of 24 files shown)


ischoegl:
Ready for review ... decided to restrict this PR to *Reactor* cleanup that became possible after the removal of the legacy MATLAB toolbox.

interfaces/cython/cantera/reactor.pyx
```diff
@@ -23,8 +23,20 @@ cdef class ReactorBase:
...
+            _reactor_counts[self.reactor_type] += 1
+            return f"{self.reactor_type}_{_reactor_counts[self.reactor_type]}"
+
+        if isinstance(contents, _SolutionBase):
+            self._reactor = newReactor(stringify(self.reactor_type),
+                                       contents._base, stringify(reactor_name(name)))
+        else:
+            # deprecated: will raise warnings in C++ layer
+            self._reactor = newReactor(stringify(self.reactor_type))
+            self._reactor.get().setName(stringify(reactor_name(name)))
```
speth, line 39: Is the second branch here (and having a default value for `contents` even needed? In Cantera 3.0, calling `ct.Reactor()` already issues the following warning:
```
DeprecationWarning: Starting in Cantera 3.1, the reactor contents must not be empty.
  r = ct.Reactor()
```
So I think at least for the Python interface, it's acceptable to remove the option not to specify the contents as part of the constructor.

src/zeroD/ReactorBase.cpp
```diff
@@ -51,6 +51,20 @@ void ReactorBase::setThermoMgr(ThermoPhase& thermo)
     m_pressure = m_thermo->pressure();
 }
 
+void ReactorBase::setThermoMgr(ThermoPhase& thermo)
+{
+    warn_deprecated("ReactorBase::setThermoMgr",
+        "To be removed after Cantera 3.1. Replaceable by 'setThermo'.");
+    setThermo(thermo);
```
speth: As a `protected` method, `setThermo` isn't really a replacement -- the replacement is to specify the phase when constructing the reactor. Likewise for `setKineticsMgr`.

src/zeroD/ReactorFactory.cpp
```diff
@@ -68,7 +68,8 @@ void ReactorFactory::deleteFactory() {
 
 shared_ptr<ReactorBase> newReactor(const string& model)
 {
-    // deprecation warning is handled in Python API
+    warn_deprecated("newReactor",
+        "Starting in Cantera 3.1, the reactor contents must not be empty.");
```
speth:
```suggestion
        "Starting in Cantera 3.1, the reactor contents must be provided in the constructor.");
```

speth (changes requested):
Thanks, @ischoegl. I had a couple of thoughts on this. 

Given the changes made here, can we move toward eliminating the C++ `ReactorBase::setSolution` and the Python `ReactorBase.insert` methods? With the requirement that the reactor needs contents from the beginning, I don't think there's a use case for switching them after the fact. (Not necessarily suggesting to resolve this in this PR)

src/zeroD/ReactorFactory.cpp
ischoegl: I think the original message introduced in #1663 was not ideal; non-empty reactor contents are being deprecated in 3.1, and removed thereafter.

interfaces/cython/cantera/reactor.pyx
ischoegl: We introduced this requirement in 3.1 via #1663, i.e. the warning should be updated.

ischoegl:
@speth - thank you for your feedback. I believe I started some confusion with a deprecation warning that incorrectly implied that something was introduced in Cantera 3.0, while only being implemented after its release (non-empty Reactors). I updated/corrected the warnings, which should take care of your concerns.

Regarding changes to the reactor content after instantiation, we can definitely think about removing this capability, but either `insert` or `setSolution` are needed until the deprecation cycle is complete. I agree that any changes should be discussed in a separate PR.

speth (approved):
Thanks, @ischoegl. These changes work for me.

PR #1686, opened 2024-03-31
Implement ThermoPhase species locks in C++

### LINKED ISSUE #1457: addSpecies should be disabled on ThermoPhase objects in use by reactors and flames

ischoegl:

Move sentinels preventing species addition for `ThermoPhase` objects used by `ReactorBase`, `Domain1D`, `SolutionArray` and `MultiPhase` instances from Python API to C++.

The implementation uses reference counters that should only be used within C++ (and not exposed at interfaces). This allows for most `_WeakrefProxy` sentinels to be be removed in Python. One exception is `Quantity`, which does not have a C++ equivalent.

The approach is relatively simple, and turns out to be independent of migrating various C++ classes to use `Solution`.

Closes #1457

**Checklist**

### DIFF
--- a/include/cantera/base/SolutionArray.h
+++ b/include/cantera/base/SolutionArray.h
@@ -39,7 +39,7 @@ private:
     SolutionArray(const SolutionArray& arr, const vector<int>& indices);
 
 public:
-    virtual ~SolutionArray() {}
+    virtual ~SolutionArray();
 
     /**
      *  Instantiate a new SolutionArray reference.
--- a/include/cantera/equil/MultiPhase.h
+++ b/include/cantera/equil/MultiPhase.h
@@ -65,7 +65,7 @@ public:
 
     //! Destructor. Does nothing. Class MultiPhase does not take "ownership"
     //! (that is, responsibility for destroying) the phase objects.
-    virtual ~MultiPhase() = default;
+    virtual ~MultiPhase();
 
     //! Add a vector of phases to the mixture
     /*!
--- a/include/cantera/oneD/Boundary1D.h
+++ b/include/cantera/oneD/Boundary1D.h
@@ -186,7 +186,7 @@ public:
     Empty1D() = default;
 
     Empty1D(shared_ptr<Solution> solution, const string& id="") : Empty1D() {
-        m_solution = solution;
+        setSolution(solution);
         m_id = id;
     }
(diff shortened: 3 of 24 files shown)


speth (approved):
Thanks, @ischoegl. This looks good to me. The reference counting approach is straightforward and effective.

PR #1706, opened 2024-06-08
Fix Numpy 2.0 compatibility issues

### LINKED ISSUE #1675: Upcoming breaking changes in Numpy 2.0
NumPy 2.0 hasn't been released yet, but the [initial release notes](https://github.com/numpy/numpy/releases/tag/v2.0.0b1) from the pre-release mentions some breaking changes that they address in their [migration guide](https://numpy.org/devdocs/numpy_2_0_migration_guide.html). I don't know how the Cantera team wants to handle compatibility going forward, but I thought I would go ahead and make a thread in case it's helpful.

# Previous Cantera builds

I installed NumPy 2.0.0b1 and attempted to import Cantera 3.0.0 which yields the following error:

```
...
    from ._cantera import *
  File "cantera/_cantera.pyx", line 33, in init cantera._cantera
  File "cantera/_utils.pyx", line 1, in init cantera._utils
ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```

Clearly previous builds of Cantera can only be used with NumPy 1.x and will not work with 2.x releases. I don't know if there is a way to enforce `numpy<2.0.0` as a package requirement on PyPI, or if that can't be done retroactively, but it might need to be added to some documentation somewhere.

# Future Cantera builds

I attempted to build Cantera from source from the latest commit using NumPy 2.0.0b1 to see what changes would need to be made. The first build error that I get from the breaking changes is below:

```
build\python\cantera\_onedim.cpp(8544): error C2039: 'subarray': is not a member of '_PyArray_Descr'
...
build\python\cantera\_onedim.cpp(8545): error C2039: 'subarray': is not a member of '_PyArray_Descr'
```

The changes to the `PyArray_Descr` struct are covered in [this section](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#the-pyarray-descr-struct-has-been-changed) of the migration guide. It seems the `PyDataType_SUBARRAY` accessor function needs to be used. They also provide a `npy2_compat.h` file that can be included in the code base that will allow compilation on NumPy 1.x and not affect 2.x builds.

I haven't gotten any further yet, but I will edit this if I do.

speth:

Fix usage of functions that were removed or deprecated in NumPy 2.0.

We can make sure that at least some of the CI tests are using NumPy 2.0 once the official release is made on June 16 (https://github.com/numpy/numpy/issues/24300), and then merge this PR.

Partially resolves #1675. I think the issues with matching versions of NumPy for packaged versions of Cantera will need to be resolved in the various packaging scripts, but we can leave the issue open until those changes are made.

**Checklist**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -450,8 +450,10 @@ jobs:
           find samples/python -type f -iname "*.py" \
           -exec sh -c 'for n; do echo "$n" | tee -a results.txt && python3 "$n" >> results.txt || exit 1; done' sh {} +
         env:
-          # The ignore setting here is due to a new warning introduced in Matplotlib==3.6.0
-          PYTHONWARNINGS: "error,ignore:warn_name_set_on_empty_Forward::pyparsing,ignore:datetime.datetime.utcfromtimestamp:DeprecationWarning:"
+          # The pyparsing ignore setting is due to a new warning introduced in Matplotlib==3.6.0
+          # @todo: Remove the trapz-related ignore when dropping support for NumPy 1.x
+          #     and replacing np.trapz with np.trapezoid
+          PYTHONWARNINGS: "error,ignore:warn_name_set_on_empty_Forward::pyparsing,ignore:datetime.datetime.utcfromtimestamp:DeprecationWarning:,ignore:`trapz`:DeprecationWarning"
           MPLBACKEND: Agg
       - name: Save the results file for inspection
         uses: actions/upload-artifact@v4
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,8 +39,8 @@ jobs:
       - name: Install Apt dependencies
         run: |
           sudo apt update
-          sudo apt install libboost-dev gfortran libopenmpi-dev libpython3-dev \
-          libblas-dev liblapack-dev libhdf5-dev
+          sudo apt install libboost-dev gfortran libopenmpi-dev libblas-dev \
+            liblapack-dev libhdf5-dev
           gcc --version
       - name: Upgrade pip
         run: python3 -m pip install -U pip setuptools wheel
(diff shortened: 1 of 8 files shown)


bryanwweber:
Is this backwards compatible? In other words, will Cantera 3.1 require NumPy >=2.0, or can we support 1.x and 2.x simultaneously?

speth:
This is meant to be backwards compatible, at least in terms of the source -- all the changes for us are pretty minor. I think the conda and PyPI packages will have to be specific to one version or another. I was hoping for your help in figuring some of that out. 😁

bryanwweber:
> I think the conda and PyPI packages will have to be specific to one version or another. I was hoping for your help in figuring some of that out. 😁

Yeah, the NumPy headers are different, right? I haven't looked into how that's meant to be handled. I'll see if there's anything out there, I assume NumPy have gotten out ahead of this.

speth:
This may be easier than I thought. According to the [NumPy Docs](https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-0-specific-advice):
> When you build wheels for your package using a NumPy 2.x version at build time, those **will work** with NumPy 1.xx.

interfaces/cython/pyproject.toml
```diff
@@ -1,3 +1,21 @@
...
+    "license",
+    "optional-dependencies",
+    "readme",
+    "scripts",
+    "urls",
+    "version",
+]
+dependencies = [
+    "numpy>=1.19.3",
+]
```
bryanwweber: I think this should be coming from `setup.cfg`, is it not?

speth: Just trying things. I'm not finding the supposed ABI compatibility of the compiled module with older NumPy versions.

bryanwweber: I'll experiment with this tomorrow too

speth: Well, from the latest runs, it seems like backwards compatibility is working as expected when building with NumPy 2.0.0. The problem is with the builds for Python 3.8 using NumPy 1.24 and then running with NumPy 1.19. And now I've finally read [this page](https://numpy.org/devdocs/dev/depending_on_numpy.html#adding-a-dependency-on-numpy) enough times to understand that this is expected not to work and we have to go the other way around in this case.

bryanwweber: Ok, got it. The dependency, at least in the sdist at wheel build time, is that "oldest-supported-numpy" option they mention. But good to know that it needs to depend on the Python versions that NumPy supports

speth: Yep. It sounds like this _should_ be a lot easier going forward -- a build-time dependency on NumPy 2.x will provide backwards compatibility with NumPy 1.x and also forward compatibility with future 2.x releases. If we want to get ahead of things, I suppose we could set the version requirement to `numpy<3`.

bryanwweber:
One tricky thing I found here is that Pint 0.24 supports Python 3.10 and up, and only Pint >=0.24 supports NumPy 2.0. https://github.com/hgrecco/pint/commit/749d77c69abc755b374553fc7152df1c0a89e557 for the Python 3.10 stuff and I'm having test failures on Python 3.9 with cibuildwheel because it's installing 0.23.2 of Pint and 2.0 of NumPy.

I think the solution is to edit the `setup.cfg` to force `numpy<2.0` if the units extra is selected and Python is < 3.9

bryanwweber (approved):

bryanwweber:
Thanks @speth I'm going to merge this and move any further work to #1713

PR #1710, opened 2024-06-12
Incorporating the LMR-R mixture rule into Cantera for the first time

pjsingal:

- LinearBurkeRate.cpp and LinearBurkeRate.h allow for vastly improved consideration of the mixture-dependent aspect of the rate constants for complex-forming ("pressure-dependent") reactions, using the new reduced-pressure linear mixture rule (LMR-R)

- This code builds, runs, and produces results that have been validated and peer-reviewed 

- LMR-R can handle three different representations of the pressure-dependent aspect of complex-forming reactions: Troe, PLOG, and Chebyshev

- LMR-R calculates the mixture-dependent aspect using temperature-dependent third-body efficiencies that have been supplied by the user for at least one collider - these parameters can be provided in either relative or absolute terms, as long as the user's choice is consistent for all colliders in a given reaction

- Description of the theory and instructions for YAML implementation have been added to rate-constants.md and reactions.rst, respectively

- Three new Python examples have been added: jet_stirred_reactor.py, shock_tube.py, flame_speed.py

- The Python examples employ a mechanism containing linear-Burke reactions, ammonia-CO-H2-Alzueta-2023.yaml, which is being added to Cantera/cantera-example-data via a separate PR

- Unit tests have been added, along with the linearBurke-test.yaml mechanism, stored in Cantera/Cantera/test/data

**Checklist**

[x] The pull request includes a clear description of this code change
[x] Commit messages have short titles and reference relevant issues
[x] Build passes (scons build & scons test) and unit tests address code coverage
[x] Style & formatting of contributed code follows contributing guidelines
[x] The pull request is ready for review


.gitignore
```diff
@@ -48,4 +48,4 @@ environment.y*
 *.gch
 coverage/
 coverage.info
-.coverage
+.coverage
```
ischoegl: Please undo this change - there should be a newline at the end of each file.

data/alzueta_LMRR.yaml
```diff
@@ -0,0 +1,1808 @@
...
+  ************************************************************************
+  ******************* For further documentation of mechanism:  ***********
+  ******************* see references below and their citations ***********
+  ************************************************************************
+  ************************************************************************
+  Nitrogen subset
+  P. Glarborg, J.A. Miller, B. Ruscic, S.J. Klippenstein,
+  "Modeling Nitrogen Chemistry in Combustion",
+  Prog. Energy Combust. Sci. 67 (2018) 31-68.
+  Updates marked by #
```
ischoegl: I assume that this is the file that you changed to illustrate LMR - please provide a clear description.

data/alzueta_LMRR.yaml
```diff
@@ -0,0 +1,1808 @@
...
+units: {length: cm, time: s, quantity: mol, activation-energy: cal/mol}
+
+phases:
+- name: gas
+  thermo: ideal-gas
+  elements: [O, H, C, N, He, Ar]
+  species: [NH3, NO2, NO, N2O, H2, O2, O3, H, O, OH, HO2, H2O, H2O2, CO,
+    CO2, HOCO, CH2O, HCO, NH2, NH, N, NNH, N2H4, N2H3, tHNNH, cHNNH, H2NN,
+    NH2OH, H2NO, HNOH, HNO, HON, HONO, HNO2, NO3, HONO2, H2NCO, HNCO, NCO,
+    AR, HE, N2]
```
ischoegl: It appears that there aren't any differences in species between the two versions of the file; in this case, you can actually import things from`alzueta.yaml` and shorten this file by around 900 lines (which would also emphasize what you are changing). For an example, have a look at `gri30_ion.yaml`.

data/example_data
ischoegl, line 1: Please revert this change.

ext/googletest
ischoegl, line 1: please revert this change.

samples/python/kinetics/flame_speed.py
```diff
@@ -0,0 +1,82 @@
...
+colours = ["xkcd:grey",'xkcd:purple']
+
+fuel_list = np.linspace(0.14,0.4,20) # mole fractions of fuel to simulate across
+alpha = 1.0
+a_st = 0.75
+Tin = 296  # unburned gas temperature [K]
+p=760
+
+for k, m in enumerate(models):
+  vel_list = []
```
ischoegl: Indentation levels should use increments of 4 whitespaces.

samples/python/kinetics/flame_speed.py
```diff
@@ -0,0 +1,82 @@
...
+X_NH3 = np.divide(expData['X_NH3'],100)
+X_O2 = np.multiply(np.subtract(1,X_NH3), 0.21)
+phi_data = np.divide(np.divide(X_NH3,X_O2),np.divide(4,3))
+plt.plot(phi_data,expData['vel'],'o',fillstyle='none',color='k',label='Ronney')
+plt.legend(fontsize=8,frameon=False, loc='lower right')
+plt.ylabel(r'Burning velocity [cm $\rm s^{-1}$]')
+plt.xlabel(r'Equivalence Ratio')
+plt.xlim([0.6, 2.1])
+plt.ylim([0, 12])
+plt.show()
```
ischoegl: Please add a single newline at the end of each file.

samples/python/kinetics/jet_stirred_reactor.py
```diff
@@ -0,0 +1,155 @@
...
+## Uncomment this code to plot the time history instead of T-dependence
+# for k,m in enumerate(models):
+#     reactants = {'H2': 0.03, 'O2': 0.03, 'Ar': 0.846, 'NH3':0.094}
+#     gas = ct.Solution(list(models.values())[k])
+#     gas.TPX = reactorTemperature, reactorPressure*ct.one_atm, reactants
+#     timeHistory = getTimeHistory(gas)
+#     ax[0].plot(timeHistory.index*1e3, np.subtract(timeHistory['temperature'],np.ones(len(timeHistory['temperature']))*reactorTemperature), color=colours[k])   
+#     ax[1].plot(timeHistory.index*1e3, timeHistory['O2']*100, color=colours[k])   
+#     ax[2].plot(timeHistory.index*1e3, timeHistory['H2']*100, color=colours[k])
+# plt.show()
```
ischoegl: Either remove or enable.

ischoegl (commented):
Thanks for this contribution, @pjsingal! I have a couple of basic recommendations that mostly pertain to formatting. The example data should likely move to the new `example_data` repository, but it is not urgent at the moment. You may also want to think about adding unit tests.

data/alzueta.yaml
speth, line 1: This mechanism is a good candidate for moving to the separate `example_data` repository (https://github.com/Cantera/cantera-example-data).

I appreciate the effort to at least share the species data between the two versions of the mechanism, but I'd like to suggest going a step further, and sharing the reactions that they have in common as well. To do this, you would need to divide the reactions into three sections in the input file: one for the shared reactions (named for example `common`), one for the baseline mechanism's pressure-dependent reactions (named for example `baseline-pdep`), and one for the reactions implemented using the `linear-Burke` rate model (named `linear-Burke`). Then, you can put two phase definitions in the file, one which specifies the reactions as `reactions: [common, baseline-pdep]` and the other that specifies the reactions as `reactions: [common, linear-Burke]`.

samples/python/kinetics/ignition_delay_time.py
speth, line 1: There are several other examples showing IDT as a function of temperature, so I think we probably don't need this example. Your paper is the main place for demonstrating this result.

samples/python/kinetics/flame_speed.py
```diff
@@ -0,0 +1,78 @@
...
+reduced-pressure linear mixture rule (LMR-R) in such calculations.
+
+Here we will consider a mixture of NH3/air at 1 atm and 296 K, and compare results against the
+experimental measurements of Ronney [1]
+
+Two models are compared in this example:
+(i)  A 2023 model of H2 and NH3 chemistry published by Alzueta et al. [2]
+(ii) An adapted version of this model that has applied the reduced-pressure linear mixture
+     rule (LMR-R) and ab initio third-body efficiencies [3]
+
```
speth: Please make sure that this header can be parsed as [reStructuredText](https://docutils.sourceforge.io/rst.html) and that the output looks like what you are expecting. You can take a look at the generated HTML by downloading the "docs" artifact from the CI builds (go to the "Checks" tab at the top of this PR page, then look through the list of "artifacts").

Also, please wrap content, including the comments, to no more than 88 characters.

include/cantera/kinetics/LinearBurkeRate.h
```diff
@@ -0,0 +1,140 @@
+//! @file LINEARBURKERATE.h
```
speth: The capitalization here should match the capitalization of the file name -- otherwise, Doxygen issues a warning.

include/cantera/kinetics/LinearBurkeRate.h
```diff
@@ -0,0 +1,140 @@
...
+
+namespace Cantera
+{
+
+//! Data container holding shared data specific to LinearBurkeRate
+/**
+ * The data container `LmrData` holds precalculated data common to
+ * all `LinearBurkeRate` objects.
+ */
+struct LmrData : public ReactionData
```
speth: The name of this class should also be changed to something less obscure -- `LinearBurkeData` would be consistent with the pattern that we use with the other rates and their corresponding data classes.

include/cantera/kinetics/LinearBurkeRate.h
```diff
@@ -0,0 +1,140 @@
...
+ * [ADD IN THE MATHEMATICAL FORMULA]
+ */
+class LinearBurkeRate final : public ReactionRate
+{
+public:
+    //! Default constructor.
+    LinearBurkeRate() = default;
+
+    //! Constructor from Arrhenius rate expressions at a set of pressures
+    explicit LinearBurkeRate(const std::multimap<double, ArrheniusRate>& rates);
```
speth: This constructor is declared, but not defined -- I think this is just copypasta from `PlogRate` and can be deleted.

doc/doxygen/cantera.bib
```diff
@@ -434,6 +434,17 @@ @online{smith1999
...
+    author = {P.~J.~Singal and J.~Lee and L.~Lei and R.~L.~Speth and M.~P.~Burke},
+    journal = {Proceedings of the Combustion Institute},
+    number = {40},
+    pages = {TBD},
+    title = {Implementation of New Mixture Rules Has a Substantial Impact on
+        Combustion Predictions for H2 and NH3},
+    doi = {TBD},
+    url = {TBD},
+    volume = {TBD},
+    year = {2024}}
```
speth: Given the usual publication schedule, I assume this should be 2025, which will also match your `@cite` annotations in the code.

doc/sphinx/reference/kinetics/rate-constants.md
```diff
@@ -235,6 +235,61 @@ Chebyshev reactions can be defined in the YAML format using the
...
 
+(sec-linear-burke)=
+## Linear Burke Rate Expressions
+
+Linear Burke rate expressions employ the reduced-pressure linear mixture rule (LMR-R). This mixture rule is used to evaluate the rate constants of complex-forming reactions, and is a mole-fraction-weighted sum of the bath gas rate constants (when pure) evaluated at the reduced pressure ($R$) and temperature ($T$) of the mixture.
+
+$$
+k_{\text{LMR-R}}(T,P,\textit{\textbf{X}}) = \sum_{i} k_{i}(T,R_{\text{LMR}})\tilde{X}_{i,\text{LMR}}
+$$
+
```
speth: Please look at the rendered HTML -- MathJax does not like this approach for getting a bold, italic symbol. I think you might want to try `\boldsymbol`.

doc/sphinx/reference/kinetics/rate-constants.md
```diff
@@ -235,6 +235,61 @@ Chebyshev reactions can be defined in the YAML format using the
...
+In fact, if $\Lambda_{0,i}(T)$ (or $\epsilon_{0,i}(T)$) for any colliders have available data or can be estimated using typical values (as is typically done in kinetic models for reactions in modified Lindemann expressions) but no data for $k_i(T,P)$ are available, this implementation also allows some colliders to be specified with unique $\Lambda_{0,i}(T)$ (or $\epsilon_{0,i}(T)$) without $k_i(T,P)$, by assuming the same reduced-pressure dependence as M (i.e. $k_{i}(T,R)=k_{M}(T,R)$). Cantera employs the following derived form of the generalized LMR-R equation, where the sum over $n$ is only for the colliders for which unique $k_n(T,P)$ are available. If unique $k_i(T,P)$ data are available for all colliders, then the second term in the above equation effectively disappears.
+
+$$
+k_{\text{LMR-R}}(T,P,\textit{\textbf{X}}) &= \sum_{n} k_{n}(T,P_{n,\text{LMR}}^{\text{ eff}})\tilde{X}_{n,\text{LMR}} + k_{M}(T,P_{M,\text{LMR}}^{\text{ eff}}) \left(1-\sum_{n}\tilde{X}_{n,\text{LMR}}\right)
+$$
+
```
speth: This equation does not render -- there is some syntax error.

doc/doxygen/cantera.bib
```diff
@@ -434,6 +434,17 @@ @online{smith1999
         and W.~C.~Gardiner, Jr. and V.~V.~Lissianski and Z.~Qin},
     url = {http://combustion.berkeley.edu/gri-mech/version30/text30.html},
     year = {1999}}
+@article{singal2024,
+    author = {P.~J.~Singal and J.~Lee and L.~Lei and R.~L.~Speth and M.~P.~Burke},
+    journal = {Proceedings of the Combustion Institute},
+    number = {40},
+    pages = {TBD},
+    title = {Implementation of New Mixture Rules Has a Substantial Impact on
+        Combustion Predictions for H2 and NH3},
```
speth:
```suggestion
        Combustion Predictions for {H2} and {NH3}},
```

doc/sphinx/reference/kinetics/rate-constants.md
```diff
@@ -235,6 +235,61 @@ Chebyshev reactions can be defined in the YAML format using the
 [`Chebyshev`](sec-yaml-Chebyshev) reaction `type`.
 ```
 
+(sec-linear-burke)=
+## Linear Burke Rate Expressions
+
+Linear Burke rate expressions employ the reduced-pressure linear mixture rule (LMR-R). This mixture rule is used to evaluate the rate constants of complex-forming reactions, and is a mole-fraction-weighted sum of the bath gas rate constants (when pure) evaluated at the reduced pressure ($R$) and temperature ($T$) of the mixture.
```
speth: Line wrap to 88 characters everywhere.

doc/sphinx/reference/kinetics/rate-constants.md
```diff
@@ -235,6 +235,61 @@ Chebyshev reactions can be defined in the YAML format using the
...
+
+In fact, if $\Lambda_{0,i}(T)$ (or $\epsilon_{0,i}(T)$) for any colliders have available data or can be estimated using typical values (as is typically done in kinetic models for reactions in modified Lindemann expressions) but no data for $k_i(T,P)$ are available, this implementation also allows some colliders to be specified with unique $\Lambda_{0,i}(T)$ (or $\epsilon_{0,i}(T)$) without $k_i(T,P)$, by assuming the same reduced-pressure dependence as M (i.e. $k_{i}(T,R)=k_{M}(T,R)$). Cantera employs the following derived form of the generalized LMR-R equation, where the sum over $n$ is only for the colliders for which unique $k_n(T,P)$ are available. If unique $k_i(T,P)$ data are available for all colliders, then the second term in the above equation effectively disappears.
```
speth: What "second term" in the above equation?

doc/sphinx/reference/kinetics/rate-constants.md
```diff
@@ -235,6 +235,61 @@ Chebyshev reactions can be defined in the YAML format using the
...
+$$
+
+such that an alternate version of the generalized LMR-R equation can be written as
+
+\begin{equation}
+k_{\text{LMR-R}}(T,P,\textit{\textbf{X}}) = \sum_{i} k_{i}(T,P_{i,\text{LMR}}^{\text{ eff}})\tilde{X}_{i,\text{LMR}}
+\label{eq:LMRR_k_P_i_eff}
+\end{equation}
+
+The user can either specify $k_i(T,P)$ and $\Lambda_{0,i}(T)$ for each collider, or specify the third-body efficiency, $\epsilon_{0,i}(T)$, for each non-M collider and assign $\epsilon_{0,\text{M}}(T)=1$.  The pressure-dependent aspect of each i-th collider ($k_i(T,P)$) can be specified in any combination of Troe, PLOG, or Chebyshev formats for colliders for which data are available and $\Lambda_{0,i}(T)$ for each collider (or $\epsilon_{0,i}(T)$ for each non-M collider) can be specified in modified Arrhenius format for colliders for which data are available.
```
speth: This discussion of $\epsilon$ is somewhat confusing, as there's no indication of how a value for $\epsilon$ is used in place of $k_i(T,P)$ and $\Lambda_{0,i}(T)$.

Is it necessary for the user to assign $\epsilon_{0,\text{M}}(T)=1$, or is that simply true by definition?

samples/python/kinetics/flame_speed.py
```diff
@@ -0,0 +1,78 @@
...
+Requires: cantera >= 3.1
+.. tags:: flame speed, kinetics, mixture rule
+"""
+
+import cantera as ct
+import matplotlib.pyplot as plt
+import pandas as pd
+import numpy as np
+
+plt.figure(figsize=(3.5,2.5))
```
speth: I would suggest leaving the default figure size than specifying a very small figure.

Also, I would recommend using the "object oriented" Matplotlib API, for example starting with `fig, ax = plt.subplots()` and going from there.

samples/python/kinetics/flame_speed.py
```diff
@@ -0,0 +1,78 @@
...
+    Has a Substantial Impact on Combustion Predictions for H2 and NH3, Proc. Combust. Inst. 40
+    (2024).
+
+Requires: cantera >= 3.1
+.. tags:: flame speed, kinetics, mixture rule
+"""
+
+import cantera as ct
+import matplotlib.pyplot as plt
+import pandas as pd
```
speth: Pandas isn't used in this example

samples/python/kinetics/flame_speed.py
```diff
@@ -0,0 +1,78 @@
...
+import numpy as np
+
+plt.figure(figsize=(3.5,2.5))
+
+models = {'Original':'alzueta.yaml','LMR-R':'alzueta_LMRR.yaml'}
+colours = ["xkcd:grey",'xkcd:purple']
+n=20 # number of points to simulate
+fuel_list = np.linspace(0.14,0.4,n) # mole fractions of fuel to simulate across
```
speth: I like this example a lot -- in addition to the kinetics comparison, we didn't previously have an example that plotted flame speed as a function of equivalence ratio. However, this example takes a long time to run, which really slows down building the docs. I have a few suggestions to speed things up at least a bit:

- Reduce the range of fuel compositions to just the range of your plot -- there are several points simulated at high equivalence ratio that are trimmed out of the figure in the end, and reduce the number of points in the curve a little bit, say to 16.
- Turn off the use of the multicomponent transport model and Soret diffusion, at least by default, as these require a lot of additional calculations. You could show them commented out, with an additional comment about their optional use
- Slightly relax the grid refinement criteria. I didn't see an appreciable difference in the results using slope=0.06, curve=0.10.

samples/python/kinetics/flame_speed.py
```diff
@@ -0,0 +1,78 @@
...
+a_st = 0.75 # coefficient of oxidizer for stoichiometric combustion
+Tin = 296  # unburned gas temperature [K]
+p=760
+
+for k, m in enumerate(models):
+    vel_list = []
+    phi_list = []
+    for j, fuel_frac in enumerate(fuel_list):
+        gas = ct.Solution(list(models.values())[k])
```
speth: You only need to instantiate the `gas` object in the outer loop. It can be re-used for each equivalence ratio.

samples/python/kinetics/flame_speed.py
```diff
@@ -0,0 +1,78 @@
...
+        NH3 = fuel_frac
+        ox_frac = 1 - fuel_frac # oxidizer fraction
+        O2 = ox_frac*0.21
+        N2 = ox_frac*0.79
+        phi = np.divide(fuel_frac/O2,1/a_st)
+        phi_list.append(phi)
+        X = {'NH3':NH3,'O2':O2,'N2':N2}
+        gas.TPX = Tin, (p/760)*ct.one_atm, X
```
speth: Rather than doing your own math on the stoichiometry, this should use Cantera's built in functionality for working with equivalence ratio -- you can define the desired equivalence ratio range as the input (instead of `fuel_frac`), then set the mixture composition with `gas.set_equivalence_ratio` (and separately set `gas.TP`).

samples/python/kinetics/flame_speed.py
speth, line 1: I think this example belongs in the "onedim" section, rather than the "kinetics" section.

samples/python/kinetics/shock_tube.py
```diff
@@ -0,0 +1,73 @@
...
+models = {'Original':'alzueta.yaml','LMR-R':'alzueta_LMRR.yaml'}
+colours = ["xkcd:grey",'xkcd:purple']
+
+for k,m in enumerate(models):
+    X_H2O2 = 1163e-6
+    X_H2O = 1330e-6
+    X_O2 = 665e-6
+    X_CO2= 0.2*(1-X_H2O2-X_H2O-X_O2)
+    X_Ar = 1-X_CO2
+    gas = ct.Solution(list(models.values())[k])
```
speth:
```suggestion
    gas = ct.Solution(models[m])
```

samples/python/kinetics/shock_tube.py
```diff
@@ -0,0 +1,73 @@
...
+
+Requires: cantera >= 3.1
+.. tags:: species profile, shock tube, kinetics, mixture rule
+"""
+
+import cantera as ct
+import matplotlib.pyplot as plt
+import numpy as np
+
+plt.figure(figsize=(3.5,2.5))
```
speth: As I mentioned in another example, I would recommend leaving the figure at the default size. This small size is resulting in the axis labels being cut off.

samples/python/kinetics/jet_stirred_reactor.py
```diff
@@ -0,0 +1,114 @@
...
+
+import numpy as np
+import pandas as pd
+import time as time
+import cantera as ct
+import matplotlib.pyplot as plt
+
+def getStirredReactor(gas,inputs):
+    reactorRadius = np.cbrt(inputs['V']*3/4/np.pi) # [m3]
+    reactorSurfaceArea = 4*np.pi*np.square(reactorRadius) # [m3]
```
speth:
```suggestion
    reactorRadius = (inputs['V']*3/4/np.pi)**(1/3) # [m3]
    reactorSurfaceArea = 4*np.pi*reactorRadius**2 # [m3]
```

samples/python/kinetics/jet_stirred_reactor.py
```diff
@@ -0,0 +1,114 @@
...
+        state = np.hstack([stirredReactor.thermo.P,
+                        stirredReactor.mass,
+                        stirredReactor.volume,
+                        stirredReactor.T,
+                        stirredReactor.thermo.X])
+        tempDependence.loc[T] = state
+    return tempDependence
+
+def main():
+    models = {'Original':'alzueta.yaml','LMR-R':'alzueta_LMRR.yaml'}
```
speth: I would suggest omitting the `main` method and just calling the following code at the top level of the script.

samples/python/kinetics/jet_stirred_reactor.py
```diff
@@ -0,0 +1,114 @@
+"""
+JSR temperature and species profiles
```
speth: Write out "Jet-stirred reactor" (you can introduce the acronym JSR later if you like)

include/cantera/kinetics/LinearBurkeRate.h
```diff
@@ -0,0 +1,140 @@
...
+    double evalTroeRate(const LmrData& shared_data, DataTypes& dataObj, RateTypes& rateObj);
+    double evalChebyshevRate(const LmrData& shared_data, DataTypes& dataObj, RateTypes& rateObj);
+    double evalFromStruct(const LmrData& shared_data);
+
+    void setContext(const Reaction& rxn, const Kinetics& kin) override;
+
+    void validate(const string& equation, const Kinetics& kin) override;
+
+    vector<size_t> colliderIndices;
+    map<string, AnyMap> colliderInfo;
```
speth: Every member variable should have a Doxygen docstring (`//! ...` on the line before, or `//!< ...` at the end of the line if the description is short enough to fit in the 88 character line width).

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,354 @@
...
+    }
+    double k_LMR_ = 0.0;
+    counter = 0;
+    while(counter<colliderIndices.size()){
+        for (size_t i = 0; i < nSpecies; i++){
+            for (size_t j = 0; j < colliderIndices.size(); j++){
+                if (i == colliderIndices[j]) {
+                    double eps1 = epsObjs1[j].evalRate(shared_data.logT, shared_data.recipT);
+                    double eps2 = epsObjs2[j].evalRate(shared_data.logT, shared_data.recipT);
+                    // eps2 equals either eps_M or eps_i, depending on the scenario
```
speth: This loop structure is far more complicated than necessary. You can reduce this to
```suggestion
    for (size_t j = 0; j < colliderIndices.size(); j++) {
        size_t i = colliderIndices[j];
        double eps1 = epsObjs1[j].evalRate(shared_data.logT, shared_data.recipT);
        double eps2 = epsObjs2[j].evalRate(shared_data.logT, shared_data.recipT);
        // eps2 equals either eps_M or eps_i, depending on the scenario
```
and eliminate the `counter` variable and the `break` statements after evaluating the corresponding rate.

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,354 @@
...
+    eps_mix = 0.0;
+    size_t counter = 0;
+    // Break loop when all colliders have been located to avoid unnecessary iterations
+    while(counter<colliderIndices.size()){
+        // Test each species listed at the top of the YAML file
+        for (size_t i = 0; i<nSpecies; i++){
+            for (size_t j = 0; j<colliderIndices.size(); j++){
+                // Check if the species is specified as an LMR-R collider
+                if (i == colliderIndices[j]) {
+                    eps_mix += shared_data.moleFractions[i] * epsObjs1[j].evalRate(shared_data.logT, shared_data.recipT);
```
speth: A similar simplification of the loop logic applies here -- you just need a single loop over the species represented in `colliderIndices`.

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,354 @@
...
+    for (size_t i = 0; i < T.size(); i++){
+        if (epsObj_M.evalRate(log(T[i]), 1 / T[i]) < 0) {
+            throw InputFileError("LinearBurkeRate::validate", m_input,
+                "Invalid eig0 or eps entry for M.");
+        }
+        for (size_t j = 0; j<colliderIndices.size(); j++) {
+            if (epsObjs1[j].evalRate(log(T[i]), 1 / T[i]) < 0){
+                throw InputFileError("LinearBurkeRate::validate", m_input,
+                    "Invalid eig0 or eps entry for one of the specified colliders.");
+            }
```
speth: Since these are all `ArrheniusRate` objects, you don't need to test that the rate is non-negative at different temperatures -- it's sufficient to just check that the value of `A` is non-negative.

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,354 @@
...
+            if (rateObjs[j].which() != 0 && rateObjs[j].which() != 1 && rateObjs[j].which() != 2) {
+                throw InputFileError("LinearBurkeRate::validate", m_input,
```
speth: I don't think it's possible for an object in `rateObjs` not to be one of the three declared variant types, so this check shouldn't be necessary.

src/kinetics/ReactionRateFactory.cpp
```diff
@@ -75,6 +76,11 @@ ReactionRateFactory::ReactionRateFactory()
         return new PlogRate(node, rate_units);
     });
 
+    // LinearBurkeRate evaluator
+    reg("linear-burke", [](const AnyMap& node, const UnitStack& rate_units) {
```
speth: Proper nouns in YAML identifiers should be capitalized
```suggestion
    reg("linear-Burke", [](const AnyMap& node, const UnitStack& rate_units) {
```

doc/sphinx/yaml/reactions.rst
```diff
@@ -397,6 +397,118 @@ Example::
...
+``linear-burke``
+-------------
+
+A complex-forming reaction (one that depends on both P and X) parameterized
+according to the reduced-pressure linear mixture rule as
+:ref:`described here <sec-linear-burke>`.
+
+Additional fields are:
+
+``collider-list``
```
speth: I would suggest naming this just `colliders`.

doc/sphinx/reference/kinetics/rate-constants.md
```diff
@@ -235,6 +235,61 @@ Chebyshev reactions can be defined in the YAML format using the
...
+
+$$
+\tilde{X}_{i,\text{LMR}}(T,P,\textit{\textbf{X}})=\frac{\Lambda_{0,i}(T)X_i}{\sum_{j} \Lambda_{0,j}(T)X_j}
+$$
+
+can be cast in terms of the absolute value of the least negative chemically significant eigenvalue of the master equation for the $i^{th}$ collider (when pure) in the low-pressure limit, $\Lambda_{0,i}(T)[M]$, and high-pressure limit, $\Lambda_{\infty}(T)$, and $[M]$ is the total concentration.
+
+Evaluating all rate constants at the reduced pressure ($R$)---instead of the pressure ($P$)---of the mixture takes advantage of the fact that rate constants (and their chemically significant eigenvectors) for different colliders are usually far more similar at the same $R$ than the same $P$. In practice, since rate constants are usually expressed with respect to pressure $P$ (which has units of Pa, Torr, bar, atm, etc.) rather than reduced pressure $R$ (which is dimensionless), one needs to find the effective pressure for the $i^{th}$ collider, $P_{i}^{\text{ eff}}$ (with units of $P$), such that the reduced pressure of pure collider $i$ is equal to the reduced pressure of the mixture, which can be shown to be
```
speth: In addition to the rates where you need an effective pressure (PLog and Chebyshev), there is also the Troe case where what you need is an effective third body concentration. Might also be worth explaining how this is handled.

doc/sphinx/yaml/reactions.rst
```diff
@@ -397,6 +397,118 @@ Example::
...
+    A list of dictionaries, where each entry contains parameters corresponding
+    to individual colliders (species in the bath gas).
+
+``collider``
+    The name of the collider species, which must be entered inside quotations (e.g.,
+    ``"H2O"``). The first collider defined must be ``"M"``, which represents the generic
```
speth: I'd suggest `name` instead of `collider` here. Also, I don't think there is any requirement to use quotes on the species name.

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,354 @@
...
+    setParameters(node, rate_units);
+}
+
+void LinearBurkeRate::setParameters(const AnyMap& node, const UnitStack& rate_units)
+{
+    UnitStack eps_units{{Units(1.0), 1.0}};
+    ReactionRate::setParameters(node, rate_units);
+    if (!node.hasKey("collider-list")) {
+        throw InputFileError("LinearBurkeRate::setParameters", m_input,
+            "Incorrect YAML input for LMR-R reaction. Please review implementation guide.");
```
speth: The error messages should reference the missing key, not just tell the user tha the input is "incorrect". I'd leave off the suggestion to "Please review implementation guide.", as there is no documentation that we refer to specifically as an "implementation guide" and this may just lead to more confusion.

doc/sphinx/yaml/reactions.rst
```diff
@@ -397,6 +397,118 @@ Example::
...
+    reference collider (often ``Ar`` or ``N2``) that represents all species lacking their
+    own explicit parameterization.
+
+``eps`` or ``eig0``
+    The fractional contribution of each bath gas component (collider) to the reduced
+    pressure. ``eps`` represents the third-body efficiency of the collider relative
```
speth: This is a bit confusing, and I think there's some room for simplification.

In the case of providing "eps" values, I think this should simply be named "efficiency" if this value is indeed interpretable as an efficiency.

Since `eig0` is not a "fractional" contribution, I'd list it as a separate option, though that will require adjusting where you explain the need for all species to use the same choice of providing either `eig0` or a relative efficiency.

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,354 @@
...
+    }
+    else{
+        throw InputFileError("LinearBurkeRate::setParameters", m_input,
+            "Cannot have both eig0 and eps chosen for M. Any additional colliders must make same choice as that of M. Please review implementation guide.");
+    }
+    AnyMap params;
+    params["A"] = 1.0;
+    params["b"] = 0.0;
+    params["Ea"] = 0.0;
+    epsObj_M = ArrheniusRate(AnyValue(params), colliders[0].units(), eps_units);
```
speth: Since you're always converting to the form that uses epsilon and $\epsilon_M=1$ by definition, you could just eliminate the `epsObj_M` object.

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,354 @@
...
+        colliderInfo.insert({colliders[i]["collider"].as<string>(), colliders[i]});
+        colliderNames.push_back(colliders[i]["collider"].as<string>());
+        ArrheniusRate epsObj_i;
+        // eig0 and eps are ONLY interchangeable due to the requirement that eps_M have parameters {A: 1, b: 0, Ea: 0}
+        params["A"] = colliders[i][eig_eps_key]["A"].asDouble() / colliders[0][eig_eps_key]["A"].asDouble();
+        params["b"] = colliders[i][eig_eps_key]["b"].asDouble() - colliders[0][eig_eps_key]["b"].asDouble();
+        params["Ea"] = colliders[i][eig_eps_key]["Ea"].asDouble() - colliders[0][eig_eps_key]["Ea"].asDouble();
+        epsObj_i = ArrheniusRate(AnyValue(params), colliders[i].units(), eps_units);
+        if (colliders[i].hasKey("rate-constants")) {
+            rateObjs.push_back(PlogRate(colliders[i], rate_units));
```
speth: Rather than checking which fields the collider map has to decide what kind of rate it is, I think you should rely on the same method as is used for the base rate types, where the `type` field is examined (and in this case should be `pressure-dependent-Arrhenius`)

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,354 @@
...
+            epsObjs2.push_back(epsObj_i);
+        }
+        // Collider has an eps specified, but no other info is provided. Assign it the same rate and data objects as "M"
+        else {
+            rateObjs.push_back(rateObj_M);
+            dataObjs.push_back(dataObj_M);
+            epsObjs1.push_back(epsObj_i);
+            epsObjs2.push_back(epsObj_M);
+        }
+    }
```
speth: Since this seems to be the most common case (based on the mechanism used in the examples), for the sake of efficiency I think you want to avoid recalculating the value of `rateObj_M` for each collider when (I think) it will just have the same value for all of them and all you really need is to compute it once and scale it by the sum of the efficiency-weighted mole fractions.

doc/example-keywords.txt
```diff
@@ -18,9 +18,11 @@ fuel cell
 heat transfer
 ignition delay
 internal combustion engine
+jet-stirred reactor
 kinetics
 Matlab
 mixture
+mixture rule
 multicomponent transport
```
speth: I don't think this is going to be a well-understood keyword to most users, and I'd recommend just leaving it out.

doc/example-keywords.txt
```diff
@@ -41,6 +43,9 @@ saving output
 sensitivity analysis
 strained flame
 surface chemistry
+shock tube
+species profile
+temperature profile
```
speth: I would recommend dropping these two keywords as well

interfaces/cython/cantera/reaction.pyx
```diff
@@ -660,6 +660,32 @@ cdef class PlogRate(ReactionRate):
...
+        elif init:
+            if isinstance(input_data, dict):
+                self._rate.reset(new CxxLinearBurkeRate(py_to_anymap(input_data)))
+            elif rates is None:
+                self._rate.reset(new CxxLinearBurkeRate(py_to_anymap({})))
+            elif input_data:
+                raise TypeError("Invalid parameter 'input_data'")
+            else:
+                raise TypeError("Invalid parameter 'rates'")
+            self.set_cxx_object()
```
speth: You should be able to get the constructor from `input_data` working by following the example of the other rate types. This is just a Python data structure containing the same information as the YAML input.

I think you can just remove the `rates` kwarg -- I'm guessing that's just leftover from copying from the PlogRate implementation.

include/cantera/kinetics/LinearBurkeRate.h
```diff
@@ -0,0 +1,140 @@
...
+    /*!
+     *  @param node  AnyMap containing rate information
+     *  @param rate_units  Unit definitions specific to rate information
+     */
+    void setParameters(const AnyMap& node, const UnitStack& rate_units) override;
+
+    void getParameters(AnyMap& rateNode, const Units& rate_units) const;
+    void getParameters(AnyMap& rateNode) const override {
+        return getParameters(rateNode, Units(0));
+    }
```
speth: I think you can just implement the version that doesn't take a `Units` object directly and skip providing this other overload of `getParameters`.

speth (changes requested):
Thanks for the continued work on this, @pjsingal. I would suggest that the first priority at this point is to add some unit tests. I would recommend getting those into place before making any of the other changes that I've suggested below, so you can be sure that making those changes doesn't introduce any regressions.

data/alzueta.yaml
pjsingal: Do you suggest merging alzueta.yaml and alzueta_LMRR.yaml into a single input file? Or should alzueta.yaml have the sections "common" and "baseline-pdep", and alzueta_LMRR.yaml import "common" and have an explicitly defined section called "linear-burke"?

speth: I was thinking a single file, with two phase definitions and three sections defining different sets of reactions.

src/kinetics/LinearBurkeRate.cpp
pjsingal: By using duck typing, a k(T,P) parameterization can be provided for individual colliders (besides just M) without needing to specify a "type" for each one, which may get messy. This code permits mixing of different k(T,P) parameterizations even within the same reaction, without requiring the user to additionally specify "type: falloff", "type: Chebyshev", etc for every collider instance

src/kinetics/LinearBurkeRate.cpp
pjsingal: It would be much more convenient to keep epsObj_M as an ArrheniusRate object, as changing it would complicate the structure of evalFromStruct()

src/kinetics/LinearBurkeRate.cpp
pjsingal: This case actually only applies to colliders which have been explicitly specified with an eig0 or eps value but lack a k(TP) parameterization. For example, in the reaction H + O2 (+M) <=> HO2 (+M) this applies to only He, N2, H2, CO2, NH3, and H2O (thereby leaving 36 other species to be calculated later on in one step using the mole fraction sum). If Troe/PLOG/Chebyshev were provided for He, e.g., then it would be treated using the other branches of this conditional structure. Modifying this aspect of the code would make evalFromStruct() more complicated later on. I would prefer to leave it as-is, but am happy to modify it if you think the computational cost is too high.

src/kinetics/LinearBurkeRate.cpp
pjsingal: I have found a middle-ground solution, where the use of epsObj_M in setParameters() for the colliders containing eig or eps but laking k(TP) is maintained, but the separate calculation of eps_M in evalFromStruct() for the unspecified "pure-M" colliders has been eliminated. The changes can be seen in commit #9e71a5d3

interfaces/cython/cantera/reaction.pyx
pjsingal: I attempted to follow these instructions, and now have a very bare-bones version of it. However, I could use more advice re: what needs to be added now.

doc/sphinx/yaml/reactions.rst
pjsingal: I implemented your suggestions, but chose to keep "eps" for now. I feel like efficiency might look too similar to the "efficiencies" key of the Troe format. What do you think?

speth: But it _is_ an efficiency, right? The `efficiencies` field isn't valid here in any case, so I don't think there should be any confusion with that. I think it's worth avoiding introducing obscure names in the YAML format, and an abbreviation of the name of a Greek letter is about as uninformative as you can get.

src/kinetics/LinearBurkeRate.cpp
speth: I'm OK with leaving this as a performance enhancement to pursue later.

doc/sphinx/reference/kinetics/rate-constants.md
```diff
@@ -235,6 +235,103 @@ Chebyshev reactions can be defined in the YAML format using the
 [`Chebyshev`](sec-yaml-Chebyshev) reaction `type`.
 ```
 
+(sec-linear-Burke)=
+## Linear Burke Rate Expressions
+
+Linear Burke rate expressions employ the reduced-pressure linear mixture rule (LMR-R).
+This mixture rule is used to evaluate the rate constants of complex-forming reactions,
+and is a mole-fraction-weighted sum of the bath gas rate constants (when pure) evaluated
+at the reduced pressure ($R$) and temperature ($T$) of the mixture.
```
speth: The weighting is by the reduced pressures, which are significantly more complicated than just the mole fractions, so I think this should just be:
```suggestion
and is a weighted sum of the bath gas rate constants (when pure) evaluated
at the reduced pressure ($R$) and temperature ($T$) of the mixture.
```

doc/sphinx/reference/kinetics/rate-constants.md
```diff
@@ -235,6 +235,103 @@ Chebyshev reactions can be defined in the YAML format using the
...
+
+$$ k_{\text{LMR-R}}(T,P,\boldsymbol{X}) = \sum_{i}
+k_{i}(T,R_{\text{LMR}})\tilde{X}_{i,\text{LMR}} $$
+
+where the reduced pressure, $R$, in its most general form
```
speth:
```suggestion
where the reduced pressure, $R_{\text{LMR}}$, in its most general form
```

doc/sphinx/reference/kinetics/rate-constants.md
```diff
@@ -235,6 +235,103 @@ Chebyshev reactions can be defined in the YAML format using the
...
+
+Evaluating all rate constants at the reduced pressure ($R$)---instead of the pressure
+($P$)---of the mixture takes advantage of the fact that rate constants (and their
+chemically significant eigenvectors) for different colliders are usually far more
+similar at the same $R$ than the same $P$. In practice, since rate constants are usually
+expressed with respect to pressure $P$ (which has units of Pa, Torr, bar, atm, etc.)
+rather than reduced pressure $R$ (which is dimensionless), one needs to find the
+effective pressure for the $i^{th}$ collider, $P_{i}^{\text{ eff}}$ (with units of $P$),
+such that the reduced pressure of pure collider $i$ is equal to the reduced pressure of
+the mixture, which can be shown to be
```
speth: To break up this very long sentence:
```suggestion
such that the reduced pressure of pure collider $i$ is equal to the reduced pressure of
the mixture. This can be shown to be
```

test/python/test_kinetics.py
```diff
@@ -1396,6 +1396,57 @@ def test_plog_invalid_third_body(self):
         with self.assertRaisesRegex(ct.CanteraError, "Found superfluous"):
             gas = ct.Solution("pdep-test.yaml", "plog-invalid")
 
+    def test_linearburke_plog(self):
+        reaction = "H + OH <=> H2O"
+        gas_baseline = ct.Solution('linearBurke-test.yaml',phase='baseline_mechanism')
+        gas_linearBurke = ct.Solution('linearBurke-test.yaml',phase='linear-Burke_mechanism')
```
speth: There is no `phase` keyword argument, so this is just being ignored and both of your mechanisms use the default phase model in the input file, `baseline_mechanism`. The phase name is specified either as just the second positional argument, or if you really want to use a keyword argument, it's called `name`.

test/python/test_kinetics.py
```diff
@@ -1396,6 +1396,57 @@ def test_plog_invalid_third_body(self):
...
+        T = 1000 # [K]
+        P_ls = [0.1,1,10,100] # [atm]
+        for i in range(len(P_ls)-1):
```
speth: The call `range(n)` returns the values [0, 1, ..., n - 1]. So you're skipping over the the last pressure value in the list.

Also, since the variable `i` is only ever used to index into `P_ls`, you can just write `for P in P_ls` to iterate over the values in `P_ls`.

test/python/test_kinetics.py
```diff
@@ -1396,6 +1396,57 @@ def test_plog_invalid_third_body(self):
...
+            # collider 'O2' treated as M
+            gas_baseline.TPX = gas_linearBurke.TPX = T,P_ls[i],{'O2':1}
+            self.assertNear(gas_baseline.reaction_equations().index(reaction),
+                            gas_linearBurke.reaction_equations().index(reaction))
```
speth: I don't understand this test comparison at all. You're checking that the index of the reaction with the equation "H + OH <=> H2O" is the same in each case. Don't you want to look at the rate constant?

Also, please use pytest-style asserts, for example, for a test where you need to allow some tolerance on the agreement:
```py
assert gas1.T == approx(gas2.T)
```

doc/sphinx/reference/kinetics/rate-constants.md
```diff
@@ -235,6 +235,103 @@ Chebyshev reactions can be defined in the YAML format using the
...
+LMR-R is implemented with the above equation, it is not necessary to specify
+$\Lambda_{\infty}(T)$, which cancels out in evaluating $P_{i}^{\text{ eff}}$. Second,
+only ratios of $\Lambda_{0,i}(T)$ appear in the calculations of
+$\tilde{X}_{i,\text{LMR}}(T,P,\boldsymbol{X})$ and $P_{i,\text{LMR}}^{\text{
+eff}}(T,P,\boldsymbol{X})$, such that third-body efficiencies
+$\epsilon_{0,i}(T)=\Lambda_{0,i}(T)/\Lambda_{0,\text{M}}(T)$ (where the user must assign
+$\epsilon_{0,\text{M}}(T)=1$, as this is true by definition), can be implemented by the
+user in lieu of $\Lambda_{0,i}(T)$.
+
+While direct implementation of LMR-R via the above equation would require $k_i(T,P)$ be
```
speth:
```suggestion
While full implementation of LMR-R via the above equation would require $k_i(T,P)$ be
```

doc/sphinx/reference/kinetics/rate-constants.md
```diff
@@ -235,6 +235,103 @@ Chebyshev reactions can be defined in the YAML format using the
...
+specified in addition to $\Lambda_{0,i}(T)$ or $\epsilon_{0,i}(T)$ for each important
+collider, often data for $k_i(T,P)$ (i.e. the complete $T,P$-dependence) for each
+collider is not available even when $\Lambda_{0,i}(T)$ or $\epsilon_{0,i}(T)$ have
+available data or can be estimated using typical values (as is typically done in kinetic
+models for reactions in modified Lindemann expressions).  It would therefore be useful
+to allow some colliders to be specified with unique $\Lambda_{0,i}(T)$ (or
+$\epsilon_{0,i}(T)$) without $k_i(T,P)$, by assuming the same reduced-pressure
+dependence as M (i.e. $k_{i}(T,R)=k_{M}(T,R)$) for those colliders:
```
speth: A more succinct phrasing that I think gets the same point across:
```suggestion
models for reactions in modified Lindemann expressions).  Therefore, for colliders
with unique $\Lambda_{0,i}(T)$ (or $\epsilon_{0,i}(T)$) but without $k_i(T,P)$, 
the same reduced-pressure dependence as M (that is, $k_i(T,R)=k_M(T,R)$) 
is assumed:
```
Also, note that we try to avoid the Latin abbreviations "i.e." and "e.g.", preferring simply "that is" and "for example".

doc/sphinx/reference/kinetics/rate-constants.md
```diff
@@ -235,6 +235,103 @@ Chebyshev reactions can be defined in the YAML format using the
...
+For the other formats, $P_{i}^{\text{ eff}}$ is implemented directly as the 'pressure'
+of interest.
+
+While not required if unique $k_i(T,P)$ data are available, this approximation, like
+LMR-R, takes advantage of the fact that rate constants for colliders with even very
+different third-body efficiencies often are much more similar at the same reduced
+pressure ($R$) than at the same pressure ($P$) and, in fact, are exactly the same if
+they differ in only their collision frequency (but have the same energy- and
+angular-momentum-transfer kernel). This equation forms the basis of the computational
+implementation of LMR-R in Cantera via the Linear Burke Rate reaction class, as it
```
speth: You can link to the C++ class here:
```suggestion
implementation of LMR-R in Cantera via the :ct:`LinearBurkeRate` reaction class, as it
```

doc/sphinx/yaml/reactions.rst
```diff
@@ -397,6 +397,130 @@ Example::
...
+Additional fields are:
+
+``colliders``
+    A list of dictionaries, where each entry contains parameters corresponding to
+    individual colliders (species in the bath gas).
+
+``name``
+    The name of the collider species (e.g., ``H2O``). The first collider defined must be
+    ``M``, which represents the generic reference collider (often ``Ar`` or ``N2``) that
+    represents all species lacking their own explicit parameterization.
```
speth: I think the descriptions of the `name`, `eig0`, and `eps` fields should be indented here, as they are all part of an individual entry within the `colliders` list. And I think you to introduce that clearly within the `colliders`, for example, by ending with something like "Each entry within the `colliders` list may contain the following fields:"
In addition to these three fields, I think it needs to be a little bit more clear that the rate-specific fields are merged with each collider entry.

Alternatively, if you want to avoid the two levels of nesting, you just start the description by saying that the one additional field of this rate type is the `colliders` list, where each item in the list contains the following fields.

src/kinetics/LinearBurkeRate.cpp
speth: I'm going to insist on adopting this suggestion. Consistency with how we represent different rate parameterizations elsewhere is valuable, and will open up the possibility of generalizing the approach later to work for other rate types, including user-defined ones. For all the examples of LMR-R rates that we give, the only rate that would require a `type` to be specified is the one for "M". And one extra line in the input file that accompanies 8-10 lines of PLOG entries or Chebyshev coefficients really isn't that much overhead.

doc/sphinx/yaml/reactions.rst
```diff
@@ -397,6 +397,130 @@ Example::
...
+A mathematical description of this YAML implementation can be found in Eq. 8 of
+:cite:t:`singal2024`.
+
+Examples::
+
+    equation: H + OH <=> H2O
+    type: linear-Burke
+    colliders:
+    - name: M # N2 is reference collider (Troe format)
+      eps: {A: 1, b: 0, Ea: 0}
```
speth: I'd suggest making this three separate examples, with the description pulled out of the code block, like this:

```suggestion
Examples:

*`linear-Burke` rate with Troe format for the reference collider (N2)*:

    equation: H + OH <=> H2O
    type: linear-Burke
    colliders:
    - name: M
      eps: {A: 1, b: 0, Ea: 0}
```

and continuing on from there.

doc/sphinx/yaml/reactions.rst
```diff
@@ -397,6 +397,130 @@ Example::
...
+      data:
+      - [-1.58e+01, 8.71e-01, -9.44e-02, -2.81e-03, -4.48e-04, 1.58e-03, -2.51e-04]
+      - [2.32e+01, 5.27e-01, 2.89e-02, -5.46e-03, 7.08e-04, -3.03e-03, 7.81e-04]
+      - [-3.80e-01, 8.63e-02, 4.03e-02, -7.23e-03, 5.76e-04, 2.79e-03, -1.49e-03]
+      - [-1.48e-01, -7.18e-03, 2.21e-02, 6.23e-03, -5.98e-03, -8.22e-06, 1.92e-03]
+      - [-6.06e-02, -1.42e-02, 1.34e-03, 9.62e-03, 1.70e-03, -3.65e-03, -4.32e-04]
+      - [-2.46e-02, -9.71e-03, -5.88e-03, 3.05e-03, 5.87e-03, 1.50e-03, -2.01e-03]
+      - [-1.54e-02, -5.24e-03, -6.91e-03, -5.94e-03, -1.22e-03, 2.17e-03, 1.59e-03]
+
+    - name: N2
```
speth: I'd remove the blank line here, as it makes it a bit unclear that the same reaction entry is continuing

include/cantera/kinetics/LinearBurkeRate.h
```diff
@@ -0,0 +1,154 @@
...
+    int mf_number;
+
+protected:
+    double m_pressure_buf = -1.0;
+};
+
+
+//! Pressure-dependent and composition-dependent reaction rate calculated according to
+//! the reduced-pressure linear mixture rule (LMR-R) developed at Columbia University.
+//! @cite singal2024
```
speth: In Doxygen, the `cite` command just inserts the numeric reference, so the "Author et al" context needs to be written out manually. Also, the first sentence (identified by the first period, or a blank line) is treated specially by Doxygen as the "brief" documentation for a class / method.

Finally, it's good to link back to the reference documentation like this (a bit tricky since we're mixing documentation systems).

```suggestion
//! Pressure-dependent and composition-dependent reaction rate calculated according to
//! the reduced-pressure linear mixture rule (LMR-R)
//!
//! This parameterization is described by Singal et al. @cite singal2024 and in the
//! [science reference](../reference/kinetics/rate-constants.html#linear-burke-rate-expressions)
//! documentation.
```

src/kinetics/LinearBurkeRate.cpp
speth: Ok, I think that works for now, though there may be an opportunity for further simplification/optimization later.

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,351 @@
...
+    params["b"] = 0.0;
+    params["Ea"] = 0.0;
+    epsObj_M = ArrheniusRate(AnyValue(params), colliders[0].units(), eps_units);
+    // Start at 1 because index 0 is for "M"
+    for (size_t i = 1; i < colliders.size(); i++){
+        if (!colliders[i].hasKey("name")) {
+                throw InputFileError("LinearBurkeRate::setParameters", m_input,
+                    "The collider located at index '{}' in 'colliders' of reaction '{}'"
+                    " has no 'name' defined.", std::to_string(i), eqn);
+        }
```
speth: Incorrect indentation of the body of this `if` statement. Also, the use of `to_string` is unnecessary.

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,351 @@
...
+            "The first collider defined in reaction '{}' must be 'M'.",eqn);
+    }
+    else if (colliders[0].hasKey("eps")) {
+        if (colliders[0]["eps"]["A"] != 1 || colliders[0]["eps"]["b"] != 0 ||
+            colliders[0]["eps"]["Ea"] != 0) {
+            throw InputFileError("LinearBurkeRate::setParameters", m_input,
+                "The third-body efficiency (eps) must be entered for M as"
+                " 'eps: {A: 1, b: 0, Ea: 0}' in reaction '{}'.", eqn);
+        }
+    }
```
speth: I don't think there's a benefit to requiring the `eps` key for the default collider, given that it's by definition equal to one. You already have other ways to check that a consistent set of parameters is provided, either as `eps` for all non-M species or `eig0` for all species including `M`.

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,351 @@
...
+        throw InputFileError("LinearBurkeRate::setParameters", m_input,
+            "'colliders' key missing from reaction '{}'.",eqn);
+    }
+    auto colliders = node["colliders"].asVector<AnyMap>();
+    if (!colliders[0].hasKey("name")) {
+        throw InputFileError("LinearBurkeRate::setParameters", m_input,
+            "'colliders' key missing from reaction '{}'.",eqn);
+    }
+    else if (!colliders[0].hasKey("eps") && !colliders[0].hasKey("eig0")) {
+        throw InputFileError("LinearBurkeRate::setParameters", m_input,
```
speth: The formatting for `else if` used throughout Cantera is:
```suggestion
            "'colliders' key missing from reaction '{}'.",eqn);
    } else if (!colliders[0].hasKey("eps") && !colliders[0].hasKey("eig0")) {
        throw InputFileError("LinearBurkeRate::setParameters", m_input,
```

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,351 @@
...
+                nm, eqn);
+        }
+        if (!colliders[i].hasKey(eig_eps_key)) {
+            throw InputFileError("LinearBurkeRate::setParameters", m_input,
+                "All collision efficiencies in reaction '{}' must be defined uniformly"
+                " as either eig0 or eps. No mixing and matching is allowed.", eqn);
+        }
+        // Save data to colliderInfo, which will make it accessible by getParameters
+        colliderInfo.insert({colliders[i]["name"].as<string>(), colliders[i]});
+        colliderNames.push_back(colliders[i]["name"].as<string>());
```
speth: The simpler notation for inserting to a map is simply:
```suggestion
        colliderInfo[colliders[i]["name"].asString()] = colliders[i];
        colliderNames.push_back(colliders[i]["name"].as<string>());
```

include/cantera/kinetics/LinearBurkeRate.h
```diff
@@ -0,0 +1,154 @@
...
+    RateTypes rateObj_M; //!< collider M
+
+    //! Stores data objects corresponding to each collider, which can be either
+    //! PlogData, TroeData, or ChebyshevData
+    vector<DataTypes> dataObjs; //!< list for non-M colliders
+    DataTypes dataObj_M; //!< collider M
+
+    size_t nSpecies; //!< total number of species in the kinetics object
+    double logPeff_; //! effective pressure as a function of eps
+    double eps_mix; //! mole-fraction-weighted overall eps value of the mixtures
```
speth:
```suggestion
    double logPeff_; //!< effective pressure as a function of eps
    double eps_mix; //!< mole-fraction-weighted overall eps value of the mixtures
```

include/cantera/kinetics/LinearBurkeRate.h
```diff
@@ -0,0 +1,154 @@
...
+
+    //! Third-body collision efficiency objects (eps = eig0_i/eig0_M)
+    vector<ArrheniusRate> epsObjs1; //!< used for k(T,P,X) and eig0_mix calculation
+    vector<ArrheniusRate> epsObjs2; //!< used for logPeff calculation
+    ArrheniusRate epsObj_M; //!< used just for M (eig0_M/eig0_M = 1 always)
+
+    //! Stores rate objects corresponding to each collider, which can be either
+    //! PlogRate, TroeRate, or ChebyshevRate
+    vector<RateTypes> rateObjs; //!< list for non-M colliders
+    RateTypes rateObj_M; //!< collider M
```
speth: Please take a look at how this is rendered by Doxygen. The preceding comments are only applied to `rateObjs`, with the end-of-line comment appended, and `rateObj_M` only has it's end-of-line comment as documentation. The easiest fix for these is to just provide a full description with each member. For larger groups of variables/methods where you want to provide a shared description, you can also use the grouping support Doxygen has, using the `@name` command with blocks delimited by `@{` and `@}`. You can see several examples of this in `Phase.h` among other places.

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,351 @@
...
+double LinearBurkeRate::evalFromStruct(const LinearBurkeData& shared_data)
+{
+    double sigmaX_M = 0.0;
+    // Test each species listed at the top of the YAML file
+    for (size_t i = 0; i<nSpecies; i++){
+        // Total sum will be essentially 1, but perhaps not exactly due to Cantera's
+        // rounding conventions
+        sigmaX_M += shared_data.moleFractions[i];
+    }
+    eps_mix = 0.0;
```
speth: `eps_mix` is only ever used within `evalFromStruct`, so there's no need to define it as a member variable.

include/cantera/kinetics/LinearBurkeRate.h
```diff
@@ -0,0 +1,154 @@
...
+        RateTypes& rateObj);
+    double evalFromStruct(const LinearBurkeData& shared_data);
+
+    void setContext(const Reaction& rxn, const Kinetics& kin) override;
+
+    void validate(const string& equation, const Kinetics& kin) override;
+
+    //! String name of each collider, appearing in the same order as that of the
+    //! original reaction input.
+    vector<string> colliderNames;
```
speth: For the `Rate` class (but not necessarily the `Data` class), all member variables should be `protected`. All internal (protected) member variables should have names starting with `m_`.

include/cantera/kinetics/LinearBurkeRate.h
```diff
@@ -0,0 +1,154 @@
...
+        DataTypes& dataObj,
+        RateTypes& rateObj);
+    double evalTroeRate(
+        const LinearBurkeData& shared_data,
+        DataTypes& dataObj,
+        RateTypes& rateObj);
+    double evalChebyshevRate(
+        const LinearBurkeData& shared_data,
+        DataTypes& dataObj,
+        RateTypes& rateObj);
```
speth: I think these three members should be `protected` -- they're really only useful from within `evalFromStruct` and require other object state to be set in a particular way to give a correct answer.

As elsewhere, this docstring is only applied to the first method, leaving the other two undocumented. Also, if you provide a `@param` directive for one argument, you must provide one for _all_ arguments to avoid warnings from Doxygen.

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,351 @@
...
+        epsObj_i = ArrheniusRate(AnyValue(params), colliders[i].units(), eps_units);
+        if (colliders[i].hasKey("rate-constants")) {
+            rateObjs.push_back(PlogRate(colliders[i], rate_units));
+            dataObjs.push_back(PlogData());
+            epsObjs1.push_back(epsObj_i);
+            epsObjs2.push_back(epsObj_i);
+        }
+        else if (colliders[i].hasKey("Troe")) {
+            // Value of "type" is unimportant; just needed to make falloff.cpp run
+            colliders[i]["type"];
```
speth: Rather than relying on the fact that this isn't explicitly checked:
```suggestion
            colliders[i]["type"] = "falloff";
```

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,351 @@
...
+double LinearBurkeRate::evalTroeRate(const LinearBurkeData& shared_data,
+    DataTypes& dataObj, RateTypes& rateObj)
+{
+    FalloffData& data = boost::get<FalloffData>(dataObj);
+    TroeRate& rate = boost::get<TroeRate>(rateObj);
+    data.conc_3b = {exp(logPeff_) / GasConstant / shared_data.temperature};
+    data.logT = shared_data.logT;
+    data.recipT = shared_data.recipT;
+    data.temperature = shared_data.temperature;
+    rate.setRateIndex(0);
```
speth: I think this could be moved to only occur once, when the rate object is initially being created.

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,351 @@
...
+        throw InputFileError("LinearBurkeRate::evalFromStruct", m_input,
+            "eps_mix == 0 for some reason");
+    }
+    double k_LMR_ = 0.0;
+    for (size_t j = 0; j < colliderIndices.size(); j++) {
+        size_t i = colliderIndices[j];
+        double eps1 = epsObjs1[j].evalRate(shared_data.logT, shared_data.recipT);
+        double eps2 = epsObjs2[j].evalRate(shared_data.logT, shared_data.recipT);
+        // eps2 equals either eps_M or eps_i, depending on the scenario
+        logPeff_ = shared_data.logP + log(eps_mix) - log(eps2);
```
speth: Given the way its used -- changing every iteration of this loop as a way of passing data into the rate evaluators, I think `logPeff_` should be a local variable, and passed as an argument to `evalPlogRate` and friends.

include/cantera/kinetics/LinearBurkeRate.h
```diff
@@ -0,0 +1,154 @@
...
+    LinearBurkeRate() = default;
+
+    LinearBurkeRate(const AnyMap& node, const UnitStack& rate_units={});
+
+    unique_ptr<MultiRateBase> newMultiRate() const override {
+        return make_unique<MultiRate<LinearBurkeRate, LinearBurkeData>>();
+    }
+
+    //! Identifier of reaction rate type
+    const string type() const override { return "linear-burke"; }
```
speth:
```suggestion
    const string type() const override { return "linear-Burke"; }
```

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,351 @@
...
+            eps_mix;
+    }
+    return k_LMR_;
+}
+
+void LinearBurkeRate::getParameters(AnyMap& rateNode) const
+{
+    vector<AnyMap> topLevelList;
+    for (const auto& entry : colliderInfo) {
+        string name = entry.first;
```
speth: You're missing a section to add the info for the default collider's rate.

samples/python/kinetics/jet_stirred_reactor.py
```diff
@@ -0,0 +1,126 @@
...
+    32127.
+
+    [2] M. U. Alzueta, I. Salas, H. Hashemi, P. Glarborg, CO-assisted NH3 oxidation,
+    Combust. Flame 257 (2023) 112438.
+
+    [3] P. J. Singal, J. Lee, L. Lei, R. L. Speth, M. P. Burke, Implementation of New
+    Mixture Rules Has a Substantial Impact on Combustion Predictions for H2 and NH3,
+    Proc. Combust. Inst. 40 (2024).
+
+Requires: cantera >= 3.1 .. tags:: jet-stirred reactor, kinetics
```
speth: I'm guessing automatic line wrapping got away from you. There needs to be a blank line before the start of the `tags` directive here and in the other new examples:
```suggestion
Requires: cantera >= 3.1

.. tags:: jet-stirred reactor, kinetics
```

speth (changes requested):
Thanks for the updates here, @pjsingal, I think we're getting fairly close.

Could you please temporarily remove the commit that updates the `example_data` submodule? In that case, the jobs that build the docs and run the examples will fail, but at least all the others should run. Once we finalize the input file format and can merge the PR from the example_data repo, then we can update the submodule commit.

test/python/test_kinetics.py
pjsingal: You're exactly right, this was supposed to be looking for rate constants, not equations

test/python/test_kinetics.py
pjsingal: This was done intentionally, as taking P_ls[i]*10 will effectively allow that last pressure value in the list to be reached

pjsingal: Nevermind, I've realized i can still achieve the same thing while implementing your changes

src/kinetics/LinearBurkeRate.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+        ArrheniusRate epsObj_i;
+        // eig0 and eps are ONLY interchangeable due to the requirement that eps_M have
+        // parameters {A: 1, b: 0, Ea: 0}
+        params["A"] = colliders[i][eig_eps_key]["A"].asDouble() /
+            colliders[0][eig_eps_key]["A"].asDouble();
+        params["b"] = colliders[i][eig_eps_key]["b"].asDouble() -
+            colliders[0][eig_eps_key]["b"].asDouble();
+        params["Ea"] = colliders[i][eig_eps_key]["Ea"].asDouble() -
+            colliders[0][eig_eps_key]["Ea"].asDouble();
```
speth, line 179: The current cause of test failures is coming from here, because `colliders[0]` (M) doesn't have a field corresponding to `eig_eps_key`. I think you need to put this normalization of the parameters inside an `if (eig_eps_key == "eig0")` condition, with the alternative being that you can just set `params = colliders[i][eig_eps_key]` as no normalization is nescessary when the parameters are already specified as efficiencies.

pjsingal: Do you think it would be confusing for the user that if eig0 is used, then they must specify a value for M, but if efficiency is used, then they do not? I can update the YAML implementation guide on Monday with a note about this. The alternative would be returning to how it was before, where we simply say that all colliders, including M, must have either 'eig0' or 'efficiency', and that if 'efficiency' is chosen then all they have to do is set it equal to 1 for M.

speth: I don't even think the formulation is necessarily that restrictive. Since for the underlying calculation you always convert to the efficiency, all that's really required is that every non-M species has either an `eig0` or an `efficiency`, and that if any species has an `eig0`, then you also need an `eig0` for `M`. I don't think it's confusing to skip writing an efficiency for `M`, any more than it is in the case of the other formulations for three-body/falloff reactions.

speth (approved):
Thanks for all the updates, @pjsingal. I think this is good to go. I simplified the commit history and pushed one last commit that just adjusts some minor formatting issues in the code and docs.

PR #1713, opened 2024-06-18
[Python] Updates for NumPy 2.0 support in the sdist builder

bryanwweber:
- **Fix a few compiler warnings from Eigen in the Cython interface**
- **Change how the sdist build environment is created**
- **Install example data in the sdist**
  This can be configured by the top-level example_data option
- **Simplify extension setup to resolve duplicate cythonizing**
  When building the sdist via scons, the cython step was running twice because of the conditional. In addition, we were getting a warning about building the extension from multiple source files. These changes resolve both problems by cythonizing once if the files aren't present and only including the C++ sources in the extension.
- **Fix build-system dependencies for NumPy 2.0 support**
  NumPy 2.0 does not support Python 3.8, so we need to special case that version. Also special case for Pint, which doesn't support NumPy 2.0 on Python 3.9.

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -10,10 +10,6 @@ on:
         description: "Try to upload wheels and sdist to PyPI after building"
         required: false
         default: "false"
-      upload_to_anaconda:
-        description: "Try to upload package to Anaconda after building"
-        required: false
-        default: "false"
   push:
     # Run on tags that look like releases
     tags:
@@ -25,7 +21,7 @@ on:
 jobs:
   build-packages:
     name: Trigger building packages from external repos
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Set up the job
         run: |
@@ -41,13 +37,6 @@ jobs:
           else
             echo "UPLOAD_TO_PYPI=false" >> $GITHUB_ENV
           fi
-          if [ -n "${{ github.event.inputs.upload_to_anaconda }}" ]; then
-            echo "UPLOAD_TO_ANACONDA=${{ github.event.inputs.upload_to_anaconda }}" >> $GITHUB_ENV
-          elif [[ "${{ github.ref }}" == refs/tags* ]]; then
-            echo "UPLOAD_TO_ANACONDA=true" >> $GITHUB_ENV
-          else
-            echo "UPLOAD_TO_ANACONDA=false" >> $GITHUB_ENV
-          fi
       - name: Trigger PyPI/Wheel builds
         run: >
           gh workflow run -R cantera/pypi-packages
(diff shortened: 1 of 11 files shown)


bryanwweber:
Related to #1675

corykinney:
I saw #1675 was closed :tada: 

@bryanwweber Is the plan to do some sort of 3.0.0post1 once this work is complete so that a fresh `pip install` will work again (maybe with Python 3.12 support 😄)?

bryanwweber:
> I saw #1675 was closed 🎉
> 
> @bryanwweber Is the plan to do some sort of 3.0.0post1 once this work is complete so that a fresh `pip install` will work again (maybe with Python 3.12 support 😄)?

I'll probably re-open #1675 until the Conda packages are fixed as well. Then I'll see about backporting :smile: I would like to backport, and see if we can preemptively support Python 3.13 as well

.github/workflows/packaging.yml
```diff
@@ -56,18 +45,3 @@ jobs:
           -f upload=${{ env.UPLOAD_TO_PYPI }}
         env:
           GITHUB_TOKEN: ${{ secrets.PYPI_PACKAGE_PAT }}
-      - name: Trigger Conda builds
-        run: >
-          gh workflow run -R cantera/conda-recipes
-          main.yml
```
speth, line 62: This is the change dropping the builds for the `cantera` conda channel, correct? I think this would be the right time to update the installation instructions (`doc/sphinx/install/conda.md`) to reflect `conda-forge` being the sole source of conda packages.

speth (commented):
Thanks, @bryanwweber, all these changes look good to me. Just one request regarding the documentation.

.github/workflows/packaging.yml
bryanwweber: They aren't uploaded from this job, but yes this is the one that builds when pushes happen on `main`. I'll update the docs! :tada:

doc/sphinx/install/conda.md
```diff
@@ -69,16 +53,11 @@ from Python. For this example, the environment is named `ct-env`. From the comma
 line (or the Anaconda Prompt on Windows), run:
 
 ```shell
-conda create --name ct-env --channel cantera cantera ipython matplotlib jupyter
+conda create --name ct-env cantera ipython matplotlib jupyter
```
speth: This is only sufficient if you're using miniforge or have otherwise configured `conda-forge` to be in the default channel list. I think we either need to suggest using the `--channel conda-forge` option or a different way of configuring the channel.

bryanwweber: Good call. I'll make that change

doc/sphinx/install/conda.md
```diff
@@ -270,5 +239,5 @@ running the commands:
 
 ```shell
 conda activate ct-dev
-conda update --channel cantera/label/dev libcantera-devel
+conda update --conda-forge libcantera-devel
```
speth:
```suggestion
conda update --channel conda-forge libcantera-devel
```

speth (approved):

PR #1727, opened 2024-06-27
Add support for HDF5 to sdist/wheels

### LINKED ISSUE #205: Port `setuptools` Python build to meson or scikit-build - sdist/wheel only!
**Abstract**

In https://github.com/Cantera/pypi-packages/issues/3, we discussed how to include `libhdf5` support in Cantera's Python wheels. I expressed some reservations about the complexity of the existing setup and wanted to see if we can simplify that, while also giving end users the ability to configure how a wheel is built from an sdist. I will investigate [meson](https://mesonbuild.com/index.html) and [scikit-build-core](https://scikit-build-core.readthedocs.io/en/latest/index.html) to see if either would be an improvement over setuptools.

Out of scope:

1. I am not considering [enscons](https://github.com/dholth/enscons) because, although it's been updated fairly recently, some of the issues on the repo suggest to me that it's not widely used or supported (for example, [not supporting Python 3.12](https://github.com/dholth/enscons/issues/41)). On the other hand, both meson and scikit-build-core are extremely well supported (the former by NumPy/SciPy, the latter by a lot of the ML community).

Acceptance criteria:

1. I've recorded in this issue the outcome of this investigation with any tradeoffs associated with each system
2. `libhdf5` support is included in the Python wheels that we build in CI
3. End users can compile an sdist into a wheel that includes `libhdf5` support from a `libhdf5` library on their system

Related: #93

bryanwweber:

- Build Python sdist and wheel using `scikit-build-core`
  - `scikit-build-core` will find system libraries in appropriate places because it uses CMake for configuration
- Enable building with HDF5 support
- Enable building the PythonExtensionManager (I think this is new?)

:construction: **TODO:** :construction: 

- [x] Refactor SCons* so that `system_packages` are _enabled_ (right now they're forced to be disabled, but we don't need our private copies anymore)
- [x] Refactor SCons* so that `scons sdist` doesn't run unnecessary system checks
- [x] Include files from Eigen and built libraries from SUNDIALS are installed to the wheel, that shouldn't be happening
- [x] We don't need to package pyx files in the wheel, or cythonized cpp files in the sdist any more. Check that this is correct
- [ ] Figure out license packaging requirements (I don't think we need to package anything except the Cantera license for the sdist, but the wheel may require dependency licenses)
- [x] Get it working with the  sdist/wheel builder workflow repo
- [X] Decide if we need to install cmake and/or ninja with the `build.system-requires` table (see https://github.com/scikit-build/scikit-build-core/issues/806). We do not need to do this because I was wrong about when cmake is installed.
- [x] Resolve [this performance warning](https://github.com/Cantera/pypi-packages/actions/runs/10022676513/job/27702718071#step:17:572):
  ```
    CMake Warning at C:/Users/runneradmin/AppData/Local/Temp/tmp61vecodi/build/_deps/sundials-src/cmake/SundialsBuildOptionsPre.cmake:85 (message):
    SUNDIALS is being built with extensive error checks, performance may be
    affected.
    ```

Closes Cantera/enhancements#205

**Checklist**


SConstruct
```diff
@@ -159,6 +159,61 @@ logger.info(
...
+cantera_version = "3.1.0a2"
+# For use where pre-release tags are not permitted (MSI, sonames)
+cantera_pure_version = re.match(r'(\d+\.\d+\.\d+)', cantera_version).group(0)
+cantera_short_version = re.match(r'(\d+\.\d+)', cantera_version).group(0)
+
+try:
+    cantera_git_commit = get_command_output("git", "rev-parse", "--short", "HEAD")
+    logger.info(f"Building Cantera from git commit {cantera_git_commit!r}")
+except (subprocess.CalledProcessError, FileNotFoundError):
+    env["git_commit"] = "unknown"
```
speth: Should this be:
```suggestion
    cantera_git_commit = "unknown"
```

SConstruct
```diff
@@ -159,6 +159,61 @@ logger.info(
...
+# Python Package Settings
+python_min_version = parse_version("3.8")
+# Newest Python version not supported/tested by Cantera
+python_max_p1_version = parse_version("3.13")
+# The string is used to set python_requires in setup.cfg.in
+py_requires_ver_str = f">={python_min_version}"
+
+if "sdist" in COMMAND_LINE_TARGETS:
+    if "clean" in COMMAND_LINE_TARGETS:
+        COMMAND_LINE_TARGETS.remove("clean")
```
speth, line 183: I think you could remove this special case -- if someone actually runs `scons sdist clean`, they deserve the error about `sdist` being exclusive.

interfaces/python_sdist/build_sdist.py
```diff
@@ -0,0 +1,199 @@
...
+        )
+    )
+
+    configure_source = HERE / "src" / "configure.py.in"
+    configure_target = target_directory / "src" / "configure.py"
+    configure_target.write_text(
+        do_configure_substitution(
+            configure_source, cantera_version, cantera_short_version
+        )
+    )
```
speth: This seems like an extra level of substitutions. Could you just have `do_configure_substitution` generate `config.h` from `config.h.in` directly?

speth (changes requested):
Thanks, @bryanwweber. Overall, this looks good to me. I had just a few minor suggestions.

SConstruct
bryanwweber: I actually really liked running `scons clean sdist` so I knew it was a clean directory. The clean happens before the sdist here, since the clean also happens at "configure" time. I'd rather keep it

interfaces/python_sdist/build_sdist.py
bryanwweber: That's a good point. This is actually a small problem... Since cmake will search the system for dependencies, I either need to pin CMake to only find the exact version we specify in the config.h for SUNDIALS and probably fmt, in which case this could be simplified; or, I need to allow those package versions to be configured at build time by CMake (which wouldn't be that hard), and then we need this double substitution since one needs to run at sdist build time and the other at wheel build time. Thoughts about which you'd prefer? I'm leaning towards the latter to support more downstream platforms.

speth: Right now, the only true variables in `config.h.in` are the cantera version and git commit. Or are you saying that you anticipate broadening the set of configured properties such that there would be other, system-dependent variables that needed to set in `config.h`, and `configure.py` would be responsible for doing that configuration?

bryanwweber: There's also at least the Sundials version, right? If that were to be something other than 7.0, it'd need to be configured on the build machine where the sdist is built to a wheel. That would have to be filled in by CMake via configure.py (since our config.h is not in a format that CMake can fill in directly).

speth: But for now, you are just pinning the SUNDIALS version, right?

bryanwweber: This line: https://github.com/Cantera/cantera/pull/1727/files#diff-85b3f5eee3cabfe48b9a8d7164766ebe64a6bc3e408e62720bdc865dea854ed6R30 doesn't specify which version of SUNDIALS to look for on the system. So someone who builds the sdist locally and who has SUNDIALS installed on their system might end up with a version other than 7.0.

speth: But you are explicitly downloading SUNDIALS 7.0, CMake is compiling it and then somehow putting the libraries on the link path, right? I think would take a somewhat peculiar system configuration to end up pulling in a different copy of SUNDIALS.

Though now, looking at the build log for the Linux wheels, I see the following as the final (shortened) link line:
`/opt/rh/gcc-toolset-12/root/usr/bin/g++  -pthread -fPIC -O3 -DNDEBUG   -shared  -o cantera/_cantera.cpython-312-aarch64-linux-gnu.so cantera/CMakeFiles/_cantera.dir/_cantera.cpp.o [...] cantera/CMakeFiles/_cantera.dir/yamlwriter.cpp.o  -Wl,-rpath,/usr/local/lib:  cantera/libext_manager.a  src/libcantera_lib.a  /usr/local/lib/libfmt.so.10.2.1  /usr/local/lib/libyaml-cpp.so.0.8.0  /usr/local/lib/libhdf5.so.310.2.0  -ldl  /usr/local/lib/libsundials_nvecserial.a  /usr/local/lib/libsundials_cvodes.a  /usr/local/lib/libsundials_idas.a  /usr/local/lib/libsundials_sunmatrixsparse.a  /usr/local/lib/libsundials_sunlinsoldense.a  /usr/local/lib/libsundials_sunmatrixdense.a  /usr/local/lib/libsundials_sunlinsolband.a  /usr/local/lib/libsundials_sunmatrixband.a  /usr/local/lib/libsundials_sunlinsolspgmr.a  /usr/local/lib/libsundials_sunnonlinsolnewton.a  /usr/local/lib/libsundials_core.a  -lm`.

I'm a little surprised that it's explicitly pulling SUNDIALS in from `/usr/local/lib`, and I don't see any sign of it either building SUNDIALS or testing for it before that. On the other hand, it seems to do what I expect on Windows: compiles SUNDIALS, then links directly to the library files in the earlier "Release" build directory. So I guess I'm missing something about how this is all configured to work.

bryanwweber: Right, sorry... In CMake, the `FIND_PACKAGE_ARGS` sets `FetchContent` up to search the system first (using the standard `find_package` function) and only download the library with the given URL if it can't find it on the system. For the Linux builds, I created an image with our external dependencies already built, basically to save time compiling them on the ARM builds that are super slow. That's why it's picking up from the system on Linux. It's possible to set up `FIND_PACKAGE_ARGS` so that it only searches for a single version; it would also be possible to use the variable that CMake sets with the SUNDIALS version to write to the `config.h` at wheel build time.

On a related note, what's the best way for me to document how this is put together? I'm happy to add READMEs or pages in the docs or whatever works best to share this knowledge :smile: 

Edit: FWIW, you can see it's doing the same for libfmt, libhdf5, and libyaml-cpp where they're built on Windows and macOS because those run natively on the system using cibuildwheel, whereas the Linux builds are done in a container to ensure a minimum level of system compatibility. Here's the PEP that defines manylinux, including the context for why that's the decision they went with: https://peps.python.org/pep-0513/

speth: Ah, I see. I guess my inclination would be to keep this simple and only support single versions of these dependencies in the sdist. Hopefully, there shouldn't be very many users installing from the sdist itself, and if a user has more complicated requirements, the SCons build route is always there.

Regarding documentation, adding a new page on the sdist/wheel/PyPI builds in `doc/sphinx/develop/` would be helpful. I guess this sort of fits under the "how Cantera works" section on https://cantera.org/dev/develop/index.html.

bryanwweber:
@speth I think I addressed everything here :smile:

doc/sphinx/develop/distribution-packages/pypi-sdist-wheel.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+Steps 1 and 2 above are set up in the `interfaces/python_sdist/src/CMakeLists.txt` file.
+Steps 3 and 4 are set up in the `interfaces/python_sdist/cantera/CMakeLists.txt` file.
+
+## Publishing the sdist and Wheels
+
+When a version of Cantera is ready to be published to PyPI, the workflow in the
+[pypi-packages](https://github.com/Cantera/pypi-packages) repository must be executed.
+The workflow essentially conducts the steps in the [TLDR](#tldr) section above.
```
speth: I get a "missing cross-reference" warning for this. Do you need to add `myst_heading_anchors = N` to `conf.py`?

doc/sphinx/develop/index.md
```diff
@@ -29,6 +29,25 @@ compiling/config-options
 compiling/special-cases
 ```
 
+(sec-distributing)=
+## Distributing Built Cantera Packages
+
```
speth: I think I'd put this section last, after "Adding New Features"

doc/sphinx/develop/distribution-packages/pypi-sdist-wheel.md
```diff
@@ -0,0 +1,158 @@
...
+This set of steps should build an sdist and wheel on a developer's machine for local
+testing.
+
+1. Clone the repository (submodules are optional)
+1. Install SCons and [build](https://pypi.org/project/build/)
+1. Run `scons sdist`
+   a. `scons clean sdist` can also be used to automatically remove the
+   `build` folder prior to building the `sdist.
+1. Building the wheel requires that Boost and libhdf5 are available on the build system.
+   If these are not installed in standard locations, you can set `Boost_ROOT` and
```
speth: I know environment variables are case insensitive, but:
```suggestion
   If these are not installed in standard locations, you can set `BOOST_ROOT` and
```

speth (changes requested):
Thanks, @bryanwweber. Just a couple of minor comments. Otherwise I think this is good to go.

bryanwweber:
Thanks @speth! Good suggestions :smile:

speth (approved):

PR #1740, opened 2024-07-26
Fix various YAML serialization issues

### LINKED ISSUE #1610: YamlWriter / yaml2ck issues with 'note' field parsed as integer
**Problem description**

There are a pair of coupled problems with species' `note` fields in cases where that field could be (but should not be) interpreted as an integer, for example `120186`, which if you're psychic you might guess is the date December 1, 1986. Or maybe January 12, 1986. Or maybe something else entirely. In any case, Cantera should just preserve this as a string.

**Steps to reproduce**

In Python:
```py
>>> import cantera as ct
>>> gas = ct.Solution('nDodecane_Reitz.yaml')
>>> gas.species('h2o2').input_data['note']
'120186'  # preserved as a string
>>> gas.write_yaml('test-note.yaml')
```

Looking at the resulting YAML file shows that this value has been converted to an integer:
```yaml
  - name: h2o2
    composition: {H: 2.0, O: 2.0}
    ...
    note: 120186
```

which is inconsistent, but not necessarily a significant problem. However, trying to convert this to a Chemkin file:
```py
gas2 = ct.Solution('test-note.yaml')
gas2.write_chemkin('test-note.ck', sort_species=None, sort_elements=None)
```
fails with the error:
```pytb
Cell In [3], line 1
----> 1 gas2.write_chemkin('test-note.ck', sort_species=None, sort_elements=None)

File ~/src/cantera/build/python/cantera/solutionbase.pyx:345, in cantera.solutionbase._SolutionBase.write_chemkin()
    343 
    344         from cantera import yaml2ck
--> 345         output_paths = yaml2ck.convert(
    346             self,
    347             mechanism_path=mechanism_path,

File ~/src/cantera/build/python/cantera/yaml2ck.py:681, in convert(solution, phase_name, mechanism_path, thermo_path, transport_path, sort_elements, sort_species, overwrite)
    674 else:
    675     header_text = header_wrapper.fill("") + "\n"
    676 mechanism_text = [
    677     header_text,
    678     build_elements_text(
    679         all_elements,
    680     ),
--> 681     build_species_text(
    682         all_species,
    683     ),
    684 ]
    685 if thermo_path is None:
    686     mechanism_text.append(
    687         build_thermodynamics_text(
    688             all_species,
    689             separate_file=False,
    690         ),
    691     )

File ~/src/cantera/build/python/cantera/yaml2ck.py:177, in build_species_text(species, max_width)
    173 species_names = {s.name: s.input_data.get("note", "") for s in species}
    175 # Notes shorter than 7 characters are probably related to thermo entries
    176 # and will be included with the thermo entry.
--> 177 if any(len(s) > 6 for s in species_names.values()):
    178     max_species_len = max(len(s) for s in species_names.keys())
    179     max_species_len = max(5, max_species_len)

File ~/src/cantera/build/python/cantera/yaml2ck.py:177, in <genexpr>(.0)
    173 species_names = {s.name: s.input_data.get("note", "") for s in species}
    175 # Notes shorter than 7 characters are probably related to thermo entries
    176 # and will be included with the thermo entry.
--> 177 if any(len(s) > 6 for s in species_names.values()):
    178     max_species_len = max(len(s) for s in species_names.keys())
    179     max_species_len = max(5, max_species_len)

TypeError: object of type 'int' has no len()
```

**System information**

- Cantera version: 3.0.0
- OS: macOS 13.5.1
- Python 3.11

**Additional context**

Originally reported on the Cantera Users' Group: https://groups.google.com/g/cantera-users/c/gn_TJt1gspc

### LINKED ISSUE #1629: Yaml writer bug after removing some originally duplicate reactions
In python: When doing mechanism reduction based on reaction rates of progress, it is possible that duplicate reactions in the original mechanism are no longer duplicate in the reduced mechanism. Those reactions can be identified and the "duplicate" property can be set to False. However, since this property change is not pushed to the underlying C++ object, when the reduced mechanism is written to a *yaml file using the YamlWriter, the originally duplicate reactions are still indicated as being duplicate, and reading the reduced mechanism gives an error. 

**Steps to reproduce**
Taking GRI-3.0 as example, in which reaction 87 and 88 are duplicate. I will remove reaction 88 and set reaction 87 to non-duplicate, then write that "reduced" model to yaml. Importing this new reduced model gives an error.

```
reactions = [gas.reaction(i) for i in range(gas.n_reactions) if i != 88]
reactions[87].duplicate = False
gas_reduced = ct.Solution(name="gas", thermo="ideal-gas", kinetics="gas", species=gas.species(), reactions=reactions)
gas_reduced.write_yaml('gri30_test_red.yaml')
```

Now read the new mechanism:
`gas_reduced = ct.Solution('gri30_test_red.yaml')`

**Behavior**

Gives the following error:

> InputFileError thrown by Kinetics::checkDuplicates:
> Error on line 1186 of ./gri30_test_red.yaml:
> No duplicate found for declared duplicate reaction number 87 (H2O2 + OH <=> H2O + HO2)

### LINKED ISSUE #1695: Multi line notes in Chemkin with mixed white space after comment symbols cause issues with write_yaml
**Problem description**

Sorry for the title gore, but recently I've been working with the NUIGMech 1.2 and have been having issues performing mechanism reductions and writing the reduced mechanisms to YAML files. The issue seems to stem from the Chemkin source files having comments that have a mixture of whitespace after the exclamation point within the same comment. I'd be happy to help contribute to this, but would need advice on the recommended way to fix it that wouldn't break other behaviors. 

**Steps to reproduce**

Run `ck2yaml` with a Chemkin file with the following reaction, note that the original Chemkin file has a space after the `!` in the first line of the comment and no space in the second line of the comment:
```
! A.W. Jasper, S.J. Klippenstein, L.B. Harding, J. Phys. Chem. A 111 (2007) 8699-8707.

!DONG 0406
CH2+CH2=>C2H2+H+H                     7.0E13   0.0022      8 !  
```

Produces the following output, which Cantera loads correctly:
```yaml
- equation: CH2 + CH2 => C2H2 + H + H  # Reaction 90
  rate-constant: {A: 7.0e+13, b: 2.2e-03, Ea: 8.0}
  note: |2-
     A.W. Jasper, S.J. Klippenstein, L.B. Harding, J. Phys. Chem. A 111 (2007) 8699-8707.
    DONG 0406
```
The `|2-` keeps the indentation consistent with the first line keeping its extra space when loaded. After loading the mechanism, creating a new `ct.Solution` object that has that species in it and writing it using `ck2yaml`  gives the following output:

```yaml
  - equation: 2 CH2 => C2H2 + 2 H
    rate-constant: {A: 7.0e+10, b: 2.2e-03, Ea: 3.3472e+04}
    note: |
       A.W. Jasper, S.J. Klippenstein, L.B. Harding, J. Phys. Chem. A 111 (2007) 8699-8707.
      DONG 0406
```
When it gets written, the `2-` is now gone but the initial space gets written out, so this mechanism can't be loaded after the round trip through `ct.Solution` and then `write_yaml` as either a `illegal map value` or `end of map not found` error is thrown. If everything else is behaving correctly, maybe `write_yaml` just needs some handling for leading whitespace on multi line notes that the `ck2yaml` already has? Otherwise, I don't know what the expected behavior is, so I'd appreciate advice on how to go about fixing it. Manually, I've been fixing it by adding the `2-` back to reactions that throw errors, but that is a tedious process to do reaction by reaction for such large mechanisms.

As a side note, I wish all mechanisms were developed in Cantera in the first place 😜

**System information**

- Cantera version: 3.0.0 (806842d)
- OS: Windows 11
- Python: 3.10

speth:

- Fix serialization of changes to reaction/species flags with default values that are normally omitted, such as `duplicate: true`
- Add quotes when outputting strings that look like numbers, such as `123185` so they will be read in as strings rather than being implicitly converted to integers
- Remove initial indentation from multiline strings when serializing, to work around `yaml-cpp`'s limited support of multiline strings.

- Closes #1629
- Closes #1610
- Closes #1695

**Checklist**

### DIFF
--- a/src/kinetics/Reaction.cpp
+++ b/src/kinetics/Reaction.cpp
@@ -226,15 +226,23 @@ void Reaction::getParameters(AnyMap& reactionNode) const
 
     if (duplicate) {
         reactionNode["duplicate"] = true;
+    } else {
+        reactionNode.exclude("duplicate");
     }
     if (orders.size()) {
         reactionNode["orders"] = orders;
+    } else {
+        reactionNode.exclude("orders");
     }
     if (allow_negative_orders) {
         reactionNode["negative-orders"] = true;
+    } else {
+        reactionNode.exclude("negative-orders");
     }
     if (allow_nonreactant_orders) {
         reactionNode["nonreactant-orders"] = true;
+    } else {
+        reactionNode.exclude("nonreactant-orders");
     }
 
     reactionNode.update(m_rate->parameters());
@@ -251,7 +259,7 @@ void Reaction::getParameters(AnyMap& reactionNode) const
             reactionNode["type"] = "elementary";
         }
     } else if (ba::ends_with(rtype, "Arrhenius")) {
-        reactionNode.erase("type");
+        reactionNode.exclude("type");
     } else if (m_explicit_type) {
         reactionNode["type"] = type();
     } else if (ba::ends_with(rtype, "Blowers-Masel")) {
--- a/include/cantera/base/AnyMap.h
+++ b/include/cantera/base/AnyMap.h
@@ -250,6 +250,9 @@ public:
     template<class T>
     AnyValue& operator=(const map<string, T> items);
 
+    //! @see AnyMap::exclude()
+    static AnyValue exclude();
+
     //! Return the held `AnyMap` as a `map` where all of the values have
     //! the specified type.
     template<class T>
@@ -470,6 +473,10 @@ public:
     //! Add items from `other` to this AnyMap. If keys in `other` also exist in
     //! this AnyMap, the `keepExisting` option determines which item is used.
     void update(const AnyMap& other, bool keepExisting=true);
+
+    //! Mark `key` as excluded from this map. This prevents `key` from being added
+    //! by update(). Explicitly setting a value for `key` overrides this exclusion.
+    void exclude(const string& key);
 
     //! Return a string listing the keys in this AnyMap, for use in error
     //! messages, for example
--- a/include/cantera/base/AnyMap.h
+++ b/include/cantera/base/AnyMap.h
@@ -626,9 +626,7 @@ public:
     OrderedProxy ordered() const { return OrderedProxy(*this); }
 
     //! Returns the number of elements in this map
-    size_t size() const {
-        return m_data.size();
-    };
+    size_t size() const;
 
     bool operator==(const AnyMap& other) const;
     bool operator!=(const AnyMap& other) const;
--- a/include/cantera/base/AnyMap.inl.h
+++ b/include/cantera/base/AnyMap.inl.h
@@ -224,7 +224,8 @@ bool AnyValue::eq_comparer(const std::any& lhs, const std::any& rhs)
     auto& ltype = lhs.type();
     auto& rtype = rhs.type();
     AssertThrowMsg(ltype == typeid(T),
-        "AnyValue::eq_comparer", "Compare function does not match held type");
+        "AnyValue::eq_comparer", "Compare function ({}) does not match held type ({})",
+        demangle(ltype), demangle(typeid(T)));
 
     if (ltype == rtype) {
         return any_cast<T>(lhs) == any_cast<T>(rhs);
--- a/src/base/AnyMap.cpp
+++ b/src/base/AnyMap.cpp
@@ -326,35 +326,15 @@ YAML::Emitter& operator<<(YAML::Emitter& out, const AnyMap& rhs)
 
 //! Write YAML strings spanning multiple lines if input includes endline '\n'
 void emitString(YAML::Emitter& out, const string& str0) {
-    size_t endline = str0.rfind('\n');
-    if (endline == string::npos) {
+    if (str0.rfind('\n') == string::npos) {
         out << str0;
         return;
     }
 
-    // Remove trailing line break
-    string str1 = str0;
-    if (endline == str1.size() - 1) {
-        str1.erase(endline, 1);
-        endline = str1.rfind('\n');
-    }
-
-    // Deblank lines (remove whitespace surrounding line breaks)
-    while (endline != string::npos) {
-        size_t len = 1;
-        while (str1[endline + len] == ' ') {
-            len++; // account for whitespace after line break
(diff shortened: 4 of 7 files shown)


src/kinetics/Reaction.cpp
```diff
@@ -196,9 +196,11 @@ void Reaction::check()
 AnyMap Reaction::parameters(bool withInput) const
 {
     AnyMap out;
+    static set<string> excludes{"type", "duplicate", "orders", "negative-orders",
+                                "nonreactant-orders"};
     getParameters(out);
     if (withInput) {
-        out.update(input);
+        out.update(input, true, excludes);
```
ischoegl: While adding `excludes` resolves the issue at hand, I am not convinced that it is the most robust implementation - you are essentially specifying what must not be overwritten, which needs to be adapted for any future change of parameters.

I honestly would prefer the following alternative: if `withInput` is set, initialize `out` with contents of `input` (empty otherwise) and _then_ overwrite/update things with `getParameter`. While it may take more effort (I actually tried that a couple of weeks ago, and ran into a buzz saw of broken logic), it is imho the cleaner resolution, as it will always overwrite things the correct way, irrespective of future changes.

speth: Yeah, I tried that approach as well. I was able to fix the logic in terms of _what_ gets output, but it really messed up control over the order to output the keys. Currently, our logic is to output programmatically added keys first, followed by user-defined keys. However, if we start by inserting all of the raw data, then `AnyMap` sees all of the keys that existed in `input` as "user-defined", and there's not really a way to change their state, since later replacing a value isn't easily distinguishable from just accessing it, since both are just invocations of `AnyMap::operator[]`.

One other option I considered was to delete these defaultable elements out of `input` as we read them, but that changes the interpretation of what `input` is, and causes problems in a couple cases where we later want to access the original item to show it as context for an error message.

I agree that it would be preferable to keep any specification of keys to be excluded with the code in `setParameters`/`getParameters`. I _think_ I have another idea. Stay tuned...

ischoegl (dismissed):
Thanks - this all looks good to me.

ischoegl (approved):

PR #1739, opened 2024-07-25
Resolve compilation issues with C++20 mode and fmt 11.0

speth:
**Changes proposed in this pull request**

- Fix usage of `fmt` when compiling in C++20 mode
- Add support for the recently-released `fmt` 11.0
- Add CI tests specifying use of C++20 mode and using `fmt` 11.0

**If applicable, fill in the issue number this pull request is fixing**

- Closes #1711
- Closes #1735

**If applicable, provide an example illustrating new features this pull request is introducing**

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] The pull request is ready for review

PR #1741, opened 2024-07-27
Remove magic number support for Func1 in clib

### LINKED ISSUE #567: Missing functors in matlab toolbox

ischoegl:

- Remove instantiation of `Func1` via magic number from `clib` (no longer needed after removal of legacy MATLAB toolbox in #1670)
- Fix implementation of `Poly1` (implementation did not match docstring; now, order matches NumPy and MATLAB); use transitional names `Poly13` and `polynomial3` to make changes tractable
- Fix non-functional `func_write` in `clib`
- Skip some deprecation cycles where transitional names were only relevant to legacy MATLAB toolbox
- Implement `checkFunc1` to check for existence of functors
- Remove unnecessary raw pointers from `Func1` code

Experimental MATLAB toolbox:
- Switch to new `Func1` code base
- Use C++ implementation of string output (instead of redundant MATLAB code)
- Eliminate trivial helper functions
- Avoid hard-coding of functor names - use logic based on argument types instead
- Rename `Func` to `Func1` in order to be consistent with C++ and Python API's

Python API:
- Improved exception handling for `Func1.cxx_functor`

Closes #567

**Examples**

In the experimental MATLAB toolbox:
```Matlab
>> a = Func1('polynomial3', [1 -1 0 2.3 0 0])
 
a = 
 
   x^5 - x^4 + 2.3x^2
 
>> [a(1.5), polyval([1 -1 0 2.3 0 0], 1.5)]

ans =

    7.7062    7.7062

>> b = Func1('pow', .5) + 3 * Func1('sin')
 
b = 
 
   \sqrt{x} + 3 \sin(x)

>> [b(3), sqrt(3) + 3*sin(3)]

ans =

    2.1554    2.1554

>> c = (Func1('pow', .5) + 3) * Func1('sin')
 
c = 
 
   \left(\sqrt{x} + 3\right) \sin(x)

>> [c(3), (sqrt(3) + 3)*sin(3)]

ans =

    0.6678    0.6678

>> Func1('spam')
Error using Func1
Functor 'spam' is not implemented
```

In Python:
```python
In [1]: import cantera as ct
   ...: import numpy as np

In [2]: fcn = ct.Func1.cxx_functor('polynomial3', [1, -1, 0, 2.3, 0, 0])

In [3]: fcn(1.5), np.polyval([1, -1, 0, 2.3, 0, 0], 1.5)
Out[3]: (7.706249999999999, 7.706249999999999)

In [4]: fcn.write('x')
Out[4]: 'x^5 - x^4 + 2.3x^2'

In [5]: ct.Func1.cxx_functor('spam', 1, 3)
NotImplementedError                       Traceback (most recent call last)
Cell In[5], line 1
----> 1 ct.Func1.cxx_functor('spam', 1, 3)

File /Volumes/Data/work/GitHub/cantera/build/python/cantera/func1.pyx:128, in cantera.func1.Func1.cxx_functor()
    126 cdef vector[double] arr
    127 if CxxHasFunc1(cxx_string) == 0:
--> 128     raise NotImplementedError(f"Functor '{functor_type}' is not implemented.")
    129 if len(args) == 0 and CxxHasFunc1(cxx_string) == 1:
    130     # simple functor with no parameter

NotImplementedError: Functor 'spam' is not implemented.
```

**Checklist**


ischoegl:
Ok. I think this is finally ready for a review.

include/cantera/numerics/Func1Factory.h
```diff
@@ -143,6 +143,13 @@ shared_ptr<Func1> newFunc1(const string& func1Type,
 shared_ptr<Func1> newFunc1(const string& func1Type,
                            const shared_ptr<Func1> f, double coeff);
 
+//! Check definition of functor object
+//! @param func1Type  String identifying functor type.
+//! @return integer flag indicating existence: '0' if undefined, '1' if @ref func1simple
+//!     or @ref func1advanced; '2' if @ref func1compound or '3' if @ref func1modified.
```
bryanwweber: Small suggestion to make the docstring a little more clear:

```suggestion
//! @return integer flag indicating existence: '0' if undefined; '1' if @ref func1simple
//!     or @ref func1advanced; '2' if @ref func1compound; '3' if @ref func1modified.
```

Separately, this is kind of another magic number, does it make sense to return strings here or an enum type?

src/numerics/Func1.cpp
```diff
@@ -223,6 +223,39 @@ Poly1::Poly1(const vector<double>& params)
     copy(params.data(), params.data() + m_cpoly.size(), m_cpoly.begin());
 }
 
+string Poly1::write(const string& arg) const
+{
+    // write terms in reverse order
+    string out = "";
+    if (m_cpoly[m_cpoly.size()-1] != 0.) {
+        out = fmt::format("{}", m_cpoly[m_cpoly.size()-1]);
```
bryanwweber, line 231: Do these lines need the `fmt::runtime` changes that Ray added in #1739?

bryanwweber (commented):
Thanks @ischoegl! I don't know enough to say whether this is good, so I just had 2 small comments.

include/cantera/numerics/Func1Factory.h
ischoegl: Fair point. I switched to an `enum` and marked the function as `@internal` (i.e. not intended to be made available at high-level APIs)

src/numerics/Func1.cpp
ischoegl: I don't believe so - if I understand correctly, `fmt::runtime` is only required when there are non-constant formatting strings involved? *PS: our CI now checks for fmt 11, so I think this is good*

include/cantera/numerics/Func1.h
```diff
@@ -795,6 +786,8 @@ class Gaussian1 : public Func1
  * @f[
  * f(x) = a_n x^n + \dots + a_1 x + a_0
  * @f]
+ * @since Switched to standard order of coefficients in %Cantera 3.1
+ * (lead with highest order to be consistent with MATLAB and NumPy)
```
speth: I agree that it would make sense to be consistent with NumPy and MATLAB , but changing the interpretation of the coefficient vector like this would change the behavior of any end-user code that uses this class without providing any sort of warning about the change. If you don't want to provide transitional behavior (on the basis that not many people likely use this directly), I think you still need to change the name of the class (and the name used in the factory) so the user at least gets an error message and can update their code.

Also, I think the formula above agrees with the old implementation, doesn't it?

interfaces/cython/cantera/func1.pyx
```diff
@@ -124,10 +124,12 @@ cdef class Func1:
         cdef Func1 f1
         cdef string cxx_string = stringify(functor_type)
         cdef vector[double] arr
-        if len(args) == 0:
+        if CxxHasFunc1(cxx_string) == 0:
+            raise NotImplementedError(f"Functor '{functor_type}' is not implemented.")
+        if len(args) == 0 and CxxHasFunc1(cxx_string) == 1:
```
speth: While you've converted the return type of `CxxHasFunc1` to an `enum` on the C++ side, they're still being treated as magic numbers here and in the equivalent MATLAB function. I haven't seen a good way of handling enumeration types across language boundaries. In this case, I'd suggest one of the alternatives:
- Change this function to return a string
- Make this a boolean function with a second string argument specifying the category
- Use a separate function for each of the different categories

interfaces/matlab_experimental/Utility/ctString.m
```diff
@@ -10,7 +10,11 @@
     if buflen > 0
         aa = char(ones(1, buflen));
         ptr = libpointer('cstring', aa);
-        [iok, bb] = calllib(ctLib, funcName, varargin{2:end}, buflen, ptr);
+        if ischar(varargin{end}) && length(varargin) > 1
+            [iok, ~, bb] = calllib(ctLib, funcName, varargin{2:end}, buflen, ptr);
+        else
+            [iok, bb] = calllib(ctLib, funcName, varargin{2:end}, buflen, ptr);
+        end
```
speth: I'm a little confused by what's happening in the second case here. If `length(varargin) == 1` then isn't `varargin{2:end}` an error? And in the first case, what's the extra output argument that's being ignored?

src/numerics/Func1.cpp
```diff
@@ -223,6 +223,39 @@ Poly1::Poly1(const vector<double>& params)
...
+        string term;
+        if (m_cpoly[m_cpoly.size()-1-n] == 1.) {
+            term = fmt::format("{}", arg);
+        } else if (m_cpoly[m_cpoly.size()-1-n] == -1.) {
+            term = fmt::format("-{}", arg);
+        } else {
+            term = fmt::format("{}{}", m_cpoly[m_cpoly.size()-1-n], arg);
+        }
+        if (n > 1) {
+            term = fmt::format("{}^{}", term, n);
```
speth, line 248: If this is meant to be a LaTeX string, then the exponent should be surrounded with (escaped) braces, in case you have a polynomial of order 10 or more.

speth (changes requested):
Thanks, @ischoegl, I think this is some worthwhile cleanup, especially for the Matlab side of things. My one significant concern is with how we need to handle reversing the coefficient ordering for `Poly1`.

interfaces/matlab_experimental/Utility/ctString.m
ischoegl: `varargin{2:end}` returns nothing if length is 1 (no error). The extra argument is necessary as MATLAB returns all strings it encounters in the parameter list. I recoded this to make this clear.

include/cantera/numerics/Func1.h
ischoegl: Fair point - renamed to `Poly13` (to be reverted after Cantera 3.1). 

Taken together with the docstring for the coefficient vector (with a_0 being last), the formula was not consistent with the old implementation. I clarified the docstring.

interfaces/cython/cantera/func1.pyx
ischoegl: string return is likely easiest (despite the added headache in `clib` handling)

ischoegl:
@speth ... thanks for the review: I addressed all points. 

Beyond, I realized that there were still some unnecessary raw pointers left in the `Func1` interface, which I decided to remove as well (I believe those go back to a version when `Cabinet`s weren't set up for shared pointers, which has since changed). I decided for direct removal of associated constructors for "compounding" and "modifying" functors as they are rather obscure and unlikely to be used in user code.

include/cantera/numerics/Func1.h
```diff
@@ -791,37 +680,43 @@ class Gaussian1 : public Func1
...
 /**
  * Implements a polynomial of degree @e n.
- * The functor class with type @c "polynomial" returns
+ * The functor class with type @c "polynomial3" returns
  * @f[
  * f(x) = a_n x^n + \dots + a_1 x + a_0
  * @f]
+ * with coefficients provided in the order @f$ [a_n, \dots, a_1, a_0] @f$ (consistent
+ * with MATLAB and NumPy conventions).
+ * @since New in %Cantera 3.1
```
speth: The indication of what's actually changed here is a bit too subtle. I think it would be more clear to any affected users to describe as "Changed in Cantera 3.1", and noting the reversal of the coefficient ordering and (temporary) renaming of the class.

ischoegl: Np. I think it's fair to add those clarifications.

speth (approved):

ischoegl:
Thanks, @speth. The new CI failures are fixed in #1746, specifically 0d1d05d774e532697e078aedf3d10c595338672f, so I am merging regardless.

PR #1745, opened 2024-07-29
Remove python_minimal interface

### LINKED ISSUE #208: Remove the `python_minimal` interface
**Abstract**

The `python_minimal` interface exists to provide copies of just the conversion scripts (`ck2yaml` et al.). We do not distribute the interface, and building and installing the full Python interface is much simpler now than when the minimal interface was added. I believe we can remove the minimal interface to save some complexity in the SCons configuration and the source tree. 

**Motivation**

In the past, installing the dependencies to build the full Python/Cython interface to Cantera was somewhat difficult, particularly on Windows. It generally required visiting Cristoph Gohlke's site to download appropriate platform binaries for dependencies, the PyPI ecosystem was much less developed, and conda didn't exist.

Since that time, all of these problems have been resolved, and Cristoph even took his site down! As such, I believe we can reduce the complexity of SCons configuration by removing the option for a "minimal" interface build. Specifically, we could remove the numerous checks when `python_package` is left as the default and the complicated conditionals to decide if the "minimal" package can be built instead of the full.

The only downside that I can see of full removing the minimal interface is that users who are only interested in C/C++/Fortran interfaces to Cantera will need to build the full Python interface to access conversion scripts. However, I believe we provide enough installation options for just the Python interface (PyPI, conda) and as mentioned previously, building the Python interface is as easy as the other interfaces these days. So I don't think this is a big downside.

bryanwweber:

- Remove the `python_minimal` interface
- Refactor the configuration for the Python/Cython interface to simplify
- Use version specifiers in the `setup.cfg.in` file. Requires the `packaging` package; `pkg_resources` is deprecated anyways.

Closes https://github.com/Cantera/enhancements/issues/208

**Checklist**


ischoegl (dismissed):
Thanks, @bryanwweber … this looks good to me.

bryanwweber:
Thanks @ischoegl! I noticed some potential Name/Key errors when I was doing a last look, so I pushed a fix in 8832890e3059601e54aeb2a70b5ebf141e18400c

site_scons/buildutils.py
```diff
@@ -1533,27 +1534,29 @@ def check_module(name):
...
+        exit_on_error = True
+    else:
+        logger_method = logger.warning
+        exit_on_error = False
 
-    python_version = versions["python"]
-    if python_version < env["python_min_version"]:
-        logger.error(
+    python_version = versions.get("python")
+    if not python_version or python_version < env["python_min_version"]:
```
ischoegl, line 1545: Shouldn’t this be `python_version is None`?

bryanwweber: No, this is correct. If `python_version` isn't in the dictionary and `.get()` returns None, we want to warn/error and exit if appropriate. 

Edit: whoops, misread your comment. None is falsey so this is equivalent to what you wrote.

ischoegl (approved):

PR #1746, opened 2024-07-29
Remove CT_SUNDIALS_VERSION constant in favor of SUNDIALS-defined constants

### LINKED ISSUE #209: Remove the custom `CT_SUNDIALS_VERSION` macro
**Abstract**

Prior to SUNDIALS 3.0, SUNDIALS did not provide a built-in way to determine its version, so we added some SCons configuration to set our own configuration macro called `CT_SUNDIALS_VERSION`. After version 3.0, SUNDIALS provides macros with the version definition, so we can remove our custom macro. This will simplify configuration and allow the sdist package to use native SUNDIALS versions macros to determine versions.

**Motivation**

The motivation here is to simplify the SCons configuration and the sdist configuration at wheel build time.

bryanwweber:

- Replace `CT_SUNDIALS_VERSION` with `SUNDIALS_VERSION_MAJOR`. In all the existing cases, the comparison was to `x0` (where `x` was one of 4, 6, or 7) indicating only the major version was being considered.
- Ensured that the version checks for SUNDIALS in SConstruct are still relevant

Closes https://github.com/Cantera/enhancements/issues/209

**Checklist**

### DIFF
--- a/src/numerics/BandMatrix.cpp
+++ b/src/numerics/BandMatrix.cpp
@@ -6,7 +6,6 @@
 #include "cantera/numerics/BandMatrix.h"
 #include "cantera/base/utilities.h"
 #include "cantera/base/stringUtils.h"
-#include "sundials/sundials_config.h"
 
 #if CT_USE_LAPACK
     #include "cantera/numerics/ctlapack.h"
--- a/src/numerics/BandMatrix.cpp
+++ b/src/numerics/BandMatrix.cpp
@@ -6,6 +6,7 @@
 #include "cantera/numerics/BandMatrix.h"
 #include "cantera/base/utilities.h"
 #include "cantera/base/stringUtils.h"
+#include "sundials/sundials_config.h"
 
 #if CT_USE_LAPACK
     #include "cantera/numerics/ctlapack.h"
@@ -233,7 +234,7 @@ int BandMatrix::factor()
     long int nu = static_cast<long int>(nSuperDiagonals());
     long int nl = static_cast<long int>(nSubDiagonals());
     long int smu = nu + nl;
-    #if CT_SUNDIALS_VERSION >= 60
+    #if SUNDIALS_VERSION_MAJOR >= 6
         m_info = SUNDlsMat_bandGBTRF(m_lu_col_ptrs.data(),
                                      static_cast<long int>(nColumns()),
                                      nu, nl, smu, m_ipiv->data.data());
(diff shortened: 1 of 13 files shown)


src/numerics/BandMatrix.cpp
```diff
@@ -6,6 +6,7 @@
 #include "cantera/numerics/BandMatrix.h"
 #include "cantera/base/utilities.h"
 #include "cantera/base/stringUtils.h"
+#include "sundials/sundials_config.h"
```
speth: Unless you found otherwise for some old version of SUNDIALS, I don't think this should be necessary in any case where there's already a SUNDIALS header present, since every SUNDIALS header should already include this header, explicitly or implicitly (usually through `sundials_types.h`).

bryanwweber: My general philosophy is not to rely on transitive imports like that, in case they get removed elsewhere. Is it more typical in C++ to rely on the transitive includes?

speth: I feel like trying to locally `#include` everything directly just leads to having a massive preamble in every source file, and there's no good way (AFAIK) to tell whether you've actually succeeded at avoiding transitive imports.

And at least in the case of a `config.h` file, I think it's a pretty standard expectation that that will be included directly or indirectly in every other header file for the project. On the other end of the spectrum, though, I wouldn't want to rely on a SUNDIALS header including a specific system header or header from another library.

bryanwweber: I think that makes sense. My point about avoiding transitive dependencies was only that I like to include all the direct dependencies for a file. In this case, since we're directly using the values from the `sundials_config.h`, I prefer to include (import in Python, which is where I mostly do this) that file. `sundials_config.h` can manage its own dependencies, if it has any. 

Point taken about config.h files in general though, I'll remove this!

src/numerics/CVodesIntegrator.cpp
```diff
@@ -611,7 +611,7 @@ AnyMap CVodesIntegrator::solverStats() const
                   "supported with sundials versions less than 4.");
     #endif
 
-    #if CT_SUNDIALS_VERION >= 60
+    #if SUNDIALS_VERSION_MAJOR >= 6
```
bryanwweber, line 614: By the way, this line had a typo before, so these lines were not being included, I believe.

src/numerics/BandMatrix.cpp
bryanwweber: Actually, in this case, if `CT_USE_LAPACK` is true, then no other header would bring in `sundials_config.h` so I had to restore it, as far as I can tell. Does it make sense to `#include "cantera/numerics/sundials_headers.h"` instead?

speth: But if `CT_USE_LAPACK` is true, then the preprocessor directives that access `SUNDIALS_VERSION_MAJOR` are all in branches that are inactive, so it doesn't matter that it isn't defined.

src/numerics/CVodesIntegrator.cpp
speth: Gotta love the C preprocessor 🤦

speth (approved):
Thanks, @bryanwweber. This looks good to me.

PR #1749, opened 2024-08-01
Enable cross compiling support without patches

### LINKED ISSUE #210: Enable compatibilty for cross-compliling in SCons
**Abstract**

We currently use `Configure.TryRun()` to test for various configuration properties in SConstruct. For example, to find the versions of our dependent libraries. Since an executable has to be compiled and executed this doesn't work for cross-compiling, since the compiled program can't be executed on the foreign architecture. This causes us to maintain patches to SConstruct for the conda-forge recipe that require fairly constant upkeep. Ideally, we could find a way to ensure the compilers are configured correctly and all the versions for our dependencies without running a program.

**Motivation**

For the conda-forge recipe, we [maintain several patches](https://github.com/conda-forge/cantera-feedstock/tree/main/recipe) for `SConstruct` that essentially hard code different dependency versions to avoid having to run compiled programs to determine configuration options. Since these are simple diff patches, they regularly conflict with ongoing changes to `SConstruct`, forcing us to recreate them. I want to see if we can eliminate the need for `TryRun()` or at least find a way to turn it into a no-op when we're cross-compiling.

**Possible Solutions**

It should be possible to retrieve dependency versions out of header files using the [`-E` or `/E`](https://stackoverflow.com/questions/3742822/how-to-view-c-preprocessor-output) flags to gcc/clang and msvc, respectively. That flag runs only the pre-processor, so by including the relevant header file and then the relevant `#define`s, we should be able to print the versions and extract them without actually running the program. SCons also has a built-in class to run the preprocessor: https://scons.org/doc/4.8.0/HTML/scons-api/_modules/SCons/cpp/#PreProcessor

**References**

Note: cross-compiling means using a compiler that produces machine code for an architecture different than the compiler itself. It is distinct from emulating, where there is no problem running executables (although they're naturally much slower to execute).

bryanwweber:

- Only run the NaN check when we're not cross-compiling. This requires running an executable, which is not possible when cross-compiling unless we can emulate somehow.
- Run only the pre-processor to find header files for version checks. This means we don't have to run an executable just to find `#define`d values. 
- Remove the conditional on the OpenMP check. It seems to work on all the platforms in the conda-forge build without problems, as well as the macOS CI here too

Closes https://github.com/Cantera/enhancements/issues/210

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1348,16 +1348,15 @@ elif conf.CheckDeclaration('_LIBCPP_VERSION', '#include <iostream>', 'C++'):
 else:
     env['cxx_stdlib'] = []
 
-if not env["using_apple_clang"]:
-    # This checks for these three libraries in order and stops when it finds the
-    # first success. Intel = iomp5, LLVM/clang = omp, GCC = gomp. Since gomp is
-    # likely to be installed on the system even if other compilers are installed
-    # or in use, it needs to go last in the check.
-    env['HAS_OPENMP'] = conf.CheckLibWithHeader(
-        ["iomp5", "omp", "gomp"], "omp.h", language="C++"
-    )
-else:
-    env["HAS_OPENMP"] = False
+
+# This checks for these three libraries in order and stops when it finds the
+# first success. Intel = iomp5, LLVM/clang = omp, GCC = gomp. Since gomp is
+# likely to be installed on the system even if other compilers are installed
+# or in use, it needs to go last in the check.
+env['HAS_OPENMP'] = conf.CheckLibWithHeader(
+    ["iomp5", "omp", "gomp"], "omp.h", language="C++"
+)
+
 _, boost_lib_version = run_preprocessor(conf, ["<boost/version.hpp>"], "BOOST_LIB_VERSION")
 if not retcode:
     config_error("Boost could not be found. Install Boost headers or set "
--- a/SConstruct
+++ b/SConstruct
@@ -1427,65 +1427,30 @@ env['NEED_LIBM'] = not conf.CheckLibWithHeader(None, 'math.h', 'C',
                                                'double x; log(x);', False)
 env['LIBM'] = ['m'] if env['NEED_LIBM'] else []
 
-if env['system_sundials'] == 'y':
-    for subdir in ('sundials', 'nvector', 'cvodes', 'idas', 'sunlinsol', 'sunmatrix',
-                   'sunnonlinsol'):
-        remove_directory('include/cantera/ext/' + subdir)
 
+if env['system_sundials'] in ("y", "default"):
     # Determine Sundials version
-    sundials_version_source = get_expression_value(['"sundials/sundials_config.h"'],
-                                                   'QUOTE(SUNDIALS_VERSION)')
-    retcode, sundials_version = conf.TryRun(sundials_version_source, '.cpp')
-    if retcode == 0:
-        config_error("Failed to determine Sundials version.")
-    env["sundials_version"] = sundials_version.strip(' "\r\n')
+    retcode, sundials_version = run_preprocessor(
+        conf,
+        ['"sundials/sundials_config.h"'],
+        "SUNDIALS_VERSION_MAJOR SUNDIALS_VERSION_MINOR SUNDIALS_VERSION_PATCH",
+    )
+    if retcode:
+        sundials_info = check_sundials(conf, sundials_version)
+        env["system_sundials"] = sundials_info["system_sundials"]
+        env["sundials_version"] = sundials_info["sundials_version"]
+        env["has_sundials_lapack"] = sundials_info["has_sundials_lapack"]
 
-    sundials_ver = parse_version(env['sundials_version'])
-    if sundials_ver < parse_version("3.0") or sundials_ver >= parse_version("8.0"):
-        logger.error(f"Sundials version {env['sundials_version']!r} is not supported.")
-        sys.exit(1)
-    elif sundials_ver > parse_version("7.0.0"):
-        logger.warning(f"Sundials version {env['sundials_version']!r} has not been tested.")
-
-    logger.info(f"Using system installation of Sundials version {env['sundials_version']!r}.")
-
-    # Determine whether or not Sundials was built with BLAS/LAPACK
-    if sundials_ver <= parse_version("5.4"):
-        # In Sundials 2.6-5.4, SUNDIALS_BLAS_LAPACK is either defined or undefined
-        env['has_sundials_lapack'] = conf.CheckDeclaration('SUNDIALS_BLAS_LAPACK',
-                '#include "sundials/sundials_config.h"', 'C++')
-    elif sundials_ver <= parse_version("6.6.0"):
-        # In Sundials 5.5-6.6.0, two defines are included specific to the
-        # SUNLINSOL packages indicating whether SUNDIALS has been built with LAPACK
-        lapackband = conf.CheckDeclaration(
-            "SUNDIALS_SUNLINSOL_LAPACKBAND",
-            '#include "sundials/sundials_config.h"',
-            "C++",
-        )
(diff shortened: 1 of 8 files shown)


SConstruct
```diff
@@ -929,7 +929,8 @@ elif env["OS"] == "Darwin":
     config.select("macOS")
 
 # SHLIBVERSION fails with MinGW: http://scons.tigris.org/issues/show_bug.cgi?id=3035
-if (env["toolchain"] == "mingw"):
+# I don't think this does anything? env["toolchain"] seems to be a list of strings, not a single string.
+if env["toolchain"] == "mingw":
     config.select("mingw")
```
bryanwweber: @speth @ischoegl Question here, is this line still relevant? As I wrote in the comment, this line doesn't seem to ever be true. Do we need it?

ischoegl: Not sure ... you could test on the mingw runner on GH though.

Edit: We apparently do use this flag? (see mingw runner)
```yaml
    - name: Build Cantera
      run: scons build -j4 boost_inc_dir=%BOOST_ROOT% debug=n logging=debug
        python_package=y env_vars=USERPROFILE,PYTHONPATH,GITHUB_ACTIONS
        toolchain=mingw f90_interface=n --debug=time
```

bryanwweber: Yeah, I know we set the toolchain to include mingw. The question is more like, `env["toolchain"]` is always a list of tools, so the `==` comparison will always be false as far as I can tell. So line 934 never gets called and things seem fine :shrug:

ischoegl: It’s probably a bug then … the `config.select` is supposed to do something if I recall correctly.

*Edit: in a nutshell, we need this line and the bug should be fixed. `mingw` isn’t configured correctly otherwise … although there’s not much where this matters at the moment.*

SConstruct
```diff
@@ -1163,21 +1163,6 @@ if env['coverage']:
...
-            open("config.log").read().strip(),
-            f"\n{' End of config.log ':*^80}",
-        ]
-        logger.debug("\n".join(debug_message), print_level=False)
-    else:
-        error_message = [message]
-        error_message.append("\nSee 'config.log' for details.")
-        logger.error("\n".join(error_message))
-    sys.exit(1)
-
```
ischoegl, line 1180: This is still referenced in the code, although you won't notice if everything passes?

bryanwweber: Yes I moved it to buildutils and it's imported with the * import at the top of SConstruct

ischoegl: Ah, ok. The famous `*` import 😝 .

site_scons/buildutils.py
```diff
@@ -1643,3 +1642,138 @@ def make_relative_path_absolute(path_to_check: Union[str, Path]) -> str:
         pth = Path(Dir("#" + path_to_check).abspath)
 
     return pth.as_posix()
+
+
+def run_preprocessor(conf, includes, text, defines=()) -> tuple[int, str]:
```
ischoegl: PEP 484 for arguments?

site_scons/buildutils.py
```diff
@@ -1643,3 +1642,138 @@ def make_relative_path_absolute(path_to_check: Union[str, Path]) -> str:
...
+    # library, but Sundials was configured without this support, print a Warning.
+    # TODO: Why is this a warning and what can/should users do about it?
+    if not has_sundials_lapack and conf.env['use_lapack']:
+        logger.warning("External BLAS/LAPACK has been specified for Cantera "
+                       "but Sundials was built without this support.")
+
+    return {"system_sundials": "y", "sundials_version": str(sundials_ver), "has_sundials_lapack": has_sundials_lapack}
+
+
+def config_error(message):
```
ischoegl: PEP 484.

ischoegl (commented):
This PR is somewhat out of my wheelhouse, but as far as I can tell it looks overall good to me (and the CI passes!). 

I have two trivial comments about `buildutils.py` ... we may want to start being consistent with PEP 484 for additions as it really takes out the guesswork about what those parameters are to be expected?

bryanwweber:
Thanks @ischoegl! Good suggestions, I added them :smile:

site_scons/buildutils.py
```diff
@@ -1643,3 +1648,147 @@ def make_relative_path_absolute(path_to_check: Union[str, Path]) -> str:
...
+        )
+        has_sundials_lapack = lapackband and lapackdense
+    else:
+        # In Sundials 6.6.1, the SUNDIALS_BLAS_LAPACK_ENABLED macro was introduced
+        has_sundials_lapack = conf.CheckDeclaration("SUNDIALS_BLAS_LAPACK_ENABLED",
+                '#include "sundials/sundials_config.h"', 'c++')
+
+    # In the case where a user is trying to link Cantera to an external BLAS/LAPACK
+    # library, but Sundials was configured without this support, print a Warning.
+    # TODO: Why is this a warning and what can/should users do about it?
```
speth: The reason it is a warning is it means that reactor integration won't be able to use the external BLAS/LAPACK library and will instead run using SUNDIALS' comparatively slower builtin linear solvers.

site_scons/buildutils.py
```diff
@@ -1643,3 +1648,147 @@ def make_relative_path_absolute(path_to_check: Union[str, Path]) -> str:
...
+        conf.env["CXXFLAGS"].remove(flag)
+    if retcode:
+        retval = None
+        # On MSVC, the `/P` flag produces an output file named for the input file,
+        # that is, `conftest_<hash of contents>_0.i`. However, SCons assumes that the
+        # `.obj` file will still be the target and sets that as the `conf.lastTarget`
+        # with the filename: `conftest_<hash of contents>_0_<hash of action>.obj`.
+        # Since we need the contents of the `.i` file, we need this string munging on
+        # the `.obj` filename to find the right file. If SCons ever decides to change
+        # how they name conftest files, this will probably break. --Bryan
```
speth: I don't think you need to sign your comments. That's what `git blame` is for, if it's ever needed.

site_scons/buildutils.py
```diff
@@ -1643,3 +1648,147 @@ def make_relative_path_absolute(path_to_check: Union[str, Path]) -> str:
...
 
     return pth.as_posix()
+
+
+def run_preprocessor(
+    conf: "SConfigure",
+    includes: "TextOrSequence",
+    text: "TextOrSequence",
+    defines: "TextOrSequence" = ()
+) -> Tuple[int, str]:
```
speth, line 1679: This function could use a docstring. Especially to explain what the return value is.

site_scons/buildutils.py
```diff
@@ -1643,3 +1648,147 @@ def make_relative_path_absolute(path_to_check: Union[str, Path]) -> str:
...
+            break
+        if retval is None:
+            raise ValueError("Could not find version. See config.log.")
+        return retcode, retval
+    else:
+        return retcode, ""
+
+
+def check_sundials(conf: "SConfigure", sundials_version: str) -> Dict[str, str]:
+    sundials_ver = parse_version(".".join(sundials_version.strip().replace('"', "").split()))
```
speth: A docstring would be helpful. What is the return value here?

speth (changes requested):
Thanks, @bryanwweber. It will be really nice to be able to avoid the painfully slow emulator-based builds. I had just a couple of small comments.

site_scons/buildutils.py
bryanwweber: Thanks! So the fix is to build sundials with the same support, if possible. I'll update the message!

site_scons/buildutils.py
bryanwweber: :smile: Thanks! A habit I picked up at work where it's easier to ping the relevant person on Slack if git blame is confused because the lines move.

bryanwweber:
@speth Comments are addressed, thanks for the review!

speth (approved):
Thanks, @bryanwweber!

PR #1723, opened 2024-06-22
Configure libhdf5_serial if it's available

bryanwweber:
**Changes proposed in this pull request**

- Configure `libhdf5_serial` if it's available and `libhdf5` is not. This happens, for example, by default on Ubuntu.
- Change configuration of HDF5 support so that specifying `hdf_include` or `hdf_libdir` does not change the state of `hdf_support` from `'n'` to `'y'`

**If applicable, fill in the issue number this pull request is fixing**

Closes #1718

**Checklist**

- [X] The pull request includes a clear description of this code change
- [X] Commit messages have short titles and reference relevant issues
- [X] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [X] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [X] The pull request is ready for review

PR #1752, opened 2024-08-01
Fix post-merge tests

### LINKED ISSUE #1751: Post-merge tests are failing, but can be fixed :tada:
**Problem description**

As of [f1b98ee](https://github.com/Cantera/cantera/commit/f1b98eea980bb70bb349bcc7d27e5a628e7cbe93), several [post-merge runs](https://github.com/Cantera/cantera/actions/runs/10191624062) are failing. They seem to be due to:

- Too-strict version specifier for Cantera (fedora-rawhide)
- Missing `packaging` (homebrew)
- Needing to use nightly wheels for dependencies (Python 3.13)

We should resolve these failures as much as possible so the signal of those tests is useful, especially most of these errors can be fixed in the GHA workflow rather than Cantera itself.

bryanwweber:

- Set up the `SpecifierSet`s for version comparisons to allow pre-release versions of packages. This shouldn't change any behavior for stable versions.
- `hdf` variables no longer need to be specified to SCons CLI, they're handled by SConstruct since #1723
- Install pre-release wheels from the anaconda index for NumPy, SciPy, Pandas, and Cython. This means we don't have to build them ourselves :tada: 
- If the Python package is built for a newer version of Python than we support, ignore pip's error about unsupported Python because the user is warned by SCons anyways
- Add support for Python 3.13 to the extensions

Closes #1751 

**Checklist**

### DIFF
--- a/.github/workflows/post-merge-tests.yml
+++ b/.github/workflows/post-merge-tests.yml
@@ -10,15 +10,22 @@ on:
     # Run when the main branch is pushed to
     branches:
       - main
+  # Run on pull requests when this file is modified
+  pull_request:
+    branches:
+      - main
+    paths:
+      - .github/workflows/post-merge-tests.yml
+
+env:
+  PIP_EXTRA_INDEX_URL: https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
+  PIP_ONLY_BINARY: ":all:"
 
 jobs:
   prerelease-cython:
     name: Pre-release Cython
     runs-on: ubuntu-22.04
     timeout-minutes: 60
-    env:
-      HDF5_LIBDIR: /usr/lib/x86_64-linux-gnu/hdf5/serial
-      HDF5_INCLUDEDIR: /usr/include/hdf5/serial
     steps:
     - uses: actions/checkout@v4
       name: Checkout the repository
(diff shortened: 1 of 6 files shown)


.github/workflows/post-merge-tests.yml
```diff
@@ -50,7 +50,7 @@ jobs:
     - name: Build Cantera
       run: python3 `which scons` build env_vars=all
         CXX=clang++-14 CC=clang-14 f90_interface=n extra_lib_dirs=/usr/lib/llvm/lib
-        -j4 debug=n --debug=time logging=debug
+        -j4 debug=n --debug=time logging=debug python_package=y
```
speth, line 53: I've been wondering if we should update the default to build the Python package unless it is explicitly disabled, so that missing a Python dependency results in a build failure unless you set `python_package=n`.

bryanwweber: I'm for it! Do you want to make an enhancements issue so we can at least try to document some tradeoffs?

bryanwweber:
As of yesterday, Pandas pre-release wheels were available for Python 3.13: https://github.com/Cantera/cantera/actions/runs/10207129578/job/28266033217 They are no longer available right now :disappointed:

bryanwweber:
I also don't _think_ I did anything that would start causing the sample tests to fail. It does look like the artifact uploaded of `libcantera_shared` is not correct, it's only a few hundred bytes as opposed to the kB that are uploaded on `main`. :sos:

speth:
It's uploading the symlink, not the actual file the symlink points to. Seems to be an unexpected change introduced in the latest version of `upload-artifact`, as reported in actions/upload-artifact#589.

bryanwweber:
That was my guess, that it was uploading the symlink. Glad to know it's not just us. I'll pin to 4.3.4 for now.

bryanwweber:
Based on https://github.com/pandas-dev/pandas/issues/59372, it looks like wheels get removed after some time in the nightly index, probably to avoid overloading Anaconda's willingness to provide storage :grimacing: And as it happens, Pandas wheels haven't been uploading for a while, due to test failures on their part. So I'm fine with accepting those jobs failing.

speth (approved):
Thanks, @bryanwweber. This all looks fine to me.

PR #1754, opened 2024-08-03
Remove clib *newFromFile from MATLAB / fix memory leaks

### LINKED ISSUE #1756: Ambiguous references in clib's Cabinet<M>::index
**Problem description**

When calling a destructor for a `Solution` object via `soln_del` using the `clib` API, associated `ThermoPhase`, `Kinetics` and `Transport` objects are deallocated. As integer entry numbers ('handles') are not available, the destructor relies on `Cabinet<M>::index`, which returns the last entry of a reverse lookup table. There is no guarantee that the *correct* entry is removed, leading to potentially unpredictable behavior and memory leaks.

**System information**

- Cantera version: 3.1a3
- OS: any

**Additional context**

- found in context of #1754

ischoegl:

- Remove `clib`'s `*_newFromFile` constructors from MATLAB; use pathways using `Solution` exclusively instead
- Fix memory leak in `clib` where automatically added *adjacent* phases could not be removed
- Make MATLAB `Interface.adjacent` consistent with Python version (select phase by name rather than index)
- Fix ambiguous references that will lead to memory leaks (going back to #1448)

Resolves #1756
Partially addresses Cantera/enhancements#199

**Example**

New in *experimental* MATLAB:
```
surf = Interface('ptcombust.yaml', 'Pt_surf');
gas = surf.adjacent('gas');
```

**Checklist**

### DIFF
--- a/interfaces/matlab_experimental/Base/Solution.m
+++ b/interfaces/matlab_experimental/Base/Solution.m
@@ -77,7 +77,7 @@ classdef Solution < handle & ThermoPhase & Kinetics & Transport
             ID = ctFunc('soln_newSolution', src, name, transport_model);
             % Inherit methods and properties from ThermoPhase, Kinetics, and Transport
             s@ThermoPhase(ID);
-            s@Kinetics('clib', ID);
+            s@Kinetics(ID);
             s@Transport('clib', ID);
             s.phaseID = ID;
             s.solnName = ctString('soln_name', s.phaseID);
--- a/interfaces/matlab_experimental/Base/Solution.m
+++ b/interfaces/matlab_experimental/Base/Solution.m
@@ -1,7 +1,7 @@
 classdef Solution < handle & ThermoPhase & Kinetics & Transport
     % Solution Class ::
     %
-    %     >> s = Solution(src, name, trans)
+    %     >> s = Solution(src, name, transport_model)
     %
     % Class :mat:class:`Solution` represents solutions of multiple species. A
     % solution is defined as a mixture of two or more constituents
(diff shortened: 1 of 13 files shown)


ischoegl:
Added a fix for #1756: while the issue was pre-existing, this PR surfaced some symptoms, where incorrect `Cabinet` entries were deleted when `Solution` destructors were called. The PR is (again) ready for review.

*PS: force-push was only to fix typos in commit messages / prevent CI failures.*

interfaces/matlab_experimental/Base/Solution.m
```diff
@@ -39,57 +31,62 @@
...
-    % :param trans:
-    %     String, transport modeling. Possible values are ``'default'``, ``'none'``,
-    %     ``'mixture-averaged'``, or ``'multicomponent'``. If not specified,
-    %     ``'default'`` is used.
+    % :param name:
+    %     name of the phase to import as specified in the YAML file.
+    % :param transport_model:
+    %     String specifying transport model. Possible values are ``'default'``,
+    %     ``'none'``, ``'mixture-averaged'``, or ``'multicomponent'``. If not specified,
+    %     ``'none'`` is used.
```
speth: For consistency with the Python and C++ interfaces, shouldn't this use the `default` transport model if the argument is omitted?

src/clib/Cabinet.h
```diff
@@ -61,12 +63,14 @@ class SharedCabinet
     /**
      * Add a new object. The index of the object is returned.
      */
-    static int add(shared_ptr<M> obj) {
+    static int add(shared_ptr<M> obj, int parent=-1) {
         dataRef data = getData();
         data.push_back(obj);
+        parentRef parents = getParents();
```
speth: You could just write:
```suggestion
        auto& parents = getParents();
```
and eliminate the need for the `parentRef` typedef.

src/clib/ct.cpp
```diff
@@ -218,12 +214,35 @@ extern "C" {
...
         } catch (...) {
             return handleAllExceptions(-2, ERR);
         }
     }
 
+    int soln_adjacentName(int n, int a, int lennm, char* nm)
+    {
+        try {
+            auto soln = SolutionCabinet::at(n);
+            if (a < 0 || a >= (int)soln->nAdjacent()) {
```
speth: I'd recommend `static_cast<int>` over the C-style cast.

test/clib/test_clib.cpp
```diff
@@ -28,11 +28,21 @@ TEST(ct, cabinet_exceptions)
...
     int ret = soln_del(999);
     ASSERT_EQ(ret, -1);
     err = reportError();
     EXPECT_THAT(err, HasSubstr("delete a non-existing object."));
 
+    int ref = soln_newSolution("h2o2.yaml", "ohmech", "default");
+    soln_del(ref);
+    soln_thermo(ref);
+    err = reportError();
+    EXPECT_THAT(err, HasSubstr("has been deleted."));
```
speth: You might also want to check the return value from `soln_thermo` here, right?

speth (commented):
Thanks, @ischoegl. This looks mostly good to me. Just had a couple of small suggestions.

interfaces/matlab_experimental/Base/Solution.m
ischoegl: Whoops - good catch!

ischoegl:
> Thanks, @ischoegl. This looks mostly good to me. Just had a couple of small suggestions.

Thanks for the prompt review! I think it's all taken care of.

speth (approved):

PR #1759, opened 2024-08-05
more detailed debugging outputs, solver documentation, and high loglevel print improvements

wandadars:
Current work for a potential improvement for the damped newton step function as well as an improvement in the debugging output that the 1D solver generates when loglevel is set to 8. For the debugging outputs, a more comprehensive history of the solution steps that result in a failure is kept in the debugging file, which can then be examined to find where a solution started to diverge

**Checklist**

### DIFF
--- a/include/cantera/oneD/MultiNewton.h
+++ b/include/cantera/oneD/MultiNewton.h
@@ -111,6 +111,15 @@ public:
      * Find the solution to F(X) = 0 by damped Newton iteration. On entry, x0
      * contains an initial estimate of the solution. On successful return, x1
      * contains the converged solution.
+     *
+     * The convergence criteria is when the 2-norm of the Newton step is less than one.
+     *
+     * Returns:
+     * - int : Status code
+     *   - `1`  a converged step was able to be taken.
+     *   - `-2` no suitable damping coefficient was found within the maximum iterations.
+     *   - `-3` the current solution `x0` was too close to the boundary and the step
+     *          points out of the allowed domain.
      */
     int solve(double* x0, double* x1, OneDim& r, MultiJac& jac, int loglevel);
 
@@ -124,7 +133,7 @@ public:
 
 protected:
     //! Work arrays of size #m_n used in solve().
-    vector<double> m_x, m_stp, m_stp1, temp_x0, temp_stp0;
+    vector<double> m_x, m_stp, m_stp1;
 
     int m_maxAge = 5;
 
--- a/include/cantera/oneD/MultiNewton.h
+++ b/include/cantera/oneD/MultiNewton.h
@@ -60,18 +60,19 @@ public:
      *   F(x) = 0
      * @f]
      *
-     * Where `F` is the system of nonlinear equations, `x` is the solution vector.
+     * Where @f$ F @f$ is the system of nonlinear equations and @f$ x @f$ is the
+     * solution vector.
      *
      * For the damped Newton method we are solving:
      *
      * @f[
-     *   x_{k+1} - x_k = \Delta x_k = -\alpha_k J^(-1)(x_k) F(x_k)
+     *   x_{k+1} - x_k = \Delta x_k = -\alpha_k J^{-1}(x_k) F(x_k)
      * @f]
      *
-     * Where @f$ J @f$ is the Jacobian matrix of @f$ F @f$ with respect to @f$ x @f$, and @f$ \alpha_k @f$ is
-     * the damping factor, and @f$ \Delta x_k @f$ is the Newton step at @fx_k @f, sometimes
-     * called the correction vector. In the equations here, k is just an iteration
-     * variable.
+     * Where @f$ J @f$ is the Jacobian matrix of @f$ F @f$ with respect to @f$ x @f$,
+     * and @f$ \alpha_k @f$ is the damping factor, and @f$ \Delta x_k @f$ is the Newton
+     * step at @f$ x_k @f$, sometimes called the correction vector. In the equations
+     * here, @f$ k @f$ is just an iteration variable.
      *
      * In this method, the Jacobian does not update, even when the solution vector is
      * evaluated at different points.
@@ -83,8 +84,9 @@ public:
      *   x_{k+1} = x_k + \alpha_k \Delta x_k
      * @f]
      *
-     * Pick @f$ \alpha_k @f$ such that @f$ \norm{\Delta x_{k+1}} < \norm{\Delta x_k} @f$.
-     * Where @f$ \Delta x_k = J^{-1}(x_k) F(x_k) @f$, and
+     * Pick @f$ \alpha_k @f$ such that
+     * @f$ \Vert \Delta x_{k+1} \Vert < \Vert \Delta x_k \Vert @f$
+     * where @f$ \Delta x_k = J^{-1}(x_k) F(x_k) @f$, and
      * @f$ \Delta x_{k+1} = J^{-1}(x_{k}) F(x_{k+1}) @f$.
      *
      * @param[in] x0 initial solution about which a Newton step will be taken
--- a/include/cantera/oneD/MultiNewton.h
+++ b/include/cantera/oneD/MultiNewton.h
@@ -45,11 +45,61 @@ public:
                      const OneDim& r, int loglevel);
 
     /**
-     * On entry, step0 must contain an undamped Newton step for the solution x0.
-     * This method attempts to find a damping coefficient such that the next
-     * undamped step would have a norm smaller than that of step0. If
-     * successful, the new solution after taking the damped step is returned in
-     * x1, and the undamped step at x1 is returned in step1.
+     * Performs a damped Newton step to solve the system of nonlinear equations.
+     *
+     * On entry, `step0` must contain an undamped Newton step for the solution `x0`.
+     * This method attempts to find a damping coefficient `alpha_k` such that the next
+     * undamped step would have a norm smaller than that of `step0`. If successful,
+     * the new solution after taking the damped step is returned in `x1`, and the
+     * undamped step at `x1` is returned in `step1`.
+     *
+     * This uses the method outlined in @cite kee2003.
+     *
+     * The system of equations can be written in the form:
+     * @f[
+     *   F(x) = 0
+     * @f]
+     *
+     * Where `F` is the system of nonlinear equations, `x` is the solution vector.
+     *
+     * For the damped Newton method we are solving:
+     *
+     * @f[
+     *   x_{k+1} - x_k = \inc x_k = -\alpha_k J^(-1)(x_k) F(x_k)
+     * @f]
+     *
+     * Where `J` is the Jacobian matrix of `F` with respect to `x`, and `alpha_k` is
+     * the damping factor, and @f$ \inc x_k @f$ is the Newton step at `x_k`, sometimes
+     * called the correction vector. In the equations here, k is just an iteration
+     * variable.
+     *
+     * In this method, the Jacobian does not update, even when the solution vector is
+     * evaluated at different points.
+     *
+     * The general algorithm is described below.
+     *
+     * We want to solve the equation:
+     * @f[
+     *   x_{k+1} = x_k + \alpha_k \inc x_k
+     * @f]
+     *
+     * Pick @f$ \alpha_k @f$ such that @f$ \norm{\inc x_{k+1}} < \norm{\inc x_k} @f$.
+     * Where @f$ \inc x_k = J^{-1}(x_k) F(x_k) @f$, and
+     * @f$ \inc x_{k+1} = J^{-1}(x_{k}) F(x_{k+1}) @f$.
+     *
+     * @param x0 initial solution about which a Newton step will be taken
+     * @param step0 initial Newton step without any damping
+     * @param x1 solution after taking the damped Newton step
+     * @param step1 Newton step after taking the damped Newton step
+     * @param loglevel controls amount of printed diagnostics
+     * @param writetitle controls if logging title is printed
+     *
+     * Returns:
+     * - int : Status code
+     *   - `1` if a damping coefficient is found and the solution converges.
+     *   - `0` if a damping coefficient is found but the solution does not converge.
+     *   - `-2` if no suitable damping coefficient is found within the maximum iterations.
+     *   - `-3` if the current solution `x0` is too close to the boundary and the step points out of the allowed domain.
      */
     int dampStep(const double* x0, const double* step0, double* x1, double* step1,
                  double& s1, OneDim& r, MultiJac& jac, int loglevel, bool writetitle);
@@ -74,7 +124,7 @@ public:
 
 protected:
     //! Work arrays of size #m_n used in solve().
-    vector<double> m_x, m_stp, m_stp1;
+    vector<double> m_x, m_stp, m_stp1, temp_x0, temp_stp0;
 
     int m_maxAge = 5;
 
--- a/include/cantera/oneD/MultiNewton.h
+++ b/include/cantera/oneD/MultiNewton.h
@@ -53,7 +53,7 @@ public:
      * the new solution after taking the damped step is returned in `x1`, and the
      * undamped step at `x1` is returned in `step1`.
      *
-     * This uses the method outlined in @cite kee2003.
+     * This uses the method outlined in Kee et al. @cite kee2003.
      *
      * The system of equations can be written in the form:
      * @f[
@@ -65,11 +65,11 @@ public:
      * For the damped Newton method we are solving:
      *
      * @f[
-     *   x_{k+1} - x_k = \inc x_k = -\alpha_k J^(-1)(x_k) F(x_k)
+     *   x_{k+1} - x_k = \Delta x_k = -\alpha_k J^(-1)(x_k) F(x_k)
      * @f]
      *
-     * Where `J` is the Jacobian matrix of `F` with respect to `x`, and `alpha_k` is
-     * the damping factor, and @f$ \inc x_k @f$ is the Newton step at `x_k`, sometimes
+     * Where @f$ J @f$ is the Jacobian matrix of @f$ F @f$ with respect to @f$ x @f$, and @f$ \alpha_k @f$ is
+     * the damping factor, and @f$ \Delta x_k @f$ is the Newton step at @fx_k @f, sometimes
      * called the correction vector. In the equations here, k is just an iteration
      * variable.
      *
@@ -80,26 +80,29 @@ public:
      *
      * We want to solve the equation:
      * @f[
-     *   x_{k+1} = x_k + \alpha_k \inc x_k
+     *   x_{k+1} = x_k + \alpha_k \Delta x_k
      * @f]
      *
-     * Pick @f$ \alpha_k @f$ such that @f$ \norm{\inc x_{k+1}} < \norm{\inc x_k} @f$.
-     * Where @f$ \inc x_k = J^{-1}(x_k) F(x_k) @f$, and
-     * @f$ \inc x_{k+1} = J^{-1}(x_{k}) F(x_{k+1}) @f$.
+     * Pick @f$ \alpha_k @f$ such that @f$ \norm{\Delta x_{k+1}} < \norm{\Delta x_k} @f$.
+     * Where @f$ \Delta x_k = J^{-1}(x_k) F(x_k) @f$, and
+     * @f$ \Delta x_{k+1} = J^{-1}(x_{k}) F(x_{k+1}) @f$.
      *
-     * @param x0 initial solution about which a Newton step will be taken
-     * @param step0 initial Newton step without any damping
-     * @param x1 solution after taking the damped Newton step
-     * @param step1 Newton step after taking the damped Newton step
-     * @param loglevel controls amount of printed diagnostics
-     * @param writetitle controls if logging title is printed
+     * @param[in] x0 initial solution about which a Newton step will be taken
+     * @param[in] step0 initial undamped Newton step
+     * @param[out] x1 solution after taking the damped Newton step
+     * @param[out] step1 Newton step after taking the damped Newton step
+     * @param[out] s1 norm of the subsequent Newton step after taking the damped Newton step
+     * @param[in] r domain object, used for evaluating residuals over all domains
+     * @param[in] jac Jacobian evaluator
+     * @param[in] loglevel controls amount of printed diagnostics
+     * @param[in] writetitle controls if logging title is printed
      *
-     * Returns:
-     * - int : Status code
-     *   - `1` if a damping coefficient is found and the solution converges.
-     *   - `0` if a damping coefficient is found but the solution does not converge.
-     *   - `-2` if no suitable damping coefficient is found within the maximum iterations.
-     *   - `-3` if the current solution `x0` is too close to the boundary and the step points out of the allowed domain.
+     * @returns
+     *   - `1` a damping coefficient was found and the solution converges.
+     *   - `0` a damping coefficient was found, but the solution has not converged yet.
+     *   - `-2` no suitable damping coefficient was found within the maximum iterations.
+     *   - `-3` the current solution `x0` is too close to the solution bounds and the
+     *          step would exceed the bounds on one or more components.
      */
     int dampStep(const double* x0, const double* step0, double* x1, double* step1,
                  double& s1, OneDim& r, MultiJac& jac, int loglevel, bool writetitle);
@@ -108,18 +111,18 @@ public:
     double norm2(const double* x, const double* step, OneDim& r) const;
 
     /**
-     * Find the solution to F(X) = 0 by damped Newton iteration. On entry, x0
+     * Find the solution to F(x) = 0 by damped Newton iteration. On entry, x0
      * contains an initial estimate of the solution. On successful return, x1
-     * contains the converged solution.
+     * contains the converged solution. If failure occurs, x1 will contain the
+     * value of x0 i.e. no change in solution.
      *
      * The convergence criteria is when the 2-norm of the Newton step is less than one.
      *
-     * Returns:
-     * - int : Status code
-     *   - `1`  a converged step was able to be taken.
+     * @returns
+     *   - `1`  a converged solution was found.
      *   - `-2` no suitable damping coefficient was found within the maximum iterations.
-     *   - `-3` the current solution `x0` was too close to the boundary and the step
-     *          points out of the allowed domain.
+     *   - `-3` the current solution `x0` is too close to the solution bounds and the
+     *          step would exceed the bounds on one or more components.
      */
     int solve(double* x0, double* x1, OneDim& r, MultiJac& jac, int loglevel);
 
@@ -135,11 +138,19 @@ protected:
     //! Work arrays of size #m_n used in solve().
     vector<double> m_x, m_stp, m_stp1;
 
+    //! Maximum allowable Jacobian age before it is recomputed.
     int m_maxAge = 5;
+
+    //! Factor by which the damping coefficient is reduced in each iteration
+    double m_dampFactor = sqrt(2.0);
+
+    //! Maximum number of damping iterations
+    size_t m_maxDampIter = 7;
 
     //! number of variables
     size_t m_n;
 
+    //! Elapsed CPU time spent computing the Jacobian.
     double m_elapsed = 0.0;
 };
 }
--- a/include/cantera/oneD/OneDim.h
+++ b/include/cantera/oneD/OneDim.h
@@ -46,14 +46,16 @@ public:
     MultiNewton& newton();
 
     /**
-     * Solve F(x) = 0, where F(x) is the multi-domain residual function. Returns
-     * a 1 on success, and either a -2 or -3 on failure depending on whether the
-     * maximum number of damping steps was reached(-2) or if the solution was up
-     * against the bounds(-3).
+     * Solve F(x) = 0, where F(x) is the multi-domain residual function.
      *
      * @param x0         Starting estimate of solution.
      * @param x1         Final solution satisfying F(x1) = 0.
      * @param loglevel   Controls amount of diagnostic output.
+     *
+     * @returns
+     * - 1 for success
+     * - -2 failure (maximum number of damping steps was reached)
+     * - -3 failure (solution was up against the bounds
      */
     int solve(double* x0, double* x1, int loglevel);
 
--- a/include/cantera/oneD/OneDim.h
+++ b/include/cantera/oneD/OneDim.h
@@ -46,7 +46,11 @@ public:
     MultiNewton& newton();
 
     /**
-     * Solve F(x) = 0, where F(x) is the multi-domain residual function.
+     * Solve F(x) = 0, where F(x) is the multi-domain residual function. Returns
+     * a 1 on success, and either a -2 or -3 on failure depending on whether the
+     * maximum number of damping steps was reached(-2) or if the solution was up
+     * against the bounds(-3).
+     *
      * @param x0         Starting estimate of solution.
      * @param x1         Final solution satisfying F(x1) = 0.
      * @param loglevel   Controls amount of diagnostic output.
@@ -232,6 +236,13 @@ public:
     void setMaxTimeStep(double tmax) {
         m_tmax = tmax;
     }
+
+    /**
+     * Sets a factor by which the time step is reduced if the time stepping
+     * fails. The default value is 0.5.
+     *
+     * @param tfactor factor time step is multiplied by if time stepping fails
+     */
     void setTimeStepFactor(double tfactor) {
         m_tfactor = tfactor;
     }
--- a/include/cantera/oneD/Sim1D.h
+++ b/include/cantera/oneD/Sim1D.h
@@ -168,10 +168,35 @@ public:
      */
     AnyMap restore(const string& fname, const string& name);
 
+    /**
+     * Deletes a `debug_sim1d.yaml` file if it exists. Used to clear the file for
+     * successive calls to the solve() method.
+     */
     void clearDebugFile();
 
-    void writeDebugInfo(const string& header, const string& message, int loglevel,
-                        int attempt_counter);
+    /**
+     * Write solver debugging information to a YAML file based on the specified log
+     * level.
+     *
+     * This method writes solver debug information to a specified YAML file
+     * (`debug_sim1d.yaml`). The section headers are formatted according to the provided
+     * `header_suffix` and `attempt_counter` arguments. Depending on the log level, the
+     * method will save either the solution information or the residual information
+     * for each attempted solution.
+     *
+     * @param header_suffix  Header used to construct a unique section in the YAML file
+     *                       where the information will be written to.
+     * @param message  A string that is written to the `description` tag in the YAML
+     *                 file.
+     * @param loglevel  Controls the type of output that will be written. A `loglevel`
+     *                  greater than 6 saves the solution, and a `loglevel` greater
+     *                  than 7 saves the residual additionally.
+     * @param attempt_counter  An integer counter used to uniquely identify the attempt
+     *                         which is included in the file header to differentiate
+     *                         between multiple solution attempts.
+     */
+     void writeDebugInfo(const string& header_suffix, const string& message, int loglevel,
+                         int attempt_counter);
 
 
     //! @}
--- a/include/cantera/oneD/Sim1D.h
+++ b/include/cantera/oneD/Sim1D.h
@@ -168,6 +168,12 @@ public:
      */
     AnyMap restore(const string& fname, const string& name);
 
+    void clearDebugFile();
+
+    void writeDebugInfo(const string& header, const string& message, int loglevel,
+                        int attempt_counter);
+
+
     //! @}
 
     void setTimeStep(double stepsize, size_t n, const int* tsteps);
--- a/src/oneD/MultiNewton.cpp
+++ b/src/oneD/MultiNewton.cpp
@@ -323,7 +323,7 @@ int MultiNewton::solve(double* x0, double* x1, OneDim& r, MultiJac& jac, int log
     while (true) {
         // Check whether the Jacobian should be re-evaluated.
         if (jac.age() > m_maxAge) {
-            if (loglevel > 0) {
+            if (loglevel > 1) {
                 writelog("\n  Maximum Jacobian age reached ({}), updating it.", m_maxAge);
             }
             forceNewJac = true;
@@ -365,7 +365,7 @@ int MultiNewton::solve(double* x0, double* x1, OneDim& r, MultiJac& jac, int log
                     break;
                 }
                 nJacReeval++;
-                if (loglevel > 0) {
+                if (loglevel > 1) {
                     writelog("\n  Re-evaluating Jacobian (damping coefficient not found"
                             " with this Jacobian)");
                 }
--- a/src/oneD/MultiNewton.cpp
+++ b/src/oneD/MultiNewton.cpp
@@ -232,14 +232,11 @@ int MultiNewton::dampStep(const double* x0, const double* step0,
 {
     // write header
     if (loglevel > 0 && writetitle) {
-        string header = fmt::format("  {:->23}Damped Newton iteration{:->24}", "", "");
-        string separator = fmt::format("  {:->70}", "");
-        writelog("\n\n{}", header);
-
+        writelog("\n\n  {:-^70}", " Damped Newton iteration ");
         writelog("\n  {:<4s}  {:<10s}   {:<10s}  {:<7s}  {:<7s}  {:<7s}  {:<5s}  {:<3s}\n",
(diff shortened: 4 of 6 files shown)


wandadars:
I realized that I hadn't been doing the proper thing with regards to the line search, and so for now I just reverted back to the original method. I added some documentation about the method as it was clearly explained in the Kee book.

wandadars:
I don't know if I'm not reading it correctly, but I thought these two would be the same for the case when loglevel is 0:

```
if (loglevel > 0) {
           writelog("\nRe-evaluating Jacobian(damping coefficient not found"
          " with this Jacobian)\n");
 }
```


 ```
 debuglog("\nRe-evaluating Jacobian(damping coefficient not found"
                            " with this Jacobian)\n", loglevel);
```

But for some reason the debuglog was writing data when loglevel was 0 and when I wrote the first one, it stopped writing when loglevel is 0.

**Comment by wandadars** [comment]
Based on the description of the Newton solver in the Kee book, there was a thing that didn't quite make sense to me in the damped newton solver where fbound was multiplied by the damping factor. I think that fbound just serves as the starting point for the correction vector length, but in the implementation it was always multiplying the damping factor by the fbound, which I think might have artificially reduced the damping factor lower than necessary.

**Comment by wandadars** [comment]
I may have gone overboard on the solver print statements. The logging outputs when loglevel was set to a high value were...quite perplexing in their current form.

**Comment by wandadars** [comment]
For anyone interested, I created a simple 2 equation, 1 point control model problem that I was using just to examine the behavior of the algorithm on a small-scale problem.

```
import numpy as np

control_point_index = 3

# The equation to solve is Laplace's equation in 1D: d^2T/dx^2 = 0
# The dummy equation for A is dA/dx = 0 at all points, except the control point,
# where A = T - UserVal
def residuals(X, UserVal, num_nodes):
    T = X[:num_nodes]
    A = X[num_nodes:]
    R = np.zeros(len(X))

    # Main equation residuals (discretized Laplace's equation)
    for i in range(1, len(T) - 1):
        R[i] = T[i - 1] - 2 * T[i] + T[i + 1]

    # A equation residuals
    for i in range(len(A)):
        if i == control_point_index:  # Control point at the specified node
            R[num_nodes + i] = T[i] - UserVal
        elif i < control_point_index:  # upwind towards the control point
            R[num_nodes + i] = A[i] - A[i + 1]
        else:  # downwind away from the control point
            R[num_nodes + i] = A[i] - A[i - 1]

    # Boundary condition residuals
    R[0] = T[0] - A[0]  # Left boundary at Node 0
    R[num_nodes-1] = T[-1] - 15  # Right boundary at the last node

    # Printing the residuals in a detailed way
    print("Residuals:")
    for i, r in enumerate(R):
        print(f"R[{i}] = {r:.4e}")

    return R

# Here we need to numerically evaluate the Jacobian matrix
# The Jacobian matrix is the matrix of partial derivatives of the residuals with respect to X (T and A combined).
# We approximate it numerically by finite differences.
def jacobian(X, UserVal, num_nodes):
    n = len(X)
    J = np.zeros((n, n))

    delta = 1e-6
    for i in range(n):
        X_plus = X.copy()
        X_minus = X.copy()

        X_plus[i] += delta
        X_minus[i] -= delta

        R_plus = residuals(X_plus, UserVal, num_nodes)
        R_minus = residuals(X_minus, UserVal, num_nodes)

        J[:, i] = (R_plus - R_minus) / (2.0 * delta)

    # Printing the Jacobian matrix in a detailed way
    print("Jacobian Matrix:")
    for row in J:
        print(" ".join(f"{val: .4e}" for val in row))

    return J

def newtons_method_multi(residuals, jacobian, X0, UserVal, num_nodes, tolerance=1e-6, max_iterations=25):
    X_n = X0.copy()

    epsilon = 1e-8
    for i in range(max_iterations):
        F_n = residuals(X_n, UserVal, num_nodes)
        J_n = jacobian(X_n, UserVal, num_nodes)

        # Regularize the Jacobian if necessary
        try:
            delta_x = np.linalg.solve(J_n + epsilon * np.eye(J_n.shape[0]), -F_n)
        except np.linalg.LinAlgError:
            print("Jacobian matrix is singular, regularization failed to resolve.")
            return X_n

        # Print out solution vector to screen
        print(f"\nSolution Vector before iteration {i+1}:")
        for j, x in enumerate(X_n):
            print(f"X[{j}] = {x:.6f}")

        # Plot the solution vector for Temperature and A
        plt.plot(X_n[:num_nodes], label='Temperature')
        plt.plot(X_n[num_nodes:], label='A')
        plt.legend()
        plt.show()


        # Update X
        X_n += delta_x

        # Check for convergence
        if np.linalg.norm(delta_x) < tolerance:
            print(f'Converged to solution after {i+1} iterations.')
            return X_n

    print('Did not converge.')
    return X_n


num_nodes = 6
UserVal = 220.0 # Desired value of T at the control point

# Initialize T to be between 5 and 15, linearly increasing from left to right
T0 = np.linspace(200, 400, num_nodes)
A0 = np.ones(num_nodes)
X0 = np.concatenate([T0, A0])

# Solve using Newton's method in matrix form
solution_X = newtons_method_multi(residuals, jacobian, X0, UserVal, num_nodes)

# Separate the solution vector into T and A
solution_T = solution_X[:num_nodes]
solution_A = solution_X[num_nodes:]

# Print the solution vectors in a clean and interpretable way
print("Solution T:")
for i, t in enumerate(solution_T):
    print(f"T[{i}] = {t:.6f}")

print("\nSolution A:")
for i, a in enumerate(solution_A):
    print(f"A[{i}] = {a:.6f}")

```

**Comment by wandadars** [comment]
One strange thing that I notice is that if we add the temporal term of the continuity equation to the continuity residual, the standard non-two-point control solve() fails. This seems odd to me. The time component from the Kee book is just d(rho)/dt

```
            // For "axisymmetric-flow", the continuity equation  propagates the
            // mass flow rate information to the left (j+1 -> j) from the value
            // specified at the right boundary. The lambda information propagates
            // in the opposite direction.
            //rsd[index(c_offset_U,j)] = -(rho_u(x,j+1) - rho_u(x,j))/m_dz[j]
            //                           -(density(j+1)*V(x,j+1) + density(j)*V(x,j));
            rsd[index(c_offset_U,j)] = -(rho_u(x,j+1) - rho_u(x,j))/m_dz[j]
                                       -(density(j+1)*V(x,j+1) + density(j)*V(x,j));
            rsd[index(c_offset_U,j)] -= rdt * (m_rho[j] - m_rho_prev[j]);

            diag[index(c_offset_U, j)] = 1; // Algebraic constraint

```

**Comment by ischoegl** [comment]
> For anyone interested, I created a simple 2 equation, 1 point control model problem that I was using just to examine the behavior of the algorithm on a small-scale problem. [...]

Examples like this could be added under `samples/cxx`. We have a __Blasius__ example there that illustrates Cantera's `BoundaryValueProblem` class.

**Comment by speth** [comment]
> One strange thing that I notice is that if we add the temporal term of the continuity equation to the continuity residual, the standard non-two-point control solve() fails. This seems odd to me. The time component from the Kee book is just d(rho)/dt

The time-dependent solver only supports terms that end up on the diagonal of the Jacobian (hence the name of the variable `diag`, so there's no way to add a $d\rho/dt$ term to the solver.

**Comment by wandadars** [comment]
> > One strange thing that I notice is that if we add the temporal term of the continuity equation to the continuity residual, the standard non-two-point control solve() fails. This seems odd to me. The time component from the Kee book is just d(rho)/dt
> 
> The time-dependent solver only supports terms that end up on the diagonal of the Jacobian (hence the name of the variable `diag`, so there's no way to add a d ρ / d t term to the solver.

Ah yep. That's definitely it.  🤦‍♂️

File: src/oneD/Sim1D.cpp [hunk 1]
```diff
@@ -608,6 +599,28 @@ int Sim1D::refine(int loglevel)
...
+    if (!std::filesystem::remove("debug_sim1d.yaml")) {
+        // Only log a warning if the file existed and couldn't be deleted for some other reason
+        if (std::filesystem::exists("debug_sim1d.yaml")) {
+            writelog("Warning: Unable to delete previous debug file 'debug_sim1d.yaml'\n");
+        }
+    }
+}
+
+void Sim1D::writeDebugInfo(const std::string& header, const std::string& message,
+                         int loglevel, int attempt_counter) {
```
speth: Since it's so common, we have `using std::string` in `ct_defs.h` (which is always included). For consistency with other usage, the continued arguments (`int loglevel`) should be indented to the same level as the first argument after the `(`, and for a function definition, the opening `{` should be on a line by itself.

src/oneD/Sim1D.cpp
```diff
@@ -608,6 +599,28 @@ int Sim1D::refine(int loglevel)
     return np;
 }
 
+void Sim1D::clearDebugFile() {
+    // Use std::filesystem::remove to attempt deletion without throwing exceptions
+    if (!std::filesystem::remove("debug_sim1d.yaml")) {
+        // Only log a warning if the file existed and couldn't be deleted for some other reason
+        if (std::filesystem::exists("debug_sim1d.yaml")) {
+            writelog("Warning: Unable to delete previous debug file 'debug_sim1d.yaml'\n");
+        }
```
speth: The logic here doesn't seem to align with the behavior of [`std::filesystem::remove`](https://en.cppreference.com/w/cpp/filesystem/remove). `remove` should return `false` only if the file did not exist, and will raise an exception if the file existed but couldn't be deleted. If that delete fails, presumably the later attempt to write to the file would also fail, so I don't think there's anything you need to check here, and you can just write `std::filesystem::remove("debug_sim1d.yaml")` with no checks at the one point where `clearDebugFile` is called.

include/cantera/oneD/MultiNewton.h
```diff
@@ -45,11 +45,61 @@ class MultiNewton
...
+     * @f[
+     *   F(x) = 0
+     * @f]
+     *
+     * Where `F` is the system of nonlinear equations, `x` is the solution vector.
+     *
+     * For the damped Newton method we are solving:
+     *
+     * @f[
+     *   x_{k+1} - x_k = \inc x_k = -\alpha_k J^(-1)(x_k) F(x_k)
```
speth: What is `\inc` supposed to be? `\Delta`?

include/cantera/oneD/MultiNewton.h
```diff
@@ -45,11 +45,61 @@ class MultiNewton
...
+     * undamped step at `x1` is returned in `step1`.
+     *
+     * This uses the method outlined in @cite kee2003.
+     *
+     * The system of equations can be written in the form:
+     * @f[
+     *   F(x) = 0
+     * @f]
+     *
+     * Where `F` is the system of nonlinear equations, `x` is the solution vector.
```
speth: Since you're referring to a symbol appearing in a mathematical expression, rather than a variable name in the code, this should be formatted as `@f$ F @f$` rather than ``` `F` ```. Here and in several other locations.

include/cantera/oneD/MultiNewton.h
```diff
@@ -45,11 +45,61 @@ class MultiNewton
...
-     * x1, and the undamped step at x1 is returned in step1.
+     * Performs a damped Newton step to solve the system of nonlinear equations.
+     *
+     * On entry, `step0` must contain an undamped Newton step for the solution `x0`.
+     * This method attempts to find a damping coefficient `alpha_k` such that the next
+     * undamped step would have a norm smaller than that of `step0`. If successful,
+     * the new solution after taking the damped step is returned in `x1`, and the
+     * undamped step at `x1` is returned in `step1`.
+     *
+     * This uses the method outlined in @cite kee2003.
```
speth: In Doxygen, the `@cite kee2003` just gets turned into `[17]`, so this should read:
```suggestion
     * This uses the method outlined in Kee et al. @cite kee2003.
```
even though this looks slightly redundant when reading the raw text.

wandadars:
If I take the standard solver, and call solve() with successively higher boundary mdot values, going from a starting value of  0.5 to around 7, and then I activate the two-point control, it has no problems with lowering the temperature control points down. Even an increment of 10 K converges without issues ( used to crash at an increment of 3K). This is all at high pressure (5.4e6 Pa). So maybe there's a possibility that the newton solver is seeing some sort of solution at a negative left boundary velocity for low boundary velocities. I'm not sure if there's anything that can be done about that.

wandadars:
I may be doing something wrong with the Python interface. To set the value of the m_mdot variable at the left and right boundaries for the 1D counterflow diffusion flame, we use the `f.fuel_inlet.mdot` and `f.oxidizer_inlet.mdot` properties correct?

In the script below, I have an initial solution at the 5.4e6 Pa, solved using the `auto=True` option. I then increment the values of the mdot variables at the boundaries by a factor of 1.25 and attempt to re-solve. I get printout results that seem to match what I am doing. The strange thing is that if I examine the backup_2.yaml file after I kill the code inside one of the loop iterations when mdot is being increased, the entry for the boundary looks like:
```

  fuel_inlet:
    type: inlet
    size: 1
    points: 1
    mass-flux: 5.895811775787321e+07
    temperature: 800.0
    pressure: 5.4e+06
    mass-fractions:
      H2: 1.0
```


```
from pathlib import Path
import cantera as ct
import numpy as np

p = 5.4e6 # pressure [Pascals]
#p = 101325 # pressure [Pascals]
#p = 7*101325 # pressure [Pascals]

fuel_inlet_temp= 800.0  # fuel inlet temperature [Kelvin]
oxidizer_inlet_temp= 711.0  # oxidizer inlet temperature [Kelvin]
fuel_mdot= 0.5 # kg/m^2/s
width= 50.0e-3 # Distance between inlets [m]

oxidizer_composition= 'O2:1.0'  # oxidizer composition (right)
fuel_composition= 'H2:1.0'  # fuel composition (left)

# Reaction mechanism
mechanism= 'h2o2.yaml'

# Define output locations
output_path = Path() / "testing_output"
output_path.mkdir(parents=True, exist_ok=True)

gas = ct.Solution(mechanism)

gas.TPY = fuel_inlet_temp, p, oxidizer_composition
density_f = gas.density

gas.TPY = oxidizer_inlet_temp, p, oxidizer_composition
density_o = gas.density

#Unity Lewis number testing
gas.transport_model = 'unity-Lewis-number'

f.set_refine_criteria(ratio=15, slope= 0.15, curve= 0.08, prune= 0.04)
#f.flame.set_steady_tolerances(default=(1e-8,1e-15))
#Set the state of the two inlets
f.fuel_inlet.mdot = fuel_mdot
f.fuel_inlet.X = fuel_composition
f.fuel_inlet.T = fuel_inlet_temp

#Create a guestimate for what the oxidizer mdot would be
f.oxidizer_inlet.mdot = (fuel_mdot / density_f) * density_o*4
f.oxidizer_inlet.X = oxidizer_composition
f.oxidizer_inlet.T = oxidizer_inlet_temp

#'''
# Generate initial condition
f.solve(auto=True, loglevel=0)
f.save('backup.yaml', name="solution", overwrite=True)

print('mdot info:')
print('Fuel mdot: ' + str(f.fuel_inlet.mdot))
print('Oxidizer mdot: ' + str(f.oxidizer_inlet.mdot))
print('BC State:')
print('Left U: ' + str(f.velocity[0]))
print('Right U: ' + str(f.velocity[-1]))
#'''

f.restore('backup.yaml')

#'''
# Generate a suite of strained flames until extinction
while(True):

    fuel_mdot = f.fuel_inlet.mdot
    oxidizer_mdot = f.oxidizer_inlet.mdot

    fuel_mdot *= 1.25
    oxidizer_mdot *= 1.25

    print('New fuel mdot: ')
    print(fuel_mdot)

    f.fuel_inlet.mdot = fuel_mdot
    f.oxidizer_inlet.mdot = oxidizer_mdot
    try:
        f.solve(loglevel=0)
    except ct.CanteraError as e:
        print('Solver did not converge. Stopping.')
        print('Error:')
        print(e)
        break
    f.save('backup_2.yaml', name="solution", overwrite=True)
#'''

f.restore('backup_2.yaml')

print('Starting two-point control')
f.two_point_control_enabled = True


spacing = 0.95
temperature_increment = 0 # Kelvin
maximum_temperature = []
a_max = []
fuel_mdot_history = []
oxidizer_mdot_history = []
control_temperature = np.min(f.T) + spacing*(np.max(f.T) - np.min(f.T))
print(f'Control temperature = {control_temperature} K')
f.set_left_control_point(control_temperature)
f.set_right_control_point(control_temperature)

f.left_control_point_temperature -= temperature_increment
f.right_control_point_temperature -= temperature_increment
f.solve(loglevel=8)
f.save('backup_twoPoint.yaml', overwrite=True)

#f.restore('backup_twoPoint.yaml')
temperature_increment = 10 # Kelvin
f.left_control_point_temperature -= temperature_increment
f.right_control_point_temperature -= temperature_increment
f.solve(loglevel=8,refine_grid=False)

```

ischoegl:
> In the script below, I have an initial solution at the 5.4e6 Pa, solved using the `auto=True` option. I then increment the values of the mdot variables at the boundaries by a factor of 1.25 and attempt to re-solve. I get printout results that seem to match what I am doing. The strange thing is that if I examine the backup_2.yaml file after I kill the code inside one of the loop iterations when mdot is being increased, the entry for the boundary looks like:

How recently have you rebased? This may be an artifact of something that @speth fixed recently in #1740 (specifically #1629). Essentially, some updated data were overwritten by older data during serialization.

include/cantera/oneD/MultiNewton.h
```diff
@@ -45,11 +45,61 @@ class MultiNewton
...
+     * Pick @f$ \alpha_k @f$ such that @f$ \norm{\inc x_{k+1}} < \norm{\inc x_k} @f$.
+     * Where @f$ \inc x_k = J^{-1}(x_k) F(x_k) @f$, and
+     * @f$ \inc x_{k+1} = J^{-1}(x_{k}) F(x_{k+1}) @f$.
+     *
+     * @param x0 initial solution about which a Newton step will be taken
+     * @param step0 initial Newton step without any damping
+     * @param x1 solution after taking the damped Newton step
+     * @param step1 Newton step after taking the damped Newton step
+     * @param loglevel controls amount of printed diagnostics
+     * @param writetitle controls if logging title is printed
```
speth: There are several undocumented parameters here (`s1`, `r`, `jac`) that generate warnings from Doxygen.

To help with clarity about what are inputs and outputs, you can mark some of these arguments with `@param[in]` or `@param[out]`. This would be most helpful for indicating the output nature of `x1` and `step1`.

wandadars:
> > In the script below, I have an initial solution at the 5.4e6 Pa, solved using the `auto=True` option. I then increment the values of the mdot variables at the boundaries by a factor of 1.25 and attempt to re-solve. I get printout results that seem to match what I am doing. The strange thing is that if I examine the backup_2.yaml file after I kill the code inside one of the loop iterations when mdot is being increased, the entry for the boundary looks like:
> 
> How recently have you rebased? This may be an artifact of something that @speth fixed recently in #1740 (specifically #1629). Essentially, some updated data were overwritten by older data during serialization.

I will try that. The issue seems to be that, even though I have specified that overwrite=True option, the backup_2.yaml file is not being overwritten. The large values were from a previous run that ran up past the extinction mdots and went to very high values of mdot during the loop.

include/cantera/oneD/MultiNewton.h
```diff
@@ -45,11 +45,61 @@ class MultiNewton
...
+     *
+     * Returns:
+     * - int : Status code
+     *   - `1` if a damping coefficient is found and the solution converges.
+     *   - `0` if a damping coefficient is found but the solution does not converge.
+     *   - `-2` if no suitable damping coefficient is found within the maximum iterations.
+     *   - `-3` if the current solution `x0` is too close to the boundary and the step points out of the allowed domain.
```
speth: This addition is very helpful -- these return codes are otherwise very opaque.

You should use the `@returns` Doxygen keyword, and look at the formatting of the resulting output.

For the `0` case, I'd suggest saying that the solution "has not converged yet". Right now, this sounds like a failure.

For the `-3` case, I'd suggest some terminology changes -- "boundary" and "domain" make this sound like we're talking about the physical domain for the 1D problem. Perhaps something like "...`x0` is close to the solution bounds and the step would exceed the bounds on one or more components"

wandadars:
> If I take the standard solver, and call solve() with successively higher boundary mdot values, going from a starting value of 0.5 to around 7, and then I activate the two-point control, it has no problems with lowering the temperature control points down. Even an increment of 10 K converges without issues ( used to crash at an increment of 3K). This is all at high pressure (5.4e6 Pa). So maybe there's a possibility that the newton solver is seeing some sort of solution at a negative left boundary velocity for low boundary velocities. I'm not sure if there's anything that can be done about that.

I spoke too soon on this. Examining the solution, the left boundary is still going negative, it just is somehow converging to a solution that has a negative u at the left boundary. Somehow that dF(u)/dLambda is getting flipped around.

include/cantera/oneD/OneDim.h
```diff
@@ -46,7 +46,11 @@ class OneDim
     MultiNewton& newton();
 
     /**
-     * Solve F(x) = 0, where F(x) is the multi-domain residual function.
+     * Solve F(x) = 0, where F(x) is the multi-domain residual function. Returns
+     * a 1 on success, and either a -2 or -3 on failure depending on whether the
+     * maximum number of damping steps was reached(-2) or if the solution was up
+     * against the bounds(-3).
+     *
```
speth: I'd recommend moving the extra documentation here to a `@returns` section, after the parameters.

include/cantera/oneD/Sim1D.h
```diff
@@ -168,6 +168,12 @@ class Sim1D : public OneDim
      */
     AnyMap restore(const string& fname, const string& name);
 
+    void clearDebugFile();
+
+    void writeDebugInfo(const string& header, const string& message, int loglevel,
+                        int attempt_counter);
+
+
```
speth: Every function should have a docstring.

src/oneD/MultiNewton.cpp
```diff
@@ -224,81 +224,88 @@ int MultiNewton::dampStep(const double* x0, const double* step0,
...
     }
 
     // If a damping coefficient was found, return 1 if the solution after
     // stepping by the damped step would represent a converged solution, and
     // return 0 otherwise. If no damping coefficient could be found, return -2.
     if (m < NDAMP) {
         if (s1 > 1.0) {
+            if (loglevel > 0) {
+                writelog("\nDamping coefficient found, but solution does not converge.\n");
+            }
```
speth: I think you're misinterpreting what is meant in this case -- it represents a successful attempt at finding a damping coefficient, which is then used to take a Newton step.

Also, to avoid these repetitive conditionals, you can use the `debuglog` function, which takes the loglevel as a second argument. You only need to use this structure when there are formatting arguments that need to be provided.

src/oneD/MultiNewton.cpp
```diff
@@ -204,14 +204,14 @@ int MultiNewton::dampStep(const double* x0, const double* step0,
...
-                "log10(s0)","log10(s1)","N_jac","Age");
-        writeline('-', 65);
+    if (loglevel > 0) {
+        writelog("\n\n{}\n", "Damped Newton iteration:");
+        writeline('-', 83, false);
+
+        writelog("\n{:<6s}  {:<9s}    {:<9s}      {:<9s}   {:<9s}   {:<9s}  {:<5s}  {:<5s}\n",
+                 "Iter", "F_damp", "F_bound", "log10(ss)",
+                "log10(s0)", "log10(s1)", "N_jac", "Age");
+        writeline('-', 83, false);
```
speth: I don't think re-starting the table every time is ideal. I understand the motivation for updating the formatting of the logging output, but I think you need to look at what this ends up looking like at each debug level. At log level 3, this now ends up repeating blocks like:
```
Damped Newton iteration:
-----------------------------------------------------------------------------------
Iter    F_damp       F_bound        log10(ss)   log10(s0)   log10(s1)  N_jac  Age  
-----------------------------------------------------------------------------------
0       1           1             6.51253     4.13493     3.48696    9      2/20
Damping coefficient found, but solution does not converge.


```
whereas previously this would be just a single line in the table as the solver takes successive Newton steps.

src/oneD/MultiNewton.cpp
```diff
@@ -350,23 +346,21 @@ int MultiNewton::solve(double* x0, double* x1, OneDim& r, MultiJac& jac, int log
...
             } else {
                 break;
             }
         }
     }
 
-    if (m < 0) {
+    if (status < 0) { // Why is this done?
         copy(m_x.begin(), m_x.end(), x1);
     }
```
speth: I think the point of this is to reset the failed step that would be in `x1` otherwise. This way, the return value is the last successful step, even if that doesn't end up getting used.

src/oneD/Sim1D.cpp
```diff
@@ -415,6 +415,14 @@ void Sim1D::solve(int loglevel, bool refine_grid)
     m_nsteps = 0;
     finalize();
 
+    // For debugging outputs
+    int attempt_counter = 0;
+    const int max_history = 10; // Store up to 10 previous solutions
+    std::string header;
```
speth: `header` isn't used.

speth (changes requested):
Thanks, @wandadars. The expanded docstrings for `MultiNewton` are very welcome, and I think the updates to how `debug_sim1d.yaml` is updated are looking pretty good. Please do look at the rendered output, either by downloading the "docs" artifact (click on "Checks" on this page, then on the "Artifacts" dropdown in the top center of the page) or by building the Doxygen docs locally. There are a number of formatting issues which you'll easily see and be able to address.

For the changes to the logging, I'd suggest a bit more work, and taking a look at what these files contain at all log levels (1-5). We want to make sure that the lower log levels are still fairly concise. That may involve leaving a few things that are less clear at high log levels, where some of the data gets interleaved a bit. One thought that came to mind was for the "tabular" outputs, you could try putting some short, distinct prefix (say 2 characters) on those lines so you could at least tell what lines were coming from which part of the solver. No idea if this would end up being horribly ugly or actually helpful.

include/cantera/oneD/MultiNewton.h
wandadars: Yeah. I was in Microsoft equation editor autopilot apparently when I wrote that.

include/cantera/oneD/MultiNewton.h
wandadars: One thing that I couldn't quite map from the Kee book was the condition for stopping the iteration. I think that obviously, given a vector of residuals, if a measure of the length of that vector drops below a threshold, then a solution may be considered to be converged. The condition in this method is that the length of the subsequent Newton step being less than 1.0 means that the solution has converged. I don't know why this value is important. I thought that usually a convergence criteria is something like 1e-4. Perhaps there's something in how the norm2 is computed for the length of the Newton step that translates into a norm2 less than 1 being satisfactory.

src/oneD/MultiNewton.cpp
wandadars: The thing that I was trying to keep straight is at high levels, we have loops:

steadySolve->MultiNewton::solve() -> MultiNewton::dampStep() -> MultiNewton::boundStep()

unsteadySolve->OneDim::timeStep() ->MultiNewton::solve()-> MultiNewton::dampStep() -> MultiNewton::boundStep()

And the logging output within the timestepping loop is where things get dicey because there's a ton of output within a single timestep attempt and that's what I was trying to make clearer.

wandadars: I think I cleaned it up with the newest commit. Hopefully more readable and less cluttered.

include/cantera/oneD/Sim1D.h
```diff
@@ -168,6 +168,37 @@ class Sim1D : public OneDim
      */
     AnyMap restore(const string& fname, const string& name);
 
+    /**
+     * Deletes a debug_sim1d.yaml file if it exists. Used to clear the file for
```
ischoegl:
```suggestion
     * Deletes a `debug_sim1d.yaml` file if it exists. Used to clear the file for
```

include/cantera/oneD/Sim1D.h
```diff
@@ -168,6 +168,37 @@ class Sim1D : public OneDim
...
+     * successive calls to the solve() method.
+     */
+    void clearDebugFile();
+
+    /**
+     * Write solver debugging information to a YAML file based on the specified log
+     * level.
+     *
+     * This method writes solver debug information to a specified YAML file
+     * (debug_sim1d.yaml). The section headers are formatted according to the provided
```
ischoegl:
```suggestion
     * (`debug_sim1d.yaml`). The section headers are formatted according to the provided
```

ischoegl (commented):
Minor suggestions.

src/oneD/MultiNewton.cpp
```diff
@@ -205,13 +218,14 @@ int MultiNewton::dampStep(const double* x0, const double* step0,
...
     // write header
     if (loglevel > 0 && writetitle) {
-        writelog("\n\nDamped Newton iteration:\n");
-        writeline('-', 65, false);
-
-        writelog("\n{}  {:>9s}   {:>9s}     {:>9s}   {:>9s}   {:>9s}  {:>5s} {:>5s}\n",
-                "m","F_damp","F_bound","log10(ss)",
-                "log10(s0)","log10(s1)","N_jac","Age");
-        writeline('-', 65);
+        string separator = fmt::format("  {:->80}", "");
```
ischoegl: The default width of many terminals is 80 characters (for historical reasons). This statement would create a line length that's slightly longer, correct? If so, I'd recommend reducing to 80 to prevent ugly line breaks, and adjust the remainder of the formatting accordingly. (sorry for not seeing this earlier)

wandadars: I was thinking about that. I wasn't sure, with modern monitors if the 80 character width is still a hard limit.

wandadars: I will try to compress the data down some more. I liked the tab over for the outputs to help with nesting the outputs to help visually separate the different parts of the solver procedure.

wandadars:
Ok, I think I got the line lengths down a bit. I made a new compressions, such as just saying "log(ss)", because we are really just looking at the change in that value, and not the magnitude for the most part. We compare to other values, and so the exact type of logarithm didn't seem critical.

I have some long lines, but I'm not sure what the best way to split them would be so as to not make reading the code overly complicated. I might be missing some fmt-fu that would help to better handle the output positioning information.  I've reviewed the different levels from 0 to 6 and I think they behave more clearly compared to the old outputs.

src/oneD/MultiNewton.cpp
```diff
@@ -205,13 +232,14 @@ int MultiNewton::dampStep(const double* x0, const double* step0,
...
-                "log10(s0)","log10(s1)","N_jac","Age");
-        writeline('-', 65);
+        string header = fmt::format("  {:->23}Damped Newton iteration{:->24}", "", "");
+        string separator = fmt::format("  {:->70}", "");
+        writelog("\n\n{}", header);
+
+        writelog("\n  {:<4s}  {:<10s}   {:<10s}  {:<7s}  {:<7s}  {:<7s}  {:<5s}  {:<3s}\n",
+                 "Iter", "F_damp", "F_bound", "log(ss)",
+                "log(s0)", "log(s1)", "N_jac", "Age");
+        writelog(separator);
```
speth: A couple of simplifications can be made here, I think:
```suggestion
        writelog("\n\n{:-^70}", " Damped Newton iteration ");
        writelog("\n  {:<4s}  {:<10s}   {:<10s}  {:<7s}  {:<7s}  {:<7s}  {:<5s}  {:<3s}\n",
                 "Iter", "F_damp", "F_bound", "log(ss)",
                "log(s0)", "log(s1)", "N_jac", "Age");
        writelog("  {:->70}", "");
```

src/oneD/MultiNewton.cpp
```diff
@@ -350,23 +369,26 @@ int MultiNewton::solve(double* x0, double* x1, OneDim& r, MultiJac& jac, int log
...
+                }
             } else {
                 break;
             }
         }
     }
+    // Close off the damped iteration table that is written by the dampedStep() method
+    if (loglevel > 1) {
+        string separator = fmt::format("\n  {:->70}", "");
+        writelog(separator);
```
speth:
```suggestion
        writelog("\n  {:->70}", "");
```

src/oneD/OneDim.cpp
```diff
@@ -338,33 +337,44 @@ double OneDim::timeStep(int nsteps, double dt, double* x, double* r, int logleve
...
-
     int n = 0;
     int successiveFailures = 0;
 
+    // Only output this if nothing else under this function call will be output
+    if (loglevel == 1){
+        debuglog("\n============================", loglevel);
+        debuglog(fmt::format("\n{:<5s}  {:<11s}   {:<7s}\n", "step", "dt (s)", "log(ss)"), loglevel);
+        debuglog("============================", loglevel);
+    }
```
speth: You don't need to use `debuglog` if you're already using an `if (loglevel)` conditional, in which case you can also elide the explicit call to `fmt::format`.

Also, as a minor formatting nit, there should be a space between the `)` and the `{` in the `if` statement.

src/oneD/OneDim.cpp
```diff
@@ -377,24 +387,40 @@ double OneDim::timeStep(int nsteps, double dt, double* x, double* r, int logleve
...
-            successiveFailures++;
             // No solution could be found with this time step.
             // Decrease the stepsize and try again.
-            debuglog("...failure.\n", loglevel);
+            successiveFailures++;
+            if (loglevel == 1) {
+                writelog("\nTimestep failed");
+            }else if (loglevel > 1) {
+                writelog("\nTimestep ({}) failed", n);
+            }
```
speth:
```suggestion
            } else if (loglevel > 1) {
                writelog("\nTimestep ({}) failed", n);
            }
```

src/oneD/OneDim.cpp
```diff
@@ -377,24 +387,40 @@ double OneDim::timeStep(int nsteps, double dt, double* x, double* r, int logleve
...
                 }
             }
         }
     }
 
+    // Write the final step to the log
+    if (loglevel == 1) {
+        double ss = ssnorm(x, r);
+        writelog("\n{:<5d}  {:<6.4e}   {:>7.4f}", n, dt, log10(ss));
+        debuglog("\n============================", loglevel);
```
speth: Again, you can just use `writelog` here.

include/cantera/oneD/MultiNewton.h
speth: It looks like you fixed several other instances of this, but not the ones on this specific line.

include/cantera/oneD/MultiNewton.h
```diff
@@ -45,11 +45,64 @@ class MultiNewton
...
+     * Where `F` is the system of nonlinear equations, `x` is the solution vector.
+     *
+     * For the damped Newton method we are solving:
+     *
+     * @f[
+     *   x_{k+1} - x_k = \Delta x_k = -\alpha_k J^(-1)(x_k) F(x_k)
+     * @f]
+     *
+     * Where @f$ J @f$ is the Jacobian matrix of @f$ F @f$ with respect to @f$ x @f$, and @f$ \alpha_k @f$ is
+     * the damping factor, and @f$ \Delta x_k @f$ is the Newton step at @fx_k @f, sometimes
```
speth: Please check for warnings from Doxygen.
```suggestion
     * the damping factor, and @f$ \Delta x_k @f$ is the Newton step at @f x_k @f, sometimes
```

speth (changes requested):
Thanks, @wandadars, I think the logging output at higher levels is quite a bit more readable now. I did have a couple thoughts about adjusting what gets added at each level, in contrast to the current levels.

One change here is that the information about what components are hitting their bounds is pushed to a higher loglevel -- from 5 for steady and 6 for transient solves to 6 and 7. While I don't think this is inherently a problem, it means that you aren't getting the ones for the transient solves until you are also generating a `debug_sim1d.yaml` file, which I don't think we want. I would suggest either moving these outputs back to their current level, or bumping the log levels for the output file up one more level.

Also, it seems like you have to go to fairly high log levels (4+) to get any information about the Newton iterations, and that comes after the (very verbose) outputs of the system state after each steady state solve that comes with log level 3. I think the current approach where the abbreviated info on Newton iterations comes in at loglevel 2 makes more sense, at the same time that we're providing an abbreviated summary of each timestep. We could also consider deferring the system state printout to a higher debug level.

wandadars:
Thank you for the review comments @speth . One thing that was in my mind while adjusting these print statements was the question: "What is the guiding principle of the log levels here?"

One thing that I tried to do was if a function gets a log level of 1, then it can be succinct in its summary because the statements don't have to worry about lower-level function calls injecting their own print statements into the flow of things. If the log level is higher than that, then the outputs at that level need to be expressed in such a way that they can still be understood even with a bunch of extra statements in-between. 

I adjusted where the loglevel was decremented ( I had it essentially decrementing during each nested call down into a function from the highest solve() level before). We have the steady newton summary and the timestep at logleve 2 now. I moved the full state output to one level higher.

speth:
> One thing that I tried to do was if a function gets a log level of 1, then it can be succinct in its summary because the statements don't have to worry about lower-level function calls injecting their own print statements into the flow of things. If the log level is higher than that, then the outputs at that level need to be expressed in such a way that they can still be understood even with a bunch of extra statements in-between.

Yes, I understood that to be one of your main goals with the modifications to the logging, and I think what you have here is an improvement in that regard, by making some of the individual log records more self-contained and by the adjustments to the formatting of the "tabular" output.

speth (approved):

wandadars:
@speth  I have that script that parses and plots the output from the debug_sim1d.yaml file. Is that something worth including somewhere in Cantera? It's probably more of a tool for developers versus users.

speth:
Do you have a link to the script? I think you posted some version of it in one of these threads, but I can't find it now. I think it would be worth preserving in some form. Perhaps on a page within the "Develop" section of the docs (https://cantera.org/dev/develop/index.html) along with some other suggestions on how to dig into unexpected issues with using the 1D code.

wandadars:
> Do you have a link to the script? I think you posted some version of it in one of these threads, but I can't find it now. I think it would be worth preserving in some form. Perhaps on a page within the "Develop" section of the docs (https://cantera.org/dev/develop/index.html) along with some other suggestions on how to dig into unexpected issues with using the 1D code.

For reference (when I get around to adding a page on the script), this is the script.

```
import yaml
import sys
import os
import re

'''
This script reads the debug output from Cantera when the loglevel is set to 7 or 8.
The data that is output is the history of the previous attempts of the solve()
method's residual and solution vectors.
'''

# Load YAML data
def load_yaml_data(file_path):
    """Load and return the YAML data from the given file path."""
    try:
        with open(file_path, 'r') as file:
            return yaml.safe_load(file)
    except FileNotFoundError:
        print(f"Error: The file '{file_path}' does not exist.")
        sys.exit(1)
    except yaml.YAMLError as exc:
        print(f"Error: Failed to parse YAML file '{file_path}'.\n{exc}")
        sys.exit(1)

# Plot component data against grid points and save the plot
def plot_component(grid, component_data, component_name, data_type, plot_dir):
    """Plot a component's data against grid points and save the plot."""
    plt.figure()

    # Plot line and scatter points
    plt.plot(grid, component_data, label=component_name)
    plt.scatter(grid, component_data, color='red', marker='o', label=f'{component_name} (data points)')

    plt.xlabel('Grid Points')
    plt.ylabel(component_name)
    plt.title(f'{component_name} {data_type} vs Grid Points')  # Include data type
    plt.legend()

    plot_filename = os.path.join(plot_dir, f'{component_name}_{data_type}_vs_grid.png')
    plt.savefig(plot_filename)
    plt.close()

def main():
    if len(sys.argv) < 2:
        print("Usage: python script.py <path_to_yaml_file>")
        sys.exit(1)

    file_path = sys.argv[1]
    data = load_yaml_data(file_path)

    all_data = {}
    for key, value in data.items():
        match = re.match(r'(solution|residual)_(\d+)_(.+)', key)
        if match:
            data_type, solution_num, stage = match.groups()
            solution_num = int(solution_num)
            if solution_num not in all_data:
                all_data[solution_num] = {}
            if stage not in all_data[solution_num]:
                all_data[solution_num][stage] = {}
            all_data[solution_num][stage][data_type] = value['flame']

    for solution_num, stages in all_data.items():
        print(f'Plotting data for solve attempt: {solution_num}')

        for stage, data_types in stages.items():
            for data_type, oneD_data in data_types.items():
                components = oneD_data.get('components')
                grid = oneD_data.get('grid')

                if not components or not grid:
                    print(f"Warning: Missing 'components' or 'grid' data for solution {solution_num}, stage {stage}, type {data_type}. Skipping.")
                    continue

                # Separate directories for "solution" and "residual"
                plot_dir = os.path.join('debug_plots', f'solution_{solution_num}', stage, data_type)
                os.makedirs(plot_dir, exist_ok=True)

                for component in components:
                    if component in oneD_data and component != 'grid':
                        plot_component(grid, oneD_data[component], component, data_type, plot_dir)

                print(f"Plots for {data_type} {solution_num}, stage {stage}, type {data_type} saved in directory: {plot_dir}")

if __name__ == "__main__":
    main()

```

PR #1760, opened 2024-08-05
Fix inconsistent string handling in `clib`

ischoegl:

- Fix instances where `clib` functions use inconsistent string handling, i.e. `ct_getDataDirectories` and `thermo_report`
- Replace by `ct_getDataDirectories3` and `thermo_report3` that have buffer length and char buffer in last two positions.
- Add threshold to MATLAB thermo report for more compact output (consistent with Python)

The change ensures that a single helper function `ctString.m` can be used as the interface to pass strings from C++ to MATLAB.

*Experimental* MATLAB now shows compact report (suppressing minor species):
```
>> gas = Solution('gri30.yaml')

  gri30:

       temperature   300 K
          pressure   1.0133e+05 Pa
           density   0.081894 kg/m^3
  mean mol. weight   2.016 kg/kmol
   phase of matter   gas

                          1 kg             1 kmol     
          enthalpy             26469             53361  J
   internal energy       -1.2108e+06        -2.441e+06  J
           entropy             64910        1.3086e+05  J/K
    Gibbs function       -1.9447e+07       -3.9204e+07  J
 heat capacity c_p             14311             28851  J/K
 heat capacity c_v             10187             20536  J/K

                H2                 1                 1           -15.717
     [  +52 minor]                 0                 0  
```

**Checklist**

### DIFF
--- a/include/cantera/clib/ct.h
+++ b/include/cantera/clib/ct.h
@@ -66,7 +66,7 @@ extern "C" {
     CANTERA_CAPI size_t thermo_elementIndex(int n, const char* nm);
     CANTERA_CAPI size_t thermo_speciesIndex(int n, const char* nm);
     //! @since Changed signature in %Cantera 3.1
-    CANTERA_CAPI int thermo_report3(int nth, int show_thermo, double threshold, int ibuf, char* buf);
+    CANTERA_CAPI int thermo_report(int nth, int show_thermo, double threshold, int ibuf, char* buf);
     CANTERA_CAPI int thermo_print(int nth, int show_thermo, double threshold);
     CANTERA_CAPI double thermo_nAtoms(int n, size_t k, size_t m);
     CANTERA_CAPI int thermo_addElement(int n, const char* name, double weight);
@@ -182,7 +182,7 @@ extern "C" {
     CANTERA_CAPI int ct_setLogCallback(LogCallback writer);
     CANTERA_CAPI int ct_addCanteraDirectory(size_t buflen, const char* buf);
     //! @since Changed signature in %Cantera 3.1
-    CANTERA_CAPI int ct_getDataDirectories3(const char* sep, int buflen, char* buf);
+    CANTERA_CAPI int ct_getDataDirectories(const char* sep, int buflen, char* buf);
     CANTERA_CAPI int ct_getCanteraVersion(int buflen, char* buf);
     CANTERA_CAPI int ct_getGitCommit(int buflen, char* buf);
     CANTERA_CAPI int ct_suppress_thermo_warnings(int suppress);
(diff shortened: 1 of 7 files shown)


speth:
What level of compatibility are we trying to maintain for the clib API between Cantera 3.0 and 3.1? So far, I think most of the changes have been consistent with what we do in the Python / C++ interface where user code should continue to work as long as it wasn't using any functions that were deprecated in Cantera 3.0. However, I'd also be OK with adopting a less conservative approach for clib to enable more rapid improvements.

ischoegl:
> What level of compatibility are we trying to maintain for the clib API between Cantera 3.0 and 3.1? So far, I think most of the changes have been consistent with what we do in the Python / C++ interface where user code should continue to work as long as it wasn't using any functions that were deprecated in Cantera 3.0. However, I'd also be OK with adopting a less conservative approach for clib to enable more rapid improvements.

Good question. With the legacy MATLAB toolbox gone, I believe that all of the remaining API's are de-facto "experimental" (.NET is unfinished and Fortran only captures a small subset of features). We can certainly include placeholders so symbols are still in place, but at the same time breaking changes may be a feature. Specifically, I think it's dangerous to use the 3.1 version of the experimental MATLAB toolbox with a 3.0 `clib` (or, more realistically, a 3.0 experimental toolbox with 3.1 `clib`). Until we have non-experimental API's that run through `clib`, I believe we should be able to cycle more rapidly. Changes in this PR are one case in point: the old `clib` was inconsistent, so specialized calls were necessary for a handful of methods that didn't fit the norm.

include/cantera/clib/ct.h
```diff
@@ -65,8 +65,8 @@ extern "C" {
     CANTERA_CAPI int thermo_setName(int n, const char* nm);
     CANTERA_CAPI size_t thermo_elementIndex(int n, const char* nm);
     CANTERA_CAPI size_t thermo_speciesIndex(int n, const char* nm);
-    CANTERA_CAPI int thermo_report(int nth,
-                                  int ibuf, char* buf, int show_thermo);
+    //! @since Changed signature in %Cantera 3.1
+    CANTERA_CAPI int thermo_report3(int nth, int show_thermo, double threshold, int ibuf, char* buf);
```
speth: So, taking the approach of not requiring `clib` to have a stable API, we should be able to change the argument order without having to rename the function. Anyone using this function will get a compiler error if they try to call the function with the old argument order.

Likewise for `ct_getDataDirectories3`.

speth (changes requested):
Thanks, @ischoegl. It's nice to see the benefit of removing these special cases in simplifying the Matlab toolbox. Just one suggestion given the discussion on clib's API stability.

speth (approved):

PR #1765, opened 2024-08-08
Reproducible names for zeroD objects

### LINKED ISSUE #1764: Broken name attribute for reactor `Wall` objects
**Problem description**

Reactor `Wall` objects do have documented name attributes, but they are not retrievable.

**Steps to reproduce**

On IPython, run
```
In [1]: %run samples/python/reactors/piston.py
time [s]    T1 [K]      T2 [K]      V1 [m^3]    V2 [m^3]    V1+V2 [m^3]  X(CO)
     0.004       900.0    900.0015         0.5         0.1         0.6      0.2853
[...]
     0.200      2319.4   2774.2346      0.4917      0.1083         0.6     0.03109
To view a plot of these results, run this script with the option --plot

In [2]: w.name  # 'w' is a variable holding a 'Wall' object
Cell In[2], line 1
----> 1 w.name

AttributeError: 'cantera.reactor.Wall' object has no attribute 'name'
```

**Behavior**

Although `name` is a documented parameter, it cannot be retrieved (even if no name is provided, a default name should be generated automatically).
https://github.com/Cantera/cantera/blob/32a8d4b8fde73e52ee7aa73dfd650bce68e46a03/interfaces/cython/cantera/reactor.pyx#L947-L972

**System information**

- Cantera version: current `main`
- OS: macOS
- Python/MATLAB/other software versions: Python 3.11

**Additional context**

Originally reported on UG, [see post](https://groups.google.com/g/cantera-users/c/V8T2Y2R1jhc)

ischoegl:

This PR seeks to implement reproducible (and unique) reactor, flow device and wall names in the C++ layer, which was previously only handled by the Python API. Rather than auto-creating unique names based on the number of certain object types created within a Cantera session, this PR switches to an approach that creates names dependent on `ReactorNet` objects, which guarantees that reproducible auto-generated names are attached to various `zeroD` objects.

- Ensure that reactors, walls and connectors have names that are stored in C++ layer
- Names are generated automatically if none are provided
- Names should be reproducible for a given `ReactorNet`, which is important for serialization (e.g. Cantera/enhancements#206)
- Update API's for Python and MATLAB
- ~Remove deprecated functions from `clib` (in anticipation of Cantera/enhancements#211)~

Closes #1764

**Checklist**

### DIFF
--- a/include/cantera/zeroD/FlowDevice.h
+++ b/include/cantera/zeroD/FlowDevice.h
@@ -45,6 +45,9 @@ public:
         m_name = name;
     }
 
+    //! Set the default name of a flow device. Returns `false` if it was previously set.
+    bool setDefaultName(map<string, int>& counts);
+
     //! Mass flow rate (kg/s).
     double massFlowRate() {
         if (m_mdot == Undef) {
@@ -131,6 +134,7 @@ public:
 
 protected:
     string m_name;  //!< Flow device name.
+    bool m_defaultNameSet = false;  //!< `true` if default name has been previously set.
 
     double m_mdot = Undef;
 
--- a/include/cantera/zeroD/FlowDevice.h
+++ b/include/cantera/zeroD/FlowDevice.h
@@ -23,7 +23,7 @@ class ReactorBase;
 class FlowDevice
 {
 public:
-    FlowDevice() = default;
+    FlowDevice(const string& name="(none)") : m_name(name) {}
 
     virtual ~FlowDevice() = default;
     FlowDevice(const FlowDevice&) = delete;
@@ -33,6 +33,16 @@ public:
     //! corresponds to the name of the derived class.
     virtual string type() const {
         return "FlowDevice";
+    }
+
+    //! Retrieve flow device name.
+    string name() const {
+        return m_name;
+    }
+
+    //! Set flow device name.
+    void setName(const string& name) {
+        m_name = name;
     }
 
     //! Mass flow rate (kg/s).
@@ -77,6 +87,11 @@ public:
         return *m_out;
     }
 
+    //! Return a mutable reference to the downstream reactor.
+    ReactorBase& out() {
+        return *m_out;
+    }
+
     //! Return current value of the pressure function.
     /*!
      * The mass flow rate [kg/s] is calculated given the pressure drop [Pa] and a
@@ -115,6 +130,8 @@ public:
     }
 
 protected:
+    string m_name;  //!< Flow device name.
+
     double m_mdot = Undef;
 
     //! Function set by setPressureFunction; used by updateMassFlowRate
--- a/include/cantera/zeroD/ReactorBase.h
+++ b/include/cantera/zeroD/ReactorBase.h
@@ -50,12 +50,12 @@ struct SensitivityParameter
 class ReactorBase
 {
 public:
-    explicit ReactorBase(const string& name = "(none)");
+    explicit ReactorBase(const string& name="(none)");
     //! Instantiate a ReactorBase object with Solution contents.
     //! @param sol  Solution object to be set.
     //! @param name  Name of the reactor.
     //! @since New in %Cantera 3.1.
-    ReactorBase(shared_ptr<Solution> sol, const string& name = "(none)");
+    ReactorBase(shared_ptr<Solution> sol, const string& name="(none)");
     virtual ~ReactorBase();
     ReactorBase(const ReactorBase&) = delete;
     ReactorBase& operator=(const ReactorBase&) = delete;
--- a/include/cantera/zeroD/ReactorBase.h
+++ b/include/cantera/zeroD/ReactorBase.h
@@ -76,6 +76,9 @@ public:
         m_name = name;
     }
 
+    //! Set the default name of a reactor. Returns `false` if it was previously set.
+    bool setDefaultName(map<string, int>& counts);
+
     //! Set the Solution specifying the ReactorBase content.
     //! @param sol  Solution object to be set.
     //! @since New in %Cantera 3.1.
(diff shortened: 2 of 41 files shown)


ischoegl:
The PR goes somewhat beyond fixing #1764, as it pushes all names to C++ and exposes them to MATLAB as well. It should be ready for a review.

ischoegl:
Minor cleanup. As an aside, I believe we should deprecate creation of empty `Wall` and `FlowDevice` objects as well (similar to the recent decision about empty `Reactor` objects); Python doesn't allow for it (and this PR removes support from MATLAB), so we should not support it at the C++ level either. A corresponding change goes beyond the scope of this PR.

ischoegl:
Fixed Python docstrings. _Edit: another force-push to fix a new CI error. Not anticipating any other changes beyond 882182d before a review._

ischoegl:
Rebased, with commits removing obsolete CLib methods moved to #1773.

include/cantera/zeroD/FlowDevice.h
```diff
@@ -115,6 +133,9 @@ class FlowDevice
     }
 
 protected:
+    string m_name = "";  //!< Flow device name.
```
speth: Since the one and only constructor sets a value for `m_name`, there's no need set it here. Also, the default value for a `std::string` is the empty string anyway.

include/cantera/zeroD/ReactorBase.h
```diff
@@ -76,6 +76,9 @@ class ReactorBase
         m_name = name;
     }
 
+    //! Set the default name of a reactor. Returns `false` if it was previously set.
+    bool setDefaultName(const string& typ, int count);
```
speth: Does this method need to take `typ` as an argument? It could just call `type()` internally instead.

src/zeroD/ReactorBase.cpp
```diff
@@ -31,6 +37,18 @@ ReactorBase::~ReactorBase()
...
+bool ReactorBase::setDefaultName(const string& typ, int count)
+{
+    if (m_defaultNameSet) {
+        return false;
+    }
+    m_defaultNameSet = true;
+    if (m_name == "(none)" || m_name == "") {
+        m_name = fmt::format("{}_{}", typ, count);
+    }
+    return true;
```
speth: The logic between this and `ReactorNet::updateNames` results in the count being incremented even for reactors that already have names. Is this the intended behavior?

src/zeroD/ReactorNet.cpp
```diff
@@ -305,6 +306,72 @@ void ReactorNet::addReactor(Reactor& r)
...
+        auto inType = in.type();
+        if (in.setDefaultName(inType, m_count[inType])) {
+            m_count[inType]++;
+        }
+        if (in.in().type() == "Reservoir") {
+            auto& res = in.in();
+            auto resType = res.type();
+            res.setDefaultName(resType, m_count[resType]);
+            m_count[resType]++;
+        }
```
speth: One possibility for simplifying this somewhat repetitious pattern would be to have `setDefaultName` take the `m_count` map as the one and only argument, and be responsible for updating the count for that reactor/wall/etc. type.

src/zeroD/ReactorNet.cpp
```diff
@@ -305,6 +306,72 @@ void ReactorNet::addReactor(Reactor& r)
...
+    }
+
+    for (size_t i=0; i<r.nSurfs(); i++) {
+        auto surf = r.surface(i);
+        auto surfType = surf->type();
+        if (surf->name() == "(none)" || surf->name() == "") {
+            surf->setName(fmt::format("{}_{}", surfType, m_count[surfType]));
+        }
+        m_count[surfType]++;
+    }
```
speth: Should `ReactorSurface` objects get the same treatment of having a `setDefaultName` method?

test/clib/test_ctreactor.cpp
```diff
@@ -13,6 +13,15 @@ TEST(ctreactor, reactor_soln)
     int sol = soln_newSolution("gri30.yaml", "gri30", "none");
     int reactor = reactor_new3("IdealGasReactor", sol, "test");
     ASSERT_EQ(reactor, 0);
+
+    int ret = reactor_setName(reactor, "spam");
+    ASSERT_EQ(ret, 0);
+    int buflen = reactor_name(reactor, 0, 0);
+    char* buf = new char[buflen];
```
speth: Instead of `new` and `delete`, you can use `vector<char> buf(buflen)`. I think there are one or two other `clib` tests where that's what we use.

test/python/test_reactor.py
```diff
@@ -236,10 +236,34 @@ def test_maxsteps(self):
...
+        assert self.r1.name.startswith(f"{self.r1.type}_")  # default name
+        assert self.r2.name.startswith(f"{self.r2.type}_")  # default name
+        assert self.w.type == "Wall"
+        assert self.w.name.startswith("Wall_")  # default name
+        self.w.name = "name-of-wall"
+        assert self.w.name == "name-of-wall"
+
+    def test_wall_type2(self):
+        self.make_reactors(n_reactors=1)
+        res = ct.Reservoir(self.gas1)  # warning raised from C++ code
```
speth: What warning, and why?

speth (changes requested):
Thanks, @ischoegl. Moving this naming process to the C++ level certainly makes sense. Just a few comments on the implementation.

src/zeroD/ReactorBase.cpp
ischoegl: This is indeed the intended behavior.

test/clib/test_ctreactor.cpp
ischoegl: :+1: There are quite a few instances.

src/zeroD/ReactorNet.cpp
ischoegl: :tada: Great suggestion!

include/cantera/zeroD/ReactorBase.h
ischoegl: Per your other suggestion using the map as argument instead.

test/python/test_reactor.py
ischoegl: Good catch. Likely copy-pasta.

ischoegl:
@speth ... thanks for the review! Your suggestion to simplify `setDefaultName` is much appreciated!

ischoegl:
@speth ... assuming that this is ready to merge, I'll use this to start on Cantera/enhancements#213.

speth (approved):

PR #1778, opened 2024-08-15
Implement local AnyMap ordering rules

### LINKED ISSUE #1776: Ordering of SolutionArray components needs to be revisited.
**Problem description**

#1775 introduced a potential memory leak via incorrect use of `AnyMap::addOrderingRules`, see https://github.com/Cantera/cantera/pull/1775#discussion_r1718528240

ischoegl:

- Allow ordering rules defined for individual `AnyMap` instances. 
- 'Local' rules override global rules.

Closes #1776

**Checklist**

### DIFF
--- a/include/cantera/base/AnyMap.h
+++ b/include/cantera/base/AnyMap.h
@@ -698,8 +698,24 @@ public:
     static bool addOrderingRules(const string& objectType,
                                  const vector<vector<string>>& specs);
 
+    //! Set local rules for element order when outputting AnyMap objects to YAML.
+    //! Rules follow the same logic as those implemented by addOrderingRules(), and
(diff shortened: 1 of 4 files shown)


speth:
I think this might be more complexity than we need in manipulating the output order of `AnyMap` items. The main way of ordering items is to add them to the `AnyMap` in the order they should be output. The additional option of using `addOrderingRules` is there to handle cases where it's difficult to just add everything in order because the data is coming from multiple sources, such as different base / derived classes.

It seems like all that's really needed in the case of the 1D output is to promote a few of the metadata fields to the front, and all of the fields with dynamic names (state vector components) will just appear in their natural order after that. I've provided an implementation of this approach in #1781.

PR #1786, opened 2024-08-22
Refiner documentation

wandadars:
This is a WIP

Mostly documentation changes, with a few variable name adjustments for clarity.

**Checklist**

### DIFF
--- a/doc/sphinx/_static/images/onedim/refinement_curve_metric_plot.svg
+++ b/doc/sphinx/_static/images/onedim/refinement_curve_metric_plot.svg
@@ -2,12 +2,12 @@
 <!-- Created with Inkscape (http://www.inkscape.org/) -->
 
 <svg
-   width="118.59274mm"
-   height="75.416153mm"
-   viewBox="0 0 118.59273 75.416153"
+   width="118.81954mm"
+   height="96.657318mm"
+   viewBox="0 0 118.81953 96.657318"
    version="1.1"
    id="svg1"
-   inkscape:version="1.3.2 (091e20e, 2023-11-25, custom)"
+   inkscape:version="1.3.2 (1:1.3.2+202311252150+091e20ef0f)"
    sodipodi:docname="refinement_curve_metric_plot.svg"
    inkscape:export-filename="refinement_diagram.png"
    inkscape:export-xdpi="150"
@@ -26,13 +26,13 @@
      inkscape:pagecheckerboard="0"
      inkscape:deskcolor="#d1d1d1"
      inkscape:document-units="mm"
-     inkscape:zoom="0.70710678"
-     inkscape:cx="333.7544"
-     inkscape:cy="474.46865"
+     inkscape:zoom="1.4142136"
+     inkscape:cx="169.70563"
+     inkscape:cy="218.84955"
      inkscape:window-width="1920"
-     inkscape:window-height="1017"
-     inkscape:window-x="-8"
-     inkscape:window-y="-8"
+     inkscape:window-height="941"
+     inkscape:window-x="0"
+     inkscape:window-y="27"
      inkscape:window-maximized="1"
      inkscape:current-layer="layer1" />
   <defs
@@ -189,23 +189,50 @@
          style="fill:context-stroke;fill-rule:evenodd;stroke:none"
          id="path8-0" />
     </marker>
+    <marker
+       style="overflow:visible"
+       id="RoundedArrow-8-2-9-6-6"
+       refX="0"
+       refY="0"
+       orient="auto-start-reverse"
+       inkscape:stockid="Rounded arrow"
+       markerWidth="1"
+       markerHeight="1"
+       viewBox="0 0 1 1"
+       inkscape:isstock="true"
+       inkscape:collect="always"
+       preserveAspectRatio="xMidYMid">
+      <path
+         transform="scale(0.7)"
+         d="m -0.21114562,-4.1055728 6.42229122,3.21114561 a 1,1 90 0 1 0,1.78885438 L -0.21114562,4.1055728 A 1.236068,1.236068 31.717474 0 1 -2,3 v -6 a 1.236068,1.236068 148.28253 0 1 1.78885438,-1.1055728 z"
+         style="fill:context-stroke;fill-rule:evenodd;stroke:none"
+         id="path8-87-1-3-9-2" />
+    </marker>
   </defs>
   <g
      inkscape:label="Layer 1"
      inkscape:groupmode="layer"
      id="layer1"
-     transform="translate(-53.455189,-240.76158)">
+     transform="translate(-53.727354,-219.52042)">
+    <path
+       style="fill:none;stroke:#000000;stroke-width:0.264583px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1"
+       d="m 57.800184,304.42636 c 0,0 19.669152,-5.52272 28.046649,-11.62467 8.377497,-6.10195 12.458429,-41.68225 20.104337,-38.673 7.64591,3.00925 10.92339,37.19674 23.59007,40.65589 12.66668,3.45915 26.85354,0.69416 26.85354,0.69416"
+       id="path1"
+       sodipodi:nodetypes="czzzc"
+       inkscape:export-filename="refinement_slope_metric_plot.svg"
+       inkscape:export-xdpi="500"
+       inkscape:export-ydpi="500" />
     <path
        style="fill:none;stroke:#000000;stroke-width:0.529167;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:1.05833, 1.05833;stroke-dashoffset:0;stroke-opacity:1"
-       d="m 73.375528,288.66412 v 19.0839"
+       d="m 66.496352,301.4563 v 6.08684"
        id="path2985-9-8-0-4-1"
        inkscape:connector-curvature="0"
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
        inkscape:export-xdpi="500"
        inkscape:export-ydpi="500" />
     <path
-       style="fill:none;stroke:#000000;stroke-width:0.529167;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:1.05833, 1.05833;stroke-dashoffset:0;stroke-opacity:1"
-       d="m 83.958872,279.26245 v 28.221"
+       style="fill:none;stroke:#000000;stroke-width:0.529169;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:1.05833, 1.05833;stroke-dashoffset:0;stroke-opacity:1"
+       d="m 80.143423,296.01231 v 11.81157"
        id="path2985-9-8-0-4-3-2"
        inkscape:connector-curvature="0"
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
@@ -213,31 +240,39 @@
        inkscape:export-ydpi="500" />
     <path
        style="fill:none;stroke:#000000;stroke-width:0.529167;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:1.05833, 1.05833;stroke-dashoffset:0;stroke-opacity:1"
-       d="m 94.542218,274.0573 v 33.69072"
+       d="m 94.177764,273.59174 v 33.94167"
        id="path2985-9-8-0-4-3-6-0"
        inkscape:connector-curvature="0"
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
        inkscape:export-xdpi="500"
        inkscape:export-ydpi="500" />
     <path
-       style="fill:none;stroke:#000000;stroke-width:0.529167;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:1.05833, 1.05833;stroke-dashoffset:0;stroke-opacity:1"
-       d="m 105.65471,277.74702 v 29.73642"
+       style="fill:none;stroke:#000000;stroke-width:0.529168;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:1.05833, 1.05833;stroke-dashoffset:0;stroke-opacity:1"
+       d="m 105.65471,254.32998 v 53.15044"
        id="path2985-9-8-0-4-3-6-8-1"
        inkscape:connector-curvature="0"
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
        inkscape:export-xdpi="500"
        inkscape:export-ydpi="500" />
     <path
-       style="fill:none;stroke:#000000;stroke-width:0.529167;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:1.05833, 1.05833;stroke-dashoffset:0;stroke-opacity:1"
-       d="m 118.35472,288.83172 v 18.65172"
+       style="fill:none;stroke:#000000;stroke-width:0.529168;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:1.05833, 1.05833;stroke-dashoffset:0;stroke-opacity:1"
+       d="m 118.88813,281.68212 v 26.0659"
        id="path2985-9-8-0-4-3-6-8-9-3"
        inkscape:connector-curvature="0"
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
        inkscape:export-xdpi="500"
        inkscape:export-ydpi="500" />
     <path
+       style="fill:none;stroke:#000000;stroke-width:0.529168;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:1.05834, 1.05834;stroke-dashoffset:0.740835;stroke-opacity:1"
+       d="m 112.15154,263.94791 v 43.5895"
+       id="path2985-9-8-0-4-3-6-8-9-3-3"
+       inkscape:connector-curvature="0"
+       inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
+       inkscape:export-xdpi="500"
+       inkscape:export-ydpi="500" />
+    <path
        style="fill:none;stroke:#000000;stroke-width:0.529167;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:1.05833, 1.05833;stroke-dashoffset:0;stroke-opacity:1"
-       d="m 132.726,286.44911 v 21.29891"
+       d="m 132.726,297.94974 v 9.9394"
        id="path2985-9-8-0-4-3-6-8-9-6-0"
        inkscape:connector-curvature="0"
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
@@ -245,7 +280,7 @@
        inkscape:export-ydpi="500" />
     <path
        style="fill:none;stroke:#000000;stroke-width:0.329;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1;marker-start:url(#RoundedArrow)"
-       d="m 132.43274,283.00575 0.41958,-21.50319"
+       d="m 117.97138,274.23352 13.54581,-5.06899"
        id="path2985-9-8-0-4-3-6-8-9-6-0-3"
        inkscape:connector-curvature="0"
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
@@ -254,7 +289,7 @@
        sodipodi:nodetypes="cc" />
     <path
        style="fill:none;stroke:#000000;stroke-width:0.329;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1;marker-start:url(#RoundedArrow-81)"
-       d="M 118.15954,285.63511 105.71862,267.91909"
+       d="M 86.75379,282.21395 75.91057,274.78198"
        id="path2985-9-8-0-4-3-6-8-9-6-0-3-5"
        inkscape:connector-curvature="0"
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
@@ -279,8 +314,8 @@
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
        inkscape:export-xdpi="500"
        inkscape:export-ydpi="500"
-       cx="73.375526"
-       cy="288.39954"
+       cx="66.496353"
+       cy="301.62891"
        rx="1.3269223"
        ry="1.3336277" />
     <ellipse
@@ -289,8 +324,8 @@
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
        inkscape:export-xdpi="500"
        inkscape:export-ydpi="500"
-       cx="83.95887"
-       cy="279.54623"
+       cx="80.122093"
+       cy="296.22815"
        rx="1.3269223"
        ry="1.3336277" />
     <ellipse
@@ -299,18 +334,18 @@
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
        inkscape:export-xdpi="500"
        inkscape:export-ydpi="500"
-       cx="94.542213"
-       cy="274.14639"
+       cx="94.277634"
+       cy="274.00165"
        rx="1.3269223"
        ry="1.3336277" />
     <ellipse
-       style="fill:#000000;fill-opacity:1;fill-rule:evenodd;stroke:#000000;stroke-width:0.529167;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1"
+       style="fill:#3ec639;fill-opacity:1;fill-rule:evenodd;stroke:#3ec639;stroke-width:0.529167;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1"
        id="path3755-74-0-9-8-4-2-1-4-6-3-3"
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
        inkscape:export-xdpi="500"
        inkscape:export-ydpi="500"
        cx="105.65472"
-       cy="277.74701"
+       cy="254.46336"
        rx="1.3269223"
        ry="1.3336277" />
     <ellipse
@@ -319,8 +354,18 @@
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
        inkscape:export-xdpi="500"
        inkscape:export-ydpi="500"
-       cx="118.35472"
-       cy="288.96286"
+       cx="118.88389"
+       cy="281.55441"
+       rx="1.3269223"
+       ry="1.3336277" />
+    <ellipse
+       style="fill:#000000;fill-opacity:1;fill-rule:evenodd;stroke:#000000;stroke-width:0.529167;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1"
+       id="path3755-74-0-9-8-4-2-1-4-6-3-9-1-2"
+       inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
+       inkscape:export-xdpi="500"
+       inkscape:export-ydpi="500"
+       cx="112.15873"
+       cy="263.95883"
        rx="1.3269223"
        ry="1.3336277" />
     <ellipse
@@ -330,27 +375,27 @@
        inkscape:export-xdpi="500"
        inkscape:export-ydpi="500"
        cx="132.64217"
-       cy="286.31705"
+       cy="295.31302"
+       rx="1.3269223"
+       ry="1.3336277" />
+    <ellipse
+       style="fill:#000000;fill-opacity:1;fill-rule:evenodd;stroke:#000000;stroke-width:0.529167;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1"
+       id="path3755-74-0-9-8-4-2-1-4-6-3-9-0-8-6"
+       inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
+       inkscape:export-xdpi="500"
+       inkscape:export-ydpi="500"
+       cx="156.42009"
+       cy="295.87799"
        rx="1.3269223"
        ry="1.3336277" />
     <path
-       style="fill:none;stroke:#000000;stroke-width:0.529167;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:1.05833, 1.05833;stroke-dashoffset:0;stroke-opacity:1"
-       d="m 152.49221,254.80804 v 52.4108"
+       style="fill:none;stroke:#000000;stroke-width:0.529165;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:1.05833, 1.05833;stroke-dashoffset:0;stroke-opacity:1"
+       d="m 156.49038,295.29794 v 12.36813"
        id="path2985-9-8-0-4-3-6-8-9-6-9-7"
        inkscape:connector-curvature="0"
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
        inkscape:export-xdpi="500"
        inkscape:export-ydpi="500" />
-    <ellipse
-       style="fill:#000000;fill-opacity:1;fill-rule:evenodd;stroke:#000000;stroke-width:0.529167;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1"
-       id="path3755-74-0-9-8-4-2-1-4-6-3-9-0-3-4"
-       inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
-       inkscape:export-xdpi="500"
-       inkscape:export-ydpi="500"
-       cx="152.40837"
-       cy="254.56749"
-       rx="1.3269223"
-       ry="1.3336277" />
     <path
        style="fill:none;stroke:#000000;stroke-width:0.529167;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1;marker-start:url(#RoundedArrow-1)"
        d="M 163.83396,307.74802 H 57.768311"
@@ -361,17 +406,12 @@
        inkscape:export-ydpi="500" />
     <path
        style="fill:none;stroke:#000000;stroke-width:0.529169;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1;marker-start:url(#RoundedArrow-1)"
-       d="m 57.764422,252.09953 v 55.45975"
+       d="m 57.764422,233.75832 v 73.80097"
        id="path2985-7-3-7-9"
        inkscape:connector-curvature="0"
        inkscape:export-filename="C:\Users\chrisneal\Google Drive\School\Mechanical Engineering - Graduate\Advanced CFD - Bala\Homework\Homework 5\Diagrams\Finite Volume Diagram.png"
        inkscape:export-xdpi="500"
        inkscape:export-ydpi="500" />
-    <path
-       style="fill:none;stroke:#000000;stroke-width:0.264583px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1"
-       d="m 57.768147,307.74826 c 0,0 21.774191,-31.92446 38.012767,-33.90079 11.145286,-1.35644 19.202396,19.25525 30.115086,16.4272 16.00401,-4.14748 29.63228,-43.21619 29.63228,-43.21619"
-       id="path2"
-       sodipodi:nodetypes="cssc" />
     <text
        id="text4421-6-2-2-1-1-5-0-0-9"
        y="309.29691"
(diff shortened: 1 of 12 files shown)


wandadars:
This `getNewGrid` method doesn't seem to be used anywhere in the refiner or elsewhere.

Also, this failing test is a bit perplexing to me as it doesn't seem that I actually changed anything.

wandadars:
Does the refiner warrant it's own documentation markdown page @speth? I made some diagrams to help me understand the methods of the refiner.

refinement_diagram

wandadars:
One strange thing about the way that certain points get marked as being in the "keep"  list is that the curvature check will only keep the point next to the point of interest if it decides to split the two intervals to the right of the point of interest. This is different from how the slope criterion check marks points. I haven't quite grasped any underlying logic for which points get added to the "keep" list during each refinement check.

wandadars:
@speth  I went ahead of created a new page that talks about the grid refinement.

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,132 @@ class Domain1D;
...
+
     //! Names of components that require the addition of new grid points
-    set<string> m_c;
+    set<string> m_component_name;
+
+    //! Vector of flags indicating whether each component should be considered
+    //! for grid refinement
     vector<bool> m_active;
+
+    //! Refinement criteria
```
ischoegl: This docstring only applies to the first of the subsequent member variables.

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,132 @@ class Domain1D;
...
     double m_ratio = 10.0;
     double m_slope = 0.8;
     double m_curve = 0.8;
     double m_prune = -0.001;
+
+    //! Threshold parameter that is used distinguish between small fluctuations
```
ischoegl: If possible, use 'brief' docstrings that are succinct statements.

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,132 @@ class Domain1D;
...
+    //! around a constant value during the refinement process.
     double m_min_range = 0.01;
-    Domain1D* m_domain;
-    size_t m_nv;
-    size_t m_npmax = 1000;
+
+    Domain1D* m_domain; //! Pointer to the domain to be refined
```
ischoegl:
```suggestion
    Domain1D* m_domain;  //!< Pointer to the domain to be refined.
```

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,132 @@ class Domain1D;
...
     //! Indices of grid points that need new grid points added after them
-    set<size_t> m_loc;
+    set<size_t> m_insertion_points;
+
+    //! Map of grid point indices that should be kept, 1=keep, -1=remove, 0=unset
     map<size_t, int> m_keep;
+
     //! Names of components that require the addition of new grid points
-    set<string> m_c;
+    set<string> m_component_name;
```
ischoegl:
```suggestion
    set<string> m_componentNames;
```
See [C++ Style Guide](https://cantera.org/dev/develop/style-guidelines.html#c). I believe that for variable names that do not appear as part of the `public` interface we are not as concerned about descriptive names, which have to be balanced with readability of the implementation itself. `m_c` can certainly approved upon, but `m_cNames` would be a shorter compromise.

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,132 @@ class Domain1D;
...
     double prune() {
         return m_prune;
     }
 
 protected:
     //! Indices of grid points that need new grid points added after them
-    set<size_t> m_loc;
+    set<size_t> m_insertion_points;
+
+    //! Map of grid point indices that should be kept, 1=keep, -1=remove, 0=unset
```
ischoegl:
```suggestion
    //! Map of grid point indices that should be kept. Values are ...
```
Detailed description should not be part of the first statement, which forms Doxygen's `@brief` description. Also, which integer is what? 

As an aside, using an `enum` as the value instead of these sentinel values would further clarify things.

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,132 @@ class Domain1D;
...
+     *         output array size is insufficient to hold the new grid.
+     */
     int getNewGrid(int n, const double* z, int nn, double* znew);
-    int nNewPoints() {
-        return static_cast<int>(m_loc.size());
+
+    /**
+     * Returns the number of new grid points that were needed.
+     */
+    int numNewPoints() {
```
ischoegl:
```suggestion
    int nNewPoints() {
```
Avoid unnecessary churn - we widely use the shorthand `n` to indicate number, so please retain the original version. Also, as this is part of the `public` interface, you would have to go through a deprecation cycle to avoid breakage of external user code.

src/oneD/refine.cpp
```diff
@@ -77,39 +80,38 @@ int Refiner::analyze(size_t n, const double* z, const double* x)
...
-            // on a constant background.
-            if ((vmax - vmin) > m_min_range*aa) {
-                // maximum allowable difference in value between adjacent
-                // points.
-                double dmax = m_slope*(vmax - vmin) + m_thresh;
+            // find the range of values and slopes of component i over the domain
+            double val_min = *min_element(val.begin(), val.end());
+            double val_max = *max_element(val.begin(), val.end());
+            double slope_min = *min_element(slope.begin(), slope.end());
+            double slope_max = *max_element(slope.begin(), slope.end());
```
ischoegl:
```suggestion
            double valMin = *min_element(val.begin(), val.end());
            double valMax = *max_element(val.begin(), val.end());
            double slopeMin = *min_element(slope.begin(), slope.end());
            double slopeMax = *max_element(slope.begin(), slope.end());
```
Per Cantera's C++ style guide. Same comment applies to other variable names below.

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,132 @@ class Domain1D;
...
+
+    //! Returns the threshold for removing unnecessary grid points
     double prune() {
         return m_prune;
     }
 
 protected:
     //! Indices of grid points that need new grid points added after them
-    set<size_t> m_loc;
+    set<size_t> m_insertion_points;
```
ischoegl:
```suggestion
    set<size_t> m_insertionPoints;
```
I am torn about changing member variable names here: `m_loc` is used in many other contexts within the Cantera code base where the meaning is well known. At the same time, it is a more meaningful name.

While we aren't completely consistent on this, the default naming style would be to use camelCase after the initial `m_`, which in this case would yield `m_insertionPoints` - see [C++ Style Guidelines](https://cantera.org/dev/develop/style-guidelines.html#c)

ischoegl (commented):
@wandadars ... thank you for taking this on! While this is still work-in-progress, I wanted to provide some feedback that may be a little nit-picky, but should help with improving consistency of our documentation as well as our code base.

_Edit: I am not sure why unit tests broke, but it does appear that you inadvertently changed criteria for regridding, which can hopefully be rolled back._

ischoegl:
> This `getNewGrid` method doesn't seem to be used anywhere in the refiner or elsewhere.

If it's unused, please go ahead and deprecate it. For this, we usually issue a `warn_deprecated` warning, and mark the method with Doxygen's `@deprecated` tag. Search for these and you will find instances which can serve as a template.

> Also, this failing test is a bit perplexing to me as it doesn't seem that I actually changed anything.

I disagree - some of the failing tests show 
```none
E   *******************************************************************************
E   CanteraError thrown by Refiner::analyze:
E   max number of grid points reached (1000).
E   -------------------------------------------------------------------------------
```
which indicates that regridding was inadvertently changed. Same goes for `TestImpingingJet.test_restore_legacy_hdf` et al., where differences in array length are explained by regridding that should not take place as far as I remember.

include/cantera/oneD/refine.h
wandadars: Ah, my vs-code editor was leading me astray there, it applied it to all the subsequent items.

include/cantera/oneD/refine.h
wandadars: We want two spaces between the semicolon and the start of the comment?

include/cantera/oneD/refine.h
wandadars: How about m_insertPts ?

wandadars:
Thank you for the feedback @ischoegl . I'm still looking into how exactly I changed the algorithm.

wandadars:
The source of the change that was causing the tests to fail was the change of:

`double ratio = fabs(slope[j+1] - slope[j]) / (max_change + m_thresh/dz[j]);`
to 
 `double ratio = fabs(slope[j+1] - slope[j]) / max_change;`

Where I had just moved the m_thresh up into the max_change variable, as the purpose of that variable seems to be a protection against divide-by-zero FPEs.

`double max_change = m_curve*(slopeMax - slopeMin) + m_thresh;`

The thing that I'm not sure about is why this mattered. There's the division of `m_thresh` by `dz[j]`, which doesn't make much sense to me given that we're just protecting the denominator from going to zero. 

`m_thresh` is roughly 1.49011745e-8.  For cases where the grid spacing is close, then the division by the dz term would make the effective FPE threshold larger.


The failing test is shown below. The width of the domain is 0.05m, and if there's say 100 points, then the dz is on the order of 5e-4m, which would make the effective FPE threshold something like  2.98e-5.  Doing it the way I had adjusted it to, the threshold would still be 1.4e-8 for guarding against the FPEs.

```
def test_unity_lewis(self):
        self.create_sim(ct.one_atm, 300, 'H2:1.1, O2:1, AR:5.3')
        self.sim.transport_model = 'unity-Lewis-number'
        self.sim.set_refine_criteria(ratio=3.0, slope=0.08, curve=0.12)
        self.sim.solve(loglevel=0, auto=True)
        dh_unity_lewis = np.ptp(self.sim.enthalpy_mass)

        self.sim.transport_model = 'mixture-averaged'
        self.sim.solve(loglevel=0)
        dh_mix = np.ptp(self.sim.enthalpy_mass)

        # deviation of enthalpy should be much lower for unity Le model (tends
        # towards zero as grid is refined)
        self.assertLess(dh_unity_lewis, 0.1 * dh_mix)
```

wandadars:
I put some print statements into that part of the code to compare the two, and it does seem like my change makes the ratio larger than the other way. This makes sense because for a smaller value of the max_change, the m_thresh by itself would be smaller. I don't really see the logic of making the FPE guard larger though.

Edit: I'm inclined to think that the old way of handling the ratio criterion was permitting un-resolved points to pass erroneously. 

```
Old demoninator: 0.00191138
New demoninator: 4.04799e-06
Old Ratio: 0.000436741
New Ratio: 0.20622

Old demoninator: 0.00191138
New demoninator: 4.04799e-06
Old Ratio: 0.000578154
New Ratio: 0.272993

Old demoninator: 0.00191138
New demoninator: 4.04799e-06
Old Ratio: 0.000713674
New Ratio: 0.336983

Old demoninator: 0.00191138
New demoninator: 4.04799e-06
Old Ratio: 0.000835034
New Ratio: 0.394286

Old demoninator: 0.00191138
New demoninator: 4.04799e-06
Old Ratio: 0.000930045
New Ratio: 0.439149

Old demoninator: 0.00191138
New demoninator: 4.04799e-06
Old Ratio: 0.000978374
New Ratio: 0.461968

Old demoninator: 0.00191138
New demoninator: 4.04799e-06
Old Ratio: 0.00211619
New Ratio: 0.999222

Old demoninator: 0.000957707
New demoninator: 4.04799e-06
Old Ratio: 0.00484523
New Ratio: 1.14632

Old demoninator: 0.000957707
New demoninator: 4.04799e-06
Old Ratio: 0.00504515
New Ratio: 1.19362
```

wandadars:
I looked at that failing test and it does look like for the case when the denominator is simply using the m_thresh, a ton of points get put in a region of almost constant slope. So maybe the guard against regions of constant slope isn't working properly causing the algorithm to be activated?

I printed out all the cases where the ratio was greater than 1 using the newer ratio definition and it looks like there's a sensitivity going on where maybe the more points that are added causes noise that can make the slope vary significantly, which drives more points to be added.

[run.log](https://github.com/user-attachments/files/17053614/run.log)

include/cantera/oneD/refine.h
ischoegl: I don’t think we have a guideline on this. On the Python side, it’s a recommendation that I have seen, and I think it’s slightly more readable. It’s really your choice; the `<` was definitely missing.

include/cantera/oneD/refine.h
ischoegl: If this is the case, there may be a setting I’m unaware of. The doxygen output will tell.

wandadars:
I think one issue was that the refiner was considering (for the failing test) the species Argon, which was a constant, but there were small fluctuations in the value, and because the slope criteria only tested for regions of constant slope, it entered into the slope refinement section and inserted points, which seems to have generated more noise in the slope, which generated more points.  If I have understood the nature of the issue, basically the zero slope case seems to have had positive/negative values or just noise that the first if-statement catch seems to have deemed as being significant enough  of a variation from a constant to let it enter the clause. Maybe a case where because the value under consideration was zero, this is different from a case of a non-zero constant slope.

I put an extra clause on the slope constant-value-check clause where it just uses the previous constant-value-clause for the previous if-statement that refines based on values. Doing this removed the excessive refinement ( it was putting 3500 points in the free flame case that was failing). It returned back to putting around 100 points over the flame after adding this additional check.

Looking at Argon, we have the following outputs for the free-flame test case.

```
Component: AR
valMax-valMin: 2.80705e-11
min_range*valMagnitude: 0.00860884
slopeMax-slopeMin: 2.79602e-08
min_range*slopeMagnitude: 2.71097e-10
Slope[j]: 2.25352e-08
Slope[j+1]: 9.84457e-10
SlopeMin and SlopeMax: -8.5052e-10 2.71097e-08
Old demoninator: 4.76871e-05
New demoninator: 1.82564e-08
Old Ratio: 0.00045192
New Ratio: 1.18045
```

Looking at the valMax-ValMin relative to the min_range*valMagnitude, it is clear this is a constant value. But looking at the slope quantity tells a different story because the magnitude of the slope is close to zero, so the noise is considered to be significant. We can also see the old ratio and new ratio have very different opinions on the ratio. The old ratio had that m_thresh term, which added just enough to the denominator so as to ignore slopes that were very close it seems. I don't think that was the intention of the original coders, as they also used the m_thresh in the previous check for the value refinement. I may be incorrect though. It seems like the expected behavior is to not let the "bad" points even get into the refinement insertion algorithm instead of letting them in and then within the algorithm ignoring ones that fail to meet that `m_thresh/dz[j]` quantity.

wandadars:
It appears I have done something wrong here. I thought I was rebasing my branch onto main, but this page now shows that 65 files were changed.

wandadars:
It looks like maybe that m_thresh variable may have been meant as a filter for the minimum allowable difference between two small values, and therefore for two grid points that differ by the smallest allowable amount, the smallest allowable slope would be m_thresh / dz. I'll revert my change and work to make it  clear that this is likely what job the m_thresh quantity is performing.

wandadars:
I think that forward difference first derivative approximation was susceptible to noise in the solution. I tested out a simple central difference approximation (with forward/backward difference approximations for the left/right boundary points), and the excessive refinement goes away for that free flame case. The original algorithm adds 120 points I think and the one with the central difference slope approximation adds 128, so they are fairly close.

I'm not going to change anything though, but in case it is looked into again in the future, this is what I was using.

```
// slope of component i (using central difference to reduce noise)
for (size_t j = 0; j < n-1; j++) {
    // Central difference for interior points
    if (j > 0 && j < n-1) {
        slope[j] = (val[j+1] - val[j-1]) / (dz[j-1] + dz[j]);
    } else if (j == 0) {
        // forward difference for first point
        slope[j] = (val[j+1] - val[j]) / dz[j];
    }
}
```

include/cantera/oneD/refine.h
speth: I think the second space is more helpful in the Python case, where you don't already have a semicolon ending the actual code statement. The one-space version is about 15x more common in the C++ code base at present.

include/cantera/oneD/Sim1D.h
```diff
@@ -203,6 +203,24 @@ class Sim1D : public OneDim
...
+     * Performs the hybrid Newton steady/time-stepping solution.
+     *
+     * The solver attempts to solve the steady-state problem first. If the steady-state
+     * solver fails, the time-stepping solver is used to take multiple time steps to
+     * move the solution closer to the steady-state solution. The steady-state solver is
+     * called again after the timesteps to make further progress towards the steady-state
+     * solution. This process is repeated until the steady-state solver converges or the
+     * maximum number of timesteps is reached.
+     *
+     * At the end of a successful solve, if the refine_grid flag is set, the grid will be
```
speth:
```suggestion
     * At the end of a successful solve, if the `refine_grid` flag is set, the grid will be
```

include/cantera/oneD/refine.h
```diff
@@ -21,12 +21,60 @@ class Domain1D;
...
+     * The ratio parameter is the maximum allowed ratio between grid spacing at
+     * adjacent intervals. The ratio parameter considers the situation where the
+     * left interval is much larger than the right interval, or if the right
+     * interval is much larger than the left interval.
+     *
+     * The slope parameter is the maximum fractional change in the value of each
+     * solution component between adjacent grid points. This quantity is used to
+     * check the criteria given below:
+     *    val[j+1] - val[j]  < slope*(val_max - val_min)
+     * Where val_max and val_min are the maximum and minimum values of the
```
speth: Careful -- this doesn't render as I think you expect. I'd recommend leaving the formulas and extended description here to the Sphinx documentation and keeping this section fairly brief. A one sentence description for each parameter, along with a link back to the new Sphinx page should be sufficient (use a relative Markdown style link, for example `[grid refinement](../reference/onedim/grid-refinement.html)`).

include/cantera/oneD/refine.h
```diff
@@ -21,12 +21,60 @@ class Domain1D;
...
+     * removal. Too much refinement can lead to unnecessary grid points, which will
+     * slow down the solution process.
+     *
+     * A rule of thumb with the prune parameter is that, if you have specified your
+     * slope and curve parameters to a desireable level, then the ratio above would
+     * be 1.0 for the case when the grid points just meet the refinement criteria that
+     * would be "good enough" for the solution at hand. A ratio less than than one
+     * still meets the criteria, but introduces more grid points. The prune parameter
+     * allows you set the point at which the refiner decides that is has overly-refined
+     * the grid and can remove points that are not needed.
```
speth: The other factor that I'd point out is that the `prune` parameter is generally not important when solving a single case. It's main benefit is when solving a problem multiple times with different boundary conditions that cause the regions of the domain where a high density of points is required to move.

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,142 @@ class Domain1D;
         return m_gridmin;
     }
 
-    //! Determine locations in the grid that need additional grid points.
-    //!
-    //! @returns The number of new grid points needed (size of #m_loc)
+    /**
+     * Determine locations in the grid that need additional grid points and
+     * updates the internal state of the Refiner with this information.
```
speth:
```suggestion
     * Determine locations in the grid that need additional grid points and
     * update the internal state of the Refiner with this information.
```

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,142 @@ class Domain1D;
...
+     * @param[in] z  Point to array of grid points
+     * @param[in] x  Pointer to solution vector
+     *
+     * @returns The number of new grid points needed (size of #m_insertPts)
+     */
     int analyze(size_t n, const double* z, const double* x);
+
+    /**
+     * Constructs a new grid based on refinement locations determined by the analyze
+     * method.
```
speth:
```suggestion
     * Constructs a new grid based on refinement locations determined by the analyze()
     * method.
```

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,142 @@ class Domain1D;
...
+     *
+     * This function generates a new grid by inserting additional points into the
+     * current grid at locations where the `analyze` method has identified a need for
```
speth: This will generate a Doxygen link to the member function in the same class:
```suggestion
     * current grid at locations where the analyze() method has identified a need for
```

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,142 @@ class Domain1D;
...
+     * @param[in] z Pointer to the array of original grid points.
+     * @param[in] nn The maximum number of points that the new grid array `znew` can hold.
+     * @param[out] znew Pointer to the array where the new grid points will be stored.
+     *
+     * @return The function returns 0 upon successful creation of the new grid. Throws
+     *         an exception if the provided output array size is insufficient to hold
+     *         the new grid.
+     *
+     * @deprecated Starting in %Cantera 3.1, this function is unused. To be removed
+     *             after %Cantera 3.1.
```
speth: This makes it sounds like the function was used up until Cantera 3.1, but digging through the Git history, this function has apparently been sitting here unused for 20 years.
```suggestion
     * @deprecated Unused. To be removed after %Cantera 3.1.
```
I would suggest leaving out the rest of the docstring changes here except the deprecation notice -- we certainly don't want to encourage someone to start using this method now.

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,142 @@ class Domain1D;
...
+     */
     int getNewGrid(int n, const double* z, int nn, double* znew);
+
+    /**
+     * Returns the number of new grid points that were needed.
+     */
```
speth: For a function with a single-line docstring, you can save some vertical real estate by just writing
```suggestion

    //! Returns the number of new grid points that were needed.
```

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,142 @@ class Domain1D;
...
-        return (m_keep[j] != -1);
+        return (m_keep[j] != REMOVE);
     }
-    double value(const double* x, size_t i, size_t j);
 
+    /**
+     * Returns the value of the solution component, n, at grid point j.
+     *
+     * @param x Solution vector
+     * @param n Solution Component index
```
speth: To avoid misleading links to the `Solution` class:
```suggestion
     * @param x %Solution vector
     * @param n %Solution component index
```

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,142 @@ class Domain1D;
...
+     * @param j Grid point index
+     */
+    double value(const double* x, size_t n, size_t j);
+
+    //! Returns the maximum ratio of grid spacing at adjacent intervals
```
speth:
```suggestion
    //! Returns the maximum allowable ratio of grid spacing between adjacent intervals
```

include/cantera/oneD/refine.h
speth: I'm not sure what it means to say that VS Code "applied it to all the subsequent items", but @ischoegl is correct. If you want to define a group, you can use the syntax:
```cxx
    //! @name Refinement criteria
    //! @{
    double m_ratio = 10.0; //!< grid spacing refinement criteria
    double m_slope = 0.8; //!< function change refinement criteria
    double m_curve = 0.8; //!< function slope refinement criteria
    double m_prune = -0.001; //!< pruning refinement criteria
    //! @}
```
This is mainly useful if you want to provide a somewhat extended description in the opening comment block (after the `@name`, before the `@{`).

include/cantera/oneD/refine.h
```diff
@@ -69,51 +123,142 @@ class Domain1D;
...
     double m_min_range = 0.01;
-    Domain1D* m_domain;
-    size_t m_nv;
-    size_t m_npmax = 1000;
+
+    Domain1D* m_domain; //!< Pointer to the domain to be refined.
+    size_t m_nv; //!< Number of components in the domain
+    size_t m_npmax = 1000; //!< Maximum number of grid points
+
+    //! Closest that a solution component at two adjacent grid points can be.
```
speth: I think you're misinterpreting the use of `m_thresh`. Its purpose is to provide an absolute tolerance for the slope and curve evaluations such that solution components whose values don't change very much across the domain aren't targeted for refinement if there happens to be a small blip at some point in the domain. It's very much analogous to the absolute tolerances used in the test suite comparisons.

src/oneD/refine.cpp
```diff
@@ -77,153 +80,154 @@ int Refiner::analyze(size_t n, const double* z, const double* x)
...
+                // on the global min and max values of the component over the domain.
+                double max_change = m_slope*(valMax - valMin);
                 for (size_t j = 0; j < n-1; j++) {
-                    double r = fabs(v[j+1] - v[j])/dmax;
-                    if (r > 1.0 && dz[j] >= 2 * m_gridmin) {
-                        m_loc.insert(j);
-                        m_c.insert(name);
+                    // m_thresh prevents two adjacent points from being considered
+                    // if they are too close together.
+                    double ratio = fabs(val[j+1] - val[j]) / (max_change + m_thresh);
```
speth: See previous comment about `m_thresh`.

src/oneD/refine.cpp
```diff
@@ -235,12 +239,16 @@ void Refiner::show()
 
 int Refiner::getNewGrid(int n, const double* z, int nn, double* zn)
 {
-    int nnew = static_cast<int>(m_loc.size());
+    warn_deprecated(
+    "Refiner::getNewGrid",
+    "Deprecated in Cantera 3.1; unsed function that will be removed.");
```
speth:
```suggestion
    warn_deprecated(
        "Refiner::getNewGrid",
        "Deprecated in Cantera 3.1; unused function that will be removed.");
```

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,229 @@
...
+entire domain. If the ratio of the curvature between two adjacent grid points is greater
+than the user-specified value, then an additional grid point is inserted between the two
+points. This criterion is used to ensure that the solution is well-resolved in regions
+where the solution has steep gradients.
+
+The specific criteria that is being enforced is:
+
+$$
+  |\left( \frac{d x_n}{d z} \right)_{j+1} - \left(\frac{d x_n}{d z} \right)_j| <= \gamma_{max} (\displaystyle\max_j \frac{dx_n}{dz} - \displaystyle\min_j \frac{dx_n}{dz})
+$$
```
speth:
```suggestion
$$
  \left|\left( \frac{d x_n}{d z} \right)_{j+1} - \left(\frac{d x_n}{d z} \right)_j\right| \le \gamma_{max} (\displaystyle\max_j \frac{dx_n}{dz} - \displaystyle\min_j \frac{dx_n}{dz})
$$
```
Similar style suggestions apply in a few other locations.

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,229 @@
...
+
+This refinement option controls the maximum change in a solution component between
+two adjacent grid points compared to the maximum change in that component across the
+entire domain. If the ratio of the change in a solution component between two adjacent
+grid points is greater than the user-specified value, then an additional grid point is
+inserted between the two points. This criterion is used to ensure that the solution
+is well-resolved in regions where the solution has steep gradients.
+
+The specific criteria that is being enforced is:
+$$
```
speth: You need a blank line before the opening `$$` for this to be rendered in display style. Please run Sphinx locally or download the output from the CI run and look at the HTML pages. There's really no other way to find all the formatting issues here.

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,229 @@
...
+
+```{figure} /_static/images/onedim/refinement_right_interval_stencil.svg
+:width: 75%
+:alt: The points that are marked as needing new grid points inserted to the right of
+them and which ones should be marked to be kept.
+:align: center
+
+The points that are marked as needing new grid points inserted to the right of them
+and which ones should be marked to be kept.
+```
```
speth: I don't know what if any syntax is needed for the alt text to extend across multiple lines, but this doesn't work. Probably the easiest fix is that there really isn't a need for alt text if it's just a repeat of the caption.

speth (changes requested):
Thanks for tackling this, @wandadars. I haven't completely finished going through this, but here were some suggestions that should help.

doc/sphinx/_static/images/onedim/refinement_curve_metric_plot.svg
speth, line 1: As drawn, the minimum slope occurs at the point before the one marked.

I think this diagram would demonstrate the algorithm more clearly with a few changes:
- Draw the function such that the slope criterion ($\gamma_j$) had its highest value at point $j$
- Mark points $j$, $j+1$, and $j+2$, as those are the ones that go into the definition of $\gamma_j$ below
- Mark the locations of the two points that would be added.

doc/sphinx/_static/images/onedim/refinement_slope_metric_plot.svg
speth, line 1: It might help to have a function where the max and min values aren't just at the left and right boundaries, to make it clear that $x_{n,min}$ isn't just $x_0$. And similar to the suggestion for the "curve" diagram, I think you should label points $j$ and $j+1$ where $\beta_j$ is maximum, and the location where a point would be added between those points. If it makes things easier, $x(z)$ doesn't need to be the same for these two diagrams.

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,229 @@
...
+
+At each point in the domain, a ratio can be calculated for each component as:
+
+$$
+  \beta_j =  \frac{|{x_{n,j+1} - x_{n,j}} |}{\displaystyle\max_j x_n - \displaystyle\min_j x_n}
+$$
+
+Points where $ \beta_j $ is greater than $ \beta_{max} $ are marked to have an additional
+grid point inserted to the right of them.
```
speth: You end up essentially giving this same formula twice, once as an inequality, and once as a definition for $\beta_j$. I think the explanation of this criterion could be simplified by starting with this definition and then describing how the addition of points occurs whenever $\beta_j > \beta_{max}$.

Likewise for the definition of the curvature-based refinement.

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,229 @@
...
+
+The rule-of-thumb for setting $ \beta_{max} $ is that it prevents any two adjacent points
+from having a change in a solution component that is greater than $ \beta_{max} \cdot 100\% $ of
+the maximum change in that component across the domain.
```
speth: This is essentially a restatement of the first sentence of this section. For a "rule-of-thumb", I'd think you would want to give a reasonable range of values for $\beta_{max}$, although that's a little tricky to do in the general case.

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,229 @@
+# Grid Refinement for One-dimensional Flows
+
+## Overview
+
+The Cantera 1D solver uses a progressive mesh refinement strategy when solving the
+governing equations. An initial coarse grid is solved, and then the solution is
+analyzed to identify regions where the solution is not well resolved, such as areas
+with steep gradients. Additional grid points are inserted in these regions and then
+the solution is solved again. If too many points are inserted, the solver may have
```
speth:
```suggestion
The Cantera 1D solver uses a progressive mesh refinement strategy when solving the
governing equations. The system is initially solved on a coarse grid, and then the solution is
analyzed to identify regions where it is not well resolved, such as areas
with steep gradients. Additional grid points are inserted in these regions and then
the system is solved again. If too many points are inserted, the solver may have
```

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,229 @@
...
+difficulty converging, and so the algorithm prevents itself from adding too many points
+during one refinement step. These additional points are added during subsequent
+refinement steps. This process is repeated until the solution is deemed to be well
```
speth: At least to my understanding, the reason for not adding more points in a single pass isn't that it necessarily makes the subsequent problem more difficult to solve, but because the solution near the added points tends to change significantly, such that you can't easily tell how many additional points are really necessary in a particular region.

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,229 @@
...
+
+  - ratio
+  - slope
+  - curve
+  - prune
+
+### Ratio
+
+This refinement option controls the maximum ratio of the spacing between the grid
+points that are adjacent to a point under consideration.If the ratio between two
```
speth:
```suggestion
points that are adjacent to a point under consideration. If the ratio between two
```

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,229 @@
...
+adjacent grid points is greater than the ratio
+specified by the user ($ \alpha_{max} $), then an additional grid point is inserted
+between the two points. Depending on which interval adjacent to a point under
```
speth: All the variable names for the refinement criteria are purely local inventions, right? In that case, you may as well stick to simple mnemonics, $r$ for ratio, $s$ for slope, $c$ for curve and $p$ for prune, as none of these letters are used elsewhere in this explanation.

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,229 @@
...
+
+The specific criteria that is being enforced is:
+
+$$
+  |\left( \frac{d x_n}{d z} \right)_{j+1} - \left(\frac{d x_n}{d z} \right)_j| <= \gamma_{max} (\displaystyle\max_j \frac{dx_n}{dz} - \displaystyle\min_j \frac{dx_n}{dz})
+$$
+
+Where $ \gamma_{max} $ is a number between 0 and 1 that is specified by the user, $ \frac{dx_n}{dz} $
+is the derivative of the solution component being considered, and $ \displaystyle\max_j \frac{dx_n}{dz} $ and $ \displaystyle\min_j \frac{dx_n}{dz} $ are the
+maximum and minimum values of the solution component derivative across the domain. Visually, these
```
speth: Using the notation $x'_n$ in these formulas, as you do in the figure, might make this all easier to read, especially for the inline formulas.

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,229 @@
...
+Points where $ \gamma_j $ is greater than $ \gamma_{max} $ are marked to have an additional
+grid point inserted to the right of them.
+
+The rule-of-thumb for setting $ \gamma_{max} $ is that it prevents any two adjacent points
+from having a change in the slope of a solution component that is greater than $ \gamma_{max} \cdot 100\% $ of
+the maximum change in that component's slope across the domain.
+
+Note: The slope is computed using a backward difference scheme and so the estimate of the
+derivative at a grid point depends on the solution at the grid point and the solution at the
+grid point to the right.
```
speth: I tend to think of this less as being a backward difference and more as an approximation for the derivative at the midpoint between $j$ and $j+1$. Then you can see that the decision of where to add new points is symmetric -- the slopes defined by points $j$, $j+1$, and $j+2$ are used to decide whether to add points to the left and right of point $j+1$.

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,229 @@
...
+adding grid points, this option removes grid points that are not needed. This option
+lets a user specify a value of the ratio quantities described above that if the ratio
+drops below this, then the point is removed. The user specifies a pruning value, $ \delta $,
+and that serves as a lower-bound on the ratio quantities.
+
+$$
+  \delta = \beta_{min} = \gamma_{min}
+$$
+
+If $\beta_j < \delta$, or $\gamma_j < \delta$, then the point is removed.
```
speth: I don't think this is correct -- the `prune` value is relative to either the `curve` or `slope` parameters, so the condition under which a point is allowed to be removed is like $\beta_j / \beta_{max} < \delta$. Also, the condition must be satisfied for _both_ `curve` and `slope`, otherwise successive refinement would potentially just re-add the point.

speth (changes requested):
Some additional comments for your consideration.

include/cantera/oneD/refine.h
wandadars: I have reduced those sections down like you suggested. Those sections where holdovers from before I decided to make the Sphinx page iirc.

include/cantera/oneD/refine.h
wandadars: Good point. I added an extra sentence describing this case.

include/cantera/oneD/refine.h
wandadars: That's my understanding I believe. Components that are closer than the m_thresh hit the m_thresh floor with the way it is used in the implementation. The slope criteria uses it in a derived way to limit the slope based on the solution component values versus as a floor of the slope. ( the use of the m_thresh/dz in the implementation instead of just using m_thresh.

Maybe a longer comment of:

//! Absolute tolerance for solution components in the domain to prevent the refiner from considering numerical noise

doc/sphinx/reference/onedim/grid-refinement.md
wandadars: Maybe I should say "is marked for removal" instead of "is removed"?

doc/sphinx/reference/onedim/grid-refinement.md
wandadars: Is there a clear reason why this staggering is done in this case? The solution components are compared at the grid points, but the slopes are compared between the grid points?

doc/sphinx/reference/onedim/grid-refinement.md
wandadars: This statement was derived from the explanation given in the Kee book iirc. They mention the need for not adding too many new points. I may have misunderstood that paragraph in the book.

Edit: I took out those sentences.

doc/sphinx/reference/onedim/grid-refinement.md
wandadars: I agree. I have combined them together.

doc/sphinx/_static/images/onedim/refinement_curve_metric_plot.svg
wandadars: For the case of the slope and the curve diagrams, I was more leaning towards just simply showing the criterion and not an example of it being applied (like was done for the ratio one).

doc/sphinx/reference/onedim/grid-refinement.md
wandadars: I removed these sections because they were redundant.

doc/sphinx/reference/onedim/grid-refinement.md
wandadars: Agreed.

include/cantera/oneD/refine.h
wandadars: What is the relative reference with respect to in your example? I tried to use that but in my local version the link doesn't resolve to the Sphinx page and gives an error.

src/oneD/refine.cpp
wandadars: Am I correct in my interpretation that this quantity is not a generic absolute tolerance. It's specifically the absolute tolerance for the solution component "closeness". In the m_curve section in refine.cpp we see that this tolerance is not used to prevent slope values below the tolerance, but rather an effective tolerance, relative to the solution components is computed.

doc/sphinx/reference/onedim/grid-refinement.md
wandadars: I reworked that section to align with your suggestion.

include/cantera/oneD/refine.h
wandadars: I think I got it to work. Changed the reference to:

`[prune](../../html/reference/onedim/grid-refinement.html#prune)`

And it looks to link properly on my local documentation build.

doc/sphinx/reference/onedim/grid-refinement.md
wandadars: Sounds good. Switched to those letters in the writeup.

wandadars:
All the comments should be covered in the new commits. Thanks for reviewing this @ischoegl  and @speth .

Edit: I'm not sure what is causing the tests to fail here. When I try to look at them it just says "This job failed". Maybe the way I linked to the Sphinx documentation page from the Doxgyen header comment is causing an issue?

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,228 @@
...
+$$
+
+All of the values are below the original $ S_{max} $ of 0.5, but some are much
+lower than that threshold. If the pruning parameter is set to a value of 0.2, then
+the effective cutoff threshold for removing a point would be 0.5*0.2 = 0.1. With this
+pruning criteria, the pruned data would be:
+
+$$
+  S_j = [0.11, 0.16, 0.1, 0.22, 0.37, 0.29]
+$$
```
speth: This example is interesting, but I think the end result you get isn't quite right -- the sum of $S_j$ should be a constant, since the change in the solution ends up getting absorbed into the neighboring points. This might be more clear if you define a corresponding array with values of $x_j$ before and after pruning.

doc/sphinx/reference/onedim/grid-refinement.md
```diff
@@ -0,0 +1,228 @@
...
+sacrificing the accuracy of the solution. The following criteria are used to decide
+if a point is marked for removal:
+
+$$
+  \frac{S_j}{S_{\text{max}}} < P \quad \text{or}
+  \quad \frac{C_j}{C_{\text{max}}} < P
+$$
+
+If $ \frac{S_j}{S_{max}} $, or $ \frac{C_j}{C_{max}} $ are less
+than $ P $, then the point is marked for removal.
```
speth: The criterion is implemented such that if either ratio is greater than or equal to $P$, the point is kept. The converse of this is that only if _both_ ratios are less than $P$ is the point marked for removal.
```suggestion
$$
  \frac{S_j}{S_{\text{max}}} < P \quad \text{and}
  \quad \frac{C_j}{C_{\text{max}}} < P
$$

If $ \frac{S_j}{S_{max}} $ and $ \frac{C_j}{C_{max}} $ are less
than $ P $, then the point is marked for removal.
```

src/oneD/refine.cpp
```diff
@@ -77,153 +80,152 @@ int Refiner::analyze(size_t n, const double* z, const double* x)
...
                 for (size_t j = 0; j < n-2; j++) {
-                    double r = fabs(s[j+1] - s[j]) / (dmax + m_thresh/dz[j]);
-                    if (r > 1.0 && dz[j] >= 2 * m_gridmin &&
-                            dz[j+1] >= 2 * m_gridmin) {
-                        m_c.insert(name);
-                        m_loc.insert(j);
-                        m_loc.insert(j+1);
+                    // Using the solution component absolute tolerance (m_thresh),
+                    // a smallest allowable slope can be estimated for an interval dz
+                    // as m_thresh/dz.
```
speth: This still reads strangely to me. There is no "smallest allowable slope" -- it is perfectly fine for a component to have zero slope. The point of adding this term is to avoid adding a point when the absolute change in the slope is very small, even if the change in slope is large in a relative sense, which can occur if `max_change` is small.

speth (changes requested):
Thanks for the updates, @wandadars. The new diagrams are looking nice. I have just a couple of remaining suggestions.

*note: I force pushed to this branch to fix a couple of minor typos, so you should fetch before making any additional changes*

doc/sphinx/reference/onedim/grid-refinement.md
wandadars: Good catch!

src/oneD/refine.cpp
wandadars: I agree that it's not clear.  Is it that the outer check `slopeMax - slopeMin > m_min_range*slopeMagnitude` is only checking for noise around a constant non-zero-ish value. And then inside the if-statement the `m_thresh/dz` is catching that other case where the max change over the entire domain is very close to zero? It's the description of why we need to perform a `m_thresh/dz` that's tripping me up. An absolute tolerance that is dependent on grid spacing.

doc/sphinx/reference/onedim/grid-refinement.md
wandadars: I'm thinking of something simpler: just marking the points that are below the pruning criteria in red.

src/oneD/refine.cpp
speth: I think what you have now works. You may be right that the addition of this absolute term is not necessary, as if `max_change` is very small, then we shouldn't have gotten into this loop in the first place.

wandadars: It definitely is needed though. I tried to get rid of it earlier in the PR, and it was a disaster because it started putting thousands of points in a region that had zero-ish slope in the species equation iirc.

> I think one issue was that the refiner was considering (for the failing test) the species Argon, which was a constant, but there were small fluctuations in the value, and because the slope criteria only tested for regions of constant slope, it entered into the slope refinement section and inserted points, which seems to have generated more noise in the slope, which generated more points. If I have understood the nature of the issue, basically the zero slope case seems to have had positive/negative values or just noise that the first if-statement catch seems to have deemed as being significant enough of a variation from a constant to let it enter the clause. Maybe a case where because the value under consideration was zero, this is different from a case of a non-zero constant slope.

speth (approved):
Thanks, @wandadars. This looks good to me. I pushed to your branch just to squash a few changes together and clean up the commit history.

PR #1788, opened 2024-08-23
Non-empty Reactor and Connector Objects

### LINKED ISSUE #213: Deprecate creation of empty `Wall` or `FlowDevice` objects in C++
**Abstract**

Creation of empty `Wall`, `FlowDevice` and `ReactorSurface` objects is not supported by the Python API, and Cantera/cantera#1765 removes preliminary support from experimental MATLAB. It would be consistent to disallow this at the C++ level as well, which would (eventually) simplify/streamline the interface by removing several methods and associated exception handling. Passing shared pointers to `ReactorBase` objects to constructors instead would be a significant step towards elimination of raw pointers from the reactor code base. A similar argument holds for `ReactorNet` and the pending `ReactorEdge` (see Cantera/cantera#1697).

**Motivation**

* What problem is it trying to solve? ... _simplify reactor API_
* Who is affected by the change? ... _mostly developers_
* Why is this a good solution? ... _simplifies API_

**Possible Solutions**

Update constructors and factory methods and `clib` beyond changes introduced in Cantera/cantera#1765. Deprecate various `install` methods. Constructors should be updated to use signatures similar to

```C++
FlowDevice(shared_ptr<ReactorBase> upstream=nullptr, shared_ptr<ReactorBase> downstream=nullptr, 
           const string& name="(none)")
```

where null `upstream`/`downstream` will raise deprecation warnings in 3.1, with defaults being removed thereafter.

**References**

- Cantera/cantera#1864
- Cantera/cantera#1848
- Cantera/cantera#1788
- Cantera/cantera#1765
- Cantera/cantera#1697

ischoegl:

This PR ensures that the Reactor/zeroD API uses shared pointers consistently. After removing raw pointers/references from various methods, it becomes possible to update the underlying code base, which at this point is a mix of shared pointers, raw pointers, and references. Details are:

- Create two new bases `ReactorNode` and `Connector`, which are bases for all 0D objects (which from the perspective of a graph are either nodes or edges); all objects are captured in updated factories.
  - `ReactorBase` and `ReactorSurface` are now derived from `ReactorNode`.
  -  `Wall` and `FlowDevice` are both derived from `Connector`.
- ~All `ReactorNode` and `Connector` objects require contents to be available to the constructor (other than deprecated versions)~
- Cantera 3.0 still had default constructors for all 0D objects (without parameters). Additions thus far implemented for 3.1 are moved to protected constructors, ~with new convenience methods creating objects accessed by shared pointers (examples: `newReservoir` and `newValve`). The old constructors (without arguments) are deprecated. The implementation loosely follows `newSolution`.~ ... _Edit: constructors are accessed via `Reservoir::create`, `Valve::create` etc.._
- A new `SharedFactory` factory variant stores shared pointers rather than raw pointers.
- The approach is motivated by future use of `shared_from_this` when connecting a `ReactorNet`.
- Changes are propagated to all relevant API’s, i.e. Python, CLib and MATLAB.
- ~Add wall names to `ReactorNet` graphviz illustrations.~ _Edit: Moved to #1792_

Closes Cantera/enhancements#213

**Checklist**


ischoegl:
@speth ... while there are likely more paper cuts, I don't necessarily want to go beyond 2k lines. 

Among the things that _could_ be tackled are lowercase/hyphenated type names to be consistent with what we do with `ThermoPhase` and `Kinetics`, providing a name to `ReactorNet`, as well as ~merging `Reactor.thermo`/`Reactor.kinetics` into `Reactor.contents` in the Python API~ (_Edit: in view of Cantera/enhancements#201, there may be better approaches_). Those are relatively minor additions that could be done in a follow-up PR (if so desired). At the same time, there will be further updates in 3.2, although I'd like to keep those mostly _under the hood_.

ischoegl:
I just realized that there are some connections to Cantera/enhancements#201 and Cantera/enhancements#202. While this PR doesn't address either directly, it does ensure that a resolution will become easier.

Without trying to pre-empt any eventual resolution, my personal take would be:
- Have each `Reactor` 'own' a copy of a`Solution` object exclusively to address Cantera/enhancements#201, which would likely require the implementation of a copy constructor for `Solution`. We then could attach `Solution` methods directly to `ReactorNode`, which would eliminate the need to access a `thermo` or `kinetics` member variable.
- Decouple `ReactorSurface` from bulk reactors to address Cantera/enhancements#202. While this PR de facto declares `ReactorSurface` to be its own "node" already, I have not implemented an "edge" that ties a surface reactor to a bulk phase reactor. I was toying with the idea to replace `ReactorSurface.install` /  `ReactorBase::addSurface` by new `Connector` specialization but ultimately decided against it. At the same time, it would be a logical next step.

include/cantera/zeroD/ReactorNet.h
```diff
@@ -12,6 +12,13 @@
 
 namespace Cantera
 {
+//! @defgroup zerodGroup Zero-Dimensional Reactor Networks
+//!
+//! @details See the
+//! [Reactor Science](https://cantera.org/science/reactors/reactors.html)
```
speth: This link is to the legacy website. To create a link that will stay within the correct section (stable or dev) of the website, this should be a relative link:
```suggestion
//! @details See the
//! [Reactor Science](../reference/reactors/index.html)
```
This is a bit tricky, but it is described at https://cantera.org/dev/develop/doc-formatting.html#useful-doxygen-syntax

include/cantera/zeroD/ReactorFactory.h
```diff
@@ -34,28 +34,41 @@ class ReactorFactory : public Factory<ReactorBase, shared_ptr<Solution>, const s
...
+//! @deprecated Empty reactors will no longer be supported after %Cantera 3.1.
+//!     Use newReactorNode() with contents instead.
+shared_ptr<ReactorNode> newReactorNode(const string& model);
+
+//! Create an empty ReactorBase object of the specified type.
 //! @since Starting in %Cantera 3.1, this method returns a `shared_ptr<ReactorBase>`
-//! @deprecated Transitional method. Use newReactor() with contents instead.
+//! @deprecated Empty reactors will no longer be supported after %Cantera 3.1.
+//!     Superseded by newReactorNode().
 shared_ptr<ReactorBase> newReactor(const string& model);
```
speth: I think this is creating an interface that will be difficult to use without a lot of `dynamic_pointer_cast`s, which should generally be avoided. Objects of type `ReactorNode` define a very limited interface that isn't very useful on it's own.

include/cantera/zeroD/Reactor.h
```diff
@@ -307,6 +316,15 @@ class Reactor : public ReactorBase
...
 };
+
+//! Create a new %Reactor.
+//! @param contents  Solution object describing contents.
+//! @param name  Name of the reactor. Optional; if left empty, a default name will be
+//!     assigned when the reactor is integrated into a ReactorNet.
+//! @ingroup reactorGroup
+shared_ptr<Reactor> newReactor(
+    shared_ptr<Solution> contents, const string& name="");
+
```
speth: Adding this function for every derived class is a lot of boilerplate -- I think if someone wants to create a reactor object with a specific type, they can just use `Reactor::create`, and we can leave `newReactor` as the sole polymorphic factory method.

Or, if you prefer the `newConstPressureReactor` etc. functions, you could make these functions friends of the corresponding classes and do away with the `create` method.

include/cantera/zeroD/ReactorBase.h
```diff
@@ -155,10 +130,15 @@ class ReactorBase
...
 
+    //! @deprecated  To be removed after %Cantera 3.1. Replaced by version using
+    //!     shared pointer.
     virtual void addSurface(ReactorSurface* surf);
 
-    //! Return a reference to the *n*-th ReactorSurface connected to this
-    //! reactor
+    //! Add a %ReactorSurface to a bulk reactor.
+    //! @since New in %Cantera 3.1. Replaces version using raw pointer.
+    virtual void addSurface(shared_ptr<ReactorNode> surf);
```
speth:
```suggestion
    //! Add a ReactorSurface to a bulk reactor.
    //! @since New in %Cantera 3.1. Replaces version using raw pointer.
    virtual void addSurface(shared_ptr<ReactorNode> surf);
```

samples/cxx/combustor/combustor.cpp
```diff
@@ -63,42 +62,37 @@ void runexample()
...
-    // two reactors with different reaction mechanisms and different
-    // numbers of species. Downstream and upstream species are matched by
-    // name.
-    MassFlowController m2;
-    m2.install(air_in, combustor);
-    m2.setMassFlowRate(air_mdot);
+    // Now create the air mass flow controller. Note that this connects two
+    // reactors with different reaction mechanisms and different numbers of
+    // species. Downstreamand upstream species are matched by name.
+    auto m2 = newMassFlowController(air_in, combustor);
```
speth:
```suggestion
    // species. Downstream and upstream species are matched by name.
    auto m2 = newMassFlowController(air_in, combustor);
```

include/cantera/zeroD/ConnectorFactory.h
```diff
@@ -0,0 +1,91 @@
...
+
+class FlowDevice;
+class WallBase;
+
+//! Factory class to create Connector objects.
+//!
+//! This class is mainly used via the newConnector() function, for example:
+//!
+//! ```cpp
+//!     shared_ptr<Connector> mfc = newConnector("Valve", r0, r1, "my_valve");
```
speth:
```suggestion
//!     shared_ptr<Connector> valve = newConnector("Valve", r0, r1, "my_valve");
```
And again in the group description below.

samples/cxx/combustor/combustor.cpp
```diff
@@ -63,42 +62,37 @@ void runexample()
...
-    v.install(combustor, exhaust);
+    auto v = newValve(combustor, exhaust);
     double Kv = 1.0;
-    v.setValveCoeff(Kv);
+    v->setValveCoeff(Kv);
 
     // the simulation only contains one reactor
     ReactorNet sim;
-    sim.addReactor(combustor);
+    sim.addReactor(*combustor);
```
speth: Don't you want to use the `shared_ptr` version of `addReactor` here?

samples/cxx/kinetics1/kinetics1.cpp
```diff
@@ -50,7 +50,7 @@ int kinetics1(int np, void* p)
     // create a container object to run the simulation
     // and add the reactor to it
     ReactorNet sim;
-    sim.addReactor(r);
+    sim.addReactor(*r);
```
speth: Likewise here.

src/zeroD/Connector.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include "cantera/zeroD/ReactorNode.h"
+
+namespace Cantera
+{
+
+bool Connector::setDefaultName(map<string, int>& counts)
+{
+    if (m_defaultNameSet) {
+        return false;
+    }
```
speth: With the logic around incrementing `counts` brought into this method, the return value for this method (and its companions in other classes) is unnecessary.

include/cantera/zeroD/ConnectorFactory.h
```diff
@@ -0,0 +1,91 @@
...
+//! @since New in %Cantera 3.1.
+shared_ptr<Connector> newConnector(const string& model,
+                                   shared_ptr<ReactorNode> r0,
+                                   shared_ptr<ReactorNode> r1,
+                                   const string& name="(none)");
+
+//! Create a FlowDevice object of the specified type
+//! @since Starting in %Cantera 3.1, this method returns a `shared_ptr<FlowDevice>`
+//! @deprecated Replaced by newConnector. To be removed after %Cantera 3.1.
+shared_ptr<FlowDevice> newFlowDevice(const string& model, const string& name="(none)");
```
speth: The `newFlowDevice` method was already marked as deprecated in the 3.0 release (noted as "To be changed after Cantera 3.0; for new behavior, see newFlowDevice3()."), so I think we could just remove it now. This applies to a few of the other factory methods as well.

include/cantera/zeroD/ConnectorFactory.h
```diff
@@ -0,0 +1,91 @@
...
+
+//! Create a WallBase object of the specified type
+//! @since Starting in %Cantera 3.1, this method returns a `shared_ptr<WallBase>`
+//! @deprecated Replaced by newConnector. To be removed after %Cantera 3.1.
+shared_ptr<WallBase> newWall(const string& model, const string& name="(none)");
+
+//! Create a WallBase object of the specified type
+//! @since New in %Cantera 3.0.
+//! @deprecated Replaced by newWall. To be removed after %Cantera 3.1.
+shared_ptr<WallBase> newWall3(const string& model);
```
speth: Given that there is also a deprecated version of `newWall`, this deprecation should specify the signature of the new overload.

include/cantera/zeroD/ConnectorFactory.h
```diff
@@ -0,0 +1,91 @@
...
+
+//! Create a FlowDevice object of the specified type
+//! @since New in %Cantera 3.0.
+//! @deprecated Replaced by newFlowDevice. To be removed after %Cantera 3.1.
+shared_ptr<FlowDevice> newFlowDevice3(const string& model);
+
+//! Create a WallBase object of the specified type
+//! @since Starting in %Cantera 3.1, this method returns a `shared_ptr<WallBase>`
+//! @deprecated Replaced by newConnector. To be removed after %Cantera 3.1.
+shared_ptr<WallBase> newWall(const string& model, const string& name="(none)");
```
speth: Given the limited direct usefulness of a `Connector` object, I think we should direct users to the more specific method `newWall` that takes the two adjacent reactors as arguments.

src/zeroD/ReactorSurface.cpp
```diff
@@ -7,22 +7,58 @@
...
+    m_solution->thermo()->addSpeciesLock();
+
+    m_thermo = dynamic_cast<SurfPhase*>(&m_kinetics->thermo(0));
+    if (m_thermo == nullptr) {
+        // This should never happen, as the Solution should already contain the correct
+        // SurfPhase object
+        throw CanteraError("ReactorSurface::ReactorSurface",
+            "Specified kinetics manager does not represent a surface "
+            "kinetics mechanism.");
     }
```
speth, line 48: Better still, `m_thermo` could be set based on `m_solution` rather than needing to extract the phase from the `Kinetics` object. Or you could start by requiring an `Interface` object rather than just a `Solution`.

src/zeroD/Reactor.cpp
```diff
@@ -22,17 +22,18 @@ namespace bmt = boost::math::tools;
...
-            "raise\nexceptions thereafter; reactor contents should be provided in the "
-            "constructor.");
+    return Reactor::create(contents, name);
+}
+
+Reactor::Reactor(shared_ptr<Solution> sol, const string& name) : ReactorBase(sol, name)
+{
+    if (!m_solution) {
+        // warning was raised by ReactorNode; converted to exception after Cantera 3.1.
         return;
```
speth: Might be worth marking with a `@todo` to indicate that this check can be removed after Cantera 3.1.

src/zeroD/Reservoir.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+#include "cantera/zeroD/Reservoir.h"
+
+namespace Cantera
+{
+
+shared_ptr<Reservoir> newReservoir(shared_ptr<Solution> contents, const string& name)
+{
+    return Reservoir::create(contents, name);
+}
```
speth: It's probably better to just put this single one-liner in the header file rather than introducing another translation unit to compile.

include/cantera/zeroD/Wall.h
```diff
@@ -164,10 +171,21 @@ class Wall : public WallBase
...
+            "To be removed after Cantera 3.1. "
+            "Replaced by version using shared pointer.");
         m_qf = q;
     }
 
+    //! Specify the heat flux function @f$ q_0(t) @f$.
+    //! @since New in %Cantera 3.1. Replaces version using raw pointer.
+    void setHeatFlux(shared_ptr<Func1> q) {
+        m_qf = q.get();
+    }
```
speth: `Wall` needs to hold onto a copy of the `shared_ptr` -- otherwise, if the caller isn't also holding onto one, the wall will be left with a dangling pointer. Likewise for the velocity function, and the `Func1` usage in `FlowDevice`.

interfaces/cython/cantera/drawnetwork.py
```diff
@@ -321,7 +321,7 @@ def draw_walls(walls, graph=None, graph_attr=None, node_attr=None, edge_attr=Non
 
         edge_attr = {"color": "red", "style": "dashed", **(edge_attr or {})}
         graph.edge(inflow_name, outflow_name,
-                   **{"label": f"q = {rate:.2g} W", **edge_attr, **w.edge_attr,
-                      **edge_attr_overwrite})
+                   **{"label": f"{w.name}\\nq̇ = {rate:.2g} W",
```
speth: :heart:

interfaces/cython/cantera/reactor.pyx
```diff
@@ -3,37 +3,77 @@
...
+
+    @name.setter
+    def name(self, name):
+        self.node.setName(stringify(name))
+
+    def sync_state(self):
+        """
+        Set the state of the Reactor to match that of the associated `ThermoPhase`
+        object. After calling sync_state(), call ReactorNet.reinitialize() before
+        further integration.
```
speth:
```suggestion
        Set the state of the Reactor to match that of the associated `ThermoPhase`
        object. After calling `sync_state()`, call `ReactorNet.reinitialize()` before
        further integration.
```

include/cantera/zeroD/ReactorBase.h
```diff
@@ -47,41 +39,23 @@ struct SensitivityParameter
...
 
-    //! Set the default name of a reactor. Returns `false` if it was previously set.
-    bool setDefaultName(map<string, int>& counts);
-
     //! Set the Solution specifying the ReactorBase content.
     //! @param sol  Solution object to be set.
     //! @since New in %Cantera 3.1.
+    //! @deprecated  To be removed after %Cantera 3.1. Superseded by instantiation of
+    //!              ReactorBase with Solution object.
     void setSolution(shared_ptr<Solution> sol);
```
speth, line 62: Now that I think about this more, my preferred resolution to Cantera/enhancements#201 will require allowing `ReactorNet` to replace `Solution` objects in `Reactors`.

test/python/test_reactor.py
```diff
@@ -1074,39 +1069,42 @@ def test_draw_network(self):
...
+            '\tRC -> Out [label="pc_c\\nṁ = 2 kg/s" color=green]\n',
+            '\tInC -> RC [label="mfc_c\\nṁ = 2 kg/s" color=green]\n',
+            '\tRH -> RC [label="wall\\nq̇ = 4e+03 W" color=orange style=dashed]\n'
         }
         # use sets because order can be random
-        assert set(graph.body) == expected
+        # expected defines two alternatives for inH -> RH and RH -> Out as test defines
+        # pairs of redundant mass flow controllers and pressure controllers; dependent
+        # on type of reactor, either first or second is chosen for the visualization.
+        assert not set(graph.body) - expected
```
speth: This makes the graph a bit misleading, if we're labeling an edge as corresponding to a specific MFC, but then displaying the mass flow through two distinct MFCs. This may be an argument for the alternative view of the network, where the connectors are also nodes in the graph.

src/clib/ctreactor.cpp
```diff
@@ -123,7 +100,7 @@ extern "C" {
     double reactor_mass(int i)
     {
         try {
-            return ReactorCabinet::at(i)->mass();
+            return ReactorCabinet::as<ReactorBase>(i)->mass();
```
speth, line 103: The fact that almost all usage of `ReactorNode` objects requires downcasting to `ReactorBase`, `ReactorSurface`, or `Reactor` suggests to me that we shouldn't be defining our interface in terms of `ReactorNode` objects (which is not to say that the base class isn't useful for implementing some features like the automatic name generation).

samples/python/reactors/mix1.py
```diff
@@ -84,4 +84,8 @@
 # %%
 # Show the network structure
 # --------------------------
-diagram = sim.draw(print_state=True, species="X")
+try:
+    diagram = sim.draw(print_state=True, species="X")
+    diagram.view("mix1.gv")
+except ImportError as err:
+    print(f"Unable to show network structure:\n{err}")
```
speth: The behavior of causing tabs to open in my browser when building the docs from the command line is not my favorite. This also makes the diagram show up in a separate browser window when using the Jupyter notebook version of the example, rather than the inline view.

speth (changes requested):
I think the changes to have `Reactor` objects always contain a valid `Solution` object make sense, and I agree that this will help with Cantera/enhancements#201. I also like the idea of always providing the `Reactor` objects to the constructor for the connector objects and bringing that feature of the Python API to all the interfaces.

However, it's not quite clear to me what the benefit is of making all `Reactor` and connector objects into mandatory `shared_ptr`s. Currently, the ownership model is simple: none of the `Reactor`/`ReactorNet`/`FlowDevice`/etc. objects take ownership of anything they are connected to; the lifetimes of all these objects are managed by the caller. Within this set of objects, we need connections in all directions. Avoiding reference cycles will require the use of `weak_ptr` for many of the connections, with the associated overhead of "upgrading" those to `shared_ptr` where they are used. And you will still be left with the situation where it's possible to hold on to some element of the network but not be able to use it because the rest of the network has been deleted, though at least that would be checked and generate an exception rather than undefined behavior. My apologies for not thinking this through earlier and discussing it on Cantera/enhancements#213 before you put the time into implementing it.

Regarding the extension of the graph idiom, I think there's a simplification that may be worth considering. Instead of seeing just `Reactor` and `ReactorSurface` objects as nodes in the graph, you could also see valves and flow devices as nodes, with the edges of the graph just being the abstract connections among all of the entities that make up the reactor network. This way, there's only the need for one new base class, and no need for anything to sit between a `Reactor` and a `ReactorSurface`. This also gets closer to the original intent of Cantera/enhancements#212, and may work better as the amount of information you might want to display about a connector increases, given some of the limitations of how Graphviz handles (or doesn't) text associated with edges. That said, I do like the `Connector` class as a generalization combining walls and flow devices, and there's probably some further room to adjust this idea.

include/cantera/zeroD/ReactorBase.h
ischoegl: Could you expand? My thinking was a strict hierarchy where the `ReactorNet` is responsible for `Reactor`, which themselves are responsible for `Solution`. I am not sure that I see why we should give `ReactorNet` the ability to muddle with reactor contents?

samples/python/reactors/mix1.py
ischoegl: I added this as when things are run in IPython or Python from the command line (my preferred avenue), nothing is shown. Still, removing the line based on feedback (moved to #1792).

test/python/test_reactor.py
ischoegl: This is a pre-existing condition that was not previously evident. Without major re-implementation, the explanation provided here is all we can do? I personally do not follow the rationale to use two MFC's in parallel between the same reservoir. Closing here as I am moving this to a separate PR (#1792)

ischoegl:
@speth ... I took care of the easy bits, but wanted to see whether we could discuss some of the more difficult parts before going into more details. I am ultimately not sure what the direction should be - just take out mandatory `shared_ptr`? Also, regarding `ReactorNode` I do see an alternative of deriving `ReactorSurface` from `ReactorBase` - I had seen the latter as something that contains a bulk phase, but I do see that simplifications may make sense. I do not think that Connectors should be combined with a more generic node redefinition, as they are very different from reactors.

include/cantera/zeroD/ReactorBase.h
speth: See https://github.com/Cantera/enhancements/issues/201#issuecomment-2349477048

ischoegl:
Based on feedback above, I suggest to shelve this PR for the time being as it may be a distraction that keeps us from releasing 3.1 sooner rather than later.

src/zeroD/ReactorSurface.cpp
ischoegl: This is an example for what needs to change eventually. This PR was an attempt to update the API first and tackle the underpinnings in follow-up work. Based on the feedback received, I do believe that a more comprehensive/strategic approach is warranted.

_Edit: the plan is to completely remove `ReactorBase::m_thermo` and replace it by `ReactorNode::m_solution` after the release of 3.1._

src/zeroD/Reservoir.cpp
ischoegl: ~I created this as otherwise we'd end up with duplicate symbols.~ ... moot: now using `Reservoir::create`.

include/cantera/zeroD/ConnectorFactory.h
ischoegl: Now replaced by `Wall::create`.

include/cantera/zeroD/Reactor.h
ischoegl: Based on this feedback, I removed all `new*Reactor` functions, and move to various `Reactor::create` variants. The sole polymorphic factory method is now `newReactorNode`.

include/cantera/zeroD/ReactorFactory.h
ischoegl: Removed.

include/cantera/zeroD/ReactorBase.h
ischoegl: I responded in https://github.com/Cantera/enhancements/issues/201#issuecomment-2351041080. Irrespective of the logic we go with eventually, the detection of previously used `Solution` objects should be handled by `ReactorNode` and not `ReactorNet`.

src/clib/ctreactor.cpp
ischoegl: Once the API is fixed, `ReactorBase` methods should be either shifted to `ReactorNode` or `Reactor`, with `ReactorBase` eventually being removed. From that perspective, this is transitional behavior.

Adding the following to `ReactorBase.h`:
```
 * @todo After completion of the %Cantera 3.1 deprecation cycle, all methods should be
 *      either moved to ReactorNode or Reactor, with ReactorBase entering its own
 *      deprecation cycle in %Cantera 3.2.
```

ischoegl:
@speth ... thank you again for your feedback. After further consideration, I decided to move ahead with revisions on this PR after all, as some of the API changes are necessary to tackle updates envisioned by various enhancement requests. 

I adopted most of your suggestions, specifically:
- Reduce boilerplate, and use `XYZ::create` methods directly rather than `newXYZ` convenience functions.
- Remove functions previously marked as deprecated.
- Leave default constructors public.
- Moved `ReactorNet` illustration updates to a separate PR.

Beyond, I tried to clarify future plans that motivated API changes, see various comments in the contexts of  Cantera/enhancements#201, Cantera/enhancements#202 and Cantera/enhancements#213.

speth:
> Once the API is fixed, `ReactorBase` methods should be either shifted to `ReactorNode` or `Reactor`, with `ReactorBase` eventually being removed. From that perspective, this is transitional behavior.
> 
> Adding the following to `ReactorBase.h`:
> 
> ```
>  * @todo After completion of the %Cantera 3.1 deprecation cycle, all methods should be
>  *      either moved to ReactorNode or Reactor, with ReactorBase entering its own
>  *      deprecation cycle in %Cantera 3.2.
> ```

*(moving out of a specific line-by-line comment to prevent GitHub from hiding this higher-level discussion)*

I think this gets to my main concern with the API changes that this PR creates. Right now, there's a bit too much importance being placed on the idea of the reactor network as a graph -- most users don't care about this, but it now becomes the most important thing about a reactor and represents almost the only set of directly accessible functionality when using `newReactorNode` over the now-deprecated `newReactor` method.

I'd like to explore whether we can directly skip over creating the `ReactorNode` class and make `ReactorBase` serve this purpose immediately. The main difference that has with the current structure is that it would make `ReactorSurface` into a `ReactorBase`, but I think that's a step in the right direction anyway, given the other discussion we've had about the future of handling surfaces and edges in Cantera/enhancements#202. There aren't that many functions of `ReactorBase` that aren't applicable to a `ReactorSurface` and we can make those virtual functions and override them as needed.

I'd rather do this than leave users with an unnecessarily difficult-to-use interface for a full Cantera version.

ischoegl:
Thanks for your response, @speth.

> I think this gets to my main concern with the API changes that this PR creates. Right now, there's a bit too much importance being placed on the idea of the reactor network as a graph -- most users don't care about this, but it now becomes the most important thing about a reactor and represents almost the only set of directly accessible functionality when using `newReactorNode` over the now-deprecated `newReactor` method.

Interesting thought - my attempt to simplify was definitely influenced by viewing 'reactor networks as a graph', although my objective was to simplify/generalize rather than making things more complicated.

> I'd like to explore whether we can directly skip over creating the `ReactorNode` class and make `ReactorBase` serve this purpose immediately. The main difference that has with the current structure is that it would make `ReactorSurface` into a `ReactorBase`, but I think that's a step in the right direction anyway, given the other discussion we've had about the future of handling surfaces and edges in [Cantera/enhancements#202](https://github.com/Cantera/enhancements/issues/202). There aren't that many functions of `ReactorBase` that aren't applicable to a `ReactorSurface` and we can make those virtual functions and override them as needed.

I agree that we have one superfluous layer, which needs to be addressed. I was thinking along the lines of differentiating between bulk and surface reactors. I still think that _nodes_ and _edges_ make sense overall, so my introduction of a new base class was a move to phase out `ReactorBase` while providing a little more flexibility.

> I'd rather do this than leave users with an unnecessarily difficult-to-use interface for a full Cantera version.

I think at this point it makes sense to defer to 3.2. I will change the project planner accordingly.

ischoegl:
@speth ... I pulled out the less controversial commits to #1848; I also added some comments to Cantera/enhancements#213.

PR #1791, opened 2024-09-03
Avoid star import from buildutils

ischoegl:

Replace 
```Python
from buildutils import *
```

with imports of specific methods and classes. The change improves readability and avoids unnecessary clutter in various namespaces.

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -90,11 +90,16 @@ from packaging.specifiers import SpecifierSet
 from packaging.version import parse as parse_version
 import SCons
 
+from buildutils import (Option, PathOption, BoolOption, EnumOption, Configuration,
+                        logger, remove_directory, remove_file, test_results,
+                        add_RegressionTest, get_command_output, listify, which,
+                        ConfigBuilder, multi_glob, quoted, add_system_include,
+                        checkout_submodule, check_for_python, check_sundials,
+                        config_error, run_preprocessor, make_relative_path_absolute)
+
 # ensure that Python and SCons versions are sufficient for the build process
 EnsurePythonVersion(3, 7)
 EnsureSConsVersion(3, 0, 0)
-
-from buildutils import *
 
 if not COMMAND_LINE_TARGETS:
     # Print usage help
@@ -1853,7 +1858,7 @@ cdefine('CT_USE_SYSTEM_YAMLCPP', 'system_yamlcpp')
 
 config_h_build = env.Command('build/src/config.h.build',
                              'include/cantera/base/config.h.in',
-                       ConfigBuilder(configh))
+                             ConfigBuilder(configh))
 # This separate copy operation, which SCons will skip if config.h.build is
 # unmodified, prevents unnecessary rebuilds of the precompiled header
 config_h = env.Command('include/cantera/base/config.h',
--- a/doc/SConscript
+++ b/doc/SConscript
@@ -1,11 +1,9 @@
-
 from collections import namedtuple
-from os.path import join as pjoin
 from pathlib import Path
-import os
 import subprocess
-from buildutils import *
 from SCons import Errors
+
+from buildutils import multi_glob
 
 Import('env', 'build', 'install')
 
--- a/interfaces/cython/SConscript
+++ b/interfaces/cython/SConscript
@@ -2,7 +2,9 @@
 import re
 from pathlib import Path
 from packaging.version import parse as parse_version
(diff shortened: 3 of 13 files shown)


SConstruct
```diff
@@ -90,12 +90,17 @@ from packaging.specifiers import SpecifierSet
 from packaging.version import parse as parse_version
 import SCons
 
+from buildutils import (Option, PathOption, BoolOption, EnumOption, Configuration,
+                        logger, remove_directory, remove_file, test_results,
+                        add_RegressionTest, get_command_output, listify, which,
+                        ConfigBuilder, multi_glob, quoted, add_system_include,
+                        checkout_submodule, check_for_python, check_sundials,
+                        config_error, run_preprocessor, make_relative_path_absolute)
```
bryanwweber, line 98: I'm not sure this one is actually better than the `*`. Otherwise, the rest look good to me :smile:

ischoegl: Citing from [PEP 8](https://peps.python.org/pep-0008/#imports):

> Wildcard imports (`from <module> import *`) should be avoided, as they make it unclear which names are present in the namespace, confusing both readers and many automated tools. There is one defensible use case for a wildcard import, which is to republish an internal interface as part of a public API (for example, overwriting a pure Python implementation of an interface with the definitions from an optional accelerator module and exactly which definitions will be overwritten isn’t known in advance).

The alternative would be to use `import buildutils as bu` etc., but I honestly think that the long list at the beginning of the module is preferable.

bryanwweber: Yes, I'm familiar with PEP 8. I think in this case, it makes more sense to ignore pep 8's recommendation. I should explain my reasoning.

In this case, I think this long block is less readable and maintainable than the *. I feel that way because, well , readability is just vibes as the kids say. But I do think it's harder to maintain this long list here rather than in the dunder all in buildutils. 

There are a couple of other reasons that pep 8 makes that recommendation against * (although they're not stated explicitly), which I think don't apply here.

1. I also considered that the * import can't be resolved by the LSP implementation. However, even the explicit import can't be resolved because `site_scons` is added to the path by SCons at runtime.
2. Similarly, I don't think naming everything on the import makes it more obvious where they come from because buildutils is hidden inside site_scons.
2. I also considered namespace pollution from buildutils. Since we define dunder all in buildutils, we are essentially defining what should be exported rather than imported. In any case, this avoids the pollution issue since we use nearly every exported member in the main SConstruct. 
3. SCons also does a bunch of magical injections of objects at runtime that you can use without importing. So in that sense using a * fits in the same pattern as SCons is already using. 

So, considering all that, my opinion is that the * is the better tradeoff ☺️

ischoegl: From the point of a maintainer, I can probably see your points. I do, however, believe that we should approach this from the perspective of users who are _not_ familiar with SCons and other intricacies, all of which have a steep learning curve. The fact that SCons makes a couple of design decisions that are debatable (points 1/2/4 - `site_scons` and other magic injections) is imho not a reason to throw Python guidelines overboard.

Based on clarity, PEP8 clearly wins, as someone who stumbles across any of the custom utils, - say `add_system_include`, - would presumably search for this in `SConstruct` but would not find where it is defined if it is imported via `from buildutils import *` unless they searched across the entire Cantera source code. Using VSCode, Pylance complains as follows:
```
"add_system_include" is not defined
```
so anyone not searching across all of Cantera won't be able to find it. If we use the list, they would at least know that it's from `buildutils` and not part of something magically injected by SCons. While Pylance still doesn't find the definition with the list import, it no longer complains.

Fwiw, the `__all__` in `buildutils.py` - point 3 - is similarly cumbersome to maintain (and should probably be removed as part of this PR). _Edit: pushed a commit that will remove it._

bryanwweber: I think we're approaching the problem with the same facts and just prioritizing differently. We may have to agree to disagree and call in @speth to break the tie 😁

speth: To me, this ends up being a pretty minor difference in the end. I think there end up being a few points in favor of importing the specific names, though.

One is that every Python user understands what `from x import y` means, whereas the use of `__all__` is a somewhat obscure detail.

Also, I think the long list, as ugly as it is, is easier to maintain as part of an `import` statement, since in that case your IDE can inform you when there's an unused import, whereas I don't think this is meaningful/possible for `__all__`.

speth (approved):

PR #1792, opened 2024-09-12
Graphviz tweaks

### LINKED ISSUE #212: Identify flow devices and walls in `graphviz` reactor network illustrations
**Abstract**

~The recently added reactor network visualization (see #180) does a great job visualizing reactor networks, but omits blocks representing flow devices (`Valve`, `MassFlowController`, etc) and walls (`Wall`). Having those connectors visualized would ensure a better understanding of a network setup. As an aside, the _"Gas Splitter"_ and _"Gas Mixer"_ blocks in the CHEMKIN example shown in #180 may be a loose equivalent.~

_Edit: Based on feedback, I am amending this feature request to simply display names of walls and flow devices along the edges._

**Motivation**

* What problem is it trying to solve? ... _improve a recent Cantera enhancement_
* Who is affected by the change? ... _anyone interested in the new `ReactorNet.draw()` method_
* Why is this a good solution? ... _adds important information that is currently missing_

**Possible Solutions**

Add additional blocks representing flow devices to the implementation introduced by @Naikless in Cantera/cantera#1624.

**References**

- #180
- Cantera/cantera#1624

ischoegl:

This PR separates `ReactorNet` visualization tweaks from #1788 to facilitate further reviews.

- Add wall names to ReactorNet graphviz illustrations.
- Partially addresses review comment by removing `diagram.view`

There is a caveat here, - see Cantera/enhancements#212, - as well as per @speth's comment:

> Regarding the extension of the graph idiom, I think there's a simplification that may be worth considering. Instead of seeing just `Reactor` and `ReactorSurface` objects as nodes in the graph, you could also see valves and flow devices as nodes, with the edges of the graph just being the abstract connections among all of the entities that make up the reactor network. This way, there's only the need for one new base class, and no need for anything to sit between a `Reactor` and a `ReactorSurface`. This also gets closer to the original intent of Cantera/enhancements#212, and may work better as the amount of information you might want to display about a connector increases, given some of the limitations of how Graphviz handles (or doesn't) text associated with edges. That said, I do like the `Connector` class as a generalization combining walls and flow devices, and there's probably some further room to adjust this idea.

On the other hand, there is @Naikless's original intent:

> In classical graph terminology, when I introduced the visualization, I envisioned the flow devices and the walls as "edges" that each connect two "nodes" that are reactors or reservoirs. So now they are represented as arrows of different shape and color. This way, the direction of both material and energy flow can be indicated relatively easily. Their main property (the amount of flow) is still displayed.

> Having them represented as nodes could imo quickly lead to very cluttered networks, especially since there are quite a few scenarios conceivable where two reactors have a multitude of connections to one another. But I am happy to hear your arguments on this.

This PR attempts to stick with the original intent.

Closes https://github.com/Cantera/enhancements/issues/212 (minor aspect of this PR)

**Checklist**

### DIFF
--- a/test/python/test_reactor.py
+++ b/test/python/test_reactor.py
@@ -1017,11 +1017,11 @@ class TestReactor(utilities.CanteraTest):
         self.make_reactors(T1=T1, P1=P1, X1=X1, T2=T2, P2=P2, X2=X2)
         self.r1.name = "Name 1"
         self.r2.name = "Name 2"
-        w = ct.Wall(self.r1, self.r2, U=0.1, edge_attr={'style': 'dotted'})
+        w = ct.Wall(self.r1, self.r2, U=0.1, edge_attr={'style': 'dotted'}, name="wall")
         w.edge_attr = {'color': 'green'}
         graph = w.draw(node_attr={'style': 'filled'},
                        edge_attr={'style': 'dashed', 'color': 'blue'})
-        expected = [('\t"Name 2" -> "Name 1" [label="q = 30 W" '
+        expected = [('\t"Name 2" -> "Name 1" [label="wall\\nQ̇ = 30 W" '
                      'color=blue style=dashed]\n')]
         assert graph.body == expected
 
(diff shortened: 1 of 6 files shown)


test/python/test_reactor.py
```diff
@@ -1017,11 +1017,11 @@ def test_draw_wall(self):
         self.make_reactors(T1=T1, P1=P1, X1=X1, T2=T2, P2=P2, X2=X2)
         self.r1.name = "Name 1"
         self.r2.name = "Name 2"
-        w = ct.Wall(self.r1, self.r2, U=0.1, edge_attr={'style': 'dotted'})
+        w = ct.Wall(self.r1, self.r2, U=0.1, edge_attr={'style': 'dotted'}, name="wall")
         w.edge_attr = {'color': 'green'}
         graph = w.draw(node_attr={'style': 'filled'},
                        edge_attr={'style': 'dashed', 'color': 'blue'})
-        expected = [('\t"Name 2" -> "Name 1" [label="q = 30 W" '
+        expected = [('\t"Name 2" -> "Name 1" [label="wall\\nq̇ = 30 W" '
```
speth: These tests are failing due to the change in case between $\dot q$ and $\dot Q$. I like the capitalized version, but either is fine.

speth (commented):
This looks good to me, pending the fix for the failing tests.

test/python/test_reactor.py
ischoegl: Not sure why I didn't pick up on the test failures. I agree with your preference btw, where I had already implemented `Q̇`.

ischoegl:
@speth … test is fixed but we appear to have a new upstream regression ~with HighFive (only shows up on windows 2022, and likely related to a new conda package)~ ... _Edit: there isn't anything new in the HighFive package since their last release in July, so it's likely some other conda-related update. As these new failures are not caused by changes in this PR, I am suggesting to handle/investigate this separately._

speth (approved):

PR #1804, opened 2024-10-30
Make generated sources and one error message deterministic

keszybz:
Replace `list(set(…))` by `sorted(set(…))` to avoid randomness in builds and at runtime.


speth (approved):
Thanks!

PR #1826, opened 2024-12-07
Address minor typos

ischoegl:

Edits were prompted by [UG question](https://groups.google.com/g/cantera-users/c/Qj1YQ13AUXE/m/1-E9jpATAwAJ).

**Checklist**

### DIFF
--- a/doc/sphinx/reference/onedim/nonlinear-solver.md
+++ b/doc/sphinx/reference/onedim/nonlinear-solver.md
@@ -65,7 +65,7 @@ $$
 J(x) =\pxpy{F(x)}{x}
 $$
 
-For the vector-value residual $F(x)$, the Jacobian matrix has elmenets that are
+For the vector-value residual $F(x)$, the Jacobian matrix has elements that are
 partial derivatives of the residuals with respect to each solution component at each
 grid point.
 
@@ -160,7 +160,7 @@ $$
 During the search for the correct value of $\lambda$, the value of $\lambda$ starts
 at 1, it is adjusted down to a value that keeps the solution within the trust region.
 The process then begins for finding $\lambda$, failures result in the damping factor
-being reduced by a constant factor. The current factor in  Cantera is the $\sqrt{2}$.
+being reduced by a constant factor. The current factor in Cantera is the $\sqrt{2}$.
 
 During the damped Newton method, the Jacobian is kept at the $x^{(k)}$ value. This
 sometimes can cause issues with convergence if the Jacobian becomes out of date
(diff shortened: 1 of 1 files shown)


src/oneD/Flow1D.cpp
```diff
@@ -732,21 +732,25 @@ void Flow1D::evalSpecies(double* x, double* rsd, int* diag,
...
         double sum = 0.0;
         for (size_t k = 0; k < m_nsp; k++) {
             sum += Y(x,k,jmax);
-            rsd[index(k+c_offset_Y, jmax)] = m_flux(k, jmax-1) +
+            rsd[index(c_offset_Y+k, jmax)] = m_flux(k, jmax-1) +
                                              rho_u(x, jmax)*Y(x, k, jmax);
+            // diag[index(c_offset_Y+k, jmin)] are zero and do not need to be reset
         }
+        // Residuals may be modified by right boundary (for k != rightExcessSpecies);
+        // excess species residual is handled as algebraic constraint.
```
bryanwweber: Thanks for the updates here @ischoegl! I think it helps a bit to have a note about what is happening here, but the "what" is also clear from the code itself IMO. I think it would even better to have a note about _why_ we can treat the excess species with an algebraic constraint (which in other situations may result in overdetermined systems, see e.g., the species production equations where we allow small negative values rather than enforcing 1-sum(k-1)). I think a link to documentation about the why would suffice, if this is already explained some where. Recognizing as well that we don't need this to become a textbook on numerical methods... Thanks!

ischoegl: Thanks, @bryanwweber. I guess my point was mostly to re-introduce some older comments from `StFlow.cpp` (which were not preserved in the refactoring that took place recently):
```
            // The default boundary condition for species is zero flux. However,
            // the boundary object may modify this.
```
It is true that part of this comment was moved to the docstring of the method, but I don't think it is sufficiently clear when reading this implementation code that residuals are later modified. The diag setters were only partially implemented but also superfluous (so reminding users that there is no time stepping involved is imho a better option). The algebraic constraint for mass fraction is imho self-evident, but I guess mileage varies? Overall, I tried to improve code that is not straight-forward (see recent UG question), and if you have a better suggestion, I'd be happy to adopt.

bryanwweber: The only suggestion I have is to explain (either here or on the website) why we can use an algebraic constraint for the residual here, specifically because I feel fairly confident that in other cases we can't. I don't know the answer, hence I assume others won't either. My comment was that the code already is clear that this is an algebraic constraint, so adding a comment which states that fact explicitly is of limited utility IMO. The why is far more important IMO.

ischoegl: Fair point, but since this is de facto legacy code, it's hard to guess. My intuition tells me that it may have something to do with the fact that these are boundary conditions and thus have to be treated differently than interior points, where we don't have that imposed constraint. Will remove the second commit as apparently this isn't an improvement over the status quo and I don't have time to dig into the theory.

bryanwweber (approved):

PR #1835, opened 2024-12-29
Experimental `clib` from doxygen

ischoegl:

This PR seeks to implement a basis for future auto-generated Cantera APIs that are fully documented based on information provided in C++ docstrings.

- New *sourcegen* scaffolder generates an experimental `clib` version ("`clib-experimental`")
- The `clib` scaffolder is based on doxygen tag files and associated XML anchorfiles
- A proposed minimal YAML syntax is used to reference C++ functions and methods (for details, see file `interfaces/sourcegen/sourcegen/_data/README.md`)
- A portion of `clib` googletests is adopted for illustration and code coverage as `test-clib-experimental`
- Partial CI is added to the `clang-compiler` action
- API information is obtained without having to parse pre-existing CLib headers (i.e. the original approach for the .NET API)

This proof of concept re-implements approaches of the traditional CLib library and is newr feature-complete, although the conversion of `zeroD` and `oneD` objects is currently omitted. ~One other currently omitted feature involves the inclusion of code that is difficult to automate, although it is expected to be rare;~ in many instances, tweaks to the C++ interface may be the go-to approach instead. _Edit: custom code is now supported._

The configuration involves YAML markup that only uses minimal information for each entry, e.g.

```yaml
- name: newSolution
  implements: newSolution(const string&, const string&, const string&)
  uses: [thermo, kinetics, transport]
- name: name
```

which produces fully documented declarations
```c++
/**
 *  Create and initialize a new Solution manager from an input file.
 *
 *  @param infile       name of the input file
 *  @param name         name of the phase in the file. If this is blank, the first phase in the file is used.
 *  @param transport    name of the transport model.
 *  @returns            Handle to stored Solution object or -1 for exception handling.
 *
 *  @implements constructor: shared_ptr<Solution> newSolution(const string&, const string&, const string&)
 *  @relates Solution::thermo(), Solution::kinetics(), Solution::transport()
 */
int sol3_newSolution(const char* infile, const char* name, const char* transport);

/**
 *  Return the name of this Solution object.
 *
 *  @param handle       Handle to queried Solution object.
 *  @param[in] bufLen   Length of reserved array.
 *  @param[out] buf     Returned string value.
 *  @returns            Actual length of string or -1 for exception handling.
 *
 *  @implements getter: string Solution::name()
 */
int sol3_name(int handle, int bufLen, char* buf);
```

<details><summary>and implementations (click to expand)</summary>

```C++
int sol3_newSolution(const char* infile, const char* name, const char* transport)
{
    // constructor: shared_ptr<Solution> newSolution(const string&, const string&, const string&)
    try {
        auto obj = newSolution(infile, name, transport);
        int id = SolutionCabinet::add(obj);
        // add all associated objects
        if (obj->thermo()) {
            ThermoPhaseCabinet::add(obj->thermo(), id);
        }
        if (obj->kinetics()) {
            KineticsCabinet::add(obj->kinetics(), id);
        }
        if (obj->transport()) {
            TransportCabinet::add(obj->transport(), id);
        }
        return id;
    } catch (...) {
    }
}

int sol3_name(int handle, int bufLen, char* buf)
{
    // getter: string Solution::name()
    try {
        string out = SolutionCabinet::at(handle)->name();
        copyString(out, buf, bufLen);
        return int(out.size()) + 1;
    } catch (...) {
    }
}
```
</details>

Here is an example for a full YAML input used for code generation
<details><summary><tt>sol3_auto.yaml</tt> (click to expand)</summary>

```yaml
# Auto-generated CLib API for %Cantera's Solution class.
# Partially implements a replacement for CLib's traditional @c ct library.

prefix: sol3
base: Solution
parents: []  # List of parent classes
derived: [Interface]  # List of specializations
recipes:
- name: newSolution
  implements: newSolution(const string&, const string&, const string&)
  uses: [thermo, kinetics, transport]
- name: newInterface
  implements:
    newInterface(const string&, const string&, const vector<shared_ptr<Solution>>&)
  uses: [thermo, kinetics]
- name: del
  uses: [thermo, kinetics, transport]
- name: name
- name: setName
- name: thermo
- name: kinetics
- name: transport
- name: setTransportModel
  code: |-
    try {
        auto obj = SolutionCabinet::at(handle);
        TransportCabinet::del(
            TransportCabinet::index(*(obj->transport()), handle));
        obj->setTransportModel(model);
        return TransportCabinet::add(obj->transport(), handle);
    } catch (...) {
    }
- name: nAdjacent
- name: adjacent
  implements: Solution::adjacent(size_t)
  uses: [thermo, kinetics, transport]
  what: constructor  # registers object in CLib storage
# - name: adjacentName
- name: cabinetSize
```
</details>

See Cantera/enhancements#39
Partially addresses Cantera/enhancements#220

Run the following command from the Cantera root folder:

```
scons doxygen
python3 interfaces/sourcegen/run.py --api=clib --output=.
scons build clib_experimental=y
```

A partial test suite based on the traditional CLib ensures that code performs as expected:

```
scons test-clib-experimental
```

**Checklist**


interfaces/sourcegen/sourcegen/_HeaderFileParser.py
```diff
@@ -8,20 +8,53 @@
...
+_data_path = Path(__file__).parent / "_data"
+
 class HeaderFileParser:
 
     def __init__(self, path: Path, ignore_funcs: List[str] = None):
         self._path = path
         self._ignore_funcs = ignore_funcs
 
+    @classmethod
+    def from_yaml(cls, ignore_files, ignore_funcs) -> List[HeaderFile]:
```
bryanwweber: Since we're supporting 3.10 and up now, we can use `list[HeaderFile]` and no longer need the `typing.List` (similarly `Tuple`, `Dict`, etc.) import :tada:

bryanwweber: On a separate note, I think a better name for this function would be `headers_from_yaml` or something, since it's not returning an instance of `HeaderFileParser`. Although I guess the `from_yaml` name has symmetry with `from_headers`... :confounded: I'm not sure what to do, I guess just pointing out the oddity in case anyone has any better ideas.

interfaces/sourcegen/sourcegen/_TagFileParser.py
```diff
@@ -0,0 +1,303 @@
+"""Parser for tag files and XML generated by doxygen."""
+
+# This file is part of Cantera. See License.txt in the top-level directory or
+# at https://cantera.org/license.txt for license and copyright information.
+
+import sys
+from pathlib import Path
+import re
```
bryanwweber, line 8: I'd really suggest using a library to parse the xml rather than string operations. Normally I'd recommend lxml but to avoid a third party dependency only for build time it might be better to use `ElementTree` from the stdlib.

interfaces/sourcegen/sourcegen/_TagFileParser.py
```diff
@@ -0,0 +1,303 @@
...
+    """Represents information parsed from a doxygen tag file."""
+
+    base: str = ""  #: qualified scope (skipping Cantera namespace)
+    type: str = ""  #: return type
+    name: str = ""  #: function name
+    arglist: str = ""  #: function argument list (original XML string)
+    anchorfile: str = ""  #: doxygen anchor file
+    anchor: str = ""  #: doxygen anchor
+
+    @staticmethod
```
bryanwweber: Should this be a class method so it can return `cls(...)`?

bryanwweber (commented):
Thanks Ingmar! I know this is still in draft but I wanted to leave a small comment that might be intrusive later

interfaces/sourcegen/sourcegen/_HeaderFileParser.py
ischoegl: Sure thing! I've been switching to `list` for the most part, but some of this code is backported from an old branch. :+1: on the suggestion of the method name ... since all of this is experimental, I switched both names.

interfaces/sourcegen/sourcegen/_TagFileParser.py
ischoegl: This is definitely a fair suggestion. The custom parser is backported from an old branch and works well for the time being, but I'll run a couple of tests before this is ready for review. 

Fwiw, `re` is really fast, given that the tag file itself is already 55k lines long, and there are a myriad of additional XML files that contain details. Not having to parse the entire tree may have some advantages, but there may be a middle ground that's better than the pure `re` approach.

bryanwweber: Yeah, I'm not so worried about performance, moreso correctness. Although that the file is autogenerated should be a benefit insofar as the tags should all be complete, etc.

ischoegl: Done.

ischoegl:
@speth and @bryanwweber ... this is ready for a review. 

~As an aside, there is a minor caveat for the .NET workflow on Windows, where a Python 3.10 version apparently does not respect [PEP 604](https://peps.python.org/pep-0604/). I'll address this together with other comments.~ _Edit: nevermind ... `windows-2022` appears to use an older system Python and my initial attempt to override failed._

interfaces/sourcegen/sourcegen/_HeaderFileParser.py
```diff
@@ -1,33 +1,90 @@
...
 class HeaderFileParser:
+    """
+    Parser for header files or corresponding YAML specifications.
+
+    Provides for convenience methods to generate lists of `HeaderFile` objects, which
+    themselves are used for subsequent code scaffolding.
+    """
 
-    def __init__(self, path: Path, ignore_funcs: List[str] = None):
+    def __init__(self, path: Path, ignore_funcs: list[str] = None):
```
bryanwweber:
```suggestion
    def __init__(self, path: Path, ignore_funcs: Iterable[str] = None) -> None:
```

interfaces/sourcegen/sourcegen/_HeaderFileParser.py
```diff
@@ -1,33 +1,90 @@
...
         self._path = path
         self._ignore_funcs = ignore_funcs
 
     @classmethod
-    def from_headers(cls, ignore_files, ignore_funcs) -> List[HeaderFile]:
+    def headers_from_yaml(cls, ignore_files, ignore_funcs) -> list[HeaderFile]:
```
bryanwweber:
```suggestion
    def headers_from_yaml(cls, ignore_files: Iterable[str], ignore_funcs: Iterable[str]) -> list[HeaderFile]:
```

interfaces/sourcegen/sourcegen/_HeaderFileParser.py
```diff
@@ -1,33 +1,90 @@
...
+        """Parse header file YAML configuration."""
+        files = [ff for ff in _DATA_PATH.glob("*.yaml") if ff.name not in ignore_files]
+        files.sort()
```
bryanwweber:
```suggestion
        files = sorted(ff for ff in _DATA_PATH.glob("*.yaml") if ff.name not in ignore_files)
```

interfaces/sourcegen/sourcegen/_HeaderFileParser.py
```diff
@@ -1,33 +1,90 @@
...
+        return [cls(ff, ignore_funcs.get(ff.name, []))._parse_yaml() for ff in files]
+
+    def _parse_yaml(self) -> HeaderFile:
+        def read_docstring():
+            doc = []
+            with self._path.open("r", encoding="utf-8") as fid:
+                while True:
+                    line = fid.readline()
+                    if line.startswith("#"):
+                        doc.append(line.lstrip("#").strip())
```
bryanwweber: This will remove all of the `#` at the start of a string, which I don't think we necessarily want. I think this would be more correct:
```suggestion
                        doc.append(line.remove_prefix("#").strip())
```

interfaces/sourcegen/sourcegen/_HeaderFileParser.py
```diff
@@ -1,33 +1,90 @@
...
+            return doc
+
+        config = read_config(self._path)
+        recipes = []
+        prefix = config["prefix"]
+        base = config["base"]
+        parents = config.get("parents", [])
+        derived = config.get("derived", [])
+        for recipe in config["recipes"]:
+            if recipe['name'] in self._ignore_funcs:
```
bryanwweber:
```suggestion
            if recipe["name"] in self._ignore_funcs:
```
I'd suggest running Black/Ruff on these changes if they don't cause too many other spurious changes :smile:

interfaces/sourcegen/sourcegen/_HeaderFileParser.py
```diff
@@ -1,33 +1,90 @@
...
+                       prefix,
+                       base,
+                       parents,
+                       derived))
+
+        return HeaderFile(self._path, [], prefix, base, parents, derived, recipes,
+                          read_docstring())
+
+    @classmethod
+    def headers_from_h(cls, ignore_files, ignore_funcs) -> list[HeaderFile]:
```
bryanwweber:
```suggestion
    def headers_from_h(cls, ignore_files: Iterable[str], ignore_funcs: Iterable[str]) -> list[HeaderFile]:
```

interfaces/sourcegen/sourcegen/_SourceGenerator.py
```diff
@@ -14,9 +13,9 @@ class SourceGenerator(metaclass=ABCMeta):
     """Specifies the interface of a language-specific SourceGenerator"""
 
     @abstractmethod
-    def __init__(self, out_dir: Path, config: dict):
+    def __init__(self, out_dir: Path, config: dict, templates: dict):
```
bryanwweber:
```suggestion
    def __init__(self, out_dir: Path, config: dict, templates: dict) -> None:
```

interfaces/sourcegen/sourcegen/_SourceGenerator.py
```diff
@@ -14,9 +13,9 @@ class SourceGenerator(metaclass=ABCMeta):
...
         pass
 
     @abstractmethod
-    def generate_source(self, headers_files: List[HeaderFile]):
+    def generate_source(self, headers_files: list[HeaderFile]):
```
bryanwweber:
```suggestion
    def generate_source(self, headers_files: list[HeaderFile]) -> None:
```

interfaces/sourcegen/sourcegen/_TagFileParser.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+@dataclass(frozen=True)
+@with_unpack_iter
+class TagDetails(TagInfo):
+    """Create tag information based on XML data."""
+
+    location: str = ""  #: File containing doxygen description
+    briefdescription: str = ""  #: Brief doxygen description
+    parameterlist: list[Param] = None  #: Annotated doxygen parameter list
```
bryanwweber:
```suggestion
    parameterlist: list[Param] | None = None  #: Annotated doxygen parameter list
```

interfaces/sourcegen/sourcegen/_TagFileParser.py
```diff
@@ -0,0 +1,296 @@
...
+    def __init__(self, bases: dict[str, str]) -> None:
+        tag_file = _TAG_PATH / "Cantera.tag"
+        if not tag_file.exists():
+            msg = (f"Tag file does not exist at expected location:\n    {tag_file}\n"
+                "Run 'scons doxygen' to generate.")
+            _LOGGER.critical(msg)
+            sys.exit(1)
+
+        with tag_file.open() as fid:
+            doxygen_tags = fid.read()
```
bryanwweber:
```suggestion
        doxygen_tags = fid.read_text()
```

interfaces/sourcegen/sourcegen/_TagFileParser.py
```diff
@@ -0,0 +1,296 @@
...
+
+        logging.info("Parsing doxygen tags...")
+        self._parse_doxyfile(doxygen_tags, bases)
+
+    def _parse_doxyfile(self, doxygen_tags: str, bases: list[str]) -> None:
```
bryanwweber:
```suggestion
    def _parse_doxyfile(self, doxygen_tags: str, bases: Sequence[str]) -> None:
```

interfaces/sourcegen/sourcegen/_TagFileParser.py
```diff
@@ -0,0 +1,296 @@
...
+        """Retrieve class and function information from Cantera namespace."""
+
+        def xml_compounds(kind: str, names: list[str]) -> dict[str,str]:
```
bryanwweber:
```suggestion
        def xml_compounds(kind: str, names: Sequence[str]) -> dict[str, str]:
```

interfaces/sourcegen/sourcegen/_TagFileParser.py
```diff
@@ -0,0 +1,296 @@
...
+            regex = re.compile(rf'<compound kind="{kind}"[\s\S]*?</compound>')
+            found = []
+            compounds = {}
+            for compound in re.findall(regex, doxygen_tags):
+                qualified_name = ET.fromstring(compound).find("name").text
+                compound_name = qualified_name.split(":")[-1]
+                if compound_name in names:
+                    found.append(compound_name)
+                    compounds[compound_name] = compound
+                    if not set(names) - set(found):
```
bryanwweber: I don't think this is strictly necessary but IMO improves clarity:
```suggestion
                    if not (set(names) - set(found)):
```

interfaces/sourcegen/sourcegen/_TagFileParser.py
```diff
@@ -0,0 +1,296 @@
...
+            unknown = "', '".join(unknown)
+            _LOGGER.critical("Class(es) in configuration file are missing "
+                             f"from tag file: {unknown!r}")
+            exit(1)
+
+        # Parse content of classes that are specified by the configuration file
+        class_names = set(bases) & set(class_names)
+        classes = xml_compounds("class", class_names)
+
+        def xml_members(kind: str, text: str, prefix="") -> dict[str, str]:
```
bryanwweber:
```suggestion
        def xml_members(kind: str, text: str, prefix: str = "") -> dict[str, str]:
```

interfaces/sourcegen/sourcegen/_TagFileParser.py
```diff
@@ -0,0 +1,296 @@
...
+        self._known = xml_members("function", namespace)
+        for name, cls in classes.items():
+            prefix = f"{name}::"
+            self._known.update(xml_members("function", cls, prefix))
+
+    def exists(self, cxx_func: str) -> bool:
+        """Check whether doxygen tag exists."""
+        return cxx_func in self._known
+
+    def detect(self, name, bases, permissive=True):
```
bryanwweber:
```suggestion
    def detect(self, name: str, bases: Iterable[str], permissive: bool = True) -> str | None:
```

interfaces/sourcegen/sourcegen/_TagFileParser.py
```diff
@@ -0,0 +1,296 @@
...
+def tag_lookup(tag_info: TagInfo) -> TagDetails:
+    """Retrieve tag details from doxygen tree."""
+    xml_file = _XML_PATH / tag_info.anchorfile
+    if not xml_file.exists():
+        msg = (f"XML file does not exist at expected location: {xml_file}")
+        _LOGGER.error(msg)
+        return TagDetails()
+
+    with xml_file.open() as fid:
+        xml_details = fid.read()
```
bryanwweber:
```suggestion
    xml_details = xml_file.read_text()
```

interfaces/sourcegen/sourcegen/_data/README.md
```diff
@@ -0,0 +1,58 @@
...
+The *sourcegen* utility implements logic to automatically detect types CLib functions
+based on a recipe, which is subsequently used to scaffold API functions using dedicated
+*Jinja* templates. The following recipe/CLib function types are differentiated:
+
+- `function`: Regular function defined in the `Cantera` namespace.
+- `constructor`: A CLib constructor adds new (or existing) C++ objects to CLib storage.
+    As all objects are handled via smart `shared_ptr<>`, a CLib constructor requires a
+    C++ utility functions that returning a pointer to a new object or an object that is
+    not yet added to CLib storage. Constructor names should start with `new`.
+- `destructor`: A CLib destructor removes a C++ object from CLib. Constructor names
```
bryanwweber:
```suggestion
- `destructor`: A CLib destructor removes a C++ object from CLib. Destructor names
```

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -1,30 +1,63 @@
...
-    p_type: str
-    name: str = ""
+    description: str = ""  #: Parameter description (optional annotation)
+    direction: str = ""  #: Direction of parameter (optional annotation)
+    default: Any = None  #: Default value (optional)
 
     @classmethod
-    def from_str(cls, param: str) -> 'Param':
-        """Generate Param from string parameter"""
+    def from_str(cls, param: str, doc: str="") -> 'Param':
```
bryanwweber:
```suggestion
    def from_str(cls, param: str, doc: str = "") -> "Param":
```

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -1,30 +1,63 @@
...
+        """Generate Param from parameter string."""
+        param = param.strip()
+        default = None
+        if "=" in param:
+            default = param[param.rfind("=")+1:]
+            param = param[:param.rfind("=")]
```
bryanwweber:
```suggestion
            default, _, param = param.partition("=")
```

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -59,12 +92,28 @@ def from_str(cls, arglist: str) -> 'ArgList':
             return cls([], spec)
         return cls([Param.from_str(arg) for arg in arglist.split(",")], spec)
 
+    @classmethod
+    def from_xml(cls, arglist: str) -> 'ArgList':
```
bryanwweber:
```suggestion
    def from_xml(cls, arglist: str) -> "ArgList":
```

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -59,12 +92,28 @@ def from_str(cls, arglist: str) -> 'ArgList':
...
+        """Generate ArgList from XML string argument list."""
+        arglist, spec = cls._split_arglist(arglist)
+        if not arglist:
+            return cls([], spec)
+        return cls([Param.from_xml(arg) for arg in arglist.split(",")], spec)
+
+    def __len__(self):
+        return len(self.params)
+
     def __getitem__(self, k):
```
bryanwweber:
```suggestion
    def __getitem__(self, k: int) -> Param:
```

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -59,12 +92,28 @@ def from_str(cls, arglist: str) -> 'ArgList':
...
         return self.params[k]
 
     def __iter__(self) -> "Iterator[Param]":
```
bryanwweber:
```suggestion
    def __iter__(self) -> Iterator[Param]:
```

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -59,12 +92,28 @@ def from_str(cls, arglist: str) -> 'ArgList':
...
         return iter(self.params)
 
+    def short_str(self) -> str:
+        """String representation of the argument list without parameter names."""
+        args = ', '.join([par.short_str() for par in self.params])
```
bryanwweber: I don't think you need to construct a list here:
```suggestion
        args = ', '.join(par.short_str() for par in self.params)
```

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -99,8 +148,88 @@ def declaration(self) -> str:
 
 @dataclass(frozen=True)
 @with_unpack_iter
-class HeaderFile:
-    """Represents information about a parsed C header file"""
+class CFunc(Func):
+    """Represents an annotated function declaration in a C/C++ header file."""
+
+    brief: str = ""  #: Brief description (optional)
+    implements: 'CFunc' = None  #: Implemented C++ function/method (optional)
```
bryanwweber:
```suggestion
    implements: "CFunc" = None  #: Implemented C++ function/method (optional)
```

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -99,8 +148,88 @@ def declaration(self) -> str:
...
+    returns: str = ""  #: Description of returned value (optional)
+    base: str = ""  #: Qualified scope of function/method (optional)
+    uses: list['CFunc'] = None  #: List of auxiliary C++ methods (optional)
```
bryanwweber:
```suggestion
    uses: list["CFunc"] | None = None  #: List of auxiliary C++ methods (optional)
```

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -99,8 +148,88 @@ def declaration(self) -> str:
...
+
+    @classmethod
+    def from_str(cls, func: str, brief="") -> 'CFunc':
```
bryanwweber:
```suggestion
    def from_str(cls, func: str, brief: str = "") -> "CFunc":
```

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -99,8 +148,88 @@ def declaration(self) -> str:
...
+        """Generate annotated CFunc from header block of a function."""
+        lines = func.split("\n")
+        func = super().from_str(lines[-1])
+        if len(lines) == 1:
+            return func
+        brief = ""
```
bryanwweber: This overwrites the passed in value of `brief`.

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -99,8 +148,88 @@ def declaration(self) -> str:
...
+
+    def short_declaration(self) -> str:
+        """Return a short string representation."""
+        ret = (f"{self.name}{self.arglist.short_str()}").strip()
+        if self.base:
+            return f"{self.ret_type} {self.base}::{ret}"
+        return f"{self.ret_type} {ret}"
+
+    @property
+    def ret_param(self):
```
bryanwweber:
```suggestion
    def ret_param(self) -> Param:
```

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -99,8 +148,88 @@ def declaration(self) -> str:
...
+
+    path: Path  #: output folder
+    funcs: list[Func]  #: list of functions to be scaffolded
+
+    prefix: str = ""  #: prefix used for CLib function names
+    base: str = ""  #: base class of C++ methods (if applicable)
+    parents: list[str] = None  #: list of C++ parent class(es)
+    derived: list[str] = None  #: list of C++ specialization(s)
+    recipes: list[Recipe] = None  #: list of header recipes read from YAML
+    docstring: list[str] = None  #: lines representing docstring of YAML file
```
bryanwweber:
```suggestion
    parents: list[str] | None = None  #: list of C++ parent class(es)
    derived: list[str] | None = None  #: list of C++ specialization(s)
    recipes: list[Recipe] | None = None  #: list of header recipes read from YAML
    docstring: list[str] | None = None  #: lines representing docstring of YAML file
```

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -99,8 +148,88 @@ def declaration(self) -> str:
...
+
+    def output_name(self, auto="3", suffix=""):
```
bryanwweber:
```suggestion
    def output_name(self, auto: str = "3", suffix: str = "") -> Path:
```

interfaces/sourcegen/sourcegen/_orchestrate.py
```diff
@@ -23,28 +23,46 @@ def format(self, record):
         return formatter.format(record)
 
 
-def generate_source(lang: str, out_dir: str=""):
+def generate_source(lang: str, out_dir: str=None, verbose=False):
```
bryanwweber:
```suggestion
def generate_source(lang: str, out_dir: str | None = None, verbose: bool = False) -> None:
```

interfaces/sourcegen/sourcegen/_orchestrate.py
```diff
@@ -23,28 +23,46 @@ def format(self, record):
...
     config = read_config(root / "config.yaml")
     templates = read_config(root / "templates.yaml")
-    ignore_files: List[str] = config.pop("ignore_files", [])
-    ignore_funcs: Dict[str, List[str]] = config.pop("ignore_funcs", {})
+    ignore_files: list[str] = config.pop("ignore_files", [])
+    ignore_funcs: dict[str, list[str]] = config.pop("ignore_funcs", {})
 
-    files = HeaderFileParser.from_headers(ignore_files, ignore_funcs)
+    if lang == 'clib':
+        files = HeaderFileParser.headers_from_yaml(ignore_files, ignore_funcs)
```
bryanwweber: I don't see a difference between this clause and the `else` clause. Is there meant to be?

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+from .._SourceGenerator import SourceGenerator
+from .._TagFileParser import TagFileParser
+
+
+_LOGGER = logging.getLogger()
+
+class CLibSourceGenerator(SourceGenerator):
+    """The SourceGenerator for generating CLib."""
+
+    def __init__(self, out_dir: str, config: dict, templates: dict):
```
bryanwweber:
```suggestion
    def __init__(self, out_dir: str, config: dict, templates: dict) -> None:
```

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+        self._out_dir = out_dir or None
+        if self._out_dir is not None:
+            self._out_dir = Path(out_dir)
+            self._out_dir.mkdir(parents=True, exist_ok=True)
+        self._config = Config.from_parsed(**config)  # typed config
```
bryanwweber:
```suggestion
        self._config = Config.from_parsed(**config)
```

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+        self._templates = templates
+        self._doxygen_tags = None
+
+    @staticmethod
+    def _javadoc_comment(block):
```
bryanwweber:
```suggestion
    def _javadoc_comment(block: str) -> str:
```

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+        block = "\n *  ".join(block).strip() + "\n */"
+        return "\n".join([line.rstrip() for line in block.split('\n')])
+
+    def _scaffold_annotation(self, c_func: CFunc, what: str) -> str:
+        """Build annotation block via Jinja."""
+        loader = Environment(loader=BaseLoader, trim_blocks=True, lstrip_blocks=True)
+        par_template = loader.from_string(self._templates["clib-param"])
+        template = loader.from_string(self._templates["clib-comment"])
+
+        def param(item: Param):
```
bryanwweber:
```suggestion
        def param(item: Param) -> str:
```

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+    @staticmethod
+    def _reverse_crosswalk(c_func: CFunc, base: str) -> tuple[dict[str, str], set[str]]:
+        """Translate CLib arguments back to Jinja argument list."""
+        handle = ""
+        args = []
+        lines = []
+        buffer = []
+        bases = set()
+
+        def shared_object(cxx_type):
```
bryanwweber:
```suggestion
        def shared_object(cxx_type: str) -> str:
```

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+        }
+        return ret, bases
+
+    def _scaffold_body(self, c_func: CFunc, recipe: Recipe) -> tuple[str, set[str]]:
+        """Scaffold body of generic CLib function via Jinja."""
+        loader = Environment(loader=BaseLoader, trim_blocks=True, lstrip_blocks=True)
+        args, bases = self._reverse_crosswalk(c_func, recipe.base)
+        args["what"] = recipe.what
+
+        if recipe.what == "noop":
```
bryanwweber: Does it make sense to create a `StrEnum` for this type, so we can make sure they've all been enumerated here?

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+
+        else:
+            _LOGGER.critical(f"{recipe.what!r} not implemented: {c_func.name!r}.")
+            exit(1)
+
+        return template.render(**args), bases
+
+    def _resolve_recipe(self, recipe: Recipe, quiet: bool=True) -> CFunc:
+        """Build CLib header from recipe and doxygen annotations."""
+        def merge_params(implements, cxx_func: CFunc) -> tuple[list[Param], int]:
```
bryanwweber:
```suggestion
        def merge_params(implements: str, cxx_func: CFunc) -> tuple[list[Param], int]:
```

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+            template = loader.from_string(
+                self._templates[f"clib-reserved-{recipe.name}-cpp"])
+
+        else:
+            _LOGGER.critical(f"{recipe.what!r} not implemented: {c_func.name!r}.")
+            exit(1)
+
+        return template.render(**args), bases
+
+    def _resolve_recipe(self, recipe: Recipe, quiet: bool=True) -> CFunc:
```
bryanwweber:
```suggestion
    def _resolve_recipe(self, recipe: Recipe, quiet: bool = True) -> CFunc:
```

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+            prop_params = self._prop_crosswalk(par_list)
+            brief = cxx_func.brief
+            args = prop_params + buffer_params
+
+        if cxx_func and not recipe.what:
+            # Autodetection of CLib function purpose ("what")
+            cxx_arglen = len(cxx_func.arglist)
+            if not cxx_func.base:
+                if cxx_func.name.startswith("new") and \
+                    any(base in cxx_func.ret_type
```
bryanwweber: It's a little unusual here to use a backslash continuation IMO. I think the more typical style would be to put the conditions in parentheses.

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+        template = loader.from_string(self._templates["clib-header-file"])
+        output = template.render(
+            name=filename.stem, guard=guard, declarations=declarations,
+            prefix=headers.prefix, base=headers.base, docstring=headers.docstring)
+
+        out = (Path(self._out_dir) /
+               "include" / "cantera" / "clib_experimental" / filename.name)
+        _LOGGER.info(f"  writing {filename.name!r}")
+        if not out.parent.exists():
+            out.parent.mkdir(parents=True, exist_ok=True)
```
bryanwweber:
```suggestion
        out.parent.mkdir(parents=True, exist_ok=True)
```

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+        with open(out, "wt", encoding="utf-8") as stream:
+            stream.write(output)
+            stream.write("\n")
```
bryanwweber:
```suggestion
        out.write_text(output + "\n")
```

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+
+    def _write_implementation(self, headers: HeaderFile) -> None:
```
bryanwweber, line 495: The plural for `headers` and the singular for the type is a little confusing here :smile: Is the type meant to be a `list[HeaderFile]`?

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+            prefix=headers.prefix, base=headers.base, docstring=headers.docstring,
+            includes=includes, other=other, str_utils=str_utils)
+
+        out = Path(self._out_dir) / "src" / "clib_experimental" / filename.name
+        _LOGGER.info(f"  writing {filename.name!r}")
+        if not out.parent.exists():
+            out.parent.mkdir(parents=True, exist_ok=True)
+        with open(out, "wt", encoding="utf-8") as stream:
+            stream.write(output)
+            stream.write("\n")
```
bryanwweber:
```suggestion
        out.parent.mkdir(parents=True, exist_ok=True)
        out.write_text(output + "\n")
```

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+
+    def resolve_tags(self, headers_files: list[HeaderFile], quiet: bool=True):
```
bryanwweber:
```suggestion
    def resolve_tags(self, headers_files: list[HeaderFile], quiet: bool = True) -> None:
```

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,529 @@
...
+
+        for headers in headers_files:
+            if not quiet:
+                _LOGGER.info(f"  resolving recipes in {headers.path.name!r}:")
+            c_funcs = []
+            for recipe in headers.recipes:
+                c_funcs.append(self._resolve_recipe(recipe, quiet=quiet))
+            headers.funcs = c_funcs
+
+    def generate_source(self, headers_files: list[HeaderFile]):
```
bryanwweber:
```suggestion
    def generate_source(self, headers_files: list[HeaderFile]) -> None:
```

interfaces/sourcegen/sourcegen/clib/_Config.py
```diff
@@ -0,0 +1,45 @@
...
+        "double *const": "double*",
+        "const double *const": "const double*",
+        "const string&": "const char*",
+        "shared_ptr<T>": "int",
+        "const shared_ptr<T>": "int",
+        "const vector<double>&": 'const double*',
+        "const vector<shared_ptr<T>>&": 'int[]',
+    }
+
+    includes: dict[str,list[str]]
```
bryanwweber:
```suggestion
    includes: dict[str, list[str]]
```

interfaces/sourcegen/sourcegen/clib/_Config.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @classmethod
+    def from_parsed(cls, *, includes=None) -> 'Config':
```
bryanwweber:
```suggestion
    def from_parsed(cls: Type[_Self], *, includes: dict[str, list[str]] | None = None) -> _Self:
```
and then somewhere above the class definition:
```
_Self = TypeVar("_Self", bound="Config")
```
I think this should be done for all the `classmethod`s in the PR. This (should) allows type checkers to correctly infer the type of a subclass.

The `_Self` type can be replaced by `typing.Self` when we support Python >=3.11 or we can add a dependency on `typing_extensions` and use the `Self` type now, rather than defining the `_Self` type. See: https://stackoverflow.com/a/75337086/2449192

interfaces/sourcegen/sourcegen/csharp/_CSharpSourceGenerator.py
```diff
@@ -21,8 +20,18 @@
 class CSharpSourceGenerator(SourceGenerator):
     """The SourceGenerator for scaffolding C# files for the .NET interface"""
 
+    def __init__(self, out_dir: str, config: dict, templates: dict):
```
bryanwweber:
```suggestion
    def __init__(self, out_dir: str, config: dict, templates: dict) -> None:
```

interfaces/sourcegen/sourcegen/csharp/_CSharpSourceGenerator.py
```diff
@@ -160,7 +159,7 @@ def _write_file(self, file_name: str, template_name: str, **kwargs) -> None:
 
         self._out_dir.joinpath(file_name).write_text(contents, encoding="utf-8")
 
-    def _scaffold_interop(self, header_file_path: Path, cs_funcs: List[CsFunc]):
+    def _scaffold_interop(self, header_file_path: Path, cs_funcs: list[CsFunc]):
```
bryanwweber:
```suggestion
    def _scaffold_interop(self, header_file_path: Path, cs_funcs: list[CsFunc]) -> None:
```

interfaces/sourcegen/sourcegen/csharp/_CSharpSourceGenerator.py
```diff
@@ -170,7 +169,7 @@ def _scaffold_interop(self, header_file_path: Path, cs_funcs: List[CsFunc]):
         self._write_file(
             file_name, "csharp-scaffold-interop", cs_functions=function_list)
 
-    def _scaffold_handles(self, header_file_path: Path, handles: Dict[str, str]):
+    def _scaffold_handles(self, header_file_path: Path, handles: dict[str, str]):
```
bryanwweber:
```suggestion
    def _scaffold_handles(self, header_file_path: Path, handles: dict[str, str]) -> None:
```

interfaces/sourcegen/sourcegen/csharp/_CSharpSourceGenerator.py
```diff
@@ -190,8 +189,8 @@ def _scaffold_derived_handles(self):
         self._write_file(
             file_name, "csharp-scaffold-handles", cs_handles=handle_list)
 
-    def _scaffold_wrapper_class(self, clib_area: str, props: Dict[str, str],
-                                known_funcs: Dict[str, CsFunc]):
+    def _scaffold_wrapper_class(self, clib_area: str, props: dict[str, str],
+                                known_funcs: dict[str, CsFunc]):
```
bryanwweber:
```suggestion
                                known_funcs: dict[str, CsFunc]) -> None:
```

interfaces/sourcegen/sourcegen/csharp/_CSharpSourceGenerator.py
```diff
@@ -204,10 +203,10 @@ def _scaffold_wrapper_class(self, clib_area: str, props: Dict[str, str],
             wrapper_class_name=wrapper_class_name, handle_class_name=handle_class_name,
             cs_properties=property_list)
 
-    def generate_source(self, headers_files: List[HeaderFile]):
+    def generate_source(self, headers_files: list[HeaderFile]):
```
bryanwweber:
```suggestion
    def generate_source(self, headers_files: list[HeaderFile]) -> None:
```

interfaces/sourcegen/sourcegen/yaml/_YamlSourceGenerator.py
```diff
@@ -0,0 +1,66 @@
...
+from .._dataclasses import HeaderFile, CFunc
+from .._SourceGenerator import SourceGenerator
+
+
+_LOGGER = logging.getLogger()
+
+class YamlSourceGenerator(SourceGenerator):
+    """The SourceGenerator for generating CLib."""
+
+    def __init__(self, out_dir: str, config: dict, templates: dict):
```
bryanwweber:
```suggestion
    def __init__(self, out_dir: str, config: dict, templates: dict) -> None:
```

interfaces/sourcegen/sourcegen/yaml/_YamlSourceGenerator.py
```diff
@@ -0,0 +1,66 @@
...
+        self._out_dir = out_dir or None
+        if self._out_dir is not None:
+            self._out_dir = Path(out_dir)
+            self._out_dir.mkdir(parents=True, exist_ok=True)
+        self._config = config
+        self._templates = templates
+
+    def _write_yaml(self, headers: HeaderFile):
```
bryanwweber:
```suggestion
    def _write_yaml(self, headers: HeaderFile) -> None:
```
Similar here about the plural consistency.

interfaces/sourcegen/sourcegen/yaml/_YamlSourceGenerator.py
```diff
@@ -0,0 +1,66 @@
...
+        template = loader.from_string(self._templates["yaml-file"])
+        output = template.render(filename=filename.name, header_entries=declarations)
+
+        out = Path(self._out_dir) / "yaml" / filename.name
+        _LOGGER.info(f"  writing {filename.name!r}")
+        if not out.parent.exists():
+            out.parent.mkdir(parents=True, exist_ok=True)
+        with open(out, "wt", encoding="utf-8") as stream:
+            stream.write(output)
+            stream.write("\n")
```
bryanwweber:
```suggestion
        out.parent.mkdir(parents=True, exist_ok=True)
        out.write_text(output + "\n")
```

interfaces/sourcegen/sourcegen/yaml/_YamlSourceGenerator.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def generate_source(self, headers_files: list[HeaderFile]):
```
bryanwweber:
```suggestion
    def generate_source(self, headers_files: list[HeaderFile]) -> None:
```

test/clib_experimental/test_clib3.cpp
```diff
@@ -0,0 +1,309 @@
...
+    ASSERT_EQ(nsp, ERR);
+    string err = reportError();
+    EXPECT_THAT(err, HasSubstr("has been deleted."));
+
+    nsp = thermo3_nSpecies(thermo2);
+    ASSERT_EQ(nsp, ERR);
+    err = reportError();
+    EXPECT_THAT(err, HasSubstr("has been deleted."));
+
+    // trans2 = sol3_setTransportModel(ref, "mixture-averaged");
```
bryanwweber: Are these tests meant to be commented out? If so, probably better to just clean them up :smile:

bryanwweber (commented):
Thanks for all this work Ingmar! Nearly all of my comments here are suggestions to improve the type hinting. In some cases, you'll need to add an import (I think just for `Iterable`). Obviously these don't affect the runtime, but my thought was that if we're going to have type hints they should at least be complete :smile:

interfaces/sourcegen/sourcegen/_orchestrate.py
ischoegl: I clarified this as there is a subtle difference: for other languages, CLib is generated in a preprocessing step, whereas for the main `if` clause it is the main processing step. I added the YAML generator to test/illustrate this.

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
ischoegl: I didn't know about `StrEnum`, but it appears to be Python 3.11. I overall don't think it is necessary as generated C++ won't compile (or CLib won't work as expected).

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
ischoegl: This terminology was chosen by the original author of *sourcegen*. I started out with singular, but reverted `header` to `headers` in one of the commits.

test/clib_experimental/test_clib3.cpp
ischoegl: I added most of the missing conversions and 'recipes' to test this more completely. Now, only a couple of commented passages are left, which should be uncommented as the experimental CLib reaches/surpasses the status of the traditional CLib (i.e., the commented blocks remain to be addressed in follow-up work).

interfaces/sourcegen/sourcegen/clib/_Config.py
ischoegl: While I have adopted all other suggestions, I'd prefer to wait for `typing.Self`, once we drop support for 3.10. Personally, I have never managed to really warm up to `TypeVar`.

_Edit: I ended up using `from typing_extensions import Self`, as it is imho the best compromise._

interfaces/sourcegen/sourcegen/_HeaderFileParser.py
ischoegl: While I believe we dropped the preference for `"` in Python strings, I manually searched for `'` and replaced what I found.

ischoegl:
> Thanks for all this work Ingmar! Nearly all of my comments here are suggestions to improve the type hinting. In some cases, you'll need to add an import (I think just for `Iterable`). Obviously these don't affect the runtime, but my thought was that if we're going to have type hints they should at least be complete 😄

Thanks, @bryanwweber, for the prompt review! I agree that having done most of the type hints already, it makes sense to take it all the way. I adopted almost all of the suggestions directly and added some conversions, so the `googletest` suite is almost fully comparable to the traditional CLib version for the parts that I have converted..

ischoegl:
@bryanwweber and @speth ... the PR ready for further comments/reviews. _PS: had to fix some unexpected CI failures. / PPS: decided to include support for custom code to be nearly feature-complete._

bryanwweber (dismissed):
Thanks Ingmar! The Python looks good to me :smile: I'm not that familiar with the overall use case though as I've been out of the loop with these changes, so I assume that it works the way we want and anyways we're early in the 3.2 cycle so we can always fix it :smiley:

ischoegl:
> Thanks Ingmar! The Python looks good to me 😄 I'm not that familiar with the overall use case though as I've been out of the loop with these changes, so I assume that it works the way we want and anyways we're early in the 3.2 cycle so we can always fix it 😃

Thanks, @bryanwweber! The overall use case is Cantera/enhancements#39 (specifically, a drop-in replacement for the existing `clib` that won't require manual coding). I'll wait for @speth to chime in.

interfaces/sourcegen/sourcegen/_data/ctsol_auto.yaml
```diff
@@ -0,0 +1,43 @@
...
+  code: |-
+    try {
+        auto obj = SolutionCabinet::at(handle);
+        TransportCabinet::del(
+            TransportCabinet::index(*(obj->transport()), handle));
+        obj->setTransportModel(model);
+        return TransportCabinet::add(obj->transport(), handle);
+    } catch (...) {
+        return handleAllExceptions(-2, ERR);
+    }
```
ischoegl, line 36: On second thought, the custom code section I added here should be moved from this main configuration file to `clib/config.yaml` as it is an implementation detail. This isn't difficult to change, but I'll hold off until there are other review comments.

doc/doxygen/Doxyfile
```diff
@@ -1013,7 +1013,8 @@ RECURSIVE              = YES
 # Note that relative paths are relative to the directory from which doxygen is
 # run.
 
-EXCLUDE                = include/cantera/ext
+EXCLUDE                = include/cantera/ext \
+                         include/cantera/clib_experimental
```
speth, line 1018: Ultimately, I assume we will want to include the clib docs in the Doxygen output, which will create a sort of interesting requirement for the SCons build process -- we will need to run Doxygen once to generate the XML output, then again after running sourcegen.

interfaces/sourcegen/sourcegen/yaml/_YamlSourceGenerator.py
```diff
@@ -0,0 +1,66 @@
...
+from jinja2 import Environment, BaseLoader
+
+from .._dataclasses import HeaderFile, CFunc
+from .._SourceGenerator import SourceGenerator
+
+
+_LOGGER = logging.getLogger()
+
+class YamlSourceGenerator(SourceGenerator):
+    """The SourceGenerator for generating CLib."""
```
speth: This docstring is incorrect, right? I'm a little confused as to how / whether this generator is used.

interfaces/sourcegen/sourcegen/_data/ctsol_auto.yaml
ischoegl: It’s been a while since this all posted. As the PR is approved, I’ll merge and address this in a follow up.

ischoegl: @speth ... I saw your comment and will hold off on the merge.

interfaces/sourcegen/sourcegen/_data/README.md
```diff
@@ -0,0 +1,58 @@
+# Automatic Code Generation (Experimental)
+
+This folder holds YAML configuration data for code generation. Code generation uses
+output of the *doxygen* utility for information on function declarations as well as
+documentation. The collected information represents annotated CLib functions that form
+the basis for specific APIs generated by *sourcegen*.
```
speth, line 6: I'd recommend moving the contents of this file (and probably the `README.md` in the parent `sourcegen` directory as well) to a page in `doc/sphinx/develop` so we can have this as part of the HTML documentation and information on how Cantera works in one place.

interfaces/sourcegen/sourcegen/_data/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+- `prefix`: Prefix used for CLib functions.
+- `base`: Base class of the CLib library; examples: `Solution`, `ThermoPhase`.
+    Stand-alone functions defined in the `Cantera` namespace use `""` as a base.
+- `parents`: List of class parents; used to locate C++ methods (default: `[]`).
+- `derived`: List of derived classes; used to locate C++ methods (default: `[]`).
+- `recipes`: List of CLib *recipes* (see below).
+
+In addition, the description of the YAML file (commented block at top of file) may be
+used as part of the docstring of auto-generated API files.
```
speth: This behavior is a bit surprising. I think it would be preferable to provide any additional content that we want to include in the generated header as a specific YAML field.

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -99,8 +148,87 @@ def declaration(self) -> str:
...
+
+    prefix: str = ""  #: prefix used for CLib function names
+    base: str = ""  #: base class of C++ methods (if applicable)
+    parents: list[str] | None = None  #: list of C++ parent class(es)
+    derived: list[str] | None = None  #: list of C++ specialization(s)
+    recipes: list[Recipe] | None = None  #: list of header recipes read from YAML
+    docstring: list[str] | None = None  #: lines representing docstring of YAML file
+
+    def output_name(self, auto: str = "3", suffix: str = "") -> Path:
+        """Return updated path."""
```
speth: What does the `auto=3` argument mean? I assume this is part of how the output files get named `ctfunc3.h` etc., but it's a little confusing.

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,563 @@
...
+            includes=includes, other=other, str_utils=str_utils)
+
+        out = Path(self._out_dir) / "src" / "clib_experimental" / filename.name
+        msg = f"  writing {filename.name!r}"
+        _LOGGER.info(msg)
+        out.parent.mkdir(parents=True, exist_ok=True)
+        out.write_text(output + "\n")
+
+    def resolve_tags(self, headers_files: list[HeaderFile]) -> None:
+        """Resolve doxygen tags."""
```
speth: What does it mean to "resolve" a Doxygen tag?

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
```diff
@@ -0,0 +1,563 @@
...
+        for obj in [headers.base] + list(other):
+            includes += self._config.includes[obj]
+
+        template = loader.from_string(self._templates["clib-source-file"])
+        output = template.render(
+            name=filename.stem, implementations=implementations,
+            prefix=headers.prefix, base=headers.base, docstring=headers.docstring,
+            includes=includes, other=other, str_utils=str_utils)
+
+        out = Path(self._out_dir) / "src" / "clib_experimental" / filename.name
```
speth, line 529: Should this output directory be a config variable somewhere, rather then embedded in the generator code?

interfaces/sourcegen/sourcegen/clib/_Config.py
```diff
@@ -0,0 +1,47 @@
...
+        "const vector<double>&": "const double*",
+        "const vector<shared_ptr<T>>&": "int[]",
+    }
+
+    includes: dict[str, list[str]]
+
+    @classmethod
+    def from_parsed(cls: Self, *, includes: dict[str, list[str]] | None = None) -> Self:
+        """Ensure that configurations are correct."""
+        return cls(includes or {})
```
speth: Is this docstring correct? It seems like this is doing more than just checking the config.

interfaces/sourcegen/sourcegen/clib/templates.yaml
```diff
@@ -0,0 +1,368 @@
...
+clib-array-getter: |-
+  // getter: {{ cxx_implements }}
+  try {
+      auto& obj = {{ cxx_base }}Cabinet::at({{ handle }});
+      {% if uses %}
+      if ({{ c_args[1] }} != obj->{{ uses[0][1] }}()) {
+          throw CanteraError("{{ c_func }}",
+              "Invalid output array size; expected size {} but received {}.",
+              obj->{{ uses[0][1] }}(), {{ c_args[1] }});
+      }
```
speth: For the array size checks, I think it would be better to maintain the existing behavior of requiring just that the available array size is at least large enough, with a larger array not being considered an error. This makes it easier to use in cases where the user's array sizes might be hard coded.

interfaces/sourcegen/sourcegen/_HeaderFileParser.py
```diff
@@ -1,33 +1,102 @@
...
+        ) -> list[HeaderFile]:
+        """Parse header file YAML configuration."""
+        files = sorted(
+            ff for ff in _DATA_PATH.glob("*.yaml") if ff.name not in ignore_files)
+        return [cls(ff, ignore_funcs.get(ff.name, []))._parse_yaml() for ff in files]
+
+    def _parse_yaml(self) -> HeaderFile:
+        def read_docstring():
+            doc = []
+            with self._path.open("r", encoding="utf-8") as fid:
```
speth: Is there a reason not to just build `doc` as a local variable within `_parse_yaml` and eliminate the nested function?

interfaces/sourcegen/sourcegen/_orchestrate.py
```diff
@@ -23,28 +23,51 @@ def format(self, record):
...
-    _logger.addHandler(loghandler)
-    _logger.setLevel(logging.DEBUG)
-    _logger.info(f"Generating {lang!r} source files...")
+    _LOGGER.handlers.clear()
+    _LOGGER.addHandler(loghandler)
+    _LOGGER.setLevel(logging.DEBUG if verbose else logging.INFO)
+
+    if not out_dir:
+        _LOGGER.critical("Aborting: sourcegen requires output folder information.")
+        exit(1)
```
speth, line 36: Instead of this, you could just remove the default value for the `out_dir` argument, right?

interfaces/sourcegen/sourcegen/_orchestrate.py
```diff
@@ -23,28 +23,51 @@ def format(self, record):
...
 
-    files = HeaderFileParser.from_headers(ignore_files, ignore_funcs)
+    if lang == "clib":
+        # prepare for generation of CLib headers in main processing step
+        files = HeaderFileParser.headers_from_yaml(ignore_files, ignore_funcs)
+    elif lang == "csharp":
+        # csharp parses existing (traditional) CLib header files
+        files = HeaderFileParser.headers_from_h(ignore_files, ignore_funcs)
+    else:
+        # generate CLib headers from YAML specifications as a preprocessing step
```
speth, line 55: I think this ought to be `elif lang == "yaml"` and a final `else` that indicates an error rather than silently defaulting to generating YAML output.

SConstruct
```diff
@@ -364,6 +368,10 @@ config_options = [
         "sphinx_docs",
         "Build HTML documentation for Cantera using Sphinx.",
         False),
+    BoolOption(
+        "clib_experimental",
+        "Build experimental CLib.",
+        False),
```
speth: Probably worth noting here that for this to work, you need to separately run both `scons doxygen` and `sourcegen` first.

speth (changes requested):
Thanks for taking this on, @ischoegl. I'm very excited to see everything that you've got working here, and I'm looking forward to the opportunity to continue expanding this code generation beyond `clib`.

One question I have is on the logistics of replacing the existing manual `clib`. Are you expecting that we will be able to complete the transition before the next Cantera release, such that we can avoid the need for the parallel set of functions and header files with the "3" suffix?

doc/doxygen/Doxyfile
ischoegl: This is correct. Also, for the first run, we don't need to generate HTML, which should speed things up a little.

interfaces/sourcegen/sourcegen/_HeaderFileParser.py
ischoegl: This is moot - based on another comment, I created a new field `docstring` so this is no longer necessary.

interfaces/sourcegen/sourcegen/_orchestrate.py
ischoegl: As we may load `sourcegen` as a module rather than run it from the command line at a later point, I'd prefer to leave this as is.

interfaces/sourcegen/sourcegen/_orchestrate.py
ischoegl: The `YamlSourceGenerator` is only a stand-in for other APIs to be generated later. Instead of parsing CLib headers as currently done by the .NET API, the information can be obtained directly from the YAML configuration as a preprocessing step. This ensures that all information is consistent. (The `csharp` branch should be eliminated when the approach taken there changes.)

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
ischoegl: It definitely could be; for the time being, I wanted to not over-engineer as we don't have a final `scons` process yet. It's implicitly covered by Cantera/enhancements#220.

interfaces/sourcegen/sourcegen/_dataclasses.py
ischoegl: I dropped the `auto` and made it part of `suffix`, while also adding substance to the docstring.

interfaces/sourcegen/sourcegen/_data/README.md
ischoegl: I'd like to hold off on this until this is more developed. I will add this to the roadmap Cantera/enhancements#220

interfaces/sourcegen/sourcegen/_data/ctsol_auto.yaml
ischoegl: As mentioned above, I will move the 'custom code' portion in a future PR.

ischoegl:
> [...] I'm looking forward to the opportunity to continue expanding this code generation beyond `clib`.

Yes, I believe this can be leveraged in the near future (especially for MATLAB, as the hand-coded part covers only a portion of the actual API). Fortran could be done in analogy to CLib (without depending on it).

> One question I have is on the logistics of replacing the existing manual `clib`. Are you expecting that we will be able to complete the transition before the next Cantera release, such that we can avoid the need for the parallel set of functions and header files with the "3" suffix?

I had already created a tracking issue for this: Cantera/enhancements#220. For now, the "3"s are needed; a future removal is imho trivial. I can likely handle the transition, but am not planning on investing in code generation for other APIs.

interfaces/sourcegen/sourcegen/_data/README.md
speth: OK, I think that's all right for now.

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
speth: Ultimately, I guess it's fine for it to be hardcoded. I'm just thinking about how to avoid confusion when we make this non-experimental and are wondering why the files keep appearing in the `clib_experimental` folder.

interfaces/sourcegen/sourcegen/_orchestrate.py
speth: Why does it matter whether this is imported as a module versus run from the command line? This function is:
```py
def generate_source(lang: str, out_dir: str = "", verbose = False) -> None:
    ...
    if not out_dir:
        _LOGGER.critical("Aborting: sourcegen requires output folder information.")
        exit(1)
```
What's the benefit of having a default value if that default value is considered an error?

interfaces/sourcegen/sourcegen/clib/_CLibSourceGenerator.py
ischoegl: I think at that point we'll redirect to `clib`.

interfaces/sourcegen/sourcegen/_orchestrate.py
ischoegl: Ah, ok. The `run.py` coming in has the empty string `""` as default, so the logger will check that it is set by the user. I can absolutely remove the default argument, but would prefer for the logger to remain.

ischoegl:
@speth ... please let me know in case there's anything else. At this point, I have taken care of all review comments.

interfaces/sourcegen/sourcegen/_orchestrate.py
speth: Right, I understand why `csharp` is a special case until we change it to rely on the Doxygen/YAML input. What I'm less clear on is why running any target other than `clib` requires first generating the CLib source -- while you might need those headers to compile another interface, it's not a requirement to run `sourcegen`, is it? Why doesn't this all just simplify to:
```py
    if lang == "csharp":
        # csharp parses existing (traditional) CLib header files
        files = HeaderFileParser.headers_from_h(ignore_files, ignore_funcs)
    else:
        files = HeaderFileParser.headers_from_yaml(ignore_files, ignore_funcs)
```

ischoegl: The CLib parser is a special case, as it populates `HeaderFile` objects with information from the C++ API as well as the translation to CLib headers, regardless of interfaces. At the same time, `ignore_files`, `ignore_funcs`, etc. should be specific to the actual `lang` input - If you look at the YAML pathway I included for illustration purposes, it does not share configurations with CLib.

While I specifically created the YAML output generator as a stub to illustrate, there are probably other solutions out there, but they are not essential for the current focus on CLib.

ischoegl: @speth ... thinking about this a little more, the best approach would be to be more explicit about splitting the preprocessing step (i.e., the `HeaderFileParser`) from the CLib generator. That way, we could use the `if` - `else` block you mentioned above and clarify some of the logic. Given the size of this PR, I'd prefer to investigate/discuss in a follow-up PR. A good place for discussion would be Cantera/enhancements#220 so it won't get lost in comments?

speth (approved):

Issue #989, opened 2021-03-07
Numeric solver runaway using ReactorNet.advance_to_steady_state


philipp1337:
I want to simulate burning methane in an IdealGasConstPressureReactor. I use different mass-flows and equivalence-ratios using two vectors you can see in my example code below. However, using ReactorNet.advance_to_steady_state breaks the solver at certain points, leading to huge reactor masses and volumes and therefore CanteraErrors. If I varying the equivalence-ratio leads to different combustors failing, varying the mass-flows to each reactor has no effect on this. Using a ConstPressureReactor also leads to the same results, using an IdealGasReactor eliminates the error while the pressure before and after the reactor stays about the same (0.1% differences).

**Steps to reproduce**

My example code:

```python
import cantera as ct
import numpy as np

# Initial constants
inlet_temperature = 712 # K
inlet_pressure = 2171756.0738435777 # Pa
inlet_density = 10.529 # kg/m³
phi_mean = 0.3
phi_std = 0.25
mdot = 350 # kg/s

# Vectors for phi and mass flow ratio
x=[0.1125, 0.1375, 0.1625, 0.1875, 0.2125, 0.2375, 0.2625, 0.2875, 0.3125, 0.3375,
 0.3625, 0.3875, 0.4125, 0.4375, 0.4625, 0.4875]
y_norm=[0.05223705, 0.05602466, 0.05948904, 0.0625391,  0.06509137, 0.0670737,
 0.06842868, 0.0691164,  0.0691164,  0.06842868, 0.0670737,  0.06509137,
 0.0625391,  0.05948904, 0.05602466, 0.05223705]

psr_exhaust_gases = []


# Run Combustors for each Massflow/Phi
n = 0
for phi, ratio in zip(x,y_norm):
    gas = ct.Solution("gri30.xml")
    gas.TP = inlet_temperature, inlet_pressure
    print(f"{n} Combustor using phi={phi} and mass flow {ratio*mdot}")
    gas.set_equivalence_ratio(phi, 'CH4:1.0', 'O2:1.0, N2:3.76')
    gas.equilibrate("HP")
    inlet = ct.Reservoir(gas)
    exhaust = ct.Reservoir(gas)
    reactor = ct.IdealGasConstPressureReactor(gas)
    reactor.volume = 1
    inlet_mfc = ct.MassFlowController(inlet, reactor, mdot=mdot*ratio)
[...]

Issue #609, opened 2019-03-07
InterfaceKinetics::solvePseudoSteadyStateProblem does not enforce coverage constraints


skyreflectedinmirrors:
### Cantera version

Dev

### Operating System

Ubuntu

### Python/MATLAB version

3.6

### Expected Behavior

Solution given by InterfaceKinetics::solvePsudeoSteadyStateProblem should at least satisfy the constraint sum(coverages) = 1.  Ideally, matching the output of InterfaceKinetics::advanceCoverages.

### Actual Behavior

It appears to find a multiple of the steady-state solution that does not satisfy the constraint.

### Steps to reproduce
1. Download the SiF4_NH3_mec.cti model from the [1D PFR notebook](https://cantera.org/examples/jupyter/reactors/1D_pfr_surfchem.ipynb.html)

2. Save the following as test.cpp

```
#include "cantera/zerodim.h"
#include "cantera/IdealGasMix.h"
#include "cantera/IncompressibleSolid.h"
#include "cantera/Interface.h"

using namespace Cantera;

void run(bool advance_coverages=false, int ifunc=-1)
{
    IdealGasMix* gas = new IdealGasMix("SiF4_NH3_mec.cti", "gas");
    gas->setState_TPX(1713, 2 * 101325 / 760.0, "NH3:6, SiF4:1");
    IncompressibleSolid* NBulk = new IncompressibleSolid("SiF4_NH3_mec.cti", "NBulk");
    NBulk->setState_TP(1713, 2 * 101325 / 760.0);
    IncompressibleSolid* SiBulk = new IncompressibleSolid("SiF4_NH3_mec.cti", "SiBulk");
    SiBulk->setState_TP(1713, 2 * 101325 / 760.0);
    std::vector<ThermoPhase*> phases({gas, NBulk, SiBulk});
    Interface* interface = importInterface("SiF4_NH3_mec.cti", "SI3N4", phases);
    interface->setState_TP(1713, 2 * 101325 / 760.0);

    std::cout << (advance_coverages ? "Advance" : "Pseudo") << std::endl;
    std::cout << "ifunc = " << ifunc << std::endl;
    if (advance_coverages)
    {
        interface->advanceCoverages(100);
    }
    else
    {
        try
        {
            interface->solvePseudoSteadyStateProblem(ifunc);
        }
        catch(Cantera::CanteraError &e)
        {
            std::cerr << e.what() << std::endl;
        }
    }

    std::vector<double> cov(interface->nSpecies());
[...]

PR #1836, opened 2025-01-02
Enable use of different Jacobian implementations with 1D solver

speth:

- Generalize `PreconditionerBase` (now renamed `SystemJacobian`) to be able represent both preconditioners ($M = I - \gamma J$) and the matrices needed for the pseudo-transient solver used by `Sim1D` ($M = J - \alpha \Delta t^{-1}$)
- Refactor the `MultiJac` to be a derived class of `PreconditionerBase` and avoid direct inheritance from `BandMatrix`
- Add methods to `OneDim` to allow setting the implementation of `SystemJacobian` to use
- Introduce a `SystemJacobian` implementation that uses Eigen's sparse direct LU factorization
- Make selection of perturbations used when computing the finite difference Jacobian used by `OneDim` more flexible. In some cases, different perturbations can reduce the number of residual and Jacobian evaluations required to converge. However, this is not universal so the default values are mostly unchanged.

This is preliminary work toward several planned improvements:
- Cantera/enhancements#31, where the banded Jacobian is a poor default (related: #1021, #989, #654)
- Cantera/enhancements#86
- Cantera/enhancements#34
- Cantera/enhancements#61 (related: #609)

*Note: some of the earlier performance gains that I saw turned out to require Jacobian settings that were not as robust as I'd hoped. I think getting reliably better Jacobians and cutting down the Jacobian evaluation time will be best addressed by work towards semi-analytical evaluation, as we've done in the 0D solver.*

**Checklist**


.github/workflows/main.yml
```diff
@@ -718,7 +718,7 @@ jobs:
     timeout-minutes: 60
     env:
       BOOST_ROOT: ${{github.workspace}}/3rdparty/boost
-      BOOST_URL: https://boostorg.jfrog.io/artifactory/main/release/1.78.0/source/boost_1_78_0.7z
+      BOOST_URL: https://github.com/boostorg/boost/releases/download/boost-1.87.0/boost-1.87.0-b2-nodocs.7z
```
bryanwweber: There's an action that will install pre-compiled versions of boost. I know we don't use the compiled libraries, but using the action might make caching easier since they'll handle all of that, and simplify our workflow file here. I used the action in the PyPI repo if you want to see it, I'm on my phone so I don't have a link handy 😀

ischoegl: Since this is the fix for some of the current CI breakages, I’d be in favor of pulling it out into its own PR.

bryanwweber: Here's where we're using it in the PyPI workflow: https://github.com/Cantera/pypi-packages/blob/a561cfa843dfe10b7682686447ef25cdb976dbc6/.github/workflows/python-package.yml#L210-L224 and here's the extension: https://github.com/MarkusJx/install-boost

speth: Using the current approach, restoring just the headers from the cache takes ~7 seconds. Restoring the full complement of compiled libraries from the cache takes ~36 seconds. I think the speed difference is worth sticking to the current approach.

include/cantera/numerics/SystemJacobian.h
```diff
@@ -0,0 +1,229 @@
...
+    //! values is solver dependent.
+    virtual int info() const {
+        throw NotImplementedError("SystemJacobian::info");
+    }
+
+    //! Elapsed CPU time spent computing the Jacobian elements.
+    double elapsedTime() const {
+        return m_elapsed;
+    }
+    void updateElapsed(double evalTime) {
```
ischoegl, line 163: Missing docstring (here and in several other instances throughout this file).

I realize that this is implicitly covered by what is directly above, but from the perspective of generated documentation (and code generation) this is irrelevant.

include/cantera/numerics/SystemJacobian.h
```diff
@@ -0,0 +1,229 @@
...
+
+    //! Absolute tolerance of the ODE solver
+    double m_atol = 0;
+
+    double m_elapsed = 0.0; //!< Elapsed CPU time taken to compute the Jacobian
+    int m_nevals = 0; //!< Number of Jacobian evaluations.
+
+    int m_age = 100000; //!< Age of the Jacobian (times incrementAge() has been called)
+
+    string m_precon_side = "none";
```
ischoegl, line 225: Docstring.

include/cantera/numerics/EigenSparseJacobian.h
```diff
@@ -0,0 +1,55 @@
...
+    //! @ingroup derivGroup
+    Eigen::SparseMatrix<double> jacobian();
+
+    //! Return the internal preconditioner matrix
+    Eigen::SparseMatrix<double> matrix() {
+        updatePreconditioner();
+        return m_matrix;
+    }
+
+    //! Print preconditioner contents
```
ischoegl: This docstring is redundant.

interfaces/cython/cantera/preconditioners.pyx
ischoegl, line 1: As I don't see this marked as 'experimental', the module likely needs a deprecation cycle (similar to what is done for the corresponding header files).

ischoegl (changes requested):
@speth ... thanks for taking this on! Overall, this looks great, and I look forward to the follow-up work. As this is part of a larger project, could you create a tracking issue over at Cantera/enhancements?

Beyond, all of my requests are nitpicks and are more about making sure that we start new files fully documented.

interfaces/cython/cantera/preconditioners.pyx
speth: The `AdaptivePreconditioner` class is still present in the parent namespace. I don't think there was previously any way to possibly use the `PreconditionerBase` class directly.

As for as the module, I think this should be regarded as an implementation detail. While you _could_ reference this class as `ct.preconditioners.AdaptivePreconditioner`, I don't think we ever use or demonstrate that style.

ischoegl: Sounds good. I _thought_ someone could have written code accessing this directly even if we didn't explicitly demonstrate it. But as this is Cython and not regular Python, I don't think this is the case actually.

ischoegl (approved):

PR #1837, opened 2025-01-07
Simplify Molar Mixture Properties

ischoegl:

There are some instances where `ThermoPhase` objects use redundant code for the evaluation of molar mixture properties.

This PR seeks to provide a more uniform approach to evaluating molar properties. By default, average properties should be calculated as a weighted sum, e.g. $\bar{h} = \sum X_k \bar{h}_k$. `getPartialMolar<>` getters are used to obtain partial molar properties.

- default behavior for molar mixture properties is changed to summation rather than throwing an exception
- improve exception handling for any virtual `ThermoPhase` method that is not implemented
- redundant/equivalent overrides are removed unless speed advantages are apparent

One caveat is that this approach has limitations for `intEnergy_mole` and `cv_mole`. These properties are implemented differently as they are largely based on simplified thermodynamic relationships rather than summations. Interestingly, an attempt to base `intEnergy_mole` on `getPartialMolarIntEnergies` causes failures in the zeroD module, albeit only on windows and sundials runners.

**Checklist**

### DIFF
--- a/include/cantera/thermo/Phase.h
+++ b/include/cantera/thermo/Phase.h
@@ -874,6 +874,9 @@ protected:
     //! Flag determining whether case sensitive species names are enforced
     bool m_caseSensitiveSpecies = false;
 
+    //! Vector of size m_kk, used as a temporary holding area.
+    mutable vector<double> m_tmpV;
+
 private:
     //! Find lowercase species name in m_speciesIndices when case sensitive
     //! species names are not enforced and a user specifies a non-canonical
--- a/include/cantera/thermo/Phase.h
+++ b/include/cantera/thermo/Phase.h
@@ -875,7 +875,7 @@ protected:
     bool m_caseSensitiveSpecies = false;
 
     //! Vector of size m_kk, used as a temporary holding area.
-    mutable vector<double> m_tmpV;
+    mutable vector<double> m_workS;
 
 private:
     //! Find lowercase species name in m_speciesIndices when case sensitive
--- a/include/cantera/thermo/ThermoPhase.h
+++ b/include/cantera/thermo/ThermoPhase.h
@@ -527,15 +527,13 @@ public:
     /**
      * Returns the amount of enthalpy per mole,
      * @f[
-     * \bar{h}(T, P, X_k) = \sum_k X_k \bar{h}_k(T)
+     * \hat{h} = \sum_k X_k \hat{h}_k
      * @f]
-     * The formula is written in terms of the partial molar enthalpies.
-     * @f$ \bar{h}_k(T, p, m_k) @f$.
-     * @relates getPartialMolarEnthalpies()
+     * @see getPartialMolarEnthalpies()
      */
     virtual double enthalpy_mole() const {
-        getPartialMolarEnthalpies(m_tmpV.data());
-        return mean_X(m_tmpV);
+        getPartialMolarEnthalpies(m_workS.data());
+        return mean_X(m_workS);
     }
 
     //! Molar internal energy. Units: J/kmol.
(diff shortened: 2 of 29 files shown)


src/clib/ct.cpp
```diff
@@ -234,10 +234,7 @@ extern "C" {
     {
         try {
             auto soln = SolutionCabinet::at(n);
-            if (a < 0 || a >= static_cast<int>(soln->nAdjacent())) {
-                throw CanteraError("soln_adjacentName", "Invalid index {}.", a);
-            }
-            return static_cast<int>(copyString(soln->adjacent(a)->name(), nm, lennm));
+            return static_cast<int>(copyString(soln->adjacentName(a), nm, lennm));
```
speth, line 237: I'm not sure I understand the change here. I think this is going to give you a strictly less readable error -- usually all you get from `std::vector::at` is the text `std::out_of_range`.

include/cantera/thermo/Phase.h
```diff
@@ -874,6 +874,9 @@ class Phase
     //! Flag determining whether case sensitive species names are enforced
     bool m_caseSensitiveSpecies = false;
 
+    //! Vector of size m_kk, used as a temporary holding area.
+    mutable vector<double> m_tmpV;
+
```
speth: I'd suggest a different name -- maybe `m_work1` or something similar. I think the name `m_tmpV` comes from a temporary array that was used to hold partial molar volumes, so this is a bit misleading.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -524,8 +524,18 @@ class ThermoPhase : public Phase
     //! @{
 
     //! Molar enthalpy. Units: J/kmol.
+    /**
+     * Returns the amount of enthalpy per mole,
+     * @f[
+     * \bar{h}(T, P, X_k) = \sum_k X_k \bar{h}_k(T)
+     * @f]
+     * The formula is written in terms of the partial molar enthalpies.
+     * @f$ \bar{h}_k(T, p, m_k) @f$.
```
speth: I know we aren't terribly consistent on notation, but I think $\bar{h}$ is a bit of an outlier -- there are several thermo classes that use `~`, and much of the reactor documentation uses `^`, while the *Chemically Reacting Flow* book that we often reference uses capital letters for molar properties and lower case for mass-specific ones.

Also, I think this is mostly copypasta from `IdealMolalSoln` and isn't true in the most general cases. I would recommend not specifying what either the mixture or partial molar properties are defined as functions of, and just writing that $\hat{h} = \sum_k X_k \hat{h}_k$.

include/cantera/thermo/ThermoPhase.h
```diff
@@ -524,8 +524,18 @@ class ThermoPhase : public Phase
...
+     * @relates getPartialMolarEnthalpies()
+     */
```
speth: I don't think the `relates` is the right directive here -- it adds these functions to a section labeled "Friends And Related Symbol Documentation", with no reference to `getPartialMolarEnthalpies`.

speth (changes requested):
Thanks for this, @ischoegl. My comments here are mostly about the documentation related to these changes.

I'm curious about the test failures when replacing `intEnergy_mole` and `cv_mole` that you mentioned cropping up on specific runners.

src/clib/ct.cpp
ischoegl: :+1: ... I reverted to the same exception handling as before but moved the function from `clib`.

ischoegl:
Thanks for the review, @speth. I've taken care of all requests.

> I'm curious about the test failures when replacing `intEnergy_mole` and `cv_mole` that you mentioned cropping up on specific runners.

While I have deleted those commits, I recall having to change some of the internal plumbing, which was a little more intrusive than what was necessary for h/s/g/cp. It wasn't difficult, but I decided against the changes as I could not troubleshoot - there weren't any errors on macOS.

*PS: New CI failures are unrelated to changes in this PR. It appears that boost downloads from JFrog are currently not working, which may or may not resolve. A workaround may involve downloading assets from [GH boostorg](https://github.com/boostorg/boost/releases).*

speth (approved):

PR #1847, opened 2025-02-03
Peng-Robinson EoS:  added setState_DP 

TJP-Karpowski:

The Peng-Robinson EoS does not implement a DP update, which is added by this merge request. 
The setState_DP function was extended with a Tguess value to enable the efficient search for the correct temperature. 
The downstream application requires the SolutionBranch of the cubic EoS to be settable; thus, the API is exposed on the ThermoPhase level. 

- added setState_DP for Peng-Robinson EoS
- added Tguess for the setState_DP function 
- the setForcedSolutionBranch is exposed on the ThermoPhase level

**TODO**
- [x] Add test of functionality to the Peng-Robinson Testset.

**Checklist**


ischoegl:
Hi @TJP-Karpowski ... thank you for the PR! 

While I believe that your PR targets a feature we'd like to see, there are some aspects that you should consider. Most people will use the `setState_DP` class method from the Python API, - via `DPX/DPY`, - which currently does not support the optional parameter. Also, from an API design aspect, the third parameter isn't consistent with what's used for other comparable setters, all of which use `tol` as a third parameter, e.g.:
```C++
void setState_HP(double h, double p, double tol=1e-9);
```

Can you think of a way to provide an automatic initial guess instead? As an aside, it would be interesting to compare to CoolProp directly in the Python test suite - it's relatively straightforward to include optional dependencies for testing purposes.

TJP-Karpowski:
Hi @ischoegl, thank you for the comment. 

I understand the issue.  As short primer my usecase: I have Cantera embedded in a CFD solver doing the thermo-update, Looping over each cell and updating the properties. So I know the temperature from the last timestep which is already a very good guess.  
Before I try to adapt it, could you give feedback on which solution you prefer?

Solutions: 

1.  Is there an option to set an object's temperature without triggering the calculation of other properties? If so, I could set the Temperature of the object (without updating) and then use the Temperature in the setState_DP calculation. 
While this is a solution for an unknowing user, this is very undefined behavior, as the setState_DP might work or fail depending on the currently set Temperature. However, I could add it to the error statements. 

2. I could add a guess State parameter (T,P, Rho) that stores estimates that can be updated separately and used when needed. This has the same issue as the calculation convergence dependence on a hidden state, but it could be documented in the call function.  

3. I think the easiest and clearest solution would be to add a fourth TGuess parameter to the C++ API and also to the Python bindings, as this would make the calculation independent of the thermo-object's current state. 

I will definitely add the tolerance (tol) as the third parameter to the call, as it is currently hard coded.
Which solution would you prefer?

Kind Regards

ischoegl:
@TJP-Karpowski ... thanks for providing the context: the use case is definitely warranting a guess to be provided.

I agree that option (3) is the cleanest solution. Leaving things in an unphysical state is a dangerous proposition, especially for APIs that may be used by inexperienced users.

ischoegl:
Hi @TJP-Karpowski ... thanks for addressing. Before further commenting, I'd like to ask you to rebase your PR to the current `main` rather than introducing a merge commit as in 0b679cb5b6c832b5439656b2f0da99a6dac8d3e2. Please let us know in case you should need assistance.

_Edit: please have a look at the Python code as well - please run `scons build` and `scons test` with the Python interface enabled, which should ensure that the submitted changes compile._

PR #1848, opened 2025-02-04
Zerod connector nodes

ischoegl:

This PR pulls out the less controversial modifications from #1788, while improving the nomenclature (see https://github.com/Cantera/enhancements/issues/213#issuecomment-2628991674):

- Create a new `ConnectorNode` base, which combine `Wall` and `FlowDevice` objects; all objects are captured in updated factories.
- Changes are propagated to all relevant API’s, i.e. Python, CLib and MATLAB.

Addresses Cantera/enhancements#213

**Checklist**


include/cantera/zeroD/ConnectorNode.h
```diff
@@ -0,0 +1,72 @@
...
+{
+class ReactorBase;
+
+/**
+ * Base class for walls and flow devices connecting reactors.
+ * In a reactor network, walls and flow devices (valves, pressure regulators, etc.)
+ * form edges of a directed graph that connect reactors that form nodes.
+ *
+ * @since New in %Cantera 3.1.
+ *
```
speth:
```suggestion
 * @since New in %Cantera 3.2.
 *
```

include/cantera/zeroD/ConnectorNode.h
```diff
@@ -0,0 +1,72 @@
...
+#include "cantera/base/global.h"
+
+namespace Cantera
+{
+class ReactorBase;
+
+/**
+ * Base class for walls and flow devices connecting reactors.
+ * In a reactor network, walls and flow devices (valves, pressure regulators, etc.)
+ * form edges of a directed graph that connect reactors that form nodes.
```
speth: This seems to get into the nomenclature question -- are connectors nodes (vertices) as the name `ConnectorNode` implies, or edges, as described here?

include/cantera/zeroD/ConnectorNode.h
```diff
@@ -0,0 +1,72 @@
...
+    void setName(const string& name) {
+        m_name = name;
+    }
+
+    //! Set the default name of a connector. Returns `false` if it was previously set.
+    void setDefaultName(map<string, int>& counts);
+
+protected:
+    //! Pair of reactors forming end points of the connector.
+    pair<shared_ptr<ReactorBase>, shared_ptr<ReactorBase>> m_nodes;
```
speth, line 66: I don't think the `pair` adds much here -- two separate member variables ends up being easier to use.

test/zeroD/test_zeroD.cpp
```diff
@@ -19,16 +19,17 @@ TEST(zerodim, simple)
 
     auto sol = newSolution("gri30.yaml", "gri30", "none");
     sol->thermo()->setState_TPX(T, P, X);
-    auto cppReactor = newReactor("IdealGasReactor", sol, "simple");
-    ASSERT_EQ(cppReactor->name(), "simple");
-    cppReactor->initialize();
+    auto cppNode = newReactor("IdealGasReactor", sol, "simple");
+    ASSERT_EQ(cppNode->name(), "simple");
+    auto reactor = std::dynamic_pointer_cast<ReactorBase>(cppNode);
+    reactor->initialize();
```
speth: `cppNode` is already a `shared_ptr<ReactorBase>` here, so the `dynamic_pointer_cast` is thankfully unnecessary. Likewise for several other casts to `ReactorBase` in this file.

test/zeroD/test_zeroD.cpp
```diff
@@ -56,6 +57,101 @@ TEST(zerodim, test_guards)
...
+    auto node0 = newReactor("IdealGasReactor", gas, "upstream");
+    auto node1 = newReactor("IdealGasReactor", gas, "downstream");
+
+    auto edge = newConnectorNode("Valve", node0, node1, "valve");
+    ASSERT_EQ(edge->name(), "valve");
+
+    auto valve = std::dynamic_pointer_cast<FlowDevice>(edge);
+    ASSERT_EQ(valve->in().name(), "upstream");
+    ASSERT_EQ(valve->out().name(), "downstream");
+
```
speth: The fact that you can do almost nothing with a `ConnectorNode` without having to `std::dynamic_pointer_cast` it to a more derived type tells me that we should largely be hiding `newConnectorNode` from users. The preferred interface would then either be the `newWall` and `newFlowDevice` functions, or just expecting them to use `make_shared<Wall>(...)` and the like.

src/zeroD/Wall.cpp
```diff
@@ -10,21 +10,27 @@
...
-    if (m_defaultNameSet) {
-        return false;
-    }
-    m_defaultNameSet = true;
-    if (m_name == "(none)" || m_name == "") {
-        m_name = fmt::format("{}_{}", type(), counts[type()]);
+    if (!m_nodes.first || !m_nodes.second) {
+        warn_deprecated("FlowDevice::FlowDevice",
+            "After Cantera 3.1, Reactors must be provided to a FlowDevice "
+            "constructor.");
```
speth:
```suggestion
            "After Cantera 3.2, Reactors must be provided to a FlowDevice "
            "constructor.");
```

src/zeroD/Wall.cpp
```diff
@@ -10,21 +10,27 @@
...
+        return;
     }
-    counts[type()]++;
-    return true;
+    // todo: switch to shared pointers after Cantera 3.1.
+    m_left = std::dynamic_pointer_cast<ReactorBase>(r0).get();
+    m_right = std::dynamic_pointer_cast<ReactorBase>(r1).get();
```
speth: `dynamic_pointer_cast` here is unnecessary.

src/zeroD/Wall.cpp
```diff
@@ -10,21 +10,27 @@
...
+    m_left->addWall(*this, 0);
+    m_right->addWall(*this, 1);
 }
 
 bool WallBase::install(ReactorBase& rleft, ReactorBase& rright)
 {
+    warn_deprecated("WallBase::install",
+        "To be removed after Cantera 3.1. Reactors should be provided to constructor "
+        "instead.");
```
speth:
```suggestion
    warn_deprecated("WallBase::install",
        "To be removed after Cantera 3.2. Reactors should be provided to constructor "
        "instead.");
```

src/zeroD/FlowDevice.cpp
```diff
@@ -11,22 +11,46 @@
...
+            "constructor.");
+        return;
     }
-    m_defaultNameSet = true;
-    string typ(type());
-    if (m_name == "(none)" || m_name == "") {
-        m_name = fmt::format("{}_{}", type(), counts[type()]);
+    // todo: switch to shared pointers after Cantera 3.1.
+    m_in = std::dynamic_pointer_cast<ReactorBase>(r0).get();
+    m_out = std::dynamic_pointer_cast<ReactorBase>(r1).get();
```
speth: Unnecessary `dynamic_pointer_cast`.

src/clib/ctreactor.cpp
```diff
@@ -547,7 +498,7 @@ extern "C" {
     double wall_area(int i)
     {
         try {
-            return WallCabinet::at(i)->area();
+            return ConnectorCabinet::as<Wall>(i)->area();
         } catch (...) {
             return handleAllExceptions(DERR, DERR);
```
speth, line 503: Especially as we move toward the generated `clib`, I'm not sure it's worth jamming more disparate types into the same `Cabinet`.

speth (changes requested):
Thanks for separating this out, @ischoegl. I think introducing the `ConnectorNode` object type is useful and avoids some repetition. Broadly, though, the interface for the different derived types of connectors is different enough that I think the interface for using them should continue to be through objects of those particular types to avoid the need for dynamically casting to the derived types, both internally and for external users.

include/cantera/zeroD/ConnectorNode.h
ischoegl: I corrected this based on the discussion in https://github.com/Cantera/enhancements/issues/213#issuecomment-2628991674

include/cantera/zeroD/ConnectorNode.h
ischoegl: I have a (slight?) preference for `pair` as it is a directed graph, where a `ConnectorNode` connects to exactly two reactors. As it is an implementation detail, we can always adjust at a later point.

test/zeroD/test_zeroD.cpp
ischoegl: I created two new versions of `newWall` and `newFlowDevice` that use reactors as parameters and deprecated the old versions.

src/clib/ctreactor.cpp
ischoegl: The generated CLib uses exactly one header file for all class specializations - all use the same `Cabinet` (at least in the current implementation, which of course may be refined as the project progresses). My plan for `zeroD` is to establish two node types - `ConnectorNode` and `ReactorBase` (potentially renamed `ReactorNode` for consistency). As `ReactorNet` is another required class, the CLib interface would be implemented using three headers.

ischoegl:
@speth - Thank you for the review (and for catching some pre-3.1 remnants of #1788). I believe your comment regarding not restricting the interface to `newConnectorNode` is valid, so I added two updated versions that retain the original split between `newWall` and `newFlowDevice`. Regarding your CLib comment, the exact implementation will be fleshed out according to Cantera/enhancements#220, so I'd like to table this for the moment.

ischoegl:
@speth ... let me know if there's anything else that needs to be done before this can be merged.

speth (approved):
Thanks, @ischoegl. This looks good to me. Sorry for the delayed in reviewing.

PR #1850, opened 2025-02-04
[C++] Fix unsigned integer check in Solution.h

### LINKED ISSUE #1849: Unsigned integer checked for negative values in Solution.h
#### Problem description

In the following snippet https://github.com/Cantera/cantera/blob/e6f3e9d4482b2e3d7b64026e62c5427099117ed3/include/cantera/base/Solution.h#L117-L119 the function argument `i` is unsigned,  but `i < 0` is checked. This currently generates a warning (at least with gcc 14). I think most often, indices in Cantera are defined as `size_t`, so the check should be superfluous. But if this is called in other language interfaces, an error message with a signed index might be more helpful to the user (if e.g. -1 is passed). What do you think?

g3bk47:

Remove check for negative index value and use the `IndexError` exception type in Solution.h.

Closes https://github.com/Cantera/cantera/issues/1849.

Before merging the PR: Is `m_adjacent` guaranteed to have at least one element? Otherwise, the maximum allowed index (last parameter in `IndexError`'s constructor) should probably be handled differently for `m_adjacent.size() == 0`.

**Checklist**

### DIFF
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -115,8 +115,8 @@ public:
 
     //! Get the name of an adjacent phase by index
     string adjacentName(size_t i) const {
-        if (i < 0 || i >= m_adjacent.size()) {
-            throw CanteraError("Solution::adjacentName", "Invalid index {}.", i);
+        if (i >= m_adjacent.size()) {
+            throw IndexError("Solution::adjacentName", "m_adjacent", i, m_adjacent.size()-1);
         }
         return m_adjacent.at(i)->name();
     }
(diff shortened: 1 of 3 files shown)


ischoegl:
@g3bk47 ... thanks for the PR!

> Before merging the PR: Is `m_adjacent` guaranteed to have at least one element? Otherwise, the maximum allowed index (last parameter in `IndexError`'s constructor) should probably be handled differently for `m_adjacent.size() == 0`.

No: it is not guaranteed to have at least one element. Indeed, it will have zero entries in most cases. As this case isn't adequately covered in `IndexError`, it may be best to address it there.

g3bk47:
@ischoegl I extended `IndexError`'s logic to check for empty arrays. Since `size-1` will be `npos` for `size==0`, I think the current PR handles all current uses of `IndexError `without introducing too much additional code.

ischoegl (approved):
Thanks, @g3bk47, this looks good to me!

PR #1863, opened 2025-03-22
Experimental CLib oneD API

ischoegl:

Implement oneD API in the experimental CLib. Further:

- Differentiate CLib function names for class specializations
- Clarify templates for header and source files
- Add experimental CLib to uploaded shared libraries

Addresses Cantera/enhancements#220

**Checklist**


ischoegl:
Not sure why the commit history shows something different, but I only added [41f6383](https://github.com/Cantera/cantera/pull/1863/commits/41f63835ae425a6da7826b8ad11c1c7ddbe2e503) after requesting the review - the additions amount to a couple of lines. (_PS: likely due to a superfluous `--force`, but then I hadn't done any work on that branch._)

.gitignore
```diff
@@ -51,3 +51,4 @@ environment.y*
 coverage/
 coverage.info
 .coverage
+yaml/
```
speth: I think it would be more consistent with our other build steps for this to get generated as a subdirectory of `build`, in which case it will automatically be excluded.

include/cantera/oneD/Domain1D.h
```diff
@@ -511,9 +511,20 @@ class Domain1D
         return m_z;
     }
 
+    //! Set up initial grid.
+    //! @since New in %Cantera 3.2.
+    void setupGrid(const vector<double>& grid);
```
speth, line 516: Can we avoid introducing this overload for `setupGrid`? If we did this for every method that took `double*`, we'd have a _lot_ of extra boilerplate code, and I don't want to force users to use `std::vector`.

speth (changes requested):
Thanks, @ischoegl. I think this generally looks good. Just a couple of minor comments.

.gitignore
ischoegl: I'll hold off on this then, as this comes from the `YamlSourceGenerator` ... this should follow the same approach as `CLibSourceGenerator`, which currently drops things into the root folder. A resolution is, at least implicitly, part of the roadmap in Cantera/enhancements#220. I'll add this on the todo list there to keep track of this.

include/cantera/oneD/Domain1D.h
ischoegl: This is a little more than meets the eye. The current `setupGrid(size_t n, const double* z)` is an outlier as it takes the array size as an additional argument rather than just the pointer. I currently don't have a rule to handle that outlier in the logic of `CLibSourceGenerator` and am hesitant to add it as it's different from what we do elsewhere. ~We have the alternative of introducing a `setupGrid(const double* z)` and deprecate the current version to be consistent with the remainder of our API.~ (_Edit: that won't work as different to other cases the size of the array isn't known a priori_). I opted for the inherently safer version of a `std::vector`.

_PPS: Fwiw, the signature `setupGrid(size_t n, const double* z)` with size followed by array pointer is consistent with how we handle things in CLib. This is about the decision whether we allow for an overload that can be handled by standard logic or we add an edge case._

speth (approved):

PR #1875, opened 2025-04-19
Add Zizmor linting and address findings

bryanwweber:

- **[CI] Address findings from zizmor**
- **[CI] Add zizmor job running on PRs and main**

The [zizmor](https://woodruffw.github.io/zizmor/) project is a static analysis tool for GitHub workflows. The main suggestion I've seen it make when I've used it is to avoid places where bad actors might inject arbitrary code using the Action template strings `${{ ... }}`. The other change it suggested here was to remove unnecessary job permissions.

My motivation to do this was the `changed-files-action` attack a few weeks ago. I don't think `zizmor` would have stopped that attack, but beefing up code quality/security seems worthwhile.

That said, I'm not convinced this is a good thing to add for `Cantera/cantera` itself. I've already [added it to the `pypi-packages`](https://github.com/Cantera/pypi-packages/blob/main/.github/workflows/zizmor.yml) repo, because we're publishing packages from there and I really don't want those to include malware :grimacing: Since we don't actually publish anything from this repo, I think it has a little lower impact, but it still seemed worth it to at least moot the idea with the associated changes. Thus, draft status and I'm happy to close it if we think it's too much noise.

**Checklist**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "monthly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,7 @@ updates:
     schedule:
       # Check for updates to GitHub Actions every month
       interval: "monthly"
+    groups:
+      actions:
+        patterns:
+          - "*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,5 +3,5 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      # Check for updates to GitHub Actions every week
+      # Check for updates to GitHub Actions every month
       interval: "monthly"
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Get changed C++ files
         if: success() || failure()
         id: changed-cxx-files
-        uses: tj-actions/changed-files@v46
+        uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c # v46.0.5
         with:
           files: |
             **.h
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -6,6 +6,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+
 jobs:
   linter:
     name: Lint checks
@@ -16,6 +19,7 @@ jobs:
         name: Checkout the repository
         with:
           fetch-depth: 100
+          persist-credentials: true
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
@@ -35,7 +39,7 @@ jobs:
         if: (success() || failure()) && steps.changed-cxx-files.outputs.any_changed == 'true'
         run: |
           RETCODE=0
-          for file in ${{ steps.changed-cxx-files.outputs.all_changed_files }}; do
+          for file in ${ALL_CHANGED_FILES}; do
             RESULT=`perl -0777 -ne 'while(m/\n\n *\/\*!.*?\*\//gs){print "$&\n";}' $file`
             if [ "$RESULT" ]; then
               echo ------ ${file} ------
@@ -44,6 +48,8 @@ jobs:
             fi
           done
           exit $RETCODE
+        env:
+          ALL_CHANGED_FILES: ${{ steps.changed-cxx-files.outputs.all_changed_files }}
       - name: Whitespace errors
         if: success() || failure()
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,12 +9,17 @@ on:
     branches:
       - main
       - testing
-      - "3.1"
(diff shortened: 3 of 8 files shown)


ischoegl:
@bryanwweber  … I’d be :+1: with this change. The post-merge test failures are unrelated as far as I can tell, but there’s one packaging test that could be addressed?

bryanwweber:
Thanks @ischoegl! The packaging test is because that workflow running on a PR from my work doesn't have permission to access the token to start the workflow in the other repository.

ischoegl (dismissed):

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
```
ischoegl: Week or month?

.github/workflows/linters.yml
```diff
@@ -25,7 +29,7 @@ jobs:
       - name: Get changed C++ files
         if: success() || failure()
         id: changed-cxx-files
-        uses: tj-actions/changed-files@v46
+        uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c # v46.0.5
```
ischoegl, line 32: Would leaving the trailing comments start to be misleading as tags are updated? Or is dependabot smart enough to update?

bryanwweber: Dependabot is, fortunately, smart enough to update the comments 😁

ischoegl (dismissed):
Thanks, @bryanwweber - approving again.

speth:
Why are there no CI runs showing up here any more?

.github/workflows/linters.yml
```diff
@@ -44,6 +48,8 @@ jobs:
             fi
           done
           exit $RETCODE
+        env:
+          ALL_CHANGED_FILES: ${{ steps.changed-cxx-files.outputs.all_changed_files }}
```
speth, line 52: I don't quite understand the potential attack vector here, but if the need is to not use the `${{ }}` style substitutions in scripts, then can we please document that in the developer docs (`doc/sphinx/develop/...`). I guess we might need a new page on CI stuff.

.github/workflows/main.yml
```diff
@@ -59,9 +65,9 @@ jobs:
           LIB_LAPACK=$(ldconfig -p | grep liblapack.so.3 | awk '{print $4}' | head -n 1)
           echo "LD_PRELOAD=$LIB_STDCXX:$LIB_OPENBLAS:$LIB_LAPACK" >> $GITHUB_ENV
       - name: Set up MATLAB
-        uses: matlab-actions/setup-matlab@v2
+        uses: matlab-actions/setup-matlab@718d4320188c73c86eb94ce76b553cbf89778487 # v2.5.0
```
speth, line 68: Likewise, the need to specify particular hashes for 3rd party actions (I guess all bets are off if the core GitHub actions are compromised...), and how these get updated (can't say I'm that excited for the dependabot-driven churn, but oh well).

.github/workflows/main.yml
```diff
@@ -96,6 +102,7 @@ jobs:
       name: Checkout the repository
       with:
         submodules: recursive
+        persist-credentials: false
```
speth, line 105: Is there any way to make this a default? Otherwise, also needs to be documented.

speth (changes requested):

.github/workflows/linters.yml
bryanwweber: My understanding is that direct template substitutions in running code can lead to arbitrary code execution. For example, a malicious actor could inject a string like `value$(rm -rf /)value`. If this is templated directly into the script, it will actually be executed. However, putting into an environment variable this way ensures that string escaping is done and the code won't be injected. 

I'm not familiar with all the weird ways that bash can find code to execute, so this might not be a particularly relevant example, but hopefully you get the gist. The problem is not the templated string, but rather expanding the template directly in the script context. 

Yes, I'll definitely document it!

.github/workflows/main.yml
bryanwweber: Couple of thoughts:
1. We don't have to pin to hashes, we can configure zizmor to accept tags instead. But since the tj-actions thing, I'm inclined to just use hashes and let dependabot keep them updated
2. Dependabot is configured to run monthly and I'll set it to group updates into a single PR. Hopefully that will minimize the additional churn/overhead.

.github/workflows/main.yml
bryanwweber: No, there isn't as far as I know. I'll document it.

.github/workflows/main.yml
ischoegl: While it’s producing churn, I consider it relatively minor.

speth (dismissed):

speth:
Ah, I just figured out what the failing `CI / Build docs` was complaining about. The relevant output is:
```
/home/runner/work/cantera/cantera/build/doc/sphinx/develop/continuous-integration.md: WARNING: document isn't included in any toctree
```
The issue is that in `doc/sphinx/develop/index.md`, the `continuous-integration` file needs to be listed not only in the bulleted list but also in the hidden TOC.

bryanwweber:
Thanks @speth! Hopefully I just fixed that, but of course it dismissed your review :smile:

ischoegl (approved):
Thanks @bryanwweber! ... the fix addresses the comment.

PR #1881, opened 2025-05-06
Complete Unit Test Suite for the Matlab Toolbox - Pt.2

ssun30:

Added unit tests for the remaining classes and methods in the Matlab toolbox. 

Fixed bugs that prevent these tests from passing. 

Updates to Cantera/enhancements#177

Added tests for rates of progress, rate constants, species rates, and thermodynamic values. 

**Checklist**


ssun30:
@ischoegl The second part of the unit test suite is ready for review. This will include tests for 0D classes. That leaves 1D classes to be the only part remaining, and these will come in a later PR as this one is also approaching 3000 lines. 

All CIs will pass but there is one issue that throws warnings about already deleted objects:

```
  CanteraError thrown by Cabinet::at:
  Object with index 5 has been deleted.
  
  
  Error in Solution/delete (line 98)
              ctFunc('soln_del', s.solnID);
```
This happens when an `Interface` class object is deleted in MATLAB, which calls the `Solution` class destructor, since `Interface` now inherits from `Solution`. Here's how the `Interface` constructor works:

First, it calls the `soln_newInterface` method in clib to create a new `Interface` object in `SolutionCabinet`. 
`ID = ctFunc('soln_newInterface', src, name, na, adj);`

It then inherits methods and properties from the `Solution` class.
`s@Solution(ID);`

These steps should address this comment from #1665:
> Both `Interface` and `Solution` are stored in the same `SolutionCabinet`, so the `solnID` should reference the correct object. You should not require `interfaceID`. I'm not sure how an object that should be correctly loaded in C++ ends up scrambled in clib? If there's an issue, it needs to be fixed in clib rather than having a workaround being created in MATLAB. Fwiw, `Interface` inherits from `Solution` in C++ and the same should be true in MATLAB (I believe there are technical reasons why it's implemented differently in Cython).

But for some reason, when the destructor is called, Cantera somehow thinks the object is already deleted, even though it isn't. Prior to invoking the destructor, all the methods of the `Interface` object could be called without problems. Any idea why this could happen?

interfaces/matlab_experimental/Base/Water.m
```diff
@@ -17,6 +17,11 @@
...
     % :return:
     %     Instance of class :mat:class:`Solution`.
-
-    w = Solution('liquidvapor.yaml', 'water', 'water');
+    if nargin == 0 | strcmp(backend, 'Reynolds')
+        w = Solution('liquidvapor.yaml', 'water', 'water');
+    elseif strcmp(backend, 'IAPWS95')
+        w = Solution('liquidvapor.yaml', 'liquid-water-IAPWS95', 'water');
+    else
+        error(['Unknow backend: ', backend]);
```
bryanwweber:
```suggestion
        error(['Unknown backend: ', backend]);
```

test/data/testConstPressureReactor.yaml
bryanwweber, line 1: This file doesn't need to be duplicated, there's probably an existing file that has this same specification too.

test/matlab_experimental/ctTestConstPressureReactor.m
```diff
@@ -0,0 +1,186 @@
...
+        surf2
+        net1
+        net2
+    end
+
+    methods (TestClassSetup)
+
+        function testSetUp(self)
+            ctTestSetUp
+            copyfile('../data/testConstPressureReactor.yaml', ...
```
bryanwweber: Instead of copying files, we should set it up so that the `test/data` directory is part of the search path for input files.

test/matlab_experimental/ctTestConstPressureReactor.m
```diff
@@ -0,0 +1,186 @@
...
+            self.makeReactors('addSurf', true);
+            self.verifyTrue(isa(self.surf1, 'ReactorSurface'));
+            self.verifyTrue(startsWith(self.surf1.name, 'ReactorSurface_'));
+            self.surf1.name = 'name-of-reactor-surface';
+            self.verifyEqual(self.surf1.name, 'name-of-reactor-surface');
+        end
+
+        function testClosed(self)
+            self.makeReactors();
+            self.integrate();
```
bryanwweber, line 153: Can you move the assertions out of the `integrate()` function and into a function with a more appropriate name?

bryanwweber (changes requested):
Thanks @ssun30! I have a few suggestions for improvements here.

interfaces/matlab_experimental/Reactor/ReactorNet.m
```diff
@@ -54,49 +54,56 @@
...
                 return
             end
-            ctFunc('reactornet_del', r.id);
+            ctFunc('reactornet_del', n.id);
         end
 
         %% ReactorNet Utility Methods
 
-        function advance(r, tout)
+        function addReactor(n, r)
```
ischoegl: This won't be supported by the new CLib API. Instead, all reactors have to be passed to the constructor.

test/matlab_experimental/ctTestWellStirredReactor.m
```diff
@@ -0,0 +1,163 @@
...
+        fuel_mfc
+        oxidizer_mfc
+        valve
+        net
+    end
+    methods (TestClassSetup)
+
+        function testSetUp(self)
+            ctTestSetUp
+            copyfile('../data/testWellStirredReactor.yaml', ...
```
ischoegl: see @bryanwweber's comment.

ischoegl (commented):
Thanks, @ssun30. I only have minor comments.

ischoegl:
@ssun30 … the changes in commit 258ce817d39159cd741031a5d501bba413bada70 make me believe that #1898 may have been a MATLAB implementation issue after all. Could you verify?

ssun30:
> @ssun30 … the changes in commit [258ce81](https://github.com/Cantera/cantera/commit/258ce817d39159cd741031a5d501bba413bada70) make me believe that #1898 may have been a MATLAB implementation issue after all. Could you verify?

That's correct. The `soln_del` method was called twice by the destructors of both `Interface` and `Solution`.

interfaces/matlab_experimental/Base/Water.m
```diff
@@ -17,6 +17,11 @@
...
     % :return:
     %     Instance of class :mat:class:`Solution`.
-
-    w = Solution('liquidvapor.yaml', 'water', 'water');
+    if nargin == 0 | strcmp(backend, 'Reynolds')
+        w = Solution('liquidvapor.yaml', 'water', 'water');
+    elseif strcmp(backend, 'IAPWS95')
+        w = Solution('liquidvapor.yaml', 'liquid-water-IAPWS95', 'water');
+    else
+        error(['Unknown backend: ', backend]);
```
bryanwweber: Since there's only two options, it's probably worth enumerating them in the error message

interfaces/matlab_experimental/Base/Interface.m
```diff
@@ -84,7 +92,7 @@
         end
 
         function c = get.concentrations(s)
-            surfID = s.tr_id;
+            surfID = s.tpid;
```
speth: Elsewhere, this is capitalized as `s.tpID`.

interfaces/matlab_experimental/Base/Water.m
```diff
@@ -15,8 +15,15 @@
...
     %     Instance of class :mat:class:`Solution`.
-
-    w = Solution('liquidvapor.yaml', 'water', 'water');
+    if nargin == 0 | strcmp(backend, 'Reynolds')
+        w = Solution('liquidvapor.yaml', 'water', 'water');
+    elseif strcmp(backend, 'IAPWS95')
+        w = Solution('liquidvapor.yaml', 'liquid-water-IAPWS95', 'water');
+    else
+        error(['Backend must be either ''water'' or ''liquid-water-IAPWS95''']);
+    end
```
speth: I believe the names in the if/elseif block are correct (for consistency with the Python interface), but both the documentation of the `backend` argument and the error message are incorrect.

interfaces/matlab_experimental/Reactor/MassFlowController.m
```diff
@@ -7,9 +7,7 @@
     % simulate a mass flow controller that maintains a constant mass flow
     % rate independent of upstream or downstream conditions. If two reactor
     % objects are supplied as arguments, the controller is installed
-    % between the two reactors. Otherwise, the :mat:func:`install` method
-    % should be used to install the :mat:class:`MassFlowController` between
-    % reactors.
+    % between the two reactors.
     %
```
speth: There's no "if" about it -- two Reactor objects are required as arguments.

interfaces/matlab_experimental/Reactor/ReactorNet.m
```diff
@@ -133,55 +144,55 @@ function advance(r, tout)
...
+            ctFunc('reactornet_setTolerances', n.id, rerr, aerr);
         end
 
         %% ReactorNet get methods
 
-        function t = get.dt(r)
-            t = ctFunc('reactor_step', r.id);
+        function t = get.dt(n)
+            t = ctFunc('reactornet_step', n.id);
         end
```
speth: This is extremely misleading. Each time this property is accessed, the `ReactorNet` will take an additional timestep, with the size of that timestep being the value returned. I don't think `dt` should be a property of the `ReactorNet` object, given it's not a function existing on the underlying C++ class or in any other language interface.

interfaces/matlab_experimental/Reactor/ReactorNet.m
```diff
@@ -133,55 +144,55 @@ function advance(r, tout)
...
-            t = ctFunc('reactornet_rtol', r.id);
+        function t = get.rtol(n)
+            t = ctFunc('reactornet_rtol', n.id);
         end
 
-        function s = sensitivity(r, component, p, rxtr)
+        function s = sensitivity(n, component, p, r)
             % Sensitivity of the solution variable `c` in reactor `rxtr`
             % with respect to the parameter `p` ::
             %
```
speth: The docstring also needs to be updated to indicate the changed parameter names.

test/data/testConstPressureReactor.yaml
```diff
@@ -0,0 +1,10 @@
+phases:
+- name: gas
+  species:
+  - gri30.yaml/species: [H2, H, O, O2, OH, H2O, HO2, H2O2, CH3, CH4, CO, CO2,
+      HCO, CH2O, CH3O, CH3OH, N2, AR]
+  thermo: ideal-gas
```
speth: It would be helpful to add a comment to this mechanism indicating which tests it's used in, and why (i.e. that it's a stripped-down version of GRI3.0 for speed) and to modify the corresponding Python tests to use this file rather than their own inline mechanism definition.

Also, you can put both phase definitions into a single file, and give it a name that's more descriptive of the mechanism rather than its use case, for example something like `ch4_minimal.yaml`.

test/matlab_experimental/ctTestConstPressureReactor.m
```diff
@@ -0,0 +1,195 @@
...
+            props = properties(self);
+            for i = 1:length(props)
+                prop = self.(props{i});
+                if isa(prop, 'handle')
+                    delete(prop)
+                end
+            end
+        end
+
+    end
```
speth: Rather than repeating these methods on every test class, can you make a common base class for all Cantera tests that derives from `matlab.unittest.TestCase`? That would be similar to what we did in the Python test suite back when it used the Python `unittest` module.

test/matlab_experimental/ctTestConstPressureReactor.m
```diff
@@ -0,0 +1,195 @@
...
+        function makeReactors(self, arg)
+            arguments
+                self
+                arg.addQ (1,1) logical = false
+                arg.addMdot (1,1) logical = false
+                arg.addSurf (1,1) logical = false
+            end
+
+            src = '../data/testConstPressureReactor.yaml';
+
```
speth: Rather than specifying a relative path, I think we should ensure that the `test/data` directory is added to the path as part of the test setup. This would be consistent with how we handle accessibility of the test data files in the GTest and Python test suites.

test/matlab_experimental/ctTestConstPressureReactor.m
```diff
@@ -0,0 +1,195 @@
+classdef ctTestConstPressureReactor < matlab.unittest.TestCase
+
```
speth: A comment explaining what this test does would be helpful, as well as noting where it comes from:

```suggestion
classdef ctTestConstPressureReactor < matlab.unittest.TestCase
    % The constant pressure reactor should give essentially the same results as
    % as a regular "Reactor" with a wall with a very high expansion rate
    % coefficient. Replicates the tests done in the Python test suite in
    % class TestConstPressureReactor.
```

interfaces/matlab_experimental/Reactor/Connector.m
```diff
@@ -59,7 +59,9 @@
 
         function delete(c)
             % Delete the :mat:class:`Connector` object.
-
+            if isempty(c.id)
```
ischoegl: Is it possible for this to happen at all? The way I read this, the only way this could happen is if the constructor doesn't fully execute, but that produces an error and the object wouldn't exist?

interfaces/matlab_experimental/Reactor/Reactor.m
```diff
@@ -89,10 +89,10 @@
     methods
         %% Reactor Class Constructor
 
-        function r = Reactor(content, typ, name)
+        function r = Reactor(content, typ, name, V0)
```
ischoegl: I am not supportive of introducing signatures that deviate from other Cantera APIs (see [C++ constructor](https://cantera.org/dev/cxx/dc/d5e/classCantera_1_1Reactor.html), [generated CLib API](https://cantera.org/dev/cxx/d6/d6d/group__CAPIctreactor.html) and [Python API](https://cantera.org/dev/python/zerodim.html#reactor)), so please remove `V0` here.

interfaces/matlab_experimental/Reactor/ReactorNet.m
```diff
@@ -54,49 +54,50 @@
...
         end
 
         %% ReactorNet Class Destructor
 
-        function delete(r)
+        function delete(n)
             % Delete the :mat:class:`ReactorNet` object object.
 
-            if isempty(r.id)
+            if isempty(n.id)
```
ischoegl: Similar here: is it possible to end up with empty `n.id` at all?

interfaces/matlab_experimental/Base/Water.m
```diff
@@ -15,8 +15,15 @@
...
     % :return:
     %     Instance of class :mat:class:`Solution`.
-
-    w = Solution('liquidvapor.yaml', 'water', 'water');
+    if nargin == 0 | strcmp(backend, 'Reynolds')
+        w = Solution('liquidvapor.yaml', 'water', 'water');
+    elseif strcmp(backend, 'IAPWS95')
+        w = Solution('liquidvapor.yaml', 'liquid-water-IAPWS95', 'water');
+    else
+        error(['Backend must be either ''water'' or ''liquid-water-IAPWS95''']);
```
ischoegl: Can't you just rely on the exception handling that is implemented in C++ and re-raise the error?

_PS: I realize that it's mimicking the less-than-ideal implementation of the Python API._

ischoegl (commented):
@ssun30 ... some minor comments.

As in my previous feedback, I'd prefer if the MATLAB toolbox could adhere to the same approach as the other APIs and introduce minimal extra code: ideally, it should act as a minimal wrapper for CLib that just re-establishes the underlying C++ class structure.

One thing I noticed recently is that the bulk of the exception handling in MATLAB is re-implemented rather than relying on exceptions thrown by the C++ core - this results in a lot of redundant code.

test/matlab_experimental/ctTestConstPressureReactor.m
```diff
@@ -0,0 +1,195 @@
...
+            if arg.addSurf
+                self.interface1 = Interface('diamond.yaml', 'diamond_100', ...
+                                            self.gas1, self.solid);
+                self.interface2 = Interface('diamond.yaml', 'diamond_100', ...
+                                            self.gas2, self.solid);
+
+                C = zeros(1, self.interface1.nSpecies);
+                C(1) = 0.3;
+                C(5) = 0.7;
+                C = {C};
```
speth: I'm curious why the coverages need to be converted into a one-element cell array rather than simply being passed as an array.

test/matlab_experimental/ctTestConstPressureReactor.m
```diff
@@ -0,0 +1,195 @@
...
+                self.surf2 = ReactorSurface(self.interface2, self.r2);
+                self.surf2.area = 0.2;
+                self.interface2.coverages = C;
+            end
+
+            self.net1 = ReactorNet(self.r1);
+            self.net2 = ReactorNet(self.r2);
+            self.net1.maxTimeStep = 0.05;
+            self.net2.maxTimeStep = 0.05;
+            % self.net2.maxErrTestFails = 10;
```
speth: If setting this isn't necessary for the test to work, you can just leave it out.

test/matlab_experimental/ctTestFlowDevice.m
```diff
@@ -0,0 +1,230 @@
...
+            n1 = ReactorNet(self.r1);
+
+            self.verifyTrue(startsWith(self.r1.name, sprintf('%s_', self.r1.type)));
+            self.verifyTrue(startsWith(res.name, sprintf('%s_', res.type)));
+            self.verifyEqual(v.type,'Valve');
+            self.verifyTrue(startsWith(v.name, 'Valve_'));
+            v.name = 'name-of-valve';
+            self.verifyEqual(v.name, 'name-of-valve');
+
+            clear res v n1
```
speth: Do local variables not get cleared automatically when returning from the function?

test/matlab_experimental/ctTestFlowReactor1.m
```diff
@@ -0,0 +1,124 @@
...
+
+        function testNonReacting(self)
+            self.gas = Solution('../data/testConstPressureReactor.yaml');
+            self.gas.TPX = {300, OneAtm, 'O2:1.0'};
+            self.reactor = FlowReactor(self.gas);
+            self.reactor.massFlowRate = 10;
+
+            self.net = ReactorNet(self.reactor);
+
+            self.assumeFail('Skipped until Reactor.speed is implemeneted');
```
speth: Here and in `testReacting`:
```suggestion
            self.assumeFail('Skipped until Reactor.speed is implemented');
```

test/matlab_experimental/ctTestConstPressureReactor.m
```diff
@@ -0,0 +1,195 @@
...
+
+            T0 = 1200;
+            P0 = 25 * OneAtm;
+            X0 = 'CH4:0.5, H2O:0.2, CO:0.3';
+
+            self.gas1.TPX = {T0, P0, X0};
+            self.gas2.TPX = {T0, P0, X0};
+
+            self.r1 = IdealGasReactor(self.gas1);
+            self.r2 = Reactor(self.gas2);
```
speth: The class name claims this is a test for `ConstPressureReactor`, but these are not the constant pressure reactor types.

test/matlab_experimental/ctTestFlowReactor2.m
```diff
@@ -0,0 +1,240 @@
...
+
+            self.assumeFail('Skipped until more Reactor methods are implemented');
+
+            try
+                self.net.initialize();
+            catch ME
+                self.verifySubstring(ME.identifier, 'Cantera:ctError');
+                self.verifySubstring(ME.message,  'mass flow rate');
+            end
+        end
```
speth: Some of these stub/incomplete tests are not particularly useful -- it's not very clear what the contents of this test should be, or what methods need to be implemented to allow this test case to be completed. I would suggest updating all of the cases you have that just say "until more Reactor methods are implemented" and similar to at least name the missing methods, even if you don't have the (commented-out) code for what those cases should look like.

test/matlab_experimental/ctTestFlowReactor2.m
```diff
@@ -0,0 +1,240 @@
...
+            self.assumeFail('Skipped until addReactions is implemented');
+        end
+
+        function testIntegratorErrorAdvance(self)
+            self.assumeFail('Skipped until addReactions is implemented');
+        end
+
+        function testRecoverableIntegratorErrors(self)
+            self.assumeFail('Skipped until addReactions is implemented');
+        end
```
speth: While I know that the Python tests that these were based on used the `add_reactions` method. But just from looking at this, it's really not obvious what these are supposed to do.

Broadly, I don't think it's necessary for the Matlab test suite to cover all the same cases that are present in the Python test suite. Many of those tests are checking behavior of the underlying C++ implementation. All that's really useful here would be one case where the `advance` method raises an exception, and checking that that occurs is sufficient (and it seems that there are several such cases below).

test/matlab_experimental/ctTestPath.m
```diff
@@ -19,6 +19,7 @@
 cantera_root = getenv('CANTERA_ROOT');
 
 % Add the Cantera toolbox to the Matlab path
+addpath(genpath([cantera_root, '/data']));
```
speth: Is this necessary / effective? I would expect `addpath` only to affect the search path for Matlab functions / classes, not for Cantera data files. For that, you should be setting the `CANTERA_DATA` environment variable. I'd say you should use `ct_addDataDirectory`, but that doesn't exist in the legacy Clib interface.

test/matlab_experimental/ctTestReactorSensitivities.m
```diff
@@ -0,0 +1,116 @@
...
+
+    methods (Test)
+
+        function testSensitivities1(self)
+            self.gas1 = Solution('gri30.yaml', '', 'none');
+            self.gas1.TPX = {1300, 20 * OneAtm, 'CO:1.0, H2:0.1, CH4:0.1, H2O:0.5'};
+            self.r1 = IdealGasReactor(self.gas1);
+            self.net = ReactorNet(self.r1);
+
+            self.assumeFail('Skipped until more methods are implemented for ReactorNet');
```
speth: Given that every test in this suite is currently being skipped, I might suggest it's a bit premature to include it at all.

test/matlab_experimental/ctTestSurfaceKinetics.m
```diff
@@ -0,0 +1,80 @@
...
+
+        function testCoveragesRegression1(self)
+            self.assumeFail('Skipped until ReactorSurface.coverages is implemented');
+        end
+
+        function testCoveragesRegression2(self)
+            self.assumeFail('Skipped until ReactorSurface.coverages is implemented');
+        end
+
+    end
```
speth: Again, if you're skipping all the tests, I think we should defer adding this test suite.

test/matlab_experimental/StateData.m
```diff
@@ -0,0 +1,54 @@
+classdef StateData
+
+    properties
+        phase
```
speth: A docstring for this class would be useful, to explain where it's used within the test suite.

test/matlab_experimental/Tolerances.m
```diff
@@ -0,0 +1,37 @@
+classdef Tolerances
+
+    properties
```
speth: Likewise, please add a docstring for this utility class.

speth (changes requested):
Thanks for your work on this, @ssun30. I'm really glad to see the expanded testing of the Matlab toolbox, and I think the approach taken here of translating some of the Python tests is generally a good one.

One thing to consider, to some extent here, but especially looking forward to tests of the 1D solver, is that the Python test suite includes many tests which are focused on making sure that the underlying algorithms are working correctly, which isn't really necessary to repeat in the Matlab test suite. Here, it's sufficient to just ensure that all of the classes / methods behave in expected ways, but not necessarily to test whether the solvers produce expected results across different ranges of inputs (though it is important to test that inputs that result in errors propagate those errors back in the expected way).

interfaces/matlab_experimental/Reactor/Connector.m
speth: I was also struggling with understanding conditions under which this could happen.

ssun30: I've removed these redundant checks. I had a duplicate destructor call problem before and I've solved it.

test/matlab_experimental/ctTestConstPressureReactor.m
ssun30: The setter method for coverages used to require the `norm` flag to determine whether coverages are normalized. I've separated the setter for normalized and unnormalized coverages.

interfaces/matlab_experimental/Reactor/ReactorNet.m
```diff
@@ -54,49 +43,46 @@
...
 
             if isa(reactors, 'Reactor')
                 % Allow simpler syntax for creating a network with one
                 % reactor.
-                reactors = {reactor};
+                reactors = {reactors};
             end
 
-            r.id = ctFunc('reactornet_new');
+            n.id = ctFunc('reactornet_new');
```
ischoegl, line 61: I believe that this line should check whether it returns `-1` ... see legacy CLib definition
```C++
    int reactornet_new()
    {
        try {
            return NetworkCabinet::add(make_shared<ReactorNet>());
        } catch (...) {
        }
    }
```

The negative return value indicates the case where a constructor fails; the underlying error message thrown by C++ can be retrieved via `ct_getCanteraError`. It may be worthwhile creating a utility function for that purpose.

ischoegl (dismissed):
Thank you for removing the unnecessary checks for the destructors. One thing I wanted to point out is that you may want to check whether constructors actually work (currently, they could fail silently, potentially wreaking havoc later on). Those checks should be added for all CLib constructor calls.

_Edit: ok to be deferred._

interfaces/matlab_experimental/Reactor/ReactorNet.m
speth: I think that's a good idea in general, but this suggestion should probably be tracked outside of this PR. There are already a lot of other changes to the toolbox itself that are needed.

I'd like to limit this PR to fixing just enough of the toolbox so that the additional tests pass and we can then use this working test suite to help us ensure that changes to the interface in other PRs are working correctly.

ischoegl: @speth ... it's totally fine to defer, but as it has been mentioned that CLib is prone to "hard crashes," I wanted to start pushing in a direction where issues caused by the external API are caught properly. It is also likely that the - now removed - workaround in the destructor appeared to be necessary because of interactions with objects where exceptions were not caught.

speth: I don't think cases like this can be sources of "hard crashes". From the library's point of view, any error here is handled. If it's ignored by the code on the Matlab side, then attempts to use that invalid handle should also just lead to errors that are caught by the `try/catch` construct in CLib and return appropriate error codes.

The possible sources of hard crashes that come to mind to me are (1) calls into CLib where array sizes are not correct, and do not correspond with the declared array size argument; and (2) problems from mismatches in functions between what Cantera was compiled with and what's provided by Matlab (as in the 64-bit integer MKL / `LD_PRELOAD` mess). On this front, I'd like to see if the switch to the "out of process" library approach solves the problems.

ischoegl: @speth … agreed that the constructor returning -1 would not be the situation where hard crashes are expected, but there wouldn’t be an actual C++ object while MATLAB isn’t aware of it. A subsequent call to a C++ method would automatically fail, but not necessarily cause a crash; users would be left guessing about what’s actually going on. Based on that, I believe that MATLAB not checking/accessing C++ exception handling is problematic and should be addressed.

I don’t have access to the wiki documentation to fully understand the 64 bit integer situation. Wrong array sized can obviously cause hard crashes, but this is hardly the fault of CLib.

Overall, I hope that we’ll find out more about those “hard crashes”. They have been mentioned, but I’d like to see them being reported in a way that is reproducible.

test/matlab_experimental/ctTestConstPressureReactor.m
ssun30: I will defer this and reimplement how the test data paths are added when we switch to the experimental clib and can call `ct_addDataDirectories`.

ischoegl (commented):
All good from my side (my request for changes is removed as it goes beyond the scope of this PR).

speth (approved):
Thanks for the updates, @ssun30. This looks good to me.

bryanwweber (approved):

Issue #1852, opened 2025-02-12
MATLAB Interface Is Incompatible with the Latest Version of MATLAB


ssun30:
**Problem description**

The latest version of MATLAB (R2024b) comes with Intel MKL as the default library for BLAS and LAPACK. This version of MKL is built with ILP64 where a 64-bit `int` type is used for all integers in order to support very large array sizes (>2^31-1). The default OpenBLAS package that comes with Cantera uses LP64 where standard integers are 32-bit and the `long` type is required for 64-bit integers. This causes an incompatibility issue when data in an array is passed back and forth between MATLAB and OpenBLAS. In some cases this will just return an incorrect type error but in the worst case scenario will cause a buffer overflow and instantly crashes MATLAB. 

A partial workaround exists to resolve this problem, as mentioned in [this thread](https://www.mathworks.com/matlabcentral/answers/1982374-custom-c-library-gives-error-intel-mkl-error-parameter-8-was-incorrect-on-entry-to-dgemm-when-us). We can bypass ILP64 MKL by setting the environment variable `LD_PRELOAD` and force MATLAB to use the LP64 OpenBLAS library in the Cantera environment. This will correctly set the types of integers and Cantera will successfully solve any linear algebra problems. However, this workaround is unstable and will cause crashes when the `plot` function is invoked, for example to plot to concentration of species over time. 

**Steps to reproduce**

1. In MATLAB, run `ctLoad`;
2. Run example `equil.m`;
3. Exit MATLAB and set `LD_PRELOAD` to the location of the OpenBLAS library in your Cantera environment;
4. Repeat 1) and 2).

**Behavior**

The `gas` object will fail to equilibrate, this error appears in the MATLAB terminal:
```
Error using ctFunc (line 11)

*******************************************************************************
CanteraError thrown by MultiPhase::equilibrate_MultiPhaseEquil:
No convergence for T
*******************************************************************************

Error in ThermoPhase/equilibrate (line 493)
            ctFunc('thermo_equilibrate', tp.tpID, xy, solver, rtol, ...
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in equil (line 39)
        gas.equilibrate('HP');
        ^^^^^^^^^^^^^^^^^^^^^
```
while this error will appear in the linux terminal:
```
Intel oneMKL ERROR: Parameter 6 was incorrect on entry to DGEMV
```
[...]

PR #1888, opened 2025-05-20
Experimental clib .NET

ischoegl:

This PR refactors `sourcegen` for clarity

- Split `HeaderGenerator` from `CLibSourceGenerator` (which accounts for a large portion of the diff)
- Use `clib_experimental` rather than the traditional `clib` for .NET
- Add `samples/clib_experimental/demo.c` example

For the `dotnet build` step, the current approach uses a Doxygen tree that is saved as an artifact.

A secondary refactoring step to make the file structure more Pythonic will be addressed in a follow-up.

Addresses Cantera/enhancements#220

**Checklist**


ischoegl:
@speth ... this is ready for a review. A large portion of the line count is due to splitting the header generation from the CLib source generation. There are a couple of loose ends still, but I am trying to keep individual PR's manageable. As an aside: both failing tests appear to be due to a MATLAB regression.

interfaces/dotnet/Cantera/src/Interop/InteropUtil.cs
```diff
@@ -28,11 +28,11 @@ public unsafe delegate int FillDoubleBufferFunc<THandle>(THandle handle, nuint s
...
     /// The Cantera C API is not very consistent on whether the size should be specified
-    /// as an int or a nuint (size_t), so users of this delegate may need to wrap their
-    /// LibCantera call in a lambda to perform the appropriate conversions,
-    /// and/or pass in the Cantera handle as a capture. For example:
+    /// as an int, so users of this delegate may need to wrap their LibCantera call in a
+    /// lambda to perform the appropriate conversions, and/or pass in the Cantera handle
+    /// as a capture. For example:
     /// <code>
-    /// (size, buffer) => kin_getType(_handle, (nuint) size, buffer)
+    /// (size, buffer) => kin_getType(_handle, size, buffer)
```
speth: Is this lack of consistency still an issue with the new CLib?

interfaces/dotnet/Cantera/src/SpeciesCollection.cs
```diff
@@ -250,30 +250,4 @@ int EnsureIndexOf(string name)
...
-    /// in the order in which they appear in this collection,
-    /// but does not normalize them.
-    /// </summary>
-    public void SetUnnormalizedMassFractions(double[] fractions)
-    {
-        var retval = LibCantera.thermo_setMassFractions(_handle,
-            (nuint) fractions.Length, fractions, InteropConsts.False);
-
-        InteropUtil.CheckReturn(retval);
-    }
```
speth, line 265: Does the "experimental" CLib have a way of setting unnormalized mass/mole fractions?

interfaces/sourcegen/sourcegen/_dataclasses.py
```diff
@@ -160,34 +161,76 @@ class CFunc(Func):
...
+        params = []
+        while len(lines) > 1:
+            line = lines.pop().strip()
+            if line in ["/*", "/**"]:
+                continue
+            if line.endswith("*/"):
+                break
+
+            line = line.lstrip("*").strip()
+            if not brief:
```
speth, line 202: I hadn't really noticed how this worked before, but it might be better for `from_snippet` to start from a more structured YAML input that's used to build the docstring, rather than going the other way around and doing an ad hoc parsing of a Doxygen-style docstring. For example, in addition to the `code` attribute in the YAML file, we'd also have strings for `brief` and returns, and lists for `params` and `uses`. If you want to toss this on a "do later" list rather than addressing in this PR, that's fine with me.

interfaces/sourcegen/sourcegen/clib/config.yaml
```diff
@@ -3,13 +3,23 @@
 # This file is part of Cantera. See License.txt in the top-level directory or
 # at https://cantera.org/license.txt for license and copyright information.
 
-# List of files to ignore entirely:
+# List of files to ignore entirely. Should be left empty.
 ignore_files: []
 
-# Dictionary of file names and list of functions to ignore.
-# Example: ctkin_auto.yaml: [phase]
+# Dictionary of file names and list of functions to ignore. Should be left empty.
```
speth: Maybe worth noting that these are useful for testing / debugging purposes. Otherwise, it makes one wonder why they exist at all.

interfaces/sourcegen/sourcegen/clib/config.yaml
```diff
@@ -3,13 +3,23 @@
...
 ignore_funcs: {}
 
+# **************************************************************************************
+#  Note: Cross-walks are still hard-coded, as a prior version used them for the
+#  generation of CLib headers (now moved to HeaderGenerator).
```
speth: I don't understand this comment.

interfaces/sourcegen/sourcegen/csharp/config.yaml
```diff
@@ -1,13 +1,33 @@
...
+  char*: byte*
+
+# Parameter type cross-walks
+prop_type_crosswalk:
+  byte*: string
+  double*: double[]
+
+# **************************************************************************************
+#  Note: Some of the following cross-walks can be automated based on available header
+#  file information.
```
speth: Is this a `@todo` item?

interfaces/sourcegen/sourcegen/headers/generator.py
```diff
@@ -0,0 +1,339 @@
...
+
+    def resolve_tags(self, headers_files: list[HeaderFile]) -> None:
+        """Resolve recipe information based on doxygen tags."""
+        def get_bases() -> tuple[list[str], list[str]]:
+            bases = set()
+            classes = set()
+            for headers in headers_files:
+                bases |= {headers.base}
+                for recipe in headers.recipes:
+                    classes |= set(recipe.bases)
```
speth: When the items being added aren't already sets, I think it would be better to use the simpler methods:
```suggestion
                bases.add(headers.base)
                for recipe in headers.recipes:
                    classes.update(recipe.bases)
```

speth (changes requested):
Thanks, @ischoegl. I think this looks good. I had just a few questions / comments for your consideration before merging.

interfaces/dotnet/Cantera/src/Interop/InteropUtil.cs
ischoegl: Good catch! The new CLib only uses regular integers. Most of the remark can be removed.

interfaces/dotnet/Cantera/src/SpeciesCollection.cs
ischoegl: It does not have this at the moment. I anticipate that this will come up again in the context of MATLAB ... I am trying to keep some mode switches out of the interface implementation. Also, while I know that it is a needed function, I view it as a relatively odd implementation detail. If we can avoid exposing this in user-facing API's, I'd be in favor of alternatives.

interfaces/sourcegen/sourcegen/_dataclasses.py
ischoegl: Good point. It started to be a parser for function signatures within the Doxygen XML tree (which remains essential), but kept growing as I adopted it first for the 'reserved' methods, and finally - in this PR - for custom code (which was intermittently broken).

We have an obvious candidate for YAML syntax - i.e., what the YAML output generator produces. I left a note that this should be refactored.

interfaces/sourcegen/sourcegen/csharp/config.yaml
ischoegl: Yes. The .NET API could be automated to a much larger extent. I don't know about the long game here and consider the current implementation as a proof-of-concept. I don’t necessarily anticipate that this will be addressed prior to the 3.2 release.

interfaces/sourcegen/sourcegen/clib/config.yaml
ischoegl: Before the refactoring (split of `HeaderGenerator` from `CLibSourceGenerator`), there were cross-walks from C++ to CLib (for the header generation) and also the reverse CLib to C++ (for the CLib source generation). The former used these configurable options while the latter was hard-coded.
It is a 'todo' item that I'll tackle in a cleanup PR.

ischoegl:
@speth … thank you for the review. I either addressed your suggestions or left notes for follow-up work.

interfaces/dotnet/Cantera/src/Interop/InteropUtil.cs
```diff
@@ -52,20 +46,6 @@ public static double CheckReturn(double code)
...
-
-        if (code == error)
-        {
-            CanteraException.ThrowLatest();
-        }
-
-        return code;
-    }
-
     public static int CheckReturn(int code)
```
burkenyo, line 49: @ischoegl With your CLib changes, do we still need to account for error code -1 _and_ error code -999? Also do we still need `Math.Abs(code)`?

interfaces/dotnet/Cantera/src/Interop/InteropUtil.cs
```diff
@@ -12,28 +12,22 @@ static class InteropUtil
...
     /// </summary>
     /// <remarks>
-    /// The Cantera C API is not very consistent on whether the size should be specified
-    /// as an int or a nuint (size_t), so users of this delegate may need to wrap their
-    /// LibCantera call in a lambda to perform the appropriate conversions,
-    /// and/or pass in the Cantera handle as a capture. For example:
-    /// <code>
-    /// (size, buffer) => kin_getType(_handle, (nuint) size, buffer)
-    /// </code>
+    /// The Cantera C API specifies the size as an int.
```
burkenyo, line 30: @ischoegl In C#, the `int` type keyword always means 32-bit signed integer. Is this the same for the CLib’s definition of `int`?

interfaces/sourcegen/sourcegen/csharp/_CSharpSourceGenerator.py
```diff
@@ -5,41 +5,62 @@
...
-        getter = known_funcs.get(clib_area + "_" + c_name)
+        getter = known_funcs.get(clib_area + "3_" + c_name)
 
         if getter:
             # here we have found a simple scalar property
             prop_type = getter.ret_type
         else:
             # here we have found an array-like property (string, double[])
-            getter = known_funcs[clib_area + "_get" + c_name.capitalize()]
+            getter = known_funcs[clib_area + "3_get" + c_name.capitalize()]
```
burkenyo, line 63: I noticed the generated ThermoPhase class was missing all its setters. To fix this, apply the `3_set` naming scheme on line 69.

interfaces/sourcegen/sourcegen/csharp/templates.yaml
```diff
@@ -54,50 +38,3 @@ csharp-property-string: |-
           LibCantera.{{ setter }}(_handle, value));
       {%- endif %}
   }
-
```
burkenyo, line 57: Up on line 35 (in the `csharp-property-string` template), since you have standardized all the CLib functions to use `int` instead of `size_t`, remove this bit `({{ p_type }})` to get rid of an unnecessary cast.

interfaces/dotnet/Cantera/src/ThermoPhase.cs
```diff
@@ -13,10 +13,9 @@ public partial class ThermoPhase
     readonly SolutionHandle _sol;
 
     /// <summary>
-    /// Represents a func that sets a pair of thermo variables using a pointer
-    /// to a pair of doubles to stand in for a stack-allocated array with two elements.
+    /// Represents a func that sets a pair of thermo variables.
     /// </summary>
-    unsafe delegate int SetPairFunc(ThermoPhaseHandle n, (double, double)* values);
+    unsafe delegate int SetPairFunc(ThermoPhaseHandle n, double val0, double val1);
```
burkenyo: Now that your made the get/set pairs simpler, you can remove the `unsafe` keyword!

interfaces/dotnet/Cantera/src/ThermoPhase.cs
```diff
@@ -13,10 +13,9 @@ public partial class ThermoPhase
     readonly SolutionHandle _sol;
```
burkenyo, line 14: I would remove this, since we never access it. (In  .NET, the mere ownership of an object will not ensure it is cleaned up. That’s the job of the Garbage Collector, and for non-GC items like objects on the native Cantera side, we need to leverage the dispose pattern for deterministic release of the resources)

```suggestion
```

interfaces/dotnet/Cantera/src/ThermoPhase.cs
```diff
@@ -55,8 +54,8 @@ static ThermoPhase()
 
     internal ThermoPhase(string filename, string? phaseName)
     {
-        _sol = LibCantera.soln_newSolution(filename, phaseName ?? "", "none");
-        _handle = LibCantera.soln_thermo(_sol);
+        _sol = LibCantera.sol3_newSolution(filename, phaseName ?? "", "none");
+        _handle = LibCantera.sol3_thermo(_sol);
```
burkenyo, line 58: Until we add a proper `Solution` class, we only need to temporarily create the solution object on the native side (it also appears that it would be safe to destroy it once we extract the thermo object because it uses shared pointers). And to ensure the native solution object is released, we should dispose the handle by leveraging the `using` statement (also guaranteeing its release if an exception is thrown later in the constructor). Also, we need to call `sol.EnsureValid()` before using the handle to get the proper error message in case of a missing file.

```suggestion
        using var sol = LibCantera.sol3_newSolution(filename, phaseName ?? "", "none");
        sol.EnsureValid();
        _handle = LibCantera.sol3_thermo(sol);
``` 

(this assumes we remove the `_sol` field above)

interfaces/dotnet/Cantera/src/SpeciesCollection.cs
```diff
@@ -80,23 +80,23 @@ internal unsafe SpeciesCollection(ThermoPhaseHandle handle)
     {
         _handle = handle;
 
-        var count = (int) InteropUtil.CheckReturn(LibCantera.thermo_nSpecies(handle));
+        var count = (int) InteropUtil.CheckReturn(LibCantera.thermo3_nSpecies(handle));
```
burkenyo: Since you have standardized on `int` in CLib, you can remove the cast.

```suggestion
        var count = InteropUtil.CheckReturn(LibCantera.thermo3_nSpecies(handle));
```

burkenyo (commented):
@ischoegl Thanks for keeping the .NET interface code generation and hand-built portions updated! I’ve added a few comments below. Addtionally, I would recommend removing the `.h` part of the generated file names since we are no longer using the CLib header files to inform the C# Code generation. For example `Interop.Handles.Reaction.g.cs` or `Interop.Handles.ctrxn.g.cs` instead of `Interop.Handles.ctrxn.h.g.cs`

interfaces/dotnet/Cantera/src/Interop/InteropUtil.cs
ischoegl: Ah - that’s an interesting question. I believe there’s some uncertainty here as it’s not standardized. We may have to use all `int32_t` - exactly 32 bits. For the MATLAB interface, we have a recent 64 bit issue (see #1852), so it may make sense to go directly to `int64_t` - exactly 64 bits. So for C#, it may actually make sense to move to `long` if we want to avoid using different compile options. I wasn’t necessarily thinking of tackling this in this PR, but it will make things simpler going forward!

burkenyo: I agree that `int32_t` or `int64_t` in CLib would be ideal for the explicitness and we can definitely move to `Int64` in .NET in the future. For now, it appears we’re “lucky” because the 32- and 64-bit architectures upon which Cantera runs agree that `int` in C means 32-bits (see table here which also has the note about 32-bit architectures https://en.m.wikipedia.org/wiki/64-bit_computing#64-bit_data_models)

interfaces/dotnet/Cantera/src/Interop/InteropUtil.cs
ischoegl: I have not tampered with this logic - currently, `clib_experimental` uses the same magic numbers as the old `clib` - these are defined in `include/cantera/clib/clib_defs.h`.

interfaces/dotnet/Cantera/src/ThermoPhase.cs
ischoegl: I'm a little uncomfortable deleting this object. The way the CLib `Solution` destructor is implemented, it will clean up associated `ThermoPhase` objects. I am not sufficiently familiar with the C# garbage collection to be sure that there aren't unintended side effects. Fwiw, here's the generated CLib destructor:
```C++
    int sol3_del(int handle)
    {
        // destructor
        try {
            auto obj = SolutionCabinet::at(handle);
            // remove all associated objects in reversed order
            if (obj->transport()) {
                int index = TransportCabinet::index(*(obj->transport()), handle);
                if (index >= 0) {
                    TransportCabinet::del(index);
                }
            }
            if (obj->kinetics()) {
                int index = KineticsCabinet::index(*(obj->kinetics()), handle);
                if (index >= 0) {
                    KineticsCabinet::del(index);
                }
            }
            if (obj->thermo()) {
                int index = ThermoPhaseCabinet::index(*(obj->thermo()), handle);
                if (index >= 0) {
                    ThermoPhaseCabinet::del(index);
                }
            }
            SolutionCabinet::del(handle);
            return 0;
        } catch (...) {
        }
    }
```

interfaces/dotnet/Cantera/src/ThermoPhase.cs
ischoegl: See above: I'm not sufficiently familiar - this may introduce side effects that go beyond my wheelhouse.

_Edit:_ As .NET would access the storage in CLib, I'm pretty certain that `_sol` needs to remain. The `ThermoPhase` handles will be gone as soon as the `Solution` object is destroyed.

ischoegl:
Hi @burkenyo ... I truly appreciate you looking over the C# code - while I was apparently able to keep things from breaking, I don't really have much prior exposure to C#.

Regarding the 32 and 64 bit types, I am planning to dedicate a PR to a migration to `int64_t` (which is necessary in view of #1852) along with a clean-up of CLib cross-walks that I commented on earlier. 

Other than that: I didn't touch the error logic (some of that comes out of CLib, some of that is ancient C++ code), and I am fairly certain that the `Solution` object does need to persist to ensure that we don't inadvertently get rid of associated `ThermoPhase` objects.

ischoegl:
> @ischoegl Thanks for keeping the .NET interface code generation and hand-built portions updated! I’ve added a few comments below. Addtionally, I would recommend removing the `.h` part of the generated file names since we are no longer using the CLib header files to inform the C# Code generation. For example `Interop.Handles.Reaction.g.cs` or `Interop.Handles.ctrxn.g.cs` instead of `Interop.Handles.ctrxn.h.g.cs`

Regarding the ‘.h’ …. Technically, there aren’t any headers parsed but things are still based on header information (they are not written explicitly but are implicit to the implementation). I certainly can rename though.

interfaces/dotnet/Cantera/src/SpeciesCollection.cs
speth: I think we will need to provide this function. It's essential for users who want to use their own integrators.

interfaces/dotnet/Cantera/src/ThermoPhase.cs
burkenyo: @ischoegl I realized the handle handling is messed up in a number of ways. For example, right now when a `ThermoPhaseHandle` is disposed in .NET it calls `thermo3_del(handle)`. But `thermo3_del` takes no arguments! I am surprised this doesn’t crash. Same problem for `KineticsHandle` and `TransportHandle`.

So I think we should merge this now, and I’ll create a follow-up to fix the these handles. In truth `ThermoPhaseHandle`, `KineticsHandle`, and `TransportHandle`  shouldn’t be treated as such because they don’t identify an owned resource which can be closed independently. It will require some thinking on how the handle classes are scaffolded and which ones to scaffold at all, and how the wrapper classes are scaffolded.

burkenyo:
> > @ischoegl Thanks for keeping the .NET interface code generation and hand-built portions updated! I’ve added a few comments below. Addtionally, I would recommend removing the `.h` part of the generated file names since we are no longer using the CLib header files to inform the C# Code generation. For example `Interop.Handles.Reaction.g.cs` or `Interop.Handles.ctrxn.g.cs` instead of `Interop.Handles.ctrxn.h.g.cs`
> 
> Regarding the ‘.h’ …. Technically, there aren’t any headers parsed but things are still based on header information (they are not written explicitly but are implicit to the implementation). I certainly can rename though.

Let's keep the filenames for now so we can merge this PR. :)

interfaces/dotnet/Cantera/src/ThermoPhase.cs
burkenyo: Keep it for now, see comment below!

speth (approved):
Thanks, @ischoegl, This looks good and takes care of most of what's needed to align the .NET interface with the new CLib. I think @burkenyo has some ideas on how to further clean up the .NET interface and resolve a couple of rough edges that he'll bring in a separate PR.

PR #1894, opened 2025-05-28
Fix handles in .NET interface

### LINKED ISSUE #1892: Incorrect handling of handle ownership in dotnet interface
**Problem description**

While reviewing #1888, I realized the handling of certain Cantera handles is messed up in a number of ways. For example, right now when a `ThermoPhaseHandle` is disposed in .NET it calls `thermo3_del(handle)`. But `thermo3_del` takes no arguments! I am surprised this doesn’t crash. Same problem for KineticsHandle and TransportHandle.

`ThermoPhaseHandle`, `KineticsHandle`, and `TransportHandle` shouldn’t be treated as such because they don’t identify an owned resource which can be closed independently. Instead the .NET side needs to mirror the ownership model in CLib, in which `ThermoPhase`, `Kinetics`, and `Transport` object are owned by a `Solution` object. This will require some thinking on how the handle classes are scaffolded and which ones to scaffold at all.

burkenyo:

Using `SafeHandle` as the base class for `CanteraHandle` objects was convenient because the runtime has deep support for them in interop. However, this is inappropriate because `SafeHandle` is always marshalled as pointer-sized int. We just got lucky the the calling convention and endianness allowed P/Invoke calls to work as expected.

This PR:
- Adds a new hand-built implementation of `CanteraHandle`.
- Switches to source generated P/Invoke signatures (available in .NET 7+) with a custom marshaller for the handles.
- Takes advantage of the source generator’s first-class support for `[ReadOnly]Span<T>` to remove pointers from P/Invoke signatures and usages of the `unsafe` keyword and `fixed` blocks.
- Removes support for .NET Standard 2.0

(note: the P/Invoke source generator is a part of the .NET SDK and separate from `sourcegen`.)

Fixes #1892 

**Checklist**


ischoegl (commented):
Thanks, @burkenyo! While this all looks good from my side, I’m not sufficiently familiar with C# to give meaningful feedback. I can approve once tests succeed, but will otherwise take a backseat..

burkenyo:
> Thanks, @burkenyo! While this all looks good from my side, I’m not sufficiently familiar with C# to give meaningful feedback. I can approve once tests succeed, but will otherwise take a backseat..

Hi @ischoegl, the best option for now is to rely on the tests which sadly are underdeveloped. This interop is pretty low level and many C# developers won’t have much experience with it. That said, once we sort this out the thrust with the C# interface should switch to the higher-level idiomatic wrapper classes, which should be easier for more people to contribute.

ischoegl:
@burkenyo ... ~the failing macOS runner can be fixed by adding `graphviz` to the 'brew install` list (I believe)~ _Edit: that was a false positive_. I'm not sure why this broke all of a sudden. I am currently suppressing the Doxygen HTML output over at #1897.

Follow-up: Doxygen 1.14.0 came out a couple of days ago. Apparently, they fixed a bug. In `doc/doxygen/Doxyfile` look for 
```
QT_AUTOBRIEF           = NO
```
And set to `YES`. The existing Cantera documentation assumes this behavior.

ischoegl (approved):
Thanks! @burkenyo! This all looks good to me.

PR #1903, opened 2025-06-05
Fix CLib Doxygen Rendering

### LINKED ISSUE #1902: Parts of CLib Doxygen output do not render
**Problem description**

Now that #1897 is merged, it becomes apparent that the Doxygen `@implements` and `@relates` tags do not render as expected (while they are a semantic fit for `sourcegen`, they have specific purposes in Doxygen). While the implements output mostly renders, it is slightly garbled and doesn't resolve the expected link to the C++ method. The relates output is missing.

Curiously, some functions are missing entirely (e.g., `thermo_getMolecularWeights`) … _PS: while testing different Doxygen versions, I was not able to reproduce this locally; fixes in #1903 appear to be make it reappear in the articacts._

**Steps to reproduce**

See newly available [CLib Reference](https://cantera.org/dev/cxx/dd/d3e/group__clibGroup.html)

**Proposed fix**

Switch to different attribute names (preferably but not necessarily Doxygen tags) and update the header template so it renders and creates the expected link. Some of the `sourcegen` nomenclature should change accordingly so everything remains consistent.

A prototype for a fix is in [ischoegl/cantera:fix-clib-doxygen](https://github.com/ischoegl/cantera/tree/fix-clib-doxygen)

ischoegl:

As the Doxygen tag `@implements` does not render information as expected, it would be misleading to retain `implements` as an attribute of sourcegen data classes.

- Rename `Recipe.implements` to `Recipe.wraps`
- Rename `Func.implements` to `Func.wraps`
- Fix rendering issues on Doxygen html and address minor glitches and improvements
- In nomenclature, avoid 'modules' that really are 'libraries' (especially in view of C++20), and differentiate all CLib Doxygen tags with the prefix `CAPI` (while I would have preferred to include underscores they appear to get mangled)
- Add missing dependencies for `post-merge-tests`

Fwiw, the [Doxygen documentation](https://www.doxygen.nl/manual/commands.html#cmdimplements) for `\implements` appeared to be ok as it states:
> This command can be used to manually indicate an inheritance relation, when the programming language does not support this concept natively (e.g. C).

The documentation further points to [manual.c](https://github.com/doxygen/doxygen/blob/master/examples/manual.c), but `\implements` no longer shows up.

Fixes #1902

image

**Checklist**


ischoegl:
After double-checking the CI artifacts, this should take care of all the glitches remaining after #1897 (which had quite a few moving parts).

PS: also adding update to `post-merge-tests`. Some of those are expected to fail (especially after updating the pre-release Python to 3.14)

interfaces/sourcegen/src/sourcegen/yaml/templates.yaml
```diff
@@ -17,8 +17,8 @@ yaml-definition: |-
     {% if c_func.returns %}
     {{ 'returns: ' + c_func.returns }}
     {% endif %}
-    {% if implements %}
-    implements: {{ implements }}
+    {% if wraps %}
+    wraps: {{ wraps }}
```
speth, line 21: I couldn't quite follow how this ends up as an enumerated list in the output, but given that `wraps` is always a single item, could we get it formatted in the output as just
```
Wraps C++ getter: `size_t Phase::nSpecies()`
```
instead of 
```
Wraps C++ getter:
- `size_t Phase::nSpecies()`
```
to make things a little more compact (particularly in the Doxygen HTML output, which is already pretty inefficient at using vertical space).

speth (commented):
> In nomenclature, avoid 'modules' that really are 'libraries' (especially in view of C++20), and differentiate all CLib Doxygen tags with the prefix CAPI (while I would have preferred to include underscores they appear to get mangled)

Nomenclature is hard, but I'm not sure "libraries" is better, given all of is really part of one "library". And at least there's no meaning attached to "module" in plain C, though as you note there is in C++ now, and this will come up again when migrating the Fortran interface to use this `sourcegen` infrastructure. So :shrug:.

ischoegl:
Thanks @speth - that was easy enough! (I still rebased to check one last thing: all good)

Regarding 'modules': yes, there isn't great terminology overall, but at least it sidesteps issues with recent C++ and Fortran. It's easy enough to revisit if there ever is a need.

interfaces/sourcegen/src/sourcegen/yaml/templates.yaml
ischoegl: Addressed in `clib/templates.yaml`.

speth (approved):

PR #1904, opened 2025-06-07
Simplify logger setup

speth:

- Make logging global rather than thread-specific. Thread-specific logging is difficult to initialize when the user doesn't
control thread creation, as in environments like .NET.
- Fix ownership handling of `Logger` objects. The `Application` instance takes ownership of these objects. Changing
the interface to use `unique_ptr` reflects this and prevents errors like the double-delete during application teardown that was previously occurring (as identified by @burkenyo).

**Checklist**

### DIFF
--- a/include/cantera/base/global.h
+++ b/include/cantera/base/global.h
@@ -147,7 +147,7 @@ bool usesHDF5();
 //! @addtogroup logGroup
 //! @{
 
-//! @copydoc Application::Messages::writelog(const string&)
+//! @copydoc Application::writelog(const string&)
 void writelog_direct(const string& msg);
 
 //! Write a message to the log only if loglevel > 0
@@ -307,7 +307,7 @@ bool legacy_rate_constants_used();
 
 // @} End of globalSettings group
 
-//! @copydoc Application::Messages::setLogger
+//! @copydoc Application::setLogger
 //! @ingroup logGroup
 void setLogger(Logger* logwriter);
 
--- a/include/cantera/base/global.h
+++ b/include/cantera/base/global.h
@@ -307,9 +307,16 @@ bool legacy_rate_constants_used();
 
 // @} End of globalSettings group
 
-//! @copydoc Application::setLogger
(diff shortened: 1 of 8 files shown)


burkenyo:
Some background here: In the .NET assembly, I want to use a module initializer to set the log callback. A module initializer ensures some initialization is run before any user code. In order to achieve this initialization previously, I was using a static constructor in the the `Application` class. This has two drawbacks:
1) All user access to Cantera through the .NET interface has to be routed through the `Application` class to ensure the static constructor is run.
2) The runtime has to emit runtime checks on each access of the `Application` class to ensure the static constructor has run. (This can be seen by the lack of the `Not so for simple static field initializers, which are prevented from referencing other fields. See https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1810)

The second one will likely only cause a minimal case in every-day use, but the first requires a very awkward library design.

When switching to a module initializer, I found that logging tests started failing. It turns out this was because the runtime was running the initializer on a different thread from the one the tests were running on. Thus, the logCallback was not getting invoked because it was only applied to the thread the on which the initializer run. By applying this commits on this PR, the test started working again.

burkenyo (approved):
Verified these changes with @speth

ischoegl (approved):
Thanks @speth ... this looks all good to me! 

While this barely touches the inconsistent `writelog` nomenclature (we use camelCase in related definitions), this isn't worth addressing.

ischoegl:
Fwiw, the PR appears to have broken the post-merge Cython test, although it’s never clear whether it’ll wash out as the pre-release progresses
https://github.com/Cantera/cantera/actions/runs/15519808401/job/43691507167

speth:
From my reading of that error log, I don't think it's anything in this PR that broke that test, but a change in the pre-release Cython. It is a bit concerning though, if there are changes in Cython that are going to break our somewhat non-standard practice of compiling multiple Cython source files and linking them as a single extension module.

PR #1907, opened 2025-06-08
Steady-state reactor network solver

speth:

- Generalize the 1D flame solver into a steady-state solver for generic ODE/DAE systems. This is the new `SteadyStateSolver` class, which is also used as a base class for `OneDim`.
- Use this solver to implement a new `ReactorNet.solveSteady` method as an alternative to the time-marching approach used in the Python-only `advance_to_steady_state` method.
- Make steady-state solver more robust by allowing it to continue in time-stepping mode after encountering certain errors while trying to solve the steady-state problem. These include both Jacobian factorization errors and errors evaluating the residual function with a "bad" state vector.

This solver isn't currently a full replacement for `advance_to_steady_state`, as there are some complicated interactions with equations that don't translate directly into the steady-state problem, implicit constraints that are automatically satisfied when integrating the normal ODEs but are not currently imposed in the steady solver, and possibly other mathematical issues that I haven't worked out yet.

A simple example of the first case is the equation for the volume of a control volume reactor, where the ODE reduces to
$dV/dt = 0$. This is fine for an ODE solver, but for the steady solver, it is a problem because there's no information that can be used to solve for $V$. Instead, this equation needs to be replaced with the algebraic equation $V(t) - V_0 = 0$. This is handled in the implementation here by a new `steadyConstraints` method of each reactor class that indicates which equations need to be transformed into such constraints.

Unfortunately, this only works in the case where the constraint corresponds to an existing equation. Other cases, such as the constraint that the mass of a constant pressure reactor should be constant work fine in the case of some formulations (`ConstPressureReactor`, where mass is one of the state variables) but not others (`ConstPressureMoleReactor`, where the mass is a derived quantity based on the moles of each species).

The cases where the solver does not work have been configured to raise exceptions when initializing the steady-state solver, and are indicated in docstring for the `solveSteady` method. I think we will eventually be able to resolve at least some of these limitations, though this will require a somewhat more flexible approach to modifying the set of equations to be solved.

- Supersedes #1021. 
- Completes https://github.com/Cantera/enhancements/issues/31

This solver seems to work particularly well for cases like [`combustor.py`](https://cantera.org/dev/examples/python/reactors/combustor.html), where the interest is in solving for a number of nearby steady-state conditions:
```py
residence_time = 0.1  # starting residence time
while combustor.T > 500:
    sim.initial_time = 0.0  # reset the integrator
    sim.solve_steady()
    print('tres = {:.2e}; T = {:.1f}'.format(residence_time, combustor.T))
    states.append(combustor.thermo.state, tres=residence_time)
    residence_time *= 0.9  # decrease the residence time for the next iteration
```

At least in this case, it is a drop-in replacement for `sim.advance_to_steady_state`.

**Checklist**


bryanwweber:
This is _so cool_ Ray, thanks for picking this work up and pushing it through 🎉🎉🎉🎉

doc/sphinx/develop/running-tests.md
```diff
@@ -399,3 +399,23 @@ message printed by any `CanteraError` exceptions, you can call the C++ method
 {ct}`CanteraError::setStackTraceDepth` with a depth of 20 to print the top 20 frames of
 the call stack, or similarly the Python method
 {py:meth}`CanteraError.set_stack_trace_depth`.
+
+## Diagnosing memory errors using Clang's AddressSanitizer
+
+The Clang compiler includes the
+[AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html) tool which
+enables checking for a range of for memory errors such as out-of-bounds array access and
```
bryanwweber:
```suggestion
enables checking for a range of memory errors such as out-of-bounds array access and
```

doc/sphinx/develop/running-tests.md
```diff
@@ -399,3 +399,23 @@ message printed by any `CanteraError` exceptions, you can call the C++ method
...
+using deleted/freed variables. To compile Cantera in debug mode with these checks
+enabled, run SCons with the following options:
+
+```sh
+scons build CC=clang CXX=clang++ env_vars=all debug_flags="-g -fsanitize=address" debug_linker_flags="-lprofiler -fsanitize=address -shared-libasan"
+```
+
+On macOS, running the Python test suite also requires the following:
+
+DYLD_INSERT_LIBRARIES=$(clang -print-file-name=libclang_rt.asan_osx_dynamic.dylib)
```
bryanwweber, line 417:
````suggestion
```
DYLD_INSERT_LIBRARIES=$(clang -print-file-name=libclang_rt.asan_osx_dynamic.dylib)
```
````

File: doc/sphinx/develop/running-tests.md [hunk 3]
```diff
@@ -399,3 +399,23 @@ message printed by any `CanteraError` exceptions, you can call the C++ method
...
+
+Or on Linux:
+
+LD_PRELOAD=$(clang -print-file-name=libclang_rt.asan-x86_64.so)
```
bryanwweber:
````suggestion
```
LD_PRELOAD=$(clang -print-file-name=libclang_rt.asan-x86_64.so)
```
````

File: include/cantera/numerics/SteadyStateSystem.h [hunk 4]
```diff
@@ -0,0 +1,335 @@
...
+    virtual double lowerBound(size_t i) const = 0;
+
+    //! Return a reference to the Newton iterator.
+    MultiNewton& newton();
+
+    //! Set the linear solver used to hold the Jacobian matrix and solve linear systems
+    //! as part of each Newton iteration.
+    void setLinearSolver(shared_ptr<SystemJacobian> solver);
+
+    //! Get the type of the linear solver being used.
```
bryanwweber: This looks like it's returning a reference to the actual solver, not the type. Is the documentation correct?

include/cantera/numerics/SteadyStateSystem.h
```diff
@@ -0,0 +1,335 @@
...
+    //! Array of number of steps to take after each unsuccessful steady-state solve
+    //! before re-attempting the steady-state solution. For subsequent time stepping
+    //! calls, the final value is reused. See setTimeStep().
+    vector<int> m_steps = { 10 };
+
+    double m_tstep = 1.0e-5; //!< Initial timestep
+    double m_tmin = 1e-16; //!< minimum timestep size
+    double m_tmax = 1e+08; //!< maximum timestep size
+
+    //! factor time step is multiplied by  if time stepping fails ( < 1 )
```
bryanwweber: I think we should be consistent about capitalizing the first word in the doc strings. Most places have the first word capitalized, but three of these don't :smile:

include/cantera/zeroD/ConstPressureMoleReactor.h
```diff
@@ -26,18 +26,26 @@ class ConstPressureMoleReactor : public MoleReactor
...
     void getState(double* y) override;
 
     void initialize(double t0=0.0) override;
 
     void eval(double t, double* LHS, double* RHS) override;
 
+    vector<size_t> steadyConstraints() const override {
+        throw CanteraError("ConstPressureMoleReactor::steadyConstraints",
+            "ConstPressureMoleReactor is not currently compatible with the steady-state"
+            " solver.");
```
bryanwweber: Should we create an enhancements issue to link to for people to indicate their support for this work?

include/cantera/zeroD/FlowReactor.h
```diff
@@ -48,6 +48,10 @@ class FlowReactor : public IdealGasReactor
     void evalDae(double t, double* y, double* ydot, double* residual) override;
 
     void getConstraints(double* constraints) override;
+    vector<size_t> steadyConstraints() const override {
+        throw CanteraError("FlowReactor::steadyConstraints",
+            "FlowReactor is not compatible with time-dependent steady-state solver.");
```
bryanwweber, line 53: I've not heard the phrase "time-dependent steady-state solver" before. Does that mean `advance_to_steady_state()` or the new solver here?

include/cantera/zeroD/ReactorNet.h
```diff
@@ -141,6 +143,46 @@ class ReactorNet : public FuncEval
...
+    //! steady-state, but imposes a few limitations:
+    //!
+    //! - The volume of control volume reactor types (such as Reactor and
+    //!   IdealGasMoleReactor) must be constant; no moving walls can be used.
+    //! - The mass of constant pressure reactor types (such as ConstPressureReactor and
+    //!   IdealGasConstPressureReactor) must be constant; if flow devices are used,
+    //!   inlet and outlet flows must be balanced.
+    //! - The solver is currently not compatible with the ConstPressureMoleReactor or
+    //!   IdealGasConstPressureMoleReactor classes.
+    //! - Only "ideal gas" reactor types can be used for when the energy equation is
```
bryanwweber: I don't think ideal gas needs to be in quotes here.
```suggestion
    //! - Only ideal gas reactor types can be used when the energy equation is
```

interfaces/cython/cantera/reactor.pyx
```diff
@@ -1589,6 +1590,57 @@ cdef class ReactorNet:
...
+        `advance_to_steady_state` method), but imposes a few limitations:
+
+        - The volume of control volume reactor types (such as `Reactor` and
+          `IdealGasMoleReactor`) must be constant; no moving walls can be used.
+        - The mass of constant pressure reactor types (such as `ConstPressureReactor`
+          and `IdealGasConstPressureReactor`) must be constant; if flow devices are
+          used, inlet and outlet flows must be balanced.
+        - The solver is currently not compatible with the `ConstPressureMoleReactor` or
+          `IdealGasConstPressureMoleReactor` classes.
+        - Only "ideal gas" reactor types can be used for when the energy equation is
```
bryanwweber:
```suggestion
        - Only ideal gas reactor types can be used when the energy equation is
```

src/numerics/SteadyStateSystem.cpp
```diff
@@ -0,0 +1,305 @@
...
+    return ss;
+}
+
+void SteadyStateSystem::setTimeStep(double stepsize, size_t n, const int* tsteps)
+{
+    m_tstep = stepsize;
+    m_steps.resize(n);
+    for (size_t i = 0; i < n; i++) {
+        m_steps[i] = tsteps[i];
+    }
```
bryanwweber, line 232: Is there a way to handle the case when `n != len(tsteps)`?

src/numerics/SteadyStateSystem.cpp
```diff
@@ -0,0 +1,305 @@
...
+    m_jac_ok = false;
+    setSteadyMode();
+    evalJacobian(x);
+    m_rdt = rdt_save;
+}
+
+void SteadyStateSystem::initTimeInteg(double dt, double* x)
+{
+    double rdt_old = m_rdt;
+    m_rdt = 1.0/dt;
```
bryanwweber, line 272: If `m_rdt` becomes NaN or inf (I don't remember what this type supports), how will that show up in the output? Is it possible for that to happen?

src/zeroD/ConstPressureReactor.cpp
```diff
@@ -157,4 +166,36 @@ string ConstPressureReactor::componentName(size_t k) {
...
+    }
+}
+
+double ConstPressureReactor::lowerBound(size_t k) const {
+    if (k == 0) {
+        return 0; // mass
+    } else if (k == 1) {
+        return -BigNumber; // enthalpy
+    } else if (k >= 2 && k < m_nv) {
+        return 0.0; // species mass fraction or surface coverage
```
bryanwweber: Should the two zeros here be `-Tiny`? If not, can you explain why not for me?

src/zeroD/Reactor.cpp
```diff
@@ -505,6 +519,40 @@ string Reactor::componentName(size_t k) {
...
+    } else {
+        throw CanteraError("Reactor::upperBound", "Index {} is out of bounds.", k);
+    }
+}
+
+double Reactor::lowerBound(size_t k) const {
+    if (k == 0) {
+        return 0; // mass
+    } else if (k == 1) {
+        return 0; // volume
```
bryanwweber, line 542: Same comment here about `-Tiny`

test/python/test_reactor.py
```diff
@@ -3110,3 +3122,153 @@ def deltaC():
...
+        ct.MassFlowController(upstream, r, mdot=160)
+        ct.MassFlowController(r, downstream, mdot=160)
+        net = ct.ReactorNet([r])
+        net.solve_steady()
+
+        assert r.mass == approx(m0)
+        assert r.thermo.T == approx(2407.35011)
+
+    @pytest.mark.parametrize("reactor_class",
+        [ct.IdealGasReactor, ct.IdealGasMoleReactor])
```
bryanwweber, line 3178: Is it worth adding some tests of the disallowed cases? On one hand it's more overhead, on the other it'd be straightforward to add and remove. :shrug:

bryanwweber (commented):
This was impressively few files changed to add a feature of this magnitude. Thank you again for taking it on!

include/cantera/numerics/SteadyStateSystem.h
ischoegl: The first line will show up as `\brief` in Doxygen (and subsequently in generated APIs via sourcegen). From that perspective, it should be short and to the point - essentially, I view it as the same as a git commit message summary. From that perspective,
```
//! factor time step is multiplied by  if time stepping fails ( < 1 )
```
lacks clarity.

Fwiw, I agree with improved consistency in terms of docstrings: I am aware of instances where current `\brief` descriptions are much longer than ideal (see [List of C++ Classes](https://cantera.org/dev/cxx/annotated.html) ... `HighPressureGasTransport` is one outlier that's especially long).

include/cantera/numerics/SteadyStateSystem.h
speth: Indeed, the docstring was incorrect.

include/cantera/numerics/SteadyStateSystem.h
speth: :+1: on consistent capitalization. Beyond that, let's just be clear that we're talking about the docstring for a `protected` member variable, so the documentation really only needs to be understood in the context where that variable is used (which, excluding the setter is one place!), and for that purpose I think this is reasonably clear. If a longer description is warranted, I'd suggest that belongs only with the `setTimeStepFactor` method. But even then, I'm not sure what else there is to say.

include/cantera/zeroD/FlowReactor.h
speth: I wasn't really sure what to call it. It's confusing because solving the `FlowReactor` is always a steady-state problem, just in a different dimension. But it's one where this solution method is not really applicable. I guess the alternative here is to just reference the method that will cause this exception to be raised, which is calling `ReactorNet.solveSteady`.

include/cantera/zeroD/ConstPressureMoleReactor.h
speth: Done. See https://github.com/Cantera/enhancements/issues/234.

src/numerics/SteadyStateSystem.cpp
speth: No. One of the standard challenges of working with C arrays is that there is no inherent indication of length. Note that while the `SteadyStateSystem` class is new, this is just the implementation of what was `OneDim::setTimeStep`, so the signature needs to remain as-is.

src/numerics/SteadyStateSystem.cpp
speth: If `m_rdt` becomes NaN, I think the solver will pretty much instantly fail -- any evaluation of the governing equations will become riddled in NaNs. I don't think it should become inf, since in the places where the timestep is adjusted, it's bounded by `m_tmin`.

src/zeroD/ConstPressureReactor.cpp
speth: I think you're right that individual mass fractions or coverages should be allowed to be small negative values. But allowing the mass or volume to be negative is going to wreak havoc on attempting to evaluate the equation of state.

src/numerics/SteadyStateSystem.cpp
speth: No. One of the standard challenges of working with C arrays is that there is no inherent indication of length. Note that while the `SteadyStateSystem` class is new, this is just the implementation of what was `OneDim::setTimeStep`, so the signature needs to remain as-is.

bryanwweber: Thanks. That's what I figured. Oh well! ☺️

src/numerics/SteadyStateSystem.cpp
bryanwweber: Thanks! You correctly surmised that my concern was for `dt` to become zero. And the solver raising an error is definitely fine in either case. 👍

src/zeroD/ConstPressureReactor.cpp
bryanwweber: Thanks! I fully believe you, but I want to understand a bit more. Why are the mass and volume different from the species? I guess then you'd end up with negative density which I know is a common error message we see on the UG. And I guess there's no "mass production equation" like there is for the species, so the constraints are just different. Is that anywhere close?

speth: We allow small negative values for mass fractions because there are things like very fast reactions that drive them to values that are so small that numerical error can result in them becoming negative. This is particularly true when using an implicit method for the reactor ODEs where the step is calculated as the solution to some linear system with a high condition number. Letting these small values remain doesn't have any ill effects on the rest of the system (once you're sufficiently careful about a few things, like reaction rates where both reactants have negative concentrations).

On the other hand, if you have something driving the mass or volume to zero (say, set an outlet mass flow rate greater than the inflow for a fixed volume reactor), the system is going to become ill-defined at some point. The solver will fail at this point, but that should be expected. And yes, if only one of these becomes negative, then you would get a negative density, which we explicitly block at the `Phase` level.

test/python/test_reactor.py
speth: I've added some (specifically, the ones with reactor surfaces), but I think doing all the cases would be a bit too much.

bryanwweber (approved):
Thanks again @speth!

PR #1908, opened 2025-06-11
Resolve some CI/build issues

speth:

- Prevent CI builds using broken Cython version 3.1.2
- Replace `windows-2019` runners with `windows-2025`
- Fix warnings about unused SUNDIALS callbacks
- Fix warnings about license classifiers being deprecated

**Checklist**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -794,15 +794,14 @@ jobs:
         retention-days: 2
 
   windows:
-    name: "Windows 2019, MSVC ${{ matrix.vs-toolset }}, Python ${{ matrix.python-version }}"
-    runs-on: windows-2019
+    name: "Windows 2025, MSVC ${{ matrix.vs-toolset }}, Python ${{ matrix.python-version }}"
+    runs-on: windows-2025
     timeout-minutes: 60
     env:
       BOOST_ROOT: ${{github.workspace}}/3rdparty/boost
       BOOST_URL: https://github.com/boostorg/boost/releases/download/boost-1.87.0/boost-1.87.0-b2-nodocs.7z
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


ischoegl (approved):
Thanks, @speth!

bryanwweber (approved):
This kind of PR makes me wonder when it makes sense to have a `requirements.txt` for developer dependencies. Thanks for doing this!

speth:
> This kind of PR makes me wonder when it makes sense to have a `requirements.txt` for developer dependencies.

I briefly started in this direction, but was dissuaded by a bunch of sources that suggested that requirements.txt is old and busted and that we ought to be using a pyproject.toml instead, and then thinking about the headache of doing that with something that isn't a Python project managed by one of the typical Python packaging systems.

bryanwweber:
My take is that requirements.txt is old and busted for use cases where you need to have consistent versioning, for library or application dependencies primarily. I agree pyproject.toml is not what we want. I think using requirements.txt for our case is more than appropriate and I'd be shocked if `[uv] pip install -r` stopped working any time in the near future. 🤷‍♂️

ischoegl:
I don't have strong opinions, but believe that having a sufficiently spread-out test matrix on the CI should be adequate? (I'm not sure that all of us have to have the exact same toolchain.) As long as we test 3-4 versions plus bleeding edge on critical infrastructure we should be good. We mostly are already doing this ... the only one missing is the recent additional dependency on doxygen (which could be made optional if users don't care about CLib).

bryanwweber:
The idea for `requirements.txt` or similar would mostly help in defining package versions for CI. Instead of having to change every job when we need to change a version or add/remove a dependency, it'd be just one (or a few) file(s) that need to change.

PR #1909, opened 2025-06-17
Fix Windows MSVC job in CI

ssun30:
This PR fixes failures in the Windows 2025 CI jobs caused by:
1. Python 3.13.4 linking error: `python313t.lib` was missing due to a [known bug](https://github.com/python/cpython/issues/135151) in that version. 
2. Wrong compiler: Windows Server 2025 comes shipped with MinGW and MSVC. If the MSVC environment variables are not explicitly set, MinGW is used by default. We previously didn't catch this problem because the first issue wasn't present and the job completed despite not using the wrong compiler.  

- Updated the Python setup step to use the latest build which solved the 1st issue. 
- Added a [step](https://github.com/ilammy/msvc-dev-cmd) to initialize the MSVC environment.

**Checklist**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -820,10 +820,6 @@ jobs:
         run: |
           python -m pip install -U pip setuptools wheel
           python -m pip install 'scons<4.4.0' pypiwin32 numpy ruamel.yaml cython!=3.1.2 pandas graphviz pytest pytest-xdist pytest-github-actions-annotate-failures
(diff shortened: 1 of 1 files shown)


speth (commented):
Thanks for looking into this, @ssun30. Rather than relying on this new action to set up the MSVC environment, you can also tell SCons to use the correct compiler by adding `toolchain=msvc` to the `scons build...` command line.

speth (approved):

Issue #1722, opened 2024-06-22
Broken flame examples in experimental MATLAB toolbox


ischoegl:
**Problem description**

`flame1.m`, `flame2.m` and `diff_flame.m` examples no longer work: the first fails to converge, and temperatures of the other two are equal to inlet temperatures.

**Steps to reproduce**

1. Install Cantera from scratch
2. Follow instructions in `interfaces/matlab_experimental/readme.md`
3. Run one of the flame examples

**Behavior**

No combustion is observed

**System information**

- Cantera version: latest `main`
- OS: macOS
- Python/MATLAB/other software versions: MATLAB R2023b

**Other context**

Issue was discovered when testing changes in #1619, but same behavior was observed on current `main`.

speth:
Using the experimental Matlab toolbox from `main` with `libcantera` from 3.0.0 (conda-forge package), I observe the same problems with all three examples.

Using the experimental Matlab toolbox from `v3.0.0` with `libcantera` from 3.0.0 (conda-forge package), `flame1.m` and `flame2.m` exhibit the same problems. However, `diff_flame.m` successfully solves a non-extinct flame.

ischoegl:
Thanks for confirming, @speth. I’m surprised that this didn’t work as far back as when we released 3.0. Tagging @ssun30 and @rwest …

rwest:
We'll take a look this week. I think some MATLAB toolbox work is in our docket anyway. Thanks for tagging us.

speth:
`git bisect` tells me that the commit that broke `diff_flame.m` is this one: 3a9bc6a66d10e40ce9194542701c54b8b60e612b.

ischoegl:
@rwest and @ssun30 ... please make sure to rebase after #1619 is merged, as that PR introduces some updates to nomenclature, which affect `clib` (it actually adopts MATLAB nomenclature for `Flow1D` within C++, as it makes more sense than the old `StFlow`)

PS: #1619 is now merged.

RRijvi:
Hi! I was trying to run flame2.m (Cantera 3.0.0) in my Matlab interface. The "solve(flame, loglevel, refine_grid);" function isn't working for my case. Could that be a problem with my software?

Thanks!

ischoegl:
> Hi! I was trying to run flame2.m (Cantera 3.0.0) in my Matlab interface. The "solve(flame, loglevel, refine_grid);" function isn't working for my case. Could that be a problem with my software?

[...]

PR #1911, opened 2025-06-18
Fix MATLAB interface samples

ssun30:
There are numerous changes to class structures to the MATLAB interface that necessitate modifications to our current collection of MATLAB samples. 

- Fix crashes/solver errors in MATLAB samples.
- Compare the results to their equivalent Python samples to make sure the solutions are correct. 

~~Not a fix to https://github.com/Cantera/cantera/issues/1722~~ Flame fixes have been deferred to a later PR. 

**Checklist**


ischoegl (commented):
@ssun30 ... thanks for addressing!

Out of curiosity: for the Python API, all examples are tested as part of a CI job, see `run-examples` in `.github/workflows/main.yml`.


The relevant code is:
```YAML
   - name: Run the examples
        # See https://unix.stackexchange.com/a/392973 for an explanation of the -exec part.
        run: |
          ln -s libcantera_shared.so build/lib/libcantera_shared.so.3
          export LD_LIBRARY_PATH=build/lib
          find samples/python -type f -iname "*.py" \
```

I believe it's possible to do something similar here as it's possible to run MATLAB scripts from the command line:
```
matlab -r "try, run('rankine.m'), catch, exit(1), end, exit(0)"
```
where I had set `matlab` to
```
alias matlab="/Applications/MATLAB_R2023b.app/bin/matlab -nojvm -nodesktop"
```
on macOS (fwiw, I just ran a quick test so the options may not be the greatest selection)

interfaces/matlab_experimental/OneDim/Sim1D.m
```diff
@@ -58,14 +58,9 @@ function delete(s)
 
         %% Sim1D Utility Methods
 
-        function display(s, fname)
+        function display(s)
```
ischoegl: For the sake of consistency with other APIs, the name of the method should be `show`, see https://cantera.org/stable/python/onedim.html#cantera.Sim1D.show.

interfaces/matlab_experimental/Func/Func1.m
```diff
@@ -137,7 +137,9 @@
 
         function delete(f)
             % Delete the :mat:class:`Func1` object.
-            ctFunc('func_del', f.id);
+            if isprop(f, 'id') && ~isempty(f.id)
```
ischoegl, line 140: Why is this check needed? Isn't it the case that as long as the constructor call is successful, `f.id` does exist?

interfaces/matlab_experimental/OneDim/Sim1D.m
ischoegl: Please address my previous comment.

interfaces/matlab_experimental/Utility/ctPaths.m
```diff
@@ -0,0 +1,73 @@
+function paths = ctPaths(varargin)
+    % ctPaths ::
+    % Configure or retrieve the library/header/toolbox paths for Cantera.
+    % The paths are stored as MATLAB preferences.
+    %
+    %   >> ctPaths()                       % Get current config as struct
+    %   >> ctPaths(libPath, includePath)   % Set library/header/toolbox paths
+    %   >> ctPaths('clear')                % Clear saved paths
```
ischoegl: Is this `clear` option needed? I am asking as the interface has more than one use for the first positional argument, which isn't good practice. As an alternative, you could either use a stand-alone function `ctClear` or require keyword/value pairs for all of these.

interfaces/matlab_experimental/Utility/ctPaths.m
```diff
@@ -0,0 +1,73 @@
...
+            currentPaths = strsplit(path, pathsep);
+            subdirsToRemove = intersect(subDirs, currentPaths);
+            if ~isempty(subdirsToRemove)
+                rmpath(subdirsToRemove{:});
+            end
+            rmpref('Cantera', 'Paths');
+        end
+        paths = struct('libPath', '', 'includePath', '', 'toolboxPath', '');
+        return
+    elseif nargin == 3 && all(cellfun(@ischar, varargin))
```
ischoegl: This option is not documented, or the documentation on top of the file is incorrect. I'd strongly recommend using the known relative paths (either for compiled Cantera or the future MLTBX version) rather than requiring the user to provide three paths (see also: comment in installation instructions).

interfaces/matlab_experimental/readme.md
```diff
@@ -22,13 +22,18 @@ function calling from Cantera CLib.
...
+6. The built-in Utility function `ctPaths` to set the following paths for the toolbox:
+   - `toolboxPath`: If you compiled Cantera from source, this is the location of your
+   Cantera source code. If you installed the MATLAB Toolbox from FileExchange, this is
+   the location of your installed Toolbox.
+   - `libpath`: This is the location of the Cantera shared library. If you compiled
+   Cantera from source, this is usually `fullfile(toolboxPath, 'build', 'lib')`.
+   If you installed from FileExchange, this is `fullfile(toolboxPath, 'dependencies')`.
+   - `includePath`: This is the location of the header files used by Cantera.
+   If you compiled Cantera from source, this is `fullfile(toolboxPath, 'include')`.
+   If you installed from FileExchange, this is `fullfile(toolboxPath, 'dependencies')`.
```
ischoegl: It appears that you're asking the user to provide information that is actually known a priori - wouldn't it be sufficient to simply indicate the root folder `toolboxPath` in the `ctPaths` API?

Also, I think that for compiled Cantera, you should use the **'installed' location**, rather than the `build` path. I.e., the location of shared libraries after running `scons install`, see for example what I get on my macOS system:
```
*************** Cantera 3.2.0a2 has been successfully installed ****************

File locations:

  library files               /opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/lib
  C++ headers                 /opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/include
  samples                     /opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/share/cantera/samples
  data files                  /opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/share/cantera/data
  input file converters       /opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/bin
  Python package              /opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/lib/python3.11/site-packages
  Python examples             /opt/homebrew/Caskroom/miniforge/base/envs/cantera-dev/share/cantera/samples/python

********************************************************************************
```

As an aside: for testing purposes, the `build` folder is the appropriate version; the `samples` should be copied over to the installed location; unit tests should probably not be packaged with the MLTBX (we don't distribute them with Python either).

interfaces/matlab_experimental/readme.md
```diff
@@ -39,4 +44,5 @@ function calling from Cantera CLib.
    then run the script `runMatlabInterfaceTests.m`.
 4. To stop using the new Cantera interface, run the following commands:
    `ctCleanUp`
-   `ctUnload`
+   `ctUnload`.
+5. To remove the Cantera paths from MATLAB preferences, run `ctPaths('clear')`.
```
ischoegl: Preferences are cleared automatically unless a user has run `savepath`, correct?

.github/workflows/main.yml
```diff
@@ -66,6 +66,15 @@ jobs:
...
         uses: matlab-actions/setup-matlab@718d4320188c73c86eb94ce76b553cbf89778487 # v2.5.0
+      - name: Set MATLAB search paths for tests and build MLTBX toolbox # v2.2.1
+        uses: matlab-actions/run-command@67f012f1ee4bc1627a7a95a0ccdeec745c9f6f36
+        with:
+          command: |
+            toolboxPath = fullfile(getenv('CANTERA_ROOT'));
+            libPath = fullfile(toolboxPath, 'build', 'lib');
+            includePath = fullfile(toolboxPath, 'include');
+            addpath(fullfile(toolboxPath, 'interfaces', 'matlab_experimental', 'Utility'));
+            ctPaths(libPath, includePath, toolboxPath);
```
ischoegl: After this, you should run the `test_examples.m` script (after running `scons install`).

interfaces/matlab_experimental/Utility/ctPaths.m
```diff
@@ -0,0 +1,73 @@
...
+    if any(cellfun(@isempty, {paths.libPath, paths.includePath, paths.toolboxPath}))
+        error('ctPaths:MissingPath', ...
+              'Library, header, and toolbox paths must be configured with ctPaths(libPath,includePath, toolboxPath).');
+    end
+
+    mapping = {
+        'interfaces/matlab_experimental', 'toolbox';
+        'samples/matlab_experimental', 'samples';
+        'test/matlab_experimental', 'test/matlab_toolbox';
+        'test/data', 'test/data';
```
ischoegl: Some comments to minimize MATLAB path / namespace pollution:
* Sample locations should not be put on the MATLAB path automatically - this decision should be left to the user.
* Unit tests should not be packaged with MLTBX, and likewise, not be added to the MATLAB path.

ischoegl (dismissed):
@ssun30 ... thanks for pushing the MATLAB toolbox closer to the finishing line. Changes to the samples are ok, but I do have comments regarding the updated approach for folder management:

* The `ctPaths` interface is too complicated - it should be sufficient to provide the root location, and use logic based on known relative paths
* Except for unit tests, compiled Cantera should use the installed location
* Unit tests don't need to be installed or distributed with the MLTBX

At this juncture, it also makes sense to start testing all operating systems in CI jobs, i.e., Windows, macOS, and Linux. This can likely be done as a standard matrix job. Further, examples should be run as part of the CI jobs as pointed out in a comment above.

interfaces/matlab_experimental/OneDim/Sim1D.m
bryanwweber: `display()` is a built-in Matlab command, I think... So either this should be renamed as suggested, or this provides the implementation for the built-in capabilities. Either way, it should be addressed 😀

.github/workflows/main.yml
ssun30: I've added a separate unit test `ctTestSamples`, instead of running this separately. I'm keeping `test_examples.m` for now in case the user wish to use it.

interfaces/matlab_experimental/readme.md
ssun30: The preferences set by `ctPaths` are persistent across different MATLAB sessions, but the search paths for the `.m` files are automatically cleared. I've changed `savepath` to be the default behavior of `ctPaths`.

interfaces/matlab_experimental/Func/Func1.m
ssun30: Currently it's possible to have a partially constructed `Func1` object. For example:
```
                    if ~isa(arg1, 'Func1') || ~isa(arg2, 'Func1')
                        error('Invalid arguments for compounding functor')
                    end
                    x.id = ctFunc('func_new_compound', typ, arg1.id, arg2.id);
```
If an exception is thrown, `x.id` will be empty and the destructor won't run properly. This issue can be replicated by running `ctTestFunc1.testCompound` (before this fix).

ischoegl: Thanks for this example. But this means that an exception that is thrown in the C++ constructor isn’t handled correctly by the MATLAB interface? Imho, the correct approach would be to check for error codes in the constructor and rethrow as needed. Otherwise, a user assumes that everything is ok but the underlying C++ object does not exist, leading to hard-to-debug issues.

.github/workflows/main.yml
ischoegl: I dislike the `test_examples.m` personally but rest my case.

interfaces/matlab_experimental/readme.md
ischoegl: I disagree with the approach of automatically ‘polluting’ the MATLAB search path. It should be left to the user to decide whether they want this.

interfaces/matlab_experimental/Utility/ctPaths.m
```diff
@@ -0,0 +1,81 @@
...
+
+            if isfield(savedPaths, 'toolboxPath') && isfolder(savedPaths.toolboxPath)
+                rmpath(genpath(savedPaths.toolboxPath));
+            end
+
+            if isfield(savedPaths, 'samplePath') && isfolder(savedPaths.samplePath)
+                rmpath(genpath(savedPaths.samplePath));
+            end
+
+            savepath;
```
ischoegl: This `savepath` decision should imho be left to the user.

interfaces/matlab_experimental/readme.md
```diff
@@ -5,30 +5,25 @@ function calling from Cantera CLib.
...
-6. In the Matlab command window, run
-   `cd([pwd, '/interfaces/matlab_experimental/Utility'])` to navigate to the Utility
-   folder.
-7. Open the file named 'ctRoot.m', in the second line, edit `output=` to
-   `output='/path/to/conda/environment'`, then save the file. This sets the search path
-   for the `ctLoad` command to find the shared library file for Cantera.
-8. In the Matlab command window, run `savepath` to save all search paths.
+6. Run the built-in Utility function `ctPaths('Set', envPath)` to set search paths
+   for the Toolbox:
+   - `envPath`: this is the location of the Conda environment where the CLib and header
```
ischoegl, line 27: What if the user installed via pip? If current instructions are limited to conda environments, then the instructions should state this.

interfaces/matlab_experimental/Utility/ctPaths.m
```diff
@@ -0,0 +1,81 @@
...
+            samplePath = fullfile(envRoot, 'share', 'cantera', 'samples');
+            dataPath   = fullfile(envRoot, 'share', 'cantera', 'data');
+
+            thisFile    = mfilename('fullpath');
+            toolboxPath = fileparts(fileparts(thisFile));
+
+            paths = struct( ...
+                'libPath',     libPath, ...
+                'includePath', includePath, ...
+                'samplePath',  samplePath, ...
```
ischoegl: I am still not convinced that samples should be on the path. Please note that the current SCons setup won't copy MATLAB samples to those locations (!), so this won't work.

ischoegl (commented):
Thanks @ssun30. Thanks for addressing the `ctPaths` issue which to me was the biggest concern. There are some remaining issues that should be easy to address.

I'm :+1: with deferring some decisions regarding installation and CI runs, but hope that you'll address that feedback in other PRs. Both should be addressed before the 'experimental' suffix can be removed.

interfaces/matlab_experimental/readme.md
speth: Pip only installs the Python package, not a full version of the Cantera library with header files. Beyond that, though, I think some of what is done here currently is specific to the Conda directory layout, and won't work for a locally-copied version of Cantera (at least not on all platforms).

ischoegl: As written in my summary, I'm :+1: with deferring some of of the installation minutiae; at the same time, if this is deferred it should be clarified that things only work for conda at the moment.

interfaces/matlab_experimental/OneDim/Sim1D.m
ssun30: > `display()` is a built-in Matlab command, I think... So either this should be renamed as suggested, or this provides the implementation for the built-in capabilities. Either way, it should be addressed 😀

Other classes such as `ThermoPhase`, `Mixture`, and `Func1` use their own `display` method to override MATLAB's default. Should I change these as well?

interfaces/matlab_experimental/Func/Func1.m
ssun30: > Thanks for this example. But this means that an exception that is thrown in the C++ constructor isn’t handled correctly by the MATLAB interface? Imho, the correct approach would be to check for error codes in the constructor and rethrow as needed. Otherwise, a user assumes that everything is ok but the underlying C++ object does not exist, leading to hard-to-debug issues.

In the provided example, the C++ constructor `func_new_compound` is not called at all since an error is already thrown before it. MATLAB then throws the MATLAB error as well as a warning about the destructor failing (due to the missing id). The check in the destructor is to suppress this kind of warning.   

Exceptions thrown in the C++ constructor are still handled correctly. For example if an invalid id was passed onto the constructor `ctFunc('func_new_compound', 'sum', f0.id, f1.id)` where `f1` was already deleted, the following exception is thrown:
```
CanteraError thrown by Cabinet::at:
Object with index 8 has been deleted.
```

interfaces/matlab_experimental/OneDim/Sim1D.m
speth: I would leave the others as they are for now. I think it's good for the default output of just typing `sim` where `sim` is a `Sim1D` object to just have a short output like:
```
sim = 

  CounterFlowDiffusionFlame with properties:

       stID: 0
    domains: {[1×1 Inlet]  [1×1 AxisymmetricFlow]  [1×1 Inlet]}
```
rather than printing the value of every state variable at every point. In contrast, the summary printed by a `Solution` object is usually more compact, so makes sense to use as an override for Matlab's default implementation of `display`.

samples/matlab_experimental/conhp.m
```diff
@@ -6,14 +6,14 @@
     % It assumes that the ``gas`` object represents a reacting ideal gas mixture.
 
     % Set the state of the gas, based on the current solution vector.
+    gas.basis = 'mass';
```
speth, line 9: The default basis should already be "mass", which would be consistent with the Python interface, though I see that the behavior defined in `ThermoPhase.m` is in fact to default to molar properties.

speth (changes requested):
I think there's a fair bit of improvement still needed with how paths are set up, but I would suggest that we hold off on that for now in the interest of getting the main feature of this PR in place, which is to have the Matlab CI job run successfully and include running the examples as part of the testing in that job.

samples/matlab_experimental/isentropic.m
```diff
@@ -1,5 +1,5 @@
 function isentropic(g)
-    %% Isentropic, adiabatic flow
+    %% Adiabatic, isentropic flow
```
bryanwweber: I know this is rephrasing the existing text, but isentropic flow is adiabatic by definition, so including both seems redundant. Maybe isentropic nozzle flow or something similar is good.

samples/matlab_experimental/rankine.m
```diff
@@ -10,35 +10,39 @@
...
 p_max = 8.0 * OneAtm;
 t1 = 300.0;
 
-% create an object representing water
+%% create an object representing water
 w = Water;
 
+%%
 % start with saturated liquid water at t1
 basis = 'mass';
```
bryanwweber, line 21: This line doesn't appear to do anything. Should we remove it?

samples/matlab_experimental/rankine.m
```diff
@@ -10,35 +10,39 @@
...
 w.TQ = {t1, 0};
 h1 = w.H;
 p1 = w.P;
 
+%%
 % pump it to p2
```
bryanwweber: In most cases you're adding the cell marker to the same line as the rest of the comment for a single line comment. Here and elsewhere you're adding a new line. Can it be consistent please?

samples/matlab_experimental/test_examples.m
```diff
@@ -1,31 +1,52 @@
...
-
-disp('Test example run successful.')
+function success = test_examples()
+    ctLoad
+
+    examples = {
+        'equil', 'isentropic', 'reactor1', 'reactor2', 'surf_reactor', ...
+        'periodic_cstr', 'plug_flow_reactor', 'lithium_ion_battery', ...
+        'rankine', 'prandtl1', 'prandtl2', 'catcomb','ignite', 'diamond_cvd', ...
+        % 'flame1', 'flame2', 'diff_flame', 'ignite_hp', 'ignite_uv'
```
bryanwweber: The last two examples here should be included as they're zero-d reactors, right?

samples/matlab_experimental/test_examples.m
```diff
@@ -1,31 +1,52 @@
...
+
+    if ~isempty(failed)
+        fprintf('❌ Failed: %s\n', strjoin(failed, ', '));
+    else
+        disp('❌ Failed: (none)');
+    end
+    ctUnload
+
+    disp('Test example run complete.');
+end
```
bryanwweber: Please add a blank newline to the end of this file. Note the message from GitHub in the diff view.

test/matlab_experimental/ctTestSamples.m
```diff
@@ -0,0 +1,48 @@
+classdef ctTestSamples < ctTestCase
```
bryanwweber, line 1: What's the difference between this file and the previous one? They both appear to run the samples for testing purposes

bryanwweber (changes requested):
I'm likewise interested in resolving the installation question, but I also feel fine with deferring that to after this work. I just had a few simple fixes for consistency and maybe some cleanup.

samples/matlab_experimental/test_examples.m
ischoegl: With the 1D flame examples exempt, is it still true that thie PR closes #1722 (as indicated at the very top)?

test/matlab_experimental/ctTestSamples.m
ssun30: They serve the same purpose. I originally planned to run the sample tests as a separate job from the unit test suite, similar to the `run-examples` job for the Python samples. But for now I will just incorporate it into the unit tests.

samples/matlab_experimental/conhp.m
ischoegl: I agree that default behaviors across interfaces should be consistent. The mass/mole switching is unfortunately handled by the interfaces, see https://github.com/Cantera/enhancements/issues/219. Obviously not something that should be addressed here.

bryanwweber (dismissed):
Thanks @ssun30 this looks good to me now

ischoegl (dismissed):
Thanks, @ssun30! While some issues remain to be resolved in follow-up PRs, this is progress.

speth (approved):
This is definitely progress in the right direction. Looking forward to more in additional PRs.

PR #1912, opened 2025-06-20
Revert "[Python/Test] Work around pint TypeError - pint #1969 resolved"

### LINKED ISSUE #1525: Make pint optional when running tests
**Problem description**

The use of the Pint units library within the Cantera Python module is optional. However, attempting to run the Python test suite without Pint installed fails.

This is dissimilar to how we handle testing of other optional dependencies in Python, where we are able to use the `@pytest.mark.skipif` decorator.

**Steps to reproduce**

1. Set up a Cantera build environment including all Python dependencies, but not including Pint
2. Run `scons test`

**Behavior**

```
___________________________________________ ERROR collecting test_units.py ___________________________________________
ImportError while importing test module '/root/cantera/test/python/test_units.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test/python/test_units.py:7: in <module>
    import cantera.with_units as ctu
build/python/cantera/with_units/__init__.py:7: in <module>
    from pint import UnitRegistry, set_application_registry
E   ModuleNotFoundError: No module named 'pint'
```

**System information**

- Cantera version: 3.0.0b1
- OS: Debian 11 (Bullseye)
- Python/MATLAB/other software versions: Python 3.9

mefuller:
This reverts commit dd03a35ca724b06c6f64029c8c31d44437efe87e.

Revert commit made as workaround until [pint#1969](https://github.com/hgrecco/pint/issues/1969) resolved (it is resolved)

closes #1525 (already closed, but superior)

**Checklist**

### DIFF
--- a/test/python/test_units.py
+++ b/test/python/test_units.py
@@ -1,14 +1,10 @@
+from contextlib import nullcontext
 from dataclasses import dataclass
 from typing import Optional, Tuple, Dict
+import sys
 
 import pytest
-try:
-    pint = pytest.importorskip("pint", "0.17.0")
-except TypeError as e:
-    # The extra exception handling is here because pint is incompatible with
-    # Python 3.13. Once https://github.com/hgrecco/pint/issues/1969 is resolved the
-    # try/except here can be restored to just the importorskip.
-    pytest.skip(f"pint import failed due to {e}", allow_module_level=True)
+pytest.importorskip("pint", "0.17.0")
 import cantera.with_units as ctu
 import cantera as ct
 try:


test/python/test_units.py
```diff
@@ -1,14 +1,10 @@
...
 
 import pytest
-try:
-    pint = pytest.importorskip("pint", "0.17.0")
-except TypeError as e:
-    # The extra exception handling is here because pint is incompatible with
-    # Python 3.13. Once https://github.com/hgrecco/pint/issues/1969 is resolved the
-    # try/except here can be restored to just the importorskip.
-    pytest.skip(f"pint import failed due to {e}", allow_module_level=True)
+pytest.importorskip("pint", "0.17.0")
```
speth, line 7: I think this should not be a pure "revert" -- the removal of unused imports that was done in the original commit should be preserved. I also think the sense of "revert" is a bit wrong here; at least to me it implies that the original commit was erroneous, which isn't the case.

speth (changes requested):
Thanks for catching that this has been resolved upstream, @mefuller. Just needs a bit of cleanup.

test/python/test_units.py
bryanwweber: In addition to that change, I think we need to set the minimum version here as the patch that fixed this bug, otherwise there's a (admittedly small) chance that an incompatible version could be installed.

speth: If so, that minimum version should be dependent on the Python version. For Python 3.10-3.12, whatever this new version is isn't required.

bryanwweber: Hmm, that's a good point and complicates this somewhat, probably more than it's worth. Especially since this is just in the test suite. The place where it might make sense to add Python-version-specific Pint version ranges would be in conda and/or the wheels, since the old versions of Pint can't be updated to say they're incompatible with 3.13. Probably still not worth it, though, for this low-probability event.

ischoegl:
Superseded by #1960.

Issue #1843, opened 2025-01-26
One of the checks for sundials doesn't support sundials built with MPI support


rdbisme:
**Problem description**
I am the maintainer of the `cantera-git` AUR package, and I was trying to restore support to use system installed `sundials`.  If I add `sundials` and not `sundials-seq` as dependency, your check [here](https://github.com/Cantera/cantera/blob/main/site_scons/buildutils.py#L1754) doesn't work, complaining about missing openmpi symbols. 

I tried to check `scons` `CheckLibWithHeader`. It doesn't support providing `extra_libs`, while the underlying `CheckLibs` it uses, does. I think you need to pass `-libmpi` to the compiler command... 

**Steps to reproduce**

1. Open '...'
2. Run '....'
3. See error '....'

**Behavior**

**System information**

- Cantera version: [9c3a57c](https://github.com/Cantera/cantera/commit/9c3a57c8be15131c1a994c077d87c324400580a3)
- OS: Archlinux
- Python/MATLAB/other software versions:

**Attachments**

**Additional context**

rdbisme:
If anyone more proficient than me in `scons` knows how to fix that... I need to pass `-lmpi` to `CheckLibWithHeader`.

rdbisme:
I've created this: https://github.com/SCons/scons/pull/4676

speth:
Isn't this an issue with the AUR [1] build of SUNDIALS? If the `sundials_core` library has additional dependencies, those should be indicated by the `sundials_core` library. Downstream users like Cantera shouldn't have to explicitly link to all the dependencies of their first-tier dependencies. For instance, on Ubuntu, the `libsundials_nvecmpimanyvector.so` library indicates linkage to the MPI library:
```
$  ldd /usr/lib/x86_64-linux-gnu/libsundials_nvecmpimanyvector.so
        linux-vdso.so.1 (0x00007fff331a1000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f24d9509000)
        libmpi.so.40 => /lib/x86_64-linux-gnu/libmpi.so.40 (0x00007f24d93d7000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f24d91b7000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f24d9624000)
        libopen-rte.so.40 => /lib/x86_64-linux-gnu/libopen-rte.so.40 (0x00007f24d90fb000)
[...]

PR #1915, opened 2025-06-22
Fix building sdist into wheels and avoid broken Cython

bryanwweber:

- **[Cython] Enforce cython cannot be v3.1.2**
- **[Sdist] Template requirements versions**

Building wheels is currently broken due to the regression fixed for CI in #1908 here and cython/cython#6957 in Cython. Since this version will always be broken, we should enforce in the build systems that it is incompatible.

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1628,7 +1628,7 @@ env['python_cmd_esc'] = quoted(env['python_cmd'])
 env["python_min_version"] = python_min_version
 env["python_max_version"] = python_max_version
 env["py_requires_ver_str"] = py_requires_ver_str
-env["cython_version_spec"] = SpecifierSet(">=0.29.31", prereleases=True)
+env["cython_version_spec"] = SpecifierSet(">=0.29.31,!=3.1.2", prereleases=True)
 # When updating NumPy spec, also update interfaces/python_sdist/pyproject.toml.in.
 env["numpy_version_spec"] = SpecifierSet(">=1.21.0,<3", prereleases=True)
 env["cython_version_spec_str"] = str(env["cython_version_spec"])
--- a/SConstruct
+++ b/SConstruct
@@ -181,6 +181,10 @@ python_max_version = parse_version("3.14")
 # The string is used to set python_requires in setup.cfg.in
 py_requires_ver_str = f">={python_min_version},<{python_max_version}"
 
+cython_version_spec = SpecifierSet(">=0.29.31,!=3.1.2", prereleases=True)
+numpy_version_spec = SpecifierSet(">=1.21.0,<3", prereleases=True)
+ruamel_version_spec = SpecifierSet(">=0.17.21,<1", prereleases=True)
+
 if "sdist" in COMMAND_LINE_TARGETS:
     if "clean" in COMMAND_LINE_TARGETS:
         COMMAND_LINE_TARGETS.remove("clean")
@@ -198,6 +202,9 @@ if "sdist" in COMMAND_LINE_TARGETS:
             py_requires_ver_str,
             cantera_version,
             cantera_short_version,
+            str(cython_version_spec),
+            str(numpy_version_spec),
+            str(ruamel_version_spec),
         ],
         check=True,
     )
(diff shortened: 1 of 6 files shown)


SConstruct
```diff
@@ -181,6 +181,10 @@ python_max_version = parse_version("3.14")
 # The string is used to set python_requires in setup.cfg.in
 py_requires_ver_str = f">={python_min_version},<{python_max_version}"
 
+cython_version_spec = SpecifierSet(">=0.29.31,!=3.1.2", prereleases=True)
```
ischoegl, line 184: Could you add explanatory comments as is done elsewhere? It’s not immediately clear why some versions have maximum version restrictions, which presumably should be relaxed at some point in the future?

ischoegl (commented):
Thanks for addressing, @bryanwweber! Apart from the minor comment, could you have a look at the post-merge-tests that are currently failing due to the same Cython issue?

bryanwweber:
Proof that the changes work to prevent using this broken Cython with Scons: https://github.com/Cantera/cantera/actions/runs/15809549593/job/44558991192?pr=1915#step:7:375 :tada: 

```
ERROR: Cython is an incompatible version: Found 3.1.2 but !=3.1.2,>=0.29.31 is required.
```

Now I just have to fix that it wasn't supposed to do that :disappointed:

bryanwweber:
Any ideas how to resolve this error when building on `ubuntu:rolling` or `ubuntu:devel`?

https://github.com/Cantera/cantera/actions/runs/15809741771/job/44559424078?pr=1915#step:7:252

```
g++ -o .sconf_temp/conftest_ce2dd1e38acb22ad88224c4d2c5f1d43_0.o -c -std=c++17 -D_GLIBCXX_ASSERTIONS -pthread -O3 -Wno-inline -DNDEBUG .sconf_temp/conftest_ce2dd1e38acb22ad88224c4d2c5f1d43_0.cpp
In file included from /usr/include/sundials/sundials_direct.h:23,
                 from /usr/include/cvodes/cvodes_ls.h:21,
                 from /usr/include/cvodes/cvodes.h:20,
                 from .sconf_temp/conftest_ce2dd1e38acb22ad88224c4d2c5f1d43_0.cpp:3:
/usr/include/sundials/sundials_types.h:57:10: fatal error: mpi.h: No such file or directory
   57 | #include <mpi.h>
      |          ^~~~~~~
compilation terminated.
scons: Configure: no
```

I think it's this error: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1082085 but I don't recall if it's come up here before.

SConstruct
bryanwweber: Numpy and ruamel.yaml have maximum version caps here, but they're unchanged from before this edit. In those cases, those are the next major versions, so it seems obvious why we'd want to restrict that.

As for the minimum versions, I'm not sure how those are set. The comment about `ruamel.yaml` from the original location isn't correct (it refers to 0.17.16, but the minimum version is 0.17.21) and neither Cython nor Numpy had a comment.

Finally, I'm not sure if we want to add a comment about why `!=3.1.2` is needed for Cython, or let the git history handle that. I'm inclined to leave it only in the git history so it doesn't cause clutter here. In principle, we can't ever remove the restriction since that version will always cause a problem...

ischoegl: That’s fine, although I have a slight preference for explicit comments (git commit messages are not as obvious)

ischoegl:
> Any ideas how to resolve this error when building on `ubuntu:rolling` or `ubuntu:devel`?
> 
> https://github.com/Cantera/cantera/actions/runs/15809741771/job/44559424078?pr=1915#step:7:252
> 
> ```
> g++ -o .sconf_temp/conftest_ce2dd1e38acb22ad88224c4d2c5f1d43_0.o -c -std=c++17 -D_GLIBCXX_ASSERTIONS -pthread -O3 -Wno-inline -DNDEBUG .sconf_temp/conftest_ce2dd1e38acb22ad88224c4d2c5f1d43_0.cpp
> In file included from /usr/include/sundials/sundials_direct.h:23,
>                  from /usr/include/cvodes/cvodes_ls.h:21,
>                  from /usr/include/cvodes/cvodes.h:20,
>                  from .sconf_temp/conftest_ce2dd1e38acb22ad88224c4d2c5f1d43_0.cpp:3:
> /usr/include/sundials/sundials_types.h:57:10: fatal error: mpi.h: No such file or directory
>    57 | #include <mpi.h>
>       |          ^~~~~~~
> compilation terminated.
> scons: Configure: no
> ```
> 
> I think it's this error: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1082085 but I don't recall if it's come up here before.

I have the suspicion that it’s related to #1843. It’s obviously unrelated to Cython …

ci_support/requirements.txt
ischoegl, line 1: Not sure that we should start a new folder here. I’d rather keep this in the `.github/workflows` location, and use a more intuitive file name than the generic `requirements.txt`? Of course, this assumes that the folder is available to GH runners.

bryanwweber: I think it makes sense to rename the file, but I don't think it belongs in `.github`. Those files are used specifically for GitHub configuration, whereas this is more generic.

Obviously, this file also doesn't have all the dependencies we use in other places, I wanted to defer that and see how this worked out before committing to all of that work.

I also need to leave some organizational comments in this file.

bryanwweber:
> I have the suspicion that it’s related to #1843. It’s obviously unrelated to Cython …

Yes thank you, I knew I'd seen it here before!

Edit: it looks like Ray fixed it with the `mpi*` compiler wrappers, so I'll do that too.

ci_support/requirements.txt
speth: What about putting this under `platform/ci_support`? The `platform` directory already contains some scripts that are only useful under certain circumstances, and would avoid polluting the root of the repository.

bryanwweber: That's a great idea, thanks Ray!

bryanwweber:
> Edit: it looks like Ray fixed it with the `mpi*` compiler wrappers, so I'll do that too.

After talking to Ingmar offline, I think we need a little more discussion about how to handle this changed dependency requirement, instead of just forcing `mpi*` compilers. I'm going to defer the work to a separate PR to resolve #1843 specifically

ischoegl (approved):
Thanks, @bryanwweber !

speth:
> > Edit: it looks like Ray fixed it with the `mpi*` compiler wrappers, so I'll do that too.
> 
> After talking to Ingmar offline, I think we need a little more discussion about how to handle this changed dependency requirement, instead of just forcing `mpi*` compilers. I'm going to defer the work to a separate PR to resolve #1843 specifically

I think we probably need to start a discussion upstream with the SUNDIALS devs or the Ubuntu package maintainers. There should be some way of making MPI optional so it's not necessary to compile a serial program with the MPI compiler wrapper.

Issue #1916, opened 2025-06-24
Sporadic test failures on CI runners using macOS on Intel hardware


speth:
**Problem description**

I've recently seen a couple instances of CI jobs failing with the error:
```
test/python/test_reaction.py::TestBlowersMaselRate::test_derivative_ddT 
[gw2] node down: Not properly terminated
[gw2] [ 54%] FAILED test/python/test_reaction.py::TestBlowersMaselRate::test_derivative_ddT 

replacing crashed worker gw2
```

See full logs [here](https://github.com/Cantera/cantera/actions/runs/15840383622/job/44651822366?pr=1907#logs).

I'm guessing this is some sort of out-of-bounds memory access issue that only sometimes causes problems.

**Steps to reproduce**

- Push branch to CI
- Wait and see

**System information**

- Last seen on the macos-13 with Python 3.13 runner

**Attachments**

**Additional context**

ischoegl:
There are other sporadic failures on Python 3.13 / macOS that are unrelated to TestBlowersMaselRate also, see https://github.com/Cantera/cantera/pull/1918#issuecomment-3006697405. Reproducing here ...

> One runner crashed with:
> ```
> [gw2] darwin -- Python 3.13.5 /Library/Frameworks/Python.framework/Versions/3.13/bin/python3
> worker 'gw2' crashed while running 'test_reactor.py::TestWellStirredReactorIgnition::test_solve_steady[MoleReactor]'
> ```
> and another one crashed with:
> ```
> [gw2] darwin -- Python 3.13.5 /Library/Frameworks/Python.framework/Versions/3.13/bin/python3
> worker 'gw2' crashed while running 'test_reactor.py::TestIdealGasMoleReactor::test_advance_with_limits'
> ```
> 
> The same happened during the merge of #1907, see https://github.com/Cantera/cantera/actions/runs/15888865696/job/44807273791
> 
> Another crashing runner is on #1911, see https://github.com/Cantera/cantera/actions/runs/15890473825/job/44811946486?pr=1911

I had not seen these failures until recently, and it is curious that they only occur on macOS 13 with Python 3.13 - fwiw, it's still the Intel version with a very recent Python version.

Fwiw, I am adding the corresponding section from the log referenced on top:
```
[gw2] darwin -- Python 3.13.5 /Library/Frameworks/Python.framework/Versions/3.13/bin/python3
[...]

PR #1918, opened 2025-06-24
Add sundials mpi check

ischoegl:

Create a more meaningful error message if Sundials is compiled with MPI support but `mpi.h` is not found.

~I am hesitant to fix the post-merge tests via~
```
scons build CC=mpicc CXX=mpicxx env_vars=all
```
~though, as there is still some hope that Debian maintainers will reconsider the MPI default - relinking https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1082085~

* _Post-merge tests are fixed_
* _Fixed issue for `FORTRAN=mpifort`_

Relates to #1843

~See failing "post-merge" tests.~ _Edit: no longer failing_

**Checklist**

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1582,22 +1582,41 @@ end program main
 
 set_fortran("{}FLAGS", env["FORTRANFLAGS"])
 
-if env["using_apple_clang"] and env["f90_interface"] == "default":
+if (env["using_apple_clang"] and env["f90_interface"] == "default" and
+    not env["FORTRAN"]):
     env["f90_interface"] = "n"
 
 if env['f90_interface'] in ('y','default'):
     foundF90 = False
-    if env['FORTRAN']:
+    fortran_used = ""
+    if "mpifort" in env["FORTRAN"]:
+        foundF90 = check_fortran(env["FORTRAN"], True)
+        fortran_used = get_command_output("mpifort", "--show").split()[0]
+    elif env["FORTRAN"]:
         foundF90 = check_fortran(env['FORTRAN'], True)
+        fortran_used = env["FORTRAN"]
 
     for compiler in ("pgfortran", "gfortran", "ifort", "ifx", "g95"):
         if foundF90:
             break
+        fortran_used = compiler
         foundF90 = check_fortran(compiler)
 
     if foundF90:
         logger.info(f"Using {env['FORTRAN']!r} to build the Fortran 90 interface")
         env['f90_interface'] = 'y'
+
+        if "pgfortran" in fortran_used:
+            env["FORTRANMODDIRPREFIX"] = "-module "
+        elif "gfortran" in fortran_used:
+            env["FORTRANMODDIRPREFIX"] = "-J"
+        elif "g95" in fortran_used:
+            env["FORTRANMODDIRPREFIX"] = "-fmod="
+        elif "ifort" in fortran_used:
+            env["FORTRANMODDIRPREFIX"] = "-module "
+        elif "ifx" in fortran_used:
+            env["FORTRANMODDIRPREFIX"] = "-module "
+
     else:
         if env['f90_interface'] == 'y':
             logger.error("Could not find a suitable Fortran compiler to build the Fortran 90 interface.")
@@ -1607,20 +1626,9 @@ if env['f90_interface'] in ('y','default'):
             env['FORTRAN'] = ''
             logger.info("Skipping compilation of the Fortran 90 interface.")
 
-if 'pgfortran' in env['FORTRAN']:
-    env['FORTRANMODDIRPREFIX'] = '-module '
-elif 'gfortran' in env['FORTRAN']:
-    env['FORTRANMODDIRPREFIX'] = '-J'
-elif 'g95' in env['FORTRAN']:
-    env['FORTRANMODDIRPREFIX'] = '-fmod='
-elif 'ifort' in env['FORTRAN']:
-    env['FORTRANMODDIRPREFIX'] = '-module '
-elif "ifx" in env["FORTRAN"]:
-    env["FORTRANMODDIRPREFIX"] = "-module "
-
-set_fortran("{}", env["FORTRAN"])
-set_fortran("SH{}", env["FORTRAN"])
-env['FORTRANMODDIR'] = '${TARGET.dir}'
+    set_fortran("{}", env["FORTRAN"])
+    set_fortran("SH{}", env["FORTRAN"])
+    env["FORTRANMODDIR"] = "${TARGET.dir}"
 
 env = conf.Finish()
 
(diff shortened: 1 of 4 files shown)


SConstruct
```diff
@@ -1430,6 +1430,13 @@ env['LIBM'] = ['m'] if env['NEED_LIBM'] else []
 
 
 if env['system_sundials'] in ("y", "default"):
+    retcode, uses_mpi = run_preprocessor(
+        conf, ['"sundials/sundials_config.h"'], "SUNDIALS_MPI_ENABLED")
+    if retcode and uses_mpi == "1":
+        if not conf.CheckCXXHeader('mpi.h', '""'):
+            logger.error("System Sundials was compiled with MPI support, but header "
+                         "'mpi.h' was not found.")
+            sys.exit(1)
```
speth: I think it's worth suggesting the two possible resolutions here; most users won't know what to do with "'mpi.h' was not found". Namely:
- add `CC=mpicc CXX=mpicxx` to the `scons build` command line
- Use the bundled version of SUNDIALS by specifying the `system_sundials=n` option to SCons.

ischoegl: Updated. Now we have the following:
```
ERROR: System installation of Sundials version 7.1.1 has MPI support enabled,
but the required 'mpi.h' header file cannot be found. To resolve this issue,
choose one of the following options when running the 'scons build' command:
- Add 'CC=mpicc CXX=mpicxx' to use MPI compiler wrappers, or
- Add 'system_sundials=n' to disable the system Sundials installation.
```

ischoegl: PS: had to fix a regression as the check only applies to Sundials>=7.0.0. Before that, there aren't any issues.

speth:
Should we apply the fix of using`mpicc`/`mpicxx` for the post-merge tests? Otherwise, these are going to fail indefinitely, which doesn't seem all that useful.

ischoegl:
> Should we apply the fix of using`mpicc`/`mpicxx` for the post-merge tests? Otherwise, these are going to fail indefinitely, which doesn't seem all that useful.

Fair. Reading the Debian message, I'm not too hopeful that there's any change in their approach. 

I ended up moving the check into `buildutils.py` so we have a fallback if `system_sundials=default`; both resolution versions are now covered in "post-merge" tests. 

_PS: had to disable the f90 interface for MPI as there were some issues:_
```
gfortran -o build/samples/f90/demo -pthread -Wl,-rpath=/usr/local/lib build/samples/f90/demo.o -Lbuild/lib -lcantera_fortran -lcantera -lsundials_cvodes -lsundials_idas -lsundials_nvecserial -lsundials_core -ldl -lhdf5_serial -lfmt -lyaml-cpp -lstdc++
/usr/bin/ld: build/lib/libcantera.a(CVodesIntegrator.os): undefined reference to symbol 'ompi_mpi_comm_null'
/usr/bin/ld: /lib/x86_64-linux-gnu/libmpi.so.40: error adding symbols: DSO missing from command line
```
_which is the same as reported in #1843. Tried adding `FORTRAN=mpifort` but that wasn’t resolved correctly:_
```
mpifort -o build/src/fortran/fct_interface.os -c -O3 -fPIC -Ibuild/src/fortran -Isrc/fortran -Isrc/fortran -module build/src/fortran src/fortran/fct_interface.f90
gfortran: error: unrecognized command-line option ‘-module’; did you mean ‘-Mmodules’?
```

speth:
> _PS: had to disable the f90 interface for MPI as there were some issues:_
> 
> ```
> gfortran -o build/samples/f90/demo -pthread -Wl,-rpath=/usr/local/lib build/samples/f90/demo.o -Lbuild/lib -lcantera_fortran -lcantera -lsundials_cvodes -lsundials_idas -lsundials_nvecserial -lsundials_core -ldl -lhdf5_serial -lfmt -lyaml-cpp -lstdc++
> /usr/bin/ld: build/lib/libcantera.a(CVodesIntegrator.os): undefined reference to symbol 'ompi_mpi_comm_null'
> /usr/bin/ld: /lib/x86_64-linux-gnu/libmpi.so.40: error adding symbols: DSO missing from command line
> ```
> 
> _which is the same as reported in #1843. Tried adding `FORTRAN=mpifort` but that wasn’t resolved correctly:_
> 
> ```
> mpifort -o build/src/fortran/fct_interface.os -c -O3 -fPIC -Ibuild/src/fortran -Isrc/fortran -Isrc/fortran -module build/src/fortran src/fortran/fct_interface.f90
> gfortran: error: unrecognized command-line option ‘-module’; did you mean ‘-Mmodules’?
> ```

Did you try `FORTRAN=mpif90`? I think `mpifort` might be assuming it's supposed to be an F77 compiler, where "modules" have no meaning. If that doesn't work, disabling this seems fine, too.

ischoegl:
> Did you try `FORTRAN=mpif90`? I think `mpifort` might be assuming it's supposed to be an F77 compiler, where "modules" have no meaning. If that doesn't work, disabling this seems fine, too.

It's something where you have to check what Fortran is wrapped by `mpifort`. I think I've got it ...

_PS: confirmed that we just have to match `gfortran` syntax. The PR should be ready ..._

SConstruct
```diff
@@ -1582,17 +1582,24 @@ end program main
...
-if env["using_apple_clang"] and env["f90_interface"] == "default":
+if (env["using_apple_clang"] and env["f90_interface"] == "default" and
+    not env["FORTRAN"]):
     env["f90_interface"] = "n"
 
 if env['f90_interface'] in ('y','default'):
     foundF90 = False
-    if env['FORTRAN']:
+    fortran_used = ""
+    if env['FORTRAN'] == "mpifort":
```
bryanwweber: Should this be `if "mpifort" in env["FORTRAN"]` to match the other cases? Like what if the absolute path to `mpifort` is passed

ischoegl: Good catch!

ischoegl:
After correcting the glitch @bryanwweber noticed, it looks like we're now getting - unrelated - crashes on macOS-13 / Python 13:

One runner crashed with:
```
[gw2] darwin -- Python 3.13.5 /Library/Frameworks/Python.framework/Versions/3.13/bin/python3
worker 'gw2' crashed while running 'test_reactor.py::TestWellStirredReactorIgnition::test_solve_steady[MoleReactor]'
```
and another one crashed with:
```
[gw2] darwin -- Python 3.13.5 /Library/Frameworks/Python.framework/Versions/3.13/bin/python3
worker 'gw2' crashed while running 'test_reactor.py::TestIdealGasMoleReactor::test_advance_with_limits'
```

The same happened during the merge of #1907, see https://github.com/Cantera/cantera/actions/runs/15888865696/job/44807273791

Another crashing runner is on #1911, see https://github.com/Cantera/cantera/actions/runs/15890473825/job/44811946486?pr=1911

ischoegl:
Pinning Python 3.13 to 3.13.4 on macOS 13 resolves failures similar to those documented in #1916.

speth (approved):

Issue #1874, opened 2025-04-16
Inconsistent Size of `net_production_rates` in `ReactorSurface()` Object


dylanrubini:
**Problem description**

My assumption was that the `net_production_rates` vector stored in `ReactorSurface()` under `kinetics`, would only store the surface species. However, this is not the case, instead both gas and surface species are stored in concatonated form. Is this behaviour intentional? For the `partial_molar_enthalpies` only the surface species are stored.

**Steps to reproduce**

``` python
import csv

import cantera as ct

# unit conversion factors to SI
cm = 0.01
minute = 60.0

#######################################################################
# Input Parameters
#######################################################################

tc = 800.0  # Temperature in Celsius
length = 0.3 * cm  # Catalyst bed length
area = 1.0 * cm**2  # Catalyst bed area
cat_area_per_vol = 1000.0 / cm  # Catalyst particle surface area per unit volume
velocity = 40.0 * cm / minute  # gas velocity
porosity = 0.3  # Catalyst bed porosity

# input file containing the surface reaction mechanism
yaml_file = 'methane_pox_on_pt.yaml'
output_filename = 'surf_pfr2_output.csv'

#####################################################################

t = tc + 273.15  # convert to Kelvin

# import the model and set the initial conditions
surf = ct.Interface(yaml_file, 'Pt_surf')
surf.TP = t, ct.one_atm
gas = surf.adjacent['gas']
gas.TPX = t, ct.one_atm, 'CH4:1, O2:1.5, AR:0.1'

mass_flow_rate = velocity * gas.density * area * porosity

# create a new reactor
r = ct.FlowReactor(gas)
r.area = area
r.surface_area_to_volume_ratio = cat_area_per_vol * porosity
r.mass_flow_rate = mass_flow_rate
r.energy_enabled = False

# Add the reacting surface to the reactor
rsurf = ct.ReactorSurface(surf, r)

sim = ct.ReactorNet([r])

print(surf.n_species)
print(rsurf.kinetics.net_production_rates.shape)
print(rsurf.kinetics.partial_molar_enthalpies.shape)
print(gas.n_species)
```

The ouput to terminal is:

``` shell
11
```
``` shell
(18,)
```
``` shell
(11,)
```

``` shell
7
```

**System information**

- Cantera version: 3.2.0a1 (dev)
- OS: Apple Silicon M4 (OS 15.3.2)
- Python/MATLAB/other software versions: Python

ischoegl:
[...]

PR #1921, opened 2025-07-05
Clone Solution objects used by Reactor / ReactorSurface

### LINKED ISSUE #1880: Negative molar fractions appear when using ReactorSurface
**Problem description**

When using ReactorSurface, negative concentrations appear. 

**Steps to reproduce**

Run attached `reproducer.py` with attached `reproducer.yaml`. Essential parts are:

```python
INITIAL_COMPOSITION = "H2:1,HN2+:0.01,N2:0.01"
file_name = "reproducer.yaml"

print(f"Loading mechanism '{file_name}' ...")
gas = ct.Solution(file_name, "gas")

print("Setting initial conditions ...")
gas.TPX = 300.0, 5, INITIAL_COMPOSITION

print("Creating reactor ...")
reactor = ct.IdealGasReactor(gas, volume=1.0, energy="off")

print("Creating surface ...")
surface_kin = ct.Interface(file_name, "surface")
surface = ct.ReactorSurface(surface_kin, reactor, A=0.1)
print(reactor.surfaces)

print("Creating network ...")
sim = ct.ReactorNet([reactor])
sim.verbose = True
reactor.thermo()

print("Integrating ...")
t_min = 1e-07
t_max = 1
time = np.geomspace(t_min, t_max, int(10*np.log(t_max / t_min)))
states = ct.SolutionArray(gas, extra=['t'])

for t in time:
    sim.advance(t)
    states.append(reactor.thermo.state, t=sim.time)
```

Where reactions:

```yaml
reactions:
  - equation: H3+ + N2 => HN2+ + H2
    rate-constant: { A: 1.86e-9, b: 0.0, Ea: 0.0 }

surface-reactions:
  - equation: H3+ => H2 + H
    sticking-coefficient: { A: 1.0, b: 0.0, Ea: 0.0 }
```

Notice that both reactions require `H3+` which is not present in the initial mixture. 

**Behavior**

At some point, molar fraction of "HN2+" becomes negative, and molar fraction of "N2" grows to 1.  See the plot 

Image

If energy equation is enabled, this also leads to negative temperature. 

**System information**

- Cantera version: 3.1.0 from PyPI
- OS: Ubuntu 25.04
- Python: 3.13.3

**Attachments**

[reproducer.zip](https://github.com/user-attachments/files/20027630/reproducer.zip)

speth:

- Partially implements changes envisioned in Cantera/enhancements#201:
  - Implements `clone` method for `Solution`, `ThermoPhase`, `Kinetics`, and `Transport`, based on `AnyMap` serialization
  - constructors for `Reactor` objects now take an optional `clone` argument, which determines whether the `Solution` object used should be cloned. The default is currently `false`, which will raise a warning. After Cantera 3.2, this default will be changed to `true`, with no warning (so, for the duration of Cantera 3.2, this argument is not optional if you don't want a warning).
  - `ReactorNet::initialize` checks all `Solution` and `Interface` objects used by `Reactor` and `ReactorSurface` objects in the network to make sure that they are all unique. If not, a warning is issued. This warning will be an error after Cantera 3.2.
- Provides user interface side of Cantera/enhancements#202
  - Constructing cloned `Interface` objects requires having access to the (probably cloned) `Solution` objects for the adjacent phases at the time the `ReactorSurface` is created, which necessitates the introduction of a new constructor for `ReactorSurface`. In doing so, we also resolve the issue of linking a single `ReactorSurface` to multiple adjacent `Reactor`s.
 - Include reaction rate multipliers in serialized Solution state.

After Cantera 3.2, when we can guarantee that the `ReactorNet` contains no re-used `Solution` objects, there is additional refactoring to the logic for how `ReactorNet::eval` works that can be implemented to complete these two enhancements.

**Checklist**

### DIFF
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -58,7 +58,10 @@ public:
     //! Create a new Solution object with cloned ThermoPhase and Kinetics objects that
     //! use the same species definitions, thermodynamic parameters, and reactions as
     //! this one.
-    shared_ptr<Solution> clone() const;
+    //! @param adjacent  For surface phases, an optional list of new adjacent phases
+    //!     to link to the InterfaceKinetics object. If not provided, the existing
+    //!     adjacent phases will be cloned.
+    shared_ptr<Solution> clone(const vector<shared_ptr<Solution>>& adjacent={}) const;
 
     //! Return the name of this Solution object
     string name() const;
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -59,8 +59,8 @@ public:
     //! use the same species definitions, thermodynamic parameters, and reactions as
     //! this one.
     //! @param adjacent  For surface phases, an optional list of new adjacent phases
-    //!     to link to the InterfaceKinetics object. If not provided, the existing
-    //!     adjacent phases will be cloned.
+    //!     to link to the InterfaceKinetics object. Any adjacent phases not provided
+    //!     will have their ThermoPhase model automatically cloned.
     //! @param withKinetics  Flag indicating whether to clone the Kinetics object
     //!    associated with this phase
     //! @param withTransport  Flag indicating whether to clone the Transport object
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -117,9 +117,7 @@ public:
     }
 
     //! Get the Solution object for an adjacent phase by name
-    shared_ptr<Solution> adjacent(const string& name) {
-        return m_adjacentByName.at(name);
-    }
+    shared_ptr<Solution> adjacent(const string& name);
 
     //! Get the number of adjacent phases
     size_t nAdjacent() const {
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -55,16 +55,19 @@ public:
         return shared_ptr<Solution>( new Solution );
     }
 
-    //! Create a new Solution object with cloned ThermoPhase and Kinetics objects that
-    //! use the same species definitions, thermodynamic parameters, and reactions as
-    //! this one.
+    //! Create a new Solution object with cloned ThermoPhase, Kinetics, and Transport
+    //! objects that use the same species definitions, thermodynamic parameters, and
+    //! reactions as this one.
     //! @param adjacent  For surface phases, an optional list of new adjacent phases
     //!     to link to the InterfaceKinetics object. Any adjacent phases not provided
     //!     will have their ThermoPhase model automatically cloned.
     //! @param withKinetics  Flag indicating whether to clone the Kinetics object
-    //!    associated with this phase
+    //!    associated with this phase. If `false`, the cloned Solution will not include
+    //!    a kinetics manager.
     //! @param withTransport  Flag indicating whether to clone the Transport object
-    //!    associated with this phase
+    //!    associated with this phase. If `false`, the cloned Solution will not include
+    //!    a transport property manager.
+    //! @since New in %Cantera 3.2.
     shared_ptr<Solution> clone(const vector<shared_ptr<Solution>>& adjacent={},
                                bool withKinetics=true, bool withTransport=true) const;
 
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -55,6 +55,11 @@ public:
         return shared_ptr<Solution>( new Solution );
     }
 
+    //! Create a new Solution object with cloned ThermoPhase and Kinetics objects that
+    //! use the same species definitions, thermodynamic parameters, and reactions as
+    //! this one.
+    shared_ptr<Solution> clone() const;
+
     //! Return the name of this Solution object
     string name() const;
 
--- a/include/cantera/base/Solution.h
+++ b/include/cantera/base/Solution.h
@@ -61,7 +61,12 @@ public:
     //! @param adjacent  For surface phases, an optional list of new adjacent phases
     //!     to link to the InterfaceKinetics object. If not provided, the existing
     //!     adjacent phases will be cloned.
-    shared_ptr<Solution> clone(const vector<shared_ptr<Solution>>& adjacent={}) const;
+    //! @param withKinetics  Flag indicating whether to clone the Kinetics object
+    //!    associated with this phase
+    //! @param withTransport  Flag indicating whether to clone the Transport object
+    //!    associated with this phase
+    shared_ptr<Solution> clone(const vector<shared_ptr<Solution>>& adjacent={},
+                               bool withKinetics=true, bool withTransport=true) const;
 
     //! Return the name of this Solution object
     string name() const;
--- a/include/cantera/zeroD/ReactorBase.h
+++ b/include/cantera/zeroD/ReactorBase.h
@@ -84,6 +84,9 @@ public:
     //!              ReactorBase with Solution object.
     void setSolution(shared_ptr<Solution> sol);
 
+    //! Access the Solution object used to represent the contents of this reactor.
+    shared_ptr<Solution> solution() { return m_solution; }
+
     //! @name Methods to set up a simulation
     //! @{
 
--- a/include/cantera/zeroD/ReactorBase.h
+++ b/include/cantera/zeroD/ReactorBase.h
@@ -49,11 +49,24 @@ class ReactorBase
 {
 public:
     explicit ReactorBase(const string& name="(none)");
+
     //! Instantiate a ReactorBase object with Solution contents.
     //! @param sol  Solution object to be set.
     //! @param name  Name of the reactor.
     //! @since New in %Cantera 3.1.
     ReactorBase(shared_ptr<Solution> sol, const string& name="(none)");
+
+    //! Instantiate a ReactorBase object with Solution contents.
+    //! @param sol  Solution object representing the contents of this reactor
+    //! @param clone  Determines whether to clone `sol` so that the internal state of
+    //!     this reactor is independent of the original Solution object and any Solution
+    //!     objects used by other reactors in the network.
+    //! @param name  Name of the reactor.
+    //! @since Added the `clone` argument in %Cantera 3.2. If not specified, the default
+    //!     behavior in %Cantera 3.2 is not to clone the Solution object. This will
+    //!     change after %Cantera 3.2 to default to `true`.
+    ReactorBase(shared_ptr<Solution> sol, bool clone, const string& name="(none)");
+
     virtual ~ReactorBase();
     ReactorBase(const ReactorBase&) = delete;
     ReactorBase& operator=(const ReactorBase&) = delete;
--- a/include/cantera/zeroD/ReactorBase.h
+++ b/include/cantera/zeroD/ReactorBase.h
@@ -48,6 +48,7 @@ struct SensitivityParameter
 class ReactorBase
 {
 public:
+    //! @deprecated After %Cantera 3.2, this constructor will become protected
     explicit ReactorBase(const string& name="(none)");
 
     //! Instantiate a ReactorBase object with Solution contents.
--- a/include/cantera/zeroD/ReactorBase.h
+++ b/include/cantera/zeroD/ReactorBase.h
@@ -142,9 +142,15 @@ public:
     WallBase& wall(size_t n);
 
     //! Add a ReactorSurface object to a Reactor object.
+    //! @attention This method should generally not be called directly by users.
+    //!     Reactor and ReactorSurface objects should be connected by providing adjacent
+    //!     reactors to the newReactorSurface factory function.
     virtual void addSurface(ReactorSurface* surf);
 
     //! Add a ReactorSurface object to a Reactor object.
+    //! @attention This method should generally not be called directly by users.
+    //!     Reactor and ReactorSurface objects should be connected by providing adjacent
+    //!     reactors to the newReactorSurface factory function.
     void addSurface(shared_ptr<ReactorBase> surf);
 
     //! Return a reference to the *n*-th ReactorSurface connected to this reactor.
--- a/include/cantera/zeroD/ReactorBase.h
+++ b/include/cantera/zeroD/ReactorBase.h
@@ -168,6 +168,7 @@ public:
     //! @attention This method should generally not be called directly by users.
     //!     Reactor and ReactorSurface objects should be connected by providing adjacent
     //!     reactors to the newReactorSurface factory function.
+    //! @deprecated  Unused. To be removed after %Cantera 3.2.
     void addSurface(shared_ptr<ReactorBase> surf);
 
     //! Return a reference to the *n*-th ReactorSurface connected to this reactor.
--- a/include/cantera/zeroD/ReactorFactory.h
+++ b/include/cantera/zeroD/ReactorFactory.h
@@ -69,6 +69,13 @@ shared_ptr<Reactor> newReactor4(
 shared_ptr<Reservoir> newReservoir(
     shared_ptr<Solution> contents, const string& name="(none)");
 
+//! Create a ReactorSurface object with the specified contents and adjacent reactors
+//! participating in surface reactions.
+//! @since  New in %Cantera 3.2.
+shared_ptr<ReactorSurface> newReactorSurface(
+    shared_ptr<Solution> soln, const vector<shared_ptr<ReactorBase>>& reactors,
+    const string& name="(none)");
+
 //! @}
 
 }
--- a/include/cantera/zeroD/ReactorFactory.h
+++ b/include/cantera/zeroD/ReactorFactory.h
@@ -22,7 +22,7 @@ class Reservoir;
 //! ```cpp
 //!     shared_ptr<ReactorBase> r1 = newReactor("IdealGasReactor");
 //! ```
-class ReactorFactory : public Factory<ReactorBase, shared_ptr<Solution>, const string&>
+class ReactorFactory : public Factory<ReactorBase, shared_ptr<Solution>, bool, const string&>
 {
 public:
     static ReactorFactory* factory();
@@ -49,7 +49,8 @@ private:
 //! Create a ReactorBase object of the specified type and contents.
 //! @since  New in %Cantera 3.2.
 shared_ptr<ReactorBase> newReactorBase(
-    const string& model, shared_ptr<Solution> contents, const string& name="(none)");
+    const string& model, shared_ptr<Solution> contents, bool clone=true,
+    const string& name="(none)");
 
 //! Create a Reactor object of the specified type and contents
 //! @since Starting in %Cantera 3.1, this method requires a valid Solution object and
@@ -62,19 +63,20 @@ shared_ptr<ReactorBase> newReactor(
 //! Create a Reactor object of the specified type and contents
 //! @since  New in %Cantera 3.2. Transitional method returning a `Reactor` object.
 shared_ptr<Reactor> newReactor4(
-    const string& model, shared_ptr<Solution> contents, const string& name="(none)");
+    const string& model, shared_ptr<Solution> contents, bool clone=true,
+    const string& name="(none)");
 
 //! Create a Reservoir object with the specified contents
 //! @since New in %Cantera 3.2.
 shared_ptr<Reservoir> newReservoir(
-    shared_ptr<Solution> contents, const string& name="(none)");
+    shared_ptr<Solution> contents, bool clone=true, const string& name="(none)");
 
 //! Create a ReactorSurface object with the specified contents and adjacent reactors
 //! participating in surface reactions.
 //! @since  New in %Cantera 3.2.
 shared_ptr<ReactorSurface> newReactorSurface(
     shared_ptr<Solution> soln, const vector<shared_ptr<ReactorBase>>& reactors,
-    const string& name="(none)");
+    bool clone=true, const string& name="(none)");
 
 //! @}
 
--- a/include/cantera/zeroD/ReactorFactory.h
+++ b/include/cantera/zeroD/ReactorFactory.h
@@ -61,21 +61,42 @@ shared_ptr<ReactorBase> newReactor(
     const string& model, shared_ptr<Solution> contents, const string& name="(none)");
 
 //! Create a Reactor object of the specified type and contents
+//! @param  model  Reactor type to be created. See [this list of reactor
+//!     types](../reference/reactors/index.html) for available options.
+//! @param  contents  Solution object to model the thermodynamic properties and
+//!     reactions occurring in the reactor
+//! @param clone  Determines whether to clone `sol` so that the internal state of this
+//!     reactor is independent of the original Solution object and any Solution objects
+//!     used by other reactors in the network.
+//! @param name  Name of the reactor.
 //! @since  New in %Cantera 3.2. Transitional method returning a `Reactor` object.
 shared_ptr<Reactor> newReactor4(
     const string& model, shared_ptr<Solution> contents, bool clone=true,
     const string& name="(none)");
 
 //! Create a Reservoir object with the specified contents
+//! @param  contents  Solution object to model the contents of this reservoir
+//! @param clone  Determines whether to clone `sol` so that the internal state of this
+//!     reservoir is independent of the original Solution object and any Solution
+//!     objects used by other reactors in the network.
+//! @param name  Name of the reservoir.
 //! @since New in %Cantera 3.2.
 shared_ptr<Reservoir> newReservoir(
     shared_ptr<Solution> contents, bool clone=true, const string& name="(none)");
 
 //! Create a ReactorSurface object with the specified contents and adjacent reactors
 //! participating in surface reactions.
+//! @param  contents  Solution (Interface) object to model the thermodynamic properties
+//!     and reactions occurring in the reactor
+//! @param  reactors  List of Reactors adjacent to this surface, whose contents
+//!     participate in reactions occurring on this surface.
+//! @param clone  Determines whether to clone `sol` so that the internal state of this
+//!     surface is independent of the original Interface object and any Solution objects
+//!     used by other reactors in the network except those in the `reactors` list.
+//! @param name  Name of the reactor surface.
 //! @since  New in %Cantera 3.2.
 shared_ptr<ReactorSurface> newReactorSurface(
-    shared_ptr<Solution> soln, const vector<shared_ptr<ReactorBase>>& reactors,
+    shared_ptr<Solution> contents, const vector<shared_ptr<ReactorBase>>& reactors,
     bool clone=true, const string& name="(none)");
 
 //! @}
--- a/include/cantera/zeroD/ReactorSurface.h
+++ b/include/cantera/zeroD/ReactorSurface.h
@@ -22,13 +22,19 @@ public:
     //! Create a new ReactorSurface
     //! @param soln  Thermodynamic and kinetic model representing species and reactions
     //!     on the surface
+    //! @param clone  Determines whether to clone `soln` so that the internal state of
+    //!     this reactor surface is independent of the original Solution (Interface)
+    //!     object and any Solution objects used by other reactors in the network.
     //! @param reactors  One or more reactors whose phases participate in reactions
     //!     occurring on the surface. For the purpose of rate evaluation, the
     //!     temperature of the surface is set equal to the temperature of the first
     //!     reactor specified.
     //! @param name  Name used to identify the surface
+    //! @since  Constructor signature including `reactors` and `clone` arguments
+    //!     introduced in %Cantera 3.2.
     ReactorSurface(shared_ptr<Solution> soln,
                    const vector<shared_ptr<ReactorBase>>& reactors,
+                   bool clone,
                    const string& name="(none)");
 
     ReactorSurface(shared_ptr<Solution> sol, const string& name="(none)");
--- a/include/cantera/zeroD/ReactorSurface.h
+++ b/include/cantera/zeroD/ReactorSurface.h
@@ -40,7 +40,10 @@ public:
     //! @deprecated To be removed after Cantera 3.2. Replaced by constructor where
     //!    contents and adjacent reactors are specified
     ReactorSurface(shared_ptr<Solution> sol, const string& name="(none)");
-    using ReactorBase::ReactorBase; // inherit constructors
+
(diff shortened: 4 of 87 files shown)


speth:
With this implementation now largely working, I have a couple of observations in relation to the discussion in Cantera/enhancements#201:
- We cannot rely on _always_ cloning the `Solution` objects used in a reactor network. The most severe issue with this is that `CustomRate` objects implemented as Python lambdas cannot be serialized. This is currently the reason the `custom_reactions.py` example fails to run.
- We cannot make cloning `Solution` objects the _default_ behavior in this Cantera version. Based on the changes to various test cases and examples, it is clear that there is a fair amount of existing code that relies on the incidental sharing of `Solution` states. While identifying these cases in the test suite is relatively easy, it is not so easy elsewhere and would be a significant breaking change.

As such, I think the next steps for this PR are:
- Implement an additional `clone` argument to the `Reactor` and `ReactorSurface` constructors, which will default to `None`, In Cantera 3.2, if no value for this argument is provided, the `Solution` will not be cloned and a warning will be issued. In Cantera 3.3, we can make `clone=True` the default, while `clone=False` will still be available for special cases.
- Add a check during `ReactorNet::initialize` to identify whether any `Solution` objects are being used multiple times within the network. In Cantera 3.2, doing so will result in a warning, which can then be converted to an error in Cantera 3.3.

ischoegl:
@speth ... thanks for your work on this, as well as the thoughts for the roadmap, which looks :+1:. 

I was somewhat surprised that `CustomRate` objects with Python lambdas cause failures, which is not among the consequences that I would have expected. On second thought, I assume that it has to do with those not being part of the YAML serialization, and thus cannot be cloned? If this were the case, I'd suggest simply throwing an error if there's an attempt to clone a mechanism that includes `CustomRate` (I have not started a review, so it may already be there). I assume this is purely limited to Python lambdas, as C++ has no way of ensuring that Python objects persist in memory? I.e., `CustomRate` with `Func1` objects should work?

Beyond, it appears that #1923 fixes other sample tests that are currently failing.

speth:
> I was somewhat surprised that `CustomRate` objects with Python lambdas cause failures, which is not among the consequences that I would have expected. On second thought, I assume that it has to do with those not being part of the YAML serialization, and thus cannot be cloned?

I didn't anticipate this either, but it's pretty clear in retrospect. Indeed, the issue is that such lambdas cannot be serialized.

> If this were the case, I'd suggest simply throwing an error if there's an attempt to clone a mechanism that includes `CustomRate` (I have not started a review, so it may already be there). I assume this is purely limited to Python lambdas, as C++ has no way of ensuring that Python objects persist in memory? I.e., `CustomRate` with `Func1` objects should work?

This affects all `CustomRate` objects, including ones built on `Func1`, since the `CustomFunc1Rate::getParameters` method is not implemented (or rather, explicitly raises a `NotImplementedError`). In principle, we could implement serialization of `Func1`-based rates, but it's not strictly necessary. The fallback is for the user to instantiate independent `Solution` objects and call the `Reactor` constructor with the `clone=False` option.

Likewise, `ExtensibleRate` objects that don't implement the `set_parameters` / `get_parameters` pair will require this alternative approach, as will any other user-defined model that doesn't fully implement serialization.

---
The "next steps" I mentioned before are now completed, but I'm currently struggling with how to get this to play nice with the generated CLib. I added a new method, `shared_ptr<Solution> ReactorBase::solution()` for accessing the `Solution` object used by a `Reactor`. Depending on whether a clone was created or not, this is either the new cloned object or just the same object that was passed into the constructor. In the first case, I understand why CLib is struggling -- that cloned `Solution` isn't already in the cabinet, but I'm not sure what the right way to update the code generation for that is. In the second case, this accessor also doesn't seem to be working, and I can't figure out why.

ischoegl:
Thanks, @speth for further context.

> [...] In principle, we could implement serialization of `Func1`-based rates, but it's not strictly necessary. The fallback is for the user to instantiate independent `Solution` objects and call the `Reactor` constructor with the `clone=False` option.
> Likewise, `ExtensibleRate` objects that don't implement the `set_parameters` / `get_parameters` pair will require this alternative approach, as will any other user-defined model that doesn't fully implement serialization.

It should be pretty straight-forward to come up with `Func1` serialization so things work out of the box. For `ExtensibleRate`, it's obviously up to the users. This clearly goes beyond this PR.

> The "next steps" I mentioned before are now completed, but I'm currently struggling with how to get this to play nice with the generated CLib. I added a new method, `shared_ptr<Solution> ReactorBase::solution()` for accessing the `Solution` object used by a `Reactor`. Depending on whether a clone was created or not, this is either the new cloned object or just the same object that was passed into the constructor. In the first case, I understand why CLib is struggling -- that cloned `Solution` isn't already in the cabinet, but I'm not sure what the right way to update the code generation for that is. In the second case, this accessor also doesn't seem to be working, and I can't figure out why.

Hmm. This makes sense - it wasn't straightforward to get the initial accessors to work properly (most relevant example: `sol_adjacent`); I managed to address this with the `uses` field. While the Python code should be relatively easy to follow (I did go back and revised), the conditionals in Jinja are quite difficult to make readable (this may be improved by Cantera/enhancements#232). 

At the same time, the existing logic may be sufficient to handle this situation using your new `shared_ptr<Solution> ReactorBase::solution()` method. The cabinet entries should be created during constructor calls for `Reactor` and `Domain1D` objects. Fortunately, the Jinja logic for constructors is among the more straightforward/better commented ones:
```yaml
clib-constructor: |-
  ## CLib constructor template: instantiates new object
  // constructor: {{ cxx_wraps }}
  try {
      {% for line in lines %}
      ## add lines used for CLib/C++ variable crosswalk
      {{ line }}
      {% endfor %}
      {% if shared %}
      ## instantiated object manages associated objects (see: sol3_newSolution)
      ## storage of associated objects requires id (handle) of new object
      auto obj = {{ cxx_name }}({{ ', '.join(cxx_args) }});
      int id = {{ base }}Cabinet::add(obj);
      {% for typ, getter in shared %}
      ## add associated objects (see: sol3_newSolution, sol3_adjacent)
      if (obj->{{ getter }}()) {
          {{ typ }}Cabinet::add(obj->{{ getter }}(), id);
      }
      {% endfor %}
      return id;
      {% else %}{# not shared #}
      ## instantiated object has no associated objects; no id is needed
      return {{ base }}Cabinet::add({{ cxx_name }}({{ ', '.join(cxx_args) }}));
      {% endif %}{# shared #}
  } catch (...) {
      return handleAllExceptions({{ error[0] }}, {{ error[1] }});
  }
```

Specifically, you'd need to get the `Solution` object clones into the following portion via the `shared` argument:

```
      {% for typ, getter in shared %}
      ## add associated objects (see: sol3_newSolution, sol3_adjacent)
      if (obj->{{ getter }}()) {
          {{ typ }}Cabinet::add(obj->{{ getter }}(), id);
      }
      {% endfor %}
```

As the cloned `Solution` is a managed object, the CLib destructors need to use the `uses` field as well.

As an aside, I noticed that there are some stray `sol3_*` references left in the `##` Jinja comment blocks :cry: (should be `sol_*`).

PS: In a nutshell, you may need to update `ctreactor_auto.yaml` as follows:

```yaml
- name: new
  wraps: newReactorBase
  uses: [solution]
  [...]
- name: del
  uses: [solution]
```
I haven't tested it, though.

include/cantera/zeroD/ReactorBase.h
```diff
@@ -84,6 +94,9 @@ class ReactorBase
     //!              ReactorBase with Solution object.
     void setSolution(shared_ptr<Solution> sol);
 
+    //! Access the Solution object used to represent the contents of this reactor.
+    shared_ptr<Solution> solution() { return m_solution; }
```
ischoegl, line 102: In other contexts (specifically Python), we use `contents`. I don't have strong opinions, though.

Fwiw, we're not consistent in Python, either:
```
    property thermo:
        """
        The `ThermoPhase` object representing the reactor's contents.
        """
        def __get__(self):
            self.rbase.restoreState()
            return self._contents
```
The docstring here isn't correct any longer as it has become the `Solution` object (it refers to the same object as the new C++ method). The situation is the also same for an existing `kinetics` property. It still all works due to inheritance, but is overall misleading.

Obviously, this is a pre-existing condition. Introducing the C++ method does, however, provide an opportunity to make things consistent across all APIs.

speth: Unfortunately, `ReactorBase::contents` already exists, returning a `ThermoPhase&`.

I think getting to a consistent state with `contents` as the name will take a couple of iterations (versions) -- we can deprecate the existing `ReactorBase::contents` now, along with the Python `ReactorBase.thermo` and `ReactorSurface.kinetics` properties, but at least for now we need a distinct name for this method.

speth:
> Hmm. This makes sense - it wasn't straightforward to get the initial accessors to work properly (most relevant example: `sol_adjacent`); I managed to address this with the `uses` field. While the Python code should be relatively easy to follow (I did go back and revised), the conditionals in Jinja are quite difficult to make readable (this may be improved by [Cantera/enhancements#232](https://github.com/Cantera/enhancements/issues/232)).

Thanks for pointing me to the `uses` field, and the template that's use for the `newReactorBase` method. This seems to get some of the necessary behavior right with minimal effort. However, it's not quite there for both cases of either cloning or not cloning the `Solution`:
- In the case where the `Solution` object is not cloned, this results in an additional reference to the `Solution` being added to the cabinet, with a new ID. This new `Solution` ID then cannot be used to obtain the `thermo` member using `sol_thermo`.
- In the case where the `Solution` is cloned, we get a new ID which is correctly retrieved via `reactor_solution`. However, the components of this `Solution` like the thermo object are not themselves being added to the corresponding cabinets, again causing `sol_thermo` to fail, and I'm not sure where that logic should live.

I think I don't fully understand what the whole "parent" logic within `Cabinet` does, which is maybe contributing to my confusion here. The current situation is a bit complicated by the fact in one case, the `Solution` should be "owned" by the `Reactor` and (presumably) deleted when the `Reactor` is deleted, but in the other case the `Solution` should outlive the `Reactor`.

speth:
@ischoegl - Could you provide a bit of guidance on how to wrap the following function for the generated CLib?
```c++
//! Create a ReactorSurface object with the specified contents and adjacent reactors
//! participating in surface reactions.
//! @since  New in %Cantera 3.2.
shared_ptr<ReactorSurface> newReactorSurface(
    shared_ptr<Solution> soln, const vector<shared_ptr<ReactorBase>>& reactors,
    bool clone=true, const string& name="(none)");
```
which I would would imagine calling as something like:
```c++
int32_t surf = sol_newInterface("ptcombust.yaml", "Pt_surf", 0, 0);
int32_t gas = sol_adjacent(surf, 0);
int32_t reactor = reactor_new("IdealGasReactor", gas, 1, "test");
vector<int32_t> adjacent = {reactor};
int32_t rsurf = reactorsurface_new(surf, 1, adjacent.data(), 1, "surf");
```
I'm not sure how to add a "free function" like this to the interface.

ischoegl:
> @ischoegl - Could you provide a bit of guidance on how to wrap the following function for the generated CLib?

I pushed an implementation for what is possible at the moment. The constructors are currently named after the base class, which is likely an easy fix, but goes beyond the scope here. You also needed to register the header for the object type.

The result is
```C++
   /**
     *  Create a ReactorSurface object with the specified contents and adjacent reactors participating in surface reactions.
     *
     *  Wraps C++ constructor: `shared_ptr<ReactorSurface> newReactorSurface(shared_ptr<Solution>, const vector<shared_ptr<ReactorBase>>&, bool, const string&)`
     *
     *  @param soln         Integer handle to Solution object. Undocumented.
     *  @param[in] reactorsLen Length of array reserved for reactors.
     *  @param reactors     Memory holding ReactorBase objects. Undocumented.
     *  @param clone        Undocumented.
     *  @param name         Undocumented.
     *  @returns            Handle to stored ReactorSurface object or -1 for exception handling.
     */
    int32_t reactor_newSurface(int32_t soln, int32_t reactorsLen, const int32_t* reactors, int32_t clone, const char* name);
```

speth:
@ischoegl, thanks for the help sorting out the new `clib` aspect of this. I think this is ready for review now. 

The complaint from Codecov is at least partly incorrect, due to it miscounting coverage of the lambda functions in `ReactorFactory`, and much of the other uncovered code is in deprecated methods that are only needed for transitional purposes.

include/cantera/base/Solution.h
```diff
@@ -55,6 +55,20 @@ class Solution : public std::enable_shared_from_this<Solution>
...
+    //! Create a new Solution object with cloned ThermoPhase and Kinetics objects that
+    //! use the same species definitions, thermodynamic parameters, and reactions as
+    //! this one.
+    //! @param adjacent  For surface phases, an optional list of new adjacent phases
+    //!     to link to the InterfaceKinetics object. Any adjacent phases not provided
+    //!     will have their ThermoPhase model automatically cloned.
+    //! @param withKinetics  Flag indicating whether to clone the Kinetics object
+    //!    associated with this phase
+    //! @param withTransport  Flag indicating whether to clone the Transport object
+    //!    associated with this phase
```
ischoegl: I'm not convinced that we need fine-grained control like this. I'd rather leave the cloning question as a yes/no proposition for everything in the Solution object.

include/cantera/zeroD/ReactorBase.h
```diff
@@ -48,12 +48,23 @@ struct SensitivityParameter
...
 {
 public:
+    //! @deprecated After %Cantera 3.2, this constructor will become protected
     explicit ReactorBase(const string& name="(none)");
     //! Instantiate a ReactorBase object with Solution contents.
     //! @param sol  Solution object to be set.
     //! @param name  Name of the reactor.
     //! @since New in %Cantera 3.1.
     ReactorBase(shared_ptr<Solution> sol, const string& name="(none)");
+    //! Instantiate a ReactorBase object with Solution contents.
```
ischoegl, line 60:
```suggestion

    //! Instantiate a ReactorBase object with Solution contents.
```
Missing newline?

include/cantera/zeroD/ReactorBase.h
```diff
@@ -48,12 +48,23 @@ struct SensitivityParameter
...
+    //! @param sol  Solution object representing the contents of this reactor
+    //! @param clone  Determines whether to clone `sol` so that the internal state of
+    //!     this reactor is independent of the original Solution object and any Solution
+    //!     objects used by other reactors in the network.
+    //! @param name  Name of the reactor.
+    //! @since Added the `clone` argument in %Cantera 3.2. If not specified, the default
+    //!     behavior in %Cantera 3.2 is not to clone the Solution object. This will
+    //!     change after %Cantera 3.2 to default to `true`.
```
ischoegl, line 68: I'd be ok with cloning by default starting immediately. Outside of Python, It's not clear to me how users would be warned about the change later, i.e., 3.2->3.3.

include/cantera/zeroD/ReactorBase.h
ischoegl: Fair point. I'd say use `ReactorBase::contents4` as the intermediate name so it's clear that it's temporary? Having consistent naming is one of my pet peeves ...

include/cantera/zeroD/ReactorBase.h
```diff
@@ -142,9 +156,15 @@ class ReactorBase
     WallBase& wall(size_t n);
 
     //! Add a ReactorSurface object to a Reactor object.
+    //! @attention This method should generally not be called directly by users.
+    //!     Reactor and ReactorSurface objects should be connected by providing adjacent
+    //!     reactors to the newReactorSurface factory function.
```
ischoegl, line 164: This should be deprecated in favor of the version with the shared pointer.

Beyond, I'm very conflicted about adding ReactorSurface directly, although I believe a resolution goes beyond the scope of this PR. Will add my comment to the new `newReactorSurface` function.

include/cantera/zeroD/ReactorFactory.h
```diff
@@ -60,14 +61,43 @@ shared_ptr<ReactorBase> newReactor(
...
+//! participating in surface reactions.
+//! @param  contents  Solution (Interface) object to model the thermodynamic properties
+//!     and reactions occurring in the reactor
+//! @param  reactors  List of Reactors adjacent to this surface, whose contents
+//!     participate in reactions occurring on this surface.
+//! @param clone  Determines whether to clone `sol` so that the internal state of this
+//!     surface is independent of the original Interface object and any Solution objects
+//!     used by other reactors in the network except those in the `reactors` list.
+//! @param name  Name of the reactor surface.
+//! @since  New in %Cantera 3.2.
```
ischoegl, line 97: I'm conflicted on this one, as I personally would prefer using a `Connector` object to link a `Reactor` to a `ReactorSurface` (perhaps via a `Bilayer`?). This goes beyond the scope of this PR, so I'm :+1: on this for the time being.

include/cantera/zeroD/ReactorSurface.h
```diff
@@ -19,8 +19,31 @@ class SurfPhase;
 class ReactorSurface : public ReactorBase
 {
 public:
+    //! Create a new ReactorSurface
```
ischoegl, line 22: See my comment for the factory function.

include/cantera/zeroD/ReactorSurface.h
```diff
@@ -19,8 +19,31 @@ class SurfPhase;
...
+                   bool clone,
+                   const string& name="(none)");
+
+    //! @deprecated To be removed after Cantera 3.2. Replaced by constructor where
+    //!    contents and adjacent reactors are specified
     ReactorSurface(shared_ptr<Solution> sol, const string& name="(none)");
-    using ReactorBase::ReactorBase; // inherit constructors
+
+    //! @deprecated To be removed after Cantera 3.2. Replaced by constructor where
+    //!    contents and adjacent reactors are specified
```
ischoegl, line 45: I'm not sure we should go there yet. Python has a 'pending deprecation' ... I'd prefer to use a comment like this here.

include/cantera/zeroD/ReactorSurface.h
```diff
@@ -67,6 +90,8 @@ class ReactorSurface : public ReactorBase
     }
 
     //! Set the reactor that this Surface interacts with
+    //! @deprecated To be removed after %Cantera 3.2. Superseded by constructor taking
+    //!     a list of adjacent reactors.
```
ischoegl, line 94: Similar comment about 'pending deprecation'.

interfaces/cython/cantera/reactor.pyx
```diff
@@ -15,19 +15,27 @@ cdef class ReactorBase:
...
     """
     reactor_type = "none"
-    def __cinit__(self, _SolutionBase contents, *args, name="(none)", **kwargs):
-        self._rbase = newReactorBase(stringify(self.reactor_type),
-                                     contents._base, stringify(name))
-        self.rbase = self._rbase.get()
+    def __cinit__(self, _SolutionBase contents, *args, name="(none)", clone=None,
+                  **kwargs):
+        if clone is None:
+            clone = False  # TODO: change as indicated after Cantera 3.2
```
ischoegl, line 21:
```suggestion
            clone = True
```
This warning mechanism only works for Python and otherwise just moves the pain to the next version.

src/oneD/MultiNewton.cpp
ischoegl, line 1: I'd move this into a separate PR, as it's unrelated - there appears to be a new failure that may be caused by this?

ischoegl (changes requested):
@speth ... thanks for taking this on: this is a pretty important feature. 

I have minor comments and one larger issue: I believe that long-term, we'd like to keep treating a `ReactorSurface` just like any other reactor, and instead implement a new `Connector` to object to create the linkage? I personally believe that this would a cleaner solution, but would require severing the internal stacking of the work vector. This would likely happen in a future release, but would also reverse some of the current deprecation decisions.

ischoegl:
P.S.: Thinking about this some more, I'm curious about the rationale for not using `Solution` clones in the future: do you see any applications where people would like to use a single copy? I'm honestly leaning towards just using clones (i.e., no shared objects). Apart from Python users, whom we can warn with the `None` workaround, anyone using implicit sharing as a 'feature' is in for a surprise once the default changes to 'clone'. I'd rather just rip that band-aid off and avoid introducing the additional parameter ...

include/cantera/zeroD/ReactorBase.h
speth: I don't really understand what the "4" means in this context, and would rather stick with "solution" as a reasonably interpretable name for now, with a "TODO" to rename it later. Do you want me to introduce the other changes as part of this PR or as a separate follow-on?

include/cantera/base/Solution.h
speth: The cost of cloning `Solution` objects is not totally insignificant. Transport initialization in particular is quite slow, and of course unnecessary for reactor networks. And for `Reservoir` objects, there is of course no need to clone the `Kinetics`.

include/cantera/zeroD/ReactorBase.h
speth: I think that would be a pretty severe breaking change with no warning. Our own examples made use of the assumption that the the input `Solution` also represented the reactor contents in numerous places, and I expect that there's a lot of user code that likewise follows this approach.

In Cantera 3.2, users will get a warning, either for Python or C++ usage, if no value for `clone` is specified. In Cantera 3.3, the new default will apply, and no warning will be required in any interface. Only users who skip from 3.1 (or older) to 3.3 might be caught by surprise.

include/cantera/zeroD/ReactorSurface.h
speth: It is impossible to clone the `Interface` object without access to the `Solution` objects which define the adjacent phases. 

While I suppose making these links later is possible in the case where the `Interface` object isn't being cloned, I think requiring these `Solution` objects to be provided at construction time is consistent with the changes that you've introduced to require `Solution` objects when constructing `Reactor`s and requiring `Reactor` objects to be provided when constructing `Connector`s.

include/cantera/zeroD/ReactorSurface.h
speth: In addition to the consistency with removing `setSolution` from `ReactorBase` and `install` from `Wall`/`FlowDevice`, the idea that there is only one `Reactor` for this `ReactorSurface` to interact with is flawed. Ditching this method is an important step to resolving Cantera/enhancements#202.

interfaces/cython/cantera/reactor.pyx
speth: The purpose of setting `clone = False` here is to avoid triggering a second deprecation warning from C++, since we issue a deprecation warning in Python that is specific to the Python keyword argument syntax.

At the C++ level, the case of clone not being specified is handled by a separate constructor that issues a deprecation warning (if you call the constructors directly) or by the fact that all the `new...` factory functions are either deprecated (in which case they retain the old non-cloning behavior) or "new in Cantera 3.2" in which case they already default to clone=True. Is there some other case that I'm missing?

speth:
> P.S.: Thinking about this some more, I'm curious about the rationale for not using `Solution` clones in the future: do you see any applications where people would like to use a single copy? I'm honestly leaning towards just using clones (i.e., no shared objects). 

Yes, I believe there are cases where not cloning the `Solution` object is preferable. Besides the relatively small performance benefit, the main reason is to handle cases where the `Solution` is not (correctly) serializable. There are currently several places where this may happen:
- Any use of `CustomRate`, where the rate is defined as a Python lambda function
- Any use of `ExtensibleRate` where the user has not gotten as far as implementing the `get_parameters` and `set_parameters` methods to allow their rate class to be serialized. While users should be able to implement these methods, I think it is an unnecessary burden in some cases, and it would significantly increase the complexity of just trying out a new rate parameterization to see if it's useful before going to the effort of implementing these additional utility methods.
- Any in-development thermodynamic or kinetics/reaction models implemented in C++, where again, it should be possible to start with implementing the methods that describe the physics rather than having to have all of the utility methods working right from the start.

> Apart from Python users, whom we can warn with the `None` workaround, anyone using implicit sharing as a 'feature' is in for a surprise once the default changes to 'clone'. I'd rather just rip that band-aid off and avoid introducing the additional parameter ...

What existing (as of Cantera 3.1) use case doesn't provide a warning or notification about this change?
- `make_shared<Reactor>(soln, name)` issues a deprecation warning about "clone argument not specified and notes the pending change in behavior"
- `newReactor("Reactor", soln, name)` issues a deprecation warning about the return type change and redirects the user to `newReactor4` where the `clone` argument is added. I guess the warning here should also mention the change in the cloning behavior, given the path is not removal of this method but to change the existing behavior.
- `newReactor4` is "new in Cantera 3.2" and defaults to `clone=true`
- `newReservoir` is "new in Cantera 3.2" and defaults to `clone=true`
- `newReactorSurface` is "new in Cantera 3.2" and defaults to `clone=true`

src/oneD/MultiNewton.cpp
speth: This change fixed test failures that started occurring in `TestWellStirredReactorIgnition.test_solve_steady[Reactor]` after introducing the `Solution` cloning. I've temporarily dropped this from the PR so you can see the failures.

ischoegl: Ah, ok. The other failure appears to be unrelated. Please feel free to re-add here as the issue is related to changes introduced in this PR.

interfaces/cython/cantera/reactor.pyx
ischoegl: Thanks for the additional context. The C++ mechanism is nice - sorry I didn't follow the first time.

include/cantera/zeroD/ReactorSurface.h
ischoegl: I agree that `ReactorSurface` objects need to be able to communicate with more than one `Reactor`.

include/cantera/zeroD/ReactorBase.h
ischoegl: The `4` is a placeholder similar to the `3` I have used extensivly during transitions. Right now, we have a `Reactor4` ... my idea is that this is a feature that goes closer to a future Cantera, rather than the current major version. I'm :+1: with `solution` but would like to deprecate `ReactorBase::contents` regardless. This PR or another one - up to you as long as it goes in for 3.2.

include/cantera/zeroD/ReactorBase.h
ischoegl: Ok. I didn't catch the bit with the C++ mechanism, so I believe this is ok. Too bad that we can't insert `clone` at the end of the parameter list due to the default parameter for name ...

include/cantera/base/Solution.h
ischoegl: I see the point, but it's the fault of the user to not skip the transport object for `Reactor`. For the case of `Reservoir`, I'm actually really conflicted about sharing Kinetics. It would be cleaner to use a default Kinetics object that is empty. 

That said, I still prefer to eliminate this, as it messes with the objective of cloning whenever possible.

include/cantera/zeroD/ReactorSurface.h
ischoegl: This goes back to me viewing `ReactorNet` as a bipartite graph, where `ReactorBase` specializations are connected by `Connector` specializations. Ultimately, I envision using a `BiLayer` (or similar) object to establish the connection. From that perspective, this isn't an extrapolation of what I've done before ... I'm :+1: with this for the time being, but the deprecation may be temporary.

It is true that cloning is impossible without access, but this is due to the solution vector being stacked in a way that I don't think is ideal.

ischoegl (commented):
@speth ... thanks for your detailed comments: your explanations address most of my concerns. The only sticking points are the partial cloning option, where I'd rather see a clean break, and the deprecation of `ReactorBase::contents`, which can be addressed in a follow-on PR. Regarding the pending deprecations, I explained that I envision an alternative way of attaching a `Reactor` to a `ReactorSurface` with a future `Connector` object at some point in the future. If we need to reactivate what's deprecated now is necessary, so be it.

Regarding your feedback on my 'PS' - yes, with the C++ deprecation warnings in place, this is good by me.

src/oneD/MultiNewton.cpp
speth: Furthermore, it looks like the intermittent failure of the CI job "Run Python 3.11 examples on ubuntu-22.04, NumPy latest" seems to be a separate issue -- it also occurred in the latest run where I omitted the change to MultiNewton. I'll create a separate issue for that.

include/cantera/base/Solution.h
speth: Oh, perhaps the docstring needs to be more clear. When cloning of the Transport/Kinetics objects is skipped, then the resulting `Solution` does indeed default to the "none" / not implemented models. There is no case where the cloned `Solution` object shares some components with the original.

include/cantera/zeroD/ReactorSurface.h
speth: I think the coupling between phases involved in an interface kinetic mechanism is intrinsically a tighter relationship than that, but we can continue to work on how this is represented after Cantera 3.2. In any case, this PR provides a necessary first step: the ability to simultaneously set all of the `Solution` and `Interface` objects involved in a `ReactorNet` to match the state of the network, and to then evaluate properties and rates as needed.

include/cantera/base/Solution.h
speth: I've updated the docstring to explain the behavior of these options.

include/cantera/zeroD/ReactorSurface.h
ischoegl: FWIW, referencing #1874. Definitely beyond the scope of this PR.

ischoegl (approved):
My only remaining concern was to start the deprecation cycle for `ReactorBase::contents` so we can start moving things. I'm :+1: for this to be a separate PR.

speth:
> My only remaining concern was to start the deprecation cycle for `ReactorBase::contents` so we can start moving things. I'm 👍 for this to be a separate PR.

~~I'm happy to add that to this PR; just haven't had time to work on this in the past week. Please hold off on merging this.~~

Actually, scratch that. I started in on migrating `ReactorBase::contents` along with renaming the Python `Reactor.thermo` and `ReactorSurface.kinetics` methods to also be `contents`, and it touches a lot of code, so I'll take that up in a separate PR.

ischoegl:
@speth ... in order to handle this consistently going forward, do you believe it to be necessary to extend cloning to `Domain1D` and `SolutionArray`?

PR #1930, opened 2025-07-21
Add Soret model for mixture-averaged diffusion

g3bk47:
Model based on the formulation by Chapman and Cowling (1970). For more information about this implementation and its validation, see T. Zirwes, A. Kronenburg, _Assessment of Approximate Soret Diffusion Models for Hydrogen and Ammonia Combustion_, Flow, Turbulence and Combustion (2025) (https://doi.org/10.1007/s10494-025-00680-5).

Addition of a Soret diffusion (thermodiffusion) model for use with the mixture-averaged diffusion model. For more information, see the discussion in https://github.com/Cantera/enhancements/issues/231.

**Checklist**


src/transport/MixTransport.cpp
```diff
@@ -50,8 +50,80 @@ double MixTransport::thermalConductivity()
...
+
+    double mmw = m_thermo->meanMolecularWeight();
+    const std::vector<double>& M = m_thermo->molecularWeights();
+    const double* y = m_thermo->massFractions();
+
+    std::vector<double>& a = m_spwork;
+
+    for (size_t k=0; k<m_nsp; ++k) {
+        dt[k] = 0.;
+        double xk = y[k]*mmw/M[k];
```
bryanwweber: Would it be faster to precompute this value for all of the species? I suppose it would use a little extra memory, but especially since the mole fraction is computed again in the nested loop, it seems worth it

bryanwweber (commented):
It's amazing to me that this feature only takes about 100 lines of code. Nice work!

src/transport/MixTransport.cpp
g3bk47: I thought the same. In theory, I could add a second "scratch" memory member like `m_spwork` to the `GasTransport` class (or even just to `MixTransport`) to hold the mole fractions. In the end, I decided against it to introduce minimal changes to the existing classes. My feeling is that it would not affect performance significantly, but I am open either way.

g3bk47:
I added `m_spwork2` (following the naming in `MultiTransport`) to MixTransport to avoid re-calculation of mole fractions.

src/transport/MixTransport.cpp
```diff
@@ -50,8 +51,80 @@ double MixTransport::thermalConductivity()
...
+    update_T();
+    update_C();
+    if (!m_bindiff_ok) {
+        updateDiff_T();
+    }
+    if (!m_viscwt_ok) {
+        updateViscosity_T();
+    }
+
+    const std::vector<double>& M = m_thermo->molecularWeights();
```
speth: This is already available as `m_mw` (via `GasTransport`).

src/transport/MixTransport.cpp
```diff
@@ -50,8 +51,80 @@ double MixTransport::thermalConductivity()
...
+        for(size_t j=0; j<m_nsp; ++j) {
+            if (j != k) {
+                sum += x[j]*m_phi(k,j);
+            }
+        };
+
+        a[k] = lambda_mono_k / (1. + 1.065 * sum / x[k]);
+    }
+
+    double kbT = Boltzmann * m_thermo->temperature();
```
speth: Already available as `m_kbt` from `GasTransport`.

src/transport/MixTransport.cpp
```diff
@@ -50,8 +51,80 @@ double MixTransport::thermalConductivity()
...
+    if (!m_viscwt_ok) {
+        updateViscosity_T();
+    }
+
+    const std::vector<double>& M = m_thermo->molecularWeights();
+    const double* y = m_thermo->massFractions();
+
+    std::vector<double>& a = m_spwork;
+    std::vector<double>& x = m_spwork2;
+    m_thermo->getMoleFractions(x.data());
```
speth: Mole fractions are already available in `m_molefracs`, with the caveat that these values have been adjusted to be >= `Tiny`.

src/transport/MixTransport.cpp
```diff
@@ -50,8 +51,80 @@ double MixTransport::thermalConductivity()
...
+
+    for (size_t k=0; k<m_nsp; ++k) {
+        dt[k] = 0.;
+        if (x[k] < Tiny)
+        {
+            a[k] = 0.;
```
speth: Formatting nitpick: We put the opening brace on the same line as the else:
```suggestion
        if (x[k] < Tiny) {
            a[k] = 0.;
```

src/transport/MixTransport.cpp
```diff
@@ -50,8 +51,80 @@ double MixTransport::thermalConductivity()
...
+            continue;
+        }
+
+        double lambda_mono_k = (15./4.) * m_visc[k] / M[k];
+
+        double sum = 0.;
+        for(size_t j=0; j<m_nsp; ++j) {
```
speth: Need a space after `for`.
```suggestion
        for (size_t j=0; j<m_nsp; ++j) {
```

src/transport/MixTransport.cpp
```diff
@@ -50,8 +51,80 @@ double MixTransport::thermalConductivity()
...
+            double log_tstar = std::log(kbT/m_epsilon(k,j));
+
+            int ipoly = m_poly[k][j];
+
+            double Cstar = 0.;
+            if (m_mode == CK_Mode) {
+                Cstar = poly6(log_tstar, m_cstar_poly[ipoly].data());
+            }
+            else {
+                Cstar = poly8(log_tstar, m_cstar_poly[ipoly].data());
```
speth: The `else` goes on the same line as the closing brace:
```suggestion
            } else {
                Cstar = poly8(log_tstar, m_cstar_poly[ipoly].data());
```

src/transport/MixTransport.cpp
```diff
@@ -50,8 +51,80 @@ double MixTransport::thermalConductivity()
...
+            }
+
+            double dt_T = ((1.2*Cstar - 1.0)/(m_bdiff(k,j)*rp)) / (M[k] + M[j]);
+
+            dt[k] += dt_T * (y[k]*a[j] - y[j]*a[k]);
+            dt[j] += dt_T * (y[j]*a[k] - y[k]*a[j]);
+        }
+    }
+
+    std::vector<double>& Dm = m_spwork;
```
speth: No need to write `std::` for `std::vector` -- this is so common that we have a `using` statement in `ct_defs.h`.

include/cantera/transport/UnityLewisTransport.h
```diff
@@ -60,6 +60,13 @@ class UnityLewisTransport : public MixTransport
         }
     }
 
+    //! Note: thermo diffusion is not enabled in the unity Lewis number model
```
speth:
```suggestion
    //! Thermal diffusion is not enabled in the unity Lewis number model.
```
You might also clarify that "not enabled" means that the returned coefficients are zero, to distinguish from other possibilities like raising `NotImplementedError`.

test/transport/gasTransportTest.cpp
```diff
@@ -177,11 +177,31 @@ TEST_F(GasTransportTest, thermalDiffCoeffs)
...
+        0, 2.00216138684567e-33, 2.79494070813552e-32,
+        1.26003408367763e-22, 4.26137699636566e-21
+    };
+
     vector<double> thermalDiff(nsp, -1);
-    // Thermal diffusion coefficients are zero for the mixture averaged model
     s_mix->getThermalDiffCoeffs(thermalDiff.data());
     for (size_t k = 0; k < nsp; k++) {
-        EXPECT_DOUBLE_EQ(thermalDiff[k], 0.0) << k;
+        EXPECT_NEAR(thermalDiff[k], thermalDiffMixRef[k], 1e-5) << k;
```
speth: This test doesn't really do anything, given the values are all much less than the absolute tolerance of 1e-5. You probably want to follow the logic of the comparison below that is used for the multicomponent model, where a combined relative and absolute error is approximated.

speth (changes requested):
Thanks, @g3bk47. This implementation looks pretty straightforward. I had just a few simplifications and formatting adjustments to suggest.

The one other thing you might consider would be updating one of the examples to make use of this. `adiabatic_flame.py` would be one option, to provide a comparison of the different options. Another is the ammonia flame simulated by `flame_speed.py` (where multicomponent calculations are disabled by default for the sake of run time, and this obviously relieves that concern).

g3bk47:
I think I addressed all comments. Let me know if I missed something.

For the `adiabatic_flame` example, the results are:
mixture-averaged flamespeed without Soret diffusion = 0.718743 m/s
mixture-averaged flamespeed with Soret diffusion = 0.638229 m/s
multicomponent flamespeed without Soret diffusion = 0.740712 m/s
multicomponent flamespeed with Soret diffusion = 0.653070 m/s

speth (approved):
Thanks for the updates, @g3bk47. This looks good to me.

g3bk47:
Great! Let me know if anything else needs to be done. The paper containing more validation cases already has a doi and should be available online in the next few days.

PR #1948, opened 2025-08-20
Make imports of external Python modules private

speth:

- Import external Python modules and functions using "private" names (for example, `_sys`) so they don't get included as members of the `cantera` module.
  - The one exception to this is `np`, which is used so extensively that I didn't want to have to change it. This one is removed by a `del` in `__init__.py`.
- Test that the resulting Python module does not (directly) contain any such modules or functions

This is an alternative approach to the same goal as #1947. There are a couple of things I discovered while preparing this.

- Because Pylance / VS Code builds its list of members through static analysis rather than by actually importing the module, some clever tricks to dynamically edit the list of namespace/module members don't work. Here, I still end up with the VS Code autocomplete showing me that `warnings`, `Path`, `np`, and `os` are members of the `cantera` namespace, despite being deleted in `__init__.py`
- The list of members obtained using the `dir` function, Jupyter's autocomplete (after import) and the autocomplete of `from cantera import ...` in IPython are correctly limited. 

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -2,12 +2,12 @@
 # at https://cantera.org/license.txt for license and copyright information.
 from __future__ import annotations as _annotations
 
-from ._cantera import *
 import numpy as np
-import csv as _csv
 import importlib as _importlib
-import warnings
-
+from ._cantera import (
+    DustyGasTransport, InterfaceKinetics, InterfacePhase, Kinetics, PureFluid,
+    SolutionArrayBase, ThermoPhase, Transport,
+)
 
 _pandas = None
 def _import_pandas():
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -1,11 +1,11 @@
 # This file is part of Cantera. See License.txt in the top-level directory or
 # at https://cantera.org/license.txt for license and copyright information.
-from __future__ import annotations
+from __future__ import annotations as _annotations
 
 from ._cantera import *
 import numpy as np
 import csv as _csv
-import importlib.metadata
+import importlib as _importlib
 import warnings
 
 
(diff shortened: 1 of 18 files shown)


interfaces/cython/cantera/composite.py
```diff
@@ -1,11 +1,11 @@
 # This file is part of Cantera. See License.txt in the top-level directory or
 # at https://cantera.org/license.txt for license and copyright information.
-from __future__ import annotations
+from __future__ import annotations as _annotations
```
bryanwweber, line 3: If we're only supporting 3.10+, do we need this anymore?

interfaces/cython/cantera/units.pyx
```diff
@@ -1,9 +1,9 @@
 # This file is part of Cantera. See License.txt in the top-level directory or
 # at https://cantera.org/license.txt for license and copyright information.
 
-from typing import Dict
-from collections.abc import Sequence
-import numbers
+from typing import Dict as _Dict
```
bryanwweber: If we're supporting 3.10+, this can be changed to drop the import and use `dict` lower case.

bryanwweber (commented):
Thanks @speth I think this makes a lot of sense. It seems that pyright also respects libraries that have `py.typed`, so if our `*.pyi` files do not export numpy publicly then it may resolve the problem. See: https://github.com/microsoft/pylance-release/discussions/4518

cc @TimothyEDawson

interfaces/cython/cantera/composite.py
TimothyEDawson: I had been wondering the same thing, but I believe it also enables some features which are not in 3.10. Though the only one I could definitively turn up was [forward references](https://peps.python.org/pep-0484/#forward-references), which I do utilize in several places.

ischoegl (approved):
Thanks, @speth … LGTM!

PR #1964, opened 2025-09-05
feat: pickling of SolutionArray objects (fixes #543)

### LINKED ISSUE #543: SolutionArray save / load

mruijzendaal:

- Implement `__reduce__` and `__copy__` methods for `SolutionArray`, such that it is picklable and copyable.

Fixes #543, continuation of #1426, see [cantera/enhancements/#162](https://github.com/Cantera/enhancements/issues/162#issuecomment-3256561347) 

In short, this pull request enables pickling and copying of SolutionArray objects with YAML as a proxy:  
```python
import cantera as ct
import pickle
import copy

sol = ct.Solution('gri30.yaml')  # load GRI-Mech 3.0 mechanism
solarr = ct.SolutionArray(sol, 100)  # create a solution array to store results

with open('sol_saved.pkl', 'wb') as f:
    pickle.dump(solarr, f)

with open('sol_saved.pkl', 'rb') as f:
    solarr_loaded = pickle.load(f)

copy.copy(solarr_loaded)
```

**Checklist**

(Edit: format proposed changes)

### DIFF
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -1306,11 +1306,61 @@ class SolutionArray(SolutionArrayBase):
         self.shape = self._api_shape()
         return meta
 
+    def _to_picklable(self):
+        import os
+        from pathlib import Path
+        import tempfile
+
+        # Save to a temporary YAML file
+        with tempfile.TemporaryDirectory() as tmpdir:
+            tf = Path(tmpdir, "cantera_solarr_save_tempfile.yaml")
+
+            # Remove the file if it already exists
+            if (tf.exists()):
+                os.unlink(tf)
+            self.save(tf, name="solarr", overwrite=True)
+
+            # Read the YAML file into a string and delete the file
+            with open(tf, 'r') as f:
+                yaml_data = f.read()
+
+        return {
+            'yaml_data': yaml_data,
+            'phase': self._phase, # Solution object, which is already picklable
+        }
+
+    @classmethod
+    def _from_pickle(cls, state):
+        from pathlib import Path
+        import os
+        import tempfile
+        # Recreate phase
+        phase = state.get('phase')
+        
+        # Create empty SolutionArray
+        arr = cls(phase)
+
+        # Write the YAML string to a temporary file
+        with tempfile.TemporaryDirectory() as tmpdir:
+            tf = Path(tmpdir, "cantera_solarr_load_tempfile.yaml")
+
+            # Remove the file if it already exists
+            if (tf.exists()):
+                os.unlink(tf)
+            with tf.open('w') as f:
+                f.write(state['yaml_data'])
+                f.flush()
+                
+            # Restore the SolutionArray from the yaml
+            arr.restore(tf, 'solarr')
+
+        return arr
+    
     def __reduce__(self):
-        raise NotImplementedError('SolutionArray object is not picklable')
+        return (self.__class__._from_pickle, (self._to_picklable(),))
 
     def __copy__(self):
-        raise NotImplementedError('SolutionArray object is not copyable')
+        return self.__class__._from_pickle(self._to_picklable())
 
 
 def _state2_prop(name, doc_source):
--- a/interfaces/cython/cantera/composite.py
+++ b/interfaces/cython/cantera/composite.py
@@ -8,6 +8,8 @@ from ._cantera import (
     DustyGasTransport, InterfaceKinetics, InterfacePhase, Kinetics, PureFluid,
     SolutionArrayBase, ThermoPhase, Transport,
 )
+import os as _os
+import tempfile as _tempfile
 
 _pandas = None
 def _import_pandas():
@@ -1307,22 +1309,11 @@ class SolutionArray(SolutionArrayBase):
         return meta
 
     def _to_picklable(self):
-        import os
-        from pathlib import Path
-        import tempfile
-
-        # Save to a temporary YAML file
-        with tempfile.TemporaryDirectory() as tmpdir:
-            tf = Path(tmpdir, "cantera_solarr_save_tempfile.yaml")
-
-            # Remove the file if it already exists
-            if (tf.exists()):
-                os.unlink(tf)
-            self.save(tf, name="solarr", overwrite=True)
-
-            # Read the YAML file into a string and delete the file
-            with open(tf, 'r') as f:
-                yaml_data = f.read()
+        with _tempfile.NamedTemporaryFile(suffix='.yaml') as temp_file:
+            _os.unlink(temp_file.name)
+            self.save(temp_file.name, name='solarr', overwrite=True)
+            with open(temp_file.name, 'r') as fid:
+                yaml_data = fid.read()
 
         return {
             'yaml_data': yaml_data,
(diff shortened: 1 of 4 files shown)


interfaces/cython/cantera/composite.py
```diff
@@ -1306,11 +1306,62 @@ def restore(self, fname, name=None, sub=None):
         self.shape = self._api_shape()
         return meta
 
+    def _to_picklable(self):
+        import os
+        from pathlib import Path
+        import tempfile
```
ischoegl: We usually use imports at the top of the module; they should be declared private, though, e.g.,
```
import tempfile as _tempfile
```

interfaces/cython/cantera/composite.py
```diff
@@ -1306,11 +1306,62 @@ def restore(self, fname, name=None, sub=None):
...
+
+        # Save to a temporary YAML file
+        with tempfile.TemporaryDirectory() as tmpdir:
+            tf = Path(tmpdir, "cantera_solarr_save_tempfile.yaml")
```
ischoegl: I was able to use `NamedTemporaryFile` - see top level comment.

interfaces/cython/cantera/composite.py
```diff
@@ -1306,11 +1306,62 @@ def restore(self, fname, name=None, sub=None):
...
+
+            # Remove the file if it already exists
+            if (tf.exists()):
+                os.unlink(tf)
+            self.save(tf, name="solarr", overwrite=True)
+
+            # Read the YAML file into a string and delete the file
```
ischoegl: Some of the comments can be skipped as code is descriptive enough by itself.

ischoegl (dismissed):
Hi @mruijzendaal ... thank you for the PR! I have a couple of minor comments. 

Regarding temporary files, I confirmed that the following works (at least on macOS), which should simplify things somewhat.
```
temp_file = tempfile.NamedTemporaryFile(suffix='.yaml')
solarr.save(temp_file.name, name='solarr', overwrite=True)
with open(temp_file.name, 'r') as fid:
    yaml_data = fid.read()
```

_PS: In addition, feel free to add yourself to the `AUTHORS.md` file._

mruijzendaal:
Thanks for the review and feedback!

> Regarding temporary files, I confirmed that the following works (at least on macOS), which should simplify things somewhat.
> 
> ```
> temp_file = tempfile.NamedTemporaryFile(suffix='.yaml')
> solarr.save(temp_file.name, name='solarr', overwrite=True)
> with open(temp_file.name, 'r') as fid:
>     yaml_data = fid.read()
> ```

That's weird, copying your code verbatim doesn't work on my system (MacOS 15.6.1, Python 3.11.10, cantera 3.1.0)
```python
import tempfile
import cantera as ct

sol = ct.Solution('gri30.yaml') 
solarr = ct.SolutionArray(sol, 10)

temp_file = tempfile.NamedTemporaryFile(suffix='.yaml')
solarr.save(temp_file.name, name='solarr', overwrite=True)
with open(temp_file.name, 'r') as fid:
    yaml_data = fid.read()
```

and gives the following error

```
CanteraError: 
InputFileError thrown by AnyMap::fromYamlFile:
Error on line 0 of /var/folders/q9/b2jlk94n25n5w5bjvp02rsc00000gn/T/tmp1tcb07ez.yaml:
bad conversion
|  Line |
```

However, with an `os.unlink` added back in, it works:
```diff
import tempfile
import cantera as ct
+ import os

sol = ct.Solution('gri30.yaml') 
solarr = ct.SolutionArray(sol, 10)

temp_file = tempfile.NamedTemporaryFile(suffix='.yaml')
+ os.unlink(temp_file.name)
solarr.save(temp_file.name, name='solarr', overwrite=True)
with open(temp_file.name, 'r') as fid:
    yaml_data = fid.read()
```

I'll make the necessary changes implementing your requests.

ischoegl:
@mruijzendaal ... looks like there was an issue on my side (apparently I scrambled file names while testing; I had run into a similar error). I pushed a fix to the C++ code base for this PR that enables writing to empty files. I can now confirm that the following works:

```Python
import tempfile
import cantera as ct

sol = ct.Solution('gri30.yaml')
solarr = ct.SolutionArray(sol, 10)

with tempfile.NamedTemporaryFile(suffix='.yaml') as temp_file:
    solarr.save(temp_file.name, name='solarr', overwrite=True)
    with open(temp_file.name, 'r') as fid:
        yaml_data = fid.read()
```

You can verify that `temp_file` is gone outside the context manager.

ischoegl:
@mruijzendaal … Turns out that temporary file access is more restrictive on Windows, so additional tweaks were necessary. 

@bryanwweber / @speth … since I pushed to this PR also, would either of you provide a final review so this can be merged?

interfaces/cython/cantera/composite.py
```diff
@@ -1306,11 +1308,45 @@ def restore(self, fname, name=None, sub=None):
         self.shape = self._api_shape()
         return meta
 
+    def _to_picklable(self):
+        with _NamedTemporaryFile(suffix=".yaml", delete=False) as t_file:
```
bryanwweber: I don't think this needs to be a context manager, I think it can be a regular assignment


```suggestion
        t_file = _NamedTemporaryFile(suffix=".yaml", delete=False)
```

Alternatively, you can set `delete_on_close=False, delete=True` and put all usage of the file inside the context manager. Then `os.unlink` is not needed.

bryanwweber (commented):
This all seems fine, although I guess a function that supports a buffer would be more generally useful. Not needed here though!

interfaces/cython/cantera/composite.py
ischoegl: Definitely not necessary - it was a leftover from earlier tests. For the context manager, there's a race condition on Windows - the process doesn't want to share the file, which is avoided by the `delete=False` option.

ischoegl:
@bryanwweber … agreed that writing to a buffer would be useful, but that won’t change the structure of the pickled object (the dictionary key names fed to the pickle are the only things we need to settle on for good). Thus, I believe that using the temporary file output is an adequate solution.

interfaces/cython/cantera/composite.py
bryanwweber: My read of the docs is that setting `delete_on_close=False` allows the context manager to be used even on Windows, as long as any additional uses of the file are closed before exiting the temp file context manager. 

> delete is true but delete_on_close is false. Note, that in this case the additional opens that do not share delete access (e.g. created via builtin open()) must be closed before exiting the context manager, else the os.unlink() call on context manager exit will fail with a PermissionError.

Regardless, this solution works just as well! 😊

speth (approved):
This looks fine to me.

interfaces/cython/cantera/composite.py
```diff
@@ -1306,11 +1308,44 @@ def restore(self, fname, name=None, sub=None):
         self.shape = self._api_shape()
         return meta
 
+    def _to_picklable(self):
+        temp_file = _NamedTemporaryFile(suffix=".yaml", delete=False).name
```
bryanwweber, line 1312: I think we should keep a reference to the original temp file object so that it doesn't get garbage collected. I guess with `delete=False` there's not a concern about deleting the file, but I still think it makes sense

interfaces/cython/cantera/composite.py
```diff
@@ -1306,11 +1308,44 @@ def restore(self, fname, name=None, sub=None):
...
+            _unlink(temp_file)
+
+        return {
+            "yaml_data": yaml_data,
+            "phase": self._phase,  # Solution object, which is already picklable
+        }
+
+    @classmethod
+    def _from_pickle(cls, state):
+        phase = state.get("phase")  # Recreate Solution object from pickled state
```
bryanwweber, line 1328: Sorry I didn't catch this the first time. What happens if `phase` is `None`? I can foresee that happening for incorrectly formed input. It may be better to check for this case and raise a useful error?

bryanwweber (commented):
I guess Ray already merged this, but I thought I should mention these two very small things. 

> Thus, I believe that using the temporary file output is an adequate solution.

Yep, agreed

interfaces/cython/cantera/composite.py
ischoegl: I _think_ it is fine as `delete=False` as you said. _PS: reverting to the context manager (i.e., previous version) resolves this._

interfaces/cython/cantera/composite.py
ischoegl: Fwiw, `None` causes a pretty nasty crash - it needs to be caught elsewhere.

mruijzendaal:
Thanks for working with me on this one, and thanks for all the work maintaining Cantera. It is very useful for me and my colleagues at Maastricht University in The Netherlands. I hope to be able to contribute more in the coming months.

PR #1966, opened 2025-09-07
Deprecate ReactorBase::addThermo

speth:
The virtual `ReactorBase::addThermo` does not function as intended, since overrides are not executed when this method is called from the base class constructor. Derived classes needing customized behavior must implement it in their own constructors or in overrides of the initialize method.

This erroneous usage was mostly invisible because we have only been using these overrides to ensure that the `ThermoPhase` objects used in the "ideal gas" reactor types are in fact ideal gasses.

- Move checks out of overrides of `ReactorBase::addThermo`
- Deprecate `ReactorBase::addThermo`

**Checklist**


ischoegl (approved):
Thanks for catching this, @speth!

PR #1970, opened 2025-09-08
Use unicode superscript in Python docstrings / tweak pickle

ischoegl:

- Replace unicode superscripts in docstrings (in analogy to recent changes in C++ recommendations)
- Apply changes requested after #1964 was merged.
- Avoid creation of `Solution` and `SolutionArray` from `None` (the latter would have caused a hard crash)

As an aside: there could be an argument for typesetting units as math text, but that would result in a much larger scope/effort.

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/cti2yaml.py
+++ b/interfaces/cython/cantera/cti2yaml.py
@@ -512,7 +512,7 @@ class gas_transport:
         :param dipole:
             The permanent dipole moment in Debye. Default: 0.0
         :param polar:
-            The polarizability in A^3. Default: 0.0
+            The polarizability in Å³. Default: 0.0
         :param rot_relax:
             The rotational relaxation collision number at 298 K. Dimensionless.
             Default: 0.0
@@ -520,10 +520,10 @@ class gas_transport:
             Pitzer's acentric factor.  Dimensionless.
             Default: 0.0
         :param disp_coeff:
-            The dispersion coefficient in A^5
+            The dispersion coefficient in Å⁵.
             Default: 0.0
         :param quad_polar:
-            The quadrupole polarizability
+            The quadrupole polarizability.
             Default: 0.0
         """
         self.geometry = geom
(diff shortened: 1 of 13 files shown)


ischoegl:
This is ready for a review - making sure `Solution(None)` and `SolutionArray(None)` raise the appropriate errors took more iterations than anticipated.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -512,18 +512,18 @@ def __init__(self, geom, diam, well_depth, dipole=0.0, polar=0.0,
         :param dipole:
             The permanent dipole moment in Debye. Default: 0.0
         :param polar:
-            The polarizability in A^3. Default: 0.0
+            The polarizability in A³. Default: 0.0
```
speth: If we're going all-in on Unicode units, this should be Å³. Likewise below for the dispersion coefficient.

interfaces/cython/cantera/solutionbase.pyx
```diff
@@ -29,6 +29,9 @@ cdef class _SolutionBase:
     def __cinit__(self, infile='', name='', adjacent=(), *, origin=None,
                   yaml=None, thermo=None, species=(),
                   kinetics=None, reactions=(), init=True, **kwargs):
+        if not isinstance(infile, (str, _PurePath)):
+            raise TypeError("Invalid object type: expected a 'str' or path object for "
+                            f"parameter 'infile', but received {type(infile)!r}.")
```
speth: I think the resulting message would look better as
```suggestion
                            f"parameter 'infile', but received {type(infile).__name__!r}.")
```
so you get `... but received 'NoneType'.` instead of `... but received <class 'NoneType'>`.
```

speth (changes requested):
Thanks for propagating this style update. I had just a couple of suggestions.

speth (approved):

Issue #1935, opened 2025-08-03
Handle upcoming changes in Pandas 3.0 string representation


ischoegl:
**Problem description**

Several post-merge tests are failing. Also, a new test to track #1916 should be added ... _PS: done, see #1952_.

**System information**

- Cantera version: latest main

speth:
To make this issue a bit more descriptive, the current GitHub Actions logs provide the following details:
- The failing jobs are `ubuntu-22.04 with Python 3.14` and `ubuntu-24.04 with Python 3.14`, both of which also pull in "nightly" versions of NumPy, Cython, Pandas, and SciPy.
- The failing tests are `test/python/test_composite.py::TestSolutionArrayIO::test_write_csv_fancy` and `test/python/test_composite.py::TestSolutionArrayIO::test_write_csv_escaped`, both of which fail with the message: `TypeError: 'SolutionArray.restore_data' only works for dictionaries that contain ndarrays`
- This exception is raised from `SolutionArray.restore_data`

Reproducing this locally, it seems to be a result of a change in some types returned by Pandas, where what was a `np.ndarray` is now a `pd.core.arrays.string_.StringArrayNumpySemantics`. This is occurring with Pandas version 3.0.0.dev0+2333.g188b2dae7d compared to the current stable release 2.3.2. The Pandas [docs](https://pandas.pydata.org/pandas-docs/version/2.3/user_guide/migration-3-strings.html#string-migration-guide) have a page describing changes to how strings will be handled in Pandas 3.0, and this seems to be a symptom of that.

ischoegl:
Ugh. I believe the alternative would be to deprecate pandas support? There's also polars now, and I don't think it's our job to maintain all those interfaces. A simply export of keys and data blocks should suffice. Funny enough, I believe it was me who introduced pandas way back when our `SolutionArray` display, etc., was not satisfactory. That has changed, and we could still think about Cantera/enhancements#170?

speth:
Unfortunately, the issue is not with the `to_pandas` / `from_pandas` methods per se. What's broken is the cases where we rely on Pandas to read CSV input files because it handles some special cases better than `np.genfromtxt`. In any case, I already have a fix ready.

ischoegl:
So it's really https://github.com/Cantera/enhancements/issues/163 ...

PR #1974, opened 2025-09-11
[MATLAB] Add multi-OS MATLAB test workflows to CI

ssun30:

- Run MATLAB tests in CI for Ubuntu 22.02/24.02, MacOS 13/14/15, and Windows 2022/2025. 

**Checklist**


ischoegl (commented):
I'd prefer utilizing runners that build on previously compiled code within a single matrix job, similar to the approach taken by the .NET runners. Of course, this requires the generated CLib.

ischoegl (commented):
@ssun30 ... now that #1997 is merged (🎉), I believe it's possible to directly adopt the approach from .NET, which avoids recompilation, as you can download shared libraries and doxygen XML from artifacts.

ischoegl:
Finally got the MATLAB matrix job to run with artifacts (🎉) - turned out to be a little more complicated than .NET, as I had to resolve some naming issues on Linux/macOS, and use a Windows DLL that was compiled outside of conda. A couple of caveats:

* The failure on `macOS-14` is odd, as I don't see it on my machine
* The failure on `ubuntu-22.04` appears for an example that is set up incorrectly; this example saw some work in #2031, so a merge may fix this (the flow domain was set up incorrectly).
* The timeout on `windows-2025` is linked to significant overhead of the `outofprocess` mode (see some diagnostics provided by @ssun30 here: https://github.com/Cantera/cantera/pull/1997#issuecomment-3440097044). It is still curious that the overall run takes more than 4 times longer than on macOS. Fwiw, the `plug_flow_reactor.m` example takes about 30 minutes on Windows by itself (and around 5 minutes on macOS).

Regarding slow samples, it appears that the tagging with `Slow` does not skip the tests (the test job bypasses this).

ischoegl:
@speth ... after rebasing to switch to the generated CLib and updating the CI logic, I believe this is actually ready (I updated this PR, which was originally opened by @ssun30; the initial work pointed in a lot of the right directions).

I believe that I found a happy middle ground for MATLAB testing - run examples on `macOS` and skip them for other OS's. The `outofprocess` mode is slowing some of the MATLAB examples down a lot, but this is beyond the scope of this PR. Long-term, it is likely best to test samples separately from unit testing (i.e., same as we already do on Python), but this is likewise beyond the scope.

PS: I pushed some simplifications

.github/workflows/main.yml
```diff
@@ -1236,3 +1194,117 @@ jobs:
...
+        brew install --display-times libomp hdf5
+      if: runner.os == 'macOS'
+    - name: Install Apt dependencies (Ubuntu)
+      run: |
+        echo "Running on `cat /proc/cpuinfo | grep 'model name' | head -n 1 | cut -d ':' -f 2`"
+        sudo apt update
+        sudo apt install libhdf5-dev libfmt-dev libopenblas0-openmp
+      if: runner.os == 'Linux'
+    - name: Install Python dependencies
+      # Install Python dependencies, which are used by 'dotnet build'
```
speth: I assume this should be:
```suggestion
      # Install Python dependencies, which are used by sourcegen
```

.github/workflows/main.yml
```diff
@@ -1236,3 +1194,117 @@ jobs:
...
+      uses: actions/download-artifact@v5
+      with:
+        name: libcantera_shared-${{ matrix.os }}.so
+        path: build/lib
+      if: runner.os == 'Linux'
+    - name: Create library symlinks (Linux)
+      working-directory: build/lib
+      run: |
+        ln -s libcantera_shared.so libcantera_shared.so.3.2.0
+        ln -s libcantera_shared.so.3.2.0 libcantera_shared.so.3
```
speth: Why are these symlinks required? Can't the Matlab interface just load the unversioned library?

.github/workflows/main.yml
```diff
@@ -1236,3 +1194,117 @@ jobs:
...
+          ctToolboxDir = fullfile(ctDir, 'interfaces', 'matlab_experimental');
+          ctIncludeDir = fullfile(ctDir, 'interfaces', 'clib', 'include');
+          ctLibDir = fullfile(ctDir, 'build', 'lib');
+          addpath(genpath(ctToolboxDir));
+          ctBuildInterface(ctToolboxDir, ctIncludeDir, ctLibDir);
+    - name: Run tests and examples (macOS)
+      uses: matlab-actions/run-tests@a80b208946040c701ae65c1bce73ba7ec4810757 # v2.1.2
+      with:
+        select-by-folder: test/matlab_experimental
+      if: runner.os == 'macOS'
```
speth: Given the macOS package doesn't suffer from library compatibility issues (at least as it's built in this CI process), can we do this run using the `inprocess` mode?

ischoegl (commented):
Had to go back and fix an issue with a runner failing silently. Based on #2030, I further added more MATLAB versions to the tests. R2024a/b and R2025a work, but R2023b failed. Fwiw, the tests with issues reported in #2034 are now disabled.

The inclusion of older MATLAB versions surfaced another issue caused by Linux glibc, see #1935.

.github/workflows/main.yml
```diff
@@ -1236,3 +1192,120 @@ jobs:
...
+    - name: Install sourcegen
+      run: python -m pip install -e interfaces/sourcegen
+    - name: Generage CLib headers
+      run: sourcegen --api=clib --output=interfaces/clib
+    - name: Set LD_PRELOAD environment variable for MATLAB (Linux)
+      run: |
+        LIB_STDCXX=$(ldconfig -p | grep libstdc++.so.6 | awk '{print $4}' | head -n 1)
+        LIB_OPENBLAS=$(ldconfig -p | grep libopenblas.so.0 | awk '{print $4}' | head -n 1)
+        LIB_LAPACK=$(ldconfig -p | grep liblapack.so.3 | awk '{print $4}' | head -n 1)
+        echo "LD_PRELOAD=$LIB_STDCXX:$LIB_OPENBLAS:$LIB_LAPACK" >> $GITHUB_ENV
```
speth, line 1295: Is this still necessary, with the `outofprocess` mode available on Linux (even with its performance limitations)? If this is still required, then we haven't solved any of the library compatibility problems.

.github/workflows/main.yml
```diff
@@ -1236,3 +1192,120 @@ jobs:
...
+      uses: actions/download-artifact@v5
+      with:
+        name: libcantera_shared-${{ matrix.os }}.so
+        path: build/lib
+      if: runner.os == 'Linux'
+    - name: Create library symlinks (Linux)
+      working-directory: build/lib
+      run: |
+        ln -s libcantera_shared.so libcantera_shared.so.3.2.0
+        ln -s libcantera_shared.so.3.2.0 libcantera_shared.so.3
```
speth, line 1256: Why are these symlinks required? Can't the Matlab interface just load the unversioned library?

interfaces/matlab_experimental/Utility/ctLib.m
```diff
@@ -1,55 +1,69 @@
...
-        if isempty(versioned)
-            warning('No versioned library found, falling back to unversioned');
-            str = libDir + baseName + ".dylib";
+
+        scores = arrayfun(@(f) versionScore(f.name), files);
+        hasVersion = scores > 0;
+        if any(hasVersion)
+            [~, idx] = max(scores);
+            str = string(fullfile(libDir, files(idx).name));
         else
```
speth, line 46: I realize this was a preexisitng condition, but I think this mechanism of interacting with the versioned shared libraries is completely ineffective. There is nothing to make sure that the Matlab toolbox corresponds to the version of `libcantera` that is found here. In the absence of a more robust mechanism for this, I would suggest just taking the simple approach and linking to the unversioned library name.

I also have no idea what a version "score" is, but that's sort of a moot point compared to the larger issue here.

test/matlab_experimental/ctTestSamples.m
```diff
@@ -30,9 +30,10 @@ function removeSampleFolder(self)
             % re-enabled when the `Flame` classes are re-implemeneted in MATLAB.
             ExampleScript = {
                 'equil', 'isentropic', 'reactor1', 'reactor2', 'surf_reactor', ...
-                'periodic_cstr', 'plug_flow_reactor', 'lithium_ion_battery', ...
+                'periodic_cstr', 'lithium_ion_battery', ...
                 'rankine', 'prandtl1', 'prandtl2', 'catcomb', 'diamond_cvd', ...
-                'ignite_hp', 'ignite_uv', 'flame1', ...
+                'flame1', ...
+                % 'ignite_hp', 'ignite_uv', 'plug_flow_reactor', ...
                 % 'ignite', 'flame2', 'diff_flame',
```
speth: I think it would be helpful to add a comment distinguishing why particular examples are disabled here -- the `plug_flow_reactor` example works, it's just very slow.

speth (changes requested):
Thanks, @ischoegl and @ssun30, I think it's good to have this more comprehensive testing. My suggestions here are mainly in the hope of simplifying a few things.

.github/workflows/main.yml
ischoegl: This is an excellent point, but I'd prefer to handle this in a follow-up PR - I had created #2034 for that reason. It should be addressed for any usage, not just the CI.

.github/workflows/main.yml
ischoegl: I don't know. Again, I'd prefer to handle this in a follow up that focuses on `LD_PRELOAD` - see new issue #2035.

.github/workflows/main.yml
ischoegl: MATLAB is unable to resolve these. It requires `libcantera_shared.3.2.0.dylib` for macOS and `libcantera_shared.so.3` for Linux. This is the reason why there libraries need to be resolved by extra logic. Took me a while to figure out why @ssun30 had introduced this logic - I updated in the hope to make it easier to follow.

interfaces/matlab_experimental/Utility/ctLib.m
ischoegl: It's targeting a different purpose - making sure that MATLAB finds a library that it is able to load. I'll add a comment to the version score (in effect, a way to ensure that we can sort version numbers according to integers).

Ensuring that libcantera matches would have to be tacked in `ctLoad`. Once we can access `ctVersion` from the shared library, it can be checked.

ischoegl:
> Thanks, @ischoegl and @ssun30, I think it's good to have this more comprehensive testing. My suggestions here are mainly in the hope of simplifying a few things.

Thanks for the review! I took care of the straightforward suggestions. Regarding simplifications, it appears that some of the apparently complex functions are required to make things run. I had already opened #2034 and #2035 to track associated issues.

I'd prefer to push `inprocess` and `LD_PRELOAD` to follow-up PRs. When picking up @ssun30's work here, my main hope was to ensure that the generated CLib works on multiple OSs while using artifacts rather than recompiling from scratch. That would allow things to move to #2031, which will be my last MATLAB contribution before the final 3.2.0 release.

speth (approved):
No further comments.

PR #1976, opened 2025-09-12
advance limit handling to improve performance (#1453)

### LINKED ISSUE #1453: Reactor.advance limits can be exceeded greatly
**Problem description**

Depending on the output times given to `Reactor.advance`, specified limits on the change in one or more state variables can be exceeded by a wide margin. In some cases, this seems to stem from the relevant variable having a near-zero derivative instantaneously, permitting a large interval to be passed on to the internal `advance` method, which then takes many time steps over which the variable changes by more than the expected amount. However, there are other cases where it seems like the instantaneous rate of change should predict the need for a shorter time step.

I originally encountered this due to surprising failures for `TestReactor.test_advance_with_limits` in #1452, where it seems that the fact that test passes at all is kind of a fluke.

Given the intention for this feature to be a robust way of getting nice output even for inexperienced users (see discussion in #628/#629), I think some improvement here is warranted.

**Steps to reproduce**

The following is based on `TestReactor.test_advance_with_limits`, with a few modifications.

```py
import cantera as ct

def integrate(limit=None):
    P0 = 10 * ct.one_atm
    T0 = 990
    X0 = 'CH4:1.0, O2:1.0, AR:5.0'
    gas.TPX = T0, P0, X0
    r1 = ct.IdealGasReactor(gas)
    net = ct.ReactorNet([r1])
    ix = net.global_component_index('CH4', 0)
    r1.set_advance_limit('CH4', limit)

    tEnd = 1.0
    tStep = 0.1
    nSteps = 0

    states = ct.SolutionArray(gas, extra=['t', 'steps'])
    t = tStep
    yp = net.get_state()[ix]
    steps_p = net.solver_stats['steps']
    dy_max = 0
    while t < tEnd:
        steps = net.solver_stats['steps']
        states.append(TPX=r1.thermo.TPX, t=net.time, steps=(steps-steps_p))
        steps_p = steps
        t_curr = net.advance(t)
        y = net.get_state()[ix]
        dy_max = max(dy_max, abs(y-yp))
        yp = y
        nSteps += 1
        if t_curr >= t:
            t += tStep

    print(f'case: {limit=}, {dy_max=}, {nSteps=}')
    return nSteps, states

n_baseline, states_baseline = integrate()
n_advance_coarse, states_coarse = integrate(0.005)
n_advance_fine, states_fine = integrate(0.001)

f, ax = plt.subplots(figsize=(5,3))
ax.plot(states_fine.t, states_fine('CH4').Y, '.-', label='fine')
ax.plot(states_coarse.t, states_coarse('CH4').Y, '.-', label='coarse')
ax.legend();
```

**Behavior**

```
case: limit=None, dy_max=0.06457704195575183, nSteps=10
case: limit=0.005, dy_max=0.0645770419565304, nSteps=10
case: limit=0.001, dy_max=0.014385584336446846, nSteps=90
```
Figure 42

As you can see, in the second case, the desired limit is exceeded by more than a factor of 10, and the output jumps over the ignition event entirely. Looking at `states_coarse.steps`, you can see that integrator takes over 2000 time steps between the second and third output points.

With a finer limit, the behavior is better, but this isn't even a guaranteed path to success -- if `tStep` is increased to 0.2 s, then both values of the advance limit miss everything. Also, even here, there is a strange gap toward the end of the ignition event where the fuel mass fraction drops by quite a bit with no intervening output.

**System information**

- Cantera version: 2.6.0 or current `main` branch at 399e1cba3
- Python 3.10 used for the example above, but affects all versions

wandadars:
This was just an attempt to try and resolve the issue discussed in #1453. This may be too invasive to be useful to the code base, but just wanted to open a pull request just in case.

Event-based limiter: Enforce advance limits with a CVODE root function so integration stops exactly when any component reaches its configured limit, preventing large overshoots between output points. Hooked via CVodeRootInit and handled CV_ROOT_RETURN.

Closes #1453 

<details>
<summary>First example case from #1453 </summary>

```
import cantera as ct

def integrate(limit=None):
    P0 = 10 * ct.one_atm
    T0 = 990
    X0 = 'CH4:1.0, O2:1.0, AR:5.0'
    gas.TPX = T0, P0, X0
    r1 = ct.IdealGasReactor(gas)
    net = ct.ReactorNet([r1])
    ix = net.global_component_index('CH4', 0)
    r1.set_advance_limit('CH4', limit)

    tEnd = 1.0
    tStep = 0.1
    nSteps = 0

    states = ct.SolutionArray(gas, extra=['t', 'steps'])
    t = tStep
    yp = net.get_state()[ix]
    steps_p = net.solver_stats['steps']
    dy_max = 0
    while t < tEnd:
        steps = net.solver_stats['steps']
        states.append(TPX=r1.thermo.TPX, t=net.time, steps=(steps-steps_p))
        steps_p = steps
        t_curr = net.advance(t)
        y = net.get_state()[ix]
        dy_max = max(dy_max, abs(y-yp))
        yp = y
        nSteps += 1
        if t_curr >= t:
            t += tStep

    print(f'case: {limit=}, {dy_max=}, {nSteps=}')
    return nSteps, states

n_baseline, states_baseline = integrate()
n_advance_coarse, states_coarse = integrate(0.005)
n_advance_fine, states_fine = integrate(0.001)

f, ax = plt.subplots(figsize=(5,3))
ax.plot(states_fine.t, states_fine('CH4').Y, '.-', label='fine')
ax.plot(states_coarse.t, states_coarse('CH4').Y, '.-', label='coarse')
ax.legend()
# Show plot on screen
plt.show()
```

</details>

The plot from that case using this updated approach is:

Figure_1

I also ran the case from the original issue report with the unit test parameters.

<details>
<summary>Click to view code example 1 </summary>

```
import cantera as ct


def integrate(limit=None):
    P0 = 10 * ct.one_atm
    T0 = 1100
    X0 = 'H2:1.0, O2:0.5, AR:8.0'
    gas.TPX = T0, P0, X0
    r1 = ct.IdealGasReactor(gas)
    net = ct.ReactorNet([r1])
    ix = net.global_component_index('H2', 0)
    r1.set_advance_limit('H2', limit)

    # Unit test parameters
    tEnd = 0.01
    tStep = 1e-3
    nSteps = 0

    #tEnd = 0.01
    #tStep = 1.1e-4
    #nSteps = 0

    states = ct.SolutionArray(gas, extra=['t', 'steps'])
    t = tStep
    yp = net.get_state()[ix]
    steps_p = net.solver_stats['steps']
    dy_max = 0
    while t < tEnd:
        steps = net.solver_stats['steps']
        states.append(TPX=r1.thermo.TPX, t=net.time, steps=(steps-steps_p))
        steps_p = steps
        t_curr = net.advance(t)
        y = net.get_state()[ix]
        dy_max = max(dy_max, abs(y-yp))
        yp = y
        nSteps += 1
        if t_curr >= t:
            t += tStep

    print(f'case: {limit=}, {dy_max=}, {nSteps=}')
    return nSteps, states

n_baseline, states_baseline = integrate()
n_advance_coarse, states_coarse = integrate(0.01)
n_advance_fine, states_fine = integrate(0.001)

fig, axes = plt.subplots(
    nrows=4, ncols=1, figsize=(6, 10), sharex=True,
    gridspec_kw={'height_ratios': [2, 1, 1, 1], 'hspace': 0.25}
)
ax_main, ax_fine, ax_coarse, ax_baseline = axes

# Main plot: fine and coarse together
ax_main.plot(states_fine.t, states_fine('H2').Y, '-', color='orange', label='fine (line)')
ax_main.plot(states_fine.t, states_fine('H2').Y, '.', color='orange', label='fine (points)')
ax_main.plot(states_coarse.t, states_coarse('H2').Y, '--', color='blue', label='coarse (dashed)')
ax_main.plot(states_coarse.t, states_coarse('H2').Y, 'o', color='blue', label='coarse (points)')
ax_main.plot(states_baseline.t, states_baseline('H2').Y, '-', color='green', label='baseline')
ax_main.plot(states_baseline.t, states_baseline('H2').Y, '.', color='green')
ax_main.set_ylabel('Y(H2)')
ax_main.legend(loc='best')

# Subplot 1: fine only
ax_fine.plot(states_fine.t, states_fine('H2').Y, '-', color='orange', label='fine')
ax_fine.plot(states_fine.t, states_fine('H2').Y, '.', color='orange')
ax_fine.set_ylabel('Y(H2)')
ax_fine.legend(loc='best')

# Subplot 2: coarse only
ax_coarse.plot(states_coarse.t, states_coarse('H2').Y, '--', color='blue', label='coarse')
ax_coarse.plot(states_coarse.t, states_coarse('H2').Y, 'o', color='blue')
ax_coarse.set_ylabel('Y(H2)')
ax_coarse.legend(loc='best')

# Subplot 3: baseline only
ax_baseline.plot(states_baseline.t, states_baseline('H2').Y, '-', color='green', label='baseline')
ax_baseline.plot(states_baseline.t, states_baseline('H2').Y, '.', color='green')
ax_baseline.set_ylabel('Y(H2)')
ax_baseline.set_xlabel('time [s]')
ax_baseline.legend(loc='best')

plt.show()

```

</details>

This was using the  default values from the unit test
Figure_1

And this was using the tighter tolerance values from the issue discussion.

Figure_1

```
case: limit=None, dy_max=0.00551594315108739, nSteps=90
case: limit=0.01, dy_max=0.00551594315108739, nSteps=90
case: limit=0.001, dy_max=0.0010000000000554895, nSteps=95
```

For reference, this is the behavior of the Cantera version (3.1.0) running the same script with the tighter tolerances.

Figure_1

```
case: limit=None, dy_max=0.0055159431511252995, nSteps=90
case: limit=0.01, dy_max=0.005365298679090868, nSteps=93
case: limit=0.001, dy_max=0.005365298679090868, nSteps=93
```

### DIFF
--- a/include/cantera/numerics/FuncEval.h
+++ b/include/cantera/numerics/FuncEval.h
@@ -145,16 +145,18 @@ public:
     /**
      * Evaluate the event/root functions currently in play.
      * Integrators invoke this whenever root finding is enabled; implementations
-     * should fill `gout` with the function values and return 0 on success.
+     * should fill `gout` with the function values.
      * @param[in] t Time at which to evaluate the root functions
      * @param[in] y Current solution vector at time *t* of length neq()
      * @param[out] gout Array of length nRootFunctions() to be filled with the
      *      values of the root functions
-     * @returns 0 on success
      */
-    virtual int evalRootFunctions(double t, const double* y, double* gout) {
-        return 0;
-    }
+    virtual void evalRootFunctions(double t, const double* y, double* gout) { }
+
+    //! Wrapper for evalRootFunctions that converts exceptions to return codes.
+    //! @returns 0 for a successful evaluation, 1 after a potentially-
+    //!     recoverable error, or -1 after an unrecoverable error.
+    int evalRootFunctionsNoThrow(double t, const double* y, double* gout);
 
     //! Fill in the vector *y* with the current state of the system.
     //! Used for getting the initial state for ODE systems.
--- a/include/cantera/numerics/FuncEval.h
+++ b/include/cantera/numerics/FuncEval.h
@@ -137,12 +137,15 @@ public:
      */
     int preconditioner_solve_nothrow(double* rhs, double* output);
 
-    //! Number of root functions used for event detection
+    //! Number of event/root functions exposed to the integrator (0 disables root finding)
     virtual size_t nRootFunctions() const {
         return 0;
     }
 
-    //! Evaluate root functions used for event detection
+    /** Evaluate the event/root functions currently in play.
+     * Integrators invoke this whenever root finding is enabled; implementations
+     * should fill `gout` with the function values and return 0 on success.
+     */
     virtual int evalRootFunctions(double t, const double* y, double* gout) {
         return 0;
     }
--- a/include/cantera/numerics/FuncEval.h
+++ b/include/cantera/numerics/FuncEval.h
@@ -137,6 +137,16 @@ public:
      */
     int preconditioner_solve_nothrow(double* rhs, double* output);
 
+    //! Number of root functions used for event detection
+    virtual size_t nRootFunctions() const {
+        return 0;
+    }
+
+    //! Evaluate root functions used for event detection
+    virtual int evalRootFunctions(double t, const double* y, double* gout) {
+        return 0;
+    }
+
     //! Fill in the vector *y* with the current state of the system.
     //! Used for getting the initial state for ODE systems.
     virtual void getState(double* y) {
--- a/include/cantera/numerics/FuncEval.h
+++ b/include/cantera/numerics/FuncEval.h
@@ -142,9 +142,15 @@ public:
         return 0;
     }
 
-    /** Evaluate the event/root functions currently in play.
+    /**
+     * Evaluate the event/root functions currently in play.
      * Integrators invoke this whenever root finding is enabled; implementations
      * should fill `gout` with the function values and return 0 on success.
+     * @param[in] t Time at which to evaluate the root functions
+     * @param[in] y Current solution vector at time *t* of length neq()
+     * @param[out] gout Array of length nRootFunctions() to be filled with the
+     *      values of the root functions
+     * @returns 0 on success
      */
     virtual int evalRootFunctions(double t, const double* y, double* gout) {
         return 0;
--- a/include/cantera/numerics/FuncEval.h
+++ b/include/cantera/numerics/FuncEval.h
@@ -137,7 +137,8 @@ public:
      */
     int preconditioner_solve_nothrow(double* rhs, double* output);
 
-    //! Number of event/root functions exposed to the integrator (0 disables root finding)
+    //! Number of event/root functions exposed to the integrator.
+    //! 0 indicates root finding is disabled.
     virtual size_t nRootFunctions() const {
         return 0;
     }
--- a/include/cantera/numerics/Integrator.h
+++ b/include/cantera/numerics/Integrator.h
@@ -93,9 +93,7 @@ public:
      * ReactorNet enforces advance limits).
      * @param nroots Number of root functions
      */
-    virtual void setRootFunctionCount(size_t nroots) {
-        (void) nroots;
-    }
+    virtual void setRootFunctionCount([[maybe_unused]] size_t nroots) {}
 
     //! Current value of the independent variable tracked by the integrator
     virtual double currentTime() const = 0;
--- a/include/cantera/numerics/Integrator.h
+++ b/include/cantera/numerics/Integrator.h
@@ -87,12 +87,17 @@ public:
         warn("setLinearSolverType");
     }
 
-    //! Configure the number of root functions evaluated by the integrator
+    //! Configure how many event/root functions the integrator should monitor
+    /*!
+     * Callers toggle this to enable/disable root finding on the fly (when
+       ReactorNet enforces advance limits).
+     * @param nroots Number of root functions
+     */
     virtual void setRootFunctionCount(size_t nroots) {
         (void) nroots;
     }
 
-    //! Current value of the independent variable represented by the integrator
+    //! Current value of the independent variable tracked by the integrator
     virtual double currentTime() const = 0;
 
     //! Set preconditioner used by the linear solver
--- a/include/cantera/numerics/Integrator.h
+++ b/include/cantera/numerics/Integrator.h
@@ -87,10 +87,10 @@ public:
         warn("setLinearSolverType");
     }
 
-    //! Configure how many event/root functions the integrator should monitor
-    /*!
+    /** Configure how many event/root functions the integrator should monitor
+     *
      * Callers toggle this to enable/disable root finding on the fly (when
-       ReactorNet enforces advance limits).
+     * ReactorNet enforces advance limits).
      * @param nroots Number of root functions
      */
     virtual void setRootFunctionCount(size_t nroots) {
--- a/include/cantera/numerics/Integrator.h
+++ b/include/cantera/numerics/Integrator.h
@@ -87,6 +87,14 @@ public:
         warn("setLinearSolverType");
     }
 
+    //! Configure the number of root functions evaluated by the integrator
+    virtual void setRootFunctionCount(size_t nroots) {
+        (void) nroots;
+    }
+
+    //! Current value of the independent variable represented by the integrator
+    virtual double currentTime() const = 0;
+
     //! Set preconditioner used by the linear solver
     /*!
      * @param preconditioner preconditioner object used for the linear solver
--- a/include/cantera/numerics/Integrator.h
+++ b/include/cantera/numerics/Integrator.h
@@ -87,8 +87,8 @@ public:
         warn("setLinearSolverType");
     }
 
-    /** Configure how many event/root functions the integrator should monitor
-     *
+    //! Configure how many event/root functions the integrator should monitor
+    /**
      * Callers toggle this to enable/disable root finding on the fly (when
      * ReactorNet enforces advance limits).
      * @param nroots Number of root functions
--- a/include/cantera/zeroD/ReactorNet.h
+++ b/include/cantera/zeroD/ReactorNet.h
@@ -363,13 +363,19 @@ public:
     //! @param settings the settings map propagated to all reactors and kinetics objects
     virtual void setDerivativeSettings(AnyMap& settings);
 
-    //! Root function used by the integrator to detect advance limit crossings
-    //! during calls to advance(t, applylimit=True). Returns 0 on success.
-    //! The function evaluates gout[0] = 1 - max_i(|y[i]-y_base[i]|/limit[i]).
-    //! If advance limits are disabled or the limit check is inactive, gout[0]
-    //! is set to a positive value so that no root is detected.
+    //! Root finding is enabled only while enforcing advance limits
     size_t nRootFunctions() const override;
+
+    //! Relay the advance-limit root function to the integrator
     int evalRootFunctions(double t, const double* y, double* gout) override;
+
+    /** Root callback used to stop integration when an advance limit is reached
+     *
+     * When limit enforcement is active this evaluates
+     * `gout[0] = 1 - max_i(|y[i] - y_base[i]| / limit[i])`, so a zero signals that
+     * some component has reached its limit. If limits are inactive the method fills
+     * `gout[0]` with a positive value to prevent a false root. Returns 0 on success.
+     */
     int advanceLimitRootFunc(double t, const double* y, double* gout);
 
 protected:
--- a/include/cantera/zeroD/ReactorNet.h
+++ b/include/cantera/zeroD/ReactorNet.h
@@ -368,6 +368,8 @@ public:
     //! The function evaluates gout[0] = 1 - max_i(|y[i]-y_base[i]|/limit[i]).
     //! If advance limits are disabled or the limit check is inactive, gout[0]
     //! is set to a positive value so that no root is detected.
+    size_t nRootFunctions() const override;
+    int evalRootFunctions(double t, const double* y, double* gout) override;
     int advanceLimitRootFunc(double t, const double* y, double* gout);
 
 protected:
--- a/include/cantera/zeroD/ReactorNet.h
+++ b/include/cantera/zeroD/ReactorNet.h
@@ -366,9 +366,12 @@ public:
     //! Root finding is enabled only while enforcing advance limits
     size_t nRootFunctions() const override;
 
-    //! Evaluate the advance-limit root function used to stop integration once a limit is met.
-    //! When limits are active this sets `gout[0] = 1 - max_i(|y[i]-y_base[i]| / limit[i])`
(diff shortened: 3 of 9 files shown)


src/numerics/CVodesIntegrator.cpp
```diff
@@ -90,6 +91,20 @@ extern "C" {
...
+    {
+        // user_data points to the FuncEval (e.g., ReactorNet)
+        auto* f = static_cast<Cantera::FuncEval*>(user_data);
+        // Only ReactorNet implements the limit-check root function
+        if (auto* net = dynamic_cast<Cantera::ReactorNet*>(f)) {
+            return net->advanceLimitRootFunc(t, NV_DATA_S(y), gout);
+        }
+        // Default: no root
+        gout[0] = 1.0;
+        return 0;
```
speth: I suspect you can avoid breaking the abstraction between `CVodesIntegrator` and `ReactorNet` by making the call to `CVodeRootInit` optional, i.e. triggered only if a call to `advance` with limits applied is made. The CVodes [documentation](https://sundials.readthedocs.io/en/latest/cvode/Usage/index.html#rootfinding-initialization-function) seems to suggest that repeated calls to `CVodeRootInit` changing the number of root functions to evaluate (including setting it to zero) are allowed.

src/zeroD/ReactorNet.cpp
```diff
@@ -209,51 +209,67 @@ double ReactorNet::advance(double time, bool applylimit)
...
+                     " |Δ| = {:11.6g}, limit = {:9.4g}, ratio = {:9.4g}\n",
+                     jmax, dt, best_delta, best_limit, max_ratio);
         }
-        t = .5 * (m_time + t);
     }
-    advance(t);
-    return t;
+
+    // m_time is tracked via callbacks during integration
+    return m_time;
```
speth, line 315: I think the current test failure (integrator hanging) stems from this. There is a tricky issue with both `CVodesIntegrator` and `ReactorNet` having a `m_time` member, where the former is shadowed by the latter. What I think is happening here is that only the former is being correctly updated to the user-specified time with the interpolated solution while the latter ends up having the time reached by the integrator internally on the last step.

speth (commented):
Thanks for looking into this, @wandadars. I think using the built-in CVODES functionality for this is a good approach to consider. I just wanted to share a couple of suggestions that might help with the implementation.

wandadars:
Thanks @speth . I totally spaced out and forgot that I had opened this PR. I'll look at this again this weekend and see if I can clean it up.

ischoegl (commented):
Thanks! It's great to see that there's actually a way to handle this. My only request would be to add a test so this is covered.

I hope to see this in the final version of 3.2 ...

include/cantera/numerics/FuncEval.h
```diff
@@ -137,6 +137,19 @@ class FuncEval
...
+        return 0;
+    }
+
+    /** Evaluate the event/root functions currently in play.
+     * Integrators invoke this whenever root finding is enabled; implementations
+     * should fill `gout` with the function values and return 0 on success.
+     */
+    virtual int evalRootFunctions(double t, const double* y, double* gout) {
+        return 0;
+    }
```
speth: Since this function is called by CVODES, I think it also needs to have the protection against throwing exceptions that we implement for the other callback methods, for example the virtual `eval` method which is called by the non-virtual `evalNoThrow` method, where the latter is the one that is set as the CVODES callback.

Also, as a minor nitpick, I would prefer formatting the docstring as:

```suggestion
    /**
     * Evaluate the event/root functions currently in play.
     * Integrators invoke this whenever root finding is enabled; implementations
     * should fill `gout` with the function values and return 0 on success.
     */
    virtual int evalRootFunctions(double t, const double* y, double* gout) {
        return 0;
    }
```
i.e., avoid starting the comment on the `/**` line. If you want a more compact docstring, I'd suggest the `//!` syntax.

include/cantera/numerics/Integrator.h
```diff
@@ -87,6 +87,19 @@ class Integrator
...
 
+    /** Configure how many event/root functions the integrator should monitor
+     *
+     * Callers toggle this to enable/disable root finding on the fly (when
+     * ReactorNet enforces advance limits).
+     * @param nroots Number of root functions
+     */
+    virtual void setRootFunctionCount(size_t nroots) {
+        (void) nroots;
+    }
```
speth: A nicer way to annotate this in C++17:
```suggestion
    virtual void setRootFunctionCount([[maybe_unused]] size_t nroots) {}
```

include/cantera/zeroD/ReactorNet.h
```diff
@@ -363,6 +363,21 @@ class ReactorNet : public FuncEval
...
+    int evalRootFunctions(double t, const double* y, double* gout) override;
+
+    /** Root callback used to stop integration when an advance limit is reached
+     *
+     * When limit enforcement is active this evaluates
+     * `gout[0] = 1 - max_i(|y[i] - y_base[i]| / limit[i])`, so a zero signals that
+     * some component has reached its limit. If limits are inactive the method fills
+     * `gout[0]` with a positive value to prevent a false root. Returns 0 on success.
+     */
+    int advanceLimitRootFunc(double t, const double* y, double* gout);
```
speth: I think you can just implement this as `evalRootFunctions`, without the extra layer of indirection. If we end up with multiple root functions, we'd probably need to do some refactoring regardless. Also, same minor note about the comment block (and in this header, using `//!` would definitely be more consistent with the rest of the docs).

include/cantera/zeroD/ReactorNet.h
```diff
@@ -386,6 +401,7 @@ class ReactorNet : public FuncEval
     //! and deliberately not exposed in external interfaces.
     virtual int lastOrder() const;
 
+
```
speth: bonus blank line added here.

src/zeroD/ReactorNet.cpp
```diff
@@ -242,51 +242,74 @@ double ReactorNet::advance(double time, bool applylimit)
...
             }
         }
-        if (!exceeded) {
-            break;
+        if (jmax != npos) {
+            double dt = m_time - m_ybase_time;
+            writelog("    Advance limit triggered for component {:d} (dt = {:9.4g}):"
+                     " |Δ| = {:11.6g}, limit = {:9.4g}, ratio = {:9.4g}\n",
+                     jmax, dt, best_delta, best_limit, max_ratio);
         }
```
speth, line 311: As long as CVODES is working correctly, some of the values printed here are not very interesting. From the example of `reactor1.py`, `best_delta` will always equal `best_limit`, and `max_ratio` should always be 1.0.

src/numerics/CVodesIntegrator.cpp
```diff
@@ -90,6 +90,16 @@ extern "C" {
...
         return f->preconditioner_solve_nothrow(NV_DATA_S(r),NV_DATA_S(z));
     }
+
+    //! Sundials callback. Forwards root evaluations to the active FuncEval
+    static int cvodes_root(sunrealtype t, N_Vector y, sunrealtype *gout, void *user_data)
+    {
+        auto* f = static_cast<FuncEval*>(user_data);
+        if (!f) {
+            return 0;
+        }
```
speth: I don't think it's necessary to check against `f` being null -- we always call `CVodeSetUserData` as part of initialization, and it's not checked in any of the other callback wrappers.

src/numerics/CVodesIntegrator.cpp
```diff
@@ -323,6 +350,9 @@ void CVodesIntegrator::initialize(double t0, FuncEval& func)
     flag = CVodeSetUserData(m_cvode_mem, &func);
     checkError(flag, "initialize", "CVodeSetUserData");
 
+    m_nRootFunctions = static_cast<size_t>(-1);
```
speth: Here and below:
```suggestion
    m_nRootFunctions = npos;
```

src/numerics/CVodesIntegrator.cpp
```diff
@@ -513,18 +546,27 @@ void CVodesIntegrator::integrate(double tout)
                 "Components with largest weighted error estimates:\n{}",
                 flag, m_error_message, f_errs, getErrorInfo(10));
         }
+        if (flag == CV_ROOT_RETURN) {
+            // Stop early at root (e.g., advance limit reached)
+            root_triggered = true;
+            break;
+        }
```
speth, line 558: I think you could slightly simplify the logic here by just setting `tout = m_tInteg` and eliminating the `root_triggered` flag.

src/numerics/CVodesIntegrator.cpp
```diff
@@ -513,18 +546,27 @@ void CVodesIntegrator::integrate(double tout)
...
         nsteps++;
     }
-    int flag = CVodeGetDky(m_cvode_mem, tout, 0, m_y);
+    // Align solution time to the requested tout when no root event occurred; this
+    // avoids small positive overshoots from CV_ONE_STEP, which can otherwise cause
+    // callers to see non-monotonic target times.
+    double t_eval = root_triggered ? m_tInteg : tout;
+    int flag = CVodeGetDky(m_cvode_mem, t_eval, 0, m_y);
```
speth: There's really no expectation for CVODES to stop at the specified `tout`. My take on this would be:
```suggestion
    // Interpolate the solution to either the user-specified output time or
    // the time at which a root event occurred.
    double t_eval = root_triggered ? m_tInteg : tout;
    int flag = CVodeGetDky(m_cvode_mem, t_eval, 0, m_y);
```

src/zeroD/ReactorNet.cpp
```diff
@@ -242,51 +242,74 @@ double ReactorNet::advance(double time, bool applylimit)
...
-        step();
+    // Integrate toward the requested time; integrator will return early if a limit is
+    // reached (CV_ROOT_RETURN)
+    try {
+        m_integ->integrate(time);
+    } catch (...) {
+        m_limit_check_active = false;
+        m_integ->setRootFunctionCount(nRootFunctions());
+        throw;
     }
```
speth, line 267: Can you add a comment explaining why this try/catch/rethrow structure is necessary?

speth (changes requested):
Thanks, @wandadars, this looks really good. I was worried that this would introduce a significant performance penalty due to the iteration required to exactly hit the target advance limit, but I think what actually happens is that this iteration is done entirely on the interpolated solution vector and as such doesn't require any additional evaluations of the ODE RHS function.

*Addendum*: I also agree with @ischoegl's suggestion on improving the testing of this. Probably the biggest gap in the coverage is on the verbose logging side. You can test this in pytest using the `capsys` fixture; there are a couple of instances in the test suite now that use this that you can use as examples.

wandadars:
Thanks for the comments @speth  and @ischoegl . I believe all the comments have been addressed and tests have been added/updated regarding the advance limits.

speth (approved):
Thanks for the updates, @wandadars. This looks good to me.

I added the "no throw" wrapper around `evalRootFunctions` that I commented on but may not have been completely clear about and updated a few comments.

@ischoegl, since I've now modified this PR, will you please review/approve?

ischoegl (approved):
Thank you, @wandadars (and @speth). As the original author of the advance limits capability, I am very happy to see this being made more functional.

PR #1981, opened 2025-09-18
Restore ThermoPhase state after error setting state

speth:

- Add functions for saving/restoring the "partial" thermo state, which excludes the composition variables
- Reset `ThermoPhase` objects to their previous (valid) state after an error setting the state. This is particularly important for the iterative two-property setters like `setState_HP` which repeatedly manipulate the internal state.

One important result of this revised behavior is that it improves the ability to use thermo models where errors in `setState_HP` or `setState_UV` are possible in reactor reactor networks. Previously, a bad estimated step by the integrator would lead to repeated failures in the `eval` function, even after the integrator backs off from the problematic state. Now, it is actually possible for the integrator to proceed after one of these "recoverable" integrator errors. 

Resolves an issue identified in a [post on the Cantera Users' Group](https://groups.google.com/g/cantera-users/c/_yYs1mJECzQ).

**Checklist**


include/cantera/thermo/Phase.h
```diff
@@ -298,6 +298,25 @@ class Phase
     //! (such as "TD", "TP", "SV").
     virtual vector<string> partialStates() const;
 
+    //! Get the size of the vector defining the "partial" state of the phase
+    //! (excluding composition). Used by savePartialState() and restorePartialState()
```
bryanwweber: I think this comment can just be declarative without parentheses:


```suggestion
    //! Get the size of the partial state vector of the phase. The partial state vector excludes phase composition.
//! Used by savePartialState() and restorePartialState()
```

src/thermo/Phase.cpp
```diff
@@ -225,11 +225,42 @@ vector<string> Phase::partialStates() const
     }
 }
 
+void Phase::savePartialState(size_t lenstate, double* state) const
+{
+    if (lenstate < partialStateSize()) {
+        throw ArraySizeError("Phase::savePartialState", lenstate, partialStateSize());
+    }
+
+    // function assumes default definition of nativeState
```
bryanwweber: It's probably worth moving this comment into the docstring.

test/thermo/ThermoPhase_Test.cpp
```diff
@@ -302,4 +302,57 @@ TEST_F(EquilRatio_MixFrac_Test, EquilRatio_MixFrac_StoichMix_Mass)
...
+    }
+
+    shared_ptr<ThermoPhase> thermo;
+    double T1, P1, h1, cp1, s1;
+};
+
+TEST_F(SetStateRecovery, HP)
+{
+    double h2 = h1 + cp1 * (600 - T1); // Water state is invalid at 600 K
+    EXPECT_THROW(thermo->setState_HP(h2, P1), CanteraError);
```
bryanwweber, line 324: Does it make sense to assert the original values after this throw?

bryanwweber (commented):
Thanks Ray! Some small suggestions for improving the docs, otherwise looks good to me.

bryanwweber (approved):

PR #1984, opened 2025-09-20
Deprecate legacy convenience headers 

ischoegl:

The convenience headers `thermo.h`, `kinetics.h` and `transport.h` have been marked as 'superseded by `core.h`' since Cantera 3.0. This PR deprecates them with Cantera 3.2.

**Checklist**

### DIFF
--- a/doc/sphinx/userguide/cxx-tutorial.md
+++ b/doc/sphinx/userguide/cxx-tutorial.md
@@ -65,15 +65,27 @@ calculations. The headers and their functions are:
 
 `thermo.h`
 : Base thermodynamic classes and functions for creating {ct}`ThermoPhase` objects from
-  input files *(Superseded by `core.h` in Cantera 3.0)*.
+  input files.
+  :::{deprecated} 3.2
+  Superseded by `core.h` in Cantera 3.0, deprecated in Cantera 3.2, and to be removed in
+  Cantera 3.3.
+  :::
 
 `kinetics.h`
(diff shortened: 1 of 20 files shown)


speth (changes requested):
If `core.h` is meant to be a true replacement for these other convenience headers, then I think it should also include a few other headers as well:
- `ThermoFactory.h`
- `KineticsFactory.h`
- `TransportFactory.h`
- `Species.h`
- `Reaction.h`

The less users need to think about the factory classes (as opposed to the `newX` functions) the better.

ischoegl:
> If `core.h` is meant to be a true replacement for these other convenience headers, then I think it should also include a few other headers as well:
> 
> * `ThermoFactory.h`
> * `KineticsFactory.h`
> * `TransportFactory.h`
> * `Species.h`
> * `Reaction.h`
> 
> The less users need to think about the factory classes (as opposed to the `newX` functions) the better.

Unfortunately, I disagree with the factory headers - I truly believe that any new code should go through `newSolution`/`newInterface`. All instances where the factories are called directly (which I had to touch in this PR) are legacy code, and I am not in favor of continuing this practice. I can see the point of `Species.h` and `Reaction.h`, as well as modifying the 'replacement options' in the pragma warnings to also point to factories. If a user really insists on side-lining `Solution`, they should not be able to rely on a convenience header.

speth (approved):
OK, with the updated deprecation message, I think this is fine.

PR #1986, opened 2025-09-21
Refactor CLib Cabinet

### LINKED ISSUE #236: Update CLib's Cabinet
**Abstract**

The current implementation of CLib's `Cabinet.h` is used for both legacy and generated versions. In both versions, handles to managed objects are shared, meaning that they are destroyed when the managing object is destroyed. This behavior may go back to before cabinets used shared pointers and was not updated when the code was refactored. A previously accessed managed `shared_ptr` object should survive destruction of the 'managing' object as long as the handle is valid.

**Motivation**

* What problem is it trying to solve? ... make CLib Cabinets more like shared_ptr
* Who is affected by the change? ... developers
* Why is this a good solution? ... avoids confusion / simplifies implementation

**Possible Solutions**

No longer use a carbon-copy of the same `Cabinet.h` file for the generated CLib - the 'parent' tracking will no longer be needed.

**References**

See also Cantera/cantera#1921

ischoegl:

This PR removes a historic approach to handling 'managed' objects that was initially developed when `Cabinet`s held raw pointers, rather than smart pointers.

The refactored approach drops 'managing' of objects, and rather makes a reference available once it is accessed.

In addition, this PR drops several instances of 'generated' suffixes that are no longer needed, and uses the `.j2` suffix for Jinja2 templates rather than `.in`.

Closes Cantera/enhancements#236

**Checklist**


ischoegl:
The current failure appears to be unrelated to changes in this PR. This is ready for a review.

speth (approved):
Thanks, @ischoegl, this looks good to me. I'm glad to see that this change ends up making things simpler overall.

I tested this with #1921 rebased on top and after resolving a few unrelated merge conflicts, this appears to work fine for the behavior required there.

PR #1987, opened 2025-09-23
Clib paper cuts

ischoegl:

Resolve remaining items marked as TODO in the generated CLib.

**Checklist**

### DIFF
--- a/doc/sphinx/python/utilities.rst
+++ b/doc/sphinx/python/utilities.rst
@@ -60,6 +60,7 @@ Global Functions
 ----------------
 
 .. autofunction:: add_directory
+.. autofunction:: add_data_directory
 .. autofunction:: get_data_directories
 .. autofunction:: import_phases
 .. autofunction:: appdelete
--- a/include/cantera/base/global.h
+++ b/include/cantera/base/global.h
@@ -49,8 +49,8 @@ class AnyMap;
  *      will be added to the search path.
  *    - Finally, the location where the data files were installed when %Cantera was
  *      built is added to the search path.
- *    - Additional directories may be added dynamically by calling the addDirectory()
- *      function.
+ *    - Additional directories may be added dynamically by calling the
+ *      addDataDirectory() function.
  *
  * %Cantera input files are written using YAML syntax. For more information on using
  * YAML files in %Cantera, see the [YAML Users' Guide](../userguide/input-tutorial.html)
@@ -68,7 +68,12 @@ class AnyMap;
 string findInputFile(const string& name);
 
 //! @copydoc Application::addDataDirectory
+//! @deprecated Deprecated in %Cantera 3.2 and to be removed thereafter. Renamed to
+//!     addDataDirectory().
 void addDirectory(const string& dir);
+
+//! @copydoc Application::addDataDirectory
+void addDataDirectory(const string& dir);
 
 //! @copydoc Application::getDataDirectories
 string getDataDirectories(const string& sep);
--- a/include/cantera/oneD/Sim1D.h
+++ b/include/cantera/oneD/Sim1D.h
@@ -167,6 +167,19 @@ public:
     AnyMap restore(const string& fname, const string& name);
 
     /**
+     * Retrieve data from a previously saved simulation.
+     *
+     * This method is almost identical to restore() but avoids the return of an AnyMap,
+     * which is not implemented in CLib.
+     *
+     * @param fname  Name of container file (YAML or HDF)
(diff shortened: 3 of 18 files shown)


interfaces/sourcegen/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "sourcegen"
-version = "3.2.0a2"
+version = "3.2.0a4"
```
bryanwweber, line 7: Somehow this got missed in the a2->a3 change too :sob:

include/cantera/base/global.h
```diff
@@ -68,8 +68,13 @@ class AnyMap;
 string findInputFile(const string& name);
 
 //! @copydoc Application::addDataDirectory
+//! @deprecated Deprecated in %Cantera 3.2 and to be removed thereafter. Renamed to
+//!     addDataDirectory().
```
speth, line 72: If we're changing this method name (not totally convinced it's worth the churn), I'll note the Python method is also just `add_directory`.

interfaces/cython/cantera/_utils.pxd
```diff
@@ -72,7 +72,7 @@ cdef extern from "cantera/base/stringUtils.h" namespace "Cantera":
 
 
 cdef extern from "cantera/base/global.h" namespace "Cantera":
-    cdef void CxxAddDirectory "Cantera::addDirectory" (string)
+    cdef void CxxAddDirectory "Cantera::addDataDirectory" (string)
```
speth: To reduce confusion, the intermediate name ought to be `CxxAddDataDirectory`.

include/cantera/oneD/Sim1D.h
```diff
@@ -166,6 +166,19 @@ class Sim1D : public OneDim
...
+     * This method restores a simulation object from YAML or HDF data previously saved
+     * using the save() method.
+     *
+     * @param fname  Name of container file (YAML or HDF)
+     * @param name  Identifier of location within the container file; this node/group
+     *      contains header information and subgroups with domain-specific SolutionArray
+     *      data
+     */
+    void restoreSimple(const string& fname, const string& name);
+
```
speth, line 181: If there's no easy way to get CLib to just ignore the return value from `restore`, I'd rather make this more of an "internal" method so user's don't get caught up trying to spot the subtle difference between this and `restore`. For example, this could be named `_restore` and be be documented as being a variant of `restore` that exists only for the sake of CLib.

test/clib/test_ctonedim.cpp
```diff
@@ -200,4 +200,7 @@ TEST(ctonedim, freeflame)
         ASSERT_GE(T + 1e-3, Tprev);
         Tprev = T;
     }
+
+    ret = sim1D_restore(flame, "gtest-freeflame.yaml", "clib");
+    ASSERT_GE(ret, 0);
```
speth: I assume the return value should just be equal to zero, not positive.

speth (changes requested):

include/cantera/base/global.h
ischoegl: Good catch - renamed that also. It _is_ a paper cut, but this makes it more consistent with `getDataDirectories`.

include/cantera/oneD/Sim1D.h
ischoegl: There is no easy way, unfortunately. Using `_restore` is a better alternative; I had to leave most of the docstring as is, as this is what sourcegen needs.

speth: Right, I forgot that sourcegen needs the docstring.

doc/sphinx/python/utilities.rst
```diff
@@ -59,7 +59,7 @@ EigenSparseDirectJacobian
 Global Functions
 ----------------
 
-.. autofunction:: add_directory
+.. autofunction:: add_data_directory
```
speth, line 63: I think the existing function should also be listed here, so the deprecation notification will be visible in the docs.

ischoegl:
@speth ... let me know if there's anything else. Otherwise, this is ready.

speth (changes requested):
Sorry for the delay. One more minor suggestion.

doc/sphinx/python/utilities.rst
ischoegl: Good catch! Done!

speth (approved):

PR #1989, opened 2025-09-30
Adding another readable unit m^3

itnonaka:
- Added the "m³" line in src/base/Units.cpp so Cantera can not only get m^3 but also read m³ as volume unit.
- Added the "m³" in unit basic conversion in test/general/test_units.cpp so it can also be tested.

When reading a .yaml file, it must have the units that the coefficients are referring to. However, sometimes people may use the same unit but in a different format, for example, m³ instead of m^3. Now, both ways can be read.

**Checklist**

- [x] Style & formatting of contributed code follows [contributing guidelines]

### DIFF
--- a/doc/sphinx/install/pip.md
+++ b/doc/sphinx/install/pip.md
@@ -120,7 +120,7 @@ You should get the following output:
 
       temperature   300 K
          pressure   1.0133e+05 Pa
-          density   0.081894 kg/m^3
+          density   0.081894 kg/m³
  mean mol. weight   2.016 kg/kmol
   phase of matter   gas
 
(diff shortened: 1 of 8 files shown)


ischoegl (changes requested):
Thanks, @itnonaka - this is a nice follow-up to the recently introduced support for the unicode alternative to `m^2`. I have a couple of minor requests.

* Documentation: search for instances of `m^3` in `*.md` files ... there are several instances where `m^3` is used in sample code. Replace 2 or 3 instances of sample code while leaving the rest as `m^3` so users see both versions; however, replace all instances of `m^3` that are part of comments.
* Documentation (cont'd): search for "Supported compound units are:" and add `m³` to the _Volume_ entry, so it's clear that it's an alternative to `m^3` that is supported (you _could_ also add the unicode version of `dyn/cm^2` on the line above
* Python unit tests: add alternative lines in `test_utils.py` (as in the C++ google tests, add new lines with `cm³` without replacing `cm^3`).

ischoegl (approved):
Thanks, @itnonaka.

PR #1995, opened 2025-10-04
Make `Sim1D`/`Domain1D` and `SolutionArray` usage more consistent

### LINKED ISSUE #1865: Radial pressure gradient name, lambda, breaks some SolutionArray functionality in Python.
**Problem description**

When adding "extra" columns to SolutionArray, they can be accessed as attributes of the object. However, presumably because `lambda` is reserved for lambda functions in Python, it breaks this feature.

**Steps to reproduce**

Because the radial pressure gradient is named `lambda` in the AxisymmetricFlow object, converting it to a SolutionArray demonstrates the issue:

```python
>>> import cantera as ct
>>> gas = ct.Solution("gri30.yaml")
>>> flame = ct.CounterflowDiffusionFlame(gas, width=0.02)
>>> sol = flame.to_array()
>>> print(sol.extra)
['grid', 'velocity', 'spread_rate', 'lambda']
>>> print(sol.lambda)
  File "<python-input-1>", line 1
    print(sol.lambda)
              ^^^^^^
SyntaxError: invalid syntax
```

Though one workaround is to use getattr instead:

```python
>>> print(getattr(sol, "lambda"))
[0. 0. 0. 0. 0. 0.]
```

**Behavior**

I would expect to be able to access the radial pressure gradient the same way I would any other `extra` parameter, e.g. `velocity`:

```python
>>> print(sol.velocity))
[0. 0. 0. 0. 0. 0.]
```

I presume this may be why the radial pressure gradient is still called `L` in `FlameBase`, and perhaps there is a way to map `sol.L` to `getattr(sol, "lambda")` without breaking other functionality or user expectations. Also, apologies if this has already been discussed, I wasn't able to find anything about it when I searched the issues and pull requests for "lambda."

**System information**

- Cantera version: 3.1.0
- OS: RHEL 8.10
- Python version: 3.13

ischoegl:

* Make usage of `Sim1D`/`Domain1D` in Python more consistent with `SolutionArray`.
* Address syntax error when accessing the radial pressure gradient `lambda` from Python. The component is internally renamed to `Lambda`, but it can still be accessed as before through an 'alias' mapping.
* Introduce `getValues`/`setValues` for `Domain1D`; new methods are also made available to CLib.
* Introduce more pythonic property names, e.g.: `L->radial_pressure_gradient`, `E->electric_field`, `Uo->oxidizer_velocity` and move them to `Domain1D`.
* Provide access of properties/states where they are defined, and start rolling back convenience access via `Sim1D`.
* Streamlined access for `Sim1D` content via `SolutionArray`

**Note:** While properties are made available for `Domain1D`, convenience getters for `Sim1D` that are not renamed remain unchanged. Future changes to `SolutionArray` and `Domain1D` should provide access to various calculated results via CLib; these changes go well beyond the scope of this PR.

Closes #1865

Fixes `SolutionArray`:
```
>>> import cantera as ct
>>> gas = ct.Solution("h2o2.yaml")
>>> arr = ct.SolutionArray(gas)
>>> arr.restore("test/data/flame_lambda.h5", "solution/flame")
>>> print(arr.extra)
['grid', 'velocity', 'spread_rate', 'Lambda']
>>> arr.radial_pressure_gradient
array([-90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678])
>>> getattr(arr, "lambda")
array([-90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678])
>>> arr.Lambda
array([-90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678])
```

Makes `Domain1D` more consistent with `SolutionArray`:

```
>>> import cantera as ct
>>> gas = ct.Solution("h2o2.yaml")
>>> fl = ct.CounterflowDiffusionFlame(gas)
>>> fl.restore("test/data/flame_lambda.yaml", "solution/flame")
>>> fl.flame.radial_pressure_gradient
array([-90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678, -90257.82873678, -90257.82873678, -90257.82873678,
       -90257.82873678])
>>> fl.flame.H2  # access species as attributes (similar to SolutionArray)
array([9.99999978e-01, 9.99997462e-01, 9.99716817e-01, 9.91765815e-01,
       8.54984875e-01, 5.96459227e-01, 3.17798181e-01, 7.52412972e-02,
       3.04125904e-02, 1.06545147e-02, 3.05519330e-03, 4.85621121e-04,
       1.30666385e-04, 3.26773349e-05, 2.21745526e-05, 1.73235417e-05,
       1.62255848e-05, 1.51638920e-05, 1.31289484e-05, 1.11919925e-05,
       9.39325661e-06, 6.38458417e-06, 4.22996912e-06, 1.88624211e-06,
       4.34155891e-07, 3.24375927e-08, 2.47941064e-09, 7.45717397e-11,
       7.17868321e-13])
```

**Checklist**


include/cantera/base/SolutionArray.h
```diff
@@ -130,7 +130,7 @@ class SolutionArray
      *  Check whether SolutionArray contains a component.
      *  A component is a property defining state or auxiliary variable.
      */
-    bool hasComponent(const string& name) const;
+    bool hasComponent(const string& name, bool checkAlias=true) const;
```
speth, line 135: The docstring should be updated to explain what the `checkAlias` argument means.

include/cantera/base/SolutionArray.h
```diff
@@ -404,6 +404,9 @@ class SolutionArray
     vector<int> m_active; //!< Vector of locations referencing active entries
 };
 
+//! Return mapping of component alias names to standardized component names.
+const std::map<std::string, std::string>& _componentAliasMap();
```
speth: `map` and `string` are already covered by `using` declarations in `ct_defs.h`.

include/cantera/oneD/Domain1D.h
```diff
@@ -196,8 +196,19 @@ class Domain1D
...
     }
 
-    //! index of component with name `name`.
-    virtual size_t componentIndex(const string& name) const;
+    /**
+     * Index of component with name `name`.
+     * @param name  name of component
+     * @param checkAlias  if `True` (default), check alias mapping
+     */
+    virtual size_t componentIndex(const string& name, bool checkAlias=true) const;
```
speth, line 216:
```suggestion
    /**
     * Index of component with name `name`.
     * @param name  name of component
     * @param checkAlias  if `true` (default), check alias mapping
     */
    virtual size_t componentIndex(const string& name, bool checkAlias=true) const;
```
Also, should add a `@since` annotation for the new argument. Ditto for `hasComponent` below.

ischoegl:
@speth ... This should be ready for a review. What started with a simple bug fix ended up being a broader maintenance job. The work has the side effect of streamlining access for CLib, which should benefit future updates of the MATLAB oneD portion.

src/base/SolutionArray.cpp
```diff
@@ -35,9 +35,29 @@ const std::map<std::string, std::string> aliasMap = {
     {"Q", "vapor-fraction"},
 };
 
+const std::map<std::string, std::string> reverseAliasMap = {
+    // reserved names used by states
+    {"temperature", "T"},
+    {"pressure", "P"},
+    {"density", "D"},
+    // reserved names used for 1-D objects
+    {"spread-rate", "spread_rate"},
```
ischoegl: As an aside: `spread_rate` uses snake_case, while other components don't. Since we're already renaming components, this could also be updated here.

include/cantera/oneD/Flow1D.h
```diff
@@ -548,15 +557,15 @@ class Flow1D : public Domain1D
...
      *
      * @f[
      *    \frac{d\Lambda}{dz} = 0
      * @f]
      *
-     * The lambda equation serves as an eigenvalue that allows the momentum
+     * The Lambda equation serves as an eigenvalue that allows the momentum
      * equation and continuity equations to be simultaneously satisfied in
-     * axisymmetric flows. The lambda equation propagates information from
+     * axisymmetric flows. The Lambda equation propagates information from
```
speth: While we're cleaning this up:
```suggestion
     * The radial pressure gradient @f$ \Lambda @f$ serves as an eigenvalue that allows
     * the momentum and continuity equations to be simultaneously satisfied in
     * axisymmetric flows. This equation propagates information from
```

src/oneD/Flow1D.cpp
```diff
@@ -16,6 +16,15 @@ using namespace std;
 namespace Cantera
 {
 
+const std::map<std::string, size_t> componentMap = {
+    {"velocity", c_offset_U}, // axial velocity [m/s]
```
speth: To make this file-local, it should be defined in an anonymous namespace (which can be nested inside `Cantera`). Also, the use of `std::` here is unnecessary given the `using` statements in `ct_defs.h`.

src/oneD/Flow1D.cpp
```diff
@@ -827,37 +842,48 @@ string Flow1D::componentName(size_t n) const
...
+    if (componentMap.count(name)) {
+        return componentMap.at(name);
+    }
+    for (size_t n=c_offset_Y; n<m_nsp+c_offset_Y; n++) {
+        if (componentName(n)==name) {
+            return n;
+        }
+    }
+    if (checkAlias) {
+        auto aliasMap = _componentAliasMap();
```
speth: To avoid making a copy:
```suggestion
        const auto& aliasMap = _componentAliasMap();
```

src/oneD/Flow1D.cpp
```diff
@@ -827,37 +842,48 @@ string Flow1D::componentName(size_t n) const
...
+                       "No component named '{}'", name);
+}
+
+bool Flow1D::hasComponent(const string& name, bool checkAlias) const
+{
+    if (componentMap.count(name)) {
+        return true;
+    }
+    for (size_t n=c_offset_Y; n<m_nsp+c_offset_Y; n++) {
+        if (componentName(n)==name) {
```
speth: Very minor, but I think I'd prefer
```suggestion
    for (size_t n = c_offset_Y; n < m_nsp+c_offset_Y; n++) {
        if (componentName(n) == name) {
```
here and elsewhere. We're reasonably consistent about having space around the binary comparison operators and the assignment operator in `for` loops.

include/cantera/oneD/Domain1D.h
```diff
@@ -368,17 +436,25 @@ class Domain1D
...
+     * @deprecated To be removed after %Cantera 3.2.
+     *      Replaceable by version that does not require solution vector.
      */
-    virtual void fromArray(SolutionArray& arr, double* soln) {
-        throw NotImplementedError("Domain1D::fromArray", "Needs to be overloaded.");
+    void fromArray(SolutionArray& arr, double* soln) {
+        warn_deprecated("Domain1D::fromArray",
+            "To be removed after Cantera 3.2. Replaceable by version that does not "
+            "require solution vector.");
+        shared_ptr<SolutionArray> arr_ptr(&arr, [](SolutionArray*){});
```
speth, line 553: I know this is immediately deprecated, but this unusual constructor warrants a comment -- this is creating a `shared_ptr` that skips deletion of the held pointer, since it doesn't really own it, right?

interfaces/sourcegen/src/sourcegen/headers/ctdomain_auto.yaml
```diff
@@ -19,6 +19,13 @@ recipes:
 - name: nPoints
 - name: componentName
 - name: componentIndex
+  wraps: componentIndex(const char*)
+- name: getValues
```
speth: I don't quite know the code generation pathway at work here, but the resulting method does not work correctly:
```c++
    int32_t domain_getValues(int32_t handle, const char* component, int32_t valuesLen, double* values)
    {
        // method: virtual void Domain1D::getValues(const string&, vector<double>&)
        try {
            vector<double> values_(values, values + valuesLen);
            Domain1DCabinet::at(handle)->getValues(component, values_);
            return 0;
        } catch (...) {
        }
    }
```
Here, a temporary vector is created from the input `values`, overwritten by the requested data, but then discarded without copying the data back to the user-provided buffer.

interfaces/sourcegen/src/sourcegen/headers/ctdomain_auto.yaml
```diff
@@ -19,6 +19,13 @@ recipes:
 - name: nPoints
 - name: componentName
 - name: componentIndex
+  wraps: componentIndex(const char*)
```
speth: Should this be:
```suggestion
  wraps: componentIndex(const string&)
```
given the signature of the actual C++ method?

include/cantera/oneD/Domain1D.h
```diff
@@ -348,16 +359,73 @@ class Domain1D
...
+    /**
+     * Retrieve component values.
+     * @param component  name of component
+     * @param[out] values  vector of values
+     *
+     * @since New in %Cantera 3.2.
+     */
+    virtual void getValues(const string& component, vector<double>& values) const {
+        throw NotImplementedError("Domain1D::setProfile", "Needs to be overloaded.");
+    }
```
speth, line 435:
```suggestion
    virtual void getValues(const string& component, vector<double>& values) const {
        throw NotImplementedError("Domain1D::getValues", "Needs to be overloaded.");
    }
```
A similar fix is needed for `setValues`.

include/cantera/oneD/Domain1D.h
```diff
@@ -348,16 +359,73 @@ class Domain1D
         return x[index(n,j)];
     }
 
+    /**
+     * Retrieve component values.
+     * @param component  name of component
+     * @param[out] values  vector of values
```
speth: I think the docstring should specify the length of `values` and that the retrieved values are at the points specified by `grid()`. Similarly for `setValues`.

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -935,7 +1087,13 @@ cdef class Sim1D:
             component name or index
 
         >>> T = s.profile(flow, 'T')
+
+        .. deprecated:: 3.2
+
+            To be removed after Cantera 3.2. Replaceable by `Domain1D.get_values`.
         """
+        warnings.warn("To be removed after Cantera 3.2. Replaceable by 'get_values'.",
+                      DeprecationWarning)
```
speth:
```suggestion
        warnings.warn("To be removed after Cantera 3.2. Replaceable by 'Domain1D.get_values'.",
                      DeprecationWarning)
```

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -456,12 +531,89 @@ cdef class FlowBase(Domain1D):
...
+        .. versionadded:: 3.2
+        """
+        return self.get_values("Lambda")
+
+    @property
+    def electric_field(self):
+        """
+        Array containing the electric field strength at each point.
+        Note: This value is named ``eField`` in the C++ code and is only defined if
+        the transport model is ``ionized_gas``.
```
speth:
```suggestion
        the transport model is ``ionized-gas``.
```

interfaces/cython/cantera/_onedim.pyx
```diff
@@ -456,12 +531,89 @@ cdef class FlowBase(Domain1D):
...
+        component_name = name
+        if (not self._has_component(name) and
+            self._has_component(name.replace("_", "-"))):
+            component_name = name.replace("_", "-")
+
+        if self._has_component(component_name):
+            return self.get_values(component_name)
+
+        raise AttributeError(
+            f"{self.__class__.__name__!r} object has no attribute {name!r}")
```
speth, line 595: I'm concerned about implementing this access with `__getattr__`. It effectively introduces a number of properties that aren't documented or discoverable from interactive sessions. And for the species mass fractions, it introduces new challenges for species names that aren't valid Python identifiers, which includes things as simple as `CH2(S)` in `gri30.yaml`.

src/base/SolutionArray.cpp
```diff
@@ -35,9 +35,29 @@ const std::map<std::string, std::string> aliasMap = {
     {"Q", "vapor-fraction"},
 };
 
+const std::map<std::string, std::string> reverseAliasMap = {
+    // reserved names used by states
```
speth: This should be defined in an unnamed namespace (and `aliasMap` above, which got missed before).

src/base/SolutionArray.cpp
```diff
@@ -35,9 +35,29 @@ const std::map<std::string, std::string> aliasMap = {
...
+    {"E", "eField"},
+    {"electric-field", "eField"},
+    {"oxidizer-velocity", "Uo"},
+};
+
 namespace Cantera
 {
 
+const std::map<std::string, std::string>& _componentAliasMap()
+{
```
speth: The `std::`s here are unnecessary.

src/base/Storage.cpp
```diff
@@ -59,7 +60,8 @@ Storage::Storage(string fname, bool write) : m_write(write)
     if (m_write) {
         m_file = make_unique<h5::File>(fname, h5::File::OpenOrCreate);
     } else {
-        m_file = make_unique<h5::File>(fname, h5::File::ReadOnly);
+        auto fullName = findInputFile(fname);
+        m_file = make_unique<h5::File>(fullName, h5::File::ReadOnly);
```
speth, line 64: This seems unrelated to the rest of the PR. When would you want to read simulation results from a directory on the data path?

test/python/test_onedim.py
```diff
@@ -1263,6 +1288,45 @@ def test_bad_boundary_conditions(self):
...
+
+    def test_restore_lambda_yaml(self):
+        self.run_restore_diffusionflame("flame_lambda.yaml")
+        self.run_restore_solutionarray("flame_lambda.yaml")
+
+    @pytest.mark.skipif("native" not in ct.hdf_support(),
+                        reason="Cantera compiled without HDF support")
+    def test_restore_lambda_hdf(self):
+        self.run_restore_diffusionflame("flame_lambda.h5")
+        self.run_restore_solutionarray("flame_lambda.h5")
```
speth, line 1330: Rather than modifying this to consider the input file search path, you can use the `test_data_path` fixture to access files in `test/data`.

Also, as this is one of the larger files in `test/data`, I'd suggest generating it from a simulation without grid refinement enabled to reduce the size a bit.

speth (changes requested):
Thanks for the work on this, @ischoegl. It certainly expanded in scope well beyond resolving the issue with accessing the `lambda` component. Most of the changes here make sense to me, with the exception of "rolling back convenience access via `Sim1D`".

- I'm on the fence about the use of `vector<double>&` as the exchange medium of `Domain1D::get/setValues` and related methods. While it avoids the unsafe raw pointer usage, it  is somewhat inflexible, as seen by the need to create temporary `vector`s where these methods are used both in the Python interface and `CLib`. I guess the longer-term solution here is to use `std::span`, as described in Cantera/enhancements#237.
- Shifting the Python `set_profile` and related methods out of `Sim1D` to the underlying `Domain1D` seems reasonable, and ultimately leads to some good simplifications of the code. However, I'm not keen on deprecating the methods for accessing the grid,   temperature, velocity, etc. arrays for the `flame` domain via the parent `Sim1D` object. I think this just makes some of the most common data to be accessed more cumbersome to work with. Also, it seems strange to leave some data accessible via the `Sim1D` object (such as mass fractions) while moving others to the flow domain. I do think the new, more descriptive names for the solution components are a beneficial change.

include/cantera/oneD/Domain1D.h
ischoegl: Correct. This one saved me major headaches ...

interfaces/cython/cantera/_onedim.pyx
ischoegl: I see the limitations (similar to the `lambda` issue this PR resolves), but my motivation here was to model things after the `__setattr__` that has been in use for `SolutionArray` for a long time (about 10 years per gitlens). My objective was to make things as similar as possible, so it's either fine for both or neither?

src/base/Storage.cpp
ischoegl: This will be useful if you keep your YAML input file and simulation data in a folder separate from the code used to analyze your data. In this case, you'd set `add_data_directory` and have things accessible; if I split hair, this avoids having serialized YAML accessible while HDF5 isn't?

test/python/test_onedim.py
ischoegl: See above. I was not too concerned about the 29 grid points (resulting 40 KB), although it is correct that it's the third-largest file. Will replace ... .

interfaces/sourcegen/src/sourcegen/headers/ctdomain_auto.yaml
ischoegl: Good catch! Apparently, this isn't covered by a unit test. This needs to be 
```C++
int32_t domain_getValues(int32_t handle, const char* component, int32_t valuesLen, double* values)
{
    // method: virtual void Domain1D::getValues(const string&, vector<double>&)
    try {
        vector<double> values_(valuesLen);
        Domain1DCabinet::at(handle)->getValues(component, values_);
        std::copy(values_.begin(), values_.end(), values);
        return 0;
    } catch (...) {
    }
}
```
where `std::span` would indeed be nice. But the introduction goes well beyond the scope of this PR (and would put Cantera 3.2 off by a fair amount).

_PS: turns out that this is something that is not yet covered by the test logic._

interfaces/cython/cantera/_onedim.pyx
speth: Having played around with this a little more, I think the behavior is more clear -- this is effectively only used for accessing fields by alias values rather than the canonical names which are implemented below this as properties. I think it would be helpful to document this in a comment on the `__getattr__` method.

src/base/Storage.cpp
speth: I'll grant the point on consistency. Though if you're saving simulation data to a folder, you need to specify the particular output path -- it's not as though Cantera can/should direct output to an arbitrary directory on the `CANTERA_DATA` path.

src/base/SolutionArray.cpp
ischoegl: Renamed to `spreadRate`.

interfaces/sourcegen/src/sourcegen/headers/ctdomain_auto.yaml
ischoegl: I will open a separate issue/PR to resolve this issue. See #2000

ischoegl:
Thank you for the review, @speth!

> * I'm on the fence about the use of `vector<double>&` as the exchange medium of `Domain1D::get/setValues` and related methods. While it avoids the unsafe raw pointer usage, it  is somewhat inflexible, as seen by the need to create temporary `vector`s where these methods are used both in the Python interface and `CLib`. I guess the longer-term solution here is to use `std::span`, as described in [Cantera/enhancements#237](https://github.com/Cantera/enhancements/issues/237).

I don't think this is critical at the moment - these methods are only accessed when setting things up prior to a simulation, although the utility could expand in the future.

> * Shifting the Python `set_profile` and related methods out of `Sim1D` to the underlying `Domain1D` seems reasonable, and ultimately leads to some good simplifications of the code. However, I'm not keen on deprecating the methods for accessing the grid,   temperature, velocity, etc. arrays for the `flame` domain via the parent `Sim1D` object. I think this just makes some of the most common data to be accessed more cumbersome to work with. Also, it seems strange to leave some data accessible via the `Sim1D` object (such as mass fractions) while moving others to the flow domain. I do think the new, more descriptive names for the solution components are a beneficial change.

I removed the `PendingDeprecationWarnings`. It would be relatively straightforward to attach the dynamic access to mass fractions etc., but this would add further bulk to the PR.

At the same time, I added the following to the PR after the review (i.e., in the last three commits):
* Renamed `spread_rate` to `spreadRate` as the C++ component name (the snake_case was an outlier)
* Moved the single-point access functions from `Sim1D` to `Domain1D`, which makes things consistent from the CLib perspective. 

For the sourcegen issue, I opened a new issue #2000 (with high priority).

There is more maintenance needed, albeit with low urgency - for the time being, I just want to provide more convenient access points for CLib and the associated MATLAB work.

include/cantera/oneD/Domain1D.h
```diff
@@ -151,7 +151,10 @@ class Domain1D
 
     //! Check that the specified component index is in range.
     //! Throws an exception if n is greater than nComponents()-1
+    //! @deprecated To be removed after %Cantera 3.2. Only used by legacy CLib.
     void checkComponentIndex(size_t n) const {
+        warn_deprecated("Domain1D::checkPointIndex",
+            "To be removed after Cantera 3.2. Only used by legacy CLib.");
```
speth:
```suggestion
        warn_deprecated("Domain1D::checkComponentIndex",
            "To be removed after Cantera 3.2. Only used by legacy CLib.");
```

include/cantera/oneD/Domain1D.h
```diff
@@ -196,8 +208,23 @@ class Domain1D
...
-    //! index of component with name `name`.
-    virtual size_t componentIndex(const string& name) const;
+    /**
+     * Index of component with name `name`.
+     * @param name  name of component
+     * @param checkAlias  if `true` (default), check alias mapping
+     *
+     * @since New in %Cantera 3.2.
+     */
+    virtual size_t componentIndex(const string& name, bool checkAlias=true) const;
```
speth: The method isn't new, just the `checkAlias` argument.

include/cantera/oneD/Domain1D.h
```diff
@@ -343,9 +370,115 @@ class Domain1D
...
+     * @param component  Name of the component.
+     * @param localPoint  Grid point within a Flow1D domain, beginning with 0 for
+     *     the leftmost grid point. Unused for Boundary1D domains; defaults to zero.
+     *
+     * @since New in %Cantera 3.2.
+     */
+    virtual double value(const string& component, size_t localPoint=0) const {
+        throw NotImplementedError("Domain1D::value",
+            "Not implemented for domain type '{}'.", domainType());
+    }
```
speth, line 398: Given the addition of the array-valued getters and setters, are these single-value methods even necessary or useful? I can't see anywhere they're used in the existing code. I think dropping these would be a nice simplification of the interface in favor to the more user-friendly methods that return full arrays.

samples/cxx/flamespeed/flamespeed.cpp
```diff
@@ -167,13 +164,10 @@ int flamespeed(double phi, bool refine_grid, int loglevel)
...
-                                        flow->componentIndex("CO"),n));
-            CO2vec.push_back(flame.value(flowdomain,
-                                         flow->componentIndex("CO2"),n));
-            Uvec.push_back(flame.value(flowdomain,
-                                       flow->componentIndex("velocity"),n));
+            Tvec.push_back(flow->value("T", n));
+            COvec.push_back(flow->value("CO", n));
+            CO2vec.push_back(flow->value("CO2", n));
+            Uvec.push_back(flow->value("velocity", n));
             zvec.push_back(flow->z(n));
```
speth: This would be a good opportunity to use the new array-based getters.

samples/cxx/flamespeed/flamespeed.cpp
```diff
@@ -139,24 +139,21 @@ int flamespeed(double phi, bool refine_grid, int loglevel)
         flow->solveEnergyEqn();
 
         flame.solve(loglevel,refine_grid);
-        double flameSpeed_mix = flame.value(flowdomain,
-                                            flow->componentIndex("velocity"),0);
+        double flameSpeed_mix = flow->value("velocity", 0);
```
speth: Ok, so there isn't zero use of the single-value getter. But this could be easily replaced.

speth (changes requested):
To confirm, based on the creation of #2000, the intention is to leave those getters broken for the moment, and address that in a separate PR?

include/cantera/oneD/Domain1D.h
ischoegl: There are currently two use cases:
1. get/set values for boundary domains (example: `ReactingSurface1D`) - these are by default scalars
2. ~Beyond, total removal will cause extensive breakages in Python `FlameBase.set_gas_state` - this should probably be moved into C++, but I don't want to add this to this PR.~

~If I get to it before the release of 3.2, I can work on a C++ `Domain1D::setGasState` in a follow-up PR so we can remove the `localPoint` parameter.~ _Update: I implemented the method, now called `Domain1D::updateState` (in analogy to what is already implemented in `SolutionArray`). Based on this, I removed the `localPoint` from the interface._

ischoegl:
> To confirm, based on the creation of #2000, the intention is to leave those getters broken for the moment, and address that in a separate PR?

I was able to circumvent the bug documented in #2000 by introducing a parallel
```C++
vector<double> Domain1D::values(const string& component) const;
```
which generates correct code. This also has the nice benefit of making C++ access much more convenient - see `flame speed.cpp`. I retained the original version of `Domain1D::getValues` as it avoids the allocation of a new vector.

The #2000 issue remains, but is no longer at high priority as I am not aware of this route being used.

ischoegl:
PS: It appears that GitHub is currently not syncing this PR with updates on my branch. (An attempt to rebase did not fix this.)

ischoegl:
@speth ... I believe I addressed all concerns.

speth (approved):

PR #1997, opened 2025-10-05
[MATLAB] Migrate to generated Clib

### LINKED ISSUE #1722: Broken flame examples in experimental MATLAB toolbox

ssun30:

This PR updates the experimental MATLAB interface to use the **auto-generated Cantera Clib** (produced via `sourcegen`), replacing the previously manually written Clib.  

Specifically, this PR:
- Updates the interface generation workflow to consume the **auto-generated Clib headers** instead of the manually curated set.
- Adjusts MATLAB definition generation to reflect updated function and class names in the new Clib.
- Removes redundant edge-case handling in `ctEditLibraryDefinitions` that was only required for the manually written Clib.
- Re-enables several MATLAB unit tests that were previously excluded due to missing API functions in the old Clib.

Closes #1722

**Checklist**


ssun30:
Hi @ischoegl , I've started migrating the MATLAB interface to the generated Clib. It won't work with the generated Clib as-is due to the following two points:

- Only `ct.h` includes `<stdint.h>`, but several other generated headers in `cantera_clib` also declare functions using `int32_t`.
When the MATLAB interface generator (clibgen.generateLibraryDefinition) parses these headers individually, compilation fails with:

```
/Users/ssun30/cantera/interfaces/clib/include/cantera_clib/ctconnector.h(139): 
error: identifier "int32_t" is undefined'
```
On my local machine, I got around this issue by adding `<stdint.h>` to other header files.

- Several function docstrings in ct.h contain the substring %Cantera, for example, in `ct_version`:

```
/**
 * Returns the %Cantera version.
 */
```


MATLAB’s clibgen treats % as a comment character in MATLAB syntax. During parsing, the line is truncated at %, producing an unterminated string in the generated definectMatlab.m file and causing:

`Error: String is not terminated properly.`

Affected functions:
`ct_version`
`ct_gitCommit`
`ct_getDataDirectories`
`ct_make_warnings_fatal`
`ct_getCanteraError`

On my local machine, I got around this issue by removing the `%` symbol from `%Cantera`.

ischoegl:
@ssun30 ... I pushed some fixes. I can compile things without needing to edit on my machine, and the toolbox loads, but the unit tests fail. Please ensure that you reset your local branch to this one to preserve my edits.

ssun30:
> @ssun30 ... I pushed some fixes. I can compile things without needing to edit on my machine, and the toolbox loads, but the unit tests fail. Please ensure that you reset your local branch to this one to preserve my edits.

Hi @ischoegl , I've fixed most of the issues with the unit tests. But five tests remain and I'll list each of them:

sample `ignite.m`
```
Unable to resolve the name 'clib.ctMatlab.kin_getSourceTerms'.

Error in [ctArray](matlab:matlab.lang.internal.introspective.errorDocCallback('ctArray', '/Users/ssun30/cantera/interfaces/matlab_experimental/Utility/ctArray.m', 6)) ([line 6](matlab: opentoline('/Users/ssun30/cantera/interfaces/matlab_experimental/Utility/ctArray.m',6,0)))
    iok = clib.ctMatlab.(funcName)(varargin{:}, buf);
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in [Kinetics/get.massProdRate](matlab:matlab.lang.internal.introspective.errorDocCallback('Kinetics/get.massProdRate', '/Users/ssun30/cantera/interfaces/matlab_experimental/Base/Kinetics.m', 458)) ([line 458](matlab: opentoline('/Users/ssun30/cantera/interfaces/matlab_experimental/Base/Kinetics.m',458,0)))
            ydot = ctArray('kin_getSourceTerms', nsp, kin.kinID);
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in [reactor_ode](matlab:matlab.lang.internal.introspective.errorDocCallback('reactor_ode', '/Users/ssun30/cantera/samples/matlab_experimental/reactor_ode.m', 44)) ([line 44](matlab: opentoline('/Users/ssun30/cantera/samples/matlab_experimental/reactor_ode.m',44,0)))
        ydt = total_mass * gas.massProdRate;
```
I didn't find the new equivalent function to `kin_getSourceTerms` in the generated clib.

ssun30:
The next sample is `catacomb.m`
```
Error using [clib.ctMatlab.domain_index](matlab:matlab.lang.internal.introspective.errorDocCallback('clib.ctMatlab.domain_index'))
No method 'clib.ctMatlab.domain_index' with matching signature found.

Error in [ctFunc](matlab:matlab.lang.internal.introspective.errorDocCallback('ctFunc', '/Users/ssun30/cantera/interfaces/matlab_experimental/Utility/ctFunc.m', 5)) ([line 5](matlab: opentoline('/Users/ssun30/cantera/interfaces/matlab_experimental/Utility/ctFunc.m',5,0)))
    output = clib.ctMatlab.(funcName)(varargin{:});
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in [Domain1D/get.domainIndex](matlab:matlab.lang.internal.introspective.errorDocCallback('Domain1D/get.domainIndex', '/Users/ssun30/cantera/interfaces/matlab_experimental/OneDim/Domain1D.m', 200)) ([line 200](matlab: opentoline('/Users/ssun30/cantera/interfaces/matlab_experimental/OneDim/Domain1D.m',200,0)))
            i = ctFunc('domain_index', d.domainID) + 1;
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in [Boundary1D/massFraction](matlab:matlab.lang.internal.introspective.errorDocCallback('Boundary1D/massFraction', '/Users/ssun30/cantera/interfaces/matlab_experimental/OneDim/Boundary1D.m', 81)) ([line 81](matlab: opentoline('/Users/ssun30/cantera/interfaces/matlab_experimental/OneDim/Boundary1D.m',81,0)))
            if d.domainIndex == 0
               ^^^^^^^^^^^^^
Error in [catcomb](matlab:matlab.lang.internal.introspective.errorDocCallback('catcomb', '/Users/ssun30/cantera/samples/matlab_experimental/catcomb.m', 144)) ([line 144](matlab: opentoline('/Users/ssun30/cantera/samples/matlab_experimental/catcomb.m',144,0)))
    y = inlt.massFraction(k);
```
The function signature for `domain_index` has changed:
```
    /**
     *  Returns the index of the solution vector, which corresponds to component n at grid point j.
     *
     *  Wraps C++ method: `size_t Domain1D::index(size_t, size_t)`
     *
     *  @param handle       Handle to queried Domain1D object.
     *  @param n            component index
     *  @param j            grid point index
     */
    int32_t domain_index(int32_t handle, int32_t n, int32_t j);
```

ssun30:
sample `surf_reactor.m`
```
Error using [ctFunc](matlab:matlab.lang.internal.introspective.errorDocCallback('ctFunc', '/Users/ssun30/cantera/interfaces/matlab_experimental/Utility/ctFunc.m', 10)) ([line 10](matlab: opentoline('/Users/ssun30/cantera/interfaces/matlab_experimental/Utility/ctFunc.m',10,0)))

CanteraError thrown by Cabinet::as:
Item is not of the correct type.

Error in [ReactorSurface/set.area](matlab:matlab.lang.internal.introspective.errorDocCallback('ReactorSurface/set.area', '/Users/ssun30/cantera/interfaces/matlab_experimental/Reactor/ReactorSurface.m', 56)) ([line 56](matlab: opentoline('/Users/ssun30/cantera/interfaces/matlab_experimental/Reactor/ReactorSurface.m',56,0)))
            ctFunc('reactor_setArea', s.id, a);
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in [surf_reactor](matlab:matlab.lang.internal.introspective.errorDocCallback('surf_reactor', '/Users/ssun30/cantera/samples/matlab_experimental/surf_reactor.m', 48)) ([line 48](matlab: opentoline('/Users/ssun30/cantera/samples/matlab_experimental/surf_reactor.m',48,0)))
rsurf.area = A;
^^^^^^^^^^
```
I didn't find the equivalent methods for `reactorsurface_area` and `reactorsurface_setArea` in the new clib.

ssun30:
The final errors are under `ctTestThermo.m`, in particular `testElementalMassFraction` and `testNAtoms`:
```
    --> The value does not contain the substring.
    
    Actual char:
        <no Cantera error
    Expected Substring:
        No such element
```
In the legacy clib, the `thermo_elementIndex` and `thermo_speciesIndex` methods have hard-coded error handling:
```
    size_t thermo_elementIndex(int n, const char* nm)
    {
        try {
             size_t k = ThermoCabinet::at(n)->elementIndex(nm);
             if (k == npos) {
                throw CanteraError("thermo_elementIndex",
                                   "No such element {}.", nm);
             }
             return k;
        } catch (...) {
            return handleAllExceptions(npos, npos);
        }
    }
```
Now the corresponding methods in the generated clib lack those checks, and thus no error message was returned.

ischoegl:
@ssun30 ... thank you - I'll look into the issue you pointed out shortly. We just merged #1995, which changes some of the 1D access routes (good news: things should become somewhat easier). Could you rebase? I'll try to push some fixes to the issues you pointed out thereafter.

_PS: I outlined some of my thoughts on the 'onedim' portion in Cantera/enhancements#239. Only a small fraction is applicable to this PR._

ischoegl:
> I didn't find the new equivalent function to `kin_getSourceTerms` in the generated clib.

This was hard-coded in the legacy CLib, and won't be replicated as such in the generated CLib. You have two options:
1. Implement what the old function did in MATLAB
2. Add it to the C++ code base

ischoegl:
> The function signature for `domain_index` has changed:

Point-wise access should be replaced by `domain_value` for boundaries and `domain_values` for flow domains.

ischoegl:
> The final errors are under `ctTestThermo.m`, in particular `testElementalMassFraction` and `testNAtoms` 
[...]

The underlying C++ methods return `npos`, and Python implements custom exception handling. The way the generated CLib works is that it will simply pass through C++ behavior, with no additional checks added.

The quick fix is to follow the Python approach, although an argument could be made to move this to C++ (which is significantly more difficult, as this changes long-standing behavior with potentially breaking changes). The best approach to implementing a change in C++ would be to create an enhancement request, as it exceeds the scope of the MATLAB toolbox. _Edit: see #2004._

PS: #2010

ischoegl:
> I didn't find the equivalent methods for `reactorsurface_area` and `reactorsurface_setArea` in the new clib.

Yes - while there was a setter/getter, it was shadowed by `FlowReactor`. I implemented a version that handles both in C++ - please use `reactor_area` and `reactor_setArea` and add it to the base class. The C++ fix is added to this PR.

ischoegl:
@ssun30 ... I will have a look at the 1D components as there have been significant changes. I will try to make this as close to Cantera/enhancements#239 and the corresponding Python framework as possible so this is futureproof.

ischoegl:
@ssun30 ... I added a new `flamespeed.m` example that is an almost identical replica of the C++ example. It runs and gives the same output as C++. I.e., this PR likely fixes #1722 (added this to the top description).

The `catcomb.m` sample runs ~mostly~ through at this point ~but stalls at the plotting step; the rest should be straightforward debugging~. _PS: pushed the remaining (trivial) fix._

Overall, there were some issues with MATLAB functions calling incorrect CLib signatures. I created convenience access points as necessary. There are two observations:
* The MATLAB-built interface changes the signatures, which is extremely confusing, as the names still correspond to CLib
* MATLAB does not display C++ output (which is a bug)

ischoegl:
@ssun30 ... I rebased and resolved the merge conflicts. Please note that I also exposed `ctMakeDeprecationWarningsFatal`, so you can see which of the transitional methods to avoid.

_PS: discrepancies in constructor signatures for `ReactorSurface` and `Reactor` made it necessary to create an abstract base class `ReactorBase`. Also, note that I opted for a `clone=true` default behavior immediately, which is what all of Cantera assumes as a default after the 3.2 release._

ischoegl:
@ssun30 ... I just went over #2010 again and based on a comment by @speth came up with a solution that won't break index lookups here (while still raising exceptions that are currently missing). The only thing that will change is that `kin_speciesIndex` will become `kin_kineticsSpeciesIndex`, which addresses an inconsistency that's worth fixing. You should see that some of the error messages will become more consistent/specific once #2010 is merged, but that's easily addressed.

Myself, I don't have anything else at the moment. Feel free to reach out if you need anything. As I have contributed to this PR myself, we'll need to get @speth or @bryanwweber to give final approval.

ischoegl:
> > I didn't find the new equivalent function to `kin_getSourceTerms` in the generated clib.
> 
> This was hard-coded in the legacy CLib, and won't be replicated as such in the generated CLib. You have two options:
> 
> 1. Implement what the old function did in MATLAB
> 2. Add it to the C++ code base

I recently updated a section addressing this here: https://cantera.org/dev/develop/clib-extensions.html#extending-the-clib-api

ssun30:
> > I didn't find the new equivalent function to `kin_getSourceTerms` in the generated clib.
> 
> This was hard-coded in the legacy CLib, and won't be replicated as such in the generated CLib. You have two options:
> 
> 1. Implement what the old function did in MATLAB
> 2. Add it to the C++ code base

This function is used by only one example (`ignite.m`). I'm leaning towards removing it altogether, and let users define `massProdRates` on their own should they need it. 

In addition, several legacy MATLAB examples currently integrate reactor equations using MATLAB’s own ODE solvers while using Cantera only for thermochemical and kinetic properties (`ignite.m`, `ignite_hp.m`, `ignite_uv.m`). I'm thinking about deprecating these examples since I don't think many users are going to writing their own flame models and only use Cantera partially, plus these examples add helper functions that clutter the `samples` directory and not used elsewhere.

ischoegl:
> > > I didn't find the new equivalent function to `kin_getSourceTerms` in the generated clib.
> > 
> > 
> > This was hard-coded in the legacy CLib, and won't be replicated as such in the generated CLib. You have two options:
> > 
> > 1. Implement what the old function did in MATLAB
> > 2. Add it to the C++ code base
> 
> This function is used by only one example (`ignite.m`). I'm leaning towards removing it altogether, and let users define `massProdRates` on their own should they need it.

:+1: I think that's the way to go!

> In addition, several legacy MATLAB examples currently integrate reactor equations using MATLAB’s own ODE solvers while using Cantera only for thermochemical and kinetic properties (`ignite.m`, `ignite_hp.m`, `ignite_uv.m`). I'm thinking about deprecating these examples since I don't think many users are going to writing their own flame models and only use Cantera partially, plus these examples add helper functions that clutter the `samples` directory and not used elsewhere.

I believe that there's some value to those examples still, so it's worth keeping those (if I recall correctly, some were specifically designed to illustrate how to use with standard ODE solvers). Regarding helper functions, what you see in `ignite.m`, i.e., local functions, is the way to go whenever possible. At the same time, I'm not sure why `reactor_ode.m` wasn't integrated already. It would definitely be good to move the helper functions into the samples to reduce clutter (there may be a reason that I'm not aware of, though). Fwiw: https://www.mathworks.com/help/matlab/matlab_prog/local-functions.html

On a separate note, #2010 is now merged, so once you rebase, you should see the exception handling that was missing.

bryanwweber:
> > currently integrate reactor equations using MATLAB’s own ODE solvers while using Cantera only for thermochemical and kinetic properties...

> I believe that there's some value to those examples still

I agree there is value to them. We have equivalent examples in the Python set as well. Please don't remove them!

ssun30:
@ischoegl All the unit tests should be fixed by now. There are still a few issues with some samples that I'll defer to a future PR:

- `flame2.m` will now solve without crashing but no ignition occurs within the length of the stack. 
- `diff_flame.m` will now solve without crashing but the results don't match up with its python equivalent (the flame front should be ~0.6 cm from the left inlet, while it's about 5 cm in the current example). 
- `ignite.m` fails to solve using MATLAB's own ode15s solver. At 3e-4 seconds, reactor temperature blows up, forcing the solver to take an impossibly small step size. 
- `surf_reactor.m` has constant surface coverage throughout the simulation time.

I also noticed that `surf_getConcentrations`  will be changed in Cantera 3.2 to `thermo_getConcentrations`. I checked the generated C++ code and it uses `Phase::getConcentrations` which returns a volumetric concentration in kmol/m^3. But the legacy Clib uses `SurfPhase:getConcentrations` which I assume returns an area concentration in kmol/m^2. Is the new `thermo_getConcentrations` the correct method for the surface phase?

ischoegl:
@ssun30 That's great! One thing that comes to mind immediately is to check whether the Solution cloning works as expected. I hope to get a chance to look at this before the weekend.

PS: I did notice the diff jumping to 8k lines. Could you check whether you inadvertently added some files?

test/matlab_experimental/matlabCoverage.xml
ischoegl, line 1: Please remove :smile:

ischoegl:
@ssun30 ... I got a chance to look over some of this this morning (I also rebased, just in case).

> * `flame2.m` will now solve without crashing but no ignition occurs within the length of the stack.

After setting an initial temperature profile, the sample now gives reasonable results.

> * `diff_flame.m` will now solve without crashing but the results don't match up with its python equivalent (the flame front should be ~0.6 cm from the left inlet, while it's about 5 cm in the current example).

The width of the two examples was different, after reducing it to 2 cm in MATLAB, results now match.

> * `ignite.m` fails to solve using MATLAB's own ode15s solver. At 3e-4 seconds, reactor temperature blows up, forcing the solver to take an impossibly small step size.

This sample looks like it needs some updating / cleanup, as the code does not resemble `reactor1.m` as the docstring claims (the reactor ode has a loop that's likely not needed, etc). I did move the `reactor_ode.m` helper function into a local function, and it doesn't seem to matter. Also did the same for `ignite_uv.m` and `ignite_hp.m`. To me, it would be very interesting to know **whether the results match the legacy toolbox** for any of these ignite samples (i.e., Cantera 3.0). Runtime seems really slow also, so a comparison would be interesting as well. I believe the `outofprocess` mode will add overhead?

> * `surf_reactor.m` has constant surface coverage throughout the simulation time.

~You found a bug that's loosely related to #2000. Will have to address this outside of this PR (and reopen this issue).~ _Update: I misdiagnosed, this actually works; the issue was related to the new `clone` argument._

> I also noticed that `surf_getConcentrations` will be changed in Cantera 3.2 to `thermo_getConcentrations`. [...]

I looked into the C++ code, and the `SurfPhase` class does not define its own `getConcentrations`. That means that this is resolved properly. The C++ docstring does appear to be inaccurate.

ischoegl:
@ssun30 ... I noticed that there still are a lot of tests that are filtered. Could you check whether the new CLib would help?
```
Failure Summary:

     Name                                               Failed  Incomplete  Reason(s)
    ================================================================================================
     ctTestFlowDevice/testMFC                                       X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowDevice/testMFCErrors                                 X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowDevice/testValve3                                    X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowDevice/testValveTiming                               X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowDevice/testPressureController                        X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowReactor1/testNonReacting                             X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowReactor1/testReacting                                X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowReactor1/testCatalyticSurface                        X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowReactor1/testComponentNames                          X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowReactor2/testNoMassFlowRate                          X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowReactor2/testMaxSteps                                X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowReactor2/testMaxTimeStep                             X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowReactor2/testIterationLimits                         X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowReactor2/testSolverOrder                             X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowReactor2/testReinitialization                        X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestFlowReactor2/testInitialConditionTolerances              X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestPureFluid/testTPQ                                        X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestPureFluid/testPhaseOfMatter                              X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestPureFluid/testWaterBackends                              X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestReactor/testHeatTransfer2                                X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestReactor/testWallVelocity                                 X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestReactor/testHeatFluxFunc                                 X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestReactor/testReinitialize                                 X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestReactor/testPreconditionerUnsupported                    X       Filtered by assumption.
    ------------------------------------------------------------------------------------------------
     ctTestWellStirredReactor/testSteadyState                       X       Filtered by assumption.
```

ischoegl:
@ssun30 ... update: regarding surface coverages, I looked into this further, and it was the `clone` issue after all. This is fixed now.

From my perspective, I find that there are still some raw edges in the toolbox, but their resolution may not necessarily be in scope of this PR. What I found by looking at the MATLAB samples is:
* `ignite.m`... ~does not work (and contains a bug)~ _Update: I was able to reproduce the integration error reported earlier after fixing a minor bug I had inadvertently introduced. Sample runs, but is excruciatingly slow and generates 3000 columns(!) of output._
* `plug_flow_reactor.m` ... very slow and has a bug in the output
* `flame.m` ... should be moved to the core toolbox and rewritten as a class (derived from `Sim1D`)

I recommend opening an issue to review samples and update as necessary, where the resolution would be to carefully demonstrate the full capabilities of the revised toolbox, rather than just making the old samples work.

I have completed the changes that I saw as necessary to move this PR forward. As I had a rather heavy hand in edits here (apologies!), I'll request a review from @speth, and/or @bryanwweber. As mentioned above, I am :+1: with leaving some raw edges here to be addressed in follow-ups, as I view the scope of this PR as being the migration to the generated CLib.

_PS: many of the `assumeFail` bypasses in the test suite are based on missing (?) functions that can be easily implemented now (for CLib, this may be as simple as adding the corresponding entries in the header specification YAML files). This may or may not be within scope here; as the `oneD` tests are currently missing, this can be deferred._

interfaces/matlab_experimental/Reactor/MassFlowController.m
```diff
@@ -20,14 +20,10 @@
 
     methods
 
-        function m = MassFlowController(upstream, downstream, name)
+        function m = MassFlowController(varargin)
```
ischoegl: Not sure that the varargin is an improvement.

interfaces/matlab_experimental/Reactor/IdealGasReactor.m
```diff
@@ -17,19 +17,27 @@
...
+        function r = IdealGasReactor(phase, name, clone)
             % Constructor
 
-            if nargin < 2
-                name = '(none)';
+            arguments
+                phase {mustBeA(phase, 'Solution')}
+                name (1,1) string = "(none)"
+                clone (1,1) logical = true
             end
```
ischoegl, line 37: I really apreciate the simplification that the `arguments` block introduces. While it goes beyond the scope of the PR, it should be adopted wherever possible.

interfaces/matlab_experimental/Reactor/ReactorBase.m
```diff
@@ -0,0 +1,241 @@
...
+            if strcmp(flag, 'on')
+                cflag = true;
+            elseif strcmp(flag, 'off')
+                cflag = false;
+            else
+                error('Input must be "on" or "off"');
+            end
+
+            ctFunc('reactor_setChemistry', r.id, cflag);
+        end
```
ischoegl: This would simplify dramatically if you require a boolean as input. This is pre-existing code that was simply moved when `Reactor.m` was renamed to `ReactorBase.m`.

interfaces/matlab_experimental/Reactor/ReactorBase.m
```diff
@@ -0,0 +1,241 @@
...
+
+        function set.energy(r, flag)
+
+            if strcmp(flag, 'on')
```
ischoegl: Same: just require boolean input (there may be other occurrences).

interfaces/matlab_experimental/Reactor/ReactorBase.m
```diff
@@ -0,0 +1,241 @@
...
+            ctFunc('reactor_setName', r.id, name);
+        end
+
+        function set.V(r, v0)
+            ctFunc('reactor_setInitialVolume', r.id, v0);
+        end
+
+        function set.massFlowRate(r, MFR)
+            ctFunc('reactor_setMassFlowRate', r.id, MFR);
+            r.massFlowRate = MFR;
```
ischoegl: There isn't a need to buffer the value?

interfaces/matlab_experimental/Reactor/ReactorBase.m
```diff
@@ -0,0 +1,241 @@
...
+        function massFractions = get.Y(r)
+
+            nsp = r.phase.nSpecies;
+            massFractions = zeros(1, nsp);
+
+            for i = 1:nsp
+                massFractions(i) = r.massFraction(i);
+            end
+
+        end
```
ischoegl, line 175: Highly inefficient code. This method should copy the array in a single CLib operation.

interfaces/matlab_experimental/Reactor/ReactorBase.m
```diff
@@ -0,0 +1,241 @@
...
+            %    String or one-based integer id of the species.
+
+            if ischar(species)
+                k = r.phase.speciesIndex(species) - 1;
+            else
+                k = species - 1;
+            end
+
+            yi = ctFunc('reactor_massFraction', r.id, k);
+        end
```
ischoegl, line 171: Should be limited to string parameters. Access by index should follow from get.Y (with appropriate switch to access to full mass fraction vector).

ischoegl (commented):
@ssun30 ... I left a review on the portion of `ReactorBase.m` that is pre-existing content that was moved during renaming from `Reactor.m`. You need to create a new CLib function `reactor_getMassFractions` which should be based on `Reactor::massFractions` where all you need is to add it to the YAML header `ctreactor.yaml` (~didn't test this though~ _edit: tested it and there's a missing crosswalk that I'll have to add._). I'd be happy to assist if you run into issues.

ssun30:
@ischoegl I will fix some of the rougher edges in this PR since some commits are resolving them anyway, and I will squash these together. 

As for performance, yes there is a performance penalty with the "outofprocess" mode. This is because each function call requires transferring data between the MATLAB process and the external C++ process, which involves memory reallocation and type conversion. This is notable in scripts that repeatedly query object properties (such as TPY). And the examples with hand-written ODEs and solved with MATLAB's own solver are the worst offender. This is especially bad on Windows with unoptimized compilation (which is already 10x slower than MacOS and Linux with comparable hardware). 

I profiled two examples on MacOS 15.7 (M2 silicon). The first example is `diff_flame`

In-process:
profiler_diffflame_macos_inprocess
Out-of-process:
profiler_diffflame_macos_outofprocess
Most of the execution time was spent on 2 calls to `Sim1D.solve`, and since there is no transfer of data, the performance difference between the two modes is minimal. 

Then we have `plug_flow_reactor`

In-process:
profiler_pfr_macos_inprocess
Out-of-process:
profiler_pfr_macos_outofprocess
There were over 100k calls to `ctFunc` and 33k calls to `ctArray` by the `PFR_Solver` function to construct the ODE. Each call requires allocation of new memory space and type conversion. I don't think this will be a typical scenario for users if they build their simulations using Cantera's native classes.

ischoegl:
@ssun30 ... just pushed the fix for `reactor_massFractions`. Ended up being only a couple of lines in sourcegen. This should reduce the number of calls to `ctFunc` dramatically?

interfaces/matlab_experimental/Reactor/ReactorBase.m
ischoegl: I ended up fixing this myself, as it required changes in `sourcegen`.

interfaces/matlab_experimental/Reactor/ReactorBase.m
ischoegl: Direct vector access via `get.Y` is now implemented.

interfaces/matlab_experimental/Base/Kinetics.m
```diff
@@ -445,12 +431,12 @@
...
-
-        function ydot = get.massProdRate(kin)
-            nsp = kin.nTotalSpecies;
-            ydot = ctArray('kin_getSourceTerms', nsp, kin.kinID);
+            buf = clib.array.ctMatlab.Double(nr);
+            iok = ctFunc('kin_getRevRateConstants', kin.kinID, buf, 0);
+            if iok ~= 0
+                error('Cantera:ctError', ctGetErr);
+            end
+            k = buf.double;
```
ischoegl: Could you modify `ctArray` to accommodate additional parameters?

interfaces/matlab_experimental/Base/Kinetics.m
```diff
@@ -161,14 +147,14 @@
...
+            %   >> rxn = kin.reactionEquation(irxn)
             %
             % :param irxn:
             %    Integer index of the reaction.
             % :return:
             %    String reaction equation.
-
-            rxn = ctString('kin_getReactionString', kin.kinID, irxn - 1);
+            r = ctFunc('kin_reaction', kin.kinID, irxn - 1);
+            rxn = ctString('rxn_equation', r);
```
ischoegl, line 157: It would be good to create a rudimentary `Reaction` base class here. This is a change from the old CLib approach.

include/cantera/oneD/Flow1D.h
```diff
@@ -246,6 +247,27 @@ class Flow1D : public Domain1D
     //! @param j  Point at which to enable the energy equation. `npos` means all points.
     void solveEnergyEqn(size_t j=npos);
 
+    //! Check if energy is enabled for entire domain.
+    bool allOfEnergyEnabled() {
```
speth: Please annotate these methods as both (a) new in Cantera 3.2 and (b) experimental / subject to removal. The idea of solving the energy equation at some arbitrary subset of grid points while holding it fixed at others is a feature that I don't think is ever actually used (and we certainly don't test that it works), so I'd like to pave the way for substantially simplifying this at some point in the future.

include/cantera/oneD/DomainFactory.h
```diff
@@ -70,6 +70,16 @@ shared_ptr<T> newDomain(
...
+//! Create a Boundary1D object of the specified type.
+//! @param domainType  string identifying domain type.
+//! @param solution  Solution holding ThermoPhase, Kinetics and Transport objects.
+//! @param id  string identifier describing domain. If omitted, id defaults to the
+//!     domain type identifier.
+//! @since New in %Cantera 3.2.
+//! @ingroup onedGroup
+shared_ptr<Domain1D> newDomain1D(
+    const string& domainType, shared_ptr<Solution> solution, const string& id="");
+
```
speth: This is effectively just a duplicate of `newDomain`.

include/cantera/numerics/SteadyStateSystem.h
```diff
@@ -190,6 +190,16 @@ class SteadyStateSystem
     //!     after further unsuccessful solution attempts.
     void setTimeStep(double stepsize, size_t n, const int* tsteps);
 
+    //! Set the number of time steps to try when the steady Newton solver is
+    //! unsuccessful.
+    //! @param stepSize  Initial time step size [s]
+    //! @param tSteps  A sequence of time step counts to take after subsequent failures
+    //!     of the steady-state solver. The last value in `tsteps` will be used again
+    //!     after further unsuccessful solution attempts.
+    void setTimeStep(double stepSize, vector<int> tSteps) {
```
speth:
```suggestion
    void setTimeStep(double stepSize, const vector<int>& tSteps) {
```

include/cantera/oneD/Domain1D.h
```diff
@@ -590,25 +602,22 @@ class Domain1D
...
 
     /**
-     * Return a concise summary of a Domain.
+     * Print a concise summary of a Domain.
      * Skips keys input while `vector<string>` is not implemented in sourcegen.
      * @see SolutionArray.info()
      * @param rows  Maximum number of rendered rows.
      * @param width  Maximum width of rendered output.
-     * @since New in %Cantera 3.2.
      */
```
speth: This seems to be an accidental reversion of updates that were previously made during review/revision of #2008.

interfaces/matlab_experimental/cantera/ctBuildInterface.m
```diff
@@ -1,4 +1,4 @@
-function ctBuildInterface(ctDir, includeDir, ctLibDir)
```
speth, line 1: The `readme.md` file needs to be updated to incorporate this change in the function signature.

interfaces/matlab_experimental/Reactor/Reactor.m
```diff
@@ -1,268 +1,39 @@
...
-        %    reactions, respectively.
-        energy
-
-        massFlowRate % Mass flow rate in kg/s.
-
-    end
+classdef Reactor < ReactorBase
+    % Reactor Class ::
+    %
+    %     >> r = Reactor(content, name, clone)
```
speth: First parameter needs to be named `phase` to be consistent with the documentation that follows.

interfaces/matlab_experimental/Reactor/ReactorBase.m
speth: If renaming to a boolean, the property name should renamed to `energy_enabled` (and likewise `chemistry_enabled` below.

interfaces/matlab_experimental/Reactor/ReactorSurface.m
```diff
@@ -14,10 +14,14 @@
...
-    % :param reactor:
-    %    Instance of class 'Reactor' to be used as the adjacent bulk phase.
+    % :param reactors:
+    %    An instance of or a cell array of instances of class :mat:class:`ReactorBase`.
     % :param name:
     %    Reactor surface name (optional; default is ``(none)``).
+    % :param clone:
+    %    Determines whether to clone `content` so that the internal state of
+    %    this reactor is independent of the original Solution object and
+    %    any Solution objects used by other reactors in the network.
```
speth, line 24:
```suggestion
    % :param clone:
    %    Determines whether to clone `surf` so that the internal state of
    %    this reactor is independent of the original Solution object and
    %    any Solution objects used by other reactors in the network.
```
(or rename the argument `phase` in all locations for consistency with the base class?)

interfaces/matlab_experimental/Reactor/Reservoir.m
```diff
@@ -22,19 +22,27 @@
     %     Cantera :mat:class:`Solution` to be set as the contents of the reactor.
     % :param name:
     %     Reservoir name (optional; default is ``(none)``).
+    % :param clone:
+    %    Determines whether to clone `content` so that the internal state of
+    %    this reactor is independent of the original Solution object and
+    %    any Solution objects used by other reactors in the network.
```
speth, line 28:
```suggestion
    % :param clone:
    %    Determines whether to clone `phase` so that the internal state of
    %    this reactor is independent of the original Solution object and
    %    any Solution objects used by other reactors in the network.
```

interfaces/matlab_experimental/Reactor/Valve.m
```diff
@@ -33,14 +33,10 @@
...
-        function v = Valve(upstream, downstream, name)
+        function v = Valve(varargin)
             % Constructor
 
-            if nargin < 3
-                name = '(none)';
-            end
-
-            v@FlowDevice('Valve', upstream, downstream, name);
+            v@FlowDevice('Valve', varargin{:});
```
speth: Like the other place where @ischoegl commented, I think this was better without `varargin`.

samples/matlab_experimental/flame1.m
```diff
@@ -19,20 +19,17 @@
...
 
-loglevel = 1; % amount of diagnostic output (0
-% to 5)
-
-refine_grid = 1; % 1 to enable refinement, 0 to
-% disable
-max_jacobian_age = [5, 10];
+logLevel = 1; % amount of diagnostic output (0 to 5)
+refineGrid = 1; % 1 to enable refinement, 0 to disable
+maxJacobianAge = [5, 10];
```
speth: The default tolerances and Jacobian settings for the flame solver generally work quite well. I think it would be better to demonstrate _not_ setting them for no reason. Furthermore, at least on my computer, this example fails with these settings but succeeds with the default ones.

samples/matlab_experimental/flame2.m
```diff
@@ -22,14 +22,14 @@
...
 comp2 = 'C2H6:1'; % fuel composition
 
-initial_grid = 0.02 * [0.0, 0.2, 0.4, 0.6, 0.8, 1.0]; % m
+width = 0.02;
+nz = 11;
 
-tol_ss = {1.0e-5, 1.0e-13}; % {rtol atol} for steady-state problem
-tol_ts = {1.0e-4, 1.0e-13}; % {rtol atol} for time stepping
+tol_ss = {1.0e-4, 1.0e-9}; % {rtol atol} for steady-state problem
+tol_ts = {1.0e-4, 1.0e-11}; % {rtol atol} for time stepping
```
speth: Likewise here, though at least this example appears to solve successfully.

samples/matlab_experimental/ignite.m
```diff
@@ -21,21 +21,80 @@
...
+        vol = this_y(2);
+        masses = this_y(3:end);
+
+        % evaluate the total mass, and the specific internal energy and volume.
+        total_mass = sum(masses);
+        u_mass = int_energy / total_mass;
+        v_mass = vol / total_mass;
+
+        % set the state of the gas by specifying (u,v,{Y_k})
+        gas.Y = masses;
```
speth: Does access to the C++ function `Phase::setMassFractions_NoNorm` exist in the Matlab interface? It is critical to getting reasonable performance when integrating this system of equations if the Jacobian is being evaluated using finite differences (which is the case here).

samples/matlab_experimental/plug_flow_reactor.m
```diff
@@ -172,3 +172,62 @@
...
+    %% Plug flow reactor governing equations
+    %
+    % This function defines the spatial derivatives for an ideal gas plug-flow
+    % reactor, where the cross-sectional area and pressure are allowed to vary
+    % axially.
+    %
+    % The model is set up by the example file :doc:`plug_flow_reactor.m
+    % <plug_flow_reactor>`, which points the integrator to this function. The integrator
+    % integrates the derivatives spatially, to solve the density, temperature, and
+    % species mass fraction profiles as a function of distance x.
```
speth: This needs to be updated -- the setup is of course just in this same file...

.gitignore
```diff
@@ -9,6 +9,8 @@ __pycache__
 *.egg-info
 *.obj
 *.exe.manifest
+*.h5
+*.hdf5
```
speth: I don't think adding these excludes is necessary or advisable. While we generally don't want to add binary files to the repo, there are a handful of these in the `test/data` directory, and it's annoying to work around having these files in the exclude, when the best practice to prevent accidentally adding these files is to never run `git add .`.

speth (changes requested):
Thanks for all the work on this, @ssun30 and @ischoegl. I think all of my comments on this are relatively minor.

speth:
One other minor update that would be worth including in this PR, which caused me a bit of trouble while trying to test locally: it appears that the argument name `InterfaceName` was only [introduced in R2024a](https://www.mathworks.com/help/matlab/ref/clibgen.generatelibrarydefinition.html). The older name for this is `PackageName`, which needs to be used in both `ctBuildInterface.m` and `ctGenerateLibraryDefinitions.m` to support older Matlab versions (I inconveniently have R2023b installed).

include/cantera/oneD/Flow1D.h
ischoegl: Agreed!

interfaces/matlab_experimental/Reactor/ReactorBase.m
ischoegl: Renamed in C++ (with old version deprecated).

samples/matlab_experimental/ignite.m
ischoegl: It is not, but can be easily exposed to the MATLAB API by adding this to `ctthermo.yaml`:
```YAML
- name: setMassFractions_NoNorm
  uses: nSpecies
```
@ssun30 ... I have not made this change, and am merely mentioning this here to showcase how easy it is now to expose additional C++ features.

.gitignore
ischoegl: Replacing this by 
```
samples/**/*.h[5df]
```
as hdf files in `samples` are the actual issue.

include/cantera/oneD/DomainFactory.h
ischoegl: Indeed, but I don't think `sourcegen` can handle templates properly yet (although it could be that it does so accidentally). Regardless, it's possible to actually bypass this, so I removed it.

ischoegl:
> Thanks for all the work on this, @ssun30 and @ischoegl. I think all of my comments on this are relatively minor.

@speth ... thanks for the review! Indeed, the comments are relatively minor. I addressed whatever pertains to exposing C++ to the MATLAB API. I'll leave the remainder to @ssun30 - I'm happy to chime in for questions, of course!

ischoegl:
@ssun30 ... had to fix one more issue.

interfaces/matlab_experimental/Reactor/ReactorBase.m
ssun30: The cpp class `FlowReactor` doesn't have a getter method. The Python interface doesn't have a getter for `mass_flow_rate` either so it seems this is not a value that's exposed to the user. Also this function should be moved to `FlowReactor.m`?

interfaces/matlab_experimental/Base/Kinetics.m
ssun30: I think I will defer this to a future PR to implement `Reaction` base classes with methods defined in `ctrxn.h`

interfaces/matlab_experimental/Reactor/ReactorBase.m
ischoegl: I think it's ok to defer. I noticed several of those instances where a getter method is missing, but it is now trivial to implement a fix.

ischoegl:
@ssun30 - thanks! I added a couple of docstring annotations that I had overlooked.

samples/matlab_experimental/ignite.m
ischoegl: Please note that the generated CLib doesn't include the flag `norm` that the legacy CLib had.

ischoegl:
@ssun30 and @speth ... I ended up simplifying `ctBuildInterface` as there is no need to specify two include folders. After installation, the `/path/to/cantera_install/include` folder holds `cantera` and `cantera_clib` in parallel folders.

The main difference is that this requires a `scons install`, and avoids using the build folders.

~PS: I also switched to `inprocess` defaults on macOS and Windows, but can roll that change back.~ _Edit: rolled back._

speth:
> The main difference is that this requires a `scons install`, and avoids using the build folders.

I think being able to test (both the test suite and interactively) without running `scons install` is incredibly valuable, and not a feature I would want to lose for the MATLAB toolbox or any of the interfaces.

ischoegl:
> I think being able to test (both the test suite and interactively) without running `scons install` is incredibly valuable, and not a feature I would want to lose for the MATLAB toolbox or any of the interfaces.

Fair point. Thankfully, the dependency on `cantera/base/config.h` in `clib_defs.h` was unnecessary, so I was able to maintain the simplification without needing to install. I.e.,

   ```matlab
   ctToolboxDir = '/path/to/cantera/source/interfaces/matlab_experimental';
   ctIncludeDir = '/path/to/cantera/include';
   ctLibDir = '/path/to/cantera/library';
   ctBuildInterface(ctToolboxDir, ctIncludeDir, ctLibDir);
   ```

I'll leave this for a final review now. There are many rough edges, but the migration to the generated CLib is accomplished (thank you @ssun30!). Additional improvements can be made in follow-up PRs.

interfaces/matlab_experimental/Reactor/ReactorBase.m
```diff
@@ -0,0 +1,237 @@
...
+        % "advance" or "step".
+        Y
+
+        % Enable or disable changing reactor composition by reactions. ::
+        %
+        %     >> r.chemistry = flag
+        %
+        % If the chemistry is disabled, then the reactor composition is
+        % constant. The parameter should be the string "on" to enable
+        % the species equations, or "off" to disable it.
```
speth: - The example should say `r.chemistryEnabled = flag`, given the update to the name.
- Setting `chemistryEnabled` does not force the composition of the reactor to remain constant. While reactions are disabled, the composition can still change depending on any mass flow into the reactor.

interfaces/matlab_experimental/Reactor/ReactorBase.m
```diff
@@ -0,0 +1,237 @@
...
+        % :param r:
+        %    Instance of :mat:class:`ReactorBase`.
+        % :param flag:
+        %    Boolean to enable or disable chemical reactions.
+        chemistryEnabled
+
+        % Enable or disable solving the energy equation. ::
+        %
+        %     >> r.energy = flag
+        %
```
speth:
```suggestion
        % Enable or disable solving the energy equation. ::
        %
        %     >> r.energyEnabled = flag
        %
```

interfaces/matlab_experimental/Reactor/ReactorBase.m
```diff
@@ -0,0 +1,237 @@
...
+        % If the energy equation is disabled, then the reactor
+        % temperature is constant. The parameter should be the string
+        % "on" to enable the energy equation, or "off" to disable it.
+        %
+        % By default, :mat:class:`ReactorBase` objects are created with the energy
+        % equation enabled, so usually this method.
```
speth: So usually this method what?

interfaces/matlab_experimental/Reactor/ReactorBase.m
```diff
@@ -0,0 +1,237 @@
...
+        %
+        % :param r:
+        %    Instance of :mat:class:`ReactorBase`.
+        % :param flag:
+        %    Boolean to enable or disable energy equations.
+        energyEnabled
+
+        massFlowRate % Mass flow rate in kg/s.
```
speth: This is for `FlowReactor` only, I guess? Can we remove it from the base class?

interfaces/matlab_experimental/Reactor/ReactorBase.m
```diff
@@ -0,0 +1,237 @@
...
+            massFractions = ctArray('reactor_massFractions', r.phase.nSpecies, r.id);
+        end
+
+        function flag = get.chemistryEnabled(r)
+            flag = ctFunc('reactor_chemistryEnabled', r.id);
+        end
+
+        function flag = get.energyEnabled(r)
+            flag = ctFunc('reactor_energyEnabled', r.id);
+        end
```
speth, line 183: These methods aren't actually defined on the C++ `ReactorBase` class, so I have to assume that calling these raises an exception.

speth (changes requested):
Thanks for the updates, @ssun30 and @ischoegl. Just a couple of minor issues to fix here, mostly documentation related.

interfaces/matlab_experimental/Reactor/ReactorBase.m
ischoegl: I removed that section. If you inspect a `Reactor` object, it will tell you whether things are enabled or disabled at any time.

interfaces/matlab_experimental/Reactor/ReactorBase.m
ischoegl: Moved!

interfaces/matlab_experimental/Reactor/ReactorBase.m
ischoegl: Good question! ~No - this won't raise an exception, as~ the MATLAB inheritance doesn't follow the C++ inheritance (the MATLAB inheritance is flat, rather than following the complex structure in C++). Specifically,
```
    int32_t reactor_energyEnabled(int32_t handle)
    {
        // getter: bool Reactor::energyEnabled()
        try {
            bool out = ReactorBaseCabinet::as<Reactor>(handle)->energyEnabled();
            return int(out);
        } catch (...) {
        }
    }
```
The C++ getter should likely be moved, but that's a different issue.

PS: it _will_ throw an exception for a `Reservoir` though:
```matlab
>> env.chemistryEnabled
Error using [ctFunc](matlab:matlab.lang.internal.introspective.errorDocCallback('ctFunc', '/Volumes/Data/work/GitHub/cantera/interfaces/matlab_experimental/Utility/ctFunc.m', 10)) ([line 10](matlab: opentoline('/Volumes/Data/work/GitHub/cantera/interfaces/matlab_experimental/Utility/ctFunc.m',10,0)))

CanteraError thrown by Cabinet::as:
Item is not of the correct type.

Error in [ReactorBase/get.chemistryEnabled](matlab:matlab.lang.internal.introspective.errorDocCallback('ReactorBase/get.chemistryEnabled', '/Volumes/Data/work/GitHub/cantera/interfaces/matlab_experimental/Reactor/ReactorBase.m', 178)) ([line 178](matlab: opentoline('/Volumes/Data/work/GitHub/cantera/interfaces/matlab_experimental/Reactor/ReactorBase.m',178,0)))
            flag = logical(ctFunc('reactor_chemistryEnabled', r.id));
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
where the implementation is not symmetric in C++, as
```
>> env.chemistryEnabled = false
Error using [ctFunc](matlab:matlab.lang.internal.introspective.errorDocCallback('ctFunc', '/Volumes/Data/work/GitHub/cantera/interfaces/matlab_experimental/Utility/ctFunc.m', 10)) ([line 10](matlab: opentoline('/Volumes/Data/work/GitHub/cantera/interfaces/matlab_experimental/Utility/ctFunc.m',10,0)))

NotImplementedError thrown by ReactorBase::setChemistryEnabled:
Not implemented.

Error in [ReactorBase/set.chemistryEnabled](matlab:matlab.lang.internal.introspective.errorDocCallback('ReactorBase/set.chemistryEnabled', '/Volumes/Data/work/GitHub/cantera/interfaces/matlab_experimental/Reactor/ReactorBase.m', 201)) ([line 201](matlab: opentoline('/Volumes/Data/work/GitHub/cantera/interfaces/matlab_experimental/Reactor/ReactorBase.m',201,0)))
            ctFunc('reactor_setChemistryEnabled', r.id, flag);
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
I'd prefer to leave the inheritance as is, mainly due to the fact that `SurfaceReactor` does not inherit from `Reactor` in C++, and thus would require duplication of some property getters/setters.
https://cantera.org/dev/cxx/d1/d7b/classCantera_1_1ReactorBase.html

Overall, this observation mainly points to the fact that it is currently not possible to toggle kinetics/energy for a `ReactorSurface` in C++?

ischoegl: I decided to fix in C++ to avoid confusing error messages.

ischoegl:
@speth / @ssun30 ... I went ahead and made all the requested changes, which also extend to improved exception handling for enabling/disabling energy and chemistry.

speth (approved):
Thanks, this looks good to me.

PR #2001, opened 2025-10-08
Toward consistent naming of reactor "contents"

speth:

Python:
- Add new `contents` property for `ReactorBase` objects that returns a `Solution` object
- Deprecate old `thermo` and `kinetics` properties of `ReactorBase`, `Reactor`, and `ReactorSurface`

C++:
- Deprecate the existing `ReactorBase::contents` method that returns `ThermoPhase&`
- Rename the `solution` property of `ReactorBase` (which returns `shared_ptr<Solution>` to `contents4`, with the expectation of renaming this `contents` in a future version
- Change the implementation of `ReactorSurface::syncState` which actually did the opposite of the `ReactorBase::syncState` method. I don't think there's a non-breaking way to implement this change, but I also don't think there is much if any user code that calls this method directly. Normally, users would rely on `Reactor` methods to do this synchronization.

**Checklist**

### DIFF
--- a/doc/sphinx/userguide/extensible-reactor.md
+++ b/doc/sphinx/userguide/extensible-reactor.md
@@ -137,7 +137,7 @@ for mass_rock in [0.0, 1.0, 3.0]:
     # Integrate for 10 seconds
     net.advance(10)
 
-    print(f'mass_rock = {mass_rock} kg; ΔT = {r1.thermo.T - T0:.2f} K')
+    print(f'mass_rock = {mass_rock} kg; ΔT = {r1.contents.T - T0:.2f} K')
 ```
 
 As expected, we see that the temperature rise decreases as the thermal mass of the
--- a/doc/sphinx/userguide/reactor-tutorial.md
+++ b/doc/sphinx/userguide/reactor-tutorial.md
@@ -38,7 +38,7 @@ reac = ct.IdealGasReactor(gas)
 sim = ct.ReactorNet([reac])
 
 # View the initial state of the mixture
-reac.thermo()
+reac.contents()
 ```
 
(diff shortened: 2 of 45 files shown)


ischoegl (commented):
Thank you, @speth! I appreciate you implementing my suggestion.

ischoegl (approved):

ischoegl:
@speth ... thanks again for taking this on. Replacing `Reactor.thermo` with `contents` is a significant improvement, as now all objects are accessed via the associated `Solution`. Since I merged this PR, I looked into adopting the same nomenclature for `Domain1D` and `SolutionArray`. Based on what I saw, I am now wondering whether `phase` may not have been a better long-term name, as it will be less disruptive elsewhere. I realize that there is still an issue with `Phase`/`ThermoPhase` being used for handling of thermo properties on the C++ side. I.e., a switch would result in less disruption to user-facing APIs, albeit with some less consistent naming choices in the C++ codebase. (I'm also aware that not every `ThermoPhase` satisfies thermodynamic definitions of a _Phase_, which is an entirely different ball of hair).

Since I pushed on the current `contents` (and quickly merged), I would be happy to make the required updates myself. Please let me know.

speth:
Naming this accessor "phase" does have the benefit of eliminating the need for the transitional "contents4" name, as well as well as being the shortest practical name I can imagine.

ischoegl:
> Naming this accessor "phase" does have the benefit of eliminating the need for the transitional "contents4" name, as well as well as being the shortest practical name I can imagine.

Perfect - add the benefits on the side of `SolutionArray` and `Domain1D`, and it's a win-win. Will create a PR over the weekend.

Issue #1999, opened 2025-10-08
Intermittent CI failure running catalytic_combustion.py example


speth:
**Problem description**

CI jobs are occasionally indicating integration failures in the `catalytic_combustion.py` example.

**Steps to reproduce**

- Trigger CI jobs on the main branch.
- Rerunning the job after a failure can result in a successful run.

**Behavior**

Output from failing cases:
```
Traceback (most recent call last):
  File "/home/runner/work/cantera/cantera/samples/python/onedim/catalytic_combustion.py", line 68, in <module>
    surf_phase.advance_coverages(1.0)
  File "build/python/cantera/kinetics.pyx", line 943, in cantera.kinetics.InterfaceKinetics.advance_coverages
cantera._utils.CanteraError: 
*******************************************************************************
CanteraError thrown by CVodesIntegrator::integrate:
CVodes error encountered. Error code: -3
At t = 0.0583727 and h = 4.24999e-08, the error test failed repeatedly or with |h| = hmin.

Components with largest weighted error estimates:
4: -392.219827383851
9: 20.872449014364545
5: -0.0024577841850999837
1: -0.0002620613619660072
0: -0.00011714129027371423
2: -0.00011122842455574104
10: 9.99021834700569e-05
3: -5.6773903789372044e-05
7: -3.898277339646901e-157
8: -9.688141290158232e-158
*******************************************************************************
```

**System information**

- Cantera version: `main` at d7f7a0a6a
- OS: Ubuntu 24.04
- Python 3.10

**Attachments**

- Full log from one failing run: [python310_ubuntu2404.txt](https://github.com/user-attachments/files/22771153/python310_ubuntu2404.txt)

**Additional context**

bryanwweber:
This is pure speculation, but I wonder if there's a difference between Intel and AMD CPUs, so depending on which runner is chosen for the job it gives slightly different results? I imagine that GitHub have a mix of processors in their data center. I also wonder if we're the only community dealing with this kind of failure. Surely we're not the only people solving ODE problems in GitHub actions 🤔

ischoegl:
Here’s another new failure that is intermittent, also Python 3.10 sample runner:
```
Traceback (most recent call last):
  File "/home/runner/work/cantera/cantera/samples/python/onedim/ion_burner_flame.py", line 32, in <module>
    f.solve(loglevel, auto=True)
[...]

Issue #1991, opened 2025-10-01
Intermittent CI failure running ion_burner_flame.py example


speth:
**Problem description**

**Steps to reproduce**

- Trigger CI jobs on the `main` branch.
- Rerunning the job after a failure can result in a successful run.

**Behavior**

Occasionally, some of the jobs that run the examples on Ubuntu fail, with the following log message:
```pytb
samples/python/onedim/ion_burner_flame.py
Traceback (most recent call last):
  File "/home/runner/work/cantera/cantera/samples/python/onedim/ion_burner_flame.py", line 32, in <module>
    f.solve(loglevel, auto=True)
  File "/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/cantera/onedim.py", line 923, in solve
    return super().solve(loglevel, refine_grid, auto)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "build/python/cantera/_onedim.pyx", line 1242, in cantera._onedim.Sim1D.solve
cantera._utils.CanteraError: Could not find a solution for the 1D problem
Error: Process completed with exit code 1.
```

Looking at the detailed log (see attached), the solver ultimately fails after trying several different initial grids. The failure modes are either:
- Failure to solve the steady-state problem even after taking the maximum number of timesteps (500). Here, the time step size is generally stuck in the vicinity of 1e-6 seconds.
- Error factorizing the Jacobian, immediately when starting integration with the energy equation enabled

**System information**

- Cantera version: `main` branch at 99550d0c1
- OS: Ubuntu 22.04
- Python 3.10 or Python 3.11

**Attachments**

- Logs from one particular CI run: [7_Run Python 3.11 examples on ubuntu-22.04, NumPy ==1.23.5.txt](https://github.com/user-attachments/files/22648673/7_Run.Python.3.11.examples.on.ubuntu-22.04.NumPy.1.23.5.txt)
- Full output from the failing example (extracted from the separate artifact upload): [ion_burner_flame_log.txt](https://github.com/user-attachments/files/22770878/ion_burner_flame_log.txt)

**Additional context**

Initially diagnosed while working on #1921.

PR #2002, opened 2025-10-11
Rename `ReactorBase::contents` to `ReactorBase::phase`

ischoegl:

This is a follow-up to #2001. Upon further consideration, updating the nomenclature to `phase` rather than `contents` will be less disruptive, as `SolutionArray` and 1D objects already use a `phase` property.

**Checklist**

### DIFF
--- a/doc/sphinx/python/zerodim.rst
+++ b/doc/sphinx/python/zerodim.rst
@@ -38,75 +38,75 @@ Reactors
 
 Reservoir
 ^^^^^^^^^
-.. autoclass:: Reservoir(phase=None, name=None)
+.. autoclass:: Reservoir(phase, name=None)
 
 Reactor
 ^^^^^^^
-.. autoclass:: Reactor(phase=None, *, clone=None, name=None, energy='on', node_attr=None, group_name="")
+.. autoclass:: Reactor(phase, *, clone=None, name=None, energy='on', node_attr=None, group_name="")
 
 MoleReactor
 ^^^^^^^^^^^
-.. autoclass:: MoleReactor(phase=None, *, clone=None, name=None, energy='on', node_attr=None, group_name="")
+.. autoclass:: MoleReactor(phase, *, clone=None, name=None, energy='on', node_attr=None, group_name="")
 
 IdealGasReactor
 ^^^^^^^^^^^^^^^
-.. autoclass:: IdealGasReactor(phase=None, *, clone=None, name=None, energy='on', node_attr=None, group_name="")
+.. autoclass:: IdealGasReactor(phase, *, clone=None, name=None, energy='on', node_attr=None, group_name="")
 
 IdealGasMoleReactor
 ^^^^^^^^^^^^^^^^^^^
-.. autoclass:: IdealGasMoleReactor(phase=None, *, clone=None, name=None, energy='on', node_attr=None, group_name="")
+.. autoclass:: IdealGasMoleReactor(phase, *, clone=None, name=None, energy='on', node_attr=None, group_name="")
 
 ConstPressureReactor
 ^^^^^^^^^^^^^^^^^^^^
-.. autoclass:: ConstPressureReactor(phase=None, *, clone=None, name=None, energy='on', node_attr=None, group_name="")
+.. autoclass:: ConstPressureReactor(phase, *, clone=None, name=None, energy='on', node_attr=None, group_name="")
 
 ConstPressureMoleReactor
 ^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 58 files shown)


ischoegl:
@speth ... with #2001 having done the heavier lift already, this was straightforward. It does touch some MATLAB code also (which had previously used `contents`), but will have the benefit that the APIs are now consistent.

On a separate note, CI run https://github.com/Cantera/cantera/actions/runs/18423070603 has the highest failure rate for #1991 and #1999 thus far (3 failed example runs on Python 3.11 and 3.12).

include/cantera/zeroD/ReactorFactory.h
```diff
@@ -71,7 +71,7 @@ shared_ptr<ReactorBase> newReactor(
 //! @param name  Name of the reactor.
 //! @since  New in %Cantera 3.2. Transitional method returning a `Reactor` object.
 shared_ptr<Reactor> newReactor4(
-    const string& model, shared_ptr<Solution> contents, bool clone=true,
+    const string& model, shared_ptr<Solution> phase, bool clone=true,
```
speth, line 74: The `@param` in the docstring also needs to be updated. Check the Doxygen logs for warnings about other instances.

doc/sphinx/python/zerodim.rst
```diff
@@ -38,75 +38,75 @@ Reactors
 
 Reservoir
 ^^^^^^^^^
-.. autoclass:: Reservoir(contents=None, name=None)
+.. autoclass:: Reservoir(phase=None, name=None)
 
 Reactor
 ^^^^^^^
-.. autoclass:: Reactor(contents=None, *, clone=None, name=None, energy='on', node_attr=None, group_name="")
+.. autoclass:: Reactor(phase=None, *, clone=None, name=None, energy='on', node_attr=None, group_name="")
```
speth: While we're at it, `phase` cannot be `None`.

interfaces/sourcegen/src/sourcegen/csharp/generator.py
```diff
@@ -215,9 +215,9 @@ def _write_file(self, file_name: str, template_name: str, **kwargs) -> None:
         _LOGGER.info(f"  writing {file_name!r}")
         t_file = Path(__file__).parent / template_name
         template = _LOADER.from_string(t_file.read_text(encoding="utf-8"))
-        contents = template.render(file_name=file_name, **kwargs)
+        phase = template.render(file_name=file_name, **kwargs)
 
-        self._out_dir.joinpath(file_name).write_text(contents, encoding="utf-8")
+        self._out_dir.joinpath(file_name).write_text(phase, encoding="utf-8")
```
speth: I'm assuming this is "find and replace" having gone a bit too far.

samples/python/kinetics/shock_tube.py
```diff
@@ -52,7 +52,7 @@
     gas = ct.Solution(file, name=models[m])
     X = {'H2O2':X_H2O2, 'H2O':X_H2O, 'O2':X_O2, 'CO2':X_CO2, 'AR':X_Ar}
     gas.TPX = 1196, 2.127*ct.one_atm, X
-    r = ct.Reactor(contents=gas, energy="on", clone=False)
+    r = ct.Reactor(phase=gas, energy="on", clone=False)
```
speth: Currently, existing usage of `Reactor(contents=gas)` will experience a breaking change with the renaming of this parameter. There are only a couple of examples where we demonstrate this using keyword arguments, so it may not be worth the effort to implement transitional behavior.

speth (changes requested):
Thanks, @ischoegl. This mostly looks good to me. Just a few small notes from me.

interfaces/sourcegen/src/sourcegen/csharp/generator.py
ischoegl: Absolutely!

samples/python/kinetics/shock_tube.py
ischoegl: Yes. I am personally not a fan of specifying a positional parameter as if it were a keyword argument (imho, one of the more questionable Python idiosyncrasies). I agree that only very few people will face breaking changes.

ischoegl:
Thanks, @speth! Comments should be addressed.

samples/python/kinetics/shock_tube.py
ischoegl: PS: I added a commit to remove usage of `phase=` to not encourage this practice (which, as this PR points out, can lead to breaking changes).

speth (approved):

PR #2009, opened 2025-10-13
Fix sourcegen markup

ischoegl:

Fix issues where XML markup causes docstrings extracted by sourcegen to cut off (`<bold>`, `<emphasis>`, `<computeroutput>`).

**Checklist**

### DIFF
--- a/include/cantera/transport/Transport.h
+++ b/include/cantera/transport/Transport.h
@@ -88,7 +88,7 @@ public:
 
     //! Create a new Transport object using the same transport model and species
     //! transport properties as this one.
-    //! @param phase ThermoPhase used to specify the state for the newly cloned
+    //! @param thermo ThermoPhase used to specify the state for the newly cloned
     //!     Transport object. Can be created from the phase used by the current
     //!     Transport object using the ThermoPhase::clone() method.
     //! @since New in %Cantera 3.2.
--- a/interfaces/sourcegen/src/sourcegen/headers/ct_auto.yaml
+++ b/interfaces/sourcegen/src/sourcegen/headers/ct_auto.yaml
@@ -8,6 +8,7 @@ base: ""
 recipes:
 - name: version  # Renamed in Cantera 3.2 (previously getCanteraVersion)
 - name: gitCommit  # Renamed in Cantera 3.2 (previously getGitCommit)
+- name: usesHDF5
 - name: addDataDirectory
 - name: getDataDirectories
 - name: findInputFile
--- a/interfaces/sourcegen/src/sourcegen/headers/tagfiles.py
+++ b/interfaces/sourcegen/src/sourcegen/headers/tagfiles.py
@@ -286,13 +286,22 @@ def tag_lookup(xml_path: Path, tag_info: TagInfo) -> TagDetails:
         _LOGGER.warning(msg)
         matches = matches[:1]
 
-    def no_refs(entry: str) -> str:
-        # Remove stray XML markup that causes problems with xml.etree
-        if "<ref" in entry:
-            regex = re.compile(r"<ref [\s\S]*?>")
-            for ref in re.findall(regex, entry):
-                entry = entry.replace(ref, "<ref>")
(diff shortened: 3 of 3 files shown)


bryanwweber:
Can you provide an example of this happening? If the tags are properly closed, etree shouldn't cut them off afaict

ischoegl:
> Can you provide an example of this happening? If the tags are properly closed, etree shouldn't cut them off afaict

`usesHDF5` - the issue is that we need to translate XML back to markup. (not necessarily that etree misbehaves)

bryanwweber:
I'm sorry for my lack of familiarity with this code, I don't feel like I can review this 😕 I was hoping if I asked I'd be able to better understand but I think it's just too deep in the weeds for me anymore! I'm not sure how to evaluate whether the change works or not, or whether it has any other side effects, or why it's necessary in the first place 🫣

ischoegl:
> I'm sorry for my lack of familiarity with this code, I don't feel like I can review this 😕 I was hoping if I asked I'd be able to better understand but I think it's just too deep in the weeds for me anymore! I'm not sure how to evaluate whether the change works or not, or whether it has any other side effects, or why it's necessary in the first place 🫣

Sorry, @bryanwweber ... if you look at the doxygen XML output that is generated for `usesHDF5`:
```xml
       <briefdescription>
<para>Returns true if Cantera was compiled with C++ <computeroutput>HDF5</computeroutput> support. </para>
        </briefdescription>
```
`sourcegen` needs to convert this back to a conventional markdown string for the doxygen `\brief` description, i.e.,
```
Returns true if Cantera was compiled with C++ `HDF5` support.
```

Before this fix, doxygen (or Jinja2?) cut off like so:
```
Returns true if Cantera was compiled with C++
```
as the XML tags hadn't been removed.

_PS: I updated the code to better document what is going on._

bryanwweber:
Thanks for the detailed explanation! I think I understand a little bit better now. I'm going to ask some questions/speculation still from my ignorance, so I'm not asking for any changes or further investigation in this PR! @speth feel free to approve and merge if this looks good.

> `sourcegen` needs to convert this back to a conventional markdown string for the doxygen `\brief` description

I see, `sourcegen` is... generating... the C++ source/header files that have to contain the documentation strings. Since parsing comments from the actual C++ source/header files is not at all worth pursuing, the most convenient option is to use the Doxygen XML output as the source for the documentation to be copied into the generated code.

I wonder if there's a way to pass the XML output to Doxygen when building docs for the generated code to say, this is the source for the docstrings of the generated files? That would avoid needing to copy the strings around.

> Before this fix, doxygen (or Jinja2?) cut off like so:

If Jinja2 is responsible for this, it's probably due to somehow escaping the reserved HTML characters. Is Doxygen XML valid in docstrings in the main source code? If not, Doxygen may indeed be at fault...

ischoegl:
@bryanwweber … `sourcegen` is the code I adopted for the generated CLib. So doxygen first documents the Cantera code base, which is available as XML. That XML is then used to auto-generate CLib. CLib is now itself fully documented, which is where Jinja2 templates come in. For processed output, see https://cantera.org/dev/cxx/dd/df6/group__CAPIindex.html (that documentation is generated from a second doxygen pass).

The legacy (hand-coded) CLib will be removed after 3.2. See https://cantera.org/dev/clib/index.html

bryanwweber (approved):
Sorry @ischoegl I think I'm not being clear, but that's on me. I'm asking questions here since 1) I don't think we found the root cause of this problem and 2) I _really_ don't like editing XML with regex :grimacing: In any case, this seems to work for the job that it has, so it looks good to me!

ischoegl:
@bryanwweber ... I am 99% certain that Jinja2 swallows the output, as the C++ headers generated from their templates cut off after the first XML tag. As we need markdown for doxygen anyways, there's no need to further investigate. I certainly see the point of XML/regex, but the editing only touches what I consider terminal nodes. Any of the XML formatting tags that are replaced only apply to those terminal nodes.

PR #2010, opened 2025-10-13
Make index lookups more robust

### LINKED ISSUE #2004: elementIndex and speciesIndex can cause segfaults
**Problem description**

The way both methods are implemented is that they return `npos` if an element/species is not found. While exception handling is present for most contexts, others are lacking those precautions: example `ThermoPhase::ThermoPhase::o2Present`. The exception handling should be consolidated to avoid segfaults, and overall simplify numerous instances with customized exception handling in C++ and user-facing APIs.

**Additional context**

The issue was pointed out in #1997, where the legacy CLib included custom exception handling that the generated CLib lacks.

ischoegl:

This PR aims to establish a more uniform approach to index lookups, where the current `npos` check (utilized for internal methods) is transitioned towards standard exception handling (suitable for user-facing APIs).

- Mimic the implementation of the `clone` flag in #1921 ... a `raise` flag is currently set to `false` and will switch to a `true` default in Cantera 3.2. This is implemented for various index lookups in `Phase`, `MultiPhase`, and `Kinetics`.
- Make more widespread use of `checkABCIndex` 
- Deprecate some additional legacy CLib helper functions
- Applies to indexing of `Element`, `Species`, `Reaction`, and `Component`.
- Immediately switch `Reactor::componentIndex`/`Reactor::speciesIndex` to throw exceptions: almost all uses are in user-facing APIs (and a transition for `ExtensibleXYZReactor` would be a breaking change).
- Make exception messages more consistent.

One major advantage of this approach is that exception handling at user-facing APIs will become significantly easier, e.g., in #1997 (the MATLAB API itself is still hand-coded, but other APIs, such as .NET, would require cumbersome workarounds).

Closes #2004

Historically, exception handling was separated into three functions, although the  `checkABC` was rarely used, and `checkABCArraySize`, which isn't directly related, only appeared in the CLib interface. For `Phase::speciesIndex`, this looks as follows:
```C++
//! Returns the index of a species named 'name' within the Phase object.
size_t speciesIndex(const string& name) const;

//! Check that the specified species index is in range.
bool checkSpeciesIndex(size_t k) const;

//! Check that an array size is at least nSpecies().
void checkSpeciesArraySize(size_t kk) const;
```

This PR proposes to integrate exception handling into `speciesIndex`. 
```C++
//! Returns the index of a species named 'name' within the Phase object.
size_t speciesIndex(const string& name, bool raise) const;
```
After Cantera 3.2, this PR anticipates a default argument of `bool raise=true`.

`checkSpeciesIndex` is still useful as an internal function (making it `protected` would be within the scope of this PR), and is updated to return the correct size as:
```C++
//! Check that the specified species index is in range.
size_t checkSpeciesIndex(size_t k) const;
```

Regarding `checkABCArraySize`, the generated CLib goes a different route, where size checking is enabled by the `uses` field of the YAML specification. This could be further streamlined if Cantera/enhancements#237 were to be implemented.

As an aside, the Python API widely uses a `ThermoPhase.species_index(int)` et al. to validate indices, which unnecessarily complicates the interface, but remains intact.

**Checklist**

### DIFF
--- a/include/cantera/equil/MultiPhase.h
+++ b/include/cantera/equil/MultiPhase.h
@@ -158,8 +158,11 @@ public:
     }
 
     //! Check that the specified species index is in range.
-    //! Throws an exception if k is greater than nSpecies()-1
-    void checkSpeciesIndex(size_t k) const;
+    /*!
+     * @since After %Cantera 3.2, returns verified species index.
+     * @exception Throws an exception if k is greater than nSpecies()-1
+     */
+    size_t checkSpeciesIndex(size_t k) const;
 
     //! Check that an array size is at least nSpecies().
     //! Throws an exception if kk is less than nSpecies(). Used before calls
--- a/include/cantera/equil/MultiPhase.h
+++ b/include/cantera/equil/MultiPhase.h
@@ -167,6 +167,7 @@ public:
     //! Check that an array size is at least nSpecies().
     //! Throws an exception if kk is less than nSpecies(). Used before calls
     //! which take an array pointer.
+    //! @deprecated To be removed after %Cantera 3.2. Only used by legacy CLib.
     void checkSpeciesArraySize(size_t kk) const;
 
     //! Name of species with global index @e kGlob
@@ -252,12 +253,16 @@ public:
     ThermoPhase& phase(size_t n);
 
     //! Check that the specified phase index is in range
-    //! Throws an exception if m is greater than nPhases()
-    void checkPhaseIndex(size_t m) const;
+    /*!
+     * @since After %Cantera 3.2, returns verified species index.
+     * @exception Throws an IndexError if m is greater than nPhases()-1
+     */
+    size_t checkPhaseIndex(size_t m) const;
 
     //! Check that an array size is at least nPhases()
     //! Throws an exception if mm is less than nPhases(). Used before calls
     //! which take an array pointer.
+    //! @deprecated To be removed after %Cantera 3.2. Unused
     void checkPhaseArraySize(size_t mm) const;
 
     //! Returns the moles of global species @c k. units = kmol
--- a/include/cantera/equil/MultiPhase.h
+++ b/include/cantera/equil/MultiPhase.h
@@ -222,12 +222,12 @@ public:
      * @param pName Name of the phase
      * @param raise  If `true`, raise exception if the specified phase is not found.
      * @returns the index. A value of -1 means the phase isn't in the object.
-     * @since Added the `raise` argument in %Cantera 3.2. If not specified, the default
-     *      behavior if a phase is not found in %Cantera 3.2 is to return -1.
-     *      After %Cantera 3.2, the default behavior will be to throw an exception.
+     * @since Added the `raise` argument in %Cantera 3.2 and changed return type. If
+     *      not specified, the default behavior if a phase is not found in %Cantera 3.2
+     *      is to return @ref npos.
      * @exception Throws a CanteraError if the phase is not found.
      */
-    int phaseIndex(const string& pName, bool raise) const;
+    size_t phaseIndex(const string& pName, bool raise) const;
 
     //! Return the number of moles in phase n.
     /*!
--- a/include/cantera/equil/MultiPhase.h
+++ b/include/cantera/equil/MultiPhase.h
@@ -117,7 +117,7 @@ public:
     //! Check that the specified element index is in range.
     /*!
      * @since After %Cantera 3.2, returns verified element index.
-     * @exception Throws an exception if m is greater than nElements()-1
+     * @exception Throws an IndexError if m is greater than nElements()-1
      */
     size_t checkElementIndex(size_t m) const;
 
@@ -148,7 +148,7 @@ public:
      * @since Added the `raise` argument in %Cantera 3.2. If not specified, the default
      *      behavior if an element is not found in %Cantera 3.2 is to return `npos`.
      *      After %Cantera 3.2, the default behavior will be to throw an exception.
-     * @exception Throws an IndexError.
+     * @exception Throws a CanteraError if the specified element is not found.
      */
     size_t elementIndex(const string& name, bool raise) const;
 
@@ -160,7 +160,7 @@ public:
     //! Check that the specified species index is in range.
     /*!
      * @since After %Cantera 3.2, returns verified species index.
-     * @exception Throws an exception if k is greater than nSpecies()-1
+     * @exception Throws an IndexError if k is greater than nSpecies()-1
      */
     size_t checkSpeciesIndex(size_t k) const;
 
@@ -212,8 +212,21 @@ public:
     /*!
      * @param pName Name of the phase
      * @returns the index. A value of -1 means the phase isn't in the object.
+     * @deprecated  To be removed after %Cantera 3.2. Use 2-parameter version instead.
      */
     int phaseIndex(const string& pName) const;
+
+    //! Returns the index, given the phase name
+    /*!
+     * @param pName Name of the phase
+     * @param raise  If `true`, raise exception if the specified phase is not found.
+     * @returns the index. A value of -1 means the phase isn't in the object.
+     * @since Added the `raise` argument in %Cantera 3.2. If not specified, the default
+     *      behavior if a phase is not found in %Cantera 3.2 is to return -1.
+     *      After %Cantera 3.2, the default behavior will be to throw an exception.
+     * @exception Throws a CanteraError if the phase is not found.
+     */
+    int phaseIndex(const string& pName, bool raise) const;
 
     //! Return the number of moles in phase n.
     /*!
--- a/include/cantera/equil/MultiPhase.h
+++ b/include/cantera/equil/MultiPhase.h
@@ -121,6 +121,7 @@ public:
     //! Check that an array size is at least nElements().
     //! Throws an exception if mm is less than nElements(). Used before calls
     //! which take an array pointer.
+    //! @deprecated To be removed after %Cantera 3.2. Only used by legacy CLib.
     void checkElementArraySize(size_t mm) const;
 
     //! Returns the name of the global element *m*.
@@ -134,6 +135,16 @@ public:
      * @param name   String name of the global element
      */
     size_t elementIndex(const string& name) const;
+
+    //! Returns the index of the element with name @e name.
+    /*!
+     * @param name   String name of the global element
+     * @since Added the `force` argument in %Cantera 3.2. If not specified, the default
+     *      behavior in %Cantera 3.2 is to return `npos` if an element is not found.
+     *      After %Cantera 3.2, the default behavior will be to throw an exception.
+     * @exception Throws an IndexError.
+     */
+    size_t elementIndex(const string& name, bool force) const;
 
     //! Number of species, summed over all phases.
     size_t nSpecies() const {
--- a/include/cantera/equil/MultiPhase.h
+++ b/include/cantera/equil/MultiPhase.h
@@ -115,8 +115,11 @@ public:
     }
 
     //! Check that the specified element index is in range.
-    //! Throws an exception if m is greater than nElements()-1
-    void checkElementIndex(size_t m) const;
+    /*!
+     * @since After %Cantera 3.2, returns verified element index.
+     * @exception Throws an exception if m is greater than nElements()-1
+     */
+    size_t checkElementIndex(size_t m) const;
 
     //! Check that an array size is at least nElements().
     //! Throws an exception if mm is less than nElements(). Used before calls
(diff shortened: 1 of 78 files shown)


include/cantera/equil/MultiPhase.h
```diff
@@ -192,9 +213,22 @@ class MultiPhase
...
+    /*!
+     * @param pName Name of the phase
+     * @param raise  If `true`, raise exception if the specified phase is not found.
+     * @returns the index. A value of -1 means the phase isn't in the object.
+     * @since Added the `raise` argument in %Cantera 3.2. If not specified, the default
+     *      behavior if a phase is not found in %Cantera 3.2 is to return -1.
+     *      After %Cantera 3.2, the default behavior will be to throw an exception.
+     * @exception Throws a CanteraError if the phase is not found.
+     */
+    int phaseIndex(const string& pName, bool raise) const;
```
ischoegl: This is an opportunity to address an outlier that does not use `size_t`.

interfaces/cython/cantera/thermo.pyx
```diff
@@ -463,17 +463,14 @@ cdef class ThermoPhase(_SolutionBase):
...
-            raise TypeError("'element' must be a string or a number."
-                            " Got {!r}.".format(element))
+            raise TypeError("'element' must be a string or a number. "
+                            f"Got {element!r}.")
 
-        if not 0 <= index < self.n_elements:
-            raise ValueError('No such element {!r}.'.format(element))
-
-        return index
+        raise ValueError(f"No such element {element!r}.")
```
speth: This `raise` statement is unreachable, isn't it?

interfaces/sourcegen/src/sourcegen/headers/ctthermo.yaml
```diff
@@ -46,7 +46,11 @@ recipes:
 - name: elementName  # Renamed in Cantera 3.2 (previously getElementName)
 - name: speciesName  # Renamed in Cantera 3.2 (previously getSpeciesName)
 - name: elementIndex
+  # alternate version to be removed after Cantera 3.2
+  wraps: elementIndex(const string&, bool)
 - name: speciesIndex
+  # alternate version to be removed after Cantera 3.2
+  wraps: speciesIndex(const string&, bool)
```
speth: This additional, required argument seems a little strange in the case of Clib, where you have to check whether the return value is valid no matter what.

interfaces/cython/cantera/mixture.pyx
```diff
@@ -98,19 +98,17 @@ cdef class Mixture:
         """Index of element with name 'element'.
 
             >>> mix.element_index('H')
-            2
         """
```
speth, line 102: The "2" here was the return value from the above example. Arguably, this example isn't particularly enlightening, but I think it should either stay or go as a whole.

interfaces/cython/cantera/mixture.pyx
```diff
@@ -98,19 +98,17 @@ cdef class Mixture:
...
-        if not 0 <= index < self.n_elements:
-            raise ValueError('No such element.')
+        if isinstance(element, (int, float)):
+            return self.mix.checkElementIndex(<int>element)
+        else:
+            raise TypeError("'element' must be a string or a number. "
+                            f"Got {element!r}.")
 
-        return index
+        raise ValueError(f"No such element {element!r}.")
```
speth: Likewise unreachable, I think.

include/cantera/equil/MultiPhase.h
```diff
@@ -115,12 +115,16 @@ class MultiPhase
     }
 
     //! Check that the specified element index is in range.
-    //! Throws an exception if m is greater than nElements()-1
-    void checkElementIndex(size_t m) const;
+    /*!
+     * @since After %Cantera 3.2, returns verified element index.
+     * @exception Throws an IndexError if m is greater than nElements()-1
```
speth:
```suggestion
     * @since Starting in %Cantera 3.2, returns the input element index, if valid.
     * @exception Throws an IndexError if m is greater than nElements()-1
```

src/thermo/Phase.cpp
```diff
@@ -126,16 +138,27 @@ size_t Phase::findSpeciesLower(const string& name) const
     return loc;
 }
 
-size_t Phase::speciesIndex(const string& nameStr) const
+size_t Phase::speciesIndex(const string& name) const
 {
-    size_t loc = npos;
+    warn_deprecated("Phase::speciesIndex", "'raise' argument not specified; "
+        "Default behavior will change from returning npos to throwing an exception "
+        "after Cantera 3.2.");
```
speth: I think we could reduce some of the churn required by this transition by issuing the deprecation warning only if a nonexistent element/species is requested. This way, any user code that never deals with unknown species/elements (or naively ignores the possibility) could make the transition without ever needing to be modified.

src/thermo/Phase.cpp
```diff
@@ -126,16 +138,27 @@ size_t Phase::findSpeciesLower(const string& name) const
...
+}
 
-    auto it = m_speciesIndices.find(nameStr);
+size_t Phase::speciesIndex(const string& name, bool raise) const
+{
+    size_t loc = npos;
+    auto it = m_speciesIndices.find(name);
     if (it != m_speciesIndices.end()) {
-        return it->second;
+        return checkSpeciesIndex(it->second);
```
speth: Is it possible for `m_speciesIndices` to contain an invalid index? I don't see how.

include/cantera/equil/MultiPhase.h
```diff
@@ -132,21 +136,38 @@ class MultiPhase
...
 
+    //! Returns the index of the element with name @e name.
+    /*!
+     * @param name  String name of the global element.
+     * @param raise  If `true`, raise exception if the specified element is not found;
+     *      otherwise, return @ref npos.
+     * @since Added the `raise` argument in %Cantera 3.2. If not specified, the default
+     *      behavior if an element is not found in %Cantera 3.2 is to return `npos`.
+     *      After %Cantera 3.2, the default behavior will be to throw an exception.
+     * @exception Throws a CanteraError if the specified element is not found.
```
speth:
```suggestion
     * @exception Throws a CanteraError if the specified element is not found and `raise` is `true`.
```

include/cantera/equil/MultiPhase.h
```diff
@@ -132,21 +136,38 @@ class MultiPhase
...
      * @param name   String name of the global element
+     * @deprecated  To be removed after %Cantera 3.2. Use 2-parameter version instead.
      */
     size_t elementIndex(const string& name) const;
 
+    //! Returns the index of the element with name @e name.
+    /*!
+     * @param name  String name of the global element.
+     * @param raise  If `true`, raise exception if the specified element is not found;
+     *      otherwise, return @ref npos.
```
speth: What is `@ref npos` as opposed to ``` `npos` ```, as used elsewhere?

include/cantera/kinetics/Kinetics.h
```diff
@@ -211,9 +223,32 @@ class Kinetics
...
+     * @since Added the `raise` argument in %Cantera 3.2. If not specified, the default
+     *      behavior if a phase is not found in %Cantera 3.2 is to return `npos`.
+     *      After %Cantera 3.2, the default behavior will be to throw an exception.
+     * @exception Throws a CanteraError if the specified phase is not defined.
+     */
+    size_t phaseIndex(const string& ph, bool raise) const {
         if (m_phaseindex.find(ph) == m_phaseindex.end()) {
+            if (raise) {
+                throw CanteraError("Kinetics::phaseIndex", "Phase '{}' not found", ph);
+            }
```
speth: I'd say this method definition is now long enough that it ought to be moved to the `.cpp` file.

include/cantera/kinetics/Kinetics.h
```diff
@@ -211,9 +223,32 @@ class Kinetics
...
     size_t phaseIndex(const string& ph) const {
+        warn_deprecated("Kinetics::phaseIndex", "'raise' argument not specified; "
+            "Default behavior will change from returning -1 to throwing an exception "
+            "after Cantera 3.2.");
+        return phaseIndex(ph, false);
+    }
+
+    /**
+     * Return the phase index of a phase in the list of phases defined within
+     * the object.
```
speth: I know this is just copy/pasted from before, but this is a good opportunity for improvement. Perhaps something along the lines of:
```suggestion
     * Return the index of a phase among the phases participating in this kinetic mechanism.
```

src/zeroD/ConstPressureMoleReactor.cpp
```diff
@@ -109,13 +109,14 @@ void ConstPressureMoleReactor::eval(double time, double* LHS, double* RHS)
...
+    if (nm == "enthalpy") {
         return 0;
-    } else {
-        return npos;
+    }
+    try {
+        return speciesIndex(nm) + m_sidx;
+    } catch (const CanteraError&) {
+        throw CanteraError("ConstPressureMoleReactor::componentIndex",
+            "Component '{}' not found", nm);
```
speth, line 119: I think it would be best to minimize use of this pattern, which could be done by extending the `raise` argument to `ReactorBase::speciesIndex`.

src/zeroD/FlowReactor.cpp
```diff
@@ -339,19 +339,23 @@ void FlowReactor::getConstraints(double* constraints) {
...
+    }
+    if (nm == "pressure") {
         return 2;
-    } else if (nm == "temperature") {
+    }
+    if (nm == "temperature") {
         return 3;
-    } else {
-        return npos;
+    }
```
speth, line 350: I don't see this as an improvement, even if it is equivalent.

speth (changes requested):
Thanks, @ischoegl. Overall, I think this is worthwhile, although there are a few rough edges that are worth smoothing out.

interfaces/sourcegen/src/sourcegen/headers/ctthermo.yaml
ischoegl: It is unfortunately required here, as I'd like to reach the new version (which doesn't have a default argument yet).

src/thermo/Phase.cpp
ischoegl: This is a great suggestion!

src/zeroD/ConstPressureMoleReactor.cpp
ischoegl: This is somewhat tricky. The `ExtensibleReactor` mechanism boxes us in here significantly, as a change of signatures is non-trivial (need to register a new delegated function type, thread everything through, and I'm not sure that default arguments work).

I parked another way I could think of in this commit https://github.com/ischoegl/cantera/commit/b11e2e965aec5501dafd8ae85e36474c93359c5b which, for inexplicable reasons, breaks an extensible reactor test also.

ischoegl:
@speth ... thanks for the review! The suggestion to condition the deprecation warning was great (https://github.com/Cantera/cantera/pull/2010#discussion_r2442687582). I rolled back much of the churn that would have been visible on the user-facing end.

For avoiding `try/catch` blocks, I am, however, facing some roadblocks that are related to `ExtensibleReactor`. Modifying the signature of `Reactor::speciesIndex` is a no-go, and some of the callback mechanism messes up things for the other workaround I could come up with (https://github.com/ischoegl/cantera/commit/b11e2e965aec5501dafd8ae85e36474c93359c5b). I ended up searching for uses of `->componentIndex(` and `.componentIndex(` in 0D `.h/.cpp` and it only appears to be used internally by `ReactorNet::globalComponentIndex`, which itself is user-facing. As the pattern appears to be limited to user-facing service methods, I'm suggesting leaving this as is.

As an aside, I added the `#pragma` warnings for the legacy CLib here: https://github.com/Cantera/cantera/pull/2010/commits/392ff534fc5a5b8f9226d28113c75a9e021b704d as this is (hopefully!) my last substantial PR before 3.2.

ischoegl:
@speth ... this is ready for a review. 

It does contain a (minor) breaking change for the MATLAB PR #1997, where all the indexing methods need to add the `raise=true` option so the API doesn't need to do the checking (which was part of the motivation here). The second argument can be dropped after Cantera 3.2, i.e., once `raise=true` becomes the default.

speth:
> The second argument can be dropped after Cantera 3.2, i.e., once `raise=true` becomes the default.

I'm a little confused by the plan here. For CLib, the second argument will always be mandatory, since C doesn't support functions with default arguments.

ischoegl:
> > The second argument can be dropped after Cantera 3.2, i.e., once `raise=true` becomes the default.
> 
> I'm a little confused by the plan here. For CLib, the second argument will always be mandatory, since C doesn't support functions with default arguments.

True, but in the generated CLib specification, you can force a shorter signature with the `wraps` field, which then uses default arguments. 😀

speth:
> True, but in the generated CLib specification, you can force a shorter signature with the `wraps` field, which then uses default arguments. 😀

So is the plan for CLib that for Cantera 3.2, you _must_ provide the second argument, and then for Cantera 3.3, you _must not_, with the behavior being to always indicate an exception for out-of-bounds indices? It might be better to temporarily use a "custom code" implementation for these methods, and avoid requiring downstream changes.

ischoegl:
> So is the plan for CLib that for Cantera 3.2, you _must_ provide the second argument, and then for Cantera 3.3, you _must not_, with the behavior being to always indicate an exception for out-of-bounds indices? It might be better to temporarily use a "custom code" implementation for these methods, and avoid requiring downstream changes.

I guess that's another way. My thought process was that updating MATLAB would be the easiest (the second argument wouldn't be exposed in the user-facing API). But we cannot rule out that someone will adopt CLib and be forced into a change. Easy enough to implement ... in which case this won't be a breaking change to #1997.

ischoegl:
Done. I did, however, realize that there is one more affected method, `Kinetics::kineticsSpeciesIndex`, that should be updated.

ischoegl:
@speth ... I had not anticipated using custom code for something like this, but it works quite well :smile:. I believe this should be ready to merge now.

src/zeroD/ConstPressureMoleReactor.cpp
speth: Fair enough. I'm OK with this approach. Thanks for investigating the alternatives.

speth (approved):
Thanks, @ischoegl. I think this is good to go.

PR #2014, opened 2025-10-21
Improve exceptions for abcName()

ischoegl:

This is a minor follow-up PR to #2010. I had noticed that exception handling for `*Name(size_t k)` was pretty inconsistent. In some instances, an exception was raised, in other cases, an `<unknown>` was returned, and in other cases, the index wasn't checked at all (i.e., undefined behavior). This PR updates the code so that it consistently throws an `IndexError` if the index is out of bounds.

The only instance where transitional behavior is implemented is `kineticsSpeciesName` (it's only user-facing, but some user code could check for `<unknown>`). For 0D and 1D instances of `componentName`, the implementation was inconsistent (some instances raised exceptions while others didn't), so I opted to raise exceptions for all, so all specializations follow the same approach.

**Checklist**

### DIFF
--- a/include/cantera/kinetics/Kinetics.h
+++ b/include/cantera/kinetics/Kinetics.h
@@ -307,14 +307,32 @@ public:
 
     //! Return the name of the kth species in the kinetics manager.
     /*!
-     *  k is an integer from 0 to ktot - 1, where ktot is the number of
+     * k is an integer from 0 to ktot - 1, where ktot is the number of
      * species in the kinetics manager, which is the sum of the number of
      * species in all phases participating in the kinetics manager. If k is
      * out of bounds, the string "<unknown>" is returned.
      *
      * @param k species index
+     * @todo Update docstring after %Cantera 3.2 to reflect that future versions
+     *      raise exception.
      */
     string kineticsSpeciesName(size_t k) const;
+
+    /**
+     * Return the name of the kth species in the kinetics manager.
+     * k is an integer from 0 to ktot - 1, where ktot is the number of
+     * species in the kinetics manager, which is the sum of the number of
+     * species in all phases participating in the kinetics manager.
+     *
+     * @param k species index
+     * @param raise  If `true`, raise exception if the species index is out of bounds.
+     * @since New in %Cantera 3.2.
+     * @deprecated  To be removed after %Cantera 3.2. Only used for transitional
+     *      behavior.
+     * @exception Throws an IndexError if the specified species index is out of bounds
+     *      and `raise` is `true`.
+     */
+    string kineticsSpeciesName(size_t k, bool raise) const;
 
     /**
      * This routine will look up a species number based on the input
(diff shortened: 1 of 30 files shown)


ischoegl:
@speth ... this should be fairly straightforward.

include/cantera/kinetics/Kinetics.h
```diff
@@ -305,17 +305,35 @@ class Kinetics
         return m_start[n] + k;
     }
 
-    //! Return the name of the kth species in the kinetics manager.
-    /*!
-     *  k is an integer from 0 to ktot - 1, where ktot is the number of
+    /** Return the name of the kth species in the kinetics manager.
+     *
+     * k is an integer from 0 to ktot - 1, where ktot is the number of
```
speth: Starting the `brief` doc on the first line of the multi-line comment block is not a style that I think we've used before for Doxygen comments.

src/thermo/Phase.cpp
```diff
@@ -51,7 +51,10 @@ void Phase::checkElementArraySize(size_t mm) const
 
 string Phase::elementName(size_t m) const
 {
-    return m_elementNames[checkElementIndex(m)];
+    if (m < m_mm) {
+        return m_elementNames[m];
+    }
+    throw IndexError("Phase::elementName", "element", m, m_mm);
```
speth, line 57: The previous approach using `checkElementIndex` achieved the same thing, didn't it?

test/kinetics/kineticsFromScratch.cpp
```diff
@@ -327,6 +327,9 @@ TEST_F(KineticsFromScratch, third_body_composition)
     Composition prod = R->products;
     EXPECT_EQ(prod.count("H2O"), (size_t) 0);
     EXPECT_EQ(reac.count("M"), (size_t) 0);
+
+    ASSERT_THROW(kin.kineticsSpeciesName(200), CanteraError);
```
speth: You could be more specific on the exception type, here and in a couple other places.
```suggestion
    ASSERT_THROW(kin.kineticsSpeciesName(200), IndexError);
```

speth (changes requested):
Thanks, @ischoegl. The previous behavior on a couple of these was definitely odd. I have just a couple of minor comments.

src/thermo/Phase.cpp
ischoegl: It is equivalent, but I thought the `Phase::elementName` labeling is nicer for this specific case. Fwiw, it's parallel to the situation with `Phase::speciesName`. Technically, the functions only share the `if (m < m_mm) {` line ... other than the trivial ‘}‘

speth: Yes, of course the same comment would apply to `speciesName`. I guess I just think it's odd to change this after you just introduced the return value for `checkElementIndex` and friends which enabled this when this is one of the few places that function is used in that way. It's a minor point, though, and obviously not something to hold up this PR over.

speth (approved):

src/thermo/Phase.cpp
ischoegl: Fair point. Was facing an internal tug of war between ‚dry‘ and the most intuitive exception message. At the same time, it’s a really inconsequential choice.

PR #2016, opened 2025-10-23
Fix surface density/concentration calculation for multi-site species

### LINKED ISSUE #1996: plug flow reactor: silicon nitride deposition example: sdot and pressure increase issues
**Problem description**

Plug flow reactor example has some fundamental issues. The deviation from the report is not due to the friction. In the CT solution, pressure is building up in the flow direction whereas velocity is increasing. In this example, pressure should not increase. Please see the attached CK solution.

https://www.cantera.org/stable/examples/python/reactors/1D_pfr_surfchem.html#sphx-glr-examples-python-reactors-1d-pfr-surfchem-py

**Steps to reproduce**

Noticed that H2NFSINH(S) and HN(FSINH)2(S) coverages are calculated almost twice lower than the reference. I have a jupyter-notebook file, in which hand calculations of the net production rates are solved using scikit-sundae CVODE (I also recommend this module to replace with that of scikits which is used in some CT examples). 
1. Open 'heterogeneous_irrev__rxn.ipynb'
2. Run 
3. See the steady state compositions are different.

**Behavior**
In 1D_bed packed example, sdot calculations are consistent with my custom, low level irreversible reaction calculations (verified with CK). However, in this case, something is inconsistent. Once sdot issue is resolved, maybe pressure problem will be solved, as well.

**System information**

- Cantera version: the latest stable
- OS: Windows 11, Ubuntu 24
- Python/MATLAB/other software versions: Python

[heterogeneous_irrev__rxn.ipynb](https://github.com/user-attachments/files/22710561/heterogeneous_irrev__rxn.ipynb)

Image

**Attachments**

**Additional context**

speth:

- Revise calculation of density for surface phases to correctly account for species that occupy multiple sites. This affects the concentrations (calculated in `Phase` as $C_k = \rho Y_k / W_k$) which in turn affects reaction rate calculations.

This resolves a regression introduced in #1350 that affects Cantera 3.0 and 3.1. The effect is most noticeable for phases where major species occupy multiple sites, such as in `SiF4_NH3_mec.yaml` where all the species have multiple occupancy.

- Partially resolves #1996

**Checklist**

### DIFF
--- a/include/cantera/thermo/SurfPhase.h
+++ b/include/cantera/thermo/SurfPhase.h
@@ -27,6 +27,22 @@ namespace Cantera
  * The density of surface sites is given by the variable @f$ n_0 @f$,
  * which has SI units of kmol m-2.
  *
+ * While the site coverage fractions @f$ \theta_k @f$ are generally used to describe the
+ * composition of surface phases, %Cantera represents the state internally in terms of
+ * mass fractions for consistency with other phase models. Mole fractions are computed
+ * from coverages as:
+ * @f[
+ *     X_k = \frac{ \theta_k / s_k }{ \sum_i \theta_i / s_i }
+ * @f]
+ * where @f$ s_k @f$ is the number of sites occupied by species @f$ k @f$. Mass
+ * fractions and the mean molecular weight are then computed in the typical manner.
+ *
+ * The mass density of a surface phase is defined to have units of kg/m². It is computed
+ * as:
+ * @f[
+ *     \rho = \frac{ n_0 \overline{W} }{ \sum X_k s_k }
+ * @f]
+ *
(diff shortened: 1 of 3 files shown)


ischoegl:
Mentioning https://github.com/Cantera/enhancements/issues/154, although I am unsure whether this is a tangent.

speth:
I don't think that enhancement idea is relevant to the issue here. For phases, the state needs to be an intensive property, so I think the composition always needs to be in terms of relative amounts of one sort or another. The challenge here comes from the fact that there are three different conventions available for expressing this in the case of surface phases.

ischoegl (approved):
Thanks, @speth. Based on my reading of `test_multi_site_density`, this looks good to me!

PR #2019, opened 2025-10-26
Fix incorrect type annotations

speth:

Fix some of the new Python type annotations based on (incorrect) errors indicated in the examples and test suite.

**Checklist**


interfaces/cython/cantera/_types.py
```diff
@@ -48,10 +49,12 @@
     "SVX",
     "SVY",
 ]
-CompositionLike: TypeAlias = str | dict[str, float] | ArrayLike
-StateSetter: TypeAlias = tuple[float, float, CompositionLike]
+CompositionLike: TypeAlias = str | Mapping[str, float | int] | ArrayLike
+State2Setter: TypeAlias = tuple[float | None, float | None]
```
TimothyEDawson, line 53: I honestly never knew this was an option... You can do stuff like `gas.TP = None, 101325.0`? What does that do, keep the current temperature as-is? (EDIT: Confirmed that's exactly what it does, TIL!) And if so, why is there no `T.setter` and `P.setter`?

interfaces/cython/cantera/transport.pyi
```diff
@@ -1,7 +1,7 @@
 # This file is part of Cantera. See License.txt in the top-level directory or
 # at https://cantera.org/license.txt for license and copyright information.
 
-from typing import Literal, TypeAlias, TypedDict
+from typing import Literal, Never, TypeAlias, TypedDict
```
TimothyEDawson: Need to import `Never` from `typing_extensions` instead, to maintain support for Python 3.10. I have some Ruff settings I use which highlight this sort of issue. EDIT: Actually, it was mypy in this case.

interfaces/cython/cantera/_types.py
speth: Yes, the setters allowing `None` are a convenient way of changing the state while holding one other property constant. The reason there is no setter for T or P (or H or V or ...) alone is because it's not clear which other thermodynamic property should be held constant.

TimothyEDawson: Of course! That make a lot of sense. Is this feature documented somewhere? It's very convenient, and I'm shocked I never knew about it.

speth: It's mentioned in the [Python Tutorial](https://cantera.org/stable/userguide/python-tutorial.html#setting-the-state), but perhaps there are a few other places it should be highlighted.

interfaces/cython/cantera/transport.pyi
speth: Thanks for the pointer. I was glad to see that the error message in the CI job also suggested this change.

TimothyEDawson: Of course! I'm hoping to write up something for the dev guide this week, which will definitely include a recommended way to run the type checking locally so you can be pretty sure the CI won't fail.

For now, you can try these mypy and stubtest commands from `main.yaml`:

```bash
mypy --config-file interfaces/cython/pyproject.toml -p cantera
stubtest  --mypy-config-file interfaces/cython/pyproject.toml --ignore-disjoint-bases --allowlist interfaces/cython/.mypyignore cantera
```

Noting that they are both run post-install.

speth:
`stubtest` is very unhappy with my updates 😢 

```
cantera.ChebyshevRate.__init__ is inconsistent, runtime does not have parameter "temperature_range". You may need to write stubs for __new__ instead of __init__.
cantera.ChebyshevRate.__init__ is inconsistent, runtime does not have parameter "pressure_range". You may need to write stubs for __new__ instead of __init__.
cantera.ChebyshevRate.__init__ is inconsistent, runtime does not have parameter "data". You may need to write stubs for __new__ instead of __init__.
cantera.ChebyshevRate.__init__ is inconsistent, runtime does not have parameter "input_data". You may need to write stubs for __new__ instead of __init__.
cantera.ChebyshevRate.__init__ is inconsistent, runtime does not have parameter "init". You may need to write stubs for __new__ instead of __init__.
cantera.CustomRate.__init__ is inconsistent, runtime does not have parameter "k". You may need to write stubs for __new__ instead of __init__.
cantera.CustomRate.__init__ is inconsistent, runtime does not have parameter "init". You may need to write stubs for __new__ instead of __init__.
cantera.UnitSystem.__init__ is inconsistent, runtime does not have parameter "units". You may need to write stubs for __new__ instead of __init__.
cantera.Units.__init__ is inconsistent, runtime does not have parameter "name". You may need to write stubs for __new__ instead of __init__.
cantera.reaction.ChebyshevRate.__init__ is inconsistent, runtime does not have parameter "temperature_range". You may need to write stubs for __new__ instead of __init__.
cantera.reaction.ChebyshevRate.__init__ is inconsistent, runtime does not have parameter "pressure_range". You may need to write stubs for __new__ instead of __init__.
cantera.reaction.ChebyshevRate.__init__ is inconsistent, runtime does not have parameter "data". You may need to write stubs for __new__ instead of __init__.
cantera.reaction.ChebyshevRate.__init__ is inconsistent, runtime does not have parameter "input_data". You may need to write stubs for __new__ instead of __init__.
cantera.reaction.ChebyshevRate.__init__ is inconsistent, runtime does not have parameter "init". You may need to write stubs for __new__ instead of __init__.
cantera.reaction.CustomRate.__init__ is inconsistent, runtime does not have parameter "k". You may need to write stubs for __new__ instead of __init__.
cantera.reaction.CustomRate.__init__ is inconsistent, runtime does not have parameter "init". You may need to write stubs for __new__ instead of __init__.
cantera.units.UnitSystem.__init__ is inconsistent, runtime does not have parameter "units". You may need to write stubs for __new__ instead of __init__.
cantera.units.Units.__init__ is inconsistent, runtime does not have parameter "name". You may need to write stubs for __new__ instead of __init__.
```

interfaces/cython/cantera/reaction.pyi
```diff
@@ -228,6 +235,14 @@ class PlogRate(ReactionRate):
...
 
 class ChebyshevRate(ReactionRate):
+    def __init__(
+        self,
+        temperature_range: Sequence[float] | None = None,
+        pressure_range: Sequence[float] | None = None,
+        data: ArrayLike | None = None,
+        input_data: _ChebyshevRateInput | None = None,
+        init: bool = True,
+    ) -> None: ...
```
TimothyEDawson, line 245: I'm seeing this is causing an error in the CI:

```bash
cantera.ChebyshevRate.__init__ is inconsistent, runtime does not have parameter "temperature_range". You may need to write stubs for __new__ instead of __init__.
```

This is the same issue I've had with most `__cinit__` methods, as their function signature is not passed on to `__init__` on the Python side. The way I've been handling it so far is to simply add the specific `__init__` to the `interfaces/cython/.mypyignore` file. Make sure to use the `--generate-allowlist` option to get the right formatting (it will likely appear twice - once in `cantera.ChebyshevRate.__init__` and once in `cantera.reaction.ChebyshevRate.__init__`).

It's not ideal, but seems to be a limitation of Cython.

TimothyEDawson:
Just left a comment about how to fix it on the `ChebyshevRate.__init__`. It also makes me realize that the `--generate-allowlist` command in the CI should print the allowlist so you can easily grab it from the failed CI job.

interfaces/cython/cantera/transport.pyi
ischoegl: Good to hear. A dev guide for running things locally will be greatly appreciated!

speth:
> Just left a comment about how to fix it on the `ChebyshevRate.__init__`. It also makes me realize that the `--generate-allowlist` command in the CI should print the allowlist so you can easily grab it from the failed CI job.

Ah, but there's a bit of a catch-22 with this at present: If the either of the previous `stubtest` commands fail, the one to generate the allow list is never run (the job runs with `bash -e` which exits after any failing command).

TimothyEDawson:
> Ah, but there's a bit of a catch-22 with this at present: If the either of the previous `stubtest` commands fail, the one to generate the allow list is never run (the job runs with `bash -e` which exits after any failing command).

Right, it should probably either be the first command in the three, or better yet, its own separate step. Maybe it should also run without the current allowlist, then print the diff between the result and the current `.mypyignore` file and/or upload that as a patch file artifact. I just hesitate to overcomplicate things.

ischoegl (approved):
Thanks! I’ll go head and approve.

Issue #2030, opened 2025-11-01
Issue with ctLoad on MacOS


santoshshanbhogue:
**Problem description**
The experimental MATLAB toolbox has a few fixable bugs, that as of now lead to Cantera being installed incorrectly. The fix on my machine is to run the following commands before ctLoad

clibConfiguration('ctMatlab','ExecutionMode','inprocess')

(ctBuildInterface claims to do this, but it ends up not being the case)

**Steps to reproduce**

Run the following commands in the MATLAB editor:

ctSourceRoot = '/Users/santosh1/packages/canteraSource';
ctInstalledDir = '/Users/santosh1/packages/ct_dev_version/';

ctToolboxDir = fullfile(ctSourceRoot, "interfaces/matlab_experimental");

ctCLibIncludeDir = fullfile(ctInstalledDir,"include");
ctLibDir = fullfile(ctInstalledDir, "lib");

clc;
ctBuildInterface(ctToolboxDir, ctCLibIncludeDir, ctLibDir);

Paths used in the build process:
Cantera Toolbox Root: /Users/santosh1/packages/canteraSource/interfaces/matlab_experimental
Include Directory: /Users/santosh1/packages/ct_dev_version/include
Cantera Library: /Users/santosh1/packages/ct_dev_version/lib
Output Folder: /Users/santosh1/packages/canteraSource/interfaces/matlab_experimental/cantera
MEX configured to use 'Xcode Clang++' for C++ language compilation.
C++ compiler set to 'Xcode Clang++'.
Definition file definectMatlab.m contains definitions for 428 constructs supported by MATLAB.
- 324 constructs are fully defined.
- 104 constructs partially defined and commented out.

To include the 104 undefined constructs in the interface, uncomment and complete the definitions in definectMatlab.m.
To build the interface, call build(definectMatlab).
Building interface file 'ctMatlabInterface.dylib' for clib interface 'ctMatlab'.
Build Log:
[...]

PR #2031, opened 2025-11-02
Matlab papercuts

### LINKED ISSUE #2006: MATLAB toolbox changes CLib signatures
**Problem description**

The MATLAB toolbox `ctBuildInterface` step changes function signatures while retaining CLib function names (this is a translation step in `ctGenerateLibraryDefinitions`). While convenient, this behavior is highly confusing for code maintenance, and is problematic from a future code generation viewpoint (see Cantera/enhancements#39).

This behavior was uncovered while debugging #1997 and was introduced with #1868.

ischoegl:

- Add prefix `m` to functions wrapped by MATLAB to disambiguate from CLib (most of the churn) .. see #2006
- Improve formatting of MATLAB samples ... MATLAB block `%%` creates `h1` level header in `sphinx-gallery` (also a lot of churn)
- Fix `flame1.m`, which erroneously used a strained flow domain (!)
- Remove redundant 1D samples and make names consistent with Python
- Improve handling of MATLAB execution mode (`inprocess` vs `outofprocess`)
- Consolidate `ctBuildInterface` into a single file
- Implement `Inlet1D::updateState` to ensure an inlet can be fully set up prior to `Sim1D` assembly

Closes #2006, partially addresses #2030

**Checklist**


include/cantera/oneD/Boundary1D.h
```diff
@@ -189,6 +191,7 @@ class Inlet1D : public Boundary1D
     Flow1D* m_flow = nullptr; //!< the adjacent flow domain
 };
 
+
 /**
  * A terminator that does nothing.
```
speth: I don't understand these unnecessary whitespace changes.

ischoegl:
While this is ready, it should probably be merged/rebased after adoption of #1974.

src/oneD/Boundary1D.cpp
```diff
@@ -127,23 +132,34 @@ void Inlet1D::show(const double* x)
...
-        m_flow->phase().setMoleFractionsByName(xin);
-        m_flow->phase().getMassFractions(m_yin.data());
+    if (m_solution) {
+        auto thermo = m_solution->thermo();
+        thermo->setMoleFractionsByName(xin);
+        thermo->getMassFractions(m_yin.data());
         needJacUpdate();
+    } else {
+        m_xstr = xin;
     }
```
speth, line 142: I think I understand the direction this is going -- I imagine by Cantera 3.3 we'll be requiring the `Solution` object to initialize the `Domain1D` object, so this will always be available. But I'm a little confused at what in the Matlab interface made this necessary to tackle now and could use a little clarification.

speth (approved):
Thanks, @ischoegl. No significant comments here, just one question on the `Boundary1D` changes.

src/oneD/Boundary1D.cpp
ischoegl: You're correct that in Cantera 3.3, we'll always have the `Solution` and will have this ability, i.e., the `if` clause can be removed.

The reason I included it here is that I really disliked the interface for the `CounterFlowDiffusionFlame` object (the only Python-FlameBase-like composite object that is currently implemented). It required the phases to be passed separately from the boundary objects, and the phases had to be separate copies of the same mechanism to work correctly. There wasn't a clean way to simplify this in MATLAB, so I decided to add this capability in C++. As you see in the `flamespeed.cpp` sample, this means that you no longer have to reset some of the states before starting to solve.

PR #2037, opened 2025-11-07
Matlab housekeeping and namespace adoption

### LINKED ISSUE #240: Use MATLAB namespaces in new toolbox
**Abstract**

MATLAB does support namespace folders, see https://www.mathworks.com/help/matlab/matlab_oop/namespaces.html.

Having the namespace in place would create an interface that:
- avoids the `ct` prefix for various service functions 
- collects constants in the namespace
- reflects namespaces used by other user-facing APIs

Tagging @rwest and @ssun30. It would presumably be fairly easy to implement.

speth:

While there are no other Matlab-related PRs open, I thought it would be a good time for a bit of housekeeping

- Remove the `_experimental` suffix from the Matlab toolbox folders (though of course the interface itself is still marked as experimental and subject to change)
- Rename some classes for consistency with the Python and/or C++ interfaces
- Rename the reference to the current object in Matlab classes as `obj` instead of using various inconsistent names.
- Reorganize the Matlab toolbox so that all classes / functions are in the `ct` namespace, with the reactor network and 1D classes in child namespaces `ct.zeroD` and `ct.oneD`, respectively.
  - Closes Cantera/enhancements#240

I'm finding that the namespace support in Matlab is a little clunky. It is quite unintuitive that you still need to refer to the qualified names even when they are used from within the namespaced code. For example, in the implementation for `ReactorBase`, the `arguments` block for its methods must refer to `ct.ReactorBase`:
```matlab
        function set.chemistryEnabled(obj, flag)
            arguments
                obj (1,1) ct.ReactorBase
                flag (1,1) logical
            end
```
However, the external usage is fine, and looks about the same as the Python module where you mainly just have object constructors and occasional constants like `OneAtm` prefixed with `ct.`.

**Checklist**


ischoegl:
@speth ... thanks for your work on this. While this is in preview, I wanted to briefly point out namespaces, see Cantera/enhancements#240. If we want to go that route, now would be the time (albeit not necessarily in this PR).

I.e., rather than
```
>> gas = Solution('gri30.yaml');
```
you'd write
```
>> gas = ct.Solution('gri30.yaml');  % or: cantera.Solution('gri30.yaml')
```
This would be especially useful for the various constants that Cantera defines.

interfaces/matlab/Constants/+ct/Avogadro.m
ischoegl, line 1: Rather than having these trivial files (most of which will be rarely used), here's another approach using a class:
https://www.mathworks.com/matlabcentral/answers/358826-best-method-to-define-constants-used-by-several-functions#comment_488420 (although I am not convinced).

speth: I think what we have is fine, and keeps the end use as simple as possible. The question there seems to be more about passing sets of problem-specific constants into a function, rather than defining true global constants.

speth:
Got a strange CI error on `macos-14`:
```
================================================================================
  Verification failed in ctTestPureFluid/testQualityExceptions.
      ---------------------
      Framework Diagnostic:
      ---------------------
      verifySubstring failed.
      --> The value does not contain the substring.
      
      Actual char:
          <no Cantera error>                                                                                                                                                                                                                    
      Expected Substring:
          Invalid vapor fraction
      ------------------
      Stack Information:
      ------------------
      In /Users/runner/work/cantera/cantera/test/matlab/ctTestPureFluid.m (ctTestPureFluid.testQualityExceptions) at 238
  ================================================================================
```
This only happened on one runner, and I've been unable to replicate it locally. I suspect it will go away on a re-run, but documenting here in case this shows up again.

ischoegl (commented):
I’ve seen those random failures before. They’re not repeatable and appear to be linked to exception handling going awry. I’ve not seen this on the googletest suite.

.github/workflows/main.yml
```diff
@@ -320,11 +320,11 @@ jobs:
       with:
         command: |
           ctDir = getenv('CANTERA_ROOT');
-          ctToolboxDir = fullfile(ctDir, 'interfaces', 'matlab_experimental');
+          ctToolboxDir = fullfile(ctDir, 'interfaces', 'matlab');
           ctIncludeDir = fullfile(ctDir, 'interfaces', 'clib', 'include');
           ctLibDir = fullfile(ctDir, 'build', 'lib');
           addpath(genpath(ctToolboxDir));
-          ctBuildInterface(ctToolboxDir, ctIncludeDir, ctLibDir);
+          ct.buildInterface(ctToolboxDir, ctIncludeDir, ctLibDir);
```
ischoegl, line 327:
```suggestion
          ct.impl.buildInterface(ctToolboxDir, ctIncludeDir, ctLibDir);
```
I think this should go into the 'implementation' folder to avoid confusion. As an aside, I'd also prefer `clib` to `impl` for the sub-namespace.

.github/workflows/main.yml
```diff
@@ -1304,12 +1304,12 @@ jobs:
         command: |
           disp("MATLAB version: " + version);
           ctDir = getenv('CANTERA_ROOT');
-          ctToolboxDir = fullfile(ctDir, 'interfaces', 'matlab_experimental');
+          ctToolboxDir = fullfile(ctDir, 'interfaces', 'matlab');
           ctIncludeDir = fullfile(ctDir, 'interfaces', 'clib', 'include');
           ctLibDir = fullfile(ctDir, 'build', 'lib');
           addpath(genpath(ctToolboxDir));
-          ctBuildInterface(ctToolboxDir, ctIncludeDir, ctLibDir);
+          ct.buildInterface(ctToolboxDir, ctIncludeDir, ctLibDir);
```
ischoegl, line 1311: See above.

doc/sphinx/matlab/onedim.rst
```diff
@@ -1,5 +1,5 @@
 .. default-domain:: mat
-.. currentmodule:: OneDim
+.. currentmodule:: OneDim.+ct
```
ischoegl:
```suggestion
.. currentmodule:: +ct.+oneD
```
Going with namespaces, I want to suggest to introduce a sub namespace `oneD` (i.e. following C++ folder nomenclature).

Members would then be accessed as
```
>>> ct.oneD.CounterFlowDiffusionFlame(left, flow, right)
```

doc/sphinx/matlab/onedim.rst
```diff
@@ -35,29 +35,29 @@ AxisymmetricFlow
 Boundaries
 ----------
 
-Inlet
-^^^^^
-.. autoclass:: Inlet(phase[, name])
+Inlet1D
```
ischoegl: The sub namespace would make this `oneD.Inlet`, which imho is clearer. I'm :+1: with minor inconsistencies: you may recall that I've advocated for modularization in Python, but I agree that this would cause too much churn there. In the case of MATLAB, however, we're starting with an (almost) blank slate.

doc/sphinx/matlab/zerodim.rst
```diff
@@ -1,5 +1,5 @@
 .. default-domain:: mat
-.. currentmodule:: Reactor
+.. currentmodule:: Reactor.+ct
```
ischoegl:
```suggestion
.. currentmodule:: +ct.+zeroD
```
Same underlying idea as for `ct.oneD`.

interfaces/matlab/Utility/+ct/buildInterface.m
```diff
@@ -1,17 +1,17 @@
-function ctBuildInterface(ctToolboxDir, ctIncludeDir, ctLibDir, cleanup)
+function buildInterface(ctToolboxDir, ctIncludeDir, ctLibDir, cleanup)
     % Build the Cantera MATLAB interface. ::
     %
-    %   >> ctBuildInterface(ctToolboxDir, includeDir, ctLibDir)
+    %   >> buildInterface(ctToolboxDir, includeDir, ctLibDir)
```
ischoegl, line 4:
```suggestion
    %   >> ct.impl.buildInterface(ctToolboxDir, includeDir, ctLibDir)
```
I think this should be moved into the `impl` sub namespace (a folder that most users will understand holds the implementation).

interfaces/matlab/+ct/+impl/call.m
```diff
@@ -1,13 +1,13 @@
-function output = ctFunc(funcName, varargin)
+function output = call(funcName, varargin)
```
ischoegl, line 1: I really like what the namespaces accomplish here. However, I'd suggest to rename so we get the very intuitive `ct.clib.call` rather than `ct.impl.call`.

interfaces/matlab/Utility/+ct/buildInterface.m
```diff
@@ -112,10 +112,10 @@ function ctGenerateLibraryDefinitions(includeDir, ctLibDir, outputDir)
         "Verbose", true);
 end
 
-function ctEditLibraryDefinitions(fileDir)
+function editLibraryDefinitions(fileDir)
     % Post-process auto-generated library definition file.
     %
-    % `ctEditLibraryDefinitions` edits the autogenerated `definectMatlab.m` file
+    % `editLibraryDefinitions` edits the autogenerated `definectMatlab.m` file
```
ischoegl, line 120: Ideally, the `definectMatlab.m` function and associated compiled library should be generated in `+ct/+impl` (or, preferably, in `+ct/+clib`). However, there may be restrictions in how MATLAB handles this with `clibgen`.

interfaces/matlab/+ct/+impl/call.m
```diff
@@ -1,13 +1,13 @@
...
     % Calls Cantera library functions with single outputs and returns
     % errors if necessary.
 
     output = clib.ctMatlab.(funcName)(varargin{:});
```
ischoegl, line 5: The `clib` namespace appears to be reserved, but there is no clash with `ct.clib` - I tried on my macOS machine and it works just fine.

interfaces/matlab/README.md
```diff
@@ -49,10 +49,10 @@ Cantera objects and functions.
    Run the following commands in MATLAB:
 
    ```matlab
-   ctToolboxDir = '/path/to/cantera/source/interfaces/matlab_experimental';
+   ctToolboxDir = '/path/to/cantera/source/interfaces/matlab';
    ctCLibIncludeDir = '/path/to/cantera/include';
    ctLibDir = '/path/to/cantera/library';
-   ctBuildInterface(ctToolboxDir, ctCLibIncludeDir, ctLibDir);
+   ct.buildInterface(ctToolboxDir, ctCLibIncludeDir, ctLibDir);
```
ischoegl, line 55: As suggested elsewhere, I'd prefer moving this to the `+impl` sub namespace.

interfaces/matlab/README.md
```diff
@@ -63,7 +63,7 @@ Cantera objects and functions.
    After building, the compiled interface should appear under:
 
    ```
-   /path/to/cantera/interfaces/matlab_experimental/cantera/ctMatlab
+   /path/to/cantera/interfaces/matlab/cantera/ctMatlab
```
ischoegl, line 66: Ideally, this should be
```suggestion
   /path/to/cantera/interfaces/matlab/cantera/+ct/+clib/ctMatlab
```
or similar, although there may be some restrictions due to MATLAB's way of resolving `clib.ctMatlab` (a second entry on the MATLAB path may be unavoidable).

ischoegl (commented):
Thanks, @speth! Overall, I really like the direction this took and am, unsurprisingly, glad to see cantera/enhancements#240 adopted.

I have some additional suggestions. Rather than having `+ct` within subfolders, I'd suggest introducing additional sub namespaces, which would enable a direct mapping of namespaces to folders:
* Introduce `+zeroD` and `+oneD` to improve context (while avoiding clutter)
* Rename `+impl` to `+clib`
* Potentially introduce `+constants`, and/or `+pureFluid`
* Get rid of all other subfolders (which avoids clutter on the MATLAB path - currently 9! folders)

Finally, I agree that we have widely used `ct` as a prefix elsewhere. However, it may make sense to use the same official (un-aliased) package name as in Python, which is `cantera`. It will result in a little longer names, but being more descriptive is very likely beneficial for the MATLAB File Exchange (i.e., needs to be descriptive and avoid ambiguity). I also don't mind the additional 'branding' :smile:

```
matlab
 └─ +cantera
     ├─ +clib
     ├─ +zeroD
     └─ +oneD
```

PS: Unfortunately, MATLAB doesn't support aliases, but this is possible
```
import cantera.Solution
```
after which `Solution` can be used by itself. To import everything, you'd use
```
import cantera.*
```

PPS: Here's an illustration of MATLAB path clutter. I feel that it should really just be a single folder for the entire toolbox, i.e., in my case: `/Volumes/Data/work/GitHub/cantera/interfaces/matlab`. A second path entry for `ctMatlab` may be necessary, though.

image

.github/workflows/main.yml
bryanwweber: I prefer `impl` because the fact that it's a C-library is somewhat irrelevant... we could (hypothetically) replace it with C++ or Python and leave the `impl` namespace to avoid breaking user code.

ischoegl: > I prefer `impl` because the fact that it's a C-library is somewhat irrelevant... we could (hypothetically) replace it with C++ or Python and leave the `impl` namespace to avoid breaking user code.

Interesting thought. While I don't anticipate a move away from CLib, the folder will likely remain untouched by users (it's pretty much what's 'under the hood'), so breaking user code is largely hypothetical. I still dislike `impl`, though. What about `core`?

speth: The purpose of the nested `impl` namespace is as a place to hide the implementation details -- the user should never call any of these functions directly. I read some suggestions that having a `private` folder within a namespace provided a way of doing this, but I couldn't get that to work so ended up with this instead.

`buildInterface`, however, is not a private implementation detail, as it needs to be called at some point (even if only once ever).

ischoegl: I’m totally on-board with ‘hiding’ these details, but don’t think that this needs to be considered private. I’m fine with users calling in directly (caveat emptor!). Regarding the right space for `buildInterface` my main thought was to hide that in the same place where the rest of the ‘under the hood’ functions live.

bryanwweber: `impl` or `core` seems fine to me. As @speth said, let's hide the implementation details, including the name.

interfaces/matlab/README.md
speth: I agree that moving this into the `+ct/+impl` namespace would be nice. I spent a little bit of time trying to get this to work but couldn't figure out how to do so without adding more magic to `editLibraryDefinitions`, and even then I'm not sure it would work. So I think I'd prefer to leave this for now.

speth: Actually, never mind. This was simpler than I thought.

speth:
This PR has been refactored to implement nested namespaces for `zeroD` and `oneD`, with everything else remaining in `ct`. I'd prefer to leave the internal implementation in the `impl` namespace, and the `buildInterface` function in the `ct` namespace at least for now.

interfaces/matlab/Utility/+ct/buildInterface.m
speth: the `definectMatlab.m` function typically gets deleted by the build process, so at least this is moot.

ischoegl (approved):
Thanks, @speth! I just ran it on my machine and it works beautifully.

Issue #2034, opened 2025-11-04
MATLAB's outofprocess mode has massive overhead


ischoegl:
**Problem description**

While the `outofprocess` mode introduced in #1868 resolves #1852, it comes at the cost of massive overhead due to repeated allocation/copying of arrays, see https://github.com/Cantera/cantera/pull/1997#issuecomment-3440097044. The migration to the generated CLib in #1997 may resolve crashes reported in #1852 when using `LD_PRELOAD`.

**Behavior**

After #1868, the Cantera toolbox defaults to `outofprocess` regardless of operating system. The MATLAB samples `ignite.m`, `ignite_uv.m`, `ignite_hp.m` and `plug_flow_reactor.m` have extreme performance penalties as they rely on integration with MATLAB's `ode15s` solver, and only use Cantera for the evaluation of properties and kinetics. Recent testing in #1974 showed runtimes for `plug_flow_reactor.m` on Windows of around 30 minutes (!). _PS: With some of the problematic examples disabled, Windows runners are significantly slower than macOS and Linux actions._

**System information**

- Cantera version: 3.2.0b1
- OS: any (but more pronounced on Windows)
- Python/MATLAB/other software versions: MATLAB 2025a

**Additional context**

- Cantera/enhancements#237 will only help marginally with this issue, as CLib only copies arrays used for setup. Most CLib array operations are currently handled by raw pointers.

I opened this issue to document previously reported behavior. Tagging @ssun30 and @rwest.

ssun30:
There's not much we can do about the inter-process communications latency, so I went for the low-hanging fruit and implemented caching where I can expect close to 100% hit rate. The most we can save is about 75% but we will still be an order of magnitude slower than `inprocess` mode.

PR #2042, opened 2025-11-12
[MATLAB] Improve performance in outofprocess mode

ssun30:

This PR introduces several optimizations to the MATLAB interface that reduces execution time for the out-of-process mode, especially in reactor examples using MATLAB's own ode solver.

- Added persistent buffer caching for numeric and char arrays (`getArray`, `getString`), eliminating repeated allocation overhead in `clib.array` objects.
- Implemented caching in `ct.Kinetics`, `ct.ThermoPhase`, and `ct.Mixture` for frequently accessed constants (`nPhases`, `nReactions`, `nSpecies`, `nElements`, etc.).
- Added cached getters for immutable `ThermoPhase` quantities (`atomicWeights`, `molecularWeights`, `charges`, `minTemp`, `maxTemp`).
- Minor changes to error handling (`ismember` can be slow in MATLAB).

Partially addressed #2034 

Profiling comparison (Mac OS 15.7, Apple M2 chip):

profiler_pfr_unoptimized_outofprocess

profiler_pfr_optimized_outofprocess

`plug_flow_reactor` execution time reduced by ~60% in `outofprocess` mode, but still an order of magnitude longer than `inprocess` mode. These optimizations also speed up `inprocess` mode by a lesser extent, but time spent on `getArray` and `call` still reduced by ~30%.

profiler_pfr_unoptimized_inprocess

profiler_pfr_optimized_inprocess

These are just a small subset of properties and methods that could be cached. They have a hit-rate of 100% until we allow the MATLAB interface to add species/reactions (but even with those the hit rate will be close to 100%). There are plenty elsewhere but I don't think the user would be calling them repeatedly. I'm still investigating whether setter methods with array inputs (such as `set.X`) can be optimized with minimized memory copy. 

**Checklist**

- [ ] Build passes (`scons build` & `scons test`) and MATLAB examples execute successfully  
- [ ] Style & formatting follow [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)  


ischoegl (commented):
Hi @ssun30 - thanks for looking into this. I had a cursory glance at this PR. 

In hindsight, it is obvious that those repeated integer fetches would create a lot of latency. Buffering them makes a whole lot of sense! At the same time, I believe that you can simplify the logic quite a bit by assuming that these integers only change for certain operations, i.e., constructor and anything that resizes. Let's take the example of `Kinetics.nReactions`. I recommend simplifying to
```
        function n = get.nReactions(obj)
            n = obj.kinCache.nReactions;
        end
```
and move the logic
```
                obj.kinCache.nReactions = ct.impl.call('mKin_nReactions', obj.kinID);
```
to the constructor, or any method that modifies the number of reactions. I have a strong hunch that this can be simplified even further by using regular MATLAB integers instead of the `struct` you introduced for caching.

Regarding the array cache, it may be simpler to create them as hidden members of objects (`ThermoPhase`, `Kinetics`, etc.), as each of the classes just use one or two standard lengths. Passing that cache to `ctArray` would be sufficient, and eliminate the logic to select/create the correct cache size there. You can obviously hybridize this  with what you already have to take care of edge cases. Fwiw, CLib doesn't care about receiving arrays that are larger than necessary.

PR #2052, opened 2025-11-19
[Python] Make TypeVar for phase covariant.

### LINKED ISSUE #2051: Bug in generic typing for SolutionArray
**Problem description**

For context, some passthrough methods on `SolutionArray` are only available for specific phases. The way I devised to constrain this was to make `SolutionArray` generic in the underlying phase object type, and then restrict those methods to only be valid when the type of `self` matches the specific phase type. For example:

```python
@property
def net_production_rates(self: SolutionArray[Kinetics]) -> Array: ...
```

However, this doesn't seem to properly match against subclasses as I expected:

Image

This error was picked up by my type checkers, as well.

**Behavior**

My expectation was that the type `SolutionArray[Solution]` would match against `self: SolutionArray[Kinetics]` as `Solution` is a subclass of `Kinetics`. I thought this worked as expected during my testing while implementing it (see [this comment](https://github.com/Cantera/cantera/pull/1926#issuecomment-3193965348)), although it's possible I only tested the negative case (`SolutionArray[Solution]` does not match against `SolutionArray[Interface]`) and not the positive case.

**System information**

- Cantera version: 3.2.0
- OS: RHEL 9
- Python/MATLAB/other software versions: Python 3.13, mypy 1.18.2, basedpyright 1.33.0, pyright 1.1.407

**Additional context**

I believe the mistake is that, by default, generics are invariant. The fix may thus be as simple as changing

```python
# Generic representing a valid "phase" input
_P = TypeVar("_P", bound=_SolutionBase)
```
to
```python
# Generic representing a valid "phase" input
_P = TypeVar("_P", bound=_SolutionBase, covariant=True)
```
in `composite.pyi`, but I haven't tested this yet.

See: https://typing.python.org/en/latest/reference/generics.html#variance-of-generic-types

TimothyEDawson:
Closes #2051

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/composite.pyi
+++ b/interfaces/cython/cantera/composite.pyi
@@ -60,7 +60,7 @@ class _EquivRatioFraction(TypedDict, Generic[_T], total=False):
     diluent: _T
 
 # Generic representing a valid "phase" input
-_P = TypeVar("_P", bound=_SolutionBase)
+_P = TypeVar("_P", bound=_SolutionBase, covariant=True)
 
 class Solution(Transport, Kinetics, ThermoPhase): ...
 class Interface(InterfaceKinetics, InterfacePhase): ...


speth:
Is there a way to get the type checking CI job to identify this problem and confirm its resolution?

TimothyEDawson:
@speth I have wondered about that, and I do think there are some things we could do to better test the typing coverage. One might be to add types to the existing files in `test/python` and/or `samples/python` and run a type checker against them. This could include measuring the stub coverage and maybe adding additional files specifically for type checks. I'll try some things locally and open an issue for it if I find a promising path forward.

speth (approved):
Thanks, @TimothyEDawson. I was likewise able to confirm the reported issue locally and that this change fixes it, so I think this is fine even in the absence of any CI coverage.

Issue #1603, opened 2023-08-25
C++ samples not checked for use of deprecated methods


speth:
**Problem description**

Currently, the C++ samples can (and do) contain use of deprecated methods. While these samples are compiled and run as part of `scons test`, we do not force deprecation warnings to be errors in these samples, and the output written to `stdout` and `stederr` is not included in the test comparisons for most of these tests, which means the warnings printed to the screen go unnoticed

**Steps to reproduce**

1. Run `scons test`
2. Run `./build/samples/cxx/bvp/blasius` (may require fiddling with `LD_LIBRARY_PATH`)
3. See `DeprecationWarning` in the output

**Behavior**

Step 1 passes with no errors. Step 2 produces output starting with:
```
DeprecationWarning: OneDim::OneDim(vector<Domain1D*>): To be removed after Cantera 3.0; superseded by OneDim::OneDim(v
ector<shared_ptr<Domain1D>>).
DeprecationWarning: OneDim::addDomain(Domain1D*): To be removed after Cantera 3.0; superseded by OneDim::addDomain(shared_ptr<Domain1D>).
DeprecationWarning: Sim1D::Sim1D(vector<Domain1D*>&): To be removed after Cantera 3.0; superseded by Sim1D::Sim1D(vector<shared_ptr<Domain1D>>&).

..............................................................................
Attempt Newton solution of steady-state problem...    failure. 
Take 10 timesteps     0.0003844     0.7784
Attempt Newton solution of steady-state problem...    success.

Problem solved on [6] point grid(s).
```

**System information**

- Cantera version: 3.0.0 (806842dac)
- OS: macOS 13.5.1

**Additional context**

I already have a fix prepared for the specific warnings here that will be included in an upcoming PR, but thought it was helpful to track this gap in our testing separately.

ischoegl:
@speth - as there have been several PR's merged since creating this issue, this may already be fixed?

speth:
This issue does not affect the sample programs in the C++ tutorial (https://cantera.org/dev/userguide/cxx-tutorial.html) which are now included in the test suite as of #1657, because the regression tests for these make comparisons of the combined `stdout` and `stderr` of the samples.

However, nothing has changed for the other C++ sample programs.

PR #2053, opened 2025-11-20
Remove deprecated code slated for removal after Cantera 3.2

speth:

- Remove deprecated code
- Implement behavior transitions indicated as occurring "after Cantera 3.2"

**Checklist**


ischoegl (commented):
While this is still work in progress, I am wondering about the outlook regarding https://github.com/Cantera/enhancements/issues/237 ... it would be a worthwhile endeavor, but also would be too much churn to introduce regular deprecation cycles in a minor version jump, as you have previously commented: "It might only be feasible to consider this as an abrupt breaking change associated with a bump to the major version." I'd be :+1: if this means going to 4.0 ... . It'd go along with a _presumably_ stable generated CLib which would then be using span under the hood and avoid raw pointers altogether.

speth:
While there are some changes here that assume that the next version is 3.3.0, I don't think we have to decide on whether to jump to the next major version just yet. I think there are some tempting reasons to do so, but we should probably discuss the tradeoffs a bit first. I'll open a different thread for that purpose.

ischoegl:
> While there are some changes here that assume that the next version is 3.3.0, I don't think we have to decide on whether to jump to the next major version just yet. I think there are some tempting reasons to do so, but we should probably discuss the tradeoffs a bit first. I'll open a different thread for that purpose.

Yes, we can presumably tag 3.3.0a1 now but still ultimately go to 4.0.

speth:
I don't know how to fix this error that's coming from the `mypy` check:
```
.../cti2yaml.pyi:412: error: Variable "cantera.cti2yaml.phase.species" is not valid as a type  [valid-type]
.../cti2yaml.pyi:412: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
```

This stems from the following (partial) class declaration:
```py
class species: ...

class phase:
    ...
    species: list[tuple[str, CommentedSeq]]
    ...
    def modify_species(self, local_species: Sequence[species]) -> None: ...
```
I don't know how to tell it that `species` in this type declaration means the `species` class rather than the `species` member of the `phase` class. Best I can think of to create a type alias for `Sequence[species]` before the declaration of the `phase` class.

Any thoughts, @TimothyEDawson?

ischoegl (commented):
I'd suggest to ignore the type hinting failure here and instead add it to #2052 (after a rebase). Beyond, I have opened Cantera/enhancements#251, as it would be good to have some troubleshooting guidance.

TimothyEDawson:
@speth I do, actually, as I had dealt with that exact error a few times. It highlights a bad practice where the same name is being used for a variable and for an existing class, so there are two options:

1. Rename the `species` class to `Species`, which is how classes should be named anyways. (But then you might want to rename all the others as well, and it could cause some users/developers to think you're using the `ct.Species` class in that file)
    - Tip: In many IDEs which utilize a Python LSP, you can highlight the original definition and rename all occurrences, even across multiple files. In VSCode the shortcut is F2.
2. For a potentially much quicker, less painful workaround: Simply move the `species: list[tuple[str, CommentedSeq]]` definition after the `def modify_species` line.

ischoegl:
I have not looked at the specifics, but it appears to be problematic to have lowercase class names (unpythonic!) where the uppercase version has the same name as `ct.Species` with a different meaning? I am both confused and concerned as I thought that stub files would simply mimic the behavior of annotating the actual definitions. I’d definitely appreciate Cantera/enhancements#251.

Fwiw, the original definition has the proper uppercase/lowercase usage we follow elsewhere:
```
def modify_species(self, k, Species species):
```
PS: Nevermind, this is `cti2yaml`!

TimothyEDawson:
@ischoegl I see you figured it out, but I did want to note that the lowercase class names were a common pattern in the conversion scripts (i.e. not my fault!), and I did actually go ahead and rename such classes in some of the conversion scripts, just not all of them. Those scripts could all certainly use a touch-up!

ischoegl (dismissed):
Thanks, @speth! As mentioned, I believe it makes sense to address the type hinting issue in a separate PR. The rest looks great, so from my perspective this is ready. As an aside, I noticed that #1603 raised its head again.

speth:
> @speth I do, actually, as I had dealt with that exact error a few times. It highlights a bad practice where the same name is being used for a variable and for an existing class, so there are two options:

It's the same name, but in almost all actual code there is no ambiguity: the `species` member variable would always be referenced as `self.species` and the bare `species` name refers to the class. It's only this weird scope of the class definition where there is any overlap.

>     1. Rename the `species` class to `Species`, which is how classes should be named anyways. (But then you might want to rename all the others as well, and it could cause some users/developers to think you're using the `ct.Species` class in that file)

As a reminder, we're talking about `cti2yaml`, where the classes defined here are the ones used in the CTI input files. So renaming the class is not an option.

>     2. For a potentially much quicker, less painful workaround: Simply move the `species: list[tuple[str, CommentedSeq]]` definition after the `def modify_species` line.

Ah, yes, I see that that should work, and helped me see why this is seen as an ambiguity at all.

TimothyEDawson:
> Thanks, @speth! As mentioned, I believe it makes sense to address the type hinting issue in a separate PR.

It wouldn't make sense to address a typing issue *introduced by this pull request* in a separate issue, it should be fixed before being merged in the first place. That's sort of the whole point of having type checks in the CI pipeline.

TimothyEDawson:
@speth I'll try to help you figure out these current typing errors, and start writing up some documentation on how they should be addressed in general. Looks like most of them are due to lines in the `.mypyignore` which are no longer required.

TimothyEDawson:
Alright, some general notes to be cleaned up and incorporated into the dev guide:

1. After building your local version of Cantera, run the following command in the root directory (note that it requires mypy to be installed):
```bash
stubtest --mypy-config-file interfaces/cython/pyproject.toml --ignore-disjoint-bases --allowlist interfaces/cython/.mypyignore --generate-allowlist cantera
```
2. Anything which follows "note: unused allowlist entry" should be removed from the `interfaces/cython/.mypyignore`, as it means it no longer applies (either the method has been removed, or the typing has been fixed).
3. For everything else, there is a typing error to be resolved. We need to decide on the proper procedure. I would propose the following for the average developer:
    - Run the following command to check what the specific typing errors are. Optionally, omit `--concise` for more verbose output, such as line numbers:
    ```bash
    stubtest --mypy-config-file interfaces/cython/pyproject.toml --ignore-disjoint-bases --allowlist interfaces/cython/.mypyignore --concise cantera
    ```
    - Fix any errors you can. E.g. "___ is not present at runtime" indicates a class or method which exists in the `.pyi` stub file which has been removed from the corresponding `.pyx` file, so remove it from the stub.
    - **NOTE**: Most items appearing within the `cantera` namespace can be ignored, as the error is a duplicate from where it actually resides. E.g. you want to fix `cantera.reaction.Arrhenius`, not `cantera.Arrhenius`. The latter will be resolved by resolving the former.
    - Re-build Cantera to ensure you have the latest stubs, and then rerun the `stubtest` command until you've resolved as many errors as you can.
        - It would be nice to find a way to only update the stub files, or even symbolically link them to the interfaces/cython/cantera directory (like when using `pip install -e .` in pip-installable projects), so you don't have to re-install. But at least `scons install` is fairly quick once the C++ code has already been built.
    - Rerun the `--generate-allowlist` command from step 1 and add any remaining lines to the `.mypyignore` file. Some of these will need to stay here as they are, as far as we're aware, unfixable. Others should be targeted for fixing via future pull requests.


I know there are ways to automate some of this and make it easier for developers. For example, it'd be nice to automatically update the `.mypyignore` file with a one-liner which diffs it against the full `--generate-allowlist` output to insert new items in a consistent order.

TimothyEDawson:
I checked out this branch to see if there is anything tricky to address, it mostly looks straightforward so far. I did notice that `cantera.reaction.Arrhenius` still exists in the namespace, presumably because of the `cdef class Arrhenius:` at the end of `reaction.pxd`. Can that be safely removed now?

TimothyEDawson:
@speth Here's a quick fix based on this branch: https://github.com/TimothyEDawson/cantera/commit/4e1fd9843da50819f3c16566805ab070f0c2fcc6

speth:
Thanks, @TimothyEDawson. For the developer docs, I'd suggest starting the recommended workflow with the second version of the `stubtest` command. At least to me, the output from the first command doesn't make it clear that the output is a list of items with (potential) errors. Also, I think it would be really helpful to just let these error messages be printed to the CI logs, rather than having this information buried in an artifact that you have to download to understand why this step "failed".

> * It would be nice to find a way to only update the stub files, or even symbolically link them to the interfaces/cython/cantera directory (like when using `pip install -e .` in pip-installable projects), so you don't have to re-install. But at least `scons install` is fairly quick once the C++ code has already been built.

I didn't find it necessary to install Cantera at all to work through this process, or even to rebuild after editing the `.pyi` files. Two things that may be necessary for this to work:
- I set the `PYTHONPATH` environment variable to `build/python` so I can import the module from that directory without installing it.
- `scons build` should be creating hard links between the files in `interfaces/cython/cantera` in `build/python/cantera/`, rather than actually copying them, on filesystems that support this (i.e. macOS and Linux).

ischoegl:
> > Thanks, @speth! As mentioned, I believe it makes sense to address the type hinting issue in a separate PR.
> 
> It wouldn't make sense to address a typing issue _introduced by this pull request_ in a separate issue, it should be fixed before being merged in the first place. That's sort of the whole point of having type checks in the CI pipeline.

Fair point, but to me it appears that the type checking implementation is still somewhat brittle. This PR is primarily about removing thousands of lines of deprecated code, and I hate to see anything else getting buried in there.

TimothyEDawson:
> Thanks, @TimothyEDawson. For the developer docs, I'd suggest starting the recommended workflow with the second version of the `stubtest` command. At least to me, the output from the first command doesn't make it clear that the output is a list of items with (potential) errors. Also, I think it would be really helpful to just let these error messages be printed to the CI logs, rather than having this information buried in an artifact that you have to download to understand why this step "failed".

Good points. I agree that the whitelist output isn't really intended to be human-readable, and the concise stubtest summary in the CI should be moved to the top instead. I did set it up to tee it to the file and the log, but I guess I didn't consider that the CI won't progress past the allowlist step when it fails. I don't recall now why I didn't move that down to the end, I thought I had.

> * `scons build` should be creating hard links between the files in `interfaces/cython/cantera` in `build/python/cantera/`, rather than actually copying them, on filesystems that support this (i.e. macOS and Linux).

Ah, that was what I didn't realize. I knew about the `PYTHONPATH` from the developer docs, but assumed those files were being copied over to the build directory rather than hard-linked back to the source.

TimothyEDawson:
> > > Thanks, @speth! As mentioned, I believe it makes sense to address the type hinting issue in a separate PR.
> > 
> > It wouldn't make sense to address a typing issue _introduced by this pull request_ in a separate issue, it should be fixed before being merged in the first place. That's sort of the whole point of having type checks in the CI pipeline.
> 
> Fair point, but to me it appears that the type checking implementation is still somewhat brittle. This PR is primarily about removing thousands of lines of deprecated code, and I hate to see anything else getting buried in there.

Well if you're removing deprecated code, you have to remove all of it. You wouldn't remove an implementation while leaving it in the header file, and likewise you would update the type stubs in sync with the implementation. I do think the implementation is fairly robust now, but the testing and CI infrastructure to ensure that remains the case do need some work. Even so, the fact that the CI job caught these errors and failed is great to see!

speth:
> Well if you're removing deprecated code, you have to remove all of it. You wouldn't remove an implementation while leaving it in the header file, and likewise you would update the type stubs in sync with the implementation. I do think the implementation is fairly robust now, but the testing and CI infrastructure to ensure that remains the case do need some work. Even so, the fact that the CI job caught these errors and failed is great to see!

Yes, I fully agree. I'm glad that the CI identified leaving these orphaned / incorrect definitions as an error. There's just a little work to be done to make those errors easier to act on, and to make it easier for developers to check this locally before opening a PR. Given the newness of having these typing stubs at all, a few rough edges are to be expected.

ischoegl (approved):

PR #2057, opened 2025-11-24
Consolidate type hinting in cti2yaml

ischoegl:

This PR introduces a consolidation of type hinting - having stub files separate from pure Python files is not ideal. The consolidation triggers more thorough type checking.

Aside: this could be further updated by forcing `PascalCase` class names; at the same time, this would expand the scope significantly.

**AI Statement (required)**

- **Extensive use of generative AI.** Significant portions of code or documentation were generated with AI, including logic and implementation decisions. All generated code and documentation were reviewed and understood by the contributor. 

**Checklist**


interfaces/cython/cantera/cti2yaml.py
```diff
@@ -107,66 +130,66 @@ def applyUnits(value):
 }
 
 # constants that can be used in .cti files
-OneAtm = 1.01325e5
-OneBar = 1.0e5
+OneAtm: float = 1.01325e5
```
bryanwweber: Types for literals like this will be inferred by the type checker, so you can remove this

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -83,7 +87,9 @@ def represent_float(self, data):
 yaml.RoundTripRepresenter.add_representer(float, represent_float)
 
 
-def applyUnits(value):
+def applyUnits(
+    value: float | int | tuple[float, str]
+) -> float | str:
```
bryanwweber, line 100: Suggest using `@overload` here to make the return type more specific. 

```python
@overload
def applyUnits(value: tuple[float, str]) -> str: ... 
@overload
def applyUnits(value: float) -> float: ...
def applyUnits(value):
    Implementation 
```
See https://typing.python.org/en/latest/spec/overload.html

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -195,7 +218,9 @@ def units(length = '', quantity = '', mass = '', time = '',
     if pressure: _upres = pressure
 
 
-def get_composition(atoms):
+def get_composition(
+    atoms: dict[str, float] | str
+) -> dict[str, float] | OrderedDict[str, float]:
```
bryanwweber: I think this distinction probably isn't worth including. Even better would be to update the implementation since this behavior is part of the spec now...

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1268,6 +1565,9 @@ class liquid_vapor(phase):
     equations of state. The substance_flag parameter selects the fluid. See
     liquidvapor.cti and liquidvapor.py for the usage of this entry type.
     """
+    pure_fluids: dict[int, str]
```
bryanwweber: I think this needs a `ClassVar[]` wrapped around this, then you can assign it on this line too

bryanwweber (commented):
I know this is still in draft but I wanted to leave a couple comments while you're working on it

interfaces/cython/cantera/cti2yaml.py
ischoegl: Agreed, but the `mypy` settings are quite aggressive so this errors if left off.

TimothyEDawson:
I would have gladly consolidated the remaining type stubs for the various conversion scripts, but was told that they're not a high priority. As you can see, it can get pretty annoying! I'll leave a few comments based on the errors I'm seeing right now.

Note that this command should report no errors:
```bash
mypy --config-file interfaces/cython/pyproject.toml interfaces/cython/cantera
```

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -354,21 +423,23 @@ def to_yaml(cls, representer, node):
...
-            comment = " ".join(c.strip() for c in comment)
+            comment_lines = node.comment.split("\n")
+            comment = " ".join(c.strip() for c in comment_lines)
             out['note'] = comment
         return representer.represent_dict(out)
 
 
 class thermo:
     """Base class for species thermodynamic properties."""
+    pref: float | None
```
TimothyEDawson, line 443: The base `thermo` class also requires a definition of `model`, as it's accessed within the `get_yaml` function below. I believe on the base class this would only requires a type and not a value, so it could be `model: ClassVar[str]` here.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -377,7 +448,17 @@ def get_yaml(self, out):
 
 class NASA(thermo):
     """The 7-coefficient NASA polynomial parameterization."""
-    def __init__(self, Trange=(0.0, 0.0), coeffs=(), p0=None):
+    model: Final = 'NASA7'
```
TimothyEDawson: Setting the value of `model` here is fine, and probably preferable, but if you do it this way you need to also remove the assignment in `__init__`, on line 474. Attempting to assign to a variable marked as `Final` is a type error. Most of the remaining errors are of this type, and your Python LSP should be putting red underlines on these within your IDE.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -463,9 +558,24 @@ def __init__(self, Trange=(0.0, 0.0), coeffs=(), p0=None):
 
 class const_cp(thermo):
     """Constant specific heat."""
-
-    def __init__(self, t0=None, cp0=None, h0=None, s0=None, tmax=None,
-                 tmin=None):
+    model: Final = 'constant-cp'
+    pref: Final[None] = None
```
TimothyEDawson: I'm seeing an error that says `Cannot override writable attribute "pref" with a final one`. This is because the base class defines `pref` as an assignable value of type `float | None`. I see what you're trying to do here, but I haven't used `Final` enough to suggest the right way to type this.

I suspect this line should just be `pref = None`, as the type is already set by the base class and I'm not sure a subclass can force it to be more narrow and/or `Final`.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -400,8 +481,12 @@ def __init__(self, Trange=(0.0, 0.0), coeffs=(), p0=None):
 
 class NASA9(thermo):
     """NASA9 polynomial parameterization for a single temperature region."""
+    model: Final = 'NASA9'
+    T_range: Sequence[float]
+    pref: float | None
```
TimothyEDawson: You don't need to redefine `pref` and its type, as it's inherited from the base class.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1795,7 +2248,8 @@ def main():
 
     # Do full validation by importing the resulting mechanism
     try:
-        from cantera import Solution, Interface
+        # Note: import-not-found needed in standalone mode, attr-defined needed in package mode
+        from cantera import Solution, Interface  # type: ignore[import-not-found,attr-defined]
```
TimothyEDawson, line 2254: This is flagged as an error: unused-ignore within the `mypy` check, but I can see how it might be required for the `stubtest` checks. I'm not sure what to do about that, exactly, but we might need to change `unused-ignore` to a warning instead of an error. That should be an option that can be set in the `pyproject.toml` file.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -933,8 +1163,18 @@ def get_yaml(self, out):
 
 
 class edge_reaction(surface_reaction):
-    def __init__(self, equation, kf, id='', order='', beta=None, options=(),
-                 rate_coeff_type=''):
+    type: Final = 'edge'
```
TimothyEDawson: `Final` is propagated through to subclasses, so because `surface_reaction` has `type: Final = "surface"`, you can't change it here. Consider using `ClassVar[str]` instead of `Final`.

TimothyEDawson: It even says as much in the tooltip when I hover over `Final`:
image

interfaces/cython/cantera/cti2yaml.py
bryanwweber: Two small things here, first you don't need the type hints for the implementation version, and second you need a third overload for the `str -> str` case ☺️

TimothyEDawson: It's not discouraged, and is arguably good practice, to have type hints on the implementation. It's just required that they are general enough to cover all possible options, where the overloads are intended to denote more specific scenarios to help constrain types.

That documentation you linked to shows this in most of the examples provided.

bryanwweber: Hmm, the only spot in the docs that mentions the implementation is [this section](https://typing.python.org/en/latest/spec/overload.html#implementation-consistency), which does imply that type hints on the implementation will be checked for consistency with the overloads. I won't object if they stay.

That said, none of the examples on the page show the implementation including type hints in the signature, except the counter-example in the consistency section. The only section that shows an "implementation" does not include the type hints (the `def utf8` example) All the other examples implement a separate function `test` which is type hinted to show the overload rules, so they aren't relevant.

interfaces/cython/cantera/cti2yaml.py
bryanwweber: Then I think we should mark them as `Final[float]` etc. so that we're actually adding information here.

interfaces/cython/cantera/cti2yaml.py
TimothyEDawson: You're right, my mistake! I mistook the last overload shown to be the implementation in the examples which did not show an implementation at all.

ischoegl:
I’m going to leave this PR here (in draft mode). Full disclosure: I was testing automated coding for this merge of stub files and Python files. I’m pleased with the outcome, but don’t want to work on this while I cannot get `mypy` etc. to run locally. Will wait for the docs.

TimothyEDawson:
> I’m going to leave this PR here (in draft mode). Full disclosure: I was testing automated coding for this merge of stub files and Python files. I’m pleased with the outcome, but don’t want to work on this while I cannot get `mypy` etc. to run locally. Will wait for the docs.

What do you mean by "cannot get `mypy` etc. to run locally", exactly?

ischoegl:
> What do you mean by "cannot get `mypy` etc. to run locally", exactly?

@TimothyEDawson ... I responded here: https://github.com/Cantera/enhancements/issues/251#issuecomment-3592555599

ischoegl:
> Note that this command should report no errors:
> 
> ```shell
> mypy --config-file interfaces/cython/pyproject.toml interfaces/cython/cantera
> ```

For my setup, I get 172 errors on current `main` / git hash dff9016104d3e8088b442b3445ba47def7f9468e

interfaces/cython/cantera/cti2yaml.py
ischoegl: Addressed the overloads, but it does appear that the implementation version requires type annotations. Without that, mypy complains about "Function is missing a type annotation"

interfaces/cython/cantera/cti2yaml.py
ischoegl: I removed all `Final`s.

interfaces/cython/cantera/cti2yaml.py
ischoegl: ~Removed the type as it appears to work after all.~

Nevermind: worked on my machine, but not on CI. The `--verifytypes` option makes the `Final[float]` etc. necessary.

interfaces/cython/cantera/cti2yaml.py
ischoegl: Added `disable_error_code = ["unused-ignore"]` to pyproject.toml

ischoegl:
@bryanwweber / @TimothyEDawson ... I believe this is now ready for another round. The `pyright` errors related to `pint` threw me off, but they are non-blocking and exist on `main` as well.

TimothyEDawson:
Awesome, I'll check it out soon! On a related note, I also have finished a first draft of the dev documentation which I'll be pushing soon.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -290,11 +365,11 @@ def __init__(self, name, atoms='', note = '', thermo=None, transport=None,
         elif isinstance(thermo, (NASA, NASA9, Shomate)):
             self.thermo = MultiPolyThermo([thermo])
         elif thermo is not None:
-            self.thermo = thermo
+            self.thermo = thermo  # type: ignore[assignment]
```
bryanwweber, line 368: nit: when adding an ignore, I find it helpful to leave a comment on the line above about why this is here.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -195,11 +234,11 @@ def units(length = '', quantity = '', mass = '', time = '',
     if pressure: _upres = pressure
 
 
-def get_composition(atoms):
-    if isinstance(atoms, dict): return atoms
+def get_composition(atoms: str) -> OrderedDict[str, float | int]:
+    """Parse composition string like 'H:2 O:1' into OrderedDict."""
```
bryanwweber, line 238: question: this causes a behavior change of this function. Is this intentional?

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -306,7 +381,7 @@ def __init__(self, name, atoms='', note = '', thermo=None, transport=None,
         _speciesnames.append(name)
 
     @classmethod
-    def to_yaml(cls, representer, node):
+    def to_yaml(cls, representer: Any, node: 'species') -> Any:
```
bryanwweber, line 384: question: could/should this be `node: Self`?

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1420,8 +1809,10 @@ def __init__(self, name='', elements='', species='', note='',
         if tabulated_thermo is None:
             raise InputError('In phase {}: Thermo data must be provided for the tabulated_species.', name)
 
-    def get_yaml(self, out):
+    def get_yaml(self, out: yaml.comments.CommentedMap) -> None:
         super().get_yaml(out)
+        assert self.tabulated_species is not None  # Checked in __init__
+        assert self.tabulated_thermo is not None  # Checked in __init__
```
bryanwweber, line 1815: question: are the asserts here to convince the type checker that these are not None? Since either value being `None` raises an `InputError`, I'd suggest removing `None` as an option for these types and moving the check above the assignment in `__init__`.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1450,7 +1852,7 @@ def __init__(self, name='', elements='', species='', note='',
             raise InputError('sublattice '+name
                             +' site density must be specified')
 
-    def modify_species(self, local_species):
+    def modify_species(self, local_species: list[species]) -> None:
```
bryanwweber, line 1855:
```suggestion
    def modify_species(self, local_species: Iterable[species]) -> None:
```
since `local_species` is only used in a `for` loop.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1465,16 +1867,30 @@ def modify_species(self, local_species):
                     if S.name in phase_species:
                         S.molar_density = self.site_density
 
-    def get_yaml(self, out):
+    def get_yaml(self, out: yaml.comments.CommentedMap) -> None:
         super().get_yaml(out)
+        assert self.site_density is not None  # Checked in __init__
```
bryanwweber, line 1872: fix: likewise here, prefer narrowing the type over `assert`

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -1594,14 +2041,17 @@ def convert(filename=None, output_name=None, text=None, encoding="latin-1"):
...
         output_name = pathlib.Path(output_name)
+    else:
+        output_name = pathlib.Path('output.yaml')  # Fallback
 
     try:
         if filename is not None:
             text = filename.read_text(encoding=encoding)
         else:
             filename = "<string>"
+        assert text is not None  # Either loaded from file or passed as parameter
```
bryanwweber, line 2054: fix: in this case, the `assert` can be removed at runtime with a flag to Python. If the further functioning of this code requires `text is not None`, then it should be an `if` that raises. Same below. Ideally, the function could be restructured to avoid having to assert these things though.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -93,6 +109,29 @@ def applyUnits(value):
         return '{} {}'.format(float2string(value[0]), units)
 
 
+# Type aliases for old/new naming conventions
+_OldKineticsModel: TypeAlias = (
+    Literal["GasKinetics"] | Literal["Interface"] | Literal["Edge"]
```
TimothyEDawson, line 114: `Literal`s with multiple options should be annotated like a list of strings, e.g. `Literal["GasKinetics", "Interface", "Edge"]`, rather than using the union operator. I had this correct in the stub file, is there a specific reason you changed it?

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -43,32 +47,32 @@ class InputError(Exception):
...
     Exception raised if an error is encountered while parsing the input file.
     """
-    def __init__(self, msg, *args):
+    def __init__(self, msg: str, *args: Any) -> None:
         if args:
             msg = msg.format(*args)
         super().__init__(msg)
 
-BlockMap = yaml.comments.CommentedMap
+BlockMap: type[yaml.comments.CommentedMap] = yaml.comments.CommentedMap
```
TimothyEDawson, line 55: In `ctml2yaml` I opted to add the following at the top:

```python
from ruamel.yaml.comments import CommentedMap, CommentedSeq
from ruamel.yaml.nodes import MappingNode, ScalarNode
from ruamel.yaml.representer import RoundTripRepresenter, SafeRepresenter
```

For two reasons:
1. *Subjective:* I really think the tradeoff in extra imports is well worth the reduced visual noise throughout the actual code.
2. *Objective:* PyLance cannot follow the import statement because `ruamel.yaml` uses star imports:

Screenshot from 2026-01-06 10-55-29

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -43,32 +47,32 @@ class InputError(Exception):
...
 
-def FlowMap(*args, **kwargs):
+def FlowMap(*args: Any, **kwargs: Any) -> yaml.comments.CommentedMap:
```
TimothyEDawson, line 57: See how I set up `FlowMap` and `FlowList` in `ctml2yaml.py`, ideally we would standardize around one way to annotate these. I think it would make sense to extract common classes like this into `_types.py` or another such utility location, as they're used in `ctml2yaml`, `cti2yaml`, and `ck2yaml`.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -80,11 +84,23 @@ def represent_float(self, data):
...
+
+@overload
+def applyUnits(value: str) -> str: ...
+@overload
+def applyUnits(value: tuple[float, str]) -> str: ...
+@overload
+def applyUnits(value: int) -> int: ...
+@overload
+def applyUnits(value: float) -> float: ...
+def applyUnits(
```
TimothyEDawson, line 98: Nice use of overloads! Looks good to me!

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -93,6 +109,29 @@ def applyUnits(value):
         return '{} {}'.format(float2string(value[0]), units)
 
 
+# Type aliases for old/new naming conventions
+_OldKineticsModel: TypeAlias = (
+    Literal["GasKinetics"] | Literal["Interface"] | Literal["Edge"]
+)
+_KineticsModel: TypeAlias = (
```
TimothyEDawson, line 116: Right now `_KineticsModel`, `_TransportModel`, and `_ConcentrationBasis` are unused. I know I had them in the stub files, but I suspect that was in anticipation of something that was never implemented, so they can be dropped for now.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -129,44 +168,44 @@ def applyUnits(value):
...
-def disable_motz_wise():
+def disable_motz_wise() -> None:
     """
     Disable the Motz & Wise correction by default for all sticking reactions.
     """
     global _motz_wise
     _motz_wise = False
 
-def validate(species = 'yes', reactions = 'yes'):
+def validate(species: str = 'yes', reactions: str = 'yes') -> None:
```
TimothyEDawson, line 194: `Literal["yes", "no"]`? Is this even used for anything anywhere?

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -226,7 +281,7 @@ def __init__(self, symbol='', atomic_mass=0.01, atomic_number=None):
         _elements.append(self)
 
     @classmethod
-    def to_yaml(cls, representer, node):
+    def to_yaml(cls, representer: Any, node: 'element') -> Any:
```
TimothyEDawson, line 284: Two things which apply to all of the `to_yaml` methods:

The primary function of the `from __future__ import annotations` is to automatically stringify all the type hints, so putting `node: 'element'` should not be necessary anymore. See: https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#forward-references

When working on `ctml2yaml` I had traced the representer through everything, and I ended up using `ruamel.yaml.representer.SafeRepresenter` for the type. I forget if that was the specific representer used in the runtime code or if it was a base class for any potential representers which might be used.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -236,9 +291,29 @@ def to_yaml(cls, representer, node):
 
 class species:
     """A constituent of a phase or interface."""
-
-    def __init__(self, name, atoms='', note = '', thermo=None, transport=None,
-                 charge=None, size=1.0, standardState=None):
+    name: str
+    atoms: OrderedDict[str, float | int]
```
TimothyEDawson, line 295: `float | int` is equivalent to `float` and will be flagged by some linters. I'm not bothered by it personally, but you might want to remove it in a few places. Note that this strictly applies to type annotations, and you still need to include both for runtime code like `isinstance` checks.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -290,11 +365,11 @@ def __init__(self, name, atoms='', note = '', thermo=None, transport=None,
         elif isinstance(thermo, (NASA, NASA9, Shomate)):
             self.thermo = MultiPolyThermo([thermo])
         elif thermo is not None:
-            self.thermo = thermo
+            self.thermo = thermo  # type: ignore[assignment]
         else:
             self.thermo = const_cp()
 
-        self.transport = None if transport == "None" else transport
+        self.transport = None if transport is None else transport
```
TimothyEDawson, line 372: This is not the same logic. Before it checked for a string `"None"` but now it will just set `self.transport = "None"` if that's passed in.

Plus, the new logic is simply equivalent to `self.transport = transport`.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -236,9 +291,29 @@ def to_yaml(cls, representer, node):
...
 class species:
     """A constituent of a phase or interface."""
-
-    def __init__(self, name, atoms='', note = '', thermo=None, transport=None,
-                 charge=None, size=1.0, standardState=None):
+    name: str
+    atoms: OrderedDict[str, float | int]
+    size: float
+    comment: str
+    thermo: 'thermo'
```
TimothyEDawson, line 298: This is quite headache-inducing, but maybe not feasible to rename things at this point. This might be okay, but it might also work if you put `thermo: thermo` after the end of the `__init__` method.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -236,9 +291,29 @@ def to_yaml(cls, representer, node):
...
+    transport: 'gas_transport | None'
+    standard_state: 'constantIncompressible | None'
```
TimothyEDawson, line 300: These quotes shouldn't be necessary.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -377,7 +457,16 @@ def get_yaml(self, out):
...
 class NASA(thermo):
     """The 7-coefficient NASA polynomial parameterization."""
-    def __init__(self, Trange=(0.0, 0.0), coeffs=(), p0=None):
+    model: str = 'NASA7'
+    T_range: Sequence[float]
+    coeffs: Sequence[float]
+
+    def __init__(
+        self,
+        Trange: Sequence[float] = (0.0, 0.0),
```
TimothyEDawson, line 466: It might be better to specify `Trange: tuple[float, float]`, even if the code can handle other sequences at runtime.

Likewise `coeffs` could be a tuple of length 7, but 1. That makes for a long annotation, 2. You couldn't set the default value to an empty tuple, and 3. It would obviate the need for the `len(coeffs)` check below.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -423,25 +514,33 @@ def __init__(self, Trange=(0.0, 0.0), coeffs=(), p0=None):
...
 class MultiPolyThermo(thermo):
-    def __init__(self, regions):
-        regions = sorted(regions, key=lambda r: r.T_range[0])
-        self.pref = regions[0].pref
-        self.Tranges = [regions[0].T_range[0]]
+    Tranges: list[float]
+    data: list[Sequence[float]]
+
+    def __init__(self, regions: Sequence[thermo]) -> None:
+        regions = sorted(regions, key=lambda r: r.T_range[0])  # type: ignore[attr-defined]
```
TimothyEDawson, line 521: If we could find a way to annotate "sequence of two-or-more `thermo` objects" that should fix a lot of these errors. Perhaps it could be inferred with a check like `if len(regions) < 2: raise ValueError`.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -590,17 +731,17 @@ def __init__(self, A=0.0, b=0.0, E=0.0, coverage=()):
 
         if coverage:
             if isinstance(coverage[0], str):
-                self.coverage = [coverage]
+                self.coverage = [coverage]  # type: ignore[list-item]
```
TimothyEDawson, line 734: I suspect some of these errors may be related to the fact that a `str` is a `Sequence[str]`, which prevents the `isinstance` check from being able to narrow the type. Might work if you switch the if-else branches and use `isinstance(coverage[0], (list, tuple))` instead. EDIT: Narrowing a `Sequence` is actually pretty difficult, so this would ideally be combined with changing the `coverage` type to only use tuple and/or list.

interfaces/cython/cantera/cti2yaml.py
```diff
@@ -653,17 +811,17 @@ def __init__(self, equation, kf, id='', order='', options=()):
         self.number = len(_reactions['reactions']) + 1
         self.id = id
         self.options = [options] if isinstance(options, str) else options
-        self.kf = Arrhenius(*kf) if isinstance(kf, (list, tuple)) else kf
+        self.kf = Arrhenius(*kf) if isinstance(kf, (list, tuple)) else kf  # type: ignore[assignment]
```
TimothyEDawson, line 814: Type narrowing here could be achieved by either changing the `kf: Sequence[float] | Arrhenius` to `kf: list[float] | tuple[float, ...] | Arrhenius` or, perhaps more simply, change this line to `self.kf = kf if isinstance(kf, Arrhenius) else Arrhenius(*kf)`.

Similar logic applies to subsequent initializations.

TimothyEDawson (changes requested):
I didn't have time to test out some of my proposed fixes, but I wanted to add this feedback sooner rather than later. There are a few more things which need to be changed, so I'll have more comments soon.

In general, I recommend using Pylance and hovering over variables in different branches to gain an understanding for what types are being inferred/narrowed.

interfaces/cython/cantera/cti2yaml.py
TimothyEDawson: Actually, it's because the `thermo` base class doesn't have these properties. If you annotate `regions` as type `Sequence[NASA | NASA9 | Shomate]` is looks like everything should work.

interfaces/cython/cantera/cti2yaml.py
TimothyEDawson: This is trickier than I initially thought because type checkers typically won't properly narrow a container's type based on checking the contents. I believe the only way to do it is to use `TypeIs` (imported from `typing_extensions`), for example:

```python
_CoverageDef: TypeAlias = tuple[str, float, float, float]
_CoverageSeq: TypeAlias = list[_CoverageDef] | tuple[_CoverageDef, ...]


def _is_coverage(x: Sequence[Any]) -> TypeIs[_CoverageDef]:
    return isinstance(x[0], str)


class Arrhenius:
    A: float
    b: float
    E: float
    coverage: _CoverageSeq | None

    def __init__(
        self,
        A: float = 0.0,
        b: float = 0.0,
        E: float = 0.0,
        coverage: _CoverageSeq | _CoverageDef = (),
    ) -> None:
        ...
        if coverage:
            if _is_coverage(coverage):
                self.coverage = [coverage]
            else:
                self.coverage = coverage
        else:
            self.coverage = None
```

TimothyEDawson:
@ischoegl I took a quick pass and found that most/all of the errors reported by Pyright were due to the fact that the subclasses were determined to be overriding the types of class attributes defined on the base class, which doesn't make a lot of sense to me. I would expect these types to be simply inherited. The solution was just to define them again on the subclasses (e.g. `ideal_gas.thermo_model`).

This branch has all of the changes I made, with everything passing now: https://github.com/TimothyEDawson/cantera/tree/remove-cti2yaml.pyi Feel free to cherry-pick or leave them. It includes a lot of completely optional formatting changes performed by Ruff. I included the Ruff settings I use in the `pyproject.toml` file, but that certainly should not be added under the present scope and without discussion. I put it in a separate commit so you could more easily leave it out.

bryanwweber (commented):
Like @TimothyEDawson I haven't had a chance to pull this down and look at it in an editor, but I had a few preliminary comments as well

interfaces/cython/cantera/cti2yaml.py
TimothyEDawson: Good point, there's no real reason to allow a `None` type if it's just going to immediately raise an `InputError`. Simply putting the `self._ = ` assignments after the `if _ is None:` checks in `__init__` would make this work.

interfaces/cython/cantera/cti2yaml.py
TimothyEDawson: It's pretty standard practice to use `assert` for type-narrowing in situations like this one, where you have arrived in a logic branch which has already ruled out a type but which the type checker fails to infer the permissible values. Type checkers are purposefully limited in their ability to follow various negations around.

Assuming the upstream logic actually prohibits `text is None` from ever being `True` at this location, adding an extra `if` here would be superfluous, and `assert` would be correct.

ischoegl:
Closing this - I believe most of the requested changes are already part of https://github.com/TimothyEDawson/cantera/tree/remove-cti2yaml.pyi so I'd suggest to just reopen as a new PR - just leave my commits in.

TimothyEDawson:
@ischoegl Couldn't you have just pulled that commit in to your branch and gone from there? Seems more straightforward than opening a new pull request, but I can try and do that in the coming days.

It's a shame there isn't a more straightforward way to directly collaborate on a pull request.

ischoegl:
> @ischoegl Couldn't you have just pulled that commit in to your branch and gone from there? Seems more straightforward than opening a new pull request, but I can try and do that in the coming days.
> 
> It's a shame there isn't a more straightforward way to directly collaborate on a pull request.

@TimothyEDawson - the `ruff` formatting would have made it extremely tedious and I don't have the bandwidth (myself, I had tried to avoid adding reformatting as actual code changes become less obvious; in other words, a pure `ruff` PR is fine but mixing it together with other changes is a big ask to reviewers). Overall, this PR was mainly for me to see how difficult it would be to eliminate the `.pyi` files for the pure Python files (as mentioned elsewhere, I see their benefit for users of our code but in terms of maintenance they are a major headache). It's not too terrible but in the end not one of my priorities.

Regarding collaborating on PRs this is possible but requires boosted privileges. I can collaborate on other contributor's PRs but I don't believe it's possible the other way around.

PR #2061, opened 2025-12-01
VCS equilibrium solver cleanup

speth:

This PR includes some broad-ranging refactoring and cleanup of the "VCS" equilibrium solver. I've done this with the idea that the only real "public" interface is through the `Mixture` class, so there are numerous API changes within the `vcs_*` classes without any attempt to keep backward compatibility. Some specific changes

- Eliminate some unnecessary classes, namely `vcs_SpeciesProperties` and `VCS_SPECIES_THERMO`
- Simplify call paths by eliminating numerous functions that were only called from a single location
- Eliminate the remaining `goto` statements
- Remove some solver options and debugging modes that were unused based on the `Mixture` interface
- Add a regression test for multiphase equilibrium at a pressure different from the standard-state reference pressure

There are no changes to the results from the solver, other than some minor changes in the high logging level outputs. This was done with the hopes of making it a little bit easier to investigate some of the long-standing bugs in the VCS equilibrium solver, but I didn't run across any smoking guns while putting this together, so all of those issues are still unresolved.

**AI Statement (required)**

  Significant portions of code or documentation were generated with AI (ChatGPT Codex using the GPT-5.1-Codex model), including logic and implementation decisions. All generated code and documentation were reviewed and understood by the contributor.

**Checklist**


include/cantera/equil/vcs_defs.h
ischoegl, line 1: Great to see most of those defines gone, but I'm wondering whether we'd like to keep this at all? Shouldn't we at least use some `enum`'s? At the same time, it's ok to leave this for a later phase in the cleanup.

ischoegl (approved):
Thanks, @speth, for preparing to tackle some long-standing issues. Given the late review, I'll approve as is, as I'm also under the assumption that this will have additional rounds of follow-up (apart from my comment, I'm also not a big fan of `plogf` logging, but all of this is definitely out of scope).

The CI coverage of 68.4% could be higher, but then there's always the issue that some of the corner cases are hard to catch.

speth:
Yes, this is certainly not the end of what's possible for cleanup of the VCS solver, but I thought this was a good start. The low patch coverage is driven in large part by the verbose logging options, most of which are not exercised in the test suite since minor numerical differences could lead to large differences in the exact solver path. I'd look more at the increase in overall project coverage (+0.45%) as a win.

PR #2063, opened 2025-12-01
Fix species slicing error with Numpy 2.4.0 dev version

speth:

Fixes an error in the `Solution` species slicing feature due to changes being introduced in Numpy 2.4.0. These changes are currently causing some of the post-merge tests that use the development version of numpy to fail with error messages like the following:

```pytb
____________________ TestTransport.test_species_visosities _____________________
[gw2] linux -- Python 3.14.0 /home/runner/work/cantera/cantera/.venv/bin/python3

self = <python.test_transport.TestTransport object at 0x7fcad0734050>
phase = <cantera.composite.Solution object at 0x7fcad2f42e50>

    def test_species_visosities(self, phase):
        for species_name in phase.species_names:
            # check that species viscosity matches overall for single-species
            # state
            phase.X = {species_name: 1}
            phase.TP = 800, 2*ct.one_atm
            visc = phase.viscosity
>           assert phase[species_name].species_viscosities[0] == approx(visc)
                   ^^^^^^^^^^^^^^^^^^^

test/python/test_transport.py:212: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
build/python/cantera/solutionbase.pyx:371: in cantera.solutionbase._SolutionBase.__getitem__
    copy.selected_species = selection
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   self._selected_species.resize(len(species))
E   ValueError: cannot resize an array that references or is referenced
E   by another array in this way.
E   Use the np.resize function or refcheck=False

build/python/cantera/solutionbase.pyx:398: ValueError
```

Since the post-merge workflow doesn't normally run on PRs, I manually ran this on [my fork](https://github.com/speth/cantera/actions/runs/19836253058), where you can see that the tests all pass.

**Checklist**

### DIFF
--- a/interfaces/cython/cantera/solutionbase.pyx
+++ b/interfaces/cython/cantera/solutionbase.pyx
@@ -54,7 +54,7 @@ cdef class _SolutionBase:
         cxx_soln.get().setKinetics(newKinetics(stringify("none")))
         cxx_soln.get().setTransportModel(stringify("none"))
         _assign_Solution(self, cxx_soln, True)
-        self._selected_species = np.ndarray(0, dtype=np.uint64)
+        self._selected_species = np.ndarray(0, dtype=np.intp)
 
     def _cinit(self, infile="", name="", adjacent=(), origin=None, yaml=None,
                thermo=None, species=(), kinetics=None, reactions=(), **kwargs):
@@ -82,7 +82,7 @@ cdef class _SolutionBase:
             _assign_Solution(self, CxxNewSolution(), True)
             self._init_parts(thermo, species, kinetics, transport, adjacent, reactions)
 
-        self._selected_species = np.ndarray(0, dtype=np.uint64)
+        self._selected_species = np.ndarray(0, dtype=np.intp)
 
     def __init__(self, *args, **kwargs):
         if isinstance(self, Transport) and kwargs.get("init", True):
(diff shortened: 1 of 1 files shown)


interfaces/cython/cantera/solutionbase.pyx
```diff
@@ -395,7 +395,7 @@ cdef class _SolutionBase:
         def __set__(self, species):
             if isinstance(species, (str, int)):
                 species = (species,)
-            self._selected_species.resize(len(species))
+            self._selected_species = np.ndarray(len(species), dtype=np.uint64)
```
bryanwweber: I'm not sure how this is used, but... could this become an array of Booleans of length n_species to avoid repeated memory allocation? I'm sure this is not a performance bottleneck so the effort might not be worth it, but I figured I'd ask.

bryanwweber (dismissed):
This looks great if you don't want to change anything from my comment ☺️

interfaces/cython/cantera/solutionbase.pyx
speth: The typical anticipated usage is for the selected species to be either a single species or just a few species. So I think it's better to keep a list of just the selected species rather than having to iterate over all species to find the few that are selected.

bryanwweber: I was thinking it could be used as a mask, if it was an array of books, but if we're not using it that way then no worries!

speth: We do often use this as a mask. I did some testing, and found that zeroing the array was slightly more expensive than recreating the array of only selected species, at least for `n_species=53` and `n_selected=3`. However, playing around with the slicing operation itself showed me that while slicing with the boolean array was faster for these sizes, just switching the existing usage to `np.intp` (native index size) instead of `np.uint64` offered an even larger boost, so I've updated this with that change. Hooray for minor micro-optimizations!

bryanwweber: Very interesting! Thanks for looking!

bryanwweber (approved):

PR #2064, opened 2025-12-04
Add generics to pint types and enable TypeForm support for mypy.

### LINKED ISSUE #2059: Updates to type annotations from pint
**Problem description**

It appears that the `pint` library has [updated its type annotations](https://pint.readthedocs.io/en/stable/advanced/typing.html), making types like `Quantity` and `UnitRegistry` generic. I believe this change occurred in v0.25. This is causing the type checkers to flag errors in the files in `with_units`.

**Steps to reproduce**

1. Install Cantera, `mypy`, and the latest version of `pint`.
2. Run `mypy --config-file interfaces/cython/pyproject.toml -p cantera`
3. Observe errors like the following:

```bash
[...]/site-packages/cantera/with_units/solution.pyi:31: error: Missing type parameters for generic type "Quantity"  [type-arg]
[...]/site-packages/cantera/with_units/solution.pyi:33: error: Missing type parameters for generic type "Quantity"  [type-arg]
[...]/site-packages/cantera/with_units/solution.pyi:37: error: Missing type parameters for generic type "Quantity"  [type-arg]
...
[...]/site-packages/cantera/with_units/__init__.py:8: error: Missing type parameters for generic type "UnitRegistry"  [type-arg]
```

**Behavior**

This could be avoided by simply pinning `pint<0.25`. I suspect it could also be easily fixed for newer versions by adding type annotations to `Quantity` before use, such as by defining the type alias in `solution.pyi` as:

```python
from pint import Quantity
...
_Quantity: TypeAlias = Quantity[float]
...
class Solution:
    ...
    def density_mole(self) -> _Quantity: ...
```

However I suspect this would raise errors when older versions of `pint` are installed, so we might need a version check to avoid that problem. I'm also not yet clear on how `UnitRegistry` should be typed, though it looks like it may just be `UnitRegistry[Quantity[float], Unit]`.

**System information**

- Cantera version: 3.3.0a1
- OS: RHEL 9
- Python/MATLAB/other software versions: Observed with Python 3.14 and `pint==0.25.2`.

**Additional context**

First raised by @ischoegl on [enhancement proposal 251](https://github.com/Cantera/enhancements/issues/251).

Change to `pint` introduced by this pull request: https://github.com/hgrecco/pint/pull/2231

### LINKED ISSUE #2062: Type checking CI failures
**Problem description**

The CI job "Verify typing correctness and coverage of the public Python API" has [started to fail](https://github.com/Cantera/cantera/actions/runs/19695215363) on the `main` branch. This change appears to be coincident with a new version of mypy being used, going from version 1.18.2 to 1.19.0.

**Behavior**

The command
```
mypy --config-file interfaces/cython/pyproject.toml --cobertura-xml-report type_check/ interfaces/cython/cantera
```
is failing with the errors:
```
/opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/cantera/_types.py:204: error: Unused "type: ignore" comment  [unused-ignore]
/opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/cantera/_types.py:204: error: TypeForm is experimental, must be enabled with --enable-incomplete-feature=TypeForm  [misc]
/opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/cantera/_types.py:204: note: Error code "misc" not covered by "type: ignore" comment
/opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/cantera/ctml2yaml.py:1679: error: TypeForm containing a string annotation cannot be recognized here. Surround with TypeForm(...) to recognize.  [maybe-unrecognized-str-typeform]
/opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/site-packages/cantera/ctml2yaml.py:1679: error: Argument 2 to "literal_type_guard" has incompatible type "<typing special form>"; expected "TypeForm[Never]"  [arg-type]
Found 4 errors in 2 files (checked 34 source files)
```

**System information**

- Cantera version: `main` branch at `29137bc48`
- OS: GitHub Actions `ubuntu-24.04` runner
- Python 3.10, mypy 1.19.0

TimothyEDawson:

As of pint 0.25.1, `UnitRegistry` and `Quantity` are now generic types. This pull request addresses this change by adding the `[float]` type parameter to the pint types, along with a `# type: ignore[type-arg]` directive to silence the type checkers when using `pint<=0.25`. This necessitates relaxing the `warn_unused_ignores` setting for mypy as it will otherwise error here when using `pint>0.25`.

Additionally, as of mypy 1.19.0, experimental support for `TypeForm` has been added. This triggered some unused ignores and notification to enable the experimental support, so these ignore have been removed and `TypeForm` support added to the mypy configuration in `pyproject.toml`.

This should resolve the current type checking CI failures.

Closes #2059. Closes #2062.

**Checklist**


TimothyEDawson:
I realize that trying to pin the version of pint would probably break the post-merge tests since they rely on apt-installed python3-pint, and likewise anyone relying on that. So I'm working on adding a version check to switch between the generic and non-generic types depending on which version of pint is installed.

I also have a fix already for #2062, so I will go ahead and add that to the scope of this pull request.

TimothyEDawson:
Hooray! The type checking is working once again!

TimothyEDawson:
@speth @ischoegl @bryanwweber This is ready as-is, but I also realize that I can re-enable the warning on unused ignore directives by pinning `pint>=0.25.1` in the type-checking CI step and simply removing the directives I added. I wanted to get your thoughts on this due to the following implications:

1. Anyone using an older `pint` version would encounter minor type errors when using `with_units` code, but that would have no effects on runtime operation.
2. It would require bumping the type-checking CI step to run on a newer version of Python, as `pint` dropped support for 3.10. I verified that the non-typing tests all pass on 3.10, so this doesn't affect those CI steps at all, and they don't need to pin a `pint` version at all.
3. All of the type-checks would still be valid for Python 3.10 as `mypy` is executed with `python_version = "3.10"`, except that in practice any end-users on 3.10 would necessarily see the errors in the first point since they can't upgrade `pint` to `>=0.25.1`.

Given `pint` is an optional dependency, I lean toward going ahead and running the type-checks on Python 3.14 with 3.10 compatibility mode, and re-enabling the unused type ignore error as I think it's pretty valuable. But leaving it as-is avoids users ever seeing a type error about `pint`, so I would be fine with that as well.

bryanwweber:
Without looking in detail at the options, my instinct is to have a (small!) matrix of Python versions and possibly pint versions to ensure expected behavior. Is that possible and does it make sense to do that?

TimothyEDawson:
@bryanwweber typically no, you wouldn't run multiple versions of Python because whether the type hints are statically valid or not is independent of the Python version used to run your type checker (e.g. `mypy`), as long as you are sticking to syntax available in your earliest-supported version. Which is why you can e.g. check that code is valid in 3.10 from a 3.14 installation, and I think vice-versa.

The Python version is only indirectly involved here, as the only thing that actually matters is the `pint` version. It's worth noting here that it's fundamentally impossible to truly _support_ both old and new versions of `pint`, because the static type checkers cannot execute the runtime code necessary to check what version of `pint` is installed and treat them differently - you must pick one syntax to support, and optionally you can ignore the type errors incurred by the version you choose not to support. Therefore a matrix of `pint` versions could only ever tell you whether you are missing the ignore directives.

ischoegl (approved):
Confirming that this runs on my system:

```
% mypy --config-file interfaces/cython/pyproject.toml interfaces/cython/cantera
Success: no issues found in 32 source files
```

speth:
I think we need to more clearly document the required version of `mypy`, especially as this now fails with any version of `mypy` that's more than a week old.

TimothyEDawson:
> I think we need to more clearly document the required version of `mypy`, especially as this now fails with any version of `mypy` that's more than a week old.

That's a good point. Should I pin the version to `>=1.19.0` in the CI?

Typically, Python projects will have an optional dependencies section in their `pyproject.toml` files specifically for dev tooling, but there's nothing like that to update here. Should it be added to the compile-from-source documentation, e.g. in the example `environment.yaml` file for Conda? I notice that none of those have pinned versions, but we could still add a commented-out `# - mypy  # optional (needed for static type checks)`, but that seems more geared toward developers.

I've started working on the documentation for https://github.com/Cantera/enhancements/issues/251 and looking into how we might set up a `scons lint` command like you had suggested, so I'll make sure it includes a note about the minimum supported mypy version.

ischoegl:
@TimothyEDawson ... our approach for this is to test with multiple versions of a specific tool. On ubuntu 22.04 LTS, the `mypy` version is 0.942 - should be the same as this [PyPI](https://pypi.org/project/mypy/0.942/) version. In other words, we are typically testing with tools that ship with a distro, rather than pulling the newest versions at a given time. Given that the static checking is new and mostly used by developers/contributors for internal consistency checks only, I'm :+1: with being less restrictive in supporting old versions for the time being (this is 3.3.0a1 after all). My concern is primarily to fix the current CI breakage.
 
At the same time, we must not strictly enforce things that depend on upstream packages, so I'd suggest to implement mypy settings that are less aggressive where we depend on others to the extent possible.

PR #2068, opened 2025-12-10
Energy equation for reactors using PlasmaPhase

mquiram:

- Added Joule heating and elastic power loss terms to energy equation RHS in all reactor classes when thermo: plasma
- Added PlasmaPhase methods for cp_mole(), intEnergy_mole(), entropy_mole(), and gibbs_mole() which account for a separate electron temperature. Currently, they are commented out and inherited from IdealGasPhase so that the energy equation can work and all the tests can pass. Not sure if we want to maintain inheritance from IdealGasPhase, skip the test, or maybe change the h = u + pv test for PlasmaPhase to something like h = u + pv + X_e(R(T_e - T)) and similarly change the Gibbs consistency test to account for non-equilibrium electron temperatures. Edit: these methods are now uncommented

**Checklist**


src/thermo/PlasmaPhase.cpp
```diff
@@ -677,4 +721,41 @@ void PlasmaPhase::getGibbs_RT(double* grt) const
...
+{
+    // Only implemented when using the Boltzmann two-term EEDF
+    if (m_distributionType == "Boltzmann-two-term") {
+        if (!m_eedfSolver) {
+            throw CanteraError("PlasmaPhase::electronMobility",
+                "EEDF solver is not initialized.");
+        }
+        return m_eedfSolver->getElectronMobility();
+    } else {
+        throw NotImplementedError("PlasmaPhase::electronMobility",
```
BangShiuh, line 755: Electron mobility can be supported without Boltzmann-two-term if we provide the cross-section data and the EEDF. The equation is different according to the bolsig document.  I am fine with the current implementation. But I would make another PR to improve this so that the plasma reactor can be more flexible. In addition, IonGasTransport.::getMobilities can use the electron mobility. We can cache the electron mobility value in PlasmaPhase so we don't need to call getElectronMobility() when it stays the same value.

src/thermo/PlasmaPhase.cpp
```diff
@@ -677,4 +721,41 @@ void PlasmaPhase::getGibbs_RT(double* grt) const
...
+    if (!(ne > 0.0) || !std::isfinite(ne)) {
+        return 0.0;
+    }
+    const double E  = electricField(); // V/m
+    if (!std::isfinite(E) || E == 0.0) {
+        return 0.0;
+    }
+    const double sigma = ElectronCharge * ne * mu_e; // S/m
+    return sigma * E * E; // W/m^3
+}
```
BangShiuh, line 779: I don't think checking std::isfinite for each term is needed. The error should be caught when calling the function. We can check if any term is or smaller than zero and return zero..

src/thermo/PlasmaPhase.cpp
```diff
@@ -564,9 +572,13 @@ double PlasmaPhase::elasticPowerLoss()
         rate += concentration(m_targetSpeciesIndices[i]) *
                 m_elasticElectronEnergyLossCoefficients[i];
     }
+    const double q_elastic = Avogadro * Avogadro * ElectronCharge * concentration(m_electronSpeciesIndex) * rate;
+
+    if (!std::isfinite(q_elastic)) {
+        return 0.0;
+    }
```
BangShiuh, line 603: Do we need this? Can we issue an error instead?

BangShiuh (dismissed):
@mquiram Thanks for working on this. I left some comments for the changes in class PlasmaPhase.

src/thermo/PlasmaPhase.cpp
mquiram: You're right. This, along with lines 577-579 is a sort of debugging artifact, and I agree that it's now unnecessary looking back at it.

src/zeroD/IdealGasConstPressureMoleReactor.cpp
```diff
@@ -36,10 +37,10 @@ void IdealGasConstPressureMoleReactor::getState(double* y)
 
 void IdealGasConstPressureMoleReactor::initialize(double t0)
 {
-    if (m_thermo->type() != "ideal-gas") {
+    /* if (m_thermo->type() != "ideal-gas") {
         throw CanteraError("IdealGasConstPressureMoleReactor::initialize",
                            "Incompatible phase type '{}' provided", m_thermo->type());
-    }
+    } */
```
speth: Rather than disabling this check, it would be better to make it allow both `ideal-gas` and `plasma` as the `type()` to avoid letting other `ThermoPhase` models through.

src/zeroD/IdealGasConstPressureMoleReactor.cpp
```diff
@@ -79,6 +80,15 @@ void IdealGasConstPressureMoleReactor::eval(double time, double* LHS, double* RH
...
 
+    if (auto* plasma = dynamic_cast<PlasmaPhase*>(m_thermo)) {
+        const double qJ = plasma->jouleHeatingPower(); // σE^2 [W/m^3]
+        const double qElastic = plasma->elasticPowerLoss(); // elastic transfer [W/m^3]
+        const double q_total = (qJ + qElastic) * m_vol; // total power [W]
+        if (std::isfinite(q_total)) {
+            mcpdTdt += q_total;
+        }
+    }
+
```
speth: This repeated block is begging to be pulled out into a `protected` function of the `Reactor` class that provides `q_total` so the addition to each other reactor class can just be a single line.

src/thermo/PlasmaPhase.cpp
```diff
@@ -677,4 +722,41 @@ void PlasmaPhase::getGibbs_RT(double* grt) const
...
 }
 
+double PlasmaPhase::electronMobility() const
+{
+    // Only implemented when using the Boltzmann two-term EEDF
+    if (m_distributionType == "Boltzmann-two-term") {
+        if (!m_eedfSolver) {
+            throw CanteraError("PlasmaPhase::electronMobility",
+                "EEDF solver is not initialized.");
+        }
```
speth: I think this case is impossible, given that `m_eedfSolver` is initialized in the `PlasmaPhase` constructor.

src/thermo/PlasmaPhase.cpp
```diff
@@ -556,6 +556,14 @@ void PlasmaPhase::updateElasticElectronEnergyLossCoefficient(size_t i)
 
 double PlasmaPhase::elasticPowerLoss()
 {
+    // If the electron energy distribution hasn't been initialized, skip elastic power loss.
+    if (m_electronEnergyDist.size() != m_nPoints
+        || m_electronEnergyDistDiff.size() != m_nPoints) {
+        writelog("PlasmaPhase::elasticPowerLoss: "
+             "EEDF not initialized, returning 0.\n");
+        return 0.0;
+    }
```
speth, line 587: I think this needs to be a bit more firm about when it does/doesn't work. If the `Boltzmann-two-term` mode is active, then an empty EEDF should trigger an update. And otherwise, calling this method should be an presumably be an error (like for `electronMobility`).

src/thermo/PlasmaPhase.cpp
```diff
@@ -564,9 +572,14 @@ double PlasmaPhase::elasticPowerLoss()
         rate += concentration(m_targetSpeciesIndices[i]) *
                 m_elasticElectronEnergyLossCoefficients[i];
     }
+    const double q_elastic = Avogadro * Avogadro * ElectronCharge * concentration(m_electronSpeciesIndex) * rate;
```
speth: Please try to stick to our style recommendation of 88 characters per line maximum.

interfaces/cython/cantera/ctml2yaml.py
speth: I'm not sure what necessitated these changes. With this change, I get a warning from PyLance here:
```
Argument of type "UnionType" cannot be assigned to parameter "literal" of type 
"TypeForm[_T0@literal_type_guard]" in function "literal_type_guard"
  "UnionType" is not assignable to "TypeForm[_T0@literal_type_guard]"
```
I'd suggest just leaving this as is, unless this was actually failing the mypy checks.

test/thermo_consistency/consistency.cpp
```diff
@@ -230,7 +240,17 @@ TEST_P(TestConsistency, g_eq_sum_gk_Xk)
...
+        // PlasmaPhase is a two-temperature, non-equilibrium model. The mixture Gibbs
+        // free energy is defined via g = h - T*s using the gas temperature T, while
+        // the electron chemical potential is defined using Te and pe. As a result,
+        // the single-temperature identity g = sum_k X_k * mu_k is not enforced for
+        // PlasmaPhase, so this particular consistency check is skipped.
+        GTEST_SKIP() << "PlasmaPhase: g = sum_k X_k * mu_k identity not enforced";
+    } else {
+        // Standard single-temperature identity
+        EXPECT_NEAR(g, phase->mean_X(gk), atol);
+    }
```
speth: Should the Gibbs free energy be calculated from the chemical potentials instead, using the species-specific temperature as is done for $h_k$ and $s_k$, so that there isn't a discrepancy with using the gas temperature for the portion of the mixture $g$ that comes from the electrons? That may make other tests fail, but it may be preferable to resolve those rather than fail to satisfy $g = \sum \mu_k X_k$.

include/cantera/thermo/PlasmaPhase.h
```diff
@@ -377,6 +370,35 @@ class PlasmaPhase: public IdealGasPhase
...
+     */
+    double electronMobility() const;
+
+    /**
+     * The joule heating power (W/m³)
+     *   @f[
+     *     q_J = \sigma * E^2,
+     *   @f]
+     * where @f$ sigma @f$ is the conductivity (S/m), and @f$ E @f$ is the electric
+     * field strength (V/m).
```
speth: You should also explain how the conductivity is calculated here.

src/equil/MultiPhase.cpp
```diff
@@ -472,6 +473,29 @@ double MultiPhase::volume() const
     return sum;
 }
 
+struct TeLockGuard {
```
speth: Put this class in an unnamed namespace (that is, wrap with `namespace {` ... `}` to keep it local to this translation unit.

src/equil/MultiPhase.cpp
```diff
@@ -472,6 +473,29 @@ double MultiPhase::volume() const
...
+    std::vector<std::pair<PlasmaPhase*, double>> items;
```
speth: No need to use the `std::` prefix for `vector` or `pair` -- we have `using` directives for these and other common types in `ct_defs.h`.

src/equil/MultiPhase.cpp
```diff
@@ -562,6 +587,7 @@ double MultiPhase::equilibrate_MultiPhaseEquil(int XY, double err, int maxsteps,
         throw CanteraError("MultiPhase::equilibrate_MultiPhaseEquil",
                            "No convergence for T");
     } else if (XY == SP) {
+        TeLockGuard te_guard(*this); // also single-T during SP
```
speth: You may want to move this up to the more general `MultiPhase::equilibrate` method, as it will then be activated for both the `MultiPhaseEquil` (aka "Gibbs") solver and the VCS solver. You should probably also implement something on this front for the `ChemEquil` solver (aka "element potential").

src/thermo/EEDFTwoTermApproximation.cpp
```diff
@@ -115,7 +115,8 @@ void EEDFTwoTermApproximation::converge(Eigen::VectorXd& f0)
         if (err1 < m_rtol) {
             break;
         } else if (n == m_maxn - 1) {
-            throw CanteraError("WeaklyIonizedGas::converge", "Convergence failed");
+            throw CanteraError("WeaklyIonizedGas::converge", "Convergence failed; returning last iterate.\n");
+            return;
```
speth: No need for the unreachable `return` after `throw`. Also, watch the line length.

src/thermo/PlasmaPhase.cpp
```diff
@@ -597,6 +610,38 @@ double PlasmaPhase::enthalpy_mole() const {
     return value;
 }
 
+double PlasmaPhase::cp_mole() const
+{
+    return IdealGasPhase::cp_mole();
+}
```
speth: If this doesn't do anything other than what the base class does, there's no need for an explicit override. I do think the class docstring could use a bit more documentation about how this and the definitions for the other properties result in a consistent set of thermodynamic properties (or even what the definition of $c_p$ is when there isn't a single temperature)

src/thermo/PlasmaPhase.cpp
```diff
@@ -597,6 +610,38 @@ double PlasmaPhase::enthalpy_mole() const {
...
+
+double PlasmaPhase::intEnergy_mole() const
+{
+    std::vector<double> ubar(m_kk);
```
speth: Might be worth having a temporary `m_work` array as a member of the class to avoid repeatedly allocating this vector.

speth (changes requested):
Thanks, @mquiram, I think this is a good step towards making `PlasmaPhase` work more smoothly with other parts of Cantera.

I think the new `equilibrate` behavior makes sense, but I'd like to suggest making the implementation a little more generic. Instead of introducing the `setLockTeToT` method on `PlasmaPhase`, we could instead introduce a pair of `virtual` methods on the `ThermoPhase` class that are always called at the start / end of the equilibrium solver, for example `beginEquilibrate` and `endEquilibrate`. The equilibrium "lock" class would then just call this method for every phase, with the special `PlasmaPhase` behavior being implemented in its overrides of these methods. While there aren't currently any other uses for this, I could imagine other phase types implementing some special behavior during an equilibrium solve, for example skipping some dependent property lookups or adjusting some internal tolerances, so this feature might come in use again.

Likewise, I think we want to avoid having to do conditional `dynamic_cast`s to `PlasmaPhase` within the reactor classes, and introduce a virtual method on the base `ThermoPhase` class that gets called no matter what (and returns 0 for most phase types). I'm not sure what the best name for this method would be -- something like `intrinsicHeating`, with Joule heating being one example.

speth (approved):
Thanks, @mquiram. I made a couple small edits to the implementation. This looks good to me.

speth (changes requested):
Sorry, I realized there was one area where this should be updated. @mquiram, can you add a couple of tests that exercise the following capabilities to improve the test coverage?
- an equilibrium calculation involving `PlasmaPhase`
- a reactor network with a `PlasmaPhase` where the energy equation is enabled so the `jouleHeatingPower` and `intrinsicHeating` functions actually get called as part of the process.

speth (approved):

PR #2069, opened 2025-12-21
Simplify Reactor evaluation/synchronization logic

### LINKED ISSUE #201: Eliminate traps associated with accessing ThermoPhase objects in use by Reactors
**Abstract**

For a `ThermoPhase`in use by a `Reactor`, the state of the `ThermoPhase`does not necessarily correspond to the state of the `Reactor` at the current time of the `ReactorNet`. This can lead to users extracting incorrect state information during integration (see [this UG post](https://groups.google.com/g/cantera-users/c/uf6Trn08WIY), for example). In addition, the current requirement that a single `ThermoPhase`can be shared across multiple `Reactor`s increases the complexity of evaluating the `Reactor`'s governing equations.

**Motivation**

After an integration step by `ReactorNet`, the state of the `ThermoPhase` object will generally be set to the latest time reached by the CVODES integrator, $t_{int}$. However, this time may be beyond the output time specified by the user (using the `advance(t_user)` method). If the user accesses the `ThermoPhase` object directly, they will get the state at $t_{int}$, in contrast to getting the state at $t_{user}$ by accessing it through the `reactor.thermo` property, in which CVODES will provide a result interpolated back to the correct time. For simple usage, like plotting the state as a function of time, getting the state at a slightly different time won't noticeably affect results. However, for cases like trying to evaluate sensitivities using a finite difference approach, two reactor networks will not have reached the same value of $t_{int}$, introducing a large amount of noise. This is the phenomenon dubbed "Bjarne's Trap" by @rwest, and has been a source of confusion for a long time.

Because the current structure allows multiple reactors to share a `ThermoPhase` object, there is a significant amount of complexity in the logic for evaluating the governing equations for all reactors in the network.

The current evaluation logic amounts to the following (in Python pseudocode):
```py
def eval(t, y, ydot):
    # ReactorNet::updateState
    for reactor in all_reactors:
        reactor.thermo.set_TPY(y[...])
        # save the current state information 
        reactor.state = reactor.thermo.state
        # compute auxiliary quantities based on state, for any quantities that may need
        # to be accessed from other reactors, e.g. mass fractions, pressure, or enthalpy
        reactor.local_vars = f(reactor.thermo)

    # ReactorNet::eval
    for reactor in all_reactors:
        reactor.thermo.restore_state(reactor.state)
        ydot[...] = f(reactor.thermo, reactor.kinetics, reactor.local_vars, reactor.neighbors.local_vars)
```

This complexity has resulted in difficulties for users who want to implement modified governing equations using the `ExtensibleReactor` framework. For example, see [this problem](https://groups.google.com/g/cantera-users/c/F-hUqR4TrEA) posted on the UG, and [this one](https://groups.google.com/g/cantera-users/c/ASrSvruxw-g) too.

**Possible Solutions**

I think the solution to both of these problems to stop having multiple reactors share `ThermoPhase`/`Kinetics` objects, and to have `ReactorNet.step` and `ReactorNet.advance` automatically synchronize the state of the reactors after each call (by default).

- Thanks to the implementation of `AnyMap`-based serialization, we can still allow users to use a single `Solution` object to set up multiple reactors. When initializing the `ReactorNet`, we can then check to see if any `Solution` object is being used multiple times, and then clone it to create a distinct `Solution` for each `Reactor`.
  - If a gas object is being used only once, don't make a duplicate, and automatically synchronize it after every call to step/advance (*TODO*: verify if this is necessary for step; not entirely clear when that would be the case)
  - If a gas object is being used for multiple objects, create an internal duplicate for all of the internal uses, leaving the external one corresponding to the initial condition for whichever reactor was set up last. The other reactors will be accessible via the `thermo` properties of the reactor objects. We will still automatically synchronize these after each step -- not rely on the magic call to `syncState` when the getter is used.
- The automatic synchronization can be disabled by passing a `sync=False` option to `advance()` or `step()`.

- The "simple" alternative is to force this on the user -- we just check and tell the user that they can't use the same `Solution` object multiple times, but I think the automatic cloning is much more user friendly and shouldn't be that difficult to implement.

After this change, the evaluation logic would look more like:

```py
def eval(t, y, ydot):
    # ReactorNet::updateState
    for reactor in all_reactors:
        # Reactor.updateState
        reactor.thermo.set_TPY(y[...])
        # Calculation of reactor-local variables is optional, for efficiency only

    # ReactorNet::eval
    for reactor in all_reactors:
        # Evaluation can directly access thermo from this reactor and neighboring reactors
        ydot[...] = f(reactor.thermo, reactor.kinetics, reactor.local_vars, reactor.neighbors.thermo)
```

I'd like to make sure that any changes to how the work of evaluating the reactor governing equations is divided up would provide relatively clean solutions to a few scenarios that have come up on the Users' Group. Namely:
- https://groups.google.com/g/cantera-users/c/0IeoYgEr-Ng: imposing $T(t)$ while keeping other reactor properties (volume, density, etc.) consistent
- https://groups.google.com/g/cantera-users/c/ASrSvruxw-g: using thermo properties in an `ExtensibleReactor` override override of `update_state`

### LINKED ISSUE #202: Improve ReactorNet handling of bulk phase interactions
**Abstract**

The handling of surface reactions in the `ReactorNet` framework is mostly focused on catalytic surfaces that only interact with a single bulk (gas) phase, or involve deposition/ablation with an effectively infinite reservoir. Surfaces that define reactions involving net mass transfer between two bulk phases can be defined and mostly work, but they have some rough edges. The goals of this enhancement are:

1. To simplify setup of reactor networks involving multiple bulk phases connected by surfaces
2. To ensure that such systems are integrated correctly and without duplicating calculations

**Motivation**

Consider the following example, using the input file [CaCO3-calcination-surface.yaml.txt](https://github.com/Cantera/enhancements/files/15309909/CaCO3-calcination-surface.yaml.txt) (which needs to be renamed due to GitHub's file type limitations) adapted from [this Users' Group post](https://groups.google.com/g/cantera-users/c/mapM1L_AfOU):

```py
import numpy as np
import cantera as ct

p = ct.one_atm # pressure
T = 1080.0 # surface temperature

gas = ct.Solution("CaCO3-calcination-surface.yaml", "gas")
CaCO3_phase = ct.Solution("CaCO3-calcination-surface.yaml", "CaCO3_s")
CaO_phase = ct.Solution("CaCO3-calcination-surface.yaml", "CaO_s")
surf_phase = ct.Interface("CaCO3-calcination-surface.yaml", "CaO_surf",
                          adjacent=[gas, CaCO3_phase, CaO_phase])

surf_phase.TPX = T, p, 'CaO(S):1, CaCO3(S):1, CO2(S):0'
gas.TPX = T, p, 'H2:1, CO2:0'
CaCO3_phase.TP = T, p
CaO_phase.TP = T, p

r = ct.IdealGasConstPressureReactor(gas, energy='off')
rCaCO3 = ct.ConstPressureReactor(CaCO3_phase, energy='off')
rCaO = ct.ConstPressureReactor(CaO_phase, energy='off')

rCaCO3.volume = 0.0005
rCaO.volume = 0.0005

surf = ct.ReactorSurface(surf_phase, r=r, A=1)
surf = ct.ReactorSurface(surf_phase, r=rCaCO3, A=1)
surf = ct.ReactorSurface(surf_phase, r=rCaO, A=1)
net = ct.ReactorNet([r, rCaCO3, rCaO])

states = ct.SolutionArray(gas, extra=['t','surf_coverages'])

t_end = 0.001
t = 0

masses = []
while t < t_end:
    t = net.step()
    states.append(r.thermo.state, t=t, surf_coverages=surf.coverages)
    masses.append((r.mass, rCaCO3.mass, rCaO.mass))
```

The issues to be resolved here are:

(1) The need to define 3 `ReactorSurface` objects. Currently, a `ReactorSurface` always "belongs" to a single `Reactor` object, and is responsible for determining its own surface species' rates of production and modifying the rates of production for the `Reactor` that owns it. It cannot modify the rates of species production for other phases/reactors. This approach means that the system of equations includes the surface species three times each. Also, it requires the phases adjacent to each surface always having their state synchronized when the surface rates are evaluated, which will only happen if each reactor has an independent `Solution` object, which relates this to #201.

(2) The surface reaction rates are not limited by the mass of one of the adjacent bulk phases reaching zero.

**Possible Solutions**

From an API perspective, I think it would be preferable to simply define the `ReactorSurface` as being adjacent to several `Reactor`s, for example:
```py
surf = ct.ReactorSurface(surf_phase, r=[r, rCaCO3, rCaO], A=1)
```

Internally, I'm not quite sure what the best way to implement this change would be. 

The least invasive approach would be to try to leave the current structure, where the `ReactorSurface` is managed by one of the bulk phases, but then needs to update components of the `RHS`/`LHS` arrays that affect other adjacent `Reactors`. However, this may not be possible because `Reactor::eval` is only provided with its own "local" segment of the overall `RHS`/`LHS` arrays.

Another approach would be to elevate `ReactorSurface` by a level in the evaluation of the equations, where it would be handled directly by `ReactorNet`. Here, `ReactorSurface::eval` would be responsible for computing the governing equations for each surface species, and for storing an array with the net production rates of the bulk phase species for each adjacent bulk phase. These would then be used by `ReactorXyz::eval` to compute the `RHS`/`LHS` terms for its bulk phase species.

I think either of these approaches require #201 to be resolved first.

I haven't yet thought about how to handle the mass of one of the phases going to zero.

**References**

- This was motivated by the example posted [on the Users' Group](https://groups.google.com/g/cantera-users/c/mapM1L_AfOU)
- Another relevant UG post: https://groups.google.com/g/cantera-users/c/i8nhewOoaas

speth:

This PR implements changes to how reactor governing equations are implemented to make use of the new requirement that all reactors have unique/independent contents.

- Eliminate the `syncState` and  `restoreState` methods of reactor objects. These steps are no longer necessary. 
- `ReactorSurface` objects are now responsible for calculating their own governing equations (for the surface coverages or moles). As a result, the `getSurfaceInitialConditions` and `evalSurfaces` methods are no longer necessary.
- Introduce a new `ReactorSurfaceFactory` class for handling distinct reactor surface types. There are two versions of the `newReactorSurface` method, one which takes an explicit model type parameter, and another which creates the "right" type of `ReactorSurface` depending on the types of the adjacent `Reactor`s (for example `FlowReactorSurface` for surfaces adjacent to a `FlowReactor`.
- Added `ExtensibleReactorSurface` class to Python, which allows customization of the governing equations for surfaces.
- Refactor process for assembling Jacobian elements. Individual reactors (and surfaces) now add elements to a `vector<Triplet>` where components indices are global indices within the network rather than reactor-local. This allows reactors to add terms for dependence on state variables of other reactors. Mostly supersedes #1634.
- Should allow straightforward implementation of Cantera/enhancements#203.

⚠️ No attempt has been made here to preserve backward compatibility. User code that makes use of `ExtensibleReactor` or otherwise creates derived types from the `Reactor` classes will likely need to be updated. **Merging this PR commits us to a major version bump.** ⚠️ 

- Resolves Cantera/enhancements#201
- Resolves Cantera/enhancements#202

A simple example of an `ExtensibleReactorSurface` (from `test_reactor.py`) now looks like this:
```py
class CustomSurface(ct.ExtensibleReactorSurface):
    def replace_get_state(self, y):
        y[:] = 0
        y[kPts] = 1

    def replace_update_state(self, y):
        # this is the same thing the original method does
        self.phase.set_unnormalized_coverages(y)

        # Replace actual reactions with simple absorption of H2 -> H(S)
        C = self.reactors[0].phase.concentrations
        theta = self.phase.coverages
        self.rop = 1e-4 * C[kH2] * theta[kPts]
        self.net_production_rates[:] = 0.0
        self.net_production_rates[kH2_kin] = - self.rop

    def replace_eval(self, t, LHS, RHS):
        site_density = self.phase.site_density
        RHS[kHs] = 2 * self.rop / site_density
        RHS[kPts] = - 2 * self.rop / site_density
```

Previously, this required implementing an `ExtensibleReactor` type and using that to replace the rate calculations for the attached surface even though the bulk phase governing equations were not being modified.

**AI Statement (required)**

- **Limited use of generative AI.**  Standard or boilerplate code snippets were generated with AI  and manually reviewed;
  all design, logic, and implementation decisions were made by the contributor.

**Checklist**


interfaces/cython/cantera/jacobians.pxd
```diff
@@ -44,20 +44,17 @@ cdef extern from "cantera/numerics/SystemJacobianFactory.h" namespace "Cantera":
...
 
 cdef class AdaptivePreconditioner(EigenSparseJacobian):
-    cdef set_cxx_object(self)
-    cdef CxxAdaptivePreconditioner* adaptive
+    pass
 
 cdef class BandedJacobian(SystemJacobian):
-    cdef set_cxx_object(self)
-    cdef CxxMultiJac* band_jac
+    pass
```
ischoegl: nit: missing line

interfaces/cython/cantera/jacobians.pyx
```diff
@@ -72,26 +68,23 @@ cdef class EigenSparseJacobian(SystemJacobian):
 
     _type = "eigen-sparse"
 
-    cdef set_cxx_object(self):
-        self.sparse_jac = <CxxEigenSparseJacobian*>self._base.get()
-
     def print_contents(self):
-        self.sparse_jac.printPreconditioner()
+        (<CxxEigenSparseJacobian*>self.jac).printPreconditioner()
```
ischoegl, line 72: I find the alternative (member variable defined to have the correct type) more readable, but that is likely subjective.

src/zeroD/FlowReactor.cpp
```diff
@@ -270,13 +193,12 @@ void FlowReactor::evalDae(double time, double* y, double* ydot, double* residual
 
     //! use mass continuity for velocity residual
     //! Kee.'s Chemically Reacting Flow, Eq. 16.48
-    double hydraulic = surfaceAreaToVolumeRatio();
-    residual[1] = m_u * drhodz + m_rho * dudz - sk_wk * hydraulic;
+    residual[1] = m_u * drhodz + m_rho * dudz - sk_wk;
```
ischoegl, line 196: There are some changes to the equations, which I assume are checked to yield the same results? I assume that conversion factors are reintroduced elsewhere.

test/python/test_reactor.py
```diff
@@ -933,14 +933,15 @@ def test_set_initial_time(self):
...
-    def test_reinitialize(self):
+    def test_reinitialize(self, allow_deprecated):
         self.make_reactors(T1=300, T2=1000)
         self.add_wall(U=200, A=1.0)
         self.net.advance(1.0)
         T1a = self.r1.T
         T2a = self.r2.T
 
         self.r1.phase.TD = 300, None
+        # Deprecated; After Cantera 4.0, replace with net.reinitialize()
```
ischoegl, line 944: Should this test for a specific DeprecationWarning?

ischoegl (dismissed):
Thanks, @speth - sorry for not getting around to this sooner, and I apologize that I still don't have sufficient bandwidth for a full review.

With the caveat that we should bump the version to 4.0.0a1 _prior_ to merging this PR, I am overall :+1: with the approach. I trust that our test suite is sufficiently thorough to catch regressions (caveat: I did notice some significant deviations of code in the flow reactor, where I don't think we have a comparison to numerical results in place).

interfaces/cython/cantera/jacobians.pyx
speth: This isn't just a style update -- it is fixing a bug where the `sparse_jac` member wasn't being initialized in certain cases. It's difficult to get the logic correct for assigning these pointers from the constructor when you have to deal with the C++-like behavior where derived class overrides aren't called from the base class constructor.

src/zeroD/FlowReactor.cpp
speth: Correct. There are a few changes to what some of the internal members represent. Specifically, the change here is due to separating the surface chemistry calculations and allowing for more than one reacting surface along the length -- there isn't a single "hydraulic diameter" anymore.

Also, there are in fact a couple of regression tests that ensure equivalent results for `FlowReactor`: `test_Si3N4_deposition_regression` and `TestFlowReactor.test_catalytic_surface`.

ischoegl (approved):

PR #2072, opened 2026-01-05
Document Python type annotations and checks.

### LINKED ISSUE #251: Document Python type hinting
**Abstract**

Cantera 3.2 introduced type hinting for the Python API. This enhancement request is for the addition of documentation on how to maintain stub files, troubleshoot CI failures, etc..

**Motivation**

* What problem is it trying to solve? ... improve documentation
* Who is affected by the change? ... mostly maintainers
* Why is this a good solution? ... should flatten learning curve by providing some pointers

**Possible Solutions**

Add a section to the [Develop](https://cantera.org/stable/develop/index.html#) section of the website.

TimothyEDawson:

- Adds developer documentation for Python type annotations, including:
  - Adding type check related optional dependencies to the example `environment.yaml` file.
  - Describing the current scope of mandatory type annotations (external interfaces).
  - Describing the purpose of each type checking step.
  - Showing how to run the type checking steps locally.

Closes Cantera/enhancements#251

**AI Statement (required)**

**Checklist**

### DIFF
--- a/doc/sphinx/develop/type-checking.md
+++ b/doc/sphinx/develop/type-checking.md
@@ -6,8 +6,8 @@ within third-party Python scripts and libraries, and thus aims for full type cov
 of the external interface. These annotations indicate the intended types for all inputs
 and outputs without restricting the runtime behavior of the code.
 
-A secondary benefit of type annotations is to facilitate static analysis of Python code
-in order to catch potential dynamic typing-related errors using tools like
+A secondary benefit of type annotations is to facilitate static analysis of Cantera's
+Python code in order to catch potential typing-related errors using tools like
 [mypy](https://mypy.readthedocs.io/en/latest/index.html),
 [pyright](https://microsoft.github.io/pyright/), [ty](https://docs.astral.sh/ty/), and
 [pyrefly](https://pyrefly.org/). These are only able to analyze pure Python code, so
@@ -17,18 +17,45 @@ they are of limited utility due to Cantera's extensive use of Cython syntax.
 
 Annotations should be added directly to all pure Python code (`.py` files). For Cython
 code (`.pyx` files), annotation support is limited and optional; instead, type stubs
-(`.pyi` files) must be added and maintained to document the external interface.
-
-Any changes to the externally-facing API must include explicit type annotations. Types
-should be as narrow as feasible to fit the desired usage. For example, when an input
-should be a tuple of two float values, prefer using `tuple[float, float]` over
-`Iterable[float]` even if the method would technically work with any iterable. Usage of
-`Any` types and `type: ignore` directives require an appropriate justification. Type
+(`.pyi` files) must be added and maintained to document the external interface. Any
+changes to the externally-facing API must include explicit type annotations. Type
 annotations of implementation details is optional but encouraged for pure Python code,
 and may be necessary in some situations to ensure the static type checks pass.
 
-Ensure all type annotations to be added are compatible with the lowest-supported Python
-version (currently 3.12).
+Ensure the syntax and features support the lowest-support Python version (currently
+3.12), referencing the [Python](https://typing.python.org/en/latest/) and
+[mypy](https://mypy.readthedocs.io/en/latest/index.html) documentation for guidance and
+current best practices. Some additional recommendations which are more specific to
+Cantera include:
+
+* Type aliases should generally be prepended with an underscore.
+* Aliases and special functions such as parametric `TypeGuard`s which will be used in
+  many parts of the code should be placed in `_types.py`.
+* Aliases representing any object which can be coerced into an object of type `{}`
+  should be named `{}Like`, in keeping with `ArrayLike` from `NumPy`.
+  * Examples: `ArrayLike`, `CompositionLike`, `_Func1Like`
+* Prefer false positives to false negatives. In other words, if a type is too
+  restrictive for a valid, if uncommon, usage, it should be switched for a more broad
+  alternative.
+  * Example: The `_TransportModel` alias could be a set of string literals which
+    enumerate all built-in transport models, but this would incorrectly flag any custom
+    extensions. It is therefore generalized to a `str` type.
+  * Exception: For input collections with a strict format it is often better to
+    restrict the type beyond what might be accepted at runtime. For example:
+      * A `TypedDict` provides better documentation of a dictionary of options than
+        something like `dict[str, str | float | dict[str, str | float]]`.
+      * For the thermodynamic state setters, `tuple[float, float, CompositionLike]` is
+        more informative than the alternative `Sequence[float | CompositionLike]`.
+      * Clever use of `Protocol`s may also help to achieve optimal typing.
+* The following should be avoided when possible:
+  * Use of `Any` should be rare, and is typically reserved for dictionaries with
+    complex or flexible contents (such as `kwargs`).
+  * Use of `type: ignore` directives should be considered temporary workarounds and
+    targeted for removal in future pull requests.
+  * In the context of a pull request it is good practice to insert these as a
+    placeholder for items where the developer requires assistance, allowing the CI
+    job to progress and catch any other issues. Developers are encouraged to add a
+    code comment elaborating on why it was deemed necessary.
 
 ## Verifying Type Annotations
 
@@ -38,27 +65,54 @@ discuss each command, its purpose, and how to run it locally.
 
 ### Local Setup
 
+Due to the use of compiled Cython code, most checks are performed on the built library
+rather than running directly against the source code. It is thus recommended to add the
+build directory to the environment variables following [](sec-using-build-dir):
+
+::::{tab-set}
+:::{tab-item} Linux
+:sync: linux
+```sh
+export LD_LIBRARY_PATH=$(pwd)/build/lib
+export PYTHONPATH=$(pwd)/build/python
+```
+:::
+
+:::{tab-item} macOS
+:sync: macos
+```sh
+export DYLD_LIBRARY_PATH=$(pwd)/build/lib
+export PYTHONPATH=$(pwd)/build/python
+```
+:::
+
+:::{tab-item} Windows (PowerShell)
+:sync: windows
+```pwsh
+$Env:PYTHONPATH = (Get-Location).Path + '\build\python'
+```
+:::
+::::
+
 Currently Cantera primarily uses the `mypy` type checker with an additional check from
-`pyright`, both of which can be installed via `pip` or `conda`. Note that
-`mypy>=1.19.0` is required, and additional reporting formats can be generated if the
-optional dependencies are installed with `mypy[reports]`.
+`pyright`. They can be installed with
 
-Some of the optional external dependencies provide type hints which are also employed
-by Cantera when available. It is therefore recommended to install them when developing
-within the Python interface:
-
-```yaml
-dependencies:
-- graphviz
-- pandas
-- pandas-stubs
-- pint
-- typing_extensions
+```sh
+conda install mypy>=1.19.0 lxml pyright
 ```
 
-Due to the use of compiled Cython code, most checks are performed on the built library
-rather than running directly against the source code. It is thus recommended to follow
-[](sec-using-build-dir).
+or
+
+```sh
+pip install mypy[reports]>=1.19.0 pyright
+```
+
+Some of the optional external dependencies provide type hints which are also employed
+by Cantera when available, and thus must be installed for the type checks to pass:
+
+```sh
+conda install graphviz pandas pandas-stubs pint typing_extensions
+```
 
 ### Static Type Correctness
 
@@ -100,6 +154,10 @@ These are omitted from the results by adding the `--ignoreexternal` option:
 ```sh
 pyright --ignoreexternal --verifytypes cantera
 ```
+
+Note that this check will additionally print various static type errors prior to the
+coverage report. This is expected due to differences between `mypy` and `pyright`, and
+it is not necessary to address these `pyright`-specific errors at this time.
 
 The second check is a more thorough coverage report, similar in format to the unit test
 coverage reports, generated using Mypy. This report highlights warnings and errors on a
--- a/doc/sphinx/develop/type-checking.md
+++ b/doc/sphinx/develop/type-checking.md
@@ -1,0 +1,155 @@
+(sec-python-type-annotations)=
+# Python Type Annotations
+
+Cantera employs type annotations in the Python interface primarily to support its use
+within third-party Python scripts and libraries, and thus aims for full type coverage
+of the external interface. These annotations indicate the intended types for all inputs
+and outputs without restricting the runtime behavior of the code.
+
+A secondary benefit of type annotations is to facilitate static analysis of Python code
+in order to catch potential dynamic typing-related errors using tools like
+[mypy](https://mypy.readthedocs.io/en/latest/index.html),
+[pyright](https://microsoft.github.io/pyright/), [ty](https://docs.astral.sh/ty/), and
+[pyrefly](https://pyrefly.org/). These are only able to analyze pure Python code, so
+they are of limited utility due to Cantera's extensive use of Cython syntax.
+
+## Adding Type Annotations
+
+Annotations should be added directly to all pure Python code (`.py` files). For Cython
+code (`.pyx` files), annotation support is limited and optional; instead, type stubs
+(`.pyi` files) must be added and maintained to document the external interface.
+
+Any changes to the externally-facing API must include explicit type annotations. Types
+should be as narrow as feasible to fit the desired usage. For example, when an input
+should be a tuple of two float values, prefer using `tuple[float, float]` over
+`Iterable[float]` even if the method would technically work with any iterable. Usage of
+`Any` types and `type: ignore` directives require an appropriate justification. Type
+annotations of implementation details is optional but encouraged for pure Python code,
+and may be necessary in some situations to ensure the static type checks pass.
+
+Ensure all type annotations to be added are compatible with the lowest-supported Python
+version (currently 3.12).
+
+## Verifying Type Annotations
+
+The `type-checking` CI job runs a set of checks on the type annotations, and developers
+should ensure these checks also pass locally prior to pushing. In this section we will
+discuss each command, its purpose, and how to run it locally.
+
+### Local Setup
+
+Currently Cantera primarily uses the `mypy` type checker with an additional check from
+`pyright`, both of which can be installed via `pip` or `conda`. Note that
+`mypy>=1.19.0` is required, and additional reporting formats can be generated if the
+optional dependencies are installed with `mypy[reports]`.
+
+Some of the optional external dependencies provide type hints which are also employed
+by Cantera when available. It is therefore recommended to install them when developing
+within the Python interface:
+
+```yaml
+dependencies:
+- graphviz
+- pandas
+- pandas-stubs
+- pint
+- typing_extensions
+```
+
+Due to the use of compiled Cython code, most checks are performed on the built library
+rather than running directly against the source code. It is thus recommended to follow
+[](sec-using-build-dir).
+
+### Static Type Correctness
+
+This is the standard type-checking pass for a Python library which scans the raw
+source code for type-related errors such as missing annotations, attempts to access
+nonexistant methods/attributes, or empty collections (such as lists or dictionaries)
+whose type could not be inferred.
+
+This check can be run without first building Cantera by navigating to the
+`interfaces/cython` directory and executing:
+
+```sh
+mypy -p cantera
+```
+
+Which will analyze the entire package and print the results to `stdout`. You can also
+analyze specific files with `mypy [filename]`. Mypy should report no issues and will
+print relevant information if any are found.
+
+When not inside the cython directory, the `-p` option will attempt to check the built
+library instead. Note that the configuration settings for `mypy` are included within
+the `pyproject.toml` file in the `interfaces/cython` folder, which from the
+root directory can be passed in on the command line like so:
+
+```sh
+mypy --config-file interfaces/cython/pyproject.toml -p cantera
+```
+
+### External Interface Type Coverage
+
+These checks examine the public interface looking for missing or underspecified
+annotations.
+
+The first check uses Pyright's `--verifytypes` feature to print a quick summary of the
+type coverage. Ideally there will be 100% coverage with known types; however, many
+types originate from external libraries which may not themselves be fully defined.
+These are omitted from the results by adding the `--ignoreexternal` option:
+
+```sh
+pyright --ignoreexternal --verifytypes cantera
+```
+
+The second check is a more thorough coverage report, similar in format to the unit test
+coverage reports, generated using Mypy. This report highlights warnings and errors on a
+per-line basis. While the CI generates a Cobertura-format XML file, the HTML format is
+more straightforward for developers to view:
+
+```sh
+mypy --config-file interfaces/cython/pyproject.toml --html-report type_check/ -p cantera
+```
+
+This will generate a static HTML site with the index at `type_check/index.html`. As
+with the static type check, this can also be run from `interfaces/cython` without first
+building the code.
+
+### Runtime Type Stub Correctness
+
+The final category of type checks utilizes Mypy's `stubtest` utility to import Cantera
+and compare the runtime code against the type annotations provided by the type stubs.
+Because the Cython code cannot be analyzed statically, this provides a means to verify
+the stubs are both correctly implemented and cover the public interface.
+
+Because it performs runtime analysis, this can only be run on the built library. The
+basic command is:
+
+```sh
+stubtest --mypy-config-file interfaces/cython/pyproject.toml \
+--ignore-disjoint-bases --allowlist interfaces/cython/.mypyignore \
+--concise cantera
+```
+
+Where the `--concise` option may be omitted to make `stubtest` print more details for
+any identified issues. The `--allowlist` option points to a file containing a manually
+curated list of methods to skip analyzing due to known and presently-unfixable errors;
+for example, Cython methods have signatures which cannot be inspected at runtime, with
+input parameters showing up as `(*args, **kwargs)`.
+
+If `stubtest` reports any errors and it's not clear how to fix them, the allow list
+can be automatically updated with the following commands:
+
+```sh
+# Remove everything from the allowlist except the pattern matches:
+sed -i '16,$d' interfaces/cython/.mypyignore
+
+# Generate a new allowlist and append it to the file:
+stubtest --mypy-config-file interfaces/cython/pyproject.toml \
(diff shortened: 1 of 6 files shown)


TimothyEDawson:
Note: The `lxml` dependency is required by `mypy[reports]` (and not by the core `mypy`), but for some reason `conda` doesn't seem to install this automatically. Not sure what's up with that? Ideally it wouldn't appear in the `environment.yaml` at all.

doc/sphinx/develop/type-checking.md
```diff
@@ -0,0 +1,155 @@
+(sec-python-type-annotations)=
+# Python Type Annotations
+
+Cantera employs type annotations in the Python interface primarily to support its use
+within third-party Python scripts and libraries, and thus aims for full type coverage
+of the external interface. These annotations indicate the intended types for all inputs
+and outputs without restricting the runtime behavior of the code.
+
+A secondary benefit of type annotations is to facilitate static analysis of Python code
+in order to catch potential dynamic typing-related errors using tools like
```
bryanwweber:
```suggestion
in order to catch potential typing-related errors using tools like
```

I think I understand what you were going for here with `dynamic` but it reads as more confusing to me with the use of `static` earlier in the sentence, and I don't think it's needed.

doc/sphinx/develop/type-checking.md
```diff
@@ -0,0 +1,155 @@
...
+[mypy](https://mypy.readthedocs.io/en/latest/index.html),
+[pyright](https://microsoft.github.io/pyright/), [ty](https://docs.astral.sh/ty/), and
+[pyrefly](https://pyrefly.org/). These are only able to analyze pure Python code, so
+they are of limited utility due to Cantera's extensive use of Cython syntax.
```
bryanwweber, line 14: This isn't strictly true, I don't think... These tools can absolutely provide type checking for compiled extensions, provided that a stub file is included with the package. I'd suggest just deleting this sentence to avoid confusion.

doc/sphinx/develop/type-checking.md
```diff
@@ -0,0 +1,155 @@
+(sec-python-type-annotations)=
+# Python Type Annotations
+
+Cantera employs type annotations in the Python interface primarily to support its use
+within third-party Python scripts and libraries, and thus aims for full type coverage
+of the external interface. These annotations indicate the intended types for all inputs
+and outputs without restricting the runtime behavior of the code.
+
+A secondary benefit of type annotations is to facilitate static analysis of Python code
```
bryanwweber: Does this mean Python code within Cantera itself?

doc/sphinx/develop/type-checking.md
```diff
@@ -0,0 +1,155 @@
...
+## Adding Type Annotations
+
+Annotations should be added directly to all pure Python code (`.py` files). For Cython
+code (`.pyx` files), annotation support is limited and optional; instead, type stubs
+(`.pyi` files) must be added and maintained to document the external interface.
+
+Any changes to the externally-facing API must include explicit type annotations. Types
+should be as narrow as feasible to fit the desired usage. For example, when an input
+should be a tuple of two float values, prefer using `tuple[float, float]` over
+`Iterable[float]` even if the method would technically work with any iterable. Usage of
```
bryanwweber: I don't agree with this advice. Unless there's a specific type check (perhaps there are implicit checks due to using Cython?), the most general type that is supported should be used for function arguments. Return values should definitely be narrowly typed, though.

doc/sphinx/develop/type-checking.md
```diff
@@ -0,0 +1,155 @@
...
+`Any` types and `type: ignore` directives require an appropriate justification. Type
```
bryanwweber: I think it'd be useful here to lay out a couple of appropriate justifications, and give advice about how to share them, for example, by leaving a code comment.

doc/sphinx/develop/type-checking.md
```diff
@@ -0,0 +1,155 @@
...
+annotations of implementation details is optional but encouraged for pure Python code,
+and may be necessary in some situations to ensure the static type checks pass.
+
+Ensure all type annotations to be added are compatible with the lowest-supported Python
+version (currently 3.12).
+
+## Verifying Type Annotations
+
+The `type-checking` CI job runs a set of checks on the type annotations, and developers
+should ensure these checks also pass locally prior to pushing. In this section we will
```
bryanwweber, line 63: I've never found the advice to ensure things pass locally before pushing to be very useful, YMMV 🤷 It's only really useful to run things locally if they fail in CI, since CI is the ultimate gatekeeper... If things pass locally and fail on CI we're not going to merge it, but the converse isn't necessarily true.

doc/sphinx/develop/type-checking.md
```diff
@@ -0,0 +1,155 @@
...
+within the Python interface:
+
+```yaml
+dependencies:
+- graphviz
+- pandas
+- pandas-stubs
+- pint
+- typing_extensions
+```
```
bryanwweber: I don't think this needs to be YAML. If you want to include a snippet, a `pip`, `uv`, or `conda` command to install these would be easier to copy-paste but I also just think a regular unordered list would be fine.

doc/sphinx/develop/type-checking.md
```diff
@@ -0,0 +1,155 @@
...
+input parameters showing up as `(*args, **kwargs)`.
+
+If `stubtest` reports any errors and it's not clear how to fix them, the allow list
+can be automatically updated with the following commands:
+
+```sh
+# Remove everything from the allowlist except the pattern matches:
+sed -i '16,$d' interfaces/cython/.mypyignore
+
+# Generate a new allowlist and append it to the file:
```
bryanwweber, line 204: Why not just use `>` to overwrite the file?

bryanwweber (changes requested):
Thanks Tim! I have a few fairly small comments. The only bigger one is to add some examples where we have `# type: ignore` and what their justification is.

doc/sphinx/develop/type-checking.md
TimothyEDawson: Yes, it would be improved by explicitly contrasting this against the previous paragraph's focus on third-party usage.

doc/sphinx/develop/type-checking.md
TimothyEDawson: No, it _is_ strictly true. Stub files are entirely distinct code which is analyzed *instead* of analyzing the actual code, as a workaround for situations where static analysis is not possible or feasible (such as compiled code or runtime-only attributes).

The point is that code written in Cython syntax cannot be checked by these tools _at all._ They can check stub files, but stub files are not tied to the implementation in any direct way. You could even delete all of the Cython files and still run `mypy`, `pyright`, etc. with the same results. The best you can really do is run tests to compare stub files against the runtime, and while `stubtest` is provided by `mypy` it is functionally distinct from `mypy` itself.

doc/sphinx/develop/type-checking.md
TimothyEDawson: Initially I agreed with you, so I understand the argument for accepting the most general type. And maybe that should remain the advice for some situations, but specifically on the topic of `Sequence` and `Iterable` there are many locations in the code where I can lay out a solid case for `tuple` being way better for the end user.

I think the most important factor is that a tuple can encode a specific order and number of elements. So for something which requires `str, float, float, float` (a real example from `cti2yaml.py`), you'd either use `Sequence[str | float]` or `tuple[str, float, float, float]`. This is both more informative for the user and can catch important input errors like passing in `(str, float, float)` or `(float, float, float, float)` which would both match against the `Sequence`.

I'm definitely open to rewriting this; imo it's infeasible to provide a full set of guidelines for annotating code, so we should direct users to e.g. the `mypy` documentation and then call out any Cantera-specific deviations, like this recommendation specific to collections.

doc/sphinx/develop/type-checking.md
TimothyEDawson: Well let's walk through this.

**Workflow A:**

1. User pushes a commit.
2. CI runs for ~30 minutes before the type check step either passes or fails.
3. a. If it passes: They're good to go.
3. b. If it fails: User reads the logs, believes they spot the issue, and implements a fix. Go back to step 1.

**Workflow B:**

1. User runs type checks locally, ~1 minute.
2. a. If it passes: It will _almost certainly_ pass the CI type check step. Push the commit.
2. b. If it fails: It will _absolutely_ fail the CI type check step. Iterate locally until it passes.

I don't see any logic in relying on CI to do something you can do locally much faster and without burning GHA CPU time (which costs money). The developer guide correctly recommends this for the unit tests, as does all modern DevOps. Many code bases even offer/enforce commit hooks to do these local checks automatically.

TimothyEDawson: Obviously the underlying assumption is that there is a very strong correlation between local checks and CI checks. If there is not, either your local setup is really messed up, or your CI does not reflect a sane local setup, and either way that's a serious problem which needs to be fixed.

doc/sphinx/develop/type-checking.md
TimothyEDawson: I agree, I would prefer to put a `uv pip install ...` command here. But the rest of the developer guide still uses `conda`, so that would probably be the more appropriate command to use. I'll make that change.

doc/sphinx/develop/type-checking.md
TimothyEDawson: Note the comment, "except the pattern matches". The first 15 lines of the `.mypyignore` file are regex patterns of things to ignore which were produced manually to substantially reduce the size of the file and the repetitive junk which would populate it.

doc/sphinx/develop/type-checking.md
TimothyEDawson: It's difficult to do that because they're very situational as it's super rare for them to _actually_ be required in general. Any examples I could provide here from the current code base are extremely unlikely to ever be necessary to pop up again in the future, and almost all of them can likely be removed with a bit of dedicated effort or discussion on better ways to handle things, so what information would really be conveyed?

Newly written code really should not have any ignores, as those are indications that you're doing something wrong. They mainly mark legacy code which predates type checking or similar problems in third-party dependencies.

TimothyEDawson: I would also be supportive of fleshing out a process here for getting feedback and suggestions through the code review process, but I'm not sure how prescriptive we want to be. Maybe I can move this bit about `Any` and ignores to its own paragraph, and frame them as an intermediate step in the process where the contributor should ensure the type checks still pass, but also solicit feedback on how to potentially eliminate them during code review?

doc/sphinx/develop/type-checking.md
bryanwweber: Ok, I think I was confused because I was thinking of this as regarding user code, not Cantera's internals. As you said, drawing a more distinct contrast would be good.

doc/sphinx/develop/type-checking.md
bryanwweber: I agree we can't call out everything and have to rely on review and external documentation. Adding a link to the mypy docs or other sources would be most welcome and would resolve my comment I think.

---

Separately, on the tuple type hints, I agree cases where we depend on the order of elements need to be typed specially. However, a list is just as acceptable as a tuple in user code provided it has the right items in the right indices. 

To be clear, I'm not asking you to implement this here. Gemini suggests using a protocol (I didn't test whether this works):

```python
from typing import Protocol, overload, Literal, Union

class TripleInterface(Protocol):
    @overload
    def __getitem__(self, index: Literal[0]) -> str: ...
    @overload
    def __getitem__(self, index: Literal[1]) -> str: ...
    @overload
    def __getitem__(self, index: Literal[2]) -> float: ...
    
    def __getitem__(self, index: int) -> Union[str, float]:
        ...

    def __len__(self) -> int: ...

def process_data(data: TripleInterface):
    first_name = data[0]  # Inferred as str
    score = data[2]       # Inferred as float
    print(f"{first_name}: {score}")

# Both of these will now satisfy the type checker
process_data(["Alice", "Smith", 95.5]) 
process_data(("Bob", "Jones", 88.0))
```

That seems like kind of a pain, but I think the cases where we do this are fairly regular so maybe there aren't too many protocols we'd have to create like this.

doc/sphinx/develop/type-checking.md
bryanwweber: I think just requiring a code comment is enough for contributors when they first open a pull request. That will flag it for review and avoid the first question of a reviewer "why is this needed?" I think a suggestion that most ignores and Anys aren't needed would also be helpful.

doc/sphinx/develop/type-checking.md
bryanwweber: I should have been explicit that this was just a comment and no change needed here.

doc/sphinx/develop/type-checking.md
speth: Can we stop fixating on `uv`? As long as it only installs Python packages, it really isn't that useful for managing a Cantera development environment.

doc/sphinx/develop/type-checking.md
TimothyEDawson: Certainly not, but also - It's a very concerning comment! Local testing isn't just a formality, it's *extremely* useful and saves countless hours of developer time!

doc/sphinx/develop/type-checking.md
TimothyEDawson: Hard disagree, @speth , but that discussion is way off topic.

doc/sphinx/develop/type-checking.md
bryanwweber: Like I said, that's not my experience :shrug: I'd spend way more time proactively keeping an environment up-to-date that I use maybe two or three times a year. Certainly, if CI failed I would get the local stuff set up but otherwise... The concerning thing to me is that CI takes 30 minutes to run type checks! I guess it's because it requires the built library so maybe unavoidable.

TimothyEDawson: That's correct, the actual checks are quite fast and most of the time is spent just setting up the built environment. If we ever reach the point where static type checking becomes more ubiquitous in the codebase - i.e. if we gain the ability to check the Cython code - then it would make sense to separate the static checks from the checks on the built library and move them to a linter step which would run immediately.

doc/sphinx/develop/type-checking.md
TimothyEDawson: I agree that `Protocol`s could provide us the best of both worlds, but they can quickly get complicated and esoteric, especially if you want to parametrize them for use in multiple scenarios. For now, I've added a very high level note about them, and I fully support revisiting them in future pull requests!

TimothyEDawson:
Alright, I think we're a lot closer now!

ischoegl (commented):
Thanks, @TimothyEDawson, for tackling this. I have independently played with type annotations in my own code bases and wanted to share some of my conclusions:
- `.pyi` files are a very limited band-aid. Other than convenience to third party software or end-users of our package there is no benefit to code quality (i.e., I am challenging the 'secondary benefit' you mention in your description; rather, `.pyi` stub files make code maintenance _harder_).
- Direct type annotations in `.py` files is a different subject matter, as there is an actual analysis happening. There's the obvious caveat of Cython in its current form.
- It has been my experience that making `mypy`/`pyright` pass for `.py` is relatively straightforward (as in: can be taken care of by coding agents). Adding `stubtest` runtime checks into the mix, however, makes things dicey, and involves exceedingly ugly hacks (`--allowlist`) for silly reasons.
- If we are concerned about code quality, I'd rather like to see `ruff check` than tracking down ways to make `mypy/pyright` plus `stubtest` happy. It's doable to add some of this in with stubs retroactively into our codebase (as has been done), but is not reasonable with `.py`. I'd rather wait until those tools get better.

TimothyEDawson:
> * `.pyi` files are a very limited band-aid. Other than convenience to third party software or end-users of our package there is no benefit to code quality (i.e., I am challenging the 'secondary benefit' you mention in your description).
> * Direct type annotations in `.py` files is a different subject matter, as there is an actual analysis happening. There's the obvious caveat of Cython in its current form.

I think you've misread. The stub files are _not_ where the secondary benefit applies, they only apply to the native Python code. In fact, that's the entire point of that paragraph.

> * It has been my experience that making `mypy`/`pyright` pass for `.py` is relatively straightforward (as in: can be taken care of by coding agents). Adding `stubtest` runtime checks into the mix, however, makes things dicey, and involves exceedingly ugly hacks (`--allowlist`) for silly reasons.

Sure, but stubtest is the _only_ way to verify stubs. Without it, you can put whatever you want in the stubs with no regard for actual runtime behavior, and only manual checks would catch errors and mistakes. As long as we have stub files we need stubtest, and as long as we have Cython-syntax code we need stub files.

> * If we are concerned about code quality, I'd rather like to see `ruff check` than tracking down ways to make `mypy/pyright` plus `stubtest` happy. It's doable to add some of this in with stubs retroactively into our codebase (as has been done), but is not reasonable with `.py`. I'd rather wait until those tools get better.

Ruff is a linter, not a type checker. It is not a replacement for `mypy`/`pyright` in any sense. Hence why Astral (developer of `ruff`) is currently working on `ty`, which _could_ be a replacement, and will have all of the same issues.

Two side notes:

1. I would very much like to add `ruff` to the Cantera workflow, as I have been using it extensively. I can open an enhancement issue and/or pull request for this.
2. Some people have just started discussions about type-checking Cython code on both the [ty](https://github.com/astral-sh/ty/issues/2377) and [pyrefly](https://github.com/facebook/pyrefly/issues/2097) Issue trackers, although I am not holding my breath on these.

ischoegl:
Agreed that `ruff` is a linter. But their `ANN` option, in combination with `mypy`, surfaced quite a few issues. Good to hear about `ty` and other potential future options for Cython. Those stubfiles need to go - for me as a developer, there is no benefit - having type hinting part of `py` is a real benefit as I can see what type to work with at the interfaces. For stubfiles, there is just extra maintance burden.

I'll grudgingly agree on `stubtest` being a necessary evil, but would like to see it moved to `post-merge-tests`. Any inexperienced Python programmer trying to contribute will be chasing their tail.

doc/sphinx/develop/type-checking.md
```diff
@@ -0,0 +1,185 @@
...
+These checks examine the public interface looking for missing or underspecified
+annotations.
+
+The first check uses Pyright's `--verifytypes` feature to print a quick summary of the
+type coverage. Ideally there will be 100% coverage with known types; however, many
+types originate from external libraries which may not themselves be fully defined.
+These are omitted from the results by adding the `--ignoreexternal` option:
+
+```sh
+pyright --ignoreexternal --verifytypes cantera
```
bryanwweber, line 155: This produced a bunch of error outputs in the `with_units` interface, is that expected?

doc/sphinx/develop/type-checking.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+The second check is a more thorough coverage report, similar in format to the unit test
+coverage reports, generated using Mypy. This report highlights warnings and errors on a
+per-line basis. While the CI generates a Cobertura-format XML file, the HTML format is
+more straightforward for developers to view:
+
+```sh
+mypy --config-file interfaces/cython/pyproject.toml --html-report type_check/ -p cantera
```
bryanwweber: This one also seems to require a path to the built package.

doc/sphinx/develop/type-checking.md
```diff
@@ -0,0 +1,185 @@
...
+# Python Type Annotations
+
+Cantera employs type annotations in the Python interface primarily to support its use
+within third-party Python scripts and libraries, and thus aims for full type coverage
+of the external interface. These annotations indicate the intended types for all inputs
+and outputs without restricting the runtime behavior of the code.
+
+A secondary benefit of type annotations is to facilitate static analysis of Cantera's
+Python code in order to catch potential typing-related errors using tools like
+[mypy](https://mypy.readthedocs.io/en/latest/index.html),
```
bryanwweber, line 11: I appreciate listing all the type checkers here, but we really only support mypy and pyright, right? At least, I tried to run `ty` and got a bunch of errors. I think listing everything is confusing if we don't support those type checkers, so I think we should remove those links. I think it would be helpful to show the CLI commands for the other checkers, but I don't expect you to handle that.

doc/sphinx/develop/type-checking.md
```diff
@@ -0,0 +1,185 @@
...
+by Cantera when available. It is therefore recommended to install them when developing
+within the Python interface:
+
+```sh
+conda install graphviz pandas pandas-stubs pint typing_extensions
+```
+
+Due to the use of compiled Cython code, most checks are performed on the built library
+rather than running directly against the source code. It is thus recommended to follow
+[](sec-using-build-dir).
```
bryanwweber: I see this link here, but it's easy to miss. I think we should describe one way to do everything in this file (i.e., if any steps require the built library, we should set it up so all of them use the built library) and include any necessary configuration (e.g., exporting environment variables) as code snippets in this doc page, in addition to this link. I know that will be some duplication, but typing is esoteric enough that I want to make this page as easy to understand on its own as possible.

doc/sphinx/develop/type-checking.md
```diff
@@ -0,0 +1,185 @@
...
+print relevant information if any are found.
+
+When not inside the cython directory, the `-p` option will attempt to check the built
+library instead. Note that the configuration settings for `mypy` are included within
+the `pyproject.toml` file in the `interfaces/cython` folder, which from the
+root directory can be passed in on the command line like so:
+
+```sh
+mypy --config-file interfaces/cython/pyproject.toml -p cantera
+```
```
bryanwweber, line 142: I mentioned this in the previous comment, I think this is the only way we should show to users. I could maybe be convinced that we can show the "no built library" in an expandable tab if you really want to keep it.

bryanwweber (changes requested):
Thanks again @TimothyEDawson I had a chance to pull this down and try the commands, so I have a few suggestions for clarifying and making this more self contained. I think that will make for a better experience for devs who might not be as familiar with Python type hinting.

doc/sphinx/develop/type-checking.md
TimothyEDawson: Yes, I haven't attempted to get Pyright passing as it tends to be much more strict than Mypy. It prints error messages during this check, but it does not actually error, and only the last bit about the coverage is the relevant output of the `--verifytypes` command.

doc/sphinx/develop/type-checking.md
TimothyEDawson: In the future, once we have more of the code actually being checked at all (i.e. if we can convert the Cython syntax to Python syntax), I think it would be great to include all of these type checkers in the CI. In the interim I think it's worth mentioning them to show they've been considered, and let contributors know they exist.

It's also good practice for contributors to try multiple checkers and get their code passing, even when Mypy is the only one currently _enforced_. I use Pyrefly quite a lot, myself. I don't think that's necessarily worth recommending in the developer guide, though.

doc/sphinx/develop/type-checking.md
TimothyEDawson: I'm open to laying out the directions more explicitly here, it's a good idea. I do think it's important to be explicit about which checks don't require a local build at all, as that has a significant impact on the developer workflow. Specifically, you really shouldn't be rebuilding between checks, as it adds an unnecessary step and extra time. Static type checking should be performed in a tight loop during development, and ideally most of it is happening as you type using a language server. Meanwhile, the checks on built libraries are a necessary evil right now, but mostly just when dealing with the stub files, and hopefully they can be eliminated in the future.

doc/sphinx/develop/type-checking.md
TimothyEDawson: I would argue the "no built library" is the _only_ one which should be shown.

doc/sphinx/develop/type-checking.md
bryanwweber: > Specifically, you really shouldn't be rebuilding between checks, as it adds an unnecessary step and extra time... ideally most of it is happening as you type using a language server.

Please correct me if I'm wrong, but the only way to check the runtime types is by rebuilding until and if we are able to put hints into Cython? Insofar as this is a guide for developers of Cantera and not developers using Cantera, I really think we should focus on the steps that require the build step and just accept that it's not ideal right now.

ischoegl:
@bryanwweber / @TimothyEDawson … what is the status on this PR?

bryanwweber:
Thanks for the ping Ingmar! I'm waiting for Tim to reply to me or make the suggested changes. ☺️

TimothyEDawson:
@ischoegl @bryanwweber my apologies for being MIA, since last I worked on this I welcomed my third kid into the world and it's been pretty tough catching up on everything once I returned to work! I promise I haven't forgotten!

ischoegl:
> @ischoegl @bryanwweber my apologies for being MIA, since last I worked on this I welcomed my third kid into the world and it's been pretty tough catching up on everything once I returned to work! I promise I haven't forgotten!

Congratulations! I completely understand the priorities here!

doc/sphinx/develop/type-checking.md
TimothyEDawson: That's mostly true as far as the Cython code goes, but I believe there's enough work to be done outside of that to justify emphasizing the non-rebuilding workflow.

For example, I want to add checks on the examples, given they should all be valid Cantera scripts. I'm already aware of a few missing or incomplete type annotations which would have been caught by such a check. Those checks only require the type stubs and the example scripts, so there's no need to build the code to run the checks and iterate on the stubs. Once the stubs pass that check, of course, the runtime checks would still be necessary to ensure they haven't diverged from the underlying Cython.

I plan to leave that out of this developer documentation until those checks have actually been implemented - I'll make sure they're both included in one pull request.

TimothyEDawson: @bryanwweber Please let me know if the latest update sufficiently addresses this thread!

bryanwweber:
Congratulations Tim! That's really exciting, I hope everyone is happy and healthy and the older two are excited for a new little sibling!

TimothyEDawson:
@bryanwweber Pinged you two weeks ago after pushing some updates, in case you missed that. Did that address your comments? Is there still some work to be done?

bryanwweber:
Hi @TimothyEDawson, thanks for the ping I did miss that comment! I'm on vacation this week but I'll take a look when I get home.

bryanwweber (changes requested):
Thanks @TimothyEDawson the changes look great! I had one error building locally, which is that Pandas released 3.0 this week and broke their old intersphinx link. Can you change `doc/sphinx/conf.py` to include the new (old) URL for the 2.3 docs? It should be `https://pandas.pydata.org/pandas-docs/version/2.3`. I know it's a little outside your scope here, but otherwise the build on `main` will fail. Thanks!

TimothyEDawson:
@bryanwweber sure thing! I'll try to remember to do so this evening.

TimothyEDawson:
@bryanwweber done, hope that works!

bryanwweber (approved):

PR #2077, opened 2026-01-23
Introduce std::span usage in thermo classes

speth:

- Replace `double*` arguments with `span<double>` and `const double*` arguments with `span<const double>`
- Replace methods that return `const vector<double>&` with `span<const double>`
- Remove extraneous arguments used to carry array sizes
- Modify a couple methods in `SpeciesThermoInterpType` that used `double*` to output scalars to use `double&` or to eliminate redundant output.
- Add checks in methods operating on `std::span` to require that provided arrays are of sufficient size

This is the first step in implementing the transition to `std::span`, as proposed in https://github.com/Cantera/enhancements/issues/237.

**AI Statement (required)**

**Checklist**

### DIFF
--- a/doc/sphinx/develop/clib-extensions.md
+++ b/doc/sphinx/develop/clib-extensions.md
@@ -265,3 +265,22 @@ The **sourcegen** utility uses a logging module to provide feedback. Add the ver
      existing crosswalks.
   1. Add new crosswalks to `config.yaml`, which may require updates of templates as
      well as the CLib source generator source code itself.
+
+## Debugging
+
+To run **sourcegen** in the VS Code debugger, the following configuration can be added
+to `.vscode/launch.json`:
+
+```json
+{
+    "name": "sourcegen CLib",
+    "type": "debugpy",
+    "request": "launch",
+    "module": "sourcegen",
+    "console": "integratedTerminal",
+    "args": ["--api=clib", "--verbose", "--output=interfaces/clib"],
+    "env": {
+        "PYTHONPATH": "${workspaceRoot}/interfaces/sourcegen"
+    },
+},
+```
--- a/doc/sphinx/userguide/thermodemo.cpp
+++ b/doc/sphinx/userguide/thermodemo.cpp
@@ -27,7 +27,7 @@ void thermo_demo(const string& file, const string& phase)
     // chemical potentials of the species
     size_t numSpecies = gas->nSpecies();
     vector<double> mu(numSpecies);
-    gas->getChemPotentials(mu.data());
+    gas->getChemPotentials(mu);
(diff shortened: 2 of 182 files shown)


doc/sphinx/develop/clib-extensions.md
```diff
@@ -265,3 +265,22 @@ The **sourcegen** utility uses a logging module to provide feedback. Add the ver
      existing crosswalks.
   1. Add new crosswalks to `config.yaml`, which may require updates of templates as
      well as the CLib source generator source code itself.
+
+## Debugging
+
+To run **sourcegen** in the VS Code debugger, the following configuration can be added
+to `.vscode/launch.json`:
```
bryanwweber, line 272: FWIW, I wouldn't be opposed to committing these files to the repo. VS Code is by far the most popular editor, and other editors will either use these settings or ignore them ☺️

ischoegl: I think this particular configuration is best kept here as it’s highly specific to the sourcegen code.

On a related note I’m very happy that adding `span` appears to not require large-scale changes here 🎉 .

bryanwweber: I don't think it matters what component it's for. The point of committing it is to share config among devs for any component of the repo. Unless there's machine-specific config needed, it should be fine to have it regardless

speth: The challenge is that VS Code doesn't provide a way of having config (or at least debugger config) coming from multiple files within the project folder, so adding this to the repository would effectively preclude adding any local configuration (or result in local configuration modifications accidentally getting committed to the repository).

For my part, I do have a bunch of other debug configurations defined that do depend on various local paths that can't be generalized (or at least, I don't know how to separate them).

bryanwweber: TLDR: there are probably ways we can accomplish sharing a config, but they're outside the scope of this PR so the following is just my train of thought, no changea needed!

---

That's a good point about merging configuration. There does seem to be a way to add [global launch configs](https://code.visualstudio.com/docs/debugtest/debugging-configuration#_global-launch-configuration), where you could put configs that wouldn't be committed, but then those are available in irrelevant projects for you. 

Another option might be to use a multi-root workspace config file (`*.code-workspace`) which allows you to define a "workspace" level config and each folder in the workspace can have its own local `.vscode/launch.json` config file. That doesn't _directly_ apply to Cantera though since we have things with basically a single root folder, so it'd be kind of a hack. 

In theory, I think you can commit a file to the repo and then add it to `.gitignore` so future changes to the file are ignored unless you force-add it.

ischoegl (approved):
Thanks, @speth!

PR #2083, opened 2026-02-03
Counterflow flame initial condition

wandadars:
When running a counterflow diffusion flame case with two inert boundaries, Cantera fails to run because it attempts to use a stoichiometric initial condition that depends on elements that may not be present in the flow. This adds a `mode=` to the `set_initial_guess()` method to allow someone to simply set a linear interpolation between BCs as the initial solution guess. A unit test for the new `mode=` option was also added to test the double inert BC solve.

**AI Statement (required)**

For additional information on Cantera's AI policy, see
https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md#restrictions-on-generative-ai-usage -->


doc/sphinx/python/onedim.rst
```diff
@@ -23,6 +23,11 @@ CounterflowDiffusionFlame
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 .. autoclass:: CounterflowDiffusionFlame
 
+   .. note::
+      ``CounterflowDiffusionFlame.set_initial_guess`` accepts ``mode="linear"``
+      to initialize profiles by linearly interpolating inlet temperature and
+      composition.
+
```
speth: I'd suggest keeping documentation in the class docstring. Also, I don't know if having a "note" in the class docstring is necessary, given that the extra parameter is described in the method docstring, and this will show up in the API docs.

interfaces/cython/cantera/onedim.py
```diff
@@ -955,7 +963,37 @@ def set_initial_guess(self, data=None, group=None):
...
+            for k, spec in enumerate(self.gas.species_names):
+                self.flame.set_profile(spec, zrel, Y[:,k])
+            return
+
+        afr = self.gas.stoich_air_fuel_ratio(Yin_f, Yin_o, 'mass')
+        zst = 1 / (1 + afr)
+        if not np.isfinite(zst) or zst <= 0.0 or zst >= 1.0:
+            raise CanteraError(
+                "Stoichiometric initial guess requires a reactive fuel/oxidizer "
+                "pair; use mode='linear'.")
```
```suggestion
            raise CanteraError(
                "Stoichiometric initial guess requires a reactive fuel/oxidizer "
                "pair; use mode='linear' for other mixtures.")
```

interfaces/cython/cantera/onedim.py
```diff
@@ -955,7 +963,37 @@ def set_initial_guess(self, data=None, group=None):
...
+            self.flame.set_profile('spreadRate', [0.0, x0/dz, 1.0], [0.0, a, 0.0])
+            self.flame.set_profile("Lambda", [0.0, 1.0], [L, L])
+            self.flame.set_profile('T', zrel, T)
+            for k, spec in enumerate(self.gas.species_names):
+                self.flame.set_profile(spec, zrel, Y[:,k])
+            return
+
+        afr = self.gas.stoich_air_fuel_ratio(Yin_f, Yin_o, 'mass')
+        zst = 1 / (1 + afr)
+        if not np.isfinite(zst) or zst <= 0.0 or zst >= 1.0:
```
speth: The output of the `stoich_air_fuel_ratio` method is restricted to [0, inf], meaning that `zst` should always be on the interval [0, 1]. You just need to avoid the two limiting values.

interfaces/cython/cantera/onedim.pyi
```diff
@@ -346,6 +346,7 @@ class BurnerFlame(FlameBase):
         self,
         data: SolutionArray[Solution] | DataFrame | str | Path | None = None,
         group: str | None = None,
+        mode: Literal["stoich", "linear"] = "stoich",
```
speth: You added this to `BurnerFlame` instead of `CounterflowDiffusionFlame`, which is I guess the reason for the error from the typing validator.

speth (changes requested):
Thanks, @wandadars. This mostly seems fine. Just a couple of minor issues for cleanup.

speth (approved):

PR #2090, opened 2026-02-20
Convert Transport, numerics and various utilities to using std::span

### LINKED ISSUE #237: Utilize std::span & migrate to C++20
**Abstract**

Replace extensive use of unchecked (and uncheckable) raw pointers with `std::span`, which means requiring a C++20 capable compiler. 

**Motivation**

A simple example is replacing:
```c++
void getMassFractions(double* const y) const;
```
with:
```c++
void getMassFractions(std::span<double> y) const;
```

This provides a number of improvements for calling these methods without being tied to a specific container, providing flexibility for end users without requiring the use of raw pointers. For example, all of the following work:

```c++
using std::span;
size_t K = gas->nSpecies();

// std::vector 
vector<double> yy(K);
gas->getMassFractions(yy);

// Eigen arrays (starting in the just-released Eigen 5.0.0)
Eigen::ArrayXd ee(K);
gas->getMassFractions(ee);

// For older Eigen versions:
gas->getMassFractions(span<double>(ee.data(), ee.data() + ee.size()));

// Raw pointers
double* ss = new double[K];
gas->getMassFractions(span<double>(ss, K));
delete[] ss;
```

At least for some compilers and standard libraries (GCC / libstdc++) using `std::span` provides bounds checking when the `NDEBUG` macro is not defined. Others (Clang / libc++) seem not to check under any circumstances. Substituting a different implementation of `span` (for example https://github.com/tcbrindle/span) could be useful in such cases. This also gives us the option of introducing our own bounds checks at key entry points.

The last option shown here of constructing the span from a raw pointer and length would couple cleanly with the generated CLib, where the length of the input array is consistently provided with all functions using array inputs and outputs.

**Implementation Considerations**
There are also benefits within the Cantera implementation, since `span` works with other C++ features such as assignment operators and range-based for loops. For example:
```c++
// legacy
void Phase::getMassFractions(double* const y) const {
    copy(m_y.begin(), m_y.end(), y);
}

// replacement
void Phase::getMassFractions(span<double> y) const {
    y = m_y;
}
```

More benefits to working with objects of this type rather than bare pointers are on the horizon in C++23, with things like [`std::views::enumerate`](https://en.cppreference.com/w/cpp/ranges/enumerate_view.html) and [`std::views::zip`](https://en.cppreference.com/w/cpp/ranges/zip_view.html).

`span` objects also provide methods for accessing sub-ranges. For example, the implementation of `saveState` might change in something like the following way: 
```c++
// legacy
void Phase::saveState(size_t lenstate, double* state) const {
    if (lenstate < stateSize()) {
        throw ArraySizeError("Phase::saveState", lenstate, stateSize());
    }
    savePartialState(lenstate, state);
    auto native = nativeState();
    if (native.count("X")) {
        getMoleFractions(state + native["X"]);
    } else if (native.count("Y")) {
        getMassFractions(state + native["Y"]);
    }
}

// replacement
void Phase::saveState(span<double> state) const {
    if (state.size() < stateSize()) {
        throw ArraySizeError("Phase::saveState", state.size(), stateSize());
    }
    savePartialState(state.first(partialStateSize()));
    auto native = nativeState();
    if (native.count("X")) {
        getMoleFractions(state.subspan(native["X"], nSpecies()));
    } else if (native.count("Y")) {
        getMassFractions(state.subspan(native["Y"], nSpecies()));
    }
}
```
While the implementation in this case isn't any more succinct, it at least gets rid of the explicit pointer arithmetic, and also eliminates the need for a separate `Phase::saveState(vector<double> state)` overload.

Regarding the availability of C++20 support, GCC 11.2, available in Ubuntu 22.04, provides `std::span` as well as most of the C++20 features that I think we would want to use in the near-term. Support in other compilers (Clang, MSVC) has likewise been around for several years at this point, so I think bumping this requirement after the Cantera 3.2 release is reasonable.

The biggest challenge with implementing this, I think, is the sheer number of methods that are affected by this change, and the number of transitional methods that would be created if we were to do this with our usual approach to deprecating and replacing methods in subsequent minor versions. It might only be feasible to consider this as an abrupt breaking change associated with a bump to the major version.

speth:

This should be the last PR in the sequence for updating to use `std::span`. It covers the `Transport` class and friends, the various matrix classes, and other utility code. It also adds some documentation explaining the necessary migrations.

- Closes Cantera/enhancements#237

**AI Statement (required)**

  logic and implementation decisions. All generated code and documentation were reviewed and understood by the contributor. *Written with the assistance of ChatGPT Codex 5.3*.

**Checklist**

### DIFF
--- a/doc/sphinx/develop/style-guidelines.md
+++ b/doc/sphinx/develop/style-guidelines.md
@@ -42,12 +42,16 @@ it easier for others to understand your code in the context of Cantera as a whol
   GCC 13.2, Clang 18.0, Visual Studio 2019 version 16.11 and Intel 2023.1.
 * Cantera moves frequently used C++ standard namespace types and functions into the
   declarative region, meaning that the `std` scope resolution can be omitted. This
-  applies to the following: `string`, `vector`, `map`, `set`, `pair`, `shared_ptr`,
-  `make_shared`, `unique_ptr`, `make_unique` and `function`. Example: use `string`
-  instead of `std::string`; a `using namespace std;` declaration is not required.
+  applies to the following: `string`, `vector`, `span`, `map`, `set`, `pair`,
+  `shared_ptr`, `make_shared`, `unique_ptr`, `make_unique` and `function`. Example: use
+  `string` instead of `std::string`; a `using namespace std;` declaration is not
+  required.
 * Avoid manual memory management (that is, `new` and `delete`), preferring to use
   standard library containers, as well as `unique_ptr` and `shared_ptr` when dynamic
   allocation is required.
+* Methods which take array arguments should use `span<double>` or `span<const double>`
+  instead of bare pointers that do not provide size information, and instead of specific
+  types that require the user to use a particular container like `std::vector<double>&`.
 * Portions of Boost which are "header only" may be used. If possible, include Boost
   header files only within .cpp files rather than other header files to avoid
   unnecessary increases in compilation time. Boost should not be added to the public
--- a/doc/sphinx/reference/releasenotes/index.md
+++ b/doc/sphinx/reference/releasenotes/index.md
@@ -3,6 +3,7 @@
 ```{toctree}
 :maxdepth: 1
 
+v4.0
 v3.2
 v3.1
 v3.0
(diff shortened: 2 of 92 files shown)


ischoegl (approved):
Thanks, @speth!

PR #2094, opened 2026-02-28
Extend temperature-based reactors to handle non-ideal phases

speth:

- Update governing equations for `IdealGasReactor` and `IdealGasMoleReactor` to handle non-ideal equations of state
- Introduce new `ThermoPhase::internalPressure` and `getPartialMolarIntEnergies_TV` methods to compute quantities needed for these governing equations. Generic versions rely on implementations of `thermalExpansionCoeff` and `isothermalCompressibility`, while specializations have been added for a few phase types (notably Redlich-Kwong)
- Add `ThermoPhase::getPartialMolarCv_TV` method to implement quantities needed by `AdaptivePreconditioner` in the case of non-ideal equations of state.

I believe that this makes the `Reactor` and `MoleReactor` classes redundant, and expect to remove them in a future PR (and drop the `IdealGas` prefix on these classes). There will also be a follow-on PR to extend this same logic to `ConstPressureReactor` and `ConstPressureMoleReactor`.

The net result of these changes is that you can now use the Redlich-Kwong model with the `IdealGasMoleReactor` class `AdaptivePreconditioner` to greatly speed up reactor calculations using this model. Switching to the temperature-based energy equation eliminates the need for the iterative solver to find the temperature as a function of $u$ and $v$ for every call to the ODE RHS. Based on updating the `non_ideal_shock_tube.py` example to use this model, I get the following average integration times:

- Redlich-Kwong EoS
  - `Reactor` (as before): 6.00 s
  - `IdealGasReactor`: 2.03 s
  - `IdealGasMoleReactor`: 1.97 s
  - Preconditioned `IdealGasMoleReactor`: 1.03 s
- Ideal gas EoS
  - `Reactor` (as before): 0.24 s
  - `IdealGasReactor`: 0.18 s
  - `IdealGasMoleReactor`: 0.18 s
  - Preconditioned `IdealGasMoleReactor`: 0.14 s

**AI Statement (required)**

- **Extensive use of generative AI.** Significant portions of code or documentation were generated with AI, including  logic and implementation decisions. All generated code and documentation were reviewed and understood by the contributor. *ChatGPT Codex 5.3 was used for significant portions of this work*.

**Checklist**


doc/sphinx/reference/reactors/ideal-gas-reactor.md
```diff
@@ -65,26 +67,55 @@ $$ (igr-species)
...
+also eliminates the need to iteratively solve the equation of state to find $T$
+given $u$ and $v$.
+
+In the case of an ideal gas, $\pi_T = 0$ and $\tilde{u}_k$ are equal to the usual
+partial molar internal energies $\hat{u}_k$.
+
+```{seealso}
+- {ct}`ThermoPhase::internalPressure`
+- {ct}`ThermoPhase::getPartialMolarIntEnergies_TV`
+```
```
bryanwweber: Needs a newline to end this file

site_scons/site_tools/UnitsInterfaceBuilder.py
```diff
@@ -27,7 +27,8 @@ def get_property_definition_strings():
         "mean_molecular_weight": '"kg/kmol"', "min_temp": '"K"',
         "molecular_weights": '"kg/kmol"', "partial_molar_cp": '"J/kmol/K"',
         "partial_molar_enthalpies": '"J/kmol"', "partial_molar_entropies": '"J/kmol/K"',
-        "partial_molar_int_energies": '"J/kmol"', "partial_molar_volumes": '"m**3/kmol"',
```
bryanwweber, line 30: It's possible I missed it, I'm on my phone, but I don't see internal pressure added here

bryanwweber (commented):
I'm on vacation and my phone this week, so these are just a couple tiny comments I could see. I'll try to test the implementation and check the docs when I get home if no one else gets to it first

ischoegl (commented):
@speth - this looks overall good. My main question/comment is that this introduces an `IdealGasReactor` with a real gas EoS, which I find slightly counterintuitive. Is there a better way to handle this?

bryanwweber:
> My main question/comment is that this introduces an IdealGasReactor with a real gas EoS

I've always found it somewhat confusing/annoying that a user has to know enough to create an `IdealGasReactor` at all. It seems like we should have enough information to automatically select the appropriate state variable(s), with appropriate escape hatches for advanced users

speth:
As I said in the PR description, the eventual plan here is to remove the existing `Reactor` and `ConstPressureReactor` classes and rename `IdealGasReactor` to `Reactor` and `IdealGasConstPressureReactor` to `ConstPressureReactor`. As far as I can tell, there's not really that much benefit to the enthalpy- or internal energy-based formulation. While the energy equation itself is simpler, that's kind of made up for by the complexity of the iterative solve in `updateState`.

I wanted to let these new implementations co-exist for just a little bit, though, to make sure there aren't any issues that we're overlooking and to facilitate some performance comparisons. I would certainly expect to implement this removal/rename as part of Cantera 4.0.

ischoegl:
> As I said in the PR description, the eventual plan here is to remove the existing `Reactor` and `ConstPressureReactor` classes and rename `IdealGasReactor` to `Reactor` and `IdealGasConstPressureReactor` to `ConstPressureReactor`. As far as I can tell, there's not really that much benefit to the enthalpy- or internal energy-based formulation. While the energy equation itself is simpler, that's kind of made up for by the complexity of the iterative solve in `updateState`.

Yes, I saw the comment, but I believe this is a real opportunity to get rid of some of this clutter:

image

As an aside, if the enthalpy-based formulation is not worthwhile as you said, this would also mean that the `ConstPressureReactor` could be eliminated?

> I wanted to let these new implementations co-exist for just a little bit, though, to make sure there aren't any issues that we're overlooking and to facilitate some performance comparisons. I would certainly expect to implement this removal/rename as part of Cantera 4.0.

Could you create a tracking issue on Cantera/enhancements?

_PS: I'm obviously :+1: with deferring some of this work._

speth:
> Yes, I saw the comment, but I believe this is a real opportunity to get rid of some of this clutter:

There's no "but" here -- reducing the number of reactor models is exactly what I'm describing, and what this PR enables in part. I'll go ahead and write up the Enhancement issue to clarify things.

ischoegl (dismissed):

ischoegl (approved):

PR #2096, opened 2026-03-04
Support compilation for Pyodide

speth:

- Enable sdist compilation targeting Pyodide 
- Add Pixi config file with separate features to compartmentalize requirements for documentation and Pyodide builds
- Add CI job to build Pyodide wheel and run the Python test suite

This is somewhat related to Cantera/enhancements#254, but not aimed at completing that Enhancement. What this does do is show that there are essentially no issues with compiling Cantera and its dependencies as a WASM library.

The wheels targeting Pyodide allow you to run Cantera in your browser. With a few additional steps (manual for the moment), you can set up a Jupyterlite site with the Cantera module and examples available, for example: https://cantera.org/~speth/live-demo/

**AI Statement (required)**

  Significant portions of code or documentation were generated with AI, including logic and implementation decisions. All generated code and documentation were reviewed and understood by the contributor. *ChatGPT Codex 5.3 was used extensively in preparing this PR*.

**Checklist**

### DIFF
--- a/interfaces/python_sdist/build_pyodide_wheel.py
+++ b/interfaces/python_sdist/build_pyodide_wheel.py
@@ -1,0 +1,70 @@
+"""Build a Pyodide-targeted wheel from an existing Cantera Python sdist."""
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+import subprocess
+import tarfile
+import tempfile
+
+
+def _find_latest_sdist(dist_dir: Path) -> Path:
+    candidates = sorted(
+        dist_dir.glob("cantera-*.tar.gz", case_sensitive=False),
+        key=lambda candidate: candidate.stat().st_mtime,
+        reverse=True,
+    )
+    if not candidates:
+        raise FileNotFoundError(f"No Cantera sdist found in '{dist_dir}'.")
+    return candidates[0]
+
+
+def _run(command: list[str], cwd: Path, env: dict[str, str] | None = None) -> None:
+    print(f"+ {' '.join(command)}")
+    subprocess.run(command, cwd=cwd, env=env, check=True)
+
+
+def _extract_sdist(sdist: Path) -> Path:
+    temp_root = Path(tempfile.mkdtemp(prefix="cantera-pyodide-src-"))
+    with tarfile.open(sdist, mode="r:gz") as archive:
+        archive.extractall(path=temp_root)
+    candidates = [path for path in temp_root.iterdir() if path.is_dir()]
+    if len(candidates) != 1:
+        raise RuntimeError(
+            f"Expected exactly one source directory in extracted sdist at '{temp_root}'."
+        )
+    return candidates[0]
+
+
+def main() -> int:
+    repo_root = Path(subprocess.getoutput("git rev-parse --show-toplevel").strip())
+    dist_dir = repo_root / "build" / "python_sdist" / "dist"
+    sdist = _find_latest_sdist(dist_dir)
+    if not sdist.is_file():
+        raise FileNotFoundError(f"Cantera sdist not found: '{sdist}'.")
+
+    outdir = repo_root / "build" / "pyodide_dist"
+    outdir.mkdir(parents=True, exist_ok=True)
+
+    env = dict(os.environ)
+    env["CANTERA_PYODIDE"] = "ON"
+    cmake_args = env.get("CMAKE_ARGS", "").strip()
+    pyodide_cmake_args = "-DCMAKE_CXX_SCAN_FOR_MODULES=OFF"
+    env["CMAKE_ARGS"] = f"{cmake_args} {pyodide_cmake_args}".strip()
+    if "Boost_INCLUDE_DIRS" not in env and (conda_prefix := env.get("CONDA_PREFIX")):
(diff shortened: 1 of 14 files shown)


interfaces/python_sdist/build_pyodide_wheel.py
```diff
@@ -0,0 +1,72 @@
...
+    candidates = [path for path in temp_root.iterdir() if path.is_dir()]
+    if len(candidates) != 1:
+        raise RuntimeError(
+            f"Expected exactly one source directory in extracted sdist at '{temp_root}'."
+        )
+    return candidates[0]
+
+
+def main() -> int:
+    repo_root = Path(__file__).resolve().parents[2]
```
bryanwweber: At work, we like to use `git rev-parse --show-toplevel` to get the repo root rather than magic integers ☺️

interfaces/python_sdist/build_pyodide_wheel.py
```diff
@@ -0,0 +1,72 @@
...
+    outdir.mkdir(parents=True, exist_ok=True)
+
+    env = dict(os.environ)
+    env["CANTERA_PYODIDE"] = "ON"
+    cmake_args = env.get("CMAKE_ARGS", "").strip()
+    pyodide_cmake_args = "-DCMAKE_CXX_SCAN_FOR_MODULES=OFF"
+    env["CMAKE_ARGS"] = f"{cmake_args} {pyodide_cmake_args}".strip()
+    if "Boost_INCLUDE_DIRS" not in env:
+        conda_prefix = env.get("CONDA_PREFIX")
+        if conda_prefix:
```
bryanwweber: If you want to be fancy, I think you can condense this with the walrus operator:

```
    if "Boost_INCLUDE_DIRS" not in env and conda_prefix := env.get("CONDA_PREFIX"):
```

pixi.toml
```diff
@@ -0,0 +1,94 @@
+[workspace]
+channels = ["conda-forge"]
+name = "cantera"
+platforms = ["linux-64", "osx-64", "osx-arm64", "win-64"]
```
bryanwweber: I think we can support `linux-aarch64` too, right?

bryanwweber (commented):
Just a few minor suggestions here! I wonder whether this depends on anything on the `main` branch or whether it'd be a clean backport and we could release 3.3.0 to include this feature?

interfaces/python_sdist/build_pyodide_wheel.py
speth: Also requires some parentheses, but sure.

speth:
> Just a few minor suggestions here! I wonder whether this depends on anything on the `main` branch or whether it'd be a clean backport and we could release 3.3.0 to include this feature?

I think backporting this would be easy, but making a 3.3.0 release just for this feature doesn't seem worth it. I'm more inclined to just make a preview of this interface available based on the current development version.

bryanwweber:
> but making a 3.3.0 release just for this feature doesn't seem worth it

Is there a big difference in effort between a patch release and a minor release? My reasoning was there probably isn't, and since we're already planning 3.2.1, we could jazz it up a little and go for a feature release. Not a big deal either way though!

speth:
Ah yes, I see what you're saying. I was thinking of that being separate from the 3.2.1 patch release, but I suppose there's no reason we couldn't combine those two. Just creates a bit of an impetus to complete this feature so we can resolve the issue with updating the conda-forge package.

speth:
The one part of this that I'm not totally satisfied with is how the Pixi "features" work -- each one creates a separate environment, such that running SCons with different ones results in full rebuilds because it sees that things are coming from different paths. I was hoping not to create a monolithic environment with all of doc building and optional dependencies as the default.

bryanwweber (approved):
Thanks @speth this looks great, tested locally seems to work and pixi is a nice addition. I don't mind having a monolithic environment, it's not that much more time to install (I assume) and that way a dev doesn't have to go look up which environment the want to use.

The only other piece of this is to update the pypi-package build, so I created https://github.com/Cantera/pypi-packages/issues/41 to track that.

lucascolley:
> The one part of this that I'm not totally satisfied with is how the Pixi "features" work -- each one creates a separate environment, such that running SCons with different ones results in full rebuilds because it sees that things are coming from different paths. I was hoping not to create a monolithic environment with all of doc building and optional dependencies as the default.

(randomly happened across this) can you not set a build directory to be shared across multiple environments? In SciPy our equivalent of `scons build` defaults to the `build` subdir of the repo root (https://github.com/scientific-python/spin/blob/aac5fec8537aa6649f67a8268e10773014cd1648/spin/cmds/meson.py#L274-L282) and build tasks in separate environments which we want to share a build directory pass the same directory as an argument: https://github.com/scipy/scipy/blob/d13df77924a38e7c46163462652e0f2c3ce500bf/pixi.toml#L637-L656

If the problem is that you need to have the same versions of shared dependencies across environments, you can use solve groups (https://pixi.prefix.dev/latest/reference/pixi_manifest/#the-environments-table), see e.g. https://github.com/scipy/scipy/blob/d13df77924a38e7c46163462652e0f2c3ce500bf/pixi.toml#L64-L77

PR #2085, opened 2026-02-10
Convert kinetics & equilibrium code to use std::span

speth:
**Changes proposed in this pull request**

This PR continues the work started in #2077 toward implementing https://github.com/Cantera/enhancements/issues/237, updating the classes that are in the `kinetics` and `equil` directories.

**AI Statement (required)**

- **Extensive use of generative AI.**
  Significant portions of code or documentation were generated with AI, including logic and implementation decisions. All generated code and documentation were reviewed and understood by the contributor. *Primarily done using GPT-Codex-5.2 within the Codex extension for VS Code*.

**Checklist**

- [x] The pull request includes a clear description of this code change
- [x] Commit messages have short titles and reference relevant issues
- [x] Build passes (`scons build` & `scons test`) and unit tests address code coverage
- [x] Style & formatting of contributed code follows [contributing guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)
- [x] AI Statement is included
- [x] The pull request is ready for review

PR #2097, opened 2026-03-04
Return variable-length coefficient list from falloff rates

speth:
This restores the behavior from before 3377c037 / #2085 and eliminates a warning about zero values for the Troe "T2" coefficient that were being issued by some tests.

**AI Statement (required)**

**Checklist**


bryanwweber:
Since this is a change to revert to previous behavior, I have to ask if there are tests to check the additional conditionals?

speth:
There was a test for the case that was creating warnings (Troe) but not the other falloff parameterization where we handled a variable number of input coefficients (SRI), so I've added one now. Then I noticed that the Tsang falloff function was inconsistent about whether it required both inputs or not; I've now extended it to allow either 1 or 2 coefficients in all cases.

bryanwweber (approved):
Thanks Ray, looks good to me with the extra tests. Glad you caught the extra reaction type too!

PR #2101, opened 2026-03-20
Thermodynamic properties for a 2-temperature plasma

### LINKED ISSUE #1872: Cantera does not compute a multi-temperature ideal gas law for now
**Problem description**

Hello, and thanks for Cantera! I think I have found a physics bug in the plasma phase, or maybe there is something that I do not understand.

A multi-temperature phase, like the Plasma phase, should take into account electron temperature when computing pressure. I think this is not the case in Cantera, when I compare the formula used in Chemkin.

## Chemkin
In [Chemkin Manual](https://personal.ems.psu.edu/~radovic/ChemKin_Theory_PaSR.pdf), equation 2-4

$$P = \sum_k n_k k_b T_k = k_b T_g \sum_{k \neq e} n_k + n_e k_b T_e$$

## Cantera
In Cantera (electronic temperature is not taken into accont to compute the pressure)

$$P = \sum_k n_k k_b T_g = k_b T_g \sum_{k \neq e} n_k + n_e k_b T_g$$

For the [Ideal Gas Mixture](https://cantera.org/stable/cxx/d7/dfa/classCantera_1_1IdealGasPhase.html) phase, the molar volume of a species is given by the ideal gas law:

$$V_k^o(T, P)=\frac{R T}{P}$$

For the [Plasma](https://cantera.org/stable/cxx/d5/dd7/classCantera_1_1PlasmaPhase.html) phase, there is an [electron pressure](https://cantera.org/stable/cxx/d5/dd7/classCantera_1_1PlasmaPhase.html#af28530a84d4ef9ee301215ce4e07d9e9) defined. However, the class `Plasma` inherits from `Ideal Gas Mixture`, but the method [setPressure](https://cantera.org/stable/cxx/d7/dfa/classCantera_1_1IdealGasPhase.html#a59983094a3e5c4391305b2dd93e83ea0) from the `Ideal Gas Mixture` phase is only inherited, and not redefined to consider the electron pressure.

**Steps to reproduce**

```
import cantera as ct

# Create the plasma object using a default YAML file.
plasma = ct.Solution(
    "example_data/oxygen-plasma-itikawa.yaml",
    "isotropic-electron-energy-plasma",
    transport_model=None,
)

# Initialize the plasma object with a temperature, pressure, and composition.
plasma.TDX = 300, 0.6415, {"O2": 0.5, "E": 0.5}
# Also initialize the electron temperature.
plasma.Te = 300.0

print(
    "Initializing the plasma object:\n"
    + f"T = {plasma.T} K\n"
    + f"rho = {plasma.density} kg/m^3\n"
    + f"X = {plasma.X}\n"
    + f"Te = {plasma.Te} K\n"
)

print(f"Computed (total?) pressure = {plasma.P} Pa")
# >>> 100011.93190795329 Pa, ok
print(f"Computed (electron) pressure = {plasma.Pe} Pa")
# >>> 50005.96595397664 Pa, ok

print("==========================")

# Now, changing the electron temperature.
plasma.Te = 30_000.0

print(
    f"Changing the electron temperature to T_e={plasma.Te} K:\n"
    + f"T = {plasma.T} K\n"
    + f"rho = {plasma.density} kg/m^3\n"
    + f"X = {plasma.X}\n"
)
print(f"Computed (total?) pressure = {plasma.P} Pa")
# >>> 100011.93190795329 Pa, NOT ok?
print(f"Computed (electron) pressure = {plasma.Pe} Pa")
# >>> 5000596.595397665, ok
```

**Behavior**

In first case, gas temperature and electron temperature are set, as well as density and mole fraction (50% O2 and 50 % E). The pressure computed is around 1 atm, which is correct.

Then, in the second case, electron temperature is multiply by a factor 100, leaving gas temperature, density and mole fraction unchanged. However, the pressure is still exactly the same as before, which does not seem correct.

I also think there is confusion between total pressure and gas pressure:

* If `plasma.P` is the total pressure
  - In case 1, $P_{tot}=1atm$ and $P_e = 0.5 atm$ --> , $P_{gas}=0.5 atm$ which is good.
  - In case 2, $P_{tot}=1atm$ and $P_e = 5 atm$ --> $P_{gas}=P_{tot}-P_e=-4.5 atm$, not good...
* If `plasma.P` is the gas pressure
  - In case 1, $P_{gas}=1atm$ and $P_e = 0.5 atm$ -->  $P_{tot}=1.5 atm$ --> not what I expected
  - In case 2, $P_{gas}=1atm$ and $P_e = 5 atm$ --> $P_{tot}=P_{gas}+P_e= 6atm$ , maybe?

**System information**

- Cantera version: 3.1.0
- OS: Windows 11
- Python versions: 3.11.10

pag1pag:
_Note that the following assumes that a temperature can be defined for all species. While this may be true for all heavy species, electrons may not be characterized by an electron temperature (for instance, if the electron energy distribution function (EEDF) is far away from a Maxwellian). It is therefore assumed that the implementation below holds if, for instance, the EEDF is Maxwellian (which may be seen as a strong approximation in plasma). See also https://github.com/Cantera/enhancements/issues/258 for details about the physics behind the code._

The partial pressure is defined as: $P_k = R T_k X_k \rho_m$ (with $\rho_m$ the mixture molar density).
The total pressure is the sum of all partial pressure: $P = \sum_k P_k = R \rho_m \sum_k X_k T_k$
Introducing a **mean temperature**, defined by $\overline{T}=\sum_k X_k T_k$, the total pressure is rewritten $P = R \rho_m \overline{T}$.

The standard concentration (`standardConcentration`) uses this mean temperature: $c^°=\frac{P}{R \overline{T}}$.
The standard volume (`getStandardVolumes` and `getPartialMolarVolumes`) is $v_k = \frac{R T_k}{P}$.

- Define or Redefine for a 2-temperature system:
  - Molar Thermodynamic Properties: `enthalpy_mole`, `entropy_mole`, `gibbs_mole`, `intEnergy_mole`
  - Mechanical Equation of State: `meanTemperature`, `pressure`, `setPressure`, `electronPressure`
  - Chemical Potentials and Activities: `standardConcentration`
  - Partial Molar Properties: `getChemPotentials`, `getPartialMolarEnthalpies`, `getPartialMolarIntEnergies`, `getPartialMolarVolumes`
  - Properties of the Standard State of the Species: `getStandardChemPotentials`, `getStandardVolumes`
  - Thermodynamic Values for the Species Reference: `getGibbs_ref`, `getStandardVolumes_ref`
- Update the `setState` method to account for additional electron temperature. Also introduce `setState_TgTeP` and `setState_TgTeD` to set electron temperature along with other quantities.
- Looking now at consistency tests, all of them pass when the electron temperature is equal to gas temperature. New cases were added with different electron and gas temperatures. The skipped tests are due to temperature differential, which are hard to define when there are multiple temperature (When $T_e \neq T_g$, $c_p=\frac{dh}{dT}$ may not be correct: is $dT$ equal to $dT_h$ or $dT_e$ ?). Some Python tests required to update the electron temperature after setting TPX, and doing that gives the expected value.
- Refactor and document plasma thermo files by regrouping related method in the same section (no change of code, just move some parts for better lisibility)

Closes #1872 

```python
import cantera as ct

# Create the plasma object using a default YAML file.
plasma = ct.Solution(
    "example_data/oxygen-plasma-itikawa.yaml",
    "isotropic-electron-energy-plasma",
    transport_model=None,
)

# Initialize the plasma object with a temperature, pressure, and composition.
plasma.TDX = 300, 0.6415, {"O2": 0.5, "E": 0.5}
# Also initialize the electron temperature, at the same temperature as the heavy particles.
plasma.Te = 300.0

print(
    "Initializing the plasma object:\n"
    + f"T = {plasma.T} K\n"  # >>> 300 K, ok
    + f"rho = {plasma.density} kg/m^3\n"  # >>> 0.6415 kg/m^3, ok
    + f"X = {plasma.X}\n"  # >>> [0.5 0.  0.5 0.  0.  0. ], ok
    + f"Te = {plasma.Te} K\n"  # >>> 300 K, ok
)

print(f"Computed pressure = {plasma.P} Pa")
# >>> 100011.93190795329 Pa, ok
print(f"Computed (electron) pressure = {plasma.Pe} Pa")
# >>> 50005.96595397664 Pa, ok

print("==========================")

# Now, changing the electron temperature.
plasma.Te = 30_000.0

print(
    f"Changing the electron temperature to T_e={plasma.Te} K:\n"
    + f"T = {plasma.T} K\n"  # >>> 300 K, ok
    + f"rho = {plasma.density} kg/m^3\n"  # >>> 0.6415 kg/m^3, ok
    + f"X = {plasma.X}\n"  # >>> [0.5 0.  0.5 0.  0.  0. ], ok
    + f"Te = {plasma.Te} K\n"  # >>> 30000 K, ok
)
print(f"Computed (total) pressure = {plasma.P} Pa")
# >>> 5050602.561351642 Pa, ok
print(f"Computed (electron) pressure = {plasma.Pe} Pa")
# >>> 5000596.595397665, ok
# The total pressure is now dominated by the electron pressure, as expected.

```

**AI Statement (required)**

- **No generative AI was used.** This contribution was written entirely without AI
  assistance.

**Checklist**


pag1pag:
Hello !

I have reread the [Contributing guide](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md), and I just seen that I have used **merge** instead of **rebase** in this work. Is it still possible to use `git rebase`, or is it too late?

Also, I have opened an [enhancement request](https://github.com/Cantera/enhancements/issues/258), to discuss the physics behind the code.

pag1pag:
As for the CI/CD tests,

Locally, after running `scons build` and `scons test`, I see that two tests are failing, but they do not seem related to this PR:
```
Tests passed: 7378
Up-to-date tests skipped: 0
Tests failed: 2

Failed tests:
    - python: test_reactor.TestMoleReactor.test_tolerances
    - cxx-openmp-ignition
```

As for the examples, `samples\python\thermo\plasma-eedf.py` seems to be failing [online](https://github.com/Cantera/cantera/actions/runs/25922799850/job/76197882434). It seems to be due to the `UserWarning` being raised during the execution of the file, but I do not understand why the code crashes on a warning.

speth:
> I have reread the [Contributing guide](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md), and I just seen that I have used **merge** instead of **rebase** in this work. Is it still possible to use `git rebase`, or is it too late?

I think you can rebase your local branch on top of our `main` and it will eliminate the merge commit. Once that is done, you can do a "force push" to this PR.

One other thing: in commit 41ea41e1241959e632bfc29c976e57bc9797cb70, the files `consistency.cpp` and `thermo.pyx` got converted to Windows-style line endings, which makes them show up as completely changed files and it's impossible to see the actual code changes (if any). Could you please fix those files? The best way would be to do so as part of an interactive rebase so the erroneous change never occurs. Otherwise, we can just squash the PR later if you don't want to worry about the detailed commit history.

pag1pag:
Hello, I have rebased, rewrote the history and force push.
I hope it is clearer now :D

include/cantera/thermo/PlasmaPhase.h
```diff
@@ -430,6 +349,456 @@ class PlasmaPhase: public IdealGasPhase
...
+     * For an ideal gas mixture with electrons at a different temperature
+     * than the heavy species, the molar internal energy is calculated as:
+     * @f[
+     * \hat u(T,T_\text{e},P)
+     *  = \sum_{k} X_k \tilde{u}_k(T_k)
+     *  = \sum_{k} X_k (\tilde{h}_k(T_k) - R T_k)
+     * @f]
+     * where heavy-species properties are evaluated at @f$ T @f$, electron properties
+     * at @f$ T_\text{e} @f$, and @f$ P @f$ is the total pressure of the mixture.
+     * For an ideal gas, enthalpy is indepedant of pressure.
```
speth:
```suggestion
     * For an ideal gas, internal energy is independent of pressure.
```

include/cantera/thermo/PlasmaPhase.h
```diff
@@ -430,6 +349,456 @@ class PlasmaPhase: public IdealGasPhase
...
+     *  \hat h(T, T_\text{e}, P)
+     *  &= \sum_{k} X_k \tilde{h}_k(T_k)
+     *   = \sum_{k} X_k h^\text{ref}_k(T_k) \\
+     *  &= \sum_{k \neq e} X_k h^\text{ref}_k(T)
+     *     + X_\text{e} h_\text{e}^\text{ref}(T_\text{e})
+     * \end{align}
+     * @f]
+     * where heavy-species properties are evaluated at @f$ T @f$, electron properties
+     * at @f$ T_\text{e} @f$, and @f$ P @f$ is the total pressure of the mixture.
+     * For an ideal gas, enthalpy is indepedant of pressure.
```
speth:
```suggestion
     * For an ideal gas, enthalpy is independent of pressure.
```

interfaces/cython/cantera/thermo.pyx
```diff
@@ -1889,6 +1889,15 @@ cdef class ThermoPhase(_SolutionBase):
             span[double](&data_levels[0], data_levels.size),
             span[double](&data_dist[0], data_dist.size))
 
+    property mean_temperature:
+        """Get the mean temperature of the plasma [K].
+        This is a mole-weighted average of the electron and heavy species temperatures.
+        """
```
speth, line 1901: Add here:
```
.. versionadded:: 4.0
```

test/thermo/thermoToYaml.cpp
```diff
@@ -572,8 +572,8 @@ TEST_F(ThermoYamlRoundTrip, CoverageDependentSurface)
 TEST_F(ThermoYamlRoundTrip, IsotropicElectronEnergyPlasma)
 {
     roundtrip("oxygen-plasma.yaml", "isotropic-electron-energy-plasma");
-    skip_cp = true; // Not implemented for PlasmaPhase
-    skip_entropy = true; // Not implemented for PlasmaPhase
+    skip_cp = false; // Not implemented for PlasmaPhase
+    skip_entropy = false; // Not implemented for PlasmaPhase
```
speth: These default to `false`, so presumably these lines can just be deleted now.

include/cantera/thermo/PlasmaPhase.h
```diff
@@ -105,19 +105,50 @@ class PlasmaPhase: public IdealGasPhase
 
     void initThermo() override;
 
+    // ================================================================= //
+    // ================================================================= //
```
speth: I'd prefer to leave out these decorative dividers; the named method groups can stay.

include/cantera/thermo/PlasmaPhase.h
```diff
@@ -83,11 +83,11 @@ class ElectronCollisionPlasmaRate;
...
 {
 public:
-    //! Construct and initialize a PlasmaPhase object
-    //! directly from an input file. The constructor initializes the electron
-    //! energy distribution to be Druyvesteyn distribution (m_x = 2.0). The initial
-    //! electron energy grid is set to a linear space which starts at 0.01 eV and ends
-    //! at 1 eV with 1000 points.
+    //! Construct and initialize a PlasmaPhase object directly from an input file.
+    //! The constructor initializes the electron energy distribution to be a
+    //! Maxwellian distribution (m_isotropicShapeFactor = 1.0).
```
speth:
```suggestion
    //! Maxwellian distribution (#m_isotropicShapeFactor = 1.0).
```

include/cantera/thermo/PlasmaPhase.h
```diff
@@ -430,6 +349,456 @@ class PlasmaPhase: public IdealGasPhase
...
+     * For an ideal gas mixture with electrons at a different temperature
+     * than the heavy species, the molar enthalpy is calculated as:
+     * @f[
+     * \begin{align}
+     *  \hat h(T, T_\text{e}, P)
+     *  &= \sum_{k} X_k \tilde{h}_k(T_k)
+     *   = \sum_{k} X_k h^\text{ref}_k(T_k) \\
+     *  &= \sum_{k \neq e} X_k h^\text{ref}_k(T)
+     *     + X_\text{e} h_\text{e}^\text{ref}(T_\text{e})
+     * \end{align}
```
speth, line 401: I know we aren't broadly consistent about this, but I don't think we need a different decoration for mixture molar enthalpy and partial molar enthalpy. My preference would be to keep $\hat{h}_k$ as the partial molar enthalpies, as before. Likewise, I'm not sure it's worth the churn on changing the notation for the standard-state from superscript `0` to superscript `ref` here. I think it would be better to stay notationally consistent with `IdealGasPhase` so the meaningful differences stand out more clearly.

src/thermo/PlasmaPhase.cpp
```diff
@@ -695,100 +804,242 @@ void PlasmaPhase::getPartialMolarIntEnergies(span<double> ubar) const
...
+        if (m_distributionType == "isotropic") {
+            setElectronTemperature(t);
+        } else {
+            // Warning for other distribution types.
+            warn_user("PlasmaPhase::setState_TP",
+                "The electron temperature is not set equal to the gas temperature "
+                "when using '{}' electron energy distribution. "
+                "This may not be intended.", m_distributionType);
+        }
+        setPressure(p);
```
speth: I'm not sure this is the best choice for how to handle setting T for either the isotropic or full EEDF-based cases. In the latter case, clearly we should not set the EEDF based on the gas temperature. But then there seems to be no way to set the temperature of the gas without triggering this warning (as now seen in various tests and the examples). On the other hand, it also seems strange to reset the electron state by setting T only in the case of an isotropic distribution.

My intuition would be that we should leave the electron state alone when setting T -- let that just be the gas temperature, and have that be true regardless of how the EEDF is specified. That also seems more consistent with introduction of the `meanTemperature` property.

include/cantera/thermo/PlasmaPhase.h
```diff
@@ -430,6 +349,456 @@ class PlasmaPhase: public IdealGasPhase
...
+    //! Set the state using an AnyMap containing any combination of properties
+    //! supported by the thermodynamic model
+    /*!
+     * Accepted keys are:
+     * * `X` (mole fractions)
+     * * `Y` (mass fractions)
+     * * `T` or `Tg` or `gas-temperature` [K]
+     * * `Te` or `electron-temperature` [K]
+     * * `P` or `pressure` [Pa]
+     * * `D` or `density` [kg/m^3]
```
speth, line 768: I wonder if we can support the full set of options allowed at the `ThermoPhase` level, and avoid reimplementing much of that method, by simply setting the electron temperature first and then calling the base class method.

src/thermo/PlasmaPhase.cpp
```diff
@@ -128,6 +289,11 @@ void PlasmaPhase::setIsotropicElectronEnergyDistribution()
 }
 
 void PlasmaPhase::setElectronTemperature(const double Te) {
+    // If the electron temperature is negative, throw an error.
+    if (Te < 0.0) {
+        throw CanteraError("PlasmaPhase::setElectronTemperature",
+            "Electron temperature cannot be negative.");
+    }
```
speth: This guard should apply to `setMeanElectronEnergy` as well, which can be achieved by having that method call `setElectronTemperature` with the computed value.

include/cantera/thermo/PlasmaPhase.h
```diff
@@ -430,6 +349,456 @@ class PlasmaPhase: public IdealGasPhase
...
+
+    // Below are 5 methods already defined in IdealGasPhase, that do not need
+    // to be overridden, since they do not depend on the temperature.
+    // {
+    // void getEnthalpy_RT_ref(span<double> hrt) const override;
+    // void getGibbs_RT_ref(span<double> grt) const override;
+    // void getEntropy_R_ref(span<double> er) const override;
+    // void getIntEnergy_RT_ref(span<double> urt) const override;
+    // void getCp_R_ref(span<double> cprt) const override;
+    // }
```
speth: I think this description is a bit misleading -- the reason these methods do not need to be overridden is the same as for `getEnthalpy_RT` and friends: because the electron temperature correction is handled by `updateThermo`. More generally, I think it would be more clear to document this all in the active sense as documentation for the `updateThermo` method, explaining that this method handles getting properties at the correct temperature for each species, rather than explaining the things we don't have to do here.

speth (changes requested):
Thanks for working on this, @pag1pag. Overall I think this is a significant improvement in the consistency of how we handle the thermo properties when there are two distinct temperatures. However, there is still one discrepancy that we need to resolve, which involves the definitions of the standard concentration, the activities, and the species standard molar volumes. The implementation here defines a single value for the standard concentration, $C^o \equiv P / R \bar{T}$, and specifies that the activities are equal to the mole fractions, $a_k \equiv X_k$. However, this is not consistent with the definition of the partial molar volume $V^o_k \equiv R T_k / P$ when $T_e \ne T_g$.

Since the current kinetics implementation relies on there being only a single standard concentration for a phase, I think the best fix for this is to define the activity for the electron as $a_e \equiv X_e T_e / T$ while leaving $a_k = X_k$ for the gas phase species. This is approximate, but correct in the limit of small $X_e$, since $\bar{T} \approx T_g$.

I'm not quite sure what the value of introducing the methods `setState_TgTeP` and `setState_TgTeD` is, over just calling `setElectronTemperature` followed by `setState_TP` or `setState_TD` (other than the fact that the latter currently overwrite the electron temperature). If these methods left the electron temperature alone, I think we could avoid adding these specialized methods (which are difficult to access in any case).

As an editorial note, the combination of formatting updates (specifically, rearranging the methods in both the header and implementation) with substantive code changes made this quite challenging to review. It's very difficult to identify where there are actual implementation changes versus just moving functions around. I would strongly recommend not mixing such changes in a single PR in the future.

pag1pag:
Thanks for the review @speth !


1. I am not sure to see the discrepancy with the definitions of the standard concentration, the activities, and the species standard molar volumes. 

- For the partial molar volume, from total pressure, $V=\frac{R}{P} \sum_k n_k^\text{mol} T_k$, and with the definition of the molar volume, we have $V_{m, k}^o = \left( \frac{\partial V}{\partial n_k^\text{mol}} \right)_{P, T_j,  n_j \neq k } = \frac{R T_k}{P}$
- The activity is chosen as: $a_k \equiv X_k$ to be equivalent with the ideal gas (I guess it comes from the implementation of the chemical potential $\mu_k(T_k, P, X_k)=\mu_k^o(T_k)+R T_k \ln \left(a_k(T_k, P, X)\right)$ )
- From what I have understood, the activity is also equal to $a_k = C^a_k / C^0_k$. Here, $C^a_k$ are the activity coefficients (defined in IdealGasPhase to be equal to the concentration of species $k$, so here, $C^a_k=X_k C_\text{tot}=(X_k P)/(R \bar{T})$ ). Therefore, assuming the previous equalities hold, $C^0_k=P/(R \bar{T})$.


2. Regarding `setState_TgTeP` and `setState_TgTeD`, it was to avoid ambiguity, so that users are sure of what will be the gas temperature, electron temperature and pressure/density.
I have now removed these methods, and done as you suggested, i.e. `setElectronTemperature` followed by `setState_TP`/`setState_TD`. Note that however, doing the inverse (i.e., `setState_TP` and then `setElectronTemperature`) is changing the pressure to a different value that the one set in `setState_TP`. This is something worth mentioning in the documentation I suppose.

3. As for the editorial note, I am sorry. I thought it was clearer this way. I can revert these changes, and bring them in a new PR if you want.

src/thermo/PlasmaPhase.cpp
pag1pag: I have now removed this function, i.e. `setState_TP` comes from `IdealGasPhase`, which only set the gas temperature and the pressure.

And to work correctly, the electron temperature must be set **before** setting `TP`/`TD` (to avoid changing the pressure/density).

include/cantera/thermo/PlasmaPhase.h
pag1pag: Understanble and done.

include/cantera/thermo/PlasmaPhase.h
pag1pag: I was following the notation used [here](https://github.com/Cantera/cantera/blob/83ebf9d56687e8143f3b2262972206eb0a98bf60/include/cantera/thermo/IdealGasPhase.h#L57) in the IdealGasPhase.

I will change to be consistent with the equivalent method of IdealGasPhase

speth:
> 1. I am not sure to see the discrepancy with the definitions of the standard concentration, the activities, and the species standard molar volumes.

Right, what you say here is what the implementation does. My point is that the standard concentrations should just be the inverse of the standard molar volumes, but this is not the case for the current implementation. However, I do think there is a fix that makes this reasonably consistent for any realistic electron mole fraction.

> 2. [...] Note that however, doing the inverse (i.e., `setState_TP` and then `setElectronTemperature`) is changing the pressure to a different value that the one set in `setState_TP`. This is something [worth] mentioning in the documentation I suppose.

Yes, it is certainly worth noting this requirement to set the electron temperature first if you want to properly constrain the pressure.

Regarding (3), no, please don't revert the organizational changes at this point. I agree that the end result is an improvement, it was just difficult to review.

pag1pag:
It seems like one line of code disappeared when I reorganized commits before pushing them; i.e., `IdealGasPhase::setState(state)` was never called in `PlasmaPhase::setState`, resulting in multiple test failures. It is now fixed.

I have also rebased and force pushed.


As for the discrepancy, I think I am missing something.
1. The definition and implementation of the partial molar volume is: $V_{m, k}^o = \frac{R T_k}{P}$ (`getPartialMolarVolumes`).
2. From this, the total molar volume should be deduced via $V_m = \sum_k X_k V_{m, k}^o = \frac{R \overline{T}}{P}$ (`?`).
3. The total concentration is $C_\text{tot}=\frac{P}{R \overline{T}}$, which is the inverse of the molar volume (`standardConcentration` ?).

Maybe I am confusing notation, but should the method `getStandardVolumes` compute what I have called the 'total molar volume'? Because if that is true, then indeed, there is a mistake in the code, as `Plasmaphase::getStandardVolumes` actually returns $\frac{R T_k}{P}$.
I suppose it is the notion of standard state that I am confused about, so could you enlighten me, please?

speth:
The issue with the standard concentrations comes from how they affect the calculation of equilibrium constants and therefore reverse rate constants. Here's an example that shows that with the current implementation in this PR, we have the problematic behavior that for even reactions not involving electrons, the reverse rates change when the electron temperature changes:
```py
import cantera as ct

phase_def = """
units: {length: cm, time: s, quantity: mol, activation-energy: cal/mol}
phases:
- name: plasma
  thermo: plasma
  species:
  - example_data/oxygen-plasma-itikawa.yaml/species: [e]
  - nasa_gas.yaml/species: [O, O2, O2-, O-, O2+]
  kinetics: gas
  state: 
    T: 2000
    Te: 5000
    P: 101325
    X: {O2: 0.7999, O2+: 0.1, e: 0.1, O: 0.0001}

reactions:
- equation: 2 O + M <=> O2 + M
  rate-constant: {A: 1.2e+17, b: -1.0, Ea: 0.0}
"""

gas = ct.Solution(yaml=phase_def)

def info(gas):
    # force cache clearing
    TD = gas.TD
    gas.TD = 300, 0.5
    gas.forward_rate_constants
    gas.TD = TD

    print(f"\n{gas.T=}, {gas.Te=}, {gas.mean_temperature=}")
    print(f"{gas.concentrations}")
    print(f"{gas.forward_rates_of_progress=}")
    print(f"{gas.reverse_rates_of_progress=}")
    print(f"{gas.equilibrium_constants=}")

info(gas)

gas.Te = 2000  # change state but with total density / concentrations constant
info(gas)
```
which outputs:
```
gas.T=2000.0, gas.Te=5000.0, gas.mean_temperature=2300.0
[5.29852016e-04 5.29852016e-07 4.23828628e-03 0.00000000e+00
 0.00000000e+00 5.29852016e-04]
gas.forward_rates_of_progress=array([8.92513974e-08])
gas.reverse_rates_of_progress=array([3.19621581e-06])
gas.equilibrium_constants=array([4.2156062e+08])

gas.T=2000.0, gas.Te=2000.0, gas.mean_temperature=2000.0
[5.29852016e-04 5.29852016e-07 4.23828628e-03 0.00000000e+00
 0.00000000e+00 5.29852016e-04]
gas.forward_rates_of_progress=array([8.92513974e-08])
gas.reverse_rates_of_progress=array([3.67564818e-06])
gas.equilibrium_constants=array([3.66574452e+08])

```

I believe the way to handle this self-consistently is:
- Define `standardConcentration` as $P/(R T_g)$ for all species; This will make $K_c$ for heavy-species reactions independent of $T_e$.
- Keep `getActivityConcentrations` as is, returning the true number density for each species
- Consistent with the above definitions, `getActivityCoefficients` should return $\gamma_k = T_g/ \bar T$
- Likewise, `getActivities` should return $a_k = X_k T_g/\bar T$ (where $X_k$ can be computed as `m_ym[k] * m_mmw`).

So, the only slightly odd behavior here is that the activity coefficients are slightly different from 1, but all the fundamental thermo properties and kinetics behave correctly.

pag1pag:
Thanks for the explanation!

I have updated `standardConcentration` and `getActivities` as you suggested, and it fixed the issue with the change of equilibrium constants and reverse reactions.

However, I encountered an issue with the test `chem_potentials_to_activities` when electron and gas temperatures are not equal. I guess it is not that important since (from what I gathered) the activity and standard concentration are only used to compute equilibrium constants, which implies thermal equilibrium.

include/cantera/thermo/PlasmaPhase.h
erwanp: ( @pag1pag if you hadn't seen the comment )

pag1pag: As suggested, `PlasmaPhase::setState` now first sets the electron temperature and then calls `IdealGasPhase::setState` (which itself is `ThermoPhase::setState`).

include/cantera/thermo/PlasmaPhase.h
speth: As long as this is consistent with something else, I think that's OK. Aligning notation more broadly in Cantera is a task for another day...

speth (approved):
Thanks for the updates, @pag1pag. I added some documentation with the caveats about the remaining, narrower consistency issues and fixed a couple of unrelated CI issues.

pag1pag:
Thanks @speth for all! :D

PR #2109, opened 2026-05-01
Fix convergence failures in element_potential solver

### LINKED ISSUE #154: ChemEquil solver fails or generates SIGFPE
```
Certain initial conditions cause unexpected failures in the ChemEquil 
equilibrium solver. For example, with the attached mechanism:

    import Cantera
    gas = Cantera.IdealGasMix('chemC2H4_100.xml')

    def f(i, Yf):
        gas.set(P=100002, T=299.999, Y='C2H4:%f, N2:0.594407, O2:0.179732' % Yf)
        gas.equilibrate('HP', i)

f(0, 0.225861) # Fails after exceeding the maximum number of iterations
f(1, 0.225861) # Succeeds
f(2, 0.225861) # Succeeds
f(0, 0.226861) # Succeeds

In addition, if trapping of floating point exceptions is enabled, e.g. by 
placing a call to:

    #include <fenv.h>
    feenableexcept(FE_OVERFLOW | FE_UNDERFLOW);

Somewhere before the call to equilibrate (I put it in the constructor of Phase, 
just to test), an exception will be caught in ChemEquil::equilibrate (approx. 
line 876 of ChemEquil.cpp), when trying to compute:

    jac.leftMult(DATA_PTR(res_trial), DATA_PTR(grad));

Where 'jac' looks like:

-3.4455e-08,  5.0889e-09,  1.7723e-08, 4.3784e+147, -3.8413e-06
 1.7243e-07, -3.0886e-07, -3.8914e-07, 6.0754e+154,  3.5798e-05
 2.0702e-07, -2.2973e-07, -5.4029e-07, 6.6276e+154,  4.3018e-05
 0.0000e+00,  5.1571e-08,  5.4633e-08, 1.0121e+148, -3.0169e-06
 1.6591e-06, -1.1083e-06, -3.0991e-06, 3.4299e+155,  1.1981e+01

The overflow is caused by the entries in the 4th column.

I think this has something to do with the fact that N2 is the only 
nitrogen-containing species defined in the mechanism, and N is the 4th element 
defined. This doesn't explain why the error only occurs for certain mixture 
compositions, though.
```

Original issue reported on code.google.com by `yarmond` on 29 Mar 2013 at 5:30

Attachments:
- [chemC2H4_100.xml](http://cantera.org/issues/154/chemC2H4_100.xml)

speth:

It turns out that the equilibrium solver errors in #154 were entirely due to cases where the equilibrium temperature was beyond the upper temperature limit of the phase's thermodynamic data. The solver was set up to limit its Newton steps to keep $T < T_{max} + 25$. Once it reached that limit, the step size would be reduced to zero and the solver would just take steps of that size until it hit the iteration limit. The error message just indicated that the solver had terminated after reaching an iteration limit, hiding the fact that the real issue is that the solution was outside the enforced bounds.

- Introduce a flag on `ThermoPhase` objects indicating whether temperature limits are enforced. By default, they are not.
- Check for temperature limits being exceeded in the `element_potential` equilibrium solver. If bounds are not enforced, issue a warning; if they are, raise an exception. 
- Check for temperature limits being exceeded in the `ThermoPhase` state setters for HP, UV, SV, and SP property pairs and either warn or raise an exception.
- Clean up some dead code in the element potential solver

- Fixes #154 
- Partially addresses Cantera/enhancements#229

```python
import cantera as ct
gas.TPX = 500, 101325, 'C2H4:0.5, N2:0.1, O2:0.9'
gas.equilibrate('HP', solver='element_potential')

gas.enforce_temperature_limits = True
gas.TPX = 500, 101325, 'C2H4:0.5, N2:0.1, O2:0.9'
gas.equilibrate('HP', solver='element_potential')
```
Outputs:
```
UserWarning: ChemEquil::equilibrate: Temperature (3134.737713989301 K) outside valid range of 300 K to 3000 K
  gas.equilibrate('HP', solver='element_potential')

CanteraError: 
CanteraError thrown by ChemEquil::equilibrate:
Equilibration with the 'HP' property pair failed because the solver reached the upper temperature bound of 3000 K. The target enthalpy is 824610.7587369331, but the current value is -1132537.4275217229 at T = 3000.0024019112298 K. The enforced temperature bounds are 300 K to 3000 K. Disable temperature-limit enforcement to allow extrapolation beyond this range.
```

More broadly, all the failing cases in the example given in #154 now succeed, including at alternative initial conditions like P = 5 atm and T = 1200 K and with a sample size of 5000 cases.

**AI Statement (required)**

- **Extensive use of generative AI.**  Significant portions of code or documentation were generated with AI, including
  logic and implementation decisions. All generated code and documentation were  reviewed and understood by the contributor. Issue investigated and fixes implemented using Codex / GPT-5.5.

**Checklist**


src/equil/ChemEquil.cpp
```diff
@@ -37,7 +37,47 @@ int _equilflag(const char* xy)
...
+                                             double currentT, double Tmin,
+                                             double Tmax, int boundDirection)
+{
+    string bound = boundDirection > 0 ? "upper" : "lower";
+    double Tbound = boundDirection > 0 ? Tmax : Tmin;
+    throw CanteraError("ChemEquil::equilibrate",
+        "Equilibration with the '{}' property pair failed because the solver "
+        "reached the {} temperature bound of {} K. The target {} is {}, but "
+        "the current value is {} at T = {} K. The enforced temperature bounds "
+        "are {} K to {} K. Disable temperature-limit enforcement to allow "
```
bryanwweber, line 75: I'd suggest adding something here to say that allowing extrapolation may result in solver failures

speth: Where would such failures come from? I agree that the solution might be inaccurate if the temperature bound is exceeded. The default behavior issues a warning about the temperature limit being exceeded, rather than this message which is only printed if the user has explicitly set `Solution.enforce_temperature_limits = True`.

bryanwweber: I thought the issue at hand was that the solver errored out because the thermo properties became nonsensical? Maybe I misunderstood, what I'm suggesting is adding something here so if this error is printed, users don't think flipping the bool to allow extrapolation will be a panacea. I'm imagining the first year grad student who inherited a script from the postdoc and doesn't understand the tradeoff.

speth: Sorry, my description was unclear, an I've tried updating it above to explain the underlying issue better. The solver was failing _because_ it was enforcing temperature limits, but the error message didn't provide any information that these were the problem. In many cases, the equilibrium values are only slightly beyond the max temperature of 3000 K (say 3100 K or so), where the NASA polynomials aren't going to be that badly behaved.

I opted to change the default behavior to allow extrapolation with a warning, rather than continue to raise an exception. This may change further with a complete implementation of the ideas proposed in Cantera/enhancements#229.

bryanwweber: Thanks Ray that clears it up!

.github/workflows/main.yml
```diff
@@ -619,21 +619,72 @@ jobs:
...
           path: build/pyodide_dist/cantera-*.whl
           name: cantera-wheel-pyodide
           if-no-files-found: error
           retention-days: 14
+      - name: Create Pyodide venv
+        run: |
+          pixi run -e pyodide pyodide venv .pyodide-venv
+      - name: Install wheel and dependencies into Pyodide venv
+        run: |
+          WHEEL=$(ls build/pyodide_dist/cantera-*.whl)
```
bryanwweber: Instead of using `ls` here, what if the file is directly discovered by pathlib in the installer script below? I don't think parsing the output of ls is generally recommended and I don't think it's needed here.

ischoegl (commented):
@bryanwweber - since you had comments on this PR, I'll defer to your approval.

test/thermo/ThermoPhase_Test.cpp
```diff
@@ -117,6 +117,114 @@ TEST_F(TestThermoMethods, setConcentrations)
...
+    thermo->setState_TP(thermo->minTemp() + 1.23, OneAtm);
+    thermo->setState_HP(h, OneAtm);
+
+    EXPECT_NEAR(thermo->temperature(), T, 1e-5);
+}
+
+TEST_F(EnforcedTemperatureLimits, setStateHPClipsToLowerTemperatureLimit)
+{
+    double T = thermo->minTemp() + 0.01;
+    thermo->setState_TPX(T, OneAtm, "O:1");
```
bryanwweber, line 148: question: is this meant to be `O2:1`? I don't think it changes the test, just for consistency

bryanwweber (changes requested):
two small changes requested, @speth! the actual work looks great 😄

test/thermo/ThermoPhase_Test.cpp
speth: No, actually. For whatever reason, the only species I can find to trigger this case are O and OH, at least without searching more broadly than the species in `h2o2.yaml.

bryanwweber (dismissed):
There's a lot of CI failures, but I've lost track of what's normal failures at this point so I'm going to approve

speth:
Well, this PR was meant to fix the Pyodide issue. I resolved that failure and I think the only remaining ones should be the "Windows 2022" runner ones that I haven't had a chance to investigate yet (but are certainly unrelated to this PR).

bryanwweber (approved):

.github/workflows/main.yml
```diff
@@ -619,21 +619,71 @@ jobs:
...
+          loadPackage(["pytest", "pandas"])
+          PY
+
+          # Download pure-Python packages and their dependencies with host pip.
+          # The Pyodide venv pip wrapper can fail with newer pip / urllib3
+          # combinations before it starts dependency resolution.
+          pixi run -e pyodide python -m pip download --only-binary=:all: \
+            --dest "${WHEELHOUSE}" pint graphviz ruamel.yaml packaging
+
+          export WHEEL WHEELHOUSE
```
bryanwweber, line 652: I just noticed that we don't need to export wheel here. Feel free to address at another convenient time if you want 😁

speth: I'm just going to hit "merge" on this and save that for an upcoming pile of miscellaneous updates.

PR #2117, opened 2026-05-14
Fix mis-named/referenced n-heptane mechanism

speth:
This resolves an issue noted in [this Users' Group post](https://groups.google.com/g/cantera-users/c/o26ziBHEecU) that the conditions in the example `continuous_reactor.py` did not correspond to the conditions in the referenced paper ([10.1016/j.combustflame.2015.08.001](https://doi.org/10.1016/j.combustflame.2015.08.001)). It was observed that the figure in the paper showing results for n-heptane combustion was at an equivalence ratio of 1.0 while the conditions in `continuous_reactor.py` were for a mixture with equivalence ratio 2.0.

However, it turns out that the mechanism used in the example is actually not from this paper, which was mainly on n-hexane combustion, but from another paper by many of the same authors on n-heptane combustion ([10.1016/j.combustflame.2016.06.028](https://doi.org/10.1016/j.combustflame.2016.06.028)), and is incorrectly named as being the n-hexane mechanism. In addition, the supplementary material from this second paper includes a spreadsheet with the exact experimental data plotted in `continuous_reactor.py`. 

_Open question_: one thing I'm not quite decided on is whether to just go with the simple rename of the mechanism file (as implemented here so far) or whether to introduce a shim file with the old name that issues a deprecation warning when you try to use it. That would avoid users getting a "file not found" error that doesn't explain what they need to do.

- Rename `n-hexane-NUIG-2015.yaml` to `n-heptane-NUIG-2016.yaml`
- Update examples using this mechanism to specify `NC7H16` as the fuel instead of `NC6H14`
- Fix `windows-2022` CI errors by pinning `micromamba` version

**AI Statement (required)**

**Checklist**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -779,7 +779,7 @@ jobs:
           -exec sh -c 'for n; do echo "$n" | tee -a results.txt && python3 -u "$n" >> results.txt || exit 1; done' sh {} +
         env:
           # The pyparsing ignore setting is due to a new warning introduced in Matplotlib==3.6.0
-          # Ignore NasaPoly2 warnings from n-hexane-NUIG-2015.yaml
+          # Ignore NasaPoly2 warnings from n-heptane-NUIG-2016.yaml
           PYTHONWARNINGS: "error,ignore:warn_name_set_on_empty_Forward::pyparsing,ignore:datetime.datetime.utcfromtimestamp:DeprecationWarning:,ignore:NasaPoly2:UserWarning:"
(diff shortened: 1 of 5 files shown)


ischoegl (commented):
I believe it's already decided by the shim in `cantera-example-data`, which uses existing machinery (to my recollection, the 'deprecated' field triggers an automatic warning). I believe this is the correct approach, and we can remove the shim at a later point.

ischoegl (approved):
Thanks, @speth!

PR #2130, opened 2026-06-06
Make `SolutionArray::getState` reliable and read-only

### LINKED ISSUE #2067: `SolutionArray.getState(loc)` returns the state of its `Solution` object rather than the state found at `loc`
**Problem description**

In the C++ API, calling `SolutionArray.getState(loc)` returns the state of the associated `Solution` object rather than the state contained at `loc` if `loc` was the last-accessed location in the array. 

**Steps to reproduce**

```C++
#include "cantera/core.h"
#include "cantera/base/SolutionArray.h"
#include <cassert>

int main() {
    using namespace Cantera;
    
    int loc = 0;
    shared_ptr<Solution> sol = newSolution("h2o2.yaml", "ohmech");
    shared_ptr<SolutionArray> array = SolutionArray::create(sol, 1);
    vector<double> initial_state(sol->thermo()->stateSize());
    
    sol->thermo()->setTemperature(100.0);
    sol->thermo()->saveState(initial_state);
    array->updateState(loc); //state @ 0 has temperature of 100.0 K
    
    vector<double> second_state(sol->thermo()->stateSize());
    sol->thermo()->setTemperature(234.0);
    sol->thermo()->saveState(second_state);

    vector<double> retrieved_state = array->getState(loc); //should still be the state @ 0

    assert(retrieved_state[0] == initial_state[0]);
    assert(retrieved_state[0] != second_state[0]);

    return 0;
}
```

**Behavior**

Expected: `getState(loc)` should always return the state stored at `loc`, regardless of subsequent modifications to the associated `Solution` object.

In this case, the first entry of the state (temperature) is equal to 234.0, but it should be equal to 100.0.

**System information**

- Cantera version: 3.2.0
- Python/MATLAB/other software versions: C++

**Additional context**

The problem seems to stem from the interaction of `getState` with `setLoc`. `getState` first uses `setLoc` to copy the state data to the associated `Solution` object (by setting `restore = true`), and then copies the state of the `Solution` object to the output vector. The problem is that `setLoc` has a check to see whether the "active" index `m_loc` has changed, and skips the copying step entirely if it hasn't. 

https://github.com/Cantera/cantera/blob/ef09fb43bf29a43b1dd8ff66780bef83bfb7c2ee/src/base/SolutionArray.cpp#L813-L822

The simplest fix, in my opinion, would be to change `getState` to directly copy the state data to the output vector:
```C++
vector<double> SolutionArray::getState(int loc)
{
    setLoc(loc, false); //update m_loc but don't restore Solution state
    size_t nState = m_sol->thermo()->stateSize();
    double* data_ptr = m_data->data() + m_loc * m_stride;
    vector<double> out(data_ptr, data_ptr + nState); //copy data directly to vector using iterator constructor
    return out;
}
```

I would also consider removing the `static_cast<size_t>(m_active[loc_]) == m_loc` optimization from `setLoc`. It doesn't meaningfully improve performance if `restore` is false and when `restore` is true, the implicit state coupling can lead to subtle edge cases like this one.

ischoegl:

- Fix `SolutionArray::getState` so it reliably returns the state stored at the requested location, regardless of the current state of the associated `Solution` object (previously, a stale value could be returned when the location matched the last-accessed one).
- Make `getState` a `const`, read-only accessor: it now resolves the data index directly and copies from the buffer instead of routing through `setLoc`, so it no longer updates the buffered location or the associated `Solution` state, i.e., a clear separation from the alternative `setState` method. The same rationale applies to a read-only `getAuxiliary`.
- Validate inputs with `ArraySizeError` (empty array) and `IndexError` (out-of-range or negative `loc`), and add a regression test (`SolutionArray.getStateReadOnly`).
- Fixed bug in `SolutionArray::setState` where an OOB index was not properly checked.
- Fixed bug in sliced `SolutionArray` constructor.

Closes #2067

**AI Statement (required)**

- **Limited use of generative AI.** Standard or boilerplate code snippets were generated with AI and manually reviewed; all design, logic, and implementation decisions were made by the contributor. AI used: Claude Opus 4.8.

**Checklist**

- [x] Style & formatting of contributed code follows [contributing
guidelines](https://github.com/Cantera/cantera/blob/main/CONTRIBUTING.md)

### DIFF
--- a/include/cantera/base/SolutionArray.h
+++ b/include/cantera/base/SolutionArray.h
@@ -197,8 +197,13 @@ public:
     //! Retrieve list of extra component names
     vector<string> listExtra(bool all=true) const;
 
-    //! Retrieve auxiliary data for a given location.
-    AnyMap getAuxiliary(int loc);
+    /**
+     *  Retrieve auxiliary data for a given location.
+     *  This is a read-only accessor that returns the auxiliary data stored at @c loc;
+     *  it does not modify the buffered location or the associated Solution object.
+     *  @since  Starting in %Cantera 4.0, this method is const/read-only.
+     */
+    AnyMap getAuxiliary(int loc) const;
 
     //! Set auxiliary data for a given location.
     void setAuxiliary(int loc, const AnyMap& data);
--- a/include/cantera/base/SolutionArray.h
+++ b/include/cantera/base/SolutionArray.h
@@ -164,8 +164,13 @@ public:
      */
     void updateState(int loc);
 
-    //! Retrieve the state vector for a given location.
-    vector<double> getState(int loc);
+    /**
+     *  Retrieve the state vector for a given location.
+     *  This is a read-only accessor that returns the state stored at @c loc; it does
+     *  not modify the buffered location or the associated Solution object.
+     *  @since  Starting in %Cantera 4.0, this method is const/read-only.
+     */
+    vector<double> getState(int loc) const;
 
     //! Set the state vector for a given location.
     void setState(int loc, const vector<double>& state);
--- a/include/cantera/base/SolutionArray.h
+++ b/include/cantera/base/SolutionArray.h
@@ -394,6 +394,9 @@ protected:
     //! Retrieve set containing list of properties defining state
     set<string> _stateProperties(const string& mode, bool alias=false);
 
+    //! `true` if this object is a view (slice) into another SolutionArray's data
+    bool _isSliced() const { return m_shared; }
+
     shared_ptr<Solution> m_sol; //!< Solution object associated with state data
     size_t m_size; //!< Number of entries in SolutionArray
     size_t m_dataSize; //!< Total size of unsliced data
--- a/src/base/SolutionArray.cpp
+++ b/src/base/SolutionArray.cpp
@@ -112,7 +112,7 @@ SolutionArray::SolutionArray(const SolutionArray& other,
     }
     for (auto loc : m_active) {
         if (loc < 0 || loc >= (int)m_dataSize) {
-            IndexError("SolutionArray::SolutionArray", "indices", loc, m_dataSize);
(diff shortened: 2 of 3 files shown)


include/cantera/base/SolutionArray.h
```diff
@@ -384,6 +394,9 @@ class SolutionArray
     //! Retrieve set containing list of properties defining state
     set<string> _stateProperties(const string& mode, bool alias=false);
 
+    //! `true` if this object is a view (slice) into another SolutionArray's data
+    bool isSliced() const { return m_shared; }
```
speth: The rest of `SolutionArray`'s protected members are prefixed with an underscore. Do you want to continue that pattern for local consistency?

src/base/SolutionArray.cpp
```diff
@@ -886,25 +888,34 @@ void SolutionArray::normalize() {
...
-    setLoc(loc);
+    // read-only access: resolve the data index directly without modifying m_loc or the
+    // associated Solution object
+    if (m_size == 0) {
+        throw ArraySizeError("SolutionArray::getAuxiliary", m_size, 1);
+    }
+    if (loc < 0 || static_cast<size_t>(loc) >= m_size) {
+        throw IndexError("SolutionArray::getAuxiliary", "indices",
+                         static_cast<size_t>(loc), m_size);
+    }
```
speth: Why is the first conditional here needed? If `m_size` is zero, you would also trip the second condition no matter what, right? And I think the `IndexError` seems more correct for trying to access a data point in an empty `SolutionArray`.

speth (changes requested):
Thanks for taking a look at this, @ischoegl. I had just a couple questions / comments.

src/base/SolutionArray.cpp
ischoegl: True - I fixed both instances.

speth (approved):
Thanks, @ischoegl! This looks good to me.

PR #2135, opened 2026-06-11
Cython pure Python

### LINKED ISSUE #241: Convert Cython code to Pure Python syntax
**Abstract**

Cython supports a Pure Python syntax using type annotations (see: [here](https://cython.readthedocs.io/en/latest/src/tutorial/pure.html#pep484-type-annotations) and [here](https://cython.readthedocs.io/en/latest/src/quickstart/cythonize.html)). If it is feasible to convert the Python interface from the current Cython syntax to this annotation-powered Pure Python syntax, it would bring a number of benefits.

**Motivation**

The primary motivation comes down to dev tooling. Modern Python dev tooling has been rapidly advancing, with linters like `ruff` and `black`, and numerous static type checkers including `mypy`, `pyright`,  `zuban`, `ty`, and `pyrefly`, with powerful IDE-integration through language server protocols (LSP) to provide convenient features like IntelliSense, inlay hints, and static type error catching as you write.

However, to my knowledge, none of these tools support Cython syntax. By transitioning to Pure Python syntax these tools could be better incorporated into the Cantera developer workflow, helping to objectively enforce code style and catch several classes of error related to static typing.

As a secondary benefit, the recently-added [type stubs](https://github.com/Cantera/cantera/pull/1926) could largely be merged into the Cython code, significantly reducing the maintenance burden and more strongly enforcing consistency between the runtime behavior and type annotations on the external API.

**Possible Solutions**

First, we should survey the Cython documentation and existing Cython code bases which primarily use the Pure Python syntax to see where it currently stands in terms of support for modern type annotations, to ensure it will be compatible with the established stubs and desired API. We can also observe whether it will support all of the desired tooling and features, and see what sort of CI pipelines we might want to implement. `NumPy` would be a good starting point.

Second, I plan to assemble a proof-of-concept by selecting one or two Cython `.pyx` files to convert directly to `.py` files. We can use this to determine whether the end result will be worth the effort, and see if there are any Cantera-specific quirks to work around. I expect that the conversion can be accomplished incrementally, so this will also prove if that is the case.

If everything looks good, we can then decide whether to target Cantera 3.3 or later, and how to structure the pull requests (e.g. Separate pull requests for each file / small groups of files? Stacking approach?).

I would also strongly push to include some new dev tooling standards/recommendations as part of this overhaul. At minimum, this would include adding linter configurations for `ruff` to `pyproject.toml`, which I have personally employed throughout the development of the type stubs, and potentially other standards which could be enforced e.g. via pre-commit and CI checks. This would all need to be documented in the contributing guide.

**References**

* [Type hint enhancement proposal](https://github.com/Cantera/enhancements/issues/85)
* [Type stubs pull request](https://github.com/Cantera/cantera/pull/1926)

speth:

- Convert the Cython module from the `.pyx` syntax to Cython's "pure Python" mode
- Absorb the type annotations in the `.pyi` files into the Cython-compiled `.py` files
- Break up the compiled Python extension so that each Cython `.py` becomes its own `.so` or `.pyd` extension module.

- Closes Cantera/enhancements#241

These changes provide the following benefits:

1. For users, IDE tool tips now include both typing information as well as method docstrings:

*Before*:
Screenshot 2026-07-07 at 11 48 59 AM

*After*:

Screenshot 2026-07-07 at 11 49 35 AM

2. Similarly, Ctrl+clicking on methods/properties will open the `.py` file at the location of the corresponding definition, rather than opening the `.pyi` file that does not show you either the docstring or the method implementation.

3. Sphinx documentation now contains the type annotation information (in some cases):

*Before* (with mouse hover over `array`)
Screenshot 2026-07-07 at 11 59 25 AM

*After*
Screenshot 2026-07-07 at 11 58 45 AM

4. For developers, the fact that Cython source files are now syntactically valid Python (though not actually directly importable) means that IDE tooling works better, and you get autocomplete / tooltip interaction at least for pure-Python modules:

*Before*
Screenshot 2026-07-07 at 11 47 17 AM

*After*
Screenshot 2026-07-07 at 11 45 58 AM

5. And finally, also for developers, there is no need to maintain the separate `.pyi` files with their typing information.

**Review Considerations**

I realize this is a massive PR, but it has essentially no effect on any actual runtime behavior. I would suggest that most useful evaluation is to try out the developer/user experience and confirm that there are no regressions. While there are certainly areas for improvement, I'd like to address them as much as possible in follow-up PRs rather than in this one, and to get this merged as soon as possible so we can minimize the need for resolving messy merge conflicts.

**Additional Caveats & Implementation Notes**

- Splitting the extension module into separate pieces is necessary because of the order in which Python resolves imports: extension modules come first, followed by `.py` files, followed by the "MetaPathFinder" magic. We need to package the `.py` files for IDE tools to have access to the type annotations and docstrings, but those would get read in preference to allowing the meta-path finder to provide the module from the bundled extension module, then resulting in errors because the special `cython.cimports` modules can't actually be used unless the module is compiled. By compiling each `.py` to a separate extension module (`.so` or `.pyd`) causes those to be found first by the Python interpreter, while leaving the `.py` files available for static analysis tooling.

  The one exception to this is Pyodide, where there are some limitations sharing data between Emscripten "side modules". As a result, we still use the metapath finder hooks there, and the `.py` files have to be excluded from the Pyodide package. We may be able to revisit this as the emscripten platform evolves.

- Bare annotations using parameterized generics such as `tuple[...]` and  `dict[str,str]` are coerced by Cython's annotation typing, rejecting for example a `list` where the declared type was a `tuple`, or rejecting a subclass of `dict` for a method where the declared type is `dict`, respectively. The workaround is to route through a TypeAlias such as `_BoundsPair: _TypeAlias = tuple[float, float]` which is not coerced and allows the runtime to accept any compatible type. One downside of this is that Sphinx docs and editor tool tips just show `_BoundsPair` rather than the more explicit type declaration.

- Pure-Python Cython syntax currently has no way of writing a C++ reference parameter. There is an open PR for this (https://github.com/cython/cython/pull/7121), but it will be a while before we can rely on this. As such, the `Delegator` interface, which involves a lot of Cython-defined functions with reference parameters, has been left in the `.pyx` format for now.

- Cython's pure-Python mode still allows the C++ "new" keyword, which is not valid Python syntax and causes type checkers to choke. Instead, we need to use the `make_shared` method, e.g. `self._rate.reset(new CxxArrheniusRate(py_to_anymap(input_data)))` becomes `self._rate = make_shared[CxxArrheniusRate](py_to_anymap(input_data))`

- Pointer casting is done as `cython.cast(cython.pointer(CxxArrheniusRate), self.rate)` instead of `<CxxArrheniusRate*>self.rate`. It's quite a bit more verbose, but it works.

- mypy/pyright are unaware of `__cinit__` being a constructor. Therefore, classes that do all there initialization in `__cinit__` need to have a placeholder `__init__` method duplicating the arguments of `__cinit__` where type annotations can be provided.

- Stubtest is chock full of false positives, because it now sees all of the Cython-only types and reports them as "not present at runtime". There doesn't appear to be a flag to suppress just this check, so the `.mypyignore` file is a bit of a mess.

- Likewise, while the Cython `.py` files are syntactically valid Python, they contain a lot of things that aren't linter friendly. Each file now has a fairly extensive ignore list for pyright to eliminate these categories of false positives. I thought this was better than tagging every single use of a Cython construct as `#noqa`:

  ```
  # pyright: reportMissingImports=false, reportAttributeAccessIssue=false
  # pyright: reportUndefinedVariable=false, reportUnboundVariable=false
  # pyright: reportInvalidTypeArguments=false, reportAssignmentType=false
  # pyright: reportIndexIssue=false, reportInvalidTypeForm=false
  ```

- Modules must have `import cython.cimports.numpy as cnp`, the equivalent of the pyx `cimport numpy as np` to provide access to the Numpy C-API and avoid confusing crashes at runtime.

- New Cython modules have to be listed explicitly in the SCons / CMake configurations, as they can't be distinguished from plain Python modules based on their extension. This requires updates in `interfaces/cython/pyproject.toml`, `interfaces/python_sdist/pyproject.toml.in`, `interfaces/cython/SConscript`, and `interfaces/python_sdist/cantera/CMakeLists.txt`

**AI Statement (required)**

  logic and implementation decisions. All generated code and documentation were reviewed and understood by the contributor. Implemented using Claude Code (Opus 4.8).

**Checklist**


TimothyEDawson:
This is extremely cool to see, I hope to find time soon to look through and maybe help if I can!

One small thing I wanted to point out is that if support for Python < 3.12 has been officially dropped, there are various things which were designed specifically to maintain 3.10 support and can now be dropped or improved with current syntax. Some examples which come to mind:

- TypeAlias [has been deprecated](https://docs.python.org/3/library/typing.html#typing.TypeAlias) in favor of the new `type` syntax.
- There's a new [inline syntax](https://docs.python.org/3/library/typing.html#typing.TypeVar) for generics, which is particularly nice for one-offs. I would still keep using `TypeVar` when there's a bounded/covariant/contravariant generic type which is used across multiple classes and functions.
- There should be no need to put types in quotes, e.g. `"SolutionArray[Solution] | _DataFrame | str | _Path | None"` in `FlameBase.set_initial_guess`. I feel like there may be exceptions which were not resolved until 3.14, but I can't locate them at the moment. I do know that it's recommended to use strings within calls to `cast`, as that avoids using the type within runtime code (at runtime `cast` simply returns the input object, but the type is still present as an input argument).

ischoegl (commented):
@speth - there are some merge conflicts. Could you rebase?

speth:
Yes, I will rebase, but the rebase is complicated and I'd like to do it exactly once. I was waiting to merge #2144, which I've just done.

speth:
> * TypeAlias [has been deprecated](https://docs.python.org/3/library/typing.html#typing.TypeAlias) in favor of the new `type` syntax.

Cython does not appear to support the `type` statement version of this yet. I couldn't seem to find an issue/PR, but that might just be due to it being difficult to search for.

> * There's a new [inline syntax](https://docs.python.org/3/library/typing.html#typing.TypeVar) for generics, which is particularly nice for one-offs.

Likewise, this is not currently supported by Cython's "pure Python" mode.

> * There should be no need to put types in quotes

This is mostly true. However, the quotes are useful/necessary in some places to prevent Cython from generating compiled functions that only accept an exact type when the function should take derived types, for example a function that takes `dict[str, Any]` should accept an `AnyMap` object, which is derived from `dict`. The code also uses `TypeAlias` in some places for the same reason.

ischoegl (approved):
Thanks, @speth for this massive undertaking (even if it the coding part is delegated to AI) -- this is a pure mechanical port, so I agree that we should go ahead and merge, and work out the kinks in second passes.

PR #2137, opened 2026-06-14
Address tpx convergence issue

### LINKED ISSUE #605: Setting T and P for PureFluid heptane fails at low temperatures

ischoegl:

- Fix `tpx::Substance::set_TPp` so that it converges on the nearly incompressible liquid branch at low temperature. There, the saturation pressure becomes so small that the solver's requested *relative* pressure tolerance (`ErrP * Pressure`) is finer than one floating-point ULP of pressure, so the iteration could never satisfy it and bailed out with a spurious `No convergence` error. The iteration now (a) detects a sign change of the pressure residual between successive iterates to bracket the root, and (b) terminates by accepting the root once the volume step drops below the floating-point resolution of the specific volume. `TP` setters now work down to each pure fluid's minimum temperature.
- Make `tpx::Substance::Tsat` reject pressures below the minimum pressure (evaluated as the saturation pressure at the minimum temperature) with the meaningful `Illegal pressure value` error, instead of failing later with the opaque `No convergence`. This removes a deferred `@todo`.
- Add a regression test (`test_set_TP_near_min_temp`) that sets a compressed-liquid state at the minimum temperature for every pure fluid, and re-enable the previously disabled below-triple-point `Tsat` test.

Closes #605

The example from the issue now succeeds instead of raising `CanteraError`:

```python
import cantera as ct

p = ct.Heptane()
p.TP = 204.4, ct.one_atm      # previously: "no convergence for P* = ..."
print(p.density)              # 777.37 kg/m^3

# also works all the way down to the minimum temperature
p.TP = p.min_temp, ct.one_atm
print(p.T, p.density)         # 182.56  803.57 kg/m^3
```

AI Statement (required)

- **Extensive use of generative AI.** The root-cause analysis and implementation were carried out with Claude Code (Opus 4.8). All scoping and design decisions were made by the contributor, and every generated change was reviewed, understood, and validated (numerical reproduction of the solver behavior plus the full pure-fluid and units test suites) by the contributor.

Checklist

- [x] Style & formatting of contributed code follows contributing guidelines

### DIFF
--- a/src/tpx/Sub.cpp
+++ b/src/tpx/Sub.cpp
@@ -11,6 +11,7 @@
 #include "cantera/base/stringUtils.h"
 #include "cantera/base/global.h"
 #include <cmath>
+#include <limits>
 
 using namespace Cantera;
 
@@ -300,17 +301,15 @@ double Substance::x()
 double Substance::Tsat(double p)
 {
     double Tsave = T;
-    if (p <= 0. || p > Pcrit()) {
-        // @todo: this check should also fail below the triple point pressure
-        // (calculated as Ps() for Tmin() as it is not available as a numerical
-        // parameter). A bug causing low-temperature TP setters to fail for
-        // Heptane (GitHub issue #605) prevents this at the moment.
-        // Without this check, a calculation attempt fails with the less
-        // meaningful error "No convergence" below.
-        throw CanteraError("Substance::Tsat",
-                           "Illegal pressure value: {}", p);
-    }
+    // The minimum pressure is not stored as a numerical parameter, so it is evaluated
+    // as the saturation pressure at the minimum temperature. Note that minimum
+    // temperature and pressure are often defined at the triple point.
+    T = Tmin();
+    double Pmin = Ps();
     T = Tsave;
+    if (p <= 0. || p > Pcrit() || p < Pmin) {
+        throw CanteraError("Substance::Tsat", "Illegal pressure value: {}", p);
+    }
     Tsave = T;
 
     int LoopCount = 0;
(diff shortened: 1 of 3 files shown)


src/tpx/Sub.cpp
```diff
@@ -801,9 +800,44 @@ void Substance::set_TPp(double Temp, double Pressure)
...
+        double res = P_here - Pressure;
+        if (have_prev && kbr == 0 && res*res_prev < 0.0) {
+            // The residual changed sign between the previous and current
+            // specific volume, so the root lies between them. On the steep
+            // liquid branch at low temperature the iteration can step across
+            // the (very thin) target band without the logic below ever
+            // recording a bracket, leaving the solver unable to converge
+            // (GitHub issue #605). Capture the bracket explicitly so the
+            // in-bracket refinement below can proceed. Pressure decreases with
+            // specific volume here, so the smaller volume is the lower bound.
```
speth: I've noticed Claude likes to reference GitHub Issues in its comments, but I think we should keep references to now-solved issues out of the code; a future reader should only be presented with an explanation of what the code does now, not how it's a fix for some bug that used to exist. Likewise for the other comment block. Keeping the reference in the new regression test is reasonable, though.

speth (changes requested):
Thanks, @ischoegl. Just one minor suggestion.

ischoegl:
> Thanks, @ischoegl. Just one minor suggestion.

Sure! Taken care of.

speth (approved):

PR #2139, opened 2026-06-16
Semi-analytic Jacobian for 1D flames

speth:

The 1D flame solver currently computes a finite difference Jacobian for use with the Newton's method solver. The main efficiency this exploits is the block tridiagonal pattern of the Jacobian, where the state at each grid point only affects the residuals at that grid point and the immediate neighbors, so everything scales linearly with the number of grid points $N$. However, the number of function evaluations to compute each block row is $O(K)$ and the cost is also $O(K)$, making this Jacobian evaluation an $O(NK^2)$ cost. Using a banded matrix solver, where the bandwidth is proportional to $K$, the factorization cost is $O(NK^3)$ and the backsubstitution/solve cost is $O(NK^2)$. It turns out, though, that the constant factors on these different operations are quite different, and for a range of mechanism sizes, the Jacobian construction cost dominates.

This cost can be dramatically reduced by computing a semi-analytical Jacobian, where the derivatives with respect to species mass fractions are computed analytically and derivatives with respect to the remaining state variables are computed using the existing finite difference approach. Derivatives of the reaction rate expressions make use of the existing `Kinetics` derivatives, while derivatives of the transport-related terms are new additions to the `Flow1D` class.

Currently, the analytic Jacobian works for ideal gases using mixture-averaged transport, including the mixture-averaged approximation for the Soret diffusion term. It is enabled by default, but automatically falls back to the finite difference form for other flow domains (specifically, `IonFlow`), transport models, or equations of state. It can also be manually switched via the `Domain1D.jacobian_mode` property.

This PR also revises the performance statistics available from the 1D solver, introducing an `AnyMap`-based store accessible via `Sim1D.solver_stats` that provides break downs of run time and call count for function evaluations, Jacobian evaluations, Jacobian factorizations, and linear solves for each grid adaptation step.

The following is a performance comparison on a laminar flame speed calculation, using mechanisms of various sizes, including a couple from [NUI Galway](https://www.universityofgalway.ie/combustionchemistrycentre/mechanismdownloads/) converted to YAML format.
```py
import cantera as ct
import pandas as pd
ct.suppress_thermo_warnings()

def run(mech, reactants, jacobian_mode):
      p = ct.one_atm  # pressure [Pa]
      Tin = 300.0  # unburned gas temperature [K]
      width = 0.03  # m
      loglevel = 0  # amount of diagnostic output (0 to 8)

      gas = ct.Solution(mech)
      gas.TPX = Tin, p, reactants

      f = ct.FreeFlame(gas, width=width)
      f.flame.jacobian_mode = jacobian_mode
      f.set_refine_criteria(ratio=3, slope=0.06, curve=0.12)
      f.transport_model = 'mixture-averaged'
      f.flux_gradient_basis = "mass"
      f.solve(loglevel=loglevel, auto=True)
      print(f"mixture-averaged flamespeed = {f.velocity[0]:7f} m/s")

      f.soret_enabled = True
      f.solve(loglevel=loglevel) # don't use 'auto' on subsequent solves
      print("mixture-averaged flamespeed with Soret diffusion = {f.velocity[0]:7f} m/s")

      df = pd.DataFrame.from_dict(f.solver_stats)
      print(df)
      print(df[['residual_time', 'jacobian_time', 'factor_time', 'solve_time', 'total_time']].sum())


CASES = [
    ('h2o2.yaml', 'H2:1.1, O2:1, AR:5'),  # 10 species, 29 reactions
    ('gri30.yaml', 'CH4:0.45, O2:1, N2:3.76'),  # 53 species, 325 reactions
    ('methanol-galway2016.yaml', 'CH3OH:0.666, O2:1, N2:3.76'),  # 173 species, 1011 reactions
    ('ethanol-DME-galway2018.yaml', 'C2H5OH:0.3666, O2:1, N2:3.76')  # 433 species, 1004 reactions
]

for mech, reactants in CASES:
      for jacobian_mode in ['finite-difference', 'analytic']:
            print("-" * 80)
            print(f"{mech=}")
            run(mech, reactants, jacobian_mode)
            print(f"{jacobian_mode=}")
```

Collating the output shows the resulting performance benefits (times in seconds; using OpenBLAS with `OMP_NUM_THREADS=1`).

| n_species | mode     | f eval  | J eval  | J factor | lin solve | total   | speedup |
| --------: | -------- | ------: | ------: | -------: | --------: | ------: | ------- |
| 10        | FD       | 0.07    | 0.19    | 0.02     | 0.03      | 0.32    | -       |
| 10        | analytic | 0.07    | 0.09    | 0.02     | 0.03      | 0.22    | **1.44**|
| 53        | FD       | 1.21    | 9.62    | 0.55     | 0.34      | 11.76   | -       |
| 53        | analytic | 1.17    | 1.89    | 0.54     | 0.32      | 3.95    | **2.97**|
| 173       | FD       | 20.79   | 218.42  | 18.05    | 12.27     | 270.30  | -       |
| 173       | analytic | 18.60   | 21.95   | 16.32    | 10.91     | 68.48   | **3.95**|
| 433       | FD       | 92.42   | 632.96  | 200.89   | 62.40     | 992.94  | -       |
| 433       | analytic | 96.15   | 69.33   | 196.78   | 65.21     | 431.70  | **2.30**|

The benefit for very small mechanisms is limited, but grows for medium-sized mechanisms. Eventually, for larger mechanisms, even though the Jacobian evaluation cost is reduced by a factor of ~9, the factorization step becomes the dominant contributor to the overall runtime and the relative speedup starts to decline.

**AI Statement (required)**

  logic and implementation decisions. All generated code and documentation were reviewed and understood by the contributor. Implemented using Claude Code (Opus 4.8) and [superpowers](https://github.com/obra/superpowers).

**Checklist**


doc/sphinx/reference/onedim/nonlinear-solver.md
```diff
@@ -74,9 +74,69 @@ J_{i,j} = \frac{\partial F_i(x)}{\partial x_j}
...
+  forces a full update of all properties.
+- The domain has fewer than 3 grid points.
+
+In ``'auto'`` mode every one of these is handled silently. In ``'analytic'``
+mode, the first three -- genuine configuration choices by the user -- raise a
+``CanteraError``; the last two, which are transient or internal solver states,
+still fall back silently so that, for example, adjoint sensitivity calculations
+keep working with an explicit analytic request.
+
+The fallback is transparent: the solver produces the same results regardless of whether
```
ischoegl, line 137: Per https://github.com/Cantera/cantera/pull/2139/commits/20dc9abc5740004622136dd84727f023c7f90e5f ... is there a caveat that it doesn't always work?

ischoegl (commented):
@speth - thank for your work on this. Too bad that the factorization cost starts to cut into the savings - did you investigate whether there is a crossover at large number of species? Fwiw, I'm pleased that my work on analytic Jacobians from some time ago continues to spread across the code base.

speth:
> Too bad that the factorization cost starts to cut into the savings - did you investigate whether there is a crossover at large number of species?

To be clear, the factorization costs were there before, too, and haven't changed. The absolute time saved by this change grows with the number of species. It's only in relative/fractional terms that this starts to matter less for large mechanisms. You can see the crossover for the updated solver in the table in the PR description -- it's around 200 species. For the baseline solver, it will be further out, maybe around the 800 species mark, though I didn't test any mechanisms that large. All this means is that if you want to improve performance for very large mechanisms, you have to change how the linear systems are solved, presumably using sparse, iterative approaches like we do for `ReactorNet`.

doc/sphinx/reference/onedim/nonlinear-solver.md
speth: The "transparent fallback" behavior is about the cases where the analytic Jacobian would be incorrect, e.g. non-ideal gas where there are additional chain rule terms.

The very high pressure, pure H2/O2 diffusion flames used in that example are the only cases I've seen where the analytic Jacobian is for some reason a worse choice than the finite difference one. I don't really understand why that's the case, but I think those conditions are pretty extreme / unique. The option to manually select the finite-difference Jacobian is still there in that case.

I guess the one question is whether the `auto` option should fall back to finite difference at some point in its sequence, too.

ischoegl: Sure - what I meant here is that you wrote that "the solver produces the same results regardless of whether the analytic or finite-difference path is taken; only the cost of Jacobian construction differs".

If this were true, then the very high pressure cases should just work. It is a nit though - rest looks good to me!

ischoegl (approved):

PR #2141, opened 2026-06-21
Make MixtureFugacityTP::cubicSolver more robust

### LINKED ISSUE #1157: MixtureFugacityTP::solveCubic finds wrong root on mingw
**Problem description**

Within the test suite, `TEST_F(cubicSolver_Test, solve_cubic)` fails on `mingw`, as discussed on #1144. For `mingw`, the check for 
```
EXPECT_NEAR(nSolnValues, -2, 1.e-6);
```
fails, where +2 is returned instead. At the moment, the test is disabled (ensuring that the newly introduced `mingw` GH runner succeeds).

**Additional context**

> The `PengRobinson` and `RedlichKwong` models both use this value and do not treat the positive and negative return values equally. 

Issue goes back to #1047

An associated test is disabled for mingw:
https://github.com/Cantera/cantera/blob/e5122f06fbc72cccc7c7e32355a956353cdf575d/test/thermo/cubicSolver_Test.cpp#L26-L34

ischoegl:

- Make cubic-EOS triple-root detection independent of compiler-specific floating-point rounding.
- Use `cbrt` for signed cube roots and correct validation of the negative-root branch.
- Re-enable the cubic solver test on MinGW and verify all roots coincide at the critical point.

Closes #1157

**Extensive use of generative AI.**
Significant portions of code or documentation were generated with AI, including logic and implementation decisions. All generated code and documentation were reviewed and understood by the contributor. Specifically, OpenAI Codex (gpt-5.5) was used to diagnose and fix the issue.

**Checklist**

### DIFF
--- a/src/thermo/MixtureFugacityTP.cpp
+++ b/src/thermo/MixtureFugacityTP.cpp
@@ -14,6 +14,7 @@
 #include "cantera/base/global.h"
 
 #include <algorithm>
+#include <limits>
 
 using namespace std;
 
@@ -785,7 +786,8 @@ int MixtureFugacityTP::solveCubic(double T, double pres, double a, double b,
     double xN = - bn /(3 * an);
 
     // Derive the value of delta**2. This is a key quantity that determines the number of turning points
-    double delta2 = (bn * bn - 3 * an * cn) / (9 * an * an);
+    double deltaNumerator = bn * bn - 3 * an * cn;
+    double delta2 = deltaNumerator / (9 * an * an);
     double delta = 0.0;
 
     // Calculate a couple of ratios
@@ -818,11 +820,29 @@ int MixtureFugacityTP::solveCubic(double T, double pres, double a, double b,
     }
 
     double h = 2.0 * an * delta * delta2;
-    double yN = 2.0 * bn * bn * bn / (27.0 * an * an) - bn * cn / (3.0 * an) + dn; // y_N term
+    double yTerm1 = 2.0 * bn * bn * bn / (27.0 * an * an);
+    double yTerm2 = -bn * cn / (3.0 * an);
+    double yN = yTerm1 + yTerm2 + dn; // y_N term
     double disc = yN * yN - h2; // discriminant
 
+    // At a triple root, both terms of the depressed cubic are zero. Detect
+    // cancellation relative to the terms used to calculate them so that the
+    // result does not depend on compiler-specific floating-point evaluation.
+    // The factor of 64 allows for roundoff accumulated across the arithmetic
+    // operations; the scales below convert this relative tolerance to an
+    // absolute tolerance for each cancellation residual.
+    double cancellationTol = 64 * std::numeric_limits<double>::epsilon();
+    double deltaScale = max(fabs(bn * bn), fabs(3 * an * cn));
+    double yScale = max({fabs(yTerm1), fabs(yTerm2), fabs(dn)});
+    bool tripleRoot = (fabs(deltaNumerator) <= cancellationTol * deltaScale
+                       && fabs(yN) <= cancellationTol * yScale);
+    if (tripleRoot) {
+        delta = 0.0;
+        disc = 0.0;
+    }
+
     //check if y = h
-    if (fabs(fabs(h) - fabs(yN)) < 1.0E-10) {
+    if (!tripleRoot && fabs(fabs(h) - fabs(yN)) < 1.0E-10) {
         if (disc > 1e-10) {
             throw CanteraError("MixtureFugacityTP::solveCubic",
                 "value of yN and h are too high, unrealistic roots may be obtained");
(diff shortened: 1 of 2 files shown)


src/thermo/MixtureFugacityTP.cpp
```diff
@@ -818,11 +820,26 @@ int MixtureFugacityTP::solveCubic(double T, double pres, double a, double b,
...
+    double yN = yTerm1 + yTerm2 + dn; // y_N term
     double disc = yN * yN - h2; // discriminant
 
+    // At a triple root, both terms of the depressed cubic are zero. Detect
+    // cancellation relative to the terms used to calculate them so that the
+    // result does not depend on compiler-specific floating-point evaluation.
+    double eps = std::numeric_limits<double>::epsilon();
+    double deltaScale = max(fabs(bn * bn), fabs(3 * an * cn));
+    double yScale = max({fabs(yTerm1), fabs(yTerm2), fabs(dn)});
+    bool tripleRoot = (fabs(deltaNumerator) <= 64 * eps * deltaScale
```
bryanwweber: Can you add a comment about what the 64 is here for?

src/thermo/MixtureFugacityTP.cpp
```diff
@@ -895,15 +912,15 @@ int MixtureFugacityTP::solveCubic(double T, double pres, double a, double b,
...
+        if (tripleRoot) {
             // yN = 0.0 and h = 0 (that is, disc = 0)
             Vroot[0] = xN;
             Vroot[1] = xN;
             Vroot[2] = xN;
         } else {
             // h and yN need to figure out whether delta^3 is positive or negative
             if (yN > 0.0) {
-                tmp = pow(yN/(2*an), 1./3.);
+                tmp = cbrt(yN / (2 * an));
```
bryanwweber, line 929: Does cbrt also depend on the compiler implementation?

src/thermo/MixtureFugacityTP.cpp
ischoegl: It's an arbitrary number that is sufficiently close to machine precision. Will add a comment.

src/thermo/MixtureFugacityTP.cpp
ischoegl: While `cbrt` is part of the C++ standard library, it indeed may not give bitwise-identical results across platforms. It is, however, preferable here as it avoids the `NaN` output that is possible for `pow(..., 1./3.)`. Fwiw, the same platform dependence holds for `pow`.

The compiler-specific rounding comment at the top was with respect to #1157, which was only observed for MinGW.

bryanwweber: So the problem being addressed is that `1./3.` rounds to a different value with MinGW than with other compilers? And the NaN issue is that `yN/(2*an)` might be negative, which cbrt resolves?

ischoegl: This line fixes a secondary bug. The primary issue was that the triple root was not detected correctly. The new thresholding fixes this.

speth:
I'm not quite sure what behavior we're trying to ensure with the test in question. There are bound to be threshold conditions in any calculation that maps a floating point values to a discrete set. The most that you can expect is that the threshold value where it flips between different outputs is close (i.e. small multiple of machine precision) for when using different compilers / systems.

The test was (apparently) getting past the relative comparison `EXPECT_NEAR(expected_result[2], Vroot[0], 1.e-6);` and only "failing" on whether decided the two solutions were on the liquid branch or the gas branch. Of course, at exactly the critical point (or as close as one can get with floating point), the failing check `EXPECT_NEAR(nSolnValues, -2, 1.e-6)` is arguably not correct in any case -- there is only one *distinct* root in this case, so I'd expect that this should either return 1 or 3 depending on whether the expectation is to count distinct or total real roots, though that perhaps also depends on how this return value is meant to be used.

Creating a radius around the critical point that effectively snaps to this triple root just moves the threshold condition to some set of points near but not quite at the critical point.

I do agree that the change to use `cbrt` over `pow` makes sense, not only for its handling of the case where the operand is negative but also because it is generally more accurate and faster.

ischoegl:
Thanks, @speth!

> [...] Of course, at exactly the critical point (or as close as one can get with floating point), the failing check `EXPECT_NEAR(nSolnValues, -2, 1.e-6)` is arguably not correct in any case -- there is only one _distinct_ root in this case, so I'd expect that this should either return 1 or 3 depending on whether the expectation is to count distinct or total real roots, though that perhaps also depends on how this return value is meant to be used.

I agree with this interpretation. The MinGW failure exposed that the old test was asserting a liquid/gas branch sign at the critical point, where that distinction is degenerate. I’ve updated the fix so the detected triple-root case returns `1` rather than `2` or `-2`. At `Tc, Pc`, the cubic has one distinct real root with multiplicity three, so `1` seems like the clearest convention for the single usable thermodynamic state. I also changed the test to assert that convention and to check the critical molar volume directly, rather than checking a branch sign.

> I do agree that the change to use `cbrt` over `pow` makes sense, not only for its handling of the case where the operand is negative but also because it is generally more accurate and faster.

I kept the `cbrt` change since it fixes the negative-operand path and is the more appropriate operation here.

speth (approved):

PR #2142, opened 2026-06-21
Fix Fortran/testing issues

### LINKED ISSUE #1379: f90_demo crashes on macOS (Apple Silicon)
**Problem description**

The macOS F90 sample crashes on macOS (arm)

**Steps to reproduce**

Build cantera from source:

```
% scons build
% scons install
% cd path/to/samples/f90
% make
```

*Edit: or simply run `scons test-f90-demo`*

**Behavior**

Running the demo crashes:
```
% ./demo

 ********   Fortran 90 Test Program   ********
 Initial state properties:


Temperature:       1200.0     K
Pressure:         0.10132E+06 Pa
Density:          0.28921     kg/m3
Molar Enthalpy:   0.23514E+08 J/kmol
Molar Entropy:    0.20594E+06 J/kmol-K
Molar cp:          27056.     J/kmol-K


libc++abi: terminating with uncaught exception of type Cantera::CanteraError:
CanteraError thrown by ThermoPhase::equilibrate:
Invalid solver specified: ''


Program received signal SIGABRT: Process abort signal.

Backtrace for this error:
#0  0x101c576f7
#1  0x101c5671b
#2  0x1c49434a3
#3  0x1c492bedf
#4  0x1c486633f
#5  0x1c48e6b07
#6  0x1c48d6937
#7  0x1c47cc32f
#8  0x1c48e5ea3
#9  0x1c48e8c1b
#10  0x1c48e8bc7
zsh: abort      ./demo
```

**System information**

- Cantera version: 3.0.0a2
- OS: macOS (on M2)
- Python/MATLAB/other software versions: installation in `conda` environment, including `gfortran`

```
% cat cantera.conf
blas_lapack_libs = 'openblas'
```

Further:
```
(base) % conda activate cantera-dev
(cantera-dev) % g++ --version
Apple clang version 13.1.6 (clang-1316.0.21.2.5)
Target: arm64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
(cantera-dev) % gfortran --version
GNU Fortran (Homebrew GCC 12.2.0) 12.2.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

(cantera-dev) % python --version
Python 3.10.5
```

**Additional context**

The same error occurs for SCons and `make` routes (after applying fix in #1378). The example works on Linux and Intel macOS.

*On a separate note, F90 CMake throws an error when linking on Intel macOS (which is not an issue on Apple Silicon).*

ischoegl:

- Fix Fortran string-length handling for modern `gfortran`, resolving the F90 demo crash on Apple Silicon.
- Enable the Fortran interface and regression tests on selected macOS CI runners.
- Update the derivative_speed sample for the span<double> API and ensure unregistered samples are compiled in CI.

Fixes #1379

**Limited use of generative AI.**
Standard or boilerplate code snippets were generated with AI and manually reviewed; all design, logic, and implementation decisions were made by the contributor. Problems were diagnosed and fixed using OpenAI Codex (gpt-5.5), with CI being updated manually.

Checklist


samples/cxx/jacobian/derivative_speed.cpp
```diff
@@ -54,7 +55,7 @@ void timeit_array(void (Kinetics::*function)(double*),
         auto t1 = std::chrono::high_resolution_clock::now();
         for (size_t i = 0.; i < loops; ++i) {
             gas.setState_TP(T + i * deltaT, pressure);
-            (kin->*function)(out.data());
+            (kin->*function)(outSpan);
```
speth: I think you can just pass `out` here -- no need to explicitly construct `outSpan`.

src/fortran/fct.cpp
```diff
@@ -32,6 +32,8 @@ typedef Cabinet<Kinetics> KineticsCabinet;
 typedef Cabinet<Transport> TransportCabinet;
 
 typedef integer status_t;
+// GCC 8 and newer pass hidden Fortran character lengths as size_t.
+typedef size_t fortran_charlen_t;
```
speth: Rather than isolating the replacement to this file, I think this update should be applied to the existing `ftnlen` typedef in `config.h.in`. I have to imagine that we are just getting "lucky" with the other places where this comes up. Or if not, we need to distinguish why these other cases are different.

test_problems/SConscript
```diff
@@ -224,6 +224,9 @@ Test('cxx-custom', 'cxx_samples', '#build/samples/cxx/custom/custom', None,
 Test('cxx-demo', 'cxx_samples', '#build/samples/cxx/demo/demo',
      'cxx_demo_blessed.txt',
      threshold=1e-10, tolerance=2e-4, slow=True)
+# # Test is significantly slower than others
+# Test('cxx-derivative-speed', 'cxx_samples',
+#      '#build/samples/cxx/jacobian/derivative_speed', None, slow=True)
```
speth: Adding this commented out is a little weird. I suppose a better approach to making sure this sample gets tested would be to add an optional argument to either run only for `h2o2.yaml` or to cut the repeat size down by some significant factor.

speth (changes requested):
Thanks for investigating this, @ischoegl. I had a couple of suggestions.

src/fortran/fct.cpp
ischoegl: Unfortunately, introducing a separate `typedef` is necessary, as `ftnlen` is used elsewhere also (BLAS), i.e., there would be side effects.

test_problems/SConscript
ischoegl: I was torn here, but added this in the interest of transparency. I already added a check in CI so this is covered either way; I can drop the commented out section. *PS: I think I'm coming around to the idea of limiting it to `h2o2.yaml` and including it as a regular sample.*

src/fortran/fct.cpp
speth: Yes, exactly. What I'm saying is that I don't see how the BLAS calls aren't subject to the same issue (or did you try this and it caused errors?).

ischoegl: I looked into changing the existing `ftnlen` typedef globally, and it _does_ compile (different from my recollection). I still think that this would mix two related but distinct ABI boundaries. `ftnlen` is also used by `ctlapack.h` for BLAS/LAPACK character arguments, where the ABI is determined by the external LAPACK/BLAS provider. That may not match the Fortran compiler used for Cantera’s own Fortran interface.

I agree that keeping the new type local to `fct.cpp` was not ideal. I moved the Cantera Fortran interface length type into `config.h.in` instead, but kept it separate from `ftnlen`:

- `ftnlen` remains the LAPACK/f2c hidden `CHARACTER` length type.
- `ftnlen_t` is used for Cantera’s Fortran interface, where current GCC passes hidden `CHARACTER` lengths as `size_t`.

I also expanded the comments in `config.h.in` to document why these are separate typedefs.

test_problems/SConscript
ischoegl: Implemented a `--test` flag, which now only uses `h2o2` for CI.

src/fortran/fct.cpp
speth: I did some digging, and I think the answer is that BLAS/LAPACK implementations don't actually include these arguments _at all_. In the LAPACK API, the string arguments are all just single characters anyway, so they don't need to check what that length argument would hold and whatever we're putting there is just being ignored. You can see this in the prototypes for `dgetrs` in Apple's [Accelerate](https://developer.apple.com/documentation/accelerate/dgetrs_(_:_:_:_:_:_:_:_:_:)), [`MKL`](https://github.com/divakarvi/bk-spca/blob/master/linking/linalg/mkl_lapack.h), and the reference [Netlib](https://www.netlib.org/lapack/lapack.h) versions.

I think the new version needs to be used in `demo_ftnlib.cpp` as well.

speth (approved):

